PR #645, opened 2024-04-23
Bump Beyla to version 1.5.1

marctc:
#### PR Description

Bumps Beyla to version 1.5.1, which includes a bunch of bugfixes.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ Main (unreleased)
 - Update `alloy-mixin` to use more specific alert group names (for example,
   `alloy_clustering` instead of `clustering`) to avoid collision with installs
   of `agent-flow-mixin`. (@rfratto)
+- Bump Beyla to v1.5.1. (@marctc)
```
ptodev, line 42: Ideally, it'd be good to mention how this affects Agent users. Would that be very difficult to do? It seems there are lots of changes in the latest Beyla version, so maybe it's not very easy.

ptodev (approved):

CHANGELOG.md
ptodev: If it doesn't affect users in a notable way, then maybe we don't need to mention this upgrade in the changelog.

marctc: Would linking to Beyla's changelog be enough? There aren't changes to Beyla component itself, just internal changes.

ptodev: Does every change in Beyla's changelog impact Alloy users? It would be most helpful if you just describe in one sentence what the sorts of improvements are noteworthy from an Alloy user point of view. I'm still not sure which version we're upgrading from, so not sure which changelog to link to also.

marctc: > Does every change in Beyla's changelog impact Alloy users? 

No, only the users that use Beyla's componet in Alloy.

PR #646, opened 2024-04-23
Fix nested import git data path

### LINKED ISSUE #643: Nested import.git blocks with same name label conflict
### What's wrong?

import.git downloads a git repository and stores it under the configured data path + the ID of the node.

For example

```
import.git "aa" {
  repository = "https://github.com/wildum/module.git"
  revision   = "test-modules"
  path       = "math"
}
```

will store the repo under dataPath/import.git.aa

The problem is that if the imported module has an import.git block also labeled "aa", it will match the path "dataPath/import.git.aa" and use the local repo instead of downloading the repo. This works if the repo is the same but if the repo is different it will break.

### Steps to reproduce

Create a config with an import.git that imports a module that has an import.git with the same label.

### System information

_No response_

### Software version

Grafana Alloy v1.0.0

### Configuration

```text
// root config
import.git "aa" {
  repository = "https://github.com/wildum/module.git"
  revision   = "test-modules"
  path       = "math"
}

aa.add "default" {
    value1 = 2
    value2 = 3
}

// module config
declare "add" {
    argument "value1" {}
    argument "value2" {}
    aa.kubernetes "default" {}
    export "output" {
        value = argument.value1.value + argument.value2.value
    }
}

import.git "aa" {
    repository = "https://github.com/grafana/alloy-modules.git"
    path       = "modules/ui/grafana"
}
```

### Logs

_No response_

wildum:

Fixes #643 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


tpaschalis (approved):
Makes sense, thanks for the quick fix!

PR #640, opened 2024-04-22
ci: enable CD for commits to main

rfratto:
Re-enables continuous delivery for main commits for internal testing.

PR #649, opened 2024-04-23
tools: fix incompatibilities between normal and alpine sh

rfratto:
`sh` (as executed by Alpine) understands `set -o pipefail`, but other installations of sh do not.

This commit partially reverts the change from #640 so that the scripts expect bash, and that our usage of alpine installs bash before executing the tools/image-tag-docker script.


mattdurham (approved):
Let’s go

PR #655, opened 2024-04-23
Add details for setting memory limit automatically.

### LINKED ISSUE #651: Add automatically setting go memory limit if cgroup is available
### Request

Automatically set the gomemlimit if cgroup is set

### Use case

Allow memory to be limited without manually setting gomemlimit

mattdurham:
#### PR Description

This will set the automemory limit if cgroups exist and the setting is enabled, the setting defaults to true. If there is a GOMEMLIMIT set already it will not overwrite it. If there are no cgroups then no limit will be applied.

Fixes #651

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


internal/alloycli/cmd_run.go
```diff
@@ -161,9 +163,16 @@ type alloyRun struct {
...
 	configBypassConversionErrors bool
 	configExtraArgs              string
+	enableAutoMemoryLimit        bool
 }
 
 func (fr *alloyRun) Run(configPath string) error {
+	// Set the memory limit, this will honor GOMEMLIMIT if set
+	// If there is a cgroup will follow that
+	if fr.enableAutoMemoryLimit {
+		memlimit.SetGoMemLimitWithOpts()
```
spartan0x117, line 173: Is it possible to pass in a compatible logger here? If so, some of the logs might be helpful (e.g. if `GOMEMLIMIT` is already set it won't honor the cgroups settings, as well as confirmation of the limit)

docs/sources/reference/cli/run.md
```diff
@@ -55,6 +55,7 @@ The following flags are supported:
 * `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
 * `--config.extra-args`: Extra arguments from the original format used by the converter.
 * `--stability.level`: The minimum permitted stability level of functionality to run. Supported values: `experimental`, `public-preview`, `generally-available` (default `"generally-available"`).
+* `--memory.auto-limit`: Enables the automatic setting of memory limit if using cgroups (default `true`).
```
spartan0x117, line 58: (nit, feel free to ignore :smile:)
 ```suggestion
* `--memory.auto-limit`: Enables automatic memory limit configuration when using cgroups (default `true`).
```

File: internal/alloycli/cmd_run.go [hunk 3]
```diff
@@ -138,6 +139,7 @@ depending on the nature of the reload error.
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
 	cmd.Flags().StringVar(&r.storagePath, "storage.path", r.storagePath, "Base directory where components can store data")
 	cmd.Flags().Var(&r.minStability, "stability.level", fmt.Sprintf("Minimum stability level of features to enable. Supported values: %s", strings.Join(featuregate.AllowedValues(), ", ")))
+	cmd.Flags().BoolVar(&r.enableAutoMemoryLimit, "memory.auto-limit", true, "Enable the automatic setting of memory limit.")
```
spartan0x117, line 142: (nit from above, feel free to ignore)
```suggestion
	cmd.Flags().BoolVar(&r.enableAutoMemoryLimit, "memory.auto-limit", true, "Enables automatic memory limit configuration when using cgroups.")
```

spartan0x117 (approved):
LGTM!

internal/alloycli/cmd_run.go
spartan0x117: Was thinking something like from the example:
```golang
memlimit.SetGoMemLimitWithOpts(
		memlimit.WithRatio(0.9),
		memlimit.WithProvider(memlimit.FromCgroup),
		memlimit.WithLogger(slog.Default()),          // Can we pass a logger here?
	)
```

mattdurham: I can but I have to move it til after our logging is fully initialized, which likely isn't the end of the world its just later in the process to pass down the bool  flag.

mattdurham: Added the slog.Default, its...a bit different looking but likely works for ease of use.

`2024/04/23 17:01:27 INFO memory is not limited, skipping: %v package=github.com/KimMachineGun/automemlimit/memlimit !BADKEY="memory is not limited"`

internal/alloycli/cmd_run.go
```diff
@@ -192,6 +197,12 @@ func (fr *alloyRun) Run(configPath string) error {
 
 	level.Info(l).Log("boringcrypto enabled", boringcrypto.Enabled)
 
+	// Set the memory limit, this will honor GOMEMLIMIT if set
+	// If there is a cgroup will follow that
+	if fr.enableAutoMemoryLimit {
+		memlimit.SetGoMemLimitWithOpts(memlimit.WithLogger(slog.Default()))
+	}
+
```
rfratto, line 205: Rather than exposing this as a new configuration option, what do you think about having this automatically be enabled for users using experimental/public-preview, and then once we get enough information on it, we can decide whether it should always be on or needs to be configurable?

mattdurham: IE if stability is experimental automatically enable? I can also set the flag as experimental, though unsure if we have a precedent for that with alloy.

rfratto: > IE if stability is experimental automatically enable? 

Yeah, like we would say that "auto memory limit is an experimental performance improvement" and only applies when `--stability.level=experimental. This is something I expect we'll be doing at some frequency, including for the eventual controller refactor.

> I can also set the flag as experimental

We don't have an existing precedent but I expect that most things will support feature gating at some point, including flags. 

However in this case I don't think marking the flag as experimental fully addresses my concerns, since presumably we intend for this to GA, and I prefer not adding too many more GA flags because of the increased API surface.

docs/sources/reference/cli/run.md
```diff
@@ -164,6 +164,7 @@ original configuration.
 Include `--config.extra-args` to pass additional command line flags from the original format to the converter.
 Refer to [alloy convert][] for more details on how `extra-args` work.
 
+
```
rfratto, line 167: IMO I think we should find a way to document this somewhere just so people know which improvements lower stability levels include. I don't know what page it should be on, but I'll leave that up to you since I'm not too opinionated.

internal/alloycli/cmd_run.go
mattdurham: Marked as public preview and added a helper function.

internal/alloycli/cmd_run.go
```diff
@@ -192,6 +195,12 @@ func (fr *alloyRun) Run(configPath string) error {
 
 	level.Info(l).Log("boringcrypto enabled", boringcrypto.Enabled)
 
+	// Set the memory limit, this will honor GOMEMLIMIT if set
+	// If there is a cgroup will follow that
+	if fr.minStability.AtLeast(featuregate.StabilityPublicPreview) {
+		memlimit.SetGoMemLimitWithOpts(memlimit.WithLogger(slog.Default()))
```
rfratto, line 201: I think this should use `slog.New(l.Handler())` here instead of `slog.Default()`; the controller's logger implements the slog.Handler interface. I think this is important so that log entries are emitted in the proper format if the user requests json vs the default.

internal/featuregate/featuregate.go
```diff
@@ -97,3 +97,9 @@ func (s *Stability) Set(str string) error {
 func (s Stability) Type() string {
 	return "<stability_level>"
 }
+
+// AtLeast will check if the stability level is allowed. For instance experimental would allow public preview, experimental and ga. Whereas ga would only allow ga.
+func (s Stability) AtLeast(stability Stability) bool {
+	return s <= stability
+
+}
```
rfratto, line 105: I'm struggling with the name here a little bit, since I'm reading it as the phrase "experimental is at least public preview" which sounds odd to me. 

What about something calling it something like "permits" so then you could read it as "experimental permits public preview functionality"?

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ Main (unreleased)
 
 - Don't restart tailers in `loki.source.kubernetes` component by above-average
   time deltas if K8s version is >= 1.29.1 (@hainenber)
+### Features
+
+- (_Public preview_) Add support for setting GOMEMLIMIT based on cgroup setting. (@mattdurham)
```
rfratto, line 32: This should go before Enhancements in the list

rfratto (approved):
LGTM following some feedback

internal/alloycli/cmd_run.go
mattdurham: I tried that. The format isnt set by this point so it errors, its possible we could default to one?

mattdurham: I could always move it after the logger is resolved but that feels like an odd time to set memory limits.

rfratto: > I think it might be worth fixing that so we can use the right logger here.

Since we've fixed issues about this in the past I'd say not using the right logger would be a regression here. IMO I think we need to do something to use the right logger, either fixing the buffering issue with the slog.Handler implementation or moving this later in evaluation at least temporarily (I agree it's weird for setting the memory limit to not happen as early as possible)

mattdurham: I am going to fork this out into its own PR, making the logger work with slog is bigger than this change.

rfratto: @mattdurham Is creating a PR to fix the logger still on your radar? Given this was merged first, there's a known regression in main now, and we're running out of time for 1.1 (RC gets cut this Thursday). 

If we can't fix this by Thursday we'll have to discuss this feature skipping 1.1 or putting out the release with a known regression.

mattdurham: It is, wrapping up testing now, it ended up being slightly more complicated than expected.

PR #661, opened 2024-04-24
pyroscope.java: Add glibc detection for arm64

simonswine:
This PR makes sure glibc is also detected correctly on ARM64 linux.


tpaschalis (approved):
Not an expert at this part of the codebase, but LGTM

PR #662, opened 2024-04-24
pyroscope.scrape: Standardize profile names

simonswine:
When a user adopts godeltaprof_, currently the resulting profile typename is shown separately from the default ones. That makes is harder than it should be to adopt godeltaprof.

I am proposing that we send the profiles still under the name without the godeltaprof prefix.

image

@aknuds1 reported that initially

This is a resubmit of https://github.com/grafana/agent/pull/6367, which was approved to merge

### DIFF
--- a/internal/component/pyroscope/scrape/target.go
+++ b/internal/component/pyroscope/scrape/target.go
@@ -65,9 +65,23 @@ type Target struct {
 func NewTarget(lbls, discoveredLabels labels.Labels, params url.Values) *Target {
 	publicLabels := make(labels.Labels, 0, len(lbls))
 	for _, l := range lbls {
-		if !strings.HasPrefix(l.Name, model.ReservedLabelPrefix) {
-			publicLabels = append(publicLabels, l)
+		if strings.HasPrefix(l.Name, model.ReservedLabelPrefix) {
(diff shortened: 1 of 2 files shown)


internal/component/pyroscope/scrape/target.go
```diff
@@ -65,9 +65,23 @@ type Target struct {
...
+			case pprofGoDeltaProfMemory:
+				l.Value = pprofMemory
+			case pprofGoDeltaProfBlock:
+				l.Value = pprofBlock
+			case pprofGoDeltaProfMutex:
+				l.Value = pprofMutex
+			}
+		}
+
+		publicLabels = append(publicLabels, l)
```
tpaschalis, line 84: Just to sanity check; Is there still the concern that @korniltsev raised in the agent PR about using this with clustering and hash collisions?

tpaschalis (approved):
LGTM % the question.

Just for awareness, we have some stricter [backward compatibility](https://grafana.com/docs/alloy/latest/introduction/backward-compatability/) rules that we're trying to follow. Since these metrics aren't used in any of the official dashboards in our alloy-mixin, this is good-to-go as-is

PR #665, opened 2024-04-24
tools/ci: fix pushed Windows "latest" name

### LINKED ISSUE #664: Latest Windows image using incorrect tag name
### What's wrong?

The latest Windows image is pushed as the tag `latest-nanoserver-1809`, which is not the convention we had discussed internally. (This is my fault, I didn't catch the inconsistency during my final launch review).

The latest image should instead be pushed as `nanoserver-1809`, with no `latest-` prefix.

### Steps to reproduce

N/A

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

rfratto:
The latest image for the Winows container was intended to be `nanoserver-1809`, but instead we were pushing `latest-nanoserver-1809`.

If we decide to merge this, I will manually create a `nanoserver-1809` tag to point to `v1.0.0-nanoserver-1809`, and leave the old tag intact until it is removed at a to-be-determined future date.

Fixes #664.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ Main (unreleased)
 
 - Add a description to Alloy DEB and RPM packages. (@rfratto)
 
+- The latest Windows Docker image is now pushed as `nanoserver-1809` instead of
+  `latest-nanoserver-1809`. The old tag will no longer be updated, and will be
+  removed in a future release. (@rfratto)
+
 v1.0.0 (2024-04-09)
 -------------------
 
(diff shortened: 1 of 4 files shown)


rfratto:
Open question: is this a breaking change?

mattdurham:
Personally it is a breaking change from a user but not in a way that violates our promise. We dont cover tag names and this would be a bug imo.

erikbaranowski (approved):
This looks like it is a bugfix exception and caught quickly so I think it is ok to change the tag.

PR #674, opened 2024-04-25
docs: remove 'd' from duration values spec

### LINKED ISSUE #599: Type duration "d" abbrev is not respected by alloy when reading it's config
### What's wrong?

Hi,

I tried to define an alloy configuration file like this:
```
loki.process "old_logs" {
  stage.drop {
    older_than = "1d"
  }
  forward_to = [loki.write.ingestor.receiver]
}
```
according the [documentation](https://grafana.com/docs/alloy/latest/concepts/configuration-syntax/expressions/types_and_values/#naming-convention) - `1d` should be valid abbreviation denoting duration of 1 day

After starting docker container with this configuration I got error captured in screenshot in attachment.
Screenshot from 2024-04-17 18-38-05

### Steps to reproduce

- create configuration for alloy
- start it in docker in any way. I used docker-compose

### System information

Fedora Linux 39 6.8.6-200.fc39.x86_64

### Software version

Grafana Alloy latest, which should be 2.9.0

### Configuration

```text
discovery.docker "stdout" {
  host_networking_host = "postgis-gateway"
}

loki.source.docker "gateway_logs" {
  targets = discovery.docker.stdout.targets
  forward_to = [loki.process.old_logs.receiver]
  labels = { "source" = "gw-stdout" }
}

loki.process "old_logs" {
  stage.drop {
    older_than = "1h"
  }
  forward_to = [loki.write.ingestor.receiver]
}

loki.write "ingestor" {
  endpoint {
    url = "http://loki:3100/loki/api/v1/push"
  }
}
```

### Logs

```text
In attachment
```

tpaschalis:
Fixes #599

This PR addresses a spec inconsistency. This is not a breaking change, since `d` was never actually supported as a valid duration unit.

I still think that the Go team had a [good rationale](https://github.com/golang/go/issues/11473) for not adding support for it in time.ParseDuration; if it's so important for people to use higher values we could find a way to support richer expressions such as `5 * 24h` in the future instead.

### DIFF
--- a/docs/sources/concepts/configuration-syntax/expressions/types_and_values.md
+++ b/docs/sources/concepts/configuration-syntax/expressions/types_and_values.md
@@ -28,10 +28,9 @@ In addition to the preceding types, the [component reference][] documentation us
   The key type of an object is always a string or an identifier converted into a string.
 * `list(T)`: an `array` with the value type`T`.
   For example, `list(string)` is an array where all the values are strings.
-* `duration`: a `string` denoting a duration of time, such as `"1d"`, `"1h30m"`, `"10s"`.
+* `duration`: a `string` denoting a duration of time, such as `"100ms"`, `"1h30m"`, `"10s"`.
   Valid units are:
 
-  * `d` for days.
   * `h` for hours.
   * `m` for minutes.
   * `s` for seconds.


wildum (approved):

tpaschalis:
cc @rfratto for visibility, I'll go ahead and merge this if there's no objections.

rfratto:
> since d was never actually supported as a valid duration unit.

To be precise, Alloy/Flow mode has never supported it, but Prometheus and static mode of Grafana Agent _do_ [thanks to `github.com/prometheus/common/model.Duration`](https://github.com/prometheus/common/blob/5c108f88a95c10f948725de41606571e0e5f12ca/model/time.go#L197)

Personally I find most people mean 1d to be 24h, even if sometimes a day is not exactly 24h. I'm ok with us fixing the docs since no component uses `model.Duration`, but we should double check that the config converters handle converting `1d` from Prometheus to `24h` in Alloy properly (they should, but wouldn't hurt to check).

rfratto (approved):

tpaschalis:
Thanks for the feedback! I've verified that the converters _do_ handle days properly (1d gets converted to 24h0m0s).

I'm merging this.

PR #678, opened 2024-04-25
Update pyroscope/ebpf to v0.4.6

simonswine:
Note: Pyroscope updated its prometheus/common dependency and hence this required further updates.

cc @korniltsev @andriikushch 

The new version v0.4.4 allows to configure the size of the PIDsmap, which has become a limitation for some deployments. (see https://github.com/grafana/pyroscope/pull/3181)

As part of the PR I also released v0.4.6, which updates the annotations to be "alloy" compatible, so the debug info is displayed as before.


korniltsev (approved):
Should we expose 	PIDMapSize     	SymbolsMapSize  to the river config? to *actually* allow resizing the maps?

simonswine:
@korniltsev I think it is worth giving this another try.

korniltsev (approved):

PR #682, opened 2024-04-26
loader: decrease log level of 'finished node evaluation' lines

tpaschalis:
#### PR Description

This PR decreases the log level of the `msg="finished node evaluation"` log lines, since they can drown out all other logs

No issue filed.

#### Notes to the Reviewer

I purposefully left the Info log level on the Apply function that is called on manual reloads of the configuration. I'm not 100% sure yet, but I think they might be marginally more useful to see new components being added in, but still they can be much more noisy than all other log sources.

I also think this doesn't warrant a CHANGELOG entry, if you think it does don't hesitate to push an extra commit for me here.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated (N/A) 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ Main (unreleased)
   `latest-nanoserver-1809`. The old tag will no longer be updated, and will be
   removed in a future release. (@rfratto)
 
+- The log level of `finished node evaluation` log lines has been decreased to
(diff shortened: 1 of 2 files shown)


rfratto (approved):
LGTM, probably deserves a changelog entry in "other changes"

PR #685, opened 2024-04-26
Add additional buckets to fanout histogram

jcreixell:
#### PR Description

  - The prometheus_fanout_latency is used to measure latencies for data moving between components.
  - The default buckets are limited to 10s, which limits the ability detect very high latencies.
  - Adds to additional buckets (30s, 60s) for additional resolution.

#### Notes to the Reviewer

  - I initially planned to replace some of the existing buckets instead of adding new ones, but I believe most components will fall into the smaller buckets and reducing the resolution there would lead to distortions in the quantile calculations (and potentially create the ilusion of performance degradation). Note however that the 2 new buckets will result in increased cardinality of 2 metrics per collector instance.


mattdurham (approved):
LGTM

PR #687, opened 2024-04-26
Otel receivers spawn on need

### LINKED ISSUE #644: Various "error":"proto: wrong wireType = 2 for field TimeUnixNano" messages when using otel.receiver.kafka
### What's wrong?

Our architecture is currently:

Applications -> OTel Collector (otel/opentelemetry-collector-contrib:0.98.0) -> Kafka -> Grafana Alloy (1.0.0) -> Grafana Tempo (2.4.0).

OTel Collector receives the traces from our apps, and exports them onto a Kafka topic, using the otlp_proto encoding. We've then setup the otel.receiver.kafka component in Alloy to consume from this Kafka topic, to then send onto Tempo via gRPC (as we're looking to enable Tempo Multi-tenancy, which isn't supported by Kafka). 

When Alloy starts consuming these messages, our Alloy logs start throwing various errors around wrong wireTypes, illegal wireTypes etc, which make no sense (previously when Kafka was being consumed by Tempo directly, we saw none of these errors).

### Steps to reproduce

1. Run a simple pipeline in OTel Collector on 0.98.0, exporting to a Kafka topic using the otlp_proto encoding format.
2. Run a simple Alloy pipeline to consume from Kafka, batch process then output via gRPC to Tempo
3. Watch the errors come in.

### System information

_No response_

### Software version

_No response_

### Configuration

```text
ALLOY:

otelcol.receiver.kafka "otelkafka" {
  brokers          = ["kafka-cluster-kafka-bootstrap.kafka.svc.cluster.local:9093"]
  protocol_version = "2.8.0"
  topic = "otlp-tracing"
  group_id = "alloy-otelkafka"
  client_id = "alloy-otelkafka"
  encoding = "otlp_proto"

  authentication {
    tls {
        ca_file = "/tmp/ca.crt"
        insecure = true
    }
  }

  message_marking {
    after_execution = true
    include_unsuccessful = true
  }

  output {
    metrics = [otelcol.processor.batch.otelkafka.input]
    logs    = [otelcol.processor.batch.otelkafka.input]
    traces  = [otelcol.processor.batch.otelkafka.input]
  }
}

otelcol.processor.batch "otelkafka" {
    send_batch_size = 10
    timeout = "100ms"
    output {
        metrics = [otelcol.exporter.otlp.otelkafka.input]
        logs    = [otelcol.exporter.otlp.otelkafka.input]
        traces  = [otelcol.exporter.otlp.otelkafka.input]
    }
}

otelcol.exporter.otlp "otelkafka" {
  client {
    endpoint = "grafana-tempo-distributed-distributor-discovery.monitoring.svc.cluster.local:4317"
    tls {
        insecure = true
        insecure_skip_verify = true
    }
  }
}
```

### Logs

```text
{"ts":"2024-04-23T09:17:23.578960444Z","level":"error","msg":"failed to unmarshal message","component_path":"/","component_id":"otelcol.receiver.kafka.otelkafka","error":"proto: illegal wireType 7"}
{"ts":"2024-04-23T09:17:23.582466437Z","level":"error","msg":"failed to unmarshal message","component_path":"/","component_id":"otelcol.receiver.kafka.otelkafka","error":"proto: wrong wireType = 2 for field TimeUnixNano"}
{"ts":"2024-04-23T09:17:32.771035129Z","level":"error","msg":"failed to unmarshal message","component_path":"/","component_id":"otelcol.receiver.kafka.otelkafka","error":"proto: Gauge: wiretype end group for non-group"}
{"ts":"2024-04-23T09:17:32.775138857Z","level":"error","msg":"failed to unmarshal message","component_path":"/","component_id":"otelcol.receiver.kafka.otelkafka","error":"proto: wrong wireType = 2 for field TimeUnixNano"}
```

wildum:
#### PR Description

Modify all otelcol receiver components to only start a receiver for a signal when the signal is needed downstream (when it's set in the output block).

This fixes an issue with the `otelcol.receiver.kafka` component when the wrong receiver would claim the topic, leading to decoding error.

Fixes #644 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol.receiver.kafka.md
+++ b/docs/sources/reference/components/otelcol.receiver.kafka.md
@@ -279,6 +279,8 @@ Regular expressions are not allowed in the `headers` argument. Only exact matchi
 
 ### output block
 
+> **WARNING**: only one signal type shoud be set per `otelcol.receiver.kafka` component
+
 {{< docs/shared lookup="reference/components/output-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
 ## Exported fields
--- a/internal/component/otelcol/receiver/kafka/kafka.go
+++ b/internal/component/otelcol/receiver/kafka/kafka.go
@@ -2,8 +2,11 @@
 package kafka
 
 import (
+	"strings"
 	"time"
 
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/alloy/logging/level"
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/otelcol"
 	"github.com/grafana/alloy/internal/component/otelcol/receiver"
@@ -23,6 +26,7 @@ func init() {
 		Args:      Arguments{},
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			checkOutput(args.(Arguments).Output, opts.Logger)
 			fact := kafkareceiver.NewFactory()
 			return receiver.New(opts, fact, args.(Arguments))
 		},
@@ -75,6 +79,23 @@ func (args *Arguments) SetToDefault() {
 	args.MessageMarking.SetToDefault()
 	args.HeaderExtraction.SetToDefault()
 	args.DebugMetrics.SetToDefault()
+}
+
+func checkOutput(output *otelcol.ConsumerArguments, logger log.Logger) {
+	var signals []string
+
+	if len(output.Logs) > 0 {
+		signals = append(signals, "logs")
+	}
+	if len(output.Metrics) > 0 {
+		signals = append(signals, "metrics")
+	}
+	if len(output.Traces) > 0 {
+		signals = append(signals, "traces")
+	}
+	if len(signals) > 1 {
+		level.Warn(logger).Log("msg", "only one output signal should be set", "signals", strings.Join(signals, ", "))
+	}
 }
 
 // Convert implements receiver.Arguments.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Main (unreleased)
 
 - Fixed issue where text labels displayed outside of component node's boundary. (@hainenber)
 
+- Fixed a bug where a topic was claimed by the wrong consumer type in `otelcol.receiver.kafka`. (@wildum)
+
 ### Other changes
 
 - Update `alloy-mixin` to use more specific alert group names (for example,
--- a/internal/component/otelcol/receiver/receiver.go
+++ b/internal/component/otelcol/receiver/receiver.go
@@ -149,36 +149,40 @@ func (r *Receiver) Update(args component.Arguments) error {
 		return err
 	}
 
-	var (
-		next        = rargs.NextConsumers()
-		nextTraces  = fanoutconsumer.Traces(next.Traces)
-		nextMetrics = fanoutconsumer.Metrics(next.Metrics)
-		nextLogs    = fanoutconsumer.Logs(next.Logs)
-	)
+	next := rargs.NextConsumers()
 
 	// Create instances of the receiver from our factory for each of our
 	// supported telemetry signals.
 	var components []otelcomponent.Component
 
-	tracesReceiver, err := r.factory.CreateTracesReceiver(r.ctx, settings, receiverConfig, nextTraces)
-	if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
-		return err
-	} else if tracesReceiver != nil {
-		components = append(components, tracesReceiver)
+	if len(next.Traces) > 0 {
+		nextTraces := fanoutconsumer.Traces(next.Traces)
+		tracesReceiver, err := r.factory.CreateTracesReceiver(r.ctx, settings, receiverConfig, nextTraces)
+		if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
+			return err
+		} else if tracesReceiver != nil {
+			components = append(components, tracesReceiver)
+		}
 	}
 
-	metricsReceiver, err := r.factory.CreateMetricsReceiver(r.ctx, settings, receiverConfig, nextMetrics)
-	if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
-		return err
-	} else if metricsReceiver != nil {
-		components = append(components, metricsReceiver)
+	if len(next.Metrics) > 0 {
+		nextMetrics := fanoutconsumer.Metrics(next.Metrics)
+		metricsReceiver, err := r.factory.CreateMetricsReceiver(r.ctx, settings, receiverConfig, nextMetrics)
+		if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
+			return err
+		} else if metricsReceiver != nil {
+			components = append(components, metricsReceiver)
+		}
 	}
 
-	logsReceiver, err := r.factory.CreateLogsReceiver(r.ctx, settings, receiverConfig, nextLogs)
-	if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
-		return err
-	} else if logsReceiver != nil {
-		components = append(components, logsReceiver)
+	if len(next.Logs) > 0 {
+		nextLogs := fanoutconsumer.Logs(next.Logs)
+		logsReceiver, err := r.factory.CreateLogsReceiver(r.ctx, settings, receiverConfig, nextLogs)
+		if err != nil && !errors.Is(err, otelcomponent.ErrDataTypeIsNotSupported) {
+			return err
+		} else if logsReceiver != nil {
+			components = append(components, logsReceiver)
+		}
 	}
 
 	// Schedule the components to run once our component is running.
--- a/internal/component/otelcol/receiver/receiver_test.go
+++ b/internal/component/otelcol/receiver/receiver_test.go
@@ -57,6 +57,22 @@ func TestReceiver(t *testing.T) {
 	require.NoError(t, waitTracesTrigger.Wait(time.Second), "consumer did not get invoked")
 }
 
+func TestReceiverNotStarted(t *testing.T) {
+	var (
+		waitConsumerTrigger = util.NewWaitTrigger()
+		onTracesConsumer    = func(t otelconsumer.Traces) {
+			waitConsumerTrigger.Trigger()
+		}
+	)
+	te := newTestEnvironment(t, onTracesConsumer)
+	te.Start(fakeReceiverArgs{
+		Output: &otelcol.ConsumerArguments{},
+	})
+
+	// Check that no trace receiver was started because it's not needed by the output.
+	require.ErrorContains(t, waitConsumerTrigger.Wait(time.Second), "context deadline exceeded")
+}
+
 type testEnvironment struct {
 	t *testing.T
 


internal/component/otelcol/receiver/kafka/kafka.go
```diff
@@ -23,6 +26,7 @@ func init() {
 		Args:      Arguments{},
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			checkOutputSignals(args.(Arguments).Output, opts.Logger)
```
wildum, line 29: I put a warn log here instead of returning an error with validate because it would be a breaking change for people who are using the workaround to have one receiver handling more than one signal. On the other hand it's an anti pattern so I would also be ok by making this an error

rfratto (commented):
I'm not sure about this one, especially because `checkOutputSignals` isn't handling updates. I find this to be a little bit complicated and potentially fixing a symptom of a larger underlying issue.

If this is a bug with upstream, we should help them fix that bug upstream. If this isn't a bug with upstream, I'd like to understand what upstream does differently than we do.

wildum:
> If this is a bug with upstream, we should help them fix that bug upstream. If this isn't a bug with upstream, I'd like to understand what upstream does differently than we do.

The Otel collector has [pipelines](https://opentelemetry.io/docs/collector/configuration/#pipelines). This looks like this for kafka:

```
service:
  pipelines:
    traces:
      receivers:
        - kafka
      exporters:
        - otlp 
```
Because the kafka receiver is set in the traces pipeline, the otel collector will only create a TraceReceiver to claim the provided kafka topic.

In Alloy we currently don't have a way to specify which signal the receiver should handle so we spawn by default the three receivers (metrics/logs/traces) for every receiver component. 

```
otelcol.receiver.kafka "otelkafka" {
  brokers                 = ["127.0.0.1:9094"]
  protocol_version = "2.8.0"
  encoding              = "otlp_proto"

  output {
    traces  = [otelcol.processor.batch.otelkafka.input]
  }
}
```
At best, the useless receivers that we spawn don't have any impact (fe the metricsReceiver and the logsReceiver that we spawn for the jaeger receiver don't do anything). But for Kafka all active receivers try to claim the given topic.

The goal of this PR is to use the output block as an equivalent to the pipeline feature of Otel and only spawns the receivers that are needed. In this case, because it only outputs traces the receiver should only create a TraceReceiver.

wildum:
it's a bit confusing to call everything receiver but that's how it's called in the code. An otel receiver contains up to three "signals" receivers: metrics, logs and traces

wildum:
> I'm not sure about this one, especially because checkOutputSignals isn't handling updates. 

Good point. If we go with this solution, I could try to move it or make it an error in the validate func if we want to prevent this behavior for the kafka receiver. Or we just keep the warning in the doc and if the user configure it wrong it will get a decoding error msg

rfratto:
> Because the kafka receiver is set in the traces pipeline, the otel collector will only create a TraceReceiver to claim the provided kafka topic.

But what happens upstream if you configure the kafka receiver in the logs and traces pipeline? Does it reject it?

internal/component/otelcol/receiver/kafka/kafka.go
ptodev, line 1: I personally don't think we need to change anything in this file. Maybe there is really a use case where you can configure 1 receiver to accept all 3 signals from the same topic? And yes, there might be warnings in the logs, but maybe it'd work? I think it's a use case which the Collector technically allows, so I don't think we should go out of our way to  say otherwise.

ptodev:
> > Because the kafka receiver is set in the traces pipeline, the otel collector will only create a TraceReceiver to claim the provided kafka topic.
> 
> But what happens upstream if you configure the kafka receiver in the logs and traces pipeline? Does it reject it?

It depends on whether you explicitly configured a Kafka topic. If you didn't, then it'll default to a different topic for every signal. If you did, it'll try to use that same topic for all signal types.

(I have never tried this with this receiver btw :) I'm just stating my expectations)

wildum:
Sorry I did not see your answer @ptodev, but you are correct indeed

internal/component/otelcol/receiver/kafka/kafka.go
wildum: > Maybe there is really a use case where you can configure 1 receiver to accept all 3 signals from the same topic?

This is not possible because only one receiver will claim the topic. The other ones wont consume from it. 1 signal receiver = 1 topic

rfratto:
> The behavior is the same as with Alloy: it starts the logsReceiver and the tracesReceiver and one of them will claim the topic. If it's the traceReceiver then it works. If it's the logsReceiver then it will fail to decode (it's stuck in a loop -> consumes a message -> fails to decode because the schema is wrong -> try to consume the next message etc...)

So I think my statement stands here; I would prefer reporting and help fix that behavior in upstream, since I think this is a symptom of a larger underlying issue rather than an issue with the framework we built out for creating OTel component wrappers.

internal/component/otelcol/receiver/kafka/kafka.go
wildum: The only workaround that works is that you have 3 topics with the exact names "otlp_spans", "otlp_metrics" and "otlp_logs" and that you explicitly set the topic to an empty string. With this hack, the trace receiver will claim its default topic "otlp_spans", the metric one will claim "otlp_metrics" and the logs one will claim "otlp_logs".

I don't like this behavior because it relies on the default values and because if you can configure only one topic on the receiver, then it should be able to handle only one topic. It's a hack that you can make it handle three topics. If we want one kafka receiver component to handle multiple topics then we should change the component in otel

wildum:
> So I think my statement stands here; I would prefer reporting and help fix that behavior in upstream, since I think this is a symptom of a larger underlying issue rather than an issue with the framework we built out for creating OTel component wrappers.

It's two different things.
The problem with Otel is that if you put the kafka receiver in the wrong pipeline then it doesn't work. That's not a bug, at most it's a questionable design. If you need to use kafka for traces, metrics and logs, then you configure 3 receivers, assign each the correct topic and you put each one in the correct pipeline.

For Alloy this is a bug because we cannot specify the pipeline. **ATM the only way to make the kafka receiver work consistenly is to explicitly set the topic to an empty string. If you set the topic argument to any other value, it will not work.**

rfratto:
>  The problem with Otel is that if you put the kafka receiver in the wrong pipeline then it doesn't work. That's not a bug, at most it's a questionable design. If you need to use kafka for traces, metrics and logs, then you configure 3 receivers, assign each the correct topic and you put each one in the correct pipeline.

I don't understand, how can a user put the Kafka receiver in a "wrong" pipeline in OpenTelemetry? Given that OpenTelemetry intends for a receiver to be re-used in any number of pipelines of any mix of telemetry types, I don't see how the Kafka receiver would have a "wrong" pipeline. 

For example, based on what I understand about OpenTelemetry, this should be valid: 

```yaml
receivers:
  kafka: # ... 
 
exporters: 
  otlp: # ... 

pipelines: 
  metrics:
     receivers: [kafka]
     exporters: [otlp]
  logs:
     receivers: [kafka]
     exporters: [otlp]
  traces:
     receivers: [kafka]
     exporters: [otlp]
```

I don't see anything in upstream documentation for the Kafka receiver would imply that this is invalid just for the Kafka receiver. Am I misunderstanding what you meant here?

ptodev:
@rfratto The config you [suggested](https://github.com/grafana/alloy/pull/687#issuecomment-2079709863) is going to work fine, but only if you don't explicitly set a kafka topic in the OTel config. 

If you do set a topic explicitly, then that topic will be used in every pipeline. This causes the kafka receiver to throw errors because all of logs/metrics/traces receivers will be trying to interpret the incoming data as logs/metrics/traces respectively. 

I'd have preferred if we can explicitly set a "log topic", "traces topic" and a "metrics topic", similarly to how the [otlp receiver](https://grafana.com/docs/alloy/latest/reference/components/otelcol.receiver.otlp/#http-block) can have `traces_url_path `, etc. So I do agree with you that there is scope to improve the component upstream. I don't see harm in merging the `receiver.go` changes in this PR though. There's no point in setting up receivers for telemetry types for which we don't need them.

wildum:
> > The problem with Otel is that if you put the kafka receiver in the wrong pipeline then it doesn't work. That's not a bug, at most it's a questionable design. If you need to use kafka for traces, metrics and logs, then you configure 3 receivers, assign each the correct topic and you put each one in the correct pipeline.
> 
> I don't understand, how can a user put the Kafka receiver in a "wrong" pipeline in OpenTelemetry? Given that OpenTelemetry intends for a receiver to be re-used in any number of pipelines of any mix of telemetry types, I don't see how an instance of the Kafka receiver component could be placed in a "wrong" pipeline.
> 
> For example, based on what I understand about OpenTelemetry, this should be valid:
> 
> ```yaml
> receivers:
>   kafka: # ... 
>  
> exporters: 
>   otlp: # ... 
> 
> pipelines: 
>   metrics:
>      receivers: [kafka]
>      exporters: [otlp]
>   logs:
>      receivers: [kafka]
>      exporters: [otlp]
>   traces:
>      receivers: [kafka]
>      exporters: [otlp]
> ```
> 
> I don't see anything in upstream documentation for the Kafka receiver would imply that this is invalid just for the Kafka receiver. Am I misunderstanding what you meant here?

The config is valid as long as you don't specify a topic. In this case you will get this:
- kafka for metrics will create a metricsReceiver which will claim the topic otlp_metrics
- kafka for traces will create a tracesReceiver which will claim the topic otlp_spans
- kafka for logs will create a logsReceiver which will claim the topic otlp_logs

If you specify one topic (let's say "test_topic") in your kafka receiver config, they will all try to claim it:
- kafka for metrics will create a metricsReceiver which will claim the topic test_topic
- kafka for traces will create a tracesReceiver which will claim the topic test_topic
- kafka for logs will create a logsReceiver which will claim the topic test_topic
Only one will get it. This is not good and will work if you are lucky, else you get decoding error and no data is forwarded.

One you can do if you need to specify different names is this:
```
receivers:
  kafka/traces:
    protocol_version: 2.8.0
    brokers: ["kafka:9092"]
    topic: tracing_custom_topic
    encoding: otlp_proto
  kafka/metrics:
    protocol_version: 2.8.0
    brokers: ["kafka:9092"]
    topic: metrics_custom_topic
    encoding: otlp_proto

exporters:
  otlp:

service:
  pipelines:
    traces:
      receivers:
        - kafka/traces
      exporters:
        - otlp
    metrics:
      receivers:
        - kafka/metrics
      exporters:
        - otlp
```
With this I can configure different kafka receivers with different topics and assign them to the correct pipelines. By correct pipeline I mean that the topic "tracing_custom_topic" which contains traces should only be consume by the traces pipeline. This use-case is currently not possible in Alloy.

rfratto:
Thanks, I see. Yes, in that case there would be a collision. However, I'd rather not us stray too far from the upstream behavior as we aim for compatibility, so I don't think we should reject someone configuring the component this way. However, I do think it's ok for us to mark the component unhealthy if we detect this scenario (multiple pipelines with the same topic set). 

I've also noticed that we don't set the default topic the same way as upstream. Upstream sets the default topic at runtime [based on the telemetry type](https://github.com/grafana/opentelemetry-collector-contrib/blob/9adb91e165fb90c22c0dab0e658bf04414bfcb7c/receiver/kafkareceiver/factory.go#L22-L25) the receiver is being instantiated for. We always set it to `otlp_spans`. We could fix this behaviour, but it might be a breaking change for anyone who is writing non-trace data to the `otlp_spans` topic. However it might be worth fixing this one anyway, since I do think this behavior is incorrect. 

Additionally, I understand now why we're changing all receivers to conditionally enable pipelines depending on where data is being routed. I think this needs a few test:

* Add tests to assert the expected behavior when you're sending (e.g.) logs to a component that doesn't have have enabled (i.e., it's not forwarding logs to another component). 
  * I don't know what the expected behavior should be, but at the very least it shouldn't panic.  
* Add tests to ensure that the code changes handles updates properly with enabling/disabling per-type instances of the underlying OTel component.

wildum:
> We always set it to otlp_spans. We could fix this behaviour, but it might be a breaking change for anyone who is writing non-trace data to the otlp_spans topic. 

Although this is a breaking change in theory. In practice it cannot be a breaking change because if you use the default value of the topic arg it won't work. The only way to have our otelcol.receiver.kafka component working is to set it to an empty string. We only found out now because no one was using this component I guess

wildum:
> I also still think we should report this behavior upstream and ensure it's intended on their side. I also think we should help them modify their documentation to warn of this behaviour if it is intended.

I can follow up on this :)

wildum:
>Add tests to assert the expected behavior when you're sending (e.g.) logs to a component that doesn't have have enabled (i.e., it's not forwarding logs to another component).
I don't know what the expected behavior should be, but at the very least it shouldn't panic.
Add tests to ensure that the code changes handles updates properly with enabling/disabling per-type instances of the underlying OTel component.

I will add the tests

rfratto:
> Although this is a breaking change in theory. In practice it cannot be a breaking change because if you use the default value of the topic arg it won't work. The only way to have our otelcol.receiver.kafka component working is to set it to an empty string. 

Yeah, that's a good point. Though my understanding is that it's not that it never worked at all, just that it's a race condition so there may be some missing data, right? 

> We only found out now because no one was using this component I guess

Neither our usage stats for Flow nor Alloy show any usage of this component. That doesn't mean that there isn't someone using it that's just not reporting usage stats, but it does seem reasonable to say that this component has little-to-no usage. 

I definitely think it's OK for us to fix this now 👍

ptodev:
If I remember correctly, the Kafka receiver upstream used to also set the topic to "otlp_spans" by default, for metrics/logs/traces pipelines. They later changed it to have a different topic depending on the type of telemetry. The Alloy defaults are consistent with the old Otel Collector defaults.

We could change our default to `""` and put in an explanation that it depends on the telemetry? Then it'd be consistent with the Collector.

wildum:
> Yeah, that's a good point. Though my understanding is that it's not that it never worked at all, just that it's a race condition so there may be some missing data, right?
Yeah saying that it "never worked" was a bit of a shortcut, you do have a chance to hit the right receiver although you need to roll the dice every time the sessions end and the receivers need to re-claim the topic

wildum:
> We could change our default to `""` and put in an explanation that it depends on the telemetry? Then it'd be consistent with the Collector.

I'm ok with the plan, should we explicitly say that if you leave it to "" then you can use the 3 signals for one receiver but if you set a topic then you should only use one signal? Just as documentation, no log in the code.

And then hopefully we can change the behaviour in Otel

rfratto:
> I'm ok with the plan, should we explicitly say that if you leave it to "" then you can use the 3 signals for one receiver but if you set a topic then you should only use one signal? Just as documentation, no log in the code.
>
> And then hopefully we can change the behaviour in Otel

Yeah, I think if the default is an empty string then it should use different topics internally and so it should be valid to use all three types for emitting telemetry. If you manually set a topic, then it's an invalid config and we should note that on the component health by marking it as unhealthy with a descriptive message. (We can mention this in the documentation as well).

And then yeah hopefully we can work with upstream to see if there's any possible improvements 👍

ptodev:
Actually, I don't think we can change the default to `""` because we'd be breaking backwards compatibility 😄  We could instead keep defaulting to `"otlp_spans"`, and just document that setting the topic to `""` has interesting side effects. And I do think we should merge the change this PR does to `receiver.go`

rfratto:
> Actually, I don't think we can change the default to "" because we'd be breaking backwards compatibility 😄 We could instead keep defaulting to "otlp_spans", and just document that setting the topic to "" has interesting side effects. And I do think we should merge the change this PR does to receiver.go

Have you seen the discussion about this above? We know that the component can't work properly today regardless of how you configure it, and that upstream have recently changed the defaults, so fixing this could fall under any one of these exceptions: 

> Specification errors: If a specification for a feature is found to be incomplete or inconsistent, fixing the specification may require a breaking change.
>
> Bugs: If a bug is found that goes against the documented specification of that functionality, fixing the bug may require breaking compatibility for users who are relying on the incorrect behavior.
>
> Upstream changes: Much of the functionality of Alloy is built on top of other software, such as OpenTelemetry Collector and Prometheus. If upstream software breaks compatibility, we may need to reflect this in Alloy.

We've also observed that the component has little-to-no usage (likely due to the bug), which makes it easier for us to justify an exception to the rule to fix the behavior as it's unlikely anyone would actually be broken.

ptodev:
Yes, as you say I think we'd be ok with such a change in this case 👍

docs/sources/reference/components/otelcol.receiver.kafka.md
```diff
@@ -40,13 +40,21 @@ Name | Type | Description | Default | Required
...
 `initial_offset` | `string` | Initial offset to use if no offset was previously committed. | `"latest"` | no
 `resolve_canonical_bootstrap_servers_only` | `bool` | Whether to resolve then reverse-lookup broker IPs during startup. | `"false"` | no
 
+If `topic` is not set, the `otelcol.receiver.kafka` will try to consume:
+
+* metrics from the topic `otlp_metrics`
+* traces from the topic `otlp_spans`
+* logs from the topic `otlp_logs`
+
+If `topic` is set to a specific value, then only the signal type that corresponds to the data stored in the topic must be set in the output block.
```
ptodev, line 56:
```suggestion
If `topic` is set to a specific value, then only the signal type that corresponds to the data stored in the topic must be set in the output block.
For example, if `topic` is set to `"my_telemetry"`, then the `"my_telemetry"` topic can only contain either metrics, logs, or traces. 
If it contains only metrics, then `otelcol.receiver.kafka` should be configured to output only metrics.
```

internal/component/otelcol/receiver/kafka/kafka.go
```diff
@@ -77,6 +78,27 @@ func (args *Arguments) SetToDefault() {
...
+			signals = append(signals, "logs")
+		}
+		if len(args.Output.Metrics) > 0 {
+			signals = append(signals, "metrics")
+		}
+		if len(args.Output.Traces) > 0 {
+			signals = append(signals, "traces")
+		}
+		if len(signals) > 1 {
+			return fmt.Errorf("if the argument topic is specified, only one signal can be set in the output block, current: %s", strings.Join(signals, ", "))
```
ptodev, line 96:
```suggestion
			return fmt.Errorf("only one signal can be set in the output block when a Kafka topic is explicitly set; currently set signals: %s", strings.Join(signals, ", "))
```

ptodev (approved):

internal/component/otelcol/receiver/kafka/kafka_test.go
```diff
@@ -433,3 +429,32 @@ func TestDebugMetricsConfig(t *testing.T) {
...
+		}
+	`
+	var args kafka.Arguments
+	require.NoError(t, syntax.Unmarshal([]byte(cfg), &args))
+
+	// Adding two traces consumer, expect no error
+	tracesConsumer := fakeconsumer.Consumer{}
+	args.Output.Traces = append(args.Output.Traces, &tracesConsumer)
+	tracesConsumer2 := fakeconsumer.Consumer{}
+	args.Output.Traces = append(args.Output.Traces, &tracesConsumer2)
```
rfratto, line 448: tiny nit: since the consumers aren't used after the append, the test might be slightly more readable if they were defined inline to the append:

```
args.Output.Traces = append(args.Output.Traces, &fakeconsumer.Consumer{}) 
args.Output.Traces = append(args.Output.Traces, &fakeconsumer.Consumer{}) 
```

(ditto with the other lines below)

rfratto (approved):
LGTM, thanks William!

PR #700, opened 2024-04-28
fix: Ensure HTTP server startup errors are logged (#694)

### LINKED ISSUE #694: No error logs when the HTTP server cannot start
### What's wrong?

If the HTTP server cannot start (such as the listen address is already in-use by another process), the HTTP server fails silently and doesn't generate any error-level logs.

This is related to #364, but slightly more specific, as the user has no signal that anything went wrong, other than self-scraping and scraping exporter components fail.  

Note that the HTTP service [does return an error](https://github.com/grafana/alloy/blob/c6f4f3054b9f5032ee8d38b1570bf451dfd58fd3/internal/service/http/http.go#L154), but this error [isn't logged](https://github.com/grafana/alloy/blob/c6f4f3054b9f5032ee8d38b1570bf451dfd58fd3/internal/alloy/internal/controller/scheduler.go#L140) anywhere. 

A reasonable fix here would be to change the code for the runnable scheduler to not ignore errors, or at least explicitly log errors before returning the error in the HTTP service.

### Steps to reproduce

1. Run two Alloy instances running listening on the same listen address.
2. Observe that the Alloy instance which started second does not have any log lines for `now listening for http traffic` nor any error indicating `failed to listen`. 

### System information

_No response_

### Software version

v1.0.0

### Configuration

N/A 

### Logs

N/A (it's the lack of logs which is the problem 😛)

mrashad10:
#### PR Description
This commit modifies the go routine that starts the HTTP server to explicitly log any errors encountered during startup, except for `http.ErrServerClosed`, which is expected during graceful shutdown. This change ensures that issues related to the server's startup, such as the address already being in use, are logged, making it easier to diagnose and fix.

Fixes #694

#### PR Checklist

### DIFF
--- a/internal/cmd/integration-tests/configs/prom-gen/main.go
+++ b/internal/cmd/integration-tests/configs/prom-gen/main.go
@@ -44,7 +44,11 @@ func main() {
 	}()
 	log.Printf("HTTP server on %s", listenAddress)
 
-	go func() { log.Fatal(server.ListenAndServe()) }()
+	go func() {
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("Server Listening error: %v", err)
+		}
+	}()
 
 	labels := map[string]string{
 		"address": address,


mrashad10:
As this is my first contribution, I welcome all your feedback and advice.

internal/cmd/integration-tests/configs/prom-gen/main.go
```diff
@@ -44,7 +44,11 @@ func main() {
 	}()
 	log.Printf("HTTP server on %s", listenAddress)
 
-	go func() { log.Fatal(server.ListenAndServe()) }()
+	go func() {
```
hainenber, line 47: I think you made changes in the wrong place. This `main.go is to run integration tests.

Robert already pointed at possibility of logging out the error before return it here. Could use something like `level.Error(s.log).Log("failed to start HTTP service", "msg", err.Error()`

https://github.com/grafana/alloy/blob/c6f4f3054b9f5032ee8d38b1570bf451dfd58fd3/internal/service/http/http.go#L154

hainenber:
hi @mrashad10, are you still working on this PR? If anything's troubling you, feel free to raise it here!

mrashad10:
Hi @hainenber

Sorry but I was busy last 2 weeks, I'm not sure that I will have time to work on it this week too.

dehaansa:
Closing this due to age, if you're interested in making this or a similar change please feel free to open a new PR or re-open this one.

PR #703, opened 2024-04-29
WIP: Fix staleness during target handover

thampiotr:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/discovery/distributed_targets.go
+++ b/internal/component/discovery/distributed_targets.go
@@ -8,48 +8,68 @@ import (
 // DistributedTargets uses the node's Lookup method to distribute discovery
 // targets when a component runs in a cluster.
 type DistributedTargets struct {
-	useClustering bool
-	cluster       cluster.Cluster
-	targets       []Target
+	localTargets     []Target
+	remoteTargetKeys map[shard.Key]struct{}
 }
 
 // NewDistributedTargets creates the abstraction that allows components to
 // dynamically shard targets between components.
-func NewDistributedTargets(e bool, n cluster.Cluster, t []Target) DistributedTargets {
-	return DistributedTargets{e, n, t}
+func NewDistributedTargets(clusteringEnabled bool, cluster cluster.Cluster, allTargets []Target) DistributedTargets {
+	// TODO(@tpaschalis): Make this into a single code-path to simplify logic.
+	if !clusteringEnabled || cluster == nil {
+		return DistributedTargets{
+			localTargets:     allTargets,
+			remoteTargetKeys: map[shard.Key]struct{}{},
+		}
+	}
+
+	peerCount := len(cluster.Peers())
+	localCap := len(allTargets) + 1
+	if peerCount != 0 {
+		localCap = (len(allTargets) + 1) / peerCount
+	}
+
+	localTargets := make([]Target, 0, localCap)
+	remoteTargetKeys := make(map[shard.Key]struct{})
+
+	for _, tgt := range allTargets {
+		targetKey := keyFor(tgt)
+		peers, err := cluster.Lookup(targetKey, 1, shard.OpReadWrite)
+		belongsToLocal := err != nil || len(peers) == 0 || peers[0].Self
+
+		if belongsToLocal {
+			localTargets = append(localTargets, tgt)
+		} else {
+			remoteTargetKeys[targetKey] = struct{}{}
+		}
+	}
+
+	return DistributedTargets{localTargets: localTargets, remoteTargetKeys: remoteTargetKeys}
 }
 
-// Get distributes discovery targets a clustered environment.
-//
-// If a cluster size is 1, then all targets will be returned.
-func (t *DistributedTargets) Get() []Target {
-	// TODO(@tpaschalis): Make this into a single code-path to simplify logic.
-	if !t.useClustering || t.cluster == nil {
-		return t.targets
+// LocalTargets returns the targets that belong to the local cluster node.
+func (dt *DistributedTargets) LocalTargets() []Target {
+	return dt.localTargets
+}
+
+// MovedAway returns targets that have been moved from this local node to another node, given the provided
+// previous targets distribution. Previous targets distribution can be nil, in which case no targets moved away.
+func (dt *DistributedTargets) MovedAway(prev *DistributedTargets) []Target {
+	if prev == nil {
+		return nil
 	}
-
-	peerCount := len(t.cluster.Peers())
-	resCap := (len(t.targets) + 1)
-	if peerCount != 0 {
-		resCap = (len(t.targets) + 1) / peerCount
-	}
-
-	res := make([]Target, 0, resCap)
-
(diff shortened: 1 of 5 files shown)


thampiotr:
I propose we use [our fork with this change](https://github.com/grafana/prometheus/pull/34) until we have a [decision on the issue](https://github.com/prometheus/prometheus/issues/14049) with the upstream.

CHANGELOG.md
```diff
@@ -78,6 +78,10 @@ v1.1.0-rc.0
 
 - Fix a bug where custom components would not shadow the stdlib. If you have a module whose name conflicts with an stdlib function
   and if you use this exact function in your config, then you will need to rename your module. (@wildum)
+  
+ - Fixed an issue with `prometheus.scrape` in which targets that move from one
+  cluster instance to another could have a staleness marker inserted and result
+  in a gap in metrics (@thampiotr)
```
rfratto, line 84: Just want to double check: is the intent that we're going to release this with 1.1.0 and we need to backport this to the release branch? (It's in the 1.1.0-rc.0 changelog section)

internal/component/discovery/distributed_targets.go
```diff
@@ -0,0 +1,91 @@
...
+
+// LocalTargets returns the targets that belong to the local cluster node.
+func (dt *DistributedTargets) LocalTargets() []Target {
+	return dt.localTargets
+}
+
+// MovedToRemoteInstance returns targets that have been moved from this local node to another node, given the provided
+// previous targets distribution. If a target has simply disappeared, it will not be returned. Previous targets
+// distribution can be nil, in which case no targets will be returned.
+func (dt *DistributedTargets) MovedToRemoteInstance(prev *DistributedTargets) []Target {
```
rfratto, line 63: It took me a little bit of time to wrap my head around whether this comment aligns with the implementation. I think it might be worth making this a little more explicit like:

```
// MovedToRemoteInstance returns the set of local targets from prev 
// that are no longer local in dt, indicating an active target has moved. 
// Only targets which exist in both prev and dt are returned. If prev 
// contains an empty list of targets, no targets are returned. 
```

Personally I think mentioning the parameter and receiver names makes it a bit easier for me to understand, though I have a hard time understanding why.

internal/component/prometheus/scrape/scrape.go
```diff
@@ -126,11 +128,15 @@ type Component struct {
...
-	args         Arguments
-	scraper      *scrape.Manager
-	appendable   *prometheus.Fanout
-	targetsGauge client_prometheus.Gauge
+	mut                 sync.RWMutex
+	args                Arguments
+	scraper             *scrape.Manager
+	appendable          *prometheus.Fanout
+	targetsGauge        client_prometheus.Gauge
+	movedTargetsCounter client_prometheus.Counter
```
rfratto, line 136: nit: do targetsGauge and movedTargetsCounter need to be protected by mut? AFAIK client_prometheus metrics are Goroutine safe. 

I recommend a pattern where a stanza of struct fields starts with a mut to indicate that everything within that stanza needs protection from that mut; if that isn't true here then I'd suggest considering moving the metric fields to a new stanza.

internal/util/assertmetrics/assert_metrics.go
```diff
@@ -0,0 +1,45 @@
...
+func AssertValueInStr(t assert.TestingT, allMetrics string, metricName string, labels labels.Labels, value float64) {
+	ls := ""
+	if len(labels) != 0 {
+		ls = strings.Replace(labels.String(), ", ", ",", -1)
+	}
+
+	// NOTE: currently no support for exemplars or explicit timestamps
+	expectedMetric := fmt.Sprintf("%s%s %v", metricName, ls, value)
+	assert.Contains(t, allMetrics, expectedMetric, "expected metric not found")
+}
```
rfratto, line 45: Just checking:

Have you seen [github.com/prometheus/client_golang/prometheus/testutil](https://pkg.go.dev/github.com/prometheus/client_golang@v1.19.0/prometheus/testutil)? Would that be useful here instead of writing our own mini-assertion library? (Though testutil only checks for metric names and not labels/values)

internal/component/prometheus/scrape/scrape.go
```diff
@@ -213,26 +228,61 @@ func (c *Component) Run(ctx context.Context) error {
...
-			c.mut.RUnlock()
 
-			promTargets := c.distTargets(targets, jobName, clusteringEnabled)
+			newTargetGroups, movedTargets := c.distributeTargets(targets, jobName, args)
+
+			// Make sure the targets that moved to another instance are NOT marked as stale. This is specific to how
+			// Prometheus handles marking series as stale: it is the client's responsibility to inject the
+			// staleness markers. In our case, for targets that moved to another instance in the cluster, we hand
+			// over this responsibility to the new owning instance. We must not inject staleness marker here.
+			c.scraper.DisableEndOfRunStalenessMarkers(jobName, movedTargets)
```
rfratto, line 247: I haven't looked too closely at all the tests, so this might already be handled, but I want to make sure: does this gracefully handle the case where a target moves to another node, and then moves back to the original node? 

For example, a scenario where the cluster timeline looks like:

```
node-a         # Initial state 
node-a, node-b # node-b joins the cluster 
node-a         # node-b leaves the cluster
```

rfratto (approved):
LGTM! I have a few questions, but nothing major that would prevent this from moving forward shortly. Nice work with the testing framework!

internal/component/prometheus/scrape/scrape.go
thampiotr: Good point. I don't think I have a full test for this and I suspect that if this happens fast enough, within 5 seconds refresh interval, it may result in some staleness markers being disabled unnecessarily. 

There are two ways to address this:
- Do not update to new targets until the previous targets are properly reloaded and sync'd. I don't think we have a good visibility of this completing now, it's all async.
- We can make sure that we `EnableStalenessMarkers` for all the current targets that we are updating to.
- We can also ignore this edge case - it may be rare and there are already situations where program can crash and staleness will never be injected, so it must be gracefully handled by the TSDB anyway.

rfratto: >  so it must be gracefully handled by the TSDB anyway.

Yes, Prometheus query engines forcibly inject a staleness marker at query time after 4 minutes from the last received sample, so I don't think it's that big of an issue if staleness markers get incorrectly disabled.

internal/util/assertmetrics/assert_metrics.go
```diff
@@ -0,0 +1,45 @@
+package assertmetrics
```
mattdurham, line 1: Can this file move to a `_test.go` file?

thampiotr: I don't think you can depend on `_test.go` files in tests of other packages... Similar to `testappender` package.

PR #705, opened 2024-04-29
internal/units: Create new units package

rfratto:
This commit creates a replacement for the github.com/alecthomas/units package. A replacement is desirable because the upstream package doesn't work very well with the Alloy syntax:

* We use the UnmarshalText implementation of `Base2Bytes`, which treats metric (MB) and IEC (MiB) suffixes as the same. 
* Base2Bytes always reports values as IEC suffixes when marshalling them back, which can confuse users into thinking a unit conversion has occurred somewhere.  

I think it's confusing to users that setting a limit of `4MB` actually sets a limit of `4MiB`, which is ~4% less than what the user intended. 

The new `internal/units` package supports both IEC and metric suffixes in the same data type, unmarshaling to the proper type based on what suffix the user provided. Additionally, it will `String()` to the most appropriate value, a byte count divisible by 1000 shows metric units, otherwise it shows IEC units.


internal/converter/internal/otelcolconvert/testdata/jaeger.alloy
```diff
@@ -5,11 +5,11 @@ otelcol.receiver.jaeger "default" {
 		thrift_http { }
 
 		thrift_binary {
-			max_packet_size = "63KiB488B"
+			max_packet_size = "65kB"
```
rfratto, line 8: Much better 😌 

However now I'm noticing that this would be a breaking change since the new package doesn't support this multi-suffix format, and this may affect users who used the converter and have this in their configuration files now. 

There's two things we could do here:

1. Support multi-suffix formats
2. Only use the new units package going forward for new fields

rfratto: I fixed this; the new package now also supports `63KiB488B`, though I wouldn't say we would recommend or document this.

internal/units/units.go
```diff
@@ -0,0 +1,214 @@
...
+	Kibibyte = 1024 * Byte
+	Mebibyte = 1024 * Kibibyte
+	Gibibyte = 1024 * Mebibyte
+	Tebibyte = 1024 * Gibibyte
+	Pebibyte = 1024 * Tebibyte
+	Exbibyte = 1024 * Pebibyte
+)
+
+var unitMap = map[string]Bytes{
+	"":   Byte,
```
mattdurham, line 46: Is this a carryover from the old library? Ideally people would specify the unit

rfratto: Yeah, it's from the old library. It also affects the syntax's behavior of coercing a number to a string when assigning to a string attribute; so that setting `max_bytes = 15` remains valid

internal/units/units.go
```diff
@@ -0,0 +1,214 @@
...
+	"b":  Byte,
```
mattdurham, line 47: Should this be bit? Which I am not sure how we would even handle.

PR #707, opened 2024-04-29
Fix handling of slog logs before logging initialized

mattdurham:
#### PR Description

Previously if you tried to use the logger for a slog style logger the system would panic. This adds a deferred handler for handling slog entries. The log buffer now supports normal logs and slog.Records and replays them in the appropriate order.

This is a lot of code to handle replaying of logs, its possible I could merge the deferred handler and the handler but then I think the logic would get real complicated. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/alloy/logging/logger.go
```diff
@@ -103,6 +119,9 @@ func NewDeferred(w io.Writer) (*Logger, error) {
 // updated.
 func (l *Logger) Handler() slog.Handler { return l.handler }
 
+// DeferredHandler is to be used if you need to use the slog logger before it is instantiated.
+func (l *Logger) DeferredHandler() slog.Handler { return l.deferredSlog }
```
rfratto, line 123: I'm a little concerned how much the deferred implementation leaks out to this file and to the caller, where they have to be able to understand when the main handler isn't fully initialized. 

> its possible I could merge the deferred handler and the handler but then I think the logic would get real complicated.

I think it might be possible for us to combine add buffering to the existing handler in a way which isn't too obtuse, perhaps with buffering logs on a call to `Handle` if the format isn't set, and then some kind of `flush()` method we can invoke on `Logger.Update`. 

If that doesn't work out, I'd alternatively like to explore reducing the blast radius of the deferred handler be minimized, where it doesn't touch the Logger subsystem and the caller doesn't have to understand the difference between Handler/DeferredHandler.

internal/alloy/logging/slogtest.go
```diff
@@ -0,0 +1,408 @@
+// Copyright 2023 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package logging implements support for testing implementations of log/slog.Handler.
+package logging
+
+// NOTE THIS IS COPIED FROM GO AND PRIMARILY USED SINCE WE OVERRIDE THE NAME OF SOME FIELDS
```
rfratto, line 8: I would generally prefer us finding a way to not override names of fields while testing compared to copying the stdlib tests; part of the reason to use upstream tests is that we get the latest test cases whenever we update Go.

internal/alloy/logging/logger.go
```diff
@@ -93,15 +98,29 @@ func NewDeferred(w io.Writer) (*Logger, error) {
...
+// The logger is not updated during initialization.
+func newDeferredTest(w io.Writer) (*Logger, error) {
+	l, err := NewDeferred(w)
+	if err != nil {
+		return nil, err
+	}
+	l.handler.replacer = testReplace
 
 	return l, nil
 }
```
rfratto, line 119: nit: if it's only used in tests, maybe this should be defined in a test file instead?

internal/alloy/logging/logger.go
```diff
@@ -124,11 +143,23 @@ func (l *Logger) Update(o Options) error {
...
-		slogadapter.GoKit(l.handler).Log(bufferedLogChunk...)
+		if len(bufferedLogChunk.kvps) > 0 {
+			// the buffered logs are currently only sent to the standard output
+			// because the components with the receivers are not running yet
+			slogadapter.GoKit(l.handler).Log(bufferedLogChunk.kvps...)
+		} else {
+			// We now can check to see if if our buffered log is at the right level.
+			if l.level.Level() <= bufferedLogChunk.record.Level {
+				// These will always be valid due to the build handlers call above.
+				bufferedLogChunk.handler.handle.Handle(context.Background(), bufferedLogChunk.record)
```
rfratto, line 160: I find this a bit hard to reason able since it's digging into the implementation details quite a bit, and I'd prefer not have to remember how the deferred handler works when reading the higher-level Logger code. Can this call `bufferedLogChunk.handler.Handle` instead?

internal/alloy/logging/logger.go
```diff
@@ -124,11 +143,23 @@ func (l *Logger) Update(o Options) error {
...
-		// the buffered logs are currently only sent to the standard output
-		// because the components with the receivers are not running yet
-		slogadapter.GoKit(l.handler).Log(bufferedLogChunk...)
+		if len(bufferedLogChunk.kvps) > 0 {
+			// the buffered logs are currently only sent to the standard output
+			// because the components with the receivers are not running yet
+			slogadapter.GoKit(l.handler).Log(bufferedLogChunk.kvps...)
+		} else {
+			// We now can check to see if if our buffered log is at the right level.
+			if l.level.Level() <= bufferedLogChunk.record.Level {
```
rfratto, line 158: nit: I'd rather this use `bufferedLogChunk.handler.Enabled(context.Background(), bufferedLogChunk.record.Level)`; to use the Handler API as it was intended to be used.

internal/alloy/logging/handler.go
```diff
@@ -141,54 +92,126 @@ func (h *handler) buildHandler() slog.Handler {
...
-
-		newAttrs = append(newAttrs, groupAttr)
-	} else {
-		newAttrs = append(newAttrs, attrs...)
-	}
+	newNest := make([]nesting, len(h.nested)+1)
+	newNest = append(newNest, h.nested...)
+	newNest = append(newNest, nesting{
+		attrs: attrs,
+	})
```
rfratto, line 114: This isn't correct; it's allocating `len(h.nested)+1` elements but then appending `len(h.nested)+1` _new_ elements to the end of the array, so there's a total of `2*(len(h.nested)+1)` elements at the end:

https://go.dev/play/p/ajPdFQmWOG_p

internal/alloy/logging/handler.go
```diff
@@ -141,54 +92,126 @@ func (h *handler) buildHandler() slog.Handler {
...
+		replacer: h.replacer,
 	}
 }
 
 func (h *handler) WithGroup(name string) slog.Handler {
+	newNest := make([]nesting, len(h.nested)+1)
+	newNest = append(newNest, h.nested...)
+	newNest = append(newNest, nesting{
+		group: name,
+	})
```
rfratto, line 131: Same bug as above

internal/alloy/logging/handler.go
```diff
@@ -26,12 +25,18 @@ type handler struct {
 	leveler   slog.Leveler
 	formatter formatter
 
-	attrs []slog.Attr
-	group []string
+	nested []nesting
```
rfratto, line 28: Nice, this is a good improvement and makes some of the logic simpler 👍

rfratto (commented):
Thanks! This is my last round of feedback, should be good to go after this

internal/alloy/logging/deferred_handler.go
```diff
@@ -0,0 +1,107 @@
...
+		return d.handle.WithAttrs(attrs)
+	}
+
+	child := &deferredSlogHandler{
+		attrs:    attrs,
+		children: make([]*deferredSlogHandler, 0),
+		l:        d.l,
+		parent:   d,
+	}
+	d.children = append(d.children, child)
```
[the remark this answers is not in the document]
mattdurham: Pretty sure after the update is called then I can at least set the children to nil which should then allow them to be GCed once the underlying handlers are no longer used.

internal/alloy/logging/slogtest.go
mattdurham: I managed to get rid of this, the tests do a bit of hijinks but probably worth it.

mattdurham:
Made the changes, also good for pushing to use the default slogtest. I have to add a little hack but it works and is much cleaner.

internal/alloy/logging/logger.go
```diff
@@ -124,11 +109,23 @@ func (l *Logger) Update(o Options) error {
...
-		slogadapter.GoKit(l.handler).Log(bufferedLogChunk...)
+		if len(bufferedLogChunk.kvps) > 0 {
+			// the buffered logs are currently only sent to the standard output
+			// because the components with the receivers are not running yet
+			slogadapter.GoKit(l.handler).Log(bufferedLogChunk.kvps...)
+		} else {
+			// We now can check to see if if our buffered log is at the right level.
+			if bufferedLogChunk.handler.handle.Enabled(context.Background(), bufferedLogChunk.record.Level) {
+				// These will always be valid due to the build handlers call above.
+				bufferedLogChunk.handler.handle.Handle(context.Background(), bufferedLogChunk.record)
```
rfratto, line 126: Double-checking: can this skip `handler.handle` and directly use handler as the implementation for slog.Handler instead? (This is what I was asking for in my last comment here)

```suggestion
			if bufferedLogChunk.handler.Enabled(context.Background(), bufferedLogChunk.record.Level) {
				bufferedLogChunk.handler.Handle(context.Background(), bufferedLogChunk.record)
```

This would go through a level of indirection but it removes needing to worry about whether the inner handle is set.

rfratto (approved):
LGTM!

internal/alloy/logging/logger.go
mattdurham: Yes, done.

PR #716, opened 2024-04-30
feat(helm/chart/rbac): allow granular cfg for clusterRole's permissions

### LINKED ISSUE #196: Support configuring agent clusterRole permission in values.yaml
### Request

In the agent helm chart, the clusterRole perm are configured in https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/templates/rbac.yaml. Depending on the components used, the user may not need all the permissions. Please make the perm configurable from values.yaml.

### Use case

Allow users to configure clusterRole permissions based on the flow mode Components used in the agent.

hainenber:

Fixes #196 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/operations/helm/charts/alloy/values.yaml
+++ b/operations/helm/charts/alloy/values.yaml
@@ -117,6 +117,20 @@ image:
 rbac:
   # -- Whether to create RBAC resources for Alloy.
   create: true
+    # -- Grant rules to allow created RBAC resources to work with `discovery.kubernetes` component.
+  allowDiscoveryKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `loki.source.kubernetes` and `loki.source.podlogs` component.
+  allowLokiSource: true
+  # -- Grant rules to allow created RBAC resources to work with `mimir.rules.kubernetes` component.
+  allowMimirRulesKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `prometheus.kubernetes.*` components.
+  allowPrometheusKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `eventhandler` integration.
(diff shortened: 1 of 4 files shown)


operations/helm/charts/alloy/values.yaml
```diff
@@ -120,6 +120,20 @@ image:
...
+  # -- Grant rules to allow created RBAC resources to work with `mimir.rules.kubernetes` component.
+  allowMimirRulesKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `prometheus.kubernetes.*` components.
+  allowPrometheusKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `eventhandler` integration.
+  allowEventHandler: true
+  # -- Grant rules to allow created RBAC resources to work with `remote.kubernetes.*` components.
+  allowRemoteKubernetes: true
+  # -- Grant rules to allow created RBAC resources to work with `otelcol.processor.k8sattributes` component.
+  allowOtelcolProcessorK8sAttributes: true
```
rfratto, line 136: I'm a little worried about tying Helm chart settings to knowledge of specific component names:

* The names might diverge with what's in Alloy over time (even right now, `eventhandler` from the comment is stale and from static mode; it should be `loki.source.kubernetes_events`). 
* There may be other components that require the same permissions, but aren't documented in the comments here. 

Can we find a way to partition these by permissions being granted rather than the main component that uses them? For example, instead of `allowLokiSource` it might be broken down into `readPods` and `readPodLogs`. 

Doing this is probably going to lead to having way more settings here, but it will also be decoupled from components and give users more granular control over what Alloy has permissions to access.

PR #718, opened 2024-04-30
feat(remote/aws): support AWS Secrets Manager as remote component

### LINKED ISSUE #689: support AWS secret manager
### Request

I want to be able to retrieve values from AWS Secret Manager in Grafana Alloy and make them available for use in Alloy's configuration.

### Use case

The following is used in scenarios that require sensitive information:

```
metrics.remote_write "prod" {
  remote_write {
    url = "https://mimir:9009/api/v1/push"
    basic_auth {
      username = remote.awssecretsmanager.remote_write.data.username
      password = remote.awssecretsmanager.remote_write.data.password
    }
  }
}
```

hainenber:

Closes #689 

#### Notes to the Reviewer

- Should I poll the Secrets Manager API continuously to fetch latest secret. Pricing per API calls for Secrets Manager is much expensive compared to AWS SSM Parameter Store and AWS S3 API calls so I'm hesitant to implement poller. Would like to hear everyone thoughts about it.
- I expose the inner `component.Component` in `componenttest` in order to verify the component's health status. Is this A-OK?

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests added


hainenber:
Thanks Matt! I'll implement the poller and fix whatever makes the tests failed.

Maybe in the long run,  we can have a reusable implementation. But that's for another day, I suppose :D

internal/component/remote/s3/s3.go
```diff
@@ -106,11 +100,11 @@ func (s *Component) Run(ctx context.Context) error {
 func (s *Component) Update(args component.Arguments) error {
 	newArgs := args.(Arguments)
 
-	s3cfg, err := generateS3Config(newArgs)
+	awsCfg, err := aws_common_config.GenerateAWSConfig(newArgs.Options.Client)
 	if err != nil {
 		return nil
```
hainenber, line 105: @rfratto pinging 'cause you authored this in the past. Do you know why this returns `nil` in the present of non-nil error? 👀

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/remote.aws.secrets_manager/
+description: Learn about remote.aws.secrets_manager
+title: remote.aws.secrets_manager
+---
+
+# remote.aws.secret_manager
+
+`remote.aws.secrets_manager` securely exposes value of secrets located in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/) to other components.
```
clayton-cornell, line 9:
```suggestion
`remote.aws.secrets_manager` securely exposes the secrets located in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/) to other components.
```

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
...
+By default, the secret would be fetched one time only at startup. If configured, the secret will be polled for changes so that the most recent value is always available.
```
clayton-cornell, line 10:
```suggestion
By default, the secret is fetched once only at startup. If configured, the secret is polled for changes so that the most recent value is always available. 
```

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
...
+description: Learn about remote.aws.secrets_manager
+title: remote.aws.secrets_manager
+---
+
+# remote.aws.secret_manager
+
+`remote.aws.secrets_manager` securely exposes value of secrets located in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/) to other components.
+By default, the secret would be fetched one time only at startup. If configured, the secret will be polled for changes so that the most recent value is always available. 
+
+Beware that this could incur cost due to frequent API calls.
```
clayton-cornell, line 12:
```suggestion
{{< admonition type="note" >}}
The polling for changes could incur costs due to frequent API calls.
{{< /admonition >}}
```

Are the costs monetary, or throttling, or something else?  Are the costs specific/identifiable?

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
...
+
+Multiple `remote.aws.secrets_manager` components can be specified using different name
+labels. By default, [AWS environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) are used to authenticate against AWS. The `key` and `secret` arguments inside `client` blocks can be used to provide custom authentication.
```
clayton-cornell, line 15:
```suggestion
You can specify multiple `remote.aws.secrets_manager` components by using a unique label for each component.
By default, [AWS environment variables][] are used to authenticate against AWS.
For custom authentication, you can use the `key` and `secret` arguments inside `client` blocks.

 [AWS environment variables]: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
```

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
...
+Name             | Type       | Description                                                              | Default | Required
+-----------------|------------|--------------------------------------------------------------------------|---------|---------
+`id`             | `string`   | Secret ID.                                                               |         | yes
+`poll_frequency` | `duration` | How often to poll the API for changes.                                   |         | no
+
+## Blocks
+
+Hierarchy | Name       | Description                                        | Required
+----------|------------|----------------------------------------------------|---------
+client    | [client][] | Additional options for configuring the AWS client. | no
```
clayton-cornell, line 38:
```suggestion
client    | [client][] | Additional AWS client configuration options. | no
```

docs/sources/reference/components/remote.aws.secrets_manager.md
```diff
@@ -0,0 +1,92 @@
...
+
+Name      | Type                 | Description                
+----------|----------------------|---------------------------------------------------------
+`data`    | `map(secret)`        |  Data from the secret obtained from AWS Secrets Manager.
+
+The `data` field contains a mapping from data field names to values.
+
+## Component health
+
+Instances of `remote.aws.secrets_manager` report as healthy if most recent fetch of stored secrets was successful.
```
clayton-cornell, line 67:
```suggestion
Instances of `remote.aws.secrets_manager` report as healthy if the most recent fetch of stored secrets was successful.
```

docs/sources/reference/components/remote.aws.secrets_manager.md
hainenber: Monetary cost

clayton-cornell:
Docs look ok.

clayton-cornell (dismissed):
Approving for docs. Still needs a code review fom @grafana/grafana-alloy-maintainers

internal/component/remote/s3/s3.go
rfratto: I think remote.s3 is originally from @mattdurham; do you remember if this is intentional? It does look a little like a typo to silently ignore the error here.

internal/component/remote/s3/types.go
```diff
@@ -15,19 +16,13 @@ type Arguments struct {
...
-// Client implements specific AWS configuration options
 type Client struct {
-	AccessKey     string            `alloy:"key,attr,optional"`
-	Secret        alloytypes.Secret `alloy:"secret,attr,optional"`
-	Endpoint      string            `alloy:"endpoint,attr,optional"`
-	DisableSSL    bool              `alloy:"disable_ssl,attr,optional"`
-	UsePathStyle  bool              `alloy:"use_path_style,attr,optional"`
-	Region        string            `alloy:"region,attr,optional"`
-	SigningRegion string            `alloy:"signing_region,attr,optional"`
+	aws_common_config.Client
```
rfratto, line 24: Can we add a test for decoding this? I don't think alloy/syntax supports anonymous fields like this, and they have to be named like

```go
type Client struct {
  CommonConfig aws_common_config.Client `alloy:",squash"` 
  UsePathStyle bool                     `alloy:"use_path_style,attr,optional"`
}

```

internal/component/common/config/aws/aws.go
```diff
@@ -0,0 +1,86 @@
...
+
+func GenerateAWSConfig(o Client) (*aws.Config, error) {
+	configOptions := make([]func(*aws_config.LoadOptions) error, 0)
+	// Override the endpoint.
+	if o.Endpoint != "" {
+		endFunc := aws.EndpointResolverWithOptionsFunc(func(service, region string, _ ...interface{}) (aws.Endpoint, error) {
+			// The S3 compatible system used for testing with does not require signing region, so it's fine to be blank
+			// but when using a proxy to real S3 it needs to be injected.
+			return aws.Endpoint{
+				PartitionID:   "aws",
```
rfratto, line 32: This wasn't set before in the old code; does this break anything for remote.s3?

internal/component/common/config/aws/aws.go
```diff
@@ -0,0 +1,86 @@
...
+					},
+				},
+			},
+		)
+		configOptions = append(configOptions, httpOverride)
+	}
+
+	if o.Region != "" {
+		configOptions = append(configOptions, aws_config.WithRegion(o.Region))
+	}
```
rfratto, line 57: This wasn't set before in the old code; does this break anything for remote.s3?

rfratto:
> I think it would be nice if it could poll but set the default to not poll. Most of our pulling mechanisms support it so it would be odd if this one did not have that capability.

@mattdurham Can you help me understand setting the default to not poll for updates? Don't all of our components default to polling in the background?

internal/component/remote/aws/secrets_manager/secrets_manager.go
```diff
@@ -0,0 +1,194 @@
...
+
+// DefaultArguments holds default settings for Arguments.
+var DefaultArguments = Arguments{
+	SecretVersion: "AWSCURRENT",
+}
+
+// SetToDefault implements syntax.Defaulter.
+func (a *Arguments) SetToDefault() {
+	*a = DefaultArguments
+}
```
rfratto, line 60: We've started to move away from the pattern of having default arguments as a variable because in some cases (like a default map or slice) memory gets accidentally shared and the defaults get overridden. 

```suggestion

// SetToDefault implements syntax.Defaulter.
func (a *Arguments) SetToDefault() {
	*a = Arguments{
		SecretVersion: "AWSCURRENT",	
	}
}
```

internal/component/remote/aws/secrets_manager/secrets_manager.go
```diff
@@ -0,0 +1,194 @@
...
+	s.handlePolledSecret(res)
+
+	return s, nil
+}
+
+func (s *Component) Run(ctx context.Context) error {
+	if s.pollFrequency > 0 {
+		go s.handleSecretUpdate(ctx)
+		go s.watcher.run(ctx)
+	}
```
rfratto, line 114: There's a bug here: If the poll frequency is set to non-zero after the component was already constructed, then it will never poll.

internal/component/remote/aws/secrets_manager/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+		select {
+		case <-w.pollTicker.C:
+			w.getSecretAsynchronously(ctx)
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (w *watcher) getSecretAsynchronously(ctx context.Context) {
```
rfratto, line 76: nit: technically this method gets the secret synchronously. A more fitting name might be `getAndSendSecret`, as the method gets the secret and then writes the result to a channel.

internal/component/remote/aws/secrets_manager/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	w.mut.Lock()
+	defer w.mut.Unlock()
+
+	res := w.getSecret(context.Background())
```
rfratto, line 80: Is there a reason we're not propagating the context here?

rfratto (commented):
Nice! Few questions (for both you and @mattdurham) but this is coming along nicely :)

internal/component/remote/s3/s3.go
mattdurham: Ooof, I cant think of a good reason why I did that.

mattdurham:
My thought on the default to not polling is related to this comment `Pricing per API calls for Secrets Manager is much expensive compared to AWS SSM Parameter Store and AWS S3 API.` if it is expensive default to no but allow it to be poll. Also fine with the other direction, default to a low poll rate.

rfratto:
Comparing the pricing of [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/pricing/) to [S3](https://aws.amazon.com/s3/pricing/) does show that secrets manager is more expensive, but perhaps not significantly:

* AWS Secrets manager: $0.05 per 10,000 API calls
* S3: $0.004 per 10,000 GET calls

It will take 200,000 polls of secrets for it to cost $1 USD. If we set the polling interval to hourly, and you had 100 Alloy instances all running the same component, that would still only be 73,000 polls per month ($0.36). 

It seems like the cost is reasonable enough if we used a 1h polling interval by default. If you're running Alloy in an environment where you need AWS secrets manager (likely Enterprise-like), running Alloy itself probably costs orders of magnitude more money than the secrets manager bill would be. 

WDYT @mattdurham?

hainenber:
I'm closing this per summer cleanup. Feel free to duplicate another one if this is still relevant to the project's cause. I'm thankful for all maintainer's effort here <3

PR #721, opened 2024-04-30
prometheus.remote_write: notify remote_write of data upon WAL commit

rfratto:
This commit fixes an issue where the remote_write queue only checked for new WAL data every 15 seconds. remote_write can be notified of new data immediately by calling `Notify` on the remote storage, but we weren't doing this, which caused us to fall back to the 15 second interval.


mattdurham (approved):
Lgtm

PR #724, opened 2024-04-30
packaging: don't assume /var/lib/alloy exists when running chown

rfratto:
The /var/lib/alloy directory may not necessarily exist when installing or upgrading Alloy, which would cause the chown/chmod of the directory to fail.

Related to grafana/alloy#723, though I haven't been able to reproduce the specific error that user was seeing.


rfratto:
Locally ran `make test-packages` to make sure everything still works. I'm not entirely convinced this is necessary, and I still don't understand what the deal with #723, but at the very least it's probably good to have this change just to be defensive.

erikbaranowski (approved):

Issue #630, opened 2024-04-22
Grafana Alloy container fails to start: exec /bin/alloy: operation not permitted


PatMis16:
### What's wrong?
Grafana Alloy needs to be deployable to an EKS cluster with enforced `podSecurityContext` and container `securityContext` policies. Our policy mandates that the container operates as a non-root user (`runAsNonRoot: true`) with all capabilities dropped:

**`securityContext`**
```
securityContext:
      readOnlyRootFilesystem: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
```
**`podSecurityContext`**
```
podSecurityContext:
      fsGroup: 2000
      runAsNonRoot: true
      runAsUser: 473
```

Deploying Grafana Alloy to EKS with a `podSecurityContext` that specifies `runAsUser` results in the container failing to start. The error logged is:
```
exec /bin/alloy: operation not permitted
``` 
**In our organization running containers as root is not permitted and enforced by Kyverno policies.**

### Steps to reproduce

To reproduce the deployment, install Grafana Alloy using the specified `podSecurityContext` while ensuring that no Custom Resource Definitions (CRDs), Custom Resources (CRs), ClusterRoleBindings (CRBs), or Role-Based Access Control (RBAC) configurations are deployed as they are also not allowed.

### System information
Deployment on AWS EKS is enhanced with Kyverno Policies, ensuring adherence to the organization's standards.

### Software version

v1.0.0

### Configuration

```text
# -- Overrides the chart's name. Used to change the infix in the resource names.
nameOverride: null

# -- Overrides the chart's computed fullname. Used to change the full prefix of
# resource names.
fullnameOverride: null

## Global properties for image pulling override the values defined under `image.registry` and `configReloader.image.registry`.
## If you want to override only one image registry, use the specific fields but if you want to override them all, use `global.image.registry`
global:
  image:
    # -- Global image registry to use if it needs to be overriden for some specific use cases (e.g local registries, custom images, ...)
    registry: ""

    # -- Optional set of global image pull secrets.
    pullSecrets: []

  # -- Security context to apply to the Grafana Alloy pod.
  podSecurityContext:
    podSecurityContext:
      fsGroup: 2000
      runAsNonRoot: true
      runAsUser: 473

crds:
  # -- Whether to install CRDs for monitoring.
  create: true

## Various Alloy settings. For backwards compatibility with the grafana-agent
## chart, this field may also be called "agent". Naming this field "agent" is
[...]

Issue #177, opened 2024-04-11
/bin/alloy: Operation not permitted


aleixmorgadas:
### What's wrong?

When deploying to [Render](https://render.com/) as a private service, Alloy doesn't start and crashes with an obscure message.

```
root@<host>:~# /bin/alloy          
bash: /bin/alloy: Operation not permitted
```

I debugged the service to try to find more without success:

```
root@<host>:~# strace -f /bin/alloy 
execve("/bin/alloy", ["/bin/alloy"], 0x7ffdec45b6c8 /* 40 vars */) = -1 EPERM (Operation not permitted)
strace: exec: Operation not permitted
+++ exited with 1 +++
```

I also opened a [post](https://community.render.com/t/unable-to-deploy-grafana-alloy/20716/2) at Render.

### Steps to reproduce

Deploy Grafana Alloy to Render.com using the next Dockerfile

```Dockerfile
# syntax = docker/dockerfile:1.2
FROM grafana/alloy:latest
ENTRYPOINT ["/bin/alloy"]
CMD ["run", "--server.http.listen-addr=0.0.0.0:12345", "/etc/alloy/config.alloy"]
```

### System information

Linux 5.15.0-1052-aws #57-Ubuntu SMP Mon Jan 15 15:24:00 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Software version

v1.0.0

rfratto:
Thanks for reporting. This sounds like something specific to Render, since I can't reproduce locally. Can you share more about Render's operating environment that would maybe clue us in to what might be going wrong here? (Such as security settings?)

aleixmorgadas:
I don't find relevant information, but I know that they provide instances on top of AWS. If they reply, I will pass the information here.

Meanwhile, I will try to deploy it from the source code :+1:

aleixmorgadas:
I have successfully deployed Alloy by:

1. Forking the project and deploy a Golang service from source
2. Attaching a Disk

Looks like Alloy (differently than Agent) expects to write to disk. Is there an option to disable it?

https://community.render.com/t/unable-to-deploy-grafana-alloy/20716/3?u=aleixmorgadas

aleixmorgadas:
Which are the directories that Alloy uses to write files into? I think the issue is that it uses more than one directory like the one passed as parameter `--storage.path=/var/lib/alloy/data` but other ones.

rfratto:
[...]

PR #726, opened 2024-05-01
Remove set bind permissions

### LINKED ISSUE #303: Grafana agent support runAsNonRoot
### Request

Grafana agent should runAsNonRoot in order to follow [Kubernetes Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/). 

Currently use runAsNonRoot is not permited and if we set runAsUser the agent do not work correctly. 

### Use case

Some use case do not require to run the Agent as root.

BlackDex:
#### PR Description

In this PR https://github.com/grafana/agent/pull/6817 the setcap was added for the alloy/agent binary for some reason unknown to me.

Since the docker engine allows binding to unpriviliged ports already for a long time via https://github.com/moby/moby/pull/41030, which was implemented in the Docker engine v20.10.x and also other container runtimes.

Fixes #117
Fixes #303

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] ~Documentation added~
- [ ] ~Tests updated~
- [ ] ~Config converters updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,9 @@ Main (unreleased)
   `/var/lib/alloy` exists before configuring its permissions and ownership.
   (@rfratto)
 
+- Remove setcap for `cap_net_bind_service` to allow alloy to run in restricted environments.
+  Modern container runtimes allow binding to unprivileged ports as non-root. (@BlackDex)
+
 v1.0.0 (2024-04-09)
 -------------------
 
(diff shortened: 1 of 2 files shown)


captncraig (approved):

rfratto:
@BlackDex Thanks for the PR! Based on the investigation @captncraig has done today, it seems like this is safe to merge as the capability was only added for Grafana Agent back when the default listen port on Kubernetes was 80. That's no longer true anymore, so the cap doesn't need to be added. 

You mention that this fixes #117 but that's a merged PR, what issue did you mean there?

captncraig:
Possibly fixes #630

captncraig:
I think #177 is the other issue

BlackDex:
@rfratto , thanks for the extra info 😄. That clarifies things.

BlackDex:
@captncraig indeed, i made a typo there.

PR #733, opened 2024-05-01
remove setcap from docker file

captncraig:
This was needed when the helm chart was using port 80 by default, if you were running as a non-root user.

Now that 12345 is a more universal default, I don't think this is needed. But it is harmful because you can't drop all capabilities anymore since the binary requires net_bind_service.

If you are running as root, this should not affect you. The only people that might notice this are people running as non-root, binding to port 80. I do not have a solution for that.


rfratto:
@captncraig Is this a dupe of #726?

captncraig:
Lol, yes. Will close this.

PR #741, opened 2024-05-02
Fix syntax error in first code block and execute TS in command

fanuch:
#### PR Description

\- 1 -

There is a syntactical error in first block which listed `loki.source.api labels` as:
```
    labels = { "source": "api" }
```
instead of the correct format of:
```
    labels = { source = "api" }
```
as per [documentation](https://grafana.com/docs/alloy/latest/reference/components/loki.source.api/) (and second reference below which was correct)

\- 2 -

Also, placing timestamp command directly in the call makes it easier to spam logs for testing and troubleshooting.
```
... "timestamp": "'"$(date -u +"%Y-%m-%dT%H:%M:%SZ")"'"}' 
```

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/tutorials/processing-logs/index.md
+++ b/docs/sources/tutorials/processing-logs/index.md
@@ -38,7 +38,7 @@ loki.source.api "listener" {
         listen_port    = 9999
     }
 
-    labels = { "source": "api" }
+    labels = { source = "api" }
 
     forward_to = [loki.process.process_logs.receiver]
 }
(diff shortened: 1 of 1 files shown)


clayton-cornell (approved):
@fanuch Thanks for catching and fixing the syntax error. Your other suggestion is great too. Approved and merging.

PR #742, opened 2024-05-02
Add support for docker containers restart in loki.source.docker

### LINKED ISSUE #278: Logs are not forwarded to Loki after target docker container is restarted
### What's wrong?

After restarting one of containers from which logs are scrapped - new logs will not be processed.
Grafana agent is installed via system package on Amazon Linux.
Reload of service configuration doesn't help.

https://github.com/grafana/loki/issues/5259 might be relevant

### Steps to reproduce

1. Restart one of containers with logs - "docker restart container-name"
2. Observer logs are created - "docker logs container-name"
3. Check Grafana loki source - no new logs after container is stopped
4. Restart agent - sudo systemctl restart grafana-agent-flow.service
5. Check Grafana loki source - new logs are pushed

### System information

Linux ip-10-10-101-40.ec2.internal 6.1.66-93.164.amzn2023.x86_64 grafana/agent#1 SMP PREEMPT_DYNAMIC Tue Jan  2 23:50:53 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Software version

agent, version v0.39.0 (branch: HEAD, revision: 402672cb)

### Configuration

```text
logging {
        level = "warn"
}

local.file "api_key" {
        filename = "/api-key"
        is_secret = true
}

}

discovery.relabel "default" {
  targets = discovery.docker.containers.targets

  rule {
    action = "labelmap"
    regex = "__meta_docker_container_label_([^_]+)"
  }
  rule {
    action = "replace"
    source_labels = ["__meta_docker_container_name"]
    target_label = "container"
  }
}

loki.source.docker "default" {
  targets    = discovery.relabel.default.output
  forward_to = [loki.write.local.receiver]
}

loki.write "local" {
  endpoint {
    url = "***"
    basic_auth {
        username = ***
        password = local.file.api_key.content
    }
  }
}
```

### Logs

_No response_

wildum:
#### PR Description

The loki.source.docker does not attempt to re-collect logs from a docker container target that has exited. The only way to continue collecting the logs from the target is to remove the target from the component and add it again.

This means that if a target is terminated and restarted within a discovery.docker interval, it will not be removed from the loki.source.docker component and the logs won't be collected anymore.

This fix adds a retry strategy that attempts to collect logs from the current running targets every 5 seconds. If the task is already running it does not do anything. If a task was stopped it will be restarted.

Fixes #278 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -305,10 +305,10 @@ func (c *Component) getManagerOptions(args Arguments) (*options, error) {
 	}
 
 	return &options{
-		client:              client,
-		handler:             loki.NewEntryHandler(c.handler.Chan(), func() {}),
-		positions:           c.posFile,
-		taskRestartInterval: 5 * time.Second,
+		client:                client,
+		handler:               loki.NewEntryHandler(c.handler.Chan(), func() {}),
+		positions:             c.posFile,
+		targetRestartInterval: 5 * time.Second,
 	}, nil
 }
 
--- a/internal/component/loki/source/docker/runner.go
+++ b/internal/component/loki/source/docker/runner.go
@@ -53,8 +53,8 @@ type options struct {
 	// positions interface so tailers can save/restore offsets in log files.
 	positions positions.Positions
 
-	// taskRestartInterval to restart task that has stopped running.
-	taskRestartInterval time.Duration
+	// targetRestartInterval to restart task that has stopped running.
+	targetRestartInterval time.Duration
 }
 
 // tailerTask is the payload used to create tailers. It implements runner.Task.
@@ -98,7 +98,7 @@ func newTailer(l log.Logger, task *tailerTask) *tailer {
 }
 
 func (t *tailer) Run(ctx context.Context) {
-	ticker := time.NewTicker(t.opts.taskRestartInterval)
+	ticker := time.NewTicker(t.opts.targetRestartInterval)
 	tickerC := ticker.C
 
 	for {
--- a/internal/component/loki/source/docker/docker_test.go
+++ b/internal/component/loki/source/docker/docker_test.go
@@ -28,7 +28,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const taskRestartInterval = 20 * time.Millisecond
+const targetRestartInterval = 20 * time.Millisecond
 
 func Test(t *testing.T) {
 	// Use host that works on all platforms (including Windows).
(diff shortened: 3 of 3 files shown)


internal/component/loki/source/docker/docker.go
```diff
@@ -305,9 +305,10 @@ func (c *Component) getManagerOptions(args Arguments) (*options, error) {
 	}
 
 	return &options{
-		client:    client,
-		handler:   loki.NewEntryHandler(c.handler.Chan(), func() {}),
-		positions: c.posFile,
+		client:              client,
+		handler:             loki.NewEntryHandler(c.handler.Chan(), func() {}),
+		positions:           c.posFile,
+		taskRestartInterval: 5 * time.Second,
```
wildum, line 311: hardcoded to 5s because I don't think that it's an interesting param to expose to the user

internal/component/loki/source/docker/internal/dockertarget/target.go
```diff
@@ -230,8 +230,6 @@ func (t *Target) StartIfNotRunning() {
 		ctx, cancel := context.WithCancel(context.Background())
 		t.cancel = cancel
 		go t.processLoop(ctx)
-	} else {
-		level.Debug(t.logger).Log("msg", "attempted to start process loop but it's already running", "container", t.containerName)
```
thampiotr, line 234: Why remove this debug log line? We could just change log message to sth like "starting process loop - already running".

internal/component/loki/source/docker/runner.go
```diff
@@ -95,23 +98,24 @@ func newTailer(l log.Logger, task *tailerTask) *tailer {
 }
 
 func (t *tailer) Run(ctx context.Context) {
-	ch, chErr := t.opts.client.ContainerWait(ctx, t.target.Name(), container.WaitConditionNextExit)
```
thampiotr, line 98: Previously, the Run will exit when ctx is cancelled or when there's an error from ContainerWait. When exiting, it calls target.Stop. 

Now, we don't seem to call target.Stop, only ticker.Stop. I wonder if this can lead to a resource leak when you update targets. Could you verify this?

thampiotr (commented):
Got one concern about the resource leak, but it looks okay otherwise!

internal/component/loki/source/docker/internal/dockertarget/target.go
wildum: with the new retry logic, the line is less relevant and more noisy (currently logs every 5 seconds/target)

thampiotr: it's on debug level, so I think the noise wouldn't be too much of a concern, but if it prints every time with this timer than yeah, fine with removing it 👍

internal/component/loki/source/docker/runner.go
wildum: Thanks for catching this one, I forgot it. I believe that it would leak in some very specific edge cases where the container is still running and producing logs but it got removed from the targets. Calling stop makes sense

internal/component/loki/source/docker/runner.go
```diff
@@ -52,6 +52,9 @@ type options struct {
 
 	// positions interface so tailers can save/restore offsets in log files.
 	positions positions.Positions
+
+	// taskRestartInterval to restart task that has stopped running.
+	taskRestartInterval time.Duration
```
thampiotr, line 57: Sorry, missed this last time. This causes a periodical call to `t.target.StartIfNotRunning()`, so I think it should be called `targetRestartInterval` instead.

thampiotr (approved):
One comment, otherwise LGTM

PR #747, opened 2024-05-02
packaging: change default Nice value on Linux installs to -5

rfratto:
This changes the default Nice value on Linux installs (when our DEB or RPM installers are used) from 0 to -5. This gives Alloy a higher priority so monitoring can be prioritzed when there's CPU pressure elsewhere on a machine.

I tested this locally and it works and sets the priority properly, though I don't know if -5 is sufficient. We can keep tuning this over time as needed. (The supported range of nice values is -20 to 19).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ Main (unreleased)
   every 15 seconds instead of as soon as data was written to the WAL.
   (@rfratto)
 
-- Imported code using `slog` logging will now not panic and replay correctly when logged before the logging 
+- Imported code using `slog` logging will now not panic and replay correctly when logged before the logging
   config block is initialized. (@mattdurham)
 
 ### Other changes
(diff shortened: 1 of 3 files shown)


mattdurham:
Does this mean alloy is likely prioritized over the production workload?

bboreham:
I think @mattdurham's point is that some people would rather their production workload took precedence.
And I agree those people exist. While some people would prefer if they can observe their system even under heavy load.

Personally I am happy to make Alloy higher priority by default, even better if the docs show how to set it the other way.

I note on Kubernetes we make `requests` which, while not setting Alloy above any other workloads, should ensure that there is some CPU for it to fight over.

mattdurham:
I am not in love with this because it likely prioritizes Alloy over production workloads. If I was the owner of the production workload and Alloy was consuming resources that then caused my workload to start to fail, I would not be happy. Do we have any issues related to this?

rfratto:
I'm going to close this for the time being since we don't have consensus.

PR #748, opened 2024-05-02
Update remotecfg dep to use `alloy-remote-config` over `agent-remote-config`

### LINKED ISSUE #783: Update remotecfg dep to use alloy-remote-config over agent-remote-config
### Request

title

### Use case

title

erikbaranowski:
#### PR Description

Closes #783

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,11 @@ Main (unreleased)
 - Imported code using `slog` logging will now not panic and replay correctly when logged before the logging 
   config block is initialized. (@mattdurham)
 
+### Breaking changes
+
+- Update Public preview `remotecfg` to use `alloy-remote-config` instead of `agent-remote-config`. The
+  API has been updated to use the term `collector` over `agent`. (@erikbaranowski)
+
 ### Other changes
 
 - Update `alloy-mixin` to use more specific alert group names (for example,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
-### Breaking changes
+### Breaking changes to non-GA functionality
 
 - Update Public preview `remotecfg` to use `alloy-remote-config` instead of `agent-remote-config`. The
   API has been updated to use the term `collector` over `agent`. (@erikbaranowski)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
+
+- Update Public preview `remotecfg` to use `alloy-remote-config` instead of `agent-remote-config`. The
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
```
rfratto, line 13: @clayton-cornell Should we find a different way of classifying these kinds of changes for non-stable functionality now that we're post-1.0?

go.mod
```diff
@@ -257,6 +256,7 @@ require (
 require (
 	github.com/KimMachineGun/automemlimit v0.6.0
 	github.com/Shopify/sarama v1.38.1
+	github.com/grafana/alloy-remote-config v0.0.3
 	github.com/grafana/kafka_exporter v0.0.0-20240409084445-5e3488ad9f9a
 )
```
rfratto, line 261: I'm not a fan of how Go tooling keeps adding so many require blocks. Can you take a second to merge the require blocks again? We should only have two: one for direct dependencies and one for indirect dependencies.

rfratto (commented):
Generally OK with these changes. I'm a little worried that it might be an abrupt change even if we can technically make these kinds of changes for non-stable functionality. Do we have a pulse on how many people this might impact, and if we need to consider a smoother migration?

CHANGELOG.md
erikbaranowski: maybe append to this header? `Breaking changes (Public preview features)`

go.mod
erikbaranowski: I cleaned up so indirect deps are grouped as well.

CHANGELOG.md
rfratto: cc @richardqlam on this too

erikbaranowski:
@rfratto As far as usage we don't have a great pulse other than a couple questions about `remotecfg` in community slack we have received and looking at the number of watchers/stars in the following repos. We have combined 1 star and 5 watchers each right now so I do feel this is not impacting a lot of folks.

- https://github.com/grafana/agent-remote-config
- https://github.com/grafana/alloy-remote-config

CHANGELOG.md
richardqlam: Hm I'm not sure further sub-categorizing breaking changes at a top level is all that useful. Feels like a changelog is most consumable when it talks about things on a feature-by-feature level. And then within the context of each feature highlighting what state it is (GA, public preview, etc).

rfratto (approved):
LGTM

CHANGELOG.md
clayton-cornell: Just looking at this now. It's still a breaking change, even though it's in a public preview state. The info clearly says it's an update to the public preview.... to my way of thinking, it's clear enough as-is.

PR #725, opened 2024-04-30
Reintroduce boringcrypto images

rfratto:
This PR introduces boringcrypto/cngcrypto Docker images to mirror boringcrypto/cngcrypto support in Grafana Agent.

Related to #64.

I'll follow up with another PR for distributing boringcrypto binaries after this is merged.

PR #753, opened 2024-05-03
Dockerfile.windows: fix issue where Alloy build was not produced

rfratto:
PR #725 reintroduced boringcrypto support and mistakenly switched the quote syntax in the Dockerfile used to pass variables. This change was not necessary as the surrounding string already used double quotes, and making the change prevented the proper command from being passed to make.


mattdurham (approved):
Qoutes!

PR #768, opened 2024-05-06
fix(pyroscope.ebpf): detect libc version , handle short buildid properly

korniltsev:
#### PR Description
Update pyroscope/ebpf dep from 0.4.6 to 0.4.7 to include fixes for two issues https://github.com/grafana/pyroscope/pull/3268
1. detect libc version properly when libc file name is libc-2.31.so and not libc.so.6
2. treat elf files with short build id (8 bytes) properly

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Bugfixes
+
+- Upgrading `pyroscope/ebpf` from 0.4.6 to 0.4.7 (@korniltsev):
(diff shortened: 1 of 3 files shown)


wildum:
hey, thanks for doing the update :) Could you make an entry in the changelog in a new "### Other changes" part under "Main unreleased" please?
Something like: Upgrading pyroscope/ebpf from 0.4.6 to 0.4.7: 
- detect libc version properly when libc file name is libc-2.31.so and not libc.so.6
- treat elf files with short build id (8 bytes) properly

wildum (approved):

PR #769, opened 2024-05-06
[Controller]: check node refs before stdlib refs when nodes are wired together

wildum:
#### PR Description

For forwards compatibility, we established that declares and imports should shadow other things. The wiring of the nodes in the loader does not follow this statement because references for the stdlib are checked before node references.

This PR changes the order so that custom components can shadow the stdlib. This will allow us to extend the stdlib without breaking our users configs.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,15 +10,6 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
-### Bugfixes
-
-- Fix a bug where custom components would not shadow the stdlib. (@wildum)
-
-{{< admonition type="warning" >}}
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Bugfixes
+
+- Fix a bug where custom components would not shadow the stdlib. (@wildum)
```
rfratto: Since we haven't release 1.1.0-rc.0 yet (due to issues), can you put this into the 1.1.0-rc.0 section and I'll cherry-pick it over?

PR #774, opened 2024-05-06
Dockerfile: make it clear that non-root support is experimental

rfratto:
Users looking at the Dockerfile may not understand that non-root support is both undocumented and experimental. This commit makes that status explicit with a comment.


mattdurham (approved):
LGTM

PR #778, opened 2024-05-06
Fix EOF issue when building on drone in windows.

mattdurham:
#### PR Description

When using drone on windows the code must pass through a few layers til it gets to the linux build container, the EOF command is not compatible with windows and will cause issues. In my case it either only partially executed the command or ignored it. That then led to corepack not found since it never installed nodejs.

By using separate RUN lines the behavior is the same when calling drone on linux and windows even when they are both using a linux docker container under the hood.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,9 +39,9 @@ ARG USERNAME="alloy"
 LABEL org.opencontainers.image.source="https://github.com/grafana/alloy"
 
 # Install dependencies needed at runtime.
-RUN  apt-get update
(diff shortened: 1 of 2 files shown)


rfratto:
I'm worried about increased number of layers this adds which will make the build image even bigger than normal; each RUN creates a new layer. Can this be combined into a single run command using && to separate them?

mattdurham:
Sure, changed to && style

rfratto (approved):

PR #789, opened 2024-05-08
Modules: add relative path support to import.file

### LINKED ISSUE #786: Modules: support relative path
### Background

The import.file block imports custom components from a file or a directory and exposes them to the importer. Users can specify the path to watch via the argument filename.

### Problem

The path is relative to the current working directory. This works fine when importing a module from the root config but it’s not ideal when the module itself has import.file blocks. Resolving the path relative to the module would be better in this case.

### Request

The following use-cases should be supported:

- import.file -> import.file
- import.git -> import.file
- import.string -> import.file

And if possible import.http -> import.file, but this one might be more tricky because it would only be possible if the nested import.file blocks would be turned into import.http blocks.

wildum:
#### PR Description

This PR introduces an ArgScope to the loader that will be used as a parent scope when evaluating the nodes in the module. For now, it contains only one variable "module_path" that represents the current path of the module and can be used in the config. It also adds the stdlib function "path_join". Both can be used together to configure a relative path on the import.file block.

```
import.file “default” {
  filename = path_join(module_path, "relative/path")
}
```

The following use-cases are currently supported:
- import.file in a module imported via import.file (module_path = filename of the parent import.file)
- import.file in a module imported via import.git (module_path = path to the local cloned repo)
- import.file in a module imported via import.string (module_path = path of the current module)

Fixes #786 

#### Notes to the Reviewer

The PR can be reviewed commit by commit.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated


wildum:
I revived this PR in https://github.com/grafana/alloy/pull/1726

PR #797, opened 2024-05-08
Live debugging service with prometheus.relabel

### LINKED ISSUE #790: X-ray: Create the x-ray service and connect the prometheus.relabel component to it

wildum:
**THIS PR WILL BE MERGED TO A FEATURE BRANCH WHICH WILL RECEIVE A FEW OTHER PRs BEFORE BEING READY TO BE MERGED ON MAIN**

This PR is the first part of the live debugging functionality.

It creates live debugging as a service and connects the prometheus.relabel component to it (an easy first candidate).

A stream to a component can be opened via `http://localhost:12345/api/v0/web/debug/ + componentID`.
This adds a callback to the x-ray service that the component can retrieve and use to pass data as a string which will be directly flushed to the HTTP client.

It supports components that are inside modules.

Fixes #790


internal/service/livedebugging/service.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/grafana/alloy/internal/featuregate"
+	"github.com/grafana/alloy/internal/service"
+)
+
+// ServiceName defines the name used for the livedebugging service.
+const ServiceName = "livedebugging"
+
+type Service struct {
+	debugStreamManager DebugStreamHandler
```
wildum, line 16: change it to debugStreamHandler

internal/service/livedebugging/debug_stream_handler.go
```diff
@@ -0,0 +1,11 @@
+package livedebugging
+
+// DebugStreamHandler defines the operations for managing debug streams.
+type DebugStreamHandler interface {
```
wildum, line 4: rename to DebugStreamManager

internal/web/api/api.go
```diff
@@ -107,3 +113,70 @@ func (a *AlloyAPI) getClusteringPeersHandler() http.HandlerFunc {
...
+				}
+				// Avoid blocking the channel when the channel is full
+				select {
+				case dataCh <- data:
+				default:
+				}
+			}
+		})
+
+		stopStreaming := func() {
```
wildum, line 144: defer

mattdurham:
Looks reasonable, tested both wide 1,000 streams and deep 1 stream with 1m metrics. Long term some changes around when to flush and the channel would be nice. But performance is totally reasonable.

internal/service/livedebugging/debug_stream_manager.go
```diff
@@ -0,0 +1,65 @@
+package livedebugging
+
+import "sync"
+
+// DebugStreamManager defines the operations for managing debug streams.
+type DebugStreamManager interface {
```
ptodev, line 6: I wonder if we really need an interface. There is a Go practice that "interfaces should be discovered" and that we should "return structs and accept interfaces". The rationale is that creating an interface in the same place as the implementation could be premature, since you don't really know if client code needs all those functions. It might be better to let client code make their own interface, so that they could just use a subset of the struct's exported functions.

It's not necessary to change the code. It's not that important :) I just thought it's worth mentioning.

internal/component/prometheus/relabel/relabel.go
```diff
@@ -259,6 +272,9 @@ func (c *Component) relabel(val float64, lbls labels.Labels) labels.Labels {
 	// Set the cache size to the cache.len
 	// TODO(@mattdurham): Instead of setting this each time could collect on demand for better performance.
 	c.cacheSize.Set(float64(c.cache.Len()))
+
+	c.debuggingStreamHandler.Stream(c.opts.ID, fmt.Sprintf("%s => %s", lbls.String(), relabelled.String()))
```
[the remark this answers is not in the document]
wildum: Yeah it will, that's a good remark because before I had a check to see whether there was a stream or not and I removed it to reduce the verbosity. Maybe it's better to put the check back to avoid computing the string (the check just checks the map if a stream exists or not, it's very light)

internal/service/livedebugging/debug_stream_manager.go
wildum: Thanks for the tips, that's a good point. We could split it in two interfaces: DebugStreamManager (IsRegistered, SetStream, DeleteStream) and DebugStreamProducer (Register, Stream, IsStreamOpen). Both are implemented by the debug stream struct. The first one is accepted by the api.go and the second one is accepted by the components. What do you think?

internal/alloy/componenttest/componenttest.go
```diff
@@ -163,6 +164,8 @@ func (c *Controller) buildComponent(dataPath string, args component.Arguments) (
 			switch name {
 			case labelstore.ServiceName:
 				return labelstore.New(nil, prometheus.DefaultRegisterer), nil
+			case livedebugging.ServiceName:
+				return livedebugging.NewDebugStreamManager(), nil
```
thampiotr, line 168: Here we use just `New()` function: https://github.com/grafana/alloy/pull/797/files#diff-7e74a62873427a2bfee5c9d25f0e133d65d417d54910964d843e4a8f616825d1R276

Why is it NewDebugStreamManager here?

internal/alloycli/cmd_run.go
```diff
@@ -292,6 +296,7 @@ func (fr *alloyRun) Run(configPath string) error {
 		MinStability: fr.minStability,
 		Services: []service.Service{
 			httpService,
+			liveDebuggingService,
```
thampiotr, line 299: does the order of these services matter? if not, maybe we make it alphabetic?

internal/component/prometheus/relabel/relabel.go
```diff
@@ -99,14 +104,22 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	debuggingStreamHandler, err := o.GetServiceData(livedebugging.ServiceName)
+	if err != nil {
+		return nil, err
+	}
+	debuggingStreamHandler.(livedebugging.DebugStreamManager).Register(name)
```
thampiotr, line 112: Instead of creating `const name` can you access name from `opts`?

internal/component/prometheus/relabel/relabel.go
```diff
@@ -85,6 +88,8 @@ type Component struct {
 	exited           atomic.Bool
 	ls               labelstore.LabelStore
 
+	debuggingStreamHandler livedebugging.DebugStreamManager
```
thampiotr, line 91: Is it a "stream handler" or a "stream manager" or is it a "service"? 
I am definitely against adding too many concepts, especially when it's not needed. Maybe everything should be called service? 

Feel free to split it however you like in the `internal` structure of the `service/livedebugging` package, but the API exposed to Alloy components etc. should be minimalistic, consistent and properly named.

internal/component/prometheus/relabel/relabel.go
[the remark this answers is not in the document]
thampiotr: +1 `.String()` will be expensive and I've actually worked on a fix in the past that had this very issue when doing debug statements for the logger

We should make it lazy in some way. Logging libraries can be used as an inspiration on how this can be achieved.

thampiotr, line 276: Why is this function called `Stream`? 

To stream a piece of information seems to be a name suitable for sending a sequence of data in a way that parts of it are being sent at a time, often asynchronously. 

Here we are not sending a sequence of data and there is one, synchronous call. Maybe call it `Publish` or `Push` as it is adding to a potentially active stream.

internal/service/ui/ui.go
```diff
@@ -21,7 +22,8 @@ const ServiceName = "ui"
 // Options are used to configure the UI service. Options are constant for the
 // lifetime of the UI service.
 type Options struct {
-	UIPrefix string // Path prefix to host the UI at.
+	UIPrefix               string                           // Path prefix to host the UI at.
+	DebuggingStreamHandler livedebugging.DebugStreamManager // Debugging stream handler used for live debugging in the UI.
```
thampiotr, line 26: Is this a handler or manager?

internal/web/api/api.go
```diff
@@ -107,3 +115,83 @@ func (a *AlloyAPI) getClusteringPeersHandler() http.HandlerFunc {
 		_, _ = w.Write(bb)
 	}
 }
+
+func (a *AlloyAPI) startDebugStream() http.HandlerFunc {
```
thampiotr, line 119: This function is a bit too long to be readable IMO.

internal/service/livedebugging/debug_stream_manager.go
```diff
@@ -0,0 +1,65 @@
...
+	Stream(componentID string, data string)
+	// SetStream assigns a debug stream callback to a componentID.
+	SetStream(streamID string, componentID string, callback func(string))
+	// DeleteStream removes a debug stream.
+	DeleteStream(streamID string, componentID string)
+}
+
+type debugStreamManager struct {
+	loadMut              sync.RWMutex
+	streams              map[string]map[string]func(string)
```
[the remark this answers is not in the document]
thampiotr: +1 let's also make sure we document what the keys and values are meant to be where it's not obvious.

internal/service/livedebugging/debug_stream_manager.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (s *debugStreamManager) Stream(componentID string, data string) {
+	s.loadMut.RLock()
+	defer s.loadMut.RUnlock()
+	for _, stream := range s.streams[componentID] {
+		stream(data)
+	}
+}
+
+func (s *debugStreamManager) SetStream(streamID string, componentID string, callback func(string)) {
```
thampiotr, line 43: It's more like `AddStreamCallback` no?

internal/service/livedebugging/debug_stream_manager.go
```diff
@@ -0,0 +1,65 @@
...
+	s.loadMut.Lock()
+	defer s.loadMut.Unlock()
+	if _, ok := s.streams[componentID]; !ok {
+		s.streams[componentID] = make(map[string]func(string))
+	}
+	s.streams[componentID][streamID] = callback
```
thampiotr, line 49: What is a stream? Can we define it and make sure we use this naming throughbout the code to make it easier to read?

Is a stream the callback? Or is the stream the data sent from a component?

thampiotr (commented):
Good start! some comments around readability and naming.

internal/component/prometheus/relabel/relabel.go
wildum: opts does not have just the name, it has the ID which is component name + label

internal/alloy/componenttest/componenttest.go
wildum: Because when you call Data() on the labelstore service it returns a pointer to itself:
func (s *service) Data() any {
	return s
}
whereas the livedebugging service returns only the debugStreamManager

internal/service/livedebugging/livedebugging.go
```diff
@@ -0,0 +1,82 @@
...
+	// The callback can be used by the component to send debugging data to live debugging consumers.
+	AddCallback(callbackID CallbackID, componentID ComponentID, callback func(string))
+	// DeleteCallback deletes a callback for a given componentID.
+	DeleteCallback(callbackID CallbackID, componentID ComponentID)
+}
+
+// DebugDataPublisher is an interface that components can use to push information to live debugging consumers.
+type DebugDataPublisher interface {
+	// Register a component by name.
+	Register(componentName ComponentName)
```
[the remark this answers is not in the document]
wildum: I did this ⬆️

ptodev (approved):
Thank you, LGTM!

internal/web/api/api.go
```diff
@@ -107,3 +115,83 @@ func (a *AlloyAPI) getClusteringPeersHandler() http.HandlerFunc {
...
+			return
+		}
+
+		if !a.debugCallbackManager.IsRegistered(livedebugging.ComponentName(component.ComponentName)) {
+			http.Error(w, fmt.Sprintf("Live debugging is not supported for the component \"%s\"", component.ComponentName), http.StatusInternalServerError)
+			return
+		}
+
+		// Buffer of 1000 entries to handle load spikes and prevent this functionality from eating up too much memory.
+		// TODO: in the future we may want to make this value configurable to handle heavy load
```
thampiotr, line 137: Perhaps we should track these TODOs somewhere in a GitHub issue, cause after we're done reviewing these partial PRs, we may miss them when merging to main.

thampiotr (approved):

internal/web/api/api.go
wildum: Good point, I created this issue to track the todos: https://github.com/grafana/alloy/issues/974
Not sure if we do the optimizations before we merge it to main. It might be better to have more components that supports the feature to identify and solve the pain points

PR #801, opened 2024-05-08
(Experiment) Dockerfile.windows: remove clean cache step

rfratto:
PR grafana/agent#6671 split out cleaning the Go cache into its own step. This increased build times as stages are slow to produce on Windows.

As Dockerfile.windows is a multi-stage Dockerfile, and the cache doesn't carry over to the final image, cleaning the Go cache can be safely skipped to reduce the overall build time.


rfratto:
The theory was wrong here, and it in fact takes longer than before. I'm not quite sure why this is (maybe something to do with the final size of the intermediate container stage), but in any case this isn't measurably improving anything so I'll close.

PR #812, opened 2024-05-09
add otelcol.receiver.file_stats component

### LINKED ISSUE #238: Flow: add otelcol.receiver.file_stats
### Request

Add a new Flow component, `otelcol.receiver.file_stats`, which wraps around the upstream OpenTelemetry Collector [filestatsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filestatsreceiver/documentation.md) component.

Note that the proposed Flow component name has an underscore separating the two words (file and stats) to align with Flow component naming conventions. Documentation can help make it clear that the upstream component is named without the underscore.

> **NOTE**: Now that OpenTelemetry Collector conversion is supported, this component should be introduced alongside converter code similar to grafana/agent#6463.

### Use case

Support more commonly used components from `opentelemetry-collector-contrib`.

### LINKED ISSUE #369: prometheus.remote_write: add an upper limit for the number of series kept in-memory
Currently, `prometheus.remote_write` will accept new series regardless of how many series are currently tracked in memory. Given enough new series in between WAL truncations, this can cause agents to run out of memory. 

It would be nice to have a mechanism in prometheus.remote_write to limit the amount of active series to prevent agents from getting overwhelmed, at the cost of not being able to accept anything new until the next WAL truncation. 

When clustering is enabled, this could potentially be paired with a forced truncate whenever the cluster state changes to free up memory sooner and allow new series to be stored.

rfratto:
#### PR Description
Add an otelcol.receiver.file_stats component which wraps around the upstream filestatsreceiver component from the -contrib distribution.

Closes #238.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol.receiver.file_stats.md
```diff
@@ -0,0 +1,269 @@
...
+## Arguments
+
+`otelcol.receiver.file_stats` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`include` | `string` | Glob path for files to collect stats from. | | yes
+`collection_interval` | `duration` | How often to collect file information. | `"1m"` | no
+`initial_delay` | `duration` | Initial time to wait before collecting file information. | `"1s"` | no
+`timeout` | `duration` | Timeout for collecting file information; `0s` means no timeout. | `"0s"` | no
```
mattdurham, line 39: Is this per file or for a whole update task?

rfratto: Good point, I'll check and update the docs

rfratto: It's the timeout for the entire poll of all files matched by `include`

docs/sources/reference/components/otelcol.receiver.file_stats.md
```diff
@@ -0,0 +1,269 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol.receiver.file_stats/
+title: otelcol.receiver.file_stats
+description: Learn about otelcol.receiver.file_stats
+---
+
+# otelcol.receiver.file_stats
+
+`otelcol.receiver.file_stats` collects metrics from files specified with a glob pattern.
```
mattdurham, line 9: Does this only do files? Several of the metrics mention folders

rfratto: it does both, I'll make sure I mention files and folders more consistently

mattdurham (approved):
LGTM, I hate that the config is internal and we have to use creative ways around it.

PR #819, opened 2024-05-10
WIP: Magefile POC

mattdurham:
This is 70% of our makefile implemented as Magefile, there are three binaries that allow you to run different targets. Two for osx and one for linux amd64. More can be added. Wanted to get thoughts before diving in deeper. I havent implemented the build cache fully so that is a todo. Tested that I can build all the binaries, boring crypto and images. Ideally the makefile becomes one command, the one to build the magefiles.

Its also debuggable! There is still some code to clean up but thought I would get some input before diving back in.


Dockerfile
```diff
@@ -17,18 +17,8 @@ ARG GOEXPERIMENT
...
-   make generate-ui
 
-RUN --mount=type=cache,target=/root/.cache/go-build \
-    --mount=type=cache,target=/go/pkg/mod \
-    GOOS=$TARGETOS GOARCH=$TARGETARCH GOARM=${TARGETVARIANT#v} \
-    RELEASE_BUILD=${RELEASE_BUILD} VERSION=${VERSION} \
-    GO_TAGS="netgo builtinassets promtail_journal_enabled" \
-    GOEXPERIMENT=${GOEXPERIMENT} \
-    make alloy
+RUN ./build-linux-amd64 alloy
```
mattdurham, line 21: Need to add this back in once I tighten everything up

PR #820, opened 2024-05-10
Allow to override the cluster headless service port name for Istio support

### LINKED ISSUE #800: Alloy Clustering mode is currently incompatible with running inside an Istio mesh when deployed via the Helm Chart
### What's wrong?

Currently this Alloy chart is falling into this trap when run with Istio as a sidecar - https://istio.io/latest/docs/ops/common-problems/network-issues/#503-error-while-accessing-headless-services

The fix is relatively simple, either; 

- Allow the end user to specify the port name used instead of hardcoding it to `http`. This will allow istio users to fallback to the TCP support which would work for this traffic.
- Alternatively, send a host header containing the hostname which was looked up on the request to join a cluster

### Steps to reproduce

Install Alloy via the helm chart on a Kubernetes cluster running Istio in strict mode and with allow_any traffic profile enabled. Have Alloy run with an istio sidecar. Attempt to turn on clustering mode. See that no nodes are able to communicate with each other.

### System information

_No response_

### Software version

Alloy v1.0.0

### Configuration

_No response_

### Logs

_No response_

josh-thisisbud:
This will be needed for an Istio mesh based installation due to this issue with headless services + Istio - https://istio.io/latest/docs/ops/common-problems/network-issues/#503-error-while-accessing-headless-services

Fixes #800

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/cluster_service.yaml
+++ b/operations/helm/charts/alloy/templates/cluster_service.yaml
@@ -18,7 +18,7 @@ spec:
     #
     # This service should only be used for clustering, and not metric
     # collection.
-    - name: http
+    - name: {{ coalesce $values.clustering.portName "http" }}
       port: {{ $values.listenPort }}
       targetPort: {{ $values.listenPort }}
       protocol: "TCP"


josh-thisisbud:
What needs to be done to get this over the line? As it's a small change in the Helm Chart.
Are you expecting me to backport this to Grafana agent before this can be merged?

josh-thisisbud:
@rfratto as you added the label, would you answer the above question for me?

rfratto:
> What needs to be done to get this over the line? As it's a small change in the Helm Chart.

Thanks for the ping, and sorry this has been sitting around. I'll review this right after sending this comment.

>  Are you expecting me to backport this to Grafana agent before this can be merged?

No, sorry; that label is for the maintainers to track what they need to backport after a PR has been merged into main. I wouldn't want contributors to have to go through the effort of backporting.

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ Main (unreleased)
   This port can be configured with `--server.http.listen-addr` or using
   the default listen address`127.0.0.1:12345`. (@mattdurham) 
 
+- Fix clustering on instances running within Istio mesh by allowing to change the name of the clustering port
```
rfratto, line 32: We don't put change entries for the Helm chart here since this tracks changes for Alloy releases. Can you move this entry to the [Helm-specific CHANGELOG](https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/CHANGELOG.md) instead?

operations/helm/charts/alloy/values.yaml
```diff
@@ -41,6 +41,9 @@ alloy:
     # -- Deploy Alloy in a cluster to allow for load distribution.
     enabled: false
 
+    # -- Name for the port used for clustering, useful if running inside an Istio Mesh
+    portName: null
```
rfratto, line 45: I generally prefer the values.yaml exposing as much information as possible about what default values are; right now you'd have to dig into the templates to see that if `portName` is null, then the port name is `http`. 

Instead of using null/coalesce for the portName, can we set this to `portName: http` instead? That way the values.yaml are a little more self-documenting.

rfratto (commented):
Thanks! Two small comments but this should be ready to go soon.

operations/helm/charts/alloy/values.yaml
josh-thisisbud: done

CHANGELOG.md
josh-thisisbud: Moved

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ Main (unreleased)
   This port can be configured with `--server.http.listen-addr` or using
   the default listen address`127.0.0.1:12345`. (@mattdurham) 
 
+v1.1.0-rc.0
+-----------
+
```
rfratto, line 34: Looks like this snuck in somehow

josh-thisisbud: fixed

rfratto (approved):
LGTM, thanks for working on this!

PR #821, opened 2024-05-10
Add path_join function to the stdlib

wildum:
#### PR Description

Adding this function to the stdlib is needed to move forward with the relative path solution that is implemented for the modules. This function should also be useful in other scenarios.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Features
+
+- Add the function `path_join` to the stdlib. (@wildum)
+
 v1.1.0-rc.0
 -----------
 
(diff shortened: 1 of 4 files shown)


rfratto (commented):
I'm OK with this implementation but I'm planning on opening a proposal for a fleshed out set of stdlib functions to add. 

I'd like for us to hold off on this until that proposal is available and we have consensus on it, mainly because it may establish naming conventions that we'd want to follow, or other details about new functions that this doesn't account for (such as the stability level of functions).

If you think we need this sooner, I'd recommend that it at least not be GA, and for us to find a way to return an error if it's used in non-GA Alloy instances.

wildum:
I'm OK to wait. I'd like the relative path for modules to make it to 1.2 but since that we are in the process of releasing 1.1 we have time. Thanks for the heads-up, changing this PR to draft for now

wildum:
@rfratto any progress on the stdlib proposal?

wildum:
closing this one to open a more recent one following the rework of the stdlib

PR #824, opened 2024-05-10
Add Windows CNGCrypto AMD64 to dist and release assets.

mattdurham:
#### PR Description

This adds windows executable for cngcrypto amd64 support.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated via drone step

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -328,7 +328,7 @@ platform:
   os: linux
 steps:
 - commands:
-  - make generate-ui
+  - make generate-ui generate-winmanifest
   - GO_TAGS="builtinassets" GOOS=windows GOARCH=amd64 GOARM= make alloy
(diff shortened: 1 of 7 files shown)


erikbaranowski:
@rfratto I agree with needing a plan on how to improve the sanity of our CI and builds. I had hoped to tackle a piece of this with the windows MSI work, but we probably need to lay out a logical plan of what things can be improved in order to get us where we want to end up.

mattdurham:
Its totally fine to wait, I wanted it to be ready so it def makes 1.2.

mattdurham:
More universally we could switch to almost one build image but it would be the Microsoft one and that concerns me a bit. It might be an unneeded concern since they are very prompt on updated it due to CVEs or releases.

PR #835, opened 2024-05-13
Add prometheus native histogram support to otel.receiver.prometheus

### LINKED ISSUE #483: otelcol.receiver.prometheus: add support for native Prometheus histograms
Add support for Prometheus native histograms (also called exponential histograms, sparse histograms) into `otelcol.receiver.prometheus`.

wildum:
#### PR Description

The support was added upstream with v0.99.0 via https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/13fca794d57a15e68b41b95ea8df64ff7564f04d

Fixes #483 

#### Notes to the Reviewer

First commit is a copy of the code from upstream except that we need to keep "textparse" pkg for now

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated
- [x] Use featuregate to support it only for public preview

### DIFF
--- a/internal/component/otelcol/receiver/prometheus/prometheus.go
+++ b/internal/component/otelcol/receiver/prometheus/prometheus.go
@@ -112,7 +112,7 @@ func (c *Component) Update(newConfig component.Arguments) error {
 		// When supported, this could be added as an arg.
 		trimMetricSuffixes = false
 
-		enableNativeHistograms = true
+		enableNativeHistograms = c.opts.MinStability.Permits(featuregate.StabilityPublicPreview)
 
 		gcInterval = 5 * time.Minute
 	)
--- a/internal/component/otelcol/receiver/prometheus/prometheus.go
+++ b/internal/component/otelcol/receiver/prometheus/prometheus.go
@@ -112,6 +112,8 @@ func (c *Component) Update(newConfig component.Arguments) error {
 		// When supported, this could be added as an arg.
 		trimMetricSuffixes = false
 
+		enableNativeHistograms = true
+
 		gcInterval = 5 * time.Minute
 	)
 
@@ -147,6 +149,7 @@ func (c *Component) Update(newConfig component.Arguments) error {
 		useStartTimeMetric,
 		startTimeMetricRegex,
 		useCreatedMetric,
+		enableNativeHistograms,
 		labels.Labels{},
 		trimMetricSuffixes,
 	)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Main (unreleased)
 - Update Public preview `remotecfg` to use `alloy-remote-config` instead of `agent-remote-config`. The
   API has been updated to use the term `collector` over `agent`. (@erikbaranowski)
 
+### Enhancements
+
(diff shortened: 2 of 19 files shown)


internal/component/otelcol/receiver/prometheus/prometheus.go
```diff
@@ -112,6 +112,8 @@ func (c *Component) Update(newConfig component.Arguments) error {
 		// When supported, this could be added as an arg.
 		trimMetricSuffixes = false
 
+		enableNativeHistograms = true
```
rfratto, line 115: Is this something configurable upstream? Do we need to consider making it configurable downstream (even if it's not configurable upstream)?

rfratto (approved):
LGTM, pending the question about whether this should/needs to be configurable. 

Also, can we add a comment somewhere in the component describing what SHA we copied the code from, just to track it?

internal/component/otelcol/receiver/prometheus/prometheus.go
wildum: Good point, it's behind a featuregate upstream so it's only activated if the feature flag is explicitly set.
We have a few options:

- a new argument with default set to false
- a new argument with default set to true
- use our featuregate to make it true for experimental? (not sure about this one)
- only set it to true

The thing with the argument options is that it sticks long-term for us. Changing the default value or removing it is considered a breaking change. Also I got bitten in the past by these types of arguments (lost some time because of the send_native_histogram default to false in [prometheus.remote_write](https://grafana.com/docs/alloy/latest/reference/components/prometheus.remote_write/#endpoint-block))

On the other hand, scrape_classic_histograms is set to false by default in [prometheus.scrape](https://grafana.com/docs/alloy/latest/reference/components/prometheus.scrape/). And according to the [otel doc](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md), "In case a metric has both the conventional (aka classic) buckets and also native histogram buckets, only the native histogram buckets will be taken into account to create the corresponding exponential histogram. To scrape the classic buckets instead use the [scrape option](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) scrape_classic_histograms."

So if we enable it by default, users that don't have scrape_classic_histograms set to true and are scraping histograms that are both classic and expo (don't know how common this is) will have a change of behavior. Their classic metrics will be replaced by the native one (I tested it).

What's your feeling on this?

wildum: Sounds good to me. There is one big difference though: In Otel you need to explicitly activate each featuregate that you want whereas in our case if you have public-preview you get the change automatically.
But that's ok with our definition: "Public preview components are subject to breaking changes, and may be replaced with equivalent functionality that covers the same use case."

rfratto: Yeah, I'm ok with us not having explicitly named feature gates for now and just using the stability.level flag as a catch-all 👍

wildum: I made the feature gate change and added the commit sha to the doc

rfratto (approved):
LGTM part two :)

PR #836, opened 2024-05-13
Update Loki and sync some of the Promtail code

ptodev:
Two bug fixes were made in Promtail recently:

1. [Clean up metrics generated from logs after a config reload](https://github.com/grafana/loki/pull/11882)
2. [Fix bug with Promtail config reloading getting stuck indefinitely](https://github.com/grafana/loki/pull/12795)

I think both of these fixes only really impact static mode, but I'd prefer if I update Alloy first and backport to Agent, so that Alloy is not more out of date.

For the first issue I just tested Flow mode and Alloy, and it seems like the bug is still there for them and I need to fix it separately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,7 +64,7 @@ require (
 	github.com/grafana/jfr-parser/pprof v0.0.0-20240126072739-986e71dc0361
 	github.com/grafana/jsonparser v0.0.0-20240209175146-098958973a2d
 	github.com/grafana/kafka_exporter v0.0.0-20240409084445-5e3488ad9f9a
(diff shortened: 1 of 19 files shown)


mattdurham:
Are the Loki stages straight copies of the promtail changes?

thampiotr:
> Are the Loki stages straight copies of the promtail changes?

As far as I remember, they're not, there are some changes to the constructor etc, but they are functionally equivalent and relatively easy to map between each other.

thampiotr (approved):
LGTM, thanks!

PR #837, opened 2024-05-13
Add info about order of arguments when customising the Windows config

clayton-cornell:
#### PR Description

Add info about order of arguments when customising the Windows config.

If the storage.path argument is not last in the set of custom arguments, you will get the following error:

```
Error: failed to create the remotecfg service: mkdir C:\ProgramData\GrafanaLabs\Alloy\data --server.http.listen-addr=0.0.0.0:12345: The directory name is invalid.
```

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Hm, the argument _isn't_ positional though, so I think there's a different underlying issue here. What was the custom arguments passed that caused the issue here?

PR #839, opened 2024-05-14
Update cloudwatch exporter dependency

andriikushch:
#### PR Description

This PR adds a new configuration parameter for the Cloudwatch exporter. It allows to specify the AWS SDK version to use. `v1` is the default version. 

It was achieved by updating and integrating the version of the `github.com/nerdswords/yet-another-cloudwatch-exporter` package and all dependencies that it depends on.

According to the https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/a275f94e5d33b3b5330ecfced86c290e1db05928/pkg/config/feature_flags.go#L10 , using the AWS SDK v2, should improve performance.

After this change, using aliases for the AWS service names will produce warnings.

N.A.

A related change in the Agent project:

- https://github.com/grafana/agent/pull/6835

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell:
Doc changes look OK.

Should this be backported? Or only in `next`?

andriikushch:
> Doc changes look OK.
> 
> Should this be backported? Or only in `next`?

@clayton-cornell It would be nice to backport it. The same change is in the Agent repository: 

- https://github.com/grafana/agent/pull/6835.

clayton-cornell:
Added backports to Alloy v1.0 and v1.1

There's a parallel Agent PR already in process, so added the label `backport-to-agent:no`

CHANGELOG.md
```diff
@@ -19,6 +23,8 @@ Main (unreleased)
 
 - (_Public preview_) Add native histogram support to `otelcol.receiver.prometheus`. (@wildum)
 
+- A new parameter `aws_sdk_version` is added for the cloudwatch exporters configuration. It enables the use of aws sdk v2 which is expected to come with performance benefits. (@kgeckhart, @andriikushch)
```
kgeckhart, line 26:
```suggestion
- A new parameter `aws_sdk_version` is added for the cloudwatch exporters configuration. It enables the use of aws sdk v2 which has shown to have significant performance benefits. (@kgeckhart, @andriikushch)
```

internal/component/prometheus/exporter/cloudwatch/config.go
```diff
@@ -26,6 +28,7 @@ var defaults = Arguments{
 		Enabled:        false,
 		ScrapeInterval: 5 * time.Minute,
 	},
+	AWSSDKVersion: "v2",
```
kgeckhart, line 31: I think we should default to v1 for now to prevent breaking anyone on upgrade. We can make v2 the default following YACE changing.

andriikushch: Let's do it 👍

internal/component/prometheus/exporter/cloudwatch/config.go
```diff
@@ -37,6 +40,7 @@ type Arguments struct {
 	Discovery             []DiscoveryJob        `alloy:"discovery,block,optional"`
 	Static                []StaticJob           `alloy:"static,block,optional"`
 	DecoupledScrape       DecoupledScrapeConfig `alloy:"decoupled_scraping,block,optional"`
+	AWSSDKVersion         string                `alloy:"aws_sdk_version,attr,optional"`
```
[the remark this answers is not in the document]
andriikushch: Changed

docs/sources/reference/components/prometheus.exporter.cloudwatch.md
```diff
@@ -90,8 +90,8 @@ To use all of the integration features, use the following AWS IAM Policy:
 
 ```alloy
 prometheus.exporter.cloudwatch "queues" {
-    sts_region = "us-east-2"
-
+    sts_region      = "us-east-2"
+    aws_sdk_version_v2 = "v2"
```
kgeckhart, line 94:
```suggestion
    aws_sdk_version_v2 = "false"
```

kgeckhart (commented):
Did you have a chance to run it locally as well?

andriikushch:
> Did you have a chance to run it locally as well?

Yes. I created the config with 3 cloudwatch exporters: without the new parameter, with `aws_sdk_version_v2` set to `true` and to `false`. I used it to get `EC2` metrics. And compared `aws_ec2_cpuutilization_average` graphs. They look similar (I think the difference is depending on the exact time of the getting data):
Screenshot 2024-05-28 at 13 54 15
Screenshot 2024-05-28 at 13 54 27
Screenshot 2024-05-28 at 13 54 35

internal/static/integrations/cloudwatch_exporter/docs/doc.go
```diff
@@ -61,7 +61,7 @@ func generateServicesDocSection() string {
 	var sb strings.Builder
 	for _, supportedSvc := range yaceConf.SupportedServices {
 		sb.WriteString(
-			fmt.Sprintf("- Namespace: `%s` or Alias: `%s`\n", supportedSvc.Namespace, supportedSvc.Alias),
```
kgeckhart, line 64: You will need to run this to update the service list in the docs. ATM they all still have the Alias.

kgeckhart (approved):

.github/workflows/test.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         go-version: "1.22"
         cache: true
     - name: Test
-      run: make GO_TAGS="nodocker" test
+      run: CGO_LDFLAGS="-ld_classic $CGO_LDFLAGS" make GO_TAGS="nodocker" test
```
tpaschalis, line 26: Why is this change needed?

tpaschalis (dismissed):
Sorry to block this, doing it to avoid getting this accidentally merged (I've yet to review the entire PR).

As per our [backwards compatibility](https://github.com/grafana/alloy/blob/main/docs/sources/introduction/backward-compatibility.md) guarantees, we're no longer doing breaking changes in GA functionality.

Instead we'll need to find a way to support both the previous and current type values without affecting users' config files and queue any breaking changes for a v2.0 release in the future with the `v2.0-breaking-change` label

kgeckhart:
> Instead we'll need to find a way to support both the previous and current type values without affecting users' config files and queue any breaking changes for a v2.0 release in the future with the `v2.0-breaking-change` label

I think this particular change is covered under this exception to the backwards compatibility guarantee,

> Upstream changes: Much of the functionality of {{< param "PRODUCT_NAME" >}} is built on top of other software, such as OpenTelemetry Collector and Prometheus. If upstream software breaks compatibility, we may need to reflect this in {{< param "PRODUCT_NAME" >}}.

As the change is being driven by the upstream exporter in this case.

.github/workflows/test.yml
andriikushch: @tpaschalis, without `CGO_LDFLAGS="-ld_classic` build fails on mac.

tpaschalis:
But also, 

> We try, whenever possible, to resolve these issues without breaking compatibility.

It's always a tradeoff, but I think it's a good opportunity to try and fix this without affecting users. On a surface level it doesn't feel like a compatibility layer to support aliasing these types wouldn't be so intrusive. WDYT?

kgeckhart:
> But also,
> 
> > We try, whenever possible, to resolve these issues without breaking compatibility.
> 
> It's always a tradeoff, but I think it's a good opportunity to try and fix this without affecting users. On a surface level it doesn't feel like a compatibility layer to support aliasing these types wouldn't be so intrusive. WDYT?

I agree in principal, but IMO @cristiangreco  intentionally dropped these aliases upstream. They do not have any AWS specific meaning them that makes them valuable for a customer and how they were used was harmful https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1131.

cristiangreco:
> > But also,
> > > We try, whenever possible, to resolve these issues without breaking compatibility.
> > 
> > 
> > It's always a tradeoff, but I think it's a good opportunity to try and fix this without affecting users. On a surface level it doesn't feel like a compatibility layer to support aliasing these types wouldn't be so intrusive. WDYT?
> 
> I agree in principal, but @cristiangreco intentionally dropped these aliases upstream which is why I'm in favor of dropping them here as well. They do not have any AWS specific meaning to them that makes them valuable for a customer and how they were used was harmful [nerdswords/yet-another-cloudwatch-exporter#1131](https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1131).

Yes the upstream change was intentional, so in theory the usage of aliases should be dropped in alloy as well.

andriikushch:
Hi @tpaschalis, 

@kgeckhart and I discussed some options for this case. We agreed that we could introduce a translation layer that will also produce warning logs when aliases are used and inform that support of the aliases will be dropped in the upcoming releases. 

This will give users enough time to update their configs and allow them to benefit from the new component version immediately. 

What do you think?

tpaschalis:
I don't see the benefit of removing the translation layer though. Do you anticipate there being it an effort to maintain?

On surface it sounds like a simple string-to-string mapping that helps maintain user confidence in our backwards compatibility guarantees.

In any case, other members of the @grafana/grafana-alloy-maintainers can chime in here with their opinion, let's see what they say

kgeckhart:
@andriikushch and I talked through it could we meet in the middle and follow a path as the upstream exporter? Add a translation layer, note that it's deprecated in the changelog, and log a warning when someone is using it so it's not a hard cutoff. Ideally, we could drop this before a 2.0 release (which I heard was next year?). 

The implementation is not going to be hard to do it's more the extra layer of indirection we have to consider when attempting to understand a customers configuration. The alias is not a real value to AWS while what we are forcing someone to use instead is the AWS CloudWatch namespace. Some aliases are easily relatable back to the CloudWatch namespace but for things like `gwlb` I would have to look at our mapping layer or google it to recognize it as `AWS/GatewayELB`.

andriikushch:
Hi :) 

I have added a translation layer with the latest commits. 

It is implemented as a separate function that produces warnings when an alias is used. I tried to add it so that it can be removed in the future without changing the rest of the code if we decide to drop support for aliases.

I've also reverted some changes in the unit tests to ensure that aliases work and redone some manual tests.

Please take a look, and I am looking forward to the feedback :)

.github/workflows/test.yml
thampiotr: Can you describe what this does and share a link to some docs where we can read more about this? I'm not familiar with this option.

CHANGELOG.md
thampiotr, line 1: These changelog entries need to be moved to the top section of unreleased changes.

docs/sources/reference/components/prometheus.exporter.cloudwatch.md
```diff
@@ -90,8 +90,8 @@ To use all of the integration features, use the following AWS IAM Policy:
 
 ```alloy
 prometheus.exporter.cloudwatch "queues" {
-    sts_region = "us-east-2"
-
+    sts_region      = "us-east-2"
+    aws_sdk_version_v2 = "false"
     discovery {
         type        = "sqs"
```
thampiotr, line 96: type=sqs is deprecated now - we need to search for all the examples everywhere and remove the usage of aliases

internal/static/integrations/cloudwatch_exporter/cloudwatch_exporter.go
thampiotr, line 1: Technically we don't need to change the code under `internal/static` cause it's only used for converters (last time I checked). But it's not a problem if we leave this.

.github/workflows/test.yml
andriikushch: Good question, @thampiotr 👍 

When this PR was created, the build on macOS returned linker errors. This flag tells CI to use a classic linker, which is supposed to be more compatible. It solved a problem back then. 

Now I see that Apple is deprecating `-ld_classic` https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#Deprecations-in-Xcode-16-Beta. 

I need to check if this is still required for our tests.

Some additional information about `ld_classic` can be found here:

- https://forums.developer.apple.com/forums/thread/715385
- https://www.unix.com/man-page/osx/1/ld_classic

docs/sources/reference/components/prometheus.exporter.cloudwatch.md
andriikushch: I will update the tests. 👍 

We kept it to test and show the backward compatibility. 

A few commits ago, based on the discussion above, we added a compatibility layer to keep support for the aliases. 

- https://github.com/grafana/alloy/pull/839/commits/63ff123d9169a9bbbd4b5ae50a63ddb67f2e7483

The idea was to generate warnings but keep the support. 

@thampiotr, shall we use this approach or drop the support completely?

CHANGELOG.md
andriikushch: Done

.github/workflows/test.yml
andriikushch: JFYI: I tried to remove it, but the build failed.

docs/sources/reference/components/prometheus.exporter.cloudwatch.md
thampiotr: I'm aware that there is compatibility and the old configs will work, my comment was more about using in our examples the official, correct names, not the aliases which are deprecated :)

thampiotr (approved):

clayton-cornell:
Removed backport labels.

Issue #221, opened 2024-03-12
grafana-agent helm chart should include CustomResourceDefinitions for PodMonitor, ServiceMonitor, Probe


Skaronator:
### What's wrong?

I'm currently setting up Grafana Agent in a new cluster and I noticed that the flow mode is the new preferred way how to do that and the Grafana Operator is no longer recommended. After installing the helm-chart I noticed that the ServiceMonitor, PodMonitor, Probe and PrometheusRule CRD were missing. 

How am I supposed to use the [prometheus.operator.servicemonitors](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.operator.servicemonitors/),  [prometheus.operator.podmonitors](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.operator.podmonitors/) and [mimir.rules.kubernetes](https://grafana.com/docs/agent/latest/flow/reference/components/mimir.rules.kubernetes/) when the Custom Resource Definitions aren't provided?

### Steps to reproduce

Looking into the Helm-Chart it only provides CRD for Grafana CRD: https://github.com/grafana/agent/tree/main/operations/helm/charts/grafana-agent/charts/crds/crds

Comparing this to the Grafana Agent Operator Helm-Chart which provides all CoreOS CRD: https://github.com/grafana/helm-charts/tree/main/charts/agent-operator/crds

### Software version

Grafana Agent Helm Chart 0.36.0

rfratto:
Hi, this is a duplicate of #173; going to close it in favor of the existing proposal.

Issue #173, opened 2024-04-11
Add monitoring.coreos.com/v1 crds to alloy


itspooya:
### Background

As I was checking the alloy and opportunity to move to alloy instead of grafana-agent, I found out that the crds for PodMonitor, ServiceMonitor, Probe, PodLogs are not present in alloy
In the documents for Migrating grafana agent operator to alloy there is a part which states: "The Monitor types (PodMonitor, ServiceMonitor, Probe, and PodLogs) are all supported natively by Alloy."
but without the crds, I think it is pointless

### Proposal

I propose that we should add these crds and there should be a flag to enable to install them within values.yaml

hbjydev:
Wouldn't it be better to install them separately? i.e. deploy them from the helm chart available for just the CRDs instead of bundling them? Stops them from getting out of date and breaking other tooling in your cluster at least

itspooya:
It is a good idea, but I felt like that because agent-operator has these crds and alloy doesn't have them, maybe adding them was a good idea

sboulkour:
> Wouldn't it be better to install them separately? i.e. deploy them from the helm chart available for just the CRDs instead of bundling them? Stops them from getting out of date and breaking other tooling in your cluster at least

I would too use chart `prometheus-operator-crds` to deploy rather than use CRD shipped within Alloy chart. Decoupling CRD installation from the component chart that use those is a common best practice pushed by Helm team. See https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations

edwardzjl:
Hi there, I'm also facing this issue.
I previously have `prometheus-operator` installed and collect metrics through `ServiceMonitor`.

I'm considering migrating to alloy but couldn't find any relevant documentation on using `ServiceMonitor` with it.

PR #840, opened 2024-05-14
feat(operations): Add PrometheusRules CRD

blufor:
#### PR Description

Adds CRD for PrometheusRules, content aligned with pre-existing PodLogs CRD.

Part of #221 

#### Notes to the Reviewer

I'm not familiar with the repository operations (adding docs, whether there are even tests for CRD YAML files). I've successfully deployed the CRD to the cluster. Let me know if there's further work that needs to be done

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/charts/crds/crds/monitoring.grafana.com_prometheusrules.yaml
+++ b/operations/helm/charts/alloy/charts/crds/crds/monitoring.grafana.com_prometheusrules.yaml
@@ -1,0 +1,92 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.2
+  creationTimestamp: null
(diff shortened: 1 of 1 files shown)


blufor:
@rfratto Actually, this is also missing from the flow-mode agent too. It's actually where I encountered this issue so I think it should be backported into the old agent too

rfratto:
Hi @blufor, thanks for the PR, and I'm sorry this has been sitting unreviewed for a while. 

As it stands, I think your PR is fine. However, I think we need to spend some time considering what issues might arise with including CRDs from another project directly in our Helm chart. For example, will there be any problems if the CRD has already been installed from an installation of the prometheus-operator Helm chart?

We've just introduced a new [proposal review process](https://github.com/grafana/alloy/tree/main/docs/design) to help make it clearer when things need consensus and when work for something will be accepted. For the time being, I'd ask that this remain open while we work towards consensus on #173. If that proposal gets Accepted, then we should be able to merge this PR as-is.

PR #841, opened 2024-05-14
No error when http fails

### LINKED ISSUE #694: No error logs when the HTTP server cannot start

mattdurham:
#### PR Description

Fail early if the port is not available for any reason. Originally I was going to try and handle the error through the scheduler but that was going to touch a lot of code and introduce a larger change for something that does not allow recovery.

Additional credit to @mrashad10 for initially working on this issue.

Fixes #694

#### PR Checklist

- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,9 @@ Main (unreleased)
   cluster instance to another could have a staleness marker inserted and result
   in a gap in metrics (@thampiotr)
 
-- If the port defined in `--server.http.listen-addr` or using the default listen address`127.0.0.1:12345` 
-  is not available then exit Alloy immediately. (@mattdurham)
+- Exit Alloy immediately if the port it runs on is not available. 
+  This port can be configured with `--server.http.listen-addr` or using
+  the default listen address`127.0.0.1:12345`. (@mattdurham) 
 
 v1.1.0-rc.0
 -----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Main (unreleased)
   cluster instance to another could have a staleness marker inserted and result
   in a gap in metrics (@thampiotr)
 
+- If the port defined in `--server.http.listen-addr` or using the default listen address`127.0.0.1:12345` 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ Main (unreleased)
   cluster instance to another could have a staleness marker inserted and result
   in a gap in metrics (@thampiotr)
 
+- If the port defined in `--server.http.listen-addr` or using the default listen address`127.0.0.1:12345` 
+  is not available then exit Alloy immediately. (@mattdurham)
```
erikbaranowski, line 25:
```suggestion
- Exit Alloy immediately if the port it runs on is not available. 
  This port can be configured with `--server.http.listen-addr` or using
  the default listen address`127.0.0.1:12345`. (@mattdurham) 
```

erikbaranowski (approved):

PR #848, opened 2024-05-14
remotecfg: defer creating remotecfg dir until we decide we need it

captncraig:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -48,6 +48,8 @@ type Service struct {
 	ticker            *time.Ticker
 	dataPath          string
 	currentConfigHash string
+
+	initializedStorage bool
 }
 
 // ServiceName defines the name used for the remotecfg service.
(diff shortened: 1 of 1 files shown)


captncraig:
The main problem this is trying to solve is that I've seen a number of error messages about remotecfg in unrealted issues, like when the volumes are set up wrong, or read only. That's just because remotecfg is the first thing that tries to create directories in the data dir. 

Perhaps a better alternative would be to do a more robust check on alloy startup that we can properly manage the data dir.

erikbaranowski:
> The main problem this is trying to solve is that I've seen a number of error messages about remotecfg in unrealted issues, like when the volumes are set up wrong, or read only. That's just because remotecfg is the first thing that tries to create directories in the data dir.
> 
> Perhaps a better alternative would be to do a more robust check on alloy startup that we can properly manage the data dir.

I agree with the point at the end about doing a more robust check on alloy startup for the data dir. I would start there and see if there is still an issue for `remotecfg`.

PR #869, opened 2024-05-15
clustering: fix ipv6 advertise addresses

matthewpi:
#### PR Description

The previous logic for handling IPv6 addresses for clustering was flawed.  This PR updates it so IPv6 addresses are properly joined with their ports, avoiding a `too many colons in address` error upon starting Alloy in cluster-mode.

There is not an open issue for this bug.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ Main (unreleased)
 
 - Fix a panic in `loki.source.docker` when trying to stop a target that was never started. (@wildum)
 
+- Fix error on boot when using IPv6 advertise addresses without explicitly
+  specifying a port. (@matthewpi)
+
 ### Other changes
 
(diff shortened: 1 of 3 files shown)


tpaschalis (commented):
Hey Matthew! I think this fix makes sense; is there any easy way to add a test so that we don't accidentally regress on this?

matthewpi:
> Hey Matthew! I think this fix makes sense; is there any easy way to add a test so that we don't accidentally regress on this?

I've moved the logic to a separate function so it can be tested independently. Right now it just tests that the port is correctly joined with the IP and that the loopback fallback works. Without digging deeper into where the advertise address is actually used, I cannot easily add a test ensuring that the proper address is used. But the test I did add should help avoid issues if the IP selection and port join logic changes.

tpaschalis (approved):
LGTM, thanks!

Issue #868, opened 2024-05-15
Debian 12: v1.1.0: panic loop


dotwaffle:
### What's wrong?

Installed v1.0.0 a couple of weeks ago, with Docker integration it was spamming a lot of logs (>10GB/day or almost 100x the previous rate) so I disabled Alloy. I saw notes saying that this happened when using Docker Swarm with multiple networks.

On upgrading to v1.1.0 yesterday, now Alloy sits in a panic loop.

### Steps to reproduce

Not 100% sure, but I think having a Docker Swarm single-node cluster enabled, and a number of stacks that have multiple docker networks triggers it. Happy to help debug if you reset my logs quota if/when it goes over ;)

### System information

Debian 12, amd64 (using rsyslog as upgrade from Debian 11 and 10 before that)

### Software version

Grafana Alloy v1.1.0

### Configuration

```text
Standard config with docker extension and node exporter setup.
```

### Logs

```text
May 15 12:43:58 very alloy[75533]: ts=2024-05-15T12:43:58.439250819Z level=info msg="tail routine: started" component_path=/ component_id=loki.source.file.logs_integrations_integrations_node_exporter_direct_scrape component=tailer path=/var/log/user.log
May 15 12:43:58 very alloy[75533]: ts=2024-05-15T12:43:58.439259996Z level=info msg="Seeked /var/log/user.log - &{Offset:0 Whence:0}" component_path=/ component_id=loki.source.file.logs_integrations_integrations_node_exporter_direct_scrape
May 15 12:44:08 very alloy[75533]: ts=2024-05-15T12:44:08.439937148Z level=error msg="error inspecting Docker container" target=5b71a3bd08c329e1c6cc46be0e448861f7cace88f2e7956d8d86fa122a03481a component_path=/ component_id=loki.source.docker.logs_integrations_docker id=5b71a3bd08c329e1c6cc46be0e448861f7cace88f2e7956d8d86fa122a03481a error="Get \"http://%2Fvar%2Frun%2Fdocker.sock/v1.44/containers/5b71a3bd08c329e1c6cc46be0e448861f7cace88f2e7956d8d86fa122a03481a/json\": context canceled"
May 15 12:44:08 very alloy[75533]: panic: runtime error: invalid memory address or nil pointer dereference
May 15 12:44:08 very alloy[75533]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x3d4cc7e]
[...]

PR #875, opened 2024-05-16
Fix panic loki source docker

wildum:
#### PR Description

This [fix](https://github.com/grafana/alloy/pull/742) introduced a panic when the code attempts to stop a target that was never started.

It was reported by https://github.com/grafana/alloy/issues/868 (not putting the fixes # above before the issue is confirmed to be solved)

#### Notes to the Reviewer

I could reproduce the bug via the test.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/internal/dockertarget/target.go
+++ b/internal/component/loki/source/docker/internal/dockertarget/target.go
@@ -235,9 +235,11 @@ func (t *Target) StartIfNotRunning() {
 
 // Stop shuts down the target.
 func (t *Target) Stop() {
-	t.cancel()
-	t.wg.Wait()
-	level.Debug(t.logger).Log("msg", "stopped Docker target", "container", t.containerName)
+	if t.Ready() {
+		t.cancel()
+		t.wg.Wait()
+		level.Debug(t.logger).Log("msg", "stopped Docker target", "container", t.containerName)
+	}
 }
 
 // Ready reports whether the target is running.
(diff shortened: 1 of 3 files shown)


internal/component/loki/source/docker/internal/dockertarget/target.go
```diff
@@ -235,9 +235,11 @@ func (t *Target) StartIfNotRunning() {
 
 // Stop shuts down the target.
 func (t *Target) Stop() {
-	t.cancel()
-	t.wg.Wait()
-	level.Debug(t.logger).Log("msg", "stopped Docker target", "container", t.containerName)
+	if t.Ready() {
+		t.cancel()
+		t.wg.Wait()
```
ptodev, line 240: Isn't the nil pointer dereference due to a `t` which is nil? I thought all we need to do is:
```
if t == nil {
  return
}
```

I would prefer if we don't call `t.Ready()` without mutexes, since the state of the container might change between the call to `t.Ready()` and the call to stop. I know it's unlikely, but it feels more resilient.

wildum: I think that it's actually the cancel function which is nil in this case. if you check the logs in the issue https://github.com/grafana/alloy/issues/868, there is an error line right before the panic with the id of the target returned by t.target.Name().
If the state changes in between that would not be a problem because if the target ran once, there should be a cancel function associated with it. So in the very specific case where ready is true and the process loop ends setting it to false right before the cancel call, then the cancel would be a no-op

ptodev (approved):

PR #879, opened 2024-05-16
feat(otelcol/connector/servicegraph): add `virtual_node_peer_attributes` setting to the component

### LINKED ISSUE #861: Add the virtual_node_peer_attributes setting to otelcol.connector.servicegraph
### Request

The upstream servicegraphconnector has moved the virtual nodes feature to stage beta, it is automatically enabled now in grafana alloy 1.1.0. 

### Use case

I would like to configure the attributes used for determining virtual nodes

hainenber:
#### PR Description
Add the virtual_node_peer_attributes setting to otelcol.connector.servicegraph. Thanks @erikbaranowski for creating the foundation already!

Closes #861 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Main (unreleased)
 
 - (_Public preview_) Add native histogram support to `otelcol.receiver.prometheus`. (@wildum)
 
+- Add the virtual_node_peer_attributes setting to `otelcol.connector.servicegraph`. (@hainenber)
+
 ### Bugfixes
 
 - Fix panic when component ID contains `/` in `otelcomponent.MustNewType(ID)`.(@qclaogui)
--- a/docs/sources/reference/components/otelcol.connector.servicegraph.md
+++ b/docs/sources/reference/components/otelcol.connector.servicegraph.md
@@ -61,6 +61,7 @@ Name                        | Type             | Description                    
 `store_expiration_loop`     | `duration`       | The time to expire old entries from the store periodically.         | `"2s"`       | no
 `metrics_flush_interval`    | `duration`       | The interval at which metrics are flushed to downstream components. | `"0s"`       | no
 `database_name_attribute`   | `string`         | The attribute name used to identify the database name from span attributes. | `"db.name"`  | no
+`virtual_node_peer_attributes` | `list(string)` | List of attributes whose presence in client span will result in creation of a virtual server node | `["peer.service", "db.name", "db.system"]` | no
 
 Service graphs work by inspecting traces and looking for spans with parent-children relationship that represent a request.
 `otelcol.connector.servicegraph` uses OpenTelemetry semantic conventions to detect a myriad of requests.
--- a/internal/component/otelcol/connector/servicegraph/servicegraph.go
+++ b/internal/component/otelcol/connector/servicegraph/servicegraph.go
@@ -12,6 +12,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/connector/servicegraphconnector"
 	otelcomponent "go.opentelemetry.io/collector/component"
 	otelextension "go.opentelemetry.io/collector/extension"
+	semconv "go.opentelemetry.io/collector/semconv/v1.13.0"
 )
 
 func init() {
@@ -49,9 +50,7 @@ type Arguments struct {
 	// StoreExpirationLoop defines how often to expire old entries from the store.
 	StoreExpirationLoop time.Duration `alloy:"store_expiration_loop,attr,optional"`
 	// VirtualNodePeerAttributes the list of attributes need to match, the higher the front, the higher the priority.
-	//TODO: Add VirtualNodePeerAttributes when it's no longer controlled by
-	// the "processor.servicegraph.virtualNode" feature gate.
-	// VirtualNodePeerAttributes []string `alloy:"virtual_node_peer_attributes,attr,optional"`
+	VirtualNodePeerAttributes []string `alloy:"virtual_node_peer_attributes,attr,optional"`
 
 	// MetricsFlushInterval is the interval at which metrics are flushed to the exporter.
 	// If set to 0, metrics are flushed on every received batch of traces.
@@ -109,18 +108,11 @@ func (args *Arguments) SetToDefault() {
 		CacheLoop:             1 * time.Minute,
 		StoreExpirationLoop:   2 * time.Second,
 		DatabaseNameAttribute: "db.name",
-		//TODO: Add VirtualNodePeerAttributes when it's no longer controlled by
-		// the "processor.servicegraph.virtualNode" feature gate.
-		// VirtualNodePeerAttributes: []string{
-		// 	semconv.AttributeDBName,
-		// 	semconv.AttributeNetSockPeerAddr,
-		// 	semconv.AttributeNetPeerName,
-		// 	semconv.AttributeRPCService,
-		// 	semconv.AttributeNetSockPeerName,
-		// 	semconv.AttributeNetPeerName,
-		// 	semconv.AttributeHTTPURL,
-		// 	semconv.AttributeHTTPTarget,
-		// },
+		VirtualNodePeerAttributes: []string{
+			semconv.AttributePeerService,
+			semconv.AttributeDBName,
+			semconv.AttributeDBSystem,
+		},
 	}
 	args.Store.SetToDefault()
 }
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Main (unreleased)
 
 - (_Public preview_) Add native histogram support to `otelcol.receiver.prometheus`. (@wildum)
 
+- Add the virtual_node_peer_attributes setting to `otelcol.connector.servicegraph`. (@hainenber)
```
wildum, line 22:
```suggestion
- Add the `virtual_node_peer_attributes` argument to `otelcol.connector.servicegraph`. (@hainenber)
```

docs/sources/reference/components/otelcol.connector.servicegraph.md
```diff
@@ -61,6 +61,7 @@ Name                        | Type             | Description
 `store_expiration_loop`     | `duration`       | The time to expire old entries from the store periodically.         | `"2s"`       | no
 `metrics_flush_interval`    | `duration`       | The interval at which metrics are flushed to downstream components. | `"0s"`       | no
 `database_name_attribute`   | `string`         | The attribute name used to identify the database name from span attributes. | `"db.name"`  | no
+`virtual_node_peer_attributes` | `list(string)` | List of attributes whose presence in client span will result in creation of a virtual server node | `["peer.service", "db.name", "db.system"]` | no
```
wildum, line 64: After the table I would add that the attributes are ordered by priority and that an empty list disables virtual node creation.

hainenber: Done in HEAD :D

wildum (approved):
LGTM, thanks!

wildum:
waiting for @clayton-cornell feedback on the doc before merging it

docs/sources/reference/components/otelcol.connector.servicegraph.md
```diff
@@ -107,6 +108,8 @@ Additional labels can be included using the `dimensions` configuration option:
 
 When `metrics_flush_interval` is set to `0s`, metrics will be flushed on every received batch of traces.
 
+Attributes configured in `virtual_node_peer_attributes` argument are ordered by priority. Empty list disables the creation of virtual node.
```
clayton-cornell, line 111:
```suggestion
Attributes configured in the `virtual_node_peer_attributes` argument are ordered by priority. An empty list disables the creation of a virtual node.
```

docs/sources/reference/components/otelcol.connector.servicegraph.md
clayton-cornell, line 64:
```suggestion
`virtual_node_peer_attributes` | `list(string)` | A list of attributes to search for in the client span. A virtual node is created if the attributes are found in the client span. | `["peer.service", "db.name", "db.system"]` | no
```

This description is a bit confusing to the uninitiated.  Is one virtual node created for each attribute? Or a single virtual node is created if any of the attributes are found?

Three attributes are named as "default". Are there any other attributes? How does the user know what these should be? Or are they defined by the user?

Removed `server` form the description for terminology consistency.

ptodev: I agree that it is not very obvious how to use this functionality... unfortunately this is because the OTel docs are not clear about it. I had to go through the code and [past issues](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/17196) in order to see why this was added.

We could have this in the table:
```
`virtual_node_peer_attributes` | `list(string)` | A list of attributes to search for in the client span. | `["peer.service", "db.name", "db.system"]` | no
```

And this under the table:

```
`virtual_node_peer_attributes` is useful when an OTel-instrumented client sends a request to a service which is not OTel-instrumented.
Normally, `otelcol.connector.servicegraph` wouldn't be able to pair the client span with a service span, 
because no service span will be received if the service is not OTel-instrumented.
When an edge expires, `otelcol.connector.servicegraph` checks if it has peer attributes listed in `virtual_node_peer_attributes`.
If an attribute is found, the metrics are then aggregated with a virtual node.
```

We should probably add a note like this somewhere in the docs:

```
If no client span is found and `virtual_node_peer_attributes` is not an empty list,
then the service span will be paired with a virtual node called `client="user"`.
This can be useful when a client which is not OTel-instrumented (like a web browser) sends a request to an OTel-instrumented service. 
Without a virtual node, normally the client span will be missing, and the server span will expire without being paired.
```

It's a bit weird that `virtual_node_peer_attributes` has an influence over `client="user"` given that `virtual_node_peer_attributes` is not actually needed for this use case. It'd be worth raising an issue in the upstream OTel repo about this...

@mapno since you seem to be familiar with this feature, do you think this is a good description? I suppose we wouldn't get any "error" and "duration" metrics from this?

clayton-cornell: This makes a LOT more sense with the additional information.

hainenber: Oops, forget that merging suggested change will resolve threads 🫣

hainenber: There we go, I added Paulin's suggested doc. PTAL folks!

ptodev (changes requested):
I opened a [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33282) upstream for this config option to be changed upstream. It would be better to update Alloy straight to the new config, so let's put the PR on hold for now to see if the upstream PR gets accepted.

hainenber:
All fine by me :D

docs/sources/reference/components/otelcol.connector.servicegraph.md
mapno: @ptodev your description is spot-on 👍

docs/sources/reference/components/otelcol.connector.servicegraph.md
```diff
@@ -107,6 +108,19 @@ Additional labels can be included using the `dimensions` configuration option:
 
 When `metrics_flush_interval` is set to `0s`, metrics will be flushed on every received batch of traces.
 
+`virtual_node_peer_attributes` is useful when an OTel-instrumented client sends a request to a service which is not OTel-instrumented.
```
clayton-cornell, line 111:
```suggestion
`virtual_node_peer_attributes` is useful when an OTel-instrumented client sends a request to a service that is not OTel-instrumented.
```

docs/sources/reference/components/otelcol.connector.servicegraph.md
```diff
@@ -107,6 +108,19 @@ Additional labels can be included using the `dimensions` configuration option:
...
+Normally, `otelcol.connector.servicegraph` wouldn't be able to pair the client span with a service span, 
+because no service span will be received if the service is not OTel-instrumented.
+When an edge expires, `otelcol.connector.servicegraph` checks if it has peer attributes listed in `virtual_node_peer_attributes`.
+If an attribute is found, the metrics are then aggregated with a virtual node.
+
+If no client span is found and `virtual_node_peer_attributes` is not an empty list,
+then the service span will be paired with a virtual node called `client="user"`.
+This can be useful when a client which is not OTel-instrumented (like a web browser) sends a request to an OTel-instrumented service.
```
clayton-cornell, line 119:
```suggestion
This can be useful when a client that is not OTel-instrumented (like a web browser) sends a request to an OTel-instrumented service. 
```

docs/sources/reference/components/otelcol.connector.servicegraph.md
```diff
@@ -107,6 +108,19 @@ Additional labels can be included using the `dimensions` configuration option:
 
 When `metrics_flush_interval` is set to `0s`, metrics will be flushed on every received batch of traces.
 
+`virtual_node_peer_attributes` is useful when an OTel-instrumented client sends a request to a service which is not OTel-instrumented.
+Normally, `otelcol.connector.servicegraph` wouldn't be able to pair the client span with a service span,
```
clayton-cornell, line 112:
```suggestion
Normally, `otelcol.connector.servicegraph` wouldn't be able to pair the client span with a service span
```

clayton-cornell (commented):
Still in passive voice, but leaving it as it gets the technical point across.

ptodev:
@hainenber I'll close this one in favour of #5058, which is up to date and includes an additional config argument.

There is still a feature gate but since it's enabled by default I don't think it's a problem. I suppose that the feature gate was either disabled by default at the time of your PR, or we missed the fact that it's not that important 😅

PR #882, opened 2024-05-16
otelcol: decouple otel/alloy component IDs

tpaschalis:
#### PR Description
This PR decouples the Alloy component IDs from the generated types used for OTel Collector components internally; this fixes issues where long block labels could exceed the 63-character limit recently set by the upstream opentelemetry packages.

It is basically similar to what we've been doing with the jaegerremotesampling extension 
https://github.com/grafana/alloy/blob/1cdd942cd43cbf013d9749a22a58c9f0a9cee94f/internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/factory.go#L19-L27

No issue filed

#### Notes to the Reviewer
I don't think this as a user-facing change so I haven't added a changelog entry.

#### PR Checklist

- [ ] CHANGELOG.md updated (N/A)
- [ ] Tests updated


internal/component/otelcol/receiver/prometheus/prometheus.go
```diff
@@ -117,11 +118,9 @@ func (c *Component) Update(newConfig component.Arguments) error {
 		gcInterval = 5 * time.Minute
 	)
 
-	cTypeStr := strings.ReplaceAll(strings.ReplaceAll(c.opts.ID, ".", "_"), "/", "_")
-
 	settings := otelreceiver.CreateSettings{
 
-		ID: otelcomponent.NewID(otelcomponent.MustNewType(cTypeStr)),
+		ID: otelcomponent.NewID(otelcomponent.MustNewType(typeStr)),
```
rfratto, line 123: In the other packages that use CreateSettings (like otelcol/exporter), they omit the ID entirely; should that be done here too? 

The ID is really only needed for extensions IIUC

internal/component/otelcol/auth/auth.go
```diff
@@ -168,12 +169,10 @@ func (a *Auth) Update(args component.Arguments) error {
 		components = append(components, ext)
 	}
 
-	cTypeStr := strings.ReplaceAll(strings.ReplaceAll(a.opts.ID, ".", "_"), "/", "_")
-
 	// Inform listeners that our handler changed.
 	a.opts.OnStateChange(Exports{
 		Handler: Handler{
-			ID:        otelcomponent.NewID(otelcomponent.MustNewType(cTypeStr)),
+			ID:        otelcomponent.NewID(otelcomponent.MustNewType(typeStr)),
```
rfratto, line 175: This would be a collision issue if there's ever two auth extensions used for a component; do you know if that's possible? Even if it's not, should we preemptively defend against that?

tpaschalis: AFAICT the only way to use auth components is currently in exporters which can only ever instantiate one `client` block. As far as I've read the code and tested, it shoulnd't be an issue.

I see the point though to make sure we don't trip over ourselves in the future.

internal/component/otelcol/receiver/prometheus/prometheus.go
tpaschalis: Yeah, let's go with that, I've removed the struct field entirely from here.

internal/component/otelcol/auth/auth.go
tpaschalis: PTAL. Just for being defensive, I'm using a small hash when the 63-char limit is exceeded, in a helper function. This can likely move to a shared place if we need it again.

captncraig (approved):

rfratto (approved):

rfratto:
@tpaschalis 

Based on this:

> this fixes issues where long block labels could exceed the 63-character limit recently set by the upstream opentelemetry packages.

It does sound like a user-facing change, right? Since if the label for the Alloy component was long enough, it would panic?

PR #891, opened 2024-05-21
fix: close procRoot dir

simonswine:
This fix was originally suggested in https://github.com/grafana/agent/pull/6906. Sot it is a forward port.


simonswine:
Just rebase, a review would be appreciated

korniltsev (approved):

PR #894, opened 2024-05-21
remotecfg: add some basic metrics

captncraig:
New metrics:

- remotecfg_hash
- remotecfg_last_load_successful
- remotecfg_load_failures_total
- remotecfg_last_load_success_timestamp_seconds
- remotecfg_load_attempts_total
- remotecfg_request_duration_seconds

Component loading info should already be exposed via controller metrics. 

This does not give any insight into the fallback local files, and is focused mainly on the remote fetch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Main (unreleased)
 ### Enhancements
 
 - (_Public preview_) Add native histogram support to `otelcol.receiver.prometheus`. (@wildum)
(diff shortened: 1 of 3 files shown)


mattdurham:
Are we always injecting these metrics even if someone is not using remote config?

captncraig:
> Are we always injecting these metrics even if someone is not using remote config?

Good point. Not everything is very good about deferring registration, but no reason this couldn't.

mattdurham (approved):
Looks good but would love if we could only have the metrics registered if using remote configuration.

PR #904, opened 2024-05-22
chore(pyroscope): mark pyroscope components as GA

korniltsev:
#### PR Description

Dealing with stability-preview is a burden for both internal grafana users and external.
Let's mark `pyroscope.ebpf`, `pyroscope.java`, `pyroscope.scrape`, `pyroscope.write`, `discovery.process` as GA.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ Main (unreleased)
 
 ### Other changes
 
+- `pyroscope.ebpf`, `pyroscope.java`, `pyroscope.scrape`, `pyroscope.write` and `discovery.process` components are now GA. (@korniltsev)
+
 - `prometheus.exporter.snmp`: Updating SNMP exporter from v0.24.1 to v0.26.0. (@ptodev, @erikbaranowski)
 
 - `prometheus.scrape` component's `enable_protobuf_negotiation` argument is now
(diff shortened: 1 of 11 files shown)


simonswine (approved):
LGTM Either that. I probably would even argue that the default stability-level for Alloy should be "public-preview"

korniltsev:
Let's wait for more approves from @rfratto @tpaschalis @petethepig @Rperry2174

rfratto:
I don’t think we should mark components GA because it’s a burden. If the components are stable, are ready for production use, and they won’t have any breaking changes, then they qualify for GA. Are they at that state yet?

rfratto:
For reference, GA means these two things: 

* It follows the definitions of the [Grafana Labs release cycle](https://grafana.com/docs/release-life-cycle/)
* [Breaking changes will no longer happen in minor releases](https://grafana.com/docs/alloy/latest/introduction/backward-compatibility/)

While not documented, we also are trying to aim to release a new major version no more frequently than yearly. 

I’m comfortable with this PR being merged if these components are ready, but I want to make sure the implications are known first.

Rperry2174 (approved):
Discussed with the team and also agree that we meet the GA criteria! 👍

rfratto (approved):
Given the above, I'm on board 👍

clayton-cornell (approved):

PR #907, opened 2024-05-22
Downgrade error log lines on Windows filesystem access issues

thampiotr:
#### PR Description

We had reports of misleading log lines being printed on Windows when log files are being deleted / moved during log file rotation. For example:

```
{"ts":"2024-05-16T06:00:49.0098797Z","level":"error","msg":"error stopping tailer","component_path":"...","component_id":"...","component":"tailer","path":"...","error":"seek ...: file already closed"}
{"ts":"2024-05-16T05:00:09.0009908Z","level":"error","msg":"error getting os stat","component_path":"...","component_id":"...","path":"...","err":"CreateFile ...: Access is denied."}
```

It seems that on Windows we can get these errors when trying to access the files shortly after they're being moved/deleted. In this PR we detect these specific errors and downgrade the log line to `debug` level.

Reproduced and later verified this fix works on a Windows VM. Also checked the goroutine dump to verify no goroutine leaks occurred. 

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/internal/component/local/file_match/watch.go
+++ b/internal/component/local/file_match/watch.go
@@ -4,9 +4,10 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/bmatcuk/doublestar"
 	"github.com/go-kit/log"
 
-	"github.com/bmatcuk/doublestar"
+	"github.com/grafana/alloy/internal/component/common/loki/utils"
 	"github.com/grafana/alloy/internal/component/discovery"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
 )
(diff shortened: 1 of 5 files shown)


internal/component/local/file_match/watch.go
```diff
@@ -38,7 +39,13 @@ func (w *watch) getPaths() ([]discovery.Target, error) {
 		}
 		fi, err := os.Stat(abs)
 		if err != nil {
-			level.Error(w.log).Log("msg", "error getting os stat", "path", abs, "err", err)
+			// On some filesystems we can get errors accessing the discovered paths. Don't log these as errors.
+			// local.file_match will retry on the next sync period if the access is blocked temporarily only.
+			if utils.IsEphemeralOrFileClosed(err) {
+				level.Debug(w.log).Log("msg", "I/O error when getting os stat", "path", abs, "err", err)
```
wildum, line 45: What if we consistently hit an ephemeral error on every retry? Should we have a timeout that logs it as an error after a while?

tpaschalis (approved):
LGTM as a solution for the specific problem.

internal/component/local/file_match/watch.go
tpaschalis: I think this'd be quite more convoluted for this specific use case (I'm not sure if it'd get stuck in one of these ephemeral errors or keep the closed status for an extended period of time).

thampiotr: Then we would not be picking that specific file. Still can be debugged with debug logging. I don't believe that right now we need to go into handling it further.

Issue #908, opened 2024-05-22
Proposal: Alloy proposal process


rfratto:
### Background

Today, there is no formalized process for proposing changes to Alloy, and no formalized process for reviewing proposed changes. This has led to governance members being unsure how to get a proposal reviewed. This becomes exacerbated for external contributors who do not have a direct line of communication to governance members. The result of a lack of process can be seen by the number of open proposals in an unknown state; 47 at the time of writing.

At the same time, more and more proposals for Alloy have been performed offline. This has gradually drifted away from the [design in the open][] mentality previously accepted by Grafana Agent. This has made it more difficult for external contributors to understand and influence the direction of the
Alloy project.

Formalizing a process for creating and reviewing proposals will help ensure that:

* All contributors can contribute to the direction of Alloy.
* All existing and new proposals are guaranteed to be evaluated and decided on.
* A publicly accessible single source of truth is used for discussing all proposals.

[design in the open]: https://github.com/grafana/agent/blob/main/docs/rfcs/0001-designing-in-the-open.md

### Proposal

At a high-level, proposals go through the following stages:

1. *Issue*: A brief [issue][new-proposal] is created for the proposal.

2. *Discuss*: Public discussion on the proposal issue drives the proposal towards one of three outcomes:

   * Accept the proposal, or
   * decline the proposal, or
   * ask for a design document.

   If the proposal is accepted or declined, the proposal process ends here.

3. *Design document*: If requested, a design document is written.

4. *Consensus*: Once comments and changes have slowed down, a final public discussion aims to reach consensus for one of two outcomes:

   * Accept the proposal, or
   * decline the proposal.

A more detailed design document will be opened shortly after creating this issue, and I will update this issue to link to it.

rfratto:
cc @grafana/grafana-alloy-team

wildum:
> Once comments on PRs for design documents have slowed down, the PR is merged
regardless of the current state of the proposal. More PRs to update the design
document may be warranted based on ongoing discussions on the proposal issue.

Just to clarify, that means that even if the proposal goes in Likely Declined following the discussions on the design doc PR, we would still merge it?

wildum:
> Proposals are in one of the following states:

The doc mentions a __Hold__ state which is not described there:

>Proposals that break [backward compatibility] are moved to a "Hold" state until
work on the next major release begins.

>This does not include the time
a proposal may spend in _Incoming_ or _Hold_

Should it be added?

wildum:
[...]

PR #909, opened 2024-05-22
Design document for Alloy proposal process (#908)

rfratto:
This PR creates a design document for #908. 

To demonstrate the proposed process, the design document follows the procedures laid out in the proposal itself.

### DIFF
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -97,8 +97,11 @@ design document.
   design document itself. All other comments should be made on the proposal
   issue.
 
-Once a design document is created, the proposal issue should be updated and a
-comment should be made to link to the design document.
+Once a design document is created, the proposal author should:
+
+* Update the proposal issue description to link to the design document.
+* Add a comment notifying readers that a design document is created, with a
+  link to the design document.
 
 Once comments on PRs for design documents have slowed down, the PR is merged
 regardless of the current state of the proposal. More PRs to update the design
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -2,7 +2,7 @@
 
 * Author: Robert Fratto (@rfratto)
 * Last updated: 2024-05-22
-* Discussion: https://github.com/grafana/alloy/issues/908
+* Original issue: https://github.com/grafana/alloy/issues/908
 
 ## Abstract
 
@@ -54,7 +54,8 @@ At a high-level, proposals go through the following stages:
 
    If the proposal is accepted or declined, the proposal process ends here.
 
-3. *Design document*: If requested, a design document is written.
+3. *Design document*: If requested, a design document is written, and the PR
+   for the design document becomes the new home for discussing the proposal.
 
 4. *Consensus*: Once comments and changes have slowed down, a final public
    discussion aims to reach consensus for one of two outcomes:
@@ -107,11 +108,13 @@ design document.
 
 Once a design document is created, the proposal author should:
 
-* Update the proposal issue description to link to the design document.
-* Add a comment notifying readers that a design document is created, with a
-  link to the design document.
+* Close the original proposal issue and link to the design document PR.
+* Update the proposal issue description with a link to the design document.
 
-Design documents do not track the state of the proposal; the proposal issue is
+Closing the original proposal issue in favor of the design document PR prevents
+fragmenting discussions as contributors work towards consensus on a proposal.
+
+Design documents do not track the state of the proposal; the proposal PR  is
 the source of truth.
 
 The PR for design documents are merged once the associated proposal becomes
@@ -125,7 +128,7 @@ is closed.
 
 * Author(s): [List of proposal authors]
 * Last updated: [Date]
-* Discussion: https://github.com/grafana/alloy/issues/NNNN
+* Original issue: https://github.com/grafana/alloy/issues/NNNN
 
 ## Abstract
 
@@ -195,8 +198,8 @@ Proposals are in one of the following states:
 * **Likely Accept**: If a proposal discussion seems to have reached a consensus
   to accept a proposal, the proposal review group moves the proposal to the
   _Likely Accept_ state. This change should be indicated with a comment on the
-  proposal issue. This state marks the final period for comments that might
-  change the recognition of consensus.
+  proposal. This state marks the final period for comments that might change
+  the recognition of consensus.
 
 * **Likely Decline**: If a proposal discussion seems to have reached a consensus
   to decline a proposal, the proposal review group moves the proposal to the
@@ -207,8 +210,8 @@ Proposals are in one of the following states:
   not accepting the proposal is appropriate.
 
   Similarly to _Likely Accept_, this change should be indicated with a comment
-  on the proposal issue, and this state marks the final period for comments
-  that might change the recognition of consensus.
+  on the proposal, and this state marks the final period for comments that
+  might change the recognition of consensus.
 
 * **Accepted**: If a proposal has been in _Likely Accept_ for a week and
   consensus has not changed, the proposal is moved to the _Accepted_ state.
@@ -308,11 +311,11 @@ decision to be made.
 
 To re-embrace the spirit of "design in the open," conversations about a
 proposal should be discoverable by new parties to make it easy to participate.
-Having a single source of truth for discussions (the proposal issue) prevents
-conversations from fragmenting in multiple locations (such as design document
-PRs, offline discussions, Google Docs). This ensures that contributors know
-where to look for discussions and important information doesn't get lost while
-trying to reach consensus.
+Having a single source of truth for discussions (the proposal issue or design
+document PR) prevents concurrent conversations from fragmenting in multiple
+locations. This ensures that contributors know where to look for active
+discussions and important information doesn't get lost while trying to reach
+consensus.
 
 However, this process comes with some trade-offs, described in sections below.
 
@@ -340,14 +343,18 @@ the time it takes to get a proposal approved and implemented into a release.
 
 ### Conversational toil
 
-Using GitHub issues for proposals ensures there is a single source of truth for
-discussion, and that all interested parties can participate in shaping the
-future of Alloy.
+Using GitHub for all proposals ensures there is a single source of truth at a
+given time for discussion, and that all interested parties can participate in
+shaping the future of Alloy.
 
-However, GitHub's issue system is not well-suited for long-form discussions.
-This may lead to long threads that are difficult to follow.
+However, GitHub is not well-suited for long-form discussions. This may lead to
+long threads that can be difficult to follow.
 
-There are two mitigation strategies to help with this:
+There are some mitigation strategies to help with this:
+
+* Discussion on issues should be done conversationally rather than responding
+  to individual lines of text. This can help reduce the total number of
+  comments and make a conversation more digestible.
 
 * Authors of proposals can start with offline discussions and use alternative
   tools (such as Google Docs) to prepare proposals for public presentation.
@@ -355,6 +362,11 @@ There are two mitigation strategies to help with this:
 * The proposal review group keeps discussions focused by preventing derailing
   and pointing out when concurrent conversational threads can be joined
   together under a larger idea.
+
+* More involved proposals are transformed into design document PRs, where
+  tooling for discussions is better suited for long-form discussions. Less
+  involved proposals can require less discussion, potentially avoiding the
+  issue of an issue becoming long-winded.
 
 ## Compatibility
 
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -293,6 +293,11 @@ A publicly available GitHub project tracks proposals and their states.
 The goal of the proposal process is to reach general public consensus about the
 outcome in a timely manner.
 
+General public consensus does not require unanimity; one person disagreeing
+will not necessarily prevent consensus from being achieved. Consensus should
+consider all raised concerns, but not necessarily addressing all concerns
+raised.
+
 If the proposal review process cannot identify a general consensus of the
 proposal, the usual result is that the proposal is declined. It can happen that
 the proposal review process may not identify a general consensus but the
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -156,7 +156,7 @@ is closed.
 
 [A description of the steps in the implementation, who will do them, and when.]
 
-## Open issues
+## Related open issues
 
 [If applicable, a discussion of issues relating to this proposal for which the author does not know the solution. This section may be omitted if there are none.]
 ```
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -103,14 +103,12 @@ Once a design document is created, the proposal author should:
 * Add a comment notifying readers that a design document is created, with a
   link to the design document.
 
-Once comments on PRs for design documents have slowed down, the PR is merged
-regardless of the current state of the proposal. More PRs to update the design
-document may be warranted based on ongoing discussions on the proposal issue.
-Any links on the proposal issue should be updated to point to the latest
-version of the design document.
-
 Design documents do not track the state of the proposal; the proposal issue is
 the source of truth.
+
+The PR for design documents are merged once the associated proposal becomes
+_Accepted_. If the proposal becomes _Declined_, the PR for the design document
+is closed.
 
 ### Design document template
 
--- a/docs/design/908-proposal-process.md
+++ b/docs/design/908-proposal-process.md
@@ -333,7 +333,9 @@ There are two mitigation strategies to help with this:
 * Authors of proposals can start with offline discussions and use alternative
   tools (such as Google Docs) to prepare proposals for public presentation.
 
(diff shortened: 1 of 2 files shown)


rfratto:
Original discussion can be found at #908.

rfratto:
@erikbaranowski @tpaschalis @mattdurham Does the current state of the design document address all of your concerns?

@wildum ping on the current state just to make sure you're still comfortable with everything here. 

@thampiotr I know you had some concerns with the process. Is the current state better? What changes would you like to see?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+[vote]: https://github.com/grafana/alloy/blob/main/GOVERNANCE.md#voting
+
+## Rationale
+
+Formalizing proposal creation and review ensures that all proposals are
+eventually given attention. Having proposals reviewed weekly ensures that
+proposals are kept moving forward and nobody is left waiting months for a
+decision to be made.
+
+To re-embrace the spirit of "design in the open," conversations about a
```
ptodev, line 314:
```suggestion
To re-embrace the spirit of "design in the open", conversations about a
```

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+compatibility of Alloy.
+
+## Implementation
+
+This proposal is written using the new process to serve as an example.
+Following the proposed process, the next steps are:
+
+1. The proposal is currently in the _Active_ state; it bypasses _Incoming_ as it
+   is the only proposal using this process. Discussion towards a consensus
+   (accept or decline) is ongoing at #908.
```
ptodev, line 385: > Discussion towards a consensus (accept or decline) is ongoing at #908.

Is this still the case?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+2. Once discussion slows down, the proposal will move to either _Likely Accept_
+   or _Likely Decline_. This depends on when discussion slows down, but is
+   likely to be at least a week from the date this was opened.
+
+3. After being in _Likely Accept_ or _Likely Decline_ for a week, this proposal
+   will be either formally accepted or declined and move to the relevant state.
+
+If this proposal is accepted, all existing an new proposals will be subject to
```
ptodev, line 394:
```suggestion
If this proposal is accepted, all existing and new proposals will be subject to
```

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+  A proposal may also be moved to _Likely Decline_ if the proposal review group
+  identifies that no consensus is likely to be reached and that the default of
+  not accepting the proposal is appropriate.
+
+  Similarly to _Likely Accept_, this change should be indicated with a comment
+  on the proposal, and this state marks the final period for comments that
+  might change the recognition of consensus.
+
+* **Accepted**: If a proposal has been in _Likely Accept_ for a week and
+  consensus has not changed, the proposal is moved to the _Accepted_ state.
```
ptodev, line 219: I'm not sure if one week is long enough. Sometimes the people who can give the most feedback on a proposal may not be able to take a look for several days. IMO 2 weeks is more reasonable.

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+[If applicable, a discussion of issues relating to this proposal for which the author does not know the solution. This section may be omitted if there are none.]
+```
+
+### Proposal review
+
+Every week, a subset of Alloy [governance] members review open and pending
+proposals as part of the proposal review process. The proposal review process
+may be performed asynchronously or synchronously, up to the discretion of the
+governance members.
```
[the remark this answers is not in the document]
erikbaranowski: The next paragraph describes the weekly review goals which are to make sure proposals are moving along, not to do the proposal review. My understanding is the actual discussion and review does happen asynchronously on the github issue/design doc PR.

docs/design/908-proposal-process.md
rfratto: Commas go inside the quotations :)

docs/design/908-proposal-process.md
rfratto: If there's a specific person that we need input from, then we can wait for them. But by the point a consensus reaches Likely Accept, consensus has already been achieved from the individuals who need to give feedback, as the review committee would've already pinged them.

Also note that changing the transition time from Likely Accept to Accept from one to two weeks also increases the minimum amount of time from 2-3 weeks to 3-4 weeks, which is more than half of a release cycle.

docs/design/908-proposal-process.md
ptodev: I can't believe I never knew this 😆

docs/design/908-proposal-process.md
rfratto: To expand on that: "likely accept" means that consensus has already been reached, and relevant people with the most knowledge about a specific topic have already signed off and are a part of the consensus. Likely accept is the final comment period to give external contributors time to raise final concerns and change the current consensus.

rfratto:
@ptodev Checking in, what's your current feeling on this proposal?

ptodev:
> @ptodev Checking in, what's your current feeling on this proposal?

@rfratto I'm very happy with it, thank you! It looks great and I'm happy for it to be merged.

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+[governance]: https://github.com/grafana/alloy/blob/main/GOVERNANCE.md
+
+
+### Consensus and disagreement
+
+The goal of the proposal process is to reach general public consensus about the
+outcome in a timely manner.
+
+If the proposal review process cannot identify a general consensus of the
```
thampiotr, line 287: I am guessing that you mean some kind of weak consensus where one person disagreeing will not necessarily mean that the proposal was veto'd. But I think we should probably be more clear about this, as general consensus may be understood as everyone agreeing.

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+shaping the future of Alloy.
+
+However, GitHub is not well-suited for long-form discussions. This may lead to
+long threads that can be difficult to follow.
+
+There are some mitigation strategies to help with this:
+
+* Discussion on issues should be done conversationally rather than responding
+  to individual lines of text. This can help reduce the total number of
+  comments and make a conversation more digestible.
```
thampiotr, line 359: As mentioned before, I find long-form essays by far less digestible than comments on individual lines of text, where all the context is available. I feel the argument in your doc is up for the debate and could actually have quite the opposite effect. A lot (if not all) modern communication tools for large groups of people support threads, so my view on this topic is also likely to be quite common.

Having said that, I'm happy that PR with design doc can become the source of truth for a more involved discussion. In fact, I would consider a policy where anything more than two threads or two rounds of technical comments should usually be migrated to a design doc PR. Worst case it'd be a short design doc.

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+* Authors of proposals can start with offline discussions and use alternative
+  tools (such as Google Docs) to prepare proposals for public presentation.
+
+* The proposal review group keeps discussions focused by preventing derailing
```
thampiotr, line 364: Can you clarify how this will work? Say there are 4 different people mentioning 4 different concerns, will the review group mandate that we discuss them in some specific order, only one at a time?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+  might change the recognition of consensus.
+
+* **Accepted**: If a proposal has been in _Likely Accept_ for a week and
+  consensus has not changed, the proposal is moved to the _Accepted_ state.
+
+  If significant discussion happens during the _Likely Accept_ period, the
+  proposal review group may decide to leave the proposal in _Likely Accept_ for
+  another week or even move the proposal back to _Active_.
+
+  Once a proposal has been Accepted, implementation work may begin.
```
thampiotr, line 225: What do we do if we accept one proposal, but then circumstances change or a new proposal comes around and we want to decline a previously accepted proposal?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+Once a design document is created, the proposal author should:
+
+* Close the original proposal issue and link to the design document PR.
+* Update the proposal issue description with a link to the design document.
+
+Closing the original proposal issue in favor of the design document PR prevents
+fragmenting discussions as contributors work towards consensus on a proposal.
+
+Design documents do not track the state of the proposal; the proposal PR  is
+the source of truth.
```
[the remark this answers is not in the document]
rfratto: No; in the current state, if a design document is requested, the issue gets closed in favor of the PR for the design document. 

What I mean to say is that the design document should not include the current state of the proposal (active/likely accept/accepted/etc), and if you want to know what that state is, go to the PR associated with that design document.

For proposals that don't have design documents (which will be most of them), the issue is the source of truth instead.

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+The PR for design documents are merged once the associated proposal becomes
+_Accepted_. If the proposal becomes _Declined_, the PR for the design document
```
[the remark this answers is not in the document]
rfratto: My concern about not closing the issue is that the proposal review team will have to continue to monitor both the issue and the design document PR to make sure discussions are happening in the right place, generating more work. 

Do you have any other ideas of how we can ensure discussions happen in one place at a time?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+[A short summary of the proposal.]
+
+## Background
+
+[An introduction of the necessary context and the proposal being solved by the proposed change. Background should include why the proposal is necessary and what problem it solves.]
+
+## Proposal
+
+[A precise description of the proposed changed.]
```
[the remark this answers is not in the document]
rfratto: If a design document is requested, it usually indicates that a high level of detail is required, where you consider all aspects of what you are suggesting.  

Do you have a suggestion for how else this could be phrased?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+* Author(s): [List of proposal authors]
+* Last updated: [Date]
+* Original issue: https://github.com/grafana/alloy/issues/NNNN
+
+## Abstract
+
+[A short summary of the proposal.]
+
+## Background
```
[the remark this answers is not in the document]
rfratto: I can change this, but I'm not sure I fully understand why the header name makes a significant difference to the template, since the description of the section describes that a problem statement should be included. Can you help me understand more before I change it?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+
+[An introduction of the necessary context and the proposal being solved by the proposed change. Background should include why the proposal is necessary and what problem it solves.]
+
+## Proposal
+
+[A precise description of the proposed changed.]
+
+## Rationale
+
+[A discussion of alternative approaches and the trade offs, advantages, and disadvantages of the specified approach.]
```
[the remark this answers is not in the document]
rfratto: Similar to my comment above, I can change this but I'd like to understand more what impact a more rigid structure will have compared to a looser structure that has the same content.

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: I see, I was confused cause until this point in the doc we had two nouns "design doc PR" and "proposal issue", but now we have "proposal PR" ;) 

Maybe you can say something like:
> The approval state of the design document PR is used to denote whether the proposal is approved or declined and is treated as the source of truth. The design document contents should not include any indication of the approval status.

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: I think it's a nudge in the right direction and if someone skims through the template or gets honestly lost in the editorial process, then it's easier to notice that a section titled Problem doesn't describe the problem than it is to notice that a section titled Background doesn't describe the problem.

docs/design/908-proposal-process.md
rfratto: I don't expect the proposal review group to force conversations to happen in a specific order, especially as they only review proposals once a week.

But if someone brings up a concern that is not related to a proposal (such as "can we also have for loops" in a proposal for conditionals) or there are multiple related concerns that can be unified into a single theme (I don't have a concrete example for this unfortunately), the proposal review group may redirect things.

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: I wouldn't want it to be _all_ aspects, because then we're basically talking about code. Finding the right level of detail can often be an iterative process, but design docs are typically used to address high-level concerns. 

I'd prefer saying "A description of proposed changes with a level of detail sufficient to evaluate the tradeoffs and alternatives."

docs/design/908-proposal-process.md
rfratto: I don't think "Accepted" means that the accepted solution will be accepted forever; software will continue to evolve over time. I think it's fine if a new proposal supersedes a previously accepted one; that doesn't mean the older proposal becomes rejected. 

I don't anticipate superseding proposals to be frequent, or to happen shortly after the superseded proposal gets accepted. If someone does manage to supersede a previously accepted proposal shortly after that proposal was accepted, I believe that points to a failing in the consensus process where the conversation failed to identify a better solution.

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: Yeah, we can lock conversation on the proposal issue, so the comments won't happen. But we can still use labels etc to track status using the issue ticket maybe? And the PR being just a secondary thing.

image

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: In a similar way to above, I think it's a nudge in a right direction and serves as an extra reminder to include the most relevant and valuable content in the right place. It sort of works like a checklist.

docs/design/908-proposal-process.md
thampiotr: It's not too important here, more of a nit, but I think that a task to keep discussions focused and preventing derailing is something that will be very hard to be effective at.

docs/design/908-proposal-process.md
thampiotr: Sounds good!

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
rfratto: I'm not sure about this; with closing an issue, the author of the proposal can close it immediately themselves, but only maintainers can lock an issue once they notice a proposal has moved to a design document. I wonder if this might cause too much extra work for maintainers, and risks someone forgetting to lock an issue. 

What do you think about trying the process in the current state (where discussion moves to a PR by closing the original issue, if a design document is requested), seeing how it works in practice, and then making changes if necessary?

docs/design/908-proposal-process.md
```diff
@@ -0,0 +1,410 @@
...
+3. *Design document*: If requested, a design document is written, and the PR
+   for the design document becomes the new home for discussing the proposal.
+
+4. *Consensus*: Once comments and changes have slowed down, a final public
+   discussion aims to reach consensus for one of two outcomes:
+
+   * Accept the proposal, or
+   * decline the proposal.
+
+[new-proposal]: https://github.com/grafana/alloy/issues/new?assignees=&labels=proposal&projects=&template=proposal.yaml
```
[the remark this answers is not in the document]
thampiotr: Sounds good! And yeah, good thinking, people that click on New Issue / Proposal may actually have a feature request or something different, so maybe deciding whether a new issue is an actual proposal should happen later?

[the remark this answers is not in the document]
rfratto: > And yeah, good thinking, people that click on New Issue / Proposal may actually have a feature request or something different, so maybe deciding whether a new issue is an actual proposal should happen later?

Yeah, it's fine if someone starts with a non-proposal, and as the design document states, any issue can be turned into a proposal just by adding the `proposal` label :)

docs/design/908-proposal-process.md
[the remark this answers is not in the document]
thampiotr: Sounds good!

rfratto:
It sounds like we're at a rough consensus with the current state of this proposal, so I'll move this to **Likely Accept**. This marks the start of the final comment period for consensus to be changed.

If consensus hasn't changed by Friday, June 14, this proposal will become Accepted.

cc @erikbaranowski @tpaschalis @mattdurham @wildum @thampiotr @ptodev

thampiotr (approved):
Happy to trial this. I'd like to be careful not to generate too much work for the proposers and maintainers with this. I'm sure we can adjust in the future if this becomes the case.

tpaschalis (approved):

rfratto:
No change in consensus, so this proposal is now **Accepted.**

I will merge this and follow up on the next steps to roll this process out to other proposals.

Issue #399, opened 2023-07-26
clustering: allow cluster-aware component scheduling


tpaschalis:
### Request

We've discussed about introducing a second use case for clustering; fine-grained component scheduling.

This would allow components to opt-in to a new mechanism which would enable running N replicas of a component inside of a cluster, rebalancing whenever the cluster state changes.

### Use case

* Running exporters only on a subset of peers
* Running service discovery centrally and distributing the results within other peers

QuentinBisson:
@tpaschalis I think it could solve such issues https://github.com/grafana/alloy/issues/3044 as the "main" componentwould take care about creating the list of targets to scrape and would distribute the tailing to the others.

Do you know if there is any work towards that direction?

PR #917, opened 2024-05-23
Add a self pruning layer to gcp exporter delta stores

kgeckhart:
#### PR Description
The GCP exporter tracks delta -> counter metric accumulations in memory over the life cycle of the alloy instance it runs on. When used in clustering mode this can cause issues,

- Only one alloy instance will be incrementing counters at any given time
- All alloy instances will maintain any counter state they have because the exporter only prunes when it is being scraped
- ✅ target flips from an alloy instance with counter state -> one with no counter state (prometheus treats this as a counter reset)
- ❌ target flips from an alloy instance with counter state -> one with previous counter state 

This PR introduces a wrapper over-top of the the delta counter stores which does periodic pruning in the event the delta store is not in use. Since we cannot infer if the exporter instance is in use it will slowly prune this over the delta store TTL window which is 30 minutes by default. This at closes the window of time in which we are open to an incorrect data scenario in the event targets rebalance. 

This is a work around until https://github.com/grafana/alloy/issues/399 is completed at which point each alloy instance should only be maintaining an instance of the exporter if it's actually in use.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/static/integrations/gcp_exporter/self_pruning_store.go
+++ b/internal/static/integrations/gcp_exporter/self_pruning_store.go
@@ -14,7 +14,7 @@ import (
 )
 
 // Five minutes in seconds for Unix comparisons
-const DefaultPruneWindow int64 = 5 * 60 * 60
+const FiveMinutePruneWindow int64 = 5 * 60
 
 type CounterOrHistogramStore[T CounterOrHistogram] interface {
 	Increment(metricDescriptor *monitoring.MetricDescriptor, currentValue *T)
@@ -27,7 +27,6 @@ type CounterOrHistogram interface {
 
 type SelfPruningDeltaStore[T CounterOrHistogram] struct {
 	wrapping                     CounterOrHistogramStore[T]
-	pruneWindow                  int64
 	mux                          sync.Mutex
 	trackedMetricDescriptorNames map[string]struct{}
 	lastListOperationTime        atomic.Int64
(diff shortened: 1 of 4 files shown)


internal/static/integrations/gcp_exporter/self_pruning_store.go
```diff
@@ -0,0 +1,110 @@
...
+		s.trackedMetricDescriptorNames[metricDescriptorName] = struct{}{}
+	} else if ok && len(result) == 0 {
+		delete(s.trackedMetricDescriptorNames, metricDescriptorName)
+	}
+
+	return result
+}
+
+func (s *SelfPruningDeltaStore[T]) Prune(ctx context.Context) {
+	now := clock.Now(ctx).Unix()
```
mattdurham, line 85: I assume we are using this over time so we can mock the tests below and adjust time as needed?

internal/static/integrations/gcp_exporter/self_pruning_store.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Increment delegates to the wrapped store
+// We do not track metric descriptors from here to avoid more locking in a high throughput function
+func (s *SelfPruningDeltaStore[T]) Increment(metricDescriptor *monitoring.MetricDescriptor, currentValue *T) {
+	s.wrapping.Increment(metricDescriptor, currentValue)
+}
+
+// ListMetrics delegates to the wrapped store and updates tracking for the metricDescriptorName based on the results
+func (s *SelfPruningDeltaStore[T]) ListMetrics(metricDescriptorName string) []*T {
+	s.lastListOperationTime.Store(time.Now().Unix())
```
mattdurham, line 68: Should we also use clock here for consistency?

internal/static/integrations/gcp_exporter/self_pruning_store.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus-community/stackdriver_exporter/collectors"
+	"github.com/tilinna/clock"
+	"go.uber.org/atomic"
+	"google.golang.org/api/monitoring/v3"
+)
+
+// Five minutes in seconds for Unix comparisons
+const DefaultPruneWindow int64 = 5 * 60 * 60
```
mattdurham, line 17: Is this right? Wouldn't this be 300 minutes?

kgeckhart: 😆 🤦 yeah you're right

kgeckhart: I refactored the tests + pruning window implementation so that the tests would fail from this problem. Pruning window is hard coded now, originally I was going to try to use it to get around clock mocking but found the clock implementation used elsewhere already.

kgeckhart:
@mattdurham could you take another pass at this?

mattdurham:
Sorry was on vacation but back now, looking

mattdurham (approved):
lgtm

PR #919, opened 2024-05-23
ci: explore using ReFS drive to speed up Windows CI jobs

rfratto:


rfratto:
While there are reports of using ReFS drives being faster, there's a few issues preventing us from using it right now:

1. Creating a ReFS drive (the `New-VHD` command) doesn't work with `windows-2019`. My understanding is that the Windows version must match the Windows version of Docker images, which is `windows-2019`; upgrading may not be a choice right now. 

2. Commands like `go build` and `docker build` still have on-disk caches on the normal drive, where they don't take advantages of the performance boosts. While it may be possible to reconfigure the Docker and Go to place caches on the mounted drive, this would be somewhat of a large burden, and issue 1 prevents this from being productive anyway.

Closing for now, but maybe in the future all runners use ReFS by default and things will be faster.

PR #930, opened 2024-05-27
Update Agent Static to Alloy migration

clayton-cornell:
#### PR Description
Update Agent Static to Alloy migration - add Alloy installed prerequisite.

Remove bullet about being familiar with Components. Static migration is the only place it was mentioned like this and it seems unnecessary. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
Makes sense to me, I can't remember if that line was there on purpose but it doesn't seem to be relevant for a migration.

PR #935, opened 2024-05-28
Update windows install to add DisplayVersion registry key

EnessenE:
#### PR Description
The current name includes the version id but it is not available as a seperate registry key. This PR makes the installer add "DisplayVersion" so the version properly shows up in Windows and makes version extraction easier.

image

(For more info here: https://learn.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key)

N/A

#### Notes to the Reviewer
This is mainly annoying in automatic deployments as the version has to be extracted from the name or in another way. (yes a bit of an oversimplification)

#### PR Checklist

- [ x ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,7 @@ Main (unreleased)
 
 - Updated Prometheus dependency to [v2.51.2](https://github.com/prometheus/prometheus/releases/tag/v2.51.2) (@thampiotr)
 
-- Updated Windows install script to seperate DisplayName and DisplayVersion into seperate tags
+- Updated Windows install script to add DisplayVersion into registory on install
 
 v1.1.0
 ------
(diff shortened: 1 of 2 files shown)


dehaansa:
I'm doing some github archaeology and looking through older PRs - first of all, sorry that this contribution didn't get any attention at the time. If you're still interested in contributing this please update the PR from main to get CI to pass and we should able to merge this in quickly.

EnessenE:
Hello, archaeology requires us to put barriers around these olders PRs for about a year for us to clearly extract any artifacts.

In all seriousness, let me see what I can do

EnessenE:
It should be up-to-date now

dehaansa:
Thanks for updating - it looks like the CLA has also been reset since you contributed,  if you can follow the prompt in the comment from the CLA bot, I'll mark this ready to merge after CI is satisfied.

dehaansa (approved):

PR #937, opened 2024-05-28
remotecfg: set lower poll_frequency limit

tpaschalis:
#### PR Description

This PR adds a lower limit for the poll_frequency used in the remotecfg block so that Alloy instances don't accidentally hammer a server implementation. I think that's a good starting point but this value can be fine-tuned in the future. 

No issue filed.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/config-blocks/remotecfg.md
+++ b/docs/sources/reference/config-blocks/remotecfg.md
@@ -50,7 +50,7 @@ If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a r
 The `id` and `metadata` fields are used in the periodic request sent to the
 remote endpoint so that the API can decide what configuration to serve.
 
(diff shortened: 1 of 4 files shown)


erikbaranowski (approved):

docs/sources/reference/config-blocks/remotecfg.md
```diff
@@ -50,6 +50,8 @@ If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a r
 The `id` and `metadata` fields are used in the periodic request sent to the
 remote endpoint so that the API can decide what configuration to serve.
 
+The `poll_frequency` must be set to least "1s".
```
clayton-cornell, line 53:
```suggestion
The `poll_frequency` must be set to at least `"1s"`.
```

PR #940, opened 2024-05-28
Bump node version to 18 in build image

wildum:
The live debugging front-end uses [Grafana React Components](https://developers.grafana.com/ui/latest/index.html?path=/docs/docs-overview-intro--docs) which requires Node > 18.


tpaschalis (approved):
🙈

PR #941, opened 2024-05-28
example: install jb dependencies before provisioning

rfratto:
grafana/alloy#808 introduced dependencies for the mixin, which accidentally broke the example environment; as the mixin originally didn't have dependencies, the example environment didn't install them before provisioning dashboards.

This commit runs jb to locally install dependencies before provisioning dashboards.

cc @gaantunes


gaantunes (approved):
LGTM.

PR #943, opened 2024-05-28
feat(vcs): bubble up SSH key conversion error for better debugging experience

### LINKED ISSUE #914: Import git ssh error silenced
The SSH key conversion does not surface the error: https://github.com/grafana/alloy/blob/main/internal/vcs/auth.go#L29

When this happens, the given ssh key will be ignored and the go-git will try to use the ssh-agent instead: https://github.com/go-git/go-git/blob/master/plumbing/transport/ssh/common.go#L39

This is not good:

- if the key that I pass is wrong but the ssh-agent is correctly setup, then it will work but I won't know that I passed wrong information. It's also not using the auth method that I wanted.
- if the key is wrong and the ssh-agent is not set, then I will get an error related to the ssh-agent. I won't know what's wrong my ssh key which makes it super hard to debug

hainenber:

Closes #914 

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@ Main (unreleased)
 
 - Updated Prometheus dependency to [v2.51.2](https://github.com/prometheus/prometheus/releases/tag/v2.51.2) (@thampiotr)
 
+- Bubble up SSH key conversion error to facilitate failed `import.git`. (@hainenber)
+
 
 v1.1.0
 ------
--- a/internal/vcs/auth.go
+++ b/internal/vcs/auth.go
@@ -16,20 +16,23 @@ type GitAuthConfig struct {
 
 // Convert converts HTTPClientConfig to the native Prometheus type. If h is
 // nil, the default client config is returned.
-func (h *GitAuthConfig) Convert() transport.AuthMethod {
+func (h *GitAuthConfig) Convert() (transport.AuthMethod, error) {
 	if h == nil {
-		return nil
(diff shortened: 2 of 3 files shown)


internal/vcs/auth.go
```diff
@@ -16,20 +16,23 @@ type GitAuthConfig struct {
...
 
 	if h.BasicAuth != nil {
-		return h.BasicAuth.Convert()
+		return h.BasicAuth.Convert(), nil
 	}
 
 	if h.SSHKey != nil {
-		key, _ := h.SSHKey.Convert()
-		return key
+		key, err := h.SSHKey.Convert()
```
wildum, line 29: can be simplified by having directly `return h.SSHKey.Convert()` here

CHANGELOG.md
```diff
@@ -91,6 +91,8 @@ Main (unreleased)
 
 - Updated Prometheus dependency to [v2.51.2](https://github.com/prometheus/prometheus/releases/tag/v2.51.2) (@thampiotr)
 
+- Bubble up SSH key conversion error to facilitate failed `import.git`. (@hainenber)
+
```
wildum, line 95: I would actually put it in bug fixes

wildum (commented):
Thanks for taking care of this issue!

internal/vcs/auth.go
hainenber: Done in HEAD :D

PR #955, opened 2024-05-29
feat(common/loki): switch Snappy compression from `golang/snappy` to `klauspost/compress/snappy` for better compression

hainenber:
#### PR Description
I discovered that Snappy compression could be done from `github.com/klauspost/compress/snappy`. It's advertised as having "better compression" so I think the switch could be a performance win.

In addition, there's still usage of old `Shopify/sarama`, which is `IBM/sarama` now. 

Hijack this PR are also replacements of `golang.org/x/exp/maps` and `golang.org/x/exp/slices` to stdlib's equivalences.

N/A

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,9 @@ Main (unreleased)
 
 - Updated Prometheus dependency to [v2.51.2](https://github.com/prometheus/prometheus/releases/tag/v2.51.2) (@thampiotr)
 
+- Switch from `github.com/golang/snappy` to `github.com/klauspost/compress/snappy` for 
+  better compression in Loki clients. (@hainenber)
+
 v1.1.0
 ------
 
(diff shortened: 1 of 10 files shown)


rfratto:
It looks like both [Prometheus]() and [Loki]() have explored switching to klauspost/compress/snappy, but found that compression performs worse in those cases. 

Those were done at the storage-level and not at the client-side, so that's not necessarily an indication that it'll perform worse in Alloy, but definitely an indication that we should add benchmarks and validate.

dehaansa:
Closing this due to age & as it's not a clear improvement (given the comparisons in Loki and Prometheus referenced in a previous comment). If you're interested in continuing to pursue this change feel free to re-open or open a new PR! Thanks.

PR #961, opened 2024-05-30
remotecfg: add jitter to polling frequency

tpaschalis:
#### PR Description
This PR introduces a new `jitter` helper package which exports a time.Ticker-style interface with a period that includes jittering

No issue filed.

#### Notes to the Reviewer
I'm still looking to verify its correctness so I'll leave this as draft for the time being, but any comments are welcome!

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -33,6 +33,8 @@ func getHash(in []byte) string {
 	return fmt.Sprintf("%x", fnvHash.Sum(nil))
 }
 
+const baseJitter = 100 * time.Millisecond
+
 // Service implements a service for remote configuration.
 // The default value of ch is nil; this means it will block forever if the
 // remotecfg service is not configured. In addition, we're keeping track of
@@ -131,7 +133,7 @@ func New(opts Options) (*Service, error) {
 
 	return &Service{
 		opts:   opts,
-		ticker: jitter.NewTicker(math.MaxInt64-100*time.Millisecond, 100*time.Millisecond),
+		ticker: jitter.NewTicker(math.MaxInt64-baseJitter, baseJitter), // first argument is set as-is to avoid overflowing
 	}, nil
 }
 
@@ -230,7 +232,7 @@ func (s *Service) Update(newConfig any) error {
 	// it. Make sure we stop everything gracefully before returning.
 	if newArgs.URL == "" {
 		s.mut.Lock()
-		s.ticker.Reset(math.MaxInt64 - 100*time.Millisecond)
+		s.ticker.Reset(math.MaxInt64 - baseJitter) // avoid overflowing
 		s.asClient = noopClient{}
 		s.args.HTTPClientConfig = config.CloneDefaultHTTPClientConfig()
 		s.mut.Unlock()
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -20,6 +20,7 @@ import (
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
 	"github.com/grafana/alloy/internal/service"
+	"github.com/grafana/alloy/internal/util/jitter"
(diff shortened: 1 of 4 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -131,7 +131,7 @@ func New(opts Options) (*Service, error) {
 
 	return &Service{
 		opts:   opts,
-		ticker: time.NewTicker(math.MaxInt64),
+		ticker: jitter.NewTicker(math.MaxInt64-100*time.Millisecond, 100*time.Millisecond),
```
erikbaranowski, line 134: This was a bit confusing in that we are creating a super long ticker to start but I see it mirrors the old implementation. I assume this prevents the service from doing anything in the case it isn't actually in use. 

nit: Can the first argument be just `math.MaxInt64`?

internal/service/remotecfg/remotecfg.go
```diff
@@ -230,8 +230,7 @@ func (s *Service) Update(newConfig any) error {
 	// it. Make sure we stop everything gracefully before returning.
 	if newArgs.URL == "" {
 		s.mut.Lock()
-		s.ch = nil
-		s.ticker.Reset(math.MaxInt64)
+		s.ticker.Reset(math.MaxInt64 - 100*time.Millisecond)
```
erikbaranowski, line 233: nit: Can we put 100*time.Millisecond into a constant?

internal/util/jitter/jitter.go
```diff
@@ -0,0 +1,100 @@
...
+// time with a period specified by `d` adjusted by a pseudorandom jitter in
+// the range of [d-j, d+j).
+// Following the behavior of time.Ticker, we use a 1-buffer channel, so if the
+// client falls behind while reading, we'll drop ticks on the floor until the
+// client catches up.
+// Callers have to make sure that both d and d-j, d+j are valid positive int64
+// values (non-negative and non-overflowing).
+// Use Stop to release associated resources and the Reset methods to modify the
+// duration and jitter.
+func NewTicker(d time.Duration, j time.Duration) *Ticker {
```
erikbaranowski, line 29: can we come up with an alternative to `d` and `j` for these values throughout? Maybe `frequency` and `jitter`? I'm open other alternatives too.

internal/util/jitter/jitter.go
```diff
@@ -0,0 +1,100 @@
...
+	t.mut.Lock()
+	t.d = d
+	t.mut.Unlock()
+	t.reset <- struct{}{}
+}
+
+// Reset stops the Ticker, resets its jitter to the specified argument and
+// re-calculates the period with the new jitter.
+// The next tick will arrive after the new period elapses.
+func (t *Ticker) ResetJitter(d time.Duration) {
```
erikbaranowski, line 82: I don't see this used outside of testing so consider nuking.

erikbaranowski (approved):
Some nits to consider but looks really solid.

internal/service/remotecfg/remotecfg.go
tpaschalis: Yes; it's the easiest way I found to not have two separate code paths; have a ticker that is actually going to tick after `219250468-12-04 15:30:07 +0000 UTC` which I think is a long way into the future 😅.

The reason why the first argument is not just MaxInt64, is that this implementation chooses a jitter value in the `[d-j, d+j)` interval, and that would lead in an overflow and make the ticker panic.

internal/util/jitter/jitter.go
tpaschalis: Yes. FYI, the reason why I went with a ticker and _not_ time.Sleep is to allow resetting the ticker dynamically. The sleep-based implementation just shut the goroutine down and wouldn't allow it to be responsive to any changes.

rfratto:
@tpaschalis Our Windows tests have been failing starting with this PR: https://drone.grafana.net/grafana/alloy/1662/2/2

Can you take a look when you have a moment?

Issue #954, opened 2024-05-29
Integration tests for `prometheus.exporter.snmp`


ptodev:
### Request

It would be good to have SNMP integration tests in `internal/cmd/integration-tests/tests/`. The test would start an [SNMP simulator](https://github.com/lextudio/snmpsim), and check if metrics are received by Mimir. This is similar to other tests that we have like the [Redis one](https://github.com/grafana/alloy/tree/main/internal/cmd/integration-tests/tests/redis).

There are also [public SNMP endpoints](https://www.pysnmp.com/snmp-simulation-service) which could be used for manual testing, but it wouldn't be a good practice to use them in a CI. For example, this config calls a public endpoint:

```
prometheus.exporter.snmp "snmp_metrics" {
  target "testing1" {
    address = "demo.pysnmp.com"
    module = "if_mib"
    auth = "public_v2"
  }
}

prometheus.scrape "snmp_metrics" {
  targets    = prometheus.exporter.snmp.snmp_metrics.targets
  forward_to = [prometheus.remote_write.snmp_metrics.receiver]
  scrape_interval = "5s"
  scrape_timeout = "500ms"
}

prometheus.remote_write "snmp_metrics" {
  ...
}
```

The config for `prometheus.exporter.snmp` could be as simple as possible. The main goal of this issue is to bootstrap the test framework. The simulator is in Python and AFAIK it doesn't come in Docker containers, so I'm not sure how we will run it in the Alloy CI.

### Use case

Improving the test coverage :)

Issue #289, opened 2023-12-15
Allow flow components to provide names via a string instead of river identifier


erikbaranowski:
### Request

There are some additional components to prometheus.exporter.blackbox that are using block labels as argument values. Review and update each that should receive a similar change to grafana/agent#5940.

- prometheus.exporter.cloudwatch
- prometheus.exporter.snmp
  - calls to common.SanitizeIdentifierPanics should be removed in the static to flow converter for this component: https://github.com/grafana/agent/blob/main/converter/internal/staticconvert/internal/build/snmp_exporter.go
- pyroscope.scrape

### Use case

River identifiers are fairly restrictive. https://grafana.com/docs/agent/next/flow/config-language/syntax/#identifiers

This prevents someone from having an otherwise valid target name like job-name and is a stumbling block when converting from static mode to flow mode since static mode does not have this restriction.

ptodev:
I'll close this since for a long time I haven't heard of this sort of thing being an issue. If there's a component for which it's a probme in practice, then we can work on it on a case by case basis.

Issue #273, opened 2024-01-18
Embedded multi-target Prometheus exporters can't be provided discovered targets 


rfratto:
Multi-target exporters (such as snmp_exporter) can have a dynamic set of targets passed to them through [the scrape config](https://github.com/prometheus/snmp_exporter#prometheus-configuration). 

However, in Grafana Agent, our embedded version of these exporters only accept [hard-coded](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.snmp/#target-block) targets, preventing the kind of dynamic behaviour Prometheus users would expect.  

I see this as a bug: using the embedded multi-target exporter should be equally as capable as using the external multi-target exporter. 

We should investigate and fix this issue in all embedded multi-target exporters.

PR #969, opened 2024-05-31
prometheus.exporter.snmp dynamic targets support

wildum:
#### PR Description

Add an argument `targets` to support passing targets at runtime to the exporter.

Also add integration tests to cover both ways of passing targets.

Related to https://github.com/grafana/alloy/issues/273
Related to https://github.com/grafana/alloy/issues/289
Fixes https://github.com/grafana/alloy/issues/954

#### Notes to the Reviewer

I added an additional argument because replacing the target block would be a breaking change.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Create an issue to eventually delete the target block when switching to a major version

### DIFF
--- a/docs/sources/reference/components/prometheus.exporter.snmp.md
+++ b/docs/sources/reference/components/prometheus.exporter.snmp.md
@@ -34,6 +34,7 @@ Omitted fields take their default values.
 | ------------- | -------------------- | ------------------------------------------------ | ------- | -------- |
 | `config_file` | `string`             | SNMP configuration file defining custom modules. |         | no       |
 | `config`      | `string` or `secret` | SNMP configuration as inline string.             |         | no       |
+| `targets`     | `list(map(string))`  | SNMP targets.                                    |         | no       |
 
 The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
 Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter#generating-configuration) for details on how to generate a configuration file.
@@ -45,6 +46,14 @@ The `config` argument must be a YAML document as string defining which SNMP modu
 - `remote.http.LABEL.content`
 - `remote.s3.LABEL.content`
 
+The `targets` argument serves as an alternative to the [target][] block. This is useful when SNMP targets are supplied by another component.
+The following labels can be set to a target:
+* `name` is the name of the target (required).
+* `address` is the address of SNMP device (required).
+* `module` is the SNMP module to use for polling.
+* `auth` is the SNMP authentication profile to use.
+* `walk_params` is the config to use for this target.
+
 ## Blocks
 
 The following blocks are supported inside the definition of
@@ -52,7 +61,7 @@ The following blocks are supported inside the definition of
 
 | Hierarchy  | Name           | Description                                                 | Required |
 | ---------- | -------------- | ----------------------------------------------------------- | -------- |
-| target     | [target][]     | Configures an SNMP target.                                  | yes      |
+| target     | [target][]     | Configures an SNMP target.                                  | no      |
 | walk_param | [walk_param][] | SNMP connection profiles to override default SNMP settings. | no       |
 
 [target]: #target-block
@@ -131,6 +140,7 @@ prometheus.exporter.snmp "example" {
         retries = "2"
     }
 }
+
 // Configure a prometheus.scrape component to collect SNMP metrics.
 prometheus.scrape "demo" {
     targets    = prometheus.exporter.snmp.example.targets
@@ -169,6 +179,7 @@ prometheus.exporter.snmp "example" {
         retries = "2"
     }
 }
+
 // Configure a prometheus.scrape component to collect SNMP metrics.
 prometheus.scrape "demo" {
     targets    = prometheus.exporter.snmp.example.targets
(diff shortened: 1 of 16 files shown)


docs/sources/reference/components/prometheus.exporter.snmp.md
```diff
@@ -187,13 +198,79 @@ prometheus.remote_write "demo" {
 }
 ```
 
+This example uses the alternative way to pass targets:
```
erikbaranowski, line 201: I would be in favor of just updating the existing example[s] since we plan to deprecate the old way. Someone can still find/use them that way but better to not model it for them.

erikbaranowski (commented):
Thanks for taking this on. A couple general questions:

- When I was thinking about how to do this so it wasn't a breaking change, I pictured adding a name attribute to the target block. This attribute would be used if set to override the block label. Did you consider this or was there a downside to it?
- I would love to see the converter code make use of the new field so that
  - it converts people to the way of the future
  - the conversion doesn't have to manage translating to a valid river identifier which has some potential edge cases

wildum:
> * When I was thinking about how to do this so it wasn't a breaking change, I pictured adding a name attribute to the target block. This attribute would be used if set to override the block label. Did you consider this or was there a downside to it?

I was unaware of this issue, it's a happy accident that it fixes it :) The problem with the block is that you cannot have a dynamic list of targets. Every time you want to add a new target, you need to update your config to add a new block. This is not very practical for users that have a big list of targets that changes a lot (see the feedback from Brainlab). That's why I decided to get rid of the block for a `prometheus.scrape` style.

> * I would love to see the converter code make use of the new field so that
>   
>   * it converts people to the way of the future
>   * the conversion doesn't have to manage translating to a valid river identifier which has some potential edge cases

I will make the change :)

docs/sources/reference/components/prometheus.exporter.snmp.md
wildum: I'm afraid to scare users if I advertise it as deprecated now. We just made some big changes with Alloy and that upsets some people. That's why I put it as "an alternative way of doing it" for now and kept the old way with examples so that users are not lost. The next major release is far in the future, maybe we should only start deprecating things once we arrive close to it?

internal/converter/internal/staticconvert/internal/build/snmp_exporter.go
```diff
@@ -64,15 +64,15 @@ func (b *ConfigBuilder) appendSnmpExporterV2(config *snmp_exporter_v2.Config) di
...
-			Name:       common.SanitizeIdentifierPanics(t.Name),
-			Target:     t.Target,
-			Module:     t.Module,
-			Auth:       t.Auth,
-			WalkParams: t.WalkParams,
-		}
+	var targets snmp.TargetsList
+	for _, t := range config.SnmpTargets {
+		target := make(map[string]string)
+		target["name"] = common.SanitizeIdentifierPanics(t.Name)
```
erikbaranowski, line 70: we should be able to remove the call to `common.SanitizeIdentifierPanics` now that we are not mapping to a block label. let's include in our test something that is invalid river to double check this.

docs/sources/reference/components/prometheus.exporter.snmp.md
```diff
@@ -45,14 +46,22 @@ The `config` argument must be a YAML document as string defining which SNMP modu
 - `remote.http.LABEL.content`
 - `remote.s3.LABEL.content`
 
+The `targets` argument serves as an alternative to the [target][] block. This is useful when SNMP targets are supplied by another component.
+The following labels can be set to a target:
+* `name` is the name of the target (required).
+* `address` is the address of SNMP device (required).
+* `module` is the SNMP module to use for polling.
+* `auth` is the SNMP authentication profile to use.
+* `walk_params` is the config to use for this target.
```
clayton-cornell, line 55:
```suggestion
The `targets` argument is an alternative to the [target][] block. This is useful when SNMP targets are supplied by another component.
The following labels can be set to a target:
* `name`: The name of the target (required).
* `address`: The address of SNMP device (required).
* `module`: The SNMP module to use for polling.
* `auth`: The SNMP authentication profile to use.
* `walk_params`: The config to use for this target.
```

docs/sources/reference/components/prometheus.exporter.snmp.md
```diff
@@ -187,13 +198,79 @@ prometheus.remote_write "demo" {
...
+}
+
+// Configure a prometheus.scrape component to collect SNMP metrics.
+prometheus.scrape "demo" {
+    targets    = prometheus.exporter.snmp.example.targets
+    forward_to = [ /* ... */ ]
+}
+```
+
+This example uses the [`local.file` component][file] to read targets from a YAML file and feed them to the prometheus.exporter.snmp component:
```
clayton-cornell, line 238:
```suggestion
This example uses the [`local.file` component][file] to read targets from a YAML file and send them to the prometheus.exporter.snmp component:
```

internal/converter/internal/staticconvert/internal/build/snmp_exporter.go
wildum: good catch, I made the change and added a "." in the name in the yml file. The "." would be replace by a "_" if we would still use SanitizeIdentifierPanics

wildum:
@erikbaranowski @clayton-cornell fyi I added another example to the doc with the discovery.file. It's a lot of examples better too many examples than too few. I think that it will help the users

erikbaranowski (approved):
Awesome! thank you for the improvement to the converter along the way 🚀

docs/sources/reference/components/prometheus.exporter.snmp.md
```diff
@@ -187,13 +198,137 @@ prometheus.remote_write "demo" {
...
+    name: t2
+    module: default
+    auth: public_v2
+```
+
 Replace the following:
 
 - `PROMETHEUS_REMOTE_WRITE_URL`: The URL of the Prometheus remote_write-compatible server to send metrics to.
 - `USERNAME`: The username to use for authentication to the remote_write API.
 - `PASSWORD`: The password to use for authentication to the remote_write API.
```
clayton-cornell, line 326:
```suggestion
Replace the following:

- _`<PROMETHEUS_REMOTE_WRITE_URL>`_: The URL of the Prometheus remote_write-compatible server to send metrics to.
- _`<USERNAME>`_: The username to use for authentication to the remote_write API.
- _`<PASSWORD>`_: The password to use for authentication to the remote_write API.
```

This text needs to moved up around line 200 -it applies to the generic example there, and is out of place down here at line 322.

In the generic example (lines 191-195) there also needs to be a small change ( I can't comment there)... wrap the replaceable args with `<` and `>`.

```
endpoint {
        url = <PROMETHEUS_REMOTE_WRITE_URL>
        basic_auth {
            username = LUSERNAME>
            password = <PASSWORD>
```

clayton-cornell:
> It's a lot of examples better too many examples than too few. I think that it will help the users

More than one example is a good thing. :-)

wildum:
@erikbaranowski I found a bug: the Validate function was not called (this component is a bit special, it has its own UnmarshalAlloy function). I move the validate code there and remove the validation for exclusivity because passing an empty list of targets is a valid scenario since the list might be populated at runtime

clayton-cornell (approved):

wildum:
@erikbaranowski @clayton-cornell sorry to drag this down but I realized that it would be better not to allow users to pass targets via discovery components without the need to use a discovery.relabel to relabel the address. I made this change in my last commit

erikbaranowski (approved):

PR #972, opened 2024-06-02
chore(web/ui): use named imports for React

hainenber:
#### PR Description

React 18 now automatically injects JSX runtime during compilation so React import is no longer necessary. I also switch to named imports for currency and slightly better bundling.

I ran `yarn start`, `yarn build`. Everything stays identical to builds prior to this change.

N/A

#### Notes to the Reviewer
N/A


wildum (approved):
LGTM, thanks!

PR #975, opened 2024-06-03
Bump build image to v0.1.2 to support nodeJS 18

wildum:
The only change was to bump the nodeJS version to 18 to support grafana's react component used for the live debugging. 
See https://github.com/grafana/alloy/pull/940


mattdurham (approved):
LGTM

PR #977, opened 2024-06-03
Xray front

wildum:
THIS PR WILL BE MERGED TO A FEATURE BRANCH WHICH WILL RECEIVE A FEW OTHER PRs BEFORE BEING READY TO BE MERGED ON MAIN

This PR is the second part of the live debugging functionality.

It adds a live debugging page in the front-end:

https://github.com/grafana/alloy/assets/21154204/c0af2de9-83b3-43e9-a5fe-83f149b513a3

The live debugging uses a 5000 lines buffer to display the feed. Once the buffer is full, old lines are removed.
The user:
- can filter the buffer
- use a sample probability to reduce the throughput
- copy the content of the buffer to the clipboard
- pause/resume/clear the page
- enable/disable auto-scrolling

It supports modules.


jcreixell (approved):
LGTM, the changes are consistent with the previous code we wrote for the POC, which we tested thoroughly. I don't love the colors of the buttons on the top right, but they are functional and we can always make cosmetic improvements later. Looking forward to seeing this released!

PR #991, opened 2024-06-06
Port faro prefix measurement values

wildum:
Port of https://github.com/grafana/agent/pull/6810 with an added test

### DIFF
--- a/internal/component/faro/receiver/internal/payload/payload.go
+++ b/internal/component/faro/receiver/internal/payload/payload.go
@@ -220,7 +220,7 @@ type Measurement struct {
 	Context   MeasurementContext `json:"context,omitempty"`
 }
 
-// KeyVal representation of the exception object
+// KeyVal representation of the measurement object
 func (m Measurement) KeyVal() *KeyVal {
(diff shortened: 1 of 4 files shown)


tpaschalis (approved):

internal/component/faro/receiver/internal/payload/payload.go
```diff
@@ -238,6 +238,14 @@ func (m Measurement) KeyVal() *KeyVal {
 	}
 	MergeKeyVal(kv, m.Trace.KeyVal())
 	MergeKeyValWithPrefix(kv, KeyValFromMap(m.Context), "context_")
+
+	values := make(map[string]float64, len(m.Values))
+	for key, value := range m.Values {
+		values[key] = value
+	}
```
ptodev, line 245: Why are we copying this? Can we not just call `KeyValFromFloatMap(m.Values)`?

wildum: true, it's an artifact from a previous implementation that had a conversion there. Let me spawn a quick pr to remove this

PR #997, opened 2024-06-06
Add dynamic targets support to prometheus.exporter.blackbox

### LINKED ISSUE #273: Embedded multi-target Prometheus exporters can't be provided discovered targets

wildum:
#### PR Description

Add an argument targets to support passing targets at runtime to the exporter.

Also add integration tests to cover both ways of passing targets.

(similar to https://github.com/grafana/alloy/pull/969) 

Fixes #273 

#### Notes to the Reviewer

I added an additional argument because replacing the target block would be a breaking change.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [ ] Create an issue to eventually delete the target block when switching to a major version

### DIFF
--- a/docs/sources/reference/components/prometheus.exporter.blackbox.md
+++ b/docs/sources/reference/components/prometheus.exporter.blackbox.md
@@ -50,7 +50,8 @@ The `config` argument must be a YAML document as string defining which `blackbox
 - `remote.s3.LABEL.content`
 
 The `targets` argument is an alternative to the [target][] block. This is useful when blackbox targets are supplied by another component.
-The following labels can be set to a target:
+
+You can set the following labels to a target:
 * `name`: The name of the target to probe (required).
 * `address`: The address of the target to probe (required).
 * `module`: The blackbox module to use to probe.
--- a/docs/sources/reference/components/prometheus.exporter.blackbox.md
+++ b/docs/sources/reference/components/prometheus.exporter.blackbox.md
@@ -49,7 +49,8 @@ The `config` argument must be a YAML document as string defining which `blackbox
 - `remote.http.LABEL.content`
 - `remote.s3.LABEL.content`
 
-The `targets` argument is an alternative to the [target][] block. This is useful when blackbox targets are supplied by another component.
+You can't use both the `targets` argument and the [target][] block in the same configuration file.
+The `targets` argument must be used when blackbox targets cannot be passed as a target block because another component supplies them.
 
 You can set the following labels to a target:
 * `name`: The name of the target to probe (required).
--- a/docs/sources/reference/components/prometheus.exporter.blackbox.md
+++ b/docs/sources/reference/components/prometheus.exporter.blackbox.md
@@ -20,16 +20,25 @@ prometheus.exporter.blackbox "LABEL" {
 }
 ```
 
+or 
+
+```alloy
+prometheus.exporter.blackbox "LABEL" {
+  targets = TARGET_LIST
+}
+```
+
 ## Arguments
 
 The following arguments can be used to configure the exporter's behavior.
 Omitted fields take their default values.
 
-| Name                   | Type                 | Description                                                        | Default  | Required |
-| ---------------------- | -------------------- | ------------------------------------------------------------------ | -------- | -------- |
+| Name                   | Type                 | Description                                                      | Default  | Required |
+| ---------------------- | -------------------- | ---------------------------------------------------------------- | -------- | -------- |
 | `config_file`          | `string`             | `blackbox_exporter` configuration file path.                       |          | no       |
 | `config`               | `string` or `secret` | `blackbox_exporter` configuration as inline string.                |          | no       |
-| `probe_timeout_offset` | `duration`           | Offset in seconds to subtract from timeout when probing targets.   | `"0.5s"` | no       |
+| `probe_timeout_offset` | `duration`           | Offset in seconds to subtract from timeout when probing targets. | `"0.5s"` | no       |
+| `targets`              | `list(map(string))`  | Blackbox targets.                                                |          | no       |
 
 Either `config_file` or `config` must be specified.
 The `config_file` argument points to a YAML file defining which `blackbox_exporter` modules to use.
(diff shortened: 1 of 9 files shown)


docs/sources/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -40,6 +49,15 @@ The `config` argument must be a YAML document as string defining which blackbox_
 - `remote.http.LABEL.content`
 - `remote.s3.LABEL.content`
 
+The `targets` argument is an alternative to the [target][] block. This is useful when blackbox targets are supplied by another component.
+The following labels can be set to a target:
```
clayton-cornell, line 53:
```suggestion
You can set the following labels to a target:
```

Do we add the labels or set the labels?

Small change to make it more active. It's OK as-is too... this is just a small step towards making our docs active vs passive :-)

docs/sources/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -40,6 +49,15 @@ The `config` argument must be a YAML document as string defining which blackbox_
...
+* `name`: The name of the target to probe (required).
+* `address`: The address of the target to probe (required).
+* `module`: The blackbox module to use to probe.
+
+Any additional labels will be passed to the exported target of the component.
+
```
clayton-cornell, line 59:
```suggestion
The component passes any additional labels to the exported target.
```

Another small tweak to make it more active and clarify who does what.

docs/sources/reference/components/prometheus.exporter.blackbox.md
wildum: I will use add, it seems slightly better

clayton-cornell (approved):
Docs are good as-is. Approving for docs

docs/sources/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -40,6 +49,15 @@ The `config` argument must be a YAML document as string defining which blackbox_
 - `remote.http.LABEL.content`
 - `remote.s3.LABEL.content`
 
+The `targets` argument is an alternative to the [target][] block. This is useful when blackbox targets are supplied by another component.
```
erikbaranowski, line 52: Is there some wording we can add here to make it clear that the one of either target or targets must be used (but not both). I'm not sure what are the best words though. @clayton-cornell

internal/component/prometheus/exporter/blackbox/blackbox.go
```diff
@@ -118,15 +164,42 @@ func (a *Arguments) Validate() error {
 		return fmt.Errorf("invalid blackbox_exporter config: %s", err)
 	}
 
+	if len(a.Targets) != 0 && len(a.TargetsList) != 0 {
```
erikbaranowski, line 167: opposite to this validation scenario, is it invalid if both are length 0?

internal/component/prometheus/exporter/blackbox/blackbox.go
```diff
@@ -93,8 +98,49 @@ func (t TargetBlock) Convert() []blackbox_exporter.BlackboxTarget {
...
+		targets = append(targets, blackbox_exporter.BlackboxTarget{
+			Name:   target["name"],
+			Target: address,
+			Module: target["module"],
+		})
+	}
+	return targets
+}
+
+func (t TargetsList) convert() []BlackboxTarget {
```
erikbaranowski, line 123: Having `Convert` and `convert` on the same type with different returns is confusing here. Can we rename this one?

erikbaranowski (commented):
Looking good, I commented a few small things to go over.

internal/component/prometheus/exporter/blackbox/blackbox.go
wildum: yes because the targets list can be provided at runtime. It's a valid scenario to go from 0 targets to 2 targets and then back to 0.

internal/component/prometheus/exporter/blackbox/blackbox.go
wildum: yeah that's a bit annoying because we have several backbox target types. I renamed it to convertInternal. Not super happy with the name but it's better than having convert twice. Do you have a suggestion or is this ok for you?

docs/sources/reference/components/prometheus.exporter.blackbox.md
erikbaranowski: sounds like both can be length zero below so this may not be true. I'll trust you all to think about/sort this question, but I think everything else is good to go now on the PR.

erikbaranowski (approved):

docs/sources/reference/components/prometheus.exporter.blackbox.md
clayton-cornell: You could be really straight forward and direct about it and add something like:

`You can't use both the `targets` argument and the `target` block in the same configuration file.`

wildum: What about this:
```
You can't use both the `targets` argument and the `target` block in the same configuration file. The `targets` argument must be used when blackbox targets cannot be passed as a target block because another component supplies them.
```

clayton-cornell: Yes... that makes it very clear when and where to use the argument.

clayton-cornell:
New changes look good to me

PR #999, opened 2024-06-07
Live debugging

### LINKED ISSUE #986: Merge live debugging to main

wildum:
#### PR Description

This PR brings the live debugging feature to main.

Fixes #986 

#### Notes to the Reviewer

This PR is made from three parts that have been reviewed already:

- back-end: https://github.com/grafana/alloy/pull/797
- front-end: https://github.com/grafana/alloy/pull/977
- doc: https://github.com/grafana/alloy/pull/987

I added these two screenshots to the media folder:
ui_live_debugging_page
ui_component_detail_page_2

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Update the screenshots


internal/service/livedebugging/livedebugging.go
```diff
@@ -0,0 +1,88 @@
...
+		callbacks:            make(map[ComponentID]map[CallbackID]func(string)),
+		registeredComponents: make(map[ComponentName]struct{}),
+	}
+}
+
+func (s *liveDebugging) Publish(componentID ComponentID, data string) {
+	s.loadMut.RLock()
+	defer s.loadMut.RUnlock()
+	for _, callback := range s.callbacks[componentID] {
+		callback(data)
```
ptodev, line 57: I wonder if there'd be an issue and the component has crashed without deleting the callback. Should we detect that and clean up the callback?

ptodev (commented):
Looks great! I can't wait to try it!

internal/service/livedebugging/livedebugging.go
wildum: The components are not responsible for deleting callbacks, it's the job of the DebugCallbackManager which is only passed to the UI service. The callback is only deleted when the user quits the live debugging page of the component. What do you mean by crashed?

docs/sources/tasks/debug.md
```diff
@@ -70,12 +73,33 @@ The clustering page shows the following information for each cluster node:
...
+
+* Pause and clear the data stream.
+* Sample data and disable auto-scrolling to handle heavy loads.
+* Search through the data using keywords.
+* Copy the entire data stream to the clipboard.
+
+The format and content of the debugging data vary depending on the component type.
+
+{{< admonition type="note" >}}
+Live debugging is not yet available in all components.
```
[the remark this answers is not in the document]
wildum: That's a good point. I did it like this because I thought that the list could grow quickly and be annoying to keep up to date. But it might still be better than letting users try and be disappointed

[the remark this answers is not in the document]
rfratto: I think a good future improvement would be for the "Live debugging" button to only show when a component supports it. Can we track that in a ticket?

internal/service/livedebugging/livedebugging.go
rfratto, line 1: General nit: to shorten the names and reduce reptition, `Debug` can probably be removed from most of the names in this package. For example:

* `livedebugging.DebugCallbackManager` can be `livedebugging.CallbackManager` 
* `livedebugging.NewLiveDebugging` can be simply `livedebugging.New`

internal/service/livedebugging/livedebugging.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+func (s *liveDebugging) IsActive(componentID ComponentID) bool {
+	_, exist := s.callbacks[componentID]
```
rfratto, line 62: Do you need to grab the read lock here?

internal/component/prometheus/relabel/relabel.go
```diff
@@ -99,14 +104,22 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	debugDataPublisher, err := o.GetServiceData(livedebugging.ServiceName)
+	if err != nil {
+		return nil, err
+	}
+	debugDataPublisher.(livedebugging.DebugDataPublisher).Register(name)
```
rfratto, line 112: The call to register leaks; once components shut down there's nothing to unregister themselves from the map. This could be a problem when you're using remotecfg or an import where component IDs are changing frequently. 

I recommend a pull-based pattern here for registration:

* Have components that support live debugging implement an interface. It can be as simple as a marker interface like what we do for [syntax.Capsule](https://github.com/grafana/alloy/blob/fa024324795b4066f3a3ce82f25d9bcd5ec0fdbf/syntax/types.go#L49-L65).
* Then, instead of checking to see if a component was registered, you can use service.Host to identify whether a given component by ID implements that interface, similarly to what is done for [HTTP](https://github.com/grafana/alloy/blob/fa024324795b4066f3a3ce82f25d9bcd5ec0fdbf/internal/service/http/http.go#L280-L295).

This way the list of components doesn't leak because you're always looking up a component by ID to an interface which handles cleanup properly when components shut down.

internal/service/livedebugging/service.go
```diff
@@ -0,0 +1,52 @@
...
+	return s.liveDebugging
+}
+
+// Definition implements service.Service.
+func (*Service) Definition() service.Definition {
+	return service.Definition{
+		Name:       ServiceName,
+		ConfigType: nil, // livedebugging does not accept configuration
+		DependsOn:  []string{},
+		Stability:  featuregate.StabilityExperimental,
```
rfratto, line 39: What does the experimental flag do in this case? Is it actually unavailable if you're running Alloy with GA only? Does the debug button in the UI get hidden if live debugging isn't available? 

Since we don't cover UI functionality in backwards compatability, I think it might be easier if this starts out as GA, otherwise I think we need to do more legwork to make sure that live debugging is truly disabled everywhere when the stability level is set too high.

internal/web/api/api.go
```diff
@@ -107,3 +115,83 @@ func (a *AlloyAPI) getClusteringPeersHandler() http.HandlerFunc {
...
+				return
+			default:
+				if sampleProb < 1 && rand.Float64() > sampleProb {
+					return
+				}
+				// Avoid blocking the channel when the channel is full
+				select {
+				case dataCh <- data:
+				default:
+				}
```
rfratto, line 157: As a future improvement, we might want to report back to the user when data is dropped (because of a full buffer) so they know that not everything is showing when the sample rate is 100%.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Main (unreleased)
 
 - Add an `otelcol.exporter.kafka` component to send OTLP metrics, logs, and traces to Kafka.
 
+- Add a live debugging feature to the UI, including support for the `prometheus.relabel` component. (@wildum)
```
rfratto, line 22: I think this might be underselling it; can we explain what live debugging does? Something like:

> - Added "live debugging" to the UI. Live debugging streams data as they flow through components for debugging telemetry data. Individual components must be updated to support live debugging. (@wildum) 
> - Adding live debugging support for `prometheus.relabel`. (@wildum)

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
clayton-cornell: Is this something that will be added one component at a time... or batches of components?  Is there a pattern to what is available now, and what will be added later?  eg all prometheus.* first, then all loki.* next and so on.

We could start with a list now if it's short and adapt to something else later?

rfratto (commented):
Nice work! I have a few small comments and suggestions for future improvements. I'm really looking forward to this expanding into more components :)

docs/sources/tasks/debug.md
```diff
@@ -70,12 +73,33 @@ The clustering page shows the following information for each cluster node:
...
+{{< /admonition >}}
+
+
 ## Debugging using the UI
 
 To debug using the UI:
 
 * Ensure that no component is reported as unhealthy.
 * Ensure that the arguments and exports for misbehaving components appear correct.
+* Ensure that the debugging data meets your expectations.
```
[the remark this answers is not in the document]
clayton-cornell: I'm not sure that we should use quotes this way. 

If we need to refer to the page title it should be bold https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/#text-formatting

Is the live debugging the only element/data that we need to make sure meets expectations? Or are we looking at all debugging data?

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
wildum: There will be batches of components because for most it will be added on common codes. For example, adding all otel.processor.* will be done in one go. But the problem is that there can be exceptions and there is a risk that the list is unsync.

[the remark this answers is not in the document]
wildum: But that could be good enough for a start and Robert's proposal is reasonable and should not be too hard to implement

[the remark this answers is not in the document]
ptodev: We can open an issue to automate updating the docs? We already have similar processes in place to update docs in an automated way.

internal/service/livedebugging/livedebugging.go
ptodev: It's possible that a component stops abruptly for some reason while it's processing data. E.g. maybe it relies on a connection to some service, and it doesn't handle the connection very well so it `Run()` stops all of the sudden. In that case, would the callback unregistration be handled well, or would it leak? If it leaks, maybe we can clean it up the next time the callback is called?

Also, do you think there'd be an issue if a user has the live debugging page open, and then they update their config to exclude that component? What would happen in this case?

thampiotr:
Good stuff! I have a few comments from testing it out, not necessarily blockers, but I'd like to raise them here. 

0. I may remembering incorrectly, but I thought this was meant to be explicitly enabled in a config file somehow? The reason is that an attacker with HTTP access to Alloy will be able to snoop the data which was previously not exposed or encrypted on the way out. We should give users an option to have this disabled IMO.

1. There is a warning in dev console on Chrome, maybe unrelated to your changes, but sharing just in case:
image

2. I find that the button to enable Auto Scroll is tiny and hidden at the bottom, why not make it a normal button like others, next to Clear, Copy, etc?
image

3. When I filter the results and click Copy, I get all the data instead of just the filtered data. Is this intentional? I'd expect to only get what I have in the buffer.

4. Nit: I am not sure about using `=>` as a way to designate the mutation in prometheus.relabel. It may work for this component, but I think that other components may have more kinds of events (e.g. 'connected' or 'reloading targets' etc) and may end up with more logs-like data stream.... which now made me wonder if we could have implemented all of this using loggers, where you can peek into debug-level logs of any component on-demand. Would work with services too... kind a liking this idea now 😂 

Overall, this is great and looking forward to having this on more components.

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
clayton-cornell: It's going to be a pretty long list at some point if we name all components that have the feature enabled... automated or not. :thinking:

wildum:
> I find that the button to enable Auto Scroll is tiny and hidden at the bottom, why not make it a normal button like others, next to Clear, Copy, etc?

It's part of an imported react autoscroll component, I could not find a good way to bring it up and style it like the other buttons :(

wildum:
> Nit: I am not sure about using => as a way to designate the mutation in prometheus.relabel. It may work for this component, but I think that other components may have more kinds of events (e.g. 'connected' or 'reloading targets' etc) and may end up with more logs-like data stream.... which now made me wonder if we could have implemented all of this using loggers, where you can peek into debug-level logs of any component on-demand. Would work with services too... kind a liking this idea now

I think that these are two different things. The original idea of the x-ray feature is to see the data/the flow through the pipelines. Events data that you see in logs is not part of the plan. It does not replace logs but gives another tool that can be used alongside the others to troubleshoot Alloy.

internal/web/ui/src/pages/LiveDebugging.module.css
```diff
@@ -0,0 +1,106 @@
...
+  .debugLink .copyButton:hover {
+    background-color: rgb(44, 90, 176);
+    border-color: rgb(44, 90, 176);
+  }
+
+  
+  .logLine {
+    white-space: pre-wrap;
+    color: #24292e;
+  }
```
rfratto, line 98: really tiny nit: should the log lines use a monospace font?

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
wildum: Not a super big fan of going the extra step to automate the doc on that. Can we start with a manual list, group the components as @clayton-cornell suggested and once we have the support for quite a few components we get rid of the list? In the meantime, we implement @rfratto's suggestion to show only the button on components that support it?

wildum:
> I may remembering incorrectly, but I thought this was meant to be explicitly enabled in a config file somehow? The reason is that an attacker with HTTP access to Alloy will be able to snoop the data which was previously not exposed or encrypted on the way out. We should give users an option to have this disabled IMO.

Would it be ok if I create an issue for this and it comes in a follow up PR? I would create a block for the service and would add also other options to customize it (e.g. size of the buffer). I agree that it is important but not a blocker because having access to the Alloy's API but not its config file is not a very popular scenario I guess?

internal/service/livedebugging/livedebugging.go
wildum: Because the callback is handled by the API, it won't leak. If the user closes the page, the callback will be removed.

In the case when you update the config to remove the component, the callback will still be there but there won't be a publisher to push to it. As a result, you won't see any data coming to your stream anymore. If you refresh the page you will receive a 404 error because you are trying to access a page that doesn't exist anymore

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
wildum: https://github.com/grafana/alloy/issues/1008

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
ptodev: This bullet point says that the users are advise to open the Live Debugging UI, and then check if nothing wrong in the live debugging data stands out to them - this is the "meets your expectations" part.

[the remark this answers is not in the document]
wildum: We previously switched from "live debugging" to "debugging" to make it more generic (the debug info part on the debug page can be checked)

internal/service/livedebugging/livedebugging.go
wildum: NewLiveDebugging does not work because New() is already defined in the pkg

internal/web/api/api.go
wildum: Created an issue to track it: https://github.com/grafana/alloy/issues/1010

wildum:
> When I filter the results and click Copy, I get all the data instead of just the filtered data. Is this intentional? I'd expect to only get what I have in the buffer.

Fixed in 17aaf444feb18345b4199d5768abb3200f19c4e3

internal/component/prometheus/relabel/relabel.go
wildum: I implemented your suggestion, it also simplifies the code which is great. I wish Go had something better to handle this scenario because having this marker method feels a bit hacky

docs/sources/reference/config-blocks/livedebugging.md
```diff
@@ -0,0 +1,40 @@
...
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`livedebugging` is an optional configuration block that enables the [live debugging feature][debug] which streams real-time data from your components directly to {{< param "PRODUCT_NAME" >}}'s UI.
+
+By default, [live debugging][debug] is disabled and must be explicitly enabled through this configuration block to make the debugging data visible in the UI.
+
+{{< admonition type="note" >}}
+Enabling this feature provides detailed insights into the data flowing through your pipelines via {{< param "PRODUCT_NAME" >}}'s API.
+To ensure that your data remains secure while live debugging is enabled, consider configuring TLS in the [http block][].
```
clayton-cornell, line 20:
```suggestion
To ensure that your data remains secure while live debugging is enabled, configure TLS in the [http block][].
```

We could be more direct in the statement?

docs/sources/tasks/debug.md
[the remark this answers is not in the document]
clayton-cornell: Saying `live debugging` is fine here. I wouldn't use double quotes around the term though.

docs/sources/reference/config-blocks/livedebugging.md
```diff
@@ -0,0 +1,40 @@
...
+# livedebugging block
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`livedebugging` is an optional configuration block that enables the [live debugging feature][debug], which streams real-time data from your components directly to {{< param "PRODUCT_NAME" >}}'s UI.
+
+By default, [live debugging][debug] is disabled and must be explicitly enabled through this configuration block to make the debugging data visible in the UI.
+
+{{< admonition type="note" >}}
+Enabling this feature provides detailed insights into the data flowing through your pipelines via {{< param "PRODUCT_NAME" >}}'s API.
```
rfratto, line 19: I think we should be a little careful about using the word "API" here because we don't have a versioned/documented HTTP API. 

What about "via Alloy's UI."? 

cc @clayton-cornell

rfratto (approved):
LGTM % one comment (screenshots look good too). Thanks!

docs/sources/reference/config-blocks/livedebugging.md
clayton-cornell: Oh  yeah I missed that. :-(  We don't mention the API anywhere else. also we should not use possessive "Alloy's".

docs/sources/reference/config-blocks/livedebugging.md
```diff
@@ -0,0 +1,40 @@
...
+title: livedebugging block
+---
+
+<span class="badge docs-labels__stage docs-labels__item">Experimental</span>
+
+# livedebugging block
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`livedebugging` is an optional configuration block that enables the [live debugging feature][debug], which streams real-time data from your components directly to {{< param "PRODUCT_NAME" >}}'s UI.
```
clayton-cornell, line 14:
```suggestion
`livedebugging` is an optional configuration block that enables the [live debugging feature][debug], which streams real-time data from your components directly to the {{< param "PRODUCT_NAME" >}} UI.
```

docs/sources/reference/config-blocks/livedebugging.md
```diff
@@ -0,0 +1,40 @@
...
+
+By default, [live debugging][debug] is disabled and must be explicitly enabled through this configuration block to make the debugging data visible in the UI.
```
clayton-cornell, line 16:
```suggestion
By default, [live debugging][debug] is disabled and must be explicitly enabled through this configuration block to make the debugging data visible in the Alloy UI.
```

docs/sources/reference/config-blocks/livedebugging.md
clayton-cornell, line 19:
```suggestion
The live debugging feature uses the {{< param "PRODUCT_NAME" >}} UI to provide detailed insights into the data flowing through your pipelines.
```

clayton-cornell (commented):
Some suggestions to remove the possessive "Alloy's"

docs/sources/reference/config-blocks/livedebugging.md
wildum: This one should be {{< param "PRODUCT_NAME" >}}, no?

clayton-cornell: Yes.... I clicked through too quickly

docs/sources/reference/config-blocks/livedebugging.md
wildum: fixed via @clayton-cornell's suggestion, thanks!

clayton-cornell (approved):

wildum:
> I find that the button to enable Auto Scroll is tiny and hidden at the bottom, why not make it a normal button like others, next to Clear, Copy, etc?

FYI I created an issue for this: https://github.com/grafana/alloy/issues/1021

PR #1016, opened 2024-06-11
Helm - allow loading all alloy files in /etc/alloy

### LINKED ISSUE #228: Allow helm chart to load all river files in volumemount
### Request

Helm chart should be able to run agent with only a folder path, hence loading all river files in the configmap (mounted as files in /etc/agent)

This is already available in the cli: https://github.com/grafana/agent/pull/5228

This line could be easily changed with a value for loading a file or just a folder: https://github.com/grafana/agent/blob/b691a44b47b3d443cf9fa462190069ed5ff5b98f/operations/helm/charts/grafana-agent/templates/containers/_agent.yaml#L8

### Use case

I'm trying to have a main config.river and other river files that would be loaded on the fly (a la terraform) so that any environment/cluster can add components.

Modules are not helping as they still need to be declared for loading in the main config.river. If I don't want this module in some cluster, that will lead to errors or tedious overrides.

fculpo:
#### PR Description

This PR implements this cli feature: https://github.com/grafana/alloy/commit/0b0b4a5e2d097238f94de36c0a8722b78ab5f01b

Currently, the chart only allows to load a single alloy file as an entrypoint.
This behaviour, even with modules, makes overrides or unloading components in some clusters difficult when using a common configuration.

This PR enables loading all alloy files in `/etc/alloy` (a la terraform).
Configmap can then be overridden in some environments/clusters by not loading, adding some files, or only modifying the needed one.

Fixes #228

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


fculpo:
@tpaschalis I updated README with helm-docs.
Sorry, forgot this.

fculpo:
Basically, this PR tries to achieves what the meta k8s monitoring helm chart does: https://github.com/grafana/k8s-monitoring-helm/blob/2e2392002d17c98c100ade51e2ecaf2fa3e70a58/charts/k8s-monitoring/templates/alloy_config/_logs_service_loki.alloy.txt#L11

If/else on nonn "if-able" blocks

fculpo:
Opened proposal: https://github.com/grafana/alloy/issues/1176

petewall:
I'm trying to think of the use-case around this. Specifically, I don't think it "feels" right just yet. Is there a circumstance where you'd want to load all files in a directory (`/etc/alloy` in this case), but also want to use the config map to create the config? Meaning, create a config using the `alloy.configMap.content = "<config>`, but also load additional files in the same directory. I don't even think it's actually possible to do that, since the ConfigMap is mounted at `/etc/alloy`.

I think a more intuitive solution might be:
```
alloy:
  configDir: /etc/alloy
  configMap:
    create: true
    mount: true
    key: config.alloy
```

`configDir` is always used.
`configMap.mount` controls whether the pod and container actually try to mount the configMap, if not, you're expected to replace that mount with your own volumeMount or your own.

So a few examples:

```
alloy:
  configMap:
    content: "my config"
```
Will create the configMap, with a `config.alloy`, and mount it. It will run against `/etc/alloy/config.alloy`

```
alloy:
  configMap:
    create: false
```
Will not create the configMap, but still expects one to exist, with a `config.alloy`, and mount it. It will run against `/etc/alloy/config.alloy`


```
alloy:
  configMap:
    create: false
    key: null
```
Will not create the configMap, but still expects one to exist, and will mount it. It will run against `/etc/alloy`

```
alloy:
  configMap:
    create: false
    mount: false
```

Will not create the configMap or mount it. It will run against `/etc/alloy`. You will need to use an extra volume mount.


configMap.create=true and configMap.mount=false should be an error.

PR #1017, opened 2024-06-11
Fix loki.source.docker to handle targets with multiple networks configured

### LINKED ISSUE #225: loki.source.docker: `could not set up a wait request to the Docker client` is still displayed when the container belongs to multiple networks.
### What's wrong?

grafana/agent#6055 will fix grafana/agent#4403 . However, same error has occurred when the docker container belongs to two or more networks.
This error appears to be observed more frequently when the number of networks that the container belongs to is increased or the frequency of discovery is increased.

This phenomenon consumes CPU and error logs continue to appear. 

### Steps to reproduce

docker and grafana-agent-flow is required.
I tested this on Ubuntu 22.04 LTS.

```
ubuntu@testnode:~$ grafana-agent-flow -v
agent, version v0.40.1 (branch: HEAD, revision: f222e2e08)
  build user:       root@c2af75d2b110
  build date:       2024-02-29T19:50:44Z
  go version:       go1.22.0
  platform:         linux/amd64
  tags:             netgo,builtinassets,promtail_journal_enabled
ubuntu@testnode:~$ sudo docker version
Client: Docker Engine - Community
 Version:           25.0.3
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        4debf41
 Built:             Tue Feb  6 21:13:09 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          25.0.3
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.6
  Git commit:       f417435
  Built:            Tue Feb  6 21:13:09 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.28
  GitCommit:        ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
ubuntu@testnode:~$
```

Create docker container and connect it to user defined network.

```sh
sudo docker network create testnet1
sudo docker run -d --name testnginx --network testnet1 nginx:latest

# Then, create another network and connect the container.
sudo docker network create testnet2
sudo docker network connect testnet2 testnginx

# Ensure that the container belongs to two networks
sudo docker inspect testnginx | jq '.[] | .NetworkSettings.Networks | keys[]'
```

Wait until the agent discovers the container.

### System information

Linux 5.15.0-92 x86_64

### Software version

Grafana Agent Flow v0.40.1

### Configuration

```text
logging {
	level = "warn"
}

loki.echo "stdout" {}

discovery.docker "linux" {
}

loki.source.docker "docker" {
    targets = discovery.docker.linux.targets
    forward_to = [loki.echo.stdout.receiver]
}
```

### Logs

```text
Mar 03 03:58:10 testnode grafana-agent-flow[17376]: ts=2024-03-03T03:58:10.254744904Z level=error msg="could not set up a wait request to the Docker client" target=5d048aeb40e70ce8356f6192ea676b9fbc0d9a339b685e657f96f54c922f8634 component=loki.source.docker.docker error=""
Mar 03 03:58:10 testnode grafana-agent-flow[17376]: ts=2024-03-03T03:58:10.254856313Z level=warn msg="could not transfer logs" component=loki.source.docker.docker target=docker/5d048aeb40e70ce8356f6192ea676b9fbc0d9a339b685e657f96f54c922f8634 written=0 container=5d048aeb40e70ce8356f6192ea676b9fbc0d9a339b685e657f96f54c922f8634 err="context canceled"
```

wildum:
#### PR Description

Explanation from the investigation in the linked issue:

The discovery.docker component will create one target for every network on the same container. This means that the loki.source.docker receives multiple targets that have the same containerID in random order. This [fix](https://github.com/grafana/agent/pull/6055) introduced a filter in the loki.source.docker component to avoid collecting logs multiple times from the same container. But because the order is not guaranteed, if both targetA and targetB have the same containerID, whichever target is first in the list will be selected.
As a consequence, if we have a worker collecting logs on targetA and an update comes with targetB instead, then the worker will be canceled and a new one for targetB will start. This creates error logs with workers being canceled.
This fix sorts the targets according to their label fingerprints before filtering on the containerID. This way, if multiple targets have the same containerID, it will always be the same one that will be selected for log collection.

Fixes #225 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Main (unreleased)
 
 - Fix panic when `import.git` is given a revision that does not exist on the remote repo. (@hainenber)
 
+- Fixed an issue with `loki.source.docker` where collecting logs from targets configured with multiple networks would result in errors. (@wildum)
+
 ### Other changes
 
 - `pyroscope.ebpf`, `pyroscope.java`, `pyroscope.scrape`, `pyroscope.write` and `discovery.process` components are now GA. (@korniltsev)
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"sync"
 	"time"
 
@@ -179,6 +180,11 @@ func (c *Component) Run(ctx context.Context) error {
 	}
 }
 
+type promTarget struct {
+	labels      model.LabelSet
+	fingerPrint model.Fingerprint
+}
+
 // Update implements component.Component.
 func (c *Component) Update(args component.Arguments) error {
 	newArgs := args.(Arguments)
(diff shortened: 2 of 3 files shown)


wildum:
> LGTM - I just hope that the sorting doesn't take too much resources.

I ran some tests locally, throwing 100_000 different targets with 30 labels each to the update function runs in around 3.1s on main vs 3.6s on this branch. The sorting is ok but the computation of the fingerPrints takes a bit.

wildum:
With 1000 targets, about 35ms with main, about 42ms on this branch

ptodev:
What about an alternative solution which gets the current list of containers that are being tailed, and ignores them? The list of containers is stored in `tailerTask`.

wildum:
> What about an alternative solution which gets the current list of containers that are being tailed, and ignores them? The list of containers is stored in tailerTask.

I think that it's a good idea but it requires a few extra steps to handle all cases:
- compare containerID of incoming targets with the ones currently running
- if there is a match, compare the labels fingerprints:
  - if there is a match, re-use the existing target and ignore all other targets with this containerID
  - if there is no match, store the target and add it at the end if no target with this fingerprint was found when looping through all incoming targets (it's possible that the container was updated and the labels are not matching anymore)

It's a little bit more advanced than the previous version and because we need to ensure that the labels are up to date we still need to compute the fingerprints. We do avoid the sorting but the sorting is fast (it's the fingerprints computation that are expensive)

ptodev:
Do we really need the label fingerprints? Isn't the container ID enough?

wildum:
> Do we really need the label fingerprints? Isn't the container ID enough?

From what I understand in the code, the labels are relevant in the log collection: https://github.com/grafana/alloy/blob/main/internal/component/loki/source/docker/internal/dockertarget/target.go#L285

If you have multiple targets with the same containerID, you will pick one of these targets to collect logs. If the labels of the target change or if the target is gone (but the others remain), you will not be able to notice it without computing the fingerprints. That means it will collect logs from a target with the wrong labels.

ptodev:
The container ID is created only after a container starts. If a container is restarted, it'll have a different ID. So if a container has the same labels during its lifetime, the container ID could be the only ID we need. But we probably shouldn't make that assumption 😅  Let's just merge the PR as it is? I hope the performance impact won't be noticeable. If it is, we can think how to improve it later.

wildum:
> If a container is restarted, it'll have a different ID.
This is not true, if you restart a running container it will have the same ID. This created some problems in the past (see https://github.com/grafana/alloy/issues/278)

> So if a container has the same labels during its lifetime, the container ID could be the only ID we need.
If you have a discovery.relabel in front of it and you update the rules + reload the config, the targets will have a different set of labels but the containerID is still the same and the loki.source.docker is still running

wildum:
> Let's just merge the PR as it is? I hope the performance impact won't be noticeable. If it is, we can think how to improve it later.

👍

PR #1042, opened 2024-06-13
Add live debugging support to all otel processors

### LINKED ISSUE #1030: Add live debugging support to otel processors
This can be achieved by creating a debug consumer similar to https://github.com/grafana/agent/pull/6045/files#diff-7e01742158a9ccdea868d04f2b940485f593b1f5367309ab8c71f7b792cab6a3

wildum:

Fixes #1030 

#### Notes to the Reviewer

I created a new consumer that is injected into every otel processors.

When a stream is open, it will format the consumed telemetry using the OTLP text formatting and send it to the live debugging service.
Else it does nothing.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/processor/processor.go
+++ b/internal/component/otelcol/processor/processor.go
@@ -14,7 +14,9 @@ import (
 	"github.com/grafana/alloy/internal/component/otelcol/internal/fanoutconsumer"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/lazycollector"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/lazyconsumer"
+	"github.com/grafana/alloy/internal/component/otelcol/internal/livedebuggingconsumer"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/scheduler"
+	"github.com/grafana/alloy/internal/service/livedebugging"
 	"github.com/grafana/alloy/internal/util/zapadapter"
 	"github.com/prometheus/client_golang/prometheus"
 	otelcomponent "go.opentelemetry.io/collector/component"
@@ -60,11 +62,14 @@ type Processor struct {
 
 	sched     *scheduler.Scheduler
 	collector *lazycollector.Collector
+
+	liveDebuggingConsumer *livedebuggingconsumer.Consumer
 }
 
 var (
 	_ component.Component       = (*Processor)(nil)
 	_ component.HealthComponent = (*Processor)(nil)
+	_ component.LiveDebugging   = (*Processor)(nil)
 )
 
 // New creates a new Alloy component which encapsulates an OpenTelemetry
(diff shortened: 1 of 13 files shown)


clayton-cornell (approved):
Approved for docs

internal/component/otelcol/processor/processor.go
```diff
@@ -163,9 +176,9 @@ func (p *Processor) Update(args component.Arguments) error {
 
 	var (
 		next        = pargs.NextConsumers()
-		nextTraces  = fanoutconsumer.Traces(next.Traces)
-		nextMetrics = fanoutconsumer.Metrics(next.Metrics)
-		nextLogs    = fanoutconsumer.Logs(next.Logs)
+		nextTraces  = fanoutconsumer.Traces(append(next.Traces, p.liveDebuggingConsumer))
+		nextMetrics = fanoutconsumer.Metrics(append(next.Metrics, p.liveDebuggingConsumer))
+		nextLogs    = fanoutconsumer.Logs(append(next.Logs, p.liveDebuggingConsumer))
```
ptodev, line 181: Unfortunately, there is a performance penalty here. Even though `livedebuggingconsumer` has `MutatesData: false`, there would still be extra copying due to the live debugging in some situations. This is because the [fanoutconsumer](https://github.com/grafana/alloy/tree/f998c10614774d29e95c97c5bd2733fe3ccf7121/internal/component/otelcol/internal/fanoutconsumer). E.g. imagine that we have one mutable consumer like the [transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/69e914151e74435361a4ec0ebe3a01043c96a5b6/processor/transformprocessor/factory.go#L23), and that we are using an Alloy binary which does not support `livedebugging`. No data will be copied, because there is only one consumer. However, if we use a binary which support `livedebugging`, data will be copied for the transform processor and it won't be copied for `livedebugging`.

To remediate this, I believe we should not wire in a `livedebugging` consumer when [livedebugging](https://grafana.com/docs/alloy/next/reference/config-blocks/livedebugging/) has `enabled = false`. 

It'd be great if we make this logic even smarter in the future, so that [fanoutconsumer](https://github.com/grafana/alloy/tree/f998c10614774d29e95c97c5bd2733fe3ccf7121/internal/component/otelcol/internal/fanoutconsumer) doesn't even stream anything to the `livedebugging` consumer if the `livedebugging` UI isn't even on.

docs/sources/tasks/debug.md
```diff
@@ -92,6 +92,7 @@ Live debugging is not yet available in all components.
 
 Supported components:
 * prometheus.relabel
+* otelcol.processor.*
```
ptodev, line 95: It'd be good to also document the performance penalty on `otelcol` components.

ptodev (changes requested):

internal/component/otelcol/processor/processor.go
wildum: thnaks for catching this! I went directly with the approach that you suggested to add only the consumer when a stream is open. lmk what you think of the change

docs/sources/tasks/debug.md
wildum: with the latest change, the performance penalty only appears when a stream is open. This is expected and does not need additional documentation

ptodev (approved):
Thanks very much, LGTM!

PR #1043, opened 2024-06-13
New `otelcol.exporter.debug` component

ptodev:
#### PR Description

The `debug` exporter is a replacement for the `logging` exporter. This exporter was also [added to the Agent](https://github.com/grafana/agent/pull/5867).

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
@@ -85,11 +85,11 @@ Attributes:
 ```
 
 {{< admonition type="note" >}}
-All instances of `\n`  in the `"detailed"` example have been replaced with new lines.
+All instances of `\n` in the `"detailed"` example have been replaced with new lines.
 {{< /admonition >}}
 
 Setting `use_internal_logger` to `false` is useful if you would like to see actual new lines instead of `\n` in the collector logs.
-However, by not using the internal logger you wouldn't see metadata in the log such as `component_id=otelcol.exporter.debug.default`.
+However, by not using the internal logger you wouldn't see metadata in the log line such as `component_id=otelcol.exporter.debug.default`.
 Multiline logs may also be harder to parse.
 
 ## Blocks
@@ -133,27 +133,21 @@ information.
 
 ## Example
 
-This example scrapes Prometheus UNIX metrics and writes them to the console:
+This example receives OTLP metrics, logs, and traces and writes them to the console:
 
-```river
-prometheus.exporter.unix "default" { }
+```alloy
+otelcol.receiver.otlp "default" {
+    grpc {}
+    http {}
 
-prometheus.scrape "default" {
-    targets    = prometheus.exporter.unix.default.targets
-    forward_to = [otelcol.receiver.prometheus.default.receiver]
-}
-
-otelcol.receiver.prometheus "default" {
     output {
         metrics = [otelcol.exporter.debug.default.input]
+        logs    = [otelcol.exporter.debug.default.input]
+        traces  = [otelcol.exporter.debug.default.input]
     }
 }
 
-otelcol.exporter.debug "default" {
-    verbosity           = "detailed"
-    sampling_initial    = 1
-    sampling_thereafter = 1
-}
+otelcol.exporter.debug "default" {}
 ```
 <!-- START GENERATED COMPATIBLE COMPONENTS -->
 
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
@@ -1,0 +1,171 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol.exporter.debug/
+description: Learn about otelcol.exporter.debug
+title: otelcol.exporter.debug
+---
+
+<span class="badge docs-labels__stage docs-labels__item">Experimental</span>
+
+# otelcol.exporter.debug
+
+`otelcol.exporter.debug` accepts telemetry data from other `otelcol` components and writes them to the console (stderr).
+You can control the verbosity of the logs.
+
+{{< admonition type="note" >}}
+`otelcol.exporter.debug` is a wrapper over the upstream OpenTelemetry Collector `debug` exporter. 
+If necessary, bug reports or feature requests are redirected to the upstream repository.
+{{< /admonition >}}
+
+You can specify multiple `otelcol.exporter.debug` components by giving them different labels.
+
+## Usage
+
+```river
+otelcol.exporter.debug "LABEL" { }
+```
+
+## Arguments
+
+`otelcol.exporter.debug` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`verbosity`           | `string` | Verbosity of the generated logs. | `"basic"` | no
+`sampling_initial`    | `int`    | Number of messages initially logged each second. | `2` | no
+`sampling_thereafter` | `int`    | Sampling rate after the initial messages are logged. | `1` | no
+`use_internal_logger` | `bool`   | Whether to use the internal logger or print directly to `stdout`. | `true` | no
+
+The `verbosity` argument must be one of:
+* `"basic"`: A single-line summary of received data is logged to stderr, with a total count of telemetry records for every batch of received logs, metrics, or traces.
+* `"normal"`: Produces the same output as `"basic"` verbosity.
+* `"detailed"`: All details of every telemetry record are logged to stderr, typically writing multiple lines for every telemetry record.
+
+The following example shows `"basic"` and `"normal"` output:
+```
+ts=2024-06-13T11:24:13.782957Z level=info msg=TracesExporter component_path=/ component_id=otelcol.exporter.debug.default "resource spans": 1, spans: 2
+```
+
+The following example shows `"detailed"` output:
+```
+ts=2024-06-13T11:24:13.782957Z level=info msg=TracesExporter component_path=/ component_id=otelcol.exporter.debug.default "resource spans"=1 spans=2
+ts=2024-06-13T11:24:13.783101Z level=info msg="ResourceSpans #0
+Resource SchemaURL: https://opentelemetry.io/schemas/1.4.0
+Resource attributes:
+     -> service.name: Str(telemetrygen)
+ScopeSpans #0
+ScopeSpans SchemaURL:
+InstrumentationScope telemetrygen
+Span #0
+    Trace ID       : 3bde5d3ee82303571bba6e1136781fe4
+    Parent ID      : 5e9dcf9bac4acc1f
+    ID             : 2cf3ef2899aba35c
+    Name           : okey-dokey
+    Kind           : Server
+    Start time     : 2023-11-11 04:49:03.509369393 +0000 UTC
+    End time       : 2023-11-11 04:49:03.50949377 +0000 UTC
+    Status code    : Unset
+    Status message :
+Attributes:
+     -> net.peer.ip: Str(1.2.3.4)
+     -> peer.service: Str(telemetrygen-client)
+Span #1
+    Trace ID       : 3bde5d3ee82303571bba6e1136781fe4
+    Parent ID      :
+    ID             : 5e9dcf9bac4acc1f
+    Name           : lets-go
+    Kind           : Client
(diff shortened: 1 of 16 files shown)


docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,164 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`verbosity`           | `string` | Verbosity of the generated logs. | `"normal"` | no
+`sampling_initial`    | `int`    | Number of messages initially logged each second. | `2` | no
+`sampling_thereafter` | `int`    | Sampling rate after the initial messages are logged. | `500` | no
+
+The `verbosity` argument must be one of:
+* `"basic"`: A single-line summary of received data is logged to stderr, with a total count of telemetry records for every batch of received logs, metrics or traces.
+* `"normal"`: Currently, this produces the same output as `"basic"` verbosity.
```
clayton-cornell, line 39: Why do we have this arg value if it's the same as `basic`? Are there future plans to expand `normal`? Or to have it provide different details at some point? If it will change at some point, this feels like we would introduce a future breaking change for users who use `normal` instead of `basic` and we change the output at some future point.

Or is this for upstream compatibility with the OTel debug exporter?

docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,164 @@
...
+    End time       : 2023-11-11 04:49:03.50949377 +0000 UTC
+    Status code    : Unset
+    Status message :
+Attributes:
+     -> net.peer.ip: Str(1.2.3.4)
+     -> peer.service: Str(telemetrygen-server)
+        {"kind": "exporter", "data_type": "traces", "name": "debug"}"
+```
+
+Note that the above `"detailed"` example was edited to replace all instances of `\n` with a new line.
```
clayton-cornell, line 86: Is this important in the context of the example? If a user looks at this output in a terminal will they see the `\n`? Will the detailed output always have `\n`?

docs/sources/reference/components/otelcol.exporter.debug.md
ptodev: > Why do we have this arg value if it's the same as basic?

It's because upstream has it. It makes using the component simpler for people migrating from/to upstream, and it makes the converters simpler.

> Are there future plans to expand normal? 

I don't know. I guess at some point it'll do something different.

> If it will change at some point, this feels like we would introduce a future breaking change for users who use normal instead of basic and we change the output at some future point.

The component is labeled as "experimental", so breaking changes should be ok. Also, this component is used for debugging and isn't something meant to be run on production on a regular basis, as part of telemetry collection. So breaking changes are not such a big deal.

docs/sources/reference/components/otelcol.exporter.debug.md
ptodev: Yes, they will always see `\n` instead of actual new lines. They need to search and replace the `/n` manually.

docs/sources/reference/components/otelcol.exporter.debug.md
clayton-cornell: Cool. It was just a moment of "hmmm" as I read this. Matching upstream is enough of a reason.

docs/sources/reference/components/otelcol.exporter.debug.md
clayton-cornell:
```suggestion
{{< admonition type="note" >}}
All instances of `\n`  in the `"detailed"` example have been replaced with new lines.
{{< /admonition >}}
```

docs/sources/reference/components/otelcol.exporter.debug.md
clayton-cornell:
```suggestion
* `"normal"`: Produces the same output as `"basic"` verbosity.
```

docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,166 @@
...
+
+`otelcol.exporter.debug` accepts telemetry data from other `otelcol` components and writes them to the console (stderr).
+You can control the verbosity of the logs.
+
+{{< admonition type="note" >}}
+`otelcol.exporter.debug` is a wrapper over the upstream OpenTelemetry Collector `debug` exporter. 
+If necessary, bug reports or feature requests are redirected to the upstream repository.
+{{< /admonition >}}
+
+Multiple `otelcol.exporter.debug` components can be specified by giving them different labels.
```
clayton-cornell, line 19:
```suggestion
You can specify multiple `otelcol.exporter.debug` components by giving them different labels.
```

docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,166 @@
...
+`otelcol.exporter.debug` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`verbosity`           | `string` | Verbosity of the generated logs. | `"normal"` | no
+`sampling_initial`    | `int`    | Number of messages initially logged each second. | `2` | no
+`sampling_thereafter` | `int`    | Sampling rate after the initial messages are logged. | `500` | no
+
+The `verbosity` argument must be one of:
+* `"basic"`: A single-line summary of received data is logged to stderr, with a total count of telemetry records for every batch of received logs, metrics or traces.
```
clayton-cornell, line 38:
```suggestion
* `"basic"`: A single-line summary of received data is logged to stderr, with a total count of telemetry records for every batch of received logs, metrics, or traces.
```

docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,166 @@
...
+* `"normal"`: Produces the same output as `"basic"` verbosity.
+* `"detailed"`: All details of every telemetry record are logged to stderr, typically writing multiple lines for every telemetry record.
+
+Example of `"basic"` and `"normal"` output:
```
clayton-cornell, line 42:
```suggestion
The following example shows `"basic"` and `"normal"` output:
```

docs/sources/reference/components/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,166 @@
...
+```
+ts=2024-06-13T11:24:13.782957Z level=info msg=TracesExporter component_path=/ component_id=otelcol.exporter.debug.default "resource spans": 1, spans: 2
+```
+
+Example of `"detailed"` output:
```
clayton-cornell, line 47:
```suggestion
The following example shows `"detailed"` output:
```

docs/sources/reference/components/otelcol/otelcol.exporter.logging.md
```diff
@@ -8,6 +8,13 @@ title: otelcol.exporter.logging
 
 # otelcol.exporter.logging
 
+{{< admonition type="warning" >}}
+The OpenTelemetry Collector `logging` exporter is deprecated.
+It is due to be removed from the upstream Collector repository in September 2024.
```
clayton-cornell, line 13:
```suggestion
It will be removed from the upstream Collector repository in September 2024.
```

docs/sources/reference/components/otelcol/otelcol.exporter.logging.md
```diff
@@ -8,6 +8,13 @@ title: otelcol.exporter.logging
...
+`otelcol.exporter.logging` may also be removed in a future Alloy version.
+Please use `otelcol.exporter.debug` instead.
```
clayton-cornell, line 15:
```suggestion
Use `otelcol.exporter.debug` instead.
```

wildum:
Do you think that we need this component now that we have the live debugging feature which supports all otel receivers and processors?
This component: shows only the output of the pipeline, the data is mixed with logs from Alloy, it has sample rate and verbosity features
Live debugging: can show the output of any steps in the pipeline without modifying the original pipeline, the data is separated from Alloy's logs, it has filtering, sample rate, pause/unpause, copy to clipboard features. For the cons the user needs an access to the UI and it's an experimental feature still

ptodev:
@wildum I agree that in the long term we should recommend that people use Live Debugging. However, I think there will always be a need for us to have this exporter:
* The logs it produces could be sent to Loki
* It's a useful component for the upstream Collector, and folks who are used to it might want to still use it. They might have already established processes around this.
* It might be necessary for some config conversions between upstream yaml and Alloy syntax.

In the long term, I think we should keep this component but we can have a banner in the docs that recommends Live Debugging instead.

internal/component/otelcol/exporter/debug/debug.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+var _ exporter.Arguments = Arguments{}
+
+// SetToDefault implements river.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = Arguments{
+		Verbosity:          "normal",
+		SamplingInitial:    2,
+		SamplingThereafter: 500,
```
wildum, line 65: why is default set to 500? The default from upstream is 1 (https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/factory.go)

internal/component/otelcol/exporter/debug/debug.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return verbosity, nil
+}
+
+var _ exporter.Arguments = Arguments{}
+
+// SetToDefault implements river.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = Arguments{
+		Verbosity:          "normal",
```
wildum, line 63: upstream's default is set to basic, even if normal and basic are the same, we should maybe align on the exact same values: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/README.md

internal/component/otelcol/exporter/debug/debug.go
```diff
@@ -0,0 +1,97 @@
...
+func (args Arguments) Extensions() map[otelcomponent.ID]otelextension.Extension {
+	return nil
+}
+
+// Exporters implements exporter.Arguments.
+func (args Arguments) Exporters() map[otelcomponent.DataType]map[otelcomponent.ID]otelcomponent.Component {
+	return nil
+}
+
+// DebugMetricsConfig implements receiver.Arguments.
```
wildum, line 94:
```suggestion
// DebugMetricsConfig implements exporter.Arguments.
```

internal/converter/internal/otelcolconvert/testdata/debug.alloy
```diff
@@ -0,0 +1,17 @@
...
+
+	output {
+		metrics = [otelcol.exporter.debug.default.input]
+		logs    = [otelcol.exporter.debug.default.input]
+		traces  = [otelcol.exporter.debug.default.input]
+	}
+}
+
+otelcol.exporter.debug "default" {
+	verbosity           = "Detailed"
```
wildum, line 14: why is the first letter uppercase? that does not work, does it?

clayton-cornell:
Docs look OK to me now. :-)

internal/converter/internal/otelcolconvert/testdata/debug.alloy
ptodev: Good spot! It looks like the upstream collector unmarshals to `Detailed` and not `detailed`. Also, it accepts any casing. So I updated our code to also accept any casing.

docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
```diff
@@ -0,0 +1,171 @@
...
+        metrics = [otelcol.exporter.debug.default.input]
+    }
+}
+
+otelcol.exporter.debug "default" {
+    verbosity           = "detailed"
+    sampling_initial    = 1
+    sampling_thereafter = 1
+}
+```
```
wildum, line 157: nit: I would suggest to put a simpler example (otlp receiver -> debug exporter). Having a convert from OTLP to Prometheus as example is not ideal because it's usually preferable to avoid converting the data when possible. Examples should follow the most common use cases

wildum (approved):
LGTM, just one nit

docs/sources/reference/components/otelcol/otelcol.exporter.debug.md
ptodev: Fair enough, I changed the example to be simper :)

wildum (approved):

PR #1046, opened 2024-06-13
[cAdvisor] Podman Support

michael-act:
#### PR Description

This pull request updates the cadvisor dependency in Grafana Alloy to version v0.49.1. This upgrade introduces support for Podman. The changes involve modifications in dependency management files (`go.mod`, `go.sum`), and adjustments to remove unnecessary configurations and better align with the new `cadvisor.Manager()` .

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/internal/component/prometheus/exporter/cadvisor/cadvisor.go
+++ b/internal/component/prometheus/exporter/cadvisor/cadvisor.go
@@ -37,15 +37,6 @@ type Arguments struct {
 	DisabledMetrics            []string      `alloy:"disabled_metrics,attr,optional"`
 	EnabledMetrics             []string      `alloy:"enabled_metrics,attr,optional"`
 	StorageDuration            time.Duration `alloy:"storage_duration,attr,optional"`
-	ContainerdHost             string        `alloy:"containerd_host,attr,optional"`
-	ContainerdNamespace        string        `alloy:"containerd_namespace,attr,optional"`
-	DockerHost                 string        `alloy:"docker_host,attr,optional"`
-	UseDockerTLS               bool          `alloy:"use_docker_tls,attr,optional"`
-	DockerTLSCert              string        `alloy:"docker_tls_cert,attr,optional"`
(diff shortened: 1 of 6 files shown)


internal/component/prometheus/exporter/cadvisor/cadvisor.go
```diff
@@ -37,15 +37,6 @@ type Arguments struct {
 	DisabledMetrics            []string      `alloy:"disabled_metrics,attr,optional"`
 	EnabledMetrics             []string      `alloy:"enabled_metrics,attr,optional"`
 	StorageDuration            time.Duration `alloy:"storage_duration,attr,optional"`
-	ContainerdHost             string        `alloy:"containerd_host,attr,optional"`
```
mattdurham, line 40: Is there an upstream reason we are removing these settings? Our backwards promise tries to ensure we dont remove settings unless we absolutely have to.

michaelact: I have reverted the removed configuration. Regarding `DisableRootCgroupStats`, I think we should remove it since the code at https://github.com/google/cadvisor/blob/master/container/raw/factory.go#L34 does not expose it for configuration.

michaelact: Hi, could you please help review this PR?

mattdurham: It looks like it is used but as a flag which is likely why we expose it as a configuration option. Removing this would still break our promise for backwards compatibility.

michaelact: Should we make this configuration available but mark it as deprecated, so it doesn't take effect? 
Or is there a mistake in how I'm configuring it?

michaelact: Hi, could you please help review this PR?

PR #1086, opened 2024-06-19
Add "Fix important security issues" step to the release process

ptodev:
I'm also working on automation so that we get a new issue labelled with `security` whenever there is a new vulnerability.


dehaansa:
@ptodev is this still relevant?

PR #1092, opened 2024-06-21
Fix issues with DataDog task doc

ptodev:
There was a missing traces config, wrong quotation marks, and the name of one of the components wasn't right.


clayton-cornell (approved):
LGTM

@jdbaldry We were discussing smart quotes, and here's a good example of them causing problems.

jdbaldry:
I've started https://github.com/grafana/writers-toolkit/pull/748 to catch this in the future

PR #564, opened 2024-04-12
Fix issues with Beyla component defaults

marctc:
#### PR Description

Current logic to set default in Beyla component overrides `attributes` section, which
causes a panic when Kubernetes decoration is enabled.

This PR fixes this and adds tests to verify that defaults are not overridden.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated
- [ ] Config converters updated

PR #1097, opened 2024-06-21
Remove our soft fork of the TSDB and use upstream Agent

mattdurham:
#### PR Description

This removes our soft  fork of the Agent TSDB and uses the upstream Agent TSDB. I implemented a translation layer at the gatherer to rename the metrics that changed AND are covered by our backwards promise.

Fixes #564 

#### PR Checklist

- [x] CHANGELOG.md updated


internal/service/http/gatherer_translator.go
```diff
@@ -0,0 +1,39 @@
...
+		case "prometheus_agent_deleted_series":
+			m.Name = &deletedSeries
+		case "prometheus_agent_out_of_order_samples_total":
+			m.Name = &totalCreatedSeries
+		case "prometheus_agent_samples_appended_total":
+			m.Name = &appendedSamples
+		}
+	}
+	return metrics, nil
+}
```
rfratto, line 39: Nice, the metrics were the main reason why I didn't remove the fork. I think this is a good approach.

I think we should do a few more small things here though:

* Just for complete consistency, maybe the metric description should be consistent too? Though I'm OK if the description is vague enough that we don't need to change it to be Alloy-specific. 
* While fairly unlikely, it's possible that upstream can change the metric types for each of these (i.e., turning a gauge into a counter). Can we have a basic assertion for each of these and panic (or something) if the type changed from what we expected? 
* It's possible that upstream might remove one of these metrics; can we have a basic assertion that we get all four expected ones?

mattdurham: I think we can do this...the last point might be difficult. Since if you dont define a remote_write you wont get the metrics. Will see if there is a way to cleanly detect without http service knowing the running components.

rfratto: > I think we can do this...the last point might be difficult. Since if you dont define a remote_write you wont get the metrics. Will see if there is a way to cleanly detect without http service knowing the running components.

I was talking about the translator doing that last check, since presumably the internal gatherer would emit all the metrics we expect. Is there still a case where translator.Gather would run but a subset of the four metrics wouldn't be emitted?

mattdurham: If there is no `prometheus.remote_write` component then we wont get the 4 metrics. I could add a sentinel metric that always exists if there is a remote_write, or detect if a component_id starts with prometheus.remote_write but those both feel a bit off.

rfratto: _Oh_, I see. I thought the translator code was in prometheus.remote_write, not in the shared HTTP service. Is this something we need to have at the service level?

go.mod
```diff
@@ -159,7 +159,7 @@ require (
 	github.com/prometheus/mysqld_exporter v0.14.0
 	github.com/prometheus/node_exporter v1.6.0
 	github.com/prometheus/procfs v0.15.0
-	github.com/prometheus/prometheus v0.51.2 // a.k.a. v2.51.2
+	github.com/prometheus/prometheus v1.99.0
```
rfratto, line 162: Heads up: this version seems wrong. 1.99.0 was published on April 4, 2022, and has since [been retracted](https://pkg.go.dev/github.com/prometheus/prometheus@v1.99.0).

mattdurham: I actually wonder if I need to change that at all since its overridden with a replace anyways. Will test.

rfratto: Ah yeah, it probably doesn't need to change but it might make sense to have it reflect the replace directive as closely as possible (but you're right that it really doesn't matter)

mattdurham: Agreed.

mattdurham:
Working on getting update working but its likely a bit wonky

internal/service/http/gatherer_translator.go
mattdurham: Gather seems to be only called by the promhttp handler at a very high level. I have been unable to figure out how to make it work smoothly without possibly converting to text and back.

mattdurham: I tried this initially but the Desc structure doesnt expose anything but `String` publicly. The root `Register` function in prometheus can go around that since it is in the same package as Desc and access the private fields.

Gatherer was the only place I could find the ability to change the string name. I thought about parsing the string and then creating a new desc inline but that feels worse and more error prone than using the Gatherer.

rfratto: Is there any (minimal) API changes we could make to client_golang that would make renaming metrics from a Registerer easier? For example, they have [prometheus.WrapRegistererWith](https://pkg.go.dev/github.com/prometheus/client_golang@v1.19.1/prometheus#WrapRegistererWith), maybe we can create some kind of `prometheus.MutateRegisterer` since that would be able to easily change the Desc. (Undeniably this would still require more work though)

If there's nothing we could do upstream, I'm fine with using the Gatherer but it's not ideal since I'd prefer to avoid component-specific logic at the service level (and because it prevents the other safety checks I've requested here).

PR #1110, opened 2024-06-24
Use percentages in OTel alerts for traces

ptodev:
Sometimes we get alerted for increases in failed sends, even if the failed spans are a very small proportion of the total number of spans. Users can still get something similar to the the old behaviour if they chose the `percentThreshold` to be over 0.

When I test this change in our internal clusters, I'll take it out of the draft PR stage.


thampiotr (approved):
I like this and I don't see any issues from reading the code. 
We should probably note this in the changelog, otherwise LGTM.

operations/alloy-mixin/alerts/opentelemetry.libsonnet
```diff
@@ -1,7 +1,7 @@
 local alert = import './utils/alert.jsonnet';
 
 {
-  newOpenTelemetryAlertsGroup(enableK8sCluster=true):
+  newOpenTelemetryAlertsGroup(enableK8sCluster=true, percentThreshold='5', forT='5m'):
```
thampiotr, line 4: I just took a look how this pans out in practice. There are a lot of temporary issues that recover within just a few minutes, so being notified of them after 5min by default may be a bit on a high end. I would propose changing `forT=15m` then when we get notified, we know that there is a good chance the issue persists and should be looked into.

Issue #245, opened 2024-02-15
Document environment variables alongside relevant config


ptodev:
### Request

There are various environment variables (mostly from Go) which can affect the Agent's behaviour. Currently, those are not documented anywhere. It would be nice to add them in a new section in the [Command-line interface docs](https://grafana.com/docs/agent/latest/flow/reference/cli/). Component reference docs such as [tls config block](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.http/#tls_config-block) can then point to relevant sections of that env var doc.

### Use case

For example:
* `GODEBUG=x509usefallbackroots=1` which was added in grafana/agent#6340
* `HTTP_PROXY`
* Other `GODEBUG` use cases
* Maybe there are other modules such as gRPC, which have their own env vars too?

clayton-cornell:
@ptodev Where can I find the env vars we use?

ptodev:
There isn't an easy way to find them. Unlike cmd args, env vars could be referenced anywhere in the code, including in libraries which we import. This is what makes having a doc page important :) 

It's not necessary to have an exhaustive list. The main goals are:
* Making a space in the documentation for env vars, so that we know where to document them as time goes on.
* Documenting the most important ones (which are probably the ones listed in this issue).

clayton-cornell:
GODEBUG env vars here: https://pkg.go.dev/runtime?utm_source=godoc#hdr-Environment_Variab

PR #1137, opened 2024-06-25
Add an environment variable topic

clayton-cornell:
#### PR Description

This PR adds a new environment variable topic to the CLI section of the Alloy docs.

Review question: What additional env vars should be documented?

 Fixes https://github.com/grafana/alloy/issues/245

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/cli/environment-variables.md
+++ b/docs/sources/reference/cli/environment-variables.md
@@ -21,9 +21,18 @@ Refer to the [Go runtime][runtime] documentation for more information about Go r
 
 The following arguments are supported.
 
-Environment variable | Argument                   | Description                                                                     | Default
----------------------|----------------------------|---------------------------------------------------------------------------------|--------
-`GODEBUG`            | `x509usefallbackroots`     | Enforce a fallback on the X.509 trusted root certificates. Set to `1` to enable | `0`
-`HTTP_PROXY`         | `http://proxy.example.com` | The hostname or IP address of the proxy server.                                 |
+### GODEBUG
+
+Argument               | Description                                                                     | Default
+-----------------------|---------------------------------------------------------------------------------|--------
+`x509usefallbackroots` | Enforce a fallback on the X.509 trusted root certificates. Set to `1` to enable | `0`
+`netdns=go`            | Force pure Go resolver                                                          |
+`netdns=cgo`           | Force native resolver. Can be set to `cgo` or `win32`.                          |
+
+### HTTP_PROXY
+
+Argument                   | Description                                     | Default
+---------------------------|-------------------------------------------------|--------
+`http://proxy.example.com` | The hostname or IP address of the proxy server. |
 
 [runtime]: https://pkg.go.dev/runtime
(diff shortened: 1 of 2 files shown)


docs/sources/reference/cli/environment-variables.md
```diff
@@ -0,0 +1,37 @@
...
+## Arguments
+
+The following arguments are supported.
+
+### GODEBUG
+
+Argument               | Description                                                                                          | Default
+-----------------------|------------------------------------------------------------------------------------------------------|--------
+`x509usefallbackroots` | Enforce a fallback on the X.509 trusted root certificates. Set to `1` to enable.                     | `0`
+`netdns`               | Force a resolver. Set to `go` for a pure Go resolver. Set to `cgo` or `win32` for a native resolver. |
```
ptodev, line 29: I don't think this env var makes a difference in Alloy on Linux, because Alloy is built for Linux using a `netdns` flag. This forces the resolver to always be `netdns`. It might make more difference for windows though. Maybe this is worth documenting? cc @rfratto 

It may be worth documenting `GODEBUG=netdns=1` and `GODEBUG=netdns=2` though. they print extra information to logs. "2" is more verbose than "1".

mattdurham: The build tag can be overridden at runtime. Looking at https://github.com/golang/go/blob/5a18e79687dea15680ff5f799b549fa0efd0cad9/src/net/conf.go#L86 there are several paths that can change the buildtag behavior by adjusting the prefer flags.

mattdurham: This setting `GODEBUG=netdns=go` forces using go on my linux box. Specifically pivots on conf.d:235, whereas the default uses cgo resolver.

clayton-cornell: I've added what's being discussed here and in the Slack thread. It's not "right" yet... just added as-is to capture the suggestions.

mattdurham (approved):
LGTM.

clayton-cornell:
@mattdurham I don't yet have a definition of `PPROF_MUTEX_PROFILING_PERCENT` or what it could be set to... any suggestions for that one? Or I can remove it in this iteration and put it into the next?

mattdurham:
`PPROF_MUTEX_PROFILING_PERCENT` enables the ability to retrieve data from the pprof mutex endpoint. It needs to be set to a whole number, for example 50 would profile 50% of the mutexes. 5 is likely a sufficient value. By default we sample 0.01 percent.

`PPROF_BLOCK_PROFILING_RATE` also allows tracking of mutexes for arcane reasons 1 tracks everything, 0 tracks nothing. Any value over 1 is treated as track anything over <VALUE> nanoseconds. By default we apply 10,000 for tracking.

mattdurham:
https://github.com/grafana/alloy/blob/f9c5e65ad45bd1b71fcdd314b7fc4f496e3293b9/internal/alloycli/cmd_run.go#L498 contains the terribleness.

mattdurham:
I am also pro splitting this out to a different PR because these settings are weird and likely need more guidance. Unless you can work your magic on what I wrote and make them coherent :)

clayton-cornell:
I opened an issue on this so it's not lost. Yeah... this env var will need some wordsmithing to get it into some sort of state that can be clearly understood.  I'll push the change to remove that var from the table and merge this PR to get the initial topic published.

PR #1139, opened 2024-06-25
misc: migrate to community.grafana.com

rfratto:
community.grafana.com has had a Grafana Alloy category for a while now, leading to a gap where maintainers were unaware of where some questions were being asked.

community.grafana.com is the typical place for questions for a few projects, so it makes sense to shut down the discussions page on GitHub in favor of the community site instead.


thampiotr (approved):
No objections from me 👍

Issue #156, opened 2023-01-19
Proposal: add function closures and data transformation functions to River


rfratto:
## Background

The relabel rules from the `discovery.relabel` component tend to have a high learning curve, and where complicated rules (such as only including targets where two conditions A and B are both true) are harder than they need to be. 

For example, I am personally using this relabel rule to set a label from a collection of other labels, copying the value from the first label which is set:

```river 
rule {
	// Try to identify a service name to eventually form the job label. We'll
	// prefer the first of the below labels, in descending order.
	source_labels = [
		"__meta_kubernetes_pod_label_k8s_app",
		"__meta_kubernetes_pod_label_app",
		"__meta_kubernetes_pod_label_name",
		"__helm_name__",
		"__meta_kubernetes_pod_controller_name",
		"__meta_kubernetes_pod_name",
	]
	target_label = "__service__"

	// Our in-memory string will be something like A;B;C;D;E;F, where any of the
	// letters could be replaced with a label value or be empty if the label
	// value did not exist.
	//
	// We want to match for the very first sequence of non-semicolon characters
	// which is either prefaced by zero or more semicolons, and is followed by
	// zero or more semicolons before the rest of the string.
	regex = ";*([^;]+);*.*"
}
``` 

Grafana Agent Flow was originally designed to use expressions to allow for more flexibility in the configuration file. These expressions have the capability of mutating values using the River language instead of Domain-Specific Languages like the `rule` block. 

However, it is not possible to write any River today to transform a set of discovered targets. I propose adding function closures and a new set of data transformation functions (map/reduce/filter) to enable a new class of expressiveness in Grafana Agent Flow configs.

Adding support for function closures and data transformation functions would enable users to have more fine-grained control over data used in Flow pipelines, reducing the need for one-off data transformation components that Grafana Agent maintainers need to introduce to cover common use cases. 

## Example 

```river 
discovery.kubernetes "pods" {
  role = "pod"
}

prometheus.scrape "default" {
  // Map over discovered targets and inject a new label. 
  targets = map(
    func(target) => target + { 
      extra_label = "test",
    },
    discovery.kubernetes.pods.targets,
  )

  forward_to = [...]
}
```

Combining with the [coalesce](https://grafana.com/docs/agent/latest/flow/reference/stdlib/coalesce/) function from the standard library, the relabel rule from above could be rewritten as:

```river 
map(
  func(target) => target + {
    // Try to identify a service name to eventually form the job label. We'll
[...]

PR #1161, opened 2024-06-27
proposal: standardizing the Alloy stdlib

rfratto:
Today, our 19 identifiers in the standard library were added organically with little thought to a consistent design. This has been fine due to the small size of the standard library, but as the standard library grows, the lack of consistency will start to make it harder to understand. 

This proposal introduces a way to standardize on how new capabilities are added to Alloy's standard library so that usage of the standard library feels consistent.

### DIFF
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -61,10 +61,10 @@ similar identifiers together:
     * `array.concat` (previously `concat`)
 * Conversion functions
     * `convert.nonsensitive` (previously `nonsensitive`)
-* Decoding functions
-    * `decode.base64` (previously `decode_base64`)
-    * `decode.json` (previously `decode_json`)
-    * `decode.yaml` (previously `decode_yaml`)
+* Encoding functions
+    * `encoding.from_base64` (previously `base64_decode`)
+    * `encoding.from_json` (previously `json_decode`)
+    * `encoding.from_yaml` (previously `yaml_decode`)
 * System-related functions
     * `sys.env` (previously `env`)
 * String functions
@@ -78,6 +78,14 @@ similar identifiers together:
     * `string.trim_prefix` (previously `trim_prefix`)
     * `string.trim_space` (previously `trim_space`)
     * `string.trim_suffix` (previously`trim_suffix`)
+
+> **NOTE**: The decoding functions were placed into an `encoding` namespace to
+> leave the door open for having a single namespace for both performing
+> decoding and encoding, so that `encoding.from_base64` could coexist alongside
+> a `encoding.to_base64`.
+>
+> In this context, the namespace is used as a noun to categorize both the
+> encoding and decoding actions.
 
 For identifiers where a namespace has been introduced, the old identifier will
 be marked deprecated for removal for the next major release. The documentation
@@ -122,10 +130,10 @@ what these namespaces could eventually look like:
     * `convert.to_number` (**new**; convert a string to a number)
     * `convert.to_string` (**new**; convert a number to a string)
     * `convert.to_bool` (**new**; convert a string or number to a boolean)
-* Decoding functions
-    * `decode.base64`
-    * `decode.json`
-    * `decode.yaml`
+* Encoding functions
+    * `encoding.from_base64`
+    * `encoding.from_json`
+    * `encoding.from_yaml`
 * System-related functions
     * `sys.env`
     * `sys.cpu_count` (**new**; the number of CPUs on a system)
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -80,7 +80,7 @@ similar identifiers together:
     * `string.trim_suffix` (previously`trim_suffix`)
 
 For identifiers where a namespace has been introduced, the old identifier will
-be marked deprecated for removal for the next major release.
+be marked deprecated for removal for the next major release. The documentation will present the namespaced functions prominently and mention the old name as a deprecated alias.
 
 Some identifiers have not been given a namespace because I could not easily
 identify one: `constants`, `coalesce`, and `json_path`. Until a namespace is
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -65,8 +65,8 @@ similar identifiers together:
     * `decode.base64` (previously `decode_base64`)
     * `decode.json` (previously `decode_json`)
     * `decode.yaml` (previously `decode_yaml`)
-* OS-related functions
-    * `os.env` (previously `env`)
+* System-related functions
+    * `sys.env` (previously `env`)
 * String functions
     * `string.format` (previously `format`)
     * `string.join` (previously `join`)
@@ -126,10 +126,10 @@ what these namespaces could eventually look like:
     * `decode.base64`
     * `decode.json`
     * `decode.yaml`
-* OS-related functions
-    * `os.env`
-    * `os.uid` (**new**; return the user ID of the current process)
-    * `os.gids` (**new**; return the group IDs of the current process)
+* System-related functions
+    * `sys.env`
+    * `sys.cpu_count` (**new**; the number of CPUs on a system)
+    * `sys.memory` (**new**; the amount of memory available on a system)
 * String functions
     * `string.format`
     * `string.join`
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -2,7 +2,7 @@
 
 * Author: Robert Fratto (@rfratto)
 * Last updated: 2024-06-27
-* Discussion link: TODO
+* Discussion link: <https://github.com/grafana/alloy/pull/1161>
 
 ## Abstract
 
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -70,7 +70,6 @@ similar identifiers together:
 * String functions
     * `string.format` (previously `format`)
     * `string.join` (previously `join`)
-    * `string.json_path` (previously `json_path`)
     * `string.replace` (previously `replace`)
     * `string.split` (previously `split`)
     * `string.to_lower` (previously `to_lower`)
@@ -83,9 +82,9 @@ similar identifiers together:
 For identifiers where a namespace has been introduced, the old identifier will
 be marked deprecated for removal for the next major release.
 
-Two identifiers have not been given a namespace because I could not easily
-identify one: `constants` and `coalesce`. Until a namespace is identified (if
-ever), these identifiers will not be deprecated.
+Some identifiers have not been given a namespace because I could not easily
+identify one: `constants`, `coalesce`, and `json_path`. Until a namespace is
+identified (if ever), these identifiers will not be deprecated.
 
 New identifiers **should** be given a namespace; namespacing makes it easier
 for the same identifier to exist in two different namespaces, such as
@@ -99,7 +98,7 @@ like today.
 
 I've elected for namespaces to use dot separation (`NAMESPACE.IDENTIFIER`) as
 opposed to underscores (`NAMESPACE_IDENTIFIER`) to align with `constants` and
-component names, which both  use dot separation for namespacing.
+component names, which both use dot separation for namespacing.
 
 ### Example future stdlib
 
@@ -132,7 +131,6 @@ what these namespaces could eventually look like:
 * String functions
     * `string.format`
     * `string.join`
-    * `string.json_path`
     * `string.replace`
     * `string.split`
     * `string.to_lower`
--- a/docs/design/1161-standardize-stdlib.md
+++ b/docs/design/1161-standardize-stdlib.md
@@ -101,6 +101,48 @@ I've elected for namespaces to use dot separation (`NAMESPACE.IDENTIFIER`) as
 opposed to underscores (`NAMESPACE_IDENTIFIER`) to align with `constants` and
 component names, which both  use dot separation for namespacing.
 
+### Example future stdlib
+
+Three of the five namespaces above only have one identifier. However, there are
+many new identifiers we may want to include in the future. Here's an example of
+what these namespaces could eventually look like:
+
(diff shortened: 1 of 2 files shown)


docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,197 @@
...
+* Decoding functions
+    * `decode.base64` (previously `decode_base64`)
+    * `decode.json` (previously `decode_json`)
+    * `decode.yaml` (previously `decode_yaml`)
+* OS-related functions
+    * `os.env` (previously `env`)
+* String functions
+    * `string.format` (previously `format`)
+    * `string.join` (previously `join`)
+    * `string.json_path` (previously `json_path`)
```
mattdurham, line 73: This one feels a bit odd since it only works on a subtype of strings.

Would it make sense to have format namespaces. IE
`json.decode`, `yaml.decode`, and `json.path`?

rfratto: I agree that this one doesn't quite fit in. 

I suspect that it would be ideal if the namespaces we identify could _potentially_ have a handful of identifiers, otherwise we risk ending up with way too many namespaces that become hard to reason about. 

If we had formatting namespaces, I don't suspect we'd usually have more than 2-3 identifiers:

* `json.decode`
* `json.encode`
* `json.path` 
* `yaml.decode`
* `yaml.encode` 
* `base64.encode`
* `base64.decode`
* (etc) 

I'm concerned that these are really small namespaces and if we continue to add support for new encoding/decoding targets (MD5? SHA?), the number of top-level format namespaces would quickly outnumber everything else. 

But still, I agree this one sticks out as strange and it may be worth exploring all the different ways we could group together encoding-related functionality.

mattdurham: I think that's totally reasonable and wouldn't fight for the `yaml`, `json` etc namespaces. Maybe a filter/search namespace? Do we fill that would have more options?

mattdurham: Overall in favor of the proposed solution.

rfratto: As a temporary solution, we could leave `json_path` alone (like how I'm handling `constants` and `coalesce`) while we try to identify what the best namespace for it is

rfratto:
cc @thampiotr and @wildum for thoughts on this one (not including Paulin since he's on PTO)

docs/design/1161-standardize-stdlib.md
rfratto: I've updated the proposal so that `json_path` isn't given a namespace yet so we have more time to decide. I'm a little concerned that if we were to put `json_path`, `constants`, and `coalesce` into a namespace it would change the design of the other proposed namespaces, but I'm not sure how realistic that concern is.

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,237 @@
...
+* Array functions
+    * `array.concat` (previously `concat`)
+* Conversion functions
+    * `convert.nonsensitive` (previously `nonsensitive`)
+* Decoding functions
+    * `decode.base64` (previously `decode_base64`)
+    * `decode.json` (previously `decode_json`)
+    * `decode.yaml` (previously `decode_yaml`)
+* OS-related functions
+    * `os.env` (previously `env`)
```
rfratto, line 69: I'm a little worried `os` is a bad namespace since I can't identify any new functionality that would belong in it. The examples I gave later (`os.uid`, `os.gids`) are contrived without an obvious use case. 

However, one possibility is we learn that `os` is a more appropriate name for what is currently called `constants`. Then we could have: 

* `os.env` (previously `env`) 
* `os.name` (previously `constants.os`) 
* `os.hostname` (previously `constants.hostname`) 
* `os.arch` (previously `constants.arch`) 

I believe this would be the most effective use of the `os` namespace (to fill it with constants and utilities). I would like to treat this as out of scope for this proposal as much as we can, but it is still important for us to make sure that the namespace names I've selected make sense.

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,237 @@
...
+similar identifiers together:
+
+* Array functions
+    * `array.concat` (previously `concat`)
+* Conversion functions
+    * `convert.nonsensitive` (previously `nonsensitive`)
+* Decoding functions
+    * `decode.base64` (previously `decode_base64`)
+    * `decode.json` (previously `decode_json`)
+    * `decode.yaml` (previously `decode_yaml`)
```
rfratto, line 67: I have similar concerns for the `decode` namespace as I do for os. It seems at least a little likely that we will have some encoding functionality, like the ability to encode base64 or produce a checksum of an input. 

It doesn't feel like encoding should be added into a decode-specific namespace. @mattdurham suggested a `format` namespace which could contain both encoding and decoding functionality

* `format.base64.encode`
* `format.base64.decode` 
* `format.json.decode` 
* (etc)

We could also call the namespace "encoding" (similar to Go's encoding/json package) and use different verbs to describe encoding/decoding to avoid using the word "encode" more than once:

* `encoding.to_base64` 
* `encoding.from_base64` 
* `encoding.to_json` 
* (etc)

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,237 @@
...
+for the same identifier to exist in two different namespaces, such as
+`string.join` (for joining multiple strings) alongside a hypothetical
+`path.join` (for joining file system paths).
+
+Introducing a new namespace should be done as a proposal; ideally proposing
+multiple identifiers that would comprise that namespace. Introducing new
+identifiers in an existing namespace should also be done as a proposal, much
+like today.
+
+I've elected for namespaces to use dot separation (`NAMESPACE.IDENTIFIER`) as
```
wildum, line 99: Worth mentioning in the proposal: The new namespaces may collide with existing custom components but it should not be a breaking change because the custom components shadow the stdlib and because of the aliasing.

wildum:
I only have a small concern: users may need to rename their custom components if they want to use new functions of the stdlib or when we remove the old names in the next major release. We may need to add a warning log that says "custom component x is shadowing a stdlib namespace or function". But overall I'm in favor of the proposal

docs/design/1161-standardize-stdlib.md
mattdurham: I would rather the lower block. The three name block feels to much, which is more stylistic since its not very different in the actual number of characters.

docs/design/1161-standardize-stdlib.md
thampiotr: I don't think `os` is terrible, but how about `system`?
- `system.env` - env variables
- `system.name` = `constants.os`
- `system.hostname` = `constants.hostname`
- system.arch` = `constants.arch`

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,247 @@
...
+    * `string.split` (previously `split`)
+    * `string.to_lower` (previously `to_lower`)
+    * `string.to_upper` (previously `to_upper`)
+    * `string.trim` (previously `trim`)
+    * `string.trim_prefix` (previously `trim_prefix`)
+    * `string.trim_space` (previously `trim_space`)
+    * `string.trim_suffix` (previously`trim_suffix`)
+
+For identifiers where a namespace has been introduced, the old identifier will
+be marked deprecated for removal for the next major release.
```
thampiotr, line 83:
```suggestion
For identifiers where a namespace has been introduced, the old identifier will
be marked deprecated for removal for the next major release. The documentation will present the namespaced functions prominently and mention the old name as a deprecated alias.
```

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,247 @@
...
+### Example future stdlib
+
+Three of the five namespaces above only have one identifier. However, there are
+many new identifiers we may want to include in the future. Here's an example of
+what these namespaces could eventually look like:
+
+> **NOTE**: These are not part of the proposal and are only provided as an
+> example.
+
+* Array functions
```
thampiotr, line 112: Just a side-thread. Thoughts on a namespace like `object.` that, for example can allow you to drop fields, map them, filter them, etc. For example `object.drop(my_object, "__address__")` which could be used on maps in lists of targets if we also had `array.map(my_targets, objects.drop(_, "___address___"))` to do some simple relabelling using the `_` as a placeholder for argument ([Scala does it](https://www.learningjournal.guru/article/scala/functions-in-scala/scala-placeholder-syntax/)). It gives you useful functions, without the need to implement a more complex closures syntax. I'd also be open to something like [Python lambda](https://www.w3schools.com/python/python_lambda.asp) syntax. Just in general, if our closures are limited to a single expression, I think it's a good compromise to keep complexity under control and still give power to the users.

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,247 @@
...
+  number of identifiers grows.
+
+## Compatibility
+
+This proposal is properly backwards compatible, with the old identifiers being
+marked deprecated in favour of their namespaced equivalents (where a namespace
+was introduced).
+
+It is possible that any namespaces introduced in this proposal may collide with
+an existing custom component. Custom components shadow the stdlib, so this is
```
thampiotr, line 223: Ha! Good call out. It's nice that custom components take precedence.

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,247 @@
...
+not a breaking change, but users will have to change the name of their custom
+components that collide with a stdlib namespace before being able to use
+functions in that namespace.
+
+By default, the removal deprecated identifiers would be considered for an
+eventual 2.0 release, but they may be kept around longer based on usage.
+
+## Implementation
```
thampiotr, line 231: Would be best if we also moved our examples, test (and any of your internal) configs to use the new namespaces. 

Also, config converters should start using the new ones.

thampiotr (approved):
Just a few suggestions, but generally I don't see blockers here.
Nice work!

docs/design/1161-standardize-stdlib.md
rfratto: Credit to @wildum :stuck_out_tongue:

docs/design/1161-standardize-stdlib.md
rfratto: `system` seems like a good candidate. How can we evaluate `system` and `os`? The most obvious difference between the two is character length, but I don't think that should be the only criteria we look at.

thampiotr: IMO `os` is more specific and `system` is more general. Operating System means the lowest-level software on the hardware like Linux and Windows. System can encompass hardware and software. And in this example, the `system.arch` I think works better, as it typically will match the hardware. 

So I think `system` or `sys` is a better option overall. Not all the things we add into this namespace may be related to OS - for example `system.cpu_count` or `system.memory`. It's also a namespace often used in other language standard libraries, like Java, while Golang uses os mostly.

docs/design/1161-standardize-stdlib.md
thampiotr: Yeah, I think that https://github.com/grafana/alloy/issues/156 should consider an option of using single expression syntax, as it may address the use cases while remaining simpler and more constrained. 

That was a deliberate decision made [in Python for lambdas](https://www.artima.com/weblogs/viewpost.jsp?thread=147358) and, similar to Python's philosophy, I would really like to reduce the complexity where we can. I think most of the things we want to do in Alloy could be done with single expressions built on top of a strong standard library.

thampiotr: Re: the `_` placeholder - I am not that attached to this idea - it's something that works in Scala, but takes a bit to get used to. Python's lambdas may be better:
```
targets = array.map(my_targets, lambda x: objects.drop(x, "__address__"))
```
of use `func` instead of `lambda`.

docs/design/1161-standardize-stdlib.md
rfratto: That makes sense to me :+1: I do prefer `sys` just to keep it as short as possible, and I think `sys` is a common enough abbreviation that it won't be too confusing. 

If you think sys will work, I'll update the proposal

thampiotr: Yep, I like sys, it's what [Python](https://docs.python.org/3/library/sys.html) uses too.

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,260 @@
...
+
+1. Implement the new namespaced identifiers, and deprecate the old ones in
+   documentation.
+
+2. Update examples, tests, internal configs, and config converters to use the
+   new namespaced identifiers.
+
+2. Find a way to detect and report usage of deprecated identifiers to a user.
+
+3. Add standard library usage to usage stats, so usage of deprecated
```
ptodev, line 254:
```suggestion
3. Add standard library usage stats, so usage of deprecated
```

docs/design/1161-standardize-stdlib.md
```diff
@@ -0,0 +1,260 @@
...
+   identifiers can be tracked over time, allowing maintainers to make an
+   informed decision for if they can be removed alongside a major release.
+
+Additionally, as this change may introduce new instances where custom component
+names shadow a namespace, we should add a warning to notify the user of
+shadowing (if detected).
```
ptodev, line 260: Is shadowing an actual concern, given that stdlib functions are called using `()`, whereas component exports are not?

rfratto: I wouldn't say "concern" (in this case, shadowing is preferable), but it is real; the syntax evaluator treats all identifier lookups equally, regardless of whether they're called as a function. 

Plus, there are non-function identifiers in the stdlib, like `constants.hostname`.

rfratto:
It looks like we have rough consensus towards this proposal. I'm moving this to **Likely Accept** for the next week. This begins the final comment period for someone to change consensus to make changes to this proposal or move towards declining it.

wildum (approved):

wildum:
The proposal has been in **Likely Accept** for a week. No comments have been made against it, it is now moved to **Accepted**.

PR #1188, opened 2024-07-03
update automatically settting gomemlimit

mattdurham:
#### PR Description

Graduate automatically setting gomemlimit to GA. It has been in public preview with no problems for two releases. Manually setting the value is something we tell users to do often so this mimics that if they have a limit set, ie most commonly through kubernetes.

- [x] CHANGELOG.md updated


erikbaranowski (approved):

mattdurham:
Adding some documentation somewhere on this.

docs/sources/reference/cli/run.md
```diff
@@ -164,6 +164,12 @@ original configuration.
 Include `--config.extra-args` to pass additional command line flags from the original format to the converter.
 Refer to [alloy convert][] for more details on how `extra-args` work.
 
+## Automatic setting of GOMEMLIMIT
```
mattdurham, line 167: @clayton-cornell I have no idea where to put this doc. I could put it in the environment variable section we have going since it relates closely to that. Maybe around the kubernetes sections since that is where it will be used 90% of the time.

clayton-cornell: I would lean towards adding it to the Env Var topic and linking too the env var topic from wherever in the Kubernetes topics.

This would also push things to get the first draft of the env var topic published (and we can expand it as needed vs trying to capture additional vars).

docs/sources/reference/cli/run.md
```diff
@@ -164,6 +164,12 @@ original configuration.
...
+
+The `GOMEMLIMIT` variable will be automatically set if {{< param "PRODUCT_NAME" >}} can determine the appropriate value. If the `GOMEMLIMIT` value is set before running  {{< param "PRODUCT_NAME" >}} then that setting will be honored. The value will be set to 90% of the cgroup value set.
```
thampiotr, line 169: What is it set to when the appropriate value cannot be determined? if there is no cgroup? I imagine it's then not set, but just maybe good idea to doc that.

internal/alloycli/cmd_run.go
```diff
@@ -197,9 +197,7 @@ func (fr *alloyRun) Run(configPath string) error {
 
 	// Set the memory limit, this will honor GOMEMLIMIT if set
 	// If there is a cgroup will follow that
-	if fr.minStability.Permits(featuregate.StabilityPublicPreview) {
```
thampiotr, line 200: Want to remove the `Permits` function? 😂 
It could be useful though, just that I don't think it's used now so YAGNI.

thampiotr (approved):

docs/sources/reference/cli/run.md
mattdurham: That is correct, I can explicitly call that  out.

docs/sources/reference/cli/run.md
clayton-cornell: Now that the Env Var topic is live, should the GOMEMLIMIT info be shifted over?

mattdurham: I havent forgotten this, on my todo list after I get the patch release out.

docs/sources/reference/cli/environment-variables.md
```diff
@@ -35,4 +36,12 @@ Argument                   | Description                                     | D
 ---------------------------|-------------------------------------------------|--------
 `http://proxy.example.com` | The hostname or IP address of the proxy server. |
 
+### GOMEMLIMIT
+
+Normally the [Go runtime][runtime] will release memory back to the Operating System when requested. In certain environments this may cause issues such as Out of Memory errors. The `GOMEMLIMIT` variable is a numeric value in bytes with an optional unit suffix. The supported suffixes include B, KiB, MiB, GiB, and TiB. This should not be treated has a hard limit. The process can use more memory if that memory is required. A rough number is to set `GOMEMLIMIT` to 90% of the maximum amount of memory usage. For example if you wanted to keep memory usage below `10GiB` then use `GOMEMLIMIT=8GiB`.
```
mattdurham, line 41: @clayton-cornell I added some doc here but its rough, would love additional eyes.

clayton-cornell, line 41:
```suggestion
Usually, the [Go runtime][runtime] will release memory back to the operating system when requested.
In some environments, this may cause issues such as Out Of Memory (OOM) errors.
You can use the `GOMEMLIMIT` environment variable to set a soft memory cap and limit the maximum memory {{< param "PRODUCT_NAME" >}} can use.
You can set `GOMEMLIMIT` to a numeric value in bytes with an optional unit suffix.
The supported unit suffixes are `B`, `KiB`, `MiB`, `GiB`, and `TiB`.
Don't treat the `GOMEMLIMIT` environment variable as a hard memory limit.
{{< param "PRODUCT_NAME" >}}  processes can use more memory if that memory is required.
A rough number is to set `GOMEMLIMIT` to is 90% of the maximum memory required.
For example, if you want to keep memory usage below `10GiB`, use `GOMEMLIMIT=8GiB`.
```

docs/sources/reference/cli/environment-variables.md
```diff
@@ -35,4 +36,12 @@ Argument                   | Description                                     | D
...
+
+#### Automatic setting of GOMEMLIMIT
+
+The `GOMEMLIMIT` variable will be automatically set if {{< param "PRODUCT_NAME" >}} can determine the appropriate value. If the `GOMEMLIMIT` value is set before running  {{< param "PRODUCT_NAME" >}} then that setting will be honored. The value will be set to 90% of the cgroup value set. If the limit cannot be determined and no value was explicitly passed then no changes will occur.
```
clayton-cornell, line 45:
```suggestion
The `GOMEMLIMIT` environment variable is set automatically {{< param "PRODUCT_NAME" >}} can determine the appropriate value.
If you set the `GOMEMLIMIT` environment variable before you run  {{< param "PRODUCT_NAME" >}} then that setting will be used.
The variable value is set to 90% of the `cgroup` value set.
No changes will occur if the limit cannot be determined and you did not explicitly define a  `GOMEMLIMIT` value.
```


I'm not sure what this means: `The variable value is set to 90% of the cgroup value set.`  Are we referring to GOMEMLIMIT as a cvalue? or does this mean that you define GOMEMLIMIT to some value such as `8GiB` and it takes 90% of that? Or... is this the automatic determination part and this sentence belongs to the first sentence?

clayton-cornell (commented):
First pass at adjusting the text used to describe this env var and how its used.

docs/sources/reference/cli/environment-variables.md
mattdurham: If a cgroup value is available then the GOMEMLIMIT will automatically be set to 90% of the cgroup value.

docs/sources/reference/cli/environment-variables.md
```diff
@@ -35,4 +36,23 @@ Argument                   | Description                                     | D
...
+Don't treat the `GOMEMLIMIT` environment variable as a hard memory limit.
+{{< param "PRODUCT_NAME" >}}  processes can use more memory if that memory is required.
+A rough number is to set `GOMEMLIMIT` to is 90% of the maximum memory required.
+For example, if you want to keep memory usage below `10GiB`, use `GOMEMLIMIT=8GiB`.
+
+#### Automatic setting of GOMEMLIMIT
+
+The `GOMEMLIMIT` environment variable is set automatically {{< param "PRODUCT_NAME" >}} can determine the appropriate value.
+If you set the `GOMEMLIMIT` environment variable before you run  {{< param "PRODUCT_NAME" >}} then that setting will be used.
+The variable value is set to 90% of the `cgroup` value set.
```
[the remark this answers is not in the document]
mattdurham: Looks correct and it is the memory cgroup.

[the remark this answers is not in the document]
clayton-cornell: I've updated the suggestion

clayton-cornell:
@mattdurham this PR is getting stale. Do we want to move this info into the Env Vars topic? Or something else?

mattdurham:
Oh jeez, we should. I will get this updated and ready tomorrow.

mattdurham:
Updated from main. Fixed a logical typo I had in the doc.

clayton-cornell (approved):

clayton-cornell:
@mattdurham Is this one ready to merge?

PR #1189, opened 2024-07-03
Minor tidy on the new nonroot topic

clayton-cornell:
#### PR Description

Minor tidy on the new nonroot topic added in https://github.com/grafana/alloy/pull/1178 

Focus is on:

* Use the product variable: `{{< param "PRODUCT_NAME" >}}`
* Minor grammar cleanup to remove phrases like `In order to`
* Use "friendly" contractions like "won't", "it's", etc. per Grafana style
* Fix a broken x-ref in the Note

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/tasks/nonroot.md
+++ b/docs/sources/tasks/nonroot.md
@@ -5,41 +5,43 @@ title: Run Alloy as a non-root user in Kubernetes
 weight: 600
 ---
 
-# Run Alloy as a non-root user in Kubernetes
+# Run {{% param "PRODUCT_NAME" %}} as a non-root user in Kubernetes
 
-The [Alloy Docker image] contains two users:
+The [{{< param "PRODUCT_NAME" >}} Docker image][image] contains two users:
 
 * A `root` user.
 * A non-root user named `alloy` with uid `473` and gid `473`.
 
-By default, the `alloy` binary runs as `root`. This is because some Alloy components like [beyla.ebpf] require root permissions.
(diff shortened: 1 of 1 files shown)


docs/sources/tasks/nonroot.md
```diff
@@ -5,41 +5,43 @@ title: Run Alloy as a non-root user in Kubernetes
...
-```
+```yaml
 alloy:
   securityContext:
     runAsUser: 473
     runAsGroup: 473
 ```
 
-This will make the Alloy binary to run with uid 473 and gid 473 rather than as root.
+This configuration makes the {{< param "PRODUCT_NAME" >}} binary run with uid 473 and gid 473 rather than as root.
```
rfratto, line 36: should uid/gid be capitalized as UID/GID? they're technically acronyms, I believe?

rfratto (approved):

docs/sources/tasks/nonroot.md
clayton-cornell: I was wondering the same... if it's general UID and GID then it's capitalised... if it's the variable it'd be `uid and `gid`.  I'll fix it up with caps

clayton-cornell, line 36:
```suggestion
This configuration makes the {{< param "PRODUCT_NAME" >}} binary run with UID 473 and GID 473 rather than as root.
```

PR #1215, opened 2024-07-05
Refactor Alloy component reference TOC to group by type

clayton-cornell:
#### PR Description
Refactor Alloy component reference TOC to group by type

The changes in this PR are:
* Move the files to new grouped locations
* Add an alias to every topic so anything pointing at the old location still works
* Update the canonical URL metadata with the new path
* Top level index page added to every TOC entry point
* @Piotr fixed up the build so the generated content works again :slightly_smiling_face: 
* A few path corrections
* Fix broken links (moving topics down a level broke some outbound links)

image


thampiotr (approved):
I like this idea and it seems to work well

PR #1241, opened 2024-07-09
Update build-image-to-use to use 1.22.5.

mattdurham:
Update build image to 1.22.5

### DIFF
--- a/tools/make/build-container.mk
+++ b/tools/make/build-container.mk
@@ -34,7 +34,7 @@
 # variable names should be passed through to the container.
 
 USE_CONTAINER       ?= 0
-BUILD_IMAGE_VERSION ?= v0.1.2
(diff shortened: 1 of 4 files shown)


tools/make/build-container.mk
```diff
@@ -34,7 +34,7 @@
 # variable names should be passed through to the container.
 
 USE_CONTAINER       ?= 0
-BUILD_IMAGE_VERSION ?= v0.1.2
+BUILD_IMAGE_VERSION ?= v0.1.3
```
rfratto, line 37: Normally we update this in a follow up PR after the image is already available. Should this be pulled out of this PR?

mattdurham: Rolled back

rfratto (approved):

PR #1245, opened 2024-07-09
Add PodDisruptionBudget to helm chart

itspooya:
#### PR Description

This PR would add PodDisruptionBudget to the helm chart

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


itspooya:
are tests required for this?
if it is required, which api version to use for tests?

wildum:
Hey, thanks for contributing :)
Regarding your questions about the tests, please have a look at https://github.com/grafana/alloy/tree/main/operations/helm. 
The best would be to create one for each following the same pattern than the tests we already have

operations/helm/charts/alloy/README.md
```diff
@@ -85,6 +85,10 @@ useful if just using the default DaemonSet isn't sufficient.
 | controller.nodeSelector | object | `{}` | nodeSelector to apply to Grafana Alloy pods. |
 | controller.parallelRollout | bool | `true` | Whether to deploy pods in parallel. Only used when controller.type is 'statefulset'. |
 | controller.podAnnotations | object | `{}` | Extra pod annotations to add. |
+| controller.podDisruptionBudget | object | `{"enabled":false,"maxUnavailable":null,"minAvailable":null}` | PodDisruptionBudget configuration. |
```
wildum, line 88:
```suggestion
| controller.podDisruptionBudget | object | `{}` | PodDisruptionBudget configuration. |
```
This is redundant and can be removed

wildum:
I think that you could also create a new version, see this example: https://github.com/grafana/agent/pull/6393/files#diff-97d54f85324149687c80856ea9fb0f4aae8556574270efacbacf491012851f14

itspooya:
Fixes #1245

operations/helm/charts/alloy/README.md
itspooya: I will be applying the changes you mentioned soon

itspooya: Hey, When I removed this one, helm generate docs exits as you saw in the action / ci part

captncraig:
Question for you: what does this allow you to do that you can't already do with `controller.updateStrategy`? I've never used standalone pdbs before but my understanding is they are useful for matching multiple different controllers at once, but we only have the one here. There may be some other use case I am missing.

itspooya:
> Question for you: what does this allow you to do that you can't already do with `controller.updateStrategy`? I've never used standalone pdbs before but my understanding is they are useful for matching multiple different controllers at once, but we only have the one here. There may be some other use case I am missing.

There are multiple things that would be really beneficial to have the PDBs

pdbs are respected by kubernetes cluster management tools,
cluster-autoscaler or karpenter would respect the PDBs, lets say there is a node rotation and I want to have atleast 50% of alloy pods to be available, in this case node rotation(cluster autoscaler or karpenter) wouldn't try to do its thing that would disrupt our alloy cluster, this is more about those things and not the rollout(availability vs deployment)

sidewinder12s:
Yep, agreed that you need PDBs not just the update options. If you are using an Autoscaling solution like Karpenter, when consolidating/replacing nodes it will respect PDBs to ensure it does not take out too many pods that are backed by a PDB at once. 

This looks like what I would expect.

Fixes #1570

PR #1254, opened 2024-07-10
Remotecfg system attributes

erikbaranowski:
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"sync"
 	"time"
 
@@ -16,6 +17,7 @@ import (
 	collectorv1 "github.com/grafana/alloy-remote-config/api/gen/proto/go/collector/v1"
 	"github.com/grafana/alloy-remote-config/api/gen/proto/go/collector/v1/collectorv1connect"
 	"github.com/grafana/alloy/internal/alloyseed"
+	"github.com/grafana/alloy/internal/build"
 	"github.com/grafana/alloy/internal/component/common/config"
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
@@ -52,6 +54,8 @@ type Service struct {
 	ticker            *jitter.Ticker
 	dataPath          string
 	currentConfigHash string
+	systemAttrs       map[string]string
+	attrs             map[string]string
 	metrics           *metrics
 }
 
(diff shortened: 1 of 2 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -259,7 +277,12 @@ func (s *Service) Update(newConfig any) error {
 			newArgs.URL,
 		)
 	}
-	s.args = newArgs // Update the args as the last step to avoid polluting any comparisons
+	// Combine the new attributes on top of the system attributes
+	s.attrs = s.systemAttrs
+	maps.Copy(s.attrs, newArgs.Attributes)
```
spartan0x117, line 282: Does this unintentionally update `systemAttrs` with values from `newArgs.Attributes` for subsequent calls to `Update`?

Here's my thinking: 

`s.attrs = s.systemAttrs` makes `s.attrs` point to the same map in memory of the original `systemAttrs`. Copying the values from `newArgs.Attributes` into `s.attrs` (which is now the same pointer as `s.systemAttrs`) ultimately copies them into `s.systemAttrs`, which will then be the starting point for future calls.

So if you had the following:
```
remotecfg {
    attributes = {"foo" = "bar", "env" = "prod"}
}
```

then set it to:
```
remotecfg {
    attributes = {"foo" = "bar"} // Remove "env", it _shouldn't_ be present anymore
}
```
`"env"="prod"` would continue to be sent since `systemAttrs` contains it now.

I think this would fix it:
```suggestion
        newAttrs, _ := getSystemAttributes()
	maps.Copy(newAttrs, newArgs.Attributes)
	s.attrs = newAttrs
```

internal/service/remotecfg/remotecfg.go
```diff
@@ -131,9 +136,22 @@ func New(opts Options) (*Service, error) {
...
 	return &Service{
-		opts:   opts,
-		ticker: jitter.NewTicker(math.MaxInt64-baseJitter, baseJitter), // first argument is set as-is to avoid overflowing
+		opts:        opts,
+		systemAttrs: systemAttrs,
+		ticker:      jitter.NewTicker(math.MaxInt64-baseJitter, baseJitter), // first argument is set as-is to avoid overflowing
+	}, nil
+}
+
+func getSystemAttributes() (map[string]string, error) {
```
spartan0x117, line 151: Just a minor note: I noticed that the function signature includes an error return value, but it seems that it always returns nil. Is there a specific reason for this, or could the error be omitted?

erikbaranowski: that's a relic from when I had wired in host. I'll clean it

internal/service/remotecfg/remotecfg.go
erikbaranowski: I forgot to use maps.Clone here. I only want to call getSystemAttributes once on New so I don't have to recalculate anything now or added in the future.

erikbaranowski:
closing for now while we work through some details. I'll open a new PR when ready

PR #1258, opened 2024-07-11
Allow setting the CPU profiling event for Java Async Profiler in the `pyroscope.java` component

slbucur:
#### PR Description

The new `event` argument can be set to one of `itimer`, `cpu` and `wall`, like in the Pyroscope Java Agent:
https://grafana.com/docs/pyroscope/latest/configure-client/language-sdks/java/#configuration-options

Also add a parameter to allow enabling `per thread mode`, which should be used when setting the `wall` CPU event.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Main (unreleased)
 - (_Public preview_) Added rate limiting of cluster state changes to reduce the
   number of unnecessary, intermediate state updates. (@thampiotr)
 
+- Allow activating `wall` mode for Java Async Profiler (@slbucur)
+
 ### Bugfixes
 
 - Fixed a clustering mode issue where a fatal startup failure of the clustering service
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ Main (unreleased)
 - (_Public preview_) Added rate limiting of cluster state changes to reduce the
   number of unnecessary, intermediate state updates. (@thampiotr)
 
-- Allow setting the CPU profiling event for Java Async Profiler (@slbucur)
+- Allow setting the CPU profiling event for Java Async Profiler in `pyroscope.java` component (@slbucur)
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ Main (unreleased)
 - (_Public preview_) Added rate limiting of cluster state changes to reduce the
   number of unnecessary, intermediate state updates. (@thampiotr)
 
-- Allow activating `wall` mode for Java Async Profiler (@slbucur)
+- Allow setting the CPU profiling event for Java Async Profiler (@slbucur)
 
 ### Bugfixes
 
--- a/docs/sources/reference/components/pyroscope/pyroscope.java.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.java.md
@@ -84,16 +84,31 @@ The `profiling_config` block describes how async-profiler is invoked.
 
 The following arguments are supported:
 
-Name          | Type       | Description                                                                                                                                                                                                                          | Default | Required
---------------|------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|---------
-`interval`    | `duration` | How frequently to collect profiles from the targets.                                                                                                                                                                                 | "60s"   | no
-`cpu`         | `bool`     | A flag to enable cpu profiling, using `itimer` async-profiler event.                                                                                                                                                                 | true    | no
-`sample_rate` | `int`      | CPU profiling sample rate. It is converted from Hz to interval and passed as `-i` arg to async-profiler.                                                                                                                             | 100     | no
-`alloc`       | `string`   | Allocation profiling sampling configuration  It is passed as `--alloc` arg to async-profiler.                                                                                                                                        | "512k"  | no
-`lock`        | `string`   | Lock profiling sampling configuration. It is passed as `--lock` arg to async-profiler.                                                                                                                                               | "10ms"  | no
-`wall`        | `bool`     | Activates the `wall` mode when `cpu` profiling is enabled. This sample all threads equally every given period of time regardless of thread status: Running, Sleeping or Blocked. It is passed as `-e wall -t` arg to async-profiler. | false   | no
+Name          | Type       | Description                                                                                                 | Default  | Required
+--------------|------------|-------------------------------------------------------------------------------------------------------------|----------|---------
+`interval`    | `duration` | How frequently to collect profiles from the targets.                                                        | "60s"    | no
+`cpu`         | `bool`     | A flag to enable cpu profiling, using `itimer` async-profiler event by default.                             | true     | no
+`sample_rate` | `int`      | CPU profiling sample rate. It is converted from Hz to interval and passed as an `-i` arg to async-profiler. | 100      | no
+`alloc`       | `string`   | Allocation profiling sampling configuration  It is passed as an `--alloc` arg to async-profiler.            | "512k"   | no
+`lock`        | `string`   | Lock profiling sampling configuration. It is passed as an `--lock` arg to async-profiler.                   | "10ms"   | no
+`event`       | `string`   | Sets the CPU profiling event. Can be one of `itimer`, `cpu` or `wall`.                                      | "itimer" | no
+`per_thread`  | `bool`     | Sets per thread mode on async profiler. It is passed as an `-t` arg to async-profiler.                      | false    | no
 
 Refer to [profiler-options](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options) for more information about async-profiler configuration.
+
+#### `event` argument
+
+Sets the CPU profiling event:
+* `itimer` - default, uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) 
+   syscall which generates a signal every time a process consumes CPU 
+* `cpu` - uses PMU-case sampling (like Intel PEBS or AMD IBS), can be more accurate than `itimer`, but it's not available in every platform.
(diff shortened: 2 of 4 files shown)


internal/component/pyroscope/java/loop.go
```diff
@@ -183,7 +183,12 @@ func (p *profilingLoop) start() error {
 		"-o", "jfr",
 	)
 	if cfg.CPU {
-		argv = append(argv, "-e", "itimer")
+		if cfg.Wall {
+			argv = append(argv, "-e", "wall", "-t")
+		} else {
+			argv = append(argv, "-e", "itimer")
+		}
```
thampiotr, line 190: Since there are many other profiling events, I think we should not have `cfg.Wall bool`, but instead `cfg.Event string` which will default to `itimer` but can be set to any of the values supported. Unless there are some values that don't make sense in the context of Alloy and we should not allow them. cc @simonswine

thampiotr (commented):
Would be good to get some insight from the Pyroscope folks @simonswine 🙏

internal/component/pyroscope/java/loop.go
slbucur: Yes, we could have a `cfg.Event` with `itimer`, `cpu` and `wall` as possible values, and `itimer` as the default. That would be similar with the [java agent configuration](https://grafana.com/docs/pyroscope/latest/configure-client/language-sdks/java/#configuration-options).
>  PYROSCOPE_PROFILER_EVENT | Sets the profiler event. With JFR format enabled, this event refers to one of the possible CPU profiling events: itimer, cpu, wall. The default is itimer.

Is there a way to restrict `cfg.Event` to only those values?

Also, could we have `per thread mode` (`-t`) as a separate flag? From the [async profiler docs](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#wall-clock-profiling) it looks like it's recommended enable it when using the `wall` event.
> Wall-clock profiler is most useful in per-thread mode:` -t`.

thampiotr: Yes, you can limit the available options for a `string` field by implementing a `Validate` function on arguments. Example here: https://github.com/grafana/alloy/blob/db2036bde950540900ad1ad4e6c03ebf76c17ad1/internal/component/pyroscope/scrape/scrape.go#L217

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -84,13 +84,14 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+`sample_rate` | `int`      | CPU profiling sample rate. It is converted from Hz to interval and passed as `-i` arg to async-profiler.                                                                                                                             | 100     | no
+`alloc`       | `string`   | Allocation profiling sampling configuration  It is passed as `--alloc` arg to async-profiler.                                                                                                                                        | "512k"  | no
+`lock`        | `string`   | Lock profiling sampling configuration. It is passed as `--lock` arg to async-profiler.                                                                                                                                               | "10ms"  | no
+`wall`        | `bool`     | Activates the `wall` mode when `cpu` profiling is enabled. This sample all threads equally every given period of time regardless of thread status: Running, Sleeping or Blocked. It is passed as `-e wall -t` arg to async-profiler. | false   | no
```
clayton-cornell, line 94:
```suggestion
`wall`        | `bool`     | Activates the `wall` mode when `cpu` profiling is enabled. This samples all threads equally every given period of time regardless of the thread status: Running, Sleeping, or Blocked. It is passed as an `-e wall -t` argument to the async-profiler. | false   | no
```

Should we move the extra information about `wall` to a spot below the table (like we do with other extended descriptions)?

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -84,16 +84,32 @@ The `profiling_config` block describes how async-profiler is invoked.
...
 Refer to [profiler-options](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options) for more information about async-profiler configuration.
 
+#### `event` argument
+
+Sets the CPU profiling event:
+* `itimer` - default, uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) 
+   syscall which generates a signal every time a process consumes CPU 
+* `cpu` - uses PMU-case sampling (like Intel PEBS or AMD IBS), can be more accurate than `itimer`, but it's not available in every platform.
+* `wall` - This sample all threads equally every given period of time regardless of thread status: Running, Sleeping or Blocked.
+   For instance, this can be helpful when profiling application start-up time or IO intensive processes.
```
clayton-cornell, line 106:
```suggestion
Sets the CPU profiling event:
* `itimer` - Default. Uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) 
   syscall, which generates a signal every time a process consumes CPU. 
* `cpu` - Uses PMU-case sampling (like Intel PEBS or AMD IBS), can be more accurate than `itimer`, but it's not available on every platform.
* `wall` - This samples all threads equally every given period of time regardless of thread status: Running, Sleeping, or Blocked.
   For example, this can be helpful when profiling application start-up time or IO-intensive processes.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -84,16 +84,32 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+
+#### `per_thread` argument  
+Sets per thread mode on async profiler. Profile threads separately, each stack trace will end with a frame that denotes a single thread.
```
clayton-cornell, line 109: `Profile threads separately`? or `Profiler`?

slbucur: Updated to `Threads are profiled` 🙂. Original description was from here:
https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options

clayton-cornell: Ah that makes a LOT more sense now :-)

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -84,16 +84,32 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+
+Sets the CPU profiling event:
+* `itimer` - Default. Uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) 
+   syscall, which generates a signal every time a process consumes CPU. 
+* `cpu` - Uses PMU-case sampling (like Intel PEBS or AMD IBS), can be more accurate than `itimer`, but it's not available on every platform.
+* `wall` - This samples all threads equally every given period of time regardless of thread status: Running, Sleeping, or Blocked.
+   For example, this can be helpful when profiling application start-up time or IO-intensive processes.
+
+#### `per_thread` argument  
+Sets per thread mode on async profiler. Threads are profiled separately, each stack trace will end with a frame that denotes a single thread.
```
clayton-cornell, line 109:
```suggestion
Sets per thread mode on async profiler. Threads are profiled separately and each stack trace will end with a frame that denotes a single thread.
```

internal/component/pyroscope/java/args.go
```diff
@@ -29,6 +32,15 @@ func (rc *Arguments) UnmarshalAlloy(f func(interface{}) error) error {
 	return f((*config)(rc))
 }
 
+func (arg *Arguments) Validate() error {
+	switch arg.ProfilingConfig.Event {
+	case "itimer", "cpu", "wall":
+		return nil
+	default:
+		return fmt.Errorf("invalid event: '%s'. Event but be one of 'itimer', 'cpu' or 'wall'", arg.ProfilingConfig.Event)
```
thampiotr, line 40:
```suggestion
		return fmt.Errorf("invalid event: '%s'. Event must be one of 'itimer', 'cpu' or 'wall'", arg.ProfilingConfig.Event)
```

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Main (unreleased)
 - (_Public preview_) Added rate limiting of cluster state changes to reduce the
   number of unnecessary, intermediate state updates. (@thampiotr)
 
+- Allow setting the CPU profiling event for Java Async Profiler (@slbucur)
```
thampiotr, line 38:
```suggestion
- Allow setting the CPU profiling event for Java Async Profiler in `pyroscope.java` component (@slbucur)
```

thampiotr (approved):
LGTM, sadly there are not tests to update :(

thampiotr:
It seems we could also add support for multiple events, like `cpu,alloc,lock` as [described here](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#multiple-events). We should be able to add them as CSV to the existing string as a future enhancement. But if you think we need them right away, maybe it's better to support it in this PR and use an array of strings? WDYT?

slbucur:
> It seems we could also add support for multiple events, like `cpu,alloc,lock` as [described here](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#multiple-events). We should be able to add them as CSV to the existing string as a future enhancement. But if you think we need them right away, maybe it's better to support it in this PR and use an array of strings? WDYT?

I think `alloc` and `lock` are already enabled separately using the `--alloc` and `--lock` parameters:
https://github.com/grafana/alloy/blob/main/internal/component/pyroscope/java/loop.go#L190
It looks like they're enabled only if `cfg.Alloc` and `cfg.Lock`  are not empty strings, and they're enabled by default:
https://github.com/grafana/alloy/blob/main/internal/component/pyroscope/java/args.go#L50

slbucur:
Thank you for your reviews  @thampiotr and @clayton-cornell 😄. I've changed the PR title and description to match the new parameters.

clayton-cornell (approved):
Docs look OK

PR #1260, opened 2024-07-11
Add live debugging support to all otelcol receivers

### LINKED ISSUE #1040: Add live debugging support to otel receivers

wildum:
#### PR Description

This PR adds live debugging support to all otelcol receivers.

Fixes #1040 

#### Notes to the Reviewer

The PR follows the same pattern as the [live debugging support for processors](https://github.com/grafana/alloy/pull/1042).

I tested the following components:
- otelcol.receiver.otlp
- otelcol.receiver.prometheus
- otelcol.receiver.loki

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/internal/component/otelcol/receiver/receiver.go
+++ b/internal/component/otelcol/receiver/receiver.go
@@ -13,8 +13,10 @@ import (
 	otelcolCfg "github.com/grafana/alloy/internal/component/otelcol/config"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/fanoutconsumer"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/lazycollector"
+	"github.com/grafana/alloy/internal/component/otelcol/internal/livedebuggingconsumer"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/scheduler"
 	"github.com/grafana/alloy/internal/component/otelcol/internal/views"
+	"github.com/grafana/alloy/internal/service/livedebugging"
 	"github.com/grafana/alloy/internal/util/zapadapter"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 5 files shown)


internal/component/otelcol/receiver/receiver.go
```diff
@@ -152,19 +168,25 @@ func (r *Receiver) Update(args component.Arguments) error {
...
 
+	liveDebuggingActive := r.debugDataPublisher.IsActive(livedebugging.ComponentID(r.opts.ID))
+
 	if len(next.Traces) > 0 {
-		nextTraces := fanoutconsumer.Traces(next.Traces)
+		traces := next.Traces
+		if liveDebuggingActive {
+			traces = append(traces, r.liveDebuggingConsumer)
+		}
+		nextTraces := fanoutconsumer.Traces(traces)
```
mattdurham, line 189: Does it make sense/possible to build the debugging into the fanouts themselves?

mattdurham (approved):
Small comment on embedding it into our fanouts/appendables but looks good.

ptodev (approved):

internal/component/otelcol/receiver/receiver.go
wildum: Thanks for the suggestion, it would be ideal if we could just modify the fanouts and get support for all the components using them but the livedebugging service still needs to communicate with the components to notify them when a livedebugging session is opened/closed. So in the end I believe that it wouldn't improve the code much. Also the fanouts are copies from upstream. Modifying copied code makes dependency updates harder and prevents us from easily removing the copied code if it's moved out of an internal pkg from upstream. This is only from the top of my head, it could be worth digging through it to maybe find a simpler way

clayton-cornell (approved):

PR #1261, opened 2024-07-11
Rate limit how often components get notified about cluster changes

thampiotr:
#### PR Description

In a larger cluster we'd get a lot of cluster updates during startup that are propagated to all observing components. This leads to unnecessary processing at startup which can delay the startup and it can generate a lot of logs and tracing spans where it's not really helpful or required. In the extreme case, when cluster is very large, the number of log messages generated can be noticeable and some users have reported it previously as a bug.

With this PR, the call to Observe is rate-limited at 1 update per second. Intermediate updates while we're waiting for the limiter are not sent by `ckit`, but that's okay cause a call to Observe only notifies all the components, which subsequently get the most up-to-date list of peers from the service directly (or check data ownership with the service).

This change also adds a tracing span to allow observing this behaviour.

#### Notes to the Reviewer

If I query logs on a test cluster of 22 instances during a rollout using something like `sum by(pod) (count_over_time({cluster="...", namespace="..."} |= `peers changed` [1m]))` I will currently get a volume of logs that looks something like this:

image

After this change, it'll be something more like this:
image

Also, here's an example of traces:
image

### DIFF
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -248,6 +248,8 @@ func (fr *alloyRun) Run(configPath string) error {
 		AdvertiseInterfaces: fr.clusterAdvInterfaces,
 		ClusterMaxJoinPeers: fr.ClusterMaxJoinPeers,
 		ClusterName:         fr.clusterName,
+		//TODO(add #issue number): graduate to GA once we have more confidence in this feature
+		EnableStateUpdatesLimiter: fr.minStability.Permits(featuregate.StabilityPublicPreview),
 	})
 	if err != nil {
 		return err
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -248,7 +248,7 @@ func (fr *alloyRun) Run(configPath string) error {
 		AdvertiseInterfaces: fr.clusterAdvInterfaces,
 		ClusterMaxJoinPeers: fr.ClusterMaxJoinPeers,
 		ClusterName:         fr.clusterName,
-		//TODO(add #issue number): graduate to GA once we have more confidence in this feature
+		//TODO(alloy/#1274): graduate to GA once we have more confidence in this feature
 		EnableStateUpdatesLimiter: fr.minStability.Permits(featuregate.StabilityPublicPreview),
 	})
 	if err != nil {
--- a/internal/service/cluster/cluster.go
+++ b/internal/service/cluster/cluster.go
@@ -233,9 +233,9 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 				span.RecordError(err)
 			}
 			span.End()
-
-			// Grab a fresh view of the peers for logging, since we may have waited a bit for the limiter to permit.
-			peers = s.node.Peers()
+			// NOTE: after waiting for the limiter, the `peers` may be slightly outdated, but that's fine as the
+			// most up-to-date peers will be dispatched to the Observer by ckit eventually. The intermediate updates
+			// will be skipped, which is exactly what we want here.
 		}
 
 		if ctx.Err() != nil {
--- a/internal/service/cluster/cluster.go
+++ b/internal/service/cluster/cluster.go
@@ -218,7 +218,7 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 	defer cancel()
 
 	limiter := rate.NewLimiter(rate.Every(stateUpdateMinInterval), 1)
-	s.node.Observe(ckit.FuncObserver(func(_ []peer.Peer) (reregister bool) {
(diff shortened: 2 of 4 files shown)


wildum (approved):
LGTM, a nice use of tracing :) I could not find in the code where the intermediate updates are ignored (saw the comment but not the code that does it). If you found it could you share it please?

thampiotr:
> LGTM, a nice use of tracing :) I could not find in the code where the intermediate updates are ignored (saw the comment but not the code that does it). If you found it could you share it please?

It's a behaviour from ckit: https://github.com/grafana/ckit/blob/main/node.go#L526-L527

wildum:
> It's a behaviour from ckit
yeah I saw this comment but I was curious about how this was implemented and couldn't find this logic. I see that the run function takes the updates from the queue and broadcasts them to the observers (https://github.com/grafana/ckit/blob/main/node.go#L284) but I'm not seeing these updates being dropped while the observers are busy.

rfratto:
@wildum It's because notifying observers uses a queue with a size of one element: https://github.com/grafana/ckit/blob/f2ea0d78e4274f0a6a50cb2f59558bcd6a3f8990/node.go#L187

The way the ckit queue package is implemented [drops older elements](https://github.com/grafana/ckit/blob/f2ea0d78e4274f0a6a50cb2f59558bcd6a3f8990/internal/queue/queue.go#L124-L127) after reaching maximum size. 

If you combine this with how we only dequeue [after notifying observers](https://github.com/grafana/ckit/blob/f2ea0d78e4274f0a6a50cb2f59558bcd6a3f8990/node.go#L272), it means that intermediate events get dropped in favor of the latest event.

internal/alloycli/cmd_run.go
```diff
@@ -248,6 +248,8 @@ func (fr *alloyRun) Run(configPath string) error {
 		AdvertiseInterfaces: fr.clusterAdvInterfaces,
 		ClusterMaxJoinPeers: fr.ClusterMaxJoinPeers,
 		ClusterName:         fr.clusterName,
+		//TODO(add #issue number): graduate to GA once we have more confidence in this feature
```
rfratto, line 251: Should you do this before merge?

internal/service/cluster/cluster.go
```diff
@@ -210,23 +217,40 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
+			_, span := tracer.Start(spanCtx, "RateLimitWait", trace.WithSpanKind(trace.SpanKindInternal))
+			if err := limiter.Wait(ctx); err != nil {
+				// This should never happen, but it should be safe to just ignore it and continue.
+				level.Warn(s.log).Log("msg", "failed to wait for rate limiter on peers update", "err", err)
+				span.RecordError(err)
+			}
+			span.End()
+
+			// Grab a fresh view of the peers for logging, since we may have waited a bit for the limiter to permit.
+			peers = s.node.Peers()
```
rfratto, line 238: Hm, this won't cause any bugs but it's really not the intended way to use Observe :thinking: If there are new peers, Observe will be called again and it'll catch up over time. 

I'm fine with this though.

internal/service/cluster/cluster.go
```diff
@@ -210,23 +217,40 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
 			// Unregister our observer if we exited.
 			return false
 		}
 
-		tracer := s.tracer.Tracer("")
-		spanCtx, span := tracer.Start(ctx, "NotifyClusterChange", trace.WithSpanKind(trace.SpanKindInternal))
-		defer span.End()
-
 		s.logPeers("peers changed", toStringSlice(peers))
+		span.SetAttributes(attribute.Int("peers_count", len(peers)))
```
rfratto, line 247: Adding the peer count as an attribute is a great idea :+1:

rfratto (approved):

internal/alloycli/cmd_run.go
rfratto: Haha sorry, I mean should you open the issue and add the issue number here before merge

internal/service/cluster/cluster.go
thampiotr: Yeah, if we're fine with the logged peers being potentially not the most up-to-date, I'd be fine with getting rid of this. It may just be confusing for the next person debugging this.

rfratto: I'd be cool with either keeping this in or replacing it with a comment explaining what I just said (that there may have been new events in the last second and Observe will be called again to catch up)

thampiotr: Sounds good, after thinking about this I opted to not re-lookup the peers and add a comment.

PR #1285, opened 2024-07-15
Live debugging: loki process

### LINKED ISSUE #1276: Add live debugging support to loki.process

wildum:
#### PR Description

This PR adds live debugging support to loki.process.
It prints the entries with labels when they enter the component and when they exit the component so that the user can make sure that the transformation is done as expected. 

Fixes #1276 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/internal/component/loki/process/process.go
+++ b/internal/component/loki/process/process.go
@@ -148,6 +148,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 func (c *Component) handleIn(ctx context.Context, wg *sync.WaitGroup) {
 	defer wg.Done()
+	componentID := livedebugging.ComponentID(c.opts.ID)
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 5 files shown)


internal/component/loki/process/process.go
```diff
@@ -147,7 +158,10 @@ func (c *Component) handleIn(ctx context.Context, wg *sync.WaitGroup) {
 			case <-ctx.Done():
 				return
 			case c.processIn <- entry.Clone():
-				// no-op
+				componentID := livedebugging.ComponentID(c.opts.ID)
```
thampiotr, line 161: Do we need to create new `ComponentID` each time?

thampiotr (approved):
One nit, but LGTM.

internal/component/loki/process/process.go
wildum: good point, moved it out of the for loop

PR #1286, opened 2024-07-15
Log error on failed component Run

thampiotr:
#### PR Description

Improve debuggability of Alloy by logging errors that cause components to exit their `Run` functions.

### DIFF
--- a/internal/runtime/alloy.go
+++ b/internal/runtime/alloy.go
@@ -51,6 +51,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/runtime/internal/controller"
 	"github.com/grafana/alloy/internal/runtime/internal/worker"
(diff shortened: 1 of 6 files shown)


rfratto:
Funnily enough components themselves aren't the biggest issue here since BuiltinComponentNodes do log on exit: https://github.com/grafana/alloy/blob/b6f84f124c46c78349433960cb580bac945e2264/internal/runtime/internal/controller/node_builtin_component.go#L317-L327

If we're going to log at the scheduler level I think we should remove the logging above in favor of this. 

Also, the scheduler is reused for running services, so saying "component" in log lines may be misleading; we may want to make it more generic (like "node" for example)

thampiotr:
> Funnily enough components themselves aren't the biggest issue here since BuiltinComponentNodes do log on exit:
> 
> https://github.com/grafana/alloy/blob/b6f84f124c46c78349433960cb580bac945e2264/internal/runtime/internal/controller/node_builtin_component.go#L317-L327
> 
> If we're going to log at the scheduler level I think we should remove the logging above in favor of this.
> 
> Also, the scheduler is reused for running services, so saying "component" in log lines may be misleading; we may want to make it more generic (like "node" for example)

Ah, right! I do recall seeing this log line that you bring this up. I do think one central place is better, but also don't want the duplication. Will fix.

thampiotr:
@rfratto I've fixed the wording and de-duplicated the log line now.

wildum:
The log lines should also be removed from the CustomComponentNode and ImportConfigNode to avoid the de-duplication

wildum (approved):

PR #1292, opened 2024-07-15
Fix bug with loki.process metrics stage during config reload

ptodev:
#### PR Description

This bugfix is ported from a [PR](https://github.com/grafana/agent/pull/6971) which was already merged to Grafana Agent. The behaviour in Alloy is different from Agent Static mode. Static mode would always log an error during config reload. Alloy doesn't log any errors. Also, reloading the config if it hasn't changed works ok and the metrics stage keeps working fine. However, if the metrics stage is changed to remove certain metrics, those metrics will still be present even after the next config reload. Their values will remain frozen in time. If they are added to the config later, then they won't resume working - they'll remain frozen.

Related to [an issue](https://github.com/grafana/agent/issues/2754) in the Agent repo.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ Main (unreleased)
 - Fixed a clustering mode issue where a fatal startup failure of the clustering service
   would exit the service silently, without also exiting the Alloy process. (@thampiotr)
 
+- Fix a bug which prevented config reloads to work if a Loki `metrics` stage is in the pipeline. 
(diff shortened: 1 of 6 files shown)


clayton-cornell:
Doc changes look OK

thampiotr (approved):
Looks good, other than the tests need fixing

PR #1300, opened 2024-07-15
Add `otelcol.processor.groupbyattrs` component

### LINKED ISSUE #224: Add `otelcol.processor.groupbyattrs` component
### Request

Add the OTEL module [groupbyattrsprocessor](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/processor/groupbyattrsprocessor#section-readme).

### Use case

I have an agent scraping prometheus metrics from a kubernetes kubelet, whose metrics I also pass through an `otelcol.processor.k8sattributes` step (and a `otelcol.processor.transform` to turn otel resource attributes to datapoint attributes, [as mentioned in the `otelcol.processor.k8sattributes` docs](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.processor.k8sattributes/#adding-kubernetes-metadata-to-prometheus-metrics)). What I want is basically what's stated in that doc section, to enrich the kubelet prometheus metrics (with labels like `pod="myapp-59fb8bb44d-6qfgf",namespace="default"`) with metadata that becomes extra labels (like `k8s_deployment_name="myapp"`). The pipeline looks like this:

```text
prometheus.scrape 
  -> otelcol.receiver.prometheus 
    -> otelcol.processor.k8sattributes 
      -> otelcol.processor.transform 
        -> otelcol.exporter.prometheus
```

However, with that pipeline, I get attributes from the k8sattributes processor that _don't match up_ with the labels reported by kubelet. For example, I'll get a datapoint with `namespace="default",k8s_namespace_name="kube-system"` (where `namespace` is from kubelet and `k8s_namespace_name` is from k8sattributes). Also, most datapoints seem "sticky" to particular label values, i.e. many pods across many namespaces will all get a `k8s_namespace_name="kube-system"` label.

What I believe is happening is that in the otel components, all the datapoints are associated with a single Resource for the single `prometheus.scrape` job. So, the k8sattributes step will talk to the Kubernetes API and update the otel Resource, but because the resource is shared, there are conflicts and eventually the last (or the first) write wins and all the other datapoints get those labels.

If that's what's happening, adding the `groupbyattrsprocessor` module should at least let me fix this pipeline so that the k8sattributes enrichment works as expected.

kehindesalaam:
#### PR Description
Adds the `otelcol.processor.groupbyattrs` component which allows grouping of metrics matching specific attributes.

Fixes #224

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell:
@grafana/grafana-alloy-maintainers  Can we do a code review before I review the doc?

go.mod
```diff
@@ -707,6 +707,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require github.com/open-telemetry/opentelemetry-collector-contrib/processor/groupbyattrsprocessor v0.104.0
```
wildum, line 710: the version should match the current version of our otel dependency. We update all components at the same time before every release. The current version is `v0.102.1`

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/groupbyattrsprocessor"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor"
+	otelcomponent "go.opentelemetry.io/collector/component"
+	otelextension "go.opentelemetry.io/collector/extension"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:      "otelcol.processor.groupbyattrs",
+		Stability: featuregate.StabilityExperimental,
```
wildum, line 20: the upstream component is in stability beta which is quite stable in opentelemetry terms. I think we can put it directly to PublicPreview or even to GenerallyAvailable. What do you think @ptodev ?

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,92 @@
...
+		Exports:   otelcol.ConsumerExports{},
+		Args:      Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			fact := k8sattributesprocessor.NewFactory()
```
wildum, line 25: this is the wrong factory

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,92 @@
...
+type Arguments struct {
+
+	// Keys is a list of attributes to group metrics by.
+	Keys []string `alloy:"keys,attr,optional"`
+
+	// Output configures where to send processed data. Required.
+	Output *otelcol.ConsumerArguments `alloy:"output,block"`
+
+	// DebugMetrics configures component internal metrics. Optional.
+	DebugMetrics otelcolCfg.DebugMetricsArguments `alloy:"debug_metrics,block,optional"`
```
wildum, line 40: this block should be documented (like the other otel components supporting this block)

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+var (
+	_ processor.Arguments = Arguments{}
+
+	// DefaultArguments holds default settings for Arguments.
+	DefaultArguments = Arguments{
+		Keys: []string{},
```
wildum, line 48: I don't think that we need to specify this because it will be empty by default

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,92 @@
...
+func (args *Arguments) Validate() error {
+	return nil
+}
+
+// Convert implements processor.Arguments.
+func (args Arguments) Convert() (otelcomponent.Config, error) {
+	return args.convertImpl()
+}
+
+func (args Arguments) convertImpl() (*groupbyattrsprocessor.Config, error) {
```
wildum, line 68: I don't think that this function is needed, you could return the config directly in Convert()

wildum (commented):
Thanks for your contribution! I left a few comments. Let me know when this is ready for a second review :)

wildum:
You will also need to add a config converter for the component. You can follow the pattern used for the other otel components in  https://github.com/grafana/alloy/tree/main/internal/converter/internal/otelcolconvert

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+`otelcol.processor.groupbyattrs` accepts telemetry data from other `otelcol`
+components and reassociates spans, log records and metric datapoints to a Resource that matches with the specified attributes. It groups telemetry data by specified attributes.
+
+{{% admonition type="note" %}}
+`otelcol.processor.groupbyattrs` is a wrapper over the upstream OpenTelemetry
+Collector `groupbyattrs` processor.  If necessary, bug reports or feature requests
+will be redirected to the upstream repository.
+{{% /admonition %}}
+
+It is recommended to use the groupbyattrs processor together with batch processor, as a consecutive step, as this will reduce the fragmentation of data (by grouping records together under matching Resource/Instrumentation Library)
```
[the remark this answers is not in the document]
kehindesalaam: I'll do that thank you.

go.mod
kehindesalaam: I'll make the change

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
kehindesalaam: does it matter if it is nil or non-nil?

wildum: I checked the Otel code and I don't think that it matters. I see indeed that they initialize it to an empty slice but IMO because it's a config it makes more sense to leave it as nil (in general I would initialize slices as non nil if there is a chance that a value might be added later but this is not the case here).
That being said, it does not really matter in this case. If you think that it feels safer to have it non-nil to match with the default in the otel library I'm fine with it

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
+# otelcol.processor.groupbyattrs
```
clayton-cornell, line 1:
```suggestion
---
canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.groupbyattrs/
description: Learn about otelcol.processor.groupbyattrs
title: otelcol.processor.groupbyattrs
---

# otelcol.processor.groupbyattrs
```

We need at least this basic metadata

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+
+`otelcol.processor.groupbyattrs` accepts telemetry data from other `otelcol`
+components and reassociates spans, log records and metric datapoints to a Resource that matches with the specified attributes. It groups telemetry data by specified attributes.
```
clayton-cornell, line 4:
```suggestion
components and reassociates spans, log records, and metric datapoints to a resource that matches the specified attributes. It groups telemetry data by specified attributes.
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+
+{{% admonition type="note" %}}
+`otelcol.processor.groupbyattrs` is a wrapper over the upstream OpenTelemetry
+Collector `groupbyattrs` processor.  If necessary, bug reports or feature requests
```
clayton-cornell, line 8:
```suggestion
Collector `groupbyattrs` processor. If necessary, bug reports or feature requests
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
clayton-cornell, line 12: There are probably some other changes that will be required here once the other update is applied

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+`keys` | `array(string)` | Keys that will be used to group the spans, log records or metric data points together | | no
+
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name    | Type               | Description
+--------|--------------------|-----------------------------------------------------------------
+`input` | `otelcol.Consumer` | A value that other components can use to send telemetry data to.
```
clayton-cornell, line 44:
```suggestion
`input` | `otelcol.Consumer` |  Accepts `otelcol.Consumer` data for metrics, logs, or traces.
```

I'm not sure about this description. I don't think components can send data to a value... what if we pulled up the extended description. If this works, we could remove lines 46 and 47.

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+```
+
+## Notes
+- The data points with different data types would not be merged under the same metric (e.g a gauge and sum metric would not be merged).
+- The data points without the specified keys would remain under their respective resources.
+- New resources would inherit the attributes of the original resource and the specified attributes in the keys array.
+- The grouping attributes in the keys array would be removed from the output metrics.
+- If the keys array is empty, the processor would perform compaction and re-associate all spans with matching Resource and InstrumentationLibrary.
```
clayton-cornell, line 88:
```suggestion
- The data points with different data types aren't merged under the same metric. For example, a gauge and sum metric would not be merged.
- The data points without the specified keys remain under their respective resources.
- New resources inherit the attributes of the original resource and the specified attributes in the keys array.
- The grouping attributes in the keys array are removed from the output metrics.
- If the keys array is empty, the processor performs compaction and reassociates all spans with matching Resource and InstrumentationLibrary.
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,88 @@
...
+
+  output {
+    metrics = [otelcol.exporter.otlp.default.input]
+    logs    = [otelcol.exporter.otlp.default.input]
+    traces  = [otelcol.exporter.otlp.default.input]
+  }
+}
+```
+
+## Notes
```
clayton-cornell, line 83: Non-blocking commen.
I'm not sure that this info is best grouped under a `Notes` heading.  I'll leave it as-is... just noting it here in the review that this is not used elsewhere, and it might be better to rethink how this information is presented. It's something we can put off to a future PR.

clayton-cornell (commented):
Some doc suggestions

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
kehindesalaam: I'll look into that

kehindesalaam: There is no extended description for this section. All the files have copies of the same thing. It is a good ideaa to make this part into a separate file. I would assume it is the same for all otelcol processors

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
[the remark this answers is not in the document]
kehindesalaam: done

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
kehindesalaam: I made it GenerallyAvailable

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
kehindesalaam: I added a doc for this

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
kehindesalaam: removed

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
kehindesalaam: I moved the last note as part of the documentation for the Keys attribute

kehindesalaam:
@wildum could you please take another look?

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name            | Type              | Description                                                                           | Default | Required |
+|-----------------|-------------------|---------------------------------------------------------------------------------------|---------|----------|
+| `keys`          | `array(string)`   | Keys that will be used to group the spans, log records or metric data points together |         | no       |
```
clayton-cornell, line 41:
```suggestion
| `keys`          | `array(string)`   | Keys that will be used to group the spans, log records, or metric data points together. |         | no       |
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
...
+`otelcol.processor.groupbyattrs` accepts telemetry data from other `otelcol`
+components and reassociates spans, log records, and metric datapoints to a resource that matches the specified attributes. It groups telemetry data by specified attributes.
+
+{{% admonition type="note" %}}
+`otelcol.processor.groupbyattrs` is a wrapper over the upstream OpenTelemetry
+Collector `groupbyattrs` processor. If necessary, bug reports or feature requests
+will be redirected to the upstream repository.
+{{% /admonition %}}
+
+It is recommended to use the groupbyattrs processor together with [otelcol.processor.batch][], as a consecutive step, as this will reduce the fragmentation of data (by grouping records together under matching Resource/Instrumentation Library)
```
clayton-cornell, line 18:
```suggestion
We recommend you use the groupbyattrs processor together with [otelcol.processor.batch][], as a consecutive step. This will reduce the fragmentation of data by grouping records together under the matching Resource/Instrumentation Library.
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
...
+logs, or traces).
+
+## Component health
+
+`otelcol.processor.groupbyattrs` is only reported as unhealthy if given an invalid
+configuration.
+
+## Debug information
+
+`otelcol.processor.groupbyattrs` does not expose any component-specific debug
```
clayton-cornell, line 78:
```suggestion
`otelcol.processor.groupbyattrs` doesn't expose any component-specific debug
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
...
+information.
+
+## Debug metrics
+
+`otelcol.processor.groupbyattrs` does not expose any component-specific debug metrics.
```
clayton-cornell, line 83:
```suggestion
`otelcol.processor.groupbyattrs` doesn't expose any component-specific debug metrics.
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.groupbyattrs/
+description: Learn about otelcol.processor.groupbyattrs
+title: otelcol.processor.groupbyattrs
+---
+
+# otelcol.processor.groupbyattrs
+
+`otelcol.processor.groupbyattrs` accepts telemetry data from other `otelcol`
+components and reassociates spans, log records, and metric datapoints to a resource that matches the specified attributes. It groups telemetry data by specified attributes.
```
wildum, line 10: I would maybe remove this sentence "It groups telemetry data by specified attributes.". It does not add a lot of info. If we think that this is not clear enough, maybe an example would help

wildum (commented):
the code looks good to me, I just have some little feedback on the doc. I found that the examples in the original doc were helpful to understand how the processor work. Should we add some examples or is it already clear enough? @ptodev I think you also wanted to give some feedback on this PR?

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,111 @@
...
+    traces  = [otelcol.exporter.otlp.default.input]
+  }
+}
+```
+
+## Notes
+- The data points with different data types aren't merged under the same metric. For example, a gauge and sum metric would not be merged.
+- The data points without the specified keys remain under their respective resources.
+- New resources inherit the attributes of the original resource and the specified attributes in the keys array.
+- The grouping attributes in the keys array are removed from the output metrics.
```
ptodev, line 109:
```suggestion
### Grouping metrics

Consider the below metrics, all originally associated to the same Resource:

``
Resource {host.name="localhost",source="prom"}
  Metric "gauge-1" (GAUGE)
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-B",id="eth0"}
  Metric "gauge-1" (GAUGE) // Identical to previous Metric
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-B",id="eth0"}
  Metric "mixed-type" (GAUGE)
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-B",id="eth0"}
  Metric "mixed-type" (SUM)
    DataPoint {host.name="host-A",id="eth0"}
    DataPoint {host.name="host-A",id="eth0"}
  Metric "dont-move" (Gauge)
    DataPoint {id="eth0"}
``

With the below configuration, the groupbyattrs will re-associate the metrics with either `host-A` or `host-B`, based on the value of the `host.name` attribute.

``alloy
otelcol.processor.groupbyattrs "default" {
  keys = [ "host.name" ]

  output {
    metrics = [otelcol.exporter.otlp.default.input]
  }
}
``

The output of the processor will therefore be:

``
Resource {host.name="localhost",source="prom"}
  Metric "dont-move" (Gauge)
    DataPoint {id="eth0"}

Resource {host.name="host-A",source="prom"}
  Metric "gauge-1"
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}
  Metric "mixed-type" (GAUGE)
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}
  Metric "mixed-type" (SUM)
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}

Resource {host.name="host-B",source="prom"}
  Metric "gauge-1"
    DataPoint {id="eth0"}
    DataPoint {id="eth0"}
  Metric "mixed-type" (GAUGE)
    DataPoint {id="eth0"}
``

This output demonstrates how `otelcol.processor.groupbyattrs` works in various situations:

- The DataPoints for the `gauge-1` (GAUGE) metric were originally split under 2 Metric instances and have been merged in the output.
- The DataPoints of the `mixed-type` (GAUGE) and `mixed-type` (SUM) metrics have not been merged under the same Metric, because their DataType is different.
- The `dont-move` metric DataPoints don't have a `host.name` attribute and therefore remained under the original Resource.
- The new Resources inherited the attributes from the original Resource (`source="prom"`), plus the specified attributes from the processed metrics (`host.name="host-A"` or `host.name="host-B"`).
- The specified "grouping" attributes that are set on the new Resources are also removed from the metric DataPoints.
- While not shown in the above example, the processor also merges collections of records under matching InstrumentationLibrary.
```

@clayton-cornell does this look ok? I copied it from the [upstream docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.105.0/processor/groupbyattrsprocessor/README.md#grouping-metrics).

@kehindesalaam Note that you'd need to replace the `` with ```. I can't do it in the GitHub suggestion, because I can't escape the tick symbol inside suggestions.

internal/component/otelcol/processor/groupbyattrs/groupbyattrs.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// DefaultArguments holds default settings for Arguments.
+	DefaultArguments = Arguments{
+		Keys: []string{},
+	}
+)
+
+// SetToDefault implements syntax.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = DefaultArguments
```
ptodev, line 53:
```suggestion
	*args = Arguments{
		Keys: []string{},
	}
```

I think this could avoid issues with shallow copying of slices.

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
clayton-cornell: @ptodev You can use codefence markdown in suggestions.. just use 4 backticks to wrap the suggestion block and the standard 3 backticks for code bloack.

Instead of `the below`, we can use `the following` (better grammatically).  I'll look at the info-flow a little deeper once it's applied and do another pass on the docs.

We have to be careful with 1:1 copied text from upstream - we should either directly link to it or note it's source (per whatever license upstream is using) if we copy it.

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
kehindesalaam: I would remove that

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
kehindesalaam: I'll make the modification and also the adjustment Clayton suggested.

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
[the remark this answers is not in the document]
kehindesalaam: I'll run this as well.

kehindesalaam:
Thanks for the review team. I'm sorry the second set of changes came a bit later. I have been busy.

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,175 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name            | Type              | Description                                                                           | Default | Required |
+|-----------------|-------------------|---------------------------------------------------------------------------------------|---------|----------|
+| `keys`          | `list(string)`    | Keys that will be used to group the spans, log records or metric data points together |         | no       |
```
ptodev, line 41:
```suggestion
| `keys`          | `list(string)`    | Keys that will be used to group the spans, log records or metric data points together. |         | no       |
```

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
```diff
@@ -0,0 +1,175 @@
...
+
+This output demonstrates how `otelcol.processor.groupbyattrs` works in various situations:
+
+- The DataPoints for the `gauge-1` (GAUGE) metric were originally split under 2 Metric instances and have been merged in the output.
+- The DataPoints of the `mixed-type` (GAUGE) and `mixed-type` (SUM) metrics have not been merged under the same Metric, because their DataType is different.
+- The `dont-move` metric DataPoints don't have a `host.name` attribute and therefore remained under the original Resource.
+- The new Resources inherited the attributes from the original Resource (`source="prom"`), plus the specified attributes from the processed metrics (`host.name="host-A"` or `host.name="host-B"`).
+- The specified "grouping" attributes that are set on the new Resources are also removed from the metric DataPoints.
+- While not shown in the above example, the processor also merges collections of records under matching InstrumentationLibrary.
+
```
ptodev, line 156:
````suggestion

### Compaction

Sometimes telemetry data can become fragmented due to multiple duplicated ResourceSpans/ResourceLogs/ResourceMetrics objects.
This leads to additional memory consumption, increased processing costs, inefficient serialization and increase of the export requests. 
In such situations, `otelcol.processor.groupbyattrs` can be used to compact the data with matching Resource and InstrumentationLibrary properties.

For example, consider this input data:

```
Resource {host.name="localhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=1, ...}
  InstrumentationLibrary {name="OtherLibrary"}
  Spans
    Span {span_id=2, ...}
    
Resource {host.name="localhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=3, ...}
    
Resource {host.name="localhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=4, ...}
    
Resource {host.name="otherhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=5, ...}
```

You can use `otelcol.processor.groupbyattrs` with its default configuration to compact the data: 
```alloy
otelcol.processor.groupbyattrs "default" {
  output {
    metrics = [otelcol.exporter.otlp.default.input]
  }
}
```

The output will be:

```
Resource {host.name="localhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=1, ...}
    Span {span_id=3, ...}
    Span {span_id=4, ...}
  InstrumentationLibrary {name="OtherLibrary"}
  Spans
    Span {span_id=2, ...}

Resource {host.name="otherhost"}
  InstrumentationLibrary {name="MyLibrary"}
  Spans
    Span {span_id=5, ...}
```
````

This example was taken from the [OTel docs](https://pkg.go.dev/github.com/open-telemetry/opentelemetry-collector-contrib/processor/groupbyattrsprocessor#section-readme).

File: docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md [hunk 23]
```diff
@@ -0,0 +1,175 @@
...
+[otelcol.processor.batch]: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.batch/
```
ptodev, line 157:
```suggestion
[otelcol.processor.batch]: ../otelcol.processor.batch/
```

internal/converter/internal/otelcolconvert/testdata/groupbyattrs.alloy
ptodev, line 1: @kehindesalaam The test related to this file is failing. I think you'll need to rebase your branch with `main` in order to fix it.

ptodev (commented):
Thank you, @kehindesalaam! When @clayton-cornell approves it and when the CI is green, I'm happy for the PR to be merged!

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
clayton-cornell: @ptodev We have to be super careful about taking (copy/pasting) content - any content - from external sources including the OTel.  We are probably OK here, but this is something we need to keep in mind when pulling info from external sources :-)  https://developers.google.com/style/other-sources

clayton-cornell: Just t be clear, this suggestion should be OK as-is.

internal/converter/internal/otelcolconvert/testdata/groupbyattrs.alloy
kehindesalaam: rebased

docs/sources/reference/components/otelcol/otelcol.processor.groupbyattrs.md
ptodev: Hi, @kehindesalaam when the section from this suggestion is added, I'll give the PR a final look and will likely ✅  it. Thank you for your contributions!

kehindesalaam: I missed this 😅

Allzz77123 (approved):

ptodev (approved):
Thank you! Looks good to me. 🚢  I just pushed a commit on your branch with a few minor docs fixes. By the way, you can visualise the docs in a browser by running these commands:
```
cd docs
make docs
```
I find this very useful when I want to make sure everything gets rendered appropriately.

clayton-cornell (approved):
Doc are ok

Issue #1208, opened 2024-07-05
[clustering] failure to discover peers when the number of instances and exposed ports is large


thampiotr:
## Issue

Alloy uses a headless service and [queries DNS for SRV records](https://github.com/grafana/alloy/blob/main/internal/alloycli/cluster_builder.go#L172) to discover peer instances, which can be used to join the cluster via [memberlist](https://github.com/hashicorp/memberlist)-based gossip protocol provided by the [ckit](https://github.com/grafana/ckit) library.

These queries were in some reported cases returning zero results, which is evidenced by a periodic logging of msg=”rejoining peers”, peers=””

The standard [Go DNS library](https://go.googlesource.com/go/+/refs/heads/master/src/net/lookup.go#493) used by Alloy was returning no results for the query.
In contrast, queries run using dig CLI utility were still returning valid results.
This indicates that there was a bug or implementation difference between Go DNS library and other DNS clients.

The DNS queries are limited in size. The query is first attempted via UDP, when the response from DNS server indicates that the message was truncated, the query is retried over TCP. 
However, even the TCP query can be truncated. This is because the TCP messages still wrap the [RFC1035 DNS](https://www.rfc-editor.org/rfc/rfc1035#section-4.2.2) message which is prefixed with a two byte length field, so the maximum length of DNS response over TCP is 65536.

When a TCP message is truncated, the Truncated flag is set to 1.
![image](https://github.com/grafana/alloy/assets/17101802/aa030c30-8937-4412-9437-cd35588f5528)

However, as seen above, there are still useful Answer RRs in the response, which can be used on the “best effort” basis by the clients.
dig does this, but Go didn’t due to this bug: https://github.com/golang/go/issues/64896 
The fix for this issue should be available in go1.23

But why are the DNS query responses so large? When Alloy cluster has, for example, 90 instances and for each of these instances, there are 10 different ports registered in the headless service, there will be 900 SRV records registered in DNS.

This is quite common - initially, the Alloy helm chart only supported one port, but [support for more ports was added later](https://github.com/grafana/alloy/commit/4a32c4477419eedabd556b723898d28d12c52d7a). 
There are 9 ports defined in the [k8s-monitoring helm](https://github.com/grafana/k8s-monitoring-helm/blob/main/charts/k8s-monitoring/values.yaml#L2055) chart.

## Proposed fixes

There are a few tasks that should be taken to resolve this issue:

```[tasklist]
### Tasks
[...]

PR #1311, opened 2024-07-17
Tests and foundations for new cluster peer discovery

thampiotr:
#### PR Description

As part of this issue investigation: https://github.com/grafana/alloy/issues/1208 we found that there are missing tests and a few inconsistencies in the way that cluster peer discovery currently works.

Since this is a critical feature used by large scale users, we want to ensure that these are well-tested paths and bugs and inconsistencies are eliminated. 

In order to isolate these changes and future refactoring, this PR:
- Introduces a new package and functions that perform peer discovery when PublicPreview stability level is used.
- The new package is a copy of the existing behaviour for now, but is more testable.
- Tests are added with some comments about the discovered problems and inconsistencies (see TODO comments). 

This sets a foundation for further changes:
- Use the A/AAAA records for DNS discovery
- Reduce the inconsistencies and issues mentioned in the tests (see TODO comments)

#### PR Checklist

- [x] Tests updated

### DIFF
--- a/docs/sources/reference/cli/run.md
+++ b/docs/sources/reference/cli/run.md
@@ -51,6 +51,7 @@ The following flags are supported:
 * `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. Set to `all` to use all available network interfaces on the system. (default `"eth0,en0"`).
 * `--cluster.max-join-peers`: Number of peers to join from the discovered set (default `5`).
 * `--cluster.name`: Name to prevent nodes without this identifier from joining the cluster (default `""`).
+* `--cluster.use-discovery-v1`: Use the older, v1 version of cluster peer discovery mechanism (default `false`). Note that this flag will be deprecated in the future and eventually removed.
 * `--config.format`: The format of the source file. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, `static` (default `"alloy"`).
 * `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
 * `--config.extra-args`: Extra arguments from the original format used by the converter.
@@ -136,6 +137,9 @@ The `--cluster.name` flag can be used to prevent clusters from accidentally merg
 When `--cluster.name` is provided, nodes will only join peers who share the same cluster name value.
 By default, the cluster name is empty, and any node that doesn't set the flag can join.
 Attempting to join a cluster with a wrong `--cluster.name` will result in a "failed to join memberlist" error.
+
+The `--cluster.use-discovery-v1` flag can be used to switch back to the older, v1 version of the cluster peer discovery mechanism
+in case of any issues with the newer version. This flag will be deprecated in the future and eventually removed.
 
 ### Clustering states
 
--- a/internal/alloycli/cluster_builder.go
+++ b/internal/alloycli/cluster_builder.go
@@ -71,7 +71,7 @@ func buildClusterService(opts clusterOptions) (*cluster.Service, error) {
 	}
 
 	// New, refactored and improved peer discovery.
-	// TODO(alloy/#1274): Remove the old peer discovery code once this becomes GA.
+	// TODO(alloy/#1274): Remove the old peer discovery code once this becomes default.
 	if opts.EnableDiscoveryV2 {
 		config.DiscoverPeers, err = discovery.NewPeerDiscoveryFn(discovery.Options{
 			JoinPeers:     opts.JoinPeers,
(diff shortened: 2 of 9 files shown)


internal/alloycli/cluster_builder.go
```diff
@@ -68,24 +70,39 @@ func buildClusterService(opts clusterOptions) (*cluster.Service, error) {
...
 
-	default:
-		// Here, both JoinPeers and DiscoverPeers are empty. This is desirable when
-		// starting a seed node that other nodes connect to, so we don't require
-		// one of the fields to be set.
+		default:
+			// Here, both JoinPeers and DiscoverPeers are empty. This is desirable when
+			// starting a seed node that other nodes connect to, so we don't require
+			// one of the fields to be set.
+		}
```
thampiotr, line 105: The diff doesn't make it easy to see that this `else` is basically the original cluster peer discovery code.

internal/alloycli/cmd_run.go
```diff
@@ -252,6 +252,8 @@ func (fr *alloyRun) Run(configPath string) error {
 		ClusterName:         fr.clusterName,
 		//TODO(alloy/#1274): graduate to GA once we have more confidence in this feature
 		EnableStateUpdatesLimiter: fr.minStability.Permits(featuregate.StabilityPublicPreview),
+		//TODO(alloy/#1274): graduate to GA once we have more confidence in this feature
+		EnableDiscoveryV2: fr.minStability.Permits(featuregate.StabilityPublicPreview),
```
thampiotr, line 256: I wonder if we could use something else to control it, like some kind of a temporary feature flag. So that users can still have public preview, but with the old clustering peer discovery if they wanted to.

internal/service/cluster/discovery/peer_discovery_test.go
```diff
@@ -0,0 +1,355 @@
...
+	"testing"
+
+	godiscover "github.com/hashicorp/go-discover"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/trace/noop"
+
+	"github.com/go-kit/log"
+)
+
+func TestPeerDiscovery(t *testing.T) {
```
thampiotr, line 17: These tests help to highlight some of the issues that I want to address on top of using A records in the follow-up PR.

internal/alloycli/cmd_run.go
thampiotr: I'm leaning towards changing this to an opt-out flag `--cluster.use-peer-discovery-v1` so that we can disentangle the public preview from clustering peer discovery mechanism.

thampiotr: As discussed with the maintainers, I've switched to use a `cluster.use-discovery-v1` flag cc @wildum

docs/sources/reference/cli/run.md
```diff
@@ -137,6 +138,9 @@ When `--cluster.name` is provided, nodes will only join peers who share the same
 By default, the cluster name is empty, and any node that doesn't set the flag can join.
 Attempting to join a cluster with a wrong `--cluster.name` will result in a "failed to join memberlist" error.
 
+The `--cluster.use-discovery-v1` flag can be used to switch back to the older, v1 version of the cluster peer discovery mechanism
+in case of any issues with the newer version. This flag will be deprecated in the future and eventually removed.
```
thampiotr, line 142: As I add support for A records, I will add here a note about what differences are between these mechanisms in the follow-up PR.

internal/alloycli/cmd_run.go
```diff
@@ -248,10 +251,12 @@ func (fr *alloyRun) Run(configPath string) error {
 		DiscoverPeers:       fr.clusterDiscoverPeers,
 		RejoinInterval:      fr.clusterRejoinInterval,
 		AdvertiseInterfaces: fr.clusterAdvInterfaces,
-		ClusterMaxJoinPeers: fr.ClusterMaxJoinPeers,
+		ClusterMaxJoinPeers: fr.clusterMaxJoinPeers,
 		ClusterName:         fr.clusterName,
 		//TODO(alloy/#1274): graduate to GA once we have more confidence in this feature
 		EnableStateUpdatesLimiter: fr.minStability.Permits(featuregate.StabilityPublicPreview),
+		//TODO(alloy/#1274): graduate to GA once we have more confidence in this feature
```
wildum, line 258: this comment is deprecated because it uses a feature flag now

wildum (approved):
LGTM, the new structure and the tests are great. The code duplication is not ideal but it's only temporary

PR #1326, opened 2024-07-18
mimir.rules.kubernetes: Add additional labels to rules

### LINKED ISSUE #291: mimir.rules/`PrometheusRule` should support injecting selector labels
### Request

The ability to inject a label selector into rules created/managed with the `mimir.rules` component.
e.g. if I have the mimir.rules config of (the newly requested field is something like `inject_labels`):
```hcl
mimir.rules.kubernetes "default" {
  address = env("MIMIR_RULER_URL")
  basic_auth {
    username = env("MIMIR_USERNAME")
    password = env("MIMIR_PASSWORD")
  }
  mimir_namespace_prefix = env("MIMIR_NAMESPACE_PREFIX")

  inject_labels = { cluster = "mycluster" }
}
```

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  name: prometheus-example-alerts
spec:
  groups:
  - name: ./example-alert.rules
    rules:
    - alert: ExampleAlert
      expr: my_metric{foo="bar"}
```
would result in a rule loaded of 
```promql
my_metric{cluster="mycluster",foo="bar"}
```

### Use case

This is used to complement `external_labels` added with a `prometheus.remote_write` component.

Without it, the created rules can easily run against a differently namespaced set of metric data.

ptodev:
#### PR Description

This can be useful when users are migrating from many independent Prometheus instances to one central Mimir instance.

Credit for this PR goes to @psychomantys, who authored an [Agent PR](https://github.com/grafana/agent/pull/6769) which this PR is based on.

Fixes #291

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
+++ b/docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
@@ -66,6 +66,7 @@ Name                     | Type                | Description                    
 `no_proxy`               | `string`            | Comma-separated list of IP addresses, CIDR notations, and domain names to exclude from proxying. |               | no
 `proxy_from_environment` | `bool`              | Use the proxy URL indicated by environment variables.                                            | `false`       | no
 `proxy_connect_header`   | `map(list(secret))` | Specifies headers to send to proxies during CONNECT requests.                                    |               | no
+`external_labels`        | `map(string)`       | Labels to add to each rule.                                                                      | `{}` | no
 
  At most, one of the following can be provided:
  - [`bearer_token` argument](#arguments).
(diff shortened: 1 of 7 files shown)


docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
```diff
@@ -234,8 +237,8 @@ mimir.rules.kubernetes "local" {
 }
 ```
 
-This example creates a `mimir.rules.kubernetes` component that loads discovered
-rules to Grafana Cloud.
+This example creates a `mimir.rules.kubernetes` component that loads discovered rules to Grafana Cloud. 
+It will also add a `"label1"` label to each rule. If that label already exists, its value will be overridden with `"value1"`.
```
clayton-cornell, line 241:
```suggestion
It also adds a `"label1"` label to each rule. If that label already exists, it is overwritten with `"value1"`.
```

It's still passive... but good enough since we are describing what happens in the example.

clayton-cornell (approved):
Approving for docs

mattdurham (approved):
LGTM

PR #1353, opened 2024-07-24
oracle exporter custom metrics

EHSchmitt4395:
#### PR Description
Oracle Exporter wrapped in alloy deployment has capability to include custom metrics. This PR aims to add that capability to alloy, modeling mssql exporter. 

#### Notes to the Reviewer
The code is compatible with a file path to be read in from alloy config, ie `local.file` , the tests however are modeled after yaml string. 
#### PR Checklist
- [x] CHANGELOG.md updated
- [ ] Documentation added (will add once confirmed sound, or changes needed ie path) 
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/exporter/oracledb/oracledb.go
+++ b/internal/component/prometheus/exporter/oracledb/oracledb.go
@@ -36,7 +36,6 @@ var DefaultArguments = Arguments{
 	MaxOpenConns: 10,
 	QueryTimeout: 5,
 }
-
 var (
 	errNoConnectionString = errors.New("no connection string was provided")
 	errNoHostname         = errors.New("no hostname in connection string")
@@ -48,6 +47,7 @@ type Arguments struct {
 	MaxIdleConns     int               `alloy:"max_idle_conns,attr,optional"`
 	MaxOpenConns     int               `alloy:"max_open_conns,attr,optional"`
 	QueryTimeout     int               `alloy:"query_timeout,attr,optional"`
+	CustomMetrics      alloytypes.OptionalSecret `alloy:"custom_metrics,attr,optional"`
 }
 
 // SetToDefault implements syntax.Defaulter.
@@ -82,5 +82,6 @@ func (a *Arguments) Convert() *oracledb_exporter.Config {
 		MaxIdleConns:     a.MaxIdleConns,
 		MaxOpenConns:     a.MaxOpenConns,
 		QueryTimeout:     a.QueryTimeout,
+		CustomMetrics:    a.CustomMetrics.Value,
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Main (unreleased)
 - Added community components support, enabling community members to implement and maintain components. (@wildum)
 
 ### Enhancements
+- Added custom metrics capability to oracle exporter. (@EHSchmitt4395)
 
 - Added a success rate panel on the Prometheus Components dashboard. (@thampiotr)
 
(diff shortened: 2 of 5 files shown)


internal/component/prometheus/exporter/oracledb/oracledb.go
```diff
@@ -48,6 +47,7 @@ type Arguments struct {
 	MaxIdleConns     int               `alloy:"max_idle_conns,attr,optional"`
 	MaxOpenConns     int               `alloy:"max_open_conns,attr,optional"`
 	QueryTimeout     int               `alloy:"query_timeout,attr,optional"`
+	CustomMetrics      alloytypes.OptionalSecret `alloy:"custom_metrics,attr,optional"`
```
thampiotr, line 50: Do you expect this to contain sensitive data such as passwords, etc? Could it be a `string` type instead?

thampiotr, line 50: Second question is: would you use the inline YAML string in .alloy config file? 
Perhaps it's not very usable and it's better to do what postgres exporter does and make it a path to a local YAML file with custom metrics. [See example](https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.exporter.postgres/#collect-custom-metrics-from-an-allowlisted-set-of-databases).

thampiotr (commented):
Thank you for opening this. I think it's worth it to add this enhancement. Couple of questions regarding the way we expose this.

internal/component/prometheus/exporter/oracledb/oracledb.go
EHSchmitt4395: i think as a file path not needed, will change to string !

EHSchmitt4395: Hey @thampiotr 
Thanks for the review! I think you are right, i made the tests surrounding this to be in-line string with `local.file.content` , but i think this logic would still work as is with a file path. IE: if we just used [local.file](https://grafana.com/docs/agent/latest/flow/reference/components/local.file/) without content. 

i believe mssql component does the same, whats your thought on this ? Can see here guidance on local.file for 
[query_config](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.mssql/#arguments)

So in this case id modify to string type and change tests to just look for basic filepath string  for compatibility like 
```
local.file "test" {
  filename = "oracle-metrics.yml"
}

prometheus.exporter.oracledb "test" {
    connection_string = CONNECTION_STRING
    custom_metrics = local.file.test 
}
```
with local.file allowing a user to store custom metrics as config or even in source of truth code mgmt repo vs needing to store on disk 

would appreciate your thoughts, and ill make the changes required 
thanks again!

thampiotr: The tests look good to me, I don't think they need further changes to use `local.file`.

What I was wondering is whether we should use file path or inline string. I see that your current implementation provides an inline string and I understand that you can use `local.file` to provide it. It does match the `prometheus.exporter.mssql`, but it doesn't match the `prometheus.exporter.postgres` ([example](https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.exporter.postgres/#collect-custom-metrics-from-an-allowlisted-set-of-databases)) where the string is provided via file path.

I think that we have more instances where we have an inline string and this approach is a bit more flexible, although more complex. I think we can keep this as-is (pass string inline) and if we wanted to make it easier for users in the future, we could have additional argument of `custom_metrics_path` to reduce pipeline complexity. Or maybe we should have a standard library function that will just load file contents to make it even simpler.

Either way, I think your implementation is fine as-is now and there's no need to further complicate the tests with `local.file`.

EHSchmitt4395: Makes sense i appreciate this insight -- that said for now if we keep the local.file method and in-line string capability, are you good with leaving this as type `alloytypes.OptionalSecret`  i dont expect users would be embedding secrets into their queries but im curious your take if you still think we should make it to `string` ?

EHSchmitt4395: @thampiotr just a humble bump when time presents itself - appreciate the help here again ! :)

thampiotr: I don't think it's a problem, so overall this LGTM.

thampiotr (approved):

thampiotr:
@EHSchmitt4395 I should have waited for the docs to be updated. Would you like to open a PR for docs?

EHSchmitt4395:
Hey @thampiotr  ofc - i will start putting a pr together and tag you likely for early next week, thnaks for the merge :)

PR #1356, opened 2024-07-25
Upgrade `beyla.ebpf` component to Beyla 1.7

marctc:
#### PR Description

This PR upgrades Beyla to version 1.7. This includes the following:
- Numerous bug fixes and improvement of overall stability
- Fixed Kubernetes discovery mechanism
- Allow to configure Beyla to get process metrics and network metrics, besides application metrics
- Enable debug mode for Beyla

#### Notes to the Reviewer

The PR adds a bunch of replaces for `go.opentelemetry.io/otel` packages which adds an experimental
feature that allow expiration of OTEL metrics. This is needed to remove stale metrics coming from short-lived process or from crashlooping pods when using network monitoring. 

There's a first PR that needs to be merged in OTEL spec. Then another PR will come to add the feature upstream. 
https://github.com/open-telemetry/opentelemetry-specification/pull/4135

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


grcevski (approved):
LGTM!

grcevski (commented):
I think we need to make sure we are able to exclude some services in a config: https://github.com/grafana/beyla/blob/main/pkg/services/criteria.go#L52

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -116,43 +126,76 @@ func (args Services) Convert() (services.DefinitionCriteria, error) {
...
+		p.Instrumentations = args.Instrumentations
+	}
+	return p
+}
+
+func (args Network) Convert() beyla.NetworkConfig {
+	networks := beyla.DefaultConfig.NetworkFlows
+	if args.Enable {
+		networks.Enable = true
+	}
```
wildum, line 197: I'm not a big fan of the defaults being set in the convert functions. I think this is error-prone. For example, here it works because `enable` is by default "false" but if one day it is changed to true in Beyla, then the logic above would need to be updated because there would not be a way for the user to set it to false. The pattern used for all other components is to implement the "[SetToDefault()](https://github.com/grafana/alloy/blob/main/syntax/internal/value/decode.go#L30)" from the Defaulter interface. Is there a particular reason why it is not used for this component?

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,40 @@ Name               | Type           | Description
...
+- `network`, exports network-level metrics.
+
+`instrumentations` is a list of instrumentations to enable for the Prometheus metrics. The following instrumentations are available:
+
+- `*` enables all `instrumentations`. If `*` is present in the list, the other values are simply ignored.
+- `http` enables the collection of HTTP/HTTPS/HTTP2 application metrics.
+- `grpc` enables the collection of gRPC application metrics.
+- `sql` enables the collection of SQL database client call metrics.
+- `redis` enables the collection of Redis client/server database metrics.
+- `kafka` enables the collection of Kafka client/server message queue metrics.
```
wildum, line 193: ideally any field that can only take a specific set of values should be checked in the Validate() function.

internal/component/beyla/ebpf/beyla_linux_test.go
```diff
@@ -178,24 +207,64 @@ func TestConvert_Discovery(t *testing.T) {
...
+	require.Len(t, config.ExcludeServices, 1)
+	require.Equal(t, "test", config.ExcludeServices[0].Name)
+	require.Equal(t, "default", config.ExcludeServices[0].Namespace)
+}
+
+func TestConvert_Prometheus(t *testing.T) {
+	args := Prometheus{
+		Features:         []string{"application", "network"},
+		Instrumentations: []string{"redis", "sql"},
+	}
```
wildum, line 246: we need the sad paths covered as well. What happens if I set a field on features or on instrumentations that does not exist?

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,40 @@ Name               | Type           | Description
 `owner_name`       | `string`       | Regular expression of Kubernetes owners of running Pods to match.                                           | `""`    | no
 `pod_labels`       | `map(string)`  | Key-value pairs of labels with keys matching Kubernetes Pods with the provided value as regular expression. |  `{}`   | no
 
+### prometheus block
+
+This block configures which metrics Beyla collects.
```
wildum, line 171: I don't see a way of disabling this feature, is it ok? What if the user is only interested in traces?

internal/component/beyla/ebpf/beyla_linux.go
wildum: This might be fine for this PR because of the short window that we have with Alloy's release cut but it would be nice to address this is a follow up PR

go.mod
```diff
@@ -803,3 +809,14 @@ exclude (
...
+// Replaces required for Beyla to implement metric expiration
+// Remove once this PR is merged upstream:
+// https://github.com/open-telemetry/opentelemetry-specification/pull/4135
+replace (
+	go.opentelemetry.io/otel => github.com/grafana/opentelemetry-go v1.28.0-grafana.2
+	go.opentelemetry.io/otel/metric => github.com/grafana/opentelemetry-go/metric v1.28.0-grafana.2
+	go.opentelemetry.io/otel/sdk => github.com/grafana/opentelemetry-go/sdk v1.28.0-grafana.2
+	go.opentelemetry.io/otel/sdk/metric => github.com/grafana/opentelemetry-go/sdk/metric v1.28.0-grafana.2
+	go.opentelemetry.io/otel/trace => github.com/grafana/opentelemetry-go/trace v1.28.0-grafana.2
+)
```
wildum, line 822: It would be much better if upstream's PR would be merged before to avoid the replace directives. Updating the Otel dependencies in Alloy is already a long and difficult process (see the current PR to update it: https://github.com/grafana/alloy/pull/1362). We update our Otel dependencies before every release (every 6 weeks). Having the fork means that we will also need to keep the fork up to date every 6 weeks until the PR is merged. What if it's not merged or it takes a different direction?

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -38,12 +38,15 @@ Name              | Type     | Description
...
 `open_port`       | `string` | The port of the running service for Beyla automatically instrumented with eBPF.     | `""`    | no
 `executable_name` | `string` | The name of the executable to match for Beyla automatically instrumented with eBPF. | `""`    | no
+`debug`           | `bool`   | Enable debug mode for Beyla.                                                        | `false` | no
 
 `open_port` accepts a comma-separated list of ports (for example, `80,443`), and port ranges (for example, `8000-8999`).
 If the executable matches only one of the ports in the list, it is considered to match the selection criteria.
 
 `executable_name` accepts a regular expression to be matched against the full executable command line, including the directory where the executable resides on the file system.
 
+`debug` enables debug mode for Beyla. This mode logs BPF logs, network logs, trace representation logs, and other debug information.
```
ptodev, line 48: Why isn't it just part of debug logging, configured via the [logging block](https://grafana.com/docs/alloy/latest/reference/config-blocks/logging/)?

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -54,8 +57,12 @@ routes                  | [routes][]     | Configures the routes to match HTTP p
...
-discovery > services    | [services][]   | Configures the discovery for the component.                                                        | no
-discovery > services > kubernetes    | [kubernetes services][]   | Configures the discovery for the component.                                | no
+discovery > services    | [services][]   | Configures the services to discover for the component.                                                        | no
+discovery > services > kubernetes    | [kubernetes services][]   | Configures the Kubernetes services to discover for the component.                                | no
+discovery > exclude_services    | [services][]   | Configures the services to exclude for the component.                                                        | no
+discovery > exclude_services > kubernetes    | [kubernetes services][]   | Configures the Kubernetes services to exclude for the component.                                | no
+prometheus              | [prometheus][] | Configures which metrics Beyla collects.                                                           | no
```
ptodev, line 64:
```suggestion
prometheus              | [prometheus][] | Configures which Prometheus metrics Beyla exposes.                                                           | no
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -89,6 +97,8 @@ If set to `false`, the Kubernetes metadata decorator will be disabled.
 
 If set to `autodetect`, Beyla will try to automatically detect if it is running inside Kubernetes, and enable the metadata decoration if that's the case.
 
+`cluster_name` if not set, Beyla tries to detect the cluster name from the Kubernetes API.
```
ptodev, line 100:
```suggestion
If `cluster_name` is not set, Beyla tries to detect the cluster name from the Kubernetes API.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,40 @@ Name               | Type           | Description
...
+------------------|----------------|----------------------------------------------------------------|---------|---------
+`features`        | `list(string)` | List of features to enable for the Prometheus metrics.         | `["application"]`    | no
+`instrumentations`| `list(string)` | List of instrumentations to enable for the Prometheus metrics. | `["*"]`    | no
+
+`features` is a list of features to enable for the Prometheus metrics. The following features are available:
+
+- `application` exports application-level metrics.
+- `application_span`exports application-level metrics in traces span metrics format.
+- `application_service_graph` exports application-level service graph metrics.
+-  `application_process` exports metrics about the processes that run the instrumented application.
```
ptodev, line 183:
```suggestion
- `application_process` exports metrics about the processes that run the instrumented application.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,40 @@ Name               | Type           | Description
...
+- `network`, exports network-level metrics.
```
ptodev, line 184:
```suggestion
- `network` exports network-level metrics.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
ptodev, line 171: Does the `prometheus` block also have an impact on the OTLP metrics?

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: I'd say this is not possible because that would introduce a cyclic dependency. Beyla would depend on Alloy's code, and Alloy at the same time would depend on Beyla code.

Also the logger is ubiquitous in Beyla and should work when Beyla runs in standalone mode. I think that replacing the logging system in Beyla could be out of the scope of this integration.

However I'd suggest some mid term changes:
- Move the Logger implementation as an independent go module.
- Make it to implement the `slog` interface so it's easy to switch from an implementation to another in Beyla.

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: Actually yes, as Alloy uses the internal Beyla prometheus for collecting metrics, whichever format they are exported. Should we rename this block?

internal/component/beyla/ebpf/beyla_linux_test.go
mariomac: Adding validation code and tests for both

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: Forget that last message. I saw that you already provided a `deferredSlogHandler` struct. Trying to see if I can integrate it easily with Beyla

docs/sources/reference/components/beyla/beyla.ebpf.md
wildum: looking at the code, I don't think that this is the case. The Convert() method will apply the default values which ["application"] and ["*"].

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: This PR would give us Beyla integration with Alloy loggers for free: https://github.com/grafana/alloy/pull/1373

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Main (unreleased)
 ### Breaking changes to non-GA functionality
 
 - Update Public preview `remotecfg` argument from `metadata` to `attributes`. (@erikbaranowski)
+- The `beyla.ebpf.routes.unmatched` configuration option default changed from unset to `heuristic` (@marctc)
```
wildum, line 16:
```suggestion
- The default value of the argument `unmatched` in the block `routes` of the component `beyla.ebpf` was changed from `unset` to `heuristic` (@marctc) 
```

wildum, line 16: I think that it would also be nice to have in Enhancements an entry that says that Beyla was updated to version 1.7 and list some of the key changes

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -285,13 +334,36 @@ func (a *Arguments) Convert() (*beyla.Config, error) {
...
 }
 
 func (args *Arguments) Validate() error {
 	if args.Port == "" && args.ExecutableName == "" && len(args.Discovery.Services) == 0 {
 		return fmt.Errorf("you need to define at least open_port, executable_name, or services in the discovery section")
 	}
+	validInstrumentations := map[string]struct{}{"*": {}, "http": {}, "grpc": {}, "redis": {}, "kafka": {}, "sql": {}}
+	for _, instrumentation := range args.Prometheus.Instrumentations {
+		if _, ok := validInstrumentations[instrumentation]; !ok {
+			return fmt.Errorf("invalid prometheus.instrumentations entry: %s", instrumentation)
```
wildum, line 352: what happens if I specify multiple times the same instrumentation? For example ['"http", "http"]

wildum: (same question for the features below)

mariomac: They will be just ignored. Internally is treated as a set.

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: I'd anyway keep this configuration option, maybe renaming it to `trace_logs`, as this option goes beyond just debug logs. It would generate a huge amount of output, like BPF logs, printing traces, etc...

mariomac: It would log:
* the Beyla debug logs
* one line per each application request or network flow
* several lines per each eBPF operation. Tenths for each service request. This is the most critical part, as it is only used for developers' debugging purposes

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,40 @@ Name               | Type           | Description
...
+
+- `application` exports application-level metrics.
+- `application_span`exports application-level metrics in traces span metrics format.
+- `application_service_graph` exports application-level service graph metrics.
+- `application_process` exports metrics about the processes that run the instrumented application.
+- `network` exports network-level metrics.
+
+`instrumentations` is a list of instrumentations to enable for the Prometheus metrics. The following instrumentations are available:
+
+- `*` enables all `instrumentations`. If `*` is present in the list, the other values are simply ignored.
```
clayton-cornell, line 188:
```suggestion
- `*` enables all `instrumentations`. If `*` is present in the list, the other values are ignored.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
mariomac: The idea is:
* If debug==true you get Beyla debug logs
* If debug==true and trace_logs == true you get Beyla logs and the low-level BPF debug information
* If debug== false you get Beyla info logs only, whatever value has trace_logs.

mariomac: I just realized that the `Logger` variable that Beyla receives in the `component.Options` ( `Logger` interface with only the `Log` method) is not the `Logger` struct that we need (the one that provides us the `slog.Handler`) we need to pass to Beyla. That's probably why Marc didn't integrate Beyla logging with the Alloy logging and created this issue: https://github.com/grafana/beyla/issues/1057

I think that, in that case we should keep the previous behavior by now (enabling logs with "debug").

wildum: ok that's fine for me for this iteration. It will be a breaking change when we remove it later but because it's a debug feature that's not so bad

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +165,39 @@ Name               | Type           | Description
...
+`instrumentations` is a list of instrumentations to enable for the metrics. The following instrumentations are available:
+
+- `*` enables all `instrumentations`. If `*` is present in the list, the other values are ignored.
+- `http` enables the collection of HTTP/HTTPS/HTTP2 application metrics.
+- `grpc` enables the collection of gRPC application metrics.
+- `sql` enables the collection of SQL database client call metrics.
+- `redis` enables the collection of Redis client/server database metrics.
+- `kafka` enables the collection of Kafka client/server message queue metrics.
+
+`network` configures network metrics options for Beyla. The following options are accepted:
```
wildum, line 195: is `network` intended to be a block or an attribute?
- if in the future there will be more options to configure the network metrics, then it makes sense that this is a block but it needs to be documented as such (it should be in the block table `metrics > network` and have its own part just like the `attributes > kubernetes` for example)
- if it's only something that can be toggled and it is intended that it stays that way then it should just be a boolean attribute and should be renamed something like `network_enabled` or network_metrics_enabled`

mariomac: It's a block. In the future there should be more options to configure the network metrics: https://grafana.com/docs/beyla/latest/network/config/

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Main (unreleased)
 ### Breaking changes to non-GA functionality
 
 - Update Public preview `remotecfg` argument from `metadata` to `attributes`. (@erikbaranowski)
+- The default value of the argument `unmatched` in the block `routes` of the component `beyla.ebpf` was changed from `unset` to `heuristic` (@marctc)
```
wildum, line 26:
```suggestion

- The default value of the argument `unmatched` in the block `routes` of the component `beyla.ebpf` was changed from `unset` to `heuristic` (@marctc) 
```

docs/sources/reference/components/beyla/beyla.ebpf.md
wildum: ok then once you adapt the doc for it we will be good to go from my POV. Also make sure you remove the `network` entry from the metrics table above (this is for attributes).

mariomac: ok! done

go.mod
```diff
@@ -706,11 +705,18 @@ require (
...
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require (
+	go.opentelemetry.io/contrib/detectors/aws/ec2 v1.28.0 // indirect
+	go.opentelemetry.io/contrib/detectors/aws/eks v1.28.0 // indirect
+	go.opentelemetry.io/contrib/detectors/azure/azurevm v0.0.1 // indirect
+	go.opentelemetry.io/contrib/detectors/gcp v1.28.0 // indirect
+)
+
```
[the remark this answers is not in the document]
mariomac: Fixed! By some reason, `go mod` tool decided to put them apart.

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -156,6 +166,42 @@ Name               | Type           | Description
...
+`instrumentations` is a list of instrumentations to enable for the metrics. The following instrumentations are available:
+
+- `*` enables all `instrumentations`. If `*` is present in the list, the other values are ignored.
+- `http` enables the collection of HTTP/HTTPS/HTTP2 application metrics.
+- `grpc` enables the collection of gRPC application metrics.
+- `sql` enables the collection of SQL database client call metrics.
+- `redis` enables the collection of Redis client/server database metrics.
+- `kafka` enables the collection of Kafka client/server message queue metrics.
+
+The metrics block contains the following `network` block:
```
wildum, line 195:
```suggestion
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -38,25 +38,32 @@ Name              | Type     | Description
...
+discovery               | [discovery][]             | Configures the discovery for instrumentable processes matching a given criteria.                   | no
+discovery > services    | [services][]              | Configures the services to discover for the component.                                                        | no
+discovery > services > kubernetes    | [kubernetes services][]   | Configures the Kubernetes services to discover for the component.                                | no
+discovery > exclude_services    | [services][]              | Configures the services to exclude for the component.                                                        | no
+discovery > exclude_services > kubernetes    | [kubernetes services][]   | Configures the Kubernetes services to exclude for the component.                                | no
+metrics                 | [metrics][]               | Configures which metrics Beyla exposes.                                                           | no
+metrics > network       | [network attributes][]    | Configures network metrics options for Beyla.                                                      |no
```
wildum, line 65:
```suggestion
metrics > network       | [network][]    | Configures network metrics options for Beyla.                                                      |no 
```

docs/sources/reference/components/beyla/beyla.ebpf.md
wildum: + the empty line should also be removed

wildum (approved):
LGTM, thanks for taking over this and making the changes! Could you please create follow up issues to track the non-blocking points that were not addressed in the PR?

mariomac:
Thanks @wildum !

Here are the created issues:
https://github.com/grafana/alloy/issues/1385
https://github.com/grafana/alloy/issues/1386

Feel free to ping me if you miss any open topic that should be added as an issue.

wildum:
> Feel free to ping me if you miss any open topic that should be added as an issue.

Maybe one more issue to consider refactoring the way the defaults are handled. (related to this comment https://github.com/grafana/alloy/pull/1356/files#r1694866266)

PR #1357, opened 2024-07-25
Fix link to Probe CRD

webwurst:


clayton-cornell (approved):
Oh, good catch. Thanks for fixing this

PR #1358, opened 2024-07-25
Fix custom components not always getting updated when config in imported directory changes

### LINKED ISSUE #1348: Alloy doesn't update all components when changing config in a imported directory.
### What's wrong?

In the example changes to components/input_otlp.alloy are not picked up by Alloy. 

It happens because when the config is first applied `loader.loadNewGraph` reduces the graph and the input_otlp -> import.file edge is removed. When input_otlp.alloy is changed only nodes pointing to import.file get updated through `loader.EvaluateDependants`. Output_otlp still points to import.file in the graph so it gets evaluated but because its exports haven't changed the evaluation is not passed down to input_otlp.

### Steps to reproduce

Step 1: Start alloy with the following config files.
Step 2: Change something like the http port in `components/input_otlp.alloy` and check that the change is not applied.

`config.alloy`
```
import.file "comp" {
	filename = "components"
}

comp.input_otlp "input_otlp" {
	metrics_targets = [comp.output_otlp.output_otlp.metrics_receiver]
}

comp.output_otlp "output_otlp" { }
```
`components/input_otlp.alloy`
```
declare "input_otlp" {
	argument "metrics_targets" {
		optional = false
	}

	otelcol.receiver.otlp "input" {
		http {  }

		output {
			metrics = argument.metrics_targets.value
		}
	}
}
```
`components/output_otlp.alloy`
```
declare "output_otlp" {
	export "metrics_receiver" {
		value = otelcol.exporter.otlphttp.default.input
	}

	otelcol.exporter.otlphttp "default" {
		client {
			endpoint = "localhost:4317"
		}
	}
}
```

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

ante012:
#### PR Description
Fixes custom components not getting updated when config changes in the imported directory they were declared in. 

It happens when you connect imported components because the transitive reduction of the dependency graph only keeps the edge from one of the components to the importConfigNode.

It is fixed here by modifying the reduce operation of the dependency graph to always keep edges from custom components to importConfigNodes.

Fixes #1348

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Main (unreleased)
 
 ### Bugfixes
 
+- Fix a bug where custom components don't always get updated when the config is modified in an imported directory. (@ante012)
+
 - Fixed an issue which caused loss of context data in Faro exception. (@codecapitano)
 
 - Fixed an issue where providing multiple hostnames or IP addresses
@@ -162,7 +164,6 @@ v1.3.0
   Previously, the reload would fail for `loki.process` without an error in the logs and the metrics
   from the `metrics` stage would get stuck at the same values. (@ptodev)
 
-- Fix a bug where custom components don't always get updated when the config is modified in an imported directory. (@ante012)
 
 v1.2.1
 -----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,8 @@ v1.3.0
   Previously, the reload would fail for `loki.process` without an error in the logs and the metrics
   from the `metrics` stage would get stuck at the same values. (@ptodev)
 
+- Fix a bug where custom components don't always get updated when the config is modified in an imported directory. (@ante012)
+
 v1.2.1
 -----------------
 
--- a/internal/runtime/internal/dag/ops.go
+++ b/internal/runtime/internal/dag/ops.go
@@ -11,7 +11,9 @@ import (
 // as many edges as possible while maintaining the same "reachability" as the
 // original graph: any node N reachable from node S will still be reachable
 // after a reduction.
-func Reduce(g *Graph) {
+// edgesToPreserve allows the caller to specify a set of edges that
+// must not be removed during the transitive reduction.
+func Reduce(g *Graph, edgesToPreserve map[Edge]struct{}) {
 	// A direct edge between two vertices can be removed if that same target
 	// vertex is indirectly reachable through another edge.
 	//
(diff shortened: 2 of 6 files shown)


internal/runtime/internal/dag/ops.go
```diff
@@ -11,7 +11,9 @@ import (
 // as many edges as possible while maintaining the same "reachability" as the
 // original graph: any node N reachable from node S will still be reachable
 // after a reduction.
-func Reduce(g *Graph) {
+// edgesToPreserve allows the caller to specify a set of edges that
+// must not be removed during the transitive reduction.
+func Reduce(g *Graph, edgesToPreserve map[Edge]struct{}) {
```
wildum, line 16: nit: the reduce function is currently not covered by tests, if you have time it would be nice to add one test where the edge is removed and another test with the same scenario but the edge is added to the edgesToPreserve arg

wildum (approved):
Thanks for reporting and fixing this bug, this is super helpful. The solution looks good to me :) @rfratto do you see a problem with preventing the reduction for these edges?

internal/runtime/internal/dag/ops.go
wildum: no need for the tests if we decide to get rid of the reduction

ante012:
> @rfratto suggested that we could maybe get completely rid of the reduction. I will experiment on our dev infra next week to see if I see any impact.

Interesting! While working on this I never quite understood when the reduction makes sense. I tested it on our real setup and no reductions happen except the ones removed by this PR.

thampiotr:
@ante012 I think the graph reduction was beneficial before we added more caching and parallel graph evaluation. Now it's no longer beneficial and led to this issue you've discovered. I think we should get rid of it.

Would you be interested in getting rid of it in this PR or should we take this on and open a new one?

ante012:
@thampiotr Excuse my slow reply. That make sense.

I added a new commit that completely removes the reduce step.

ante012:
I don't think the failing tests are related to the changes here at least I can't get them to fail locally.

wildum:
yeah the integration tests were broken because of [this](https://github.com/actions/runner-images/issues/9692). If you rebase your branch with main it should be fine

ante012:
Hi! Is there anything else that I should do or can help with before this can be merged?

CHANGELOG.md
```diff
@@ -138,6 +138,8 @@ v1.3.0
   Previously, the reload would fail for `loki.process` without an error in the logs and the metrics
   from the `metrics` stage would get stuck at the same values. (@ptodev)
 
+- Fix a bug where custom components don't always get updated when the config is modified in an imported directory. (@ante012)
```
thampiotr, line 141: Could you move this to the top, under the unreleased version?

thampiotr (approved):
Just one small comment and we're ready to go! Thank you for the contribution.

ante012:
> Just one small comment and we're ready to go! Thank you for the contribution.

Great! Thanks for the help!

ante012:
Hi, can we merge this now?

wildum (approved):
I'm back from PTO, LGTM, it's great not to have this "original graph" anymore. Thanks for making the changes.

PR #1359, opened 2024-07-26
remotecfg: wire calls to RegisterCollector

tpaschalis:
#### PR Description
This PR updates the API definition for remotecfg to its latest version, and wires in calls to the RegisterCollector/UnregisterCollector rpcs on startup and shutdown.

No issue filed

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/config-blocks/remotecfg.md
+++ b/docs/sources/reference/config-blocks/remotecfg.md
@@ -46,7 +46,8 @@ Name             | Type                 | Description                           
 
 If the `url` is not set, then the service block is a no-op.
 
-If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a randomly generated, anonymous unique ID (UUID) that is stored as an `alloy_seed.json` file in {{< param "PRODUCT_NAME" >}}'s storage path so that it can persist across restarts. In that case, the `name` field can be used to set another human-friendly identifier for the specific Alloy instance.
+If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a randomly generated, anonymous unique ID (UUID) that is stored as an `alloy_seed.json` file in the {{< param "PRODUCT_NAME" >}} storage path so that it can persist across restarts.
+You can use the `name` field to set another human-friendly identifier for the specific {{< param "PRODUCT_NAME" >}} instance.
 
 The `id` and `attributes` fields are used in the periodic request sent to the
 remote endpoint so that the API can decide what configuration to serve.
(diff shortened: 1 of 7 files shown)


docs/sources/reference/config-blocks/remotecfg.md
```diff
@@ -42,10 +42,11 @@ Name             | Type                 | Description
 `id`             | `string`             | A self-reported ID.                               | `see below` | no
 `attributes`     | `map(string)`        | A set of self-reported attributes.                | `{}`        | no
 `poll_frequency` | `duration`           | How often to poll the API for new configuration.  | `"1m"`      | no
+`name`           | `string`             | A human-readable name for the collector.          | `""`        | no
 
 If the `url` is not set, then the service block is a no-op.
 
-If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a randomly generated, anonymous unique ID (UUID) that is stored as an `alloy_seed.json` file in {{< param "PRODUCT_NAME" >}}'s storage path so that it can persist across restarts.
+If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a randomly generated, anonymous unique ID (UUID) that is stored as an `alloy_seed.json` file in {{< param "PRODUCT_NAME" >}}'s storage path so that it can persist across restarts. In that case, the `name` field can be used to set another human-friendly identifier for the specific Alloy instance.
```
clayton-cornell, line 49:
```suggestion
If not set, the self-reported `id` that {{< param "PRODUCT_NAME" >}} uses is a randomly generated, anonymous unique ID (UUID) that is stored as an `alloy_seed.json` file in the {{< param "PRODUCT_NAME" >}} storage path so that it can persist across restarts.
You can use the `name` field to set another human-friendly identifier for the specific {{< param "PRODUCT_NAME" >}} instance.
```

internal/service/remotecfg/remotecfg.go
```diff
@@ -226,6 +227,9 @@ var _ service.Service = (*Service)(nil)
 func (s *Service) Run(ctx context.Context, host service.Host) error {
 	s.ctrl = host.NewController(ServiceName)
 
+	s.registerCollector()
+	defer s.unregisterCollector()
```
spartan0x117, line 231: I should probably know this by now, but does `Update(...)` end up with a call to `Run(...)`?

I'm wondering if this means that only the attributes that are present when the `remotecfg` service is started get sent in the `RegisterCollector` call and no further updates (say, for example, the initial config were updated)?

I'm not set on that being the desired behavior, but wondering if it's intended :smile:

spartan0x117 (approved):
Just a question about whether hot-reloads of the configs after updates will trigger a call to `RegisterCollector` and whether that's even desired behavior.

LGTM otherwise!

PR #1368, opened 2024-07-26
feat(discovery): Add a remote_relabel component

simonswine:
This allows to receive discovery relabelling rules from a remote server using a websocket connection. 

We plan this for an experimental way to control profile collection with Grafana Cloud Profiles very dynamically. Currently the only backend available is the settings server for Pyroscope (https://github.com/grafana/pyroscope/pull/3416)


internal/component/discovery/remote_relabel/remote_relabel.go
```diff
@@ -0,0 +1,818 @@
...
+	close(w.cleanStopCh)
+	w.wg.Wait()
+
+	return w.c.Close()
+}
+
+var _ component.Component = (*Component)(nil)
+
+func defaultInstance() string {
+	// TODO: This should come from Alloy
```
korniltsev, line 406: should this be addressed before merging PR?

korniltsev (approved):
The code looks good to me, nice and simple.

Remember  to update changelog, docs and API depdendency once it is tagged

docs/sources/reference/components/discovery/discovery.remote_relabel.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name      | Type                | Description        | Default | Required |
+| --------- | ------------------- | ------------------ | ------- | -------- |
+| `targets` | `list(map(string))` | Targets to relabel |         | yes      |
```
clayton-cornell, line 35:
```suggestion
| `targets` | `list(map(string))` | Targets to relabel. |         | yes      |
```

docs/sources/reference/components/discovery/discovery.remote_relabel.md
```diff
@@ -0,0 +1,126 @@
...
+
+### websocket block
+
+The `websocket` block describes a single command and control server. Only one `websocket` block can ben provided.
+
+The following arguments are supported:
+
+| Name                  | Type       | Description                                                        | Default   | Required |
+| --------------------- | ---------- | ------------------------------------------------------------------ | --------- | -------- |
+| `url`                 | `string`   | Full URL to the websocket server. Needs to start with ws:// wss:// |           | yes      |
```
clayton-cornell, line 57:
```suggestion
| `url`                 | `string`   | Full URL to the websocket server. Must start with `ws://` or `wss://`. |           | yes      |
```

clayton-cornell (approved):
A couple of minor non-blocking doc suggestions

wildum:
Hey, thanks for opening a PR, that's an interesting idea :) Adding such a component is a big deal, it would have been better to discuss it over a [proposal](https://github.com/grafana/alloy/blob/main/docs/design/README.md) (it's not too late to create one now and use this PR as a POC).
I would like to know a bit more about the problem it is trying to solve and whether [modules](https://grafana.com/docs/alloy/latest/get-started/modules/) or [remotecfg](https://grafana.com/docs/alloy/latest/reference/config-blocks/remotecfg/) could be a solution.

wildum (changes requested):
^ blocking the PR for now till above's comment is clarified

simonswine:
We will implement this slightly different

PR #1372, opened 2024-07-29
Wire component lookups to remotecfg's isolated controller

tpaschalis:
#### PR Description

This PR exposes the Host of the isolated controller that is run by the remotecfg service, to allow for component lookups to be made against them.

This enables, for example, the in-memory HTTP listener to work properly for remotecfg-supplied configurations.

@rfratto pointed me to a design doc about a [controller refactor](https://docs.google.com/document/d/1jo0WBg89KMqXRixsr7uAZ6cMVgVg50AEhmoq2S-Y1ss/edit#heading=h.4d226co8feom) which potentially could've given us a different approach, but I believe this is the most straightforward way to achieve this right now.

No issue filed.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -182,7 +182,7 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		// This will never happen as the service dependency is explicit.
 		return fmt.Errorf("failed to get the remotecfg service when setting up the http service")
 	}
-	remotecfgHost := svc.(*remotecfg.Service).Data().(remotecfg.Data).Host
+	remotecfgHost := svc.Data().(remotecfg.Data).Host
 
 	// NOTE(@tpaschalis) These need to be kept in order for the longer
 	// remotecfg prefix to be invoked correctly.
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -132,7 +132,7 @@ func (s *Service) Definition() service.Definition {
 	return service.Definition{
 		Name:       ServiceName,
 		ConfigType: Arguments{},
-		DependsOn:  []string{remotecfg.ServiceName}, // http requires remotecfg to be up to wire lookups to its controller.
+		DependsOn:  nil, //[]string{remotecfg.ServiceName}, // http requires remotecfg to be up to wire lookups to its controller.
 		Stability:  featuregate.StabilityGenerallyAvailable,
 	}
 }
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -21,6 +21,7 @@ import (
 	alloy_runtime "github.com/grafana/alloy/internal/runtime"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
 	"github.com/grafana/alloy/internal/service"
+	"github.com/grafana/alloy/internal/service/remotecfg"
 	"github.com/grafana/alloy/internal/static/server"
 	"github.com/grafana/ckit/memconn"
 	_ "github.com/grafana/pyroscope-go/godeltaprof/http/pprof" // Register godeltaprof handler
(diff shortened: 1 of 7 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -224,12 +238,17 @@ var _ service.Service = (*Service)(nil)
 // Run implements [service.Service] and starts the remotecfg service. It will
 // run until the provided context is canceled or there is a fatal error.
 func (s *Service) Run(ctx context.Context, host service.Host) error {
+	s.ctrlMut.Lock()
 	s.ctrl = host.NewController(ServiceName)
+	s.ctrlMut.Unlock()
```
tpaschalis, line 243: Note: the synchronization is required to avoid HTTP's call to Run() be fast enough that s.ctrl hasn't yet had a chance to be populated.

Without it, we get panics like the following every now and then.

```
panic: interface conversion: interface {} is nil, not runtime.ServiceController

goroutine 46 [running]:
github.com/grafana/alloy/internal/service/http.(*Service).Run(0xc002746fc0, {0x9135f30, 0xc002d902d0}, {0x9144d80, 0xc002d09000})
        /home/tpaschalis/GitRepos/alloy/internal/service/http/http.go:185 +0xfa5
github.com/grafana/alloy/internal/runtime/internal/controller.(*ServiceNode).Run(0x0?, {0x9135f30?, 0xc002d902d0?})
        /home/tpaschalis/GitRepos/alloy/internal/runtime/internal/controller/node_service.go:124 +0x30
github.com/grafana/alloy/internal/runtime/internal/controller.newTask.func1()
        /home/tpaschalis/GitRepos/alloy/internal/runtime/internal/controller/scheduler.go:150 +0x31
created by github.com/grafana/alloy/internal/runtime/internal/controller.newTask in goroutine 190
        /home/tpaschalis/GitRepos/alloy/internal/runtime/internal/controller/scheduler.go:149 +0x138
```

tpaschalis:
There are some test failures that indicate a deeper potential issue.

The controller code in the following location only attempts to clean up the loader if it's for a module controller, otherwise it never tries to stop items from the worker pool.
https://github.com/grafana/alloy/blob/c9663c58cc6ec58d97f33ecf783465678e553ce6/internal/runtime/alloy.go#L244

Given the isolated controller from remotecfg is defined as a module controller, this is never hit, leading to those pool goroutines getting left behind. I'll see if we should define it as a non-module controller or sprinkle some extra logic on the controller code to make sure we properly clean up all resources.

internal/service/remotecfg/remotecfg.go
tpaschalis: Note: we're now solving this using a different approach by lazily getting the controller's host via a callback.

thampiotr (approved):
LGTM!

internal/service/http/http.go
```diff
@@ -120,7 +122,8 @@ func New(opts Options) *Service {
 		tcpLis:    tcpLis,
 		memLis:    memconn.NewListener(l),
 
-		componentHttpPathPrefix: "/api/v0/component/",
+		componentHttpPathPrefix:          "/api/v0/component/",
+		componentHttpPathPrefixRemotecfg: "/api/v0/component/remotecfg",
```
mattdurham, line 126: Where is `remotecfg` endpoint being called from?

tpaschalis: Presently, this is used so that lookups via the in-memory listener work properly. There isn't any other caller for now.

Foe example, currently, if you try and scrape `prometheus.exporter.self` for example on a remotecfg-provided config, it would fail with an 'invalid URL' message.

mattdurham (approved):
LGTM, long term would like to tidy things up but since its all under v0 that means we can change without breaking backwards compat

PR #1387, opened 2024-07-31
Cluster peer discovery improvements and fixes

thampiotr:
#### PR Description

Continuation of https://github.com/grafana/alloy/pull/1394.

- Clustering peer resolution through `--cluster.join-addresses` flag has been
  improved with more consistent behaviour, better error handling and added
  support for A/AAAA DNS records. If necessary, users can temporarily opt out of
  this new behaviour with the `--cluster.use-discovery-v1`, but this can only be
  used as a temporary measure, since this flag will be disabled in future
  releases.

- Fixed an issue where providing multiple hostnames or IP addresses
  via `--cluster.join-addresses` would only use the first provided value.

- Fixed an issue where providing `<hostname>:<port>`
  in `--cluster.join-addresses` would only resolve with DNS to a single address,
  instead of using all the available records. 

- Fixed an issue where clustering peers resolution via hostname in `--cluster.join-addresses`
  resolves to duplicated IP addresses when using SRV records.

- Added more tests to cover the above

- Refactored for readability (separate commit from test changes, verified tests pass)

- Improved error handling, logging and added tracing.

#1208 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


mattdurham:
Will put some eyes on this tomorrow.

internal/service/cluster/discovery/join_peers.go
```diff
@@ -0,0 +1,178 @@
...
+			host = addr
+			port = strconv.Itoa(opts.DefaultPort)
+		}
+
+		atLeastOneSuccess := false
+		for _, resolver := range resolvers {
+			resolved, err := resolver(host)
+			deferredErr = errors.Join(deferredErr, err)
+			for _, foundAddr := range resolved {
+				result = append(result, appendPortIfAbsent(foundAddr, port))
```
mattdurham, line 80: Do we need this? Is this always handled in the above code starting on line 69.

internal/service/cluster/discovery/join_peers.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			level.Debug(opts.Logger).Log("msg", "failed to resolve DNS records", "addr", addr, "record_type", recordType, "err", err)
+			span.SetStatus(codes.Error, err.Error())
+			return nil, fmt.Errorf("failed to resolve %q records: %w", recordType, err)
+		}
+
+		// Use all the addresses found
+		level.Debug(opts.Logger).Log("msg", "received DNS query response", "addr", addr, "record_type", recordType, "records_count", len(addresses))
+		var result []string
+		for _, a := range addresses {
```
mattdurham, line 170: Can we return the addresses without adding to result? kthey are both []string.

mattdurham (approved):
Overall a few questions but overall good.

internal/service/cluster/discovery/join_peers.go
thampiotr: You're right, can be simplified to `net.JoinHostPort(foundAddr, port)`.

internal/service/cluster/discovery/join_peers.go
thampiotr: Yes! leftover from refactoring that I've missed, good catch!

PR #1391, opened 2024-07-31
Update loki.source.docker.md

Alwaysin:
+ Labels are mandatory (else faced with: `level=error msg="final error sending batch" component_path=/ component_id=loki.write.local component=client host=127.0.0.1:3100 status=400 tenant="" error="server returned HTTP status 400 Bad Request (400): error at least one label pair is required per stream"`)
+ default Loki URL was not correct

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


clayton-cornell (approved):
Looks good to me. Thanks for catching this!

PR #1398, opened 2024-07-31
Clear `discovery.*` cache when config is reloaded

ptodev:
#### PR Description

This PR fixes a bug whereby if `discovery.` components such as `discovery.kubernetes` are reconfigured differently, they may still end up exporting targets that it's not meant to export. For example, if an Alloy is configured like this...

```
discovery.kubernetes "pods" {
  role = "pod"
  kubeconfig_file = env("HOME") + "/.kube/config"
  namespaces {
    names = ["default", "alloy-logs"]
  }
}
```

...and then reconfigured like this...
```
discovery.kubernetes "pods" {
  role = "pod"
  kubeconfig_file = env("HOME") + "/.kube/config"
  namespaces {
    names = ["default"]
  }
}
```

... after the config is reloaded with `localhost:12345/-/reload`, Alloy would still end up exporting targets with a `__meta_kubernetes_namespace` label set to `alloy-logs` **but only if there are no pods discovered in the `default` namespace**. If pods were discovered in the `default` namespace, this bug wouldn't trigger and no pods from `alloy-logs` will flow downstream, as expected.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/discovery/discovery.go
+++ b/internal/component/discovery/discovery.go
@@ -225,7 +225,11 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				send()
 				haveUpdates = false
 			}
-		case <-ctx.Done():
+		// If Run has exited, return.
+		// We can't exit straight after the context is cancelled.
+		// We need to receive from runExited first.
+		// If we don't the first "go func() " will block on sending to runExited.
+		case <-runExited:
 			// Clear the cache
 			//TODO: Write a unit test which changes the config, reloads the component,
 			// and checks that the targets which should no longer be exported are removed.
@@ -234,7 +238,6 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 			cache = map[string]*targetgroup.Group{}
 			// shut down the discoverer - send latest targets and wait for discoverer goroutine to exit
 			send()
-			<-runExited
 			return
 		case groups := <-ch:
 			for _, group := range groups {
--- a/internal/component/discovery/discovery.go
+++ b/internal/component/discovery/discovery.go
@@ -226,6 +226,12 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				haveUpdates = false
 			}
 		case <-ctx.Done():
+			// Clear the cache
+			//TODO: Write a unit test which changes the config, reloads the component,
+			// and checks that the targets which should no longer be exported are removed.
+			// For example, remove certain namespaces from the config and make sure there is no
+			// __meta_kubernetes_namespace label whose value is set to the removed namespace.
+			cache = map[string]*targetgroup.Group{}
 			// shut down the discoverer - send latest targets and wait for discoverer goroutine to exit
 			send()
 			<-runExited
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,8 +132,8 @@ Main (unreleased)
   Previously, the reload would fail for `loki.process` without an error in the logs and the metrics
   from the `metrics` stage would get stuck at the same values. (@ptodev)
 
-- Fix a bug in `discovery.kubernetes` where old `targets` would continue to be exported to downstream components,
-  even if the component is configured to not discover them. (@ptodev)
+- Fix a bug in `discovery.*` components where old `targets` would continue to be exported to downstream components.
+  This would only happen if the config for `discovery.*`  is reloaded in such a way that no new targets were discovered. (@ptodev)
 
 v1.2.1
 -----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,9 @@ Main (unreleased)
   Previously, the reload would fail for `loki.process` without an error in the logs and the metrics
   from the `metrics` stage would get stuck at the same values. (@ptodev)
 
+- Fix a bug in `discovery.kubernetes` where old `targets` would continue to be exported to downstream components,
+  even if the component is configured to not discover them. (@ptodev)
+
 v1.2.1
 -----------------
 


internal/component/discovery/discovery.go
```diff
@@ -226,6 +226,12 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				haveUpdates = false
 			}
 		case <-ctx.Done():
+			// Clear the cache
+			//TODO: Write a unit test which changes the config, reloads the component,
+			// and checks that the targets which should no longer be exported are removed.
+			// For example, remove certain namespaces from the config and make sure there is no
+			// __meta_kubernetes_namespace label whose value is set to the removed namespace.
+			cache = map[string]*targetgroup.Group{}
```
wildum, line 234: what I understand reading through the code is that because the cache is local to this function, before your change it would send the targets one last time and then exit. But I would expect that this function would run again and it would have a different cache then. How is that not working? By emptying the cache here, it should only affect the last send before the reload

ptodev: Good observation! When [OnStateChange](https://github.com/grafana/alloy/blob/a014bda1725d7817d34523f956812d3ae5e0c809/internal/component/discovery/discovery.go#L215C10-L215C23) is called, the cached exports will be [stored](https://github.com/grafana/alloy/blob/a014bda1725d7817d34523f956812d3ae5e0c809/internal/runtime/internal/controller/node_builtin_component.go#L376) in a structure where future components will reference them. 

This bug only triggers when no pods were discovered after the config is reloaded. In that case `OnStateChange` won't be called, and the stale exports will remain. I updated the PR description and the entry in the changelog to clarify this.

I also changed the PR description and the changelog to reflect that this is a change to `discovery.*` components as a whole, and not just `discovery.kubernetes`.

internal/component/discovery/discovery.go
```diff
@@ -225,10 +225,19 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				send()
 				haveUpdates = false
 			}
-		case <-ctx.Done():
+		// If Run has exited, return.
+		// We can't exit straight after the context is cancelled.
+		// We need to receive from runExited first.
+		// If we don't the first "go func() " will block on sending to runExited.
+		case <-runExited:
```
ptodev, line 232: The [CI was failing](https://drone.grafana.net/grafana/alloy/2364/2/2) because of this. I didn't see this problem when running tests locally though - it's a flaky test.

internal/component/discovery/discovery.go
ptodev: > With this fix, on every reload the discovery components will send empty targets to other components. What is the expected impact of this?

If `discovery` sends empty data to components like `discovery.relabel` or `prometheus.scrape`, I'd expect those components to not do anything. I don't see a downside.

In general, a reload in config can be thought of as starting from a clean slate sometimes. E.g. I think the metrics for the `discovery` component will also get reset after a config reload, if the config for the given component actually changed.

internal/component/discovery/discovery.go
```diff
@@ -225,10 +225,19 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				send()
 				haveUpdates = false
 			}
-		case <-ctx.Done():
+		// If Run has exited, return.
```
ptodev, line 228: @thampiotr @wildum An alternative solution could be to modify the `case` on line 223 to send empty targets on the first tick after the component has started. This would clear the cached targets in cases when no targets were discovered, and it'd make sure we minimise the number of times we send an empty list of targets to downstream component. 

However, I'm not sure if we're guaranteed to discover targets within the first tick when starting a new component. It's possible that no targets were discovered yet, or that the list of targets is still incomplete.

In the case of an incomplete list of targets, if the list of discovered targets is small, I'm not sure if the gain in this solution is much better than simply sending an empty list of targets.

internal/component/discovery/discovery.go
wildum: This is not that trivial. The components have states and they are not recreated between reloads (this is not a clean state). For example the loki.source.docker component runs a worker per target. When a new list of targets comes, it will only compare the hash between the new ones and the ones currently running to only add the new ones and remove the ones that went away. By sending an empty set of targets on every reload, this component will stop all the workers.
This is added processing, maybe this has a noticeable performance impact if you have a lot of targets?
I don't have deep knowledge of all the components that consume targets but we would need to know the trade-offs for all of these components before committing to such a change

ptodev: @thampiotr @wildum Thank you for your feedback - before I start making changes to this PR, I think it'd be good if we agree on what sort of behaviour we want from the `discovery.*` components.

I agree that sending empty targets should be avoided where possible. However, there are two issues with attempting to avoid it:
* It is not known at what time targets will be discovered after a config reload. It might be sooner, or it might be later. And it might depend on how the discovery is configured.
* Even if some targets are discovered, they may not be an exhaustive list. The discovery component may be sending them to the channel one by one, which could result in an incomplete list of targets being sent downstream.
* If the user reconfigures their discovery in such a way that no targets are discovered, then the discovery should send an empty list of targets downstream as soon as possible.

I don't think it's possible to guarantee that we are going to send a "complete" list of discovered targets right after a config reload. I'd argue that downstream components should be able to deal with a temporary removal of a target, and if they don't then that's an issue in itself.

It's also worth noting that the discovery component will be reloaded only if its config is reloaded. Unless the users change their config all the time, I don't expect this to be a problem in practice.

I think a reasonable solution could be that we clear the cache on `Update`, but we don't send it downstream until the first (or second?) time we get a tick on `ticker.C`. This could help reduce unnecessary sending of empty targets in cases where targets are discovered quickly, but it wouldn't help us in cases where discovery is much slower than `MaxUpdateFrequency`.

thampiotr: > This bug only triggers when no pods were discovered after the config is reloaded. In that case OnStateChange won't be called, and the stale exports will remain.

Can we instead make sure that when no pods are discovered we publish empty targets? Instead of temporary publish of empty targets on every reload?

I agree that downstream components should be able to handle targets disappearing, but if they currently don't we will create further issues. This is why I prefer to make a change that has minimal impact. 

Also I think this code needs tests and refactor.

ptodev: > Can we instead make sure that when no pods are discovered we publish empty targets? Instead of temporary publish of empty targets on every reload?

I don't think so. Based on what the [specification](https://github.com/prometheus/prometheus/blob/27579c91484f95f65cf98fe159cae857c11be1b8/discovery/discovery.go#L27-L40) for Prometheus discoverers say, the channel will be updated only when there has been a change in what's been discovered. If you start a new discoverer and it doesn't discover anything, then it simply won't send anything on the channel.

ptodev:
Closing this PR because we seem likely to merge #1473 instead.

PR #1406, opened 2024-08-01
runtime: reuse worker pool for new child controllers

tpaschalis:
#### PR Description

Right now, child controllers started from the NewController start their own individual worker pool for managing component lifecycle. The fact that these controllers are non-root though [prohibits them](https://github.com/grafana/alloy/blob/09548bf1a77c5b704147ec711444f672baf2bbd4/internal/runtime/alloy.go#L244
) from being properly cleaned up, meaning they can leak goroutines.

In the rest of the cases that we spawn module controllers, [the worker pool is inherited](https://github.com/grafana/alloy/blob/09548bf1a77c5b704147ec711444f672baf2bbd4/internal/runtime/alloy.go#L207) from the root, so this PR makes it so that NewController follow along in the same pattern.

No issue filed.

#### Notes to the Reviewer

As discussed, I'd like to _not_ include this PR in the upcoming release, so as to leave some time for dogfooding with this internally.

Feel free to propose more ways we can test this or any raise any other questions you may have.

#### PR Checklist

- [ ] CHANGELOG.md updated (N/A)
- [X] Tests updated

### DIFF
--- a/internal/runtime/alloy_services.go
+++ b/internal/runtime/alloy_services.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/grafana/alloy/internal/runtime/internal/controller"
 	"github.com/grafana/alloy/internal/runtime/internal/dag"
-	"github.com/grafana/alloy/internal/runtime/internal/worker"
 	"github.com/grafana/alloy/internal/service"
 )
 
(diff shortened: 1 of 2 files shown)


thampiotr:
@wildum after release, could you weigh in how/if this could affect modules? I can't remember whether we reuse the pool when modules are started.

tpaschalis:
Btw, I think this is the [code](https://github.com/grafana/alloy/blob/09548bf1a77c5b704147ec711444f672baf2bbd4/internal/runtime/alloy.go#L207) where the worker pool is reused on the NewModuleController.

thampiotr:
> Btw, I think this is the [code](https://github.com/grafana/alloy/blob/09548bf1a77c5b704147ec711444f672baf2bbd4/internal/runtime/alloy.go#L207) where the worker pool is reused on the NewModuleController.

Right, so it is already reused and I think that's great, makes me even less concerned about this change. I'll approve to not block you in the future :)

thampiotr (approved):

wildum (approved):
LGTM!

PR #1414, opened 2024-08-02
Default to `0.0.0.0` instead of `localhost` in OTel receivers

ptodev:
#### PR Description

This rollbacks a recent OTel breaking change, which was ported to Alloy in #1362 but so far it hasn't been released in Alloy officially (it's just been on the main branch for a few days).

The initial argument for having the breaking change in Alloy was that it gives users a more up to date OTel experience, which aligns with the upstream defaults. However, this particular breaking change has such a large impact that we decided the best time to introduce it would be in Alloy 2.0, which will probably go out at some point next year. 

#### Notes to the Reviewer

The converters will now explicitly convert `endpoint` attributes to `localhost`, since `localhost` is what upstream uses by default.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,6 @@ v1.3.0-rc.0
 
 ### Breaking changes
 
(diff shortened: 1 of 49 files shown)


wildum (approved):
Thanks for taking care of this

PR #1472, opened 2024-08-14
Display remote configuration components in Alloy's local UI

tpaschalis:
#### PR Description

This PR wires in the Data() call of the remotecfg service that exposes its isolated Host to be able to display remotely loaded components via the UI.

To do this, we introduce new API endpoinst prefixed with /remotecfg, and make use of the existing layout pages to call to either host dynamically. To do that, I'm having  `useLocation()`'s output as a parameter, depending on the current URL the user has navigated to.

This shouldn't impact any of the existing UI capabilities, just adding new ones. I've tested with nested modules under remotecfg's controller and they seem to be working properly.

No issue filed.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

image
image
image

### DIFF
--- a/internal/web/api/api.go
+++ b/internal/web/api/api.go
@@ -23,6 +23,7 @@ import (
 
 // AlloyAPI is a wrapper around the component API.
 type AlloyAPI struct {
+	urlPrefix       string
 	alloy           service.Host
 	remotecfg       service.Host
 	CallbackManager livedebugging.CallbackManager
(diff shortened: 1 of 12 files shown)


internal/web/ui/src/Router.tsx
```diff
@@ -18,7 +18,11 @@ const Router = ({ basePath }: Props) => {
       <main>
         <Routes>
           <Route path="/" element={<PageComponentList />} />
+          <Route path="/remotecfg" element={<PageComponentList />} />
+
           <Route path="/component/*" element={<ComponentDetailPage />} />
+          <Route path="/remotecfg/component/*" element={<ComponentDetailPage />} />
```
tpaschalis, line 24: I tried to reuse the same Page types, by providing a different functionality based on the URL it's called from via `useLocation`.

Let me know if you think that's a good approach or if you'd prefer separate pages here.

internal/web/api/api.go
```diff
@@ -71,6 +78,32 @@ func (a *AlloyAPI) listComponentsHandler() http.HandlerFunc {
 	}
 }
 
+func (a *AlloyAPI) listRemotecfgComponentsHandler() http.HandlerFunc {
```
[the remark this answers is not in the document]
ptodev: Yes please, reducing the code duplication would be great!

internal/web/ui/src/Router.tsx
ptodev: I don't know anything about web development, but reusing code as a whole is good. My only concern with this approach is that the "Remote Configuration" link in the navigation bar doesn't seem to be highlighted, and it's not clear that the user is actually looking at remote configuration. Is that very hard to fix?

ptodev (commented):
I don't know anything about react, but on a high level it looks ok to me 😄 I only added a few minor comments.

internal/web/ui/src/Router.tsx
tpaschalis: I've pushed a commit to reduce code deduplication and also conditionally set the page's name and description (so when you click on Remote Configuration) it looks like this:
image

wildum: I think that the cleanest way would be to have different pages but the same components to balance the extensibility and the code duplication. It's possible that in the future we want to add some particular features or designs in the remote config page and wrapping everything with the useRemoteCfg in the common page would add complexity

wildum (commented):
one additional comment that I did not know where to put in the file: Is livedebugging working for components configured via the remotecfg?

internal/web/api/api.go
```diff
@@ -23,28 +23,44 @@ import (
...
 	// component IDs.
 
 	r.Handle(path.Join(urlPrefix, "/modules/{moduleID:.+}/components"), httputil.CompressionHandler{Handler: a.listComponentsHandler()})
+	r.Handle(path.Join(urlPrefix, "/remotecfg/modules/{moduleID:.+}/components"), httputil.CompressionHandler{Handler: a.listComponentsHandler()})
+
 	r.Handle(path.Join(urlPrefix, "/components"), httputil.CompressionHandler{Handler: a.listComponentsHandler()})
+	r.Handle(path.Join(urlPrefix, "/remotecfg/components"), httputil.CompressionHandler{Handler: a.listComponentsHandler()})
+
 	r.Handle(path.Join(urlPrefix, "/components/{id:.+}"), httputil.CompressionHandler{Handler: a.getComponentHandler()})
+	r.Handle(path.Join(urlPrefix, "/remotecfg/components/{id:.+}"), httputil.CompressionHandler{Handler: a.getComponentHandler()})
```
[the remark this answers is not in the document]
tpaschalis: Thanks for the feedback. I've went ahead and implemented this, looking cleaner now.

internal/web/ui/src/Router.tsx
tpaschalis: Thanks for the comment, makes sense.

I've added two new pages for the ComponentList and ComponentDetail functionality, which also made them a little simpler (no need to parse the URL itself to figure out whether to useRemoteCfg in https://github.com/grafana/alloy/pull/1472/commits/b3fa5680bea6d25d7c4f033764a2b8c390822028.

The Graph page, I've left as-is for now (the pipelines are not connected in any way _yet_), and the clustering page would be the same for both controllers.

wildum (approved):

PR #1473, opened 2024-08-14
Prevent stale targets from remaining in discovery + tests

thampiotr:
#### PR Description

This is an alternative solution to https://github.com/grafana/alloy/pull/1398/files

#### Notes to the Reviewer

There is some concurrency to this test for obvious reasons, but I tried to make sure it's working fine and no race conditions and no flakiness:

```
go test -race -count 200 ./internal/component/discovery
# github.com/grafana/alloy/internal/component/discovery.test
ok  	github.com/grafana/alloy/internal/component/discovery	19.830s
```

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/discovery/discovery.go
+++ b/internal/component/discovery/discovery.go
@@ -216,8 +216,9 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 	}
 
 	ticker := time.NewTicker(MaxUpdateFrequency)
-	// true if we have received new targets and need to send.
-	haveUpdates := false
+	// true if we have received new targets and need to send. Initially set it to true to send empty targets in case
+	// the discoverer never sends any targets.
+	haveUpdates := true
 	for {
 		select {
 		case <-ticker.C:
--- a/internal/component/discovery/discovery_test.go
+++ b/internal/component/discovery/discovery_test.go
@@ -1,0 +1,227 @@
+package discovery
+
+import (
+	"context"
+	"os"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/alloy/internal/component"
+)
+
+// discovererUpdateTestCase is a test case for testing discovery updates. A discovery component is created and the
+// initialTargets are published. We check that the component correctly publishes exports matching exepectedInitialExports.
+// Then, the discoverer is updated and new updatedTargets are published. We check that the exports published so far
+// match the expectedUpdatedExports. Finally, the component is shut down, and we check that the list of exports published
+// matches the expectedFinalExports.
+type discovererUpdateTestCase struct {
+	name                   string
+	initialTargets         []*targetgroup.Group
+	expectedInitialExports []component.Exports
+	updatedTargets         []*targetgroup.Group
+	expectedUpdatedExports []component.Exports
+	expectedFinalExports   []component.Exports
+}
+
+var updateTestCases = []discovererUpdateTestCase{
+	{
+		name: "from one target to another",
+		initialTargets: []*targetgroup.Group{
+			{Source: "test", Labels: model.LabelSet{"test_key": "value"}, Targets: []model.LabelSet{{"foo": "bar"}}},
+		},
+		expectedInitialExports: []component.Exports{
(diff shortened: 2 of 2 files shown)


internal/component/discovery/discovery.go
```diff
@@ -216,8 +216,9 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 	}
 
 	ticker := time.NewTicker(MaxUpdateFrequency)
-	// true if we have received new targets and need to send.
-	haveUpdates := false
+	// true if we have received new targets and need to send. Initially set it to true to send empty targets in case
+	// the discoverer never sends any targets.
+	haveUpdates := true
```
ptodev, line 221: Will it send empty targets if there are already targets in the cache?

thampiotr: If targets are added to `cache` map, this means that the `case groups := <-ch:` triggered before `case <-ticker.C:`. The `haveUpdates` remains `true`. 

Then the `case <-ticker.C:` would typically trigger and we would publish the targets in the `cache` map. 

However, if the underlying discovery never publishes anything, the `case groups := <-ch:` will not be hit and instead the `case <-ticker.C:` would be triggered first and an empty `cache` map would be published. 

Hope this answers your question!

ptodev: Yes, that's right! I forgot that the cache is local to this function. It makes sense now.

ptodev:
IMO this change is a good compromise and I'm happy with it. It will reduce the amount of times we send empty/incomplete targets in some real world situations. However, as [this comment](https://github.com/grafana/alloy/pull/1398#discussion_r1699963531) says, there is no guarantee that the list of targets will be complete by the time the first tick happens. It's possible that nothing is discovered by then because the SD could be taking a while; in that case, we will send an empty list of targets downstream even though we might be just about to discover a few targets.

The only situation I can think of where this change performs poorly is when:
1. Targets are discovered and sent to downstream components.
2. The config is changed in such a way that no targets are discovered.
3. A config reload is triggered every 3 seconds (which is less than the hardcoded 5 second tick time). The new config is structured in such a way that no new targets are discovered.
4. Since the config is reloaded before the component has a chance to do its first tick, no empty list of targets will be sent.

I personally think the likelyhood of such a scenario is quite low 😆 On the other hand, the PR will fix some real world issues.  So I think the tradeoff is worth it.

thampiotr:
Thanks for taking a look! Responding to some concerns inline below.

> However, as [this comment](https://github.com/grafana/alloy/pull/1398#discussion_r1699963531) says, there is no guarantee that the list of targets will be complete by the time the first tick happens. It's possible that nothing is discovered by then because the SD could be taking a while; in that case, we will send an empty list of targets downstream even though we might be just about to discover a few targets.

Correct. We have a `MaxUpdateFrequency` deadline (currently 5s) for the new discoverer to send some targets when we just started the discoverer. If it doesn't, we'll publish empty targets. I think it's a good compromise given that there's no way to differentiate between discoverer being slow or stuck vs. genuinely finding no targets at all. If it becomes unstuck or finally completes, we will happily publish the new targets.

> The only situation I can think of where this change performs poorly is when:
> 
> 1. Targets are discovered and sent to downstream components.
> 2. The config is changed in such a way that no targets are discovered.
> 3. A config reload is triggered every 3 seconds (which is less than the hardcoded 5 second tick time). The new config is structured in such a way that no new targets are discovered.
> 4. Since the config is reloaded before the component has a chance to do its first tick, no empty list of targets will be sent.

The `4.` is not quite the case - on new config the `ctx` will be done and [we will call `send()` before exit](https://github.com/grafana/alloy/blob/f86e22f95f27fa01513914bd1edb530ef36d65d1/internal/component/discovery/discovery.go#L231) - the tests I've added are checking this behaviour. So even though there is not enough time for the first tick to occur, after reload is triggered, the empty targets will be published. I think this is fine.
 
Let me know if you have further comments or we can approve and merge this?

ptodev:
> The `4.` is not quite the case - on new config the `ctx` will be done and [we will call `send()` before exit](https://github.com/grafana/alloy/blob/f86e22f95f27fa01513914bd1edb530ef36d65d1/internal/component/discovery/discovery.go#L231) - the tests I've added are checking this behaviour. So even though there is not enough time for the first tick to occur, after reload is triggered, the empty targets will be published. I think this is fine.
> 
> Let me know if you have further comments or we can approve and merge this?

Ah yes, you're right! Sounds great in that case. I'm happy to approve it 🚢

ptodev (approved):

PR #1474, opened 2024-08-14
Update encoding location

### LINKED ISSUE #1051: Docs feedback: OTLPHTTP Exporter Encoding argument incorrectly defined
### URL

https://grafana.com/docs/alloy/latest/reference/components/otelcol.exporter.otlphttp/

### Feedback

The documntation states that the encoding argument lives within the Client block; however, in practice it actually lives outside the Client block. The actual component is built [here](https://github.com/grafana/alloy/blob/main/internal/component/otelcol/exporter/otlphttp/otlphttp.go) and we can see that in the Arguments block starting on line 34 that the Encoding argument is not a part of the client args.

I am curious if the best course of action here is to update the documentation to reflect this difference, or if it is better to just move the Encoding declaration further down the file into the Client Args?

clayton-cornell:
#### PR Description

The encoding argument is defined to be within the client block, but it should be one level up. This PR moves the definition of this argument to the component level.

Fixes #1051 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.otlphttp.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.otlphttp.md
@@ -31,11 +31,12 @@ otelcol.exporter.otlphttp "LABEL" {
 
 `otelcol.exporter.otlphttp` supports the following arguments:
 
-Name               | Type     | Description                      | Default                           | Required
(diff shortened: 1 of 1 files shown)


mattdurham (approved):
Lgtm

PR #1480, opened 2024-08-15
Dynamic pipelines - a new `foreach` block

### LINKED ISSUE #1443: Integrate `prometheus.exporter` and `discovery` components
### Request

`prometheus.exporter` components tend to not take `targets` as "input" attributes. Some do (e.g. [prometheus.exporter.snmp](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.snmp/) and [prometheus.exporter.blackbox](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.blackbox/)), but others such as [prometheus.exporter.kafka](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.kafka/) simply have an attribute for the URL of the system they are monitoring (`kafka_uris`). It would be nice to leverage the `discovery` components when working with `prometheus.exporter`, so that the metrics can have extra labels.

The simplest way I can think of is to have an additional argument for every `prometheus.exporter` that takes in `targets`. For example, `prometheus.exporter.kafka` can have a `kafka_targets` argument. `kafka_targets` must contain a special `__address__` label with the Kafka API URL, and it may contain various additional labels which can be passed down to `prometheus.scrape`.

An alternative approach for `prometheus.exporter.kafka` would be to leave the existing `kafka_uris` attribute as the only attribute which can contain Kafka API URLs, and instead we can have some sort of `additional_labels` attribute. However, it may be harder to associate the right labels with the right kafka instance.

### Use case

You may want to run `prometheus.exporter` to monitor a Kafka or Redis running on k8s, and you may want k8s labels such as pod name.

ptodev:
#### PR Description

This is a proposal for starting up Alloy pipelines dynamically based on data in transit. For example, if a `discovery` component comes up with 10 targets, Alloy can start 10 sub-pipelines, each dealing with a different target.

The PR consists of a a design doc in `docs/design/` and an experimental implementation. The user will have to have the "experimental" cmd flag turned on.

Fixes #1443


docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+* Author: Paulin Todev (@ptodev), Piotr Gwizdala (@thampiotr)
+* Last updated: 2024-08-15
+* Original issue: https://github.com/grafana/alloy/issues/1443
+
+## Abstract
+
+We are proposing a new feature to the [Alloy standard library][stdlib].
+It will be similar to a `map` operation over a collection such as a `list()`.
+Each `map` transformation will be done by a chain of components (a "sub-pipeline") created for this transformation.
+Each item in the collection will be processed by a different "sub-pipeline".
```
ptodev, line 12: I wonder how feasible this is when the number of items in the collection is really big. Should there be a limit to how many sub-pipelines there are?

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+
+We should find answers to the unknowns below before this proposal is accepted:
+
+* Will the solution only work for `list()`? What about `map()`?
+  * If we go with a `foreach`, we could have a `key` attribute in addition to the `var` one. That way we can also access the key. The `key` attribute can be a no-op if `collection` is a map?
+* What about debug metrics? Should we aggregate the metrics for all "sub-pipelines"?
+  * If there is 1 series for each sub-pipeline, the amount of metrics could be huge. 
+    Some service discovery mechanisms may generate a huge number of elements in a list of targets.
+  * If we want to aggregate the metrics, how would we do that? Is it even possible to do in within Alloy?
+  * Can we have a configuration parameter which dictates whether the metrics should be aggregated or not?
```
ptodev, line 244: This could become a huge blocker. I don't think we want a new series for each sub-pipeline, and I am not sure if we can avoid it.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+  <summary>Example</summary>
+
+```alloy
+discovery.file "default" {
+  files = ["/Users/batman/Desktop/redis_addresses.yaml"]
+}
+
+// Every component defined in the "foreach" block will be instantiated for each item in the collection. 
+// The instantiated components will be scoped using the name of the foreach block and the index of the
+// item in the collection. For example: /foreach.redis/0/prometheus.exporter.redis.default
```
wildum, line 77: > /foreach.redis/0/prometheus.exporter.redis.default

Targets can come and go between discovery intervals. If you use the index to scope them you will need to shift it on every update. Maybe you can just use the target itself which should be unique: /foreach.redis/target/prometheus.exporter.redis.default

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+
+* Will the solution only work for `list()`? What about `map()`?
+  * If we go with a `foreach`, we could have a `key` attribute in addition to the `var` one. That way we can also access the key. The `key` attribute can be a no-op if `collection` is a map?
+* What about debug metrics? Should we aggregate the metrics for all "sub-pipelines"?
+  * If there is 1 series for each sub-pipeline, the amount of metrics could be huge. 
+    Some service discovery mechanisms may generate a huge number of elements in a list of targets.
+  * If we want to aggregate the metrics, how would we do that? Is it even possible to do in within Alloy?
+  * Can we have a configuration parameter which dictates whether the metrics should be aggregated or not?
+* Do we have to recreate the sub-pipelines every time a new collection is received,
+  even if the new collection has the same number of elements?
```
wildum, line 246: if you identify each sub-pipeline by its corresponding target, it should be ok to only add/remove sub-pipelines according to the new set of targets and keep the sub-pipelines that are still relevant running.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+}
+```
+
+<details>
+  <summary>Example</summary>
+
+```alloy
+// declare.dynamic "maps" each target to a sub-pipeline.
+// Each sub-pipeline has 1 exporter, 1 relabel, and 1 scraper.
+// Internally, maybe one way this can be done via serializing the pipeline to a string and then importing it as a module?
```
wildum, line 156: that's an interesting idea. One way to do this could be similar to the import node. The declare.dynamic instance would manage a list of custom component nodes that it can run using the runner package. These custom component nodes would not be part of the initial graph and would take the pipeline template as AST body to run.

docs/design/1443-dynamic-pipelines.md
ptodev: > Maybe you can just use the target itself which should be unique

What do you mean by "the target itself"? Do you mean the "__address__" label? We're not guaranteed to have such a label on the collection which is being iterated on :) The collection might not be `targets`. Also, some `targets` might not follow this convention of having an `__address__` label.

wildum: the target (or more generally the item in the collection) can be hashed and we can use the hash for scoping and for avoid recomputations on update when a new collection comes

thampiotr: Yeah, I think targets going away in the middle of the list is a good point to consider. Hashing the way that William is proposing might work and we'd need to recommend people to try change the input targets as little as possible (e.g. use a reasonable refresh interval), otherwise there will be a lot of churn.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+}
+```
+
+</details>
+
+Pros:
+* The `foreach` name is consistent with other programming languages.
+
+Cons:
+* It looks less like a component than a `declare.dynamic` block.
```
thampiotr, line 120: I don't think it's a problem, you can still define a `declare` component (or import one) and use it inside the `foreach` syntax to make it smaller, add the parameters you need and reuse it for multiple foreach blocks.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+The [Alloy type][alloy-types] of the list above is `list(map(string))`.
+However, you may want to pipe information from this list of targets to a component which doesn't work with a `list()` or a `map()`.
+For example, you may want to input the `"__address__"` string to a `prometheus.exporter.redis`,
+and you may want to use the `"instance"` string in a `discovery.relabel`.
+
+[alloy-types]: https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/
+
+## Proposal 1: A foreach block
```
thampiotr, line 44: +1 to this one IMO.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+
+* Author: Paulin Todev (@ptodev), Piotr Gwizdala (@thampiotr)
+* Last updated: 2024-08-15
+* Original issue: https://github.com/grafana/alloy/issues/1443
+
+## Abstract
+
+We are proposing a new feature to the [Alloy standard library][stdlib].
+It will be similar to a `map` operation over a collection such as a `list()`.
+Each `map` transformation will be done by a chain of components (a "sub-pipeline") created for this transformation.
```
thampiotr, line 11: Nit: I am not a fan of double-quoted quasi-concept of "sub-pipeline"... it's not essential to this proposal, but coming up with a better way to explain it using well-defined concepts is something we'll need to do for the documentation of this feature.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+
+Cons:
+* Not possible to integrate most `prometheus.exporter` components with the `discovery` ones.
+
+## Unknowns
+
+We should find answers to the unknowns below before this proposal is accepted:
+
+* Will the solution only work for `list()`? What about `map()`?
+  * If we go with a `foreach`, we could have a `key` attribute in addition to the `var` one. That way we can also access the key. The `key` attribute can be a no-op if `collection` is a map?
```
thampiotr, line 239: Alternatively, a map of `{k1: v1, k2: v2}` can be translated into list of entries: `[{key: k1, value: v1}, {key: k2, value: v2}]`.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+* What about debug metrics? Should we aggregate the metrics for all "sub-pipelines"?
+  * If there is 1 series for each sub-pipeline, the amount of metrics could be huge.
```
thampiotr, line 241: Explosion of metrics is an existing issue with Alloy and using modules...
Also, currently, if someone wanted to implement the use-case of, e.g. 10 redis exporters, and they chose to use modules, they would also get a ton of metrics AFAIU. 

So I don't think it's a problem that should be part of this proposal. This is not to say that I don't think it's a problem, we should ideally look into how we can handle metrics for modules a bit better. But IMO this should be a new proposal that will deal with this problem for all usages of modules.

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+* Will the solution only work for `list()`? What about `map()`?
+  * If we go with a `foreach`, we could have a `key` attribute in addition to the `var` one. That way we can also access the key. The `key` attribute can be a no-op if `collection` is a map?
+* What about debug metrics? Should we aggregate the metrics for all "sub-pipelines"?
+  * If there is 1 series for each sub-pipeline, the amount of metrics could be huge. 
+    Some service discovery mechanisms may generate a huge number of elements in a list of targets.
+  * If we want to aggregate the metrics, how would we do that? Is it even possible to do in within Alloy?
+  * Can we have a configuration parameter which dictates whether the metrics should be aggregated or not?
+* Do we have to recreate the sub-pipelines every time a new collection is received,
+  even if the new collection has the same number of elements?
+* Do we need to have more than one output, of a different type?
```
thampiotr, line 247: Not sure what you mean... in Proposal 1 you can output to as many things as you want, AFAICT

docs/design/1443-dynamic-pipelines.md
```diff
@@ -0,0 +1,252 @@
...
+* Do we need to have more than one input, of a different type?
```
thampiotr, line 248: IMO no. You can use multiple `foreach` or perhaps merge your targets using something like the `join` in your other proposal?

docs/design/1443-dynamic-pipelines.md
ptodev: Theoretically, couldn't there be hash collisions? I don't see how hashing is a very resilient way of handling this.

thampiotr: With [sufficiently long hash](https://stackoverflow.com/questions/62664761/probability-of-hash-collision) you don't need to worry about collisions. Unless there are implementation bugs. 

If collisions worry you or we want to use shorter IDs than a 512bit hash, we could think of a mapper that will assign stable IDs to targets, using a simple `int` sequence.

ptodev:
It looks likely that we are going to go with the `foreach` option. I updated the PR to add some draft code that I've been working on. As a first step, it'd be good to have a unit tests that doesn't use the proposed `var` argument. The test could look something like what's proposed in `internal/runtime/testdata/foreach/foreach_1.txtar`.

As a second step, we will need to figure out how to refer to the values that are currently iterated on. Should we actually have a `var` argument? I think it might be unnecessary - having a predefined argument (e.g. `val`) which refers to the current thing being iterated on may be enough. I'm not yet sure how to add this reserved word into the Alloy syntax though.

By the way, apparently the Collector already has something like dynamic pipelines. There are [observer extensions](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.112.0/extension/observer) which you can hook up to a [recover creator component](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.112.0/receiver/receivercreator/README.md). For example, you can have a k8s observer which crates a receiver for each discovered pod.

docs/design/1443-dynamic-pipelines.md
ptodev: For now we will go with no limit.

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+func (fi *forEachChild) Hash() uint64 {
+	fnvHash := fnv.New64a()
+	fnvHash.Write([]byte(fi.id))
+	return fnvHash.Sum64()
+}
+
+func (fi *forEachChild) Equals(other runner.Task) bool {
+	return fi.id == other.(*forEachChild).id
```
mattdurham, line 374: Do we need a check here? If other is non forEachChild this will panic.

internal/runtime/foreach_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/grafana/alloy/internal/runtime"
+	alloy_runtime "github.com/grafana/alloy/internal/runtime"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/tools/txtar"
+)
+
+func TestForeach(t *testing.T) {
+	directory := "./testdata/foreach"
```
mattdurham, line 24: I dont think I saw a test for iterating over an array of capsule values. `collection = [prometheus.remote_write.d1,prometheus.remote_write.d2]`, now why would I do that, no idea but we should know what the behavior is.

mattdurham (commented):
Fantastic work, gave it a first review and will go over it again. Added some comments on tests.

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -0,0 +1,412 @@
...
+
+func hashObject(obj any) string {
+	//TODO: Test what happens if there is a "true" string and a true bool in the collection.
+	switch v := obj.(type) {
+	case int, string, bool:
+		return fmt.Sprintf("%v", v)
+	case float64:
+		// Dots are not valid characters in Alloy syntax identifiers.
+		// For example, "foreach_3.14_1" should become "foreach_3_14_1".
+		return strings.Replace(fmt.Sprintf("%f", v), ".", "_", -1)
```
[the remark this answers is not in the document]
wildum: very nice catch, fixed in: https://github.com/grafana/alloy/pull/1480/commits/0b42dc374ad7c3f020f01e3013c89a40d87b5789

The rule is only alpha numeric chars + no leading digits but we can allow the digits part since we have the foreach_ prefix

internal/runtime/internal/controller/node_config_foreach_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/grafana/alloy/syntax/parser"
+	"github.com/grafana/alloy/syntax/vm"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+func TestCreateCustomComponents(t *testing.T) {
+	config := `foreach "default" {
```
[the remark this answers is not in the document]
wildum: the Alloy syntax does not allow setting a string as a slice, added a test for it: https://github.com/grafana/alloy/pull/1480/commits/8611fa3de112f5d0a22c7bef6220e5002e3cec1f

(fun fact, the Otel syntax allows it. If you set a string value instead of a slice of strings, it will just consider it as a slice of string with one element in it)

internal/runtime/internal/controller/node_config_foreach.go
wildum: It can't happen because the runner in the foreach is only used with forEachChild tasks. If the code is modified in a way that it has other tasks, then either it's a bug or the Equals must be modified anyway (else it will create big problems). In both cases I would prefer to panic to ensure that the code is fixed right away rather than running the risk that it goes unnoticed and we struggle later to explain weird behaviours of the foreach

internal/runtime/foreach_test.go
wildum: good point, added here: https://github.com/grafana/alloy/pull/1480/commits/7db785903883104843cf901ae903c1c3cca6a887

internal/runtime/internal/controller/node_config_foreach.go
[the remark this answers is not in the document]
ptodev: I added a few more tests for non-alphanumeric strings too, in 1ab03b984ca8012d57570d7f6cff0fed4ed0175d.

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/stdlib/foreach/
+description: Learn about foreach
+menuTitle: foreach
+title: foreach
+---
+
+<span class="badge docs-labels__stage docs-labels__item">Experimental</span>
```
clayton-cornell, line 8:
```suggestion
labels:
  stage: experimental
menuTitle: foreach
title: foreach
---

```

We can now use the labels.stage metadata to add an `Experimental` badge

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Arguments
+
+The following arguments are supported:
+
+Name             | Type        | Description                                                           | Default | Required
+-----------------|-------------|-----------------------------------------------------------------------|---------|---------
+`collection`     | `list(any)` | A list of items to loop over.                                         |         | yes
+`var`            | `string`    | Name of the variable referring to the current item in the collection. |         | yes
+`enable_metrics` | `bool`      | Whether to expose debug metrics in the {{< param "PRODUCT_NAME" >}} `/metrics` endpoint.       | `false` | no
```
clayton-cornell, line 36:
```suggestion
Name             | Type        | Description                                                                              | Default | Required
-----------------|-------------|------------------------------------------------------------------------------------------|---------|---------
`collection`     | `list(any)` | A list of items to loop over.                                                            |         | yes
`var`            | `string`    | Name of the variable referring to the current item in the collection.                    |         | yes
`enable_metrics` | `bool`      | Whether to expose debug metrics in the {{< param "PRODUCT_NAME" >}} `/metrics` endpoint. | `false` | no
```

Just formatting the table

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+            password = sys.env("<GRAFANA_CLOUD_API_KEY>")
+        }
+    }
+}
+```
+
+Replace the following:
+
+* _`<PROMETHEUS_USERNAME>`_: Your Prometheus username.
+* _`<GRAFANA_CLOUD_API_KEY>`_: Your Grafana Cloud API key.
```
clayton-cornell, line 198:
```suggestion
* _`<GRAFANA_CLOUD_API_KEY>`_: Your Grafana Cloud API key.

```

Markdown style says we should have an empty line at the end of the file.

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Example
+
+The following example shows you how to run Prometheus exporters dynamically on service discovery targets.
+
+`prometheus.exporter.*` components often require the address of one particular instance being monitored.
+For example, `prometheus.exporter.redis` has a `redis_addr` attribute for the Redis instance under observation.
+On the other hand, `discovery.*` components such as `discovery.kubernetes` output a list of targets such as this:
+
+{{< collapse title="Example targets output by discovery.kubernetes" >}}
```
clayton-cornell, line 73:
```suggestion
{{< collapse title="Example targets output by `discovery.kubernetes`" >}}
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+The following blocks are supported inside the definition of `foreach`:
+
+Hierarchy | Block        | Description                  | Required
+----------|--------------|------------------------------|---------
+template  | [template][] | A component pipeline to run. | yes
+
+[template]: #template-block
+
+### template
```
clayton-cornell, line 56:
```suggestion
### `template`
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Blocks
+
+The following blocks are supported inside the definition of `foreach`:
+
+Hierarchy | Block        | Description                  | Required
+----------|--------------|------------------------------|---------
+template  | [template][] | A component pipeline to run. | yes
+
+[template]: #template-block
```
clayton-cornell, line 54:
```suggestion
[template]: #template
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+[types]: ../../../get-started/configuration-syntax/expressions/types_and_values
+
+## Blocks
+
+The following blocks are supported inside the definition of `foreach`:
+
+Hierarchy | Block        | Description                  | Required
+----------|--------------|------------------------------|---------
+template  | [template][] | A component pipeline to run. | yes
```
clayton-cornell, line 52:
```suggestion
Block        | Description                  | Required
--------------|------------------------------|---------
[template][] | A component pipeline to run. | yes
```

We can drop the hierarchy column. I'm working on removing it throughout the docs. It's duplicate information and can be represented well enough in the `Block` column with it's accompanying explanation where there are nested blocks.

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+  var        = "<VAR_NAME>"
+  template {
+    ...
+  }
+}
+```
+
+## Arguments
+
+The following arguments are supported:
```
clayton-cornell, line 30:
```suggestion
You can use the following arguments with `foreach`:
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -0,0 +1,198 @@
...
+
+{{< admonition type="warning" >}}
+Setting `enable_metrics` to `true` when `collection` has lots of elements may cause a large number of metrics to appear on the {{< param "PRODUCT_NAME" >}} `/metric` endpoint.
+{{< /admonition >}}
+
+[types]: ../../../get-started/configuration-syntax/expressions/types_and_values
+
+## Blocks
+
+The following blocks are supported inside the definition of `foreach`:
```
clayton-cornell, line 48:
```suggestion
You can use the following blocks with `foreach`:
```

This sentence change is being implemented in the linting cleanup. It removes the passive voice which is a goal we aim for in all docs.

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -0,0 +1,427 @@
...
+	var args Arguments
+	if err := eval.Evaluate(scope, &args); err != nil {
+		return fmt.Errorf("decoding configuration: %w", err)
+	}
+
+	fn.args = args
+
+	// By default don't show debug metrics.
+	if args.EnableMetrics {
+		fn.moduleControllerOpts.Reg = nil
```
mattdurham, line 175: Not sure I understand the line here. This appears to nil out the line if Metrics are enabled?

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -0,0 +1,427 @@
...
+func (fr *forEachChildRunner) Run(ctx context.Context) {
+	err := fr.child.cc.Run(ctx)
+	if err != nil {
+		level.Error(fr.child.logger).Log("msg", "foreach child stopped running", "err", err)
+		fr.child.healthUpdate(component.HealthTypeUnhealthy, fmt.Sprintf("foreach child stopped running: %s", err))
+	}
+}
+
+func (fi *forEachChild) Hash() uint64 {
+	fnvHash := fnv.New64a()
```
mattdurham, line 369: Why are we using two different hash functions? sha256 and fnv, though they dont collide in usage, it feels odd.

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -0,0 +1,427 @@
...
+	return fi.id == other.(*forEachChild).id
+}
+
+func computeHash(s string) string {
+	hasher := sha256.New()
+	hasher.Write([]byte(s))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
+
+func hashObject(obj any) string {
```
mattdurham, line 384: is there a better name than hash? Not sure if these count has hashe

internal/runtime/alloy.go
```diff
@@ -207,17 +207,22 @@ func newController(o controllerOptions) *Runtime {
 			OnExportsChange: o.OnExportsChange,
 			Registerer:      o.Reg,
 			ControllerID:    o.ControllerID,
-			NewModuleController: func(id string) controller.ModuleController {
+			NewModuleController: func(opts controller.ModuleControllerOpts) controller.ModuleController {
+				reg := o.Reg
```
mattdurham, line 211: Maybe add a comment here on why we are doing this, or what situations one is nil vs the other.

mattdurham (approved):
Small comments but overall fantastic!

thampiotr (approved):

internal/runtime/internal/controller/node_config_foreach.go
ptodev: @wildum WDYT about renaming `hashObject` to `objectFingerprint`? Similarly to [this](https://pkg.go.dev/github.com/prometheus/common@v0.61.0/model#LabelSet.Fingerprint).

internal/runtime/internal/controller/node_config_foreach.go
ptodev: Apologies, you're right that it's not very clear. I added a few comments and renamed `Reg` to `RegOverride`. I hope this helps?

internal/runtime/internal/controller/node_config_foreach_test.go
```diff
@@ -239,6 +239,41 @@ func TestNonAlphaNumericString(t *testing.T) {
...
+		collection = ["123./s4", "123__s4_1"]
+		var = "num"
+		template {
+		}
+	}`
+	foreachConfigNode := NewForeachConfigNode(getBlockFromConfig(t, config), getComponentGlobals(t), nil)
+	require.NoError(t, foreachConfigNode.Evaluate(vm.NewScope(make(map[string]interface{}))))
+	customComponentIds := foreachConfigNode.moduleController.(*ModuleControllerMock).CustomComponents
+	//TODO: It's not very clear which item became "foreach_123__s4_1_1".
+	// To avoid confusion, maybe we should log a mapping?
```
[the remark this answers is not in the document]
ptodev: True, but I'd prefer to keep the todo if you don't mind? I still think there's room for improvement there.

clayton-cornell (approved):

internal/runtime/internal/controller/node_config_foreach.go
wildum: sounds good to me

internal/runtime/internal/controller/node_config_foreach.go
ptodev: IIUC, for items in the collection the hash is 256 bits since there could be lots of things in each item in a collection, and it gives us extra protection against collisions. For the foreach ID we use a 64 bit hash since it's just a string. IMO it's ok to use different hashes, but we do need to document why each situation uses a different one.

wildum: we use the hash of the items in the collection for the foreachID and the foreachID as a key: https://github.com/grafana/alloy/pull/1480/files#diff-9cebda46e5a40368c4b76027ff2bda36114d2737124b5390aa35fa6435d79127R193
The 64 bits hash is used for the runner that wraps around the foreach to run it.
What's in an item in a collection should not matter. Whether the object has 3 or 30 fields, it should not have any influence on the collision I believe. The only difference would be the number of items in the collection and we would need billions of items for a collision to be likely. 
I'm ok to keep it as it is and add a comment that we keep the 256 bits hash for extra security and that if one day it becomes a performance bottleneck it would be ok to use a 64 bits hash instead

wildum: I added the comment

PR #1487, opened 2024-08-19
Add appProtocol for OpenShift

### LINKED ISSUE #1405: Add variable appProtocol Option to Helm Chart for your Alloy Service for h2c/http2
### Request

At the moment it is possible to provide the "extraPorts:" option within the Alloy helm chart and by default the below are set,:

```
...
    alloy:
      extraPorts:
        - name: "trace-http"
          port: 4318
          targetPort: 4318
          protocol: "TCP"
        - name: "trace-grpc"
          port: 4317
          targetPort: 4317
          protocol: "TCP"
 ...         
``` 

This Service.yaml is propagated by the serice.yaml within the templates of the helm chart:

```
{{- $values := (mustMergeOverwrite .Values.alloy (or .Values.agent dict)) -}}
{{- if .Values.service.enabled -}}
apiVersion: v1
kind: Service
metadata:
  name: {{ include "alloy.fullname" . }}
  labels:
    {{- include "alloy.labels" . | nindent 4 }}
  {{- with .Values.service.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  type: {{ .Values.service.type }}
  {{- if .Values.service.clusterIP }}
  clusterIP: {{ .Values.service.clusterIP }}
  {{- end }}
  selector:
    {{- include "alloy.selectorLabels" . | nindent 4 }}
  internalTrafficPolicy: {{.Values.service.internalTrafficPolicy}}
  ports:
    - name: http-metrics
      {{- if eq .Values.service.type "NodePort" }}
      nodePort: {{ .Values.service.nodePort }}
      {{- end }}
      port: {{ $values.listenPort }}
      targetPort: {{ $values.listenPort }}
      protocol: "TCP"
{{- range $portMap := $values.extraPorts }}
    - name: {{ $portMap.name }}
      port: {{ $portMap.port }}
      targetPort: {{ $portMap.targetPort }}
      protocol: {{ coalesce $portMap.protocol "TCP" }}
{{- end }}
{{- end }}
```

Adding an option or overwrite within the Helm Chart under (for example) ```alloy:-> extraPorts: []``` (not by default, as default appProtocol is omitted) to add the AppProtocol would make it a lot easier for people who are trying to expose the service and connect to it from outside of the Kubernetes/OpenShift clusters on an Ingress or Route. More reasoning within the use case section.

          

### Use case

**Why appProtocol?**
We were trying to send traces from an alloy instance outside of the Kubernetes/OpenShift clusters to send traces over gRPC towards the Alloy instance on the Kubernetes/OpenShift clusters. We ran into many issues when trying to do so.

Adding an optional appProtocol would allow you to set "h2c" as an appProtocol option, allowing the service to be exposed via Ingress or Route within a Kubernetes/OpenShift Clusters. Once this Ingress/Route is exposed it would then allow gRPC communication for (for example) traces. Allowing an alloy instance on a VM or another Kubernetes/OpenShift clusters, to be able to communicate over gRPC via the exposed Ingress or Route. 

**Work Arround**
We currently add another service manually with an additional route in order to be able to perform the above use case of sending traces over gRPC via an exposed Ingress/Route on Kubernetes/OpenShift:

```
# Service
apiVersion: v1
kind: Service
metadata:
  name: service-otlp-alloy-h2c
spec:
  selector:
    app.kubernetes.io/instance: service-graph
    app.kubernetes.io/name: alloy
  ports:
    - name: otlp-grpc
      protocol: TCP
      appProtocol: h2c
      port: 4317
      targetPort: 4317
---
# Route
kind: Route
apiVersion: route.openshift.io/v1
metadata:
  name: route-otlp-alloy-h2c
spec:
  to:
    kind: Service
    name: test-grpc-h2c
    weight: 100
  port:
    targetPort: otlp-grpc
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Redirect
  wildcardPolicy: None
  ```

We currently use Kustomize + ArgoCD to deploy the helm charts, so another option would be to use Kustomize to Patch the inline result from the ```kustomize build . --enable-helm``` results. But we obviously would prefer if there would be an appProtocol option. That would be amazing, so we could set the appProtocol option on the ports we need the h2c settings in order for gRPC to work via route/ingress. 

If you have any questions, let me know.

PS: Also discussed this with @clementduveau before, therefore tagging him here as well.
PS2: Worked on this with colleague @maikel-09

### PR DESCRIPTION
**Pull request by clementduveau**
#### PR Description
On OpenShift, to expose Alloy gRPC to receive OTLP from other ressources, by default, you need to specifiy appProtocol = h2c on the service. It actually depends on your setup but the default Ingress HAProxy requires it, so probably 80% of RedHat customers needs it.

This PR just add the appProtocol as a possibility in the values.yaml. If not set, it outputs nothing. No default value because Kubernetes uses it as a "[hint](https://kubernetes.io/docs/concepts/services-networking/service/#:~:text=This%20is%20used%20as%20a%20hint%20for%20implementations%20to%20offer%20richer%20behavior)", therefore it's not mandatory.

Fixes #1405 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated --> I don't have an environment to test it, it should have no effect on vanilla K8s. We can ask @benoitschipper for tests on OpenShift

### DIFF
--- a/operations/helm/charts/alloy/templates/cluster_service.yaml
+++ b/operations/helm/charts/alloy/templates/cluster_service.yaml
@@ -29,5 +29,9 @@ spec:
       port: {{ $portMap.port }}
       targetPort: {{ $portMap.targetPort }}
       protocol: {{ coalesce $portMap.protocol "TCP" }}
+      # Useful for OpenShift clusters that want to expose Alloy ports externally
+      {{- if not (empty $portMap.appProtocol) }}
+      appProtocol: {{ $portMap.appProtocol }}
+      {{- end }}
   {{- end }}
 {{- end }}
--- a/operations/helm/charts/alloy/templates/cluster_service.yaml
+++ b/operations/helm/charts/alloy/templates/cluster_service.yaml
@@ -29,8 +29,8 @@ spec:
       port: {{ $portMap.port }}
       targetPort: {{ $portMap.targetPort }}
       protocol: {{ coalesce $portMap.protocol "TCP" }}
-      # Useful for OpenShift clusters that want to expose Alloy ports externally
       {{- if not (empty $portMap.appProtocol) }}
+      # Useful for OpenShift clusters that want to expose Alloy ports externally
       appProtocol: {{ $portMap.appProtocol }}
       {{- end }}
   {{- end }}
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -13,6 +13,7 @@ Unreleased
 ### Enhancements
 
 - Add the ability to set --cluster.name in the Helm chart with alloy.clustering.name. (@petewall)
+- Add the ability to set appProtocol in extraPorts to help OpenShift users to expose gRPC. (@clementduveau)
 
 0.6.0 (2024-08-05)
 ------------------
--- a/operations/helm/charts/alloy/README.md
+++ b/operations/helm/charts/alloy/README.md
@@ -179,6 +179,7 @@ Port numbers specified must be 0 < x < 65535.
 | hostPort | hostPort | (Optional) Number of port to expose on the host. Daemonsets taking traffic might find this useful. |
 | name | name | If specified, this must be an `IANA_SVC_NAME` and unique within the pod. Each named port in a pod must have a unique name. Name for the port that can be referred to by services.
 | protocol | protocol | Must be UDP, TCP, or SCTP. Defaults to "TCP". |
+| appProtocol | appProtocol | Hint on application protocol. This is used to expose Alloy externally on OpenShift clusters using "h2c". Optional. No default value. |
 
 ### alloy.listenAddr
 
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by benoitschipper** [comment]
@clementduveau Looking great! LGTM! =D

**Review by mattdurham — APPROVED** [review]
This looks reasonable to me, will defer to @petewall for final confirmation.

File: operations/helm/charts/alloy/templates/cluster_service.yaml [hunk 1]
```diff
@@ -29,5 +29,9 @@ spec:
       port: {{ $portMap.port }}
       targetPort: {{ $portMap.targetPort }}
       protocol: {{ coalesce $portMap.protocol "TCP" }}
+      {{- if not (empty $portMap.appProtocol) }}
```
petewall, line 32: Could this be simplified to:
```suggestion
      {{- if $portMap.appProtocol }}
```

petewall (approved):

operations/helm/charts/alloy/templates/cluster_service.yaml
benoitschipper: Would make it easier to read. But isn't  {{- if not (empty $portMap.appProtocol) }} better if you want to ensure that $portMap.appProtocol is not only present but also non-empty? 

If $portMap.appProtocol is undefined or empty, the first version (empty) will prevent any unexpected issues by skipping the block, while the second version ($portMap.appProtocol) might run into problems if the variable is not properly initialized.

Chances are slim, so I am (in the end) happy with both ❤️

PR #1502, opened 2024-08-19
Add Secret resource into Helm chart to allow passing sensitive values to config via environment variables

TheoBrigitte:
#### PR Description

This PR introduces a new Secret resources into the Helm chart which is used to store sensitive values as environment variables and populate them into the Alloy container to be read within the Alloy config using the `env` directive.

This allow for sensitive values to be stored in a Secret rather than in the ConfigMap or the Pod definition.

#### Notes to the Reviewer

I did my best to add some meaningful examples on how those secret values are defined and used from the config, but I am open to suggestions.

I would also appreciate some guidance on how to update the Helm chart tests.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Main (unreleased)
   seen by each instance in the cluster. This can help diagnose cluster split
   brain issues. (@thampiotr)
 
+- Added a new Secret resource into the helm chart to allow passing values into
+  config via environement variables.
+
 ### Bugfixes
 
 - Fixed an issue which caused loss of context data in Faro exception. (@codecapitano)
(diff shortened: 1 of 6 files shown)


clayton-cornell:
ping @grafana/grafana-agent-maintainers for a review.  No input for doc in the PR in it's current state.

wildum:
@petewall pinging you on this PR because it's about the helm chart

petewall (commented):
Personally, I don't know if I see the value for this, since this chart already has a `envFrom`, which lets you already do this:
```
envFrom:
- secretRef:
  name: mySecret
```
The only difference is that this chart would create a secret for you, which, I guess is helpful. It feels like this chart doing too much, IMO.

PR #1504, opened 2024-08-19
Clarify the need for capture groups in stage.replace

ptodev:
This has been a source of confusion. Also, Alloy's docs are not as clear as [Promtail's](https://grafana.com/docs/loki/latest/send-data/promtail/stages/replace/).

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1032,21 +1032,23 @@ The following arguments are supported:
 
 | Name         | Type     | Description                                                     | Default | Required |
 | ------------ | -------- | --------------------------------------------------------------- | ------- | -------- |
-| `expression` | `string` | Name from extracted data to use for the log entry.              |         | yes      |
+| `expression` | `string` | A RE2 regular expression containing capture groups.            |         | yes      |
 | `source`     | `string` | Source of the data to parse. If empty, it uses the log message. |         | no       |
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1032,21 +1032,23 @@ The following arguments are supported:
...
-| `expression` | `string` | Name from extracted data to use for the log entry.              |         | yes      |
+| `expression` | `string` | A RE2 regular expression containing capture groups.            |         | yes      |
 | `source`     | `string` | Source of the data to parse. If empty, it uses the log message. |         | no       |
 | `replace`    | `string` | Value replaced by the capture group.                            |         | no       |
 
+Each capture group and named capture group in `expression` will be replaced with the value given in `replace`.
+
+`expression` must contain valid RE2 regular expression capture groups.
+You can also name some of the groups using syntax such as `(?P<name>re)`.
+If any of the capture groups are named, their values will be set into the shared extracted map under the name of the regex group.
```
clayton-cornell, line 1043:
```suggestion
Each capture group and named capture group in `expression` is replaced with the value given in `replace`.

`expression` must contain valid RE2 regular expression capture groups.
You can also name some groups using syntax such as `(?P<name>re)`.
If any of the capture groups are named, their values will be set into the shared extracted map under the name of the regular expression group.
```

clayton-cornell (approved):

PR #1518, opened 2024-08-22
Fix windows tests

thampiotr:
#### PR Description

Since https://github.com/grafana/alloy/commit/4d79f3ecb0f6cccde4125b673219e087feeada4f we had windows tests failing due to handling of default value in test converters. I was not able to find a solution that would avoid changing the tests, but after investigating it seems that default values should be handled correctly in Alloy.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


mattdurham (approved):
Looks reasonable.

Allzz77123 (approved):

PR #1527, opened 2024-08-23
Upgrade windows_exporter to 0.27.2

jkroepke:
#### PR Description

This PR update the windows_exporter dependency to 0.27.

We try to deliver smaller version now, that the reason why there is an version again.

The 0.27 includes some breaking changes in terms of the config API. 

Before, most of the struct fields where strings. Now they are specific types, e.g. slice of strings or regex.Regex, if we requesting a regex.Regex value.

#### Notes to the Reviewer

I will do some integration test on a windows server. I raised the PR, because doc changes can be review.

I'm very unhappy using `regexp.MustCompile` at the `ToWindowsExporterConfig`, because i can result into a panic. But I have no idea where I can put my error.

I need some assistance here.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/static/integrations/windows_exporter/config_windows.go
```diff
@@ -1,54 +1,53 @@
...
 
 import (
+	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/prometheus-community/windows_exporter/pkg/collector"
 )
 
 func (c *Config) ToWindowsExporterConfig() collector.Config {
```
ptodev, line 11: This function could return a `(collector.Config, error)`. The functions which call `ToWindowsExporterConfig` will need to be changed to take the error value into account. Then we can use `Compile` instead of `MustCompile`, and return an error if there is a problem. The code may become hard to read with too many `if err != nil` statements though. To avoid the repetition, maybe you can try something like this:
```go
regexes := []struct {
	regexStr string
	out      **regexp.Regexp
}{
	{
		coalesceString(c.IIS.SiteInclude, c.IIS.SiteWhiteList),
		&cfg.IIS.SiteInclude,
	},
}
for i := range regexes {
	regex, err := regexp.Compile(coalesceString(c.IIS.SiteInclude, c.IIS.SiteWhiteList))
	if err != nil {
		return nil, err
	}
	*(regexes[i].out) = regex
}
```

I wonder if anyone else from @grafana/grafana-alloy-maintainers has other ideas?

I do think we should keep using `string` in the Alloy codebase. I'd prefer not to change the config types to the native regex ones, since we generally try to use just the [basic Alloy types](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/) as much as possible.

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
```diff
@@ -198,9 +198,9 @@ The v2 collector can query service states much more efficiently, but can't provi
 
 ### smb block
 
-Name           | Type     | Description                                | Default | Required
----------------|----------|--------------------------------------------|---------|---------
-`enabled_list` | `string` | Comma-separated list of collectors to use. | `""`    | no
+Name           | Type     | Description                                                    | Default | Required
+---------------|----------|----------------------------------------------------------------|---------|---------
+`enabled_list` | `string` | Deprecated (no-op), Comma-separated list of collectors to use. | `""`    | no
```
mattdurham, line 203: Is there a replacement for this?

internal/static/integrations/windows_exporter/config_windows.go
mattdurham: I agree with returning an error here, MustCompile throws a panic, and on something like a reload that's not great. As far as type changing I dont think any of the alloy underlying types are changing, but the upstream got more specific.

mattdurham (changes requested):
Overall seems reasonable, two main concerns: The MustCompile panics and we really want to avoid that, and the deprecrated feature, does it have a replacement.

jkroepke:
>  and the deprecrated feature, does it have a replacement.

This is correct. The initial author of that collector implement the same sub-collector concept, as we have for exchange. However there is only one collector (for smb and smb_client). 

Maintain such an setting for a single collector seems pointless to me. In conclusion, the sub-collector is just always enabled.

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
jkroepke: The initial author of that collector implement the same sub-collector concept, as we have for exchange collector. 

However there was only one collector (for smb and smb_client). The new behavior is just to enable the single sub-collector by default.

mattdurham: Cool, can we add that to the doc and/or changelog?

internal/static/integrations/windows_exporter/config_windows.go
jkroepke: @mattdurham please check that proposal https://github.com/grafana/alloy/pull/1527/commits/d0ee40247b9d23b8490f6b2b7e78947f21f6e0c8

internal/static/integrations/windows_exporter/config_windows.go
```diff
@@ -1,55 +1,78 @@
...
 
-	cfg.ScheduledTask.TaskInclude = regexp.MustCompile(c.ScheduledTask.Include)
-	cfg.ScheduledTask.TaskExclude = regexp.MustCompile(c.ScheduledTask.Exclude)
+	cfg.ScheduledTask.TaskInclude, err = regexp.Compile(c.ScheduledTask.Include)
+	errs = append(errs, err)
+	cfg.ScheduledTask.TaskExclude, err = regexp.Compile(c.ScheduledTask.Exclude)
+	errs = append(errs, err)
 
-	return cfg
+	return cfg, errors.Join(errs...)
```
ptodev, line 75: That's cool - I didn't know there is a `Join` function :)
Personally I wouldn't return an object if there are errors (I'd probably return a nil instead), but I think it's ok if we do.

jkroepke: If all values in the slice are `nil`, then nil is returned. Otherwise, an error is returned.

https://github.com/golang/go/blob/7c54e024e892cbb8ed24274f0b22505356220f10/src/errors/join.go#L20-L28

jkroepke: The reason to choice join here was that if there are multiple errors, all errors returned instead the first one

internal/static/integrations/windows_exporter/windows_exporter_windows.go
```diff
@@ -29,11 +36,22 @@ func New(logger log.Logger, c *Config) (integrations.Integration, error) {
...
+			collector.NewPrometheus(4*time.Minute, &winCol, logger),
+		),
+		integrations.WithRunner(func(ctx context.Context) error {
+			<-ctx.Done()
+
+			// Stop the collector
+			err := winCol.Close()
+
+			return errors.Join(ctx.Err(), err)
+		}),
```
jkroepke, line 53: Could you please take a look here, if this is the right place to add an Close method?

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
jkroepke: I add something to the Changelog, could you check if sufficient?

mattdurham (approved):
Thank you! This looks good to me.

internal/static/integrations/windows_exporter/windows_exporter_windows.go
mattdurham: Yes, though there is possible a slight difference from a normal component. during a an Update operation this will be closed and reopened. So has long as this doesnt prevent reloads then we are fine.

https://github.com/grafana/alloy/blob/2ab9e0d1d40b7e25612b50fe06e2dfd1d580a339/internal/component/prometheus/exporter/exporter.go#L78

jkroepke: I’m planning to move away from using the global scope. It has been indicated that some handles remain open if they are used in this way. Instead, the handles should be opened in the initialization functions.

For the exporter as a standalone binary, this isn’t as critical. However, for binaries that embed the exporter, it's important to close handles to avoid handle leaks.

Allzz77123 (approved):

PR #1537, opened 2024-08-23
Update discovery.kubernetes selectors block description

TheoBrigitte:
This PR suggest a new description for the `selectors` block in the `discovery.kubernetes` component.

AFAIU this block filters the resources selected in the role argument and not the namespace as stated.


clayton-cornell (approved):
Thanks for this PR @TheoBrigitte   I'll get this one merged right away.

PR #1541, opened 2024-08-23
Document a new "join" block in discovery.relabel

ptodev:
#### PR Description

The main use case driving this is adding labels from `discovery.kubernetes` to targets coming from `prometheus.exporter` components. If you can also think of other use cases, please drop a comment in the PR.

Instead of writing a proposal for the `join` block, I decided to just write the documentation for it. If you agree that the proposed functionality is reasonable, I will update this PR with the code implementation of the new block. I could then take the PR out of the draft stage.

Fixes ##1443

#### Notes to the Reviewer

Initially I meant to solve this problem using [dynamic pipelines](https://github.com/grafana/alloy/pull/1480). Unfortunately, that proved to be very difficult. The approach in this PR would be easier to implement. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/discovery/discovery.relabel.md
+++ b/docs/sources/reference/components/discovery/discovery.relabel.md
@@ -52,15 +52,65 @@ Name      | Type                | Description        | Default | Required
 The following blocks are supported inside the definition of
 `discovery.relabel`:
 
-Hierarchy | Block    | Description                           | Required
-----------|----------|---------------------------------------|---------
-rule      | [rule][] | Relabeling rules to apply to targets. | no
+Hierarchy         | Block         | Description                                       | Required
+------------------|---------------|---------------------------------------------------|---------
+rule              | [rule][]      | Relabeling rules to apply to targets.             | no
+join              | [join][]      | Joined labels from other lists of targets.        | no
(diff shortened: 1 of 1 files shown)


ptodev:
@thampiotr The reason I didn't opt for a stdlib function is that the ergonomics might not be as good. This is because not all labels from the SD should be taken into account. Suppose you have these labels from k8s SD:
```
{"pod_ip" = "1.2.3.4", "namespace" = "a", "container" = "a"}
{"pod_ip" = "1.2.3.4", "namespace" = "a", "container" = "b"}
{"pod_ip" = "1.2.3.4", "namespace" = "a", "container" = "c"}
```

Let's say you want to join on the "pod_ip" label. How would you join the "container" labels?

The solution I decided to go with is to remove labels which are not in `labels_to_join` and `condition` prior to the join, and then to remove any duplicates. This is quite a specific operation... we could add stdlibs for each one of those steps, but it'd be awkward to have to setup all those stdlib functions manually in a config file.

ptodev:
@thampiotr I think you're right - probably an inner join is a better way to go. I opened #1826 for it, and I'm going to close this PR because it's unlikely to be the solution we go with. Thank you for the feedback!

PR #1548, opened 2024-08-26
Add community shared topic

clayton-cornell:
#### PR Description

This PR adds a new shared topic for Community contributions

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/shared/stability/community.md
+++ b/docs/sources/shared/stability/community.md
@@ -6,3 +6,6 @@ headless: true
 
 > **Community**: This component is developed, maintained, and supported by the Alloy user community.
 > Grafana doesn't offer commercial support for this component.
+> To enable and use community components, you must set the `--feature.community-components.enabled` [flag][] to `true`.
+
+[flag]: https://grafana.com/docs/alloy/<ALLOY_VERSION>/reference/cli/run/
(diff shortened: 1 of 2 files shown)


docs/sources/shared/stability/community.md
```diff
@@ -0,0 +1,8 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/shared/stability/community/
+description: Shared content, community contributed
+headless: true
+---
+
+> **Community**: This component is developed, maintained, and supported by the Alloy user community.
+> Grafana doesn't offer commercial support for this component.
```
wildum, line 8: Could you add that to use it the user must set the flag "--feature.community-components.enabled" to true with the run command?

wildum (commented):
Could you also update this line "The documentation should also follow the same pattern as the core components but at a different [location][cc dir]." in this file: https://github.com/grafana/alloy/blob/main/docs/developer/adding-community-components.md please. Instead the users should add this new block at the top of their doc

clayton-cornell:
Removed the backports to 1.0, 1.1, and 1.2. These backports fail because the community contributions topic in the developer docs isn't included in those versions.

Issue #305, opened 2023-11-17
Error parsing syslog stream UDP


rbenea:
### What's wrong?

I have updated to grafana-agent-flow-0.38.0-rc.0-1.amd64.deb in order to have syslog via UDP working. This was fixed in https://github.com/grafana/agent/issues/5197

For the moment we are sending syslog messages from PFSense, Mikrotik and Unifi devices.

From Mikrotik and Unify devices we get the message in journald:
`msg="error parsing syslog stream" component=loki.source.syslog.syslog err="expecting a version value in the range 1-999 [col 4]"`

From PFSense we get the message:
`msg="error parsing syslog stream" component=loki.source.syslog.syslog err="expecting a version value in the range 1-999 [col 5]`

### Steps to reproduce

Install grafana-agent-flow-0.38.0-rc.0-1.amd64.deb on a Debian12 VM with the config below and send remote syslog messages to the agent via UDP.

### System information

Debian 12.2 x86_64 Linux

### Software version

Grafana Agent 0.38.0-rc.0

### Configuration

```text
loki.relabel "syslog" {
  forward_to = []

  rule {
    source_labels = ["__syslog_message_hostname"]
    target_label  = "host"
  }
}
loki.source.syslog "syslog" {
  listener {
    address  = "0.0.0.0:5514"
    protocol = "tcp"
    labels   = { component = "loki.source.syslog", protocol = "tcp"}
  }

  listener {
    address  = "0.0.0.0:5514"
    protocol = "udp"
    labels   = { component = "loki.source.syslog", protocol = "udp"}
    use_rfc5424_message = true
  }

  relabel_rules = loki.relabel.syslog.rules
  forward_to = [loki.write.local.receiver]
}

loki.write "local" {
    endpoint {
        url = "http://127.0.0.1:3100/loki/api/v1/push"
    }
    external_labels = {}
}
```

### Logs

```text
Nov 17 15:26:07 logserver grafana-agent-flow[1308]: ts=2023-11-17T13:26:07.294962163Z level=warn msg="error parsing syslog stream" component=loki.source.syslog.syslog err="expecting a version value in the range 1-999 [col 4]"
Nov 17 15:30:08 logserver grafana-agent-flow[1308]: ts=2023-11-17T13:30:08.800809677Z level=warn msg="error parsing syslog stream" component=loki.source.syslog.syslog err="expecting a version value in the range 1-999 [col 5]"
```

rarrr:
[...]

PR #1556, opened 2024-08-27
Support RFC3164 syslog entries

sushain97:
#### PR Description

Ports the changes from https://github.com/grafana/loki/pull/12810 here.

Fixes https://github.com/grafana/alloy/issues/305.

#### Notes to the Reviewer

Suggest reviewing commit-by-commit.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -8,9 +8,10 @@ title: loki.source.syslog
 
 # loki.source.syslog
 
-`loki.source.syslog` listens for syslog messages over TCP or UDP connections
-and forwards them to other `loki.*` components. The messages must be compliant
-with the [RFC5424](https://www.rfc-editor.org/rfc/rfc5424) format.
+`loki.source.syslog` listens for syslog messages over TCP or UDP connections and
+forwards them to other `loki.*` components. The messages must be compliant with
+the [RFC5424](https://www.rfc-editor.org/rfc/rfc5424) format or the
+[RFC3164](https://datatracker.ietf.org/doc/html/rfc3164) ("BSD syslog") format.
```
clayton-cornell, line 14:
```suggestion
the [RFC5424](https://www.rfc-editor.org/rfc/rfc5424) syslog protocol or the
[RFC3164](https://datatracker.ietf.org/doc/html/rfc3164) BSD syslog protocol.
```

We can name these RFCs as they are titled in the referenced source.

sushain97:
@clayton-cornell it looks like there's one test failing (really timing out):

```
FAIL	github.com/grafana/alloy/internal/component/loki/process	600.159s
```

I doubt that it's related to this PR per se but it could be from the bump to `github.com/grafana/loki/v3` ?

clayton-cornell:
> it could be from the bump to `github.com/grafana/loki/v3` ?

Possible. The macos test fails once in a while. I'll leave it as-is for now.

First we need a code review from someone in @grafana/grafana-alloy-maintainers.

internal/component/loki/source/syslog/types.go
```diff
@@ -42,6 +43,15 @@ func (sc *ListenerConfig) Validate() error {
 		return fmt.Errorf("syslog listener protocol should be either 'tcp' or 'udp', got %s", sc.ListenProtocol)
 	}
 
+	if sc.SyslogFormat != "" {
```
wildum, line 46: the default value should be set above in DefaultListenerConfig. Then you don't need to check for empty string, you can just check for `sc.SyslogFormat != scrapeconfig.SyslogFormatRFC3164 && sc.SyslogFormat != scrapeconfig.SyslogFormatRFC5424`

internal/component/loki/source/syslog/types.go
```diff
@@ -42,6 +43,15 @@ func (sc *ListenerConfig) Validate() error {
...
+		if sc.SyslogFormat != scrapeconfig.SyslogFormatRFC3164 && sc.SyslogFormat != scrapeconfig.SyslogFormatRFC5424 {
```
wildum, line 47: it would be nice to add a unit test for this part

wildum (commented):
LGTM, only two small comments for the Alloy part

go.mod
```diff
@@ -63,8 +63,8 @@ require (
 	github.com/grafana/jfr-parser/pprof v0.0.0-20240126072739-986e71dc0361
 	github.com/grafana/jsonparser v0.0.0-20240209175146-098958973a2d
 	github.com/grafana/kafka_exporter v0.0.0-20240409084445-5e3488ad9f9a
-	github.com/grafana/loki/pkg/push v0.0.0-20240514112848-a1b1eeb09583 // k201 branch
-	github.com/grafana/loki/v3 v3.0.0-20240513110952-8622293f23b1 // k201 branch
+	github.com/grafana/loki/pkg/push v0.0.0-20240617182007-6c33561108ad // k201 branch
+	github.com/grafana/loki/v3 v3.0.0-20240617182007-6c33561108ad // k201 branch
```
ptodev, line 67: Which `k` branch does the commit correspond to? It'd be nice to update the comment.

internal/component/loki/source/syslog/types.go
ptodev, line 47: Do you mind creating local RFC strings please?
```go
const (
	// A modern Syslog RFC
	SyslogFormatRFC5424 = "rfc5424"
	// A legacy Syslog RFC also known as BSD-syslog
	SyslogFormatRFC3164 = "rfc3164"
)
```

It would be good not to depend on Promtail for them, in case they get changed in the Promtail code.

go.mod
sushain97: k206, I thought I updated it but it got lost somehow

sushain97:
@wildum @ptodev Thanks for taking a look! I think I've addressed the comments.

internal/component/loki/source/syslog/types.go
```diff
@@ -22,13 +29,15 @@ type ListenerConfig struct {
...
 	TLSConfig            config.TLSConfig  `alloy:"tls_config,block,optional"`
+	SyslogFormat         string            `alloy:"syslog_format,attr,optional"`
 }
 
 // DefaultListenerConfig provides the default arguments for a syslog listener.
 var DefaultListenerConfig = ListenerConfig{
 	ListenProtocol:   st.DefaultProtocol,
 	IdleTimeout:      st.DefaultIdleTimeout,
 	MaxMessageLength: st.DefaultMaxMessageLength,
+	SyslogFormat:     SyslogFormatRFC5424,
```
sushain97, line 40: This can't be `st.DefaultSyslogFormat` since that would introduce a cyclic dependency unless we moved the constants out of this package. But that feels odd.

I'm not sure why the defaults don't live here, TBH.

internal/component/loki/source/syslog/types.go
```diff
@@ -52,6 +68,15 @@ func (sc ListenerConfig) Convert() *scrapeconfig.SyslogTargetConfig {
 		lbls[model.LabelName(k)] = model.LabelValue(v)
 	}
 
+	var syslogFormat scrapeconfig.SyslogFormat
+	if sc.SyslogFormat == SyslogFormatRFC3164 {
+		syslogFormat = scrapeconfig.SyslogFormatRFC3164
+	} else if sc.SyslogFormat == SyslogFormatRFC5424 {
+		syslogFormat = scrapeconfig.SyslogFormatRFC5424
+	} else {
+		panic("unknown syslog format " + sc.SyslogFormat)
```
sushain97, line 77: This can only happen if validation fails but this is called anyway...

ptodev: I think it's ok not to do a check for the values, and instead to assume that `sc` is an already well-formatted Alloy config. In the struct below, you could just do:
```go
scrapeconfig.SyslogFormat(sc.SyslogFormat)
```

sushain97: Then we're basically relying on the promtail strings to match and might as well use them directly?

internal/component/loki/source/syslog/types.go
sushain97: I mean that the default constants like `DefaultProtocol` don't live in this package.

ptodev: I see, thanks for pointing that out. I haven't worked with that code a lot and didn't realise that this is where the constants usually live. It's ok if you also put the RFC constants in `loki/source/syslog/internal/syslogtarget`. 

The reason why I initially suggested that you redeclare them is that I mistakenly thought the constants are imported from the Loki repo. If that was the case, and if Loki change their values, then our docs could go out of date by accident. Having the constant definitions in the `syslogtarget` package in the Alloy repo is ok. Apologies for the confusion!

sushain97: No worries! You're right though!

There are three approaches:
1. Re-use the RFC constants from `github.com/grafana/loki/v3/clients/pkg/promtail/scrapeconfig`. This was my initial approach.
2. Copy the RFC constants into `internal/component/loki/source/syslog`. This is the current approach.
3. Copy the RFC constants into `internal/component/loki/source/syslog/internal/syslogtarget`.

Both options (2) and (3) require converting our constants into `scrapeconfig.SyslogFormat`. If we don't check each of our constants against the constants in `scrapeconfig`, then I'm not sure there's a lot of value coming from options 2/3 vs option 1. Since not checking them means that if Loki changes, we have to add back the conversion or change our config (backwards incompatible so we're just going to go with the former option). So, might as well add the conversation code now or go with option 1. I guess we could allow creating an invalid `scrapeconfig.SyslogTargetConfig` and assume that it would fail later but that feels worse?

I'm OK with any of the options! Just let me know what y'all would prefer.

ptodev: I would be happy with option 1. Thank you!

sushain97: Swapped!

sushain97:
@ptodev thanks for triggering tests! I just pushed a fix for the failing one.

sushain97:
@ptodev is this good to merge?

internal/component/loki/source/syslog/types.go
ptodev: Apologies, I gave you the wrong advise by going with option 1 🤦‍♂️ 

My main concern is that values of const strings in the Loki repo could change. If that happens, the Alloy docs will be out of date and user's configs may no longer work. So it'd be good to copy the strings into the Alloy repo. I think option 2 would have been fine.

I pushed a few commits to your branch, and rebased it to avoid causing you extra work. I hope you don't mind. Thank you for the help, and apologies for the confusion and the back and forth!

sushain97: No worries! Feel free to modify as much as you'd like :)

ptodev:
Hi @wildum, would you mind reviewing the PR again please? It should be ready to merge, but I pushed a few changes to it and it'd be nice to have some else double check them :)

sboschman:
Would it be easy to add 'raw' as format, apart from 'RFC3164' and 'RFC5424' formats? So, the parsing of the message can be done in a `loki.process` component.

I have firewall devices sending syslog messages in [CEF](https://help.deepsecurity.trendmicro.com/20_0/on-premise/event-syslog-message-formats.html) format.

sushain97:
@sboschman I'd suggest creating an issue for that, it's probably best as a separate PR.

sushain97:
@wildum @ptodev please let me know if I should resolve the conflicts with master. Don't want to resolve them just to see it get conflicted again 😅

sushain97:
@ptodev @wildum 👋 Just checking in again on this. I'm happy to resolve the conflicts if someone can merge it afterwards.

wildum (approved):
Sorry for the delay, I missed the notifications. LGTM, I will merge once the conflicts are resolved

sushain97:
@wildum no worries, thanks for taking a look! I've merged master to eliminate the conflicts.

wildum (approved):

sushain97:
There's one failing test but it doesn't look related?

```
--- FAIL: TestAdditionalLabels (3.01s)
    events_test.go:276: 
        	Error Trace:	/drone/src/internal/component/mimir/rules/kubernetes/events_test.go:276
        	            				/usr/local/go/src/runtime/asm_amd64.s:1695
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestAdditionalLabels
    events_test.go:272: 
        	Error Trace:	/drone/src/internal/component/mimir/rules/kubernetes/events_test.go:272
        	Error:      	Condition never satisfied
        	Test:       	TestAdditionalLabels
level=info ts=2024-09-27T07:54:17.90219047Z msg="shutting down event loop"
FAIL
FAIL	github.com/grafana/alloy/internal/component/mimir/rules/kubernetes	3.360s
```

wildum:
yeah seems like a flaky test, restarting the build

PR #1557, opened 2024-08-27
feat: Configurable cgroup ID regex for process discovery component

### LINKED ISSUE #1461: Configurable cgroup ID regex for process discovery component
### Request

Currently, [process discovery](https://grafana.com/docs/alloy/latest/reference/components/discovery/discovery.process/) component attempts to get the container ID by [looking into the cgroup of the process](https://github.com/grafana/alloy/blob/c6fdf83be68d738043f73bed35395fad741e04fa/internal/component/discovery/process/container.go#L14-L30). This can be further generalized by adding a new argument that takes a regex as input and attempts to find the cgroup ID by matching it against the cgroup of the process. Consequently we can add a [new label](https://github.com/grafana/alloy/blob/c6fdf83be68d738043f73bed35395fad741e04fa/internal/component/discovery/process/discover.go#L20-L28), say, `__meta_cgroup_id__`, that will be added to the targets.

### Use case

Most of the resource managers use cgroups to manage the resource allocated to compute workloads. In our particular case, it is SLURM (HPC batch scheduler). By using a configurable cgroup ID regex to process discovery component, we can find the job IDs of each process. And by using relabel magic, we can filter the processes that do not belong to any user jobs and eventually use the job ID as [`service_name`](https://grafana.com/docs/alloy/latest/reference/components/pyroscope/pyroscope.ebpf/#service-name) to Pyroscope eBPF component. This will allow us to do continous profiling of user jobs and aggregate the profiles of each job on Grafana based on `service_name` (which is essentially job ID).

This should work for any resource manager which manages the cgroups in a deterministic way. For instance, one more use case can be to use with Openstack where libvirt manages the cgroups. Grafana alloy can be deployed directly on the hypervisor that will do the continous profiling of the VMs.

If the maintainers find a value in this feature, I would be happy to submit a PR.

mahendrapaipuri:
#### PR Description

This PR adds a new regex argument to Process discovery component that captures the cgroup ID of the process and exports it in the target. 

Fixes #1461

#### Notes to the Reviewer
@simonswine Please take a look at it when you have time.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -87,6 +88,11 @@ The resulting targets are:
 ]
 ```
 
+### Cgroup ID regex
+
+Argument `cgroup_id_regex` allows to capture the cgroup ID of the process managed by other resource managers like
+SLURM, Libvirt, _etc._ The capture cgroup ID can be exported as `__meta_process_cgroup_id` field when enabled.
```
clayton-cornell, line 94:
```suggestion
The `cgroup_id_regex` argument allows you to capture the cgroup ID of the process managed by other resource managers like SLURM or Libvirt.
When it is enabled, you can export the cgroup ID as the `__meta_process_cgroup_id` field.
```

clayton-cornell (approved):
Docs are OK. Over to @grafana/grafana-alloy-maintainers for a code review.

docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -29,6 +29,7 @@ The following arguments are supported:
 | Name               | Type                | Description                                                                              | Default | Required |
 |--------------------|---------------------|------------------------------------------------------------------------------------------|---------|----------|
 | `join`             | `list(map(string))` | Join external targets to discovered processes targets based on `__container_id__` label. |         | no       |
+| `cgroup_id_regex`  | `string`            | Regex expression to retrieve cgroup ID of the process.                                   |         | no       |
```
jdbaldry, line 32: https://developers.google.com/style/word-list#regex

I think we should also document that the regular expression needs to have exactly one capture group which is what is used to extract the ID.

```suggestion
| `cgroup_id_regex`  | `string`            | Regular expression to retrieve cgroup ID of the process.                                   |         | no       |
```

mahendrapaipuri: Yes, agree about documenting about capture groups. Modified the docs as suggested. Cheers!!

internal/component/discovery/process/discover.go
```diff
@@ -167,3 +180,17 @@ func getLinuxProcessContainerID(pid string) (string, error) {
 	}
 	return "", nil
 }
+
+func getLinuxProcessCgroupID(pid string, regexp *regexp.Regexp) (string, error) {
+	if runtime.GOOS == "linux" {
```
mattdurham, line 185: Is this needed?  With the go build linux flag at the top of the file.

internal/component/discovery/process/process.go
```diff
@@ -44,16 +46,25 @@ type Component struct {
 }
 
 func (c *Component) Run(ctx context.Context) error {
-	doDiscover := func() error {
-		processes, err := discover(c.l, &c.args.DiscoverConfig)
+	var cgroupIDRegexp *regexp.Regexp
+	var err error
+	if c.args.CgroupIDRegex != "" {
```
mattdurham, line 51: I dont see where updates to the regex are handled in the Update function.

mattdurham (changes requested):
Minor nit and how to handle updates.

internal/component/discovery/process/discover.go
mahendrapaipuri: Makes sense. Will remove it.

internal/component/discovery/process/process.go
mahendrapaipuri: You are right. Currently we are not handling regex updates.

I also think it makes more sense to move the regex compilation to `New` function so that we fail fast when provided regular expression is not valid. What do you think?

Question: Can we return error during updates? As we need to verify the regular expression, can we return error for `changed` method when an invalid regular expression is passed during an update?

mattdurham: New is fine but we still need to handle updating the regex, we can return an error with update.

mahendrapaipuri: I pushed the changes to ensure we update regex as well. Let me know what do you think. Cheers

internal/component/discovery/process/cgroup_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenericCGroupMatching(t *testing.T) {
```
mattdurham, line 14: Can we add a test for the update regex functionality? Looks like its coming along nicely!

mahendrapaipuri: I pushed couple of unit tests. Could you check them and let me know if that is what you mean? Cheers

internal/component/discovery/process/process.go
```diff
@@ -78,6 +91,15 @@ func (c *Component) Run(ctx context.Context) error {
 func (c *Component) Update(args component.Arguments) error {
 	a := args.(Arguments)
 	c.argsUpdates <- a
+
```
mattdurham, line 94: I feel we may need mutexes around the regex. It can be accessed  from both dodiscover and Update.

internal/component/discovery/process/cgroup_test.go
```diff
@@ -0,0 +1,91 @@
...
+		err = tc.Run(componenttest.TestContext(t), args)
+		require.NoError(t, err)
+	}()
+
+	// Sleep a short time for component to go into run state
+	time.Sleep(100 * time.Millisecond)
+
+	newArgs := args
+	newArgs.CgroupIDRegex = "^.*/docker/([a-z0-9]+)(?:.*$)"
+	require.NoError(t, tc.Update(newArgs))
```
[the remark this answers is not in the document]
mahendrapaipuri: Seems like `componenttest.Controller` do not have any public methods that will allow us to check the updated args. Maybe I am wrong. Do you have any leads on how we can verify the updated regex?

docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -29,6 +29,7 @@ The following arguments are supported:
 | Name               | Type                | Description                                                                              | Default | Required |
 |--------------------|---------------------|------------------------------------------------------------------------------------------|---------|----------|
 | `join`             | `list(map(string))` | Join external targets to discovered processes targets based on `__container_id__` label. |         | no       |
+| `cgroup_id_regex`  | `string`            | Regular expression that captures only one group that will used as cgroup ID of process.  |         | no       |
```
korniltsev, line 32: It may be a dumb question, but is it not possible to use `discovery.relabel` instead of introducing `cgroup_id_regex`  ?

mahendrapaipuri: Ahh I see what you mean but even to use relabel we need `process` component exporting cgroup path field in the targets. Because we need a cgroup ID which maps to a compute unit (pod, VM or batch job) of resource manager. which is usually in the cgroup path.

But I guess your suggestion opens a new direction. We can simply export cgroup path as one of the target fields and end users can use `relabel` to retrieve cgroup ID from the path. This will simplify error handling as well as there is no regex compiling involved and so, no need of updating and any mutexes. 

@mattdurham What do you think?

mahendrapaipuri: @mattdurham Just checking in to see where are we with this PR.

internal/component/discovery/process/process.go
mattdurham: The below lines still feel vulnerable to race conditions.

mahendrapaipuri: Sorry @mattdurham I wasnt clear enough.

I was more referring to this [comment](https://github.com/grafana/alloy/pull/1557#discussion_r1759340960) which gives a [new direction](https://github.com/grafana/alloy/pull/1557#discussion_r1759753505) to simplify this feature further.

To put it simply, we export **raw cgroup path** like `0::/system.slice/slurmstepd.scope/job_1446354/step_batch/user/task_0` as one of the fields of target, say `__meta_process_cgroup_path`, and we leave it to end user to retrieve correct cgroup ID using relabel component. This way we dont have to compile regexes, do error handling and worry about race conditions. We can add this "new" field with minimal changes to the existing component. If you think it is a good idea, I can refactor the PR. If you think it is better to stick to current approach, I can fix the race issues and update the PR.

Cheers!!

docs/sources/reference/components/discovery/discovery.process.md
ixzzd: >  We can simply export cgroup path as one of the target fields and end users can use relabel to retrieve cgroup ID from the path.

This would be very helpful. I would like to set service_name based on the systemd service name (which could be determined by cgroup). Looking forward have this PR merged

mahendrapaipuri:
@mattdurham I pushed the changes I proposed in the [previous comment](https://github.com/grafana/alloy/pull/1557#discussion_r1782828988). 

Simply put, we export cgroup path as one of the labels. In the case of cgroups v2, it is a single path as there is only one unified cgroup path for a given process. In the case of cgroups v1, there is a path for each controller and "relevant" path depends on which controllers are active. So, in order to keep the component simple, we just export all the cgroup paths delimited by `|` as a string. 

Users will be able to use `relabel` component to retrieve the information they need from these paths very easily. At the same time, there is no need to add new arguments and keep the component simple as before.

Let me know what do you think. 

@korniltsev Could you take a look at it and see if it what you mean?

Cheers!

internal/component/discovery/process/args.go
```diff
@@ -29,6 +30,7 @@ var DefaultConfig = Arguments{
 		Exe:         true,
 		Commandline: true,
 		ContainerID: true,
+		CgroupPath:  true,
```
simonswine, line 33: I would argue that we should disable it by default. As without further config it is not a useful metadata.

When you really need it then you can also set this to `true`, when adding your `disocvery.relabel`.

Wdyt?

docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -157,6 +159,7 @@ discovery.process "all" {
     commandline = true
     username = true
     uid = true
+    cgroup_path = true
     container_id = true
   }
 }
```
simonswine, line 165: A separate example showing how to parse the cgroup in a `discovery.relabel` would be nice.

Not fully understand your usecase, but it looks like you are doing something with qemu vms, maybe relevant/helpful for someone else.

simonswine (commented):
@mahendrapaipuri thank you for your contribution and sorry for a rather slow review from our end. Promise to be more on top of it, with the next/final iteration.

I happy how it all looks, just some minor requests 🙂

internal/component/discovery/process/args.go
mahendrapaipuri: Fair enough. Makes sense!

docs/sources/reference/components/discovery/discovery.process.md
mahendrapaipuri: Yes, I agree. It would be very useful to have an example.

Well, our use case is to profile HPC batch jobs. We can retrieve job IDs using the `cgroup` name and use this job ID as `service_name` to aggregate the profiles of all processes running under a job.

mahendrapaipuri:
@simonswine Thanks a lot for coming back to this PR. Totally understandable on why things take long time and its fine.

So, I added an example on how to use the new cgroup path meta data with `relabel` component. Our use case of HPC batch jobs might be too specific for generic users. So, I chose an example of discovering `systemd` services using `cgroup` path and `relabel` component. 

Let me know if it is clear for you and what do you think of example. If you think adding a "complete" config example on how to add `service_name` label and then pass them to Pyroscope's `ebpf` component to continuously profile `systemd` services adds more value, I can do that too. Cheers!!

docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -111,6 +111,7 @@ The following arguments are supported:
 | `commandline`  | `bool` | A flag to enable discovering `__meta_process_commandline` label. | true    | no       |
 | `uid`          | `bool` | A flag to enable discovering `__meta_process_uid`: label.        | true    | no       |
 | `username`     | `bool` | A flag to enable discovering `__meta_process_username`: label.   | true    | no       |
+| `cgroup_path`  | `bool` | A flag to enable discovering `__meta_cgroup_path__` label.       | true    | no       |
```
simonswine, line 114: This is still inconsistent:

```suggestion
| `cgroup_path`  | `bool` | A flag to enable discovering `__meta_cgroup_path__` label.       | false    | no       |
```

simonswine (approved):
Thanks for your contribution, sharing your use case and your patience with us.

See the suggestion on the parameter default value, once that is in, I will merge it.

mahendrapaipuri:
@simonswine Done!! Cheers for everyone's time in reviewing this as well.

docs/sources/reference/components/discovery/discovery.process.md
```diff
@@ -187,6 +190,34 @@ discovery.process "all" {
   }
 }
 
+### Example discovering processes on the local host based on `cgroups` path
+
+To discover processes running under systemd services on the local host.
```
clayton-cornell, line 195:
```suggestion
The following example configuration shows you how to discover processes running under systemd services on the local host.
```

We could expand this intro sentence a little (I know the previous example doesn't have this intro sentence... fixing things like this are on the list of things to improve in the docs).

clayton-cornell (commented):
The doc input looks OK. One small suggestion.

clayton-cornell (commented):
The doc input looks OK. One small suggestion.

clayton-cornell (commented):
The doc input looks OK. One small suggestion.

clayton-cornell (commented):
The doc input looks OK. One small suggestion.

(hmmm it posted the same comment 3x... internet glitch on my end I suspect)

PR #1564, opened 2024-08-27
Merge `prometheus.write.queue` into main.

### LINKED ISSUE #370: Durable Queue as a WAL in Flow
### Background

The current WAL implementation suffers from several downsides.

* OOM errors while replaying
* Replaying the WAL stops graph evaluation
* High memory usage
* Inability to re-send data between restarts
* Lack of expiration promises (when does a metric expire is a hard question to answer)
* Current WAL needs shims to work with the Flow concept

### Proposal

Create a Disk Based Queue WAL+remote_write component. With a focus on

* Limiting memory through no unbounded large maps
* Memory should be limited to what it is reading/writing at the time
* Flow native 
* Concrete expiration promises
* Replayable
* Performance on sending and writing samples within 90% of our current implementation
* Simplify the implementation
* Ability to make changes with scraping/wal within the Agent codebase more easily

With no need to cache the series themselves memory will no longer be dependent on the number series stored. Each commit from a scrape to the WAL creates a new record and the remote write side works its way through the queue bookmarking its position on each successful send. This gives us strong replayability guarantees. 

On startup instead of replaying the WAL it only needs to verify the correctness of the WAL. 

This is not meant to replace the current WAL implementation but be an experiment in using a queue based WAL.

mattdurham:
Ready for merge into main, there have only been minor changes to the name, changelog, and two tests that were flaky since the component pr.

Fixes #370


mattdurham:
Only significant change is changelog and renaming from prometheus.remote.queue to prometheus.write.queue.

clayton-cornell (approved):
No changes from the docs standpoint. LGTM

go.mod
```diff
@@ -809,12 +809,10 @@ require (
...
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
-
-)
-
-require (
+	github.com/gammazero/deque v0.2.1 // indirect
+	github.com/vladopajic/go-actor v0.9.0 // indirect
 	github.com/aws/aws-sdk-go-v2/service/iam v1.33.1 // indirect
-	github.com/checkpoint-restore/go-criu/v6 v6.3.0 // indirect
+    github.com/checkpoint-restore/go-criu/v6 v6.3.0 // indirect
```
dehaansa, line 815: nit: indentation

mattdurham: Tabs vs spaces! But fixed so its consistent. I was confused for a bit when opened it up in my editor.

thampiotr (approved):
Happy to go with it, dogfood and fix any issues we find. 
Would be great to link a tracking issue with follow-up work so we can prioritise.

PR #1591, opened 2024-08-29
Fix a flaky unit test in loki.process

ptodev:
There is a genuine bug in the code which is causing a test to be flaky. However, I'm not classifying this PR as a bugfix because in practice the bug shouldn't cause issues for Alloy users. The leak would only happen when `Run` completes, which can only happen once for the lifetime of the process - when the process exits. This fix was [ported](https://github.com/grafana/agent/pull/7004) to Agent as well.

### DIFF
--- a/internal/component/loki/process/process.go
+++ b/internal/component/loki/process/process.go
@@ -97,17 +97,24 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 // Run implements component.Component.
 func (c *Component) Run(ctx context.Context) error {
+	shutdownCh := make(chan struct{})
+	wgOut := &sync.WaitGroup{}
 	defer func() {
 		c.mut.RLock()
 		if c.entryHandler != nil {
 			c.entryHandler.Stop()
+			// Stop handleOut only after the entryHandler has stopped.
+			// If handleOut stops first, entryHandler might get stuck on a channel send.
+			close(shutdownCh)
+			wgOut.Wait()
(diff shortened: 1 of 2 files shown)


internal/component/loki/process/process.go
```diff
@@ -97,17 +97,24 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 // Run implements component.Component.
 func (c *Component) Run(ctx context.Context) error {
+	shutdownCh := make(chan struct{})
```
thampiotr, line 100:
```suggestion
	handleOutShutdown := make(chan struct{})
```

internal/component/loki/process/process.go
```diff
@@ -97,17 +97,24 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
 		if c.entryHandler != nil {
 			c.entryHandler.Stop()
+			// Stop handleOut only after the entryHandler has stopped.
+			// If handleOut stops first, entryHandler might get stuck on a channel send.
+			close(shutdownCh)
+			wgOut.Wait()
 		}
 		c.mut.RUnlock()
 	}()
 	wg := &sync.WaitGroup{}
```
thampiotr, line 113:
```suggestion
	wgIn := &sync.WaitGroup{}
```

thampiotr (approved):
LGTM % couple of suggestions for the naming to make it easier to read

PR #1600, opened 2024-09-03
Add prometheus.relabel external cache proposal

wilfriedroset:
#### PR Description

This PR add a proposal to introduce the support for an external cache for `prometheus.relabel `component.

N/A

#### Notes to the Reviewers

### DIFF
--- a/docs/design/1600-prometheus-relabel-external-cache.md
+++ b/docs/design/1600-prometheus-relabel-external-cache.md
@@ -43,7 +43,7 @@ The alternative is to do nothing as we deem this improvement unnecessary.
 
 ## Compatibility
 
-This proposal offer to deprecate for a couple of release the old way to configure the in-memory cache, then release+2 drop the old way. Doing so allow to migrate the settings to the correct block.
+This proposal offer to deprecate the old way to configure the in-memory cache and drop it in the next major release (e.g: 2.0). Doing so allow to migrate the settings to the correct block.
 
 ## Implementation
 


docs/design/1600-prometheus-relabel-external-cache.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+- redis config
+
+```river
+prometheus.relabel "redis_config" {
+  forward_to = [...]
+  cache {
+    backend = "redis"
+    redis {
```
mattdurham, line 110: Instead of a bespoke redis/memcached action, how about a redis and memcached component that exports a key value interface? Along with generic accessors. I could see this being used for more than just relabel. Much in the same way we allow components to consume strings but those strings can come from vault, file, http etc.

mattdurham: Could even get clever and allow prometheus.relabel to have a list of caches, if there are problems with redis then fallback to in memory.

```
prometheus.relabel "relabel" {
  caches = [storage.redis.relabel.cache,storage.memory.relabel]
}
```

wilfriedroset: I definitively like the idea of having a unified way to define the caching mecanism.
> if there are problems with redis then fallback to in memory.

You are correct but this add an extra complexity to take into account. A highly available external cache is indeed recommended but having a fallback is definitively a good idea.

The extra complexity comes from when the external caches becomes unavailable Alloy needs to  use the in-memory one. This could lead to OOM, e.g: you have plan to use an external cache then you fallback on in-memory but the limits of your pod is not high enough.
Then Alloy needs to check if the external storage is available again. Once it is, Alloy needs to move its in-memory k/v to the external cache. But we are exposed to a race when we have two Alloy instances using the same cache.

That being said, as I'm not that familiar with the code base yet and your proposal seems harder to achieve.

What do you think about extending our proposal to `loki.relabel` and standardize the whole thing as prometheus and loki are the main component using cache?
This would serve as a first step toward your proposal. Something like so
```go
type RelabelCache interface {
    Get(key uint64) (*labelAndID, error)
    Set(key uint64, value *labelAndID) error
    Remove(key uint64)
    Clear(newSize int) error
    GetCacheSize() int
} 
```

Once done, the associated work to create the `storage` components such as `storage.redis` could be picked up right after the initial proposal.

mattdurham: Fallback caches are likely a stretch though I do feel the kv store should be a generic component. As far as the interface I would go with something like the below. Disclaimer this is getting into the deeper technical ideas. The reason keys might want to be strings is to mirror how alloy uses maps. Where the key is always a string.

```
type Cache interface {
    Get(key string) (any, bool) 
    Set(key string, value any, ttl int) 
    Remove(key uint64)
    Clear(newSize int) error
    GetCacheSize() int
} 
```

pbailhache:
I agree for the more standard interface you proposed, I started a poc only with relabel and that's why I ended with this relabel specific interface.

Concerning the cache as component, what do you advise ?

I looked into https://github.com/grafana/alloy/blob/main/docs/developer/writing-component-documentation.md and more closely at 
`capsule(T): An internal encapsulated value.` 

For what I gather that's the correct way to implement such a cache component. Am I correct about this ?

docs/design/1600-prometheus-relabel-external-cache.md
pbailhache: I agree for the more standard interface you proposed, I started a poc only with relabel and that's why I ended with this relabel specific interface.

Concerning the cache as component, what do you advise ?

I looked into https://github.com/grafana/alloy/blob/main/docs/developer/writing-component-documentation.md and more closely at 
`capsule(T): An internal encapsulated value.` 

For what I gather that's the correct way to implement such a cache component. Am I correct about this ?

mattdurham: You shouldn't have to specifically use that for the first round. Capsules are mainly if you want to bridge into it being accessible within the alloy config. In this case your likely passing an interface/pointer from go code to go code, which should work without problem.

wilfriedroset: I've updated the proposal to take reflect what we have been discussing.

mattdurham: I also think we will likely need `GetMultiple(keys []string) ([]Results)` and `SetMultiple`, imagine a scenario where you are scraping a million records, hitting redis each time is going to be pretty expensive at one request for each.

wilfriedroset: @mattdurham  I've tried to take your feedbacks into account, the PR should be fine what do you think about it?
We have started to work on the code with @pbailhache, what would be the next step regarding this proposal? should we merge it if it is to your liking and iterate over the code?


p.s: I don't know what to do about the build 😅

docs/design/1600-prometheus-relabel-external-cache.md
```diff
@@ -0,0 +1,172 @@
...
+
+- Config is a bit more complex compared to previous one
+
+## Alternative solutions
+
+The alternative is to do nothing as we deem this improvement unnecessary.
+
+## Compatibility
+
+This proposal offer to deprecate for a couple of release the old way to configure the in-memory cache, then release+2 drop the old way. Doing so allow to migrate the settings to the correct block.
```
tpaschalis, line 46: I think our accepted [backwards compatibility](https://grafana.com/docs/alloy/latest/introduction/backward-compatibility/) guarantees go against this. Any breaking changes for stable functionality would have to wait until v2.0.

wilfriedroset: We don't want to deprecate anything sooner than what is guaranteed.
I will rephrase to say that we keep both compatibility until the next major release

docs/design/1600-prometheus-relabel-external-cache.md
pbailhache: Just opened a draft pull request for the feature :
https://github.com/grafana/alloy/pull/1692

This is still a work in progress and I'm ready to iterate over it.

wilfriedroset:
This is not inactive as we are iterating on an implementation in a distinct PR

PR #1601, opened 2024-09-03
Add filequeue functionality

mattdurham:
FileQueue offers the lowest level functionality, it writes a block of data and meta data to a file. Then pushes the data out. This also serves as the introduction to the actor framework, primarily via DoWork function. This allows us to avoid mutexes and have a well defined workflow that will be reused.


internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+		meta = make(map[string]string)
+	}
+	q.maxIndex++
+	name := filepath.Join(q.directory, fmt.Sprintf("%d.committed", q.maxIndex))
+	// record wraps the data and metadata in one. This allows the consumer to take action based on the map.
+	r := &Record{
+		Meta: meta,
+		Data: data,
+	}
+	// TODO @mattdurham reuse a buffer here.
```
mattdurham, line 157: Not entirely sold on reusing the buffer here. Its not a lot of allocs to recreate, it also lowers the overall memory footprint.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	case item, ok := <-q.files.ReceiveC():
+		if !ok {
+			return actor.WorkerEnd
+		}
+		name, err := q.add(item.Meta, item.Data)
+		if err != nil {
+			level.Error(q.logger).Log("msg", "error adding item", "err", err)
+			return actor.WorkerContinue
+		}
+		q.out(ctx, types.DataHandle{
```
mattdurham, line 135: This will block until out  returns. This is important for flow control, until the file is processed another wont be added.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/types"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/vladopajic/go-actor/actor"
+)
+
+var _ actor.Worker = (*queue)(nil)
+var _ types.FileStorage = (*queue)(nil)
+
+// queue represents the on disk queue. This is an ordered list by id in the format: <id>.committed
```
thampiotr, line 21:
```suggestion
// queue represents an on-disk queue. This is a list implemented as files ordered by id with a name pattern: <id>.committed
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/types"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/vladopajic/go-actor/actor"
```
thampiotr, line 15: I'm going to have to spend some more time looking into the concurrency model here. Any reasons why you chose the actor model? 

I worked with teams that introduced actors only to get rid of them a year later because it became impossible to follow and optimise and very hard to test. So I'd like to be careful with that or make it easy to replace. The `go-actor` seems like a clean interface, but I haven't seen it used in other projects yet.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+var _ types.FileStorage = (*queue)(nil)
+
+// queue represents the on disk queue. This is an ordered list by id in the format: <id>.committed
+// Each file contains a byte buffer and an optional metatdata map.
+type queue struct {
+	self      actor.Actor
+	directory string
+	maxIndex  int
+	logger    log.Logger
+	files     actor.Mailbox[types.Data]
```
thampiotr, line 28: I'd prefer to call this `inbox`? It's a mailbox with Data, not files yet.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	// that get installed into the system.
+	matches, _ := filepath.Glob(filepath.Join(directory, "*.committed"))
+	ids := make([]int, len(matches))
+
+	// Try and grab the id from each file.
+	// 1.committed
+	for i, x := range matches {
+		id, err := strconv.Atoi(strings.ReplaceAll(filepath.Base(x), ".committed", ""))
+		if err != nil {
+			continue
```
thampiotr, line 53: IMO we need to log a warning here or at least debug level with err details.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+
+var _ actor.Worker = (*queue)(nil)
+var _ types.FileStorage = (*queue)(nil)
+
+// queue represents the on disk queue. This is an ordered list by id in the format: <id>.committed
+// Each file contains a byte buffer and an optional metatdata map.
+type queue struct {
+	self      actor.Actor
+	directory string
+	maxIndex  int
```
thampiotr, line 26:
```suggestion
	maxId  int
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	// 1.committed
+	for i, x := range matches {
+		id, err := strconv.Atoi(strings.ReplaceAll(filepath.Base(x), ".committed", ""))
+		if err != nil {
+			continue
+		}
+		ids[i] = id
+	}
+	sort.Ints(ids)
+	var currentIndex int
```
thampiotr, line 58:
```suggestion
	var currentMaxId int
```
It's not for holding an index, but for holding an ID.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	q := &queue{
+		directory:      directory,
+		maxIndex:       currentIndex,
+		logger:         logger,
+		out:            out,
+		files:          actor.NewMailbox[types.Data](),
+		existingsFiles: make([]string, 0),
+	}
+
+	// Push the files that currently exist to the channel.
```
thampiotr, line 71:
```suggestion
	// Save the existing files in `q.existingFiles`, which will be push to `out` when actor starts.
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, id := range ids {
+		name := filepath.Join(directory, fmt.Sprintf("%d.committed", id))
+		q.existingsFiles = append(q.existingsFiles, name)
+	}
+	return q, nil
+}
+
+func (q *queue) Start() {
+	q.self = actor.Combine(actor.New(q), q.files).Build()
```
thampiotr, line 80: Took me a while to figure out why we need to do that... maybe a comment would be nice. I think the key thing to note is that `q.files` is owned and managed by this instance of queue.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	q.self.Start()
+
+}
+
+func (q *queue) Stop() {
+	q.self.Stop()
+}
+
+func (q *queue) Store(ctx context.Context, meta map[string]string, data []byte) error {
+	return q.files.Send(ctx, types.Data{
```
thampiotr, line 90: Is `files` mailbox buffered by default? When will this block? 
I'd like to have a comment about this for future reference, readability.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	select {
+	case <-ctx.Done():
+		return actor.WorkerEnd
+	case item, ok := <-q.files.ReceiveC():
+		if !ok {
+			return actor.WorkerEnd
+		}
+		name, err := q.add(item.Meta, item.Data)
+		if err != nil {
+			level.Error(q.logger).Log("msg", "error adding item", "err", err)
```
thampiotr, line 132:
```suggestion
			level.Error(q.logger).Log("msg", "error adding item - dropping data", "err", err)
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
thampiotr: I guess the consumers have a freedom to implement a buffer / async consumption themselves, so I'm cool with that choice. Would be good to document it in godoc, however :)

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+			Name: name,
+			Get: func() (map[string]string, []byte, error) {
+				return get(name)
+			},
+		})
+		return actor.WorkerContinue
+	}
+}
+
+// Add a committed file to the queue.
```
thampiotr, line 145:
```suggestion
// Add a file to the queue (as committed).
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	r := &Record{}
+	_, err = r.UnmarshalMsg(buf)
+	if err != nil {
+		return nil, nil, err
+	}
+	return r.Meta, r.Data, nil
+}
+
+func (q *queue) DoWork(ctx actor.Context) actor.WorkerStatus {
```
thampiotr, line 111: Just to be sure. The Actor framework will manage concurrency such that ever only one goroutine will be running the `DoWork` for given queue instance?

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Add a committed file to the queue.
+func (q *queue) add(meta map[string]string, data []byte) (string, error) {
+	if meta == nil {
+		meta = make(map[string]string)
+	}
+	q.maxIndex++
+	name := filepath.Join(q.directory, fmt.Sprintf("%d.committed", q.maxIndex))
+	// record wraps the data and metadata in one. This allows the consumer to take action based on the map.
```
thampiotr, line 152: Not sure what you mean, because the consumer doesn't get the `Record`, the `out` function callback has a `Get` that returns the map and data separately.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
thampiotr: Yep, I don't think we need to do that until we hit performance issues and then it may still not be the main issue.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+	// TODO @mattdurham reuse a buffer here.
+	rBuf, err := r.MarshalMsg(nil)
+	if err != nil {
+		return "", err
+	}
+	err = q.writeFile(name, rBuf)
+	if err != nil {
+		return "", err
+	}
+	return name, err
```
thampiotr, line 166:
```suggestion
	return name, nil
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (q *queue) writeFile(name string, data []byte) error {
+	return os.WriteFile(name, data, 0644)
+}
+
+func deleteFile(name string) {
+	_ = os.Remove(name)
```
thampiotr, line 174: probably will never happen, but I'd log on warn level the error... 
the number of times I had failed disk produce no useful error messages ...

thampiotr (commented):
First round! It's looking good, happy it's comparably lean 👍

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: Inbox feels a little generic, but agree files is wrong.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: That is correct.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: More so a note of what Record is, I can remove this comment and make sure its represented on the Record structs themselves.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: Mailboxes are always unbounded, unless you specify a capacity. Which I do in a few upcoming PRs but will ensure I document this one and those,

internal/component/prometheus/remote/queue/filequeue/filequeue.go
thampiotr: Cool! I do see that it's quite lightweight. Some high-level idea about how actors are scheduled and what is blocking, what has a buffer would be a good idea to document. But I think I'm happy with the library choice.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,190 @@
...
+// queue represents an on-disk queue. This is a list implemented as files ordered by id with a name pattern: <id>.committed
+// Each file contains a byte buffer and an optional metatdata map.
+type queue struct {
+	self      actor.Actor
+	directory string
+	maxID     int
+	logger    log.Logger
+	dataQueue actor.Mailbox[types.Data]
+	// Out is where to send data when pulled from queue, it is assumed that it will
+	// block until ready for another file.
```
thampiotr, line 30:
```suggestion
	// block until ready for another record.
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, err
+	}
+
+	// We dont actually support uncommitted but I think its good to at least have some naming to avoid parsing random files
+	// that get installed into the system.
+	matches, _ := filepath.Glob(filepath.Join(directory, "*.committed"))
+	ids := make([]int, len(matches))
+
+	// Try and grab the id from each file.
+	// 1.committed
```
thampiotr, line 49:
```suggestion
	// e.g. grab 1 from `1.committed`
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,190 @@
...
+	q := &queue{
+		directory:      directory,
+		maxID:          currentMaxID,
+		logger:         logger,
+		out:            out,
+		dataQueue:      actor.NewMailbox[types.Data](),
+		existingsFiles: make([]string, 0),
+	}
+
+	// Save the existing files in `q.existingFiles`, which will be push to `out` when actor starts.
```
thampiotr, line 72:
```suggestion
	// Save the existing files in `q.existingFiles`, which will have their data pushed to `out` when actor starts.
```

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,190 @@
...
+	r := &Record{
+		Meta: meta,
+		Data: data,
+	}
+	// Not reusing a buffer here since allocs are not bad here and we are trying to reduce memory.
+	rBuf, err := r.MarshalMsg(nil)
+	if err != nil {
+		return "", err
+	}
+	err = q.writeFile(name, rBuf)
```
thampiotr, line 167: One file = one piece of data implies that we probably want to have a good level of batching for this queue and probably it's worth calling out that it's going to work better with larger data.

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/types"
+	"github.com/stretchr/testify/require"
+	"github.com/vladopajic/go-actor/actor"
+)
+
+func BenchmarkFileQueue(t *testing.B) {
```
thampiotr, line 14: Nit: Can you add as a comment the current results of the benchmark for some future reference? I think it can be useful in certain cases.

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
```diff
@@ -0,0 +1,50 @@
...
+	for i := 0; i < t.N; i++ {
+		dir := t.TempDir()
+		log := log.NewNopLogger()
+		mbx := actor.NewMailbox[types.DataHandle]()
+		mbx.Start()
+		defer mbx.Stop()
+		q, err := NewQueue(dir, func(ctx context.Context, dh types.DataHandle) {
+			_ = mbx.Send(ctx, dh)
+		}, log)
+		require.NoError(t, err)
```
thampiotr, line 24: This creation code probably should be outside the benchmark?

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
```diff
@@ -0,0 +1,50 @@
...
+		q.Start()
+		defer q.Stop()
+		err = q.Store(context.Background(), nil, []byte("test"))
```
thampiotr, line 27: I'd imagine we want to create one `q` and store something `t.N` times.

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		require.NoError(t, err)
+
+		meta, buf, err := getHandleBench(mbx)
+		require.NoError(t, err)
+		require.True(t, string(buf) == "test")
+		require.Len(t, meta, 0)
+
+		// Ensure nothing new comes through.
+		timer := time.NewTicker(100 * time.Millisecond)
```
thampiotr, line 37: Doesn't this make the benchmark always last 100ms+?

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
thampiotr, line 27: I wonder how it works for different message lengths too.

internal/component/prometheus/remote/queue/filequeue/filequeue_test.go
```diff
@@ -0,0 +1,248 @@
...
+	require.NoError(t, err)
+
+	err = q.Store(context.Background(), map[string]string{"name": "bob"}, []byte("first"))
+	require.NoError(t, err)
+	err = q.Store(context.Background(), map[string]string{"name": "bob"}, []byte("second"))
+
+	require.NoError(t, err)
+
+	// Send is async so may need to wait a bit for it happen.
+	require.Eventually(t, func() bool {
```
thampiotr, line 98: Could this be flaky? when Eventually doesn't catch the 1.committed existing, cause it appears and disappears too fast?

Would be good to run this test with `-count 100` or sth to validate it's not flaky if you didn't already :)

internal/component/prometheus/remote/queue/filequeue/filequeue.go
```diff
@@ -0,0 +1,190 @@
...
+
+// DoWork allows most of the queue to be single threaded with work only coming in and going out via mailboxes(channels).
+func (q *queue) DoWork(ctx actor.Context) actor.WorkerStatus {
+	// Queue up our existing items.
+	for _, name := range q.existingsFiles {
+		q.out(ctx, types.DataHandle{
+			Name: name,
+			Get: func() (map[string]string, []byte, error) {
+				return get(q.logger, name)
+			},
```
thampiotr, line 123: So a call to `get` will delete the file, and thus types.DataHandle.Get() can only be called once because it has a side-effect (unexpected for a method called Get, btw)

I think that's an important behaviour that needs to be documented. But also not sure if we want this... can you share some context why we do a lazy file reading via DataHandle and delete it only after it's read?

What happens if `Get` is never called for some reason? Seems like we'd leak a file and it would be picked up again on next run?

internal/component/prometheus/remote/queue/filequeue/filequeue_test.go
thampiotr: I just realised that we won't delete the file until we call `Get` on the data handle... not sure if I like this side-effect, let's discuss in another thread.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: The assumption (that I should document) is that `Get` is called when the caller is ready to process the file, to limit the amount of lost data and limit the amount of memory in use. The `out` queue (shown in upcoming PR) has a capacity of one, so if `Get` is never called then data was never sent. This allows us to err on the side of dropping data versus sending duplicates.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: In general yes, in the case of a single node exporter self scraping its kilobytes. The default is it writes `10,000` signals in the full PR with a flush timer to ensure its timely. When testing internally the files after snappy compressed were megabytes.

thampiotr: Yeah, that likely will work if you write entire scrapes to the queue and I'm fine with that... if you started to write single samples then it's gonna be a lot of overhead. That's all fine as long as we make it somewhat clear that the queue is optimised only for some use patterns.

BTW, I'd like to see some metrics on how the queue performs, but I'm happy for this to be in future PRs.

mattdurham: There are some e2e benchmarks in the full pr. The file queue is really cheap though.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: Retry/Backoff is handled in the network loop. In regards to get not being called the actual implementation of the callee processes them one at a time. If Get is not called then it should be retried on restart which would requeue it. In the main PR the lifecycles of the FileQueue and the Endpoint (callee in practice), are tied together at the component level. From the filequeue perspective the endpoint processes the file sequentially, deserializing and then adding to the network buffer, and stops pulling files when the network buffer is full.

internal/component/prometheus/remote/queue/filequeue/filequeue_bench_test.go
mattdurham: Going to drop this benchmark, its not super useful. There are better end to end benchmarks that track it better.

internal/component/prometheus/remote/queue/filequeue/filequeue.go
mattdurham: Pop is more accurate to whats going on so will change that and add a comment.

thampiotr (approved):

PR #1606, opened 2024-09-04
fix(helm): explicitly include namespace in manifests

### LINKED ISSUE #1607: Helm chart does not work well with "helm template"
Currently, most of the manifests omit `.metadata.namespace`.

This works fine with `helm install` / `helm upgrade` where all the manifests without a namespace will be installed to `--namespace` anyway.

However, `helm template` will produce the manifests as-is (with no namespace), regardless of `--namespace`, forcing users to add the namespace manually, or use additional tools on top like kustomize.

valorl:
Currently, most of the manifests omit `.metadata.namespace`.

This works fine with `helm install` / `helm upgrade` where all the manifests without a namespace will be installed to `--namespace` anyway.

However, `helm template` will produce the manifests as-is (with no namespace), regardless of `--namespace`, forcing users to add the namespace manually or use more tooling.

This change adds the namespace to all manifests.

Fixes #1607 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/cluster_service.yaml
+++ b/operations/helm/charts/alloy/templates/cluster_service.yaml
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "alloy.fullname" . }}-cluster
+  namespace: {{ include "alloy.namespace" . }}
   labels:
     {{- include "alloy.labels" . | nindent 4 }}
     app.kubernetes.io/component: networking
--- a/operations/helm/charts/alloy/templates/configmap.yaml
+++ b/operations/helm/charts/alloy/templates/configmap.yaml
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "alloy.fullname" . }}
+  namespace: {{ include "alloy.namespace" . }}
   labels:
     {{- include "alloy.labels" . | nindent 4 }}
     app.kubernetes.io/component: config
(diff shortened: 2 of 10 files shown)


operations/helm/charts/alloy/templates/cluster_service.yaml
```diff
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "alloy.fullname" . }}-cluster
+  namespace: {{ include "alloy.namespace" . }}
```
neopointer, line 7: @valorl excuse my ignorance, but why use `alloy.namespace` instead of `Release.Namespace`?

neopointer:
@clayton-cornell I'm just tagging a random contributor here. What is this PR missing to get merged?

clayton-cornell:
@neopointer It looks like this one got missed in the shuffle. Lets see if we can get a developer/maintainer review happening here.

@grafana/grafana-alloy-maintainers can someone form the team take a look at this?

mattdurham (approved):
LGTM

mattdurham:
@petewall any thoughts on if this causes any issue?

mattdurham:
This PR #2044 seems to be the most active and achieves the same result?

mattdurham (changes requested):
Looking to see if we can consolidate PRs

valorl:
Closing in favor of #2044

PR #1624, opened 2024-09-06
loki.process: fix bug where drop_counter_reason gets set incorrectly

captncraig:
#### PR Description

The issue we are seeing is that if you define multiple `loki.process` stages with `sampling`, both components will use the same drop reason label. 

For some reason, both string pointers have the same address, and I am still not clear why. Changing the drop reason to a plain string avoids the issue, as well as not calling `WithLabels` inside the loop.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/sampling.go
+++ b/internal/component/loki/process/stages/sampling.go
@@ -66,6 +66,7 @@ type samplingStage struct {
 
 func (m *samplingStage) Run(in chan Entry) chan Entry {
 	out := make(chan Entry)
+	reason := *m.cfg.DropReason
 	go func() {
 		defer close(out)
 		for e := range in {
@@ -73,7 +74,7 @@ func (m *samplingStage) Run(in chan Entry) chan Entry {
 				out <- e
 				continue
 			}
-			m.dropCount.WithLabelValues(*m.cfg.DropReason).Inc()
+			m.dropCount.WithLabelValues(reason).Inc()
 		}
 	}()
 	return out
--- a/internal/component/loki/process/stages/sampling.go
+++ b/internal/component/loki/process/stages/sampling.go
@@ -27,9 +27,7 @@ type SamplingConfig struct {
 }
 
 func (s *SamplingConfig) SetToDefault() {
-	if s.DropReason == "" {
-		s.DropReason = defaultSamplingpReason
-	}
+	s.DropReason = defaultSamplingpReason
 }
 
 func (s *SamplingConfig) Validate() error {
--- a/internal/component/loki/process/stages/sampling.go
+++ b/internal/component/loki/process/stages/sampling.go
@@ -22,13 +22,13 @@ var (
 
 // SamplingConfig contains the configuration for a samplingStage
 type SamplingConfig struct {
-	DropReason   *string `alloy:"drop_counter_reason,attr,optional"`
+	DropReason   string  `alloy:"drop_counter_reason,attr,optional"`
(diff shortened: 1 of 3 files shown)


internal/component/loki/process/stages/sampling.go
```diff
@@ -66,14 +66,15 @@ type samplingStage struct {
 
 func (m *samplingStage) Run(in chan Entry) chan Entry {
 	out := make(chan Entry)
+	reason := *m.cfg.DropReason
```
captncraig, line 69: I'm not sure this is sufficient. I think `m.isSampled` may give the wrong result too. I think I need to make a local for all of m.

wildum:
> For some reason, both string pointers have the same address, and I am still not clear why. Changing the drop reason to a plain string avoids the issue, as well as not calling WithLabels inside the loop.

This could hide more bugs, I think that we should dig deeper to understand how this happened. The two components should not point to the same config

ptodev:
There have been other issues with pointer config attributes in this component too - #1426

internal/component/loki/process/stages/sampling.go
```diff
@@ -22,13 +22,13 @@ var (
...
-	DropReason   *string `alloy:"drop_counter_reason,attr,optional"`
+	DropReason   string  `alloy:"drop_counter_reason,attr,optional"`
 	SamplingRate float64 `alloy:"rate,attr"`
 }
 
 func (s *SamplingConfig) SetToDefault() {
-	if s.DropReason == nil || *s.DropReason == "" {
-		s.DropReason = &defaultSamplingpReason
+	if s.DropReason == "" {
+		s.DropReason = defaultSamplingpReason
```
wildum, line 31: I dug a little bit to check the reason behind this: what's happening is that `SetToDefault()` is always called before the args are set (that makes sense: first the defaults are set and then the params set by the user will override the defaults). The problem here was that the default was a global var and because DropReason was a pointer, its default value was the address of `defaultSamplingpReason`, not its value. So when the value is overridden by the user input, it's actually overriding the value of the var `defaultSamplingpReason`.
When the next component is built, it will also point to `defaultSamplingpReason` and modify the value with the user input. So both components are using the same value `defaultSamplingpReason` that they override.

Looking at the other stages, it seems like this is the only one that has this problem

internal/component/loki/process/stages/sampling.go
```diff
@@ -22,13 +22,13 @@ var (
...
 type SamplingConfig struct {
-	DropReason   *string `alloy:"drop_counter_reason,attr,optional"`
+	DropReason   string  `alloy:"drop_counter_reason,attr,optional"`
 	SamplingRate float64 `alloy:"rate,attr"`
 }
 
 func (s *SamplingConfig) SetToDefault() {
-	if s.DropReason == nil || *s.DropReason == "" {
-		s.DropReason = &defaultSamplingpReason
+	if s.DropReason == "" {
```
wildum, line 30: nit: it would be best to remove the if condition because it is misleading. It will always be true

internal/component/loki/process/stages/sampling.go
mattdurham: A while back we switched to defaultVals being a func that returned a new object for this type of reason. Glad the other stages arent affected.

wildum (approved):
lgtm, thanks!

PR #1625, opened 2024-09-06
Upgrade Go version to 1.23.1

wildum:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grafana/alloy
 
-go 1.22.5
+go 1.23.1
```
mattdurham, line 3: Do we need to also update pyroscope here? I know there was a collision between pyroscope pprof endpoint and go 1.23?

mattdurham: Run the exe and it should trigger either immediately or hitting the pprof endpoint

aleks-p: The CI caught it as well it seems. Bumping `github.com/grafana/pyroscope-go/godeltaprof` to 0.1.8 will likely fix it. Here is a related fix in Grafana: https://github.com/grafana/grafana/pull/92055

wildum: thanks, I will do it

mattdurham:
kwe may need to separate the build image from the executable version change.

wildum:
yeah this is what I'm trying to do but boringcrypto is not happy

PR #1629, opened 2024-09-06
Fix issue preventing import.git from syncing due to divergent branches

ptodev:
#### PR Description

Consider this situation:

* `import.git` is configured to use branch A.
* Alloy is then stopped, and its new configuration is configured to use branch B.
* If branches A and B are divergent, `import.git` will fail to switch to branch B because it's going to try to pull A into B.

The user will get a `"failed to update repository \"git@github.com:org/repo.git\": non-fast-forward update"` error.

I suppose the same thing could also happen if a force push is done to the branch Alloy is monitoring.

Since Alloy is only reading from the repo, but not writing, there is no point in trying to do a pull. It's much easier to just do a fetch and checkout.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,8 @@ v1.4.0
   configuration could result in incorrect URLs. The `path` and `key` arguments have been separated to allow for clear and accurate
   specification of Vault secrets. (@PatMis16)
 
+- Fixed a bug in `import.git` which caused a `"non-fast-forward update"` error message. (@ptodev)
+
 ### Other
 
 - Renamed standard library functions. Old names are still valid but are marked deprecated. (@wildum)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
+
+- `import.git`: The default value for `revision` has changed from `HEAD` to `main`. (@ptodev)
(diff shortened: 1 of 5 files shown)


mattdurham:
I think this was my original approach but it had issues with switching branches and accessing HEAD. Hence the test failing.

ptodev:
> I think this was my original approach but it had issues with switching branches and accessing HEAD. Hence the test failing.

@mattdurham  Thank you, I had a look at the [Agent PR](https://github.com/grafana/agent/pull/6668) which introduced pull just now. I think the default value for "revision" shouldn't be something like "HEAD". HEAD refers to the tip of the current branch. But what is the current branch in import.git? :) There is no config attribute to set it, so I'm not sure how it gets chosen. Maybe the remote tells it what the default branch is.

I think we should simply not allow values such as "HEAD", "FETCH_HEAD", etc to be the "revision". I updated my PR accordingly - would you like to take a look please? If you agree with it, I can update the docs and I'll also list it as a breaking change in the changelog.

If a user is able to set "revision" to something like "HEAD", "FETCH_HEAD" etc, then they'd be relying on some internal quirk of how import.git works, which wouldn't be correct.

mattdurham:
HEAD isn't a quirk of import.git but fundamental feature for git. That being said I am not against saying you cant use the `pointer` style aliases and instead have to use a non alias value.

mattdurham (approved):
LGTM, its a breaking change but if this cleans this up totally fine and once the tests pass.

PR #1635, opened 2024-09-09
Fix "alloy.controller.volumes.extra" field not being populated

tjaacks:
Volumes specified in `alloy.controller.volumes.extra` were not rendered into the resulting `DeamonSet` manifest. Fix this by using `$values` instead of `.Values`.

Example:

```yaml
alloy:
  controller:
    volumes:
      extra:
      - name: myvolume
        hostPath:
          path: /path/to/something
```

Adding the above snippet to the helm chart's `values.yaml` did not add the specified volume to the pod. This can be tested using the following `kubectl` call:

```cmd
kubectl get pod alloy-4pkwb -o jsonpath='{.spec.volumes[?(@.name=="myvolume")]}
```

Before this PR the call did not return anything. After the PR it returns:

```json
{"hostPath":{"path":"/path/to/something","type":""},"name":"myvolume"}
```

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -10,6 +10,10 @@ internal API changes are not present.
 Unreleased
 ----------
 
+### Bugfixes
+
+- Fix "alloy.controller.volumes.extra" field not being populated. (@tjaacks)
+
 0.7.0 (2024-08-26)
 ------------------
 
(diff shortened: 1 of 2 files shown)


tjaacks:
I just realized that the `controller` field should not be a child of `alloy`. My fault, sorry for the noise.

PR #1639, opened 2024-09-10
Fix typo in docs components.md

fridgepoet:
#### PR Description
If I understood correctly, the component is "local.file" and its label is "api_key" so the comment seemed to have a typo. Let me know if I have misunderstood!

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/get-started/components.md
+++ b/docs/sources/get-started/components.md
@@ -74,7 +74,7 @@ The following configuration file represents the pipeline.
 // This component has an exported field called "content", holding the content
 // of the file.
 //
-// local.file.api_key will watch the file and update its exports any time the
+// local.file "api_key" will watch the file and update its exports any time the
 // file changes.
 local.file "api_key" {
   filename  = "/var/data/secrets/api-key"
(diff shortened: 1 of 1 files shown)


docs/sources/get-started/components.md
```diff
@@ -74,7 +74,7 @@ The following configuration file represents the pipeline.
 // This component has an exported field called "content", holding the content
 // of the file.
 //
-// local.file.api_key will watch the file and update its exports any time the
+// local.file "api_key" will watch the file and update its exports any time the
```
fridgepoet, line 77: Or should it just read:
```suggestion
// local.file will watch the file and update its exports any time the
```

clayton-cornell (approved):
Good catch @fridgepoet  Looks like a simple copy-paste error crept into the content. Than happens and they can be hard to spot some days... so thank you for submitting the PR. 

I committed your alternate suggestion as I agree, that's more correct here.

PR #1643, opened 2024-09-10
Update mysqld_exporter to v0.16.0

cristiangreco:
#### PR Description
Update to mysqld_exporter v0.16.0 and also include https://github.com/prometheus/mysqld_exporter/pull/897.

Note that we continue to use the grafana/mysqld_exporter branch `refactor_collector`, including changes from our branch https://github.com/prometheus/mysqld_exporter/pull/774.

n.a.

#### Notes to the Reviewer
This introduces the use of github.com/tjhop/slog-gokit

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ Main (unreleased)
 
 - Logs from underlying clustering library `memberlist` are now surfaced with correct level (@thampiotr)
 
-- Update mysqld_exporter from v0.15.0 to v0.16.0, most notable changes: (@cristiangreco)
+- Update mysqld_exporter from v0.15.0 to v0.16.0 (including 2ef168bf6), most notable changes: (@cristiangreco)
   - Support MySQL 8.4 replicas syntax
   - Fetch lock time and cpu time from performance schema
   - Fix fetching tmpTables vs tmpDiskTables from performance_schema
@@ -42,6 +42,7 @@ Main (unreleased)
   - Fixed parsing of timestamps with non-zero padded days
   - Fix auto_increment metric collection errors caused by using collation in INFORMATION_SCHEMA searches
   - Change processlist query to support ONLY_FULL_GROUP_BY sql_mode
+  - Add perf_schema quantile columns to collector
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,15 @@ Main (unreleased)
 
 - Logs from underlying clustering library `memberlist` are now surfaced with correct level (@thampiotr)
 
+- Update mysqld_exporter from v0.15.0 to v0.16.0, most notable changes: (@cristiangreco)
+  - Support MySQL 8.4 replicas syntax
+  - Fetch lock time and cpu time from performance schema
+  - Fix fetching tmpTables vs tmpDiskTables from performance_schema
+  - Skip SPACE_TYPE column for MariaDB >=10.5
+  - Fixed parsing of timestamps with non-zero padded days
+  - Fix auto_increment metric collection errors caused by using collation in INFORMATION_SCHEMA searches
+  - Change processlist query to support ONLY_FULL_GROUP_BY sql_mode
+
 ### Bugfixes
 
 - Fixed an issue in the `prometheus.exporter.postgres` component that would leak goroutines when the target was not reachable (@dehaansa)
--- a/go.mod
+++ b/go.mod
@@ -916,7 +916,7 @@ replace (
 	github.com/prometheus/exporter-toolkit => github.com/prometheus/exporter-toolkit v0.11.0
 
 	// TODO(marctc): remove once this PR is merged upstream: https://github.com/prometheus/mysqld_exporter/pull/774
-	github.com/prometheus/mysqld_exporter => github.com/grafana/mysqld_exporter v0.16.1-0.20241111085008-47b8ea9fab19
+	github.com/prometheus/mysqld_exporter => github.com/grafana/mysqld_exporter v0.16.1-0.20241128200101-e16f5286d5e9
 
 	// TODO(marctc, mattdurham): Replace node_export with custom fork for multi usage. https://github.com/prometheus/node_exporter/pull/2812
 	github.com/prometheus/node_exporter => github.com/grafana/node_exporter v0.18.1-grafana-r01.0.20231004161416-702318429731
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/go-kit/log v0.2.1
 	github.com/go-logfmt/logfmt v0.6.0
 	github.com/go-sourcemap/sourcemap v2.1.3+incompatible
-	github.com/go-sql-driver/mysql v1.7.1
+	github.com/go-sql-driver/mysql v1.8.1
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.5.4
 	github.com/golang/snappy v0.0.4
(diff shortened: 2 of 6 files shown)


internal/component/loki/source/aws_firehose/internal/handler_test.go
```diff
@@ -737,8 +737,10 @@ func TestGetStaticLabelsFromRequest_NoError_InvalidData(t *testing.T) {
 
 			require.Equal(t, tt.want, got)
 
-			err := testutil.GatherAndCompare(registry, strings.NewReader(tt.expectedMetrics), "loki_source_awsfirehose_invalid_static_labels_errors")
-			require.NoError(t, err)
+			if tt.expectedMetrics != "" {
+				err := testutil.GatherAndCompare(registry, strings.NewReader(tt.expectedMetrics), "loki_source_awsfirehose_invalid_static_labels_errors")
+				require.NoError(t, err)
+			}
```
andriikushch, line 743:
```suggestion
if tt.expectedMetrics != "" {
				err := testutil.GatherAndCompare(registry, strings.NewReader(tt.expectedMetrics), "loki_source_awsfirehose_invalid_static_labels_errors")
				require.NoError(t, err)
			} else {
				err := testutil.GatherAndCompare(registry, strings.NewReader(tt.expectedMetrics), "loki_source_awsfirehose_invalid_static_labels_errors")
				require.EqualError(t, err, "expected metric name(s) not found: [loki_source_awsfirehose_invalid_static_labels_errors]")
			}
```

internal/component/faro/receiver/exporters_test.go
```diff
@@ -12,10 +12,10 @@ import (
...
 
 var metricNames = []string{
-	"logs_total",
-	"measurements_total",
-	"exceptions_total",
-	"events_total",
+	"faro_receiver_logs_total",
+	"faro_receiver_measurements_total",
+	"faro_receiver_exceptions_total",
+	"faro_receiver_events_total",
```
cristiangreco, line 18: Tests in this PR are updated due to a breaking change in `prometheus/client_golang`, nothing related to `mysqld_exporter` itself.

go.mod
```diff
@@ -161,7 +161,7 @@ require (
 	github.com/prometheus/common/sigv4 v0.1.0
 	github.com/prometheus/consul_exporter v0.8.0
 	github.com/prometheus/memcached_exporter v0.13.0
-	github.com/prometheus/mysqld_exporter v0.14.0
+	github.com/prometheus/mysqld_exporter v0.15.1
```
thampiotr, line 164: Any new config options that need to be exposed since it's a minor version, it can have new features?

internal/static/integrations/mysqld_exporter/mysqld-exporter.go
```diff
@@ -122,7 +123,7 @@ func New(log log.Logger, c *Config) (integrations.Integration, error) {
 	}
 
 	scrapers := GetScrapers(c)
-	exporter := collector.New(context.Background(), string(dsn), scrapers, log, collector.Config{
+	exporter := collector.New(context.Background(), string(dsn), scrapers, slog.Default(), collector.Config{
```
cristiangreco, line 126: @thampiotr `New` here now requires an `slog.Logger`, is there any adapter in the Alloy codebase?

internal/static/integrations/mysqld_exporter/mysqld-exporter.go
```diff
@@ -122,7 +124,8 @@ func New(log log.Logger, c *Config) (integrations.Integration, error) {
 	}
 
 	scrapers := GetScrapers(c)
-	exporter := collector.New(context.Background(), string(dsn), scrapers, log, collector.Config{
+	logger := slog.New(slogadapter.GoKitHandler{Logger: log, Group: ""})
```
wildum, line 127: I think that you could just cast the logger into our internal logger type `log.(*logging.Logger)` and then call the Handler() function to build the slog logger: `slog.New(l.Handler())`

internal/slogadapter/handler.go
thampiotr, line 1: Can we add a test for this?

internal/static/integrations/mysqld_exporter/mysqld-exporter.go
wildum: in this case, the slogadapter would not be needed

cristiangreco: Just tried that:

```
panic: interface conversion: log.Logger is *log.context, not *logging.Logger
```

Unfortunately it seems it's a go-kit logger rather than the internal `logging.Logger`.

wildum: Ah I see, it's because of the log.With here: https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/node_builtin_component.go#L172

wildum: this is similar to the problem that @dehaansa encountered in this support bundle PR
There are a few options:
- use the adapter that you made
- implement log.With in the logging.Logger to pass the logging.logger instead of the go-kit logger
- pass a slog logger instead of a gokit logger
- pass the logging.Logger additionally to the go-kit logger

cristiangreco: I stole the adapter in this PR from @aknuds1 's work at https://github.com/grafana/mimir/pull/9844/files#diff-36917e58779c0933db5147cf951d9c445a912ba808cf35c32ac951063b91f93f

Option 2 or option 3 that you mentioned above are more appealing though, as we would avoid relying on an adapter forever.

wildum: yeah the thing is that we want to migrate the code base to use slog everywhere. The idea behind the slogadapter that we already have is to be able to use slog in the areas where go-kit is still needed. It feels like the adapter that this PR is adding is going the wrong way, it's going from slog to go-kit and back to slog.
This is out of the scope of this PR so if that's ok I would suggest to pause it for now and that we find a better solution to pass the slog logger properly to the components that can use it directly

wildum: when do you need the update of the exporter in Alloy?

dehaansa:
https://github.com/grafana/alloy/issues/2173 should be resolved by this when its ready

CHANGELOG.md
```diff
@@ -46,6 +46,15 @@ Main (unreleased)
 
 - Fix a race condition where the ui service was dependent on starting after the remotecfg service, which is not guaranteed. (@dehaansa & @erikbaranowski)
 
+- Update mysqld_exporter from v0.15.0 to v0.16.0, most notable changes: (@cristiangreco)
```
cristiangreco, line 49: Note that `v0.12.2-0.20231005125903-364b9c41e595` we were using before was actually pointing at a commit just after 0.15.0.

cristiangreco: Is the changelog format ok? I picked only the changes I thought are most relevant for the embedded exporter.

go.mod
```diff
@@ -183,6 +183,7 @@ require (
 	github.com/stretchr/testify v1.9.0
 	github.com/testcontainers/testcontainers-go v0.33.0
 	github.com/tilinna/clock v1.1.0
+	github.com/tjhop/slog-gokit v0.1.2
```
cristiangreco, line 186: Mimir decided to use this library instead of a custom adapter, see https://github.com/grafana/mimir/pull/9844/commits/1d0ea03af41c042ed8ffe5605d0b95d164588d0c

CHANGELOG.md
wildum: the format is fine but it should not be in the ### Bugfixes area. I think that it can be in Enhancements or Other changes

go.mod
wildum: there is some level of wrapping but I think that it should work because the ref is the same. I tested it manually with your previous adapter and it was working.
The scenario Im thinking of would be:
- create this logger: https://github.com/grafana/alloy/blob/main/internal/runtime/logging/logger.go#L46
- wrap it in a go-kit with a log.With()
- use the adapter to transform it into a slog
- log with the slog logger on debug and on info (debug should drop, info should pass)
- call the update function on the logger that was created in the first step to change the level to debegu
- log with the slog logger on debug and on info (both should pass)

The key is to make sure that while Alloy is running, the logging level can be updated dynamically and that it's propagated through all the components.

I did not test the above, maybe you need a first call to the update right after the 1st step with the info level. let me know if something is not clear or not working

cristiangreco: As chatted, I've reintroduced the previous adapter and made some small refactoring to it, including tests.

wildum (approved):
LGTM, the implementation follows the one from https://github.com/tjhop/slog-gokit/blob/main/handler.go that Mimir is using besides the necessary small changes that are needed to make it work with the Alloy internal logger

PR #1666, opened 2024-09-11
Adding the serialization features.

mattdurham:
This adds the serialization side of converting series to the binary format. The binary format is time series where the strings are deduplicated with actual marshalling handled by the msgp library. I tested roughly 6 libraries listed [here](https://alecthomas.github.io/go_serialization_benchmarks/). [Msgp](https://github.com/tinylib/msgp) hit the sweetspot of ease of use and size and features. Such as reusing arrays if they were passed in.


wildum (commented):
great work, I like the deduplication algo. I only did a first pass, I'm not yet familiar with the full picture

internal/component/prometheus/remote/queue/types/serialization.go
```diff
@@ -0,0 +1,286 @@
...
+	ts.Value = 0
+	ts.Hash = 0
+	ts.Histograms.Histogram = nil
+	ts.Histograms.FloatHistogram = nil
+	tsBinaryPool.Put(ts)
+}
+
+// DeserializeToSeriesGroup transforms a buffer to a SeriesGroup and converts the stringmap + indexes into actual Labels.
+func DeserializeToSeriesGroup(sg *SeriesGroup, buf []byte) (*SeriesGroup, []byte, error) {
+	buffer, err := sg.UnmarshalMsg(buf)
```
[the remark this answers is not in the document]
mattdurham: It should fixing now.

mattdurham:
Stats are protocol agnostic, so that if used in prometheus or otel environment they can add their own specific metrics and we dont define protocol specific in the lower level structs. The component in the last PR will expose prometheus compatible ones derived from the callback.

internal/component/prometheus/remote/queue/serialization/appender.go
```diff
@@ -0,0 +1,121 @@
...
+	endTime := time.Now().Unix() - int64(a.ttl.Seconds())
+	if t < endTime {
+		return ref, nil
+	}
+	ts := types.GetTimeSeriesBinary()
+	ts.Labels = l
+	ts.TS = t
+	ts.Value = v
+	ts.Hash = l.Hash()
+	err := a.s.SendSeries(a.ctx, ts)
```
thampiotr, line 52: Is it guaranteed that `ts` will be returned eventually to the object pool? Would we have a leak if, e.g. the component was removed from Alloy config? I don't see any issues, but would be nice to make this code a bit more clear that this is what's going on, with naming or comments.

internal/component/prometheus/remote/queue/serialization/appender.go
```diff
@@ -0,0 +1,121 @@
...
+		Name:  "__alloy_metadata_type__",
+		Value: string(m.Type),
+	})
+	combinedLabels = append(combinedLabels, labels.Label{
+		Name:  "__alloy_metadata_help__",
+		Value: m.Help,
+	})
+	combinedLabels = append(combinedLabels, labels.Label{
+		Name:  "__alloy_metadata_unit__",
+		Value: m.Unit,
```
thampiotr, line 116: These are used in more than one place, can we make them constants? Helps find usages and generally nice practice IMO.

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+	"time"
+
+	snappy "github.com/eapache/go-xerial-snappy"
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/types"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/vladopajic/go-actor/actor"
+)
+
+// serializer collects data from multiple appenders and will write them to file.Storage.
```
thampiotr, line 15:
```suggestion
// serializer collects data from multiple appenders in-memory and will periodically flush the data to file.Storage.
```

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+// serializer will trigger based on the last flush duration OR if it hits a certain amount of items.
```
thampiotr, line 16:
```suggestion
// serializer will flush based on configured time duration OR if it hits a certain number of items.
```

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+	// By pulling the config from the mailbox we ensure it does NOT need a mutex around access.
+	select {
+	case <-ctx.Done():
+		return actor.WorkerEnd
+	case cfg, ok := <-s.cfgInbox.ReceiveC():
+		if !ok {
+			return actor.WorkerEnd
+		}
+		s.maxItemsBeforeFlush = int(cfg.MaxSignalsInBatch)
+		s.flushFrequency = cfg.FlushFrequency
```
thampiotr, line 89: Could we simplify and use atomic variables instead of the `cfgInbox` etc? Or a single atomic variable to hold the SerializerConfig. Would make the code a bit simpler and probably similar performance.

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+		return actor.WorkerContinue
+	case item, ok := <-s.metaInbox.ReceiveC():
+		if !ok {
+			return actor.WorkerEnd
+		}
+		err := s.AppendMetadata(ctx, item)
+		if err != nil {
+			level.Error(s.logger).Log("msg", "unable to append metadata to serializer", "err", err)
+		}
+		return actor.WorkerContinue
```
thampiotr, line 114: Do we need these two cases? The inboxes are of the same type... the handling is almost the same... only error message is different. Maybe we can have just one inbox? I see that later in `store` you do split them, but maybe there's some way to simplify?

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+	// If we would go over the max size then send, or if we have hit the flush duration then send.
+	if len(s.meta)+len(s.series) >= s.maxItemsBeforeFlush {
+		return s.store(ctx)
+	} else if time.Since(s.lastFlush) > s.flushFrequency {
+		return s.store(ctx)
+	}
+	return nil
+}
+
+func (s *serializer) Append(ctx actor.Context, data *types.TimeSeriesBinary) error {
```
thampiotr, line 137: I would prefer to call this something more descriptive, like `storeInMem` - they also don't need to be exported AFAICT?

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+	// If we would go over the max size then send, or if we have hit the flush duration then send.
+	if len(s.meta)+len(s.series) >= s.maxItemsBeforeFlush {
+		return s.store(ctx)
+	} else if time.Since(s.lastFlush) > s.flushFrequency {
+		return s.store(ctx)
+	}
+	return nil
+}
+
+func (s *serializer) store(ctx actor.Context) error {
```
thampiotr, line 148: I'd call it `flush`

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (s *serializer) AppendMetadata(ctx actor.Context, data *types.TimeSeriesBinary) error {
+	s.meta = append(s.meta, data)
+	// If we would go over the max size then send, or if we have hit the flush duration then send.
+	if len(s.meta)+len(s.series) >= s.maxItemsBeforeFlush {
+		return s.store(ctx)
+	} else if time.Since(s.lastFlush) > s.flushFrequency {
+		return s.store(ctx)
+	}
```
thampiotr, line 133: Not sure it's required if we have `<-s.flushTestTimer.C`?
This can also lead to flushing twice when we don't need to.

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+		s.storeStats(err)
+		// Return series to the pool, this is key to reducing allocs.
+		types.PutTimeSeriesBinarySlice(s.series)
+		types.PutTimeSeriesBinarySlice(s.meta)
+		s.series = s.series[:0]
+		s.meta = s.series[:0]
+	}()
+
+	// This maps strings to index position in a slice. This is doing to reduce the file size of the data.
+	strMapToInt := make(map[string]uint32)
```
thampiotr, line 171:
```suggestion
	// This maps strings to index position in a slice. This is going to reduce the file size of the data.
	strToIndex := make(map[string]uint32)
```

internal/component/prometheus/remote/queue/types/serialization.go
```diff
@@ -0,0 +1,289 @@
...
+// LabelNames and LabelsValues that point to the index in Strings.
+// This deduplicates the strings and decreases the size on disk.
+type SeriesGroup struct {
+	Strings  []string
+	Series   []*TimeSeriesBinary
+	Metadata []*TimeSeriesBinary
+}
+
+type TimeSeriesBinary struct {
+	// Labels are not serialized to msgp but are passed in.
```
thampiotr, line 24:
```suggestion
	// Labels are not serialized to msgp, instead we store separately a dictionary of strings and use `LabelNames` and `LabelValues` to refer to the dictionary by ID.
```

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+	for i, ts := range s.meta {
+		ts.FillLabelMapping(strMapToInt)
+		group.Metadata[i] = ts
+	}
+
+	stringsSlice := make([]string, len(strMapToInt))
+	for stringValue, index := range strMapToInt {
+		stringsSlice[index] = stringValue
+	}
+	group.Strings = stringsSlice
```
thampiotr, line 185:
```suggestion
	dictionary := make([]string, len(strMapToInt))
	for stringValue, index := range strMapToInt {
		dictionary[index] = stringValue
	}
	group.dictionary = dictionary
```

I like to use the concept of dictionary here, or lookup table... it makes it easier to figure out what's going on.

internal/component/prometheus/remote/queue/serialization/serializer.go
```diff
@@ -0,0 +1,223 @@
...
+
+	buf, err := group.MarshalMsg(s.msgpBuffer)
```
thampiotr, line 187: Sooo... is it worth it to do the dictionary stuff? I guess yes, but on the other hand I know that compression algos would do something similar automatically, snappy [can refer to previous part of the data to reduce repetition](https://en.wikipedia.org/wiki/Snappy_(compression)#Example_of_a_compressed_stream).

internal/component/prometheus/remote/queue/serialization/serializer_bench_test.go
```diff
@@ -0,0 +1,112 @@
...
+	// This should be around 200-300 allocs 7m ns/op
+	series := getTimeSeries(b)
+	b.ResetTimer()
+	b.ReportAllocs()
+	logger := log.NewNopLogger()
+	for i := 0; i < b.N; i++ {
+		serial, _ := NewSerializer(types.SerializerConfig{
+			MaxSignalsInBatch: 1_000,
+			FlushFrequency:    1 * time.Second,
+		}, &fakeFileQueue{}, func(stats types.SerializerStats) {
```
thampiotr, line 40: Hmmm... don't think we need to benchmark NewSerializer.
Also, shouldn't we use `b.N` to determine number of series that will be passed to `SendSeries`?

thampiotr (commented):
First pass. Looking good :)

internal/component/prometheus/remote/queue/serialization/appender.go
```diff
@@ -0,0 +1,121 @@
...
+
+type appender struct {
+	ctx    context.Context
+	ttl    time.Duration
+	s      types.Serializer
+	logger log.Logger
+}
+
+func (a *appender) AppendCTZeroSample(ref storage.SeriesRef, l labels.Labels, t, ct int64) (storage.SeriesRef, error) {
+	// TODO @mattdurham figure out what to do here later. This mirrors what we do elsewhere.
```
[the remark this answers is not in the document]
mattdurham: Yes, this mirrors the comment in wal.go. Which is a todo Paulin added.

internal/component/prometheus/remote/queue/serialization/appender.go
mattdurham: It should be a required that all time series are returned. Though not in this PR this is checked in a future test via `OutStandingTimeSeriesBinary` atomic int. There are end to end tests that ensure at the end of the test this is zero.

Makefile
```diff
@@ -141,7 +141,7 @@ lint: alloylint
 # final command runs tests for all other submodules.
 test:
 	$(GO_ENV) go test $(GO_FLAGS) -race $(shell go list ./... | grep -v /integration-tests/)
-	$(GO_ENV) go test $(GO_FLAGS) ./internal/static/integrations/node_exporter ./internal/static/logs ./internal/component/otelcol/processor/tail_sampling ./internal/component/loki/source/file ./internal/component/loki/source/docker
+	$(GO_ENV) go test $(GO_FLAGS) ./internal/static/integrations/node_exporter ./internal/static/logs ./internal/component/otelcol/processor/tail_sampling ./internal/component/loki/source/file ./internal/component/loki/source/docker ./internal/component/prometheus/remote/queue/serialization
```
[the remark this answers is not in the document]
mattdurham: There is one test that will not be ran twice since I am accessing the var directly to test its value. The others will be ran, I could add the //go:build race to the others. Note most of our exclusions above have some tests that run twice.

internal/component/prometheus/remote/queue/serialization/appender.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Append metric
+func (a *appender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	// Check to see if the TTL has expired for this record.
+	endTime := time.Now().Unix() - int64(a.ttl.Seconds())
+	if t < endTime {
+		return ref, nil
+	}
+	ts := types.GetTimeSeriesBinary()
```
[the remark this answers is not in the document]
mattdurham: Add `Pool` into the name of the Get and Put functions.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: In this case we could since we don't generally take action on the config, though this mirrors what we do in other objects so I am inclined to leave it like this for consistency.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: Its possible, I think we can table this until the network PR. Metadata and time series have different write characteristics. Its a matter of where we want to split them. We could have one work stream and split them at the network level instead of splitting at the beginning like we do here.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: Should never lead to a double flush since the code is single threaded, but it makes it cleaner so will change.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: Lemme refactor this out entirely, and the func can be moved to the do work loop.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: Do you mean to use an actual map? Or a rename like above?

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: I hadnt tested since I used msgp so recreated a test, 100k metrics with 10 labels apiece. The name being relatively constant the the value being random. The mapping approach saves roughly 33%, the less random the data is the better savings. If we reduce the random string generation to only 5 characters its half the size of the raw dictionary.

```
//go:generate msgp
package main

import (
	"fmt"
	"math/rand"

	"github.com/golang/snappy"
)

func main() {
	metrics := make([]map[string]string, 0)
	// 100k metrics with 10 labels each
	for i := 0; i < 100_000; i++ {
		metrics = append(metrics, getLabels())

	}

	ss := &StringString{Labels: metrics}
	bb, err := ss.MarshalMsg(nil)
	if err != nil {
		panic(err)
	}
	out := snappy.Encode(nil, bb)
	println(fmt.Printf("dictionary based is %d bytes", len(out)))

	ib := &IndexBased{
		String: make([]string, 0),
		Labels: make([]map[uint32]uint32, 0),
	}
	alignIndexBased(ib, metrics)
	bb, err = ib.MarshalMsg(nil)
	if err != nil {
		panic(err)
	}
	out = snappy.Encode(nil, bb)
	println(fmt.Printf("index based is %d bytes", len(out)))

}

type StringString struct {
	Labels []map[string]string
}

type IndexBased struct {
	String []string
	Labels []map[uint32]uint32
}

func alignIndexBased(ib *IndexBased, strings []map[string]string) {
	index := 0
	stringsList := make(map[string]int)
	for _, metric := range strings {
		mapping := make(map[uint32]uint32)
		for k, v := range metric {
			keyIndex, ok := stringsList[k]
			if !ok {
				stringsList[k] = index
				ib.String = append(ib.String, k)
				keyIndex = index
				index++
			}
			valIndex, ok := stringsList[v]
			if !ok {
				stringsList[k] = index
				ib.String = append(ib.String, v)
				valIndex = index
				index++
			}
			mapping[uint32(keyIndex)] = uint32(valIndex)
		}
		ib.Labels = append(ib.Labels, mapping)

	}

}

func getLabels() map[string]string {
	retLbls := make(map[string]string, 0)
	for i := 0; i < 10; i++ {
		retLbls[fmt.Sprintf("label_%d", i)] = randString()
	}
	return retLbls
}

var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")

func randString() string {
	b := make([]rune, rand.Intn(20))
	for i := range b {
		b[i] = letterRunes[rand.Intn(len(letterRunes))]
	}
	return string(b)
}
 
```

internal/component/prometheus/remote/queue/serialization/serializer_bench_test.go
mattdurham: I was worried about the overhead of creating a series, but seems minimal. Mostly concerned with allocs since that tends to be the cpu killer for any implementation of the wal. Changed to be more accurate of a benchmark, good call.

internal/component/prometheus/remote/queue/serialization/serializer.go
mattdurham: In general the index based is never worse, and IMO in many cases is 60% of the size of the pure string based.
Results from the above test, changing out the letterRunes to small sets.

// 5 char long really random
// 5371616
// 4732108

// 5 long half random
// 5050060
// 3929185

// 5 long quarter random
// 4455979
// 2918973

mattdurham: The lower the cardinality the better but even in worse case its not terrible.

mattdurham:
Going to merge this and we can revisit any followup in the big merge on specific points.

PR #1668, opened 2024-09-11
Fix broken windows unit tests

ptodev:
There were a couple of tests which were [failing](https://drone.grafana.net/grafana/alloy/2903/2/2) on the main branch since #1527 was merged.

One test was about pointer reuse:
```
--- FAIL: TestSetDefault_NoPointerReuse (0.00s)
    --- FAIL: TestSetDefault_NoPointerReuse/prometheus.exporter.windows (0.00s)
        all_test.go:27: 
            	Error Trace:	C:/drone/src/internal/component/all/all_test.go:61
            	            				C:/drone/src/internal/component/all/all_test.go:27
            	Error:      	Detected SetToDefault pointer reuse at github.com/grafana/alloy/internal/component/prometheus/exporter/windows.Arguments.Dfsr.SourcesEnabled
            	Test:       	TestSetDefault_NoPointerReuse/prometheus.exporter.windows
            	Messages:   	Types implementing syntax.Defaulter must not reuse pointers across multiple calls. Doing so leads to default values being changed when unmarshaling configuration files. If you're seeing this error, check the path above and ensure that copies are being made of any pointers in all instances of SetToDefault calls where that field is used.
FAIL
FAIL	github.com/grafana/alloy/internal/component/all	1.160s
```

The other failing tests were for the converters. The integration tests wanted to add `smb` and `smb_client` Alloy config, because the `split` function would output a `[""]` instead of `[]`, and the converter tests thought that Alloy's defaults are different from static mode's defaults.

Btw static mode doesn't even support these `smb` and `smb_client` parameters... But I'm not 100% how to ignore them right now, so I'll leave this for another day 😄  It's not really important, since no one who uses static mode would set them in their config anyway.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
@@ -75,7 +75,7 @@ text_file      | [text_file][]      | Configures the text_file collector.      |
 [scheduled_task]: #scheduledtask-block
 [service]: #service-block
 [smb]: #smb-block
-[smb_client]: #smbclient-block
+[smb_client]: #smb_client-block
 [smtp]: #smtp-block
 [text_file]: #textfile-block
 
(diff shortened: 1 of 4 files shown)


mattdurham (approved):
LGTM

PR #1674, opened 2024-09-12
Add  lifecycle and terminationGracePeriodSeconds for daemonset

jianpingzhangBill:
- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/containers/_agent.yaml
+++ b/operations/helm/charts/alloy/templates/containers/_agent.yaml
@@ -57,6 +57,10 @@
       scheme: {{ $values.listenScheme }}
     initialDelaySeconds: 10
     timeoutSeconds: 1
+  {{- with $values.lifecycle }}
+  lifecycle:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   {{- with $values.resources }}
   resources:
     {{- toYaml . | nindent 4 }}
--- a/operations/helm/charts/alloy/templates/controllers/daemonset.yaml
+++ b/operations/helm/charts/alloy/templates/controllers/daemonset.yaml
@@ -22,4 +22,7 @@ spec:
   updateStrategy:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- with .Values.controller.terminationGracePeriodSeconds }}
+  terminationGracePeriodSeconds: {{ . }}
+  {{- end }}
 {{- end }}
--- a/operations/helm/charts/alloy/templates/controllers/daemonset.yaml
+++ b/operations/helm/charts/alloy/templates/controllers/daemonset.yaml
@@ -22,7 +22,7 @@ spec:
   updateStrategy:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- with .Values.controller.terminationGracePeriodSeconds }}
-  terminationGracePeriodSeconds: {{ . }}
+  {{- if .Values.controller.terminationGracePeriodSeconds }}
+  terminationGracePeriodSeconds: {{ .Values.controller.terminationGracePeriodSeconds | int }}
   {{- end }}
 {{- end }}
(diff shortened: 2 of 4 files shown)


mattdurham:
This feels reasonable but would love people more familiar to give some input @petewall  or @captncraig

operations/helm/charts/alloy/templates/containers/_agent.yaml
```diff
@@ -57,6 +57,10 @@
       scheme: {{ $values.listenScheme }}
     initialDelaySeconds: 10
     timeoutSeconds: 1
+  {{- with $values.lifecycle }}
+  lifecycle:
```
petewall, line 61: Is this supposed to set a `lifecycle` on the Pod spec?

operations/helm/charts/alloy/templates/controllers/daemonset.yaml
```diff
@@ -22,4 +22,7 @@ spec:
   updateStrategy:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- with .Values.controller.terminationGracePeriodSeconds }}
+  terminationGracePeriodSeconds: {{ . }}
+  {{- end }}
```
petewall, line 27:
```suggestion
  {{- if .Values.controller.terminationGracePeriodSeconds }}
  terminationGracePeriodSeconds: {{ .Values.controller.terminationGracePeriodSeconds | int }}
  {{- end }}
```
Force to an integer.

operations/helm/charts/alloy/values.yaml
```diff
@@ -269,7 +271,9 @@ controller:
 
   # -- Additional containers to run alongside the Alloy container and initContainers.
   extraContainers: []
-
+  # -- terminationGracePeriodSeconds 
+  terminationGracePeriodSeconds: {}
```
petewall, line 275:
```suggestion
  terminationGracePeriodSeconds: 
```
This is an integer, but by setting the default to `{}` it seems like it's supposed to be an object.

operations/helm/charts/alloy/templates/containers/_agent.yaml
petewall:
```
% kubectl explain pod.spec.lifecycle
KIND:       Pod
VERSION:    v1

error: field "lifecycle" does not exist
```

jianpingzhangBill: ➜  ~ `kubectl explain pod.spec.containers.lifecycle`
KIND:       Pod
VERSION:    v1

FIELD: lifecycle <Lifecycle>


DESCRIPTION:
    Actions that the management system should take in response to container
    lifecycle events. Cannot be updated.
    Lifecycle describes actions that the management system should take in
    response to container lifecycle events. For the PostStart and PreStop
    lifecycle handlers, management of the container blocks until the action is
    complete, unless the container process fails, in which case the handler is
    aborted.

FIELDS:
  postStart	<LifecycleHandler>
    PostStart is called immediately after a container is created. If the handler
    fails, the container is terminated and restarted according to its restart
    policy. Other management of the container blocks until the hook completes.
    More info:
    https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks

  preStop	<LifecycleHandler>
    PreStop is called immediately before a container is terminated due to an API
    request or management event such as liveness/startup probe failure,
    preemption, resource contention, etc. The handler is not called if the
    container crashes or exits. The Pod's termination grace period countdown
    begins before the PreStop hook is executed. Regardless of the outcome of the
    handler, the container will eventually terminate within the Pod's
    termination grace period (unless delayed by finalizers). Other management of
    the container blocks until the hook completes or until the termination grace
    period is reached. More info:
    https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks

jianpingzhangBill (commented):
fixed.

PR #1677, opened 2024-09-12
Upgrade go to 1.22.7 and use corresponding build image

wildum:
It's a jump from v0.1.3 to v0.1.6 for the build image because there were a few fails related to the Go version 1.23.1 that did not work


mattdurham (approved):
lgtm

PR #1684, opened 2024-09-13
Update GitHub workflows to "latest" Ubuntu

ptodev:


.github/workflows/check-linux-build-image.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   check-linux-build-image:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
wildum, line 10: not super convinced about switching to "latest". I had bad experiences using mimir:latest in integration tests when they pushed a "latest" tag that was actually pointing to an old version of Mimir and it broke the tests.
If ubuntu pushes an image that contains breaking changes to our pipeline, it would cripple us and would be annoying to fix

ptodev: We already use `ubuntu-latest` on most of the GitHub runners. The ones being updated in this PR are the minority which use an explicit version.

Btw, this is not a Docker tag :) It's a GitHub runner tag:
https://github.com/actions/runner-images?tab=readme-ov-file#available-images

Atm, it corresponds to `ubuntu-22.04`. If there are any problems, I suppose we could just revert to using a fixed version in the future.

PR #1692, opened 2024-09-16
feat: Add support of multiple kind of cache for relabeling components

pbailhache:
#### PR Description

Related to proposal introduced by https://github.com/grafana/alloy/pull/1600. 

This is a (working) draft for this feature.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/faro/receiver/server.go
```diff
@@ -80,7 +80,6 @@ func (s *server) Run(ctx context.Context) error {
 	})
 
 	mw := middleware.Instrument{
-		RouteMatcher:     r,
```
pbailhache, line 83: Not sure about the impact of this change (This is related to the update of dskit : https://github.com/grafana/dskit/commit/27d7d41066d39e4ee64a0ae8a964518bac1b9f1e)

internal/component/prometheus/relabel/relabel.go
```diff
@@ -47,22 +55,40 @@ type Arguments struct {
...
 func (arg *Arguments) Validate() error {
-	if arg.CacheSize <= 0 {
-		return fmt.Errorf("max_cache_size must be greater than 0 and is %d", arg.CacheSize)
+	switch arg.CacheConfig.Backend {
+	case cache.InMemory:
+		if arg.CacheConfig.InMemory.CacheSize <= 0 {
+			return fmt.Errorf("cache_size must be greater than 0 and is %d", arg.CacheConfig.InMemory.CacheSize)
+		}
+	case cache.Memcached:
+		// todo
```
pbailhache, line 85: Need to determine what to include here, maybe move this to the service/cache package ?

internal/service/cache/cache.go
```diff
@@ -0,0 +1,118 @@
...
+	InMemory  InMemoryCacheConfig `alloy:"inmemory,block,optional"`
+}
+
+//TODO Those field are copied from dskit/cache for now (only the one mandatory)
+// We need to have a better way to manage conf
+// For now I used those because we cannot embed 'yaml' tags into alloy configuration
+// Ideally we should be using the dskit/cache conf directly, but it means it should not
+// be into the alloy configuration ?
+
+type RedisConf struct {
```
pbailhache, line 45: As explained in the comment, I'm open to suggestion here on how to handle the config part.

For now each cache is configured at the relabel component level but this include copying the struct to add the `alloy` tags as we cannot embed the dskit/cache configs into the grafana alloy config.

If we decide to move this config outside of the alloy config we could directly use the dskit/cache config.

internal/service/cache/cache_memcached.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/cache"
+)
+
+type MemcachedCache[valueType any] struct {
+	client *cache.MemcachedClient
+}
+
+func newMemcachedCache[valueType any](cfg MemcachedConfig) (*MemcachedCache[valueType], error) {
+	client, err := cache.NewMemcachedClientWithConfig(
```
pbailhache, line 20: Some things to add here (same in the other implementation)

internal/service/cache/cache_memcached.go
```diff
@@ -0,0 +1,139 @@
...
+
+	}
+
+	return result, nil
+}
+
+func (c *MemcachedCache[valueType]) Remove(key string) {
+	ctx := context.Background()
+	//TODO manage error
+	_ = c.client.Delete(ctx, key)
```
pbailhache, line 93: We ignore error for now but this isn't ideal

internal/service/cache/cache_redis.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	}
+
+	return result, nil
+}
+
+func (c *RedisCache[valueType]) Remove(key string) {
+	ctx := context.Background()
+	//TODO manage error
```
pbailhache, line 93: Ignoring error is never ideal

pbailhache (commented):
Some comments to open discussion to improve this code

internal/service/cache/cache.go
```diff
@@ -0,0 +1,118 @@
...
+	// requirepass server configuration option (if connecting to a Redis 5.0 instance, or lower),
+	// or the User Password when connecting to a Redis 6.0 instance, or greater,
+	// that is using the Redis ACL system.
+	Password string `alloy:"password,attr,optional"`
+
+	// DB Database to be selected after connecting to the server.
+	DB int `alloy:"db,attr"`
+
+	// MaxAsyncConcurrency specifies the maximum number of SetAsync goroutines.
+	MaxAsyncConcurrency int `yaml:"max_async_concurrency" category:"advanced"`
```
psauvage0, line 64: Are there some decent default values that we can chose for this and all the *BufferSize ?

internal/service/cache/cache.go
```diff
@@ -0,0 +1,118 @@
...
+	Username string `alloy:"username,attr"`
+
+	// Optional password. Must match the password specified in the
+	// requirepass server configuration option (if connecting to a Redis 5.0 instance, or lower),
+	// or the User Password when connecting to a Redis 6.0 instance, or greater,
+	// that is using the Redis ACL system.
+	Password string `alloy:"password,attr,optional"`
+
+	// DB Database to be selected after connecting to the server.
+	DB int `alloy:"db,attr"`
```
psauvage0, line 61: I don't understand what this does, but the call to newRedisCache always set this value to 0, which makes this config field ineffectual

internal/service/cache/cache.go
```diff
@@ -0,0 +1,118 @@
...
+	SetMultiple(values map[string]*valueType, ttl time.Duration) error
+	Remove(key string)
+	Clear(newSize int) error
+	GetCacheSize() int
+}
+
+// NewCache creates a new cache based on the given configuration
+func NewCache[valueType any](cfg CacheConfig) (Cache[valueType], error) {
+	switch cfg.Backend {
+	case InMemory:
```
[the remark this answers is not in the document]
pbailhache: We can select one backend per relabel component, each cache is independently configured. We could have one in-memory and one redis side by side for two different relabels.

internal/service/cache/cache_inmemory.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"sync"
+	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+type InMemoryCache[valueType any] struct {
+	lru       *lru.Cache[string, *valueType]
```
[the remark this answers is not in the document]
pbailhache: Previous cache LRU so I kept it this way, I think it's out of scope

internal/service/cache/cache.go
pbailhache: That's where you want to store your data, kind of a schema in a classic relational database.
So I set it to 0 (could leave it blank).

internal/service/cache/cache.go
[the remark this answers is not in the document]
psauvage0: OK, I re-read how CacheConfig is made. For some reason I thought that you could put multiple backend in the config  ^^'
No issue here then

internal/service/cache/cache_redis.go
```diff
@@ -0,0 +1,140 @@
...
+		//TODO NewLogFmtLogger ? Maybe something else
+		log.NewLogfmtLogger(os.Stdout),
+		"redis-cache",
+		cache.RedisClientConfig{
+			Endpoint:            []string{cfg.Endpoint.String()},
+			Username:            "default",
+			Password:            flagext.SecretWithValue(""),
+			MaxAsyncConcurrency: cfg.MaxAsyncConcurrency,
+			MaxAsyncBufferSize:  cfg.MaxAsyncBufferSize,
+			DB:                  0,
```
psauvage0, line 31: Why not set DB to cfg.DB here ? Default value would still be 0 and it would be possible to configure

internal/service/cache/cache.go
psauvage0: See my comment on newRedisCache function in cache_redis.go

internal/component/prometheus/relabel/relabel.go
```diff
@@ -230,7 +264,13 @@ func (c *Component) Update(args component.Arguments) error {
 	defer c.mut.Unlock()
 
 	newArgs := args.(Arguments)
-	c.clearCache(newArgs.CacheSize)
+
+	// todo maybe recreate whole relabelCache here in case of change for redis/memcached client
```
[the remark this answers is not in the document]
mattdurham: I think this would be a no-op, generally this is used to reset the cache size. In the case if redis/memcache I imagine this would not even be called. IE the cache size should not exist within Alloy.

internal/component/prometheus/relabel/relabel.go
```diff
@@ -47,22 +55,40 @@ type Arguments struct {
...
 }
 
 // SetToDefault implements syntax.Defaulter.
 func (arg *Arguments) SetToDefault() {
 	*arg = Arguments{
-		CacheSize: 100_000,
+		CacheConfig: cache.CacheConfig{
+			Backend: cache.InMemory,
+			InMemory: cache.InMemoryCacheConfig{
+				CacheSize: 100_100,
```
mattdurham, line 71: Typo here?

mattdurham:
Before we get to far down this path, I have a few longer term concerns about using this while we still have the handling of a single metric at a time. In an ideal world the Appender a batch based interface so we can batch the cache requests but thats a bigger lift.

wilfriedroset:
Thank you @mattdurham  for your review and feedbacks. regarding your concerns

> about using this while we still have the handling of a single metric at a time. In an ideal world the Appender a batch based interface so we can batch the cache requests but thats a bigger lift.

Do you think it would be possible to split the two needs? First we introduce the external caches via this PR, then after we take time to design and implement the batching to optimize further the performance.

As you have pointed, batching the requests will require more work. 
The reason why I'm suggesting that is because the current inmemory cache is the default and will continue to work as it is now with the same behavior/performance.
The performance improvement added by the refactoring will benefit the external caches only so there is no risks.
When I'm saying later, this is something we are willing to contribute but we might require help due to the complexity.

WDYT?

internal/service/cache/cache.go
```diff
@@ -0,0 +1,118 @@
+package cache
+
+import (
+	"time"
+
+	"github.com/grafana/dskit/cache"
```
mattdurham, line 6: We want to avoid pulling in dskit.

internal/service/cache/cache_inmemory.go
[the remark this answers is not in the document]
mattdurham: LRU should be kept as is, if we want to change that behavior should be a separate PR.

pbailhache:
I made a bench branch, you can see the commit here : 
https://github.com/pbailhache/alloy/commit/95098e04fff2b249910eb67e2fc4d639f8f8e38a

Here are the result of the benchmark on my laptop, running with dockertest.
benchmark

It's a bit better with a running docker container outside of the benchmark (around 65000ns/op for redis).
I used this benchmark to optimize the encode/decode for redis & memcached, switching from gob to json as it a bit quicker for small structs.

Here is the 3 flamegraphs for each backend, as we can see, the Get calls to the redis/memcached are what makes it slow. 
LRU :
LRU

Redis
Redis

Memcached
Memcached

Here I'm still using the deskit/cache package, I chose to use it because I did not want to re-implement the whole cache clients system. Do you think it's better to not use it here, I could implement a simpler version of each client and bench that, but I don't think the difference will be huge.

mattdurham:
Actually we are already using some dskit structs in a few places, will review your imports. We recently found an issue with exemplars that will likely require changes to the appender/appender interface that will batch the samples which should significantly improve the viability of this. I would hold off until we get that out, which will likely be in two weeks.

PR #1698, opened 2024-09-18
Add file path join to stdlib

wildum:
#### PR Description

Adding this function to the stdlib is needed to move forward with the [relative path solution](https://github.com/grafana/alloy/pull/789) that is implemented for the modules. This function should also be useful in other scenarios.

I put this function in a new namespace called "file". LMK if you would prefer this function to be in another namespace or without a namespace.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/stdlib/file.md
+++ b/docs/sources/reference/stdlib/file.md
@@ -11,7 +11,7 @@ The `file` namespace contains functions related to files.
 
 ## file.path_join
 
-The `file.path_join` function joins any number of path elements into a single path, separating them with an OS specific separator.
+The `file.path_join` function joins any number of path elements into a single path, separating them with an OS-specific separator.
 
 ### Examples
 
--- a/docs/sources/reference/stdlib/file.md
+++ b/docs/sources/reference/stdlib/file.md
@@ -1,0 +1,24 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/stdlib/file/
+description: Learn about file functions
+menuTitle: file
+title: file
+---
+
+# file
+
+The `file` namespace contains functions related to files.
+
+## file.path_join
+
+The `file.path_join` function joins any number of path elements into a single path, separating them with an OS specific separator.
+
+### Examples
+
+```
+> file.path_join()
+""
+
+> file.path_join("this/is", "a/path")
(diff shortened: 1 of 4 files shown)


tpaschalis (approved):

syntax/vm/vm_stdlib_test.go
```diff
@@ -215,6 +215,30 @@ func TestStdlib_StringFunc(t *testing.T) {
...
 }
 
+func TestStdlibFileFunc(t *testing.T) {
+	tt := []struct {
+		name   string
+		input  string
+		expect interface{}
+	}{
+		{"file.path_join", `file.path_join("this/is", "a/path")`, "this/is/a/path"},
+		{"file.path_join empty", `file.path_join()`, ""},
```
thampiotr, line 225: Is this going to work on Windows where paths use \ ?

wildum: yeah it should, it uses the filepath.Join function which is using an OS specific separator (https://pkg.go.dev/path/filepath#Join)

docs/sources/reference/stdlib/file.md
```diff
@@ -0,0 +1,24 @@
...
+title: file
+---
+
+# file
+
+The `file` namespace contains functions related to files.
+
+## file.path_join
+
+The `file.path_join` function joins any number of path elements into a single path, separating them with an OS specific separator.
```
clayton-cornell, line 14:
```suggestion
The `file.path_join` function joins any number of path elements into a single path, separating them with an OS-specific separator.
```

docs/sources/reference/stdlib/file.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Examples
+
+```
+> file.path_join()
+""
+
+> file.path_join("this/is", "a/path")
```
clayton-cornell, line 22:
```suggestion
file.path_join()
""

file.path_join("this/is", "a/path")
```

Is the greater-than symbol representing the OK command line or part of the actual command.

wildum: we use it in the other stdlib doc, I guess it's just a style to show input and output

PR #1699, opened 2024-09-18
Add beyla integration test

wildum:
I reworked the integration tests to use test containers instead of running Alloy's binary on the host directly for two reasons:
- some components (such as the Beyla one) require rights that the Github runner's host does not have. This is solved by running Alloy in test containers with privileges.
- some components only run on specific linux (for example Beyla, Unix, SNMP). Now that Alloy is running in a test container, the tests don't depend on the OS anymore

Additionally I added a test for Beyla

The annoying bit is that building the Alloy docker image takes time and disk space. As a result, I needed to use a bigger github runner for this task


go.mod
```diff
@@ -812,10 +811,52 @@ require (
 
 )
 
+require github.com/testcontainers/testcontainers-go/modules/compose v0.33.0
```
ptodev, line 814: It'd be nice to put this in the main `require` block.

ptodev (approved):
> The annoying bit is that building the Alloy docker image takes time and disk space. As a result, I needed to use a bigger github runner for this task

Can the integration tests reuse the docker image which the "official" CI step for image generation created?

marctc:
closing, as it's already done in another PR

PR #1704, opened 2024-09-18
Use centralized 'Add to docs project' workflow with GitHub App auth

jdbaldry:
Until this is merged, it is likely that any issues labeled `type/docs` won't be added to the [organization project](https://github.com/orgs/grafana/projects/69).

The underlying action is centralized so that any future changes are made in one place (`grafana/writers-toolkit`). The action is versioned to protect workflows from breaking changes.

The action uses Vault secrets instead of the discouraged organization secrets.

The workflow uses a consistent name so that future changes can be made programmatically.

Relates to https://github.com/orgs/grafana/projects/279/views/9?pane=issue&itemId=44280262


tpaschalis (approved):

.github/workflows/add-to-docs-project.yml
```diff
@@ -0,0 +1,16 @@
...
+    types: [labeled]
+jobs:
+  main:
+    if: ${{ github.event.label.name == 'type/docs' }}
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
jdbaldry, line 15:
```suggestion
```

PR #1716, opened 2024-09-19
Fix logs cluster

mattdurham:
#### PR Description

The `loki.source.kubernetes` component uses the default clustering approach. Which adds things like ports, addresses and other things. When in reality we only care about 4 fields to uniquely identify a log location. This can cause duplication of log pulling. This limits the labels to the only ones that uniquely identify a log. This is copied by how the actual tailer deduplicates targets, unfortunately that comes after the cluster figures out how to allocate logs.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


internal/component/loki/source/kubernetes/kubetail/target.go
```diff
@@ -25,6 +25,21 @@ const (
 	kubePodUID           = "__meta_kubernetes_pod_uid"
 )
 
+var ClusteringLabels = []string{
```
ptodev, line 28: Do I understand correctly that these are the labels which uniquely identify a log? Are we sure that we didn't miss any? What if the user also has some other custom labels?

ptodev (commented):
Should we update the [docs](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.kubernetes/#clustering-block) to clarify how the feature works?

internal/component/loki/source/kubernetes/kubetail/target.go
mattdurham: Technically its just name, namespace, container name for the tailer. But we use UID to identify the pod in the positions file so I wanted to add that to ensure it is sharded to the correct pod.

docs/sources/reference/components/loki/loki.source.kubernetes.md
```diff
@@ -145,6 +145,20 @@ If {{< param "PRODUCT_NAME" >}} is _not_ running in clustered mode, then the blo
 `loki.source.kubernetes` collects logs from every target it receives in its
 arguments.
 
+Clustering only looks at the following labels for determining the shard key:
```
clayton-cornell, line 148:
```suggestion
Clustering looks at the following labels for determining the shard key:
```

I think we can drop the only here.

ptodev (approved):

PR #1717, opened 2024-09-19
Network wal pr

mattdurham:
These are the network items for the new wal. There are some similarities with the old wal, buildWriteRequest and queuing by hash. I have e2e tests that arent shown here, will see if I can distill them to more specific tests but figure we can go ahead and get started looking at the code.


internal/component/prometheus/remote/queue/network/manager.go
```diff
@@ -0,0 +1,171 @@
...
+		// it will stop the filequeue from feeding more.
+		inbox:       actor.NewMailbox[*types.TimeSeriesBinary](actor.OptCapacity(1)),
+		metaInbox:   actor.NewMailbox[*types.TimeSeriesBinary](actor.OptCapacity(1)),
+		configInbox: actor.NewMailbox[types.ConnectionConfig](),
+		stats:       seriesStats,
+		cfg:         cc,
+	}
+
+	// start kicks off a number of concurrent connections.
+	var i uint64
```
wildum, line 47: why is i initialized outside of the for loop?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+
+// trySend is the core functionality for sending data to a endpoint. It will attempt retries as defined in MaxRetryBackoffAttempts.
+func (l *loop) trySend(ctx context.Context) {
+	attempts := 0
+attempt:
+	start := time.Now()
+	result := l.send(ctx, attempts)
+	duration := time.Since(start)
+	l.statsFunc(types.NetworkStats{
+		SendDuration: duration,
```
wildum, line 126: should the sendDuration be set even if the send was unsuccessful?

internal/component/prometheus/remote/queue/network/stats.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+			Metadata: types.CategoryStats{
+				FailedSamples: getMetadataCount(series),
+			},
+		})
+	}
+
+}
+
+func getSeriesCount(tss []*types.TimeSeriesBinary) int {
```
wildum, line 85: to reduce verbosity and loops, you could make one function getCounts and call it at the top of the recordStats function:
histCount, seriesCount := getCounts(series)

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+	l.lastSend = time.Now()
+}
+
+// send is the main work loop of the loop.
+func (l *loop) send(ctx context.Context, retryCount int) sendResult {
+	result := sendResult{}
+	defer func() {
+		recordStats(l.series, l.isMeta, l.statsFunc, result, len(l.sendBuffer))
+	}()
+	var err error
```
wildum, line 173: err is not used

wildum (commented):
Thanks for all this work, I did a first pass but couldn't review everything yet, I will continue the next day

internal/component/prometheus/remote/queue/network/manager.go
```diff
@@ -0,0 +1,171 @@
...
+func (s *manager) updateConfig(cc types.ConnectionConfig) {
+	// No need to do anything if the configuration is the same.
+	if s.cfg.Equals(cc) {
+		return
+	}
+	s.cfg = cc
+	// TODO @mattdurham make this smarter, at the moment any samples in the loops are lost.
+	// Ideally we would drain the queues and re add them but that is a future need.
+	// In practice this shouldn't change often so data loss should be minimal.
+	// For the moment we will stop all the items and recreate them.
```
wildum, line 122: we should maybe log that the config is being updated and the loops are reset

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		ticker:         time.NewTicker(1 * time.Second),
+		buf:            proto.NewBuffer(nil),
+		sendBuffer:     make([]byte, 0),
+	}
+	l.req = &prompb.WriteRequest{
+		// We know BatchCount is the most we will ever send.
+		Timeseries: make([]prompb.TimeSeries, 0, cc.BatchCount),
+	}
+
+	return l
```
thampiotr, line 71: Seems like all this could be one `return &loop {...}` statement.

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		return actor.WorkerEnd
+	// Ticker is to ensure the flush timer is called.
+	case <-l.ticker.C:
+		if len(l.series) == 0 {
+			return actor.WorkerContinue
+		}
+		if time.Since(l.lastSend) > l.flushFrequency {
+			l.trySend(ctx)
+		}
+		return actor.WorkerContinue
```
thampiotr, line 105: Worth calling out that the smallest effective`flushFrequency` is 1s, because `ticker` is 1s.

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		l.sendingCleanup()
+		return
+	}
+	// This helps us short circuit the loop if we are stopping.
+	if l.stopCalled.Load() {
+		return
+	}
+	// Sleep between attempts.
+	time.Sleep(result.retryAfter)
+	goto attempt
```
thampiotr, line 148: Could we use a `for {}` instead of `goto`? It does look safe here, but we don't know how this software will evolve over time... and I generally resonate with the sentiment against using `goto` as [some raise here](https://www.reddit.com/r/golang/comments/grglzm/why_are_so_many_golang_programmers_against_using/).

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		SendDuration: duration,
+	})
+	if result.successful {
+		l.sendingCleanup()
+		return
+	}
+	if !result.recoverableError {
+		l.sendingCleanup()
+		return
+	}
```
thampiotr, line 135: We should log an error when failed? I don't see `result.err` being used anywhere.

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		Type:             prompb.MetricMetadata_MetricType(prompb.MetricMetadata_MetricType_value[strings.ToUpper(ts.Labels.Get(types.MetaType))]),
+		Help:             ts.Labels.Get(types.MetaHelp),
+		Unit:             ts.Labels.Get(types.MetaUnit),
+		MetricFamilyName: ts.Labels.Get("__name__"),
+	}, true
+}
+
+func retryAfterDuration(defaultDuration time.Duration, t string) time.Duration {
+	parsedTime, err := time.Parse(http.TimeFormat, t)
+	if err == nil {
```
thampiotr, line 365:
```suggestion
	if parsedTime, err := time.Parse(http.TimeFormat, t); err == nil {
```

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		result.err = err
+		result.networkError = true
+		result.recoverableError = true
+		result.retryAfter = l.cfg.RetryBackoff
+		return result
+	}
+	result.statusCode = resp.StatusCode
+	defer resp.Body.Close()
+	// 500 errors are considered recoverable.
+	if resp.StatusCode/100 == 5 || resp.StatusCode == http.StatusTooManyRequests {
```
thampiotr, line 222: In vanilla remote_write we had a setting to "retry on 429" - do you think we'd need it here?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+	err              error
+	successful       bool
+	recoverableError bool
+	retryAfter       time.Duration
+	statusCode       int
+	networkError     bool
+}
+
+func (l *loop) sendingCleanup() {
+	types.PutTimeSeriesSliceIntoPool(l.series)
```
thampiotr, line 161: I can't see where did we get it from the pool, the `l.series` seems to be created in the `DoWork` loop. But after every send we put it to the pool. 

Can't we just slice it to zero length with keeping the capacity?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		return actor.WorkerContinue
+	}
+}
+
+// trySend is the core functionality for sending data to a endpoint. It will attempt retries as defined in MaxRetryBackoffAttempts.
+func (l *loop) trySend(ctx context.Context) {
+	attempts := 0
+attempt:
+	start := time.Now()
+	result := l.send(ctx, attempts)
```
thampiotr, line 123: Thoughts about debug level logging of result?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+	if len(l.sendBuffer) == 0 {
+		var data []byte
+		var wrErr error
+		if l.isMeta {
+			data, wrErr = createWriteRequestMetadata(l.req, l.series, l.buf)
+		} else {
+			data, wrErr = createWriteRequest(l.req, l.series, l.externalLabels, l.buf)
+		}
+		if wrErr != nil {
+			result.err = wrErr
```
thampiotr, line 185: Should we wrap the error so that we know where it comes from?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+
+	if retryCount > 0 {
+		httpReq.Header.Set("Retry-Attempt", strconv.Itoa(retryCount))
+	}
+	ctx, cncl := context.WithTimeout(ctx, l.cfg.Timeout)
+	defer cncl()
+	resp, err := l.client.Do(httpReq.WithContext(ctx))
+	// Network errors are recoverable.
+	if err != nil {
+		result.err = err
```
thampiotr, line 213: Should we wrap the error so that we know where it comes from?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		result.recoverableError = true
+		result.retryAfter = l.cfg.RetryBackoff
+		return result
+	}
+	result.statusCode = resp.StatusCode
+	defer resp.Body.Close()
+	// 500 errors are considered recoverable.
+	if resp.StatusCode/100 == 5 || resp.StatusCode == http.StatusTooManyRequests {
+		result.retryAfter = retryAfterDuration(l.cfg.RetryBackoff, resp.Header.Get("Retry-After"))
+		result.recoverableError = true
```
thampiotr, line 224: Should we set `result.err` to something?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+
+	result.successful = true
+	return result
+}
+
+func createWriteRequest(wr *prompb.WriteRequest, series []*types.TimeSeriesBinary, externalLabels map[string]string, data *proto.Buffer) ([]byte, error) {
+	if cap(wr.Timeseries) < len(series) {
+		wr.Timeseries = make([]prompb.TimeSeries, len(series))
+	}
+	wr.Timeseries = wr.Timeseries[:len(series)]
```
thampiotr, line 246: The `wr.Timeseries` lives throughout the loop lifetime, but the `series` is returned to the pool and re-allocated in `sendingCleanUp` - what's the idea here? why?

internal/component/prometheus/remote/queue/network/manager.go
```diff
@@ -0,0 +1,171 @@
...
+	s.stopLoops()
+	s.configInbox.Stop()
+	s.metaInbox.Stop()
+	s.inbox.Stop()
+	s.self.Stop()
+}
+
+func (s *manager) stopLoops() {
+	for _, l := range s.loops {
+		l.stopCalled.Store(true)
```
thampiotr, line 148: `stopCalled` is set inside `l.Stop()` already - I think we can have just the latter?

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		actor.New(l),
+		l.seriesMbx,
+	}
+}
+
+func (l *loop) DoWork(ctx actor.Context) actor.WorkerStatus {
+	// Main select loop
+	select {
+	case <-ctx.Done():
+		l.stopCalled.Store(true)
```
thampiotr, line 95: I don't think you need `stopCalled` here?

internal/component/prometheus/remote/queue/network/manager.go
```diff
@@ -0,0 +1,171 @@
...
+		l.Stop()
+	}
+	s.metadata.stopCalled.Store(true)
```
thampiotr, line 151: Similar to above comment about loops

thampiotr (commented):
First pass, didn't go through everything yet.

internal/component/prometheus/remote/queue/network/manager.go
mattdurham: Its because connection count is a uint64, I will switch to `i := uint64(0)` since that is clearer.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: That is how queue_manager currently works so was trying to mimic that behavior. https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L603

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+		ts.Labels = ts.Labels[:len(tsBuf.Labels)]
+		for k, v := range tsBuf.Labels {
+			ts.Labels[k].Name = v.Name
+			ts.Labels[k].Value = v.Value
+		}
+
+		// By default each sample only has a histogram, float histogram or sample.
+		if cap(ts.Histograms) == 0 {
+			ts.Histograms = make([]prompb.Histogram, 1)
```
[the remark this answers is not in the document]
mattdurham: I think functionally this is less ideal, since I would need to use append semantics. There will only ever be 0 or 1 record here.

internal/component/prometheus/remote/queue/network/loop.go
```diff
@@ -0,0 +1,374 @@
...
+			cnt++
+		}
+	}
+	return cnt
+}
+
+func isMetadata(ts *types.TimeSeriesBinary) bool {
+	return ts.Labels.Has(types.MetaType) &&
+		ts.Labels.Has(types.MetaUnit) &&
+		ts.Labels.Has(types.MetaHelp)
```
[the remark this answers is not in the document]
mattdurham: It really needs all to be valid metadata. Technically you could get away without help. In practice not having all three would be a bug.

Makefile
```diff
@@ -141,7 +141,7 @@ lint: alloylint
 # final command runs tests for all other submodules.
 test:
 	$(GO_ENV) go test $(GO_FLAGS) -race $(shell go list ./... | grep -v /integration-tests/)
-	$(GO_ENV) go test $(GO_FLAGS) ./internal/static/integrations/node_exporter ./internal/static/logs ./internal/component/otelcol/processor/tail_sampling ./internal/component/loki/source/file ./internal/component/loki/source/docker ./internal/component/prometheus/remote/queue/serialization
+	$(GO_ENV) go test $(GO_FLAGS) ./internal/static/integrations/node_exporter ./internal/static/logs ./internal/component/otelcol/processor/tail_sampling ./internal/component/loki/source/file ./internal/component/loki/source/docker ./internal/component/prometheus/remote/queue/serialization  ./internal/component/prometheus/remote/queue/network
```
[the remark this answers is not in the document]
mattdurham: Because I am reaching into the config to test for updating config which happens on a different thread than created it so the race detector is unhappy. I thought about making the ability to retrieve safely but was a lot of effort for one test for something unused in actual code.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: I will note it in the master PR component documentation that will be in the next PR.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: I think that is something we can look into, currently we default to true in remote_write vanilla. If we need to we can add a config to turn it off.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: TimeSeries are pulled from the pool in something we havent seen yet, at the component level. Conceptually the filequeue sends the bytes, the component transforms those bytes into time series from a pool and sends them to the manager.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: TimeSeries is a protobuf concept and only exist within the lifecycle of the loop and are reused. Whereas series are used by other parts of the full remote write and component.

mattdurham: TimeSeries is a loaded term, TimeSerieskbinary might be to close to the protobuf name so will consider if there is a better name.

internal/component/prometheus/remote/queue/network/manager.go
mattdurham: Good catch!

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: Correct, removing.

internal/component/prometheus/remote/queue/network/loop.go
mattdurham: I have added comments to the TimeSeriesBinary struct itself detailing to never instantiate it directly but only via the Get/Put functions.

Makefile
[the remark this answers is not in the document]
mattdurham: The manager code itself would need synchronization primitives in addition to the test which is not ideal. What do you mean by using function pointers?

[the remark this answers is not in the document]
mattdurham: Resolved this by ensuring the records coming in are written by the new batch size.

mattdurham:
Going to merge this and move on to the next PR.

bboreham:
I guess this relates to #370 ?

PR #1726, opened 2024-09-23
Modules support relative path 2

wildum:
#### PR Description

This PR introduces an ArgScope to the loader that will be used as a parent scope when evaluating the nodes in the module. For now, it contains only one variable "module_path" that represents the current path of the module and can be used in the config. Both can be used together to configure a relative path on the import.file block.

```alloy
import.file “default” {
  filename = file.path_join(module_path, "relative/path")
}
```

The following use-cases are currently supported:

import.file in a module imported via import.file (module_path = filename of the parent import.file)
import.file in a module imported via import.git (module_path = path to the local cloned repo)
import.file in a module imported via import.string (module_path = path of the current module)

Fixes https://github.com/grafana/alloy/issues/786

#### Notes to the Reviewer

This was an old [PR](https://github.com/grafana/alloy/pull/789) that I revived now that the stdlib has been updated.

I recommend reviewing commit by commit.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/runtime/alloy.go
```diff
@@ -296,22 +298,36 @@ func (f *Runtime) Run(ctx context.Context) {
...
+func (f *Runtime) LoadSource(source *Source, args map[string]any, configPath string) error {
+	return f.applyLoaderConfig(controller.ApplyOptions{
+		Args:            args,
+		ComponentBlocks: source.components,
+		ConfigBlocks:    source.configBlocks,
+		DeclareBlocks:   source.declareBlocks,
+		ArgScope: &vm.Scope{
+			Parent: nil,
+			Variables: map[string]interface{}{
+				"module_path": util.ExtractDirPath(configPath),
```
wildum, line 310: should probably use a const for "module_path"

docs/sources/reference/config-blocks/import.file.md
```diff
@@ -56,11 +74,71 @@ declare "add" {
...
+```
+
+{{< /collapse >}}
+
+{{< collapse title="relative_math.alloy" >}}
+
+```alloy
+import.file "lib" {
+  filename = file.path_join(module_path, "lib.alloy")
+}
```
thampiotr, line 101: Instead of the keyword, could we introduce a rule that imports from inside a module will be relative to that module's location?

In this example if you have `filename = "./lib.alloy"` it would automatically be relative to the checkout of the git repo. Users can still use absolute paths if they want a different behaviour.

docs/sources/reference/config-blocks/import.file.md
```diff
@@ -56,11 +74,71 @@ declare "add" {
...
+```
+
+{{< /collapse >}}
+
+{{< collapse title="relative_math.alloy" >}}
+
+```alloy
+import.file "lib" {
+  filename = file.path_join(module_path, "lib.alloy")
```
thampiotr, line 100: Do we need to ensure users cannot escape outside the sandbox and start accessing files outside of the git repo checkout in this example? What's been our threat model / security approach to modules?

thampiotr (commented):
Should this go through a proposal as a user-facing API?

I think there are different ways to design this that we'd want to consider before implementing.

thampiotr:
Another idea, which I think could be powerful and simple for users to work with: if a file has a relative path in it, it is relative to that file's location.

docs/sources/reference/config-blocks/import.file.md
wildum: this is proposal 4 in the design doc: https://docs.google.com/document/d/16vRoiIxQVVlqqEGngm27Kcun6hF_BTpY2tPOSt05JWU/edit#heading=h.qc2hfy5h8d7z
There are a few cons with this approach and consensus was reached with another approach but nothing is set in stone. If you think that the proposal should be re-evaluated we can discuss it again

docs/sources/reference/config-blocks/import.file.md
wildum: they can already do it. If an attacker accesses the imported module, he can modify it for malicious purposes. There were some discussions around module protection levels months ago but it was not prioritized

docs/sources/reference/config-blocks/import.file.md
thampiotr: Yeah, Proposal 4 is a design that I would personally prefer, but it's true that it could break some users. So it'd be something to do before 1.0, but now we have to maintain the existing behaviour. 

We could add a ticket for 2.0 to reconsider this.

I agree that the option you're implementing here is the next best thing.

docs/sources/reference/config-blocks/import.file.md
thampiotr: Yeah, I guess that's our approach - keep your config files safe from modification. 
I'd like to doc that: https://github.com/grafana/alloy/pull/1746/files

docs/sources/reference/config-blocks/import.file.md
wildum: opened an issue to reconsider our decision with a 2.0 label: https://github.com/grafana/alloy/issues/1752

internal/runtime/alloy_services.go
```diff
@@ -98,7 +98,8 @@ func (sc ServiceController) LoadSource(b []byte, args map[string]any) error {
 	if err != nil {
 		return err
 	}
-	return sc.f.LoadSource(source, args)
+	// The config is loaded via a service, the config path is left empty.
+	return sc.f.LoadSource(source, args, "")
```
thampiotr, line 102: If I use `module_path` in such config, what path would it be?

internal/runtime/internal/controller/loader.go
```diff
@@ -608,7 +614,7 @@ func (l *Loader) wireGraphEdges(g *dag.Graph) diag.Diagnostics {
 		}
 
 		// Finally, wire component references.
-		refs, nodeDiags := ComponentReferences(n, g, l.log)
+		refs, nodeDiags := ComponentReferences(n, g, l.log, l.cache.scope)
```
thampiotr, line 617: Nit: We're reading `l.cache.scope` without the mutex. It may be stale and come up in race detector, but I don't think it's dangerous.

internal/runtime/internal/importsource/import_http.go
```diff
@@ -106,3 +107,8 @@ func (im *ImportHTTP) CurrentHealth() component.Health {
 func (im *ImportHTTP) SetEval(eval *vm.Evaluator) {
 	im.eval = eval
 }
+
+func (im *ImportHTTP) ModulePath() string {
+	dir, _ := path.Split(im.arguments.URL)
+	return dir
+}
```
thampiotr, line 114: Hmmm.... shouldn't we return `""` because there's no filesystem path for HTTP?

internal/runtime/internal/importsource/import_git.go
```diff
@@ -287,3 +288,7 @@ func (im *ImportGit) CurrentHealth() component.Health {
 func (im *ImportGit) SetEval(eval *vm.Evaluator) {
 	im.eval = eval
 }
+
+func (im *ImportGit) ModulePath() string {
+	return im.repoPath
+}
```
thampiotr, line 294: Should we now document that the entire repository will be cloned and files in the git repo will be available to reference them using module_path?

internal/util/filepath.go
```diff
@@ -0,0 +1,13 @@
+package util
+
+import "path/filepath"
+
+// ExtractDirPath removes the file part of a path if it exists.
+func ExtractDirPath(p string) string {
+	// If the base of the path has an extension, it's likely a file.
+	if filepath.Ext(filepath.Base(p)) != "" {
```
thampiotr, line 8: I think it's likely that people will use a file without extension, e.g. `/etc/alloy_config` and this will not work for them.

I think instead we should `Stat` the path and check if it's directory.

thampiotr (commented):
A few small comments, but LGTM in general.

Before merging, would it make sense to share it with someone who uses modules extensively to test this build? Or check if it works with fleet management?

internal/runtime/internal/importsource/import_http.go
wildum: I wasn't sure what to return here because using import.file in import.http is not allowed (https://github.com/grafana/alloy/pull/1726/files#diff-69560fb45faf98ab9023ff409c97649c2de412f3be022e5b22b2c018c3170490R293). Technically module_path can be used anywhere in the module so I decided to return the URL path (could be set as a label for example). I don't think that it's very useful but I thought it might be better than ""

thampiotr: ok, that's fine I guess ;)

internal/runtime/alloy_services.go
wildum: I added support for it and made sure that it works with fleet management.
I did the following manual tests:
- remote config with import.git that imports a module that has import.file with relative path
- remote config with import.file that imports a module that has an import.file with relative path (it's relative to the path of the config where the remotecfg component is, not super useful I guess but that works and it's good for consistency)

internal/runtime/internal/importsource/import_git.go
wildum: added documentation: https://github.com/grafana/alloy/pull/1726/commits/133aeb018e84c84299c22bbf7a1d234a333b7f14

internal/util/filepath.go
wildum: I did the change: https://github.com/grafana/alloy/pull/1726/commits/3f08a9ef7bc1b75745a8b4bd885a34f57d11f4a5. The error should not happen because if the path is wrong then it will fail earlier but we use this func everywhere for testing so I only logged it with warn

internal/runtime/internal/controller/loader.go
wildum: add mutex rlock and runlock to protect it: https://github.com/grafana/alloy/pull/1726/commits/acf8ef67d4d529bd5b85b0126a8c5c3e7d2d430f

wildum:
> Before merging, would it make sense to share it with someone who uses modules extensively to test this build? 

@bentonam would be a good candidate

docs/sources/reference/config-blocks/import.file.md
```diff
@@ -41,6 +44,21 @@ The following arguments are supported:
 
 This example imports a module from a file and instantiates a custom component from the import that adds two numbers:
 
+{{< collapse title="main.alloy" >}}
```
clayton-cornell, line 47: I'm not sure that using collapse here is really the best way to present the examples. What's the reason for choosing collapse over something simpler like a heading?

image

thampiotr: +1 I like the expanded version. I don't think there are good rules on when to use collapsible.

docs/sources/reference/config-blocks/import.git.md
```diff
@@ -9,6 +9,9 @@ title: import.git
 The `import.git` block imports custom components from a Git repository and exposes them to the importer.
 `import.git` blocks must be given a label that determines the namespace where custom components are exposed.
 
+The entire repository is cloned, and the module path is accessible via the `module_path` keyword.
+This enables your module to import other modules within the repository by setting relative paths in the [import.file][] blocks.
```
thampiotr, line 13:
```suggestion
This enables, for example, your module to import other modules within the repository by setting relative paths in the [import.file][] blocks.
```
There's bunch of other things it can do using `local.file` or `discovery.file`, etc.

internal/runtime/alloy.go
```diff
@@ -296,22 +299,40 @@ func (f *Runtime) Run(ctx context.Context) {
...
+	modulePath, err := util.ExtractDirPath(configPath)
+	if err != nil {
+		level.Warn(f.log).Log("msg", "failed to extract directory path from configPath", "configPath", configPath, "err", err)
+	}
+	return f.applyLoaderConfig(controller.ApplyOptions{
+		Args:            args,
+		ComponentBlocks: source.components,
+		ConfigBlocks:    source.configBlocks,
+		DeclareBlocks:   source.declareBlocks,
+		ArgScope: &vm.Scope{
```
thampiotr, line 312: What are the rules for the `vm.Scope` now? Does it always need to have either a Parent or ModulePath set? 

I think the way it's instantiated now is error-prone. I would prefer to have specialised constructors in `vm` package that enforce correct creation of `vm.Scope` so that we don't make mistakes or have semantic merge conflicts that result in bugs.

thampiotr (approved):
I would like to see `vm.Scope` to have a constructor, but otherwise LGTM.

docs/sources/reference/config-blocks/import.file.md
clayton-cornell: Generally, the collapsed sections should be reserved for information that is extra to the flow... for example a legacy Agent Flow configuration. For examples I would prefer that we expand them out and use a more standard structured approach. There are multiple examples here, so the Example heading should be a little more descriptive. It's more work than I can put into the suggestions... so I can push a change to the PR or do you want to tweak things @wildum ?

wildum: I made a few changes, @clayton-cornell feel free to tweak it via suggestions or to push changes directly if that's better

internal/runtime/alloy.go
wildum: added via https://github.com/grafana/alloy/pull/1726/commits/d5436529460b9e6a282492af209024bf5070c416

docs/sources/reference/config-blocks/import.file.md
```diff
@@ -17,6 +17,9 @@ Imported directories are treated as single modules to support composability.
 That means that you can define a custom component in one file and use it in another custom component in another file
 in the same directory.
 
+You can use the keyword `module_path` in combination with the `stdlib` function [file.path_join][] to import a module relative to the current module's path.
+The `module_path` keyword works for modules that are imported via `import.file`, `import.git` and `import.string`.
```
clayton-cornell, line 21:
```suggestion
The `module_path` keyword works for modules that are imported via `import.file`, `import.git`, and `import.string`.
```

clayton-cornell (approved):
Looks OK in a local build. I may open a new PR to change the way we present the `main.alloy` , `lib.alloy`, etc. code blocks, but... it's good enough as it is right now for this PR.

thampiotr (approved):
LGTM

PR #1773, opened 2024-09-26
Support extra label selectors on mimir.rules.kubernetes

thampiotr:
#### PR Description

The goal is to allow mimir.rules.kubernetes to add extra matchers to queries, using the [same code as the promtool does here](https://github.com/prometheus/prometheus/blob/79a6238e195ecc1c20937036c1e3b4e3bdaddc49/cmd/promtool/main.go#L1242).

Internal request.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
+++ b/docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
@@ -50,25 +50,25 @@ mimir.rules.kubernetes "LABEL" {
 
 `mimir.rules.kubernetes` supports the following arguments:
 
-Name                     | Type                | Description                                                                                      | Default       | Required
--------------------------|---------------------|--------------------------------------------------------------------------------------------------|---------------|---------
-`address`                | `string`            | URL of the Mimir ruler.                                                                          |               | yes
-`tenant_id`              | `string`            | Mimir tenant ID.                                                                                 |               | no
-`use_legacy_routes`      | `bool`              | Whether to use [deprecated][gem-2_2] ruler API endpoints.                                        | false         | no
-`prometheus_http_prefix` | `string`            | Path prefix for [Mimir's Prometheus endpoint][gem-path-prefix].                                  | `/prometheus` | no
-`sync_interval`          | `duration`          | Amount of time between reconciliations with Mimir.                                               | "5m"          | no
-`mimir_namespace_prefix` | `string`            | Prefix used to differentiate multiple {{< param "PRODUCT_NAME" >}} deployments.                  | "alloy"       | no
-`bearer_token_file`      | `string`            | File containing a bearer token to authenticate with.                                             |               | no
-`bearer_token`           | `secret`            | Bearer token to authenticate with.                                                               |               | no
-`enable_http2`           | `bool`              | Whether HTTP2 is supported for requests.                                                         | `true`        | no
-`follow_redirects`       | `bool`              | Whether redirects returned by the server should be followed.                                     | `true`        | no
-`proxy_url`              | `string`            | HTTP proxy to send requests through.                                                             |               | no
-`no_proxy`               | `string`            | Comma-separated list of IP addresses, CIDR notations, and domain names to exclude from proxying. |               | no
-`proxy_from_environment` | `bool`              | Use the proxy URL indicated by environment variables.                                            | `false`       | no
-`proxy_connect_header`   | `map(list(secret))` | Specifies headers to send to proxies during CONNECT requests.                                    |               | no
-`external_labels`        | `map(string)`       | Labels to add to each rule.                                                                      | `{}` | no
+| Name                     | Type                | Description                                                                                      | Default       | Required |
+|--------------------------|---------------------|--------------------------------------------------------------------------------------------------|---------------|----------|
+| `address`                | `string`            | URL of the Mimir ruler.                                                                          |               | yes      |
+| `tenant_id`              | `string`            | Mimir tenant ID.                                                                                 |               | no       |
+| `use_legacy_routes`      | `bool`              | Whether to use [deprecated][gem-2_2] ruler API endpoints.                                        | false         | no       |
+| `prometheus_http_prefix` | `string`            | Path prefix for [Mimir's Prometheus endpoint][gem-path-prefix].                                  | `/prometheus` | no       |
+| `sync_interval`          | `duration`          | Amount of time between reconciliations with Mimir.                                               | "5m"          | no       |
+| `mimir_namespace_prefix` | `string`            | Prefix used to differentiate multiple {{< param "PRODUCT_NAME" >}} deployments.                  | "alloy"       | no       |
+| `bearer_token_file`      | `string`            | File containing a bearer token to authenticate with.                                             |               | no       |
+| `bearer_token`           | `secret`            | Bearer token to authenticate with.                                                               |               | no       |
+| `enable_http2`           | `bool`              | Whether HTTP2 is supported for requests.                                                         | `true`        | no       |
+| `follow_redirects`       | `bool`              | Whether redirects returned by the server should be followed.                                     | `true`        | no       |
+| `proxy_url`              | `string`            | HTTP proxy to send requests through.                                                             |               | no       |
+| `no_proxy`               | `string`            | Comma-separated list of IP addresses, CIDR notations, and domain names to exclude from proxying. |               | no       |
+| `proxy_from_environment` | `bool`              | Use the proxy URL indicated by environment variables.                                            | `false`       | no       |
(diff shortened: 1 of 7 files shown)


internal/component/mimir/rules/kubernetes/events_test.go
```diff
@@ -251,6 +251,18 @@ func TestAdditionalLabels(t *testing.T) {
...
+				name:      "cluster",
+				matchType: "=",
+				value:     "big",
+			},
+			{
+				name:      "job",
+				matchType: "=",
+				value:     "good",
+			},
+		},
```
thampiotr, line 265: Users would be expected to set something like this in their alloy config for mimir.rules.kubernetes.

internal/component/mimir/rules/kubernetes/events_test.go
```diff
@@ -287,11 +299,11 @@ func TestAdditionalLabels(t *testing.T) {
 		expectedRule := `- name: group1
   rules:
   - alert: alert1
-    expr: expr1
+    expr: expr1{cluster="big",job="good"}
```
thampiotr, line 302: The queries are modified to include the extra labels.

internal/component/mimir/rules/kubernetes/events.go
```diff
@@ -183,14 +187,27 @@ func (e *eventProcessor) desiredStateFromKubernetes() (kubernetes.RuleGroupsByNa
...
-						if rule_group.Rules[i].Labels == nil {
-							rule_group.Rules[i].Labels = make(map[string]string, len(e.externalLabels))
+					for i := range ruleGroup.Rules {
+						if ruleGroup.Rules[i].Labels == nil {
+							ruleGroup.Rules[i].Labels = make(map[string]string, len(e.externalLabels))
+						}
+						maps.Copy(ruleGroup.Rules[i].Labels, e.externalLabels)
+					}
+				}
+			}
```
thampiotr, line 200: Just a rename of non-idiomatic rule_group variable here.

internal/component/mimir/rules/kubernetes/events.go
```diff
@@ -183,14 +187,27 @@ func (e *eventProcessor) desiredStateFromKubernetes() (kubernetes.RuleGroupsByNa
...
+			if e.extraQueryMatchers != nil {
+				for _, ruleGroup := range groups {
+					for i := range ruleGroup.Rules {
+						query := ruleGroup.Rules[i].Expr.Value
+						newQuery, err := addMatchersToQuery(query, e.extraQueryMatchers.Matchers)
+						if err != nil {
+							level.Error(e.logger).Log("msg", "failed to add labels to PrometheusRule query", "query", query, "err", err)
 						}
-						maps.Copy(rule_group.Rules[i].Labels, e.externalLabels)
+						ruleGroup.Rules[i].Expr.Value = newQuery
```
thampiotr, line 210: This is new.

internal/component/mimir/rules/kubernetes/events_test.go
```diff
@@ -298,3 +281,123 @@ func TestAdditionalLabels(t *testing.T) {
...
+	// The map of rules has only one element.
+	for ruleName, rule := range rules {
+		require.Equal(t, "alloy/namespace/name/64aab764-c95e-4ee9-a932-cd63ba57e6cf", ruleName)
+
+		ruleBuf, err := yaml.Marshal(rule)
+		require.NoError(t, err)
+
+		expectedRule := `- name: group1
+  rules:
+    - expr: "sum by (namespace) (rate(success{cluster=~\"prod-.*\",job=\"good\"}[10m]) / rate(total{cluster=~\"prod-.*\",job=\"good\"}[10m]))"
```
thampiotr, line 380: This shows how for queries that refer multiple metrics, the extra matchers will be added to all metric selectors.

internal/component/mimir/rules/kubernetes/rules_test.go
```diff
@@ -20,32 +20,90 @@ import (
...
+			name = "namespace"
+			match_type = "!~"
+			value = ".+"
+		}
+		matcher {
+			name = "cluster"
+			match_type = "=~"
+			value = "prod-.*"
+		}
+	}
```
thampiotr, line 72: Example of user-facing config.

docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
```diff
@@ -127,16 +129,38 @@ an `oauth2` block.
 [tls_config]: #tls_config-block
 [label_selector]: #label_selector-block
 [match_expression]: #match_expression-block
+[extra_query_matchers]: #extra_query_matchers-block
+[matcher]: #matcher-block
+
+### extra_query_matchers block
+
+The `extra_query_matchers` block has no attributes. It contains zero or more [matcher][] blocks.
+These blocks allow to add extra label matchers to all queries that are discovered by `mimir.rules.kubernetes`
```
clayton-cornell, line 138:
```suggestion
These blocks allow you to add extra label matchers to all queries that are discovered by `mimir.rules.kubernetes`
```

docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
```diff
@@ -127,16 +129,38 @@ an `oauth2` block.
...
+### matcher block
+
+The `matcher` block describes a label matcher that will be added to each query found in `PrometheusRule` CRDs.
+
+The following arguments are supported:
+
+| Name         | Type     | Description                                        | Default | Required |
+|--------------|----------|----------------------------------------------------|---------|----------|
+| `name`       | `string` | Name of the label to match.                        |         | yes      |
+| `match_type` | `string` | The type of match. One of `=`, `!=`, `=~` and `!~` |         | yes      |
```
clayton-cornell, line 151:
```suggestion
| `match_type` | `string` | The type of match. One of `=`, `!=`, `=~` and `!~`. |         | yes      |
```

docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
```diff
@@ -127,16 +129,38 @@ an `oauth2` block.
...
+
+### matcher block
+
+The `matcher` block describes a label matcher that will be added to each query found in `PrometheusRule` CRDs.
+
+The following arguments are supported:
+
+| Name         | Type     | Description                                        | Default | Required |
+|--------------|----------|----------------------------------------------------|---------|----------|
+| `name`       | `string` | Name of the label to match.                        |         | yes      |
```
wildum, line 150: wouldn't it be more clear to call this `label`?

internal/component/mimir/rules/kubernetes/events.go
```diff
@@ -202,6 +219,62 @@ func (e *eventProcessor) desiredStateFromKubernetes() (kubernetes.RuleGroupsByNa
...
+		matchType = promlabels.MatchEqual
+	case parser.ItemType(parser.NEQ).String():
+		matchType = promlabels.MatchNotEqual
+	case parser.ItemType(parser.EQL_REGEX).String():
+		matchType = promlabels.MatchRegexp
+	case parser.ItemType(parser.NEQ_REGEX).String():
+		matchType = promlabels.MatchNotRegexp
+	default:
+		return query, fmt.Errorf("invalid label match type: %s", labelMatchType)
+	}
```
wildum, line 252: this could be checked at the Validate step to exit early if a match block is not configured with a valid match type

docs/sources/reference/components/mimir/mimir.rules.kubernetes.md
thampiotr: I'm using names based on how they're called in the code. I think since "value" can also be called "label value" then "label name" and "name" are fine. A "label" in this sense consists of both "name" and "value". So if you're okay with it I'd like to leave it.

internal/component/mimir/rules/kubernetes/events.go
thampiotr: It is being checked in the Validate step: https://github.com/grafana/alloy/pull/1773/files/453e253075f1a450a26240b8e6a94f4590e49c60#diff-1103eec8a21dbf8e36fdaf679f6214c70a74407b70d7786e3e826cdeab48f1e4R90 

This code is copied from upstream promtool (there's no way to import it as it's in main.go). So that check is redundant and comes from there. I'd prefer to leave this as-is.

thampiotr: Also, see this test case that uses Validate: https://github.com/grafana/alloy/pull/1773/files/453e253075f1a450a26240b8e6a94f4590e49c60#diff-5fe9918a1cab895fbb9ac0ec979e437c157885dce549f3920b0a6f4f40e08392R82

wildum: aa my bad I missed it, thanks

wildum (approved):

PR #1778, opened 2024-09-27
Wait for the old CRD Manager to stop before starting a new one

ptodev:
#### PR Description

This PR fixes an issue which was [reported](https://grafana.slack.com/archives/C01050C3D8F/p1726136997279809) on the community slack. A person ran into this error when doing a config reload:
```
failed to register service discovery metrics: failed to create service discovery refresh metrics
error running crd manager
```

I could not reproduce the error, but I think it happens due to the new CRD Manager starting up before the old one has had a chance to stop and unregister its metrics.

I tested my change locally with a config like this, just to make sure the waitgroup functions ok:

<details>
  <summary>Alloy config</summary>

```
  kubeconfig_file = "/Users/paulintodev/.kube/config"
  role = "node"
}

discovery.kubernetes "services" {
  kubeconfig_file = "/Users/paulintodev/.kube/config"
  role = "service"
}

// ServiceMonitor resources
prometheus.operator.servicemonitors "servicemonitors" {
  client {
    kubeconfig_file = "/Users/paulintodev/.kube/config"
  }
  clustering {
    enabled = false
  }
  forward_to = []
}

// PodMonitor resources
prometheus.operator.podmonitors "podmonitors" {
  client {
    kubeconfig_file = "/Users/paulintodev/.kube/config"
  }
  clustering {
    enabled = false
  }
  forward_to = []
}

// Probe resources
prometheus.operator.probes "probes" {
  client {
    kubeconfig_file = "/Users/paulintodev/.kube/config"
  }
  clustering {
    enabled = false
  }
  forward_to = []
}

logging {
  level = "info"
  format = "logfmt"
}
```

</details>

I changed the `clustering/enabled` value, then triggered a config reload via `curl localhost:12345/-/reload`. Then I shut down Alloy using Ctrl + C. The reload and shutdown went ok.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/operator/common/component.go
+++ b/internal/component/prometheus/operator/common/component.go
@@ -19,13 +19,15 @@ import (
 type Component struct {
 	mut     sync.RWMutex
 	config  *operator.Arguments
-	manager *crdManager
+	manager crdManagerInterface
 	ls      labelstore.LabelStore
 
 	onUpdate  chan struct{}
 	opts      component.Options
 	healthMut sync.RWMutex
 	health    component.Health
+
+	crdManagerFactory crdManagerFactory
 
 	kind    string
 	cluster cluster.Cluster
@@ -44,11 +46,12 @@ func New(o component.Options, args component.Arguments, kind string) (*Component
 	}
 	ls := service.(labelstore.LabelStore)
 	c := &Component{
-		opts:     o,
-		onUpdate: make(chan struct{}, 1),
-		kind:     kind,
-		cluster:  clusterData,
-		ls:       ls,
+		opts:              o,
+		onUpdate:          make(chan struct{}, 1),
+		kind:              kind,
+		cluster:           clusterData,
+		ls:                ls,
+		crdManagerFactory: realCrdManagerFactory{},
 	}
 	return c, c.Update(args)
 }
@@ -87,7 +90,7 @@ func (c *Component) Run(ctx context.Context) error {
 			c.reportHealth(err)
 		case <-c.onUpdate:
 			c.mut.Lock()
-			manager := newCrdManager(c.opts, c.cluster, c.opts.Logger, c.config, c.kind, c.ls)
+			manager := c.crdManagerFactory.New(c.opts, c.cluster, c.opts.Logger, c.config, c.kind, c.ls)
 			c.manager = manager
 
 			// Wait for the old manager to stop.
@@ -180,7 +183,7 @@ func (c *Component) Handler() http.Handler {
 		}
 		ns := parts[1]
 		name := parts[2]
-		scs := man.getScrapeConfig(ns, name)
+		scs := man.GetScrapeConfig(ns, name)
 		if len(scs) == 0 {
 			w.WriteHeader(404)
 			return
--- a/internal/component/prometheus/operator/common/component.go
+++ b/internal/component/prometheus/operator/common/component.go
@@ -74,6 +74,8 @@ func (c *Component) Run(ctx context.Context) error {
 
 	c.reportHealth(nil)
 	errChan := make(chan error, 1)
+	runWg := sync.WaitGroup{}
+	defer runWg.Wait()
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


internal/component/prometheus/operator/common/component.go
```diff
@@ -87,15 +89,23 @@ func (c *Component) Run(ctx context.Context) error {
...
+			// Wait for the old manager to stop.
+			// If we start the new manager before stopping the old one,
+			// the new manager might not be able to register its debug metrics due to a duplicate registration error.
 			if cancel != nil {
 				cancel()
 			}
+			runWg.Wait()
+
 			innerCtx, cancel = context.WithCancel(ctx)
+			runWg.Add(1)
```
mattdurham, line 102: Is it possible to get a test on this? Other than that looks good.

joke:
The same thing happens to me if I do configuration changes.

ptodev:
The test in this PR was flaky, but it is now fixed thanks to [help](https://github.com/grafana/alloy/pull/1903) from @mattdurham. He made a couple of suggestions:
1. To have the test wait until Run starts before proceeding. I added this code to the PR.
2. In the production code, to wrap the cancelations in a mutex like this:
   <details>
     <summary>Code</summary>

     ```go
     type cancelHandler struct {
     	mut            sync.RWMutex
     	cancel         context.CancelFunc
     	alreadyTrigger bool
     }
     
     func (ch *cancelHandler) Set(cncl context.CancelFunc) {
     	ch.mut.Lock()
     	defer ch.mut.Unlock()
     
     	ch.cancel = cncl
     }
     
     func (ch *cancelHandler) Cancel() {
     	ch.mut.Lock()
     	defer ch.mut.Unlock()
     
     	if ch.alreadyTrigger {
     		return
     	}
     	if ch.cancel == nil {
     		return
     	}
     	ch.cancel()
     	ch.alreadyTrigger = true
     }
     
     	// innerCtx gets passed to things we create, so we can restart everything anytime we get an update.
     	// Ideally, this component has very little dynamic config, and won't have frequent updates.
     	var innerCtx context.Context
     	// cancel is the func we use to trigger a stop to all downstream processors we create
     	var cancel cancelHandler
     
     	wg := sync.WaitGroup{}
     
     	c.reportHealth(nil)
     	errChan := make(chan error, 1)
     
     	defer wg.Wait()
     	for {
     		select {
     		case <-ctx.Done():
     			cancel.Cancel()
     			return nil
     		case err := <-errChan:
     			c.reportHealth(err)
     		case <-c.onUpdate:
     			c.manager = c.crdManagerFactory.New(c.opts, c.cluster, c.opts.Logger, c.config, c.kind, c.ls)
     
     			// Wait for the old manager to stop.
     			// If we start the new manager before stopping the old one,
     			// the new manager might not be able to register its debug metrics due to a duplicate registration error.
     			cancel.Cancel()
     			var cncl context.CancelFunc
     			innerCtx, cncl = context.WithCancel(ctx)
     			cancel.Set(cncl)
     			wg.Add(1)
     			go func() {
     				if err := c.manager.Run(innerCtx); err != nil {
     					level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
     					errChan <- err
     				}
     				wg.Done()
     			}()
     		}
     	}
     }
     ```
     </details>
    
    I'm not convinced that this is necessary, so I have not added this suggestion to this PR. We could revisit this decision in the future in case of issues. The code is already wrapped in a mutex, so to me it looks ok as it is. The test is also not flaky.

internal/component/prometheus/operator/common/component.go
thampiotr, line 1: Looks legit!

thampiotr (approved):
LGTM, no opinion about test, will let you discuss with @mattdurham

ptodev:
> internal/component/prometheus/operator/common/component.go

@thampiotr Thank you for the review! Unfortunately, I believe @mattdurham is out for the next two weeks. I'll just merge it :)

PR #1783, opened 2024-09-27
Update the release steps for "Cherry Pick Commits" and "Update Version in Code"

ptodev:
Fixing a few issues with the release docs:
* The example PRs used to point to the Agent.
* It wasn't very clear what the VERSION file should point to.
* There is a "Test Release" step which is a bit ambiguios. I think we should remove it and replace it with a CI check. 
* Combining the "Cherry Pick Commits" and "Update Version in Code" steps into one

I find it confusing to have "Cherry Pick Commits" and "Update Version in Code" as two separate steps. It's more efficient (and less confusing) to explain those things in the same step. It also makes the instructions less verbose.


docs/developer/release/3-update-version-in-code.md
```diff
@@ -1,60 +1,74 @@
...
+For example:
+* If you are going to release `v1.2.0-rc.0`, then the VERSION file should contain `v1.2.0`.
+* If you are going to release `v1.5.1`, then the VERSION file should contain `v1.5.1`.
 
-   After updating `VERSION`, run:
+## 2. Update the "main" branch
 
-   ```bash
-   make generate-versioned-files
-   ```
```
wildum, line 50: I think that we should keep this info

docs/developer/release/7-test-release.md
```diff
@@ -1,9 +0,0 @@
-# Test Release
-
-Validate the new version is working by running it incrementally in all dogfooding clusters.
-
-## Steps
-
-1. Validate performance metrics are consistent with the prior version.
-
-2. Validate components are healthy.
```
wildum, line 9: why delete this step? It's important not to forget to update our infra.

ptodev: It's very ambiguous how you'd do it. For example:
* Which components should this be done for?
* Which config should be used?
* Which platforms - Windows, Mac, Linux, BSD? 

Ultimately, is this not a step which can be automated by the CI? It sounds like something which should be done by the unit tests.

docs/developer/release/3-update-version-in-code.md
ptodev: True, I must have removed it by mistake. Thanks for spotting it! I added it back.

docs/developer/release/7-test-release.md
wildum: we can't test everything but it's an additional layer that has the benefit of testing real use cases at scale. 
For example Alloy 1.3 rc showed an increased rate of OOO errors and that's something we caught because we tested the RC in our infra

ptodev: Ok, this is the first time I realise that this is a reminder step to update Grafana's internal environments :) I thought this is about testing locally. I'll update the step instructions to mention it more clearly. Thank you for the feedback!

ptodev: Actually, it already says "dogfooding clusters" - idk how I missed it 😅  I guess I didn't read the description and just went straight to the steps.

wildum (approved):

clayton-cornell (approved):

Issue #979, opened 2024-06-04
Remove target block from prometheus.exporter.snmp


wildum:
An alternative way to send targets to the prometheus.exporter.snmp was introduced via this [PR](https://github.com/grafana/alloy/pull/969) to support dynamic targets.

The old way of passing the targets (via block) can be removed in the next major version. (breaking change)

PR #1790, opened 2024-09-30
SNMP exporter: Use a "name" argument for "target" and "walk_param" blocks

ptodev:
#### PR Description

Using an Alloy syntax identifiers restricts the sorts of labels that the SNMP exporter produces. For example, you cannot use `target "network-switch-1"` because identifiers can't contain a dash. 

In the SNMP exporter the name of a `target` block is used as for the value of a `job` Prometheus metric label, so it's not good to restrict the users in this way. They might want the label value to be a FQDM of a device that they are monitoring, and having to change the FQDN to something else is a usability issue.

The name of the `walk_param` block should probably ideally also not be restricted, although TBH I'm not sure if restricting it is a problem in practice.

Related to #289

#### Notes to the Reviewer

I documented the `name` argument as "required", even though it technically isn't required. But I think the docs look less confusing this way.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
@@ -21,7 +21,8 @@ The `prometheus.exporter.snmp` component embeds
 prometheus.exporter.snmp "LABEL" {
   config_file = SNMP_CONFIG_FILE_PATH
 
-  target "TARGET_NAME" {
+  target {
+    name    = TARGET_NAME
     address = TARGET_ADDRESS
   }
 }
@@ -48,7 +49,7 @@ Omitted fields take their default values.
 | `targets`     | `list(map(string))`  | SNMP targets.                                    |         | no       |
 
 The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
-Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter#generating-configuration) for details on how to generate a configuration file.
+Refer to [snmp_exporter][snmp_exporter-cfg] for details on how to generate a configuration file.
 
 The `config` argument must be a YAML document as string defining which SNMP modules and auths to use.
 `config` is typically loaded by using the exports of another component. For example,
@@ -64,6 +65,8 @@ The following labels can be set to a target:
 * `module`: The SNMP module to use for polling.
 * `auth`: The SNMP authentication profile to use.
 * `walk_params`: The config to use for this target.
+
+[snmp_exporter-cfg]: https://github.com/prometheus/snmp_exporter#generating-configuration
 
 ## Blocks
 
@@ -81,15 +84,25 @@ The following blocks are supported inside the definition of
 ### target block
 
 The `target` block defines an individual SNMP target.
-The `target` block may be specified multiple times to define multiple targets. The label of the block is required and will be used in the target's `job` label.
+The `target` block may be specified multiple times to define multiple targets.
 
 | Name           | Type     | Description                                                           | Default | Required |
 | -------------- | -------- | --------------------------------------------------------------------- | ------- | -------- |
+| `name`         | `string` | The name of the target.                                               |         | yes      |
 | `address`      | `string` | The address of SNMP device.                                           |         | yes      |
 | `module`       | `string` | SNMP module to use for polling.                                       | `""`    | no       |
 | `auth`         | `string` | SNMP authentication profile to use.                                   | `""`    | no       |
 | `walk_params`  | `string` | Config to use for this target.                                        | `""`    | no       |
 | `snmp_context` | `string` | Override the `context_name` parameter in the SNMP configuration file. | `""`    | no       |
+
+The value of the `name` argument will be set as the value of the target's `job` label.
+
+{{< admonition type="note" >}}
+Instead of a `name` argument, previous versions of Alloy used to use a block label.
+See the [Deprecated features](#deprecated-features) section for more information.
+{{< /admonition >}}
+
+If `name` is not set, the target's `job` label will be set to the label of the `target` block.
 
 ### walk_param block
 
@@ -98,10 +111,15 @@ The `walk_param` block may be specified multiple times to define multiple SNMP c
 
 | Name              | Type       | Description                                   | Default | Required |
 | ----------------- | ---------- | --------------------------------------------- | ------- | -------- |
-| `name`            | `string`   | Name of the module to override.               |         | no       |
+| `name`            | `string`   | Name of the module to override.               |         | yes      |
 | `max_repetitions` | `int`      | How many objects to request with GET/GETBULK. | `25`    | no       |
 | `retries`         | `int`      | How many times to retry a failed request.     | `3`     | no       |
 | `timeout`         | `duration` | Timeout for each individual SNMP request.     |         | no       |
+
+{{< admonition type="note" >}}
+Instead of a `name` argument, previous versions of Alloy used to use a block label.
+See the [Deprecated features](#deprecated-features) section for more information.
+{{< /admonition >}}
 
 ## Exported fields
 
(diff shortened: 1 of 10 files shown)


ptodev:
cc @bastischubert since he tends to use this component quite a lot :)

wildum (commented):
This way of passing targets is deprecated (see https://github.com/grafana/alloy/pull/1790/files#diff-858924fd308c6bdb411e0a2bbf65b65f040c2b3290b92af440c272558da206ffR165, https://github.com/grafana/alloy/issues/979, https://github.com/grafana/alloy/pull/969).
The problem is that passing targets as blocks forces the users to hardcode them in the Alloy config. With the new way of passing targets, you can get them from another component which means that you don't need to hardcode them.
I don't think that we should bring changes to deprecated solutions. We should rather push users towards the new solution

ptodev:
@wildum Thank you, I didn't realise the `targets` block is deprecated. I think we should make it more clear in the docs and also we need to update the examples to use the `targets` argument. At the moment the docs just state that "the `targets` argument is an alternative to the [target][] block.".

I'll ask the customer if they can use the `targets` argument instead.

wildum:
I agree that we should change the wording in the doc. We have three examples using the targets arg already but maybe we can change the order to put them first and clearly state that this is the preferred way. I was first reluctant to use the word deprecated in the doc because 2.0 is far away and I didn't want to stress the users but it seems that I brought more confusion than anything.

ptodev:
I'll raise a PR later today to reword the doc a bit :) It's a nice feature for it'd be nice for it to be advertised more prominently.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
 | Name           | Type     | Description                                                           | Default | Required |
 | -------------- | -------- | --------------------------------------------------------------------- | ------- | -------- |
+| `name`         | `string` | The name of the target.                                               |         | yes      |
 | `address`      | `string` | The address of SNMP device.                                           |         | yes      |
 | `module`       | `string` | SNMP module to use for polling.                                       | `""`    | no       |
 | `auth`         | `string` | SNMP authentication profile to use.                                   | `""`    | no       |
 | `walk_params`  | `string` | Config to use for this target.                                        | `""`    | no       |
 | `snmp_context` | `string` | Override the `context_name` parameter in the SNMP configuration file. | `""`    | no       |
 
+The value of the `name` argument will be set as the value of the target's `job` label.
```
clayton-cornell, line 98:
```suggestion
The value of the `name` argument is set to the value of the target's `job` label.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
+
+{{< admonition type="note" >}}
+Instead of a `name` argument, previous versions of Alloy used to use a block label.
```
clayton-cornell, line 101:
```suggestion
Instead of a `name` argument, previous versions of {{< param "PRODUCT_NAME" >}} used a block label.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
+See the [Deprecated features](#deprecated-features) section for more information.
```
clayton-cornell, line 102:
```suggestion
Refer to [Deprecated features](#deprecated-features) for more information.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
+{{< /admonition >}}
+
+If `name` is not set, the target's `job` label will be set to the label of the `target` block.
```
clayton-cornell, line 105:
```suggestion
If `name` is not set, the target's `job` label is set to the label of the `target` block.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
 | Name              | Type       | Description                                   | Default | Required |
 | ----------------- | ---------- | --------------------------------------------- | ------- | -------- |
-| `name`            | `string`   | Name of the module to override.               |         | no       |
+| `name`            | `string`   | Name of the module to override.               |         | yes      |
 | `max_repetitions` | `int`      | How many objects to request with GET/GETBULK. | `25`    | no       |
 | `retries`         | `int`      | How many times to retry a failed request.     | `3`     | no       |
 | `timeout`         | `duration` | Timeout for each individual SNMP request.     |         | no       |
 
+{{< admonition type="note" >}}
+Instead of a `name` argument, previous versions of Alloy used to use a block label.
```
clayton-cornell, line 120:
```suggestion
Instead of a `name` argument, previous versions of {{< param "PRODUCT_NAME" >}} used a block label.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -81,28 +84,43 @@ The following blocks are supported inside the definition of
...
+See the [Deprecated features](#deprecated-features) section for more information.
```
clayton-cornell, line 121:
```suggestion
Refer to [Deprecated features](#deprecated-features) for more information.
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -123,6 +141,46 @@ debug information.
 `prometheus.exporter.snmp` does not expose any component-specific
 debug metrics.
 
+## Deprecated features
+
+In previous versions of Alloy, the names `target` blocks and `walk_param` blocks were specified via a block label:
```
clayton-cornell, line 146:
```suggestion
In previous versions of {{< param "PRODUCT_NAME" >}}, the names `target` blocks and `walk_param` blocks were specified via a block label:
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -123,6 +141,46 @@ debug information.
...
+target "example_name_of_target" {
+    ...
+}
+
+walk_param "example_name_of_walk_param" {
+    ...
+}
+```
+
+Users are now advised to switch to using a `name` attribute instead:
```
clayton-cornell, line 158:
```suggestion
You can switch to using a `name` attribute instead of using a block label:
```

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -123,6 +141,46 @@ debug information.
...
+
+walk_param {
+    name = "example_name_of_walk_param"
+    ...
+}
+```
+
+Block labels for `target` and `walk_param` are deprecated and will be removed in version 2.0 of Alloy.
+In Alloy version 1, it will remain possible to use a block label instead of the `name` attribute.
+However, the `name` attribute is documented as "required" to encourage users to use it and to remove confusion.
```
clayton-cornell, line 174:
```suggestion
Block labels for `target` and `walk_param` are deprecated and will be removed in {{< param "PRODUCT_NAME" >}} 2.0.
In {{< param "PRODUCT_NAME" >}} 1.x, you can continue to use a block label.
However, the `name` attribute is documented as "required" to encourage users to use it and to remove confusion.
```

What happens if the user keeps the block labels AND adds a name attribute?

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -123,6 +141,46 @@ debug information.
...
+
+The benefits of using a `name` attribute include:
+* Not being subject to restrictions on [Alloy Syntax Identifiers][syntax-identifiers].
+* It is possible to get the value of a `name` argument from a component such as [local.file][] or [remote.http][].
```
clayton-cornell, line 178:
```suggestion
The benefits of using a `name` attribute include:
* You aren't subject to restrictions on [Alloy Syntax Identifiers][syntax-identifiers].
* You can get the value of a `name` argument from a component such as [local.file][] or [remote.http][].
```

ptodev:
Closing this in favour of #2274

PR #1806, opened 2024-10-02
Increase timeout for mimir.rules.kubernetes tests

ptodev:
Those tests are flaky. I hope increasing the timeout will help.


thampiotr (approved):

ptodev:
Closing this since #1890 should fix the real root cause of the problem.

PR #1809, opened 2024-10-02
Fix unnecessary `loki.process` config reloads

thampiotr:
#### PR Description

As discussed offline, this is a different implementation of https://github.com/grafana/alloy/pull/1426.

We're addressing only the mutation of configuration that is provided to `loki.process`, which is a local solution to address too frequent config reloads.

A more global (in the future) solution would ideally be implemented on the framework / runtime level that would prevent updating config when they didn't change. Or making configs truly immutable. 

NOTE 1: mutating arguments doesn't always lead to issues, as they are passed by value (struct) to the component and to each `newXXXStage` function. The problem starts when the arguments contain a slice, map or a pointer to another struct and it gets mutated.

NOTE 2: I also caught a couple of stages not correctly applying their defaults because they were mutating a function-local copy of the configuration only. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/labels_test.go
+++ b/internal/component/loki/process/stages/labels_test.go
@@ -72,10 +72,8 @@ func TestLabelsPipelineWithMissingKey_Labels(t *testing.T) {
 }
 
 var (
-	lv1  = "lv1"
-	lv2c = "l2"
-	lv3  = ""
-	lv3c = "l3"
+	lv1 = "lv1"
+	lv3 = ""
 )
 
 var emptyLabelsConfig = LabelsConfig{nil}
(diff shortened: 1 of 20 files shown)


internal/component/loki/process/stages/labels_test.go
```diff
@@ -105,18 +103,18 @@ func TestLabels(t *testing.T) {
...
 			err: nil,
-			expectedCfgs: LabelsConfig{Values: map[string]*string{
-				"l1": &lv1,
-				"l2": &lv2c,
-				"l3": &lv3c,
-			}},
+			expectedCfgs: map[string]string{
+				"l1": "lv1",
+				"l2": "l2",
+				"l3": "l3",
```
ptodev, line 109: Since the map's value is no longer a pointer to string, why not just use `"l1": lv1`?

ptodev (approved):

ptodev:
> NOTE 2: I also caught a couple of stages not correctly applying their defaults because they were mutating a function-local copy of the configuration only.

This is partially why I went with a different approach in #1426. I'd also noticed a few cases of mutations and I decided could not guarantee that I can catch every single instance where this happens. I still think it's easier to copy than to do manually inspect the codebase for mutations, but if this will get the bug resolved for now, I'm ok with merging this.

internal/component/loki/process/stages/labels_test.go
thampiotr: ok

thampiotr:
> I could not guarantee that I can catch every single instance where this happens.

I agree that we should not rely on human reviews to catch this. And also many other possible errors. That's why I believe the right answer here is to have tests. If there is a validate / default values function, it should be tested. And same goes for most of other code that has any non-trivial logic.

PR #1816, opened 2024-10-03
Remove the old peer discovery mechanism since v2 is default

### LINKED ISSUE #1274: Graduate the improvements to clustering to GA once proven
Some improvements to clustering are done as part of https://github.com/grafana/alloy/issues/784.

Some of them are in public preview maturity. This issue is to graduate them to GA once they've been released and confirmed not to cause issues. 

See comments in code and the following task list

```[tasklist]
### Tasks
- [x] Make this PR GA: https://github.com/grafana/alloy/pull/1261
- [x] Make the `--cluster.use-discovery-v1` flag a no-op and schedule removal in next major release: https://github.com/grafana/alloy/pull/1311
```

thampiotr:
#### PR Description

Now that we had run this code for a while, we're happy to remove the old v1 cluster peer discovery mechanism since the v2 works without issues as a drop-in replacement.

Fixes #1274

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Main (unreleased)
 - The `mimir.rules.kubernetes` component now supports adding extra label matchers
   to all queries discovered via `PrometheusRule` CRDs. (@thampiotr)
 
+- The `cluster.use-discovery-v1` flag is now deprecated since there were no issues found with the v2 cluster discovery mechanism. (@thampiotr)
+
 ### Bugfixes
 
 - Update windows_exporter from v0.27.2 vo v0.27.3: (@jkroepke)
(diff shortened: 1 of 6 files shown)


mattdurham:
Does this change the behavior any of join peers or the discovery string? Taking a quick look it doesnt appear to but want to double check.

thampiotr:
> Does this change the behavior any of join peers or the discovery string? Taking a quick look it doesnt appear to but want to double check.

No, because it was default from the time we merged it. We had a way to switch back to the old discovery, and this is what we're removing here.

There is also one feature that was enabled only for experimental users, but we tested it sufficiently too, so it's becoming the default.

mattdurham:
Did how we use to handle `buildJoinAddresses` change? Looks like the v1 handled it differently on looking up the SRV address. IE if I was using v1 discovery and using it as a DNS lookup does that still work?

mattdurham:
It looks like it should work through the SRVResolver but its enough different I am unsure.

mattdurham (approved):
LGTM

thampiotr:
> It looks like it should work through the SRVResolver but its enough different I am unsure.

Sorry for the confusion. As discussed offline - we currently use v2 by default and we had a flag that allowed to switch back to v1 in case of issues. Since v2 works fine, I'm removing the flag and deleting the v1. There are no changes required by users, this is fully compatible with old configs.

PR #1721, opened 2024-09-20
tweak OTEL alerts to use SR

thampiotr:
#### PR Description

Changes OTEL Collector alerts to use success rate of 95% over 10min period to alert. Otherwise, we may have 99.9999% success rate and still alert. Some level of failures is acceptable for most users, especially for tracing which is frequently sampled.

#### Notes to the Reviewer

Tested in https://github.com/grafana/deployment_tools/pull/174974 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #1818, opened 2024-10-03
Fix OTEL alerts query

thampiotr:
#### PR Description

This is a fix to yet unreleased change to OTEL alerts queries that was added in this PR: https://github.com/grafana/alloy/pull/1721

There was a mistake in original PR: we were calculating the failure rate of each time series and then doing `sum by (cluster, namespace, job)` on these, so if we had 200 instances with 1% failure rate, we would end up with 200% failure rate.

Instead we should `sum by (cluster, namespace, job)` for the nominator (number of failures) and the denominator (total number of events) to get a correct SR for the entire cluster/namespace/job. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


mattdurham (approved):
lgtm

PR #1821, opened 2024-10-03
Configure SELinux label on CloudWatch integration mounts

jdbaldry:
Also make it easy to turn off checking the CloudWatch integration.

Without the SELinux label, users running SELinux see permissions issues running this target.


mattdurham (approved):
LGTM

PR #1823, opened 2024-10-03
Component

mattdurham:
This is the last PR, aside from the merge to main. This ties everything together and exposes the component. Most of this are very expansive tests. I am likely to change the metrics in subsequent PRs so dont get to in detail about them.

Things to do that I will do after this large merge and create tickets for each.
* Add support for TLS
* Rethink the metrics and the approach.
* Add metadata support, this requires exposing additional properties on the scrape itself.
* Add hook to not use labelstore if not running any remote_write components.


docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+
+# prometheus.remote.queue
+
+`prometheus.remote.queue` collects metrics sent from other components into a
+Write-Ahead Log (WAL) and forwards them over the network to a series of
+user-supplied endpoints. Metrics are sent over the network using the
+[Prometheus Remote Write protocol][remote_write-spec].
+
+Multiple `prometheus.remote.queue` components can be specified by giving them
+different labels.
```
clayton-cornell, line 17:
```suggestion
You can specify multiple `prometheus.remote.queue` components by giving them different labels.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+{{< docs/shared lookup="reference/components/basic-auth-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`receiver` | `MetricsReceiver` | A value which other components can use to send metrics to.
```
clayton-cornell, line 93:
```suggestion
`receiver` | `MetricsReceiver` | A value that other components can use to send metrics to.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Debug information
+
+`prometheus.remote_write` does not expose any component-specific debug
+information.
+
+## Debug metrics
+
+The below metrics are provided for backwards compatibility, they behave generally the same but there are likely
+edge cases where they differ.
```
clayton-cornell, line 109:
```suggestion
The following metrics are provided for backward compatibility.
They generally behave the same, but there are likely edge cases where they differ.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+  exemplars that failed to send to remote storage but were retried due to
+  recoverable errors.
+* `prometheus_remote_storage_metadata_retried_total` (counter): Total number of
+  metadata entries that failed to send to remote storage but were retried due
+  to recoverable errors.
+* `prometheus_remote_storage_samples_dropped_total` (counter): Total number of
+  samples which were dropped after being read from the WAL before being sent to
+  remote_write because of an unknown reference ID.
+* `prometheus_remote_storage_exemplars_dropped_total` (counter): Total number
+  of exemplars which were dropped after being read from the WAL before being
```
clayton-cornell, line 145:
```suggestion
  of exemplars that were dropped after being read from the WAL before being
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+## Technical details
+
+`prometheus.remote.queue` uses [snappy][] for compression.
+`prometheus.remote.queue` sends native histograms by default.
+Any labels that start with `__` will be removed before sending to the endpoint.
+
+### Data retention
+
+Data is written to disk in blocks utilizing [snappy][] compression. These blocks are read on startup and resent if they are still within the TTL. 
+Any data that has not been written to disk, or that is in the network queues is lost if Alloy is restarted.
```
clayton-cornell, line 208:
```suggestion
Any data that has not been written to disk, or that is in the network queues is lost if {{< param "PRODUCT_NAME" >}} is restarted.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Retries
+
+Network errors will be retried. 429 errors will be retried. 5XX errors will retry. Any other non-2XX return codes will not be tried. 
+
+### Memory
+
+`prometheus.remote.queue` is meant to be memory efficient. By adjusting the `max_signals_to_batch`, `queue_count`, and `batch_size` the amount of memory
+can be controlled. A higher `max_signals_to_batch` allows for more efficient disk compression. A higher `queue_count` allows more concurrent writes and `batch_size`
+allows more data sent at one time. This can allow greater throughput, at the cost of more memory on both Alloy and the endpoint. The defaults are good for most
```
clayton-cornell, line 218:
```suggestion
allows more data sent at one time. This can allow greater throughput, at the cost of more memory on both {{< param "PRODUCT_NAME" >}} and the endpoint. The defaults are good for most 
```

clayton-cornell (commented):
First high level pass of the docs. I'll go through again once the ToDos are added

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,288 @@
...
+Metrics that are new to `prometheus.remote.write`. These are highly subject to change.
+
+* `alloy_queue_series_serializer_incoming_signals` (counter): Total number of series written to serialization.
+* `alloy_queue_metadata_serializer_incoming_signals` (counter): Total number of metadata written to serialization.
+* `alloy_queue_series_serializer_incoming_timestamp_seconds` (gauge): Highest timestamp of incoming series.
+* `alloy_queue_series_serializer_errors` (gauge): Number of errors for series written to serializer.
+* `alloy_queue_metadata_serializer_errors` (gauge): Number of errors for metadata written to serializer.
+* `alloy_queue_series_network_timestamp_seconds` (gauge): Highest timestamp written to an endpoint.
+* `alloy_queue_series_network_sent` (counter): Number of series sent successfully.
+* `alloy_queue_metadata_network_sent` (counter): Number of metadata sent successful.
```
clayton-cornell, line 187:
```suggestion
* `alloy_queue_metadata_network_sent` (counter): Number of metadata sent successfully.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,288 @@
...
+Any labels that start with `__` will be removed before sending to the endpoint.
+
+### Data retention
+
+Data is written to disk in blocks utilizing [snappy][] compression. These blocks are read on startup and resent if they are still within the TTL. 
+Any data that has not been written to disk, or that is in the network queues is lost if {{< param "PRODUCT_NAME" >}} is restarted.
+
+### Retries
+
+Network errors will be retried. 429 errors will be retried. 5XX errors will retry. Any other non-2XX return codes will not be tried.
```
clayton-cornell, line 249:
```suggestion
`prometheus.remote.queue`  will retry sending data if the following errors or HTTP status codes are returned:

 * Network errors. 
 * HTTP 429 error.s 
 * HTTP 5XX errors.
 
`prometheus.remote.queue`  will  not retry sending data if any other unsuccessful status codes are returned. 
```

I tried to be more explicit what was happening. The original implied that the errors are retried when what was really meant was a retry happens if the 429 and 5xx status codes are returned.

For the network errors... does this mean network errors in general?

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,288 @@
...
+### Retries
+
+Network errors will be retried. 429 errors will be retried. 5XX errors will retry. Any other non-2XX return codes will not be tried. 
+
+### Memory
+
+`prometheus.remote.queue` is meant to be memory efficient. By adjusting the `max_signals_to_batch`, `queue_count`, and `batch_size` the amount of memory
+can be controlled. A higher `max_signals_to_batch` allows for more efficient disk compression. A higher `queue_count` allows more concurrent writes and `batch_size`
+allows more data sent at one time. This can allow greater throughput, at the cost of more memory on both {{< param "PRODUCT_NAME" >}} and the endpoint. The defaults are good for most 
+common usages.
```
clayton-cornell, line 256:
```suggestion
`prometheus.remote.queue` is meant to be memory efficient.
You can adjust the `max_signals_to_batch`, `queue_count`, and `batch_size` to control how much memory is used.
A higher `max_signals_to_batch` allows for more efficient disk compression.
A higher `queue_count` allows more concurrent writes, and `batch_size` allows more data sent at one time.
This can allow greater throughput at the cost of more memory on both {{< param "PRODUCT_NAME" >}} and the endpoint.
The defaults are suitable for most common usages. 
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: Network errors are specific to transport errors, ie where you dont get a response.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+Data is written to disk in blocks utilizing [snappy][] compression. These blocks are read on startup and resent if they are still within the TTL. 
+Any data that has not been written to disk, or that is in the network queues is lost if {{< param "PRODUCT_NAME" >}} is restarted.
+
+### Retries
+
+`prometheus.remote.queue`  will retry sending data if the following errors or HTTP status codes are returned:
+
+ * Network errors. 
+ * HTTP 429 error.s
```
clayton-cornell, line 252:
```suggestion
 * HTTP 429 errors. 
```

Ooops.. transposed the full stop and the last letter in the suggestion, this fixes it.

clayton-cornell (approved):
Approving for docs.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Component health
+
+`prometheus.remote.queue` is only reported as unhealthy if given an invalid
+configuration. In those cases, exported fields are kept at their last healthy
+values.
+
+## Debug information
+
+`prometheus.remote_write` does not expose any component-specific debug
```
ptodev, line 119:
```suggestion
`prometheus.remote.queue` does not expose any component-specific debug
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+* `alloy_queue_network_series_retried_5xx` (counter): Number of series retried due to status code 5xx.
+* `alloy_queue_network_metadata_retried_5xx` (counter): Number of metadata retried due to status code 5xx.
+* `alloy_queue_network_series_network_duration_seconds` (histogram): Duration writing series to endpoint.
+* `alloy_queue_network_metadata_network_duration_seconds` (histogram): Duration writing metadata to endpoint.
+* `alloy_queue_network_series_network_errors` (counter): Number of errors writing series to network.
+* `alloy_queue_network_metadata_network_errors` (counter): Number of errors writing metadata to network.
+
+## Examples
+
+The following examples show you how to create `prometheus.remote_write` components that send metrics to different destinations.
```
ptodev, line 203:
```suggestion
The following examples show you how to create `prometheus.remote.queue` components that send metrics to different destinations.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+    basic_auth {
+      username = "example-user"
+      password = "example-password"
+    }
+  }
+}
+
+// Configure a prometheus.scrape component to send metrics to
+// prometheus.remote_write component.
```
ptodev, line 223:
```suggestion
// prometheus.remote.queue component.
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
ptodev, line 1: It'd be nice to add a bit more info on how `prometheus.remote.queue` is different from `prometheus.remote_write`, and when to use what component.

internal/component/prometheus/remote/queue/component.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/serialization"
+	"github.com/grafana/alloy/internal/component/prometheus/remote/queue/types"
+	"github.com/grafana/alloy/internal/featuregate"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/storage"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:      "prometheus.remote.queue",
```
ptodev, line 22: Do you have plans to put other components in the `prometheus.remote` namespace?

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote.queue/
+description: Learn about prometheus.remote.queue
+title: prometheus.remote.queue
+---
+
+
+<span class="badge docs-labels__stage docs-labels__item">Experimental</span>
+
+# prometheus.remote.queue
```
thampiotr, line 10: I know it may be late to the partey, but what do you think of `prometheus.write.queue`?
We have `loki.write`, `pyroscope.write`... to me it seems a good idea to use `prometheus.write.*` prefix and then we can have a `prometheus.write.sync` for maybe an implementation that writes synchronously without a WAL (can be useful for gateway scenario) and the current `prometheus.write.queue` that uses a queue....

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+`prometheus.remote.queue` collects metrics sent from other components into a
+Write-Ahead Log (WAL) and forwards them over the network to a series of
+user-supplied endpoints. Metrics are sent over the network using the
+[Prometheus Remote Write protocol][remote_write-spec].
+
+You can specify multiple `prometheus.remote.queue` components by giving them different labels.
+
+You should consider everything here extremely experimental and highly subject to change.
+[emote_write-spec]: https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit
```
thampiotr, line 20:
```suggestion
[remote_write-spec]: https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`ttl` | `time` | `duration` | How long the timestamp of a signal is valid before the signal is discarded. | `2h` | no
```
thampiotr, line 44:
```suggestion
`ttl` | `time` | `duration` | How long the samples can be queued for before they are discarded. | `2h` | no
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+The `>` symbol indicates deeper levels of nesting. For example, `endpoint >
+basic_auth` refers to a `basic_auth` block defined inside an
+`endpoint` block.
+
+[endpoint]: #endpoint-block
+[basic_auth]: #basic_auth-block
+[serialization]: #serialization-block
+
+### serialization block
```
thampiotr, line 65: How about calling this `persistence` instead? Serialisation is just the process of encoding data in some format. But here we are configuring more than that - it's disk persistence / durability / storage or something like that.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+The `serialization` block describes how often and at what limits to write to disk. Serialization settings
+are shared for each `endpoint.`
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`max_signals_to_batch` | `uint` | The maximum number of signals before they are batched to disk. | `10,000` | no
+`batch_frequency` | `duration` | How often to batch signals to disk if `max_signals_to_batch` is not reached. | no
```
thampiotr, line 75: Why do we use `signals` as a name? maybe it should be samples? or is this to encompass histograms samples too? I'm not sure if that's a right name. In current remote_write we call this 'metrics' which makes sense although it's not very precise: https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.remote_write/#wal-block

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`url` | `string` | Full URL to send metrics to. | | yes
+`name` | `string` | Optional name to identify the endpoint in metrics. | | no
+`write_timeout` | `duration` | Timeout for requests made to the URL. | `"30s"` | no
+`retry_backoff` | `duration` | How often to wait between retries. | `1s` | no
+`max_retry_backoff_attempts` | Maximum number of retries before dropping the batch. | `1s` | no
```
thampiotr, line 92:
```suggestion
`max_retry_backoff_attempts` | `uint`  |Maximum number of retries before dropping the batch. | `1` | no
```
Something is not right here...

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+`batch_count` | `uint` | How many series to queue in each queue. | `1,000` | no
```
thampiotr, line 93:
```suggestion
`batch_count` | `uint` | How many series to queue in each queue. | `1000` | no
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+`flush_frequency` | `duration` | How often to wait until sending if `batch_count` is not trigger. | `1s` | no
```
thampiotr, line 94:
```suggestion
`flush_interval` | `duration` | How often to wait until sending if `batch_count` is not trigger. | `1s` | no
```

frequency is `1/interval` and would be in Hz

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
thampiotr, line 92:
```suggestion
`max_retry_attempts` | Maximum number of retries before dropping the batch. | `1s` | no
```
not sure why add backoff here?

thampiotr (commented):
I focused on the user-facing API as this will be harder to change without breaking things.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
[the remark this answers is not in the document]
wildum: it's a duration, the default is 5s

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+The `endpoint` block describes a single location to send metrics to. Multiple
+`endpoint` blocks can be provided to send metrics to multiple locations. Each
+`endpoint` will have its own WAL folder
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`url` | `string` | Full URL to send metrics to. | | yes
+`name` | `string` | Optional name to identify the endpoint in metrics. | | no
```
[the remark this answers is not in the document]
mattdurham: Actually I can remove this since its the label name.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+  targets = [
+    // Collect metrics from the default HTTP listen address.
+    {"__address__" = "127.0.0.1:12345"},
+  ]
+  forward_to = [prometheus.remote.queue.staging.receiver]
+}
+
+```
+
+## TODO Metadata settings
```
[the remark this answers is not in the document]
mattdurham: It requires scrape changes, I will remove the todo, metadata support is down the line.

internal/component/prometheus/remote/queue/endpoint.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (ep *endpoint) Start() {
+	ep.self = actor.Combine(actor.New(ep), ep.incoming).Build()
+	ep.self.Start()
+	ep.serializer.Start()
+	ep.network.Start()
+}
+
+func (ep *endpoint) Stop() {
```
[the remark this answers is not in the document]
mattdurham: Its cleaner this way, you get less ignorable errors. Will add a comment.

internal/component/prometheus/remote/queue/component.go
mattdurham: None, but I couldnt think of a better place to put it!

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: I dont have problem with that @ptodev @wildum thoughts? Seems to match better with our current standard so sounds legit.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: Seems reasonable

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: Subsequent PR will enable metadata storage, so I wanted it to not be just metrics. Sample is to narrow, open to better names here.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: We use both interval and frequency all over the configuration. The original remote_write uses both, but both use the verbiage frequency/frequent to describe the value so it felt better. 

remote write below
`truncate_frequency` | `duration` | How frequently to clean up the WAL. | `"2h"` | no
`send_interval` | `duration` | How frequently metric metadata is sent to the endpoint. | `"1m"` | no

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`url` | `string` | Full URL to send metrics to. | | yes
+`name` | `string` | Optional name to identify the endpoint in metrics. | | no
+`write_timeout` | `duration` | Timeout for requests made to the URL. | `"30s"` | no
+`retry_backoff` | `duration` | How often to wait between retries. | `1s` | no
+`max_retry_backoff_attempts` | Maximum number of retries before dropping the batch. | `1s` | no
+`batch_count` | `uint` | How many series to queue in each queue. | `1,000` | no
+`flush_frequency` | `duration` | How often to wait until sending if `batch_count` is not trigger. | `1s` | no
+`queue_count` | `uint` | How many concurrent batches to write. | 10 | no
```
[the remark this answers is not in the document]
mattdurham: Parallelism is likely the more accurate word in both the name and verbiage. How do you feel about that?

`parallelism` | `uint` | How many parrallel batches to write. | 10 | no

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
```diff
@@ -0,0 +1,296 @@
...
+`external_labels` | `map(string)` | Labels to add to metrics sent over the network. | | no
```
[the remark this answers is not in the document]
mattdurham: Its how remote write defines it, so its easier to grok. I dont have a strong preference though

internal/component/prometheus/remote/queue/component.go
```diff
@@ -0,0 +1,160 @@
...
+	// If they are the same do nothing.
+	if reflect.DeepEqual(newArgs, s.args) {
+		return nil
+	}
+	s.args = newArgs
+	// TODO @mattdurham need to cycle through the endpoints figuring out what changed instead of this global stop and start.
+	// TODO @mattdurham is there an issue/race condition with stopping these while the appender is still going on.
+	if len(s.endpoints) > 0 {
+		for _, ep := range s.endpoints {
+			ep.Stop()
```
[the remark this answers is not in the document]
mattdurham: No, it will in general call the cancel on the context, which will cascade down. You will in general lose any data in the queues. Will add documentation.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
thampiotr: haha, I see... well, we can improve on that, but I guess it's a nit

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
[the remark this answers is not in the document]
thampiotr: yep! that sounds good to me

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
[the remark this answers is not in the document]
thampiotr: ok, I'll leave it up to you, hard to tell whether it's worth using better phrasing with a cost of potentially tripping some users who got used to old names

internal/component/prometheus/remote/queue/component.go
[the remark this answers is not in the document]
thampiotr: Any plans on adding a drain behaviour? Could perhaps include that in the comment.  I anticipate someone opening an issue about metric gaps one day.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
mattdurham: I like this better, and going to do it unless we get strong objection :)

internal/component/prometheus/remote/queue/types.go
```diff
@@ -0,0 +1,119 @@
...
+	TTL           time.Duration    `alloy:"ttl,attr,optional"`
+	Serialization Serialization    `alloy:"serialization,block,optional"`
+	Endpoints     []EndpointConfig `alloy:"endpoint,block"`
+}
+
+type Serialization struct {
+	// The batch size to persist to the file queue.
+	MaxSignalsToBatch int `alloy:"max_signals_to_batch,attr,optional"`
+	// How often to flush to the file queue if BatchSize isn't met.
+	BatchFrequency time.Duration `alloy:"batch_frequency,attr,optional"`
```
wildum, line 34: Should this be FlushFrequency to match with the rest?

internal/component/prometheus/remote/queue/component.go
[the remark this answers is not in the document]
mattdurham: One of the original versions I did had it but it was a bit fiddly. Its something I def want to do but removed it from scope for the first iteration; like specific TLS settings, auto scaling the number of parallel queues  and a few other things.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
clayton-cornell: Either way works if it's consistent with upstream and within Alloy.  But... we can look at the top level definitions...

* frequency... is the rate something happens or the number of times something happens within a given time period
* interval... is a measurable time between things or events

So what are we providing? A fixed time.. for example, 30 seconds? Then it's an interval. if it's something like like 150/hour then it's a frequency.  

In this case... I'd probably opt to go with 

```suggestion
`flush_interval` | `duration` | How long to wait until sending if `batch_count` is not triggered. | `1s` | no
```

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
thampiotr: maybe 'entries'? that would be a generic name for things inside a queue.

docs/sources/reference/components/prometheus/prometheus.remote.queue.md
ptodev: `prometheus.write.queue` sounds better to me too.

mattdurham: Doing a rename here and will do the import changes in a separate PR I will automerge since it generates a lot of noise.

mattdurham:
Most of the feedback has been responded so merging.

PR #1826, opened 2024-10-04
Prototype targets.merge function (array.combine_maps)

### LINKED ISSUE #1443: Integrate `prometheus.exporter` and `discovery` components

ptodev:
I hope this can enable users to decorate prometheus.exporter metrics with labels from discovery components. 

The config could look like this:

```
discovery.kubernetes "default" {
  role = "service"
}

prometheus.exporter.unix "default" {}

// Probably we'd need some discovery.relabel to make sure both targets have a common key/val to serve as condition of joining

prometheus.scrape "demo" {
  targets    = targets.merge(prometheus.exporter.unix.default.targets, discovery.kubernetes.default.targets, ["__meta_kubernetes_service_cluster_ip"])
  forward_to = [prometheus.remote_write.mimir.receiver]
}

  endpoint {
    url = "https://prometheus-prod-05-gb-south-0.grafana.net/api/prom/push"

    basic_auth {
      username = ""
      password = ""
    }
  }
}

```

Fixes #1443


docs/sources/reference/stdlib/map.md
```diff
@@ -0,0 +1,35 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/stdlib/map/
+description: Learn about map functions
+aliases:
+  - ./concat/ # /docs/alloy/latest/reference/stdlib/concat/
+menuTitle: map
+title: map
+---
+
+# map
```
ptodev, line 10: Technically, this page should be called "object" and the function should be `object.inner_join`. That's because the [Alloy syntax type](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/#objects) is called "object". I wonder if we should rename the Alloy syntax type?
* "Map" would ring more bells in people's minds than "object"
* We already document this as "map" in component reference docs
* "Object" makes me think of the sort of thing a "capsule" does

docs/sources/reference/stdlib/map.md
```diff
@@ -0,0 +1,35 @@
...
+   Its value could have any Alloy type such as a string, integer, map, or a capsule.
+2. An array of maps (objects) which will be joined. The keys of the map are strings. 
+   Its value could have any Alloy type such as a string, integer, map, or a capsule.
+3. An array containing strings. The strings are the keys whose value has to match for maps to be joined.
+   If the set of keys don't identify a map uniquely, the resulting output may contain more maps than the total sum of maps from both input arrays.
+4. (optional; default: `"update"`) A merge strategy. It describes which value will be used if there is Can be set to either of:
+   * `update`: If there is already a key with the same name in the first map, it will be updated with the value from the second map.
+   * `none`: If both maps have different values for the same key, no such key will be present in the output map.
+5. (optional; default: `[]`) An array containing strings. Only keys listed in the array will be present in the output map.
+6. (optional; default: `[]`) An array containing strings. Keyls listed in the array will not be present in the output map.
```
ptodev, line 29: I might remove those two parameters. I added them for two reasons:
* I thought it might make the function more ergonomic so that users don't have to use discovery.relabel every time, even if they want to do something basic.
* I thought it could be an optimisation to handle duplicate maps in a smart way. E.g. if the user only cares about adding 1 label, but their kubernetes SD gives them 20 labels, `map.inner_join` could remove the labels from the incoming map and then deduplicate the maps from the SD. 

But the more I think of this, the more it sounds like an overkill. Even if we do all this deduplication, there is still a chance that the keys used as condition of join are not unique, which could lead to too many output maps in the outgoing array. So it's probably best to let the users massage the data before the inner_join to make sure it's good.

docs/sources/reference/stdlib/map.md
```diff
@@ -0,0 +1,35 @@
...
+
+## map.inner_join
+
+The `map.inner_join` function allows you to join two arrays containing maps if certain keys have matching values in both maps.
+It takes several inputs, in this order:
+
+1. An array of maps (objects) which will be joined. The keys of the map are strings. 
+   Its value could have any Alloy type such as a string, integer, map, or a capsule.
+2. An array of maps (objects) which will be joined. The keys of the map are strings. 
+   Its value could have any Alloy type such as a string, integer, map, or a capsule.
```
wildum, line 22: steps 1 and 2 are identical

docs/sources/reference/stdlib/map.md
```diff
@@ -0,0 +1,35 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/stdlib/map/
+description: Learn about map functions
+aliases:
+  - ./concat/ # /docs/alloy/latest/reference/stdlib/concat/
```
wildum, line 5: this page should not have aliases

mattdurham:
I wonder on the naming, the name makes me feel like its very SQL like and should behave like a SQL inner join.

docs/sources/reference/stdlib/map.md
ptodev: I removed those params

docs/sources/reference/stdlib/map.md
ptodev: I realised that even "map" is not an appropriate namespace for this function, since it operates on arrays of maps. I changed the function name to `targets.merge`.

ptodev:
> I wonder on the naming, the name makes me feel like its very SQL like and should behave like a SQL inner join.

@mattdurham Good point - I renamed it from "inner_join" to "merge". I also changed the namespace from "map" to "targets", because I realised that even "map" is not an appropriate namespace for this function as it operates on arrays of maps.

docs/sources/reference/stdlib/targets.md
```diff
@@ -0,0 +1,38 @@
...
+[Compatible components]: ../compatibility/#targets
+
+## targets.merge
+
+The `targets.inner_join` function allows you to join two arrays containing maps if certain keys have matching values in both maps.
+It takes three inputs:
+
+* The first two inputs are a of type `list(map(string))`. The keys of the map are strings. 
+  The value for each key could have any Alloy type such as a string, integer, map, or a capsule.
+* The third input is an array containing strings. The strings are the keys whose value has to match for maps to be joined.
```
ptodev, line 25: Currently, the "third input" assumes both targets have a label with the same value. I wonder if it's worth changing the third input to be an array of arrays, where we can set different label names in the LHS and RHS targets. E.g.:

```
[["lhs_lbl_name"], ["rhs_lbl_name"], ["lhs_lbl_name_2"], ["rhs_lbl_name_2"]]
```

This could reduce the amount of relabels a user would have to do. But I don't know if it's really worth the extra complexity. The users probably have to relabel anyway.

syntax/vm/vm_stdlib_test.go
```diff
@@ -39,6 +39,51 @@ func TestVM_Stdlib(t *testing.T) {
 		{"encoding.from_yaml nil field", "encoding.from_yaml(`foo: null`)", map[string]interface{}{"foo": nil}},
 		{"encoding.from_yaml nil array element", `encoding.from_yaml("[0, null]")`, []interface{}{0, nil}},
 		{"encoding.from_base64", `encoding.from_base64("Zm9vYmFyMTIzIT8kKiYoKSctPUB+")`, string(`foobar123!?$*&()'-=@~`)},
+
+		// Map tests
+		{
+			// Basic case. No conflicting key/val pairs.
+			"targets.merge",
+			`targets.merge([{"a" = "a1", "b" = "b1"}], [{"a" = "a1", "c" = "c1"}], ["a"])`,
```
mattdurham, line 47: What is the result for invalid input? Imagine the first is a valid map list but the second is nil/invalid type/etc? I imagine it would return the map list, whereas if it is the reverse it would return a blank list?

docs/sources/reference/stdlib/targets.md
```diff
@@ -0,0 +1,38 @@
...
+  
+  
+If the set of keys don't identify a map uniquely, the resulting output may contain more maps than the total sum of maps from both input arrays.
```
mattdurham, line 28: This feels off, whats the use case here?

mattdurham:
Overall looks good! Though the complexity of this feels like it borders on a component, which means we could use the live debugging and mark as experimental. Since this has become specific to targets thoughts to add it to relabel.discovery?

docs/sources/reference/stdlib/targets.md
```diff
@@ -0,0 +1,38 @@
...
+They are often used by [prometheus][prom-comp] and [discovery][disc-comp] components.
+Refer to [Compatible components][] for a full list of components which export and consume targets.
+
+[prom-comp]: ../components/prometheus/
+[disc-comp]: ../components/discovery/
+[Compatible components]: ../compatibility/#targets
+
+## targets.merge
+
+The `targets.inner_join` function allows you to join two arrays containing maps if certain keys have matching values in both maps.
```
ptodev, line 20:
```suggestion
The `targets.merge` function allows you to join two arrays containing maps if certain keys have matching values in both maps.
```
I need to rename to `targets.merge` here and in all other places in the doc.

ptodev:
> Overall looks good! Though the complexity of this feels like it borders on a component, which means we could use the live debugging and mark as experimental. Since this has become specific to targets thoughts to add it to relabel.discovery?

@mattdurham The main reasons for having it as an std lib function are:
* It is technically not specific to targets. You could have any lists of maps and the function could still work. I namespaced it as "targets", just because I couldn't think of a better name.
* It's probably more performant as stdlib than as a component. This is just a guess and I'm not sure what the benchmarks would be in reality.

I do agree that there would be benefits to the component approach:
* Works with the UI - not just live debugging, but the actual UI that shows inputs.
* We can gate it behind the "experimental" cmd flag.
* It's easier for people who use `discovery` components to find it.

We could do a few things to make the std function a more compelling option:
* Rename this function to something that doesn't include "targets" in the name. For example, `map.merge()`? Technically, the inputs are not maps because we're merging arrays of maps... but that's probably fine.
* We'd have to rename the [objects](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/#objects) type to "maps". I asked the team about this a while back, and no one objected. It is already documented as "map" in the component reference docs.
* I could add examples of how to use the "targets.merge"/"map.merge" function with non-discovery functionality. For example, we could get two maps from [encoding.from_json](https://grafana.com/docs/alloy/latest/reference/stdlib/encoding/) and merge them.
* It just dawned on me that we can also do filtering with this too. E.g. `map.merge(arr_w_maps, [{"val" = "a"}], ["val"])` will only pass through the maps in `arr_w_maps` which have `"val" = "a"`. It's a bit weird that you can do filtering with a function called merge, but I can't think of a better alternative right now 😄  I could include an example for that too.

In the long term, we could also:
* Make the UI work better with std lib functions.
* Make it possible to gate std lib features behind an "experimental" command line argument.

cc @thampiotr he made the [original suggestion ](https://github.com/grafana/alloy/pull/1541#issuecomment-2312592047)to use std lib.

docs/sources/reference/stdlib/targets.md
ptodev: You're right, this is wrong. I deleted the comment. I added a unit test just to make sure :)

syntax/vm/vm_stdlib_test.go
ptodev: There is an error if the input is in valid:

Screenshot 2024-10-31 at 18 16 42
Screenshot 2024-10-31 at 18 16 09
Screenshot 2024-10-31 at 18 17 41

ptodev:
Maybe `map.combine` might be better than `map.merge`, since the function makes a new map for each combination of LHS map and a RHS map.

docs/sources/reference/stdlib/targets.md
thampiotr: I would say it's not worth the complexity. If there's a demand for this, we can create new function rather than impose extra complexity on everyone who may not even need it.

docs/sources/reference/stdlib/targets.md
```diff
@@ -0,0 +1,35 @@
...
+It takes three inputs:
+
+* The first two inputs are a of type `list(map(string))`. The keys of the map are strings.
```
thampiotr, line 23: We want to be clear that the values that don't have the merge key will be discarded.

docs/sources/reference/stdlib/targets.md
```diff
@@ -0,0 +1,35 @@
...
+[Compatible components]: ../compatibility/#targets
+
+## targets.merge
+
+The `targets.inner_join` function allows you to join two arrays containing maps if certain keys have matching values in both maps.
+It takes three inputs:
+
+* The first two inputs are a of type `list(map(string))`. The keys of the map are strings. 
+  The value for each key could have any Alloy type such as a string, integer, map, or a capsule.
+* The third input is an array containing strings. The strings are the keys whose value has to match for maps to be joined.
```
thampiotr, line 25: We want to be clear that all keys we're joining on need to match, not just one (it's an AND, not an OR).

thampiotr (commented):
I wonder if there could be a way to get live debugging for stdlib functions?

wildum:
> I wonder if there could be a way to get live debugging for stdlib functions?

In the Alloy UI, users can see the arguments and exports so they should already have the possibility to see the inputs and the output of the merge function. It's not live but seeing the latest values should be good enough.

Live debugging is activated per component (you click on a button in the component page to open the stream of data). I'm not sure how it would look like in the UI to have it for a function that's inside a component.

Implementation wise it's also not trivial. The functions are defined in the syntax pkg which should not depend on a particular Alloy feature. We would need some kind of wrapper that plugs the live debugging and runs because it needs to be able to enable/disable the stream depending on user input

syntax/vm/vm_stdlib_test.go
```diff
@@ -39,6 +39,57 @@ func TestVM_Stdlib(t *testing.T) {
 		{"encoding.from_yaml nil field", "encoding.from_yaml(`foo: null`)", map[string]interface{}{"foo": nil}},
 		{"encoding.from_yaml nil array element", `encoding.from_yaml("[0, null]")`, []interface{}{0, nil}},
 		{"encoding.from_base64", `encoding.from_base64("Zm9vYmFyMTIzIT8kKiYoKSctPUB+")`, string(`foobar123!?$*&()'-=@~`)},
+
+		// Map tests
```
thampiotr, line 43: I think we'll need many more test cases. For example the third arg (keys) can have multiple keys in it.

internal/runtime/internal/controller/component_references.go
```diff
@@ -63,6 +64,24 @@ func ComponentReferences(cn dag.Node, g *dag.Graph, l log.Logger, scope *vm.Scop
...
+				if err := featuregate.CheckAllowed(featuregate.StabilityExperimental, minStability, funcName); err != nil {
+					diags = append(diags, diag.Diagnostic{
+						Severity: diag.SeverityLevelError,
+						Message:  err.Error(),
+						StartPos: ast.StartPos(t[0]).Position(),
+						EndPos:   ast.StartPos(t[len(t)-1]).Position(),
+					})
+					continue
+				}
+			}
```
wildum, line 84: this code is not easy to test, ideally we would need stdlib test functions

CHANGELOG.md
```diff
@@ -44,6 +45,8 @@ v1.5.0-rc.0
 - (_Experimental_) Add a `prometheus.write.queue` component to add an alternative to `prometheus.remote_write`
   which allowing the writing of metrics  to a prometheus endpoint. (@mattdurham)
 
+- (_Experimental_) Add the function `arrary.combine_maps` to the stdlib. (@ptodev, @wildum)
+
```
wildum, line 49: I'm adding it directly to the rc changelog to avoid another PR to main to switch it when we patch it

internal/runtime/internal/controller/component_references.go
```diff
@@ -63,6 +64,24 @@ func ComponentReferences(cn dag.Node, g *dag.Graph, l log.Logger, scope *vm.Scop
...
 			level.Warn(l).Log("msg", "this stdlib function is deprecated; please refer to the documentation for updated usage and alternatives", "function", t[0].Name)
+		} else {
+			var tokenNames []string
+			for _, token := range t {
+				tokenNames = append(tokenNames, token.Name)
+			}
+
+			funcName := strings.Join(tokenNames, ".")
+			if scope.IsStdlibExperimental(funcName) {
+				if err := featuregate.CheckAllowed(featuregate.StabilityExperimental, minStability, funcName); err != nil {
```
thampiotr, line 75: So we have `Scope` that knows whether funcName is experimental.
And we have this function - `ComponentReferences` knowing the `minStability`. 

What if we had `Scope` knowing both pieces of information - the `minStability` and the funcNames that are experimental and it would return error or diags right away to here?

If that has some implementation challenges, I'm open to stay at what it is now, not to further delay this. But it seems fragmented and a bit complicated now.

syntax/internal/stdlib/stdlib.go
```diff
@@ -86,7 +94,8 @@ var str = map[string]interface{}{
 }
 
 var array = map[string]interface{}{
-	"concat": concat,
+	"concat":       concat,
+	"combine_maps": mapCombine,
```
thampiotr, line 98:
```suggestion
	"combine_maps": combineMaps,
```
Let's rename the code to match to keep cognitive load under control.

syntax/internal/stdlib/stdlib.go
```diff
@@ -146,6 +155,138 @@ var concat = value.RawFunction(func(funcValue value.Value, args ...value.Value)
...
+		res[key] = val
+	}
+
+	return value.Object(res), nil
+}
+
+// Inputs:
+// args[0]: []map[string]string: lhs array
+// args[1]: []map[string]string: rhs array
+// args[2]: []string:             merge conditions
```
thampiotr, line 207:
```suggestion
// args[2]: []string:            merge conditions
```

syntax/internal/stdlib/stdlib.go
```diff
@@ -146,6 +155,138 @@ var concat = value.RawFunction(func(funcValue value.Value, args ...value.Value)
...
+var mapCombine = value.RawFunction(func(funcValue value.Value, args ...value.Value) (value.Value, error) {
+	if len(args) != 3 {
+		return value.Value{}, fmt.Errorf("inner_join: expected 3 arguments, got %d", len(args))
```
thampiotr, line 210:
```suggestion
		return value.Value{}, fmt.Errorf("combine_maps: expected 3 arguments, got %d", len(args))
```

syntax/internal/stdlib/stdlib.go
```diff
@@ -146,6 +155,138 @@ var concat = value.RawFunction(func(funcValue value.Value, args ...value.Value)
...
+				Expected: value.TypeArray,
+			},
+		}
+	}
+	if args[2].Len() == 0 {
+		return value.Null, value.ArgError{
+			Function: funcValue,
+			Argument: args[2],
+			Index:    2,
+			Inner:    fmt.Errorf("inner_join: merge conditions must not be empty"),
```
thampiotr, line 258:
```suggestion
			Inner:    fmt.Errorf("combine_maps: merge conditions must not be empty"),
```

syntax/internal/value/value.go
```diff
@@ -221,9 +221,11 @@ func (v Value) Text() string {
 	}
 }
 
-// Len returns the length of v. Panics if v is not an array or object.
+// Len returns the length of v. Panics if v is not an array, an object, or nil.
 func (v Value) Len() int {
 	switch v.ty {
+	case TypeNull:
+		return 0
```
thampiotr, line 228: Would this have some weird side-effects, like Len() called on null capsule or sth?

syntax/internal/value/value.go
```diff
@@ -553,3 +555,15 @@ func convertGoNumber(nval Number, target reflect.Type) reflect.Value {
 
 	panic("unsupported number conversion")
 }
+
+func (v Value) Equal(rhs Value) bool {
+	if v.Type() != rhs.Type() {
+		return false
+	}
+
+	if !v.rv.Equal(rhs.rv) {
```
thampiotr, line 564: Works for our shallow equal on strings, but may not work for maps/arrays perhaps where a DeepEqual would be needed? Maybe we should write a comment if that's the case.

syntax/vm/vm_stdlib_test.go
```diff
@@ -39,6 +39,89 @@ func TestVM_Stdlib(t *testing.T) {
...
+		{
+			// Basic case. Integer and string values.
+			"array.combine_maps",
+			`array.combine_maps([{"a" = 1, "b" = 2.2}], [{"a" = 1, "c" = "c1"}], ["a"])`,
+			[]map[string]interface{}{{"a": 1, "b": 2.2, "c": "c1"}},
+		},
+		{
+			// The second map will override a value from the first.
+			"array.combine_maps",
+			`array.combine_maps([{"a" = 1, "b" = 2.2}], [{"a" = 1, "b" = "3.3"}], ["a"])`,
```
thampiotr, line 71: Can we add another test that takes this override behaviour further?
`array.combine_maps([{"a" = 1, "b" = 2.2}], [{"a" = 1, "b" = "3.3"}, {"a" = 1, "b" = "4.4"}], ["a"])` --> `{"a": 1, "b": "4.4"}`.

This is what I tried to capture in the doc. But probably putting this example in the doc would be good too.

thampiotr (commented):
Looking good! Left some comments and would like to take it for a spin. With some more tests and docs I think we can land this.

docs/sources/reference/stdlib/array.md
```diff
@@ -32,3 +32,26 @@ Elements within the list can be any type.
 > array.concat([[1, 2], [3, 4]], [[5, 6]])
 [[1, 2], [3, 4], [5, 6]]
 ```
+
+## array.combine_maps
+
```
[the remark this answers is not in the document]
wildum: I wanted to avoid having two almost identical shared paragraphs but I'm ok to have both if "feature" is too vague for "component". @clayton-cornell what do you think?

syntax/internal/value/value.go
wildum: it panics for arrays : 
```
	t.Run("equal - array", func(t *testing.T) {
		v := value.Array(value.String("aa"), value.String("bb"))
		w := value.Array(value.String("aa"), value.String("bb"))
		require.True(t, v.Equal(w))
	})
```
panic: reflect.Value.Equal: values of type []value.Value are not comparable

syntax/vm/vm_stdlib_test.go
wildum: this results in {"a": 1, "b": "3.3"}, {"a": 1, "b": "4.4"}

syntax/internal/value/value.go
wildum: I documented the types that are not supported

syntax/internal/value/value.go
wildum: I removed this. Looking at the code, I don't think that this can be triggered because the type is checked before every call to Len. Maybe Paulin made this change before checking the types and then forgot about it or I'm missing something

internal/runtime/internal/controller/component_references.go
wildum: I think that makes sense but we would need to bring the featuregate pkg outside of the internal pkg so that the syntax pkg has access to it + some changes in the syntax pkg to adapt it. It's not that difficult but it will bloat this PR. I would suggest to do this in a follow up PR. I'm already planning to refactor the stdlib because the names in the code are not great right now

thampiotr: ok! makes perfect sense, happy to do these later

thampiotr (approved):

PR #1844, opened 2024-10-08
remotecfg: support not_modified response

spartan0x117:
#### PR Description
Adds support for the remote to send a `hash` to be used to determine if the remote config has changed or not, and to ignore the contents when Alloy receives a `not_modified == true` response.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -428,11 +428,13 @@ func (s *Service) getAPIConfig() ([]byte, error) {
 		return nil, err
 	}
 	s.metrics.getConfigTime.Observe(time.Since(start).Seconds())
-	if gcr.Msg.Hash != "" {
-		s.remoteHash = gcr.Msg.Hash
-	}
 	if gcr.Msg.NotModified {
 		return nil, errNotModified
+	}
+	if gcr.Msg.Hash != "" {
+		s.mut.Lock()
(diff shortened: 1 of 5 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -60,10 +63,15 @@ type Service struct {
 	systemAttrs       map[string]string
 	attrs             map[string]string
 	metrics           *metrics
+
+	// This is the hash received from the API. It is used to determine if
+	// the configuration has changed since the last fetch
+	remoteHash string
```
erikbaranowski, line 69: how confident are you that we don't need to cover this with the mutex above?

internal/service/remotecfg/remotecfg.go
```diff
@@ -401,6 +428,12 @@ func (s *Service) getAPIConfig() ([]byte, error) {
 		return nil, err
 	}
 	s.metrics.getConfigTime.Observe(time.Since(start).Seconds())
+	if gcr.Msg.Hash != "" {
+		s.remoteHash = gcr.Msg.Hash
+	}
+	if gcr.Msg.NotModified {
```
erikbaranowski, line 434: super nit: this conditional returns so it can be done before the hash set on line 431 directly above

erikbaranowski (approved):

PR #1861, opened 2024-10-09
Add OpenShift docs

clayton-cornell:
#### PR Description

This PR adds OpenShift deployment instructions to the Alloy docs

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/set-up/install/openshift.md
+++ b/docs/sources/set-up/install/openshift.md
@@ -3,7 +3,7 @@ canonical: https://grafana.com/docs/alloy/latest/set-up/install/openshift/
 description: Learn how to deploy Grafana Alloy on OpenShift
 menuTitle: OpenShift
 title: Deploy Grafana Alloy on OpenShift
-weight: 250
+weight: 530
 ---
 
 # Deploy {{% param "FULL_PRODUCT_NAME" %}} on OpenShift
@@ -15,9 +15,9 @@ You can deploy {{< param "PRODUCT_NAME" >}} on the Red Hat OpenShift Container P
 * These steps assume you have a working OCP environment.
 * You can adapt the suggested policies and configuration to meet your specific needs and [security][] policies.
 
-## Configure Role-Based Access Control
+## Configure RBAC
 
-You must configure RBAC to allow secure access to Kubernetes and OCP resources.
+You must configure Role-Based Access Control (RBAC) to allow secure access to Kubernetes and OCP resources.
 
 1. Download the [rbac.yaml][] configuration file. This configuration file defines the OCP verbs and permissions for {{< param "PRODUCT_NAME" >}}.
 1. Review the `rbac.yaml` file and adapt as needed for your local environment. Refer to [Managing Role-based Access Control (RBAC)][rbac] topic in the OCP documentation for more information about updating and managing your RBAC configurations.
@@ -27,9 +27,9 @@ You must configure RBAC to allow secure access to Kubernetes and OCP resources.
 You must configure {{< param "PRODUCT_NAME" >}} to [run as a non-root user][nonroot].
 This ensures that {{< param "PRODUCT_NAME" >}} complies with your OCP security policies.
 
-## Apply Security Context Constraints
+## Apply SSC
 
-OCP uses Security Context Constraints (SCC) to control pod permissions.
+OCP uses Security Context Constraints (SCC) to control Pod permissions.
 Refer to [Managing security context constraints][scc] for more information about how you can define and enforce these permissions.
 This ensures that the pods running {{< param "PRODUCT_NAME" >}} comply with OCP security policies.
 
@@ -87,6 +87,8 @@ spec:
            emptyDir: {}
 ```
 
+Refer to [Deploy {{< param "FULL_PRODUCT_NAME" >}}][deploy] for more information about deploying {{< param "PRODUCT_NAME" >}} in your environment.
+
 ## Next steps
 
 - [Configure {{< param "PRODUCT_NAME" >}}][Configure]
@@ -97,3 +99,4 @@ spec:
 [nonroot]: ../../../configure/nonroot/
 [scc]: https://docs.openshift.com/container-platform/latest/authentication/managing-security-context-constraints.html
 [Configure]: ../../../configure/linux/
+[deploy]: ../../deploy/
--- a/docs/sources/set-up/install/openshift.md
+++ b/docs/sources/set-up/install/openshift.md
@@ -57,7 +57,7 @@ You can adapt the SCCs to meet your local requirements and needs.
 The following example shows a SCC configuration that deploys {{< param "PRODUCT_NAME" >}} as a non-root user:
 
 ```yaml
-piVerison: aapps/v1
+apiVersion: aapps/v1
 kind: DaemonSet
 metadata:
   name: alloy-logs
--- a/docs/sources/set-up/install/openshift.md
+++ b/docs/sources/set-up/install/openshift.md
@@ -8,37 +8,33 @@ weight: 250
 
 # Deploy {{% param "FULL_PRODUCT_NAME" %}} on OpenShift
 
-You can deploy {{< param "PRODUCT_NAME" >}} on OpenShift.
-It's important to ensure that Alloy conforms to OpenShift [security][] requirements.
-This includes correctly configuring permissions and security policies to avoid vulnerabilities.
-You can adapt the suggested policies and configuration to meet your specific needs and security policies.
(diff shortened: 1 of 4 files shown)


docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,99 @@
...
+  This SCC is generally not required to deploy {{< param "PRODUCT_NAME" >}} as a non-root user.
+* `FSGroup`: Specifies the fsGroup IDs for file system access.
+  You must configure this constraint to give {{< param "PRODUCT_NAME" >}} group access to the files it needs.
+* `Volumes`: Specifies the persistent volumes used for storage.
+  You must configure this constraint to give {{< param "PRODUCT_NAME" >}} access to the volumes it needs.
+
+The following example shows a SCC configuration that deploys {{< param "PRODUCT_NAME" >}} as a non-root user:
+
+```yaml
+piVerison: aapps/v1
```
thampiotr, line 60:
```suggestion
apiVersion: aapps/v1
```

But also, I can see something slightly different in https://docs.openshift.com/container-platform/3.11/dev_guide/daemonsets.html

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,99 @@
...
+kind: DaemonSet
```
thampiotr, line 61: There are few deployment modes that may apply here. Maybe we should give an example of one only and refer to https://grafana.com/docs/alloy/next/set-up/deploy/ for more info?

LMK what you think and I'll help iron out the details of this YAML, as I think it needs some changes.

docs/sources/set-up/install/openshift.md
clayton-cornell: I saw that too... but I don't know if that's just variations like we see in an Alloy collector config, or if they are elements that should actually be in the SCC YAML.

docs/sources/set-up/install/openshift.md
clayton-cornell: What would be the most common deployment mode?  Definitely agree, pick one and then provide links/info for the reader to adapt to their local needs.

clayton-cornell:
@Toure If you have a chance, can you take a look through this draft and let me know if it looks OK... is it missing anything?

docs/sources/set-up/install/openshift.md
clayton-cornell: @thampiotr I've tweaked the topic a little and fixed a few typos etc. The main remaining issue is to fix up the YAML a bit. I'm not sure what needs to be done here....

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,102 @@
...
+          - containerPort: 12345
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsUser: 473
+            runAsGroup: 473
+            # fsGroup: 1000
+         volumes:
+         - name: log-volume
+           emptyDir: {}
+```
```
thampiotr, line 88: I think we need to provide two files, something like this:

Example SCC definition for running Alloy as non-root user:
```yaml
kind: SecurityContextConstraints
apiVersion: security.openshift.io/v1
metadata:
  name: scc-alloy
runAsUser:
  type: MustRunAs
  uid: 473
fsGroup:
  type: MustRunAs
  uid: 1000
volumes: 
- '*'
users:
- my-admin-user
groups:
- my-admin-group
```

Minimal example of DaemonSet deployment as non-root user

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: alloy-logs
  namespace: monitoring
spec:
  replicas: 1
  selector:
    matchLabels:
      app: alloy-logs
   template:
     metadata:
       lables:
         app: alloy-logs
      spec:
        containers:
        - name: alloy-logs
          image: grafana/alloy:latest
          ports:
          - containerPort: 12345
          # The security context configuration
          securityContext:
            allowPrivilegeEscalation: false
            runAsUser: 473
            runAsGroup: 473
            fsGroup: 1000
         volumes:
         - name: log-volume
           emptyDir: {}
```

I don't have OpenShift environment to test, so I was only able to do my best from reading the docs.

thampiotr (approved):
Left the suggestion for the configs, LGTM if that's applied. Thanks!

docs/sources/set-up/install/openshift.md
clayton-cornell: Thanks @thampiotr  you're awesome :-)  I've added this in to the doc. Reading through, it make sense to me. I'm happy to publish as-is and if there's a problem, I'm sure people will let us know.  

@Toure can you do a quick check of the change?

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+          - containerPort: 12345
+          # The security context configuration
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsUser: 473
+            runAsGroup: 473
+            fsGroup: 1000
+         volumes:
+         - name: log-volume
+           emptyDir: {}
```
Toure, line 88: While emptyDir can be useful for testing, consider giving an example to prevent data lost in production, if the node restarts, the logs will be lost.  If the empty collection is include then a note should be included to make the user aware.

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+       lables:
+         app: alloy-logs
+      spec:
+        containers:
+        - name: alloy-logs
+          image: grafana/alloy:latest
+          ports:
+          - containerPort: 12345
+          # The security context configuration
+          securityContext:
```
Toure, line 81: If the container doesn't need to write to its root filesystem, set `readOnlyRootFilesystem: true` to true for added security.

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+The following example shows a DaemonSet configuration that deploys {{< param "PRODUCT_NAME" >}} as a non-root user:
+
+```yaml
+apiVersion: aapps/v1
+kind: DaemonSet
+metadata:
+  name: alloy-logs
+  namespace: monitoring
+spec:
+  replicas: 1
```
Toure, line 66: I would remove the replicas field as DaemonSets do not use this field. They inherently run one pod per node (with potential nodeSelectors for more granular control). This field will be ignored.

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+metadata:
+  name: scc-alloy
+runAsUser:
+  type: MustRunAs
+  uid: 473
+fsGroup:
+  type: MustRunAs
+  uid: 1000
+volumes: 
+- '*'
```
Toure, line 105: Carefully evaluate if all volume types are required. If not, remove the wildcard * and explicitly list only the necessary volume types. An example such as this:
```
volumes:
  - configMap
  - downwardAPI
  - emptyDir
  - persistentVolumeClaim
  - secret 
```

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: scc-alloy
+runAsUser:
+  type: MustRunAs
+  uid: 473
+fsGroup:
+  type: MustRunAs
+  uid: 1000
```
Toure, line 103: The manifest doesn't specify an SELinux context. Consider adding seLinuxContext with a type like MustRunAs to enforce a specific SELinux context for added security.
```

seLinuxContext:
  type: MustRunAs
  # Specify the appropriate SELinux user, role, type, and level
  # Example:
  # user: system_u
  # role: system_r
  # type: container_t
  # level: s0
```

docs/sources/set-up/install/openshift.md
```diff
@@ -0,0 +1,124 @@
...
+            fsGroup: 1000
+         volumes:
+         - name: log-volume
+           emptyDir: {}
+```
+
+The following example shows an SSC definition that deploys {{< param "PRODUCT_NAME" >}} as a non-root user:
+
+```yaml
+kind: SecurityContextConstraints
```
Toure, line 94: Ensure the alloy pod manifest has allowPrivilegeEscalation: false in the security context to prevent processes within the container from escalating privileges.

```
containers:
   securityContext:
      allowPrivilegeEscalation: false
```

Toure (commented):
@clayton-cornell I left a couple suggestions.

clayton-cornell:
This PR has build issues that are a pain to fix. The new PR is https://github.com/grafana/alloy/pull/2265

Issue #233, opened 2024-02-25
loki.source.kubernetes_events fails to start with large number of events in a namespace ("event watcher exited with error")


towolf:
### What's wrong?

We are using grafana-agent to ship Events off cluster and we are using a system that has a crazy pod churn.

The component worked fine initially, but as our churn increased it just stopped working after startup with:

"event watcher exited with error" 

It is timing out trying to start the informer. 

In the code you have this comment:

https://github.com/grafana/agent/blob/main/component/loki/source/kubernetes_events/kubernetes_events.go#L25-L26

I think for our use case it is not generous enough, becasue the initial sync is too heavy.

I would be cool if we could just start with new events or something and ignore the backlog, if that is at all possible.

### Steps to reproduce

Use the `loki.source.kubernetes_events` in a cluster with a very large amount of events.

Observe timeout error.

### System information

GKE 1.27 and RKE2 1.26

### Software version

docker.io/grafana/agent:v0.39.1

### Configuration

```text
logging {
      level = "debug"
    }

    //
    // Ship Kubernetes Events
    //
    loki.source.kubernetes_events "all" {
      forward_to = [loki.process.events.receiver]
      job_name = "integrations/kubernetes/eventhandler"
      // log_format = "logfmt"
      // namespaces = ["xyz"]
    }

    loki.process "events" {
      forward_to = [loki.write.central.receiver]

      stage.match {
        selector = "{namespace=~\"xzy\"}"

        stage.tenant {
          label = "namespace"
        }
      }

      stage.metrics {
        metric.counter {
          name        = "events_total"
          description = "total number of events"
          prefix      = "agent_"

          match_all         = true
          action            = "inc"
          max_idle_duration = "5m"
        }

        metric.counter {
          name        = "event_bytes_total"
          description = "total bytes of events"
          prefix      = "agent_"

          match_all         = true
          count_entry_bytes = true
          action            = "add"
          max_idle_duration = "5m"
        }
      }
    }

    loki.write "central" {
      endpoint {
        url = "https://xyz/loki/api/v1/push"
        tenant_id = "main"
        basic_auth {
          username = "loki"
          password = "xyz"
        }
      }
      external_labels = {
        cluster = "xyz",
      }
    }
```

### Logs

[...]

PR #1862, opened 2024-10-09
fix: increase timeout for informer on loki.source.podlogs

### LINKED ISSUE #1852: loki.source.podlogs fails to start due to timeout error on busy K8 clusters
We've started leveraging loki.source.podlogs within our AKS cluster to be able to discover pods and tail their logs into Loki. For clusters we have where workloads are lower (i.e. not as many services) it operates without issue.

However one of our busier clusters, we get the following error when Alloy starts up:

`failed to configure informers: Timeout exceeded while configuring informers. Check the connection to the Kubernetes API is stable and that Alloy has appropriate RBAC permissions`

We've confirmed Network Connectivity to the K8s API is stable, and the relevant RBAC permissions are in place. We have a feeling this seems to be linked to https://github.com/grafana/alloy/issues/233 due to the error being pretty much identical (the short-term fix applied in that Issue was to increase the timeout parameter to 10 minutes, and was only applied to the kubernetes_events component and not the podlogs component).

We have also validated Alloy's own pod resources, and they are not hitting limits.

elburnetto-intapp:
#### PR Description

This PR is to fix #1852 where on a busy Kubernetes cluster, the loki.source.podlogs component times-out after 10 seconds with error 'failed to configure informers: Timeout exceeded while configuring informers. Check the connection to the Kubernetes API is stable and that Alloy has appropriate RBAC permissions'.

Fixes #1852 

#### Notes to the Reviewer

Similar issue was seen (#233) where the fix was to increase the timeout on the informer. I have validated this works on a busy K8s cluster and resolves the issue of the component not starting.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ Main (unreleased)
 
 - `pyroscope.scrape` no longer tries to scrape endpoints which are not active targets anymore. (@wildum @mattdurham @dehaansa @ptodev)
 
+- Fixed a bug with `loki.source.podlogs` not starting in large clusters due to short informer sync timeout. (@elburnetto-intapp) 
+
 ### Other changes
 
 - Small fix in UI stylesheet to fit more content into visible table area. (@defanator)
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
LGTM will merged after unit test complete.

Issue #1845, opened 2024-10-08
Some missed metrics with Alloy 1.4.x on Windows Server 2022 Datacenter


halt-avmc:
### What's wrong?

Alloy does not export windows metrics, at least for

- net
- logical_disk
- phisical_disk

while no errors or timeout are reported:

```
# HELP windows_exporter_build_info A metric with a constant '1' value labeled by version, revision, branch, goversion from which windows_exporter was built, and the goos and goarch for the build.
# TYPE windows_exporter_build_info gauge
windows_exporter_build_info{branch="HEAD",goarch="amd64",goos="windows",goversion="go1.22.7",revision="f59e491e5",tags="builtinassets",version="v1.4.2"} 1
# HELP windows_exporter_collector_duration_seconds windows_exporter: Duration of a collection.
# TYPE windows_exporter_collector_duration_seconds gauge
windows_exporter_collector_duration_seconds{collector="logical_disk"} 0
windows_exporter_collector_duration_seconds{collector="net"} 0
windows_exporter_collector_duration_seconds{collector="physical_disk"} 0
# HELP windows_exporter_collector_success windows_exporter: Whether the collector was successful.
# TYPE windows_exporter_collector_success gauge
windows_exporter_collector_success{collector="logical_disk"} 1
windows_exporter_collector_success{collector="net"} 1
windows_exporter_collector_success{collector="physical_disk"} 1
# HELP windows_exporter_collector_timeout windows_exporter: Whether the collector timed out.
# TYPE windows_exporter_collector_timeout gauge
windows_exporter_collector_timeout{collector="logical_disk"} 0
windows_exporter_collector_timeout{collector="net"} 0
windows_exporter_collector_timeout{collector="physical_disk"} 0
# HELP windows_exporter_perflib_snapshot_duration_seconds Duration of perflib snapshot capture
# TYPE windows_exporter_perflib_snapshot_duration_seconds gauge
windows_exporter_perflib_snapshot_duration_seconds 0.0014556
```
and that is all what is shown, there are no metrics at all of
```
windows_net_xxx
windows_logical_disk_xxx
windows_phisical_disk_xxx
```
while other windows metrics (like cpu, system, etc.) are actually exported normally

FYI: **Alloy v1.3.4** and **windows_exporter v0.29.1** are working correctly

### Steps to reproduce

[...]

PR #1865, opened 2024-10-10
[prometheus.exporter.windows] Add missing physical_disk block

ptodev:
A user [reported](https://github.com/grafana/alloy/issues/1845#issuecomment-2404586727) that the `physical_disk` block is missing from the documentation.

I also noticed that in the [Windows exporter code](https://github.com/prometheus-community/windows_exporter/blob/v0.27.3/pkg/types/regexp.go), all the `include` regex arguments have a default value of `".+"`, but in Alloy they are documented as `".*"`. I wonder if this is intentional. Maybe @jkroepke has more context on it?

I also changed the links to the `logical_disk` and `scheduled_task` sections. The website already works ok without this change, but the VS Code markdown viewer doesn't. With this change, navigating to the blocks in VS Code is a bit easier.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
@@ -54,6 +54,7 @@ logical_disk   | [logical_disk][]   | Configures the logical_disk collector.   |
 msmq           | [msmq][]           | Configures the msmq collector.           | no
 mssql          | [mssql][]          | Configures the mssql collector.          | no
 network        | [network][]        | Configures the network collector.        | no
+physical_disk  | [physical_disk][]  | Configures the physical_disk collector.  | no
 printer        | [printer][]        | Configures the printer collector.        | no
 process        | [process][]        | Configures the process collector.        | no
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
```diff
@@ -152,6 +154,12 @@ Name      | Type     | Description                             | Default | Requi
 
 NIC names must match the regular expression specified by `include` and must _not_ match the regular expression specified by `exclude` to be included.
 
+### physical_disk block
+
+Name      | Type     | Description                                     | Default | Required
+----------|----------|-------------------------------------------------|---------|---------
+`exclude` | `string` | Regular expression of physical disc to exclude. | `""`    | no
+`include` | `string` | Regular expression of physical disc to include. | `".*"`  | no
```
clayton-cornell, line 162:
```suggestion
`exclude` | `string` | Regular expression of physical disk to exclude. | `""`    | no
`include` | `string` | Regular expression of physical disk to include. | `".*"`  | no
```

clayton-cornell (approved):
Looks OK. One small suggestion

PR #1593, opened 2024-08-30
Secret Filtering component for Alloy

romain-gaillard:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #1877, opened 2024-10-11
fix: Disable generic API key rule by default in loki.secretfilter

romain-gaillard:
#### PR Description
We have [recently merged](https://github.com/grafana/alloy/pull/1593#event-14380083833) a new component for alloy called `loki.secretfilter`. This component hasn't been released yet and will be in 1.5.0.

However, we have noticed that the inclusion of the `generic-api-key` rule by default creates too many false positive secrets, so we're changing the config so that it's disabled by default.

#### PR Checklist

- [ ] CHANGELOG.md updated -> Don't think it applies since the component hasn't been included in a release yet.
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.secretfilter.md
+++ b/docs/sources/reference/components/loki/loki.secretfilter.md
@@ -40,7 +40,7 @@ Name                     | Type                 | Description                   
 `gitleaks_config`        | `string`             | Path to the custom `gitleaks.toml` file.        | Embedded Gitleaks file           | no
 `types`                  | `map(string)`        | Types of secret to look for.                    | All types                        | no
 `redact_with`            | `string`             | String to use to redact secrets.                | `<REDACTED-SECRET:$SECRET_NAME>` | no
-`exclude_generic`        | `bool`               | Exclude the generic API key rule.               | `false`                          | no
+`include_generic`        | `bool`               | Include the generic API key rule.               | `false`                          | no
 `allowlist`              | `map(string)`        | List of regexes to allowlist matching secrets.  | `{}`                             | no
 `partial_mask`           | `number`             | Show the first N characters of the secret.      | `0`                              | no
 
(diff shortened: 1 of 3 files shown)


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -51,7 +51,7 @@ The `types` argument is a map of secret types to look for. The values are used a
 
 The `redact_with` argument is a string that can use variables such as `$SECRET_NAME` (replaced with the matching secret type) and `$SECRET_HASH`(replaced with the sha1 hash of the secret).
 
-The `exclude_generic` argument is a boolean that excludes the generic API key rule in the Gitleaks configuration file if set to `true`.
+The `include_generic` argument is a boolean that includes the generic API key rule in the Gitleaks configuration file if set to `true`. It is disabled by default because it can generate false positives.
```
wildum, line 54: should not remove it completely if it's not good?

romain-gaillard: It can be really useful! But because it's a generic secret-detection regex, it's definitely not perfect and some logs will have a format that sometimes triggers it (especially when they contain keywords like `secret`).

My opinion is that it should be disabled by default and enabled explicitly when the user has confidence that their log format doesn't generate false positives (or they excluded them), so that it can start picking up real secrets that weren't caught by the more specialised rules.

So in short, it can definitely be useful but requires a bit more work from the user.

wildum (approved):

wildum:
LGTM, waiting on @clayton-cornell's approval before merging

docs/sources/reference/components/loki/loki.secretfilter.md
clayton-cornell, line 54:
```suggestion
The `include_generic` argument is a boolean that includes the generic API key rule in the Gitleaks configuration file if set to `true`. It's disabled by default because it can generate false positives.
```

PR #1893, opened 2024-10-16
Fix dead link for PodLogs RelabelConfig

TheoBrigitte:
This PR fixes a dead link on the PodLog page which points to the Prometheus operator relabeling config.


clayton-cornell (approved):
Looks good. Thanks for catching this !

PR #1903, opened 2024-10-16
WIP IGNORE

mattdurham:


internal/component/prometheus/operator/common/component_test.go
```diff
@@ -94,25 +96,25 @@ func TestRunExit(t *testing.T) {
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	cmpRunExited := atomic.Bool{}
 	cmpRunExited.Store(false)
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
+		wg.Done()
```
ptodev, line 103: Not sure why `wg.Done()` is in the beginning of the goroutine? Does it guarantee anything?

internal/component/prometheus/operator/common/component_test.go
```diff
@@ -94,25 +96,25 @@ func TestRunExit(t *testing.T) {
...
 		require.NoError(t, err)
 		cmpRunExited.Store(true)
-		fmt.Println("component.Run exited")
 	}()
-
+	wg.Wait()
 	// Stop the component.
 	// It shouldn't stop immediately, because the CRD Manager is hung.
 	cancelFunc()
+	<-stopRun
```
ptodev, line 112: The idea behind the test is to sleep for a bit after `cancelFunc()` and make sure Run() hasn't exited, up until we call `<-stopRun`.

internal/component/prometheus/operator/common/component.go
```diff
@@ -62,55 +62,70 @@ func (c *Component) CurrentHealth() component.Health {
 	return c.health
 }
 
+type cancelHandler struct {
+	mut            sync.RWMutex
```
ptodev, line 66: Is it really necessary to use a mutex? I thought only one case in the switch statement can be active at a given time.

internal/component/prometheus/operator/common/component_test.go
mattdurham: It ensures that we have entered the go routine, its still not perfect timing but hasnt failed in my testing.

internal/component/prometheus/operator/common/component.go
mattdurham: I did this out of an abundance of caution and to ensure race didnt trigger. It may not be needed.

internal/component/prometheus/operator/common/component_test.go
mattdurham: Added the sleep back in.

ptodev:
```suggestion
	if cmpRunExited.Load() {
		require.Fail(t, "component.Run exited")
	}

	<-stopRun
```

Thank youm but we also need the check to make sure the component didn't exit after `time.Sleep` and before `<-stopRun`.

ptodev (approved):
Thank you so much! I added one minor comment. Other than that, I'd be happy to merge it into my branch.

ptodev:
Closing this PR as I implemented some of the suggestions in the [original one](https://github.com/grafana/alloy/pull/1778).

PR #1912, opened 2024-10-17
Add windows integration test

### LINKED ISSUE #1896: Ensure coverage of windows exporter
### Request

Ensure that the most common scenarios in windows exporter can start, and serve metrics. This likely will not be exhaustive but instead more of a sanity check.

### Use case

We have had issues in the past with specific collectors having issues starting and/or serving metrics.

wildum:
#### PR Description

Add windows integration tests github action and an integration tests for the default collectors of the windows exporter.

The Setup-go step is super slow... (https://github.com/actions/setup-go/issues/495)

Fixes #1896

### DIFF
--- a/.github/workflows/integration-tests-win.yml
+++ b/.github/workflows/integration-tests-win.yml
@@ -1,0 +1,21 @@
+name: Integration Tests Windows
+on:
+  # Run tests on main just so the module and build cache can be saved and used
+  # in PRs. This speeds up the time it takes to test PRs dramatically.
+  # (More information on https://docs.github.com/en/enterprise-server@3.6/actions/using-workflows/caching-dependencies-to-speed-up-workflows)
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  run_tests:
+    runs-on: windows-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Setup Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: "1.22"
+    - name: Run tests
+      run: make integration-test
--- a/internal/cmd/integration-tests/main.go
+++ b/internal/cmd/integration-tests/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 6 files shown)


.github/workflows/integration-tests-win.yml
```diff
@@ -0,0 +1,21 @@
+name: Integration Tests Windows
+on:
+  # Run tests on main just so the module and build cache can be saved and used
+  # in PRs. This speeds up the time it takes to test PRs dramatically.
+  # (More information on https://docs.github.com/en/enterprise-server@3.6/actions/using-workflows/caching-dependencies-to-speed-up-workflows)
+  push:
+    branches:
+      - main
+  pull_request:
```
thampiotr, line 9: Doesn't this enable workflow for pull requests? is this temporary?

.github/workflows/integration-tests-win.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  run_tests:
+    runs-on: windows-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Setup Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: "1.22"
```
thampiotr, line 19: Would that become another place where we need to update Go version? What's the process currently to update these?

internal/cmd/integration-tests/main.go
```diff
@@ -30,26 +31,37 @@ func main() {
 
 func runIntegrationTests(cmd *cobra.Command, args []string) {
 	defer reportResults()
-	defer cleanUpEnvironment()
+
+	testFolder := "./tests/"
+	alloyBinaryPath := "../../../../../build/alloy"
+	alloyBinary := "build/alloy"
```
thampiotr, line 37: Let's switch to using https://pkg.go.dev/path#Join

internal/cmd/integration-tests/main.go
```diff
@@ -30,26 +31,37 @@ func main() {
...
+
+	if runtime.GOOS == "windows" {
+		testFolder = "./tests-windows/"
+		alloyBinaryPath = "..\\..\\..\\..\\..\\build\\alloy.exe"
+		alloyBinary = "build/alloy.exe"
+	} else {
+		setupEnvironment()
+		defer cleanUpEnvironment()
```
thampiotr, line 45: Why no setup or clean up on windows?

internal/cmd/integration-tests/main.go
```diff
@@ -30,26 +31,37 @@ func main() {
...
+		buildAlloy(alloyBinary)
 	}
-	setupEnvironment()
 
 	if specificTest != "" {
 		fmt.Println("Running", specificTest)
-		if !filepath.IsAbs(specificTest) && !strings.HasPrefix(specificTest, "./tests/") {
-			specificTest = "./tests/" + specificTest
+		if !filepath.IsAbs(specificTest) && !strings.HasPrefix(specificTest, testFolder) {
+			specificTest = testFolder + specificTest
```
thampiotr, line 55: https://pkg.go.dev/path#Join

internal/cmd/integration-tests/utils.go
```diff
@@ -88,14 +84,17 @@ func runSingleTest(testDir string, port int) {
 		}
 	}
 
+	// sleep for a few seconds before deleting the files to make sure that they are not use anymore
+	time.Sleep(5 * time.Second)
+
```
thampiotr, line 89: Instead of sleeps, can we have a retry and backoff (without panic on first failure) to remove all files?
Sleeps are notoriously fragile and should be avoided. A slow machine can sometimes hang for 5 seconds.

.github/workflows/integration-tests-win.yml
wildum: yep, I did not change it since our discussion, I will remove it before merging to make sure that the test is running properly

.github/workflows/integration-tests-win.yml
wildum: yes, and afaik it's just "search and replace" through the codebase :/

internal/cmd/integration-tests/main.go
wildum: can't use the linux images in the windows runner so we are quite limited in terms of environment. For now not using docker and only running one Alloy to run the windows exporter seems to be sufficient. I can rename the function to "setupDockerEnv" and "cleanupDockerEnv" to make it more clear

wildum: Not sure that it would be really worth it because besides the docker part which is quite small, they would re-use everything. Also I don't think that we need a linux standalone. And maybe one day the windows runner will be able to use WSL 2 and it will work with Mimir linux image or at some point we may want to run a docker windows image and the difference will be gone.
I agree that now it looks a bit convoluted though. I can structure the test better so that it looks less chaotic

PR #1914, opened 2024-10-17
feat: add prometheus.write.queue bearer token auth

### LINKED ISSUE #1913: Add bearer auth for prometheus.write.queue
### Request

Currently, prometheus.write.queue only supports Basic Auth, would be nice to also have bearer auth there, same way as in prometheus.remote_write component.

### Use case

For us, we are using Prometheus managed solution, which only allows using bearer auth, so basic auth won't work here, therefore we need to be able to use bearer auth.

freak12techno:

Fixes #1913 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.write.queue.md
+++ b/docs/sources/reference/components/prometheus/prometheus.write.queue.md
@@ -83,16 +83,17 @@ The `endpoint` block describes a single location to send metrics to. Multiple
 
 The following arguments are supported:
 
-Name | Type | Description                                                      | Default | Required
----- | ---- |------------------------------------------------------------------| ------ | --------
-`url` | `string` | Full URL to send metrics to.                                     | | yes
-`write_timeout` | `duration` | Timeout for requests made to the URL.                            | `"30s"` | no
-`retry_backoff` | `duration` | How often to wait between retries.                               | `1s` | no
-`max_retry_attempts` | Maximum number of retries before dropping the batch. | `0`                                                              | no
-`batch_count` | `uint` | How many series to queue in each queue.                          | `1000` | no
-`flush_interval` | `duration` | How often to wait until sending if `batch_count` is not trigger. | `1s` | no
-`parallelism` | `uint` | How many parallel batches to write.                              | 10 | no
-`external_labels` | `map(string)` | Labels to add to metrics sent over the network.                  | | no
+Name | Type | Description                                                        | Default | Required
+---- | ---- |--------------------------------------------------------------------| ------ | --------
+`url` | `string` | Full URL to send metrics to.                                       | | yes
+`bearer_token` | `secret` | Bearer token to authenticate with.                                 |  | no
+`write_timeout` | `duration` | Timeout for requests made to the URL.                              | `"30s"` | no
+`retry_backoff` | `duration` | How often to wait between retries.                                 | `1s` | no
+`max_retry_attempts` | Maximum number of retries before dropping the batch. | `0`                                                                | no
+`batch_count` | `uint` | How many series to queue in each queue.                            | `1000` | no
+`flush_interval` | `duration` | How often to wait until sending if `batch_count` is not triggered. | `1s` | no
+`parallelism` | `uint` | How many parallel batches to write.                                | 10 | no
+`external_labels` | `map(string)` | Labels to add to metrics sent over the network.                    | | no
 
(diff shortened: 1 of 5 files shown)


freak12techno:
@mattdurham mind taking a look?

mattdurham:
Would like to see a validation so that you cannot set both bearer token and basic auth, with a test to catch that. Otherwise looks good!

internal/component/prometheus/write/queue/types/network.go
```diff
@@ -2,6 +2,7 @@ package types
 
 import (
 	"context"
+	"github.com/grafana/alloy/syntax/alloytypes"
```
mattdurham, line 5: This likely fails the linter, generally "core" golang imports are separated from other imports.

mattdurham: It didnt..coolio :)

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -83,16 +83,17 @@ The `endpoint` block describes a single location to send metrics to. Multiple
...
-`batch_count` | `uint` | How many series to queue in each queue.                          | `1000` | no
-`flush_interval` | `duration` | How often to wait until sending if `batch_count` is not trigger. | `1s` | no
-`parallelism` | `uint` | How many parallel batches to write.                              | 10 | no
-`external_labels` | `map(string)` | Labels to add to metrics sent over the network.                  | | no
+Name | Type | Description                                                        | Default | Required
+---- | ---- |--------------------------------------------------------------------| ------ | --------
+`url` | `string` | Full URL to send metrics to.                                       | | yes
+`bearer_token` | `secret` | Bearer token to authenticate with.                                 |  | no
+`write_timeout` | `duration` | Timeout for requests made to the URL.                              | `"30s"` | no
+`retry_backoff` | `duration` | How often to wait between retries.                                 | `1s` | no
```
clayton-cornell, line 91: How often or how long?

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -83,16 +83,17 @@ The `endpoint` block describes a single location to send metrics to. Multiple
...
+`max_retry_attempts` | Maximum number of retries before dropping the batch. | `0`                                                                | no
+`batch_count` | `uint` | How many series to queue in each queue.                            | `1000` | no
+`flush_interval` | `duration` | How often to wait until sending if `batch_count` is not triggered. | `1s` | no
```
clayton-cornell, line 94: Same question.. often or long? If we say "how often" it doesn't make sense to me.. Wait 6 times (would be how often) vs Wait 2 minutes (would be how long).

clayton-cornell (dismissed):

docs/sources/reference/components/prometheus/prometheus.write.queue.md
mattdurham: @clayton-cornell I dont think he changed most of these, I think adding bearer_token triggered them to look different. I can create a separate PR for these doc comments related to backoff and flush interval.

freak12techno: Yeah I didn't, it's my IDE that apparently tried to format it, only thing I changed is adding this line and also fixing the typo (trigger -> triggered) in another one.

freak12techno: I might change it here as well though if you prefer

mattdurham:
@freak12techno I will follow up on the docs stuff later, has long as @clayton-cornell is good with the bearer token text.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
mattdurham: Handling it in #1917

clayton-cornell (approved):
LGTM, no blocks from docs side

mattdurham (approved):
Lgtm

freak12techno:
okay @mattdurham I didn't get a chance to add tests + validations before this got merged :( should I make another PR adding it, or would you add it in your PR?

mattdurham:
I can handle it, I jumped the gun on it

PR #1919, opened 2024-10-18
Update snmp doc

wildum:
It's likely that it won't work if you try to use a snmp.yml file that does not match the snmp exporter version in your Alloy. This change will redirect users to the version of the snmp exporter that matches the one used in Alloy

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
@@ -48,7 +48,7 @@ Omitted fields take their default values.
 | `targets`     | `list(map(string))`  | SNMP targets.                                    |         | no       |
 
 The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
-Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter#generating-configuration) for details on how to generate a configuration file.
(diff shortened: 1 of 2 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -48,7 +48,7 @@ Omitted fields take their default values.
 | `targets`     | `list(map(string))`  | SNMP targets.                                    |         | no       |
 
 The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
-Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter#generating-configuration) for details on how to generate a configuration file.
+Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter/tree/v0.26.0?tab=readme-ov-file#configuration) for details on how to generate a configuration file.
```
ptodev, line 51: It might be best to add an `SNMP_EXP_VERSION` to `docs/sources/_index.md.t`, and to also use it in the link to the upsteram repo at the top of the page, as well as in the `{{< admonition type="note" >}}`

wildum: thanks for the suggestion, that's much better indeed. I added a comment in the go.mod also for the windows exporter in case there is an update to also change the version. Btw it seems that it's already out of sync, shouldn't it be v0.27.4 in the doc instead of v0.27.3?

ptodev (approved):

PR #1931, opened 2024-10-18
fix: Lock `loki.secretfilter`'s mutex for reading while processing logs

romain-gaillard:
#### PR Description
Lock the component's mutex for reading while processing logs to avoid concurrency issues when the component is updated.

None

#### Notes to the Reviewer
I initially forgot to lock the component's mutex for reading while processing the logs, so I guess that could cause issues during component updates.

The component hasn't been released yet so I suppose updating the changelog isn't needed.

#### PR Checklist

N/A

### DIFF
--- a/internal/component/loki/secretfilter/secretfilter.go
+++ b/internal/component/loki/secretfilter/secretfilter.go
@@ -145,6 +145,7 @@ func (c *Component) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case entry := <-c.receiver.Chan():
+			c.mut.RLock()
 			// Start processing the log entry to redact secrets
 			newEntry := c.processEntry(entry)
 			if c.debugDataPublisher.IsActive(componentID) {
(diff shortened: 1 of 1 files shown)


romain-gaillard:
@grafana/grafana-alloy-maintainers hey team! Could you please have a look at this short PR when you have some time? 😄

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -154,10 +155,12 @@ func (c *Component) Run(ctx context.Context) error {
 			for _, f := range c.fanout {
```
wildum, line 155: nit: I think that technically you can release the lock before the for loop because the processing is already done. This way you release a little bit earlier and you only have one release call. Not very important though, feel free to ignore

wildum (approved):
lgtm, just a nit

internal/component/loki/secretfilter/secretfilter.go
romain-gaillard: I think I did this based on the behaviour of other components I saw, where they kept the lock while iterating over the fanouts (in case that would be changed by the config update)

PR #1937, opened 2024-10-21
docs: Fix broken link in writing exporter components

Pokom:
#### PR Description

The link for all.go was pointing to an older version before `internal` was added. This updates it so that it points to the approriate file.


clayton-cornell (approved):
Thanks for fixing this! LGTM

PR #1982, opened 2024-10-28
Fix fake minimum capacity

mattdurham:
Fixes two issues, one where the endpoint mailbox had an invalid capacity, data handles are just pointers to a the file, so its fine for them to be unlimited. The series mailbox isnt a big deal since most of the time you want a batchsize greater than 64 (internally the default of the actor buffer). Adding the optaschan removes the buffer and enforces the strong capacity.

### DIFF
--- a/internal/component/prometheus/write/queue/network/manager.go
+++ b/internal/component/prometheus/write/queue/network/manager.go
@@ -37,7 +37,7 @@ func New(cc types.ConnectionConfig, logger log.Logger, seriesStats, metadataStat
 		loops:  make([]*loop, 0, cc.Connections),
 		logger: logger,
 		// This provides blocking to only handle one at a time, so that if a queue blocks
-		// it will stop the filequeue from feeding more. Without as chan the minimum capacity is actually a 64 item buffer.
+		// it will stop the filequeue from feeding more. Without OptAsChan the minimum capacity is actually a 64-item buffer.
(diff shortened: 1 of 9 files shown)


internal/component/prometheus/write/queue/network/manager.go
```diff
@@ -37,9 +37,9 @@ func New(cc types.ConnectionConfig, logger log.Logger, seriesStats, metadataStat
 		loops:  make([]*loop, 0, cc.Connections),
 		logger: logger,
 		// This provides blocking to only handle one at a time, so that if a queue blocks
-		// it will stop the filequeue from feeding more.
-		inbox:       actor.NewMailbox[*types.TimeSeriesBinary](actor.OptCapacity(1)),
-		metaInbox:   actor.NewMailbox[*types.TimeSeriesBinary](actor.OptCapacity(1)),
+		// it will stop the filequeue from feeding more. Without as chan the minimum capacity is actually a 64 item buffer.
```
wildum, line 40:
```suggestion
		// it will stop the filequeue from feeding more. Without OptAsChan the minimum capacity is actually a 64-item buffer.
```

wildum (approved):
just a nit for the comment else lgtm

freak12techno:
JFYI I tested it locally (starting Alloy, disabling internet, waiting for some time, restarting Alloy, enabling internet back) and it works as expected, so this seems to fix what I have been facing.

PR #1990, opened 2024-10-29
fix: parser for azureeventhubs message without time field

andriikushch:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/azure_event_hubs/internal/parser/parser.go
+++ b/internal/component/loki/source/azure_event_hubs/internal/parser/parser.go
@@ -36,7 +36,9 @@ func (l azureMonitorResourceLogs) validate() error {
 // azureMonitorResourceLog used to unmarshal common schema for Azure resource logs
 // https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/resource-logs-schema
 type azureMonitorResourceLog struct {
-	Time          string `json:"time"`
+	Time string `json:"time"`
+	// Some logs have `time` field, some have `timeStamp` field : https://github.com/grafana/loki/issues/14176
+	TimeStamp     string `json:"timeStamp"`
 	Category      string `json:"category"`
 	ResourceID    string `json:"resourceId"`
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Main (unreleased)
 ### Bugfixes
 
 - Fixed a bug in `import.git` which caused a `"non-fast-forward update"` error message. (@ptodev)
+- Fixed a bug in `azure_event_hubs` parser for messages without `time` field. (@andriikushch)
```
wildum, line 51:
```suggestion
- Support logs that have a `timestamp` field instead of a `time` field for the `loki.source.azure_event_hubs` component. (@andriikushch)
```
I think that your changelog entry is a bit obscure for the users. I suggested an alternative, feel free to modify as you want. Also maybe I would put it in enhancement and not bugfix

wildum (approved):
lgtm, just a nit for the changelog

CHANGELOG.md
andriikushch: Thank you @wildum! I will merge the suggestion and move it to the enhancement section

wildum:
Sorry one more thing before merging: at the top of the parser file there is a comment saying that the code is copied from Loki with a ref to the version. I see that now the fix has been merged in loki, could you update the ref of the commit to match the current state in Loki please?

andriikushch:
> Sorry one more thing before merging: at the top of the parser file there is a comment saying that the code is copied from Loki with a ref to the version. I see that now the fix has been merged in loki, could you update the ref of the commit to match the current state in Loki please?

Sure 👍 

I have updated a link to:

- https://github.com/grafana/loki/commit/2e62abbf47c47041027baf240722b3d76e7bd9a3

This commit contains the same fix and was recently merged into the Loki project.

wildum:
Thanks!

PR #1991, opened 2024-10-29
Add proxy_url to otelcol.exporter.otlphttp

### LINKED ISSUE #1892: add proxy_url to otelcol.exporter.otlp components
### Request

the otelcol.exporter.otlp & otelcol.exporter.otlphttp components do not have an individual proxy option.  it appears that it is an undocumented feature https://github.com/open-telemetry/opentelemetry-java/issues/6204

i was expecting something similar to the prometheus.remote_write client block. where you would do something like this:

`otelcol.exporter.otlp "default" {
  client {
    endpoint="https://otlp.grafana.net"
    proxy_url="http://internal.proxy.com:1234"
    auth=otel.auth.basic.default
  }
}
otelcol.auth.basic "default" {
  user="username"
  password="password"
}
`

### Use case

the only way to do this now is to set as an extraEnv the HTTPS_PROXY, this is really unclear, adding this to the block would clean it up, and make it easier to use proxies.  In addition, this would allow support for multiple proxies

wildum:
#### PR Description

It seems like this attribute was missed. It's been requested and I don't see a reason why we would not have it in Alloy.

Fixes #1892

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [na] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ Main (unreleased)
 
 - Add Prometheus bearer authentication to a `prometheus.write.queue` component (@freak12techno)
 
+- Add `proxy_url` to `otelcol.exporter.otlphttp`. (@wildum)
+
 ### Bugfixes
 
 - Fixed a bug in `import.git` which caused a `"non-fast-forward update"` error message. (@ptodev)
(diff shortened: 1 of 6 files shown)


clayton-cornell (approved):
Doc changes are small. Looks OK for docs.

dehaansa (approved):

ptodev (approved):

PR #1998, opened 2024-10-29
fix documentation for helm extraPorts

### LINKED ISSUE #1997: Docs feedback: Helm chart alloy.extraPorts.hostPort should be port instead
### URL

https://artifacthub.io/packages/helm/grafana/alloy#alloy-extraports

### Feedback

If I configure `alloy.extraPorts` according to the documentation as follows:
```
alloy:
  extraPorts:
    - targetPort: 4317
      hostPort: 4317
      name: otlp-grpc
```
I get this error from kubernetes after helm upgrade:
```
Error: UPGRADE FAILED: cannot patch "alloy" with kind Service: Service "alloy" is invalid: spec.ports[1].port: Invalid value: 0: must be between 1 and 65535, inclusive
```
This is because the generated manifest looks like this:
```
...
    ports:
      - name: http-metrics
         port: 12345
         targetPort: 12345
         protocol: "TCP"
      - name: otlp-grpc
        port: 
        targetPort: 4317
        protocol: TCP
.... 
```
So I changed configuration to the following:
```
alloy:
  extraPorts:
    - targetPort: 4317
      port: 4317
      name: otlp-grpc
```
...and it worked well!

This should be reflected to the documentation.

jkruke:

Fixes #1997 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


clayton-cornell:
@jkruke Don't forge to sign the CLA so we can move forward with your PR.

jkruke:
@clayton-cornell The CLA is now signed and I ran the helm tests. 
Due to the results, I realized that k8s pod `ports` need `hostPort`, but a service needs `port`.  
So I changed the service templates to use `hostPort` instead of the (not documented) `port`.

clayton-cornell:
@grafana/grafana-alloy-maintainers Can one of you take a look at this one? My Helm knowledge isn't strong enough to check/verify these changes.

wildum:
@petewall could you check this change to the helm chart please

petewall (commented):
Is this a breaking change for anyone who has deployed alloy with the `port` and not `hostPort`?

clayton-cornell:
@mattdurham This one seems to be stuck in my queue. What do we need to do to get it unstuck?

blewis12:
hi folks, i'm taking a quick look at older PR's that we could get merged (or close) and saw this one. To clarify, the issue here that the docs reference `extraPorts.hostPort` when in reality we use `extraPorts.port`. This PR looks to be renaming `port` to `extraPort` which would be a breaking change, could we not instead make no changes to the code and just change the documentation? Let me know if i'm missing some info there

I see there was an idea to keep both ports/hostPost, but i'm not sure I understand the benefit of this

PR #1886, opened 2024-10-15
feat: add `pyroscope.receive_http` component for profile handling

marcsanmi:
#### PR Description

This PR introduces the new `pyroscope.receive_http` component, which allows for receiving and forwarding Pyroscope profiles. Key features and changes include:

- HTTP server that listens for profile data on the `/ingest` endpoint
- Ability to forward received profiles to multiple `pyroscope.write` components
- Tests covering various scenarios (profile sizes, error conditions, etc.)
- Updated documentation explaining component usage and configuration options

#### Which issue(s) this PR fixes

Closes https://github.com/grafana/alloy/issues/270

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated

PR #2001, opened 2024-10-30
feat: add external labels support to pyroscope.write AppendIngest

marcsanmi:
#### PR Description

After implementing  in #1886, I noticed we needed to add support for `external_labels` in the `pyroscope.write` component's new `AppendIngest` implementation. This PR adds that support.

When a profile is ingested with labels in the name parameter (e.g., `name=my.app{env=staging}`), these labels are now merged with the component's configured external labels.

#### Notes to the Reviewer
* No documentation changes needed as `external_labels` is already documented in the `pyroscope.write` component
* No CHANGELOG.md entry added as this extends #1886 and can be included in the same release entry. Let me know if you'd prefer a separate changelog entry.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -373,7 +373,17 @@ func (f *fanOutClient) AppendIngest(ctx context.Context, profile *pyroscope.Inco
 			}
 
 			u.Path = path.Join(u.Path, profile.URL.Path)
-			u.RawQuery = profile.URL.RawQuery
+
+			// Handle query param and external labels
+			query := profile.URL.Query()
+			if nameParam := query.Get("name"); nameParam != "" {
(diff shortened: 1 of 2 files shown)


korniltsev (changes requested):
I would prefer a copy-paste of the parser from here https://github.com/grafana/pyroscope/blob/e17b0d0164e0e0806c33a2c12de9b63f6b1e872a/pkg/og/storage/segment/key.go#L36
Or maybe somehow reuse the code (maybe move it to pyroscope/api module ?)

I think `parseLabels("foo{label=value,    label2=value2}")` will produce different results then the pyroscope server's handler with `segmet.ParseKey`. We better have no parser differential here.

korniltsev (approved):
:+1:

thampiotr (approved):

PR #2011, opened 2024-10-31
add buffer to promise like channel

vladopajic:
#### PR Description

add buffer to promise like channel so main goroutine loop doesn't block

no issue, just performance improvement 

#### Notes to the Reviewer

hey @mattdurham, I was reviewing usage of go-actor and just noticed this so I wanted to share it with you.

in `DoWork` caller is being notified  after config gets updated. because this channel is not buffered this goroutine will get blocked because caller is also blocked at this point (it's waiting on response from `done` channel). main loop will remain blocked until goroutine that made this request is resumed again (which can take time theoretically speaking).

probably the best approach when implementing logic of `DoWork` (aka workers) is to never have it's code blocked unless it is blocked on reading data from inbound mailboxes - in other words blocked because it's waiting on work.  

in this case, this can be easily achieved by making `done`, promise like channel, to have buffer of size 1. this will ensure that `DoWork` can write to it without waiting on this goroutine to be ready. 

--- 

`defer close(done)` was removed because it is unnecessary, after data transmission, this channel is no longer used so no one will care if this channel is closed or not - eventually it will be garbage collected.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/prometheus/write/queue/network/manager.go
+++ b/internal/component/prometheus/write/queue/network/manager.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"context"
+
 	"github.com/go-kit/log"
 	"github.com/grafana/alloy/internal/component/prometheus/write/queue/types"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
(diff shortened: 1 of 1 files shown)


mattdurham:
Since your here :) having an issue on where I don't know the number of callers, so if I close call and still have work coming in I get a panic which is not ideal. My idea is for actors that behave as a channel to not call close and not set the panic send handler. Thoughts?

vladopajic:
You mean mailboxes? To me it sound like you are using mailbox and sometimes you stop mailbox but there is more work coming in? If this is true then just just have to additional logic that will not send data to mailbox is something is sopped. 

Or there is different issue, were you are stopping in different order. Inbound routine, that sends work, should be stopped first, before mailbox.

PR #1679, opened 2024-09-13
Add label cluster_name to alloy_config_hash and some cluster_ metrics

wildum:
#### PR Description

- I added the label "cluster_name" to the metric alloy_config_hash and removed redundant metrics logic in config_metrics.
- I bumped the ckit version to also have the "cluster_name" label on relevant cluster_ metrics
- I updated the clustering alerts to take the new cluster_name label into account  

#### Which issue(s) this PR fixes

Fixes #1364

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #2018, opened 2024-11-01
Mixin: Add alloy cluster label filters to dashboards

gaantunes:
#### PR Description

This is including the new `cluster_name` label introduced [here](https://github.com/grafana/alloy/pull/1679) to all dashboards.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


gaantunes:
@wildum WDYT about this? I have noticed you have only included this label to the `alloy_config_hash`, but I think we could gain from having it included in every metric when this config option is set, not only on the k8s world but everywhere.

operations/alloy-mixin/dashboards/prometheus.libsonnet
```diff
@@ -430,63 +456,123 @@ local filename = 'alloy-prometheus-remote-write.json';
...
-        label_values(prometheus_remote_storage_sent_batch_duration_seconds_sum{%(filterSelector)s, job="$job", instance=~"$instance", component_id=~"$component"}, url)
-    ||| % $._config,
+      if $._config.enableAlloyCluster then
+        |||
+            label_values(prometheus_remote_storage_sent_batch_duration_seconds_sum{%(filterSelector)s, cluster_name=~"$alloyCluster", job="$job", instance=~"$instance", component_id=~"$component"}, url)
+        ||| % $._config
+      else
+        |||
+            label_values(prometheus_remote_storage_sent_batch_duration_seconds_sum{%(filterSelector)s, job="$job", instance=~"$instance", component_id=~"$component"}, url)
+        ||| % $._config,
```
wildum, line 575: all this branching is not scaling very well. If we add another variable it will just look awful. Isn't there a way to make this simpler? I'm not familiar with the dashboards config

gaantunes: Will consult @v-zhuravlev on this, since he is a jsonnet sorcerer.

gaantunes:
@wildum any comments regarding the including of `cluster_name` label to all metrics? This PR assumes that is the case.

thampiotr:
Could we instead recommend that if users have [two or more Alloy clusters](https://grafana.com/docs/alloy/next/reference/cli/run/#clustering), then it's important to make them have a different `job` label value, so they can be distinguished easily? 

I think that would be cleaner then having an ever-increasing list of labels to group by.

gaantunes:
Honestly besides the cluster label I do not see we adding any new one.

Issue #1929, opened 2024-10-18
Increasing number of go routines when postgresql cannot be accessed


KaczDev:
### What's wrong?

I have noticed that at one of our production servers there way too many go routines reported in comparison to other instances.
![Image](https://github.com/user-attachments/assets/85bc3dc3-39c3-41d5-8dd5-b5a849311787)

When I checked the logs I found that there are issues with connectin to the PostgreSQL database using `prometheus.exporter.postgres`.

I have reproduced the issue with an Alloy instance that has configured a postgres exporter to a database that it cannot connect to (pg_hba rejected or simply the db doesn't exist).
![Image](https://github.com/user-attachments/assets/f913c570-9c25-4d19-8ca5-952b61dd56fc)

### Steps to reproduce

Create a `prometheus.exporter.postgres` component and configure the URL to a non-existing postgresql URL.

### Software version

Grafana Alloy v.1.4.2

### Configuration
```
logging {
    level  = "info"
    format = "logfmt"
}

prometheus.exporter.postgres "POSTGRESQL_METRICS" {
    data_source_names = ["postgresql://postgres:postgres@postgres:5432/postgres?sslmode=disable"]
    autodiscovery {
        enabled = true
    }
}

prometheus.scrape "SCRAPE_POSTGRESQL_METRICS" {
    targets    = prometheus.exporter.postgres.POSTGRESQL_METRICS.targets
    forward_to = [prometheus.remote_write.PROMETHEUS_REMOTE_WRITE.receiver]
    // increased the number of intervals to reproduce the issue faster.
    scrape_interval = "5s"
    scrape_timeout="4s"
}


prometheus.exporter.self "SELF_REPORT" {}

prometheus.scrape "SELF_SCRAPER" {
    targets    = prometheus.exporter.self.SELF_REPORT.targets
    forward_to = [prometheus.remote_write.PROMETHEUS_REMOTE_WRITE.receiver]
    scrape_interval = "15s"
}

prometheus.remote_write "PROMETHEUS_REMOTE_WRITE" {
    endpoint {
        url = format("http://%s:%s/api/v1/push", env("MIMIR_URL"), env("MIMIR_PORT")) 
    }
}
```

### Logs
```
ts=2024-10-18T15:36:51.273854616Z level=info msg="Established new database connection" component_path=/ component_id=prometheus.exporter.postgres.POSTGRESQL_METRICS fingerprint=postgres:5432
[...]

PR #2019, opened 2024-11-01
fix: Update postgres exporter

dehaansa:
#### PR Description
Update the postgres exporter to pull in a fix for a goroutine leak when the target was unavailable.

https://github.com/grafana/alloy/issues/1929

#### PR Checklist

- [X] CHANGELOG.md updated


go.mod
```diff
@@ -896,7 +896,9 @@ replace (
 	// https://github.com/grafana/cadvisor/tree/grafana-v0.47-noglobals
 	github.com/google/cadvisor => github.com/grafana/cadvisor v0.0.0-20240729082359-1f04a91701e2
 
-	github.com/prometheus-community/postgres_exporter => github.com/grafana/postgres_exporter v0.15.1-0.20240417113938-9358270470dd
+	// TODO(dehaansa): integrate the changes from the exporter-package-v0.15.0 branch into at least the
+	// grafana fork of the exporter, or completely into upstream
+	github.com/prometheus-community/postgres_exporter => github.com/grafana/postgres_exporter v0.15.1-0.20241105053755-e0a51174f168
```
wildum, line 901: your branch was [merged to master](https://github.com/grafana/postgres_exporter/pull/23) so the first part of the todo should be done already, right? 
do you plan on creating the PR upstream? Even if we have our own fork with our own changes, if the bugfix also applies upstream it would be nice to contribute

dehaansa: My branch was merged to master, but I discovered that the version we were using previously is from an un-merged branch from Piotr & another developer at grafana: https://github.com/grafana/postgres_exporter/compare/master...grafana:postgres_exporter:exporter-package-v0.15.0?expand=1 . I merged master into that branch to get my changes into the branch that's compatible with alloy right now, and was planning to jump back to cleaning up their branch and getting it merged later.

PR #2022, opened 2024-11-03
`loki.secretfilter`: Add inline MIT license

romain-gaillard:
#### PR Description
Adding MIT license inline for the gitleaks.toml file.


romain-gaillard:
@thampiotr @wildum The `loki.secretfilter` component should be released for the first time with 1.5.0. This is the last PR to make it fully polished enough for release. Can we make sure it's included in the `1.5.0` release? Thanks!

wildum (approved):

Issue #521, opened 2022-10-27
Flow: Add component for multi-tenant remote_write support


tpaschalis:
The Prometheus remote_write protocol doesn't have the notion of multi-tenancy itself. For this reason different backends offer use various methods to enable multi-tenancy, most often using an `X-Scope-OrgID` header, or labels with a special meaning.

When using a `prometheus.remote_write` component, the Prometheus queue_manager reads WAL segments sequentially and enqueues metrics opportunistically to be batched off as remote_write requests. This behaviour offers no fine-grained control at the request level. Our current suggestion to users is that they add an extra header to their [endpoint block](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/#endpoint-block), but these per-endpoint headers are static, and we don't support write_relabel_config filtering in Flow yet.

The new component would act as a remote_write middleware, receiving metrics from upstream components, extracting a given label (say `tenant`), and batching timeseries grouped by this label value.

It would then send discrete remote_write requests for these batches while also adding the correct `X-Scope-OrgID` header for each one.

### Notes
* This was inspired by the conversation in grafana/agent#1821 and existing solutions like [cortex-tenant](https://github.com/blind-oracle/cortex-tenant).
* I don't believe it would be easy to try and shoehorn this behaviour into the current `prometheus.remote_write` component.
* Using this component would entail some performance penalty which should be measured and explicitly documented.

LilWatson:
+1

Since there seems to be no progress - is there any alternative for dynamic (based on metric label) tenant headers?

wildum:
Moving this proposal to active because of the amount of requests for it. @mattdurham let's discus it when you have time whether it's something we can bring in the remote_write component or the remote_queue one

Nachtfalkeaw:
I would like to see a possibility to add a component in Alloy the tenant information.
I have an alloy instance and several different components active - however I want to send metrics from one component to tenantA and the metrics from other components to tenantB.

Maybe in the component we can add a parameter like "tenant = tenantA" which is then forwarded to "prometheus.remotw_write" component and this component is able to if the component which is sending the metrics to is has the value set and the dynamically replaces it in remote_write for the relevant data,.


If an additional "label in the metrics should be the indicator where to send the metrics to we should make it possible to decide which label it is because a hardcoded "tenant" label may conflict with existing labels in metrics.

ptodev:
[...]

PR #2025, opened 2024-11-04
Add a prometheus label mapping component

vaxvms:
#### PR Description

This PR add a prometheus component to create a label based on a source_label value and a static mapping table.

For a large mapping table, using regex with `prometheus.relabel` is inefficient

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+along to the exported receiver by applying a mapping table to a label value.
+
+The most common use of `prometheus.mapping` is to create a new label based filter Prometheus metrics or
+standardize the label set that is passed to one or more downstream
+receivers. The `rule` blocks are applied to the label set of each metric in
+order of their appearance in the configuration file. The configured rules can
+be retrieved by calling the function in the `rules` export field.
+
+Multiple `prometheus.mapping` components can be specified by giving them
+different labels.
```
clayton-cornell, line 29:
```suggestion
You can specify multiple `prometheus.mapping` components by giving them different labels.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+The `prometheus.mapping` component create a new label on each metric passed
+along to the exported receiver by applying a mapping table to a label value.
+
+The most common use of `prometheus.mapping` is to create a new label based filter Prometheus metrics or
+standardize the label set that is passed to one or more downstream
+receivers. The `rule` blocks are applied to the label set of each metric in
+order of their appearance in the configuration file. The configured rules can
+be retrieved by calling the function in the `rules` export field.
```
clayton-cornell, line 26:
```suggestion
The most common use of `prometheus.mapping` is to create a new label-based filter for Prometheus metrics or standardize the label set passed to one or more downstream receivers.
The `rule` blocks are applied to the label set of each metric in order of their appearance in the configuration file. You can retrieve the configured rules by calling the function in the `rules` export field.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Component health
+
+`prometheus.mapping` is only reported as unhealthy if given an invalid
+configuration. In those cases, exported fields are kept at their last healthy
+values.
+
+## Debug information
+
+`prometheus.mapping` does not expose any component-specific debug information.
```
clayton-cornell, line 74:
```suggestion
`prometheus.mapping` doesn't expose any component-specific debug information.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Debug metrics
+
+* `prometheus_mapping_metrics_processed` (counter): Total number of metrics processed.
+* `prometheus_mapping_metrics_written` (counter): Total number of metrics written.
+
+## Example
+
+Let's create an instance of a see `prometheus.mapping` component and see how
+it acts on the following metrics.
```
clayton-cornell, line 84:
```suggestion
You can create an instance of a `prometheus.mapping` component and see how it acts on the following metrics.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+```
+metric_a{__address__ = "localhost", instance = "development", app = "frontend"} 10
+metric_a{__address__ = "localhost", instance = "development", app = "backend"}  2
+metric_a{__address__ = "cluster_a", instance = "production",  app = "frontend"} 7
+metric_a{__address__ = "cluster_a", instance = "production",  app = "backend"}  9
+metric_a{__address__ = "cluster_b", instance = "production",  app = "database"} 4
+```
+
+After applying the mapping a new label `team` is created based on mapping table
+and `app` label value.
```
clayton-cornell, line 110:
```suggestion
After applying the mapping, a new `team` label is created based on the mapping table and the `app` label value.
```

clayton-cornell (commented):
Some initial doc review comments

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.mapping/
+aliases:
+  - ../prometheus.mapping/ # /docs/alloy/latest/reference/components/prometheus.mapping/
+description: Learn about prometheus.mapping
```
clayton-cornell, line 5:
```suggestion
description: Learn about prometheus.mapping
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,139 @@
...
+title: prometheus.mapping
+---
+
+# prometheus.mapping
+
+Prometheus metrics follow the [OpenMetrics](https://openmetrics.io/) format.
+Each time series is uniquely identified by its metric name, plus optional
+key-value pairs called labels. Each sample represents a datapoint in the
+time series and contains a value and an optional timestamp.
+```
```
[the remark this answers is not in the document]
vaxvms: It come from `prometheus.relabel` documentation.Shoud I amend it as well ?

thampiotr:
Thanks for contributing. 

> For a large mapping table, using regex with prometheus.relabel is inefficient

Could you share the use case and performance differences? In many cases users can use `discovery.relabel` to add the `team=X` labels - and then the overhead is much smaller, because it's not samples, but targets that are relabeled.

If the reason to add this is that prometheus.relabel it's not efficient, we will need some more data on how inefficient it is and whether it could be optimised to work faster. Any new solution we would consider will require benchmarks to prove that it's more efficient. We also generally want to avoid having many ways of doing the same thing - but there can be exceptions.

docs/sources/reference/components/prometheus/prometheus.mapping.md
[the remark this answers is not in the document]
clayton-cornell: Lets fix it here and I'll spin up a PR to align the relabel doc.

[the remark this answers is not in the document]
vaxvms: Before this paragraph, we say
> Prometheus metrics follow the [OpenMetrics](https://openmetrics.io/) text format.

Isn't it precise enough ?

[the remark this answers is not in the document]
clayton-cornell: Oh we've had a miscommunication.  The codefence (the 3 backticks) should have the syntax highlighting language added.  So, for example, if it's just plaintext...

````
```text
<metric name>{<label_1>=<label_val_1>, <label_2>=<label_val_2> ...} <value> [timestamp]
```
````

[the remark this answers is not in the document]
Comment (vaxvms) [reply docs/sources/reference/components/prometheus/prometheus.mapping.md]: Make sense !

Fixed

**Comment by thampiotr** [comment]
Thanks for sharing some performance data. Have you tried to use `discovery.relabel` instead?

The issue with `prometheus.relabel` is that you are relabelling every single sample. With `discovery.relabel` you can relabel the targets before they are scraped.

So, for example, in a cluster of 1k targets and each exposing 1k metrics, you would have 1 million relabels with `prometheus.relabel` as opposed to 1k relabels with `discovery.relabel`.

If we find that for some reason `discovery.relabel` cannot be used, I would like to understand better why. If we need to optimise `prometheus.relabel` I would like to explore that option too before committing to create a new feature to do a very similar job. We may also need to go [through proposal process](https://github.com/grafana/alloy/blob/main/docs/design/908-proposal-process.md) to let team/people comment on this.

**Comment by vaxvms** [comment]
Thanks for your answer.

I'm not using alloy to scrape targets but I'm receiving data thru remote_write from multiple data producers.
The data producers only have the information about the service_id but not the tenant_id, therefore they cannot apply the correct label at the scraping time.

I'm unable to see how we could optimize `prometheus.relabel` to be as performant as a key lookup. We have made some attempt to share cache between instances https://github.com/grafana/alloy/pull/1692 which aims to lower the memory consumption.

Here is a extract of my test configuration, using `prometheus.relabel`:
```river
prometheus.receive_http "ingestion_head" {
    http {
        listen_address = "0.0.0.0"
        listen_port    = 9000
    }
    forward_to = [prometheus.relabel.allow_list.receiver]
}

prometheus.relabel "allow_list" {
    forward_to = [prometheus.remote_write.ingestion_tail.receiver]
    rule {
        action        = "keep"
        regex         = "(.+)"
        source_labels = ["service_id"]
    }


    // One rule block per service_id
    rule {
        source_labels = ["service_id"]
        regex         = "^000001$"
        target_label  = "tenant_id"
        replacement   = "tenant_42"
    }
    
    rule {
        source_labels = ["service_id"]
        regex         = "^000002$"
        target_label  = "tenant_id"
        replacement   = "tenant_190"
    }
    
    // ...
    
    rule {
        source_labels = ["service_id"]
        regex         = "^12345678$"
        target_label  = "tenant_id"
        replacement   = "tenant_8876"
    }
}

prometheus.remote_write "ingestion_tail" {
    endpoint {
        url = "http://mimir/"
    }
}
```

The `prometheus.relabel` can be fairly heavy on the resources especially when we do a 1:1 match, hence the proposal to have a mapping. Here is an example of a working proof of concept based on the code in the PR.
Here is a schema:
    
image

```river
prometheus.receive_http "ingestion_head" {
  http {
    listen_address = "0.0.0.0"
    listen_port    = 9000
  }
  forward_to = [prometheus.relabel.allow_list.receiver]
}

prometheus.relabel "allow_list" {
  forward_to = [prometheus.mapping.tenants_mapping.receiver]

  // First ditch everything that don't have service_id
  rule {
    action        = "keep"
    regex         = "(.+)"
    source_labels = ["service_id"]
  }
}

prometheus.mapping "tenants_mapping" {
  forward_to = [prometheus.remote_write.ingestion_tail.receiver]

  src_label_name = "service_id"
  output_label_name = "tenant_id"

  mapping = {
   "000001" = "tenant_42",
   "000002" = "tenant_190",
   // ....
   "12345678" = "tenant8876",
  }
}
prometheus.remote_write "ingestion_tail" {
        endpoint {
                url = "http://mimir/"
        }
}
```

wilfriedroset:
What we forgot to explicit with @vaxvms is that our initial design would benefit from https://github.com/grafana/alloy/issues/521 we would remove the need for cortex-tenant

vaxvms:
> @vaxvms Did the doc topic at `components/prometheus/prometheus.relabel` accidentally fall off this PR in one of the force/push updates? Or was it removed intentionally? I don't see it in the file list anymore.

Are you talking about `components/prometheus/prometheus.mapping` ? I accidentally removed it. It back again.

I've updated the way the component work: The target_label isn't fixed anymore. each value of the source label can be mapped to severals labels

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+
+  mapping = {
+    "frontend" = {"team" = "teamA"}
+    "backend"  = {"team" = "teamB"}
+    "database" = {"team" = "teamC"}
+  }
+}
+```
+
+```
```
clayton-cornell, line 94:
````suggestion
```text
````

File: docs/sources/reference/components/prometheus/prometheus.mapping.md [hunk 9]
```diff
@@ -0,0 +1,132 @@
...
+metric_a{__address__ = "localhost", instance = "development", app = "backend"}  2
+metric_a{__address__ = "cluster_a", instance = "production",  app = "frontend"} 7
+metric_a{__address__ = "cluster_a", instance = "production",  app = "backend"}  9
+metric_a{__address__ = "cluster_b", instance = "production",  app = "database"} 4
+```
+
+After applying the mapping a new `team` label is created based on mapping table
+and `app` label value.
+
+```
```
clayton-cornell, line 105:
````suggestion
```text
````

File: docs/sources/reference/components/prometheus/prometheus.mapping.md [hunk 10]
```diff
@@ -0,0 +1,132 @@
...
+
+## Debug metrics
+
+* `prometheus_mapping_metrics_processed` (counter): Total number of metrics processed.
+* `prometheus_mapping_metrics_written` (counter): Total number of metrics written.
+
+## Example
+
+You can create and an instance of a see `prometheus.mapping` component and see how
+it acts on the following metrics.
```
clayton-cornell, line 78:
```suggestion
Create an instance of a `prometheus.mapping` component.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+  source_label = "app"
+
+  mapping = {
+    "frontend" = {"team" = "teamA"}
+    "backend"  = {"team" = "teamB"}
+    "database" = {"team" = "teamC"}
+  }
+}
+```
+
```
clayton-cornell, line 93:
```suggestion

Use the following metrics.

```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+```
+metric_a{__address__ = "localhost", instance = "development", app = "frontend"} 10
+metric_a{__address__ = "localhost", instance = "development", app = "backend"}  2
+metric_a{__address__ = "cluster_a", instance = "production",  app = "frontend"} 7
+metric_a{__address__ = "cluster_a", instance = "production",  app = "backend"}  9
+metric_a{__address__ = "cluster_b", instance = "production",  app = "database"} 4
+```
+
+After applying the mapping a new `team` label is created based on mapping table
+and `app` label value.
```
clayton-cornell, line 103:
```suggestion
After applying the mapping, a new `team` label is created based on mapping table and `app` label value.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+```
+metric_a{team = "teamA", __address__ = "localhost", instance = "development", app = "frontend"} 10
+metric_a{team = "teamB", __address__ = "localhost", instance = "development", app = "backend"}  2
+metric_a{team = "teamA",  __address__ = "cluster_a", instance = "production",  app = "frontend"} 7
+metric_a{team = "teamA",  __address__ = "cluster_a", instance = "production",  app = "backend"}  9
+metric_a{team = "teamC",  __address__ = "cluster_a", instance = "production",  app = "database"} 4
+```
+
+The resulting metrics are then propagated to each receiver defined in the
+`forward_to` argument.
```
clayton-cornell, line 114:
```suggestion
The resulting metrics are propagated to each receiver defined in the `forward_to` argument.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+
+Name       | Type              | Description
+-----------|-------------------|-----------------------------------------------------------
+`receiver` | `MetricsReceiver` | The input receiver where samples are sent to be relabeled.
+
+## Component health
+
+`prometheus.mapping` is only reported as unhealthy if given an invalid
+configuration. In those cases, exported fields are kept at their last healthy
+values.
```
clayton-cornell, line 64:
```suggestion
`prometheus.mapping` is only reported as unhealthy if given an invalid configuration.
In those cases, exported fields are kept at their last healthy values.
```

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name           | Type                      | Description                                                             | Default | Required
+---------------|---------------------------|-------------------------------------------------------------------------|---------|---------
+`forward_to`   | `list(MetricsReceiver)`   | Where the metrics should be forwarded to, after relabeling takes place. |         | yes
```
clayton-cornell, line 48:
```suggestion
`forward_to`   | `list(MetricsReceiver)`   | The receiver the metrics are forwarded to after they are relabeled. |         | yes
```

vaxvms:
Thanks for review @clayton-cornell 
Most of the phrasing came from `prometheus.relabel.md`

clayton-cornell:
> Most of the phrasing came from `prometheus.relabel.md`

Yeah that topic needs some attention too. I've got an open task to review and refactor the component docs.

internal/component/prometheus/mapping/mapping.go
```diff
@@ -0,0 +1,224 @@
...
+// component.
+type Arguments struct {
+	// Where the relabelled metrics should be forwarded to.
+	ForwardTo []storage.Appendable `alloy:"forward_to,attr"`
+
+	// Labels to use for mapping
+	SourceLabel string `alloy:"source_label,attr"`
+
+	// Mapping
+	LabelValuesMapping map[string]map[string]string `alloy:"mapping,attr"`
```
dehaansa, line 47: Is a static map the best user experience for this? Should it support reading from a file - especially using the exports of a `local.file` or `remote.s3` for example? This allows for polled updating of the static map without needing to reload/restart/directly edit the Alloy config.

internal/component/prometheus/mapping/mapping.go
```diff
@@ -0,0 +1,224 @@
...
+	c.sourceLabel = newArgs.SourceLabel
+	c.mappings = newArgs.LabelValuesMapping
+	c.fanout.UpdateChildren(newArgs.ForwardTo)
+
+	c.opts.OnStateChange(Exports{Receiver: c.receiver})
+
+	return nil
+}
+
+func (c *Component) mapping(val float64, lbls labels.Labels) labels.Labels {
```
dehaansa, line 206: If we're not looking to see if the val is a staleNaN for caching, we probably don't need it to still be in the function call.

Is there a reason this isn't using the global labelstore like prometheus.relabel? Did you test performance with/without it?

internal/component/prometheus/mapping/mapping.go
```diff
@@ -0,0 +1,224 @@
...
+	// Relabel against a copy of the labels to prevent modifying the original
+	// slice.
+	lb := labels.NewBuilder(lbls.Copy())
+	sourceValue := lb.Get(c.sourceLabel)
```
dehaansa, line 210: This logic will attempt to map when the sourceLabel is not present. Rather than the mapping behave the same when sourceLabel is not present && sourceLabel has explicit value "", should there be an (optional) explicit defaultMapping for when sourceValue is not present?

internal/component/prometheus/mapping/mapping.go
vaxvms: I can dynamically read the mapping using alloy components. Using this method, it avoid having parsing logic in mapping component :

```
...
local.file "mapping" {
   filename = file.path_join(module_path, "mapping.bifrost")
}

prometheus.mapping "tenants_mapping" {
  forward_to   = [prometheus.remote_write.cluster.receiver]
  source_label = "uid_label_key1"
  mapping      = encoding.from_json(local.file.mapping.content)
}
...
```

Did i miss something ?

internal/component/prometheus/mapping/mapping.go
vaxvms: As i have used prometheus.relabel as a starting point, i forgot to remove val parameter. Fixed.

The performance improvement was so high, i didn't bothered looking for increased performance. Relabeling can be a heaving process involving regex whereas mapping is just a hash lookup.

Using the labelstore, it will have to
- compute a hash based on labels
- fetch the globalId in a map
- try to fetch new labels set from cache

I might be missing something but i can't see why 2 map lookup would be faster than a single lookup

internal/component/prometheus/mapping/mapping.go
vaxvms: I don't have any strong opinion about this one. Having all mapping values at the same place make more sense to mebut i don't really bother.

vaxvms:
@dehaansa @clayton-cornell @thampiotr Is here anything I have to fix/document/explain/change ?

vaxvms:
So I have to wish them to have a nice holiday !

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name           | Type                      | Description                                                         | Default | Required
+---------------|---------------------------|---------------------------------------------------------------------|---------|---------
+`forward_to`   | `list(MetricsReceiver)`   | The receiver the metrics are forwarded to, afterthey are relabeled. |         | yes
```
clayton-cornell, line 48:
```suggestion
`forward_to`   | `list(MetricsReceiver)`   | The receiver the metrics are forwarded to after they are relabeled. |         | yes
```

clayton-cornell (dismissed):

clayton-cornell:
@dehaansa will this be released as a GA component? or as a Preview/Experimental? The docs don't mention the `--stability.level`runtime flag... I'm wondering if it needs to be added in here.

dehaansa:
> @dehaansa will this be released as a GA component? or as a Preview/Experimental? The docs don't mention the `--stability.level`runtime flag... I'm wondering if it needs to be added in here.

Probably a good idea, should start as Public Preview.

docs/sources/reference/components/prometheus/prometheus.mapping.md
```diff
@@ -0,0 +1,131 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.mapping/
+description: Learn about prometheus.mapping
+title: prometheus.mapping
+---
+
+# prometheus.mapping
```
clayton-cornell, line 7:
```suggestion
<span class="badge docs-labels__stage docs-labels__item">Public preview</span>

# prometheus.mapping

{{< docs/shared lookup="stability/public_preview.md" source="alloy" version="<ALLOY_VERSION>" >}}
```

Adding in the Public Preview elements for the docs

internal/component/prometheus/mapping/mapping.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/prometheus/prometheus/storage"
+	"go.uber.org/atomic"
+)
+
+const name = "prometheus.mapping"
+
+func init() {
+	component.Register(component.Registration{
+		Name:      name,
+		Stability: featuregate.StabilityGenerallyAvailable,
```
dehaansa, line 27: We should mark this as PublicPreview for first release.
```suggestion
		Stability: featuregate.StabilityPublicPreview,
```

dehaansa:
Apologies, this fell out of memory over the holidays. Adding to a list to discuss next Monday to see about moving it forward.

docs/design/2025-prometheus-mapping-component.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Problem
+
+Using `prometheus.relabel` to populate a label value based on another label value is inefficient as we have to have a rule block for each source label value.
+
+If we have 1k values to map, we'll have to execute 1k regex for each datapoint resulting in an algorithm complexity of O(n).
+
+## Proposal
+
+Replace regex computing by a lookup table. Algorithm complexity goes from O(n) to O(1)
```
thampiotr, line 19: I guess `loki.mapping` or `discovery.mapping` could be useful too. Happy to leave it for future PRs though.

docs/design/2025-prometheus-mapping-component.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Alternative solutions
+
+- Instanciate more CPU resources to perform the task
+- Optimize prometheus.relabel component
+- Summarize regex when severals keys have to same value.
+
+## Compatibility
+
+As this is a new component, there isn't any compatibility issue as long as you don't use it.
```
thampiotr, line 37: Should this be a community component?

thampiotr (commented):
Looks good @vaxvms, thanks for providing extra context. I'll leave up to @dehaansa to complete the review.

dehaansa:
Since we last discussed this component, we've [merged](https://github.com/grafana/alloy/pull/2882/files) the loki.enrich component ([docs](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.enrich/)). 

I wonder if we want to standardize this to be the equivalent component `prometheus.enrich` and operate on similar target-style json. While the loki.enrich docs show it being used with target objects from a discovery component, it can be used with a plain json mapping table, but the format would be different. Rather than having the mapping label point to a map of labels to apply, the mapping label would be present in the json map and it would apply the labels iterated in `labels_to_copy`. 

Internally the component will be as efficient as this implementation, as if we follow the pattern established in `loki.enrich` it pre-processes the targets on update to make it a O(1) lookup map rather than iterating over the list.

The one use case that it might not support that would be supported by this component currently would include replacing the target that is used for mapping. IE this component currently supports mapping `{"labelA": "oldValue"} => {"labelA": "newValue"}` but the current implementation for `loki.enrich` wouldn't support that. Is that an important use case for you?

PR #2012, opened 2024-10-31
Update otel v0.112.0

wildum:
@polyrain please update the datadog exporter if you have time. I only made some changes to make it compile but I did not check the changes for it (you can branch out from my branch and create a PR to it)

Blocker: we need a Beyla version with otel collector dependencies updated to v0.112.0

I encourage reviewers to review commit by commit

PR #2027, opened 2024-11-04
Fix race condition in otelcol component wrappers

thampiotr:
#### PR Description

We have a general issue with OTel components where consumers may be used before the Start functions in OTel have finished running. This is because in OTel Start functions are non-blocking and sometimes do work to set things up, like it was the case for [batch_processor](https://github.com/open-telemetry/opentelemetry-collector/pull/11594). In Alloy, however we have Run function that is blocking for the lifetime of the component. As soon as it's called, we consider the component Running. In OTel, however, the Start function should be called and exit to consider a component running.

The solution here is to pause the Consumer until we are sure that the OTel component scheduler has called `Start` on all OTel components. The consumer will block any attempts to feed data to it.

#### Notes to the Reviewer

```
thampiotr@MacWork ~/w/alloy (update-otel-112-fix-race)> go test ./internal/component/otelcol/internal/scheduler/ -count 50 -race
ok  	github.com/grafana/alloy/internal/component/otelcol/internal/scheduler	3.543s
thampiotr@MacWork ~/w/alloy (update-otel-112-fix-race)> go test ./internal/component/otelcol/internal/lazyconsumer/ -count 50 -race
ok  	github.com/grafana/alloy/internal/component/otelcol/internal/lazyconsumer	17.495s
```

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
+++ b/internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
@@ -17,9 +17,9 @@ import (
 type Consumer struct {
 	ctx context.Context
 
-	// pauseMut is used to implement Pause & Resume semantics. When a write lock is held - this consumer is paused.
-	// See Pause method for more info.
+	// pauseMut is used to implement Pause & Resume semantics. See Pause method for more info.
 	pauseMut sync.RWMutex
+	pausedWg *sync.WaitGroup
 
 	mut             sync.RWMutex
 	metricsConsumer otelconsumer.Metrics
@@ -65,6 +65,9 @@ func (c *Consumer) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
 
 	c.pauseMut.RLock() // wait until resumed
 	defer c.pauseMut.RUnlock()
+	if c.pausedWg != nil {
+		c.pausedWg.Wait()
+	}
 
 	c.mut.RLock()
 	defer c.mut.RUnlock()
@@ -89,6 +92,9 @@ func (c *Consumer) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) error
 
 	c.pauseMut.RLock() // wait until resumed
 	defer c.pauseMut.RUnlock()
+	if c.pausedWg != nil {
+		c.pausedWg.Wait()
+	}
 
 	c.mut.RLock()
 	defer c.mut.RUnlock()
@@ -113,6 +119,9 @@ func (c *Consumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
 
 	c.pauseMut.RLock() // wait until resumed
 	defer c.pauseMut.RUnlock()
+	if c.pausedWg != nil {
+		c.pausedWg.Wait()
+	}
 
 	c.mut.RLock()
 	defer c.mut.RUnlock()
@@ -141,13 +150,35 @@ func (c *Consumer) SetConsumers(t otelconsumer.Traces, m otelconsumer.Metrics, l
 }
 
 // Pause will stop the consumer until Resume is called. While paused, the calls to Consume* methods will block.
-// After calling Pause once, it must not be called again until Resume is called.
+// Pause can be called multiple times, but a single call to Resume will un-pause this consumer. Thread-safe.
 func (c *Consumer) Pause() {
 	c.pauseMut.Lock()
+	defer c.pauseMut.Unlock()
+
+	if c.pausedWg != nil {
+		return // already paused
+	}
+
+	c.pausedWg = &sync.WaitGroup{}
+	c.pausedWg.Add(1)
 }
 
-// Resume will revert the Pause call and the consumer will continue to work. Resume must not be called if Pause wasn't
-// called before. See Pause for more details.
+// Resume will revert the Pause call and the consumer will continue to work. See Pause for more details.
 func (c *Consumer) Resume() {
-	c.pauseMut.Unlock()
+	c.pauseMut.Lock()
+	defer c.pauseMut.Unlock()
+
+	if c.pausedWg == nil {
+		return // not paused
+	}
+
+	c.pausedWg.Done()
+	c.pausedWg = nil
+}
+
+// IsPaused returns whether the consumer is currently paused. See Pause for details.
+func (c *Consumer) IsPaused() bool {
+	c.pauseMut.RLock()
+	defer c.pauseMut.RUnlock()
+	return c.pausedWg != nil
 }
--- a/internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
+++ b/internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
@@ -17,7 +17,7 @@ import (
 type Consumer struct {
 	ctx context.Context
 
-	// pauseMut is used to implement Pause & Resume semantics. See Pause method for more info.
+	// pauseMut and pausedWg are used to implement Pause & Resume semantics. See Pause method for more info.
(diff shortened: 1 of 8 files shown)


internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
```diff
@@ -119,3 +139,15 @@ func (c *Consumer) SetConsumers(t otelconsumer.Traces, m otelconsumer.Metrics, l
...
+// After calling Pause once, it must not be called again until Resume is called.
+func (c *Consumer) Pause() {
+	c.pauseMut.Lock()
+}
+
+// Resume will revert the Pause call and the consumer will continue to work. Resume must not be called if Pause wasn't
+// called before. See Pause for more details.
+func (c *Consumer) Resume() {
+	c.pauseMut.Unlock()
+}
```
wildum, line 153: If pause is called by accident on a paused consumer or resume on a running consumer, it will be dramatic. Should the consumer holds its state (smth like "isPaused") to avoid locking if its already paused or avoid resuming if it's already running for extra safety?

internal/component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -90,6 +115,10 @@ func (cs *Scheduler) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case <-cs.newComponentsCh:
+			if !firstRun { // do not pause on first run
+				cs.onPause()
+				firstRun = false
```
wildum, line 120: firstRun cannot be set to false inside of the !firstRun statement. I think you should set is right after it, right?

internal/component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -100,6 +129,7 @@ func (cs *Scheduler) Run(ctx context.Context) error {
 
 			level.Debug(cs.log).Log("msg", "scheduling components", "count", len(components))
 			components = cs.startComponents(ctx, host, components...)
+			cs.onResume()
```
wildum, line 132: if the component was created with New() instead of NewPaused() then this will panic because on the first run you will try to unlock the mut that's not locked

wildum (commented):
Thanks for working on this, I think that's a step in the right direction.
The logic was not super clear to me at first in the scheduler, the idea is that the components are created in the paused state so they are not paused on the first run but are always resumed after being started because they are paused when the scheduler stops running, right? Maybe we could have this explanation in the scheduler code

docs/sources/reference/components/otelcol/otelcol.exporter.datadog.md
```diff
@@ -118,6 +120,27 @@ If `endpoint` is unset, the value is obtained through the `site` parameter in th
...
+`batch_wait`                     | `int`         | The maximum time in seconds the logs agent waits to fill each batch of logs before sending.                                                                                                               | `5`                                | no
+
+
+If `use_compression` is disabled, `compression_level` has no effect.
+
+If `endpoint` is unset, the value is obtained through the `site` parameter in the [api][] section.
+
+[api]: #api-block
+
+
```
clayton-cornell, line 143:
```suggestion

If `use_compression` is disabled, `compression_level` has no effect.

If `endpoint` is unset, the value is obtained through the `site` parameter in the [api][] section.

[api]: #api-block

```

Just removing extra cr/lf

docs/sources/reference/components/otelcol/otelcol.connector.spanmetrics.md
```diff
@@ -16,24 +16,27 @@ aggregates Request, Error and Duration (R.E.D) OpenTelemetry metrics from the sp
...
-    duration { service.name="shipping", span.name="get_shipping/{shippingId}", span.kind="SERVER", status.code="Ok" }
-    ```
+  Span durations are tracked using a `duration` histogram metric:
+
+  ```
+  duration { service.name="shipping", span.name="get_shipping/{shippingId}", span.kind="SERVER", status.code="Ok" }
+  ```
 
 > **NOTE**: `otelcol.connector.spanmetrics` is a wrapper over the upstream
 > OpenTelemetry Collector `spanmetrics` connector. Bug reports or feature requests
```
clayton-cornell, line 42: Same comment as in https://github.com/grafana/alloy/pull/2012#discussion_r1826095047 to wrap the note in a Note adminiton

docs/sources/reference/components/otelcol/otelcol.processor.interval.md
```diff
@@ -10,19 +10,23 @@ title: otelcol.processor.interval
...
 
-* All delta metrics
-* Non-monotonically increasing sums
-* Gauges
-* Summaries
+- All delta metrics
+- Non-monotonically increasing sums
+
+{{< admonition type="note" >}}
+Aggregating data over an interval is an inherently "lossy" process. For monotonically increasing, cumulative sums, histograms, and exponential histograms, you "lose" precision, but you don't lose overall data. But for non-monotonically increasing sums, gauges, and summaries, aggregation represents actual data loss. IE you could "lose" that a value increased and then decreased back to the original value. In most cases, this data "loss" is ok. However, if you would rather these values be passed through, and _not_ aggregated, you can set that in the configuration.
```
clayton-cornell, line 28:
```suggestion
Aggregating data over an interval is an inherently lossy process.
You lose precision for monotonically increasing cumulative sums, histograms, and exponential histograms, but you don't lose overall data.
You can lose data when you aggregate non-monotonically increasing sums, gauges, and summaries.
For example, a value can increase and decrease to the original value, and you can lose this change in the aggregation.
In most cases, this type of data loss is OK.
However, you can change the configuration so that these changed values pass through and are _not_ aggregated.
```

Same as: https://github.com/grafana/alloy/pull/2012#discussion_r1826117206

clayton-cornell (commented):
Just some edits to align with the same changes in https://github.com/grafana/alloy/pull/2012  Might also be "fixed" up when the PR conflicts are resolved?

internal/component/otelcol/internal/lazyconsumer/lazyconsumer.go
thampiotr: Yeah, I was thinking about this too, I wouldn't want to have any dramatic panics either. I've changed the implementation to handle multiple calls to Resume and Pause - it's a bit more complex, but I also wrote a lot of tests.

internal/component/otelcol/internal/scheduler/scheduler.go
thampiotr: Now it will no longer panic as calls to Resume can be repeated.

When writing code we still need to take care to use the correct paused or resumed consumer, but the consequences of a mistake shouldn't be catastrophic.

internal/component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -75,11 +96,16 @@ func (cs *Scheduler) Schedule(h otelcomponent.Host, cc ...otelcomponent.Componen
...
+	firstRun := true
 	var components []otelcomponent.Component
 
 	// Make sure we terminate all of our running components on shutdown.
 	defer func() {
+		if !firstRun { // always handle the callbacks correctly
+			cs.onPause()
+		}
 		cs.stopComponents(context.Background(), components...)
+		cs.onResume()
```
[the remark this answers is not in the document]
thampiotr: We don't need to, but I think it's better choice to call `onResume` here so we can unblock any goroutines that may be waiting on this.

wildum (approved):

PR #2043, opened 2024-11-05
prometheus.exporter.cloudwatch: add `add_cloudwatch_timestamp` field to metrics

captncraig:
This will allow you to force metrics to use the cloudwatch timestamps instead of the scrape time.

Probably must also use `honor_timestamps` in your scrape.


docs/sources/reference/components/prometheus/prometheus.exporter.cloudwatch.md
```diff
@@ -316,6 +316,7 @@ on how to explore metrics, to easily pick the ones you need.
 | `period`      | `duration`     | Refer to the [period][] section below.                                    |                                                                                                                    | yes      |
 | `length`      | `duration`     | Refer to the [period][] section below.                                    | Calculated based on `period`. Refer to [period][] for details.                                                     | no       |
 | `nil_to_zero` | `bool`         | When `true`, `NaN` metric values are converted to 0.                      | The value of `nil_to_zero` in the parent [static][] or [discovery][] block. `true` if not set in the parent block. | no       |
+| `add_cloudwatch_timestamp` | `bool` | When `true`, use the timestamp from cloudwatch instead of the scrape time. | `false` | no
```
erikbaranowski, line 319: I don't love the add term in this case since it's a replace but it is consistent with upstream.

clayton-cornell, line 319:
```suggestion
| `add_cloudwatch_timestamp` | `bool` | When `true`, use the timestamp from CloudWatch instead of the scrape time. | `false` | no
```

captncraig: agree, but it matches upstream so 🤷

spartan0x117 (approved):
Code portions LGTM, do you know if there's a difference in behavior in upstream if we pass `nil` instead of `false` since previously we were always explicitly passing `false` (but now if the new argument is not supplied it will pass `nil` instead)?

clayton-cornell (approved):

PR #2044, opened 2024-11-05
fix(helm): ensure helm resources have correct namespace

shinebayar-g:
#### PR Description

Helm chart resources don't have correct namespaces.

Issue can be reproduced by running `helm template --namespace alloy alloy grafana/alloy --version 0.9.2`.
Generated yaml files didn't have correct namespace.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


petewall:
I get nervous about the removal of the use of the `alloy.namespace` template function. That might break some users.
Perhaps instead of using `.Release.namespace`, we use `alloy.namespace` which is enabled by the namespaceOverride value

shinebayar-g:
> I get nervous about the removal of the use of the `alloy.namespace` template function. That might break some users. Perhaps instead of using `.Release.namespace`, we use `alloy.namespace` which is enabled by the namespaceOverride value

Didn't seem that was the case considering that it was undocumented value and some resources were already using the `{{ .Release.Namespace }}` like in the `operations/helm/charts/alloy/templates/pdb.yaml` file.

But updated the PR regardless.

shinebayar-g:
I resolve the conflict, please review. For now I'm using my local fork as a workaround.

shinebayar-g:
Thanks @clayton-cornell  , I've updated the PR as requested and rebased.

clayton-cornell:
@petewall @wildum Is this PR ready to approve/merge?

Lobarr:
hi, friendly ping on this PR. the pyroscope helm chart is broken and this could fix it.

some context: 

the pyroscope helm chart properly handles namespaces by utilizing the `.Release.Namespace` provided by helm. it depends on alloy's helm chart but alloy doesn't handle namespaces at all. well, except in one place - https://github.com/grafana/alloy/blob/020676122908ca455b699839945a78f39a1d4d8a/operations/helm/charts/alloy/templates/serviceaccount.yaml#L6.

the effect of this is that the service account ends up being inaccessible to alloy as alloy expects it to be in the default namespace but that's not the case. instead the following error is thrown:

```
Warning  FailedCreate  68s (x19 over 22m)  statefulset-controller  create Pod pyroscope-alloy-0 in StatefulSet pyroscope-alloy  failed error: pods "pyroscope-alloy-0" is forbidden: error looking up service account default/pyroscope-alloy: serviceaccount "pyroscope-alloy" not found
```

petewall (approved):

wildum:
hey, sorry this took so long, @shinebayar-g would you mind running `make generate-helm-tests` again please? I would then proceed to merge the branch.

shinebayar-g:
Thanks for merging the conflict. I committed the `make generate-helm-tests` changes.

PR #2049, opened 2024-11-06
Update updating OpenTelemetry Collector dependencies doc

wildum:


docs/developer/updating-otel/README.md
```diff
@@ -95,6 +95,18 @@ Unfortunately, updating Otel dependencies is not straightforward:
 You can refer to [PR grafana/agent#5290](https://github.com/grafana/agent/pull/5290)
 for an example on how to update Alloy.
 
+### Notify community component contributors
+
+You can find their GitHub handles at the top of the community component files (the components that have "Community" set to true).
+Once the PR is created, you can ping them on it and message them on Slack. Each contributor should create a PR to your branch with the updated component.
+They should be notified a few days before the release so that they have enough time to do the update.
```
clayton-cornell, line 102:
```suggestion
Notify the contributors a few days before the release so they have enough time to do the update.
```

docs/developer/updating-otel/README.md
```diff
@@ -95,6 +95,18 @@ Unfortunately, updating Otel dependencies is not straightforward:
...
+It should not be a blocker for the release. If they can't do the update, you can fix the code to resolve the breaking changes or disable the component if that's too complicated.
```
clayton-cornell, line 103:
```suggestion
It should not be a blocker for the release. If the contributors can't do the update, you can fix the code to resolve the breaking changes or disable the component if that's too complicated.
```

dehaansa:
Should we define a specific window of time that we're going to allow for community maintainers before we can opt to disable/remove the component from that release if it's too much effort?

wildum:
> Should we define a specific window of time that we're going to allow for community maintainers before we can opt to disable/remove the component from that release if it's too much effort?

disabeling/removing the component is really a last resort option that I hope we don't ever have to take. If community maintainers are not available to solve some breaking changes, we should be able to quickly solve them (doing the minimum to get it to build and run should be easy in most cases, in the end these are just wrappers around the otel components). I'd prefer to not specify a time window for now and do case-by-case

docs/developer/updating-otel/README.md
```diff
@@ -95,6 +95,18 @@ Unfortunately, updating Otel dependencies is not straightforward:
 You can refer to [PR grafana/agent#5290](https://github.com/grafana/agent/pull/5290)
 for an example on how to update Alloy.
 
+### Notify community component contributors
+
+You can find the community contributor GitHub handles at the top of the community component files (the components that have "Community" set to true).
+Once the PR is created, you can ping them in the PR and message them on Slack. Each contributor should create a PR in your branch with the updated component.
+Notify the contributors a few days before the release so they have enough time to do the update.
```
thampiotr, line 102: Most of the time when you read these docs you will be doing the release, so I'm not sure if it's the right place for a reminder like this.

wildum: this is the "updating-otel" doc. The update should be done prior to the release and should be started at least one week before cutting the RC

Issue #1827, opened 2024-10-04
Component for Grafana Cloud DB O11y


matthewnolf:
### Background

At Grafana Labs, we are in the beginning stages of developing a new feature set for Grafana Cloud - Database Observability. This feature set provides end users with an experience for monitoring their database setups, to perform tasks such as instance monitoring, query analysis, and performance insight and optimization. 

Such functionality obviously relies on having data available from the source, in order to interpret, and visualize. This data is available both directly and indirectly (i.e. requires additional transformation) from the database instance being monitored. 
We therefore must ask end users to run an application that is capable of this data collection, and a means to ship it to Grafana Cloud - something they are already using Alloy to do for a variety of similar use cases.
This feature set will initially be compatible with the MySQL database, with plans to extend it to other SQL and NoSQL alternatives in the future. The contents of this proposal however are centered around MySQL support.

### Proposal

To add a new component to alloy that captures and outputs all of the required telemetry for the Grafana Cloud - Database Observability feature set - which will include both metrics, and logs. The objective of this component will be to complement the Grafana Cloud feature. Maintaining compatibility with other similar features/products is not a primary objective.

The proposal is to implement the collection functionality in an external, separate (to the alloy repository) package. This package will then be imported and run as a component within Alloy, behaving in a similar way to other prometheus exporter components. This will enable us to maintain and iterate on the package outside of alloy. While we have begun work on this already, we are submitting this proposal early so that any adjustments or considerations can be incorporated to the approach we take.
There exists today a prometheus exporter for MySQL, which is available as an alloy component. This proposal plans to make use of that component, however will encompass it with additional transformations along with outputting other telemetry signals - namely logs.

For illustrative purposes, the component signature may look like the following. We would appreciate any naming recommendations.

```
db-o11y.collector "collector-name" {
  data_source_name = DATA_SOURCE_NAME
  metrics_remote_write = [prometheus.remote_write.grafana_cloud.receiver]
  logs_remote_write = [loki.write.grafana_cloud.receiver]
  
  # additional configuration to be added later
}
```

ptodev:
Hello! Thank you for the proposal. Do you think you'd be able to reuse the existing components, and wrap them in a module? Then you could host the modules in the [alloy-modules repo](https://github.com/grafana/alloy-modules).

matthewnolf:
[...]

PR #2053, opened 2024-11-07
Add `database_observability.mysql` component

cristiangreco:
#### PR Description
Add new component `database_observability.mysql`.

Relates to https://github.com/grafana/alloy/issues/1827

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -0,0 +1,193 @@
...
+		var digest, query_sample_text, query_sample_seen, query_sample_timer_wait string
+		err := rs.Scan(&digest, &query_sample_text, &query_sample_seen, &query_sample_timer_wait)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to scan query samples", "err", err)
+			break
+		}
+
+		redacted, err := sqlparser.RedactSQLQuery(query_sample_text)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to redact sql", "err", err)
```
matthewnolf, line 115:
```suggestion
			level.Error(c.logger).Log("msg", "failed to redact sql query", "err", err)
```

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (c QuerySample) tablesFromQuery(query string) []string {
+	if strings.HasSuffix(query, "...") {
+		level.Info(c.logger).Log("msg", "skipping parsing truncated query")
+		return []string{}
+	}
+
+	stmt, err := sqlparser.Parse(query)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to parse sql", "err", err)
```
matthewnolf, line 149:
```suggestion
		level.Error(c.logger).Log("msg", "failed to parse sql query", "err", err)
```

matthewnolf (approved):
LGTM on the SQL collector part!

docs/sources/reference/components/grafanacloud/grafanacloud.database_observability.mysql.md
```diff
@@ -0,0 +1,23 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/grafanacloud.database_observability.mysql/
+description: Learn about grafanacloud.database_observability.mysql
+title: grafanacloud.database_observability.mysql
+---
+<!-- START GENERATED COMPATIBLE COMPONENTS -->
```
clayton-cornell, line 6: This only contains generated content?

Also since this is a new namespace and a new doc section, we need to add an `_index.md` in the root of `/reference/components/grafanacloud/`

```
---
canonical: https://grafana.com/docs/alloy/latest/reference/components/grafanacloud/
description: Learn about the Grafana Cloud components in Grafana Alloy
title: grafanacloud
weight: 100
---

# grafanacloud

This section contains reference documentation for the `grafanacloud` components.

{{< section >}}
````

Comment (cristiangreco) [reply docs/sources/reference/components/grafanacloud/grafanacloud.database_observability.mysql.md:9]: Hey @clayton-cornell can you recommend how to fill the docs with some basic content? This is an experimental component that we'll continue to iterate on, so we expect to followup with more detailed docs in the next PRs but right now things are still quite unstable.

File: internal/component/database_observability/mysql/component.go [hunk 4]
```diff
@@ -0,0 +1,260 @@
...
+		Targets: []discovery.Target{c.baseTarget},
+	})
+
+	for _, collector := range c.collectors {
+		collector.Stop()
+	}
+	c.collectors = nil
+
+	newArgs := args.(Arguments)
+	c.args = newArgs
```
wildum, line 172: I think that you can just do c.args := args.(Arguments). newArgs could be useful if you would want the component to keep running with the previous config in case the update goes wrong. In this case you would set the c.args = newArgs at the end of the Update when you are sure that it went through. Since that you are stopping the collectors anyway I don't think that you need this

internal/component/database_observability/mysql/component.go
```diff
@@ -0,0 +1,260 @@
...
+func (a *Arguments) Validate() error {
+	_, err := mysql.ParseDSN(string(a.DataSourceName))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type Exports struct {
+	Targets []discovery.Target `alloy:"targets,attr"`
```
wildum, line 72: I don't think that this component should expose targets to be scrapped. If you want this component to expose its own metrics you can use the [Registerer](https://github.com/grafana/alloy/blob/main/internal/component/registry.go#L94) that's part of the component options that you get in the New() function

internal/component/database_observability/mysql/collector/query_sample_test.go
```diff
@@ -0,0 +1,66 @@
...
+	lokiEntries := lokiClient.Received()
+	for _, entry := range lokiEntries {
+		require.Equal(t, model.LabelSet{"job": "integrations/db-o11y"}, entry.Labels)
+	}
+	require.Equal(t, `level=info msg="query samples fetched" op="query_sample" digest="abc123" query_sample_text="select * from some_table where id = 1" query_sample_seen="2024-01-01T00:00:00.000Z" query_sample_timer_wait="1000" query_redacted="select * from some_table where id = :redacted1"`, lokiEntries[0].Line)
+	require.Equal(t, `level=info msg="table name parsed" op="query_parsed_table_name" digest="abc123" table="some_table"`, lokiEntries[1].Line)
+
+	err = mock.ExpectationsWereMet()
+	require.NoError(t, err)
+}
```
wildum, line 66: for components that spawn goroutines, I think that it's nice to make sure that they are properly stopped and that no goroutines are leaking with goleak.VerifyNone

wildum (commented):
Nice work, looking forward to see the cool features that this component can enable!

internal/component/database_observability/mysql/component.go
cristiangreco: My understanding is that `opts.Registerer` is for metrics regarding the component itself. If I switch to using `opts.Registerer` then the `connection_info` metric is available at `http://localhost:12345/metrics`.

Our intention is to expose metrics that can be piped into a remote write, e.g.

```
grafanacloud.database_observability.mysql "orders_db" {
  data_source_name = "mysql:3306"
  forward_to = [loki.write.logs_service.receiver] // not used in this example
}

prometheus.scrape "orders_db" {
  targets = grafanacloud.database_observability.mysql.orders_db.targets
  honor_labels = true
}

prometheus.remote_write "metrics_service" {
  endpoint {
    url = sys.env("GCLOUD_HOSTED_METRICS_URL")
    basic_auth {
      username = sys.env("GCLOUD_HOSTED_METRICS_ID")
      password = sys.env("GCLOUD_RW_API_KEY")
    }
  }
}
```

Is there a different way of doing this?

wildum: that's correct, but given that you are planning to use the existing exporter for the metrics + that you only have the "connection_info" as a metric here, I thought that it would fit as a metric regarding the component itself. If you want to expose more metrics from this component in the future and you think that it doesn't fit as metrics about the component itself then you can expose them the way you did

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	c.ctx = ctx
+	c.cancel = cancel
+
+	go func() {
+		ticker := time.NewTicker(c.scrapeInterval)
+
+		for {
+			if err := c.fetchQuerySamples(c.ctx); err != nil {
+				break
+			}
```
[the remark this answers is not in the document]
cristiangreco: Added a call to `Stop()` before the final return.

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -0,0 +1,193 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	c.ctx = ctx
+	c.cancel = cancel
+
+	go func() {
+		ticker := time.NewTicker(c.scrapeInterval)
+
+		for {
+			if err := c.fetchQuerySamples(c.ctx); err != nil {
+				break
```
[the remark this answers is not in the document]
cristiangreco: Added more logging before exiting collector loop.

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -0,0 +1,224 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	c.ctx = ctx
+	c.cancel = cancel
+
+	go func() {
+		ticker := time.NewTicker(c.scrapeInterval)
+
+		for {
+			if err := c.extractSchema(c.ctx); err != nil {
+				break
```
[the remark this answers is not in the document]
cristiangreco: Addressed

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -0,0 +1,224 @@
...
+				Labels: model.LabelSet{"job": "integrations/db-o11y"},
+				Entry: logproto.Entry{
+					Timestamp: time.Unix(0, time.Now().UnixNano()),
+					Line:      fmt.Sprintf(`level=info msg="table detected" op="%s" schema="%s" table="%s"`, OP_TABLE_DETECTION, schema, table),
+				},
+			}
+		}
+	}
+
+	for _, table := range tables {
```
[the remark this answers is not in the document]
cristiangreco: IIRC at poc stage we had an issue detecting tables with the same name across different schemas, I can't remember exactly. If you don't mind I'd leave this logic as-is for now, we'll anyway followup with further code reorganisation and unit testing for the collectors.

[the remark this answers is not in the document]
wildum: That's ok, please just leave a comment in the code about it to prevent confusing others before the refactor is made. Same goes for other changes that you want to keep for now but you consider changing in the future. This helps to understand which directions the component will take

internal/component/database_observability/mysql/component.go
cristiangreco: We want the metrics to come out of `database_observability.mysql` (or whatever final name). If we create them as metrics attached to the component itself then one would need to scrape the Alloy `/metrics` endpoint to ingest those metrics, right? It's not what we want in this case: don't think the discriminator should be the fact we currently have only 1 metric, as it pushes us toward a totally different usage model.

wildum: yes that's correct. I get your point and that's ok if you want to go this way. I do think that the amount of metrics you have matters a bit in the solution because if it stays with one metric that's actually a constant then it seems overkill to have the scrape complexity in the component + require the need to scrape this component just for this one info. But if you think that you might add more metrics to it later and that it would be worth it for users to scrape it then that's totally fine.

Also it may not work with clustering if the target is not exactly the same across all instances (not sure if that's the case or not for you. If it is then it should be warned in the doc because this is a common pitfall that confuses users)

internal/component/database_observability/mysql/component.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+var (
+	_ syntax.Defaulter = (*Arguments)(nil)
+	_ syntax.Validator = (*Arguments)(nil)
+)
+
+type Arguments struct {
+	DataSourceName alloytypes.Secret   `alloy:"data_source_name,attr"`
+	ScrapeInterval time.Duration       `alloy:"scrape_interval,attr,optional"`
```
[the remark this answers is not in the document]
cristiangreco: Renamed to `collect_interval`

internal/component/database_observability/mysql/component.go
```diff
@@ -0,0 +1,260 @@
...
+
+	return c, nil
+}
+
+func (c *Component) Run(ctx context.Context) error {
+	defer func() {
+		level.Info(c.opts.Logger).Log("msg", name+" component shutting down, stopping collectors")
+		c.mut.RLock()
+		for _, collector := range c.collectors {
+			collector.Stop()
```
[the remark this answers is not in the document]
cristiangreco: Is this a common pattern for components? I'm thinking that maybe we could mark `Stop()` as idempotent instead?

internal/component/database_observability/mysql/component.go
cristiangreco: Makes sense! I think it might be a nice followup to keep the component working until the new config is applied, but I'll leave it out for now.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -0,0 +1,61 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/database_observability.mysql/
+description: Learn about database_observability.mysql
+title: database_observability.mysql
+---
+
+Database Observability component. This component is under active development and can be run with Alloy flag `--stability.level=experimental`.
```
clayton-cornell, line 7:
```suggestion
<span class="badge docs-labels__stage docs-labels__item">Experimental</span>

# database_observability.mysql

{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
```

Am I correct in assuming content for blocks/attributes will be added at some point?

docs/sources/reference/components/grafanacloud/grafanacloud.database_observability.mysql.md
clayton-cornell: We could put in the basic scaffolding for the docs... the common headings that all component docs have.

````
## Usage

```alloy
component name and required parts
```

## Arguments

## Blocks
````

Even if some sections are empty in the early stages, putting in the heading structure will indicate more is coming.

Comment (cristiangreco) [reply docs/sources/reference/components/grafanacloud/grafanacloud.database_observability.mysql.md]: @clayton-cornell I've added the basic structure. I'm happy to add any scaffolding now and extend it as we refactor the component. Anything else that should be added at this point?

Comment (clayton-cornell) [reply docs/sources/reference/components/grafanacloud/grafanacloud.database_observability.mysql.md]: It's fine as-is now. The core elements/headings are in place. As long as we don't forget to eventually update and replace the `coming soon` :-)

File: docs/sources/reference/components/database_observability/database_observability.mysql.md [hunk 14]
```diff
@@ -0,0 +1,86 @@
...
+
+The following arguments are supported:
+
+| Name                 | Type           | Description                                                                                                         | Default | Required |
+| -------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `data_source_name`   | `secret`       | [Data Source Name](https://github.com/go-sql-driver/mysql#dsn-data-source-name) for the MySQL server to connect to. |         | yes      |
+
+## Blocks
+
+Coming soon.
```
clayton-cornell, line 32:
```suggestion
The following blocks are supported inside the definition of `database_observability.mysql`:

Hierarchy | Block | Description | Required
--------- | ----- | ----------- | --------
            |        |                |
```

We could replace the coming soon with the basic elements of the block section.

internal/component/database_observability/mysql/component.go
cristiangreco: Not planning to make this work for clustering. I do see your point about the number of metrics, I think we want to keep the door open for now and have this component produce its own metrics. Things will be more clear once we approach moving out of "experimental".

cristiangreco:
@wildum I think all major points are addressed and what's left is probably not a blocker for merging this PR? Could you please take another look?

internal/component/database_observability/mysql/component.go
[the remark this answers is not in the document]
wildum: ah sorry, I meant "does not crash" so yes it should be idempotent imo

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -0,0 +1,90 @@
...
+| -------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `data_source_name`   | `secret`       | [Data Source Name](https://github.com/go-sql-driver/mysql#dsn-data-source-name) for the MySQL server to connect to. |         | yes      |
+
+## Blocks
+
+The following blocks are supported inside the definition of `database_observability.mysql`:
+
+| Hierarchy | Block | Description | Required |
+| --------- | ----- | ----------- | -------- |
+|           |       |             |          |
```
wildum, line 36: nit: Replace this area by "The `database_observability.mysql` component does not support any blocks, and is configured fully through arguments."

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (c *SchemaTable) Stop() {
+	c.cancel()
+	c.dbConnection.Close()
```
wildum, line 123: I don't think that the SchemaTable should be responsible to close the connection that's given as an arg. Since that the collectors are sharing the connections,  if one closes it because of an error it will be closed for all collectors?

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name                 | Type           | Description                                                                                                         | Default | Required |
+| -------------------- | -------------- | ------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `data_source_name`   | `secret`       | [Data Source Name](https://github.com/go-sql-driver/mysql#dsn-data-source-name) for the MySQL server to connect to. |         | yes      |
```
wildum, line 28: it's missing the collect_interval and the forward_to

wildum (commented):
Thanks for working on the feedback, I just have a few little comments left and then we can merge it

docs/sources/reference/components/database_observability/database_observability.mysql.md
cristiangreco: Fixed

docs/sources/reference/components/database_observability/database_observability.mysql.md
cristiangreco: Fixed

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: Good point, initially I was passing `dbConnection` to the collectors, but now that `dbConnection` is handled by the component I've moved `dbConnection.Close()` there.

internal/component/database_observability/mysql/component.go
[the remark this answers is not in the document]
cristiangreco: I've added a comment to make sure it's kept idempotent (currently it just calls the `cancelFunc` of the context, which is an idempotent operation).

wildum (approved):

PR #2058, opened 2024-11-08
Add integration test for OTLP internal metrics

wildum:
In Alloy v1.4.0 some Otel metrics were renamed but we forgot to rename them in the k8s-monitoring helm-chart. This broke some user dashboards. (see https://github.com/grafana/k8s-monitoring-helm/pull/840)

This test will ensure that the internal otel metrics are available. 

It's also adding traces to the otel gen app + tempo in the docker compose. This is helpful when setting up a local lab to test configs.


thampiotr (approved):
I think this test is useful, however I wonder if it could be tested on a lower level, from our normal Go tests.

Still, let's merge this.

PR #2072, opened 2024-11-12
Update local.file.md

bcotton:
when trying to use `local.file` it seemed to be missing an example of how to use it.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/local/local.file.md
+++ b/docs/sources/reference/components/local/local.file.md
@@ -68,6 +68,8 @@ The read error will be exposed as a log message and in the debug information for
 
 ## Example
 
+The following example shows a simple `local.file` configuration that watches a passwords text file and uses the exported content field.
+
 ```alloy
 local.file "secret_key" {
   filename  = "/var/secrets/password.txt"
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/local/local.file.md
```diff
@@ -71,4 +73,10 @@ local.file "secret_key" {
   filename  = "/var/secrets/password.txt"
   is_secret = true
 }
+
+# Then later:
+grafana_cloud.stack "receivers" {
+  stack_name = "mystack"
+  token = local.file.secret_key.content # <-- The file's contents are here
+}
```
clayton-cornell, line 81:
```suggestion
grafana_cloud.stack "receivers" {
  stack_name = "mystack"
  token = local.file.secret_key.content
}
```

I'd suggest we remove the in-line config file comments and add a better description of the Example up under the heading on line 69.  Maybe something like this? (I'm not sure I'm getting the description right here though)

``The following example shows a simple `local.file` configuration that watches a passwords text file and exports the content field.``

clayton-cornell (approved):

PR #2084, opened 2024-11-13
Add otelcol.receiver.solace

### LINKED ISSUE #1993: Add otelcol.receiver.solace
Add [the receiver ](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/solacereceiver) to Alloy

Desired stability: GA

wildum:
#### PR Description

Add the solace receiver from the Otel contrib repo: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/solacereceiver

 Fixes #1993

#### Notes to the Reviewer

A few things to note:
- I tested it manually using the Solace image. This was not super easy to setup, I wrote a guide in the component for it.
- The upstream component uses a list for the broker but only the first element in the list is used. On Alloy side I decided to use a string to avoid confusion
- The upstream component has a retry policy when the next component fails to consume it. I believe that this can be a problem in Alloy when the component has several consumers: if one of the consumer fails, I think that they will all receive the data again, which might lead to data duplication. I added a warning about this in the doc. I don't think that there is a way around it but I also don't think that this is a big problem because it's unlikely to get two traces consumer at this level. If there is a need for branching, an easy workaround would be to just do the branching in the next component.
- I needed to fix a test because of the special case of having empty ptr structs

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
@@ -37,7 +37,7 @@ The following arguments are supported:
 | Name                 | Type     | Description                                                               | Default          | Required |
 | -------------------- | -------- | ------------------------------------------------------------------------- | ---------------- | -------- |
 | `queue`              | `string` | Name of the Solace telemetry queue to get span trace messages from.       |                  | yes      |
-| `broker`             | `string` | Name of the Solace broker using amqp over tls.                            | `localhost:5671` | no       |
+| `broker`             | `string` | Name of the Solace broker using AMQP over TLS.                            | `localhost:5671` | no       |
 | `max_unacknowledged` | `int`    | Maximum number of unacknowledged messages the Solace broker can transmit. | 10               | no       |
 
 `queue` must have the format `queue://#telemetry-myTelemetryProfile`.
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
@@ -1,0 +1,209 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.receiver.solace/
+description: Learn about otelcol.receiver.solace
+title: otelcol.receiver.solace
+---
+
+# otelcol.receiver.solace
+
+`otelcol.receiver.solace` accepts traces from a [Solace PubSub+ Event Broker](https://solace.com/products/event-broker/) and
+forwards it to other `otelcol.*` components.
+
+> **NOTE**: `otelcol.receiver.solace` is a wrapper over the upstream
+> OpenTelemetry Collector `solace` receiver from the `otelcol-contrib`
+> distribution. Bug reports or feature requests will be redirected to the
+> upstream repository, if necessary.
+
+Multiple `otelcol.receiver.solace` components can be specified by giving them
+different labels.
+
+## Usage
+
+```alloy
+otelcol.receiver.solace "LABEL" {
+  queue = "QUEUE"
+  auth {
+    // sasl_plain or sasl_xauth2 or sasl_external block
+  }
+  output {
+    traces  = [...]
+  }
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name                 | Type     | Description                                                               | Default          | Required |
+| -------------------- | -------- | ------------------------------------------------------------------------- | ---------------- | -------- |
+| `queue`              | `string` | Name of the Solace telemetry queue to get span trace messages from.       |                  | yes      |
+| `broker`             | `string` | Name of the Solace broker using amqp over tls.                            | `localhost:5671` | no       |
+| `max_unacknowledged` | `int`    | Maximum number of unacknowledged messages the Solace broker can transmit. | 10               | no       |
+
+`queue` should have the format `queue://#telemetry-myTelemetryProfile`.
+
+## Blocks
+
+The following blocks are supported inside the definition of
+`otelcol.receiver.solace`:
+
+| Hierarchy                      | Block              | Description                                                                                                                      | Required |
+| ------------------------------ | ------------------ | -------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| authentication                 | [authentication][] | Configures authentication for connecting to the Solace broker.                                                                   | yes      |
+| authentication > sasl_plain    | [sasl_plain][]     | Authenticates against Solace broker with SASL PLAIN.                                                                             | no       |
+| authentication > sasl_xauth2   | [sasl_xauth2][]    | Authenticates against Solace broker with SASL XOauth2.                                                                           | no       |
+| authentication > sasl_external | [sasl_external][]  | Authenticates against Solace broker with SASL External.                                                                          | no       |
+| flow                           | [flow][]           | Configures the behaviour to use when temporary errors are encountered from the next component.                                   | no       |
+| flow > delayed_retry           | [delayed_retry][]  | Sets the flow control strategy to `delayed retry` which will wait before trying to push the message to the next component again. | no       |
+| tls                            | [tls][]            | Configures TLS for connecting to the Solace broker.                                                                              | no       |
+| debug_metrics                  | [debug_metrics][]  | Configures the metrics which this component generates to monitor its state.                                                      | no       |
+| output                         | [output][]         | Configures where to send received telemetry data.                                                                                | yes      |
+
+One SASL authentication block is required in the `authentication` block.
+
+`sasl_external` must be used together with the `tls` block.
+
+The `>` symbol indicates deeper levels of nesting. For example,
+`authentication > tls` refers to a `tls` block defined inside an
+`authentication` block.
+
+[authentication]: #authentication-block
+[sasl_plain]: #sasl_plain-block
+[sasl_xauth2]: #sasl_xauth2-block
+[sasl_external]: #sasl_external-block
+[tls]: #tls-block
+[flow]: #flow-block
+[delayed_retry]: #delayed_retry-block
+[debug_metrics]: #debug_metrics-block
+[output]: #output-block
+
+### authentication block
+
+The `authentication` block configures how to authenticate for connecting to the Solace broker.
+It doesn't support any arguments and is configured fully through inner blocks.
+
+### sasl_plain block
+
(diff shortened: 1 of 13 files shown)


internal/component/otelcol/receiver/solace/solace.go
```diff
@@ -0,0 +1,149 @@
...
+		Broker:     "localhost:5671",
+		MaxUnacked: 1000,
+	}
+	args.Flow.SetToDefault()
+	args.DebugMetrics.SetToDefault()
+}
+
+// Validate implements syntax.Validator.
+func (args *Arguments) Validate() error {
+	if args.Auth.PlainText == nil && args.Auth.External == nil && args.Auth.XAuth2 == nil {
```
dehaansa, line 107: Should we validate that only one form of auth is provided? Also possibly worth upstreaming if we do choose to implement.

Looking at the upstream receiver the toAMPQAuthentication function in messaging_service.go uses PlainText if provided, then XAuth2, then External. If multiple are provided only one is used, though the config Validate upstream also doesn't error with multiple authentication forms provided.

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name                 | Type     | Description                                                               | Default          | Required |
+| -------------------- | -------- | ------------------------------------------------------------------------- | ---------------- | -------- |
+| `queue`              | `string` | Name of the Solace telemetry queue to get span trace messages from.       |                  | yes      |
+| `broker`             | `string` | Name of the Solace broker using amqp over tls.                            | `localhost:5671` | no       |
```
dehaansa, line 41: Should we consider keeping this the same as upstream, so when upstream is changed to utilize multiple broker support we don't need to change the config in Alloy?

internal/component/otelcol/receiver/solace/solace.go
wildum: Makes sense, I made the change here: https://github.com/grafana/alloy/pull/2084/commits/2c625213441551452cd45462455808453f5dcdee
Also makes sense to contribute upstream, I will create an issue in our board when this get merged to track the contribution

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
wildum: To support multiple brokers, they would need to change the structure of the config because with several brokers you probably want the possibility to specify a queue and credentials for each broker. Also they should change the name to "brokers" for consistency and change the doc since they refer to it as "the broker" + the component is already "old" and considered "stable" in otel standard. So I don't see this change happening in the near future / happening without other breaking changes in the config.
I would prefer to ask upstream if they are ok to switch to a string and if not, whether they are really planning to support multiple brokers at some point

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+
+# otelcol.receiver.solace
+
+`otelcol.receiver.solace` accepts traces from a [Solace PubSub+ Event Broker](https://solace.com/products/event-broker/) and
+forwards it to other `otelcol.*` components.
+
+> **NOTE**: `otelcol.receiver.solace` is a wrapper over the upstream
+> OpenTelemetry Collector `solace` receiver from the `otelcol-contrib`
+> distribution. Bug reports or feature requests will be redirected to the
+> upstream repository, if necessary.
```
clayton-cornell, line 15:
```suggestion
{{< admonition type="note" >}}
`otelcol.receiver.solace` is a wrapper over the upstream OpenTelemetry Collector `solace` receiver from the `otelcol-contrib` distribution.
Bug reports or feature requests will be redirected to the upstream repository, if necessary.
{{< /admonition >}}
```

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+
+Multiple `otelcol.receiver.solace` components can be specified by giving them
+different labels.
```
clayton-cornell, line 18:
```suggestion
You can specify multiple `otelcol.receiver.solace` components by giving them different labels.
```

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+
+The following arguments are supported:
+
+| Name                 | Type     | Description                                                               | Default          | Required |
+| -------------------- | -------- | ------------------------------------------------------------------------- | ---------------- | -------- |
+| `queue`              | `string` | Name of the Solace telemetry queue to get span trace messages from.       |                  | yes      |
+| `broker`             | `string` | Name of the Solace broker using amqp over tls.                            | `localhost:5671` | no       |
+| `max_unacknowledged` | `int`    | Maximum number of unacknowledged messages the Solace broker can transmit. | 10               | no       |
+
+`queue` should have the format `queue://#telemetry-myTelemetryProfile`.
```
clayton-cornell, line 44: Should or must? Should implies it can have some other format too.

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+
+| Hierarchy                      | Block              | Description                                                                                                                      | Required |
+| ------------------------------ | ------------------ | -------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| authentication                 | [authentication][] | Configures authentication for connecting to the Solace broker.                                                                   | yes      |
+| authentication > sasl_plain    | [sasl_plain][]     | Authenticates against Solace broker with SASL PLAIN.                                                                             | no       |
+| authentication > sasl_xauth2   | [sasl_xauth2][]    | Authenticates against Solace broker with SASL XOauth2.                                                                           | no       |
+| authentication > sasl_external | [sasl_external][]  | Authenticates against Solace broker with SASL External.                                                                          | no       |
```
clayton-cornell, line 56:
```suggestion
| authentication > sasl_plain    | [sasl_plain][]     | Authenticates against the Solace broker with SASL PLAIN.                                                                         | no       |
| authentication > sasl_xauth2   | [sasl_xauth2][]    | Authenticates against the Solace broker with SASL XOauth2.                                                                       | no       |
| authentication > sasl_external | [sasl_external][]  | Authenticates against the Solace broker with SASL External.                                                                      | no       |
```

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
wildum: "must" is better then

thampiotr:
> The upstream component has a retry policy when the next component fails to consume it.

That's odd, no? Any idea why did they add this? Shouldn't this be something done on the runtime level?

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+## Arguments
+
+The following arguments are supported:
+
+| Name                 | Type     | Description                                                               | Default          | Required |
+| -------------------- | -------- | ------------------------------------------------------------------------- | ---------------- | -------- |
+| `queue`              | `string` | Name of the Solace telemetry queue to get span trace messages from.       |                  | yes      |
+| `broker`             | `string` | Name of the Solace broker using amqp over tls.                            | `localhost:5671` | no       |
```
thampiotr, line 41:
```suggestion
| `broker`             | `string` | Name of the Solace broker using AMQP over TLS.                            | `localhost:5671` | no       |
```

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
```diff
@@ -0,0 +1,210 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.receiver.solace/
+description: Learn about otelcol.receiver.solace
+title: otelcol.receiver.solace
+---
+
+# otelcol.receiver.solace
```
thampiotr, line 7: Is it becoming stable right away?
I think you want to make it exprimental / public preview first?

internal/component/otelcol/receiver/solace/solace.go
```diff
@@ -0,0 +1,159 @@
...
+		authMethod++
+	}
+	if args.Auth.External != nil {
+		authMethod++
+	}
+	if args.Auth.XAuth2 != nil {
+		authMethod++
+	}
+	if authMethod != 1 {
+		return fmt.Errorf("the auth block must contain either a sasl_plain block, a sasl_xauth2 block or a sasl_external block")
```
thampiotr, line 118:
```suggestion
		return fmt.Errorf("the auth block must contain exactly one of sasl_plain block, sasl_xauth2 block or sasl_external block")
```

wildum:
> That's odd, no? Any idea why did they add this? Shouldn't this be something done on the runtime level?

Yet not a fan of this + it makes their implementation much more complex. This was [done 2 years ago](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16570), maybe they wanted some extra safety because the component might be spamming a lot? They did not add any other flow control strategy since then but there is also [no way around it.](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/solacereceiver/config.go#L55C1-L59C3)

internal/component/otelcol/receiver/solace/solace.go
wildum: Upstream contribution: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36386

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
wildum: Upstream contribution: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36387

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
wildum: I updated it

docs/sources/reference/components/otelcol/otelcol.receiver.solace.md
wildum: discussed internally

PR #2095, opened 2024-11-14
True up the documentation with the code.

mattdurham:
Documentation only change, verified these flags are both false by default.


dehaansa (approved):

clayton-cornell (approved):

clayton-cornell:
This was documented as `true` back to v1.0. Should this change be backported? or is this a recent code change?

mattdurham:
I think this has always been default to false, even node_exporter the underlying library sets it to false.

Issue #2000, opened 2024-10-30
otelcol.processor.attributes delete pattern not working?


fede843:
### What's wrong?

I am trying to delete an attribute based on a patter.
The pattern is totally ignored. No matter what I set, it always delete the attribute or not based on the key matching.

Even more, docs says:

> delete: Deletes an attribute from the input data.
> The key attribute and/or the pattern attribute is required. It specifies the attribute to act upon.

Well, if I set only a pattern, I get an error that key is missing.

Not sure if the error is introduced in Alloy or if comes from upstream OTEL Collector.

### Steps to reproduce

You can try to remove any attribute, for instance the http.method

    action {
	key="http.method"
        pattern= `GET`
        action = "delete"
    }

Regardless what you set or not in pattern, it always deletes the attribute.

### System information

Docker

### Software version

Alloy 1.4.3

wildum:
Hello, thanks for reporting this, it's indeed a bug! Alloy assumes that the key is always required but it should only be required for the actions insert, update, and upsert. For the action `delete`, only `key`, only `pattern`, and both `key/pattern` should be supported.

wildum:
solved by https://github.com/grafana/alloy/pull/2124

PR #2124, opened 2024-11-18
Fix key/pattern logic for the attribute processor

wildum:
#### PR Description

The `key` argument was set as required but it's not required upstream for the action "delete" and "hash".
This PR fixed the issue by introducing a validation step that checks for the key and the pattern argument depending on the action.

#2000

See the upstream code for reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/attraction/attraction.go#L173C1-L186C4

#### PR Checklist

- [x] CHANGELOG.md updated
- [na] Documentation added // the doc is not updated because it was matching with upstream
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Main (unreleased)
 
 - Fixed issue with reloading configuration and prometheus metrics duplication in `prometheus.write.queue`. (@mattdurham)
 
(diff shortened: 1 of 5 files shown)


dehaansa (approved):

mattdurham (approved):
lgtm

PR #2126, opened 2024-11-18
Update build image to 1.23.3.

mattdurham:
Update the build image to 1.23


wildum (commented):
I think that you also need to update the version in .github/workflows/check-linux-build-image.yml and there is also a comment is tools/build-image/Dockerfile about the version

mattdurham:
Good call on the workflow, what comment are you referring to?

wildum:
> Good call on the workflow, what comment are you referring to?

https://github.com/grafana/alloy/blob/main/tools/build-image/Dockerfile#L7

wildum:
ah but it's only relevant for the next PR actually

Issue #312, opened 2023-11-03
FLOW: Add loki.remote.syslog 


ToMMy86:
### Request

It would be **extremely** usefull to have logs forwarded/converted from loki component  to remote  syslogs (  loki.remote.syslog  ) in RFC format.

Configuration of  loki.remote.syslog: 
HOST, PROTOCOL, PORT 

The logs can be directly forwarded as they come from the component in RFC format ( loki.source.syslog ), but it will also be helpfull for the loki receiver.

From my understanding of the documentation the "loki.source.syslog" component can only forward to other loki.* components.

![image](https://github.com/grafana/agent/assets/3274761/14c7fda6-1502-4050-85f4-0342f4422bee)

### Use case

I manage the logging and metrics delivery in the nodes I administer using the Grafana stack. The key strength is the ability to use just one agent for centralizing monitoring, and this is, for me (and I imagine for many others), the most valuable aspect, especially when dealing with hundreds of nodes. So, all the Grafana agents in the administered nodes forward logs to a central cluster where a "Grafana agent in flow mode" is present. This way, I can route and filter logs from various sources and manage pipelines on them through a single central configuration, which is quite fantastic.

I use Grafana for visualize certain logs stored in Loki (application logs, databases, etc.), but Grafana is not a suitable tool for log analysis, advanced correlations, or threat hunting ( LogQL is powerfull but overall visualizing loki logs with grafana it's not comparable to proper SIEM systems as of now ). 
The tools suitable for these purposes are SIEM and unfortunately, it's very unlikely that they have "Loki" as a log ingestion integration (at least I'm not aware of any at the moment), but they all support syslog. Syslog ingestion is always supported.

What I would like to achieve while maintaining my current architectural monitoring design is forwarding some logs from my environment (both coming from Syslog or/and Grafana Agents on nodes ) to a SIEM such as Qradar, Wazuh, Splunk, Dataset, whatever..  via SYSLOG (because it's a standard input format they always support). In this scenario, I would have the Central Flow-Agents  managing these pipelines/routing also for 3rd party integrations and it would be awesome...

At the moment, I'm struggling a lot to find a solution, and it doesn't seem even feasible if i push logs via grafana-agents to loki endpoint (  to central flow-agent  loki receivers  ) .. i have to somehow proxy with other tools and it's a completely mess. 

[...]

PR #2132, opened 2024-11-19
feat: OpenTelemetry syslog exporter

dehaansa:
#### PR Description
Add support for the syslog exporter from OpenTelemetry to Alloy. There are several user requests for this component with varying requirements. The upstream component is in `alpha` stability, but does not appear to be unstable at this point in time.

https://github.com/grafana/alloy/issues/312

#### Notes to the Reviewer
There is significant data-massaging required to use this component well. I attempted to provide good examples in the docs, but this is definitely a good use case for better docs & task based docs moving forward. If there's anything in the docs that feels insufficient please raise it and I'll address.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
@@ -34,7 +34,7 @@ otelcol.exporter.syslog "LABEL" {
 The exporter creates one syslog message for each log record based on the following attributes of the log record.
 If an attribute is missing, the default value is used. The log's timestamp field is used for the syslog message's time.
 RFC3164 only supports a subset of the attributes supported by RFC5424, and the default values are not the same between
-the two protocols. See the [OpenTelemetry documentation][upstream_readme] for the exporter for more details.
+the two protocols. Refer to the [OpenTelemetry documentation][upstream_readme] for the exporter for more details.
 
 | Attribute name    | Type   | RFC5424 Default value  | RFC3164 supported | RFC3164 Default value
 | ----------------- | ------ | ---------------------- |------------------ | ----------------------
@@ -128,7 +128,7 @@ The following fields are exported and can be referenced by other components:
 
 ### TCP endpoint without TLS
 
-This example creates an exporter to send data to a syslog server expecting RFC5424 compliant messages over TCP without TLS:
+This example creates an exporter to send data to a syslog server expecting RFC5424-compliant messages over TCP without TLS:
 
 ```alloy
 otelcol.exporter.syslog "default" {
@@ -140,7 +140,7 @@ otelcol.exporter.syslog "default" {
 }
 ```
 
-### Using the `otelcol.processor.transform` component to format logs from `loki.source.syslog`
+###  Use the `otelcol.processor.transform` component to format logs from `loki.source.syslog`
 
 This example shows one of the methods for annotating your loki messages into the format expected 
 by the exporter using a `otelcol.receiver.loki` component in addition to the `otelcol.processor.transform` 
@@ -181,7 +181,7 @@ otelcol.processor.transform "syslog" {
 }
 ```
 
-### Using the `otelcol.processor.transform` component to format otel logs
+### Use the `otelcol.processor.transform` component to format OpenTelemetry logs
 
 This example shows one of the methods for annotating your messages in the OpenTelemetry log format into the format expected 
 by the exporter using an `otelcol.processor.transform` component. This example assumes that the log messages being 


docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,346 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.syslog/
+aliases:
+  - ../otelcol.exporter.syslog/ # /docs/alloy/latest/reference/components/otelcol.exporter.syslog/
```
clayton-cornell, line 4:
```suggestion
canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.syslog/
```

Not necessary on a new topic. We could add the current location, but the alias that was included here is for a location that this topic has never been published to.

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,346 @@
...
+## Usage
+
+```alloy
+otelcol.exporter.syslog "LABEL" {
+  endpoint = "HOST"
+}
+```
+
+<!-- 
+The next few paragraphs were copied from the OTel docs:
```
clayton-cornell, line 35: Unless there's a really good reason to copy in external documentation, I'd strongly prefer that we just link to the external docs.  We could paraphrase things... if that helps?

If we do need include external docs that are copied verbatim, then we need to give attribution and make sure we stick within the license of the doc.

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Debug information
+
+`otelcol.exporter.syslog` doesn't expose any component-specific debug information.
+
+## Examples
+
+### TCP endpoint without TLS
+
+This example creates an exporter to send data to a syslog server expecting rfc5424 compliant messages over TCP without TLS:
```
clayton-cornell, line 233:
```suggestion
This example creates an exporter to send data to a syslog server expecting RFC5424 compliant messages over TCP without TLS:
```

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
dehaansa: I had paraphrased, but then I saw a copy-paste block in the `otelcol.processor.probabilistic_sampler` doc and assumed it was a standard, looks like it's the opposite 🙃 . I do think having the exact copied tables of attribute mapping is important, but I'll work on a refactor in the morning that does less copying and more attribution.

dehaansa: I updated the docs to consolidate the details & link out, much more of a re-write now so I didn't add full attribution.

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,244 @@
...
+description: Learn about otelcol.exporter.syslog
+title: otelcol.exporter.syslog
+---
+
+# otelcol.exporter.syslog
+
+{{< docs/shared lookup="stability/public_preview.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`otelcol.exporter.syslog` accepts logs from other `otelcol` components and writes them over the network using the syslog protocol. 
+It supports syslog protocols [RFC5424][RFC5424] and [RFC3164][RFC3164] and can send data over `TCP` or `UDP`.
```
clayton-cornell, line 12:
```suggestion
It supports syslog protocols [RFC5424][] and [RFC3164][] and can send data over `TCP` or `UDP`.
```

We can drop the second element if the two elements are identical

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,244 @@
...
+  endpoint = "HOST"
+}
+```
+
+### Supported Attributes
+
+The exporter creates one syslog message for each log record based on the following attributes of the log record.
+If an attribute is missing, the default value is used. The log's timestamp field is used for the syslog message's time.
+RFC3164 only supports a subset of the attributes supported by RFC5424, and the default values are not the same between
+the two protocols. See the [OpenTelemetry documentation][upstream_readme] for the exporter for more details.
```
clayton-cornell, line 37:
```suggestion
the two protocols. Refer to the [OpenTelemetry documentation][upstream_readme] for the exporter for more details.
```

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Debug information
+
+`otelcol.exporter.syslog` doesn't expose any component-specific debug information.
+
+## Examples
+
+### TCP endpoint without TLS
+
+This example creates an exporter to send data to a syslog server expecting RFC5424 compliant messages over TCP without TLS:
```
clayton-cornell, line 131:
```suggestion
This example creates an exporter to send data to a syslog server expecting RFC5424-compliant messages over TCP without TLS:
```

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,244 @@
...
+otelcol.exporter.syslog "default" {
+  endpoint = "localhost"
+  tls {
+      insecure             = true
+      insecure_skip_verify = true
+  }
+}
+```
+
+### Using the `otelcol.processor.transform` component to format logs from `loki.source.syslog`
```
clayton-cornell, line 143:
```suggestion
###  Use the `otelcol.processor.transform` component to format logs from `loki.source.syslog`
```

docs/sources/reference/components/otelcol/otelcol.exporter.syslog.md
```diff
@@ -0,0 +1,244 @@
...
+
+  output {
+    metrics = []
+    logs    = [otelcol.exporter.syslog.default.input]
+    traces  = []
+  }
+}
+```
+
+### Using the `otelcol.processor.transform` component to format otel logs
```
clayton-cornell, line 184:
```suggestion
### Use the `otelcol.processor.transform` component to format OpenTelemetry logs
```

clayton-cornell (commented):
A few minor tweaks Otherwise looking good.

clayton-cornell:
Docs look ok. I guess it over to someone else on the team for a code review?

dehaansa:
> LGTM, only comment I would have is it would be nice to have live debugging. Might be reasonable to do that in a separate PR though.

I'm happy to look at live debugging support in this PR or immediate followup, either is fine with me.

wildum:
> I'm happy to look at live debugging support in this PR or immediate followup, either is fine with me.

Otel exporters don't support live debugging yet. To add it you would need to plug it to the [common exporter](https://github.com/grafana/alloy/blob/main/internal/component/otelcol/exporter/exporter.go) struct so that all Otel exporters get it. Because it would impact all exporters, it should be in a dedicated PR

PR #2145, opened 2024-11-22
remove metrics that are not there

mattdurham:
This resolves metrics that are not implemented yet.


clayton-cornell (approved):

clayton-cornell:
@mattdurham  Is this one that can/should be backported?

mattdurham:
Yes, its only in 1.5.x

PR #2151, opened 2024-11-25
Merge scope hierarchy for evaluation

wildum:
#### Description of the problem

The [evaluateExpr](https://github.com/grafana/alloy/blob/main/syntax/vm/vm.go#L296) func in the vm pkg recursively evaluate expressions. If you want to evaluate this:
`a = [loki.write.endpoint.receiver]`
the [loki.write.endpoint.receiver] will be represented by nested structs:
- ArrayExpr
  - AccessExpr (receiver)
    - AccessExpr (endpoint)
       - AccessExpr (write)
         - IdentifierExpr (loki)
The evaluate will go through the nesting recursively and when it arrives at the IdentifierExpr, it will use the [scope](https://github.com/grafana/alloy/blob/main/syntax/vm/vm.go#L487) to look it up.
The scope looks like this:
Parent (the parent scope)
Variables (nested maps)
"loki"
"write"
"endpoint"
The Lookup will check whether the identifier is in the variables. Here it matches "loki" so it return the associated map:
"write"
"endpoint"
Then the recursion unfolds because "write" will match and it will return "endpoint" which will also match and return the receiver object.
During the lookup, if "loki" would not have been found in the scope, it would have looked up at the parent scope recursively till there is a match. If there is no match and no more parent then it's an error.
The problem with this process is that if you have the following:
Scope:
Parent
Parent
Variables
"loki"
"write"
"endpoint2"
Variables
"loki"
"write"
"endpoint"
And you want to evaluate the expression:
a = [loki.write.endpoint2.receiver]
The algorithm will match with the wrong path because it will match "loki" and only at the last AccessExpr it will realise that "endpoint" does not match with "endpoint2" and it will throw an error.
This was not a problem so far because the ParentScope is only used for the ModulePath AFAIK. But now I'm using it for the foreach because unlike modules that uses "arguments" I think that it's nicer if the components in the foreach have access to the outside:
- The scope contains the info in the foreach
- The parent scope contains the info from outside of the foreach

#### Solution

This solution merges the variables of the scope hierarchy into one variable map. I could not find a way to change it easily in syntax pkg. Solving it in the value_cache seemed easier and safer (thanks @rfratto for the hint).

### DIFF
--- a/internal/runtime/internal/controller/value_cache.go
+++ b/internal/runtime/internal/controller/value_cache.go
@@ -199,6 +199,17 @@ func (vc *valueCache) BuildContext() *vm.Scope {
 		}
 	}
 
+	// The syntax pkg only checks the first key when choosing the scope to evaluate an expression.
+	// if you have:
+	// - parent: "test": "component1": v
+	// - child: "test": "component2": v
+	// the expression "test.component1.v" would result in an error because it would use the child scope.
+	// To avoid such conflict, we merge the scopes into one:
+	// - parent: nil
+	// - child: "test": {"component1": v, "component2": v}
+	scope.Variables = mergeScopes(scope)
+	scope.Parent = nil
+
 	return scope
 }
 
@@ -233,4 +244,50 @@ func (vc *valueCache) buildValue(from []ComponentID, offset int) interface{} {
 		attrs[label] = vc.buildValue(ids, offset+1)
 	}
 	return attrs
+}
+
+// mergeScopes merges the Variables maps of the scope hierarchy into a new map.
+// The child always overrides the parent's value.
+func mergeScopes(scope *vm.Scope) map[string]any {
+	merged := make(map[string]any)
+
+	var mergeFrom func(*vm.Scope)
+	mergeFrom = func(s *vm.Scope) {
+		if s == nil {
+			return
+		}
+		mergeFrom(s.Parent)
+		for k, v := range s.Variables {
+			if pm, ok1 := merged[k].(map[string]any); ok1 {
+				if cm, ok2 := v.(map[string]any); ok2 {
+					merged[k] = mergeNestedMaps(pm, cm)
+					continue
+				}
+			}
+			merged[k] = v
+		}
+	}
+
+	mergeFrom(scope)
+	return merged
+}
+
+func mergeNestedMaps(parent, child map[string]any) map[string]any {
+	merged := make(map[string]any)
+
+	for k, v := range parent {
+		merged[k] = v
+	}
+
+	for k, v := range child {
+		if pm, ok1 := merged[k].(map[string]any); ok1 {
+			if cm, ok2 := v.(map[string]any); ok2 {
+				merged[k] = mergeNestedMaps(pm, cm)
+				continue
+			}
+		}
+		merged[k] = v
+	}
+
+	return merged
 }
--- a/internal/runtime/internal/controller/value_cache_test.go
+++ b/internal/runtime/internal/controller/value_cache_test.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"testing"
 
+	"github.com/grafana/alloy/syntax/vm"
 	"github.com/stretchr/testify/require"
 )
 
@@ -173,4 +174,120 @@ func TestModuleArgumentCache(t *testing.T) {
 			require.Equal(t, map[string]any{}, res.Variables)
 		})
 	}
+}
+
+func TestScopeMerge(t *testing.T) {
+	vc := newValueCache()
+	scope := vm.NewScope(
+		map[string]any{
+			"test": map[string]any{
+				"scope": barArgs{Number: 13},
+			},
+		},
+	)
+	vc.SetScope(scope)
+	vc.CacheExports(ComponentID{"foo", "bar"}, barArgs{Number: 12})
+	res := vc.BuildContext()
+
+	expected := map[string]any{
+		"test": map[string]any{
+			"scope": barArgs{
+				Number: 13,
+			},
+		},
+		"foo": map[string]any{
+			"bar": barArgs{
+				Number: 12,
+			},
+		},
+	}
+	require.Equal(t, expected, res.Variables)
+}
+
+func TestScopeMergeConflict(t *testing.T) {
+	vc := newValueCache()
+	scope := vm.NewScope(
+		map[string]any{
+			"test": map[string]any{
+				"scope": barArgs{Number: 13},
+			},
+		},
+	)
+	vc.SetScope(scope)
+	vc.CacheExports(ComponentID{"test", "bar"}, barArgs{Number: 12})
+	res := vc.BuildContext()
+
+	expected := map[string]any{
+		"test": map[string]any{
+			"scope": barArgs{
+				Number: 13,
+			},
+			"bar": barArgs{
+				Number: 12,
+			},
+		},
+	}
+	require.Equal(t, expected, res.Variables)
+}
+
+func TestScopeMergeOverride(t *testing.T) {
+	vc := newValueCache()
+	scope := vm.NewScope(
+		map[string]any{
+			"test": map[string]any{
+				"scope": barArgs{Number: 13},
+			},
+		},
+	)
+	vc.SetScope(scope)
+	vc.CacheExports(ComponentID{"test", "scope"}, barArgs{Number: 12})
+	res := vc.BuildContext()
+
+	expected := map[string]any{
+		"test": map[string]any{
+			"scope": barArgs{
+				Number: 12,
+			},
+		},
+	}
+	require.Equal(t, expected, res.Variables)
+}
+
+func TestScopeMergeMoreNesting(t *testing.T) {
+	vc := newValueCache()
+	parentScope := vm.NewScope(
+		map[string]any{
+			"test": map[string]any{
+				"cp1": map[string]any{
+					"scope": barArgs{Number: 13},
+				},
+			},
+		},
+	)
+	scope := vm.NewScopeWithParent(
+		parentScope,
+		map[string]any{
+			"test": map[string]any{
+				"cp2": barArgs{Number: 12},
+			},
+		},
+	)
+	vc.SetScope(scope)
+	vc.CacheExports(ComponentID{"test", "cp1", "foo"}, barArgs{Number: 12})
+	vc.CacheExports(ComponentID{"test", "cp1", "bar", "fizz"}, barArgs{Number: 2})
+	res := vc.BuildContext()
+
+	expected := map[string]any{
+		"test": map[string]any{
+			"cp1": map[string]any{
+				"scope": barArgs{Number: 13},
+				"foo":   barArgs{Number: 12},
+				"bar": map[string]any{
+					"fizz": barArgs{Number: 2},
+				},
+			},
+			"cp2": barArgs{Number: 12},
+		},
+	}
+	require.Equal(t, expected, res.Variables)
 }


internal/runtime/internal/controller/value_cache.go
```diff
@@ -199,6 +199,17 @@ func (vc *valueCache) BuildContext() *vm.Scope {
...
+	// The syntax pkg only checks the first key when choosing the scope to evaluate an expression.
+	// if you have:
+	// - parent: "test": "component1": v
+	// - child: "test": "component2": v
+	// the expression "test.component1.v" would result in an error because it would use the child scope.
+	// To avoid such conflict, we merge the scopes into one:
+	// - parent: nil
+	// - child: "test": {"component1": v, "component2": v}
+	scope.Variables = mergeScopes(scope)
+	scope.Parent = nil
```
ptodev, line 211: Should we remove the `scope.Parent` variable altogether?

internal/runtime/internal/controller/value_cache.go
```diff
@@ -234,3 +245,49 @@ func (vc *valueCache) buildValue(from []ComponentID, offset int) interface{} {
 	}
 	return attrs
 }
+
+// mergeScopes merges the Variables maps of the scope hierarchy into a new map.
+// The child always overrides the parent's value.
+func mergeScopes(scope *vm.Scope) map[string]any {
```
ptodev, line 251: Do you think there is any performance impact from this? I suppose it depends on the scenario, since there might be cases where it's efficient to merge upfront e.g. when there's a deeply nexted structure of modules and we instantiate a lot components from the outermost layer inside the innermost layer.

ptodev (dismissed):

internal/runtime/internal/controller/value_cache.go
wildum: Without the parent, the merge is a bit simpler but still needed. Both solutions have a performance impact though because the scope is never nil (it contains at least the [module_path](https://github.com/grafana/alloy/blob/main/internal/runtime/alloy.go#L312))

wildum: Maybe it's time for a deeper refactor of the value_cache. I will try something different.

syntax/vm/vm.go
```diff
@@ -475,22 +470,13 @@ func NewScope(variables map[string]interface{}) *Scope {
...
-	}
-}
-
-// Lookup looks up a named identifier from the scope, all of the scope's
-// parents, and the stdlib.
+// Lookup looks up a named identifier from the scope and the stdlib.
 func (s *Scope) Lookup(name string) (interface{}, bool) {
 	// Traverse the scope first, then fall back to stdlib.
-	for s != nil {
+	if s != nil {
```
mattdurham, line 476: Is this still a traversal? Comment may need some cleanup

wildum: no not anymore, will update

internal/runtime/internal/controller/value_cache.go
```diff
@@ -234,3 +239,26 @@ func (vc *valueCache) buildValue(from []ComponentID, offset int) interface{} {
 	}
 	return attrs
 }
+
+// deepMergeMaps merges two maps. It uses the values of map2 in case of conflict.
```
mattdurham, line 243: So in this case it uses the child var with the parent var in case of a conflict?

wildum: yes, not sure if conflicts should be allowed at all though. I will give another go at this by changing the value_cache a bit more

internal/runtime/internal/controller/value_cache.go
```diff
@@ -130,107 +135,98 @@ func (vc *valueCache) ExportChangeIndex() int {
...
-	scope := vm.NewScopeWithParent(vc.scope, make(map[string]interface{}))
+// cleanup removes the ComponentID path from the map
+func cleanup(m map[string]any, id ComponentID, index int) error {
 
-	// First, partition components by Alloy block name.
-	var componentsByBlockName = make(map[string][]ComponentID)
-	for _, id := range vc.components {
-		blockName := id[0]
-		componentsByBlockName[blockName] = append(componentsByBlockName[blockName], id)
+	if _, ok := m[id[index]]; !ok {
```
ptodev, line 166: What does `id[index]` mean? I thought `ComponentID` is a string?

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,93 @@
...
 	"github.com/grafana/alloy/syntax/vm"
 )
 
-// valueCache caches component arguments and exports to expose as variables for
-// Alloy expressions.
-//
-// The current state of valueCache can then be built into a *vm.Scope for other
-// components to be evaluated.
+// This special keyword is used to expose the argument values to the custom components.
+const argumentLabel = "argument"
```
ptodev, line 13: Would it be cleaner to just have another map for arguments?

wildum:
> I find both the new and the old code hard to understand though... Maybe it's just me but it feels too convoluted. 

Thanks for the review, it's definitely not just you, it's hard to simplify it because it must build the scope tree which is quite a complex object to make + the fact that it must handle the args for the modules and the "parent" context. I also don't want to start a massive refactoring of everything. I will work on the review feedback and see whether I can simplify it a bit more. 

Maybe further refactoring PRs can follow to simplify other areas + add doc + add tests to consolidate the logic and simplify the code

internal/runtime/internal/controller/loader.go
```diff
@@ -166,7 +168,7 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 
 	var (
 		components   = make([]ComponentNode, 0)
-		componentIDs = make([]ComponentID, 0)
+		componentIDs = make(map[string]ComponentID)
```
mattdurham, line 171: Is there a reason we switched to strings here? This feels like a regression.

internal/runtime/internal/controller/loader.go
```diff
@@ -260,7 +262,13 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 	l.componentNodes = components
 	l.serviceNodes = services
 	l.graph = &newGraph
-	l.cache.SyncIDs(componentIDs)
+	err := l.cache.SyncIDs(componentIDs)
+	if err != nil {
```
mattdurham, line 266: Should this return immediately?

internal/runtime/internal/controller/value_cache.go
```diff
@@ -130,107 +135,98 @@ func (vc *valueCache) ExportChangeIndex() int {
...
+	if len(argsMap) == 0 {
+		delete(vc.scope.Variables, argumentLabel)
+	}
+}
 
-	// Then, convert each partition into a single value.
-	for label, ids := range componentsByLabel {
-		attrs[label] = vc.buildValue(ids, offset+1)
+// GetScope returns the current scope.
+func (vc *valueCache) GetScope() *vm.Scope {
```
mattdurham, line 215: nit: Should this be `GetScopyCopy`? Normally not super caring about this but this is dense code and updates to this scope will not affect the current scope? Is that right?

internal/runtime/internal/controller/value_cache.go
```diff
@@ -130,107 +135,98 @@ func (vc *valueCache) ExportChangeIndex() int {
...
+	vc.mut.RLock()
+	defer vc.mut.RUnlock()
+	return vm.NewScope(deepCopyMap(vc.scope.Variables))
+}
+
+func deepCopyMap(original map[string]any) map[string]any {
```
mattdurham, line 221: How does this interact with capsule types? Those are by pointer so will get copied into the new copy?

wildum: yes that's correct, only the maps are created new but the content of the maps is copied. This was also the case before where the exports values were copied when building the scope

internal/runtime/internal/controller/loader.go
wildum: I doesnt have to because it's adding to the diags so Alloy will exit anyway but I agree that's a little bit cleaner so I will make the change

internal/runtime/internal/controller/loader.go
wildum: I'm using the string representation of the node ID as a key: `componentIDs[n.ID().String()] = n.ID()`
This is how the components are stored in the value_cache (this is how they were stored before as well). Using the map in the loader directly is helpful to compare the current one vs the new one in the SyncIDs function.
We could still keep it as a slice in the loader and convert it to a map in syncIDS but I thought that using a map directly would make things easier

internal/runtime/internal/controller/value_cache.go
wildum: yeah I agree, I wanted to have the Scope fully built so that when a node wants to evaluate it can just get a copy of it but I don't think that's worth the added complexity in GetModuleArgument, CacheModuleArgument and SyncModuleArgs

internal/runtime/internal/controller/value_cache.go
```diff
@@ -130,107 +135,98 @@ func (vc *valueCache) ExportChangeIndex() int {
...
-func (vc *valueCache) SyncModuleArgs(args map[string]any) {
-	vc.mut.Lock()
-	defer vc.mut.Unlock()
-
-	for id := range vc.moduleArguments {
-		if _, keep := args[id]; keep {
-			continue
+	// Remove the component exports from the scope.
+	for _, id := range cleanupIds {
+		err := cleanup(vc.scope.Variables, id, 0)
```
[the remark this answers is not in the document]
wildum: this is the index that refers to the first part of the componentID (see the answer of the comment below). I will add some comments

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,93 @@
...
+	return nil
+}
+
+func (vc *valueCache) GetModuleArgument(key string) (any, bool) {
+	vc.mut.RLock()
+	defer vc.mut.RUnlock()
+	if _, ok := vc.scope.Variables[argumentLabel]; !ok {
+		return nil, false
+	}
+	v, exist := vc.scope.Variables[argumentLabel].(map[string]any)[key]
```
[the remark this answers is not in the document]
wildum: good point, I will use a separate map for the arguments but else I would have changed it

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,93 @@
...
 	mut                sync.RWMutex
-	components         map[string]ComponentID // NodeID -> ComponentID
-	args               map[string]interface{} // NodeID -> component arguments value
-	exports            map[string]interface{} // NodeID -> component exports value
-	moduleArguments    map[string]any         // key -> module arguments value
-	moduleExports      map[string]any         // name -> value for the value of module exports
-	moduleChangedIndex int                    // Everytime a change occurs this is incremented
-	scope              *vm.Scope              // scope provides additional context for the nodes in the module
+	componentIds       map[string]ComponentID
+	moduleExports      map[string]any // name -> value for the value of module exports
```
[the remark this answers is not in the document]
wildum: componentIds -> string representations of the node IDs
moduleExports -> labels of the export blocks
I will add some comments

internal/runtime/internal/controller/loader.go
```diff
@@ -829,12 +840,18 @@ func (l *Loader) evaluate(logger log.Logger, bn BlockNode) error {
...
-		// Always update the cache both the arguments and exports, since both might
-		// change when a component gets re-evaluated. We also want to cache the arguments and exports in case of an error
-		l.cache.CacheArguments(c.ID(), c.Arguments())
-		l.cache.CacheExports(c.ID(), c.Exports())
+		// Always update the cache both the exports, since that it might change when a component gets re-evaluated.
+		// We also want to cache it in case of an error
+		err2 := l.cache.CacheExports(c.ID(), c.Exports())
+		if err2 != nil {
+			level.Error(logger).Log("msg", "failed to cache exports after evaluation", "err", err2)
+			// Don't mask the previous evaluation error.
```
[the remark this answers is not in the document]
wildum: yeah not a big fan of this, I think that when there is an evaluation error in modules, this checks whether a mandatory argument is missing. In this case it shadows the evaluation error because it might be just a consequence of the fact that the arguments are missing. I think that there should be better ways to do this but I don't want to modify it in this PR, I can add comments though

internal/runtime/internal/controller/value_cache.go
```diff
@@ -15,46 +16,26 @@ import (
 // components to be evaluated.
 type valueCache struct {
 	mut                sync.RWMutex
-	components         map[string]ComponentID // NodeID -> ComponentID
-	args               map[string]interface{} // NodeID -> component arguments value
-	exports            map[string]interface{} // NodeID -> component exports value
```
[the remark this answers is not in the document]
wildum: I think that your comment refers to an older version of the code, the current version is:
```
// valueCache caches exports to expose as variables for Alloy expressions.
// The exports are stored directly in the scope which is used to evaluate Alloy expressions.
```
I should probably modify it to include the module arguments and module exports

wildum:
@mattdurham @ptodev Thank you both for the review, I answered your comments and made some modifications. Please take another look when you have time and feel free to add more comments or unresolved existing ones. We can also discuss it over a call if you have concerns about the approach or find some parts confusing and would like to talk about alternatives

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,90 @@
...
 // be nil to store an empty object.
-func (vc *valueCache) CacheExports(id ComponentID, exports component.Exports) {
+func (vc *valueCache) CacheExports(id ComponentID, exports component.Exports) error {
 	vc.mut.Lock()
 	defer vc.mut.Unlock()
 
-	nodeID := id.String()
-	vc.components[nodeID] = id
+	variables := vc.scope.Variables
+	// Build nested maps.
```
thampiotr, line 54: Can you help me understand why do we choose to take on the complexity of the nested maps instead of using a flattened map with a concatenated component id parts, like "foo.bar.baz" instead of `{foo: { bar: baz}}`?

thampiotr (commented):
I have a few questions

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,90 @@
...
 	defer vc.mut.Unlock()
 
-	if value == nil {
-		vc.moduleArguments[key] = nil
-	} else {
-		vc.moduleArguments[key] = value
-	}
+	keyMap := make(map[string]any)
+	keyMap["value"] = value
+	vc.moduleArguments[key] = keyMap
```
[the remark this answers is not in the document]
wildum: because to refer to an argument value in the config file you need to write "argument.LABEL.value". To resolve it, it must have the path ["argument"][LABEL]["value"]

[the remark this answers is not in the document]
thampiotr: Ah gotcha. Yeah now I remember this syntax. Would things break if we made both `foo.bar.value` and `foo.bar` work? It could be UX improvement 🤔

internal/runtime/internal/controller/value_cache.go
wildum: because that's the structure that we need to give to the Alloy syntax to evaluate the expression. It's also the structure that we use for modules to propagate the context. The current implementation builds it on every evaluation whereas this new implementation stores the values directly in it

internal/runtime/internal/controller/value_cache.go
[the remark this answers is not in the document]
wildum: that would be nice but it's tricky to do because you cant have two values for the same key in the map + you would need to handle the weird cases where the user names its argument "value".

internal/runtime/internal/controller/loader.go
```diff
@@ -819,22 +832,30 @@ func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer tr
...
-	ectx := l.cache.BuildContext()
+	ectx := l.cache.GetContext()
 	err := bn.Evaluate(ectx)
 	return l.postEvaluate(logger, bn, err)
 }
 
 // postEvaluate is called after a node has been evaluated. It updates the caches and logs any errors.
 // mut must be held when calling postEvaluate.
+// TODO: Refactor this function to avoid using an error as input solely for shadowing purposes.
+// This design choice limits extensibility and makes it harder to modify or add functionality.
```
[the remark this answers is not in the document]
wildum: After putting more thoughts into it, I think that the shadowing is ok. I thought that it could be avoided or done in another function but it covers the use-case where something could be an error depending on the result of the evaluation (for example in this context: the missing argument from the cache could be an error depending whether the argument is optional or not, so you first need to evaluate before you can test it). This may come in the future in other use-case. I will update the comment to clarify this concept

internal/runtime/internal/controller/value_cache.go
```diff
@@ -1,88 +1,90 @@
...
-	vc.exports[nodeID] = exportsVal
+	variables[id[len(id)-1]] = exportsVal
+	return nil
+}
+
+func (vc *valueCache) GetModuleArgument(key string) (any, bool) {
+	vc.mut.RLock()
+	defer vc.mut.RUnlock()
+	v, exist := vc.moduleArguments[key]
+	return v, exist
```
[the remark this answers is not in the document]
wildum: that does not work in Go, the bool value is only added in an assignment statement or initialization (https://go.dev/ref/spec#Index_expressions). In this case, we want to return both the value (which can be nil) and the boolean to see if it exists in the map

internal/runtime/internal/controller/value_cache.go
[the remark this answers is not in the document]
wildum: done

internal/runtime/internal/controller/loader.go
```diff
@@ -819,22 +832,30 @@ func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer tr
...
-		// Always update the cache both the arguments and exports, since both might
-		// change when a component gets re-evaluated. We also want to cache the arguments and exports in case of an error
-		l.cache.CacheArguments(c.ID(), c.Arguments())
-		l.cache.CacheExports(c.ID(), c.Exports())
+		// Always update the cache both the exports, since that it might change when a component gets re-evaluated.
+		// We also want to cache it in case of an error
+		err2 := l.cache.CacheExports(c.ID(), c.Exports())
+		if err2 != nil {
+			level.Error(logger).Log("msg", "failed to cache exports after evaluation", "err", err2)
+			// Don't mask the previous evaluation error.
```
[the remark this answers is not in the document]
wildum: I think that it might be confusing because these are two different errors. The evaluation error is probably the most important here so it gets returned via the diag to be nicely printed and the other one is just logged as an additional error.

internal/runtime/internal/controller/value_cache.go
```diff
@@ -130,107 +132,98 @@ func (vc *valueCache) ExportChangeIndex() int {
...
-	vc.mut.Lock()
-	defer vc.mut.Unlock()
-
-	for id := range vc.moduleArguments {
-		if _, keep := args[id]; keep {
-			continue
+	// Remove the component exports from the scope.
+	for _, id := range cleanupIds {
+		// Start with the index "0". It refers to the first part of the componentID and it's used for recursion.
+		err := cleanup(vc.scope.Variables, id, 0)
```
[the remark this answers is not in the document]
wildum: I applied the suggestion below, let me know if that's ok or if I should try to simplify further

internal/runtime/internal/controller/value_cache.go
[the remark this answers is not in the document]
thampiotr: ah, funny, they could make it work

internal/runtime/internal/controller/loader.go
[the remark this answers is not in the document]
thampiotr: I think the likelihood of errors in `CachedExports` is low, but if we do have them it would be important to go and fix them as they likely mean there's a proper bug instead of a user error. I'll leave it up to you.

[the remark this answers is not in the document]
wildum: I used the join and handled the cases separately to be more explicit.

thampiotr (approved):
LGTM - merge after or before the release?

wildum:
> merge after or before the release

let's do it after to leave this in dev for longer

Issue #2144, opened 2024-11-22
Panic when accessing remote configuration tab on alloy http interface


racinnedivet-m:
### What's wrong?

We are using Alloy v1.5.0 with clustering enabled, deployed on each node of a K8S cluster using the Alloy Helm chart v0.10.0.

When port-forwarding to one of alloy pods to access the http interface, the "Remote Configuration" tab stays blank, and the alloy pod keeps logging a panic runtime error : `http: panic serving 127.0.0.1:<random-port>: runtime error: invalid memory address or nil pointer dereference`

The home page, "Graph" and "Clustering" tabs are working fine.

### Steps to reproduce

On a K8S alloy cluster : 
1. Port-forward to an alloy pod http port :
    `kubectl port-forward pod/alloy-xxxxx 8080:80`
2. On your web browser, access the "Remote Configuration" tab [http://127.0.0.1:8080/remotecfg](http://127.0.0.1:8080/remotecfg)
3. Observe the error on the pod logs.

### Software version

Grafana Alloy v1.5.0

### Configuration

```text

```

### Logs

```text
2024/11/22 10:06:51 http: panic serving 127.0.0.1:40850: runtime error: invalid memory address or nil pointer dereference
goroutine 13932270 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1903 +0xbe
panic({0x9396c40?, 0x117c1d00?})
	/usr/local/go/src/runtime/panic.go:770 +0x132
github.com/grafana/alloy/internal/web/api.(*AlloyAPI).RegisterRoutes.listComponentsHandler.func4({0xc0952f0, 0xc033e73300}, 0xc033e85600?)
	/src/alloy/internal/web/api/api.go:64 +0x69
net/http.HandlerFunc.ServeHTTP(0xc0951d0?, {0xc0952f0?, 0xc033e73300?}, 0xc01d525188?)
	/usr/local/go/src/net/http/server.go:2171 +0x29
github.com/prometheus/prometheus/util/httputil.CompressionHandler.ServeHTTP({{0xbffb840?, 0xc0000131e8?}}, {0xc0951d0?, 0xc03408cfc0?}, 0xc03428a6c0)
	/go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20240514135907-13889ba362e6/util/httputil/compression.go:91 +0x63
[...]

Issue #2157, opened 2024-11-25
Alloy Service startup should be deterministic & ordered


dehaansa:
### Request

The Services (HTTP service, UI service, remote cfg service, etc) have dependencies, laid out in their DependsOn parameter. However, the scheduler does not enforce any starting order and ensure other services are fully available before starting ones that depend upon them.

This caused a panic https://github.com/grafana/alloy/issues/2144 when the HTTP service tried to install the UI service's endpoints that depend on the remote config service, but the remote config service had not yet executed its Run() function and stored a variable that the UI service depended upon. This was a race condition that would not have been present if the services were started in order, with that order informed by the DependsOn parameter.

This will require changes to the Scheduler to start components in order, and the ServiceNode (and likely other Node) implementations need a Wait Group or chan or mutex provided to be able to inform the scheduler that they are ready for their dependent components to start.

### Use case

This will help prevent future issues, and enforce the DependsOn contract in a way that allows developers to trust the availability of other services.

PR #2160, opened 2024-11-26
fix: race condition UI remotecfg

dehaansa:
#### PR Description
There was a race condition where the http service -> ui service required the remotecfg service to have at least run the first line of its Run() function, but they are all started in parallel in goroutines so there was no guarantee of consistent ordering. See https://github.com/grafana/alloy/issues/2157 for a follow up on that concern, but this PR fixes it by only looking to resolve the remotecfg's Data object during HTTP resolution time, where the service should definitely be available.

Fixes https://github.com/grafana/alloy/issues/2144

#### PR Checklist

- [X] CHANGELOG.md updated


wildum (approved):
lgtm, thanks!

PR #2161, opened 2024-11-26
prometheus.operator.*: allow setting informer_sync_timeout

### LINKED ISSUE #2004: Make the prometheus operator informer sync timeout configurable and increse the default
### Request

The timeout defined here:

https://github.com/grafana/alloy/blob/main/internal/component/prometheus/operator/common/crdmanager.go#L41

should ideally be configurable. A larger default timeout could also make sense, because this happens at component `Run` and could result in a failed component. Sometimes the request takes too long.

### Use case

For large / slow clusters

captncraig:

Fixes #2004

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
+++ b/docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
@@ -34,6 +34,7 @@ Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `forward_to` | `list(MetricsReceiver)` | List of receivers to send scraped metrics to. | | yes
 `namespaces` | `list(string)` | List of namespaces to search for PodMonitor resources. If not specified, all namespaces will be searched. || no
+`informer_sync_timeout` | `duration` | Timeout for initial sync of PodMonitor resources. | `1m` | no
 
 ## Blocks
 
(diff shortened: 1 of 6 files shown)


docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
```diff
@@ -34,6 +34,7 @@ Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `forward_to` | `list(MetricsReceiver)` | List of receivers to send scraped metrics to. | | yes
 `namespaces` | `list(string)` | List of namespaces to search for PodMonitor resources. If not specified, all namespaces will be searched. || no
+`informer_sync_timeout` | `duration` | Timeout for initial sync of PodMonitor resources. | `1m` | no
```
ptodev, line 37: If it's for the initial sync, should it be called `initial_sync_timeout`? Is "informer" a k8s term?

ptodev: I think it's best to leave it as "informer", because the error also mentions the word "informer":

> timeout exceeded while configuring informers

By naming it `informer_sync_timeout`, there's a bigger chance that the users will realise that they need to tweak `informer_sync_timeout`.

PR #2162, opened 2024-11-26
feat(helm): only add pod annotation with config checksum if config reloader is disabled

ebuildy:
#### PR Description

So the replicaSet (deployment, daemonset etc...) is automatically  restarted after config changes.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


dehaansa:
I think this is a really valuable optimization to not cause unnecessary reloads! Do you have capacity to rebase it from main & sign the CLA?

korniltsev (commented):
I like the check for config loader more than my PR, however i think we can not do this at this point as this will be a breaking change, removing pod annotations and restarting deployment

korniltsev:
Just linking my alternative here for consideration https://github.com/grafana/alloy/pull/3395

thampiotr:
@korniltsev I think the expected behaviour here was that there will be no restart if reloader is enabled, so this is a bugfix IMO and not a breaking change... adding an additional config option is a way to avoid this, but I'm not sure if it's worth it. We can try to call this out in the changelog.

ebuildy:
well, I dont get the point --> if you use the config watcher (the sidecar container defined here https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/containers/_watch.yaml#L3) , you dont want to restart your pod isnit ? (because the watcher already reload the config)

korniltsev:
Any progress on this?

jannis-a:
Could we progress on this PR please? The current behaviour is not optimal and causes a lot of waiting time for our team due to the introduced bug 🙏

kalleep:
Hey, we agree that we should merge this. Could you please re-generate helm test by running `make generate-helm-test` and add a change-log entry?

kalleep:
@ebuildy Thanks for your contribution

jannis-a:
@kalleep when will this get released? :)

kalleep:
@jannis-a I will do a release for our helm chart first thing on Monday

PR #2164, opened 2024-11-26
fix: fully prevent panic in remotecfg ui

dehaansa:
#### PR Description
While https://github.com/grafana/alloy/pull/2160 removed the race condition from the Remote Config UI, the possibility was still present to get a panic if you attempted to load the UI immediately on startup. This prevents the panic completely.

### DIFF
--- a/internal/web/api/api.go
+++ b/internal/web/api/api.go
@@ -113,7 +113,13 @@ func getComponentHandlerRemoteCfg(host service.Host) http.HandlerFunc {
 			return
 		}
 
-		getComponentHandlerInternal(svc.Data().(remotecfg.Data).Host, w, r)
+		data := svc.Data().(remotecfg.Data)
+		if data.Host == nil {
+			http.Error(w, "remote config service startup in progress", http.StatusInternalServerError)
+			return
+		}
+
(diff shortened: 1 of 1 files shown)


erikbaranowski (approved):

internal/web/api/api.go
```diff
@@ -66,7 +66,12 @@ func listComponentsHandlerRemoteCfg(host service.Host) http.HandlerFunc {
 			return
 		}
 
-		listComponentsHandlerInternal(svc.Data().(remotecfg.Data).Host, w, r)
+		data := svc.Data().(remotecfg.Data)
+		if data.Host == nil {
+			http.Error(w, "remote config service startup in progress", http.StatusInternalServerError)
+			return
+		}
+		listComponentsHandlerInternal(data.Host, w, r)
```
wildum, line 74: should we have the same check in getComponentHandlerRemoteCfg? It may no be possible to trigger via the UI but could be in theory by sending http requests directly

dehaansa: Definitely, great catch.

wildum (approved):

PR #2175, opened 2024-11-27
Fix prometheus.write.queue ttl

mattdurham:
Update to the newest walqueue that resolves ttl comparisons.


mattdurham:
Previously I was sometimes comparing milliseconds to seconds in walqueue.

dehaansa (approved):

PR #2178, opened 2024-11-27
chore: Update uses of prometheus registration to be less likely to panic

dehaansa:
#### PR Description
Inspired by https://github.com/grafana/alloy/pull/2177 - I don't see a reason that the possibility of a metric being already registered should cause Alloy to panic.

None


wildum (approved):
lgtm

ptodev:
It's worth noting that the `AlreadyRegisteredError` feature is more of an anti-pattern, and should only be used as a last resort. It's not meant to be used regularly for every registration. If there are duplicate registrations, there is likely to be a bug and avoiding the error could just further mask the bug.

To avoid the panic, it's possible to use `Register` instead of `MustRegister` and then the error can be propagated.

PR #2185, opened 2024-11-28
Implement tracing for pyroscope components

madaraszg-tulip:
#### PR Description

Implement tracing for the `pyroscope.scrape`, `pyroscope.receive_http` and `pyroscope.write` components.

Spans are created for http server and http client calls in these components. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/scrape/scrape_loop.go
+++ b/internal/component/pyroscope/scrape/scrape_loop.go
@@ -48,9 +48,12 @@ func newScrapePool(hco []commonconfig.HTTPClientOption, cfg Arguments, appendabl
 		return nil, err
 	}
 
-	scrapeClient.Transport = otelhttp.NewTransport(scrapeClient.Transport, otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
-		return otelhttptrace.NewClientTrace(ctx, otelhttptrace.WithoutSubSpans())
-	}),
+	scrapeClient.Transport = otelhttp.NewTransport(
+		scrapeClient.Transport,
+		otelhttp.WithTracerProvider(tracer),
+		otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
+			return otelhttptrace.NewClientTrace(ctx, otelhttptrace.WithoutSubSpans())
+		}),
 	)
 
 	return &scrapePool{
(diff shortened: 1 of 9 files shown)


internal/component/pyroscope/scrape/scrape_loop.go
```diff
@@ -228,6 +241,9 @@ func (t *scrapeLoop) scrape() {
 	)
 	defer cancel()
 
+	scrapeCtx, span := t.tracer.Tracer("").Start(scrapeCtx, "scrape")
```
madaraszg-tulip, line 244: Naming spans is the most difficult thing. This is the only span I create explicitly in this PR, so there's not much naming convention to do, but it would be good to have one.

madaraszg-tulip:
New spans: 

image
image

madaraszg-tulip:
I'm seeing segfaults during reload related to tracing, will need to investigate

ptodev (commented):
Hi, thank you for the PR! I do wonder to what extent traces will be useful here. Tracing is mostly handy when there is a chain of network calls - one service calling another, calling another, calling another, etc. But with the profiling calls there are only two services involved in every call. The only use case I can think of is spanmetrics generation from the traces?

madaraszg-tulip:
> Hi, thank you for the PR! I do wonder to what extent traces will be useful here. Tracing is mostly handy when there is a chain of network calls - one service calling another, calling another, calling another, etc. But with the profiling calls there are only two services involved in every call. The only use case I can think of is spanmetrics generation from the traces?

I implemented this to debug the root cause of the issue that eventually resulted in https://github.com/grafana/alloy/pull/2201. It was very useful to have traces to see what is actually happening. Btw there are more services involved - the client, alloy, traefik, pyroscope-distributor, pyroscope-ingester.

marcsanmi:
Hi @madaraszg-tulip,

Thank you for your contribution and for taking the time to enhance observability in Alloy. 

My two cents: while the instrumentation could be valuable, I believe we should align with Alloy's observability choices for now. If they decide to adopt `otelhttp` in the future, we'd be happy to revisit this kind of instrumentation. However, I'll leave the final decision to someone from Alloy.

Thanks again for your contribution!

madaraszg-tulip:
Closing for now, might revisit in the future

PR #1722, opened 2024-09-20
Stop kubernetes log tailer gracefully

ptodev:
#### PR Description

As mentioned in #281, users get this warning with `loki.source.kubernetes`:
```
level=warn msg="tailer stopped; will retry" target=monitoring/pod-id:grafana component=loki.source.kubernetes.pods err="http2: response body closed"
```

It looks like `reader.ReadString('\n')` gets called after `stream.Close()` has already been called. 

Me and @wildum were looking at this today, and we suspect that decoupling the lifetime of the string from the timeout context + moving `stream.Close()` to the end of `tail()` could remove the warning message.

The warning message seems to be a cosmetic problem. No logs would be lost, but it's confusing since users don't expect to see a warning for something that works as expected. Making the shutdown more graceful will make the component logs less confusing.

#### Which issue(s) this PR fixes

Fixes #281

#### Notes to the Reviewer

I am not yet sure how to test this. When we test it, I'll take the PR out of the draft stage.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #2187, opened 2024-11-28
`loki.source.podlogs`: Fix issue which disables clustering unintentionally.

ptodev:
#### PR Description

This bug cases multiple logs to be sent. Every Alloy instance in the cluster would send all of the logs which `loki.source.podlogs` discovered.

#### Notes to the Reviewer

It'd be great to write unit tests, but we don't have any tests for `loki.source.podlogs` at the moment other than trivial tests for configs. A test would need to start a fake kubernetes cluster, and I'm not yet sure how to do it. There's [another PR](https://github.com/grafana/alloy/pull/1722) where I've been trying to do a similar thing, and got a bit stuck.

I suspect that `loki.source.podlogs` may also be affected by the same issue as #1716, so I will also investigate this separately and will open another PR for it.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ Main (unreleased)
 
 - Fix a race condition where the ui service was dependent on starting after the remotecfg service, which is not guaranteed. (@dehaansa & @erikbaranowski)
 
+- `loki.source.podlogs`: Fixed a bug which prevented clustering from working and caused duplicate logs to be sent. (@ptodev)
+
 ### Other changes
 
 - Change the stability of the `livedebugging` feature from "experimental" to "generally available". (@wildum)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@ Main (unreleased)
 
 - Fix a race condition where the ui service was dependent on starting after the remotecfg service, which is not guaranteed. (@dehaansa & @erikbaranowski)
 
-- `loki.source.podlogs`: Fixed a bug which prevented clustering from working and caused duplicate logs to be sent. (@ptodev)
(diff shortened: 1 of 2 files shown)


cyriltovena (approved):
LGTM

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ Main (unreleased)
 
 - Fix a race condition where the ui service was dependent on starting after the remotecfg service, which is not guaranteed. (@dehaansa & @erikbaranowski)
 
+- `loki.source.podlogs`: Fixed a bug which prevented clustering from working and caused duplicate logs to be sent. (@ptodev)
```
wildum, line 49: This was not working only in the case where the clustering config would be updated at runtime and without changing the selectors, right? On restart of Alloy it would work, no?

wildum (approved):
lgtm, just a question

CHANGELOG.md
ptodev: It depends. If you're using default configs like this, then clustering will never work:
```
loki.source.podlogs "podlogs_objects" {
  client {
    kubeconfig_file = "/Users/paulintodev/.kube/config"
  }
  clustering {
    enabled = true
  }
  forward_to = [loki.echo.example.receiver]
}
```

That's because on startup, the "new" and the "old" configs are both the same.

ptodev: I updated the readme to mention this more clearly :) I haven't tested what happens when a `selector` or `namespace_selector` are specified, but I'm pretty sure the code will work ok from what I can see

PR #2190, opened 2024-11-28
loki.source.podlogs: For clustering only take into account some labels

ptodev:
#### PR Description

When clustering, `loki.source.podlogs` currently only takes into account public labels such as `job` and `instance`. I think it'd be safer to take the same approach as `loki.source.kubernetes` took in #1716 and only take into account some labels.

I have not seen this in the wild and I'm not sure if there could be duplicate logs as in #1716, so I'm just labeling this as an enhancement in the changelog rather than a bugfix.

#### Notes to the Reviewer

Unfortunately, it's hard to write a unit test for this since we'd have to fake k8s. I tested this locally on a local k8s cluster.

I don't think this should be classified as a breaking change, because the docs already say that certain labels are required for `loki.source.podlogs` to work:

> In addition to the meta labels, the following labels are exposed to tell `loki.source.podlogs` which container to tail:
> * `__pod_namespace__`: The namespace of the Pod.
> * `__pod_name__`: The name of the Pod.
> * `__pod_container_name__`: The container name within the Pod.
> * `__pod_uid__`: The UID of the Pod.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.podlogs.md
+++ b/docs/sources/reference/components/loki/loki.source.podlogs.md
@@ -231,7 +231,7 @@ cluster to distribute the load of log collection between all cluster nodes.
 If {{< param "PRODUCT_NAME" >}} is _not_ running in clustered mode, then the block is a no-op and
 `loki.source.podlogs` collects logs based on every PodLogs resource discovered.
 
-Clustering only looks at the following labels for determining the shard key:
+Clustering looks only at the following labels for determining the shard key:
 
 * `__pod_namespace__`
 * `__pod_name__`
--- a/docs/sources/reference/components/loki/loki.source.podlogs.md
+++ b/docs/sources/reference/components/loki/loki.source.podlogs.md
@@ -231,6 +231,21 @@ cluster to distribute the load of log collection between all cluster nodes.
 If {{< param "PRODUCT_NAME" >}} is _not_ running in clustered mode, then the block is a no-op and
 `loki.source.podlogs` collects logs based on every PodLogs resource discovered.
 
+Clustering only looks at the following labels for determining the shard key:
+
+* `__pod_namespace__`
+* `__pod_name__`
+* `__pod_container_name__`
+* `__pod_uid__`
+* `__meta_kubernetes_namespace`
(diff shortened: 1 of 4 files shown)


docs/sources/reference/components/loki/loki.source.podlogs.md
```diff
@@ -231,6 +231,21 @@ cluster to distribute the load of log collection between all cluster nodes.
 If {{< param "PRODUCT_NAME" >}} is _not_ running in clustered mode, then the block is a no-op and
 `loki.source.podlogs` collects logs based on every PodLogs resource discovered.
 
+Clustering only looks at the following labels for determining the shard key:
```
clayton-cornell, line 234:
```suggestion
Clustering looks at the following labels for determining the shard key:
```

We can drop the only. If you think it is still needed for clarity, then switch its location to read `Clustering looks only at the...`

petewall (approved):

ptodev:
The main issue I'm trying to address is that if a label such as `__meta_kubernetes_pod_phase` changes, I'd like the same Alloy instance to still be responsible for that container instance. If the number of Alloy instances in the cluster stays constant, then ideally a container should be tailed by the same Alloy instance throughout its lifetime. That Alloy instance will also track it in its own position file. If a container stops being tailed by one Alloy instance and another Alloy takes over, then that new Alloy instance will not start from the log position that the 1st Alloy left. Instead, I think it'll start tailing from the beginning. I suppose that's the sort of situation @mattdurham prevented in #1716?

Note that if the number of Alloy instances in a cluster changes, some of the targets will move from one Alloy instance to another, which again could cause the new Alloy to start tailing them from the beginning. We should raise an issue about this, assuming that this is indeed a problem.

The reason why I talk about containers and not pods is that I believe the "podlogs" component is really about container logs, and different Alloy instances in the cluster could be tailing the logs for different containers on the same pod.

clayton-cornell (approved):
Docs look OK

mattdurham (approved):
lgtm

PR #2197, opened 2024-11-29
Add a `/-/healthy` endpoint for monitoring component health

### LINKED ISSUE #2061: Expose Alloy overall component health via HTTP endpoint
### Request

Expose via an HTTP endpoint the overall health of all running components: if at least one component is not healthy - return an error.

For example, if I request
```
GET /-/ready?strict=true
```
And there is one component failing, we should get back:
```
503: Component 'foo.bar.baz' is not healthy: 'error message'
```

### Use case

Sometimes users may want to set up a liveness or readiness probe on Alloy in k8s that does not only check if Alloy runtime is started, but also checks that all the components are healthy.

Currently Alloy's `/-/ready` endpoint will return 200 even if there are components failing to start. This can be useful when we have multiple pipelines and want the healthy ones to continue running. But some users may prefer a "fail hard" behaviour.

ptodev:
#### PR Description

Adding a `/-/healthy` endpoint which returns an error if at least one component is not healthy. The `/-/ready` endpoint doesn't check for component health, so this could be useful for Kubernetes liveliness probes.

Fixes #2061

#### Notes to the Reviewer

[Prometheus](https://prometheus.io/docs/prometheus/latest/management_api/) also has a `/-/healthy` endpoint, and so does [Grafana Agent Static mode](https://grafana.com/docs/agent/latest/static/api/#ready--health-api). However, both of these were not very useful because they always return HTTP 200. There is an [open issue](https://github.com/prometheus/prometheus/issues/3650) in the Prometheus repo to make it more useful.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/http/_index.md
+++ b/docs/sources/reference/http/_index.md
@@ -1,7 +1,7 @@
 ---
 canonical: https://grafana.com/docs/alloy/latest/reference/http/
 description: Learn about HTTP endpoints exposed by Grafana Alloy
-title: Grafana Alloy HTTP endpoints
+title: The Grafana Alloy HTTP endpoints
 menuTitle: HTTP endpoints
 weight: 700
 ---
@@ -44,12 +44,12 @@ unhealthy components: math.add
 ```
 
 {{< admonition type="note" >}}
-The `/-/healthy` endpoint is not suitable for a [Kubernetes liveness probe][k8s-liveness].
+The `/-/healthy` endpoint isn't suitable for a [Kubernetes liveness probe][k8s-liveness].
 
-An Alloy instance which reports as unhealthy should not necessarily be restarted.
-For example, a component may be unhealthy due to invalid configuration or due to an unavailable external resource.
+An {{< param "PRODUCT_NAME" >}} instance that reports as unhealthy should not necessarily be restarted.
+For example, a component may be unhealthy due to an invalid configuration or an unavailable external resource.
 In this case, restarting {{< param "PRODUCT_NAME" >}} would not fix the problem.
-A restart may even make it worse, because it would could stop the flow of telemetry in healthy pipelines.
+A restart may make it worse, because it would could stop the flow of telemetry in healthy pipelines.
 
 [k8s-liveness]: https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/
 {{< /admonition >}}


docs/sources/reference/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
 canonical: https://grafana.com/docs/alloy/latest/reference/
-description: The reference-level documentaiton for Grafana Aloy
```
thampiotr, line 3: On the very top level 😂

docs/sources/reference/http/_index.md
thampiotr, line 1: Love it that we have this file now!

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+
+### /-/healthy
+
+If all components are healthy, HTTP 200 and the message "Alloy is healthy." will be returned.
+Otherwise, {{% param "FULL_PRODUCT_NAME" %}} will return HTTP 500 and an error message.
+You can also monitor component health through the [UI](../../troubleshoot/debug#alloy-ui).
+
+```
+$ curl localhost:12345/-/healthy
+Alloy is healthy.
```
thampiotr, line 37: Nit: should we say "All Alloy components are healthy."?

internal/service/http/http.go
```diff
@@ -188,6 +188,32 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		otelmux.WithTracerProvider(s.tracer),
 	))
 
+	// The implementation for "/-/healthy" is inspired by
+	// the "/components" web API endpoint in /internal/web/api/api.go
+	r.HandleFunc("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
+		components, err := host.ListComponents("", component.InfoOptions{
```
thampiotr, line 194: Does it check for components running in modules? or do we rely on the custom component instance's health that aggregates the health of all its internal components?

internal/service/http/http_test.go
```diff
@@ -43,6 +44,26 @@ func TestHTTP(t *testing.T) {
 		require.NoError(t, err)
 		defer resp.Body.Close()
 
+		buf, err := io.ReadAll(resp.Body)
+		require.Equal(t, "Alloy is ready.\n", string(buf))
+
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+	})
+
+	util.Eventually(t, func(t require.TestingT) {
```
thampiotr, line 53: Cool! Do you think it's worth hitting this endpoint in the integration tests?

thampiotr (commented):
Looks nice, thanks! Just minor comments.

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+```
+$ curl localhost:12345/-/healthy
+unhealthy components: math.add
+```
+
+### /-/reload
+
+Reloads the {{% param "FULL_PRODUCT_NAME" %}} configuration file. Returns HTTP 400 and an error message if an issue with the reload was encountered.
```
clayton-cornell, line 47:
```suggestion
The `/-/reload` endpoint reloads the {{< param "PRODUCT_NAME" >}} configuration file.
If the configuration file can't be reloaded, the `/-/reload` endpoint returns `HTTP 400 Bad Request` and an error message.
```

Is it the same error message or a different error message (depending on the reason the reload failed). Can we be more explicit here?

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
```
clayton-cornell, line 49:
````suggestion
```shell
````

File: docs/sources/reference/http/_index.md [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+$ curl localhost:12345/-/reload
+config reloaded
+```
+
+```
```
clayton-cornell, line 54:
````suggestion
```shell
````

File: docs/sources/reference/http/_index.md [hunk 8]
```diff
@@ -0,0 +1,65 @@
...
+$ curl localhost:12345/-/reload
+error during the initial load: /Users/user1/Desktop/git.alloy:13:1: Failed to build component: loading custom component controller: custom component config not found in the registry, namespace: "math", componentName: "add"
+```
+
+### /-/support
+
+Generates a [support bundle](../../troubleshoot/support_bundle).
```
clayton-cornell, line 61:
```suggestion
The `/-/support` endpoint returns a [support bundle](../../troubleshoot/support_bundle) that contains information about your {{< param "PRODUCT_NAME" >}} instance. You can use this information as a baseline when debugging an issue.
```

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+
+### /debug/pprof
+
+Generates a [profile](../../troubleshoot/profile).
```
clayton-cornell, line 65:
```suggestion
The `debug/pprof` endpoint returns a pprof Go [profile](../../troubleshoot/profile) that you can use to visualize and analyse profiling data.

```

clayton-cornell (commented):
I added a few suggestions here to expand an clarify some doc things. :-)

rfratto:
Hi, I wanted to share some concerns with using component health for liveness probes.

Kubernetes [intends](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) for liveness probes to be an indication of when the pod is "stuck" and needs to be killed. Kubernetes will eventually kill the pod if the liveness probe fails enough times. 

With this definition of liveness, component health is not a good indicator. Components like `remote.s3` can report themselves as unhealthy if an object in S3 gets deleted, and custom components can fail if you give them an invalid configuration (such as someone rolls out a bad config from fleet management).  

Killing the pod when components are unhealthy will not only not fix the problem, but since Alloy requires all components to be healthy on startup, it will also completely halt all telemetry collection until the problem is manually fixed.

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+
+A {{% param "FULL_PRODUCT_NAME" %}} instance is "ready" once it has loaded its initial configuration.
+If it is ready, HTTP 200 and the message `Alloy is ready.` are returned.
+Otherwise, HTTP 503 and the message `Alloy is not ready.` are returned.
+
+### /-/healthy
+
+If all components are healthy, HTTP 200 and the message "Alloy is healthy." will be returned.
+Otherwise, {{% param "FULL_PRODUCT_NAME" %}} will return HTTP 500 and an error message.
+You can also monitor component health through the [UI](../../troubleshoot/debug#alloy-ui).
```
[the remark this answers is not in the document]
ptodev: The error message varies, so I'd rather not be too specific.

thampiotr:
> Hi, I wanted to share some concerns with using component health for liveness probes.

Oh yeah, I 100% agree and I would not recommend using it as liveness probe. But people still wanted it. I think adding this endpoint is harmless, because it can be used for other health checking purposes or diagnostics, not just as a liveness probe. We do know people who wanted to use it as liveness probe, but this would be doing it against our advice. They may have some special requirements though and it may work okay for them.

thampiotr (approved):

ptodev:
Hi @rfratto, thank you for opining! I agree that in the component health is not a good indicator for whether Alloy as a whole should be restarted. 

Apparently, the issue which originally prompted this feature request is an informer timeout like the one referenced in #2161. 
 @thampiotr - if that timeout does happen, maybe the real issue is that the component should keep retrying to sync the informers rather than giving up? Judging by the code, the podlogs component will only sync the informers when the config is updated? 

I do think there is a benefit in the `/healhy` endpoint as a whole, although I do agree that we shouldn't recommend that users use it as a liveliness probe. It could be useful in situations where users update the Alloy config and then want to double check that it works as expected. They might do rolling deployments - e.g. 1% of Alloy instances first, and then if they are healthy, the deployment can continue to more instances. I mistakenly thought that liveliness probes in k8s can do similar things, but apparently that's not the case.

docs/sources/reference/http/_index.md
ptodev: Yes, I think this is more clear. I changed it just now, thanks!

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
...
+canonical: https://grafana.com/docs/alloy/latest/reference/http/
+description: Learn about HTTP endpoints exposed by Grafana Alloy
+title: HTTP endpoints
+weight: 700
+---
+
+# The {{% param "FULL_PRODUCT_NAME" %}} HTTP endpoints
+
+There are HTTP endpoints which are enabled by default on every instance of {{% param "FULL_PRODUCT_NAME" %}}, 
+regardless which components are configured.
```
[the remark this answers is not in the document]
clayton-cornell: Maybe we could drop the always (which implies the availability never changes) and and just say `available by default` ?  I've updated the suggestion.  Does this work?

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,65 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/http/
+description: Learn about HTTP endpoints exposed by Grafana Alloy
+title: HTTP endpoints
+weight: 700
+---
+
+# The {{% param "FULL_PRODUCT_NAME" %}} HTTP endpoints
```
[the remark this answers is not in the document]
clayton-cornell: We still need to match up the H1 and the `title:` metadata. So we either drop `The` from the H1 or add it to the `title:` metadata.

clayton-cornell (approved):
Minor tweaks to the docs and all good

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,78 @@
...
+{{< admonition type="note" >}}
+The `/-/healthy` endpoint is not suitable for a [Kubernetes liveness probe][k8s-liveness].
+
+An Alloy instance which reports as unhealthy should not necessarily be restarted.
+For example, a component may be unhealthy due to invalid configuration or due to an unavailable external resource.
+In this case, restarting {{< param "PRODUCT_NAME" >}} would not fix the problem.
+A restart may even make it worse, because it would could stop the flow of telemetry in healthy pipelines.
+
+[k8s-liveness]: https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/
+{{< /admonition >}}
```
ptodev, line 55: @clayton-cornell are you happy with this note?
Also, thanks to @rfratto for emphasising this!

clayton-cornell:
```suggestion
{{< admonition type="note" >}}
The `/-/healthy` endpoint isn't suitable for a [Kubernetes liveness probe][k8s-liveness].

An {{< param "PRODUCT_NAME" >}} instance that reports as unhealthy should not necessarily be restarted.
For example, a component may be unhealthy due to an invalid configuration or an unavailable external resource.
In this case, restarting {{< param "PRODUCT_NAME" >}} would not fix the problem.
A restart may make it worse, because it would could stop the flow of telemetry in healthy pipelines.

[k8s-liveness]: https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/
{{< /admonition >}}
```

Some minor tweaking and I think it's OK. The point is clear.

docs/sources/reference/http/_index.md
```diff
@@ -0,0 +1,78 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/http/
+description: Learn about HTTP endpoints exposed by Grafana Alloy
+title: Grafana Alloy HTTP endpoints
```
clayton-cornell, line 4:
```suggestion
title: The Grafana Alloy HTTP endpoints
```

We need the `The` to align with the H1 title.  Or we drop the `The` off the H1

PR #2201, opened 2024-12-02
Fix pyroscope.write issues with pyroscope.receive_http

madaraszg-tulip:
#### PR Description

The nodejs Pyroscope SDK sends profiles with a `Connection: close` header. This header was copied to the upstream request, causing connection churn towards Pyroscope, which can be quite bad on the CPU when using TLS. Do not copy the `Connection` header from the incoming request to fix this issue.

Additionally, `pyroscope.write` had a single `http.Client` used for forwarding data from `pyroscope.receive_http`, which may not work if multiple endpoints are configured with different options. To fix this, store a `http.Client` for each endpoint.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,12 @@ Main (unreleased)
   - Add perf_schema quantile columns to collector
 
 ### Bugfixes
+- Fixed an issue in the `pyroscope.write` component to prevent TLS connection churn to Pyroscope when the `pyroscope.receive_http` clients don't request keepalive (@madaraszg-tulip)
+
+- Fixed an issue in the `pyroscope.write` component with multiple endpoints not working correctly for forwarding profiles from `pyroscope.receive_http` (@madaraszg-tulip)
+
 - Fixed an issue in the `pyroscope.write` component to allow slashes in application names in the same way it is done in the Pyroscope push API (@marcsanmi)
+
 - Fixed an issue in the `prometheus.exporter.postgres` component that would leak goroutines when the target was not reachable (@dehaansa)
 
 - Fixed an issue in the `otelcol.exporter.prometheus` component that would set series value incorrectly for stale metrics (@YusifAghalar)
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -38,6 +38,17 @@ var (
 		return Arguments{}
 	}
 	_ component.Component = (*Component)(nil)
+
+	// List of headers to ignore when copying headers from client to server connection
+	// https://datatracker.ietf.org/doc/html/rfc9113#name-connection-specific-header-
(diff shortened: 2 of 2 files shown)


marcsanmi:
Thanks for taking care of this! 👏  

From HTTP/2 specification ([RFC 9113]:(https://datatracker.ietf.org/doc/html/rfc9113#name-headers))
> HTTP/2 does not use the Connection header field ([Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]) to indicate connection-specific header fields; in this protocol, connection-specific metadata is conveyed by other means. An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields. This includes the Connection header field and those listed as having connection-specific semantics in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)] (that is, Proxy-Connection, Keep-Alive, Transfer-Encoding, and Upgrade).

👆 Could we also exclude the [above mentioned](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) hop-by-hop headers?

marcsanmi (commented):
Thanks for taking care of this! 👏

From HTTP/2 specification ([RFC 9113]:(https://datatracker.ietf.org/doc/html/rfc9113#name-headers))

> HTTP/2 does not use the Connection header field ([Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]) to indicate connection-specific header fields; in this protocol, connection-specific metadata is conveyed by other means. An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields. This includes the Connection header field and those listed as having connection-specific semantics in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)] (that is, Proxy-Connection, Keep-Alive, Transfer-Encoding, and Upgrade).

👆 Could we also exclude the [above mentioned](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) hop-by-hop headers?

madaraszg-tulip:
> 👆 Could we also exclude the [above mentioned](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) hop-by-hop headers?

Added a map to list all ignored headers.

korniltsev (approved):
lgtm, thank you!

marcsanmi (approved):
LGTM, thank you for all the work!

PR #2202, opened 2024-12-02
Retry running a CRD Manager if it fails

ptodev:
#### PR Description

Setting up a k8s informer may time out (see #2161). In that case, we'd only retry setting up an informer the next time Alloy's config changes. We should retry sooner than that, since the config may never change. 

#### Notes to the Reviewer

A few reasons why the PR is in draft stage:
* No unit tests have been added yet.
* I'm not sure what is a good retry value. I suppose it should be something between 30 seconds and 60 seconds.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/prometheus/operator/common/component.go
+++ b/internal/component/prometheus/operator/common/component.go
@@ -104,11 +104,22 @@ func (c *Component) Run(ctx context.Context) error {
 			innerCtx, cancel = context.WithCancel(ctx)
 			wg.Add(1)
 			go func() {
-				if err := manager.Run(innerCtx); err != nil {
-					level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
-					errChan <- err
+				defer wg.Done()
+				retryInterval := 0 * time.Second
+				for {
+					select {
+					case <-innerCtx.Done():
+						//TODO: Log that the manager is exiting?
+						return
+					case <-time.After(retryInterval):
+						if err := manager.Run(innerCtx); err != nil {
+							level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
+							errChan <- err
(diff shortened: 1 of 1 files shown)


internal/component/prometheus/operator/common/component.go
```diff
@@ -104,11 +104,22 @@ func (c *Component) Run(ctx context.Context) error {
...
+					select {
+					case <-innerCtx.Done():
+						//TODO: Log that the manager is exiting?
+						return
+					case <-time.After(retryInterval):
+						if err := manager.Run(innerCtx); err != nil {
+							level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
+							errChan <- err
+							//TODO: What is a good default value for this?
+							retryInterval = 30 * time.Second
```
thampiotr, line 119: Let's use backoff to avoid thundering herd problems. You can use a similar approach to this: https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/loader.go#L732

internal/component/prometheus/operator/common/component.go
```diff
@@ -104,11 +104,22 @@ func (c *Component) Run(ctx context.Context) error {
...
 			go func() {
-				if err := manager.Run(innerCtx); err != nil {
-					level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
-					errChan <- err
+				defer wg.Done()
+				retryInterval := 0 * time.Second
+				for {
+					select {
+					case <-innerCtx.Done():
+						//TODO: Log that the manager is exiting?
```
thampiotr, line 112: I'd say let's add a debug level just in case.

PR #2213, opened 2024-12-03
fix: Update documentation for using filename in otelcol.auth.bearer component

### LINKED ISSUE #1988: Support otel-collector bearer token filename
### Request

The otel-collector bearertokenauthextension (https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/bearertokenauthextension) supports specifying either a hard-coded `token` or a `filename` to read short-lived tokens from.  

The Alloy `otelcol.auth.bearer` component currently only supports the hard-coded `token`, making it difficult to use short-lived bearer tokens.

The request is to add support for specifying a `filename` to read short-lived tokens from, maybe in the following config form:
```
otelcol.auth.bearer "LABEL" {
  filename = "/path/to/file"
}
```

For comparison, both `loki.write` and `prometheus.remote_write` already support a `bearer_token_file` parameter in their endpoint blocks, making it very easy to use short-lived tokens there.

### Use case

Need a way to specify the use of a short-lived bearer token for use in otelcol exporters.

ravishankar15:
#1988 

#### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
+++ b/docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
@@ -38,7 +38,7 @@ Name     | Type     | Description                                      | Default
 `scheme` | `string` | Authentication scheme name.                      | "Bearer" | no
 
 When sending the token, the value of `scheme` is prepended to the `token` value.
-The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC).
+The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC). If we are using file for storing the token we can refer the value of `token` using `local.file`
 
 ## Blocks
 
@@ -97,7 +97,7 @@ otelcol.auth.bearer "creds" {
 
 The example below configures [otelcol.exporter.otlphttp][] to use a bearer token authentication.
 
-If we assume that the value of the `API_KEY` environment variable is `SECRET_API_KEY`, then 
+If we assume that the value of the `API_KEY` environment variable is `SECRET_API_KEY`, then
 the `Authorization` HTTP header is set to `MyScheme SECRET_API_KEY`.
 
 ```alloy
--- a/docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
+++ b/docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
@@ -38,7 +38,11 @@ Name     | Type     | Description                                      | Default
 `scheme` | `string` | Authentication scheme name.                      | "Bearer" | no
 
 When sending the token, the value of `scheme` is prepended to the `token` value.
-The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC). If we are using file for storing the token we can refer the value of `token` using `local.file`
+The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC).
+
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
```diff
@@ -34,10 +34,13 @@ otelcol.auth.bearer "LABEL" {
...
 ---------|----------|--------------------------------------------------|----------|---------
-`token`  | `secret` | Bearer token to use for authenticating requests. |          | yes
+`filename`| `string` | File name containing the authentication token    |          | no
+`token`  | `secret` | Bearer token to use for authenticating requests. |          | no
 `scheme` | `string` | Authentication scheme name.                      | "Bearer" | no
 
-When sending the token, the value of `scheme` is prepended to the `token` value.
+The auth is taken from `token` or `filename` and one of these two value is required. The `filename` is watched for any new changes.
+
+When sending the auth, the value of `scheme` is prepended to the auth value.
```
clayton-cornell, line 43:
```suggestion
The authentication is taken from `token` or `filename` and one of these two values is required.
The `filename` is watched for any new changes.

When sending the authentication, the value of `scheme` is prepended to the authentication value.
```

We should spell out authentication... unless it's referring to a specific variable/label... then it should be wrapped in backticks.

mattdurham:
We generally do not use file based secrets anymore. Can what you want be achieved with `local.file`?

ravishankar15:
Hi @mattdurham Can you check the thread from the issue i think this was discussed a bit there. https://github.com/grafana/alloy/issues/1988

clayton-cornell:
@wildum @mattdurham I'm a bit confused here. Does this doc change actually cover what is needed? Do I understand this correctly that if the user is storing a token in a file somewhere, they can... retrieve the token somehow using `local.file`? Looking at this comment: https://github.com/grafana/alloy/issues/1988#issuecomment-2523338506 I'm not sure the proposed doc change covers it all... but I might be missing a detail here too.

wildum:
@clayton-cornell that's correct, the upstream component has a "filename" attribute that can be used to retrieve a secret but we decided not to have it in Alloy and instead tell the users to use the local.file component to retrieve the secret from the file and send it to this component via the token attribute.

docs/sources/reference/components/otelcol/otelcol.auth.bearer.md
```diff
@@ -38,7 +38,7 @@ Name     | Type     | Description                                      | Default
 `scheme` | `string` | Authentication scheme name.                      | "Bearer" | no
 
 When sending the token, the value of `scheme` is prepended to the `token` value.
-The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC).
+The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC). If we are using file for storing the token we can refer the value of `token` using `local.file`
```
clayton-cornell, line 41:
```suggestion
The string is then sent out as either a header (in case of HTTP) or as metadata (in case of gRPC).

If you use a file to store the token, you can use `[local.file`][local.file] to retrieve the `token` value from the file.`

[local.file]: ../../local/local.file/
```

@wildum Does this suggestion look right then?

wildum (approved):

PR #2242, opened 2024-12-07
Configure metrics exporter for loadbalancing when possible

madaraszg-tulip:
#### PR Description

Metrics support was turned off in `otelcol.exporter.loadbalancing` for two reasons: 
- metrics support is considered `development` upstream
- the metrics exporter crashed when configured with the default `routing_key = traceID`

Solve both issues by supporting metrics only if the specified `routing_key` supports it and stability level is set to `experimental`

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -65,7 +65,7 @@ The `routing_key` attribute determines how to route signals across endpoints. It
 - `"service"`: spans, logs, and metrics with the same `service.name` will be exported to the same backend.
 This is useful when using processors like the span metrics, so all spans for each service are sent to consistent {{< param "PRODUCT_NAME" >}} instances
 for metric collection. Otherwise, metrics for the same services would be sent to different instances, making aggregations inaccurate.
-- `"traceID"`: spans/logs belonging to the same traceID will be exported to the same backend.
+- `"traceID"`: spans and logs belonging to the same `traceID` will be exported to the same backend.
 - `"resource"`: metrics belonging to the same resource will be exported to the same backend.
 - `"metric"`: metrics with the same name will be exported to the same backend.
 - `"streamID"`: metrics with the same streamID will be exported to the same backend
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -70,7 +70,7 @@ for metric collection. Otherwise, metrics for the same services would be sent to
 - `"metric"`: metrics with the same name will be exported to the same backend.
 - `"streamID"`: metrics with the same `streamID` will be exported to the same backend.
 
-The loadbalancer will configure the exporter for the singal types supported by the `routing_key`
+The loadbalancer configures the exporter for the signal types supported by the `routing_key`.
 
 Metrics support in `otelcol.exporter.loadbalancing` is considered experimental, so an exporter
 will be configured for metrics only if Alloy is run with `--stability-level=experimental`
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -67,7 +67,7 @@ This is useful when using processors like the span metrics, so all spans for eac
 for metric collection. Otherwise, metrics for the same services would be sent to different instances, making aggregations inaccurate.
 - `"traceID"`: spans/logs belonging to the same traceID will be exported to the same backend.
 - `"resource"`: metrics belonging to the same resource will be exported to the same backend.
-- `"metric"`: metrics with the same name will be exported to the same backend
+- `"metric"`: metrics with the same name will be exported to the same backend.
 - `"streamID"`: metrics with the same streamID will be exported to the same backend
 
 The loadbalancer will configure the exporter for the singal types supported by the `routing_key`
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -72,8 +72,9 @@ for metric collection. Otherwise, metrics for the same services would be sent to
 
 The loadbalancer configures the exporter for the signal types supported by the `routing_key`.
 
-Metrics support in `otelcol.exporter.loadbalancing` is considered experimental, so an exporter
-will be configured for metrics only if Alloy is run with `--stability-level=experimental`
+> **EXPERIMENTAL**: Metrics support in `otelcol.exporter.loadbalancing` is an [experimental][] feature.
+> Experimental features are subject to frequent breaking changes, and may be removed with no equivalent replacement.
+> The `stability.level` flag must be set to `experimental` to use the feature.
 
 ## Blocks
 
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -62,7 +62,7 @@ Name          | Type     | Description                          | Default     | 
 `routing_key` | `string` | Routing strategy for load balancing. | `"traceID"` | no
 
 The `routing_key` attribute determines how to route signals across endpoints. Its value could be one of the following:
-- `"service"`: spans/logs/metrics with the same `service.name` will be exported to the same backend.
+- `"service"`: spans, logs, and metrics with the same `service.name` will be exported to the same backend.
 This is useful when using processors like the span metrics, so all spans for each service are sent to consistent {{< param "PRODUCT_NAME" >}} instances
 for metric collection. Otherwise, metrics for the same services would be sent to different instances, making aggregations inaccurate.
 - `"traceID"`: spans/logs belonging to the same traceID will be exported to the same backend.
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     | 
 `routing_key` | `string` | Routing strategy for load balancing. | `"traceID"` | no
 
 The `routing_key` attribute determines how to route signals across endpoints. Its value could be one of the following:
-* `"service"`: spans with the same `service.name` will be exported to the same backend.
+- `"service"`: spans/logs/metrics with the same `service.name` will be exported to the same backend.
 This is useful when using processors like the span metrics, so all spans for each service are sent to consistent {{< param "PRODUCT_NAME" >}} instances
 for metric collection. Otherwise, metrics for the same services would be sent to different instances, making aggregations inaccurate.
-* `"traceID"`: spans belonging to the same traceID will be exported to the same backend.
+- `"traceID"`: spans/logs belonging to the same traceID will be exported to the same backend.
+- `"resource"`: metrics belonging to the same resource will be exported to the same backend.
+- `"metric"`: metrics with the same name will be exported to the same backend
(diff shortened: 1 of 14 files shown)


clayton-cornell:
Does this need a doc update in https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.loadbalancing/ ?

madaraszg-tulip:
> Does this need a doc update in https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.loadbalancing/ ?

I have now added a doc update to the PR

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
 `routing_key` | `string` | Routing strategy for load balancing. | `"traceID"` | no
 
 The `routing_key` attribute determines how to route signals across endpoints. Its value could be one of the following:
-* `"service"`: spans with the same `service.name` will be exported to the same backend.
+- `"service"`: spans/logs/metrics with the same `service.name` will be exported to the same backend.
```
clayton-cornell, line 65:
```suggestion
- `"service"`: spans, logs, and metrics with the same `service.name` will be exported to the same backend.
```

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
...
 This is useful when using processors like the span metrics, so all spans for each service are sent to consistent {{< param "PRODUCT_NAME" >}} instances
 for metric collection. Otherwise, metrics for the same services would be sent to different instances, making aggregations inaccurate.
-* `"traceID"`: spans belonging to the same traceID will be exported to the same backend.
+- `"traceID"`: spans/logs belonging to the same traceID will be exported to the same backend.
```
clayton-cornell, line 68:
```suggestion
- `"traceID"`: spans and logs belonging to the same `traceID` will be exported to the same backend.
```

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
...
+- `"resource"`: metrics belonging to the same resource will be exported to the same backend.
+- `"metric"`: metrics with the same name will be exported to the same backend
```
clayton-cornell, line 70:
```suggestion
- `"metric"`: metrics with the same name will be exported to the same backend.
```

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
...
+- `"streamID"`: metrics with the same streamID will be exported to the same backend
```
clayton-cornell, line 71:
```suggestion
- `"streamID"`: metrics with the same `streamID` will be exported to the same backend.
```

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
...
+
+The loadbalancer will configure the exporter for the singal types supported by the `routing_key`
```
clayton-cornell, line 73:
```suggestion
The loadbalancer configures the exporter for the signal types supported by the `routing_key`.
```

docs/sources/reference/components/otelcol/otelcol.exporter.loadbalancing.md
```diff
@@ -62,10 +62,18 @@ Name          | Type     | Description                          | Default     |
...
+
+Metrics support in `otelcol.exporter.loadbalancing` is considered experimental, so an exporter
+will be configured for metrics only if Alloy is run with `--stability-level=experimental`
```
clayton-cornell, line 76:
```suggestion
> **EXPERIMENTAL**: Metrics support in `otelcol.exporter.loadbalancing` is an [experimental][] feature.
> Experimental features are subject to frequent breaking changes, and may be removed with no equivalent replacement.
> The `stability.level` flag must be set to `experimental` to use the feature.
```

The way this is presented may change, but for now this matches the general notice of an Experimental feature here: https://grafana.com/docs/alloy/latest/reference/stdlib/array/#arraycombine_maps  This is parallel to the notice on Experimental or Public preview components (example:  https://grafana.com/docs/alloy/latest/reference/components/beyla/beyla.ebpf/ )

madaraszg-tulip: Thanks for the documentation fixes.

internal/component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -34,10 +35,33 @@ func init() {
...
-			return exporter.New(opts, fact, args.(Arguments), exporter.TypeLogs|exporter.TypeTraces)
+
+			myArgs := args.(Arguments)
+
+			var typeSignal exporter.TypeSignal
+
+			// As per https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/loadbalancingexporter/README.md
+			// metrics is considered "development" stability level
+
+			switch myArgs.RoutingKey {
```
ptodev, line 46: This approach will work ok the first time a component starts up. However, if the arguments change later on, the `typeSignal` would be out of date. For example, a component may start with `key = "traceID"` and accepting only traces. After a config update, it may then switch to `key = "metric"`. I'm not sure what the best way to handle this is.

ptodev (commented):
Thank you for the PR! If I remember correctly, there were two main problems in the past:
* We had to always send metrics, logs, and traces in OTLP components. #5684 fixed this so that we can only send a subset of those. Unfortunately, it also came with a limitation - a component is bound to the same types of telemetry throughout its lifetime. E.g. it might only be able to work for metrics.
* A component could only have one stability label. This is not a problem anymore - we now have the stability cmd args which can check if certain functionality is experimental.

Fortunately, there are ready solutions to those problems now. Unfortunately, one remaining challenge is that this exporter can change the types of outputs at runtime. It can start by accepting only metrics, then after a config update it may want to only accept traces in a way that's not compatible with any metrics config. I'm not sure what a good way to solve this problem would look like. If you want to try to implement a solution, I'd be happy to review the PR again 😊

internal/component/otelcol/exporter/loadbalancing/loadbalancing.go
madaraszg-tulip: What method is called when the component configuration is updated? Exporter.Update()? We'd need some kind of a callback to reconfigure e.supportedSignals

ptodev: Yes, it's [Exporter.Update](https://github.com/grafana/alloy/blob/c3d4086b769c2e1a8e6f5686e7ed80dab97b591d/internal/component/otelcol/exporter/exporter.go#L154). It seems like it's actually recreating the components from scratch, in which case your change might be close to the final solution.

Do you want to try make a unit test which creates an HTTP server, similar to [otlphttp_test.go](https://github.com/grafana/alloy/blob/c3d4086b769c2e1a8e6f5686e7ed80dab97b591d/internal/component/otelcol/exporter/otlphttp/otlphttp_test.go)? It could start with traces, then switch to metrics. Then it also might be more clear if there is a problem.

madaraszg-tulip: There certainly is a problem because the only place we set e.supportedSignals is in New(), and we need to update it in Update(). Will see what I can do

madaraszg-tulip: Okay, here's an idea. How about not storing the actual `supportedSignals` in the `Exporter` struct, but store a callback that returns the `TypeSignal` value from the component in question each time `Update()` is called, something like this:

```
diff --git a/internal/component/otelcol/exporter/exporter.go b/internal/component/otelcol/exporter/exporter.go
index 22a2b56d3..d54bf65f4 100644
--- a/internal/component/otelcol/exporter/exporter.go
+++ b/internal/component/otelcol/exporter/exporter.go
@@ -77,6 +77,8 @@ func (s TypeSignal) SupportsTraces() bool {
        return s&TypeTraces != 0
 }
 
+type TypeSignalFunc func(component.Options, component.Arguments) TypeSignal
+
 // Exporter is an Alloy component shim which manages an OpenTelemetry Collector
 // exporter component.
 type Exporter struct {
@@ -92,7 +94,7 @@ type Exporter struct {
 
        // Signals which the exporter is able to export.
        // Can be logs, metrics, traces or any combination of them.
-       supportedSignals TypeSignal
+       supportedSignals TypeSignalFunc
 }
 
 var (
@@ -106,7 +108,7 @@ var (
 //
 // The registered component must be registered to export the
 // otelcol.ConsumerExports type, otherwise New will panic.
-func New(opts component.Options, f otelexporter.Factory, args Arguments, supportedSignals TypeSignal) (*Exporter, error) {
+func New(opts component.Options, f otelexporter.Factory, args Arguments, supportedSignals TypeSignalFunc) (*Exporter, error) {
        ctx, cancel := context.WithCancel(context.Background())
 
        consumer := lazyconsumer.NewPaused(ctx)
@@ -212,8 +214,10 @@ func (e *Exporter) Update(args component.Arguments) error {
        // supported telemetry signals.
        var components []otelcomponent.Component
 
+       supportedSignals := e.supportedSignals(e.opts, args)
+
        var tracesExporter otelexporter.Traces
-       if e.supportedSignals.SupportsTraces() {
+       if supportedSignals.SupportsTraces() {
                tracesExporter, err = e.factory.CreateTracesExporter(e.ctx, settings, exporterConfig)
                if err != nil && !errors.Is(err, pipeline.ErrSignalNotSupported) {
                        return err
@@ -223,7 +227,7 @@ func (e *Exporter) Update(args component.Arguments) error {
        }
 
        var metricsExporter otelexporter.Metrics
-       if e.supportedSignals.SupportsMetrics() {
+       if supportedSignals.SupportsMetrics() {
                metricsExporter, err = e.factory.CreateMetricsExporter(e.ctx, settings, exporterConfig)
                if err != nil && !errors.Is(err, pipeline.ErrSignalNotSupported) {
                        return err
@@ -233,7 +237,7 @@ func (e *Exporter) Update(args component.Arguments) error {
        }
 
        var logsExporter otelexporter.Logs
-       if e.supportedSignals.SupportsLogs() {
+       if supportedSignals.SupportsLogs() {
                logsExporter, err = e.factory.CreateLogsExporter(e.ctx, settings, exporterConfig)
                if err != nil && !errors.Is(err, pipeline.ErrSignalNotSupported) {
                        return err





```

madaraszg-tulip: I'll look into creating a test for this now

madaraszg-tulip: Test is now created for this scenario

internal/component/otelcol/exporter/exporter.go
```diff
@@ -92,7 +100,7 @@ type Exporter struct {
 
 	// Signals which the exporter is able to export.
 	// Can be logs, metrics, traces or any combination of them.
-	supportedSignals TypeSignal
+	supportedSignals TypeSignalFunc
```
ptodev, line 103:
```suggestion
	// This is a function because which signals are supported may depend on the component configuration.
	supportedSignals TypeSignalFunc
```

ptodev (approved):

internal/component/otelcol/exporter/loadbalancing/loadbalancing.go
ptodev: Thank you, I think it's a good solution!

clayton-cornell (approved):

PR #2245, opened 2024-12-08
feat: Add ignore_older option for local.file_match

### LINKED ISSUE #2214: `local.file_match` ignore older files
### Request

Could the alloy `local.file_match` component be updated with an option to ignore files that have been modified in a certain period of time?

Something like:
```
local.file_match "tmp" {
  path_targets = [{"__path__" = "/tmp/logs/**/*.log"}]
  ignore_older = 1d
}
```

### Use case

Alloy has run out of memory attempting to scrape log files from a batch job that runs frequently and drops a unique log file per execution.
Alloy has also maintained many more open files than is necessary to run, increasing system resource usage.

While the argument could be made that better log rotation practices should be followed, I feel as if this would be a nice addition to match the feature provided by the [filebeat filestream ignore_older](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#filebeat-input-filestream-ignore-older) feature

ravishankar15:
Fixes #2214 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Main (unreleased)
   - Add perf_schema quantile columns to collector
 
 - Add three new stdlib functions to_base64, from_URLbase64 and to_URLbase64 (@ravishankar15)
-- Add ignore_older option for local.file_match
+- Add `ignore_older` option for local.file_match (@ravishankar15)
 
 - Use a forked `github.com/goccy/go-json` module which reduces the memory consumption of an Alloy instance by 20MB.
   If Alloy is running certain otelcol components, this reduction will not apply. (@ptodev)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Main (unreleased)
   - Add perf_schema quantile columns to collector
 
 - Add three new stdlib functions to_base64, from_URLbase64 and to_URLbase64 (@ravishankar15)
-- Add `ignore_older` option for local.file_match (@ravishankar15)
+- Add `ignore_older_than` option for local.file_match (@ravishankar15)
 
 - Use a forked `github.com/goccy/go-json` module which reduces the memory consumption of an Alloy instance by 20MB.
   If Alloy is running certain otelcol components, this reduction will not apply. (@ptodev)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ Main (unreleased)
   - Add perf_schema quantile columns to collector
 
 - Add three new stdlib functions to_base64, from_URLbase64 and to_URLbase64 (@ravishankar15)
+- Add ignore_older option for local.file_match
 
 - Use a forked `github.com/goccy/go-json` module which reduces the memory consumption of an Alloy instance by 20MB.
   If Alloy is running certain otelcol components, this reduction will not apply. (@ptodev)
@@ -101,7 +102,7 @@ v1.5.1
 
 - Fixed a crash when updating the configuration of `remote.http`. (@kinolaev)
 
-- Fixed an issue in the `otelcol.processor.attribute` component where the actions `delete` and `hash` could not be used with the `pattern` argument. (@wildum) 
+- Fixed an issue in the `otelcol.processor.attribute` component where the actions `delete` and `hash` could not be used with the `pattern` argument. (@wildum)
 
 - Fixed an issue in the `prometheus.exporter.postgres` component that would leak goroutines when the target was not reachable (@dehaansa)
 
@@ -301,7 +302,7 @@ v1.4.0
 
 - Add the label `alloy_cluster` in the metric `alloy_config_hash` when the flag `cluster.name` is set to help differentiate between
   configs from the same alloy cluster or different alloy clusters. (@wildum)
-  
+
 - Add support for discovering the cgroup path(s) of a process in `process.discovery`. (@mahendrapaipuri)
 
 ### Bugfixes
--- a/docs/sources/reference/components/local/local.file_match.md
+++ b/docs/sources/reference/components/local/local.file_match.md
@@ -28,7 +28,7 @@ Name                | Type                | Description                         
 ---------------     | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
 `path_targets`      | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
 `sync_period`       | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
-`ignore_older_than` | `duration`          | Ignores files which are modified before this duration                                      |         | no
+`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |         | no
 
 `path_targets` uses [doublestar][] style paths.
 * `/tmp/**/*.log` will match all subfolders of `tmp` and include any files that end in `*.log`.
--- a/docs/sources/reference/components/local/local.file_match.md
+++ b/docs/sources/reference/components/local/local.file_match.md
@@ -28,12 +28,13 @@ Name                | Type                | Description                         
 ---------------     | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
 `path_targets`      | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
 `sync_period`       | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
-`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |         | no
+`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |  `"0s"` | no
 
 `path_targets` uses [doublestar][] style paths.
 * `/tmp/**/*.log` will match all subfolders of `tmp` and include any files that end in `*.log`.
 * `/tmp/apache/*.log` will match only files in `/tmp/apache/` that end in `*.log`.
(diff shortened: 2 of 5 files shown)


internal/component/local/file_match/file_test.go
```diff
@@ -270,6 +299,7 @@ func createComponentWithLabels(t *testing.T, dir string, paths []string, exclude
 	}, Arguments{
 		PathTargets: tPaths,
 		SyncPeriod:  1 * time.Second,
+		IgnoreOlder: 1 * time.Hour,
```
wildum, line 302: nit: I would not modify this test

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Main (unreleased)
   - Add perf_schema quantile columns to collector
 
 - Add three new stdlib functions to_base64, from_URLbase64 and to_URLbase64 (@ravishankar15)
+- Add ignore_older option for local.file_match
```
wildum, line 48:
```suggestion

- Add `ignore_older` attribute for local.file_match (@ravishankar15)
```

internal/component/local/file_match/file.go
```diff
@@ -28,6 +28,7 @@ func init() {
 type Arguments struct {
 	PathTargets []discovery.Target `alloy:"path_targets,attr"`
 	SyncPeriod  time.Duration      `alloy:"sync_period,attr,optional"`
+	IgnoreOlder time.Duration      `alloy:"ignore_older,attr,optional"`
```
wildum, line 31: @dehaansa about the name you mentioned that something along the lines of "drop_older" might fit better?

wildum (commented):
Thanks for the contribution! Just a few nits comments, it looks really good, the one line logic is cool

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -28,6 +28,7 @@ Name            | Type                | Description
 --------------- | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
 `path_targets`  | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
 `sync_period`   | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
+`ignore_older`  | `duration`          | Ignores files which are modified before this duration                                      |         | no
```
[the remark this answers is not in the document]
ravishankar15: @wildum If we add that as "0s" wouldn't that be misleading that we ignore files older than current timestamp while the actual implementation considers all the files ?

internal/component/local/file_match/watch.go
```diff
@@ -48,6 +50,11 @@ func (w *watch) getPaths() ([]discovery.Target, error) {
 			}
 			continue
 		}
+
+		if w.ignoreOlder != 0 && fi.ModTime().Before(time.Now().Add(-w.ignoreOlder)) {
+			continue
+		}
```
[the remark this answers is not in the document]
ravishankar15: @wildum I was intentionally adding that before the dir check to prevent from navigating inside that directory if the files are not updated. Do you think we should change that ?

internal/component/local/file_match/file.go
wildum: I'm fine with both, I like the added consistency but I also find "ignore" a bit clearer than "drop" in this context.

internal/component/local/file_match/watch.go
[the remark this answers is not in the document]
wildum: the fi.IsDir() is just a boolean operation to check whether it's a file or a dir (see https://github.com/golang/go/blob/master/src/io/fs/fs.go#L231). I think that it makes sense to put it before and avoid system calls from the time.Now() when it's not needed

docs/sources/reference/components/local/local.file_match.md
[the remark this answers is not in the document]
wildum: Let's ask the doc expert, @clayton-cornell would you rather:

- have the default value explicitly documented as "0s" with a line under saying that 0 means that it will consider all the files. Although this is more transparent, this might be misleading for users to see the "0s".
- not document the "0s" to implicitly say that this functionality is deactivated by default

or something else?

internal/component/local/file_match/file.go
ptodev: I also think `ignore_older_than` is more clear than `drop_older_than`.

`local.file_match` creates brand new data out of the files it discovers, so technically it isn't dropping anything. It's just ignoring older files.

ravishankar15: Should I make the change on the config name ? to `ignore_older_than` it seems more aligned in this context ?

wildum: yes that's a good idea, thanks!

ravishankar15: @wildum I have made the changes. Please take a look

internal/component/local/file_match/watch.go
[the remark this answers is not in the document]
wildum: let's move it after fi.IsDir() then?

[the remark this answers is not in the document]
ravishankar15: If we move it after then we would be unnecessarily navigating to all the files in that directory ? Do you think we should still move it after the check ?

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -24,10 +24,11 @@ local.file_match "LABEL" {
...
-`path_targets`  | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
-`sync_period`   | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
+Name                | Type                | Description                                                                                | Default | Required
+---------------     | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
+`path_targets`      | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
+`sync_period`       | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
+`ignore_older_than` | `duration`          | Ignores files which are modified before this duration                                      |         | no
```
clayton-cornell, line 31:
```suggestion
`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |         | no
```

internal/component/local/file_match/watch.go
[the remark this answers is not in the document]
dehaansa: Directory modification time isn't propagated up the tree if a file is updated/added/etc multiple subdirectories down. IE if you have `test/log/my.log` and `my.log` is updated, while the `log` directory's modification time will change in most OSes the `test` directory's modification time will not. You shouldn't trust directory modification time for this purpose.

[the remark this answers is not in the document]
ravishankar15: Ohh interesting, Thats insightful, I have made the changes Thanks

docs/sources/reference/components/local/local.file_match.md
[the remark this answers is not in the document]
ravishankar15: Yup made the change Please take a look

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -24,15 +24,17 @@ local.file_match "LABEL" {
...
+---------------     | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
+`path_targets`      | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
+`sync_period`       | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
+`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |  `"0s"` | no
 
 `path_targets` uses [doublestar][] style paths.
 * `/tmp/**/*.log` will match all subfolders of `tmp` and include any files that end in `*.log`.
 * `/tmp/apache/*.log` will match only files in `/tmp/apache/` that end in `*.log`.
 * `/tmp/**` will match all subfolders of `tmp`, `tmp` itself, and all files.
+* When `ignore_older_than` is set to "0s" the component will not ignore any files
```
clayton-cornell, line 37:
```suggestion
* `local.file_match` doesn't ignore files when `ignore_older_than` is set to the default, `0s`.
```

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -24,15 +24,17 @@ local.file_match "LABEL" {
...
+---------------     | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
+`path_targets`      | `list(map(string))` | Targets to expand; looks for glob patterns on the  `__path__` and `__path_exclude__` keys. |         | yes
+`sync_period`       | `duration`          | How often to sync filesystem and targets.                                                  | `"10s"` | no
+`ignore_older_than` | `duration`          | Ignores files which are modified before this duration.                                     |  `"0s"` | no
 
 `path_targets` uses [doublestar][] style paths.
 * `/tmp/**/*.log` will match all subfolders of `tmp` and include any files that end in `*.log`.
 * `/tmp/apache/*.log` will match only files in `/tmp/apache/` that end in `*.log`.
 * `/tmp/**` will match all subfolders of `tmp`, `tmp` itself, and all files.
+* `local.file_match` doesn't ignore files when `ignore_older_than` is set to the default, `0s`.
```
wildum, line 37:
```suggestion

`local.file_match` doesn't ignore files when `ignore_older_than` is set to the default, `0s`.
```

wildum (approved):
LGTM, thanks again for another great contribution, you're helping us a lot! :)

PR #2259, opened 2024-12-11
Added resource samples

xogoodnow:
Added resource limit/request samples.
Commented out the resources.requests section for config reloader.

### DIFF
--- a/operations/helm/charts/alloy/values.yaml
+++ b/operations/helm/charts/alloy/values.yaml
@@ -111,7 +111,12 @@ alloy:
 
   # -- Resource requests and limits to apply to the Grafana Alloy container.
   resources: {}
-
+#    limits:
+#      cpu: "1"
+#      memory: "1Gi"
+#    requests:
+#      cpu: "1"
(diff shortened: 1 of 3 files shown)


xogoodnow:
Hi,
What seems to be the issue for this PR to move forward?

operations/helm/charts/alloy/values.yaml
```diff
@@ -166,10 +171,13 @@ configReloader:
...
-    requests:
-      cpu: "1m"
-      memory: "5Mi"
+  resources: {}
+#    limits:
+#      cpu: "1m"
+#      memory: "5Mi"
+#    requests:
+#      cpu: "1m"
+#      memory: "5Mi"
```
ptodev, line 180: IIUC, you are doing this so that you can set `resources` as a whole rather than `resources/request/cpu` and `resources/request/memory` individually? Isn't there still a way to set `resources` as a whole, without having to tweak the cpu and memory  params individually?

Also, wouldn't this break users who rely on the current defaults?

Issue #2223, opened 2024-12-05
Log bug? memory is not limited, skipping: %v


systemcrash:
### What's wrong?

This line looks wrong: `%v`?

alloy  | ts=2024-12-05T15:55:28.172162987Z level=info source=/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.6.0/memlimit/memlimit.go:170 msg="memory is not limited, skipping: %v" package=github.com/KimMachineGun/automemlimit/memlimit !BADKEY="memory is not limited"

### Steps to reproduce

Run alloy 1.5.1

### Software version

1.5.1

### Configuration

```text

```

### Logs

```text
alloy  | ts=2024-12-05T15:55:28.172162987Z level=info source=/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.6.0/memlimit/memlimit.go:170 msg="memory is not limited, skipping: %v" package=github.com/KimMachineGun/automemlimit/memlimit !BADKEY="memory is not limited"
```

dehaansa:
Looks like this was resolved upstream, and we'll make sure to update the dependency to fix it.
https://github.com/KimMachineGun/automemlimit/pull/19

Issue #2225, opened 2024-12-06
Alloy does not work on Windows 11


igolets:
### What's wrong?

I've installed alloy on my win11 machine using script provided by configurationmaster. "C:\Program Files\GrafanaLabs\Alloy\config.alloy" was created and contain my token. But test connection fails.
I did some basic check and here is wha I've found:
1. Service is running.
2. in Windows Application logs I see one error message — ts=2024-12-06T09:11:45.650179Z level=error source=/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.6.0/memlimit/memlimit.go:117 msg="failed to set GOMEMLIMIT" package=github.com/KimMachineGun/automemlimit/memlimit error="failed to set GOMEMLIMIT: cgroups is not supported on this system"

Googling did not give any clue.

### Steps to reproduce

1. Get setup script using configuration master (cd "%TEMP%" && powershell -c Invoke-WebRequest "https://storage.googleapis.com/cloud-onboarding/alloy/scripts/install-windows.ps1" -OutFile "install-windows.ps1" && powershell -executionpolicy Bypass -File ".\install-windows.ps1"....)
2. Run it in elevated CMD 
3. Check Windows eventlog (Application)

### System information

Windows 11 x64 PRO 24H2

### Software version

Grafana Alloy 1.5.1

### Configuration

```text

```

### Logs

```text

```

naZzethe3rd:
We confirmed the same error with Windows Server 2019.

dehaansa:
While that error log is annoying it should be a red herring, as it only affects automatic setting of the GOMEMLIMIT variable.

I see that you mentioned test connection failed in Fleet Management, however if you go to look at your collector fleet does the alloy instance appear? I believe FM is working on some issues where the test connection fails but the collector is actually successfully connected.

humbberto:
facing the same issue here 

```bash
.\alloy-windows-amd64.exe run .\config.alloy --stability.level public-preview
ts=2024-12-10T21:31:05.266222Z level=info "boringcrypto enabled"=false
ts=2024-12-10T21:31:04.9237752Z level=error source=/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.6.0/memlimit/memlimit.go:117 msg="failed to set GOMEMLIMIT" package=github.com/KimMachineGun/automemlimit/memlimit error="failed to set GOMEMLIMIT: cgroups is not supported on this system"
[...]

PR #2264, opened 2024-12-11
fix: Prevent the constant logging of cgroup errors on non-linux machines

dehaansa:
#### PR Description
1. Update automemlimit to fix the log format of an error log https://github.com/KimMachineGun/automemlimit/pull/19
2. Update use of automemlimit to not _try_ to use cgroup based logic when not running on Linux. Still allow use of the `system` experiment in automemlimit if the environment variable is set up to use it.

Fixes https://github.com/grafana/alloy/issues/2223
Relevant to https://github.com/grafana/alloy/issues/2225

#### PR Checklist

- [X] CHANGELOG.md updated


internal/alloycli/automemlimit_cgroups_unsupported.go
```diff
@@ -0,0 +1,25 @@
+//go:build !linux
+// +build !linux
```
ptodev, line 2: Apparently cgroups are a Linux feature and not so much a universal UNIX thing, so this should be fine.

internal/alloycli/automemlimit_linux.go
```diff
@@ -0,0 +1,12 @@
+package alloycli
```
ptodev, line 1: Shouldn't we have a `//go:build linux`? Otherwise, on a non-Linux, wouldn't Go try to build both files?

internal/alloycli/automemlimit_cgroups_unsupported.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/KimMachineGun/automemlimit/memlimit"
+	"github.com/grafana/alloy/internal/runtime/logging"
+)
+
+func applyAutoMemLimit(l *logging.Logger) {
+	// For non-linux builds without cgroups, memlimit will always report an error.
+	// However, if the system experiment is requested, we can use the system memory limit provider.
+	// This logic is similar to https://github.com/KimMachineGun/automemlimit/blob/main/memlimit/experiment.go
+	if v, ok := os.LookupEnv("AUTOMEMLIMIT_EXPERIMENT"); ok {
```
ptodev, line 20: IIUC, this requires the user to set `AUTOMEMLIMIT_EXPERIMENT`? But isn't the goal that Alloy does it automatically? 

Btw if we're introducing new env vars, they need to be mentioned here:
https://grafana.com/docs/alloy/latest/reference/cli/environment-variables/

internal/alloycli/automemlimit_cgroups_unsupported.go
```diff
@@ -0,0 +1,25 @@
...
+		if slices.Contains(strings.Split(v, ","), "system") {
+			memlimit.SetGoMemLimitWithOpts(memlimit.WithProvider(memlimit.FromSystem), memlimit.WithLogger(slog.New(l.Handler())))
```
ptodev, line 22: It'd be nice to have a Windows unit test to make sure there's no error message?

ptodev (commented):
Thank you! I was also thinking yesterday that having an error isn't a good UX :)

internal/alloycli/automemlimit_linux.go
mattdurham: The `linux` suffix behaves like go build linux.

mattdurham (approved):
Looks reasonable, only nit would be to document the var if we are referencing it.

internal/alloycli/automemlimit_cgroups_unsupported.go
dehaansa: I'll add it to the docs. This is an environment variable that is specific to the automemlimit module, and it's marked as experimental there so I didn't think we should enable it for all non-linix OSes in this PR. We can consider that change more broadly, this PR was just focused on retaining current functionality while reducing errors.

PR #2270, opened 2024-12-12
Fix static converters

wildum:
#### PR Description

https://github.com/grafana/alloy/pull/2262 I discovered a few bugs with the static converters:
- the spanmetrics connector is starting a goroutine on start. If the component is not stopped, it will panic. We don't have this panic on main because the component is not started. With the fix above, the component is started on creation.
- the azure static converter is always adding two auth methods but only one is allowed
- the otelcol.exporter.prometheus is not using the default value for the gc_frequency which result in an error when the value is not provided

As explained in the comment in testing.go, we need to run the controller to cleanup the component that are starting go routines on Start. But we can't do it for all components because some have config errors and others panic. Because these are only static converters tests, I decided to add the workaround that the controller will run only when the config contains a spanmetrics connector for now. LMK if you're ok with it or if you want to explore alternatives.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/converter/internal/prometheusconvert/component/azure.go
+++ b/internal/converter/internal/prometheusconvert/component/azure.go
@@ -25,12 +25,10 @@ func toDiscoveryAzure(sdConfig *prom_azure.SDConfig) *azure.Arguments {
 		return nil
 	}
 
-	return &azure.Arguments{
+	args := &azure.Arguments{
 		Environment:     sdConfig.Environment,
 		Port:            sdConfig.Port,
 		SubscriptionID:  sdConfig.SubscriptionID,
-		OAuth:           toDiscoveryAzureOauth2(sdConfig.ClientID, sdConfig.TenantID, string(sdConfig.ClientSecret)),
-		ManagedIdentity: toManagedIdentity(sdConfig),
 		RefreshInterval: time.Duration(sdConfig.RefreshInterval),
 		ResourceGroup:   sdConfig.ResourceGroup,
 		ProxyConfig:     common.ToProxyConfig(sdConfig.HTTPClientConfig.ProxyConfig),
@@ -38,6 +36,16 @@ func toDiscoveryAzure(sdConfig *prom_azure.SDConfig) *azure.Arguments {
 		EnableHTTP2:     sdConfig.HTTPClientConfig.EnableHTTP2,
 		TLSConfig:       *common.ToTLSConfig(&sdConfig.HTTPClientConfig.TLSConfig),
 	}
+
+	// Only one auth method is allowed.
+	switch sdConfig.AuthenticationMethod {
+	case "ManagedIdentity":
+		args.ManagedIdentity = toManagedIdentity(sdConfig)
+	default:
+		args.OAuth = toDiscoveryAzureOauth2(sdConfig.ClientID, sdConfig.TenantID, string(sdConfig.ClientSecret))
+	}
+
+	return args
 }
 
 func ValidateDiscoveryAzure(sdConfig *prom_azure.SDConfig) diag.Diagnostics {
--- a/internal/converter/internal/staticconvert/internal/build/converter_remotewriteexporter.go
+++ b/internal/converter/internal/staticconvert/internal/build/converter_remotewriteexporter.go
@@ -91,13 +91,20 @@ func toremotewriteexporterConfig(cfg *remotewriteexporter.Config, forwardTo []st
 	defaultArgs := &prometheus.Arguments{}
 	defaultArgs.SetToDefault()
 
-	return &prometheus.Arguments{
+	args := &prometheus.Arguments{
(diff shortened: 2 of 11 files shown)


internal/converter/internal/prometheusconvert/component/azure.go
```diff
@@ -25,19 +25,27 @@ func toDiscoveryAzure(sdConfig *prom_azure.SDConfig) *azure.Arguments {
...
-		ManagedIdentity: toManagedIdentity(sdConfig),
 		RefreshInterval: time.Duration(sdConfig.RefreshInterval),
 		ResourceGroup:   sdConfig.ResourceGroup,
 		ProxyConfig:     common.ToProxyConfig(sdConfig.HTTPClientConfig.ProxyConfig),
 		FollowRedirects: sdConfig.HTTPClientConfig.FollowRedirects,
 		EnableHTTP2:     sdConfig.HTTPClientConfig.EnableHTTP2,
 		TLSConfig:       *common.ToTLSConfig(&sdConfig.HTTPClientConfig.TLSConfig),
 	}
+
+	// Only one auth method is allowed.
```
ptodev, line 40: If the user supplied a config with both auth methods, isn't that technically a problem with the config and not with the converters? I'm ok for this check to remain, but not sure if it's overkill to check for invalid configuration in the converter.

internal/converter/internal/staticconvert/internal/build/converter_remotewriteexporter.go
```diff
@@ -91,13 +91,20 @@ func toremotewriteexporterConfig(cfg *remotewriteexporter.Config, forwardTo []st
...
 		IncludeScopeInfo:              defaultArgs.IncludeScopeInfo,
 		IncludeScopeLabels:            defaultArgs.IncludeScopeLabels,
-		GCFrequency:                   cfg.StaleTime,
+		GCFrequency:                   defaultArgs.GCFrequency,
 		ForwardTo:                     forwardTo,
 		AddMetricSuffixes:             defaultArgs.AddMetricSuffixes,
 		ResourceToTelemetryConversion: defaultArgs.ResourceToTelemetryConversion,
 	}
+
+	// Override default only if > 0 because GCFrequency of 0 is not allowed
```
ptodev, line 104: Same here, do we really need to check if the incoming config is invalid? Also, if we set a different GC frequency than what the user specified, we're technically not converting their exact config anymore. Maybe it'd make more sense to just refuse to convert the config, or to log a warning.

But IMO we could just convert it as it is and let Alloy fail when it loads it.

internal/converter/internal/test_common/testing.go
```diff
@@ -214,14 +217,42 @@ func attemptLoadingAlloyConfig(t *testing.T, bb []byte) {
 			clusterService,
 			labelstore.New(nil, prometheus.DefaultRegisterer),
 			remotecfgService,
+			livedebugging.New(),
 		},
 		EnableCommunityComps: true,
 	})
 	err = f.LoadSource(cfg, nil, "")
 
+	// This is a bit of an ugly workaround but the spanmetrics connector is starting a go routine in its start function.
```
ptodev, line 226: I guess this is referring to the start function of OTel Collector components? Do we call this when we call  Alloy's`New()` function for a component? I think those `New()` functions aren't meant to have side effects such as opening files and starting goroutines. They should only do this once Run is called.

internal/converter/internal/prometheusconvert/component/azure.go
wildum: I don't think that setting two auth methods is possible in static mode because you must use the `authentication_method` field to specify your method (default OAuth). The problem here is that no matter which auth method the user would set, the converter would always create the two auth blocks, resulting in an invalid config.

internal/converter/internal/staticconvert/internal/build/converter_remotewriteexporter.go
wildum: I dug deeper into this one because I could not understand why it would be set to 0 by default and I found out that when we moved to Alloy we transformed some static components into empty shells: https://github.com/grafana/alloy/pull/55/files#diff-1b127540731f7403cbe5211de129eddfc35579d88d6bcbc371520642a9aa5486. The problem with this component is that the default values were set in the newRemoteWriteExporter() function. This StaleTime default value is actually supposed to be 15 minutes. I got rid of the check in the converter and set the correct value in the yaml component. This way, if you convert a config that uses the default yaml value, it will be correctly converted to 15m in Alloy.

internal/converter/internal/test_common/testing.go
ptodev: Thank you for pointing this out... I added a [comment](https://github.com/grafana/alloy/pull/2262#discussion_r1886472989) to the other PR. `New` is not supposed to have such side effects.

PR #2274, opened 2024-12-12
SNMP exporter: Label targets block as deprecated

ptodev:
Since the `target` block is deprecated, it'd be nice to mention this in the docs #979.

In general, it's not ideal how there are so many ways to do the same thing in this components:
* `target` block and `targets` argument
* `config_file` argument and `config`argument

I wish we didn't have the `config_file` argument at all. I really don't see the benefit in it.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
@@ -17,6 +17,17 @@ The `prometheus.exporter.snmp` component embeds
 
 ## Usage
 
+### Recommended usage
+
+```alloy
+prometheus.exporter.snmp "LABEL" {
+  config_file = SNMP_CONFIG_FILE_PATH
+  targets     = TARGET_LIST
+}
+```
+
+### Deprecated usage
+
 ```alloy
 prometheus.exporter.snmp "LABEL" {
   config_file = SNMP_CONFIG_FILE_PATH
@@ -27,14 +38,12 @@ prometheus.exporter.snmp "LABEL" {
 }
 ```
 
-or
+Using the `target` block is deprecated because it is less flexible than the `targets` argument:
+* The name of the `target` block cannot contain certain characters, 
+  because it has to comply with Alloy syntax restrictions for [block labels][syntax-blocks].
+* With the `targets` argument you can also pass in additional labels.
 
-```alloy
-prometheus.exporter.snmp "LABEL" {
-  config_file = SNMP_CONFIG_FILE_PATH
-  targets     = TARGET_LIST
-}
-```
+[syntax-blocks]: ../../../../get-started/configuration-syntax/syntax#blocks
 
 ## Arguments
 
@@ -50,7 +59,8 @@ Omitted fields take their default values.
 The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
 Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter/tree/{{< param "SNMP_VERSION" >}}?tab=readme-ov-file#configuration) for details on how to generate a configuration file.
 
-The `config` argument must be a YAML document as string defining which SNMP modules and auths to use.
+The `config` argument is an alternative to the `config_file` argument. 
+It must be a YAML document as string defining which SNMP modules and auths to use.
 `config` is typically loaded by using the exports of another component. For example,
 
 - `local.file.LABEL.content`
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -334,6 +255,102 @@ The YAML file in this example looks like this:
     auth: public_v2
 ```
 
+## Example using the target block (deprecated)
```
wildum, line 258: did you build the docs and checked that it's correctly rendered? The github linting is showing weird colors but it's probably a lint bug because it seems ok to me

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -127,102 +137,9 @@ debug information.
 `prometheus.exporter.snmp` does not expose any component-specific
 debug metrics.
 
-## Example
+## Examples using the targets argument (recommended)
```
clayton-cornell, line 140:
```suggestion
## Examples using the targets argument

The following examples show you how to use `prometheus.exporter.snmp`.
```

I'd suggest we remove `recommended` here. There's no alternative deprecated example provided... 

We should try to add at least a sentence of text between headings.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -17,25 +17,34 @@ The `prometheus.exporter.snmp` component embeds
 
 ## Usage
 
+### Recommended usage
+
```
clayton-cornell, line 21: I'd like to remove this and keep it just Usage to keep it consistent with the other topic layouts.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -17,25 +17,34 @@ The `prometheus.exporter.snmp` component embeds
...
-
-  target "TARGET_NAME" {
-    address = TARGET_ADDRESS
-  }
+  targets     = TARGET_LIST
 }
 ```
 
-or
+### Deprecated usage
```
clayton-cornell, line 29: We could use this example: https://grafana.com/docs/writers-toolkit/write/shortcodes/#boltdb-deprecated to document the deprecated target block.

And I'd suggest we move this information down to the actual target block section (line 93) since that's where users will encounter it. And... don't provide a deprecated Usage section... Explicitly showing how to use a deprecated bit encourages people to keep using it.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
ptodev: There is indeed a deprecated example:

> `## Example using the target block (deprecated)`

clayton-cornell: Oh.. ooops. missed that. The comment about collapsed section similar to the linked example still stands though.  I'd like to see the examples only show the current config (vs saying "recommended") and we keep the deprecated stuff documented during the period it's still in Alloy... but tuck it away under a collapsed section.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -17,25 +17,34 @@ The `prometheus.exporter.snmp` component embeds
...
+  target "TARGET_NAME" {
+    address = TARGET_ADDRESS
+  }
 }
 ```
 
+Using the `target` block is deprecated because it is less flexible than the `targets` argument:
+* The name of the `target` block cannot contain certain characters, 
+  because it has to comply with Alloy syntax restrictions for [block labels][syntax-blocks].
+* With the `targets` argument you can also pass in additional labels.
```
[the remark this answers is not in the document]
ptodev: True, I added this as an extra bullet point.

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
ptodev: It displays ok on the website. But anyway, not I refactored the markdown and this is a `###` instead of a `##`

ptodev:
@clayton-cornell @wildum I edited the page to put less emphasis on the deprecated options. I do think it looks better now! Thank you for the feedback.

clayton-cornell:
@wildum I was looking at old PRs that are still open... that I've worked on, and stumbled on this one.  Is it still valid?  If I look at the source code for this component I see 

```
type Arguments struct {
	ConfigFile          string                    `alloy:"config_file,attr,optional"`
	SnmpConcurrency     int                       `alloy:"concurrency,attr,optional"`
	Config              alloytypes.OptionalSecret `alloy:"config,attr,optional"`
	ConfigMergeStrategy string                    `alloy:"config_merge_strategy,attr,optional"`
	Targets             TargetBlock               `alloy:"target,block,optional"`
	WalkParams          WalkParams                `alloy:"walk_param,block,optional"`
	ConfigStruct        snmp_config.Config

	// New way of passing targets. This allows the component to receive targets from other components.
	TargetsList TargetsList `alloy:"targets,attr,optional"`
```

Which tells me that `targets` is still a valid arg?  So... do we close this PR? Or have I missed something?

PR #2284, opened 2024-12-13
Don't run the workflow on forks

jdbaldry:
It's expected that it can't succeed on forks. We aren't supporting that use case yet.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


wildum (approved):
LGTM, thanks for the quick fix!

PR #2294, opened 2024-12-18
Performance optimization for live debugging feature

### LINKED ISSUE #974: Consider performance optimizations
Some optimizations which might improve performance / flexibility of the live debugging feature:
- make the buffer in the back-end configurable to allow the users to adjust it to their load
- flush the data on regular interval rather than flushing it on change

ravishankar15:

Fixes #974 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/docs/sources/reference/cli/run.md
+++ b/docs/sources/reference/cli/run.md
@@ -38,6 +38,7 @@ The following flags are supported:
 * `--server.http.memory-addr`: Address to listen for [in-memory HTTP traffic][] on (default `alloy.internal:12345`).
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--server.http.ui-path-prefix`: Base path where the UI is exposed (default `/`).
+* `--server.http.live-debugging-buffer-stream-size`: Buffer stream size used for buffering the live debugging entries (default `1000`)
 * `--storage.path`: Base directory where components can store data (default `data-alloy/`).
 * `--disable-reporting`: Disable [data collection][] (default `false`).
 * `--disable-support-bundle`: Disable [support bundle][] endpoint (default `false`).
--- a/docs/sources/reference/cli/run.md
+++ b/docs/sources/reference/cli/run.md
@@ -38,7 +38,6 @@ The following flags are supported:
 * `--server.http.memory-addr`: Address to listen for [in-memory HTTP traffic][] on (default `alloy.internal:12345`).
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--server.http.ui-path-prefix`: Base path where the UI is exposed (default `/`).
-* `--server.http.live-debugging-buffer-stream-size`: Buffer stream size used for buffering the live debugging entries (default `1000`)
 * `--storage.path`: Base directory where components can store data (default `data-alloy/`).
 * `--disable-reporting`: Disable [data collection][] (default `false`).
 * `--disable-support-bundle`: Disable [support bundle][] endpoint (default `false`).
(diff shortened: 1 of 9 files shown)


docs/sources/reference/cli/run.md
```diff
@@ -38,6 +38,7 @@ The following flags are supported:
 * `--server.http.memory-addr`: Address to listen for [in-memory HTTP traffic][] on (default `alloy.internal:12345`).
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--server.http.ui-path-prefix`: Base path where the UI is exposed (default `/`).
+* `--server.http.live-debugging-buffer-stream-size`: Buffer stream size used for buffering the live debugging entries (default `1000`)
```
wildum, line 41: a command flag is not super ideal because you need to restart Alloy to update the size of the buffer. I would recommend adding this functionality as an attribute of the [live debugging block](https://grafana.com/docs/alloy/latest/reference/config-blocks/livedebugging/). This way you can just update the size of the buffer in the config if you see that it's not enough and hit the reload endpoint (localhost:12345/-/reload)

ravishankar15: Yup makes sense have made the changes

internal/component/loki/process/process_test.go
```diff
@@ -164,7 +165,7 @@ func TestJSONLabelsStage(t *testing.T) {
 		"[IN]: timestamp: 2020-11-15T02:08:41-07:00, entry: {\"log\":\"log message\\n\",\"stream\":\"stderr\",\"time\":\"2019-04-30T02:12:41.8443515Z\",\"extra\":\"{\\\"user\\\":\\\"smith\\\"}\"}, labels: {filename=\"/var/log/pods/agent/agent/1.log\", foo=\"bar\"}",
 		"[OUT]: timestamp: 2019-04-30T02:12:41.8443515Z, entry: {\"log\":\"log message\\n\",\"stream\":\"stderr\",\"time\":\"2019-04-30T02:12:41.8443515Z\",\"extra\":\"{\\\"user\\\":\\\"smith\\\"}\"}, labels: {filename=\"/var/log/pods/agent/agent/1.log\", foo=\"bar\", stream=\"stderr\", ts=\"2019-04-30T02:12:41.8443515Z\", user=\"smith\"}",
 	}
-	require.Equal(t, expectedLiveDebuggingLog, liveDebuggingLog.Get())
+	assert.ElementsMatch(t, expectedLiveDebuggingLog, liveDebuggingLog.Get())
```
wildum, line 168: this test is not flaky on main, why would it sometimes not be ordered with your changes?

ravishankar15: It was failing in my earlier commit https://github.com/grafana/alloy/actions/runs/12394965463/job/34599585193

ravishankar15: Hey @wildum I missed adding the channel size in the tests have added the same. But I am not sure if that is the reason for failure. I checked the code and logic seems to be good I don't see any race conditions. Like you mentioned the IN and OUT are published in order and should be received in order. I tried to run this in local a couple of times frequently with the channel size fix it seems stable. Can you help me see if I am missing any race condition on this test failure ?

ravishankar15: ohh, You mean we should remove the config `buffer_stream_size` ? and have only the flush interval changes ? or just remove the config changes for the tests made in this [commit](https://github.com/grafana/alloy/pull/2294/commits/62c8855695a38f6efff6d6d304d8cd66e6f421a8)

wildum: I think that we should remove the config buffer_stream_size because it doesn't seem necessary in the end

wildum (approved):
LGTM, thanks!

clayton-cornell (approved):
Doc changes are fine.

PR #2296, opened 2024-12-18
Don't start components until Run is called

ptodev:
This is a patch to #2262 which prevents the OTel components from being started during New. It was discussed in [this thread](https://github.com/grafana/alloy/pull/2262/files#r1886472989).

### DIFF
--- a/internal/component/otelcol/internal/scheduler/scheduler.go
+++ b/internal/component/otelcol/internal/scheduler/scheduler.go
@@ -73,25 +73,43 @@ func NewWithPauseCallbacks(l log.Logger, onPause func(), onResume func()) *Sched
 // Schedule completely overrides the set of previously running components;
 // components which have been removed since the last call to Schedule will be
 // stopped.
-func (cs *Scheduler) Schedule(ctx context.Context, h otelcomponent.Host, cc ...otelcomponent.Component) {
+func (cs *Scheduler) Schedule(ctx context.Context, updateConsumers func(), h otelcomponent.Host, cc ...otelcomponent.Component) {
 	cs.schedMut.Lock()
 	defer cs.schedMut.Unlock()
 
-	cs.schedComponents = cc
-	cs.host = h
-
+	// If the scheduler isn't running yet, just update the state.
+	// That way the Run function is ready to go.
 	if !cs.running {
+		cs.schedComponents = cc
+		cs.host = h
+		updateConsumers()
 		return
 	}
 
-	cs.runScheduled(ctx)
+	cs.onPause()
+
+	// Stop the old components before running new scheduled ones.
+	cs.stopComponents(ctx, cs.schedComponents...)
+
+	updateConsumers()
+
+	level.Debug(cs.log).Log("msg", "scheduling components", "count", len(cs.schedComponents))
+	cs.schedComponents = cs.startComponents(ctx, h, cc...)
+	cs.host = h
+	//TODO: Check if there were errors? What if the trace component failed but the metrics one didn't? Should we resume all consumers?
+
+	cs.onResume()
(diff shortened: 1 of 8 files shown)


internal/component/otelcol/processor/processor.go
```diff
@@ -238,10 +240,8 @@ func (p *Processor) Update(args component.Arguments) error {
 	}
 
 	// Schedule the components to run once our component is running.
-	p.consumer.Pause()
 	p.consumer.SetConsumers(tracesProcessor, metricsProcessor, logsProcessor)
 	p.sched.Schedule(p.ctx, host, components...)
```
wildum, line 244: there is a race condition here on Update because the consumer is not paused while we set new unstarted otel components to it. Technically it's possible that a call to consume happens right after, before the start of the scheduling

internal/component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -37,13 +37,33 @@ type Scheduler struct {
...
 	host            otelcomponent.Host
+	running         bool
+
+	// onPause is called when scheduler is making changes to running components.
+	onPause func()
+	// onResume is called when scheduler is done making changes to running components.
+	onResume func()
 }
 
+// TODO: Delete this function? I don't think it's used anywhere.
```
wildum, line 48: I would be ok to delete it because it may actually lead to mistakes but you can check with Piotr because he decided to create a different one

internal/component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -53,19 +73,45 @@ func New(l log.Logger) *Scheduler {
 // Schedule completely overrides the set of previously running components;
 // components which have been removed since the last call to Schedule will be
 // stopped.
-func (cs *Scheduler) Schedule(ctx context.Context, h otelcomponent.Host, cc ...otelcomponent.Component) {
+func (cs *Scheduler) Schedule(ctx context.Context, updateConsumers func(), h otelcomponent.Host, cc ...otelcomponent.Component) {
```
wildum, line 76: somewhere in this file you could add the process order that you told me:

- Pause consumers
- Stop the process
- Change the consumers
- Start the process
- Start the consumer

This is a good way to understand what's happening and to make sure that there is no race condition that's breaking the logic

wildum (approved):
Nice one, you solved it in a way that's easily readable and is not ambiguous about race conditions. I'm happy to merge it in my PR and update my other PR: https://github.com/grafana/alloy/pull/2270

internal/component/otelcol/internal/scheduler/scheduler.go
ptodev: I just realised we need to keep it - it's used by `otelcol.auth.`, `otelcol.extension.`, and `otelcol.receiver.`.

ptodev:
Hi @wildum, I polished the PR a bit. Please feel free to merge it in if you are still happy with it. I'll lift my "needs improvement" on #2262 and will let @thampiotr review it again. I'm a co-author now, so it doesn't feel right to approve it :)

internal/component/otelcol/processor/processor.go
```diff
@@ -139,6 +139,8 @@ func (p *Processor) Run(ctx context.Context) error {
 // configuration for OpenTelemetry Collector processor configuration and manage
 // the underlying OpenTelemetry Collector processor.
 func (p *Processor) Update(args component.Arguments) error {
+	//TODO: Lock a mutex? There could be a race condition with multiple calls to Update
```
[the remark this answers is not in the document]
ptodev: Thank you, I removed this comment.

PR #2298, opened 2024-12-18
Add livedebugging support for prometheus.scrape

### LINKED ISSUE #1031: Add live debugging support to prometheus.scrape

ravishankar15:

Fixes #1031 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/scrape/scrape.go
+++ b/internal/component/prometheus/scrape/scrape.go
@@ -28,6 +28,7 @@ import (
 	"github.com/grafana/alloy/internal/service/cluster"
 	"github.com/grafana/alloy/internal/service/http"
 	"github.com/grafana/alloy/internal/service/labelstore"
+	"github.com/grafana/alloy/internal/service/livedebugging"
 	"github.com/grafana/alloy/internal/useragent"
 	"github.com/grafana/alloy/internal/util"
 )
@@ -188,10 +189,13 @@ type Component struct {
 
 	dtMutex            sync.Mutex
 	distributedTargets *discovery.DistributedTargets
+
+	debugDataPublisher livedebugging.DebugDataPublisher
 }
 
 var (
-	_ component.Component = (*Component)(nil)
+	_ component.Component     = (*Component)(nil)
+	_ component.LiveDebugging = (*Component)(nil)
 )
 
 // New creates a new prometheus.scrape component.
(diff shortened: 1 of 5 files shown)


internal/component/prometheus/scrape/scrape.go
```diff
@@ -341,6 +352,32 @@ func (c *Component) distributeTargets(
...
+			movedTargetLabels = append(movedTargetLabels, t.Labels().Copy()...)
+		}
+
+		data := fmt.Sprintf("oldTargetLabels: %s => newTargetLabels: %s => movedTargetLabels: %s",
+			oldTargetLabels.String(),
+			newTargetLabels.String(),
+			movedTargetLabels.String(),
+		)
+		c.debugDataPublisher.Publish(componentID, data)
+	}
```
wildum, line 379: with the scrape component it would be more interesting to get the metrics than the targets. I solved this on an experimental branch by using a Prometheus interceptor: https://github.com/grafana/alloy/blob/hackathon-alloy-live-graph/internal/component/prometheus/scrape/scrape.go#L266
I don't know if that's the best approach or not, it's crucial that it does not impact the performances when the debugDataPublisher is not active

ravishankar15: The interceptor sounds good. It acts a proxy function call and without debug enabled the publish is not going to happen. Also I see that the idea of interceptor is used in remote_write as well.

The other way is to hook up the interceptor only if the debugging is Active but it would make the code complicated as the `Update` function only updates the config and we have scraper manager which runs independently of update.

I feel the interceptor sounds good. Do you have other thoughts ?

wildum: I'm gonna run the interceptor version in a dev infra that uses prometheus scrape heavily to see if there are any performance problems.

ravishankar15: Sure ! If it helps I can make the interceptor changes to this branch for convenience. Shall add that commit ?

wildum: thanks but you can leave it as it is for now, I will keep you updated on the tests

clayton-cornell:
Docs are OK as they are right now.

internal/component/prometheus/scrape/scrape.go
wildum: small update: there was a bug in the implementation that I shared: the code should use the globalRef instead of using the localID to make a ref. I tested it and it seems fine but we are busy testing other features on our dev env so I will perform other tests a bit later.

Issue #2285, opened 2024-12-14
"grafana_cloud.stack" and "pyroscope.receive_http"


falahati:
### What's wrong?

Using both `grafana_cloud.stack` and `pyroscope.receive_http` results in a crash on Alloy's startup.

### Steps to reproduce

Alloy is deployed as a docker container with the following command and the specified configuration file:
```bash
  docker run \
  -v ~/config.alloy:/etc/alloy/config.alloy \
  -p 12345:12345 \
  -p 4317:4317 \
  -p 4318:4318 \
  -p 4040:4040 \
  --userns=host --cgroupns=host --privileged --ipc=host --pid=host --uts=host --security-opt="seccomp=unconfined" --security-opt="systempaths=unconfined" --security-opt="label=disable" \
  grafana/alloy:latest \
    run --server.http.listen-addr=0.0.0.0:12345 --storage.path=/var/lib/alloy/data --stability.level="experimental" \
    /etc/alloy/config.alloy
```

### System information

Containerized on Darwin ARM64

### Software version

Grafana Alloy 1.5.1 (latest)

### Configuration

```text
// imports the Grafana Cloud module from the Grafana Alloy repository
import.git "grafana_cloud" {
  repository = "https://github.com/grafana/alloy-modules.git"
  revision = "main"
  path = "modules/cloud/grafana/cloud/module.alloy"
  pull_frequency = "15m"
}

// should be read from environment variables, for now hardcoded for testing
grafana_cloud.stack "receivers" {
  stack_name = "XXXX"
  token = "glc_XXXX"
}

// creating a profile endpoint and forwarding to Grafana Cloud directly
pyroscope.receive_http "default" {
  http {
    listen_address = "0.0.0.0"
    listen_port = 4040
  }
  forward_to = [grafana_cloud.stack.receivers.profiles]
  // forward_to = [pyroscope.write.default.receiver]
}

// Defining a `pyroscope.write`, when reading the endpoint or username values from the `grafana_cloud.stack.receivers.info` 
// also results in the same exception when `pyroscope.receive_http` is pointed to the new `pyroscope.write`. 
// But hardcoding the values works.
pyroscope.write "default" {
  endpoint {
    url = grafana_cloud.stack.receivers.info["hpInstanceUrl"]

    basic_auth {
      username = grafana_cloud.stack.receivers.info["hpInstanceId"]
      password = "glc_XXXX"
    }
  }
}
```

### Logs

```text
panic: duplicate metrics collector registration attempted

[...]

PR #2302, opened 2024-12-19
fix(pyroscope.receive_http): Handle metrics registry gracefully

simonswine:
This component did not handle duplciate registrations well, esp. when the server was restarted.

This fixes bug report #2285.


wildum:
thanks for fixing the bug, please add changelog entries in the future (https://github.com/grafana/alloy/blob/main/CHANGELOG.md).

PR #2304, opened 2024-12-19
feat(pyroscope.java): Show debug info about profiled targets

simonswine:
This should make it clearer, what is currently profiled.

image

### DIFF
--- a/internal/component/pyroscope/java/java.go
+++ b/internal/component/pyroscope/java/java.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


korniltsev:
maybe let's add last_profile size to help somebody estimate their usages. Altgouth it's a jfr size, not the pprof that we count, I guess it will not hurt.

korniltsev (approved):
lgtm

simonswine:
> maybe let's add last_profile size to help somebody estimate their usages. Altgouth it's a jfr size, not the pprof that we count, I guess it will not hurt.

Yes good point, I went for the uncompressed pprof size:

image

PR #2309, opened 2024-12-21
Add livedebugging support for discovery components

### LINKED ISSUE #1034: Add live debugging support to discovery components

ravishankar15:

Fixes #1034 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/docs/sources/troubleshoot/debug.md
+++ b/docs/sources/troubleshoot/debug.md
@@ -112,6 +112,36 @@ Supported components:
 * `prometheus.relabel`
 {{< /admonition >}}
 * `discovery.relabel`
+* `discovery.azure`
+* `discovery.consul`
+* `discovery.consulagent`
+* `discovery.digitalocean`
+* `discovery.dns`
+* `discovery.docker`
+* `discovery.dockerswarm`
+* `discovery.ec2`
+* `discovery.eureka`
+* `discovery.file`
+* `discovery.gce`
+* `discovery.hetzner`
+* `discovery.http`
+* `discovery.ionos`
+* `discovery.kubelet`
+* `discovery.kubernetes`
+* `discovery.kuma`
+* `discovery.lightsail`
+* `discovery.linode`
+* `discovery.marathon`
+* `discovery.nerve`
+* `discovery.nomad`
+* `discovery.openstack`
+* `discovery.ovhcloud`
+* `discovery.process`
+* `discovery.puppetdb`
+* `discovery.scaleway`
+* `discovery.serverset`
+* `discovery.triton`
+* `discovery.uyuni`
 
 ## Debug using the UI
 
--- a/internal/component/discovery/discovery.go
+++ b/internal/component/discovery/discovery.go
@@ -2,6 +2,7 @@ package discovery
 
 import (
 	"context"
+	"fmt"
 	"slices"
 	"sort"
 	"strings"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/grafana/alloy/internal/service/livedebugging"
 )
 
 // Target refers to a singular discovered endpoint found by a discovery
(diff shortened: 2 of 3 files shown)


docs/sources/troubleshoot/debug.md
```diff
@@ -112,6 +112,36 @@ Supported components:
 * `prometheus.relabel`
 {{< /admonition >}}
 * `discovery.relabel`
```
dehaansa, line 114: We can probably abbreviate this to `discovery.*` unless there's discovery components not added by this PR.

internal/component/discovery/discovery.go
```diff
@@ -74,16 +76,26 @@ type Component struct {
...
+	debugDataPublisher livedebugging.DebugDataPublisher
 }
 
+var _ component.Component = (*Component)(nil)
+var _ component.LiveDebugging = (*Component)(nil)
+
 // New creates a discovery component given arguments and a concrete Discovery implementation function.
 func New(o component.Options, args component.Arguments, creator Creator) (*Component, error) {
+	debugDataPublisher, err := o.GetServiceData(livedebugging.ServiceName)
+	if err != nil {
```
dehaansa, line 89: If there is an error in connecting with livedebugging I would not expect that to cause the component to fail to instantiate.

Just looked, and this is clearly the established pattern so I'll bring that up in a separate channel, this is fine for now.

internal/component/discovery/discovery.go
```diff
@@ -257,3 +273,5 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 		}
 	}
 }
+
+func (c *Component) LiveDebugging(_ int) {}
```
dehaansa, line 277: My understanding is that this function is triggered on a change of livedebug consumers. Should the function Publish the cached targets data as discovery may not update rapidly? CC @wildum

wildum: I don't think so because the concept of live debugging is to show what's happening now. If I have live debugging open and I see targets I expect that the targets were just sent. The user can see the last sent targets in the exports area of the component page in the UI.
We can revisit this concept if that's something users really want to see on the live graph/live debugging feed but we could simply just use the export values

internal/component/discovery/discovery.go
wildum: yeah I agree that it is a bit weird that a missing debugging functionality would prevent the components from starting. But because this is not something that can happen (all services are always created at start), we decided to always expect the service to be there. We could revisit it by adding nil pointers check instead of falling + some way to log once  that the live debugging functionality is not available

docs/sources/troubleshoot/debug.md
clayton-cornell: Abbreviating would follow the pattern for other component groups.

docs/sources/troubleshoot/debug.md
```diff
@@ -112,6 +112,36 @@ Supported components:
...
+* `discovery.kubernetes`
+* `discovery.kuma`
+* `discovery.lightsail`
+* `discovery.linode`
+* `discovery.marathon`
+* `discovery.nerve`
+* `discovery.nomad`
+* `discovery.openstack`
+* `discovery.ovhcloud`
+* `discovery.process`
```
wildum, line 139: discovery process is not using the common discovery component. Can you add livedebugging to it too please?

ravishankar15: I have added the change

clayton-cornell: If that covers all `discovery.*` components (ie none are left out of this change), then you can remove lines 115-144 and update line 113 to be `discovery.*`.

ravishankar15: Yup done

docs/sources/troubleshoot/debug.md
```diff
@@ -110,8 +110,8 @@ Supported components:
 * `otelcol.processor.*`
 * `otelcol.receiver.*`
 * `prometheus.relabel`
-* `discovery.relabel`
 {{< /admonition >}}
+* `discovery.*`
```
clayton-cornell, line 114:
```suggestion
* `discovery.*`
{{< /admonition >}}
```

The `discovery.*` needs to me inside the `admonition` shortcode element. If you put it outside the closing tag, then it "falls" out of the note in the rendered webpage. :-)

internal/component/discovery/discovery.go
```diff
@@ -224,6 +236,10 @@ func (c *Component) runDiscovery(ctx context.Context, d DiscovererWithMetrics) {
 				allTargets = append(allTargets, labels)
 			}
 		}
+		componenentID := livedebugging.ComponentID(c.opts.ID)
+		if c.debugDataPublisher.IsActive(componenentID) {
+			c.debugDataPublisher.Publish(componenentID, fmt.Sprintf("%s", allTargets))
```
wildum, line 241: there is a small typo error for the componentID here and in the process.go:
componenentID -> componentID

wildum (commented):
Thanks! Just a small typo left and we are good to go

wildum:
Some discovery tests are failing

PR #2317, opened 2024-12-27
[pyroscope.java] Clean-up JFR files left behind by previous instances of alloy to reduce risk of filling up disk when alloy's in crash loop

### LINKED ISSUE #1960: [pyroscope.java] Old JFR files accumulating on disk
### What's wrong?

When alloy is terminated, JFR files written to disk by the `pyroscope.java` file never seem to be cleaned-up. In an extreme scenario when alloy is frequently restarted, like if OOMKilled on kubernetes, then maybe the files gradually accumulate and start filling up the host's disk.

This is making us take a gradual approach to adopting continuous profiling since we're worried about disk pressure on our kubernetes nodes. Not sure how valid of a concern that is, but it would be nice if alloy cleaned-up any of old JFR files, like during start-up

### Steps to reproduce

- Set-up alloy to auto-profile java applications, for example, using this recipe https://github.com/grafana/pyroscope/blob/main/examples/grafana-agent-auto-instrumentation/java/kubernetes/grafana-alloy.yaml
- Restart alloy and wait until it starts triggering profiling
- Check the `/tmp` directory of the java application container. Except after a rare race condition, there should now be two JFR files 

### System information

Ubuntu Linux 22.04.3 x86_64

### Software version

v1.4.3

### Configuration

https://github.com/grafana/pyroscope/blob/main/examples/grafana-agent-auto-instrumentation/java/kubernetes/grafana-alloy.yaml

### Logs

_No response_

swar8080:
#### PR Description

Cleans-up unused JFR files that were accumulating on profiling targets' filesystem by previously terminated instances of alloy. This helps reduce the risk of files piling up if alloy is constantly restarting, like if kubernetes memory limits are too low. 

Fixes #1960

#### Notes to the Reviewer

I tested this manually on our EKS cluster with amd64 linux nodes. I both manually created `asprof` files and used some left behind by terminated alloy pods to test that this new code is executed once during start-up.

image (18)

I couldn't find any unit or integration tests for this component so let me know if there's suggestions for further testing.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/java/loop.go
+++ b/internal/component/pyroscope/java/loop.go
@@ -7,6 +7,8 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
+	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -23,7 +25,12 @@ import (
 	gopsutil "github.com/shirou/gopsutil/v3/process"
 )
 
-const spyName = "alloy.java"
+const (
+	spyName       = "alloy.java"
+	processJfrDir = "/tmp"
+)
+
+var jfrFileNamePattern = regexp.MustCompile("^asprof-\\d+-\\d+\\.jfr$")
 
 type profilingLoop struct {
 	logger     log.Logger
@@ -45,6 +52,7 @@ type profilingLoop struct {
 func newProfilingLoop(pid int, target discovery.Target, logger log.Logger, profiler *asprof.Profiler, output *pyroscope.Fanout, cfg ProfilingConfig) *profilingLoop {
 	ctx, cancel := context.WithCancel(context.Background())
 	dist, err := profiler.DistributionForProcess(pid)
+	jfrFileName := fmt.Sprintf("asprof-%d-%d.jfr", os.Getpid(), pid)
 	p := &profilingLoop{
 		logger:   log.With(logger, "pid", pid),
 		output:   output,
@@ -52,7 +60,7 @@ func newProfilingLoop(pid int, target discovery.Target, logger log.Logger, profi
 		target:   target,
 		cancel:   cancel,
 		dist:     dist,
-		jfrFile:  fmt.Sprintf("/tmp/asprof-%d-%d.jfr", os.Getpid(), pid),
+		jfrFile:  filepath.Join(processJfrDir, jfrFileName),
 		cfg:      cfg,
 		profiler: profiler,
 	}
@@ -62,6 +70,16 @@ func newProfilingLoop(pid int, target discovery.Target, logger log.Logger, profi
 		p.onError(fmt.Errorf("failed to select dist for pid %d: %w", pid, err))
 		return p
 	}
+
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		// Clean-up files that weren't removed by a previous instance of alloy
+		err := p.cleanupOldJFRFiles(jfrFileName)
+		if err != nil {
+			_ = level.Warn(p.logger).Log("msg", "failed cleaning-up java jfr files created by a previous instance of alloy", "err", err)
+		}
+	}()
 
 	p.wg.Add(1)
 	go func() {
@@ -274,4 +292,23 @@ func (p *profilingLoop) alive() bool {
 		_ = level.Error(p.logger).Log("msg", "failed to check if process is alive", "err", err)
 	}
 	return err == nil && exists
+}
+
+func (p *profilingLoop) cleanupOldJFRFiles(myFileName string) error {
+	dir := asprof.ProcessPath(processJfrDir, p.pid)
+	files, err := os.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		if !file.IsDir() && jfrFileNamePattern.MatchString(file.Name()) && file.Name() != myFileName {
+			_ = level.Debug(p.logger).Log("msg", "deleting jfr file created by previous alloy process", "file", file.Name())
+			err := os.Remove(filepath.Join(dir, file.Name()))
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/component/pyroscope/java/loop.go
```diff
@@ -63,6 +71,16 @@ func newProfilingLoop(pid int, target discovery.Target, logger log.Logger, profi
 		return p
 	}
 
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		// Clean-up files that weren't removed by a previous instance of alloy
+		err := p.cleanupOldJFRFiles(jfrFileName)
```
ptodev, line 78: Is it possible that those files were created by something other than Alloy? And that deleting them could interfere with this other process?

internal/component/pyroscope/java/loop.go
```diff
@@ -23,7 +25,12 @@ import (
 	gopsutil "github.com/shirou/gopsutil/v3/process"
 )
 
-const spyName = "alloy.java"
+const (
+	spyName       = "alloy.java"
+	processJfrDir = "/tmp"
```
ptodev, line 30: It's odd that the process has been using `/tmp` for this. It'd make more sense to use the directory specified by the `--storage.path` [cmd arg](https://grafana.com/docs/alloy/latest/reference/cli/run/), similarly to other processes such as `loki.file` and `prometheus.remote_write`. 

@grafana/grafana-alloy-profiling-maintainers Would you mind if we change `pyroscope.java` to use the storage path instead please? I don't feel comfortable with Alloy deleting anything outside of that directory.

Note that if we do this we'd have to update the [docs](https://github.com/grafana/alloy/blob/main/docs/sources/reference/components/pyroscope/pyroscope.java.md) too. They currently [mention](https://grafana.com/docs/alloy/latest/reference/components/pyroscope/pyroscope.java/#profiling-behavior) that `tmp_dir` is used. The docs also mention this:

> The asprof binary runs with root permissions. If you change the tmp_dir configuration to something other than /tmp, then you must ensure that the directory is only writable by root.

I suppose using the storage path still won't be a problem, since root should also have access to it? IDK why the docs say "**only** writable by root" though.

ptodev (commented):
Hi, thank you for the PR! I'm worried about deleting files in directories which Alloy doesn't "own", so it'd be nice if we use the `--storage.path` dir instead.

internal/component/pyroscope/java/loop.go
```diff
@@ -63,6 +71,16 @@ func newProfilingLoop(pid int, target discovery.Target, logger log.Logger, profi
 		return p
 	}
 
+	p.wg.Add(1)
+	go func() {
```
dehaansa, line 75: Should this happen in a goroutine? I would expect that we want cleanup to _complete_ before the regular loop begins, so starting this and allowing the internal scheduler to decide what goes first doesn't seem right.

internal/component/pyroscope/java/loop.go
swar8080: Hi @ptodev, thanks for reviewing

> It's odd that the process has been using /tmp for this. It'd make more sense to use the directory specified by the --storage.path [cmd arg](https://grafana.com/docs/alloy/latest/reference/cli/run/)

Maybe this was to distinguish the storage directory on alloy containers from a directory on the profiled java process's container. `tmp_dir` needs to be a directory on the java container for async-profiler to work

I can change the JFR files to be written to `tmp_dir` as well though, since right now only the async-profiler binaries are written to that path

internal/component/pyroscope/java/loop.go
swar8080: Hi @dehaansa, I was thinking the clean-up can safely run in parallel to the loop as a small optimization. It has a check to make sure the JFR file eventually created by the loop isn't accidentally deleted during clean-up, since that file always has the same name for the current process running alloy. 

Let me know if you think it's still worth changing though!

internal/component/pyroscope/java/loop.go
swar8080: > Hi, thank you for the PR! I'm worried about deleting files in directories which Alloy doesn't "own", so it'd be nice if we use the --storage.path dir instead.

Another option could be changing the file name to something less likely to conflict, like by prepending `alloy-`

internal/component/pyroscope/java/loop.go
ptodev: I personally think it's ok to have that check and to delete the old files in parallel. It does make the code a bit more complicated, but maybe it's worth it to start profiling as quickly as possible.

internal/component/pyroscope/java/loop.go
ptodev: Thank you - I didn't realise that the `asprof` files are not in the Alloy container. It would indeed be nice to prefix the files with `alloy-`, or better yet - to just put them in a directory called `/tmp/alloy/`. Maybe it can even be called `tmp/alloy/<alloy_pid>/` - then you can just delete all the folders which don't match the current Alloy PID.

ptodev (commented):
@grafana/grafana-alloy-profiling-maintainers one thing that worries me about this change is that it wouldn't be possible for more than one Alloy instance to use `pyroscope.java` for the same process. Is that ok? I think it should be documented?

simonswine:
Thank you so much for your contribution! I am really sorry that we have not gotten to your PR in time.

I only realised after having started on https://github.com/grafana/alloy/pull/3630, that you created https://github.com/grafana/alloy/issues/1960 and this PR. 

In the PR above, I purposefully try to avoid assumptions about files I can be deleted. There might be valid use cases out of there, that would two Alloy instances to profile java processes without interfering with each other. In the happy path (e.g. no SIGKILL, we would clean up after ourselves).

There are still some limitations, e.g. when it comes to containerised targets stopping while they are profiles. Please do raise any problem you might have in an issue again and if you fail to get attention feel free to tag me.

Issue #2288, opened 2024-12-16
Panic: runtime error utilizing secretfilter component


yurii-kryvosheia:
### What's wrong?

Intermittent crashes of the pod when using loki.secretfilter component.
Alloy ran fine for a few days before it started crashlooping, I presume under a higher load.

### Steps to reproduce

Use custom gitleaks config in loki.secretfilter

### System information

Bottlerocket OS 1.26.2 arm64

### Software version

v1.5.0

### Configuration

```text
Alloy config:
...

loki.secretfilter "secret_filter" {
  forward_to = [loki.write.logs_service.receiver]
  redact_with     = "<REDACTED-SECRET:$SECRET_NAME>"
  gitleaks_config = "/etc/alloy/custom/gitleaks.toml"
  partial_mask    = 3
}

...
gitleaks.toml

title = "Custom Gitleaks Configuration"
description = "Detects secrets with custom rules"
[extend]
useDefault = false

[allowlist]
description = "global allow lists"

[[rules]]
id = "entry-code"
regex = '''(/investor/e|/investor/complete_profile/e|/e)/([a-zA-Z0-9]+)([/\d\?]+)?"'''
secretGroup = 2
entropy = 2

[[rules]]
id = "email"
regex = '''[^\/\s@*]+@[a-zA-Z0-9-]+\.[a-zA-Z]+'''

[[rules]]
id = "activation-code"
regex = '''(/api/users/attempt|/complete_profile/sso)/[a-zA-Z0-9-_@.<>:]+/([a-zA-Z0-9]+)"'''
secretGroup = 2
entropy = 2

[[rules]]
id = "verification-code"
regex = '''/api/users/attempt/[a-zA-Z0-9-_@.<>:]+/([a-zA-Z0-9]+)"'''


[[rules]]
id = "hash"
regex = '''(/users/attempt|/reset_password)/[a-zA-Z0-9-_@.<>:]+/([a-zA-Z0-9]+)"'''
secretGroup = 2
entropy = 2
```

### Logs

```text
[...]

PR #2320, opened 2024-12-31
fix(loki.secretfilter): Fix partial masking for short secrets and support multiple allowlists per rule

romain-gaillard:
#### PR Description
- Fixes the issue reported in #2288 where the component was crashing when the secret to mask was shorter than the `partial_mask` value.
- Adds (partial) support for the new `[[rules.allowlists]]` format added in Gitleaks v8.21.0
- Updates the documentation to be clearer on the fact that the component doesn't support all features of the Gitleaks configuration format.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -39,17 +39,21 @@ loki.secretfilter "<LABEL>" {
...
-`allowlist`              | `map(string)`        | List of regexes to allowlist matching secrets.  | `{}`                             | no
-`partial_mask`           | `number`             | Show the first N characters of the secret.      | `0`                              | no
+`gitleaks_config`        | `string`             | Path to the custom `gitleaks.toml` file.            | Embedded Gitleaks file           | no
+`types`                  | `map(string)`        | Types of secret to look for.                        | All types                        | no
+`redact_with`            | `string`             | String to use to redact secrets.                    | `<REDACTED-SECRET:$SECRET_NAME>` | no
+`include_generic`        | `bool`               | Include the generic API key rule.                   | `false`                          | no
+`allowlist`              | `map(string)`        | List of regexes to allowlist matching secrets.      | `{}`                             | no
+`partial_mask`           | `number`             | Show the first N characters (runes) of the secret.  | `0`                              | no
```
mattdurham, line 47: I liked the previous text without runes, this feels like go leaking into docs. @clayton-cornell thoughts?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -39,17 +39,21 @@ loki.secretfilter "<LABEL>" {
...
 
 The `gitleaks_config` argument is the path to the custom `gitleaks.toml` file.
 The Gitleaks configuration file embedded in the component is used if you don't provide the path to a custom configuration file.
 
-The `types` argument is a map of secret types to look for. The values are used as prefixes for the secret types in the Gitleaks configuration. If you don't provide this argument, all types are used.
+{{< admonition type="note" >}}
+This component does not support all the features of the Gitleaks configuration file. Currently, it only supports the regex-based rules, `secretGroup`, and allowlist regexes (`regexTarget` only supports the default value `secret`). Other features such as `keywords`, `entropy`, `paths`, and `stopwords` are not supported. The `extend` feature is also not supported, meaning that a custom configuration file must contain all the rules to use.
```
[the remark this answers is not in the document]
romain-gaillard: They would be ignored.

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -98,18 +98,24 @@ type Component struct {
...
 	AllowList struct {
 		Description string
-		Paths       []string
+		Paths       []string // Not used
 		Regexes     []string
 	}
 	Rules []struct {
 		ID          string
-		Description string
+		Description string // Not used
```
mattdurham, line 106: Are these unused variables still here so the format processes? What is the harm in removing them?

docs/sources/reference/components/loki/loki.secretfilter.md
clayton-cornell: Yeah we should avoid GoLang specifics in the docs if we can.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -72,8 +76,10 @@ The `include_generic` argument is a boolean that includes the generic API key ru
 The `allowlist` argument is a map of regular expressions to allow matching secrets.
 A secret will not be redacted if it matches any of the regular expressions. The allowlist in the Gitleaks configuration file is also applied.
 
-The `partial_mask` argument is the number of characters to show from the beginning of the secret before the redact string is added.
+The `partial_mask` argument is the number of characters (runes) to show from the beginning of the secret before the redact string is added.
```
clayton-cornell, line 79: Another instance of `runes` that should probably be removed?  Same comment for the following lines (81 and 82)

docs/sources/reference/components/loki/loki.secretfilter.md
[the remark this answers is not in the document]
romain-gaillard: Yes, indeed, it's the first one. 

I think it's still worth keeping a mention to the Gitleaks configuration file? Because people would read the documentation on the Gitleaks configuration file (or import it from a different place they use it) and maybe expect some features to work in this component too. That's why I wanted to list what's supported and what's not _in the Gitleaks configuration file format_

internal/component/loki/secretfilter/secretfilter.go
romain-gaillard: Yes, indeed they could be removed. I initially added them to keep track of features we don't currently support but might in the future.

docs/sources/reference/components/loki/loki.secretfilter.md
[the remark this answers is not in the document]
romain-gaillard: Also, for the _feature_ part, it's probably both. The feature (e.g. entropy, stopwords, ...) is configured via arguments. The feature is currently not implemented and the related arguments are therefore ignored.

internal/component/loki/secretfilter/secretfilter.go
mattdurham: In that case I would rather remote/comment out the whole line so they cant mistakenly be used.

romain-gaillard: 👍 I removed it

mostafa (approved):

romain-gaillard:
@clayton-cornell @mattdurham 
Thank you so much for your comments and suggestions, they should all be implemented.

mattdurham (approved):
LGTM

mattdurham:
Will merge once clayton approves.

clayton-cornell (approved):
LGTM @mattdurham it's good to go

PR #2335, opened 2025-01-03
fix debug doc

wildum:


clayton-cornell (approved):
Ooops. I should have caught that in the previous PR :-P

PR #2351, opened 2025-01-08
Fix topics in loki.source.kafka

wildum:
#### PR Description

The code claims that it will use regex if the topic name starts with "^" but it actually always uses regex. As a result, if you have the topic "foo" and "foobar", you will match both by setting "foo" only.

The fix prevents this from happening by using exact match by default and documenting the possible use of regex.

This is technically a breaking change for users relying on the undocumented regex behavior.

To avoid the breaking change, we could document that an exact match can be achieved by having the name start with "^" and end with "$" but I would prefer fixing this to match the intended way. Because the use of regex was undocumented, it does not breach our backward compatibility statement.

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


CHANGELOG.md
thampiotr, line 1: Maybe we should call this change out a bit more prominently in release notes. Something similar to the way we document when upstream libraries make breaking changes (which is covered by exceptions in our compatibility guarantees)?

thampiotr (approved):

CHANGELOG.md
wildum: done

thampiotr (approved):

docs/sources/reference/components/loki/loki.source.kafka.md
```diff
@@ -49,6 +49,8 @@ loki.source.kafka "LABEL" {
 
 `assignor` values can be either `"range"`, `"roundrobin"`, or `"sticky"`.
 
+If a topic starts with a '^', it is treated as a regex and may match multiple topics.
```
clayton-cornell, line 52:
```suggestion
If a topic starts with a '^', it is treated as a regular expression and may match multiple topics.
```

docs/sources/release-notes.md
```diff
@@ -14,6 +14,14 @@ For a complete list of changes to {{< param "FULL_PRODUCT_NAME" >}}, with links
 
 [Changelog]: https://github.com/grafana/alloy/blob/main/CHANGELOG.md
 
+## v1.6
+
+### Breaking change: The `topics` argument in the component `loki.source.kafka` does not use regex by default anymore
+
+A bug in `loki.source.kafka` caused the component to treat all topics as regex. For example, setting the topic value to "foo" would match any topic containing the substring "foo".
```
clayton-cornell, line 21: We should avoid using "foo" in the docs (including the readme).  https://developers.google.com/style/word-list#foo   Can we come up with a more applicable generic topic value here?

docs/sources/release-notes.md
```diff
@@ -14,6 +14,14 @@ For a complete list of changes to {{< param "FULL_PRODUCT_NAME" >}}, with links
...
+With the fix introduced in this version, topic values are now treated as exact matches by default.
+Regex matching is still supported by prefixing a topic with "^", allowing it to match multiple topics.
```
clayton-cornell, line 23:
```suggestion
Regular expression matching is still supported by prefixing a topic with "^", allowing it to match multiple topics.
```

docs/sources/release-notes.md
clayton-cornell, line 21:
```suggestion
A bug in `loki.source.kafka` caused the component to treat all topics as regular expressions. For example, setting the topic value to "foo" would match any topic containing the substring "foo".
```

wildum: ack, replaced by "telemetry"

PR #2353, opened 2025-01-08
Update support bundle link to the correct file in github

callmehyde:
The previous link went to a 404 page. This commit is an update to the correct url.

**Docs page**: https://grafana.com/docs/alloy/latest/troubleshoot/support_bundle/

**Old**: https://github.com/grafana/alloy/tree/internal/service/http/supportbundle.go
**New**: https://github.com/grafana/alloy/blob/main/internal/service/http/supportbundle.go


clayton-cornell (approved):
Oh, good catch, Thanks for the fix.

Issue #2357, opened 2025-01-09
Image Version in the given example is old


adamab48:
### Steps to reproduce

- Clone the repo
- do a `docker compose up -d`
- do a 'docker compose --profile=alloy up -d
- inspect the logs to see this : 
```bash
alloy-upstream/example on  main took 3s 
❯ docker compose --profile=alloy logs alloy -f
alloy-1  | 
alloy-1  | Error: /etc/alloy/config.alloy:64:9: component "string.format" does not exist or is out of scope
alloy-1  | 
alloy-1  | 63 |     endpoint {
alloy-1  | 64 |         url = string.format(
alloy-1  |    |               ^^^^^^^^
alloy-1  | 65 |             "http://%s/loki/api/v1/push",
alloy-1  | 
alloy-1  | Error: /etc/alloy/config.alloy:66:13: component "sys.env" does not exist or is out of scope
alloy-1  | 
alloy-1  | 65 |             "http://%s/loki/api/v1/push",
alloy-1  | 66 |             coalesce(sys.env("LOKI_HOST"), "localhost:3100"),
alloy-1  |    |                      ^^^^^
alloy-1  | 67 |         )
alloy-1  | 
alloy-1  | Error: /etc/alloy/config.alloy:83:9: component "string.format" does not exist or is out of scope
```

### Configuration

```text

```

### Logs

```text

```

dehaansa:
Thanks for catching this!

PR #2358, opened 2025-01-09
update the image version to work with the given example

adamab48:
#### PR Description
Updated the image version in the example given since it's old and does not work with the updated configuration


adamab48:
Fixes #2357

clayton-cornell (approved):
Thanks for catching this.

dehaansa (approved):

PR #2363, opened 2025-01-09
Update wal queue tls

### LINKED ISSUE #1907: `prometheus.write.remote` add tls structure
### Request

Add TLS settings to `promethues.write.remote`

### Use case

So people can specifiy keys!

mattdurham:
#### PR Description

This adds TLS support to `prometheus.write.queue` , also fixes me removing the unreleased header. The underlying walqueue library has tests and we dont have any tests for the remote_write TLS config and underneath its functionally the same.

Fixes #1907

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/internal/component/prometheus/write/queue/types.go
+++ b/internal/component/prometheus/write/queue/types.go
@@ -119,7 +119,7 @@ func (cc EndpointConfig) ToNativeType() types.ConnectionConfig {
 	if cc.TLSConfig != nil {
 		tcc.InsecureSkipVerify = cc.TLSConfig.InsecureSkipVerify
 		tcc.TLSCert = cc.TLSConfig.Cert
-		tcc.TLSKey = cc.TLSConfig.Cert
+		tcc.TLSKey = string(cc.TLSConfig.Key)
 		tcc.TLSCACert = cc.TLSConfig.CA
 	}
 	return tcc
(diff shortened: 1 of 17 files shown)


wildum (dismissed):

internal/component/prometheus/write/queue/types.go
```diff
@@ -105,13 +108,20 @@ func (cc EndpointConfig) ToNativeType() types.ConnectionConfig {
...
 			Username: cc.BasicAuth.Username,
 			Password: string(cc.BasicAuth.Password),
 		}
 	}
+	if cc.TLSConfig != nil {
+		tcc.InsecureSkipVerify = cc.TLSConfig.InsecureSkipVerify
+		tcc.TLSCert = cc.TLSConfig.Cert
+		tcc.TLSKey = string(cc.TLSConfig.Key)
+		tcc.TLSCACert = cc.TLSConfig.CA
+	}
```
wildum, line 124: I think that we should not allow BasicAuth and TLSConfig to be set at the same time + having a test that sets the TLS and a test that sets both to check the error would be nice

mattdurham: Good call, will add.

clayton-cornell (approved):
Docs look ok to me :-)

PR #2379, opened 2025-01-10
Fix examples for filter and transform processors

wildum:
I also used backticks for the filter processor to remove the need to escape every string in the OTTL statements.

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.processor.filter.md
+++ b/docs/sources/reference/components/otelcol/otelcol.processor.filter.md
@@ -197,7 +197,7 @@ information.
 
 ### Drop spans which contain a certain span attribute
 
-This example sets the attribute `test` to `pass` if the attribute `test` does not exist.
+This example drops the signals that have the attribute `container.name` set to the value `app_container_1`.
 
 ```alloy
 otelcol.processor.filter "default" {
@@ -205,7 +205,7 @@ otelcol.processor.filter "default" {
 
   traces {
     span = [
-      "attributes[\"container.name\"] == \"app_container_1\"",
+      `attributes["container.name"] == "app_container_1"`,
     ]
   }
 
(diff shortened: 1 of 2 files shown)


docs/sources/reference/components/otelcol/otelcol.processor.filter.md
```diff
@@ -197,15 +197,15 @@ information.
...
+This example drops the signals that have the attribute `container.name` set to the value `app_container_1`.
 
 ```alloy
 otelcol.processor.filter "default" {
   error_mode = "ignore"
 
   traces {
     span = [
-      "attributes[\"container.name\"] == \"app_container_1\"",
+      `attributes["container.name"] == "app_container_1"`,
```
clayton-cornell, line 208: Since we are not escaping the double quotes... lines 220, 248-250, and 279-287 could be removed.

docs/sources/reference/components/otelcol/otelcol.processor.transform.md
```diff
@@ -399,7 +399,7 @@ otelcol.processor.transform "default" {
     context = "resource"
     statements = [
       // The Concat function combines any number of strings, separated by a delimiter.
-      `set(attributes["test"], Concat([attributes["foo"], attributes["bar"]], " "))`,
+      `set(attributes["test"], Concat([attributes["service.name"], attributes["service.version"]], " "))`,
```
clayton-cornell, line 402: I wonder if we really need the text explaining the backticks vs double quotes that is used throughout this topic...

docs/sources/reference/components/otelcol/otelcol.processor.filter.md
wildum: nice catch, removed

clayton-cornell (approved):
Nice cleanup on this one :-) I'll get it backported since these little details seem to span all the way back to the 1.0 release.

PR #2392, opened 2025-01-13
database_observability: report health of component and collectors

cristiangreco:
#### PR Description
Report unhealthy in case of errors when starting up the collectors or of any collector is stopped during operations.

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -155,6 +171,11 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
 		}
 	}
 
+	if len(schemas) == 0 {
+		level.Info(c.logger).Log("msg", "no schema detected from information_schema.schemata")
+		return nil
+	}
+
```
cristiangreco, line 178: Drive-by: log if no schema is detected

internal/component/database_observability/mysql/component.go
```diff
@@ -79,12 +80,15 @@ type Exports struct {
...
-	_ http_service.Component = (*Component)(nil)
+	_ component.Component       = (*Component)(nil)
+	_ http_service.Component    = (*Component)(nil)
+	_ component.HealthComponent = (*Component)(nil)
 )
 
 type Collector interface {
+	Name() string
 	Start(context.Context) error
+	Stopped() bool
```
wildum, line 91: In the future, you might have collectors that can be considered unhealthy but are still running. A different approach to support this would be to have a CurrentHealth function in the collector interface that returns the health object. Then you would not need the healthErr attribute anymore, you would just call CurrentHealth on all the collectors in the CurrentHealth function of the component.

wildum (approved):
LGTM, I just suggested a different approach to give the collectors more flexibility on their health status but feel free to ignore

fridgepoet (approved):

internal/component/database_observability/mysql/component.go
cristiangreco: Yes that's a great point. I wanted to start simple for now, as collectors are anyway not resilient at all (they'll stop as soon as any error is hit). Agree that in the future we might want to delegate the logic to the collectors themselves.

matthewnolf (approved):

PR #2404, opened 2025-01-15
Disable debug metrics for components inside foreach, and for foreach itself.

ptodev:
Opening a separate PR for @wildum to review the changes before they go on the main dynamic pipelines branch.

### DIFF
--- a/internal/runtime/internal/controller/node_builtin_component.go
+++ b/internal/runtime/internal/controller/node_builtin_component.go
@@ -60,20 +60,25 @@ func (id ComponentID) Equals(other ComponentID) bool {
 // DialFunc is a function to establish a network connection.
 type DialFunc func(ctx context.Context, network, address string) (net.Conn, error)
 
+type ModuleControllerOpts struct {
+	Id  string
+	Reg prometheus.Registerer
+}
+
 // ComponentGlobals are used by BuiltinComponentNodes to build managed components. All
 // BuiltinComponentNodes should use the same ComponentGlobals.
 type ComponentGlobals struct {
-	Logger               *logging.Logger                        // Logger shared between all managed components.
-	TraceProvider        trace.TracerProvider                   // Tracer shared between all managed components.
-	DataPath             string                                 // Shared directory where component data may be stored
-	MinStability         featuregate.Stability                  // Minimum allowed stability level for features
-	OnBlockNodeUpdate    func(cn BlockNode)                     // Informs controller that we need to reevaluate
-	OnExportsChange      func(exports map[string]any)           // Invoked when the managed component updated its exports
-	Registerer           prometheus.Registerer                  // Registerer for serving Alloy and component metrics
-	ControllerID         string                                 // ID of controller.
-	NewModuleController  func(id string) ModuleController       // Func to generate a module controller.
-	GetServiceData       func(name string) (interface{}, error) // Get data for a service.
-	EnableCommunityComps bool                                   // Enables the use of community components.
+	Logger               *logging.Logger                                  // Logger shared between all managed components.
+	TraceProvider        trace.TracerProvider                             // Tracer shared between all managed components.
+	DataPath             string                                           // Shared directory where component data may be stored
+	MinStability         featuregate.Stability                            // Minimum allowed stability level for features
+	OnBlockNodeUpdate    func(cn BlockNode)                               // Informs controller that we need to reevaluate
+	OnExportsChange      func(exports map[string]any)                     // Invoked when the managed component updated its exports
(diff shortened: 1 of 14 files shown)


internal/runtime/internal/controller/node_builtin_component.go
```diff
@@ -60,20 +60,25 @@ func (id ComponentID) Equals(other ComponentID) bool {
 // DialFunc is a function to establish a network connection.
 type DialFunc func(ctx context.Context, network, address string) (net.Conn, error)
 
+type ModuleControllerOpts struct {
+	Id  string
+	Reg prometheus.Registerer
+}
```
wildum, line 66: Maybe this should be defined in internal/runtime/internal/controller/module.go?

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -111,8 +118,9 @@ func (fn *ForeachConfigNode) ID() ComponentID {
 // Foreach doesn't have the ability to export values.
 // This is something we could implement in the future if there is a need for it.
 type Arguments struct {
-	Collection []any  `alloy:"collection,attr"`
-	Var        string `alloy:"var,attr"`
+	Collection    []any  `alloy:"collection,attr"`
+	Var           string `alloy:"var,attr"`
+	EnableMetrics bool   `alloy:"enable_metrics,attr,optional"`
```
wildum, line 123: maybe a small comment here to justify why we disable the metrics by default

internal/runtime/testdata/foreach_metrics/foreach_2.txtar
```diff
@@ -0,0 +1,42 @@
...
+# TYPE alloy_component_controller_running_components gauge
+alloy_component_controller_running_components{controller_id="",controller_path="/",health_type="healthy"} 2
+alloy_component_controller_running_components{controller_id="foreach_10_1",controller_path="/foreach.testForeach",health_type="healthy"} 1
+# HELP alloy_component_evaluation_queue_size Tracks the number of components waiting to be evaluated in the worker pool
+# TYPE alloy_component_evaluation_queue_size gauge
+alloy_component_evaluation_queue_size{controller_id="",controller_path="/"} 1
+alloy_component_evaluation_queue_size{controller_id="foreach_10_1",controller_path="/foreach.testForeach"} 0
+# HELP pulse_count
+# TYPE pulse_count counter
+pulse_count{component_id="testcomponents.pulse.pt",component_path="/foreach.testForeach/foreach_10_1"} 0
```
wildum, line 38: why is it 0? Should the count metric be updated in the pulse component to get the correct count here?

internal/runtime/testdata/foreach_metrics/foreach_1.txtar
```diff
@@ -0,0 +1,35 @@
+Foreach with only one item. The pulse will send "1" to the receiver of the summation component until it reaches 10.
```
wildum, line 1: the description should be updated to tell that this test is focusing on testing the metrics. Same for the test below

internal/runtime/testdata/foreach_metrics/foreach_2.txtar
```diff
@@ -0,0 +1,42 @@
+Foreach with only one item. The pulse will send "1" to the receiver of the summation component until it reaches 10.
+
+-- main.alloy --
+foreach "testForeach" {
+  collection = [10]
```
wildum, line 5: it would be nice also to have a test with two items to make sure that the metrics are correctly registered and updated

wildum (commented):
Thanks, looks good overall, I really like the added tests, just a few small comments

internal/runtime/internal/controller/node_builtin_component.go
ptodev: Why should it be there? It's used in `type ComponentGlobals struct {`, right below.

internal/runtime/testdata/foreach_metrics/foreach_2.txtar
ptodev: Yes, apologies. I'd setup the metric, but forgot to actually make it count the pulse. It's fixed now!

internal/runtime/internal/controller/node_builtin_component.go
wildum: true, then we can leave it there for now. I think that in the future it would be best to move some parts of the code in the node_builtin_component.go in a more generic file because they are not just used by the builtin components anymore but by other nodes as well

wildum (approved):

PR #2276, opened 2024-12-12
Add json format support for log export via faro receiver

ravishankar15:
#### Which issue(s) this PR fixes
Fixes #1973 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Config converters updated

PR #2407, opened 2025-01-15
[faro receiver] Move location of changelog entry and waitgroup wait

ptodev:
Minor cleanup after #2276. The PR was not quite up to date, so the changelog entry went in the RC instead of `main`. I didn't want to ask the contributor (@ravishankar15) to rebase, since I'd already asked once before and didn't get around to merging it 😅 

I'm also moving the waitgroup Wait to a defer statement so that it gets executed at the end of the goroutine. I don't think it makes a huge difference for the test, but it's technically more correct.


mattdurham (approved):
LGTM

Issue #2292, opened 2024-12-17
Alloy - not scraping logs after log rotation in Windows


umeshvh:
### What's wrong?

In Windows, When application logs are rotated, they are not fully scraped. The Alloy agent scrapes logs only from the initial file (e.g., api.log) but misses the newly created file after log rotation is triggered.

Response from Alloy team:

Alloy engineering team confirmed this is an **issue with Alloy running on Windows**. This was their findings:
 
On Windows, when the file is deleted the tailer is stopped right away. When the new file is created, it will be picked up by the next [sync_period](https://grafana.com/docs/alloy/latest/reference/components/local/local.file_match/#arguments) of the local.file_match component. But because the name of the file is the same as the previous one (the one that is cached), it won't propagate the update, resulting in the file not being tailed.

This problem does not happen on Unix systems because it will only stop the tailer when the local.file_match reports that the file is gone and in the meantime it will try to re-open it.

While we continue investigating the problem, a workaround would be to have a different file name for every log rotation.

### Steps to reproduce

Set up log rotation with Log4j, where the current log file is closed, compressed, renamed, and a new file is created.
Verify if Loki/Alloy scrapes logs from the new file after log rotation.
We’d appreciate your guidance on resolving this issue to ensure uninterrupted log scraping.

### System information

Windows

### Software version

Grafana alloy 1.5.1 (Note: 1.3.1 also had similar behaviour)

### Configuration

```text
local.file_match "logs_logs_data_APP_A_Logs" {
	path_targets = [{
		__address__ = "localhost",
		__path__    = "C:\\L*\\S*\\Tomcat*\\logs\\*.log",
		application = "APP_A",
		environment = "prod",
		role = "<<ROLE>>",
		hostname    = "<<HOSTNAME>>",
		job         = "APP_A_Tomcat_logs",
		log_type    = "application_logs",
	}]
}
	
	loki.process "logs_logs_data_APP_A_Logs" {
	forward_to = [loki.write.logs_logs_data.receiver]

	stage.match {
		selector = "{application=\"APP_A\"}"

		stage.regex {
			expression = ".*\\\\(?P<file>[^\\\\]+)$"
			source     = "filename"
		}

		stage.labels {
			values = {
				file = null,
			}
		}
	}

	stage.match {
		selector = "{file =~ \"(?i)LS2APP.LOG\"}"
		action   = "drop"
	}

	stage.multiline {
[...]

PR #2428, opened 2025-01-16
Use runner pkg in loki source file

### LINKED ISSUE #2345: Rework of loki.source.file
The current implementation of the loki.source.file has some problems:
- It does not scale very well because, on every target update, we stop all the readers and re-create them (see the comment in the code https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/file.go#L165).
- it does not handle file rotations on Windows if the file has the same name (see https://github.com/grafana/alloy/pull/2282)

I suggest reworking this component to use the runner pkg (like the loki.source.docker component) and adding a mechanism for the runners to periodically retry reading the file that they were assigned.

To improve it further, we could use a pool of workers to reduce the number of goroutines

wildum:
##### Why is a rework needed?

The current implementation of the loki.source.file has a bug on Windows: if you rotate a file by deleting it and recreating it with the same name, the file won't be tailed anymore unless a discovery component had time to update that the file was every missing (which does not usually happen because the sync period is usually much higher that the time it takes to rotate the file).

An attempt to fix this bug was done here: https://github.com/grafana/alloy/pull/2282, but the fix was not satisfying:
- it added complexity
- it added an argument that's only relevant for a Windows edge case

From there it was decided to rework the component to add a proper retry mechanism.

##### What are the changes?

Previously, all readers were stopped and deleted on every update. The tailing was also started in the Update function.
Now the readers are created unstarted in the Update function and the runner pkg that we use in other components is responsible for running them in the run function.

##### What are the gains from the rework?

- The log rotation file bug on Windows is fixed by the backoff retry that is handled by the runners. The test "TestDeleteRecreateFile" fails on Windows on main but passes with this solution.
- The code is easier to follow because it reuses a generic runner implementation. It's also easier to test because all the goroutines are now started and stopped via the Run function. Previously, if you created the component, it would call Update which would start a bunch of goroutines that would only be cleaned up in the Run function.
- CPU usage seems to be lower on heavy setups because the existing readers are kept between updates as long as the target is still there.
- Increased test coverage
- Cleaned up some old comments that don't apply anymore

##### What about the performances?

I ran locally in parallel two Alloy (latest main vs rework). They tailed 2000 files each containing a few lines of logs. Every 3 seconds 10 of these files were deleted and 10 new ones were created with different names.

The number of goroutines and the amount of memory are still the same:
- 12036 vs 12042 (that's a lot of goroutines, further work could be done to reduce it if that's a problem)
- 44MB vs 46MB

But in terms of CPU, there seems to be a significant improvement (50% according to the flamegraph):

Before

Screenshot 2025-01-16 at 12 13 17

After

Screenshot 2025-01-16 at 12 13 31

Fixes #2345
Fxise #2292

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


internal/component/loki/source/file/file.go
```diff
@@ -324,64 +287,51 @@ func (c *Component) startTailing(path string, labels model.LabelSet, handler lok
...
 		reader = tailer
 	}
 
 	return reader, nil
 }
 
-func (c *Component) reportSize(path, labels string) {
-	// Ask the reader to update the size if a reader exists, this keeps
-	// position and size metrics in sync.
-	if reader, ok := c.readers[positions.Entry{Path: path, Labels: labels}]; ok {
```
wildum, line 373: this logic was wrong because the readers were reset at the start of every update and because the readers were dedup before the size report so I just got rid of it

internal/component/loki/source/file/runner.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if r == otherTask {
+		return true
+	}
+
+	return r.readerHash == otherTask.readerHash
+}
+
+// runnerReader is a wrapper around a reader (tailer or decompressor)
+// It is responsible for running the runner. If the reader stops running,
```
thampiotr, line 35:
```suggestion
// It is responsible for running the reader. If the reader stops running,
```

internal/component/loki/source/file/tailer.go
```diff
@@ -1,6 +1,6 @@
 package file
 
-// This code is copied from loki/promtail@a8d5815510bd959a6dd8c176a5d9fd9bbfc8f8b5.
+// This code is adapted from loki/promtail@a8d5815510bd959a6dd8c176a5d9fd9bbfc8f8b5.
```
thampiotr, line 3: We occasionally grab the latest changes from `loki/promtail` because they sometimes fix bugs there. How do you envision this process will look now?

thampiotr (commented):
I would ideally like to see it running for some time in a dev cluster so we can verify there are no obvious issues with it. Do you think we could set something up? 🤔

internal/component/loki/source/file/tailer.go
wildum: That would not be a copy paste anymore, the change would need to be applied manually if it applies at all. But I think that this is unlikely to happen. Last change was 16 months ago and Promtail has been deprecated for some time (they document it as feature complete and recommend Alloy [here](https://grafana.com/docs/loki/latest/send-data/promtail/)). Having the comments still has some value to understand where most of the logic comes from but I expect the bug fixes to be done only in Alloy by now.

thampiotr: Cool. Maybe let's change wording to something like:
```
This code is adapted from loki/promtail. Last revision used to port changes to Alloy was a8d5815510bd959a6dd8c176a5d9fd9bbfc8f8b5.
```

thampiotr (approved):

internal/component/loki/source/file/tailer.go
```diff
@@ -180,26 +139,83 @@ func getLastLinePosition(path string) (int64, error) {
...
+		level.Error(t.logger).Log("msg", "failed to tail file", "path", t.path, "err", err)
+		return
+	}
+	pos, err := t.positions.Get(t.path, t.labelsStr)
+	if err != nil {
+		level.Error(t.logger).Log("msg", "failed to get file position", "err", err)
+		return
+	}
+
+	if fi.Size() < pos {
```
dehaansa, line 154: If a file is rotated but the new rotated file is larger than the previous, do we miss the first `pos` lines?

internal/component/loki/source/file/tailer.go
```diff
@@ -272,8 +286,8 @@ func (t *tailer) readLines() {
 	}
 }
 
-func (t *tailer) MarkPositionAndSize() error {
-	// Lock this update as there are 2 timers calling this routine, the sync in filetarget and the positions sync in this file.
+func (t *tailer) markPositionAndSize() error {
+	// Lock this update because it can be called in two different threads
```
dehaansa, line 290: nit
```suggestion
	// Lock this update because it can be called in two different goroutines
```

internal/component/loki/source/file/tailer.go
```diff
@@ -295,40 +309,35 @@ func (t *tailer) MarkPositionAndSize() error {
...
-	// stop can be called by two separate threads in filetarget, to avoid a panic closing channels more than once
-	// we wrap the stop in a sync.Once.
-	t.stopOnce.Do(func() {
-		// Save the current position before shutting down tailer
-		err := t.MarkPositionAndSize()
-		if err != nil {
-			level.Error(t.logger).Log("msg", "error marking file position when stopping tailer", "path", t.path, "error", err)
-		}
+	// Save the current position before shutting down tailer
+	err := t.markPositionAndSize()
```
dehaansa, line 319: Should we mark after we wait for `readLines` to complete?

internal/component/loki/source/file/decompresser.go
```diff
@@ -261,36 +274,31 @@ func (d *decompressor) readLines() {
...
 }
 
-func (d *decompressor) MarkPositionAndSize() error {
-	// Lock this update as there are 2 timers calling this routine, the sync in filetarget and the positions sync in this file.
+func (d *decompressor) markPositionAndSize() error {
+	// Lock this update because it can be called in two different threads
 	d.posAndSizeMtx.Lock()
 	defer d.posAndSizeMtx.Unlock()
 
 	d.metrics.totalBytes.WithLabelValues(d.path).Set(float64(d.size))
```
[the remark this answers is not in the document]
wildum: done

internal/component/loki/source/file/file.go
```diff
@@ -211,72 +200,47 @@ func (c *Component) Update(args component.Arguments) error {
...
 		}
 
-		c.readers[readersKey] = readerWithHandler{
-			reader:  reader,
-			handler: handler,
+		c.tasks[readersKey] = runnerTask{
+			reader: reader,
+			path:   path,
+			labels: labels.String(),
+			// Could fastFingerPrint work?
```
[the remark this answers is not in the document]
wildum: done

internal/component/loki/source/file/decompresser.go
```diff
@@ -261,36 +274,31 @@ func (d *decompressor) readLines() {
...
+	if err := d.markPositionAndSize(); err != nil {
+		level.Error(d.logger).Log("msg", "error marking file position when stopping decompressor", "path", d.path, "error", err)
+	}
 
-		// Wait for readLines() to consume all the remaining messages and exit when the channel is closed
-		<-d.done
-		level.Info(d.logger).Log("msg", "stopped decompressor", "path", d.path)
-		d.handler.Stop()
-	})
+	// Wait for readLines() to consume all the remaining messages and exit when the channel is closed
```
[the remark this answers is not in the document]
wildum: done

wildum:
@thampiotr @dehaansa while going over the review from Sam, I noticed that there were some race conditions between the Stop and the Run functions. To address it:
- I made sure that the Stop() function exits before exiting the Run function of the runner
- I protected the first part of the Run function and the first part of the Stop function with mutex and added a "stopping" boolean. This handles the scenario where the Stop function is called before the Run function (or in between two Run functions). When this happens, the Run function will check the stopping bool and exit early + cleanup to release the Stop function. 

I hope that the logic is still ok to follow. The added complexity is not ideal but all the different go routines that this component is spawning are not making the job easy. LMK what you think

internal/component/loki/source/file/decompresser.go
```diff
@@ -151,6 +154,29 @@ func mountReader(f *os.File, logger log.Logger, format CompressionFormat) (reade
 	return reader, nil
 }
 
+func (d *decompressor) Run() {
+	d.mut.Lock()
+
+	// Check if the stop function was called between two Run.
+	if d.stopping {
+		close(d.done)
```
dehaansa, line 162: Missing a mut.Unlock in this case

internal/component/loki/source/file/tailer.go
```diff
@@ -180,26 +143,99 @@ func getLastLinePosition(path string) (int64, error) {
 	}
 }
 
+func (t *tailer) Run() {
+	t.mut.Lock()
+
+	// Check if the stop function was called between two Run.
+	if t.stopping {
+		close(t.done)
```
dehaansa, line 151: Missing a mut.Unlock()

internal/component/loki/source/file/decompresser.go
wildum: good catch!

dehaansa (approved):

internal/component/loki/source/file/decompresser.go
```diff
@@ -256,41 +291,51 @@ func (d *decompressor) readLines() {
...
 }
 
 func (d *decompressor) Stop() {
-	// stop can be called by two separate threads in filetarget, to avoid a panic closing channels more than once
-	// we wrap the stop in a sync.Once.
-	d.stopOnce.Do(func() {
-		// Shut down the position marker thread
-		close(d.posquit)
-		<-d.posdone
+	d.mut.RLock()
```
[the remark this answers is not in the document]
wildum: changed here and in the tailer

PR #2431, opened 2025-01-16
feat(pyroscope.receive_http): Support pushv1.Push in receive_http

### LINKED ISSUE #2301: pyroscope.receive_http should support the connect go handlers
### Background

Pyroscope receive_http only support the `/ingest` API, which is confusing. We should promote the use of the connect-go handlers and there fore provide a pushv1.PushV1.

Currently it is not possible for example to make `pyroscope.ebpf` of many agents send to `pyroscope.receive_http` on a single gateway agent,  as this only uses the `push.v1.PusherService/Push`.

### Proposal

Support connect-go on the reciever_http socket.

simonswine:
The Pyroscope backend has currently two ingest API:
* /ingest, which is predominatnely used by the Pyroscope SDKs
* /push.v1.PusherService/Push which is a connect API used by profilecli and pyroscope.write with pyroscope.ebpf and pyroscope.alloy

This adds the missing support for `/push.v1.PusherService/Push`

Fixes #2301

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Main (unreleased)
 
 - Bump snmp_exporter and embedded modules to 0.27.0. Add support for multi-module handling by comma separation and expose argument to increase SNMP polling concurrency for `prometheus.exporter.snmp`. (@v-zhuravlev)
 
+- Add support for pushv1.PusherService Connect API in `pyroscope.receive_http`. (@simonswine)
+
 v1.6.1
 -----------------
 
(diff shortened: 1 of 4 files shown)


marcsanmi (approved):
LGTM 👍 
I think we should update the `pyroscope.receive_http.md` docs, which were originally written to explicitly cover HTTP ingestion API support.

I also believe `CHANGELOG.md` needs to be updated.

clayton-cornell (approved):

PR #2436, opened 2025-01-17
Add tests for types other than integers

ptodev:


internal/runtime/internal/testcomponents/stringer.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+}
+
+type StringerConfig struct {
+	InputString *string          `alloy:"input_string,attr,optional"`
+	InputInt    *int             `alloy:"input_int,attr,optional"`
+	InputFloat  *float64         `alloy:"input_float,attr,optional"`
+	InputBool   *bool            `alloy:"input_bool,attr,optional"`
+	InputMap    *map[string]any  `alloy:"input_map,attr,optional"`
+	InputArray  *[]any           `alloy:"input_array,attr,optional"`
```
wildum, line 31: would it not work to just have one attribute "any"?

internal/runtime/testdata/foreach/foreach_9.txtar
```diff
@@ -0,0 +1,19 @@
+A collection containing maps.
+
+-- main.alloy --
+foreach "testForeach" {
+  collection = [{"a" = 4}, {"a" = 6}]
+  var = "num"
+
+  template {
+    testcomponents.pulse "pt" {
+      max = num["a"]
```
wildum, line 10: this is really cool :) could you also add a test with objects where you access an attribute of the object? In the other foreach tests we are already passing an object (the receivers) but we don't try to access the values of the object. (the stringer component would need to support objects)

internal/runtime/internal/testcomponents/logger.go
```diff
@@ -0,0 +1,91 @@
...
+type StringReceiverComponent struct {
+	opts component.Options
+	log  log.Logger
+
+	mut      sync.Mutex
+	recvStr  string
+	receiver StringReceiver
+}
+
+// NewStringReceiver creates a new summation component.
```
wildum, line 53:
```suggestion
// NewStringReceiver creates a new string_receiver component.
```

internal/runtime/internal/testcomponents/logger.go
```diff
@@ -0,0 +1,91 @@
...
+	_ component.Component = (*StringReceiverComponent)(nil)
+)
+
+// Run implements Component.
+func (s *StringReceiverComponent) Run(ctx context.Context) error {
+	<-ctx.Done()
+	return nil
+}
+
+// Return the sum as debug info instead of export to avoid evaluation loop.
```
wildum, line 81:
```suggestion
// Return the receiver as debug info instead of export to avoid evaluation loop.
```

internal/runtime/internal/testcomponents/logger.go
```diff
@@ -0,0 +1,91 @@
+package testcomponents
+
```
wildum, line 2: I would prefer this file to be called string_receiver.go to match the name of the component. It makes it easier when searching through the code

wildum (commented):
nice one, good catch on the float and the tests names, this will make the feature much more robust

internal/runtime/internal/testcomponents/stringer.go
ptodev: It would, but I like this approach since it's more clear what we want to test.

internal/runtime/testdata/foreach/foreach_9.txtar
ptodev: > the stringer component would need to support objects

Objects are just maps, and `stringer` already has an `input_map` argument.

> could you also add a test with objects where you access an attribute of the object?

Do you mean a test which accesses something like `num["a"]["b"]`? I'm not sure what the "attribute" of an object is. [Objects](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/#objects) just have keys and values.

wildum: So if I have a component "testcomponent" that exports the object A, with an attribute b, I pass it to the collection like this:
collection = [testcomponent.A]
var = "item"

then to access b, I can do `item["b"]`, is this correct? And if `b` was an object with an attribute c, then I could access it with `item["b"]["c"]`? I assume this is how it works, maybe it's still worth it to have a test for this?

ptodev: No worries, I added one just now where we access `num["a"]["c"]`.

wildum (approved):

PR #2443, opened 2025-01-17
database_observability: Fix handling of view table types when detecting schema.

matthewnolf:
#### PR Description
MySQL returns different column count for `SHOW CREATE VIEW` (2 columns) when compared to `SHOW CREATE TABLE` (4 columns). 
This change adds conditional logic to handle both cases so that we can extract the schema of a view without generating an error.

There are a couple small refactors:
- Log the table name when any operation related to schema detection fails, so that we can isolate the source of the issue
- Introduce sub tests in the `schema_table_test.go` for the different scenarios

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -225,15 +232,24 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
 
 		row := c.dbConnection.QueryRowContext(ctx, showCreateTable+" "+fullyQualifiedTable)
 		if row.Err() != nil {
-			level.Error(c.logger).Log("msg", "failed to show create table", "err", row.Err())
+			level.Error(c.logger).Log("msg", "failed to show create table", "err", row.Err(), "table", table.tableName)
```
cristiangreco, line 235: Nit: could we log `err` as last value?

matthewnolf: Updated log messages

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -195,19 +196,25 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
 				break
 			}
 
-			var table string
+			var tableName, tableType string
 			var createTime, updateTime time.Time
-			if err := rs.Scan(&table, &createTime, &updateTime); err != nil {
+			if err := rs.Scan(&tableName, &tableType, &createTime, &updateTime); err != nil {
```
cristiangreco, line 201: I'd expect the `selectTableName` query to be updated to fetch table type?

matthewnolf: Seems like a victim of resolving conflicts. Let me look thanks

cristiangreco (approved):

PR #2474, opened 2025-01-22
Optimise Prometheus targets handling

### LINKED ISSUE #1999: Reduce resource consumption of discovery components
### Address high allocs in creation of targets lists

Targets/Discovery are represented as `list(map[string])`, with discovery running every 5 seconds in very large clusters the garbage collector can spike due to creating these labels. The names especially are repeated. The hard part is implementing the capsule value so hard coded targets still work. This is because in the alloy runtime there is a bug where you cannot convert from list/map to a capsule value.

In reviewing profiles this code was by far the largest allocator.

https://github.com/grafana/alloy/blob/5845b2ccacf65532b920406f5acae539dbfa8e00/internal/component/discovery/discovery.go#L212

### Use case

Reduce memory and cpu consumption

thampiotr:
#### PR Description

High level overview: This changes Targets to a capsule instead of a Go map in a way that is transparent to the users, but allows us to optimise the allocation and memory footprint of the targets handling in Prometheus pipelines.

NOTE: this PR appears large, however:
- 1k is a copy of tests from upstream Prometheus for extra validation
- 1.5k is new tests to ensure Target/TargetBuilder/equality work as expected
- A lot of lines are changing the use of Targets throughout the codebase, in particular, various ways they're constructed in tests.

Fixes #1999

#### Notes to the Reviewer

Changes in more detail to facilitate the review:
- Implemented a new Target which uses the underlying Prometheus LabelSets we get from discovery: the shared labels in `group` and target's own labels in `own` fields. This allows to avoid a lot of allocation. 
- Because two targets can have the same labels but they are split differently between their `own` and `group` fields, we need to use a dedicated `Equals` function. Implementing it is easy, but we also need to make sure that it is used by Alloy runtime (see below).
- To allow customising Equals function in the runtime, I added the `equality` package which will attempt to compare using our `CustomEquality` interface where possible, but if that fails, it will use the previous reflect.DeepEquals. This package is now used throughout the runtime to avoid unnecessary updates etc.
- Targets implement the necessary ConvertFrom and ConvertInto methods so the users still see them as maps when working with the UI and Alloy config files. The change should be fully transparent to the users. There are tests to verify this.
- To implement ConvertFrom and ConvertInto we had to expose a couple of internal functions from the syntax package, but that's minimal and I think safe.
- However, when encoding the capsules into Alloy syntax or into JSON (which is used in UI), the syntax package wasn't checking if given capsule can be converted in to an Alloy object and printed as such. This functionality was added - if we see a capsule that implements `ConvertibleIntoCapsule` - we know it can be converted into an object (which is a map), so we do that and we print it as such. 
- Targets are now immutable and changing them is done using TargetBuilder. This in turn uses similar method to Prometheus Labels to reduce allocation when adding or deleting labels.
- Converting targets to the format required by Prometheus scrape is now faster, because we already use the same underlying structures. We further improve performance by reconstructing target groups where shared labels from `group` are passed on a group level. This reconstruction of target groups is done in `discovery.ComponentTargetsToPromTargetGroups`. We group them using the `group` labelset hash, but we also handle hash conflicts in case they are detected.
- We have already some copy of Prometheus relabel code. I added the actual relabelling code to this so we can use our own, optimised label builder implementation. Also ported the tests to ensure we are 100% compliant.
- Changed the distributed targets hashing to directly calculate hash instead of allocating labels first and then grabbing their hash.
- Updated all the tests to use the new Targets and also the correct equality checks on them.

#### Performance:

Deployed to a large cluster between 10:30 and 13:00, we observed the following impact:

Memory and CPU down, we can also see smaller variation in memory usage which is good for HPA.
image

Allocation profile before:
image

Allocation profile after:
image

Looking at the diff, the reduction is, as expected, mostly in these two functions:
image

image

We can see the overall memory usage went down and back up as the changed was rolled out and then rolled back:
image

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/discovery/target_test.go
```diff
@@ -0,0 +1,866 @@
...
+
+goos: darwin goarch: arm64 cpu: Apple M2
+Benchmark_Targets_TypicalPipeline-8   	      36	  32868159 ns/op	 6022494 B/op	  100544 allocs/op
+Benchmark_Targets_TypicalPipeline-8   	      34	  34562724 ns/op	 6109322 B/op	  100543 allocs/op
+Benchmark_Targets_TypicalPipeline-8   	      34	  35662420 ns/op	 6022429 B/op	  100545 allocs/op
+Benchmark_Targets_TypicalPipeline-8   	      36	  33446308 ns/op	 6021909 B/op	  100541 allocs/op
+Benchmark_Targets_TypicalPipeline-8   	      34	  33537419 ns/op	 6022333 B/op	  100543 allocs/op
+Benchmark_Targets_TypicalPipeline-8   	      34	  33687083 ns/op	 6109172 B/op	  100543 allocs/op
+*/
+func Benchmark_Targets_TypicalPipeline(b *testing.B) {
```
thampiotr, line 771: Compared to current `main` we see:

```
benchstat ./01_typical_pipeline_targets_as_maps.txt  ./08_typical_pipeline_obj_pools.txt
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/component/discovery
cpu: Apple M2
                           │ ./01_typical_pipeline_targets_as_maps.txt │  ./08_typical_pipeline_obj_pools.txt  │
                           │                  sec/op                   │   sec/op     vs base                  │
_Targets_TypicalPipeline-8                                 38.63m ± 1%   33.61m ± 6%  -12.98% (p=0.000 n=10+6)

                           │ ./01_typical_pipeline_targets_as_maps.txt │  ./08_typical_pipeline_obj_pools.txt   │
                           │                   B/op                    │     B/op      vs base                  │
_Targets_TypicalPipeline-8                               61.953Mi ± 0%   5.743Mi ± 1%  -90.73% (p=0.000 n=10+6)

                           │ ./01_typical_pipeline_targets_as_maps.txt │  ./08_typical_pipeline_obj_pools.txt  │
                           │                 allocs/op                 │  allocs/op   vs base                  │
_Targets_TypicalPipeline-8                                 241.8k ± 0%   100.5k ± 0%  -58.42% (p=0.000 n=10+6)
```

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -273,17 +273,17 @@ func (c *Component) Update(args component.Arguments) error {
 func (c *Component) baseTarget() (discovery.Target, error) {
 	data, err := c.opts.GetServiceData(http_service.ServiceName)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get HTTP information: %w", err)
```
rafaelroquetto, line 276: @marctc you may want to have a look here

internal/component/discovery/relabel/relabel.go
```diff
@@ -82,24 +79,23 @@ func (c *Component) Run(ctx context.Context) error {
...
 	targets := make([]discovery.Target, 0, len(newArgs.Targets))
-	relabelConfigs := alloy_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelConfigs)
-	c.rcs = relabelConfigs
 
 	for _, t := range newArgs.Targets {
-		lset := componentMapToPromLabels(t)
-		relabelled, keep := relabel.Process(lset, relabelConfigs...)
+		var (
+			relabelled discovery.Target
+			builder    = discovery.NewTargetBuilderFrom(t)
```
simonswine, line 89: This builder should/could be reused to avoid extra allocations

internal/component/discovery/target_builder.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		for k, v := range t.toAdd {
+			newOwn[commonlabels.LabelName(k)] = commonlabels.LabelValue(v)
+		}
+	}
+	if modifyGroup {
+		// TODO(thampiotr): When relabeling a lot of targets that require changes to t.group, we might produce a lot of
+		//  				t.groups that will be essentially the same. If this becomes a hot spot, it could be
+		//  				remediated with an extra step to consolidate them using perhaps a hash as an ID.
+		newGroup = make(commonlabels.LabelSet, len(t.group))
```
wildum, line 187: The group label set is referencing the same data across the targets, that's why it must be immutable and that's why the target builder has a toAdd and toDel, right? But since `own` is unique for every target, does it need to be immutable? (could labels be added and deleted directly in it?)
Also with the current implementation, creating a new group is the same as adding the labels in `own` because the new group would be unique for the target?

thampiotr: I think it's much better to keep targets immutable: we no longer need to copy them around... we no longer need to worry that some component will accidentally mutate the underlying targets that it's given and runtime caching the targets is also safer. Using builder is always possible so we're not stopping users from changing targets deliberately.  

> Also with the current implementation, creating a new group ...

From user of Target's point of view, they should not worry about group and own in most cases - they can use the Target as a set of labels. In the implementation, from performance point of view, we delete from the `t.group` only, because adding can be done in `t.own`. And we prefer to not modify t.group as much as possible so that it can remain being the same underlying slice array used across all targets.

I have another idea that could still do better and never modify `t.group` - instead of deleting from `t.group` we could add an empty label to `t.own` which signifies delete (this is already handled as such because Prometheus supports these semantics). This way the `t.group` will never be changing and the same slice can be used for the lifetime of a Target, shared across all the targets that came from discovery component. 

I can implement this idea as a follow-up. Would need to re-test and re-benchmark to be sure, so I would prefer to leave it as a follow-up. 

In general, after this PR we can probably do more optimisations when necessary, but we need to collect the data on how this performs at scale over longer period of time first.

internal/component/common/relabel/relabel.go
```diff
@@ -149,7 +167,13 @@ func (rc *Config) Validate() error {
...
 		return fmt.Errorf("relabel configuration for %s action requires 'target_label' value", rc.Action)
 	}
-	if (rc.Action == Replace || rc.Action == Lowercase || rc.Action == Uppercase || rc.Action == KeepEqual || rc.Action == DropEqual) && !relabelTarget.MatchString(rc.TargetLabel) {
+	if rc.Action == Replace && !strings.Contains(rc.TargetLabel, "$") && !model.LabelName(rc.TargetLabel).IsValid() {
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+	}
+	if rc.Action == Replace && strings.Contains(rc.TargetLabel, "$") && !relabelTarget.MatchString(rc.TargetLabel) {
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+	}
+	if (rc.Action == Lowercase || rc.Action == Uppercase || rc.Action == KeepEqual || rc.Action == DropEqual) && !model.LabelName(rc.TargetLabel).IsValid() {
```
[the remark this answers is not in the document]
mattdurham: Probably due to var expansion? Wonder if we can add a comment to the why.

internal/component/common/relabel/relabel.go
```diff
@@ -186,6 +210,97 @@ func (rc *Config) Validate() error {
 	return nil
 }
 
+// ProcessBuilder is like Process, but the caller passes a labels.Builder
+// containing the initial set of labels, which is mutated by the rules.
+func ProcessBuilder(lb LabelBuilder, cfgs ...*Config) (keep bool) {
```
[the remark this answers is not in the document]
mattdurham: Can we workshop this comment a bit, the initial set of labels that are used to builder the label builder arent mutated but the LabelBuilder itself.

internal/component/discovery/relabel/relabel.go
mattdurham: Not sure it should be since the content of the builder is mutated on each call to ProcessBuilder, if we can make it safe then seems reasonable.

internal/component/discovery/target.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (t Target) hashLabelsInOrder(order []string) uint64 {
+	// This optimisation is adapted from prometheus/model/labels.
+	// Use xxhash.Sum64(b) for fast path as it's faster.
+	b := make([]byte, 0, 1024)
+	mustGet := func(label string) string {
+		val, ok := t.Get(label)
+		if !ok { // should never happen as Target is immutable
+			panic("label concurrently modified - this is a bug - please report an issue")
```
mattdurham, line 292: We should not panic but instead return an error

internal/component/discovery/target.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+		return val
+	}
+
+	for i, key := range order {
+		value := mustGet(key)
+		if len(b)+len(key)+len(value)+2 >= cap(b) {
```
mattdurham, line 299: I assume the +2 is from seps?

internal/component/discovery/target.go
```diff
@@ -0,0 +1,370 @@
...
+			// If labels entry is 1KB+ do not allocate whole entry.
+			h := xxhash.New()
+			_, _ = h.Write(b)
+			for _, key := range order[i:] {
+				_, _ = h.WriteString(key)
+				_, _ = h.Write(seps)
+				_, _ = h.WriteString(mustGet(key))
```
mattdurham, line 306: Not sure I understand this inner loop. If a the key value pair puts itl onger than 1024 bytes then switch to slower loop not using the pre allocated buffer.

internal/component/discovery/target.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// Sort by hash to get deterministic order
+	sortedKeys := maps.Keys(targetIndWithCommonGroupLabels)
+	slices.Sort(sortedKeys)
+
+	allGroups := make([]*targetgroup.Group, 0, len(targetIndWithCommonGroupLabels))
+	var hashConflicts []commonlabels.LabelSet
+	for _, hash := range sortedKeys {
+		targetIndices := targetIndWithCommonGroupLabels[hash]
+		sharedLabels := tgs[targetIndices[0]].group // all should have the same group labels, so grab first one
```
mattdurham, line 341: Is this enforced by the caller? If so since its an exported method this feels like it might not be always true.

internal/component/discovery/target.go
```diff
@@ -0,0 +1,370 @@
...
+			if !labelSetEqualsFn(sharedLabels, target.group) {
+				hashConflicts = append(hashConflicts, target.LabelSet())
+				continue
+			}
+			individualLabels = append(individualLabels, target.own)
+		}
+
+		if len(individualLabels) != 0 {
+			allGroups = append(allGroups, &targetgroup.Group{
+				Source:  fmt.Sprintf("%s_part_%v", jobName, hash),
```
mattdurham, line 355: Is this a magic label or carry over from something?

internal/component/discovery/target_builder.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	commonlabels "github.com/prometheus/common/model"
+
+	"github.com/grafana/alloy/internal/component/common/relabel"
+)
+
+type TargetBuilder interface {
+	relabel.LabelBuilder
+	Target() Target
+	SetKV(kv ...string) TargetBuilder
```
mattdurham, line 12: SetKV name seems odd, since I think of key value but neither the input or exposed property is inherently a KV. It returns a TargetBuilder, `SetTargetBuilder`, `CopyTargetBuilder`?

internal/component/common/relabel/label_builder.go
```diff
@@ -0,0 +1,9 @@
+package relabel
+
+// LabelBuilder is an interface that can be used to change labels with relabel logic.
+type LabelBuilder interface {
+	Get(label string) string
```
mattdurham, line 5: Should we add a comment that Get called with an unfound value returns a blank string? Or change signature to include found?

internal/component/common/relabel/label_builder.go
```diff
@@ -0,0 +1,9 @@
...
+	Range(f func(label string, value string))
+	Set(label string, val string)
```
mattdurham, line 7: Should we add that setting to '''' is the same as deleting?

internal/component/discovery/target_builder.go
mattdurham: Very much agree making things immutable makes things easier!

internal/component/prometheus/scrape/scrape.go
```diff
@@ -479,34 +479,27 @@ func (c *Component) DebugInfo() interface{} {
...
+	for _, tgs := range groups {
+		for _, tg := range tgs {
+			promTargets, errs := scrape.TargetsFromGroup(
+				tg,
+				getPromScrapeConfigs(jobName, args),
+				false,                                /* noDefaultScrapePort - always false in this component */
+				make([]*scrape.Target, len(targets)), /* targets slice to reuse */
+				lb,
+			)
+			lb.Reset(labels.EmptyLabels())
```
mattdurham, line 496: How expensive would it be to recreate a new one each time? Your depending on the vendored code not doing anything with the label builder passed in, which currently it doesnt

internal/runtime/equality/equality.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if v1.Type() != v2.Type() {
+		return couldNotCompare
+	}
+
+	pointerOrStruct := v1.Type().Kind() == reflect.Ptr || v1.Type().Kind() == reflect.Struct
+	if pointerOrStruct && v1.CanInterface() && v1.Type().Implements(customEqualityType) {
+		if v2Ptr := getAddr(v2); v2Ptr.CanInterface() {
+			return successfulCompare(v1.Interface().(CustomEquality).Equals(v2Ptr.Interface()))
```
mattdurham, line 61: Do we need to add a new code path for the other direction, ie v2 implements CustomEquality?

mattdurham (changes requested):
Some questions but overall very solid! Fantastic gains.

internal/component/discovery/target.go
thampiotr: Target labels are immutable and the only way that labels passed in `order` are no longer there is someone introducing a bug. I've removed the panic to not have crashes if this ever happened, but instead of returning an error, in order not to overcomplicate the API, I'm hashing an empty string - it will still provide consistent hashing behaviour and is a better choice than complicating things with error handling IMO.

internal/component/discovery/target.go
thampiotr: Yep. All of this code comes from Prometheus: https://github.com/grafana/prometheus/blob/staleness_disabling_v0.54.1/model/labels/labels.go#L75

And we must match it because we don't want the rollout to rehash all targets differently. There are tests to guarantee this too.

internal/component/discovery/target.go
thampiotr: It does use the pre-allocated buffer, it writes its content first in line `_, _ = h.Write(b)` and then continues with the remaining items to hash. 

Like above, all of this code comes from Prometheus: https://github.com/grafana/prometheus/blob/staleness_disabling_v0.54.1/model/labels/labels.go#L75
and it must match it

internal/component/discovery/relabel/relabel.go
thampiotr: Since the current profiles show good performance gains here already, I would like to leave this optimisation for future PRs after we have more data. I would like to consider an option of using an object pool for the couple of maps that the builder uses underneath as an alternative and compare the two options.

thampiotr (commented):
Addressed the comments, PTAL

internal/component/discovery/target.go
mattdurham: The the funcs/files that come from Prometheus can we document that and that it cant be changed for backwards compat?

mattdurham (approved):
Looks good, great work. My only remaining ask is to document the places we cant change the code due to backwards compatibility with prometheus.

PR #2519, opened 2025-01-23
Update using code format in headings and tables

clayton-cornell:
#### PR Description

This PR updates the beyla and discovery component topics to use code formatting in headings and tables. The goal is to clean up the linting errors and make sure that the topic contents are closer to being semantically correct.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/beyla/beyla.ebpf.md
+++ b/docs/sources/reference/components/beyla/beyla.ebpf.md
@@ -56,8 +56,8 @@ Block                                                                  | Descrip
 [`attributes`][attributes]                                             | Configures the Beyla attributes for the component.                                                 | no
 `attributes` > [`kubernetes`][kubernetes attributes]                   | Configures decorating of the metrics and traces with Kubernetes metadata of the instrumented Pods. | no
 [`discovery`][discovery]                                               | Configures the discovery for instrumentable processes matching a given criteria.                   | no
-`discovery` > [`exclude services`][services]                           | Configures the services to exclude for the component.                                              | no
-`discovery` > `exclude services` > [`kubernetes`][kubernetes services] | Configures the Kubernetes services to exclude for the component.                                   | no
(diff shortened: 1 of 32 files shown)


clayton-cornell:
Missing info in belya.ebpf docs tracks in https://github.com/grafana/alloy/issues/2533

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -83,13 +83,13 @@ You must specify the `output` block, but all its arguments are optional.
...
 
-### attributes
+### `attributes`
 
 The `attributes` block configures how some attributes for metrics and traces are decorated.
 
 It contains the following block:
 
-#### kubernetes attributes
+#### `kubernetes` attributes
```
ptodev, line 92: Wouldn't this be more clear if it was something like `attributes` > `kubernetes`? I think I had a proposal on this a while ago.

ptodev (approved):

docs/sources/reference/components/beyla/beyla.ebpf.md
clayton-cornell: Yeah those have been "bothering" me. I left it as-is in this iteration but... I'd definitely like to update it. I'll push a quick change today to switch the order and use the hierarchy style... if it looks OK, I will merge.

clayton-cornell: And now after trying it out, formatting it that way only works if the hierarchy is clear and child elements are always child elements. Formatting this way fails (at least in the current topic layout) when you have blocks like:
```
oauth2
oauth2 > tls_config
tls_config
```
The description of `tls_config` is the same as a top level block and as a child of `oauth2`.  Using the hierarchy in the headings doesn't work in these situations.

I'll leave as-is for this iteration and we can think about other ways to structure the headings in a future edit. I'll add an issue for this so I don't forget.

PR #2522, opened 2025-01-24
GitHub Actions steps to build Alloy for every pull request

ptodev:
This replicates the Drone builds. The Drone builds will be turned off in a follow-up PR.

In a follow up PR I also intend to enable caching for Go modules. I haven't enabled it yet, because there might be a few issues with it that lead to runners going out of disk space. I will need to test the caching functionality more thoroughly before enabling it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,128 @@
+name: Build
+on:
+  pull_request:
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
+  # We group both by ref_name (available when CI is triggered by a push to a branch/tag)
+  # and head_ref (available when CI is triggered by a PR).
+  group: "${{ github.ref_name }}-${{ github.head_ref }}"
+  cancel-in-progress: true
+jobs:
+  build_linux:
+    name: Build on Linux
+    runs-on: ubuntu-latest
+    container: grafana/alloy-build-image:v0.1.8
+    strategy:
+      matrix:
+        os: [linux]
+        arch: [amd64, arm64, ppc64le, s390x]
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,128 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+        # TODO: Try enabling caching later. It might use up too much disk space on runners so needs extra testing.
+        cache: false
+    - run: make generate-ui
+    - run: GO_TAGS="builtinassets promtail_journal_enabled" GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} GOARM= GOEXPERIMENT=boringcrypto make alloy
+
+  build_mac_intel:
```
ptodev, line 61: We are not running the Mac builds using the `grafana/alloy-build-image` Docker image, because GitHub Actions doesn't support running inside containers for Mac. The build completes ok, so I suppose that's fine.

.github/workflows/build.yml
```diff
@@ -0,0 +1,128 @@
...
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+        # TODO: Try enabling caching later. It might use up too much disk space on runners so needs extra testing.
+        cache: false
+    - run: make generate-ui
+    - run: GO_TAGS="builtinassets" GOOS=darwin GOARCH=arm64 GOARM= make alloy
+
+  build_windows:
+    name: Build on Windows (AMD64)
```
ptodev, line 92: We are not running the Windows builds using the `grafana/alloy-build-image` Docker image, because GitHub Actions doesn't support running inside containers for Windows. The build completes ok, so I suppose that's fine.

mattdurham (approved):
LGTM!

PR #2528, opened 2025-01-24
Run all unit tests on the main branch in GitHub actions.

ptodev:
A GitHub action which runs `make test`. The corresponding Drone pipeline can be removed later on.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -30,7 +30,6 @@ trigger:
   - pull_request
 type: docker
 ---
-# TODO: Remove this pipeline once its GitHub actions equivalent is battle tested.
 kind: pipeline
 name: Test (Full)
 platform:
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -30,6 +30,7 @@ trigger:
   - pull_request
 type: docker
 ---
+# TODO: Remove this pipeline once its GitHub actions equivalent is battle tested.
 kind: pipeline
 name: Test (Full)
 platform:
--- a/.github/workflows/test_full.yml
+++ b/.github/workflows/test_full.yml
@@ -1,0 +1,30 @@
+name: Test (Full)
+on:
+  push:
+    branches:
+      - main
+  # TODO: Remove this later
+  pull_request:
+jobs:
+  test_linux:
+    name: Test (Full)
+    runs-on: ubuntu-latest
+    container:
+      image: grafana/alloy-build-image:v0.1.8
+      volumes:
+        - /var/run/docker.sock
+    steps:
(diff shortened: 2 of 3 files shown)


.github/workflows/test_full.yml
```diff
@@ -0,0 +1,30 @@
...
+      # https://github.com/actions/runner/issues/2033#issuecomment-1204205989
+      run: |
+          # this is to fix GIT not liking owner of the checkout dir
+          chown -R $(id -u):$(id -g) $PWD
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+        # TODO: Enable caching later.
+        cache: false
```
wildum, line 29: why not enable it now?

ptodev: While I was testing these changes, the runners ran out of disk space. Then I disabled the caching and they worked fine.

Caches tend to be specific to branches and workflows... there are also race conditions if steps within a workflow run concurrently. For example:
https://github.com/actions/setup-go/issues/358
https://github.com/actions/setup-go/pull/416/files

OTel doesn't use the setup-go cache, instead they use the generic caching action. They run it at the start of the workflow and all other workflow steps wait for it to complete before running. This looks like a better approach, but I'd like us to test it first.

For such an essential and widely used feature, it appears really complicated to configure it the right way.

.drone/drone.yml
```diff
@@ -132,6 +132,7 @@ trigger:
   - pull_request
 type: docker
 ---
+# TODO: Remove this pipeline once its GitHub actions equivalent is battle tested.
```
ptodev, line 135:
```suggestion
```

We will need to remove this line. If we edit drone.yml without running `make drone` to generate it from the jsonnet, Drone will refuse to run the pipelines automatically. We'd have to manually click a "run" button each time a PR is opened (see #2557).

.github/workflows/test_full.yml
```diff
@@ -0,0 +1,30 @@
+name: Test (Full)
+on:
+  push:
+    branches:
+      - main
+  # TODO: Remove this later
+  pull_request:
```
ptodev, line 7:
```suggestion
```

This was added just for testing. The drone build doesn't run on a PR.

.github/workflows/test_full.yml
wildum: we use it for the integration tests and I think that so far it worked well but maybe it was not a problem because we had less workflow. I'm happy to leave it disabled for now but we should keep it in the radar because 20mins to run the tests is a lot

wildum (approved):

PR #2530, opened 2025-01-24
database_observability: make collectors configurable

fridgepoet:
#### PR Description
This enables users to select which components they want to run in the database observability component, similar to how mysqld_exporter works.

Fixes https://github.com/grafana/grafana-dbo11y-app/issues/370

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -23,17 +23,30 @@ database_observability.mysql "<LABEL>" {
 
 You can use the following arguments with `database_observability.mysql`:
 
-Name                    | Type                 | Description                                              | Default | Required
-------------------------|----------------------|----------------------------------------------------------|---------|---------
-`data_source_name`      | `secret`             | [Data Source Name][] for the MySQL server to connect to. |         | yes
-`forward_to`            | `list(LogsReceiver)` | Where to forward log entries after processing.           |         | yes
-`collect_interval`      | `duration`           | How frequently to collect information from database.     | `"10s"` | no
-`query_samples_enabled` | `bool`               | Whether to enable collection of query samples.           | `true`  | no
+ Name                 | Type                 | Description                                                   | Default | Required
```
fridgepoet, line 26: Neat preview feature. See changes here https://deploy-preview-alloy-2530-zb444pucvq-vp.a.run.app/docs/alloy/latest/reference/components/database_observability/database_observability.mysql/

cristiangreco: Nice!

internal/component/database_observability/mysql/component.go
```diff
@@ -206,6 +208,33 @@ func (c *Component) Update(args component.Arguments) error {
...
+		schemaTable: true,
+	}
+
+	for _, disabled := range a.DisableCollectors {
+		for c := range collectors {
+			if c == disabled {
+				collectors[c] = false
+				break
+			}
+		}
```
cristiangreco, line 224: Instead of looping, can we check directly with?

```
		if _, ok := collectors[disabled]; ok {
			collectors[disabled] = false
		}
```

internal/component/database_observability/mysql/component.go
```diff
@@ -29,7 +29,11 @@ import (
 	"github.com/grafana/alloy/syntax/alloytypes"
 )
 
-const name = "database_observability.mysql"
+const (
+	name        = "database_observability.mysql"
+	querySample = "query_sample"
+	schemaTable = "schema_table"
```
cristiangreco, line 35: The collectors have a `Name()` method that returns a similar string. Maybe we could put that into a constant and reuse it from here? (the format is `SchemaTable` rather than `schema_table`, I'm happy with underscores)

internal/component/database_observability/mysql/component.go
fridgepoet: good point, thanks

internal/component/database_observability/mysql/component.go
fridgepoet: good idea, thanks!

cristiangreco (approved):

mattdurham (approved):
Looks good, will merge once @clayton-cornell  reviews.

matthewnolf (approved):
Great work!

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -23,17 +23,30 @@ database_observability.mysql "<LABEL>" {
...
+### Supported Collectors
+
+The full list of supported collectors is:
+
+| Name         | Description                                           | Enabled by default |
+|--------------|-------------------------------------------------------|--------------------|
+| query_sample | Collect query samples.                                | yes                |
+| schema_table | Collect schemas and tables from `information_schema`. | yes                |
+
+
```
clayton-cornell, line 49:
```suggestion
The following collectors are enabled by default:

* `query_sample`: Collect query samples.
* `schema_table`: Collect schemas and tables from `information_schema`.

```

I would suggest this is better represented as a bullet list.

I would also drop the heading and just move it up under the table around line 35-36.

clayton-cornell, line 49:
```suggestion

The following collectors are enabled by default:

* `query_sample`: Collect query samples.
* `schema_table`: Collect schemas and tables from `information_schema`.

[collectors]: #supported-collectors

## Blocks

The `database_observability.mysql` component doesn't support any blocks. You can configure this component with arguments.

```

This info is probably better as a bullet list, and should be below the Args table, not a child of `Blocks`.

fridgepoet: Thanks for reviewing, @clayton-cornell! I agree it was a mistake to have it as a child of Blocks.
I was trying to model the page a little bit like the mysql exporter in prometheus [here](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.mysql/).

 What do you think about the following:
* keeping it as a table to match the mysql exporter docs and because we will add more collectors in the future.
* Supported Collectors' is correctly moved to a child of Args

fridgepoet: No worries, perfect, thank you!

clayton-cornell (approved):

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -23,12 +23,22 @@ database_observability.mysql "<LABEL>" {
...
-`collect_interval`      | `duration`           | How frequently to collect information from database.     | `"10s"` | no
-`query_samples_enabled` | `bool`               | Whether to enable collection of query samples.           | `true`  | no
+ Name                 | Type                 | Description                                                   | Default | Required 
+----------------------|----------------------|---------------------------------------------------------------|---------|----------
+ `data_source_name`   | `secret`             | [Data Source Name][] for the MySQL server to connect to.      |         | yes      
+ `forward_to`         | `list(LogsReceiver)` | Where to forward log entries after processing.                |         | yes      
+ `collect_interval`   | `duration`           | How frequently to collect information from database.          | `"10s"` | no       
+ `enable_collectors`  | `list(string)`       | A list of [collectors][] to enable on top of the default set. |         | no       
+ `disable_collectors` | `list(string)`       | A list of [collectors][] to disable from the default set.     |         | no
```
clayton-cornell, line 32:
```suggestion
 `collect_interval`   | `duration`           | How frequently to collect information from database.          | `"10s"` | no       
 `disable_collectors` | `list(string)`       | A list of [collectors][] to disable from the default set.     |         | no       
  `enable_collectors`  | `list(string)`       | A list of [collectors][] to enable on top of the default set. |         | no       
```

Just alphabetizing the "not required" arg in the table.

PR #2532, opened 2025-01-24
Add GitHub Actions steps for creating Linux build images.

ptodev:
After the PR is merged, I'd like to create a new build image tag to see if the image is created as expected.

I believe we don't need the Windows image, because #2522 runs fine without it. Also, GitHub Actions runners don't support running actions inside Docker containers.


mattdurham:
Should we adjust the drone jsonnet also?

mattdurham (approved):
lgtm

PR #2546, opened 2025-01-27
[helm] add NetworkPolicy support

TheRealNoob:
#### PR Description

Add NetworkPolicy support.  Since Alloy is extremely flexible in what it can be used for, the OOBE example I included basically doesn't enforce security at all.  The reason you might still want to enable it with this example is if your company has default-deny set up, meaning if you don't have a netpol at all then no traffic gets through.

I didn't open one.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -7,6 +7,13 @@ This document contains a historical list of changes between releases. Only
 changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
+0.11.1 (2025-01-27)
+------------------
+
+### Enhancements
+
+- Add NetworkPolicy support to the Helm chart. (@TheRealNoob)
+
 0.11.0 (2025-01-23)
 ----------
 
--- a/operations/helm/charts/alloy/templates/networkpolicy.yaml
+++ b/operations/helm/charts/alloy/templates/networkpolicy.yaml
@@ -1,0 +1,25 @@
+{{- if and .Values.networkPolicy.enabled (eq .Values.networkPolicy.flavor "kubernetes") -}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "alloy.fullname" . }}
+  namespace: {{ include "alloy.namespace" . }}
+  labels:
+    {{- include "alloy.labels" . | nindent 4 }}
+    app.kubernetes.io/component: networking
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "alloy.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
+  {{- if .Values.networkPolicy.ingress }}
+  ingress:
+    {{- toYaml .Values.networkPolicy.ingress | nindent 4 }}
+  {{- end }}
+  {{- if .Values.networkPolicy.egress }}
+  egress:
+    {{- toYaml .Values.networkPolicy.egress | nindent 4 }}
+  {{- end }}
+{{- end }}
--- a/operations/helm/charts/alloy/Chart.yaml
+++ b/operations/helm/charts/alloy/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alloy
 description: 'Grafana Alloy'
 type: application
-version: 0.11.0
+version: 0.11.1
 appVersion: 'v1.6.1'
 icon: https://raw.githubusercontent.com/grafana/alloy/main/docs/sources/assets/alloy_icon_orange.svg
 
--- a/operations/helm/charts/alloy/ci/create-networkpolicy-values.yaml
+++ b/operations/helm/charts/alloy/ci/create-networkpolicy-values.yaml
@@ -1,0 +1,2 @@
+networkPolicy:
+  enabled: true
--- a/operations/helm/charts/alloy/values.yaml
+++ b/operations/helm/charts/alloy/values.yaml
@@ -288,6 +288,41 @@ controller:
   # -- Additional containers to run alongside the Alloy container and initContainers.
   extraContainers: []
 
+networkPolicy:
+  enabled: true
+  flavor: kubernetes
+
+  ingress:
+    # Allow ingress from all pods in the cluster
+    - from:
+      - podSelector: {}
+    # Allow ingress from the internet (Faro)
+    - from:
+      - ipBlock:
+          cidr: 0.0.0.0/0
+  egress:
+    # Allow egress to the kube-apiserver (default port)
+    # Not all cloud providers use the same labels on kube-apiserver pod
+    - to:
+      - ports:
+        - protocol: TCP
+          port: 6443
+    # Allow egress to all pods in the cluster
+    - to:
+      - podSelector: {}
+    # Allow egress to the internet
+    - to:
+      - ipBlock:
+          cidr: 0.0.0.0/0
+          except:
+            - 10.0.0.0/8
+            - 172.16.0.0/12
+            - 192.168.0.0/16
+      - ipBlock:
+          cidr: ::/0 
+          except:
+            - fc00::/7
+
 service:
   # -- Creates a Service for the controller's pods.
   enabled: true
--- a/operations/helm/charts/alloy/values.yaml
+++ b/operations/helm/charts/alloy/values.yaml
@@ -289,7 +289,7 @@ controller:
   extraContainers: []
 
 networkPolicy:
-  enabled: true
+  enabled: false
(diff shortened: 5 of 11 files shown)


mattdurham:
Looks like some linting errors.

TheRealNoob:
@mattdurham I know this has stalled a bit and that's partially my fault as well even though I haven't said it.  I think the rules as I have them defined now may not be sufficient defaults.  It's amazing how Alloy's versatility mandates wide open netpols.  I wonder how you would feel about just setting the defaults open as ANY/ANY.  Obviously the netpol is still disabled by default.

TheRealNoob:
@mattdurham I changed the PR to default to `allow any/any` because Alloy is just so widely configurable, and even out of the box it talks to a lot of components.  I don't trust myself to list all of the endpoints.  Thoughts?

TheRealNoob:
@mattdurham @petewall Is there anythign I can do to help this along?  I know it has to be rebased, let me know if you're good with the spirit of the change and if it's a good time for me to do that.

operations/helm/charts/alloy/README.md
```diff
@@ -126,6 +126,10 @@ useful if just using the default DaemonSet isn't sufficient.
 | ingress.tls | list | `[]` |  |
 | nameOverride | string | `nil` | Overrides the chart's name. Used to change the infix in the resource names. |
 | namespaceOverride | string | `nil` | Overrides the chart's namespace. |
+| networkPolicy.egress[0] | object | `{}` |  |
+| networkPolicy.enabled | bool | `false` |  |
+| networkPolicy.flavor | string | `"kubernetes"` |  |
```
ptodev, line 131: Why do we need a `flavor` param? Can this be anything other than `kubernetes`? I know it's `{}` by default, but can someone not explicitly set `egress` to `nil` for example?

operations/helm/charts/alloy/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,25 @@
...
+  labels:
+    {{- include "alloy.labels" . | nindent 4 }}
+    app.kubernetes.io/component: networking
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "alloy.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
```
ptodev, line 16: Should we not list only the ones that are defined? E.g. either ingress or egress, or both?

ptodev (commented):
LGTM, based on my very limited Helm knowledge 😅 There are many ways to use Alloy and I don't think we can set up defaults that work in a general case. You could run Alloy just for that node in a DaemonSet to collect logs, or you could run it for central OTel telemetry collection where many nodes send to it but it has limited egress. There are many options.

Is it really necessary to add such things to the Helm chart though? Does the chart add any value versus just adding a network policy yaml manifest separately? I just don't see what added value this has inside the chart.

TheRealNoob:
I reverted my Chart.yaml version bump as I found that you guys do periodic releases, not immediately on feature release.  Correct me if I'm wrong.

operations/helm/charts/alloy/README.md
TheRealNoob: > Why do we need a flavor param? Can this be anything other than kubernetes?

Flavor is intended to be a toggle switch between different networkPolicy objects; some of them offer expanded features over the default kubernetes one.  Right now I'm only adding support for the vanilla one, so I could remove this toggle if you like.  I had decided to add it to communicate to the next person how the toggle is conventionally done, but I would hope that they know that already...

Here's an example of another project using it
https://github.com/prometheus-community/helm-charts/blob/86033644f8c2e8f34d4adc2d7f0c1dcc670baea6/charts/kube-prometheus-stack/values.yaml#L2951C1-L2960C23

TheRealNoob: > Is it really necessary to add such things to the Helm chart though? Does the chart add any value versus just adding a network policy yaml manifest separately? I just don't see what added value this has inside the chart.

If I can speak philosophically, Helm is envisioned to be a all-in-one wrapper for an application.  So if your application requires a netpol to function correctly, then it's supposed to ship with the application (as part of the Chart).  You can deploy the netpol separately, either manually or automated, but for obvious reasons it becomes a scaling challenge.

As an example, in my organization we install 3 separate releases of Alloy to each cluster.  One for metrics, for logs, and "other" for smaller cluster-specific things such as scraping a router/switch.  We do this to keep the resource usage of each Alloy pod down and to keep the metrics/logs collection uniform and simple.  Our cluster is also setup to default deny all traffic to a pod if it doesn't have a netpol (most clusters are configured to default allow).  So we need to make a netpol for alloy - I'm currently defining them in `extraObjects`, however it currently lacks `tpl` support so I'm having to hardcode the `spec.podSelector` on each of the releases.  It's be more concise if support was built into the chart, and I hope useful for others.

Hopefully this helps.  Let me know if anything's unclear.

operations/helm/charts/alloy/templates/networkpolicy.yaml
TheRealNoob: Good point.  Will fix.

TheRealNoob: I moved `policyTypes` to the values file

operations/helm/charts/alloy/README.md
```diff
@@ -126,6 +126,12 @@ useful if just using the default DaemonSet isn't sufficient.
...
+| networkPolicy.ingress[0] | object | `{}` |  |
+| networkPolicy.policyTypes[0] | string | `"Ingress"` |  |
+| networkPolicy.policyTypes[1] | string | `"Egress"` |  |
```
ptodev, line 134: Do we really need such parameters? Why not just write "ingress" if `networkPolicy.ingress` is not nil?

operations/helm/charts/alloy/README.md
ptodev: I see, but if there's a different netpol object type then wouldn't the current arguments have to be totally different? Then we'd also have to document which is for which, or maybe rename them to something like `networkPolicy.k8s.egress`?

TheRealNoob: Good question.  I've only ever looked into 2 or 3 other objects types, cilium being one of them.  It doesn't use `policyTypes` but it does use `ingress` and `egress`.  The fields defined within `ingress/egress` are different from the k8s-netpol, but I think it's reasonable to re-use the same fields between both objects and just ignore `policyTypes` for a cilium object.  At least that's what I've seen all the other projects do.  Sometimes they put cilium specific arguments under `networkPolicy.cilium.blah`.

Link to the ciliumNetworkPolicy schema
https://docs.cilium.io/en/stable/network/kubernetes/policy/

ptodev: Thank you, let's leave it as it is and not overcomplicate it then.

operations/helm/charts/alloy/README.md
ptodev: Thank you, makes sense! I suppose if helm can distinguish between a nil list and an empty one then we can still do without a `networkPolicy.policyTypes[0]`, but even if that's the case I'm happy with making it more explicit and having the `networkPolicy.policyTypes[0]`.

ptodev (approved):
LGTM, apologies for the delay and thank you for the thorough explanations! We have a few issues with the CI due to rotating a few CI tokens, but if the PR is rebased and the CI goes green once we fixed the token issues I'm happy to merge it.

ptodev:
Hi @TheRealNoob, apologies for the delay again! Do you mind rebasing please? It just needs a rebase and it's ready for merging.

TheRealNoob:
@ptodev updated.  Thank you!

ptodev:
@TheRealNoob no worries! Do you mind running `make generate-helm-tests` please? One of the tests is [failing](https://github.com/grafana/alloy/actions/runs/15726366518/job/44331913212?pr=2546).

TheRealNoob:
```
        modified:   tests/create-networkpolicy/alloy/templates/controllers/daemonset.yaml
```

operations/helm/charts/alloy/CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ Unreleased
 - Change configReloader from jimmydyson/configmap-reload to prometheus-operator/prometheus-config-reloader (@dehaansa)
 - Update to Grafana Alloy v1.7.5. (@kimxogus)
 - Add `checksum/config` pod annotation (@kimxogus)
+- Add NetworkPolicy support. (@TheRealNoob)
```
ptodev: This line will have to be moved to the "unreleased" section.

PR #2557, opened 2025-01-29
Remove unnecessary builds form drone.jsonnet, run "make drone"

ptodev:
This change syncs the Drone pipelines with drone.jsonnet. It is required so that the signature at the end of drone.yaml is up to date. Otherwise contributors will get have to manually run the drone pipelines.

Unfortunately, I think there's no way to add comments to the generated drone.yaml, so that we can keep a comment on what should be removed. That's ok - we are going to remove the builds soon anyway.


.drone/pipelines/build_images.jsonnet
ptodev, line 1: There is now a github action for building images, so we don't need to run the Drone builds.

cristiangreco (approved):

mattdurham (approved):
LGTM

simonswine (approved):
LGTM

PR #2575, opened 2025-01-30
database_observability: change default collect_interval to 1 minute

cristiangreco:
#### PR Description
Change default value of `collect_interval` to 1 minute.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


clayton-cornell (approved):
Approving for doc changes

matthewnolf (approved):
I think we are going to need to split the scrape intervals out for the different collectors. The requirements for each are fairly different and so there's no 1 value which makes sense for all. Query sampling will require more frequent collection given the data is more real time, whereas schema information is much less likely to change day to day (or minute to minute)

PR #2578, opened 2025-01-30
New loki.process stage structured_metadata_regex, to move all labels matching a regex to structured metadata

### LINKED ISSUE #926: Support Regex-based Label \ Metadata matching
### Request

This FR is similar to the following issue - https://github.com/grafana/loki/issues/4987

The `loki.process` component should allow regex pattern matching on the `structured_metadata ` and `pack` stages.

### Use case

In a Alloy pipeline, when discovering k8s targets I can easily perform a `labelmap` to have access to all the discoverable `__meta` labels and keep the ones I need. 

Now with these, in a `loki.process` component, I want to transform them into Loki Structured Metadata fields. This seems to be only possible by specifying a list of static names ([Source](https://grafana.com/docs/alloy/latest/reference/components/loki.process/#stagestructured_metadata-block)). 

With k8s discovered targets, they tend to have the `__meta_kubernetes_*_annotation_*` and `__meta_kubernetes_*_label_*` Labels. These values vary, and knowing all possible values, so that they get specified statically in the `stage.structured_metadata` block stage of the component doesn't seem feasible. Making me lose some k8s labels and annotations that I want to keep.

akevdmeer:
#### PR Description

This adds a stage to loki.process `structured_metadata_regex`, that takes a string attr `regex`, e.g.

```
        stage.structured_metadata_regex { 
                regex = "kubernetes_pod_.*"
        }
```
All labels that match the regular expression are moved to structured metadata (so removed as labels).

Fixes #926

#### Notes to the Reviewer

Please advise: 

Is it reasonable to introduce a new loki.process stage for this? 

I considered extending `loki.relabel` but that is a Loki wrapper around prometheus relabel so I decided against that. Then I considered extending the existing `structured_metadata` stage of `loki.process`, but also decided against that to not modify/embed the existing `LabelsConfig`.

Happy to re-work this per your guidance to get this feature merged.

#### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/pipeline.go
+++ b/internal/component/loki/process/stages/pipeline.go
@@ -38,6 +38,7 @@ type StageConfig struct {
 	ReplaceConfig         *ReplaceConfig         `alloy:"replace,block,optional"`
 	StaticLabelsConfig    *StaticLabelsConfig    `alloy:"static_labels,block,optional"`
 	StructuredMetadata    *LabelsConfig          `alloy:"structured_metadata,block,optional"`
+	StructuredMetadataRegex *StructuredMetadataRegexConfig `alloy:"structured_metadata_regex,block,optional"`
 	SamplingConfig        *SamplingConfig        `alloy:"sampling,block,optional"`
 	TemplateConfig        *TemplateConfig        `alloy:"template,block,optional"`
 	TenantConfig          *TenantConfig          `alloy:"tenant,block,optional"`
--- a/internal/component/loki/process/stages/stage.go
+++ b/internal/component/loki/process/stages/stage.go
@@ -40,6 +40,7 @@ const (
 	StageTypeSampling           = "sampling"
 	StageTypeStaticLabels       = "static_labels"
 	StageTypeStructuredMetadata = "structured_metadata"
+	StageTypeStructuredMetadataRegex = "structured_metadata_regex"
 	StageTypeTemplate           = "template"
 	StageTypeTenant             = "tenant"
 	StageTypeTimestamp          = "timestamp"
@@ -155,6 +156,11 @@ func New(logger log.Logger, jobName *string, cfg StageConfig, registerer prometh
 		}
 	case cfg.StructuredMetadata != nil:
 		s, err = newStructuredMetadataStage(logger, *cfg.StructuredMetadata)
+		if err != nil {
+			return nil, err
+		}
+	case cfg.StructuredMetadataRegex != nil:
+		s, err = newStructuredMetadataRegexStage(logger, *cfg.StructuredMetadataRegex)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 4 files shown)


ptodev:
Thank you for the PR! I personally would prefer if there is a new optional argument in the `structured_metadata` block called `regex`.Any labels which match the regex can then be added. This wouldn't conflict with the existing `values` argument. All the labels in `values ` will always be added if they exist; then `regex` can add additional ones. I'll let other maintainers open too, in case there are other ideas :)

wildum:
Hey, to add to @ptodev's comment, I would recommend adding a stage "label_extract" to move labels to the extracted values map. This way you can combine it with the structured_metadata stage; and any other processing stages since most of them can interact with the extracted values map.

ptodev:
Yes, just like the `values` argument, a potential `regex` argument needs to use the extracted map and not the labels. I like the idea of a `label_extract` stage too. I'd keep the `structured_metadata` stage simple.

akevdmeer:
To add a `regex` argument to `structured_metadata`, instead of introducing a new `structured_metadata_regex` stage is clear. 

But the current `structured_metadata` processes _both_ the extracted map, from which it copies to structured metadata, _and_ the labels from which it movies to structured metadata.

Should the `regex` argument not work in that same way? If you wish `structured_metadata` to act only on the extracted map, and introduce a then required `label_extract`, could we separate that from introducing `regex`?

Unlike adding `regex`, changing `structured_metadata` to no longer move from labels is a breaking change?

wildum:
Aaah sorry I missed this behavior. This is undocumented, I thought that the structured_metadata step was only processing the data from the extracted map and not the labels... Thanks for clarifying. That leaves us with two options:

- change the behavior to only process data from the extracted values map, add the regex arg to it, and create the new label_extract stage
- add the regex arg to it, make it behave like the values block, and update the doc (process from extracted values and labels)

I think that the first option offers more modularity + simpler logic. But it's a breaking change (of undocumented behavior but still annoying if a lot of users rely on it) + the new "label_extract" stage would need to be of stability GA right away to avoid forcing users to lower their stability level to achieve the same behavior. 

Which options do you prefer? @akevdmeer @ptodev
I will bring it up at an internal meeting that we have later today

wildum:
Discussed in our proposal review meeting today, we think that the option to not do breaking changes to the structured_metadata stage is the best.

We're ok to have an argument "regex" which will match with the extracted values map and the labels. Ideally it would be nice to have the possibility to modify the label name (like the `values`attribute does). This could be achieved with a similar approach as the loki.relabel component by using capture groups. This can be done separately (one PR brings the regex attribute and the relabeling can come later, especially because as a workaround you can put a loki.relabel component before the loki.process

akevdmeer:
Shall we close this PR, and I'll open a new one with the implementation as discussed above?

wildum:
> Shall we close this PR, and I'll open a new one with the implementation as discussed above?

yeah sounds good, you can link this PR in the new PR to keep the context. Thanks for your work, let us know if you need help!

PR #2604, opened 2025-02-04
pyroscope.scrape: add __profile_path_prefix__ label handling

### LINKED ISSUE #2597: pyroscope.scrape does not allow scraping with a prefix
### What's wrong?

If the scraping endpoints are mounted behind a prefix, for example  `/mimir-prometheus/debug/pprof`, then it is not possible to specify the prefix `/mimir-prometheus`  with a label (pod annotation?).

There is a `__profile_path__` label used to configure the url path here https://github.com/grafana/alloy/blob/583ad76ad6b55f3e6a45fc1c2eb4042574e72656/internal/component/pyroscope/scrape/target.go#L126, however it's value is taken from `pyroscope.scrape` configuration , specifically `ProfilingTarget.Path` https://github.com/grafana/alloy/blob/db2036bde950540900ad1ad4e6c03ebf76c17ad1/internal/component/pyroscope/scrape/scrape.go#L186-L185

In the current alloy version, scraping prefixed urls requires  creating a separate `pyroscope.scrape` component with path overrides for each profiling type, which is terrible.

I propose we modify code to handle `__profile_prefix__` label to concatenate the label with the `__profile_path__` label.

### Steps to reproduce

See above

### System information

_No response_

### Software version

`c6d62ec66`

### Configuration

```text

```

### Logs

```text

```

korniltsev:
#### PR Description
Introduce `__profile_path_prefix__` special label which is prepended to `__profile_path__` label before scraping URL creation.
Fixes #2597 #2605

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/pyroscope/scrape/target.go
+++ b/internal/component/pyroscope/scrape/target.go
@@ -223,12 +223,15 @@ func (t *Target) Health() TargetHealth {
 // LabelsByProfiles returns the labels for a given ProfilingConfig.
 func LabelsByProfiles(lset labels.Labels, c *ProfilingConfig) []labels.Labels {
 	res := []labels.Labels{}
+
 	add := func(profileType string, cfgs ...ProfilingTarget) {
 		for _, p := range cfgs {
 			if p.Enabled {
-				l := lset.Copy()
-				l = append(l, labels.Label{Name: ProfilePath, Value: p.Path}, labels.Label{Name: ProfileName, Value: profileType})
-				res = append(res, l)
+				lb := labels.NewBuilder(lset)
+				setIfNotPresentAndNotEmpty(lb, ProfilePath, p.Path)
+				setIfNotPresentAndNotEmpty(lb, ProfileName, profileType)
+				setIfNotPresentAndNotEmpty(lb, ProfilePathPrefix, c.PathPrefix)
+				res = append(res, lb.Labels())
 			}
 		}
 	}
@@ -449,4 +452,14 @@ func inferServiceName(lset labels.Labels) string {
 		return swarmService
 	}
 	return "unspecified"
+}
+
+func setIfNotPresentAndNotEmpty(b *labels.Builder, k, v string) {
+	if b.Get(k) != "" {
+		return
+	}
+	if v == "" {
+		return
+	}
+	b.Set(k, v)
 }
--- a/internal/component/pyroscope/scrape/target.go
+++ b/internal/component/pyroscope/scrape/target.go
@@ -398,9 +398,6 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments, targetTypes map[s
 				lbls = append(lbls, labels.Label{Name: model.AddressLabel, Value: lset.Get(model.AddressLabel)})
 				lbls = append(lbls, labels.Label{Name: model.SchemeLabel, Value: cfg.Scheme})
 				lbls = append(lbls, labels.Label{Name: ProfilePath, Value: lset.Get(ProfilePath)})
-				//ilbls = append(lbls, labels.Label{Name: ProfilePathPrefix, Value: profilePathPrefix})f profilePathPrefix := lset.Get(ProfilePathPrefix); profilePathPrefix != "" {
-				//	lbls = append(lbls, labels.Label{Name: ProfilePathPrefix, Value: profilePathPrefix})
-				//}
 				// Encode scrape query parameters as labels.
 				for k, v := range cfg.Params {
 					if len(v) > 0 {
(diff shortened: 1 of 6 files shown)


internal/component/pyroscope/scrape/target.go
```diff
@@ -404,6 +404,9 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments, targetTypes map[s
 				lbls = append(lbls, labels.Label{Name: model.AddressLabel, Value: lset.Get(model.AddressLabel)})
 				lbls = append(lbls, labels.Label{Name: model.SchemeLabel, Value: cfg.Scheme})
 				lbls = append(lbls, labels.Label{Name: ProfilePath, Value: lset.Get(ProfilePath)})
+				//ilbls = append(lbls, labels.Label{Name: ProfilePathPrefix, Value: profilePathPrefix})f profilePathPrefix := lset.Get(ProfilePathPrefix); profilePathPrefix != "" {
```
marcsanmi, line 407: Leftover?

internal/component/pyroscope/scrape/target.go
```diff
@@ -259,6 +258,7 @@ func (ts Targets) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
 
 const (
 	ProfilePath         = "__profile_path__"
+	ProfilePathPrefix   = "__profile_path_prefix__"
```
marcsanmi, line 261: I'd extend documentation and examples for this label in the component's reference docs.

marcsanmi (approved):
LGTM 👍

internal/component/pyroscope/scrape/target.go
korniltsev: Yeah, I wanted to gather some feedback before finalizing this.
I just have gut feeling that I'm missing something and it should be possible to configure a prefix without code modification, but I don't see it

internal/component/pyroscope/scrape/target.go
korniltsev: Yeah. I really want to remove this branch altogether as I believe it is a dead code and never executed

internal/component/pyroscope/scrape/target.go
marcsanmi: I've just seen `PprofPrefix` from `ProfilingConfig` 🤔

korniltsev: I've created an issue for unused argument https://github.com/grafana/alloy/issues/2605

simonswine: I agree that this is not possible right now based on labels and we make this easier.

I suggest rename the argument PprofPrefix to ProfilePathPrefix. Make its default `/debug/pprof/`.

When there is a label __profile_path_prefix__ override ProfilePathPrefix to what ever the value is. So for mimir it should be: `/mimir-prometheus/debug/pprof/`.

korniltsev: > I suggest rename the argument PprofPrefix to ProfilePathPrefix or maybe just PathPrefix

It is already called `path_prefix` in alloy configurationi, so `PathPrefix` be it

> Make its default /debug/pprof/.

I think we should not do this:
    1. There is fgprof which is behind `/debug/fgprof` , not `/debug/pprof/fgprof`
    2. There may be custom profiling targets which are not 
    3. We would have to change default values and this may break some existing configurations, if users configured custom path '/foo/debug/pprof/profile` and we start prepending `/debug/pprof`

> When there is a label profile_path_prefix override ProfilePathPrefix to what ever the value is.

Agree. I was going to do this in a separate PR to resolve [this](https://github.com/grafana/alloy/issues/2605)  issue, but lets do this here and now.

internal/component/pyroscope/scrape/target.go
```diff
@@ -404,6 +398,9 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments, targetTypes map[s
 				lbls = append(lbls, labels.Label{Name: model.AddressLabel, Value: lset.Get(model.AddressLabel)})
 				lbls = append(lbls, labels.Label{Name: model.SchemeLabel, Value: cfg.Scheme})
 				lbls = append(lbls, labels.Label{Name: ProfilePath, Value: lset.Get(ProfilePath)})
+				//ilbls = append(lbls, labels.Label{Name: ProfilePathPrefix, Value: profilePathPrefix})f profilePathPrefix := lset.Get(ProfilePathPrefix); profilePathPrefix != "" {
+				//	lbls = append(lbls, labels.Label{Name: ProfilePathPrefix, Value: profilePathPrefix})
+				//}
```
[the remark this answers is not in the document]
korniltsev: removed the gibberish, will remove the whole branch and relabeling-drop logic in a followup

PR #2607, opened 2025-02-04
database_observability: extract table schema

cristiangreco:
#### PR Description
This is the first iteration of the implementation for extracting the table schema details.

In this PR:
- parse columns information_schema.columns (ignore indexes and other constraints)
- introduce a json structure to represent the table schema
- base64-encode both the schema and the create statement itself

Still a bunch of things to followup on but didn't want to make this PR too big.

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,23 +22,18 @@ Main (unreleased)
 
 - Add the possibility to export span events as logs in `otelcol.connector.spanlogs`. (@steve-hb)
 
-- (_Experimental_) Log instance label key in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Improve parsing of truncated queries in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Capture schema name for query samples in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Fix handling of view table types when detecting schema in `database_observability.mysql` (@matthewnolf)
-
-- (_Experimental_) Fix error handling during result set iteration in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Better support for table name parsing in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Better error handling for `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Add namespace to `connection_info` metric in `database_observability.mysql` (@cristiangreco)
-
 - Add json format support for log export via faro receiver (@ravishankar15)
+
+- (_Experimental_) Various changes to the experimental component `database_observability.mysql`:
+  - Always log `instance` label key (@cristiangreco)
+  - Improve parsing of truncated queries (@cristiangreco)
+  - Capture schema name for query samples (@cristiangreco)
+  - Fix handling of view table types when detecting schema (@matthewnolf)
+  - Fix error handling during result set iteration (@cristiangreco)
+  - Better support for table name parsing (@cristiangreco)
+  - Better error handling for components (@cristiangreco)
+  - Add namespace to `connection_info` metric in `database_observability.mysql` (@cristiangreco)
+  - Added table columns parsing (@cristiagreco)
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,19 +16,15 @@ Main (unreleased)
...
-- (_Experimental_) Capture schema name for query samples in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Fix handling of view table types when detecting schema in `database_observability.mysql` (@matthewnolf)
-
-- (_Experimental_) Fix error handling during result set iteration in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Better support for table name parsing in `database_observability.mysql` (@cristiangreco)
-
-- (_Experimental_) Better error handling for `database_observability.mysql` (@cristiangreco)
+- (_Experimental_) Various changes to the experimental component `database_observability.mysql`:
```
cristiangreco, line 19: I've moved everything related to this component under a single entry.

cristiangreco: Will fix conflicts after approva.

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -219,6 +234,11 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
 		}
 	}
 
+	if len(tables) == 0 {
+		level.Info(c.logger).Log("msg", "no tables detected from information_schema.tables")
```
fridgepoet, line 238: just curious why we want to log this?

fridgepoet:
Would it be possible to get some unit tests on parseTableSpec? I'm seeing a lot of special conditions that could benefit from illustration of how they work and protection during future refactoring.
Would it also be possible to add a test for the encoded json for tableSpec? (I mean various tests where we json Marshal a tableSpec to assert its output json string.) I'm just thinking it might be useful to illustrate/protect that json output with a test for whoever is consuming the json.

cristiangreco:
> Would it be possible to get some unit tests on parseTableSpec? I'm seeing a lot of special conditions that could benefit from illustration of how they work and protection during future refactoring. Would it also be possible to add a test for the encoded json for tableSpec? (I mean various tests where we json Marshal a tableSpec to assert its output json string.) I'm just thinking it might be useful to illustrate/protect that json output with a test for whoever is consuming the json.

Definitely, but I'd like to do it in a followup PR where all the parsing is moved to a separate file.

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: Found useful for a case where the user was not granted SELECT to that particular schema, so the query on `information_schema.tables` would return an empty result without error (just a blank output). I'm happy to move to debug, maybe?

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -241,27 +261,125 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
...
+				level.Error(c.logger).Log("msg", "failed to scan create view", "schema", table.schema, "table", table.tableName, "err", err)
 				break
 			}
 		}
 
 		table.createStmt = createStmt
 		c.cache.Add(cacheKey, table)
 
+		spec, err := c.parseTableSpec(table.schema, table.tableName, createStmt)
+		if err != nil {
```
matthewnolf, line 278: Do we not want to report this error in some way?

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -241,27 +261,125 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
...
+			break
+		}
+		jsonSpec, err := json.Marshal(spec)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to marshal table spec", "schema", table.schema, "table", table.tableName, "err", err)
```
matthewnolf, line 283: It might be worth adding const labels for these log lines, given we're using them in many places

internal/component/database_observability/mysql/collector/schema_table.go
fridgepoet: Thanks for the info! I'm ok with it in Info!

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: Done!

matthewnolf (approved):

PR #2614, opened 2025-02-05
fix(pyroscope.scrape): godeltaprof hiding

korniltsev:
#### PR Description
The PR fixes the godeltaprof hiding.
The code inside `NewTarget` was checking the wrong label name, also the correct label name is reserved, so it should have been done in a different branch.
Furthermore the test did not check the target values, just inequality

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


korniltsev:
cc @simonswine

simonswine (approved):
Thanks for fixing this

LGTM

PR #2617, opened 2025-02-05
pyroscope: Improve pyroscope.relabel instrumentation

marcsanmi:
#### PR Description

- Improves instrumentation in the `pyroscope.relabel` component.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/relabel/metrics.go
+++ b/internal/component/pyroscope/relabel/metrics.go
@@ -1,6 +1,11 @@
 package relabel
 
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"github.com/grafana/alloy/internal/util"
+
+	"github.com/prometheus/client_golang/prometheus"
+	prometheus_client "github.com/prometheus/client_golang/prometheus"
+)
 
 type metrics struct {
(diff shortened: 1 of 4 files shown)


internal/component/pyroscope/relabel/metrics.go
```diff
@@ -12,41 +17,41 @@ type metrics struct {
 }
 
 func newMetrics(reg prometheus.Registerer) *metrics {
```
simonswine, line 19: Not necessarily blocking, but it is super easy to mess up the metrics, so I would suggest you strongly consider adding some test coverage for it, it is fairly straight forward: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil#CollectAndCompare

simonswine (approved):
LGTM

internal/component/pyroscope/relabel/metrics.go
marcsanmi: Updated. I also took the chance to add a test checking the purge of the cache.

simonswine (approved):
Thank you ❤️

PR #2627, opened 2025-02-06
pyroscope.scrape: remove relabeling, remove discovered targets field, remove public labels field, use labels builder

korniltsev:
#### PR Description
- remove relabeling logic, which was impossible to configure and therefore it was dead code.
- remove dropped targets logic as a result
- remove `Target.discoveredLabels` field as it was only used as debug information (was replaced with `Target.allLabels` in debug information). As a result, remove the second argument to `NewTarget` simplifying testing code.
- use `labels.Builder` for `populateLabels` to reduce some allocations
- use `filepath.Join` instead of `URL.joinPath` to save some allocations
- add a small benchmark
- remove publicLabels field. 
- calculate public labels hash with less allocations

~Found a bug: Updating labels, does not work. The updating logic was updating `DiscoveredTargets` which are not used when data is forwarded to `pyroscope.write`. Created an issue: #2628~ EDIT: never-mind, I think we should not update anything at all, unless we decide to change __meta labels dropping behavior, see #2629
```
Before
BenchmarkPopulateLabels-12    	   11810	    110312 ns/op	   77465 B/op	     661 allocs/op
BenchmarkPopulateLabels-12    	   10000	    106040 ns/op	   77466 B/op	     661 allocs/op
After
BenchmarkPopulateLabels-12    	   24166	     52106 ns/op	   30082 B/op	     280 allocs/op
BenchmarkPopulateLabels-12    	   23424	     54803 ns/op	   30082 B/op	     280 allocs/op
```

Note: allocations and performance are not a big issue here. The goal of the PR is to remove dead code.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


marcsanmi (approved):
LGTM 👍

PR #2631, opened 2025-02-06
introduce new way of getting process name in loki source windows event

### LINKED ISSUE #2615: Improve process name retrieval in loki.source.windowsevent
The investigation performed in https://github.com/grafana/alloy/issues/2539#issuecomment-2618982447 showed that most of the CPU usage is spent updating the bookmark and retrieving the process name.

The bookmark was optimized in https://github.com/grafana/alloy/pull/2558.

Now more than 80% CPU usage of the component (and most of the overall Alloy CPU usage) is used to retrieve the process name for every processed events.

This comes from the fact that GetFromSnapProcess iterates over all processes running on the Windows machine till it finds the one that matches the pid. These system calls are expensive.

We should find a better way to retrieve the process name.

wildum:
#### PR Description

This PR further reduces the CPU usage of the loki.source.windows component by 70%.

The previous approach was iterating over all processes to find the one corresponding to the received event by comparing the PIDs to retrieve the process name.

The new approach uses the Windows API to query the process name directly using the PID. It also returns directly the names "System" and "Idle Process" which are guaranteed to always have the same PIDs on all Windows machines.

Fixes #2615

#### Notes to the Reviewer

The change was tested by running both the old function and the new function side by side and comparing the results with at least 1_000_000 prod events. No discrepancies were found between the two versions.
Profiles showed an improvement of more than 70% compared to the version on main.

Additional manual tests were done locally to test different types of events.

#### PR Checklist

- [x] CHANGELOG.md updated
- [na] Documentation added
- [na] Tests updated

### DIFF
--- a/internal/component/loki/source/windowsevent/format.go
+++ b/internal/component/loki/source/windowsevent/format.go
@@ -7,9 +7,14 @@ package windowsevent
 
 import (
 	"fmt"
+	"path/filepath"
 	"syscall"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+
 	jsoniter "github.com/json-iterator/go"
+	"golang.org/x/sys/windows"
 
 	"github.com/grafana/loki/v3/clients/pkg/promtail/scrapeconfig"
 	"github.com/grafana/loki/v3/clients/pkg/promtail/targets/windows/win_eventlog"
@@ -59,7 +64,7 @@ type Correlation struct {
 }
 
 // formatLine format a Loki log line from a windows event.
-func formatLine(cfg *scrapeconfig.WindowsEventsTargetConfig, event win_eventlog.Event) (string, error) {
+func formatLine(cfg *scrapeconfig.WindowsEventsTargetConfig, event win_eventlog.Event, l log.Logger) (string, error) {
 	structuredEvent := Event{
 		Source:        event.Source.Name,
 		Channel:       event.Channel,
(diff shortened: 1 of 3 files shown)


internal/component/loki/source/windowsevent/format.go
```diff
@@ -112,10 +114,37 @@ func formatLine(cfg *scrapeconfig.WindowsEventsTargetConfig, event win_eventlog.
...
+		return "Idle Process", nil
+	}
+
+	handle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, pid)
+	if err != nil {
+		return "", fmt.Errorf("failed to open process %d: %w", pid, err)
+	}
+	defer windows.CloseHandle(handle)
+
+	var buf [windows.MAX_PATH]uint16
```
ptodev, line 143: Does this get the correct max path length even if support for long paths is enabled?

wildum: good point, I was not aware of the long paths param on Windows. I changed the code to dynamically resize the buffer if it was too small.
I tried setting up unit tests but that was tricky so instead I manually tested by:
- setting a very small buffer size
- running it on my computer to make sure that the windows API would throw the correct error
- checked that the buffer size was correctly increased till it was big enough to contain the full path

internal/component/loki/source/windowsevent/format.go
```diff
@@ -112,10 +114,48 @@ func formatLine(cfg *scrapeconfig.WindowsEventsTargetConfig, event win_eventlog.
...
+		err = windows.QueryFullProcessImageName(handle, 0, &buf[0], &size)
+		if err == nil {
+			return filepath.Base(windows.UTF16ToString(buf[:size])), nil
+		}
+		if err == windows.ERROR_INSUFFICIENT_BUFFER {
+			if size >= maxSize {
+				return "", fmt.Errorf("failed to get process name for %d: buffer size exceeded maximum limit", pid)
+			}
+			size *= 2
+			continue
```
ptodev, line 157: It's a bit weird to retry like this, but I can't think of a better way to do it. There might be some examples on the internet that do it in a smarter way. Anyway, I think this code is ok at least for now.

ptodev (approved):

internal/component/loki/source/windowsevent/format.go
ptodev: Thank you, it's understandable that it can't be unit tested easily. Would you like to add a comment with how it was tested, and a TODO to maybe make a test later? That way at least it's documented.

internal/component/loki/source/windowsevent/format.go
wildum: yeah it's fairly low level that's why it's a bit limited. I think we are fine because the implementation on main did not support long path anyway (buf was 260 bytes). It's not super pretty but with 512 buf size + dynamic resize + safe guard we should be fine

internal/component/loki/source/windowsevent/format.go
wildum: done!

PR #2636, opened 2025-02-06
database_observability: refactor cache config in schema_table collector

cristiangreco:
WIP

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/query_sample_test.go
```diff
@@ -553,6 +553,12 @@ func TestQuerySampleSQLDriverErrors(t *testing.T) {
 					"select * from some_table where id = 1",
 					"2024-01-01T00:00:00.000Z",
 					"1000",
+				).AddRow(
+					"def456",
+					"another_schema",
+					"select * from another_table where id = 2",
+					"2024-01-01T00:00:00.000Z",
+					"1000",
```
cristiangreco, line 561: Drive by fix.

internal/component/database_observability/mysql/collector/schema_table_test.go
```diff
@@ -121,6 +121,233 @@ func TestSchemaTable(t *testing.T) {
...
+		err = mock.ExpectationsWereMet()
+		require.NoError(t, err)
+
+		lokiEntries := lokiClient.Received()
+		for _, entry := range lokiEntries {
+			require.Equal(t, model.LabelSet{"job": database_observability.JobName}, entry.Labels)
+		}
+		require.Equal(t, `level=info msg="schema detected" op="schema_detection" instance="mysql-db" schema="some_schema"`, lokiEntries[0].Line)
+		require.Equal(t, `level=info msg="table detected" op="table_detection" instance="mysql-db" schema="some_schema" table="some_table"`, lokiEntries[1].Line)
+		require.Equal(t, fmt.Sprintf(`level=info msg="create table" op="create_statement" instance="mysql-db" schema="some_schema" table="some_table" create_statement="%s" table_spec="%s"`, base64.StdEncoding.EncodeToString([]byte("CREATE TABLE some_table (id INT)")), base64.StdEncoding.EncodeToString([]byte(`{"columns":[{"name":"id","type":"int","not_null":true,"auto_increment":true,"primary_key":true,"default_value":"null"}]}`))), lokiEntries[2].Line)
```
fridgepoet, line 222: I didn't understand how this test doesn't read from cache like the other test does. How are the two tests different? Is it just that the other test waits until the number of lokiEntries double to 6?

internal/component/database_observability/mysql/collector/query_sample_test.go
fridgepoet: Just curious, what was the issue here?

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Main (unreleased)
   - Add namespace to `connection_info` metric (@cristiangreco)
   - Added table columns parsing (@cristiagreco)
   - Add enable/disable collector configurability to `database_observability.mysql`. This removes the `query_samples_enabled` argument, now configurable via enable/disable collector. (@fridgepoet)
+  - Refactor cache config in schema_table collector (@cristiangreco)
```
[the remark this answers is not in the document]
cristiangreco: No, I'm still leaving out the ability to configure it for now (there's a comment in component.go)

internal/component/database_observability/mysql/collector/schema_table_test.go
cristiangreco: This test only writes to cache, because it loops only once. The next test writes in the first loop and then reads in the second loop. I could use a suggestion for better naming.

internal/component/database_observability/mysql/collector/query_sample_test.go
cristiangreco: I didn't realise that when using `RowError(1, fmt.Errorf("rs error"))` you still need to write the result for the row at position `1`. The sqlmock library will then return the error instead of the actual row.

The library behaviour here is a bit confusing to me, I only caught this with further debugging. (I fixed a similar problem in the other test as well)

internal/component/database_observability/mysql/collector/schema_table_test.go
fridgepoet: Does it do a second in the next test only because of the following or did I miss anything else?
```
		require.Eventually(t, func() bool {
			return len(lokiClient.Received()) == 6
		}, 50000*time.Second, 100*time.Millisecond)
```

CHANGELOG.md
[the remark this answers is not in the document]
fridgepoet: Ah sorry missed that

internal/component/database_observability/mysql/collector/schema_table_test.go
```diff
@@ -121,6 +121,233 @@ func TestSchemaTable(t *testing.T) {
...
+		require.Eventually(t, func() bool {
+			return len(lokiClient.Received()) == 6
+		}, 50000*time.Second, 100*time.Millisecond)
+
+		collector.Stop()
+		lokiClient.Stop()
+
+		require.Eventually(t, func() bool {
+			return collector.Stopped()
+		}, 50000*time.Second, 100*time.Millisecond)
```
cristiangreco, line 335:
```suggestion
		}, 5*time.Second, 100*time.Millisecond)
```

internal/component/database_observability/mysql/collector/schema_table_test.go
```diff
@@ -121,6 +121,233 @@ func TestSchemaTable(t *testing.T) {
...
+					time.Date(2024, 2, 2, 0, 0, 0, 0, time.UTC),
+				),
+			)
+
+		err = collector.Start(context.Background())
+		require.NoError(t, err)
+
+		require.Eventually(t, func() bool {
+			return len(lokiClient.Received()) == 6
+		}, 50000*time.Second, 100*time.Millisecond)
```
cristiangreco, line 328:
```suggestion
		}, 5*time.Second, 100*time.Millisecond)
```

internal/component/database_observability/mysql/collector/schema_table_test.go
cristiangreco: Thanks to your comment I noticed I had left `50000` as I was debugging the test.

> Does it do a second loop in the next test only because of the following

It loops twice because the mock database is instructed to expect/return values for this sequence of queries:

- selectSchemaName (first loop)
- selectTableName (first loop)
- SHOW CREATE TABLE (first loop)
- selectColumnNames (first loop)
- selectSchemaName (second loop)
- selectTableName (second loop)

Note that in the second loop we expect fewer queries because that content must be cached.

Other tests stop after the first loop because the mock database return an error after all expectations are met.

fridgepoet: Ah ok thanks for the explanation. 

Hmm I don't have a great idea for improving the names. Maybe some long explicit test names if you want. "detects table schema, writes to cache, then retrieves from cache"

In this case where it's a little complicated, I would maybe just include what you explained here as comments.
Maybe group the (first loop) mocks together and above a comment like `// first loop, cache is written` then group the (second loop) mocks together and above a comment like `// second loop, note that show create table and select column names are not called: they were retrieved from the cache`. It might not hurt to have these comments if we consider these tests to be a bit complicated. (I am usually more in favor of self-explanatory code, but in this case it might make figuring them out a lot easier in the future for a newcomer or while debugging a problem.)

By the way, I don't know if this would add value or make anything clearer, but are you able make assertions on the cache directly itself? (collector.cache) 

Anyway just my 2c, I defer to your best judgement.

fridgepoet (approved):

internal/component/database_observability/mysql/collector/schema_table_test.go
cristiangreco: > By the way, I don't know if this would add value or make anything clearer, but are you able make assertions on the cache directly itself? (collector.cache)

One thing we should add later on is debug metrics (those exposed by Alloy itself), which we could use for tests assertions. Right now it'd be a bit tedious to assert it in tests though.

fridgepoet: Sorry I might've meant a different thing.
I meant that in the test, I saw I can access the methods of `collector.cache`. And you're putting stuff into the cache, right? For example, would it be possible to assert that collector.cache has some key or some values that exist that you expect to be there? (I think it is less end-to-end than the lokiEntries, so disregard this if it doesn't add value. I don't mean to beat a dead horse or anything.)

cristiangreco: I see, pushed one more commit with a new assertion. Not a big fan of testing it via accessing a private field, but don't want to expose the cache with public methods either.

PR #2648, opened 2025-02-07
Move context retrieval under mutex lock in concurrent eval

wildum:
#### PR Description

There was a small window where a component would copy the context during the re-evaluation process. This led to the component being evaluated without all the data needed and logged errors.

#### PR Checklist

- [x] CHANGELOG.md updated
- [na] Documentation added
- [ ] Tests updated -> bit hard to test given the time sensitivity 

### DIFF
--- a/internal/runtime/internal/controller/loader.go
+++ b/internal/runtime/internal/controller/loader.go
@@ -807,10 +807,10 @@ func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer tr
 	var err error
 	switch n := n.(type) {
 	case BlockNode:
-		ectx := l.cache.GetContext()
 
 		// RLock before evaluate to prevent Evaluating while the config is being reloaded
 		l.mut.RLock()
+		ectx := l.cache.GetContext()
 		evalErr := n.Evaluate(ectx)
 
 		err = l.postEvaluate(l.log, n, evalErr)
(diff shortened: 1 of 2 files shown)


internal/runtime/internal/controller/loader.go
```diff
@@ -807,10 +807,10 @@ func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer tr
 	var err error
 	switch n := n.(type) {
 	case BlockNode:
-		ectx := l.cache.GetContext()
 
 		// RLock before evaluate to prevent Evaluating while the config is being reloaded
 		l.mut.RLock()
+		ectx := l.cache.GetContext()
```
mattdurham, line 813: This looks good, but noticed a few lines down on 826 that defer looks...suspicious. When should that unlock logically happen? If its at 831 would it make sense to wrap that defer in a func call. Feels easy to add logic after later in the function that could be unexpected with that defer.

mattdurham (approved):
Looks good

wildum:
> Would be good to follow up with some test that would at least be sometimes flaky without this fix?

I added a test. It's a bit more generic because it checks that Alloy does not log errors while reloading the config. Tested on main locally, I got it to fail everytime. On this branch I ran it 100 times with the race flag and it passed.

thampiotr (approved):
Nice work, thanks for adding the test! Having done some huge refactor for targets recently, I cannot stress enough how much tests help!

PR #2665, opened 2025-02-09
Add walqueue sharding

mattdurham:
#### PR Description

This adds dynamic sharding configuration into prometheus.write.queue

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -48,13 +48,13 @@ The following arguments are supported:
...
- endpoint > basic_auth | [basic_auth][]   | Configure basic_auth for authenticating to the endpoint. | no       
- endpoint > tls_config | [tls_config][]   | Configure TLS settings for connecting to the endpoint.   | no
-
+ Hierarchy             | Block           | Description                                               | Required 
+-----------------------|-----------------|-----------------------------------------------------------|----------
+ persistence           | [persistence][] | Configuration for persistence                             | no       
+ endpoint              | [endpoint][]    | Location to send metrics to.                              | no       
+ endpoint > basic_auth | [basic_auth][]  | Configure basic_auth for authenticating to the endpoint.  | no       
+ endpoint > tls_config | [tls_config][]  | Configure TLS settings for connecting to the endpoint.    | no
+ endpoint > parallelism | [parallelism][] | Configure parralelism for the endpoint.                   | no
```
thampiotr, line 57:
```suggestion
 endpoint > parallelism | [parallelism][] | Configure parallelism for the endpoint.                   | no
```

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -64,7 +64,7 @@ basic_auth` refers to a `basic_auth` block defined inside an
 [basic_auth]: #basic_auth-block
 [persistence]: #persistence-block
 [tls_config]: #tls_config-block
-
+[parallelism]: #parralelism
```
thampiotr, line 67:
```suggestion
[parallelism]: #parallelism
```

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,31 @@ Name                   | Type     | Description
 `insecure_skip_verify` | `bool`   | Disables validation of the server certificate.           |         | no
 `key_pem`              | `secret` | Key PEM-encoded text for client authentication.          |         | no
 
+
+### parallelism block
+
+Name            | Type       | Description                                                                                                                        | Default | Required
+----------------|------------|------------------------------------------------------------------------------------------------------------------------------------|---------|---------
+`drift_scale_up_seconds`      | `int`      | The maximum amount of time between the timestamps of incoming signals and outgoing signals before increasing desired connections.  | `60`    | no
+`drift_scale_down_seconds`    | `int`      | The minimum amount of time between the timestamps of incoming signals and outgoing signals before descreasing desired connections. | `30`    | no
```
thampiotr, line 125: I'd prefer this to be duration to stay consistent with Alloy's configs everywhere else. We also have duration used in this config block already for flush interval.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,31 @@ Name                   | Type     | Description
...
+`max_connections`    | `uint`     | The maximum number of desired connections.                                                                                         | `50`    | no
+`min_connections`    | `uint`     | The minimum number of desired connections.                                                                                         | `2`     | no
+`network_flush_interval`    | `duration` | The length of time that network successes and failures are kept for determining desired connections.                               | `1m`    | no
+`desired_connections_lookback`    | `duration` | The length of time that previous desired connections are kept for determining desired connections.                                 | `5m`    | no
+`desired_check_interval`    | `duration` | The length of time that between checking for desired connections.                                                                  | `5s`    | no
```
thampiotr, line 130:
```suggestion
`desired_check_interval`    | `duration` | The length of time between checking for desired connections.                                                                  | `5s`    | no
```

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,31 @@ Name                   | Type     | Description
...
+`allowed_network_error_percent`    | `float`    | The allowed error rate before scaling down.                                                                                        | `0.50`   | no
+
+Parralelism determines when to scale up or down the number of desired connections. This is accomplished by a variety of inputs.
+By determining the drift between the incoming and outgoing timestamps that will determine whether to increase or decrease the
```
thampiotr, line 134: Not sure if this connects / flows with previous sentence. Is it meant to be a bullet point?

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,31 @@ Name                   | Type     | Description
...
+`max_connections`    | `uint`     | The maximum number of desired connections.                                                                                         | `50`    | no
+`min_connections`    | `uint`     | The minimum number of desired connections.                                                                                         | `2`     | no
+`network_flush_interval`    | `duration` | The length of time that network successes and failures are kept for determining desired connections.                               | `1m`    | no
+`desired_connections_lookback`    | `duration` | The length of time that previous desired connections are kept for determining desired connections.                                 | `5m`    | no
+`desired_check_interval`    | `duration` | The length of time that between checking for desired connections.                                                                  | `5s`    | no
+`allowed_network_error_percent`    | `float`    | The allowed error rate before scaling down.                                                                                        | `0.50`   | no
```
thampiotr, line 131: This is 0.5% = 0.005 or is it 50% = 0.5? Just to be sure ;)

internal/component/prometheus/write/queue/types.go
```diff
@@ -66,6 +75,23 @@ func (r *Arguments) Validate() error {
...
+			return fmt.Errorf("drift_scale_up_seconds less than or equal drift_scale_down_seconds")
+		}
+		// Any lower than 1 second and you spend a fair amount of time churning on the draining and
+		// refilling the write buffers.
+		if conn.Parallelism.DesiredCheckInterval < 1*time.Second {
+			return fmt.Errorf("desired_check_interval must be greater than or equal to 1 second")
+		}
+		if conn.Parallelism.AllowedNetworkErrorPercent < 0 || conn.Parallelism.AllowedNetworkErrorPercent > 1 {
+			return fmt.Errorf("allowed_network_error_percent must be between 0.00 and 1.00")
+		}
```
thampiotr, line 94: Are we restricting to max 1% = 0.01 or max 100% = 1?
If the latter, this should be called maybe `AllowedNetworkErrorFraction` instead.

internal/component/prometheus/write/queue/types.go
```diff
@@ -100,6 +126,32 @@ type TLSConfig struct {
 	InsecureSkipVerify bool              `alloy:"insecure_skip_verify,attr,optional"`
 }
 
+type ParralelismConfig struct {
```
thampiotr, line 129: I'd be okay to refer to the docs here instead of duplicating info.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
mattdurham: Lord one day I will spell it right.

internal/component/prometheus/write/queue/types.go
mattdurham: The value is restricted from 0 to 1, changed the underlying name in the walqueue with a pr and will change to fraction since that is more accurate verbiage.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
mattdurham: 0.5 would mean allow 50 percent of failures to happen before scaling down.

thampiotr (approved):

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,35 @@ Name                   | Type     | Description
...
+two values then the value will stay the same. 
+
+Network success and failures are recorded and kept in memory, this helps determine
+the nature of the drift. For instance if the drift is increasing but the network failures are increasing we should not increase
+desired connections since that would only increase load on the endpoint. 
+
+Flapping prevention accomplished with `desired_check_interval`, each time a desired connection is calculated it is added to a list, before actually changing the
+desired connection the system will choose the highest value in the lookback buffer. Example; for the past 5 minutes desired connections have been: [2,1,1] the check runs
+and determines that the desired connections are 1, but will not change the value since the value 2 is still in the lookback. On the next check we have [1,1,1],
+now it will change to 1. In general the system is fast to increase and slow to decrease.
```
clayton-cornell, line 146:
```suggestion
Parallelism determines when to scale up or down the number of desired connections.

The drift between the incoming and outgoing timestamps determines whether to increase or decrease the desired connections.
The value stays the same if the drift is between `drift_scale_up_seconds` and `drift_scale_down_seconds`. 

Network successes and failures are recorded and kept in memory.
This data helps determine the nature of the drift.
For example, if the drift is increasing and the network failures are increasing, the desired connections should not increase because that would increase the load on the endpoint.

The `desired_check_interval` prevents connection flapping.
Each time a desired connection is calculated, the connection is added to a list.
Before changing the desired connection, the system will choose the highest value in the lookback buffer.
For example, for the past 5 minutes, desired connections have been: [2,1,1].
The check determines that the desired connections are 1, and the number of desired connections will not change because the value `2` is still in the lookback buffer.
On the next check, the desired connections are [1,1,1].
Now, it will change to 1. In general, the system is fast to increase and slow to decrease.
```

This is a first pass at reworking the description here. I don't think I've accurately distilled what you were trying to explain here... so I expect we will have to go over this at least once more.

Questions... 
1. What is `the system` here? Alloy? The connector? 
2. In the 4th paragraph example, what changes to `1` when the lookback is `[1,1,1]`?

mattdurham: 1. Yes, but more specifically the component `prometheus.write.queue` so will be specific there.

2. The `desired_connections` change to 1.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,37 @@ Name                   | Type     | Description
...
+
+The drift between the incoming and outgoing timestamps determines whether to increase or decrease the desired connections.
+The value stays the same if the drift is between `drift_scale_up_seconds` and `drift_scale_down_seconds`. 
+
+Network successes and failures are recorded and kept in memory.
+This data helps determine the nature of the drift.
+For example, if the drift is increasing and the network failures are increasing, the desired connections should not increase because that would increase the load on the endpoint.
+
+The `desired_check_interval` prevents connection flapping.
+Each time a desired connection is calculated, the connection is added to a list.
```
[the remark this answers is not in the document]
mattdurham: Prometheus has a concept called desired shards. This is similar to that which is the number concurrent outgoing connections to keep up with the load, so if the incoming data goes down a calculation is done and it lowers desired shards/connections and the opposite. 

Its not a comma separated list, its an in memory buffer of the last X number of desired connections evaluations. This is to prevent a situation where it says it needs 4,then 5, then 4, then 5; which can lead to flapping and more work for the system. Instead it looks back at the last X number of desired connections and goes with the max. Eventually the older ones will time out and it will scale down. Note this only applies to scaling down. The idea is scale up fast, slow scale down.

clayton-cornell, line 143:
```suggestion
Each time a desired connection is calculated, the connection is added to a lookback buffer.
```

Maybe then instead of list we use the same terminology as in a later sentence here.. `lookback buffer` ?

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,37 @@ Name                   | Type     | Description
...
+Before changing the desired connection, `prometheus.write.queue` will choose the highest value in the lookback buffer.
```
clayton-cornell, line 144:
```suggestion
Before increasing or decreasing the desired connections, `prometheus.write.queue` chooses the highest value in the lookback buffer.
```

docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -117,6 +116,37 @@ Name                   | Type     | Description
...
+For example, for the past 5 minutes, desired connections have been: [2,1,1].
+The check determines that the desired connections are 1, and the number of desired connections will not change because the value `2` is still in the lookback buffer.
+On the next check, the desired connections are [1,1,1].
+Since the `2` value has expired the desired connections will change to 1. In general, the system is fast to increase and slow to decrease.
```
clayton-cornell, line 148:
```suggestion
For example, for the past 5 minutes, the desired connections have been: [2,1,1].
The check determines that the desired connections are 1, and the number of desired connections will not change because the value `2` is still in the lookback buffer.
On the next check, the desired connections are [1,1,1].
Since the `2` value has expired, the desired connections will change to 1.
In general, the system is fast to increase and slow to decrease the desired connections.
```

It makes sense to me now :-) Thanks for explaining things. Given the clarification, do these suggestions work? I aimed for consistency in terminology and how we are referring to the states, the lookback, and the connections.

docs/sources/reference/components/prometheus/prometheus.write.queue.md
dehaansa: @clayton-cornell do you think the following might be clearer?
```
`prometheus.write.queue` chooses the highest value in the lookback buffer as the number of desired connections.
```

docs/sources/reference/components/prometheus/prometheus.write.queue.md
mattdurham: Looks good!

clayton-cornell (approved):
I'd say the docs for this new block are fine as-is now. I'll do a pass though later when I update the Prometheus topics for overall style/consistency... for now it matches the other Prometheus topics :-)

PR #2666, opened 2025-02-10
feat(helm): add the ability to set additional service labels

devsunb:
#### PR Description

Adding the ability to add extra labels to the service of the chart.
This can be used for [Linkerd Multicluster Service Export](https://linkerd.io/2.17/tasks/installing-multicluster/#step-3-export-services).

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -7,6 +7,13 @@ This document contains a historical list of changes between releases. Only
 changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
+Unreleased
+----------
+
+### Enhancements
+
+- Add the ability to set additional service labels. (@devsunb)
+
 0.11.0 (2025-01-23)
 ----------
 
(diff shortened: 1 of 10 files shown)


alexander-zimmermann:
+1 — same limitation hits Cilium-based clusters that rely on `CiliumLoadBalancerIPPool.serviceSelector` and `CiliumBGPAdvertisement.selector` to gate which Services receive a LoadBalancer IP and get BGP-advertised. Both
are plain Kubernetes LabelSelectors, so they match on labels only, not on annotations.

Without `service.labels` being honoured on the rendered Service, alloy-receiver Services cannot carry the standard `bgp.cilium.io/ip-pool: default` / `bgp.cilium.io/advertise-service: default` labels and silently fall outside the cluster's normal LB-IPAM / BGP gating. We are currently working around this with a dedicated CiliumLoadBalancerIPPool plus a dedicated advertisement entry that selects on labels the operator already sets (`app.kubernetes.io/part-of: alloy`). This PR would remove both workarounds.

The diff is three template lines plus a default — would be great to see this
revisited.

PR #2675, opened 2025-02-10
feat(pyroscope.receive_http): ensure consistent service_name label handling

marcsanmi:
#### PR Description

This PR ensures that `pyroscope.receive_http` component sets the `service_name` label consistently from `__name__` to match Pyroscope's behavior. When `service_name` already exists, it sets `app_name` from `__name__` instead.

#### Notes to the Reviewer

I decided to duplicate the service name logic in Alloy rather than create a shared abstraction because:
- The logic is simple and stable (see [pkg/ingester/pyroscope/ingest_adapter.go lines 125-149](https://github.com/grafana/pyroscope/blob/048874070c948c05c9209c9054288953034a09ad/pkg/ingester/otlp/ingest_handler.go#L108-L119))
- Both implementations use different label types (Pyroscope's LabelSet vs Prometheus' labels.Labels) and forcing an abstraction would add unnecessary complexity

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ Main (unreleased)
 - Reduce CPU usage of `loki.source.windowsevent` by up to 85% by updating the bookmark file every 10 seconds instead of after every event and by 
   optimizing the retrieval of the process name. (@wildum)
 
+- Ensure consistent service_name label handling in `pyroscope.receive_http` to match Pyroscope's behavior. (@marcsanmi)
+
 ### Bugfixes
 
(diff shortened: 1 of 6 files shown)


simonswine (commented):
I am not entirely sure if that is fully "consistent" with at least my expectations.

I mix both types of APIs (pushv1/ingest). Now I have a relabel rule modifying `service_name` to e.g. let's say append `-grafana-labs`. 

Now pyroscope.write will still use the old value in `__name__` to write the __name__ for the `/ingest` request.

I think maybe it is better if we "rename" the `__name__` to `service_name` and then just before sending we "rename" it back

simonswine (approved):
Thanks for your work on this. LGTM

PR #2706, opened 2025-02-12
Upload devel containers via GitHub Actions

ptodev:
Moving the uploading of devel containers away from Drone. Also updating deployment_tools via GitHub Actions instead of Drone.

I tested the container image steps up by [triggering a build](https://github.com/grafana/alloy/actions/runs/13288192581/job/37101984665?pr=2462) which uploads them as `ptodev/alloy` instead of `grafana/alloy`. As expected, the build ran ok up until the very end when it failed as it couldn't authenticate for `ptodev`.

I have not tested the step which uploads to deployment_tools. I think that's ok, since it's not too crucial. Let's see how the build goes on `main` first.


.github/workflows/publish-alloy-devel.yml
```diff
@@ -0,0 +1,126 @@
...
+          # this is to fix GIT not liking owner of the checkout dir
+          chown -R $(id -u):$(id -g) $PWD
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+        cache: false
+
+    - name: Login to DockerHub (from vault)
```
mattdurham, line 33: Does this automatically handle the secrets for docker?

dehaansa (approved):

mattdurham (approved):
lgtm

PR #2710, opened 2025-02-13
pyroscope.scrape: do not drop __meta labels

### LINKED ISSUE #2629: pyrosope.scrape, pyroscope.write: clarify dropped labels behavior
`pyroscope.scrape` drops all meta labels ( starts with `__meta_*` ) 
`pyroscope.write` drops all private labels (starts with `__`) , except `__name__` and `__delta__`

Both dropping logic is not documented and an implementation detail.

With the new `pyroscope.relabel` component, it may be useful in some cases to NOT drop meta labels, so that they can be used in `pyroscope.relabel` component.

I think we should do one of the following:
- document that `pyroscope.scrape` drops meta labels
- make the `pyroscope.scrape` not drop labels

Maybe we should also document `pyroscope.write` behavior as well

korniltsev:
#### PR Description
Modify pyroscope.scrape to not drop meta labels
Update docs for pyroscope.write to clarify what labels are dropped and preserved

Fixes #2629

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/pyroscope/pyroscope.write.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.write.md
@@ -10,7 +10,9 @@ title: pyroscope.write
 
 # `pyroscope.write`
 
-`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/).
+`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/). When forwarding profiles, most labels starting with double underscore (`__`) are dropped before sending the data, with the following exceptions:
+- `__name__` - preserved as it identifies the profile type
+- `__delta__` - preserved as it's required for delta profiles
 
 You can specify multiple `pyroscope.write` components by giving them different labels.
 
--- a/docs/sources/reference/components/pyroscope/pyroscope.write.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.write.md
@@ -10,9 +10,11 @@ title: pyroscope.write
 
 # `pyroscope.write`
 
-`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/). When forwarding profiles, most labels starting with double underscore (`__`) are dropped before sending the data, with the following exceptions:
-- `__name__` - preserved as it identifies the profile type
-- `__delta__` - preserved as it's required for delta profiles
+`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints.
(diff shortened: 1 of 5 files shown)


docs/sources/reference/components/pyroscope/pyroscope.write.md
```diff
@@ -10,7 +10,9 @@ title: pyroscope.write
 
 # `pyroscope.write`
 
-`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/).
+`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/). When forwarding profiles, most labels starting with double underscore (`__`) are dropped before sending the data, with the following exceptions:
+- `__name__` - preserved as it identifies the profile type
+- `__delta__` - preserved as it's required for delta profiles
```
clayton-cornell, line 15:
```suggestion
`pyroscope.write` receives performance profiles from other components and forwards them to a series of user-supplied endpoints using [Pyroscope' Push API](/oss/pyroscope/).
When `pyroscope.write` forwards profiles, most labels starting with double underscore (`__`) are dropped before the data is sent, with the following exceptions:

* `__name__` is preserved because it identifies the profile type.
* `__delta__`is preserved because it's required for delta profiles.
```


The link to `/oss/pyroscope` lands on https://grafana.com/oss/pyroscope/  Is this really where we want to direct people? Or shoudl we maybe point people at the API docs instead? Maybe here https://grafana.com/docs/pyroscope/latest/reference-server-api/ instead?

korniltsev: I agree we should put a better link to push API, however I'm afraid we don't have the doc for the push API or at least I'm not aware of such docs.

The link you shared is not `Push API`, but `Ingest API` which is doing the same thing, just with different APIS.

The API itself is here https://github.com/grafana/pyroscope/tree/main/api maybe we can link to the source code at least. 

Or remove "Push API" mentioning and the link alltogether

marcsanmi, line 15: If we end up specifying what labels aren't dropped, should we mention `__meta` labels as well?

marcsanmi (approved):
LGTM 👍

docs/sources/reference/components/pyroscope/pyroscope.write.md
clayton-cornell: Updated the suggestion

simonswine: I would also say `all labels [...]` instead of `most labels [...]`.

simonswine (approved):
LGTM

docs/sources/reference/components/pyroscope/pyroscope.write.md
clayton-cornell: Updated the suggestion

PR #2714, opened 2025-02-13
database_observability: fetch indexes and foreign keys info

cristiangreco:
#### PR Description
Extend the tableSpec with indexes and foreign keys info.

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/schema_table.go
+++ b/internal/component/database_observability/mysql/collector/schema_table.go
@@ -461,6 +461,7 @@ func (c *SchemaTable) fetchColumnsDefinitions(ctx context.Context, schemaName st
 			return nil, err
 		}
 
+		// Append column to the last index if it's the same as the previous one (i.e. multi-column index)
 		if nIndexes := len(tblSpec.Indexes); nIndexes > 0 && tblSpec.Indexes[nIndexes-1].Name == indexName {
 			lastIndex := &tblSpec.Indexes[nIndexes-1]
 			if len(lastIndex.Columns) != seqInIndex-1 {
--- a/internal/component/database_observability/mysql/collector/schema_table.go
+++ b/internal/component/database_observability/mysql/collector/schema_table.go
@@ -403,19 +403,19 @@ func (c *SchemaTable) fetchTableDefinitions(ctx context.Context, fullyQualifiedT
 }
 
 func (c *SchemaTable) fetchColumnsDefinitions(ctx context.Context, schemaName string, tableName string) (*tableSpec, error) {
-	rs, err := c.dbConnection.QueryContext(ctx, selectColumnNames, schemaName, tableName)
+	colRS, err := c.dbConnection.QueryContext(ctx, selectColumnNames, schemaName, tableName)
 	if err != nil {
 		level.Error(c.logger).Log("msg", "failed to query table columns", "schema", schemaName, "table", tableName, "err", err)
 		return nil, err
 	}
-	defer rs.Close()
+	defer colRS.Close()
 
 	tblSpec := &tableSpec{Columns: []columnSpec{}}
 
-	for rs.Next() {
+	for colRS.Next() {
 		var columnName, isNullable, columnType, columnKey, extra string
 		var columnDefault sql.NullString
-		if err := rs.Scan(&columnName, &columnDefault, &isNullable, &columnType, &columnKey, &extra); err != nil {
+		if err := colRS.Scan(&columnName, &columnDefault, &isNullable, &columnType, &columnKey, &extra); err != nil {
 			level.Error(c.logger).Log("msg", "failed to scan table columns", "schema", schemaName, "table", tableName, "err", err)
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -119,6 +150,21 @@ type columnSpec struct {
 	DefaultValue  string `json:"default_value,omitempty"`
 }
 
+type indexSpec struct {
+	Name     string   `json:"name"`
+	Type     string   `json:"type"`
+	Columns  []string `json:"columns"`
+	Unique   bool     `json:"unique"`
```
fridgepoet, line 157: my brain got twisted with the name of the variable `Unique` and the column name as `non_unique`. I'm sure you got it the right way, just want to double check with you that it's ok?

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -415,5 +445,71 @@ func (c *SchemaTable) fetchColumnsDefinitions(ctx context.Context, schemaName st
...
+			lastIndex.Columns = append(lastIndex.Columns, columnName)
+		} else {
+			tblSpec.Indexes = append(tblSpec.Indexes, indexSpec{
+				Name:     indexName,
+				Type:     indexType,
+				Columns:  []string{columnName},
+				Unique:   nonUnique == 0,
+				Nullable: nullable.Valid && nullable.String == "YES",
+			})
+		}
```
fridgepoet, line 479: just curious, what's going on here? sorry i lack the context to easily understand what we're trying to do here. maybe this needs to be commented or extracted to a function with a good name.

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -415,5 +445,71 @@ func (c *SchemaTable) fetchColumnsDefinitions(ctx context.Context, schemaName st
 		return nil, err
 	}
 
+	rs, err = c.dbConnection.QueryContext(ctx, selectIndexNames, schemaName, tableName)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to query table indexes", "schema", schemaName, "table", tableName, "err", err)
+		return nil, err
+	}
+	defer rs.Close()
```
fridgepoet, line 453: Is it ok to defer this `rs.Close()` or do we need to be careful to `Close()` before we re-use the `rs` variable again and iterate over the rows for for `selectForeignKeys` (down around line 487)?

internal/component/database_observability/mysql/collector/schema_table_test.go
```diff
@@ -34,7 +34,7 @@ func TestSchemaTable(t *testing.T) {
 		collector, err := NewSchemaTable(SchemaTableArguments{
 			DB:              db,
 			InstanceKey:     "mysql-db",
-			CollectInterval: time.Second,
+			CollectInterval: time.Millisecond,
```
fridgepoet, line 37: just curious, how come the collect interval had to be changed?

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: What do you mean exactly? Later on this is assigned as:

```
Unique: nonUnique == 0
```

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: Good call. For clarity, I'll refactor to use different names.

internal/component/database_observability/mysql/collector/schema_table_test.go
cristiangreco: This is just a drive-by improvement. Tests that expect to loop twice were sitting idle for 1 second. I changed everything down to 1ms so that tests that loop twice will complete much faster (this shaved ~1s from tests duration as the test cases run in parallel anyway).

internal/component/database_observability/mysql/collector/schema_table.go
fridgepoet: I just wanted a double-check that it was assigned correctly as the sense of unique was switched between the column and the variable to which it was scanned, along with interpreting 0.

Got it now from the docs https://dev.mysql.com/doc/refman/8.4/en/information-schema-statistics-table.html
> 0 if the index cannot contain duplicates, 1 if it can.

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: I've added more comments to explain the comparisons.

internal/component/database_observability/mysql/collector/schema_table.go
cristiangreco: I've added a comment about this as well, later on when the value is assigned.

fridgepoet (approved):

matthewnolf (approved):

PR #2716, opened 2025-02-13
database_observability: don't error out for `insert` with `values`

cristiangreco:
#### PR Description
Statements like `insert into table values (...)` were logging an error even though the tables were correctly parsed.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -226,6 +226,8 @@ func (c QuerySample) tablesFromQuery(digest string, stmt sqlparser.Statement) []
 	case *sqlparser.Insert:
 		parsedTables = []string{c.parseTableName(stmt.Table)}
 		switch insRowsStmt := stmt.Rows.(type) {
+		case sqlparser.Values:
+			// ignore raw values
```
cristiangreco, line 230: Don't need to update tests as with this change we're just going to skip a log.

matthewnolf (approved):

PR #2734, opened 2025-02-15
WIP - DO NOT REVIEW RFC: Integrate Beyla without eBPF binaries

rafaelroquetto:
#### PR Description
Beyla will no longer include pre-built eBPF binaries and related artifacts. The main reasons are:
 - prevent malicious pre-compiled binaries from being injected into the Beyla tree
 - removing the binaries ensures that they remain consistent with the source code, preventing discrepancies.
 - repos importing Beyla (i.e. Alloy) won’t need to manage git-lfs, as the removal of pre-built binaries eliminates that requirement and reduces repository overhead
 - as a result, there's no longer need for special Beyla/Alloy release branches and any potential problems that arise with it

The eBPF binaries are now generated by an upstream docker image, that encapsulates all the necessary tooling.

As a result of how `go mod vendor` works by trimming non-used dependencies, and because `go build` can only be used to compile `.go` files, an extra `Makefile` target dependency and an extra "GOTAG" is required as fair _trade-off_.

I have opted to list both the dependency and the extra go-tag close to the place of usage, with the assumption that this will make it easier for their usage scope to be identified. 

Non-Linux compilations should result as a no-op. This is abstracted on the Beyla side and there's no need for explicit checks on the Alloy end, except for the aforementioned extra `Makefile` target.

#### Notes to the Reviewer

Tested with `make alloy` and `make alloy-image`

[Related Beyla PR](https://github.com/grafana/beyla/pull/1666)


rafaelroquetto:
I've rebased and update this branch, but the build is now likely to fail because Beyla uses a more recent version of cilium/ebpf than pyroscope.

PR #2739, opened 2025-02-17
Reorder CI steps so Docker image tag generation can work

ptodev:
This should hopefully fix the CI step which updates deployment_tools.

### DIFF
--- a/.github/workflows/publish-alloy-devel.yml
+++ b/.github/workflows/publish-alloy-devel.yml
@@ -77,22 +77,21 @@ jobs:
     - publish_linux_boringcrypto_container
     steps:
 
-    - name: Checkout code
-      uses: actions/checkout@v4
-
-    - name: Set ownership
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-alloy-devel.yml
```diff
@@ -77,22 +77,21 @@ jobs:
...
     steps:
 
-    - name: Checkout code
-      uses: actions/checkout@v4
-
-    - name: Set ownership
-      # https://github.com/actions/runner/issues/2033#issuecomment-1204205989
-      run: |
-          # this is to fix GIT not liking owner of the checkout dir
-          chown -R $(id -u):$(id -g) $PWD
```
ptodev, line 87: This step isn't necessary because we're not running inside a container.

cristiangreco (approved):
🚀

PR #2751, opened 2025-02-18
[otelcol.connector.servicegraph] Remove seconds from metric names

ptodev:
There was a [change upstream](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34511) to remove the `_seconds` suffix.. OTel has a "unit" property which says the unit (seconds in this case).


clayton-cornell (approved):

ptodev:
@clayton-cornell the upstream change was introduced in [v0.109](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.109.0):

> servicegraphconnector: Fix histogram metrics miss unit

[Alloy v1.4](https://github.com/grafana/alloy/blob/v1.4.0/go.mod#L107) was still running v0.108, which didn't have this change. Alloy v1.5 and v1.6 have been using more recent versions of otel.

PR #2754, opened 2025-02-19
[remotecfg] Add mut.Unlock() calls before returning errors

spartan0x117:
#### PR Description

Releases the `remotecfg` mutex in a few places before returning the error so the lock doesn't remain held.

#### PR Checklist

- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,8 @@ Main (unreleased)
 
 - Fixed a bug where components could be evaluated concurrently without the full context during a config reload (@wildum)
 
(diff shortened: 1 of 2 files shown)


dehaansa (approved):
It feels like the Update() function would be much cleaner if we just locked & deferred unlock at the start of the function.

The only code that currently falls outside of the mutex would be the `s.fetch()` if controller is Ready, which could be put into a defer as well though it would need to add a check for an error. As that's a potentially trickier thing to get right I'm happy to merge this and consider other options in the future.

spartan0x117:
Makes sense, thanks @dehaansa!

PR #2759, opened 2025-02-19
fix filelog test on Windows

wildum:
On Windows, the path would be something like
\"C:\\Users\\William\\AppData\\Local\\Temp\\Test2588855952\\001\\example1761374715\"

This was not correctly parsed by the [syntax ](https://github.com/grafana/alloy/blob/main/syntax/scanner/scanner.go#L502)

By using %q instead of %s will escape the escape chars again, resulting in:
\"C:\\\\Users\\\\William\\\\AppData\\\\Local\\\\Temp\\\\Test2588855952\\\\001\\\\example1761374715\"

This is what the parser actually does normally if you run:

otelcol.receiver.filelog "default" {
	include = ["C:\\Users\\William\\Desktop\\alloytest\\*.log"]
	  output {
      logs = [otelcol.exporter.debug.default.input]
  }
}

it will be scanned as: "C:\\\\Users\\\\William\\\\Desktop\\\\alloytest\\\\*.log

I also tested the component on Windows and I works well

### DIFF
--- a/internal/component/otelcol/receiver/filelog/filelog_test.go
+++ b/internal/component/otelcol/receiver/filelog/filelog_test.go
@@ -31,7 +31,7 @@ func Test(t *testing.T) {
 	require.NoError(t, err)
 
 	cfg := fmt.Sprintf(`
-		include = ["%s"]
+		include = [%q]
 
 		output {
 			// no-op: will be overridden by test code.


internal/component/otelcol/receiver/filelog/filelog_test.go
```diff
@@ -31,7 +31,7 @@ func Test(t *testing.T) {
 	require.NoError(t, err)
 
 	cfg := fmt.Sprintf(`
-		include = ["%s"]
+		include = [%q]
```
ptodev, line 34: Personally, I'd prefer to just use a raw string. But this is ok too.

ptodev (approved):
Thanks!

internal/component/otelcol/receiver/filelog/filelog_test.go
wildum: how would you make it work here? because you need somehow to include the file path

wildum: merging now to unblock the release, happy to modify it later

PR #2760, opened 2025-02-19
docs: Fix uniqueness of top-level configuration blocks

setpill:
#### PR Description

Changes the wording on the uniqueness requirement of top-level configuration blocks to be specifically about _unlabeled_ blocks.

Labeled top-level configuration blocks like `declare` can be specified multiple times, so the current wording is misleading.

#### PR Checklist

- [x] Documentation updated


setpill:
Nevermind, not reading/signing that entire legal word salad to fix a documentation error, I'll just file a bug report instead.

PR #2762, opened 2025-02-19
remotecfg: fix logic around skipping loaded configuration

tpaschalis:
#### PR Description

This PR:

1) Renames `currentConfigHash` to `lastLoadedConfigHash`.
2) Makes sure we _always_ set its value in `parseAndLoad`, before we attempt to pass it the configuration to the controller, and nowhere else
3) Skip over applying configurations that are the same as the previously applied ones.
4) Adds a test
5) Renames get/set helpers to mirror the new name

The other thing to do would be to remove this logic entirely and only depend on the server's response as to whether we should skip applying configurations or not.

Given this is likely a breaking change, this is our chance to choose what to do here before removing the `private-preview` label from the next release. Thoughts?

No issue filed.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -56,14 +56,14 @@ type Service struct {
 
 	ctrl service.Controller
 
-	mut               sync.RWMutex
-	asClient          collectorv1connect.CollectorServiceClient
-	ticker            *jitter.Ticker
-	dataPath          string
-	currentConfigHash string
-	systemAttrs       map[string]string
-	attrs             map[string]string
-	metrics           *metrics
+	mut                  sync.RWMutex
+	asClient             collectorv1connect.CollectorServiceClient
+	ticker               *jitter.Ticker
+	dataPath             string
+	lastLoadedConfigHash string
+	systemAttrs          map[string]string
+	attrs                map[string]string
+	metrics              *metrics
 
 	// This is the hash received from the API. It is used to determine if
 	// the configuration has changed since the last fetch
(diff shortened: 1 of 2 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -480,22 +479,21 @@ func (s *Service) parseAndLoad(b []byte) error {
...
 	if err != nil {
 		return err
 	}
 
 	s.setAstFile(file)
-	s.setCfgHash(getHash(b))
 	return nil
 }
 
 func (s *Service) getCfgHash() string {
```
thampiotr, line 492:
```suggestion
func (s *Service) getLastLoadedConfigHash() string {
```

thampiotr (approved):
only one nit

internal/service/remotecfg/remotecfg.go
```diff
@@ -480,22 +479,21 @@ func (s *Service) parseAndLoad(b []byte) error {
 	if len(b) == 0 {
 		return nil
 	}
-
+	s.setLastLoadedCfgHash(getHash(b))
```
erikbaranowski, line 482: is it possible to get a length 0 config and get stuck doing:

good config -> 0 length config -> good config

erikbaranowski: answered myself: we will never set the 0 length config into the hash OR attempt to load it. looks good!

erikbaranowski (approved):

spartan0x117 (approved):
LGTM!

PR #2807, opened 2025-02-21
database_observability: use TiDB parser for parsing query samples

cristiangreco:
#### PR Description
Add a new experimental option to use the TiDB parser for parsing query.

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


CHANGELOG.md
```diff
@@ -31,11 +31,12 @@ Main (unreleased)
 
 - Added OpenTelemetry logs and metrics support to Alloy mixin's dashboards and alerts. (@thampiotr)
 
-v1.7.1
------------------
-
```
cristiangreco, line 36: Moving this entry as the change "better handling of truncated queries" was not included in 1.7.1.

internal/component/database_observability/mysql/component.go
```diff
@@ -55,10 +55,14 @@ type Arguments struct {
 	ForwardTo         []loki.LogsReceiver `alloy:"forward_to,attr"`
 	EnableCollectors  []string            `alloy:"enable_collectors,attr,optional"`
 	DisableCollectors []string            `alloy:"disable_collectors,attr,optional"`
+
+	// TODO(cristian): experimental, will be removed soon
+	UseTiDBParser bool `alloy:"use_tidb_parser,attr,optional"`
```
cristiangreco, line 60: This will be undocumented because it's temporary.

fridgepoet (approved):

ptodev (commented):
LGTM, but what is the reason for this feature? And why is it temporary?

cristiangreco:
> LGTM, but what is the reason for this feature? And why is it temporary?

Temporary might not be the best word here, sorry. What I mean is that we intend to test it internally and make a decision in relatively short time whether to promote it to be the default or drop it.

internal/component/database_observability/mysql/collector/parser/parser_test.go
matthewnolf, line 1: Maybe we should split this into files for respective parsers?

matthewnolf (commented):
Looks good, couple small comments

internal/component/database_observability/mysql/collector/parser/parser_test.go
matthewnolf: After some discussion, given the 2 parsers share test cases, having a single file means we're not duplicating for the sake of it.

ptodev:
> Temporary might not be the best word here, sorry. What I mean is that we intend to test it internally and make a decision in relatively short time whether to promote it to be the default or drop it.

@cristiangreco I see, thank you. And what is the reason for the new feature?

cristiangreco:
> > Temporary might not be the best word here, sorry. What I mean is that we intend to test it internally and make a decision in relatively short time whether to promote it to be the default or drop it.
> 
> @cristiangreco I see, thank you. And what is the reason for the new feature?

We believe this parser is more feature rich (i.e. better sql syntax support) and easy to use than the one we're currently using.

internal/component/database_observability/mysql/collector/parser/parser_test.go
cristiangreco: Discussed this, we will leave it as is for now.

PR #2814, opened 2025-02-22
pull in fix for walqueue client reuse

mattdurham:
#### PR Description

This includes a fix for `prometheus.write.queue` that reuses the client.

### DIFF
--- a/internal/component/prometheus/write/queue/component.go
+++ b/internal/component/prometheus/write/queue/component.go
@@ -64,9 +64,6 @@ func (s *Queue) Run(ctx context.Context) error {
 			ep.Stop()
 		}
 	}()
-
-	// This needs to be started before we export the onstatechange so that it can accept
-	// signals.
 	for _, ep := range s.endpoints {
 		ep.Start(ctx)
 	}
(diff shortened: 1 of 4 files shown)


internal/component/prometheus/write/queue/component.go
```diff
@@ -32,18 +32,11 @@ func NewComponent(opts component.Options, args Arguments) (*Queue, error) {
 		log:       opts.Logger,
 		endpoints: map[string]promqueue.Queue{},
 	}
-
+	s.opts.OnStateChange(Exports{Receiver: s})
 	err := s.createEndpoints()
 	if err != nil {
 		return nil, err
 	}
-	// This needs to be started before we export the onstatechange so that it can accept
```
mattdurham, line 40: this isnt actually true

dehaansa (approved):

internal/component/prometheus/write/queue/component.go
```diff
@@ -113,7 +110,7 @@ func (s *Queue) Update(args component.Arguments) error {
 		if err != nil {
 			return err
 		}
-		end.Start()
+		end.Start(s.ctx)
```
thampiotr, line 113: This will be nil if Update is called before Run. Will passing nil to Start cause issues?

PR #2822, opened 2025-02-24
AWS Cloudwatch receiver

### LINKED ISSUE #1080: Include AWS CloudWatch Logs reciever
### Request

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.102.0/receiver/awscloudwatchreceiver to be included in Alloy.

### Use case

I need to get logs from CloudWatch into Loki. In addition I need autodiscovery of log groups with some prefix.

Existing solution of `CloudWatch Logs -> Firehose -> Alloy` is too complex and costly to maintain. Also this solution must be reconfigured every time new log group is added or removed, which makes whole solution not viable for my case.

wildum:
#### PR Description

This PR adds a component wrapping around the Otel awscloudwatch receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/awscloudwatchreceiver/README.md

Fixes #1080 

#### Notes to the Reviewer

I tested locally using Localstack and the awscli. I couldn't find a way to change the endpoint so I had to modify the otel component to override the endpoint...

@clayton-cornell the doc is not well formatted, let me know which formatted i should use and I can make the change

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
@@ -33,7 +33,7 @@ otelcol.receiver.awscloudwatch "<LABEL>" {
 
 ## Arguments
 
-The following arguments are supported:
+You can use the following arguments with `otelcol.receiver.awscloudwatch`:
 
 |      Name       |   Type   |           Description            | Default | Required |
 | --------------- | -------- | -------------------------------- | ------- | -------- |
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
@@ -22,7 +22,7 @@ You can specify multiple `otelcol.receiver.awscloudwatch` components by giving t
 ## Usage
 
 ```alloy
-otelcol.receiver.awscloudwatch "LABEL" {
+otelcol.receiver.awscloudwatch "<LABEL>" {
   region = "us-west-2"
 
   output {
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
@@ -6,7 +6,7 @@ labels:
 title: otelcol.receiver.awscloudwatch
 ---
 
-# otelcol.receiver.awscloudwatch
+# `otelcol.receiver.awscloudwatch`
 
 {{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
@@ -35,7 +35,7 @@ otelcol.receiver.awscloudwatch "LABEL" {
 
 The following arguments are supported:
 
-| Name            | Type     | Description                      | Default | Required |
+|      Name       |   Type   |           Description            | Default | Required |
 | --------------- | -------- | -------------------------------- | ------- | -------- |
 | `region`        | `string` | AWS region to collect logs from. |         | yes      |
 | `profile`       | `string` | AWS credentials profile to use.  |         | no       |
@@ -47,7 +47,7 @@ If `imds_endpoint` is not specified, and the environment variable `AWS_EC2_METAD
 
 You can use the following blocks with `otelcol.receiver.awscloudwatch`:
 
-| Block                            | Description                                                                | Required |
+|              Block               |                                Description                                 | Required |
 | -------------------------------- | -------------------------------------------------------------------------- | -------- |
 | [`output`][output]               | Configures where to send received telemetry data.                          | yes      |
 | [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
@@ -63,14 +63,14 @@ The `logs` block configures how logs are collected from CloudWatch.
 
 The following arguments are supported:
 
-| Name                     | Type       | Description                                                    | Default | Required |
+|           Name           |    Type    |                          Description                           | Default | Required |
 | ------------------------ | ---------- | -------------------------------------------------------------- | ------- | -------- |
 | `poll_interval`          | `duration` | How frequently to poll for new log entries.                    | `"1m"`  | no       |
 | `max_events_per_request` | `int`      | Maximum number of events to process per request to Cloudwatch. | `1000`  | no       |
 
 The `logs` block supports the following blocks:
 
-| Block                   | Description                                  | Required |
+|          Block          |                 Description                  | Required |
 | ----------------------- | -------------------------------------------- | -------- |
 | [groups](#logs--groups) | Configures which log groups to collect from. | no       |
 
@@ -78,7 +78,7 @@ The `logs` block supports the following blocks:
 
 The `groups` block supports the following blocks:
 
-| Block                                       | Description                                     | Required |
+|                    Block                    |                   Description                   | Required |
 | ------------------------------------------- | ----------------------------------------------- | -------- |
 | [autodiscover](#logs--groups--autodiscover) | Configures automatic discovery of log groups.   | no       |
 | [named](#logs--groups--named)               | Configures specific log groups to collect from. | no       |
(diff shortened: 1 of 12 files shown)


docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.receiver.awscloudwatch/
+description: Learn about otelcol.receiver.awscloudwatch
+labels:
+  stage: experimental
+title: otelcol.receiver.awscloudwatch
+---
+
+# otelcol.receiver.awscloudwatch
```
clayton-cornell, line 9:
```suggestion
# `otelcol.receiver.awscloudwatch`
```

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
...
+`otelcol.receiver.awscloudwatch` is a wrapper over the upstream OpenTelemetry Collector `awscloudwatch` receiver.
+Bug reports or feature requests will be redirected to the upstream repository, if necessary.
+{{< /admonition >}}
+
+You can specify multiple `otelcol.receiver.awscloudwatch` components by giving them different labels.
+
+## Usage
+
+```alloy
+otelcol.receiver.awscloudwatch "LABEL" {
```
clayton-cornell, line 25:
```suggestion
otelcol.receiver.awscloudwatch "<LABEL>" {
```

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
...
+
+  output {
+    logs = [...]
+  }
+}
+```
+
+## Arguments
+
+The following arguments are supported:
```
clayton-cornell, line 36:
```suggestion
You can use the following arguments with `otelcol.receiver.awscloudwatch`:
```

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
...
+
+|      Name       |   Type   |           Description            | Default | Required |
+| --------------- | -------- | -------------------------------- | ------- | -------- |
+| `region`        | `string` | AWS region to collect logs from. |         | yes      |
+| `profile`       | `string` | AWS credentials profile to use.  |         | no       |
+| `imds_endpoint` | `string` | Custom EC2 IMDS endpoint to use. |         | no       |
```
clayton-cornell, line 42:
```suggestion
| `imds_endpoint` | `string` | Custom EC2 IMDS endpoint to use. |         | no       |
| `profile`       | `string` | AWS credentials profile to use.  |         | no       |
```

Alphabetical order

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
...
+| -------------------------------- | -------------------------------------------------------------------------- | -------- |
+| [`output`][output]               | Configures where to send received telemetry data.                          | yes      |
+| [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
+| [`logs`][logs]                   | Configures the log collection settings.                                    | no       |
+
+[logs]: #logs
+[debug_metrics]: #debug_metrics
+[output]: #output
+
+### logs
```
clayton-cornell, line 60:
```suggestion
### `logs`
```

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
```diff
@@ -0,0 +1,206 @@
...
+| `poll_interval`          | `duration` | How frequently to poll for new log entries.                    | `"1m"`  | no       |
+| `max_events_per_request` | `int`      | Maximum number of events to process per request to Cloudwatch. | `1000`  | no       |
+
+The `logs` block supports the following blocks:
+
+|          Block          |                 Description                  | Required |
+| ----------------------- | -------------------------------------------- | -------- |
+| [groups](#logs--groups) | Configures which log groups to collect from. | no       |
+
+### logs > groups
```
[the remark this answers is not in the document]
wildum: I took the otelcol.processor.resourcedetection as an example

1. true, I don't have a good solution for this one

3. yes that was intentional because the nested blocks are also not documented in the top blocks table in the resourcedetection component and I somehow thought that it was the new style. But I can change it, let me know what's the best

thampiotr:
> I tested locally using Localstack and the awscli. I couldn't find a way to change the endpoint so I had to modify the otel component to override the endpoint...

Perhaps that should be an issue in upstream?

internal/component/otelcol/receiver/awscloudwatch/awscloudwatch.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+	})
+}
+
+// Arguments configures the otelcol.receiver.awscloudwatch component.
+type Arguments struct {
+	Region       string      `alloy:"region,attr"`
+	Profile      string      `alloy:"profile,attr,optional"`
+	IMDSEndpoint string      `alloy:"imds_endpoint,attr,optional"`
+	Logs         *LogsConfig `alloy:"logs,block,optional"`
```
thampiotr, line 39: Nit: since this has a default, like `DebugMetrics`, maybe don't make it a pointer?

internal/component/otelcol/receiver/awscloudwatch/awscloudwatch.go
```diff
@@ -0,0 +1,130 @@
...
+	DebugMetrics otelcolCfg.DebugMetricsArguments `alloy:"debug_metrics,block,optional"`
+
+	// Output configures where to send received data. Required.
+	Output *otelcol.ConsumerArguments `alloy:"output,block"`
+}
+
+var _ receiver.Arguments = Arguments{}
+
+// The defaultLogGroupLimit is not set in the SetToDefault but in the Validate method because
+// the block that contains it is optional.
```
thampiotr, line 50: Do we do the same trick somewhere else? IIRC the mutating Validate was a source of a few issues in the past.

thampiotr (approved):
Only some nits, looks good, thanks!

wildum:
> Perhaps that should be an issue in upstream?

agreed, I created the issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38219

internal/component/otelcol/receiver/awscloudwatch/awscloudwatch.go
wildum: Thanks for pointing this out, I agree that this seems to be the wrong way to handle this. I think that the correct way is to have a SetDefault function on the Autodiscover block. This way if the block is set, then the default will be applied

thampiotr: yep, that sounds better

docs/sources/reference/components/otelcol/otelcol.receiver.awscloudwatch.md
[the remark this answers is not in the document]
clayton-cornell: I haven't edited the otelcol component docs yet. So I haven't gone through any of those topics in any detail to do any markdown cleanup.

internal/component/otelcol/receiver/awscloudwatch/awscloudwatch.go
wildum: changed here: https://github.com/grafana/alloy/pull/2822/commits/43d3982ade4250b23f6ab8d9bfac6a683d6f47b8

internal/component/otelcol/receiver/awscloudwatch/awscloudwatch.go
wildum: I remember now why I did this, it's because it's a very specific scenario where the audiscover block should only set by default if the named block is not set, that's why it cannot be set directly via the SetDefault.
But I took a different approach anyway to avoid changing the Alloy data in the validate: now the change is applied to the converted object, and to still have the validation check I convert + validate in the Alloy validation step. This is a pattern that's used in other otel components that we have such as the kafka exporter or the transform processor

clayton-cornell (approved):
Docs are fine as-is. Technical correct wins over things the users won't really see.  I will pass over this topic later and do some really minor tweaking to align with the way the other topics are documented.  This includes figuring out how we can best document the nested blocks.

thampiotr (approved):
LGTM! 🚢

PR #2828, opened 2025-02-25
Use version 2 of the deploy preview workflow

jdbaldry:
Supports building with multiple documentation sources together (not used here)


clayton-cornell (approved):
Seems to work. Are we good to merge?

PR #2832, opened 2025-02-25
mixin: otel metrics and logs support

thampiotr:
#### PR Description

- Add metrics and logs successes and failures to receivers row in OTel dashboard
- Add metrics and logs successes and failures to exporters row in OTel dashboard
- Make sure these rows have both HTTP and RPC latency tracked
- Add alerts for failing OTLP metrics and logs
- Set up example docker-compose environment to sync up alerts correctly from the mixin

#### Notes

The alerts added correctly when tested locally and look good:
image

image

Dashboard for receivers (some aggregation rules may need tweaking internally)
image

And same for exporters:
image

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/operations/alloy-mixin/alerts/opentelemetry.libsonnet
+++ b/operations/alloy-mixin/alerts/opentelemetry.libsonnet
@@ -1,6 +1,10 @@
 local alert = import './utils/alert.jsonnet';
 
 {
+  local successThreshold = 0.95,
+  local successThresholdText = '95%',
+  local pendingPeriod = '10m',
+
   local successRateQuery(enableK8sCluster, failed, success) =
         local sumBy = if enableK8sCluster then "cluster, namespace, job" else "job";
         |||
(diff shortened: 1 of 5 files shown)


operations/alloy-mixin/alerts/opentelemetry.libsonnet
```diff
@@ -22,19 +26,55 @@ local alert = import './utils/alert.jsonnet';
...
-          'The exporter sending spans success rate is below 95%.',
+          'The exporter sending spans success rate is below %s.' % successThresholdText,
           'The exporter failed to send spans to their destination under job {{ $labels.job }}. There could be an issue with the payload or with the destination endpoint.',
-          '10m',
+          pendingPeriod,
+        ),
+
+        // Metrics exporter alerts
+        alert.newRule(
+          'OtelcolExporterFailedMetrics',
```
mattdurham, line 64: With the new prom changes will this immediately fire?

thampiotr: I don't think it will. In our internal environments the metric aggregation needs tweaking to not aggregate these by instance label. But we update the mixin manually there, so this change won't trigger it.

mattdurham (approved):
LGTM

PR #2834, opened 2025-02-25
fix targets reading from files

thampiotr:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/discovery/target.go
```diff
@@ -188,7 +188,17 @@ func (t Target) ConvertInto(dst interface{}) error {
 		})
 		*dst = result
 		return nil
+	case *map[string]string:
```
thampiotr, line 191: This was the code path required to fix the issue with 
```
ts=2025-02-25T12:04:02.660000145Z level=error msg="failed to evaluate config" controller_path=/ controller_id="" node=prometheus.exporter.snmp.snmp err="decoding configuration: /etc/alloy/config.alloy:204:16: discovery.file.snmp.targets[0] target::ConvertInto: conversion to '*map[string]string' is not supported"
```
see test for more details

internal/component/discovery/target.go
```diff
@@ -211,8 +221,15 @@ func (t *Target) ConvertFrom(src interface{}) error {
 		}
 		*t = NewTargetFromLabelSet(labelSet)
 		return nil
+	case map[string]string:
```
thampiotr, line 224: Adding the other way around just in case we ever need it.

internal/component/discovery/target_test.go
```diff
@@ -117,21 +117,70 @@ func TestEncode_Decode_Targets(t *testing.T) {
...
+
+			t.Run("decode from map into target via scope", func(t *testing.T) {
+				scope := vm.NewScope(map[string]interface{}{"map": tc.input})
+				expr, err := parser.ParseExpression("map")
+				require.NoError(t, err)
+				eval := vm.New(expr)
+				actual := Target{}
+				require.NoError(t, eval.Evaluate(scope, &actual))
+				require.Equal(t, NewTargetFromMap(tc.input), actual)
+			})
```
thampiotr, line 183: These were the missing tests. The issue is not obvious because we technically already have a way to convert from map to Target, because we can do `map -> alloy syntax -> Target`, but that is not how Alloy works, it does not figure out intermediate formats and that's probably good for performance reasons. So an explicit test of `map -> Target` was needed, as well as `Target -> map` is added for any such cases in the future. 

The reason this was an issue in SNMP exporter is that in its arguments instead of Target it uses a de-facto map[string]string here: https://github.com/grafana/alloy/blob/main/internal/component/prometheus/exporter/snmp/snmp.go#L157-L160 

There are other possible such cases, and they could be added in the future, so a solution of allowing this automatic conversion is preferred.

wildum (approved):

PR #2841, opened 2025-02-26
feat: Add string concatenation for secrets

### LINKED ISSUE #1514: Use string concatenation (`+`) on secrets
### Request

Currently, it doesn't seem possible to use an operator like `+` (which works for strings) on `secret` types. It would be good if that kind of thing was available.

### Use case

For example, constructing a `connection_string` in a `prometheus.exporter.mssql` component:

```alloy
remote.kubernetes.secret "sql" {
  namespace = "alloy"
  name = "sql-password"
}

prometheus.exporter.mssql "demo" {
  connection_string = "sqlserver://grafana-alloy:" + remote.kubernetes.secret.sql.data.pwd + "@mssql:1433?trustservercertificate=true"
}
```

The majority of the `connection_string` isn't actually secret, even if the result is, but doing the above results in an error like:

```
Error: /etc/alloy/config.alloy:32:58: remote.kubernetes.secret.sql.data.pwd should be one of [number string] for binop +, got capsule
31 | prometheus.exporter.mssql "demo" {
32 |   connection_string = "sqlserver://grafana-alloy:" + remote.kubernetes.secret.sql.data.pwd + "@mssql:1433?trustservercertificate=true"
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
33 | }
```

Marking the export as `nonsensitive` works:

```alloy
prometheus.exporter.mssql "demo" {
  connection_string = "sqlserver://grafana-alloy:" + nonsensitive(remote.kubernetes.secret.sql.data.pwd) + "@mssql:1433?trustservercertificate=true"
}
```

But that exposes the secret in logs.

Alternatives:

1. A way to "un-capsule" the secret somehow but keep it secret;
2. For the mssql (and the like) component to understand that capsule somehow (e.g., maybe the `username` and `password` fields can be provided separately from the rest of the connection string, and the string constructed internally);
3. Better documentation on how to do the above with existing functionality

Thanks in advance.

ravishankar15:
#### PR Description
Adds string concatenation for the secrets type

Fixes #1514

#### Note to reviewers
I am not sure why we have `secrets` and `optionalsecrets`. I believe secrets was the older implementation. Hence i skipped adding this functionality on the secrets type. Let me know if we have to add this to the secrets type it is easily extendable.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/get-started/configuration-syntax/expressions/operators.md
+++ b/docs/sources/get-started/configuration-syntax/expressions/operators.md
@@ -27,7 +27,7 @@ Operator | Description
 
 Operator | Description
 ---------|-------------------------
-`+`      | Concatenate two strings.
+`+`      | Concatenate two strings/secrets.
 
 ## Comparison operators
 
--- a/docs/sources/get-started/configuration-syntax/expressions/operators.md
+++ b/docs/sources/get-started/configuration-syntax/expressions/operators.md
@@ -27,7 +27,7 @@ Operator | Description
 
 Operator | Description
 ---------|-------------------------
-`+`      | Concatenate two strings/secrets.
+`+`      | Concatenate two strings or two secrets or string and secret.
 
 ## Comparison operators
 
--- a/docs/sources/get-started/configuration-syntax/expressions/operators.md
+++ b/docs/sources/get-started/configuration-syntax/expressions/operators.md
@@ -27,7 +27,7 @@ Operator | Description
 
 Operator | Description
 ---------|-------------------------
-`+`      | Concatenate two strings or two secrets or string and secret.
+`+`      | Concatenate two strings or two secrets, or a string and a secret.
 
 ## Comparison operators
 
--- a/syntax/vm/op_binary.go
+++ b/syntax/vm/op_binary.go
@@ -19,7 +19,7 @@ func evalBinop(lhs value.Value, op token.Token, rhs value.Value) (value.Value, e
 	)
 
 	if lhs.Type() == value.TypeCapsule || rhs.Type() == value.TypeCapsule {
-		lhs, rhs = mapCapsules(lhs, rhs)
+		lhs, rhs = tryUnwrapCapsules(lhs, rhs)
 	}
 
 	// TODO(rfratto): evalBinop should check for underflows and overflows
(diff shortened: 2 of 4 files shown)


wildum (commented):
Thanks for your contribution, I think that this is the right approach and it works well for optional secrets. 
We do still have values of the type "Secret", for example the [remote.value](https://grafana.com/docs/alloy/latest/reference/components/remote/remote.vault/#exported-fields) component will export secrets and not optional secrets. It would be nice to also have concatenation for secrets and not only optional secrets. optional secret + secret = secret and string + secret = secret.
The type casts and switch make the code hard to follow and I'm afraid that it might become even harder with the secret extension. Maybe you can cast the types at the beginning of the map function:
```
lhsOptSecret, lhsIsOptSecret := lhs.Interface().(alloytypes.OptionalSecret)
lhsSecret, lhsIsSecret := lhs.Interface().(alloytypes.Secret)
etc...
```
That's just an idea, if you find a better way please go ahead.
One last little thing: can you update the doc [here](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/operators/#string-operators) with `Concatenate strings/secrets`

ravishankar15:
I see your thought about nesting getting complicated. I wanted to make this implementation little generic so that it can be extended to any type of capsules in future. I have split the handling to separate functions with comments. Let me know if this looks over engineered and we wanted to stick to handling only secrets for now.

syntax/vm/op_binary.go
```diff
@@ -254,19 +274,92 @@ func evalBinop(lhs value.Value, op token.Token, rhs value.Value) (value.Value, e
...
+	switch lhs.Interface().(type) {
+	case alloytypes.OptionalSecret:
+		return tryUnwrapOptionalSecrets(lhs, rhs)
+	case alloytypes.Secret:
+		return tryUnwrapSecret(lhs, rhs)
+	case string:
+		return tryUnwrapString(lhs, rhs)
+	default:
+		return lhs, rhs
+	}
```
wildum, line 289: A few suggestions to simplify the code a little:
- since that you are casting the type of lhs in the function above, you can pass lhs typed directly to avoid casting it again in the nested tryUnwrap functions
- instead of having the explanations at the top of the function, I would recommend adding one comment line above every return like this:

```
// Secret, OptionalSecret = Secret, Secret
return lhs, value.Encapsulate(alloytypes.Secret(rhsValue.Value))
```

docs/sources/get-started/configuration-syntax/expressions/operators.md
```diff
@@ -27,7 +27,7 @@ Operator | Description
 
 Operator | Description
 ---------|-------------------------
-`+`      | Concatenate two strings.
+`+`      | Concatenate two strings/secrets.
```
clayton-cornell, line 30: Can you mix/match strings and secrets? or is it... two strings or two secrets?

We should avoid using the slash in the description... so here it'd be something like this:

```suggestion
`+`      | Concatenate two strings or two secrets.
```

Or maybe `strings or secrets`, depending on the definition

ravishankar15: Yeah we can mix and match I have updated the sentence accordingly.

syntax/vm/op_binary.go
ravishankar15: I have made the changes. Please take a look

wildum (approved):
LGTM, thanks!

PR #2846, opened 2025-02-26
`CI`: Move login-to-gar after image tag creation

dsotirakis:
#### PR Description

Moves the `login-to-gar` step after the image-tag is done, so the generated files won't interfere with git magic.

Also pins the version, according to the [latest docs](https://github.com/grafana/shared-workflows/tree/main/actions/login-to-gar).

[Failure on main](https://github.com/grafana/alloy/actions/runs/13543101477/job/37849440397).

### DIFF
--- a/.github/workflows/publish-alloy-devel.yml
+++ b/.github/workflows/publish-alloy-devel.yml
@@ -101,6 +101,10 @@ jobs:
         echo "$(bash ./tools/image-tag-docker)" > .tag-only
         echo "grafana/alloy-dev:$(bash ./tools/image-tag-docker)" > .image-tag
 
+      # This step needs to run after "Checkout code".
+      # That's because the login to GAR generates a new file.
+      # We don't want this file to end up in the repo directory.
+      # Then "tools/image-tag" would get confused because "git status" no longer reports a clean repo.
     - name: Log in to Google Artifact Registry
       uses: grafana/shared-workflows/actions/login-to-gar@login-to-gar-v0.2.2
       with:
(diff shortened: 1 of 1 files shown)


ptodev (approved):

.github/workflows/publish-alloy-devel.yml
```diff
@@ -112,6 +101,12 @@ jobs:
         echo "$(bash ./tools/image-tag-docker)" > .tag-only
         echo "grafana/alloy-dev:$(bash ./tools/image-tag-docker)" > .image-tag
 
+    - name: Log in to Google Artifact Registry
```
ptodev, line 104:
```suggestion
      # This step needs to run after "Checkout code".
      # That's because the login to GAR generates a new file.
      # We don't want this file to end up in the repo directory.
      # Then "tools/image-tag" would get confused because "git status" no longer reports a clean repo.
    - name: Log in to Google Artifact Registry
```

.github/workflows/publish-alloy-devel.yml
```diff
@@ -112,6 +101,16 @@ jobs:
         echo "$(bash ./tools/image-tag-docker)" > .tag-only
         echo "grafana/alloy-dev:$(bash ./tools/image-tag-docker)" > .image-tag
 
+      # This step needs to run after "Checkout code".
```
ptodev, line 104:
```suggestion
      # This step needs to run after "Get the image tag".
```

ptodev:
LGTM, I just added a comment to say why the step is there

PR #2860, opened 2025-02-27
feat: Add new stdlib function to_json

### LINKED ISSUE #152: syntax: Add `json_encode` to stdlib
### Request

Add a function to the standard library similar to `json_decode` but in the reverse direction. Taking a map type and encoding to a string.

### Use case

In order to configure the embedded blackbox, it would be nice to be able to use the powerful features of Alloy syntax while providing the inline configuration.

```
prometheus.exporter.blackbox "example" {
  config = json_encode({
    modules = {
      http_2xx = {
        prober = "http"
        http = {
          headers = {
            "Authorization" = sys.env("HTTP_AUTH_HEADER")
          }
        }
      }
    }
  })

  target "example" {
    address = "http://example.com"
    module  = "http_2xx"
  }

  target "grafana" {
    address = "http://grafana.com"
    module  = "http_2xx"
  }
}
```

ravishankar15:
#### PR Description
Adds a new stdlib function for json encoding `to_json` which converts a map to JSON

Fixes #152 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ Main (unreleased)
 
 - Added additional backwards compatibility metrics to `prometheus.write.queue`. (@mattdurham)
 
-- Add new stdlib functions to_json (@ravishankar15)
+- Add new stdlib functions encoding.to_json (@ravishankar15)
 
 - Added OpenTelemetry logs and metrics support to Alloy mixin's dashboards and alerts. (@thampiotr)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ Main (unreleased)
 
 - Added additional backwards compatibility metrics to `prometheus.write.queue`. (@mattdurham)
 
+- Add new stdlib functions to_json (@ravishankar15)
+
 - Added OpenTelemetry logs and metrics support to Alloy mixin's dashboards and alerts. (@thampiotr)
 
 - Add support for proxy and headers in `prometheus.write.queue`. (@mattdurham)
--- a/docs/sources/reference/stdlib/encoding.md
+++ b/docs/sources/reference/stdlib/encoding.md
@@ -63,18 +63,33 @@ c3RyaW5nMTIzIT8kKiYoKSctPUB-
 
 ## encoding.to_json
 
-The `encoding.to_json` function encodes a map into a JSON.
-`encoding.to_json` fails if the map argument provided can't be parsed as json string.
+The `encoding.to_json` function encodes the alloy type into a JSON string.
+`encoding.to_json` fails if the input argument provided can't be parsed as JSON string.
 
-A common use case of `encoding.to_json` is to encode a config of component which is expected to be a JSON
-For example, `config` of a [`prometheus.exporter.blackbox`][]
+A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string
+For example, `config` param of [`prometheus.exporter.blackbox`][]
+
+{{< admonition type="note" >}}
+It produces inconsistent behaviour for capsules and secrets.
+{{< /admonition >}}
 
 ### Examples
 
 ```alloy
-> encoding.to_json({"key"={"key1"="val1"}, "key2"=2})
-{"foo":{"foo1":"Hello!"},"key":12}
+> encoding.to_json({"example"={"example1"="Hello!"}, "key"=12})
+{\"example\":{\"example1\":\"Hello!\"},\"key\":12}
 
+> encoding.to_json([0,1,2])
+"[0,1,2]"
+
+> encoding.to_json(15)
+"15"
+
+> encoding.to_json("Hello World")
+\"Hello World\"
+
+> encoding.to_json(null)
+"null"
 ```
 
 ## encoding.from_json
--- a/docs/sources/reference/stdlib/encoding.md
+++ b/docs/sources/reference/stdlib/encoding.md
@@ -66,8 +66,8 @@ c3RyaW5nMTIzIT8kKiYoKSctPUB-
 The `encoding.to_json` function encodes the map into a JSON string.
 `encoding.to_json` fails if the input argument provided can't be parsed as JSON string.
 
-A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string
-For example, `config` param of [`prometheus.exporter.blackbox`][]
+A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string.
+For example, `config` argument of [`prometheus.exporter.blackbox`][].
 
 ### Examples
 
--- a/docs/sources/reference/stdlib/encoding.md
+++ b/docs/sources/reference/stdlib/encoding.md
@@ -72,9 +72,8 @@ For example, `config` param of [`prometheus.exporter.blackbox`][]
 ### Examples
 
 ```alloy
-> encoding.to_json({"example"={"example1"="Hello!"}, "key"=12})
-{\"example\":{\"example1\":\"Hello!\"},\"key\":12}
-
+> encoding.to_json({"modules"={"http_2xx"={"prober"="http","timeout"="5s","http"={"headers"={"Authorization"=sys.env("TEST_VAR")}}}}})
+"{\"modules\":{\"http_2xx\":{\"http\":{\"headers\":{\"Authorization\":\"Hello!\"}},\"prober\":\"http\",\"timeout\":\"5s\"}}}"
 ```
 
 ## encoding.from_json
--- a/docs/sources/reference/stdlib/encoding.md
+++ b/docs/sources/reference/stdlib/encoding.md
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
 c3RyaW5nMTIzIT8kKiYoKSctPUB-
 ```
 
+## encoding.to_json
+
+The `encoding.to_json` function encodes a map into a JSON.
+`encoding.to_json` fails if the map argument provided can't be parsed as json string.
+
+A common use case of `encoding.to_json` is to encode a config of component which is expected to be a JSON
+For example, `config` of a [`prometheus.exporter.blackbox`][]
+
+### Examples
+
+```alloy
+> encoding.to_json({"key"={"key1"="val1"}, "key2"=2})
+{"foo":{"foo1":"Hello!"},"key":12}
+
+```
+
 ## encoding.from_json
 
 The `encoding.from_json` function decodes a string representing JSON into an {{< param "PRODUCT_NAME" >}} value.
@@ -125,4 +141,5 @@ null
 "Hello, world!"
 ```
 
-[`local.file`]: ../../components/local/local.file/
+[`local.file`]: ../components/local/local.file/
+[`prometheus.exporter.blackbox`]: ../components/prometheus/prometheus.exporter.blackbox
--- a/docs/sources/reference/stdlib/encoding.md
+++ b/docs/sources/reference/stdlib/encoding.md
@@ -63,15 +63,11 @@ c3RyaW5nMTIzIT8kKiYoKSctPUB-
 
 ## encoding.to_json
 
-The `encoding.to_json` function encodes the alloy type into a JSON string.
+The `encoding.to_json` function encodes the map into a JSON string.
 `encoding.to_json` fails if the input argument provided can't be parsed as JSON string.
 
 A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string
 For example, `config` param of [`prometheus.exporter.blackbox`][]
-
-{{< admonition type="note" >}}
-It produces inconsistent behaviour for capsules and secrets.
-{{< /admonition >}}
 
 ### Examples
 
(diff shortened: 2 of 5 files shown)


docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
 c3RyaW5nMTIzIT8kKiYoKSctPUB-
 ```
 
+## encoding.to_json
+
+The `encoding.to_json` function encodes a map into a JSON.
```
wildum, line 66: Thanks for your contribution, the code looks good but it looks like it can handle other types that map. Could you add tests for the different [types](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/) and document the ones that don't work? (Not sure what happens if you try to encode a Secret or a Capsule for example)

clayton-cornell, line 66:
```suggestion
The `encoding.to_json` function encodes a map into a JSON string.
```

Is string the right word here?

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+`encoding.to_json` fails if the map argument provided can't be parsed as json string.
```
clayton-cornell, line 67:
```suggestion
`encoding.to_json` fails if the map argument provided can't be parsed as JSON string.
```

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+
+A common use case of `encoding.to_json` is to encode a config of component which is expected to be a JSON
```
clayton-cornell, line 69:
```suggestion
A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string.
```

JSON string?

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+For example, `config` of a [`prometheus.exporter.blackbox`][]
```
clayton-cornell, line 70:
```suggestion
For example, configuration of a [`prometheus.exporter.blackbox`][]
```

Missing a word here... a `prometheus.exporter.blackbox` string?

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+
+### Examples
+
+```alloy
+> encoding.to_json({"key"={"key1"="val1"}, "key2"=2})
+{"foo":{"foo1":"Hello!"},"key":12}
```
clayton-cornell, line 76:
```suggestion
> encoding.to_json({"key"={"key1"="val1"}, "key2"=2})
{"example":{"example1":"Hello!"},"key":12}
```

We should avoid using `foo`, `bar`, and `baz` in the documentation. If `example` isn't appropriate here, then use some other generic placeholder.  (I'm working on removing foo, bar, baz from other places in the docs)

docs/sources/reference/stdlib/encoding.md
ravishankar15: I have updated the tests and documentation. For Secret and Capsule the function converts the public fields of the underlying struct. I updated the doc with the closest understandable text. Should we handle this differently ?

Example configuration the I tried for verifying the behaviour,
```
discovery.relabel "keep_backend_only" {
  targets = [
    { "__meta_foo" = "foo", "__address__" = "localhost", "instance" = "ones",   "app" = "backend" },
    { "__meta_bar" = "bar", "__address__" = "localhost", "instance" = "two",   "app" = "database" },
    { "__meta_baz" = "baz", "__address__" = "localhost", "instance" = "three", "app" = "frontend" },
  ]

  rule {
    source_labels = ["__address__", "instance"]
    separator     = "/"
    target_label  = "destination"
    action        = "replace"
  }

  rule {
    source_labels = ["app"]
    action        = "keep"
    regex         = "backend"
  }
}

declare "discovery_result" {
  argument "output_value" {
    optional = false
  }

  argument "rules_value" {
    optional = false
  }

  argument "number_value" {
    optional = true
    default = 14
  }

  argument "array_value" {
    optional = true
  }

  argument "string_value" {
    optional = true
  }

  argument "null_value" {
    optional = true
  }
}

discovery_result "backend" {
  output_value = encoding.to_json(discovery.relabel.keep_backend_only.output)
  rules_value = encoding.to_json(discovery.relabel.keep_backend_only.rules)
  number_value = encoding.to_json(23)
  array_value = encoding.to_json([1,2,3])
  string_value = encoding.to_json("Hello World")
  null_value = encoding.to_json(null)
}
```

docs/sources/reference/stdlib/encoding.md
ravishankar15: I wanted to refer the param name in the component updated accordingly.

docs/sources/reference/stdlib/encoding.md
wildum: Thanks for testing it, I think that the best approach would be to prevent capsule and secret/optional secrets from being converted because the only way to expose a secret should be through the nonsensitive function + I don't see a good usage to get the public fields exported of the capsule for now.

ravishankar15: I made the changes to support only map type instead of complicating the support for all other types.

The way i see it,
1. we can add conditions on top of map[string]interface{} for other types int, float, string etc,
2. We can have a recursive logic to convert the applicable types [types](https://grafana.com/docs/alloy/latest/get-started/configuration-syntax/expressions/types_and_values/) since a map can have have nested maps and arrays can have map nested.

Let me know your thoughts or if we have any other approaches here ?

syntax/internal/stdlib/stdlib.go
```diff
@@ -324,6 +325,18 @@ func base64URLDecode(in string) (interface{}, error) {
 	return decoded, nil
 }
 
+func jsonEncode(in interface{}) (interface{}, error) {
+	v, ok := in.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("only supports map")
```
wildum, line 331:
```suggestion
		return nil, fmt.Errorf("jsonEncode only supports map")
```

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,22 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+
+A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string
+For example, `config` param of [`prometheus.exporter.blackbox`][]
+
+### Examples
+
+```alloy
+> encoding.to_json({"example"={"example1"="Hello!"}, "key"=12})
+{\"example\":{\"example1\":\"Hello!\"},\"key\":12}
+
```
wildum, line 77: could you add an example with prometheus.exporter.blackbox please?

syntax/vm/vm_stdlib_test.go
```diff
@@ -42,6 +42,7 @@ func TestVM_Stdlib(t *testing.T) {
 		{"encoding.from_URLbase64", `encoding.from_URLbase64("c3RyaW5nMTIzIT8kKiYoKSctPUB-")`, string(`string123!?$*&()'-=@~`)},
 		{"encoding.to_base64", `encoding.to_base64("string123!?$*&()'-=@~")`, string(`c3RyaW5nMTIzIT8kKiYoKSctPUB+`)},
 		{"encoding.to_URLbase64", `encoding.to_URLbase64("string123!?$*&()'-=@~")`, string(`c3RyaW5nMTIzIT8kKiYoKSctPUB-`)},
+		{"encoding.to_json object", `encoding.to_json({"foo" = {"foo1" = sys.env("TEST_VAR")}, "key"=12})`, string(`{"foo":{"foo1":"Hello!"},"key":12}`)},
```
wildum, line 45: could you add a test for the error scenario where the input is not a map please?

wildum (commented):
Thanks, I agree that only supporting map is better and if users need another type we can always extend it later :)

docs/sources/reference/stdlib/encoding.md
wildum: good idea, I don't think that the recursion is needed for now, let's keep it simple and extend it if needed in the future

wildum (approved):
LGTM, thanks! Over to @clayton-cornell for another doc review

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,21 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
 c3RyaW5nMTIzIT8kKiYoKSctPUB-
 ```
 
+## encoding.to_json
+
+The `encoding.to_json` function encodes the map into a JSON string.
+`encoding.to_json` fails if the input argument provided can't be parsed as JSON string.
+
+A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string
```
clayton-cornell, line 69:
```suggestion
A common use case for `encoding.to_json` is to encode a configuration of component which is expected to be a JSON string.
```

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,21 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
...
+For example, `config` param of [`prometheus.exporter.blackbox`][]
```
clayton-cornell, line 70:
```suggestion
For example, `config` param of [`prometheus.exporter.blackbox`][].
```

I'm not sure what "`config` param" means in this context. 

Can this be reworked as:

For example, a configuration parameter in [`prometheus.exporter.blackbox`][].

Does this mean the same thing?

clayton-cornell (approved):
Two small doc changes. If the question I have on line 70 is "it depends" then we can leave it as-is for now and just add the two missing `.` on lines 69 and 70.

docs/sources/reference/stdlib/encoding.md
ravishankar15: `prometheus.exporter.blackbox` component accepts a argument called `config` which is expected to be a JSON string.  May be changing the word param can be more appropriate I believe. https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.blackbox/#arguments

docs/sources/reference/stdlib/encoding.md
```diff
@@ -61,6 +61,21 @@ The `encoding.to_base64` function encodes the original string into RFC4648-compl
 c3RyaW5nMTIzIT8kKiYoKSctPUB-
 ```
 
+## encoding.to_json
+
+The `encoding.to_json` function encodes the map into a JSON string.
+`encoding.to_json` fails if the input argument provided can't be parsed as JSON string.
```
clayton-cornell, line 67:
```suggestion
`encoding.to_json` fails if the input argument provided can't be parsed as a JSON string.
```

clayton-cornell (approved):
Just one small tweak in the docs - a missing word.

syntax/internal/stdlib/stdlib.go
```diff
@@ -324,6 +325,18 @@ func base64URLDecode(in string) (interface{}, error) {
 	return decoded, nil
 }
 
+func jsonEncode(in interface{}) (interface{}, error) {
+	v, ok := in.(map[string]interface{})
```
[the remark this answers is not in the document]
ravishankar15: I have changed the return type and added test for the target_test. I am not entirely sure if TryConvertToObject can be helpful here ? Its for capsules but our input is object I think.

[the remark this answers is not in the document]
wildum: The targets were refactored to be capsules under the hood to improve performance, but they behave like maps. But maybe having the TryConvertToObject function opens the door again to weird scenario so we should maybe just keep it simple and not have it

internal/component/discovery/target_test.go
```diff
@@ -152,6 +152,45 @@ func TestNestedIndexing(t *testing.T) {
...
+			evalError := eval.Evaluate(scope, &decodeInto)
+			if tc.expectedEvalError != "" {
+				require.ErrorContains(t, evalError, tc.expectedEvalError)
+			} else {
+				require.NoError(t, evalError)
+			}
+			require.Equal(t, tc.expected, decodeInto)
+		})
+	}
+}
```
wildum, line 192: this test is not testing the target type. But if we don't support the targets you can remove completely

syntax/internal/stdlib/stdlib.go
```diff
@@ -345,6 +347,19 @@ func base64Encode(in string) (string, error) {
 	return encoded, nil
 }
 
+func jsonEncode(in interface{}) (string, error) {
+	v, ok := in.(map[string]interface{})
+	fmt.Println(reflect.TypeOf(in))
```
wildum, line 352:
```suggestion
```

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Main (unreleased)
 
 - Added additional backwards compatibility metrics to `prometheus.write.queue`. (@mattdurham)
 
+- Add new stdlib functions to_json (@ravishankar15)
```
wildum, line 38:
```suggestion
- Add new stdlib functions encoding.to_json (@ravishankar15)
```

wildum:
@ravishankar15 thanks for the changes, sorry for the trouble I guess it's not worth it to support the new Targets for now. I just left a few small comments. When resolved I will be happy to merge it

internal/component/discovery/target_test.go
ravishankar15: I think it makes sense since Target is not supported I have removed the test

wildum (approved):
Thanks!

PR #2868, opened 2025-02-27
Upgrade to prometheus 2.55

ptodev:
#### PR Description

This is an incremental step towards an upgrade to Prometheus 3.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -142,7 +142,7 @@ func (c *crdManager) Run(ctx context.Context) error {
 	// Start prometheus scrape manager.
 	alloyAppendable := prometheus.NewFanout(c.args.ForwardTo, c.opts.ID, c.opts.Registerer, c.ls)
 	opts := &scrape.Options{}
-	c.scrapeManager, err = scrape.NewManager(opts, c.logger, alloyAppendable, unregisterer)
+	c.scrapeManager, err = scrape.NewManager(opts, c.logger, nil, alloyAppendable, unregisterer)
```
mattdurham, line 145: Whats the nil value here?

internal/component/prometheus/scrape/scrape.go
```diff
@@ -71,6 +74,8 @@ type Arguments struct {
 	ScrapeClassicHistograms bool `alloy:"scrape_classic_histograms,attr,optional"`
 	// Whether to scrape native histograms.
 	ScrapeNativeHistograms bool `alloy:"scrape_native_histograms,attr,optional"`
+	// File to which scrape failures are logged.
+	ScrapeFailureLogFile string `alloy:"scrape_failure_log_file,attr,optional"`
```
mattdurham, line 78: I wonder if we can make this more generic, before we commit to this. Looking at the callback this is the filename in the NewManager callback, could we override that with a new entity that would allow use to have a generic output and then have a `local.output.file`, output to logs, and other outputs.

mattdurham (commented):
Looks reasonable, but want to take a second to talk about scrape failure log before committing to it.

clayton-cornell (approved):
Approving for docs

mattdurham (approved):
LGTM

internal/component/prometheus/scrape/scrape.go
ptodev: As we discussed offline, we could polish this further at a later date. Given that it's a feature for debugging, breaking changes to it would be fine. My personal opinion is that:
* It'd be best to integrate this with the deb logs of the component. But I haven't done it yet since I'm not sure if it'd spam too much, or if there's some other reason for having a separate file that I could be missing.
* I'd rather not add this feature to live debugging since live debugging is for seeing the data flowing through a component. It's good to have a clear separation between this and the component's own telemetry.
* I'd prefer not to have a complex way to send this log to other components... it's just part of the component's own logs and we should try treat it the same way as its regular logs.

Ideally we should have the ability to set component-level debug logging, and maybe even set what sort of features in a component should be doing debug logging. Then we can integrate such debug logs very easily.

PR #2893, opened 2025-03-03
fix: cloudwatch check task

dehaansa:
#### PR Description
Fix issues with the docs make task for checking cloudwatch services after updates to the docs.


clayton-cornell (approved):
Tested with `make docs` and it works as it is supposed to

PR #2896, opened 2025-03-04
hackathon: cluster targets query

thampiotr:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/web/api/api.go
```diff
@@ -50,7 +56,11 @@ func (a *AlloyAPI) RegisterRoutes(urlPrefix string, r *mux.Router) {
 	r.Handle(path.Join(urlPrefix, "/remotecfg/components/{id:.+}"), httputil.CompressionHandler{Handler: getComponentHandlerRemoteCfg(a.alloy)})
 
 	r.Handle(path.Join(urlPrefix, "/peers"), httputil.CompressionHandler{Handler: getClusteringPeersHandler(a.alloy)})
+
 	r.Handle(path.Join(urlPrefix, "/debug/{id:.+}"), liveDebugging(a.alloy, a.CallbackManager))
+
+	r.Handle(path.Join(urlPrefix, "/tools/instance-prom-targets-debug-info"), httputil.CompressionHandler{Handler: getInstanceTargetDebugInfo(a.alloy)})
+	r.Handle(path.Join(urlPrefix, "/tools/cluster-prom-targets-debug-info"), httputil.CompressionHandler{Handler: getClusterTargetDebugInfo(a.alloy)})
```
jharvey10, line 63: Would love to work with you on this to see if we can bring together the logic you have here with the graphql API layer from the REPL hackathon. Potentially allowing querying of this via something like:
```graphql
query {
  components(type: "prometheus.scrape") {
    targets {
      jobName
      endpoint {
        hostname
        ip
        port
      }
      health
      labels
      lastError
      lastScrapeDate
      lastScrapeDuration
    }
  }
}
```

wonder if we could also get fancy and do the clustering stuff as well like:

```graphql
query {
  alloy {
    cluster {
      members {
        id
        components(type: "prometheus.scrape") {
          targets {
            # ...
          }
        }
      }
    }
  }
}
```

PR #2914, opened 2025-03-05
fix targets indexing and field access

thampiotr:
#### PR Description

Fixes an issue where indexing a target's specific key would not work. I added tests and ran them on main and it reproduced the problem. Since `foreach` was affected, adding an extra test for it too.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/syntax/vm/vm.go
+++ b/syntax/vm/vm.go
@@ -356,6 +356,27 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, 
 		}
 
 		switch val.Type() {
+		case value.TypeCapsule:
+			if val.Implements(reflect.TypeFor[value.ConvertibleIntoCapsule]()) {
+				// Check if this capsule can be converted into Alloy object for more detailed description:
+				newVal := make(map[string]value.Value)
+				if err := val.ReflectAddr().Interface().(value.ConvertibleIntoCapsule).ConvertInto(&newVal); err == nil {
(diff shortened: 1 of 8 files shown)


syntax/vm/vm.go
```diff
@@ -356,6 +356,27 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node,
 		}
 
 		switch val.Type() {
+		case value.TypeCapsule:
+			if val.Implements(reflect.TypeFor[value.ConvertibleIntoCapsule]()) {
```
wildum, line 360: shouldnt it check if it's ConvertibleFromCapsule since that you are converting from a capsule to a map?

thampiotr: The names are in some way reversed and it was my first impression too. See the interface:
```
// ConvertibleIntoCapsule is a Capsule which supports custom conversion rules
// into any Go type which is not the same as the capsule type.
type ConvertibleIntoCapsule interface {
	Capsule

	// ConvertInto should convert its value and store it into dst. dst will be a
	// pointer to a value which ConvertInto is expected to update.
	//
	// ConvertInto should return ErrNoConversion if no conversion into dst is
	// available.
	ConvertInto(dst interface{}) error
}

```

If you think about it for long enough, both can make sense. 

But there's a test to prove that this works.

wildum (approved):

PR #2917, opened 2025-03-06
fix more stdlib functions working with targets

thampiotr:
#### PR Description

- Fixes an issue with `coalesce` and `combine_maps` stdlib functions.
- Adds `val.TryConvertToObject()` to simplify code
- Adds tests and fixes some minor misc things in the code

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/syntax/internal/stdlib/stdlib.go
+++ b/syntax/internal/stdlib/stdlib.go
@@ -18,15 +18,13 @@ import (
 	"github.com/grafana/alloy/syntax/internal/value"
 )
 
-// TODO: refactor the stdlib to have consistent naming between namespaces and identifiers.
-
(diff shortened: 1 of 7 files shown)


syntax/internal/stdlib/stdlib.go
```diff
@@ -309,33 +315,33 @@ func yamlDecode(in string) (interface{}, error) {
 	return res, nil
 }
 
-func base64Decode(in string) (interface{}, error) {
+func base64Decode(in string) ([]byte, error) {
```
thampiotr, line 318: These were unnecessarily untyped making it harder to read and reason about the code. I've given them a stronger type constraint to prevent bugs.

wildum (approved):

PR #2918, opened 2025-03-06
feat(pyroscope.write): Add labels validation

marcsanmi:
#### PR Description

This PR adds validation for labels in the `pyroscope.write` component to ensure compatibility with Pyroscope's validation requirements. The validation:

- Checks for duplicate labels
- Validates label names & values according to Pyroscope's standards

https://github.com/grafana/pyroscope-squad/issues/357

#### Notes to the Reviewer

- I don't check for `service_name` label existence. As talked, as far as using the `receive_http`, this should come for free with our `ensureServiceName` function and also validated at pyroscope side.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -498,14 +498,11 @@ func validateLabels(lbls labels.Labels) error {
 		}
 
 		// Skip label name validation for pyroscope reserved labels
-		if l.Name == pyroscope.LabelName {
-			lastLabelName = l.Name
-			continue
-		}
-
-		// Validate label name
-		if err := labelset.ValidateLabelName(l.Name); err != nil {
-			return fmt.Errorf("invalid label name: %w", err)
+		if l.Name != pyroscope.LabelName {
+			// Validate label name
+			if err := labelset.ValidateLabelName(l.Name); err != nil {
+				return fmt.Errorf("invalid label name: %w", err)
+			}
 		}
 
 		lastLabelName = l.Name
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/component/pyroscope/write/write.go
```diff
@@ -466,3 +476,36 @@ func ensureNameMatchesService(lbls labels.Labels) labels.Labels {
...
 	return lbls
 }
+
+// validateLabels checks for valid labels and doesn't contain duplicates.
+func validateLabels(lbls labels.Labels) error {
+	if lbls.Len() == 0 {
+		return labelset.ErrServiceNameIsRequired
+	}
+
+	seen := make(map[string]struct{}, lbls.Len())
```
simonswine, line 486: Nit: You could also go for the approach of sorting the label slice by name and then checking for duplicates like the backend.

internal/component/pyroscope/write/write.go
```diff
@@ -466,3 +476,36 @@ func ensureNameMatchesService(lbls labels.Labels) labels.Labels {
...
+
+	for _, l := range lbls {
+		if _, exists := seen[l.Name]; exists {
+			return fmt.Errorf("duplicate label name: %s", l.Name)
+		}
+		seen[l.Name] = struct{}{}
+
+		// Skip validation for reserved labels
+		if l.Name == pyroscope.LabelName {
```
simonswine, line 495: Not too sure I understand the reason for skipping.

I think it might make sense to still check if the label value is correct in that case.

simonswine (approved):
Thanks for adding this.

LGTM

internal/component/pyroscope/write/write.go
marcsanmi: I've skipped it because it's a `reservedLabelNames ` in the pyroscope `ValidateLabelName` func. However, I can add switch the order and check label values always and then check for valid keys.

internal/component/pyroscope/write/write.go
marcsanmi: Done in 4879713eec22cce3535dc2ea7f59a778f98f5836

internal/component/pyroscope/write/write.go
marcsanmi: Updated in 4879713eec22cce3535dc2ea7f59a778f98f5836

internal/component/pyroscope/write/write.go
```diff
@@ -466,3 +477,39 @@ func ensureNameMatchesService(lbls labels.Labels) labels.Labels {
...
+		// Skip label name validation for pyroscope reserved labels
+		if l.Name == pyroscope.LabelName {
+			lastLabelName = l.Name
+			continue
+		}
+
+		// Validate label name
+		if err := labelset.ValidateLabelName(l.Name); err != nil {
+			return fmt.Errorf("invalid label name: %w", err)
+		}
```
simonswine, line 509: Nit: Find that easier to read and less error prone

```suggestion
		// Skip label name validation for pyroscope reserved labels
		if l.Name != pyroscope.LabelName {
  		  		// Validate label name
		  		if err := labelset.ValidateLabelName(l.Name); err != nil {
		  		  		return fmt.Errorf("invalid label name: %w", err)
		  		  }
		}

```

simonswine: If you disagree, just ignore

simonswine (approved):
LGTM

PR #2943, opened 2025-03-10
Set signal types according to encoding in otel kafka exporter

### LINKED ISSUE #2935: otelcol.exporter.kafka | unrecognized encoding
Hi,

I am using otelcol.exporter.kafka to send logs to kafka broker, below are the config used as per docs.

https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.kafka/#arguments

`Encodings which work only for logs:
"raw": If the log record body is a byte array, it is sent as is. Otherwise, it is serialized to JSON. Resource and record attributes are discarded.`

`otelcol.exporter.kafka "default" {
  protocol_version = "2.0.0"
  brokers = ["localhost:9092"]
  topic = "loki-events"
  encoding = "raw"
}`

I am seeing the below error with otelcol.exporter.kafka.default  Unhealthy status. Please advise
`failed to create components: unrecognized encoding; unrecognized encoding
`

wildum:
#### PR Description

The supported signal type should be set according to the encoding.

Fixes #2935

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/internal/component/otelcol/exporter/kafka/kafka.go
+++ b/internal/component/otelcol/exporter/kafka/kafka.go
@@ -27,7 +27,17 @@ func init() {
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
 			fact := kafkaexporter.NewFactory()
-			return exporter.New(opts, fact, args.(Arguments), exporter.TypeSignalConstFunc(exporter.TypeAll))
+			typeSignalFunc := func(opts component.Options, args component.Arguments) exporter.TypeSignal {
+				switch args.(Arguments).Encoding {
+				case "raw":
+					return exporter.TypeLogs
+				case "jaeger_proto", "jaeger_json", "zipkin_proto", "zipkin_json":
+					return exporter.TypeTraces
+				default:
+					return exporter.TypeAll
+				}
+			}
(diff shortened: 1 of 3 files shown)


internal/component/otelcol/exporter/kafka/kafka.go
```diff
@@ -27,11 +27,22 @@ func init() {
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
 			fact := kafkaexporter.NewFactory()
-			return exporter.New(opts, fact, args.(Arguments), exporter.TypeSignalConstFunc(exporter.TypeAll))
+			return exporter.New(opts, fact, args.(Arguments), GetSignalType)
```
ptodev, line 30: Would this still work if the config is reloaded to work with a different signal type?

thampiotr: Yes: https://github.com/grafana/alloy/blob/2a991573b129157960f7c2fc3fa6600620803f02/internal/component/otelcol/exporter/exporter.go#L214

thampiotr (approved):

PR #2945, opened 2025-03-10
remote.vault: enhance auth.custom to support namespace switching for authentication

Notedop:
#### PR Description

With Vault Enterprise it is possible to define different namespaces and multi-tenancy support as described [here](https://developer.hashicorp.com/vault/docs/enterprise/namespaces). With this setup it is also possible to have a dedicated authentication namespace, while the kv store exists on a child namespace. 

Currently Alloy uses the same client for both the AuthenticationManager and SecretManager; as a result currently it is not possible to define a separate namespace for each client. 

This PR addresses the above and will allow you to override the namespace used by the client of the authentication manager.

#### PR Checklist

- [ x ] CHANGELOG.md updated
- [ x ] Documentation added


Notedop:
Rebased on latest main branch to address conflict in `CHANGELOG.md`

Notedop:
@clayton-cornell could you review this PR for me and let me know if this is fine?

docs/sources/reference/components/remote/remote.vault.md
```diff
@@ -143,13 +143,21 @@ The `auth.custom` blocks allows authenticating against Vault using an arbitrary
...
+| Name         | Type            | Description                                            | Default | Required |
+|--------------|-----------------|--------------------------------------------------------| ------- |----------|
+| `path`       | `string`        | Path to write to for creating an authentication token. |         | yes      |
+| `data`       | `map(secret)`   | Authentication data.                                   |         | yes      |
+| `namespace`  | `string`        | The namespace to authenticate to.                      |         | no       |
 
 All values in the `data` attribute are considered secret, even if they contain nonsensitive information like usernames.
 
+With Vault Enterprise, you can authenticate against a parent namespace while storing secrets in a child namespace. 
+By specifying the namespace argument in `auth.custom`, you can authenticate to a different namespace than the one used to retrieve the secrets.
```
clayton-cornell, line 156:
```suggestion
By specifying the namespace argument in `auth.custom`, you can authenticate to a namespace different from the one used to retrieve the secrets.
```

docs/sources/reference/components/remote/remote.vault.md
```diff
@@ -143,13 +143,21 @@ The `auth.custom` blocks allows authenticating against Vault using an arbitrary
...
+
+You can also define Vault environment variables, which will be automatically loaded by the clients used by Alloy.
```
clayton-cornell, line 158:
```suggestion
You can also define Vault environment variables, which the clients used by Alloy will automatically load. 
```

docs/sources/reference/components/remote/remote.vault.md
```diff
@@ -143,13 +143,21 @@ The `auth.custom` blocks allows authenticating against Vault using an arbitrary
...
+| `namespace`  | `string`        | The namespace to authenticate to.                      |         | no       |
 
 All values in the `data` attribute are considered secret, even if they contain nonsensitive information like usernames.
 
+With Vault Enterprise, you can authenticate against a parent namespace while storing secrets in a child namespace. 
+By specifying the namespace argument in `auth.custom`, you can authenticate to a different namespace than the one used to retrieve the secrets.
+
+You can also define Vault environment variables, which will be automatically loaded by the clients used by Alloy. 
+This approach allows you to use certificate-based authentication i.e. by setting the `VAULT_CACERT` and `VAULT_CAPATH` environment variables.
+Please refer to the [Vault Environment variables](https://developer.hashicorp.com/vault/docs/commands#configure-environment-variables)  documentation for more information.
```
clayton-cornell, line 160:
```suggestion
Refer to the [Vault Environment variables](https://developer.hashicorp.com/vault/docs/commands#configure-environment-variables)  documentation for more information.
```

Notedop:
@clayton-cornell thank you for the review. I've committed your suggestions.

clayton-cornell:
Now over to @grafana/grafana-agent-maintainers for a code review

Notedop:
@clayton-cornell do you have any idea how long this step usually takes? I can't seem to be able to tag the Grafana maintainers myself.

clayton-cornell:
@Notedop It really depends on current workload for the dev team. I'll check with them :-)

Notedop:
@clayton-cornell thank you, I appreciate it 🤗

docs/sources/reference/components/remote/remote.vault.md
```diff
@@ -143,13 +143,21 @@ The `auth.custom` blocks allows authenticating against Vault using an arbitrary
...
+| Name         | Type            | Description                                            | Default | Required |
+|--------------|-----------------|--------------------------------------------------------| ------- |----------|
+| `path`       | `string`        | Path to write to for creating an authentication token. |         | yes      |
+| `data`       | `map(secret)`   | Authentication data.                                   |         | yes      |
+| `namespace`  | `string`        | The namespace to authenticate to.                      |         | no       |
 
 All values in the `data` attribute are considered secret, even if they contain nonsensitive information like usernames.
 
+With Vault Enterprise, you can authenticate against a parent namespace while storing secrets in a child namespace. 
+By specifying the namespace argument in `auth.custom`, you can authenticate to a namespace different from the one used to retrieve the secrets.
```
dehaansa, line 156: Is there a technical reason this is only supported on `auth.custom`? I don't have enough knowledge of the vault integration to say for sure but it seems like this could apply to other authMethod implementations.

Notedop: Honestly, I'm not 100% sure. We use the auth.custom method, which allows to define a custom authentication URL, in combination with TLS / Certificate based authentication setup via environment variables. Our setup has the authentication as a parent namespace while the KV Store is a child namespace, which is not supported by Alloy currently. This has triggered me to create this change.

Unfortunately I don't have the ability to verify (test) any other authentication methods with the namespace switching as I am limited to what is available to me.

I would suggest to continue with this change and if needed enhance any additional authentication methods if there is a need for it and there's someone available to do the verification.

docs/sources/reference/components/remote/remote.vault.md
```diff
@@ -143,13 +143,21 @@ The `auth.custom` blocks allows authenticating against Vault using an arbitrary
...
+
+You can also define Vault environment variables, which the clients used by Alloy will automatically load. 
+This approach allows you to use certificate-based authentication i.e. by setting the `VAULT_CACERT` and `VAULT_CAPATH` environment variables.
```
clayton-cornell, line 159:
```suggestion
You can also define Vault environment variables, which the clients used by {{< param "PRODUCT_NAME" >}} will automatically load. 
This approach allows you to use certificate-based authentication by setting the `VAULT_CACERT` and `VAULT_CAPATH` environment variables.
```

Use the product variable for Alloy - just ensures consistency in how Alloy is spelled/used.
We shouldn't use `i.e` https://developers.google.com/style/word-list#ie  and in this context, I think we can simply drop the `i.e.` and the sentence still works and is clear.

docs/sources/reference/components/remote/remote.vault.md
dehaansa: This makes sense, but for future maintenance can you put a comment to this effect in the code? Something that communicates "This is likely applicable to other forms of authentication on enterprise vault but only testable with authCustom when added in #2945"?

Notedop: I've added a comment in code as suggested.

dehaansa (approved):
Thanks for the contribution! Once CI is happy we'll get this merged and ready for v1.8.

clayton-cornell (approved):

PR #2954, opened 2025-03-10
Update running_components metric and adds component_health metric

### LINKED ISSUE #2953: Add component health state monitoring
### Request

Add a new alloy metric called, `alloy_component_controller_health` that passes in the health_state's current value for monitoring component health. The current implementation that exists within `alloy_component_controller_running_components` does not provide the affected component, only a health_state and the current count of components running

### Use case

This feature is useful to identify and monitor components by their reported health_state

6fears7:
#### PR Description
Updates `alloy_component_controller_running_components`  to strip the `health_state` label and create a new metric  `alloy_component_controller_health` to monitor the health state by a given component

## Example New Metric
```
# HELP alloy_component_controller_health Component health
# TYPE alloy_component_controller_health gauge
alloy_component_controller_health{component_name="prometheus.scrape",controller_id="",controller_path="/",health_type="healthy"} 1
alloy_component_controller_health{component_name="prometheus.scrape",controller_id="",controller_path="/",health_type="unhealthy"} 0
```

## Updated Old Metric
```
# HELP alloy_component_controller_running_components Total number of running components.
# TYPE alloy_component_controller_running_components gauge
alloy_component_controller_running_components{controller_id="",controller_path="/"} 3
alloy_component_controller_running_components{controller_id="remotecfg",controller_path="/"} 0
```

Fixes #2953

#### Notes to the Reviewer
Some Prometheus exporter maintainers prefer avoiding creating more metrics based off multiple states (such as "healthy" and "unhealthy" and instead using the value as the condition. I am indifferent and am open to re-writing if you prefer to go that route. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


6fears7:
@clayton-cornell Some assistance may be needed. Pipeline is continuing to fail the OTLP integration test. I even compiled from main  and ran the same integration test command and am failing. I am not sure how to proceed.

clayton-cornell:
@6fears7 First thing that needs to be done is the minor conflict needs to be resolved. It looks like just a collision in the changelog.  Should be easy to resolve. If you can fix that then more of the pipeline shoudl start to build/pass. If it's still failing then we can take it to the next step :-)

docs/sources/troubleshoot/controller_metrics.md
```diff
@@ -18,10 +18,10 @@ Metrics for the controller are exposed at the `/metrics` HTTP endpoint of the {{
 
 The controller exposes the following metrics:
 
-* `alloy_component_controller_evaluating` (Gauge): Set to `1` whenever the  component controller is evaluating components.
+- `alloy_component_controller_evaluating` (Gauge): Set to `1` whenever the component controller is evaluating components.
```
clayton-cornell, line 21:
```suggestion
* `alloy_component_controller_evaluating` (Gauge): Set to `1` whenever the component controller is evaluating components.
```

clayton-cornell:
Over to @grafana/grafana-alloy-maintainers for code review

internal/runtime/internal/controller/metrics.go
```diff
@@ -98,42 +98,82 @@ func (cm *controllerMetrics) Describe(ch chan<- *prometheus.Desc) {
...
+		collectorHealth: prometheus.NewDesc(
+			"alloy_component_controller_health",
+			"Component health",
+			[]string{"health_type", "component_name"},
+			map[string]string{"controller_path": parent, "controller_id": id},
+		),
 		runningComponentsTotal: prometheus.NewDesc(
 			"alloy_component_controller_running_components",
-			"Total number of running components.",
-			[]string{"health_type"},
```
ptodev, line 109: I'm worried that some folks might rely on this label and this would be a breaking change for them.

internal/runtime/internal/controller/metrics.go
```diff
@@ -98,42 +98,82 @@ func (cm *controllerMetrics) Describe(ch chan<- *prometheus.Desc) {
...
 }
 
 func newControllerCollector(l *Loader, parent, id string) *controllerCollector {
 	return &controllerCollector{
 		l: l,
+		collectorHealth: prometheus.NewDesc(
+			"alloy_component_controller_health",
+			"Component health",
+			[]string{"health_type", "component_name"},
+			map[string]string{"controller_path": parent, "controller_id": id},
```
ptodev, line 111: A few things could be a problem:
* If Alloy is running a lot of components, this would create a lot of series. For many users this would not be desirable.
* Is it really useful to have this metric? The health status feature is currently underutilised and probably not that useful... There are many times when components are reported as healthy even though they are experiencing issues.

It would be better if users rely on metrics that are specific to the each component's behaviour when deciding whether it runs well or not. Component health is more like a vague general indicator which isn't right sometimes. It's nice for the UI, but I wouldn't use it too much on dashboards and alerts.

ptodev: > My current rough count of the current number of components if a user ran every single component is ~132

This is assuming there's only one instance of every component. There is often more than one instance though. For example, it's common to have several relabel/transform components. It still feels like it's too many metrics, for a feature which at the moment is of questionable usefulness. Maybe if we somehow gate this behind a "debug metrics" flag it'd be more acceptable. We need to have such a flag sooner or later IMO. The upstream Collector already has it and they're trying to categorise their metrics more and more to make the most of it.

Btw if you just want to solve an immediate need, I wonder if you could just use the Alloy UI endpoints instead? I can't tell what the URL is off the top of my head, but there's an endpoint you can query and you'll get a json reply with a list of running components and their health. It might be good enough for a CI/CD pipeline.

6fears7: I'll defer to your judgment on this. Feel free to close, postpone, etc. at your discretion.

PR #2959, opened 2025-03-11
Enforces string type

scarolan:
When you pass --set "alloy.extraEnv[0].value=8443", Helm interprets 8443 as a number, which is incompatible with a template expecting a string.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/containers/_agent.yaml
+++ b/operations/helm/charts/alloy/templates/containers/_agent.yaml
@@ -32,8 +32,9 @@
       valueFrom:
         fieldRef:
           fieldPath: spec.nodeName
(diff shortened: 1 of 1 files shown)


scarolan:
Apparently this "fix" breaks other things.  Closing while we work out the solution for passing numbers to Helm

thampiotr:
It is related to an issue reported in Helm [here](https://github.com/helm/helm/issues/1707) and [here](https://github.com/helm/helm/issues/1694). There are a few workarounds mentioned in those, including using [--set-string feature](https://github.com/helm/helm/issues/1707).

PR #2960, opened 2025-03-11
Remove unnecessary Drone steps for building and testing

ptodev:
These builds migrated from Drone to GitHub Actions and need to be removed when we verify that GH Actions work ok. When I've done that I'll take the PR out of draft status.


jharvey10:
Feel free to reopen if necessary, but I think this PR has been superseded by others we've done since then.

PR #2966, opened 2025-03-11
Fix attribute name in Beyla configuration.

rafaelroquetto:
#### PR Description

The Beyla configuration for network and application filters rely on the block label for the attribute name, i.e.

```
            filters {
              network "k8s.src.owner.name" {
                match = "*"
              }
            }
```

however, attribute names such as "k8s.src.owner.name" are not valid label identifiers, rendering the above config unparsable by the alloy config syntax.

This PR changes the code so that attributes need to be explicitly specified using the `attr` field:

```
	filters {
		network {
			attr = "k8s.src.owner.name"
			match = "*"
		}

		network {
			attr = "k8s.src.owner.name2"
			match = "*"
		}
	}
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ Main (unreleased)
   - `query_sample`: better handling of truncated queries (@cristiangreco)
   - `query_sample`: add option to use TiDB sql parser (@cristiangreco)
 
+### Bugfixes
+
+- Fixed the parsing of application and network filter blocks for Beyla
+
 ### Other changes
 
 - Upgrading to Prometheus v2.55.1. (@ptodev)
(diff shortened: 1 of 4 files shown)


clayton-cornell:
No doc changes or suggestions

wildum (approved):
LGTM, tests should be updated. It would be nice to have more examples in the doc also but can be done via another PR

rafaelroquetto:
@wildum thanks, I will track the request for improved examples at https://github.com/grafana/beyla/issues/1736 so that it does not fall in the limbo.

PR #2967, opened 2025-03-12
Add docker compose configuration notes

tiborrr:
Update the Alloy Cadvisor exporter docs with relevant information on how to set Cadvisor up in docker compose.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
@@ -15,8 +15,36 @@ The `prometheus.exporter.cadvisor` component collects container metrics using [c
 ## Usage
 
 ```alloy
-prometheus.exporter.cadvisor "<LABEL>" {
+prometheus.exporter.cadvisor "container_metrics" {
+    docker_host      = "unix:///var/run/docker.sock"
+    storage_duration = "5m"
 }
+```
+
+When exporting to Mimir then make sure to set max_label_names_per_series higher than the default, otherwise Mimir will not except the metrics.
+
+```yaml
+limits:
+  max_label_names_per_series: 100
+```
+
+## Docker Configuration
+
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
```diff
@@ -19,6 +19,32 @@ prometheus.exporter.cadvisor "<LABEL>" {
 }
 ```
 
+When exporting to Mimir then make sure to set max_label_names_per_series higher than the default, otherwise Mimir will not except the metrics.
```
clayton-cornell, line 22:
```suggestion
When you export to Mimir, make sure to set `max_label_names_per_series` higher than the default, otherwise Mimir will not accept the metrics.
```

docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
```diff
@@ -19,6 +19,32 @@ prometheus.exporter.cadvisor "<LABEL>" {
...
+
+```yaml
+limits:
+  max_label_names_per_series: 100
+```
+
+## Docker Configuration
+
+To properly collect container metrics when using Alloy in docker compose, the Alloy container must be run with specific volume mounts and privileges. Here's the required Docker Compose configuration:
```
clayton-cornell, line 31:
```suggestion
To properly collect container metrics when using {{< param "PRODUCT_NAME" >}} in a Docker container, the {{< param "PRODUCT_NAME" >}} container must be run with specific volume mounts and privileges.
The following example shows a typical Docker configuration that provides the correct volume mounts.
```

clayton-cornell (commented):
The new text is not in the right place in the doc structure. I'd suggest all the new content is moved to around line 184 (just before the generated components section).

I'd like to keep the overall structure of the component docs fairly consistent. Deployment-specific info (not everyone runs Alloy in a Docker container) should come after the generic information about args and blocks.

PR #2968, opened 2025-03-12
[Feature branch] Live graph front-end

### LINKED ISSUE #2609: Live graph frontend

wildum:
#### PR Description

This is the front-end part of the live graph. This is a replacement for the existing graph in the UI.

Fleet management is not yet supported, it will be done via another PR.
Modules are supported: There is a button on the module page to open the module graph page.
If the config is reloaded, then the graph page must be refreshed (this could be handled if there is any need for it one day. For now I don't think that this is worth the effort).

Fixes #2609

#### Notes to the Reviewer

- I fixed the count that was not updated in the API.
- Fixed the count for the relabel component.
- I realised that the live debugging does not need to be enabled for the live graph so I removed it.

### DIFF
--- a/internal/component/prometheus/relabel/relabel.go
+++ b/internal/component/prometheus/relabel/relabel.go
@@ -275,11 +275,15 @@ func (c *Component) relabel(val float64, lbls labels.Labels) labels.Labels {
 	// TODO(@mattdurham): Instead of setting this each time could collect on demand for better performance.
 	c.cacheSize.Set(float64(c.cache.Len()))
 
+	count := uint64(1)
+	if relabelled.Len() == 0 {
+		count = 0 // if no labels are left, the count is not incremented because the metric will be filtered out
+	}
 	componentID := livedebugging.ComponentID(c.opts.ID)
 	c.debugDataPublisher.PublishIfActive(livedebugging.NewData(
 		componentID,
 		livedebugging.PrometheusMetric,
-		1,
+		count,
 		func() string {
 			return fmt.Sprintf("%s => %s", lbls.String(), relabelled.String())
 		},
--- a/internal/component/prometheus/relabel/relabel.go
+++ b/internal/component/prometheus/relabel/relabel.go
@@ -276,14 +276,10 @@ func (c *Component) relabel(val float64, lbls labels.Labels) labels.Labels {
 	c.cacheSize.Set(float64(c.cache.Len()))
 
 	componentID := livedebugging.ComponentID(c.opts.ID)
-	count := uint64(1)
-	if relabelled.Len() == 0 {
-		count = 0 // if no labels are left, the count is not incremented because the metric will be filtered out
-	}
 	c.debugDataPublisher.PublishIfActive(livedebugging.NewData(
 		componentID,
 		livedebugging.PrometheusMetric,
-		count,
+		1,
 		func() string {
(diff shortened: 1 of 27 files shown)


internal/web/ui/src/features/graph/debugDataType.ts
```diff
@@ -0,0 +1,26 @@
+export interface DebugData {
+  componentID: string;
+  targetComponentIDs: Array<string>;
+  rate: number;
+  type: DebugDataType;
+}
+
+export enum DebugDataType {
```
ptodev, line 8: Do we display anything for non-telemetry components such as `local.file`?

ptodev (approved):
Rubber stamping, I don't see anything wrong but I also don't know anything about the UI. I like the final result though! 😆

internal/component/prometheus/relabel/relabel.go
```diff
@@ -276,14 +276,10 @@ func (c *Component) relabel(val float64, lbls labels.Labels) labels.Labels {
...
 	componentID := livedebugging.ComponentID(c.opts.ID)
-	count := uint64(1)
-	if relabelled.Len() == 0 {
-		count = 0 // if no labels are left, the count is not incremented because the metric will be filtered out
-	}
 	c.debugDataPublisher.PublishIfActive(livedebugging.NewData(
 		componentID,
 		livedebugging.PrometheusMetric,
-		count,
+		1,
```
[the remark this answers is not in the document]
wildum: thanks for pointing it out, I got confused between branches and something that I wrongly fixed last time. There should be not changes actually, I will revert this part

internal/web/ui/src/features/graph/debugDataType.ts
wildum: they are connected but there is no data passing on their edges (they stay gray). Actually, I need to reverse the edge of this one as well.
I think that we should then try to add live debugging support to all components. For the type of data, we might need to introduce a type "other". 

I hope that not seeing data on components that don't support live debugging won't be too weird for the users. I added in the description that the data is shown for components that support live debugging but I might need to make it more obvious.

wildum: After reflecting and testing with other components, I don't think that changing the direction of the edges depending on the component names is a good strategy. I will come up with a better approach and ping you when I'm ready. I'm putting this PR back to draft for now

internal/runtime/internal/controller/loader_test.go
```diff
@@ -124,6 +124,46 @@ func TestLoader(t *testing.T) {
 		requireGraph(t, l.Graph(), testGraphDefinition)
 	})
 
+	// TODO: add one more test with a receiver component to make sure the edge is added properly
```
wildum, line 127: I'm planning to add this test on the feature branch after I rebase it with main because I expect some conflicts since this branch is a bit old

wildum:
@ptodev I did the following rework since your review:
- created the dataflow graph in the backend. This is a much stronger approach than reverting the edges from the dependency graph in the frontend. It simplifies the logic in the frontend and it will help to build other features on top of the data flow graph (such as the diagnostic feature). 
- added an info text at the top of the page to tell clearly that only the components that support live debugging will have their edges colored
- removed the graph button on the remotecfg component (support for fleet management will come later)
- fixed a visual bug on the edges that were horizontal 

Please take another look at it when you have time!

PR #2969, opened 2025-03-12
Reduce contention in prom scrape

thampiotr:
The mutexes in prometheus codebase come up too often on the profiles, even when there are no moved targets as evidenced by prometheus_scrape_targets_moved_total. This is probably the reason why.

### DIFF
--- a/internal/component/prometheus/scrape/scrape.go
+++ b/internal/component/prometheus/scrape/scrape.go
@@ -22,6 +22,10 @@ import (
 	"github.com/prometheus/prometheus/scrape"
 	"github.com/prometheus/prometheus/storage"
 
+	"github.com/prometheus/prometheus/model/exemplar"
+	"github.com/prometheus/prometheus/model/metadata"
+	"github.com/prometheus/prometheus/util/logging"
+
 	"github.com/grafana/alloy/internal/component"
 	component_config "github.com/grafana/alloy/internal/component/common/config"
 	"github.com/grafana/alloy/internal/component/discovery"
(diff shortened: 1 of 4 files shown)


internal/component/prometheus/scrape/scrape.go
```diff
@@ -335,7 +336,9 @@ func (c *Component) Run(ctx context.Context) error {
 			// Prometheus handles marking series as stale: it is the client's responsibility to inject the
 			// staleness markers. In our case, for targets that moved to another instance in the cluster, we hand
 			// over this responsibility to the new owning instance. We must not inject staleness marker here.
-			c.scraper.DisableEndOfRunStalenessMarkers(jobName, movedTargets)
+			if len(movedTargets) > 0 {
```
ptodev, line 339: Why not just have `DisableEndOfRunStalenessMarkers` immediately return when the length is 0, prior to attempting to lock a mutex?

thampiotr: Yeah, I can do that some time later because it'd require changing the fork and I don't have capacity to do that right now.

ptodev: I can do it for you and update your branch?

thampiotr: sounds good!

ptodev: Done!

thampiotr: Thanks so much for the help!

PR #2970, opened 2025-03-12
clustering traffic admission control

### LINKED ISSUE #201: Proposal: Only attempt to send metrics after joining a sufficiently large cluster
### Background

In clusters with several agent pods and many potential targets, it can take a few minutes for a cluster to have all of its pods join and become healthy on startup. I noticed that when a pod starts up that is intended to be part of a larger cluster, it takes a few minutes to join. During this time it will briefly attempt to scrape a large number of targets and run into OOM errors.

image

This is more apparent in a statefulset with OrderedReady enabled, where the first pod will have all potential targets. Then as additional pods get spun up, the number of scrape targets will be distributed evenly among them. We can know ahead time that a cluster will have at least say 5 pods.

What I have found as a work around, is to launch the pods with a config that has several `discovery` components to find all the potential targets, but no scrape component. Then once all 5 pods are running and part of the cluster, then I update the config to add the scrape component.

Also when the statefulset is changed from OrderedReady to Parallel, then all the pods launch at the same time. But what I found is there is an initial burst in 409 conflict errors from `thanos-receive` because the pods have not joined the cluster yet. Once they have joined, this goes away.

### Proposal

I propose a new flag to grafana-agent run which specifies the minimum number of agents in the cluster before sending any metrics:

```
--cluster.min-cluster-size # minimum number of agents in the cluster before metrics will be sent
```

This will lead to a delay in sending metrics when first starting up because it takes a few minutes for a large cluster to become healthy and have its agents register themselves. However, this opt-in feature can be helpful for stability purposes. This can prevent the first grafana agent pod in a statefulset from running into an OOM error and dramatically cut down the number of 409 conflict errors from `thanos-receive`.

thampiotr:
#### PR Description

Added `--cluster.wait-for-size` and `--cluster.wait-timeout` flags which allow to specify the minimum cluster size
  required before components that use clustering begin processing traffic to ensure adequate cluster capacity is
  available.

Extended the existing tests, including [the e2e tests added previously](https://github.com/grafana/alloy/pull/2996).

Fixes #201

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/service/cluster/cluster_readonly.go
thampiotr, line 1: I want to rename this file to cluster.go and the cluster.go -> service.go as it makes more sense, but it messes up the diff a lot, so I will leave this for later.

internal/service/cluster/cluster.go
```diff
@@ -299,6 +327,9 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		s.logPeers("peers changed", toStringSlice(peers))
 		span.SetAttributes(attribute.Int("peers_count", len(peers)))
 
+		// Peers changed - update the ready-to-admit-traffic state
+		s.alloyCluster.updateReadyToAdmitTraffic()
```
thampiotr, line 331: TODO: remove this, it's no longer needed

internal/service/cluster/cluster.go
```diff
@@ -352,6 +385,9 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		}
 	}
 
+	// Update the ready-to-admit state after node start
+	s.alloyCluster.updateReadyToAdmitTraffic()
```
thampiotr, line 389: TODO: will remove this too

internal/service/cluster/cluster.go
```diff
@@ -410,6 +446,7 @@ func (s *Service) getPeers() ([]string, error) {
 	}
 
 	// We shuffle the list and return only a subset of the peers.
+	// TODO(thampiotr): verify that this can be called from multiple threads
```
thampiotr, line 449: TODO: Will investigate this before merging.

thampiotr (commented):
It's ready for feedback, I have just a few small things to address in the meantime.

dehaansa:
In cases with very large k8s clusters for example, where discovery is immense and costly, I think we would want to wait to even do discovery until after cluster has converged.

Do you agree? Should there be an additional option to wait for cluster converge before doing _any_ work?

thampiotr:
> In cases with very large k8s clusters for example, where discovery is immense and costly, I think we would want to wait to even do discovery until after cluster has converged.
> 
> Do you agree? Should there be an additional option to wait for cluster converge before doing _any_ work?

The discovery is the same for each instance, regardless of number of instances in the cluster, so stopping them from doing work will not improve things. Every instance needs to be able to handle the entire cluster discovery in current architecture. This PR follows the design [that I shared here](https://github.com/grafana/alloy/issues/201#issuecomment-2717587504) where we explicitly said that 

> we scope this behaviour only to components that support clustering. Other components will run as usual.

I think scaling discovery will have to be a separate problem to address and once we have it sharded in some way we can definitely include the min cluster requirements to it in the future.

internal/service/cluster/cluster.go
thampiotr: `getPeers` is called from one thread only so not an issue

thampiotr:
I will still need to add the following, as per [design](https://github.com/grafana/alloy/issues/201#issuecomment-2717587504):

> Debug info should ideally explain the situation to the user in the UI.
> The clustering overview dashboard should show clearly when the minimum size is not met. May need to add metrics for it.

I would like to do this in a follow-up PR.

docs/sources/reference/cli/run.md
```diff
@@ -138,16 +140,25 @@ The `--cluster.rejoin-interval` flag defines how often each node should rediscov
...
 
-To join or rejoin a cluster, {{< param "PRODUCT_NAME" >}} tries to connect to a certain number of peers limited by the `--cluster.max-join-peers` flag.
+To join or rejoin a cluster, {{< param "PRODUCT_NAME" >}} tries to connect to a number of random peers limited by the `--cluster.max-join-peers` flag.
 This flag can be useful for clusters of significant sizes because connecting to a high number of peers can be an expensive operation.
 To disable this behavior, set the `--cluster.max-join-peers` flag to 0.
 If the value of `--cluster.max-join-peers` is higher than the number of peers discovered, {{< param "PRODUCT_NAME" >}} connects to all of them.
 
+The `--cluster.wait-for-size` flag specifies the minimum cluster size required before components that use clustering
+begin processing traffic. When set to a value greater than zero, a node will join the cluster but the components that
+use clustering will not take on any work until enough nodes are available. This ensures adequate cluster capacity - see
```
clayton-cornell, line 154:
```suggestion
use clustering will not take on any work until enough nodes are available. This ensures adequate cluster capacity - refer to
```

docs/sources/reference/cli/run.md
```diff
@@ -55,6 +55,8 @@ The following flags are supported:
 * `--cluster.tls-cert-path`: Path to the certificate file used for peer communication over TLS.
 * `--cluster.tls-key-path`: Path to the key file used for peer communication over TLS.
 * `--cluster.tls-server-name`: Server name used for peer communication over TLS.
+* `--cluster.wait-for-size`: Wait for the cluster to reach the specified number of instances before admitting traffic to components that use clustering. Zero means disabled (default `0`).
```
dehaansa, line 58: Is "admitting traffic" going to be clear language to users? Should it be more literal?
> Wait for the cluster to reach the specified number of instances before allowing components that use clustering to begin collecting telemetry...

internal/service/cluster/cluster.go
```diff
@@ -167,15 +191,19 @@ func New(opts Options) (*Service, error) {
...
 		tracer: t,
 		opts:   opts,
 
 		sharder: ckitConfig.Sharder,
 		node:    node,
 		randGen: rand.New(rand.NewSource(time.Now().UnixNano())),
-	}, nil
+	}
+
+	s.alloyCluster = newAlloyCluster(l, ckitConfig.Sharder, opts)
```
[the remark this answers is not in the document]
thampiotr: It was not needed, but after my refactor it's now needed...

internal/service/cluster/cluster_readonly.go
```diff
@@ -0,0 +1,159 @@
...
+	if c.opts.MinimumClusterSize == 0 {
+		c.isReadyToAdmitTraffic.Store(true)
+		return
+	}
+
+	// Number of peers is greater than the minimum cluster size = ready to admit traffic.
+	if len(c.sharder.Peers()) >= c.opts.MinimumClusterSize {
+		// Reset the deadline if it is configured:
+		if c.opts.MinimumSizeWaitTimeout != 0 {
+			c.minimumSizeDeadline.Store(time.Now().Add(c.opts.MinimumSizeWaitTimeout))
```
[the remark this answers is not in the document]
thampiotr: I read that and it confused me too. The behaviour was not correct and hard to read, so I added more tests, fixed the issues and refactored to make it more readable.

internal/service/cluster/cluster.go
```diff
@@ -269,62 +299,16 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	limiter := rate.NewLimiter(rate.Every(stateUpdateMinInterval), 1)
-	s.node.Observe(ckit.FuncObserver(func(peers []peer.Peer) (reregister bool) {
-		tracer := s.tracer.Tracer("")
```
thampiotr, line 274: This logic was moved below to a goroutine that handles dispatching cluster change notifications. This is because when cluster minimum size timer expires we also need to dispatch cluster change notifications, even though the peers in ckit didn't change.

internal/service/cluster/cluster_e2e_test.go
```diff
@@ -36,48 +37,62 @@ import (
...
+					verifyClusterReady(t, p)
 				}
 				verifyLookupInvariants(t, state.peers)
 			},
 		},
 		{
 			name:             "three nodes with slow components",
 			nodeCountInitial: 3,
+			// slow components can currently lead to timeouts and communication errors
+			// TODO: consider decoupling cluster operations from runtime/components performance
```
thampiotr, line 77: I actually have done this by putting the dispatching of cluster updates on a separate goroutine. TODO: remove this comment.

internal/service/cluster/cluster.go
```diff
@@ -387,7 +387,59 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
+	}
+	spanWait.End()
+
+	peers := s.node.Peers()
+	s.logPeers("peers changed", toStringSlice(peers))
+	span.SetAttributes(attribute.Int("peers_count", len(peers)))
+	span.SetAttributes(attribute.Int("minimum_cluster_size", s.opts.MinimumClusterSize))
+
+	// Notify all components about the clustering change.
+	components := component.GetAllComponents(host, component.InfoOptions{})
```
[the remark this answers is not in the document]
thampiotr: Yes, I see where you're coming from. In step 5 they can hit the limiter, but they can also go through state update logic to conclude that the state didn't change (unless timer expires in the meantime). I see the circular call hierarchy here and I agree it's worth breaking it down. 

Here's a drawing of what's going on now:

image

And here's a drawing of what I am going to do to break this down:
image

Hope this makes sense, let me know if you have still any concerns. I'll go ahead to implement this now though.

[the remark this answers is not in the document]
thampiotr: I've implemented it in this commit: https://github.com/grafana/alloy/pull/2970/commits/1363c70bf20bcd77c0ed43d15c2cf1c58f1909c6

The e2e tests didn't require any changes, so that's reassuring. The code is a bit simpler too, we don't need the second rate limiter.

internal/service/cluster/cluster.go
```diff
@@ -387,7 +387,59 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
+
+		clusterComponent, ok := comp.Component.(Component)
+		if !ok {
+			continue
+		}
+
+		_, subSpan := tracer.Start(spanCtx, "NotifyClusterChange", trace.WithSpanKind(trace.SpanKindInternal))
+		subSpan.SetAttributes(attribute.String("component_id", comp.ID.String()))
+
+		clusterComponent.NotifyClusterChange()
```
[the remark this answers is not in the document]
thampiotr: I don't think they need to get notified close to simultaneously... it is now firing off all the notifications in a dedicated goroutine, previously it was using ckit's Observe routine to do that which could lead to other issues. I'm thinking we should keep it as-is unless we find a good reason to change it.

internal/service/cluster/cluster_readonly.go
```diff
@@ -0,0 +1,236 @@
...
+	if c.deadlineTimer != nil {
+		c.deadlineTimer.Stop()
+		c.deadlineTimer = nil
+	}
+	level.Info(c.log).Log(
+		"msg", "minimum cluster size reached, marking cluster as ready to admit traffic",
+		"minimum_cluster_size", c.opts.MinimumClusterSize,
+		"peers_count", len(c.sharder.Peers()),
+	)
+	c.clusterChangeCallback()
```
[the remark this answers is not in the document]
thampiotr: That callback was putting a struct in a channel, so it wasn't a problem. But this code changed now.

internal/service/cluster/cluster.go
```diff
@@ -387,7 +387,62 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
 
-func (s *Service) getPeers() ([]string, error) {
+func (s *Service) notifyComponentsOfClusterChanges(ctx context.Context, limiter *rate.Limiter, host service.Host) {
+	tracer := s.tracer.Tracer("")
+	spanCtx, span := tracer.Start(ctx, "NotifyClusterChange", trace.WithSpanKind(trace.SpanKindInternal))
+
+	// Limit how often we notify components about peer changes. At start up we may receive N updates in a period
+	// of less than one second. This leads to a lot of unnecessary processing.
+	_, spanWait := tracer.Start(spanCtx, "RateLimitWait", trace.WithSpanKind(trace.SpanKindInternal))
+	if err := limiter.Wait(ctx); err != nil {
```
dehaansa, line 397: Is there a risk here that we block notification of not ready -> ready due to the limiter?

internal/service/cluster/cluster_readonly.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Restart the deadline timer if it is configured and we just transitioned to not ready
+	if c.opts.MinimumSizeWaitTimeout != 0 {
+		if c.deadlineTimer != nil {
+			c.deadlineTimer.Stop()
+		}
+		c.deadlineTimer = time.AfterFunc(c.opts.MinimumSizeWaitTimeout, func() {
+			c.rwMutex.Lock()
+			defer c.rwMutex.Unlock()
+			c.transitionToStateDeadlinePassed()
```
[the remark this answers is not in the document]
thampiotr: You're right. Even though we Stop the timer in transition to Ready, the function may already be running. It won't cause errors, but it will give confusing logs. So I added a check whether we're already in `stateReady`. From abundance of caution, when this race is detected I added an Info level log message, just in case we need to debug something like this in the future. It should be extremely rare that it gets logged though. 

```
func (c *alloyCluster) transitionToStateDeadlinePassed() {
	if c.clusterState == stateReady {
		level.Info(c.log).Log(
			"msg", "minimum cluster size deadline passed, but cluster is already ready to admit traffic - ignoring",
			"minimum_cluster_size", c.opts.MinimumClusterSize,
			"minimum_size_wait_timeout", c.opts.MinimumSizeWaitTimeout,
			"peers_count", len(c.sharder.Peers()),
		)
		return
	}
(...)
```

internal/service/cluster/cluster.go
thampiotr: So in this case of limiter, we will block instead of aborting the operation when there's no budget left in the limiter. Blocking for up to 1s means that the `Ready()` when called directly from components will be stale (so, for example, cluster may not be Ready(), even though a call to Peers returns enough peers). After limiter unblocks we update the `Ready()` and notify all the components right away, at which point they should have consistent view. 

I don't see a problem with this delay really. The godocs say "Always use Ready to verify before assigning work to instance." and our components currently follow this correctly.

thampiotr: Okay, I don't think it's a problem, but it also doesn't really cost a thing if we move that state update before the limiter. I've done this now with a comment.

```
	// Update Ready() state of cluster service that components use. Doing it before the limiter to reduce the time
	// during which the components' view of cluster is not fully consistent (e.g. cluster not Ready() even though 
	// the number of peers is sufficient). Calls to `updateReadyState()` will still be effectively rate-limited
	// because only one goroutine performs the notification.
	s.alloyCluster.updateReadyState()

```

internal/service/cluster/cluster_readonly.go
```diff
@@ -0,0 +1,239 @@
...
+	defer c.rwMutex.Unlock()
+
+	// Number of peers is greater or equal the minimum cluster size = update to ready to admit traffic.
+	if len(c.sharder.Peers()) >= c.opts.MinimumClusterSize {
+		c.transitionToStateReady()
+		return
+	}
+
+	// The number of peers is less than the minimum and the deadline timer has expired, the cluster is ready to admit
+	// traffic and should remain in this state.
```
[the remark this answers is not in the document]
thampiotr: Yes. If this condition was not present:
```
if c.clusterState == stateDeadlinePassed {
		return
}
```
we would NOT "remain in this state" and would transition to "stateNotReady". So once the deadline expires and, for example, we add one extra instance but it's still not enough to be ready, the state would go back to "not ready" and we would no longer admit traffic. But we should admit traffic because the deadline expired. So this would be incorrect.

internal/service/cluster/cluster.go
thampiotr: After 1 second, which is this limiter's configured wait time. Note that if there's no budget left in the limiter, the `Wait` will block until budget is available and then it will continue. It's different then calling `Allow()` and aborting, for example. 

See limiter docs:
```
// Wait is shorthand for WaitN(ctx, 1).
func (lim *Limiter) Wait(ctx context.Context) (err error) {
	return lim.WaitN(ctx, 1)
}

// WaitN blocks until lim permits n events to happen.
// It returns an error if n exceeds the Limiter's burst size, the Context is
// canceled, or the expected wait time exceeds the Context's Deadline.
// The burst limit is ignored if the rate limit is Inf.
func (lim *Limiter) WaitN(ctx context.Context, n int) (err error) { ...
```

dehaansa: Ah, I misunderstood the Limiter, thought it was rejecting not forcing a wait. 👍

dehaansa (approved):

internal/service/cluster/cluster_readonly.go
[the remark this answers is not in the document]
thampiotr: Hope this answers the question, but if not I'm happy to chat. I'm going to merge this now, let me know if you have doubts about this!

PR #2973, opened 2025-03-12
feat: New`loki.stdin` component

dehaansa:
#### PR Description
We had an internal request to evaluate if we could support the behavior from promtail of capturing logs from stdin and sending them to a destination.

While Alloy is internally very different from Promtail, we want to see if we can support this behavior. This PR introduces a highly experimental component to mimic that behavior. While it works in testing, it has not been tested with large, complex pipelines, or slow final telemetry endpoints.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


dehaansa:
Closing this PR as the experimental component isn't in active development.

PR #2976, opened 2025-03-12
Fix the linting action to work appropriately in github actions.

dehaansa:
#### PR Description
Existing linting in GHA and drone does not actually report linting failure as an error due to the way they use find -execdir to attempt to lint all modules. As we currently only have base alloy module & syntax module, I've unfolded that into 2 steps in GHA, and updated the version of golangci-lint as the version we were using was not compatible with go 1.23+

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.github/workflows/test_pr.yml
+++ b/.github/workflows/test_pr.yml
@@ -27,6 +27,29 @@ jobs:
     - run: apt-get update -y && apt-get install -y libsystemd-dev
     - run: make lint
 
+  experimental-lint:
+    name: Experimental Lint
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Set up Go 1.23
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+        cache: false
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v6
+      with:
+        version: v1.64.7
+        args: --timeout=10m
+    - name: golangci-lint-syntax
+      uses: golangci/golangci-lint-action@v6
+      with:
+        version: v1.64.7
+        working-directory: ./syntax/
+    - run: make run-alloylint
+
   test_linux:
     name: Test Linux
     runs-on: ubuntu-latest


ptodev:
I like the integration with GitHub! It's nice ot have those inline comments in the diff section.

internal/component/discovery/relabel/relabel.go
```diff
@@ -79,6 +79,9 @@ func (c *Component) Run(ctx context.Context) error {
 
 // Update implements component.Component.
 func (c *Component) Update(args component.Arguments) error {
+	c.mut.Lock()
```
dehaansa, line 82: The linter noticed that this mutex wasn't used.

internal/service/remotecfg/remotecfg.go
```diff
@@ -166,6 +167,17 @@ func New(opts Options) (*Service, error) {
 		opts:        opts,
 		systemAttrs: getSystemAttributes(),
 		ticker:      jitter.NewTicker(math.MaxInt64-baseJitter, baseJitter), // first argument is set as-is to avoid overflowing
+		clientFactory: func(args Arguments) (collectorv1connect.CollectorServiceClient, error) {
```
dehaansa, line 170: This refactor was required to make tests pass, otherwise the initial testEnv update call got a 403.

internal/service/remotecfg/remotecfg.go
```diff
@@ -260,9 +272,13 @@ var _ service.Service = (*Service)(nil)
 // run until the provided context is canceled or there is a fatal error.
 func (s *Service) Run(ctx context.Context, host service.Host) error {
 	s.ctrl = host.NewController(ServiceName)
+	defer s.ticker.Stop()
```
dehaansa, line 275: Changed as there were hanging goroutines if registerCollector failed.

dehaansa:
Full green pass on 452b3bd (https://github.com/grafana/alloy/actions/runs/13841723976/job/38732069729), moving to replace the original Lint task.

internal/component/database_observability/mysql/collector/parser/parser.go
```diff
@@ -0,0 +1,23 @@
+package parser
```
cristiangreco, line 1: Thanks for the fixes @dehaansa! I do have a slight preference for leaving the interface declaration close to the caller, but it's not a big deal.

internal/component/common/kubernetes/event.go
```diff
@@ -22,10 +22,10 @@ const (
 
 type queuedEventHandler struct {
 	log   log.Logger
-	queue workqueue.RateLimitingInterface
+	queue workqueue.TypedRateLimitingInterface[Event]
```
ptodev, line 25: I suppose `RateLimitingInterface` is deprecated?

internal/service/remotecfg/remotecfg.go
```diff
@@ -260,24 +270,45 @@ var _ service.Service = (*Service)(nil)
...
+		return err
+	}
 
 	// Run the service's own controller.
 	go func() {
 		s.ctrl.Run(ctx)
 	}()
 
 	for {
+		// If the ticker is nil just wait, allow an update to set it.
```
ptodev, line 293: Why not set the ticker in the Run function? It also feels very random to wait 100 ms.

internal/component/database_observability/mysql/collector/parser/parser.go
dehaansa: Understood - this change was to avoid a cyclical dependency after adding the StatementType type.

internal/component/common/kubernetes/event.go
dehaansa: Correct

internal/service/remotecfg/remotecfg.go
dehaansa: I ended up iterating a few times on how to fix issues with the Ticker. When I added error checking for registerCollector() it caused ticker.Stop() to be called zero times (leaking goroutines) then twice (panic for closing closed channels). I ended up leaning in this direction as it was clearest and avoided the current behavior of having your tick time be at maxInt-jitter if the ticker wasn't configured.

I'll see if there's a better way to just initialize it in Run().

dehaansa: I cleaned up the ticker code, should be better now & doesn't do the random wait.

wildum (approved):
LGTM, thanks for doing this! The lint rule that makes sure that errors are handled seems super useful (I ran into this problem this week actually in another pkg)

ptodev (approved):

cristiangreco (approved):

PR #2985, opened 2025-03-13
[secretfilter] Add metrics to track the redaction of secrets

kelnage:
#### PR Description

When the secretfilter component reacts secrets from log data, it is currently challenging to track how many secrets have been redacted, where they were redacted from, and how the component performs without expensive queries over Loki. This PR introduces multiple metrics to enable these to be tracked.

#### Notes to the Reviewer

I'd be very interested in your thoughts on the value of each of these metrics and my design choices around them. I would consider the `loki_secretfilter_secrets_redacted_by_rule_total` and `loki_secretfilter_secrets_redacted_by_origin` to be the key metrics - but I think each of the others would also have some value. Arguably `loki_secretfilter_secrets_redacted_total` could be determined directly from `loki_secretfilter_secrets_redacted_by_rule_total` - would it be better to just have the latter?

#### PR Checklist

- [X] Documentation added
- [X] Tests updated


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -141,8 +153,8 @@ loki.write "local_loki" {
 
 Replace the following:
 
-* _`<PATH_TARGETS>`_: The paths to the log files to monitor.
-* _`<LOKI_ENDPOINT>`_: The URL of the Loki instance to send logs to.
+- _`<PATH_TARGETS>`_: The paths to the log files to monitor.
+- _`<LOKI_ENDPOINT>`_: The URL of the Loki instance to send logs to.
```
clayton-cornell, line 157: Minor observation... this change doesn't seem necessary. Most topics are standardised on using `*` for bullets instead of `-`.  Both work, but the markdown becomes inconsistent if we flip-flop back and forth.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -110,7 +112,17 @@ The following fields are exported and can be referenced by other components:
...
+| -------------------------------------------------- | ------- | -------------------------------------------------------------------------------------- |
+| `loki_secretfilter_secrets_redacted_total`         | Counter | Total number of secrets that have been redacted.                                       |
+| `loki_secretfilter_secrets_redacted_by_rule_total` | Counter | Number of secrets redacted, partitioned by rule name.                                  |
+| `loki_secretfilter_secrets_redacted_by_origin`     | Counter | Number of secrets redacted, partitioned by origin label value.                         |
+| `loki_secretfilter_secrets_allowlisted_total`      | Counter | Number of secrets that matched a rule but were in an allowlist, partitioned by source. |
+| `loki_secretfilter_processing_duration_seconds`    | Summary | Summary of the time taken to process and redact logs in seconds.                       |
+
+The `origin_label` argument specifies which Loki label's value should be used for the `secrets_redacted_by_origin` metric. This allows tracking how many secrets were redacted in logs from different sources or environments.
```
clayton-cornell, line 125:
```suggestion
The `origin_label` argument specifies which Loki label value should be used for the `secrets_redacted_by_origin` metric.
This allows tracking how many secrets were redacted in logs from different sources or environments.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -92,6 +92,8 @@ If set to `0`, the entire secret is redacted.
 If a secret isn't at least 6 characters long, it's entirely redacted.
 For short secrets, at most half of the secret is shown.
 
+The `origin_label` argument specifies which Loki label's value should be used for the `secrets_redacted_by_origin` metric. This allows tracking how many secrets were redacted in logs from different sources or environments.
```
clayton-cornell, line 95:
```suggestion
The `origin_label` argument specifies which Loki label value should be used for the `secrets_redacted_by_origin` metric.
This allows tracking how many secrets were redacted in logs from different sources or environments.
```

docs/sources/reference/components/loki/loki.secretfilter.md
kelnage: Apologies for that! Fixed in 9aaf44f5d :pray:

clayton-cornell:
Docs look OK for now. I'll wait on final approve for docs until after code review.

wildum (commented):
Thanks for adding this and for all the tests, this is great!
Only two small remarks:
- the changelog entry is missing
- it would be nice to have a little bit more test coverage for the update behavior in the tests: ensuring that existing metrics are not reset between two updates + what happens if I update with a different origin label and then update back to the previous origin label? Will I start from the previous value or from 0?

kelnage:
@wildum hope the new test matches your expectations - helpfully it also revealed a bug with updating the allowlists (it didn't clear the old allowlist) :+1:

dehaansa:
Re-running the test to confirm flakiness, and we'll need an updated changelog as this just missed the RC for 1.8.

dehaansa (approved):

PR #2993, opened 2025-03-14
ui: enable Graph view to show both local and remote graphs

tpaschalis:
#### PR Description

https://github.com/user-attachments/assets/c2744507-77e2-4f02-9ad8-6f84c1990d75

This PR adds a toggle to the Graph view so that we can show both the 'local' configuration dependencies as well as the ones from remotecfg components

No issue filed.

#### Notes to the Reviewer

I know that @wildum has some FE ideas in line, I hope this don't collide with what you wanted to do here.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated (N/A)

### DIFF
--- a/internal/web/ui/src/pages/Graph.tsx
+++ b/internal/web/ui/src/pages/Graph.tsx
@@ -1,15 +1,33 @@
+import { useState } from 'react';
 import { faDiagramProject } from '@fortawesome/free-solid-svg-icons';
+
+import { RadioButtonGroup } from '@grafana/ui';
 
 import { ComponentGraph } from '../features/graph/ComponentGraph';
 import Page from '../features/layout/Page';
 import { useComponentInfo } from '../hooks/componentInfo';
 
 function Graph() {
-  const [components] = useComponentInfo('', false);
+  const [configType, setConfigType] = useState<string>('local');
+  const [components] = useComponentInfo('', configType === 'remote');
+
+  const options = [
+    { label: 'Local', value: 'local' },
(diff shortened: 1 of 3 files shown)


internal/web/ui/src/pages/Graph.tsx
```diff
@@ -1,15 +1,35 @@
...
     <Page name="Graph" desc="Relationships between defined components" icon={faDiagramProject}>
-      {components.length > 0 && <ComponentGraph components={components} />}
+      <div style={{ display: 'flex', flexDirection: 'column', height: '100%' }}>
+        <div style={{ marginBottom: '16px' }}>
+          <RadioButtonGroup options={options} value={configType} onChange={handleConfigChange} size="sm" />
+        </div>
+        <div style={{ flex: 1, minHeight: 0 }}>
+          {components.length > 0 && <ComponentGraph components={components} useRemotecfg={configType === 'remote'} />}
+        </div>
+      </div>
```
tpaschalis, line 32: Cursor wrote this bit, there might be a way to simplify.

The thing is, when I included the radio button it added an annoying small scrollbar, and I wanted to get rid of it.

wildum:
This graph will be soon replaced by the live graph: https://github.com/grafana/alloy/pull/2968 (planned for next release).
We could add a similar concept in the new graph.
If the live graph does not make it in the next release, we can merge this PR. Would that work for you?

tpaschalis:
Yes, that looks much more promising! Thanks 🙏

PR #2998, opened 2025-03-14
Fix deadlock in loki.source.file

mblaschke:
#### PR Description
Locks are not released if file tailer is failing

Alloy is running into a deadlock if errors happens

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@ Main (unreleased)
 
 - Reduced lock contention in `prometheus.scrape` component (@thampiotr)
 
+### Bugfixes
+
+- Fix deadlocks in `loki.source.file` when tailing fails (@mblaschke)
+
 ### Other changes
 
 - Upgrading to Prometheus v2.55.1. (@ptodev)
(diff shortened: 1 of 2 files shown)


dehaansa:
Thanks so much for finding this! We would like to get this in as soon as possible, can you sign the CLA?

wildum:
@mblaschke thanks, that's super helpful. Did you test it at scale? It definitely fixes deadlocks, but I'm still wondering if that's the only problem with the component at the moment or not

wildum:
If needed, I'd be happy to build a new docker image or binary or windows installer for you. We just don't have a big setup right now to test it ourselves at scale

mblaschke:
CLA is signed

We haven't tested it yet if that's the only deadlock, but still looks promising.
Could you release a preview release (container image) so we can test it? Then we could try it next week.

wildum:
sure, do you need a linux amd64 docker image?

wildum (approved):
approving for the code, if you could just add a changelog entry under Main (unreleased) and ### Bugfixes that would be perfect

mblaschke:
@wildum done

we need amd64 and arm64

wildum:
@mblaschke thanks, you can find the generated images for amd64 and arm64 with your fix here: https://hub.docker.com/layers/grafana/alloy-dev/v1.8.0-devel-f47b2ba-boringcrypto/images/sha256-52e4ab38e4bbb8b1b3ab271e719f0f775be22da8fd2c40461e1479c1418f7739

wildum:
hi @mblaschke, any feedback on the new image with the fix? Did you have a chance to test it?

PR #3022, opened 2025-03-18
[helm] expand `extraObjects` templating

TheRealNoob:
#### PR Description

Expand `extraObjects` to be far more flexible to the consumer of the chart.  Couple top of mind use cases
1. By allowing `extraObjects` as a map, we can now split it across different values files, including encrypted ones.
2. The previous `{{ tpl (toYaml .) $ }}` did not allow templating of `key` names (referencing key, value pairs).  The most common case where someone would want to do this is to when defining a new object that imports the standard labels `{{ include "alloy.labels" . }}` or `{{ include "alloy.selectorLabels" . }}`.

None

#### Notes to the Reviewer

I've tested this pretty extensively in other projects.  I looked at seeing what I could do for testing here, but to do any kind of adequate testing in this scenario, requires more than simple "does it template" tests IMO.  I have a helm-unittests for this that I've written which I'm happy to share.  Note that it uses `.Values.manifests` and not `.Values.extraObjects`.
https://gist.github.com/TheRealNoob/ca7dee0e862a4e5526a115d54e863ec9

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/extra-manifests.yaml
+++ b/operations/helm/charts/alloy/templates/extra-manifests.yaml
@@ -1,4 +1,15 @@
-{{ range .Values.extraObjects }}
+{{- /* Normalize extraObjects to a list, easier to loop over */ -}}
+{{- $extraObjects := .Values.extraObjects | default (list) -}}
+
+{{- if kindIs "map" $extraObjects -}}
+  {{- $extraObjects = values $extraObjects -}}
+{{- end -}}
+
+{{- range $extraObjects }}
 ---
-{{ tpl (toYaml .) $ }}
-{{ end }}
+  {{- if kindIs "map" . }}
+    {{- tpl (toYaml .) $ | nindent 0 }}
+  {{- else if kindIs "string" . }}
+    {{- tpl . $ | nindent 0 }}
+  {{- end }}
+{{- end }}
--- a/operations/helm/charts/alloy/values.yaml
+++ b/operations/helm/charts/alloy/values.yaml
@@ -376,6 +376,8 @@ ingress:
   #      - chart-example.local
 
 # -- Extra k8s manifests to deploy
+# extraObjects can be of type map or slice.  If slice, keys are ignored and only values are used.
+# items contained within extraObjects can be defined as dict or string and are passed through tpl.
 extraObjects: []
 # - apiVersion: v1
 #   kind: Secret
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -9,6 +9,9 @@ internal API changes are not present.
 
 Unreleased
 ----------
+### Enhancements
+
+- Expand `extraObjects` to have more flexibility in templating (@TheRealNoob)
 
 0.12.5 (2025-03-13)
 ----------


skhtor:
Also eager for this to be merged. YAML merging is very useful and keeps our env-specific values files small.

operations/helm/charts/alloy/values.yaml
```diff
@@ -376,7 +376,9 @@ ingress:
   #      - chart-example.local
 
 # -- Extra k8s manifests to deploy
-extraObjects: []
+# extraObjects can be of type map or slice.  If slice, keys are ignored and only values are used.
```
ptodev, line 379:
```suggestion
# extraObjects can be of type map or slice.  If map, keys are ignored and only values are used.
```

operations/helm/charts/alloy/values.yaml
```diff
@@ -376,7 +376,9 @@ ingress:
...
+# items contained within extraObjects can be defined as dict or string and are passed through tpl.
```
ptodev, line 380:
```suggestion
# Items contained within extraObjects can be defined as dict or string and are passed through tpl.
```

operations/helm/charts/alloy/templates/extra-manifests.yaml
```diff
@@ -1,4 +1,15 @@
...
+  {{- $extraObjects = values $extraObjects -}}
+{{- end -}}
+
+{{- range $extraObjects }}
 ---
-{{ tpl (toYaml .) $ }}
-{{ end }}
+  {{- if kindIs "map" . }}
+    {{- tpl (toYaml .) $ | nindent 0 }}
+  {{- else if kindIs "string" . }}
```
ptodev, line 12: Is it possible that it's neither a map nor a string? E.g. could it contain numbers?

ptodev (commented):
Thank you for the PR! I'm not very familiar with Helm... I suppose this makes sense but it'd be nice if @petewall could sanity check it before merging 😆

petewall:
I think it's fine. I'd love to see some more examples in the values.yaml comments, especially with the whole "if its a map, the keys are thrown away". That's non-obvious to me, and we should be very careful adding things that might contribute to confusion.

operations/helm/charts/alloy/templates/extra-manifests.yaml
TheRealNoob: It could, and we could template that, but it would yield invalid kube yaml so ultimately would be rejected.  The user does need to define multiple fields, at least a minimum to render valid yaml.  As far as I can think, these are the only two object types that would allow defining a complete object.

TheRealNoob:
I use this template internally on one of our charts and recognizing that it's variability is quite wide and complex, i wrote out a helm-unittest case for it.  Happy to share it here.
https://gist.github.com/TheRealNoob/ca7dee0e862a4e5526a115d54e863ec9

TheRealNoob:
I added another example of using list, but I see your point about the confusion of using dict.  Do you have any thoughts on how that could be better-communicated without making the examples doubles the length?

TheRealNoob:
@ptodev do you need anything from me to keep this moving forward?

PR #3025, opened 2025-03-18
fix(loki.process.md): stage.tenant example

wtf-is-flying:
#### PR Description

Fix an issue with an example in loki.process documentation (loki.tenant section).


clayton-cornell (approved):
Oh, thanks for catching this!  I'll get it merged and backported as quickly as possible.

PR #3030, opened 2025-03-18
fix(pyroscope.scrape): Fix duplicated `seconds` query param

### LINKED ISSUE #3026: Duplicated seconds query parameter when scraping Pyroscope profiles
### What's wrong?

I am deploying 3 Alloy pods to Kubernetes with **clustering mode enabled**.

When scraping CPU profiles using `pyroscope.scrape`, the `seconds` query parameter gets duplicated a lot of times:

Image

I found that if I restarted one of the Alloys **that is not the one scraping the target**, the number of `seconds` params would increase by 2.

This breaks scraping for my app which requires query params to not be duplicated.

A temporary fix I found was to replace `params.Add` with `params.Set` here:
https://github.com/grafana/alloy/blob/38ab4b55318329593f6b15393322e7cc73e92561/internal/component/pyroscope/scrape/target.go#L345

### Steps to reproduce

Deploy 3 Alloys with clustering mode to scrape CPU profiles.
Kill one of the Alloys that is not scraping the target and observe the `seconds` query param is duplicated.
This is reverted to normal if you make a config change because Alloy will reload it.

### System information

Kubernetes

### Software version

Grafana Alloy v1.7.1

### Configuration

```text

```

### Logs

```text

```

Pluggi:
#### PR Description

This PR fixes the duplicated `seconds` query parameter when scraping profiles.

Fixes #3026

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ Main (unreleased)
 
 - Fix deadlocks in `loki.source.file` when tailing fails (@mblaschke)
 
+- Fix duplicated `seconds` query parameter when scraping pyroscope profiles (@Pluggi)
+
 ### Other changes
 
 - Upgrading to Prometheus v2.55.1. (@ptodev)
--- a/internal/component/pyroscope/scrape/target.go
+++ b/internal/component/pyroscope/scrape/target.go
@@ -342,7 +342,7 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments, targetTypes map[s
 					if cfg.DeltaProfilingDuration != defaultProfilingDuration {
 						seconds = (cfg.DeltaProfilingDuration) / time.Second
 					}
-					params.Add("seconds", strconv.Itoa(int(seconds)))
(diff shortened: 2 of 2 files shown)


dehaansa (commented):
I see how this would duplicate the "seconds" param if it was already present, looks like a straightforward fix, thanks for the contribution.

Could you add a unit test validating the fix?

Pluggi:
I am not sure how to test that actually.

korniltsev (approved):
The change looks good to me, however I could not reproduce the issue with the suggested clsutering mode node kill.
I have no objections to merge this.
However I think it makes sense to understand the root cause.
Can you share your full scraping & discovery configuration. And full output of debug UI of scrape and discovery components.

dehaansa:
This will require an update off main and a re-sign of the CLA to get merged.

korniltsev:
I will close the PR as no progress in a while. Feel free to reopen and we will merge it

PR #3036, opened 2025-03-19
database_observability: rename query_sample collector

matthewnolf:
#### PR Description
This change updates the name of the collector responsible for collecting table names from a given query from `query_samples` to `query_tables`.

This is to better reflect the responsibility of the component - which is just that - to map queries to the tables involved.

The name is plural, to reflect that a query can use multiple tables. However happy to make it singular if it better fits with convention. 

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


cristiangreco (approved):
LGTM - docs need an update as well, but approving meanwhile.

internal/component/database_observability/mysql/collector/query_tables.go
```diff
@@ -111,19 +111,19 @@ func (c *QuerySample) Start(ctx context.Context) error {
...
 	c.cancel()
 }
 
-func (c *QuerySample) fetchQuerySamples(ctx context.Context) error {
-	rs, err := c.dbConnection.QueryContext(ctx, selectQuerySamples)
+func (c *QueryTables) fetchQueryTables(ctx context.Context) error {
+	rs, err := c.dbConnection.QueryContext(ctx, selectQueryTablesSamples)
 	if err != nil {
-		level.Error(c.logger).Log("msg", "failed to fetch query samples", "err", err)
+		level.Error(c.logger).Log("msg", "failed to fetch query tables", "err", err)
```
cristiangreco, line 126: Maybe this log line can stay as is? Because it's actually still fetching samples here.

matthewnolf: The reason I renamed this was that we need to avoid the collision when we introduce the new query sample collector.

The purpose of this query is to fetch the table info - the fact it is still fetching sample info is a caveat of not yet removing that logic here.

cristiangreco: Ok I see the point on log collision, but this is actually fetching samples still, not tables. Maybe change it to something line `failed to fetch query samples from events_statements_summary_by_digest`?

matthewnolf: Updated in 8d3f533f82cb44c7263add010a6c700bbfaa70ca

fridgepoet (approved):

Issue #2949, opened 2025-03-10
Support converting telemetry service from Otel config file


wildum:
The internal telemetry of the Otel collector can be configured (see https://opentelemetry.io/docs/collector/internal-telemetry/#activate-internal-telemetry-in-the-collector).

We should have a way to translate this into an Alloy-equivalent configuration.

We should maybe wait before starting on this because it might change a lot after 1.0

PR #3038, opened 2025-03-19
Add basic support for OTel's `service/telemetry` config argument

ptodev:
#### PR Description

So far Alloy's OTel converters have been ignoring settings under the `service/telemetry` section of an OTel config file. This PR adds support for a two arguments in the logging section - `level` and `encoding`. For the other ones Alloy will issue diagnostics to warn the users that they're not supported.

For `encoding`, OTel's `console` is not exactly the same as Alloy's `logfmt`. But the intention behind them is kind of similar - to produce human-readable logs. So I think it's ok not to have a warning for now. We could change this later on.

Related to #2949, but I'm not sure if we should close the ticket yet. I'd prefer to keep it open because this PR only adds very basic support.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/converter/internal/otelcolconvert/converter_service_telemetry.go
```diff
@@ -0,0 +1,131 @@
...
+	if len(tel.OutputPaths) > 0 &&
+		// If this is set to the default (stderr) then it's the same as Alloy default.
+		// There's no need to a diagnostic.
+		!(len(tel.OutputPaths) == 1 && tel.OutputPaths[0] == "stderr") {
+		diags.Add(diag.SeverityLevelCritical, "the service/telemetry/logs/output_paths configuration is not supported")
+	}
+	if tel.ErrorOutputPaths != nil &&
+		// If this is set to the default (stderr) then it's the same as Alloy default.
+		// There's no need to a diagnostic.
+		!(len(tel.ErrorOutputPaths) == 1 && tel.ErrorOutputPaths[0] == "stderr") {
```
wildum, line 92: is `len(tel.ErrorOutputPaths) == 1` necessary? because you are already checking above than tel.ErrorOutputPaths is not empty. Same remark for the tel.OutputPaths above

internal/converter/internal/otelcolconvert/otelcolconvert_test.go
```diff
@@ -9,14 +9,32 @@ import (
...
 )
 
+var diagsToIgnore map[string]struct{}
+
+func init() {
+	// These diagnostics are expected and we only want to check them in TestConvertTelemetry.
+	// If we check them in every test we'd have to create too many ".diag" files.
+	diagsToIgnore = map[string]struct{}{
+		"(Warning) the service/telemetry/logs/sampling configuration is not supported":                                                                                                          {},
+		"(Warning) the service/telemetry/metrics/readers configuration is not supported - use the `--server.http.listen-addr` cmd argument or the `prometheus.exporter.self` component instead": {},
```
wildum, line 19: These diags are generated on every conversion, even on the one where the telemetry service is not added?
I think that it's a bit hard for users to understand what they mean, especially if they don't monitor their otel collector and they don't know how it works. Maybe instead of having these two we could write about it in the doc and link to this: https://grafana.com/docs/alloy/latest/collect/metamonitoring/

wildum (commented):
Thanks for adding this, I think that's a reasonable start. Could you add a TODO comment in internal/converter/internal/otelcolconvert/converter_service_telemetry.go to explain what work would be needed to support this telemetry conversion better?

internal/converter/internal/otelcolconvert/otelcolconvert_test.go
ptodev: I figured they're useful since we don't know whether the users didn't add a `telemetry` setting because they like the defaults or because they don't care about it. 

The `metrics/readers` warning could be especially useful to a lot of people. It's very common to monitor the Collector and it's normal to expect Alloy to show its metrics in the same way as the Collector.

I don't really understand the argument to document instead of producing a warning. The point of the warnings is to alert you if something doesn't work, and that's what this is doing. People are likely to just gloss over the docs and then try things out themselves to see what fails.

ptodev: > These diags are generated on every conversion, even on the one where the telemetry service is not added?

Yep.

internal/converter/internal/otelcolconvert/converter_service_telemetry.go
ptodev: If there are two elements in the slice, e.g. `stderr` and `~/log.txt`, then the conversion should log a diagnostic because the output to a file won't work.

wildum: ah my bad I misread the condition, makes sense

internal/converter/internal/otelcolconvert/otelcolconvert_test.go
ptodev: Ah sorry, at first I thought you mean that we should just write them in the doc and not have diagnostics! I would be happy to link to a doc from a diagnostic. It's a good idea - I agree that many people may not be familiar with this.

ptodev: I changed the message to:

> the service/telemetry/metrics/readers configuration is not supported - to gather Alloy's own telemetry refer to: https://grafana.com/docs/alloy/latest/collect/metamonitoring/

wildum (approved):

clayton-cornell (approved):

PR #3040, opened 2025-03-20
feat(pyroscope.write): Propagate error from backend on /ingest

simonswine:
This change will make sure error bodies are propagated to the user and it is also clear which endpoint returned the error.


marcsanmi (approved):
LGTM, thanks for improving it 👍

PR #3041, opened 2025-03-20
refactor(pyroscope.write): Forward content-type header only.

### LINKED ISSUE #2612: pyroscope.receive_http treats headers inconsistently
`pyroscope.receive_http` supports forwarding pushv1 and ingest handler.

For the pushv1 non of the headers are forwarded, while the ingest handler forwards all but a block list of headers:

https://github.com/simonswine/alloy/blob/755bcfcb5ab193488a226f467216b93ce98c6cd1/internal/component/pyroscope/write/write.go#L45-L52

I suggest we should only forward headers, that are important for parsing of the profiles, which includes:

- [Content-Type](https://github.com/simonswine/pyroscope/blob/b57274a5acbfa867660938ba7e7f5171c3114d68/pkg/ingester/pyroscope/ingest_handler.go#L132)

simonswine:
We currently forward all incoming headers of the profile /ingest request
to the endpoints defined in `pyroscope.write`.

I think this is risky. We already had bugs (like #2201) and there probably
a couple more in there already, as I noticed we send on headers like
`Content-Length`.

I think we should change this and only forward `Content-Type`, before
anyone starts relying on this, as this is the only header used:

https://github.com/simonswine/pyroscope/blob/b57274a5acbfa867660938ba7e7f5171c3114d68/pkg/ingester/pyroscope/ingest_handler.go#L132

This is still technically a breaking change, but I think it should go
forward, as the stabililty of the component is only "public-preview"
after all.

Fixes #2612

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,8 @@ Main (unreleased)
 
 - Change the stability of the `beyla.ebpf` component from "public preview" to "generally available". (@marctc)
 
+- The ingest API of `pyrscope.receive_http` no longer forwards all received headers, instead only passes through the `Content-Type` header. 
+
 v1.7.5
 -----------------
 
--- a/internal/component/pyroscope/appender.go
+++ b/internal/component/pyroscope/appender.go
@@ -2,7 +2,6 @@ package pyroscope
 
 import (
 	"context"
-	"net/http"
 	"net/url"
 	"sync"
 	"time"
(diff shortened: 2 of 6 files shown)


marcsanmi:
Should we update the `CHANGELOG.md`?

marcsanmi (approved):
LGTM 👍 
I've tested it locally with the go SDK and everything is working.

dehaansa (approved):
Needs a changelog entry, otherwise 👍

Issue #280, opened 2024-01-04
kafka-exporter bug with tls ca


max107:
### What's wrong?

impossible to use tls with ca only (without cert&key pairs)

### Steps to reproduce

working docker-compose example

```
version: '3.9'
services:
  agent: # dead
    image: docker.io/grafana/agent:main-6e4a9b9
    environment:
      AGENT_MODE: flow
      CONFIG_FILE_PATH: /config.river
    command:
      - run
      - /config.river
      - --storage.path=/tmp/agent
      - --server.http.listen-addr=0.0.0.0:80
      - --server.http.ui-path-prefix=/
      - --disable-reporting
      - --cluster.enabled=false
    ports:
      - "8080:80"
    volumes:
      - ./example.river:/config.river:ro
      - ./ca.pem:/pki/ca.pem:ro
  exporter: # alive
    image: dparrott/kafka-exporter:latest
    command:
      - --sasl.enabled
      - --sasl.handshake
      - --sasl.username=example
      - --sasl.password=example
      - --sasl.mechanism=plain
      - --tls.enabled
      - --tls.ca-file=/pki/ca.pem
      - --kafka.server=example-broker.com
    ports:
      - "9308:9308"
    volumes:
      - ./ca.pem:/pki/ca.pem:ro
```

### Software version

docker.io/grafana/agent:main-6e4a9b9 (ttps://github.com/grafana/agent/issues/6044)

### Configuration

```text
logging {
    level  = "debug"
    format = "logfmt"
}

prometheus.exporter.kafka "example" {
    kafka_uris = ["example-broker.com"]
    ca_file = "/pki/ca.pem"
    use_tls = true
    instance = "example"
    use_sasl = true
    use_sasl_handshake = true
    sasl_mechanism = "plain"
    sasl_username = "example"
    sasl_password = "example"
}

prometheus.scrape "example" {
    targets    = prometheus.exporter.kafka.example.targets
    forward_to = [prometheus.remote_write.default.receiver]
}
```
```


### Logs

```text
↪ docker compose up agent
[+] Running 1/0
 ✔ Container ha-agent-1  Created                                                                                                                                       0.0s
Attaching to agent-1
agent-1  | ts=2024-01-04T15:41:51.71027843Z level=info "boringcrypto enabled"=false
agent-1  | ts=2024-01-04T15:41:51.711410456Z level=info msg="starting complete graph evaluation" controller_id="" trace_id=25843d6229f77cc08cf3d378255624bc
[...]

PR #3070, opened 2025-03-23
fix: Allow various SSL configs for kafka exporter

dehaansa:
#### PR Description
The kafka exporter internally does not _require_ cert & key like the Alloy integration to it does. The exporter will attempt to connect with CAs provided on the host if CA & key/cert are not provided, or will use the CA without key/cert if provided. I think the best path here is to avoid examining the TLS-applicable arguments and just allow the exporter to attempt a connection in whichever way it is configured by the user.

This matches expectations already set in the documentation (all three fields are called optional in the docs, with no special note).

Fixes https://github.com/grafana/alloy/issues/280

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@ Main (unreleased)
 
 - Fix deadlocks in `loki.source.file` when tailing fails (@mblaschke)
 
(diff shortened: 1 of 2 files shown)


wildum (approved):
lgtm

PR #3084, opened 2025-03-25
Change backend frontend com for live graph

wildum:
The message in the front-end would sometimes be split in two, resulting in a parsing error and the UI crashing.

This approach simplifies the process by attempting to convert the result into a json array instead of using delimiters.

The data is buffered in the frontend till it receives the full message. I added a safeguard in case the server pushes only invalid data.

### DIFF
--- a/internal/web/ui/src/hooks/graph.tsx
+++ b/internal/web/ui/src/hooks/graph.tsx
@@ -31,18 +31,36 @@ export const useGraph = (
 
         const reader = response.body.getReader();
         const decoder = new TextDecoder();
+        let buffer = '';
+        let skipCount = 0;
+        const MAX_SKIP_COUNT = 30;
 
         while (enabled) {
           const { value, done } = await reader.read();
-          if (done) {
-            break;
-          }
+          if (done) break;
 
-          const decodedChunks = decoder
-            .decode(value, { stream: true })
-            .split('|;|')
-            .filter((entry) => entry.length !== 0);
-          setData(() => decodedChunks.map((chunk) => JSON.parse(chunk)));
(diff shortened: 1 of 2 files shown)


internal/web/ui/src/hooks/graph.tsx
```diff
@@ -31,18 +31,36 @@ export const useGraph = (
...
-          setData(() => decodedChunks.map((chunk) => JSON.parse(chunk)));
+          try {
+            const data = JSON.parse(buffer) as DebugData[];
+            skipCount = 0;
+            buffer = '';
+            setData(data);
+          } catch (err) {
+            skipCount++;
+            // This is a safeguard to avoid growing the buffer indefinitely if the server is sending invalid data.
+            if (skipCount >= MAX_SKIP_COUNT) {
```
ptodev, line 54: Why is it acceptable to have up to 30 errors?

internal/web/ui/src/hooks/graph.tsx
```diff
@@ -31,18 +31,36 @@ export const useGraph = (
...
+        let buffer = '';
+        let skipCount = 0;
+        const MAX_SKIP_COUNT = 30;
 
         while (enabled) {
           const { value, done } = await reader.read();
-          if (done) {
-            break;
-          }
+          if (done) break;
```
ptodev, line 40: Pardon my React ignorance, but do we not need to use `value` before we break?

wildum: the reader returns if there is a value or if the connection has ended. If the connection has ended, we don't care what value is (it's most likely if not always empty in this case), because we will remove the graph and just write an error saying that there is a network issue

ptodev (approved):

PR #3087, opened 2025-03-25
docs: Update discovery.relabel docs to include static label example

dehaansa:
#### PR Description
Had a few requests over the past few months for how to add static labels, adding an example to discovery.relabel docs so it's a use case that is discoverable in docs.


clayton-cornell (approved):
Adding comments to the example config is a HUGE improvement. I've had this on a ToDo list for a long time and just haven't got to it yet.

PR #3114, opened 2025-03-26
database_observability: refactor parsing of truncated queries

cristiangreco:
#### PR Description
Move parsing of truncated queries to the parser implementations and log out as error when a query is truncated.

The parsers still have code copied, but we'll eventually drop one of the implementations

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/parser/parser_tidb_test.go
+++ b/internal/component/database_observability/mysql/collector/parser/parser_tidb_test.go
@@ -4,8 +4,9 @@ import (
 	"testing"
 
 	"github.com/go-kit/log"
-	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
 	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
 )
 
 func TestParserTiDB_ExtractTableNames(t *testing.T) {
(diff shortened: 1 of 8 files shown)


fridgepoet (approved):

internal/component/database_observability/mysql/collector/parser/parser_tidb_test.go
```diff
@@ -146,3 +147,51 @@ func TestParserTiDB_ExtractTableNames(t *testing.T) {
 		})
 	}
 }
+
+func TestParserTiDB_CleanTruncatedText(t *testing.T) {
+	testcases := []struct {
+		name    string
+		sql     string
+		want    string
+		wantErr bool
```
matthewnolf, line 156: We should assert on the error containing the expected info/type too. 
Instead of wantErr being a `bool`, it could be an `error`

internal/component/database_observability/mysql/collector/parser/parser_xwb_test.go
```diff
@@ -135,3 +135,51 @@ func TestParserXwb_ExtractTableNames(t *testing.T) {
 		})
 	}
 }
+
+func TestParserXwb_CleanTruncatedText(t *testing.T) {
+	testcases := []struct {
+		name    string
+		sql     string
+		want    string
+		wantErr bool
```
matthewnolf, line 144: Same comment applies here

matthewnolf (approved):

Issue #2651, opened 2025-02-07
Set a high priority to the Alloy process on Windows


ptodev:
### Request

It would be good if there is a way to configure Alloy to always have a high priority such as "realtime" or "high". I don't think it should be the default though. Alloy may not always be a very critical piece of software for the person who installed it.

There could be a few ways to do this:
* Via Alloy's config file. This would have to be a top-level service block similar to [remotecfg](https://grafana.com/docs/alloy/latest/reference/config-blocks/remotecfg/), because there cannot be more than one such block for an Alloy instance. We'd need to make sure the priority is set accordingly. E.g. if an Alloy instance is ran with a `windows` block that gives it a high priority, and the block is removed and the config is reloaded, Alloy should go back to using normal priority.
* Via a new Alloy command line argument. This is simpler. The downside is that editing the command line parameters is usually more difficult than editing the config file. On Windows, the command line parameters for the Alloy service are stored in `HKEY_LOCAL_MACHINE\SOFTWARE\GrafanaLabs\Alloy`.

### Use case

Alloy may not work properly when the machine is resource constrained. Since Alloy is used for monitoring, many users would it to work even if there is not much free memory and CPU. That way they can spot issues with the OS and with applications.

A customer recently noticed that their Alloy instance wasn't sending metrics when the CPU consumption is high. To resolve this, they:
1. Set the priority of the Alloy process to "realtime".
2. Disabled the service collector in [prometheus.exporter.windows](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.windows/).

Nachtfalkeaw:
I have alloy running in an Environment with high load and alloy stops sending Windows metrics daily between 8am and 5pm.

Changing priority of the process is ok If it helps.
Except only Set it to "high" or "Realtime" please allow to Set it to lower priority as normal If possible. Sometimes Business critical appliances are more important than alloy.


Another Thing i would Like to See ist a metric or a Log Message If alloy can Not get the ressources it needs and stops collecting metrics. However i am Not Sure If this is possible.

Nachtfalkeaw:
@ptodev I am one of the persons who is Testing this new 1.7.x Loki.source.windowsevent

dehaansa:
This should be available as of release 1.8 (next weekish)

PR #3119, opened 2025-03-26
feat: Set Windows process priority for Alloy

dehaansa:
#### PR Description
Add a new `--windows.priority` flag that allows the user to set the runtime priority of the alloy process. Set this to public preview so that we can change the way this works (move to block, apply from service config) if this doesn't meet user needs.

The flag is applied within the first few actions of the `run` command, and the setting cannot be changed while running.

image

Fixes https://github.com/grafana/alloy/issues/2651

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/cli/run.md
+++ b/docs/sources/reference/cli/run.md
@@ -63,6 +63,13 @@ The following flags are supported:
 * `--feature.prometheus.metric-validation-scheme`: Prometheus metric validation scheme to use. Supported values: `legacy`, `utf-8`. NOTE: this is an experimental flag and may be removed in future releases (default `"legacy"`).
 * `--windows.priority`: The priority to set for the {{< param "PRODUCT_NAME" >}} process when running on Windows. This is only available on Windows. Supported values: `above_normal`, `below_normal`, `normal`, `high`, `idle`, `realtime` (default `"normal"`).
 
+{{< admonition type="note" >}}
+The `--windows.priority` flag is in [Public preview][] and is not covered by {{< param "FULL_PRODUCT_NAME" >}} [backward compatibility][] guarantees.
+
+[Public preview]: https://grafana.com/docs/release-life-cycle/
+[backward compatibility]: ../introduction/backward-compatibility/
+{{< /admonition >}}
+
 ## Update the configuration file
 
 The configuration file can be reloaded from disk by either:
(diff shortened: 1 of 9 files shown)


docs/sources/reference/cli/run.md
```diff
@@ -61,6 +61,14 @@ The following flags are supported:
 * `--stability.level`: The minimum permitted stability level of functionality to run. Supported values: `experimental`, `public-preview`, `generally-available` (default `"generally-available"`).
 * `--feature.community-components.enabled`: Enable community components (default `false`).
 * `--feature.prometheus.metric-validation-scheme`: Prometheus metric validation scheme to use. Supported values: `legacy`, `utf-8`. NOTE: this is an experimental flag and may be removed in future releases (default `"legacy"`).
+* `--windows.priority`: The priority to set for the {{< param "PRODUCT_NAME" >}} process when running on Windows. This is only available on Windows. Supported values: `above_normal`, `below_normal`, `normal`, `high`, `idle`, `realtime` (default `"normal"`).
```
clayton-cornell, line 64:
```suggestion
* `--windows.priority`: The priority to set for the {{< param "PRODUCT_NAME" >}} process when running on Windows. This is only available on Windows. Supported values: `above_normal`, `below_normal`, `normal`, `high`, `idle`, or `realtime` (default `"normal"`).
```

dehaansa: I was copying the example from `--stability.level` just above, which listed without an "or". Happy to change, just noting the inconsistency.

internal/util/windowspriority/set_priority_windows.go
```diff
@@ -0,0 +1,55 @@
...
+	if !ok {
+		return 0, fmt.Errorf("invalid priority: %s", priorityString)
+	}
+	return priority, nil
+}
+
+func SetPriority(priorityString string) error {
+	if priorityString == "" {
+		return nil
+	}
```
ptodev, line 43: Why not omit this and let `TranslatePriority` return an error?

ptodev (approved):

ptodev:
We we want to have an argument in the installer to setup this feature?

internal/util/windowspriority/set_priority_windows.go
dehaansa: I think it's a leftover artifact of when I was attempting to have the main code be cross-platform and the os-specific implementation be smaller. Will clean up.

dehaansa:
> Do we want to have an argument in the installer to setup this feature?

I took a pass at this - do we have a good way to test the windows installer?

dehaansa:
Confirmed the installer works, had to fix some string escaping.

PR #3121, opened 2025-03-26
deps(ui): update minor+patch versions

jharvey10:
- Some packages are held back a release or two for compatibility reasons.
- When these package are updated upstream, their pinned deps/resolutions can be relaxed


internal/web/ui/yarn.lock
```diff
@@ -3139,13 +3089,19 @@
...
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-18.2.6.tgz#5cd53ee0d30ffc193b159d3516c8c8ad2f19d571"
-  integrity sha512-wRZClXn//zxCFW+ye/D2qY65UsYP1Fpex2YXorHc8awoNamkMZSvBxwxdYVInsHOZZd2Ppq8isnSzJL5Mpf8OA==
+"@types/react@*":
+  version "19.0.12"
+  resolved "https://registry.yarnpkg.com/@types/react/-/react-19.0.12.tgz#338b3f7854adbb784be454b3a83053127af96bd3"
+  integrity sha512-V6Ar115dBDrjbtXSrS+/Oruobc+qVbbUxDFC1RSbRqLt5SYvxxyIDrSC85RWml54g+jfNeEMZhEj7wW07ONQhA==
+  dependencies:
+    csstype "^3.0.2"
+
+"@types/react@^18.3.20":
```
dehaansa, line 3099: This will make the change in ComponentBody unnecessary, I don't understand why yarn reaaaaaally wants to resolve `@types/react@*` to 19.x.

```suggestion
"@types/react@^18.3.20", "@types/react@*":
```

jharvey10: Thanks! Made that change by pinning the dep in `package.json` and it did the trick. No more type issue!

dehaansa (approved):

PR #3123, opened 2025-03-27
Small ui graph improvements

### LINKED ISSUE #3113: Alloy graph UI improvements
- component names are currently truncated to make it look nicer but it can be confusing for the user. It's better to show the full names.
- when clicking on a component, the page switches to the component details page. To avoid losing the graph setup, it should open a new tab for the component details page
- allow users to zoom out more

wildum:
#### PR Description

- allow zooming out to a minimum of 0.1 instead of 0.5
- onClick the component detail page is open is a new tab instead of the same one
- remove the truncating of the component name

Fixes #3113


thampiotr (approved):
Thank you! <3

PR #3132, opened 2025-03-27
database_observability: add support for expressions in index definition

cristiangreco:
#### PR Description
Fetch index expresssion, if present, and store it in the index spec.

n.a.

#### PR Checklist
- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -432,30 +434,48 @@ func (c *SchemaTable) fetchColumnsDefinitions(ctx context.Context, schemaName st
...
-				level.Error(c.logger).Log("msg", "unexpected index column sequence", "schema", schemaName, "table", tableName, "index", indexName, "column", columnName)
+			if len(lastIndex.Columns)+len(lastIndex.Expressions) != seqInIndex-1 {
+				level.Error(c.logger).Log("msg", "unexpected index ordinal position", "schema", schemaName, "table", tableName, "index", indexName, "seq", seqInIndex, "len_columns", len(lastIndex.Columns), "len_expressions", len(lastIndex.Expressions))
 				continue
 			}
-			lastIndex.Columns = append(lastIndex.Columns, columnName)
+
+			if columnName.Valid {
+				lastIndex.Columns = append(lastIndex.Columns, columnName.String)
+			} else if expression.Valid {
```
fridgepoet, line 462: What if you have both a columnName and expression?

cristiangreco: Docs says they are mutually exclusive, and there's no syntax in current mysql versions to create an index with both a column and an expression. We could add support for such a possible edge case, but I'm thinking it might be better to do it when/if we find anything odd (just to avoid supporting all possible edge cases as it makes the code harder to follow).

matthewnolf (approved):

PR #3140, opened 2025-03-29
feat: Add `otelcol.exporter.googlecloud` component for exporting metrics/traces/logs to Google Cloud

### LINKED ISSUE #732: Support `googlecloudexporter` in alloy
### Request

I would love to rip out every other exporter and just use alloy because it's awesome. Google has a cloud exporter for all of their metrics/tracing/logging sinks, and I'd love to be able to use i: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter

### Use case

We use google as a sink for telemetry, but the google ops agent doesn't work on prem, hence our wanting to use a collector that sends to google.

motoki317:
#### PR Description

Adds `otelcol.exporter.googlecloud` component, which is a wrapper of [googlecloudexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter) in opentelemetry-collector-contrib repo.
I referred to the existing `otelcol.exporter.datadog` component implementation and documentation for writing this patch.

I've tested all data types (logs, metrics, and traces) on an actual GCP project.

<details>

<summary>Testing procedures</summary>

1. Using the example from OpenTelemetry here: https://opentelemetry.io/docs/languages/go/getting-started/
2. Modify the code to export to localhost:4317 (with insecure),
3. Run alloy with the following config:
```alloy
otelcol.receiver.otlp "receiver" {
  grpc {
    endpoint = "0.0.0.0:4317"
  }

  output {
    metrics = [otelcol.exporter.googlecloud.gcp.input]
    logs = [otelcol.exporter.googlecloud.gcp.input]
    traces = [otelcol.exporter.googlecloud.gcp.input]
  }
}

otelcol.exporter.googlecloud "gcp" {
  log {
    default_log_name = "opentelemetry.io/collector-exported-log"
  }
}
```
4. Access the sample app, and get the following results:

image
image
image

</details>

resolves #732 

#### Notes to the Reviewer

- The documentation for arguments were excerpted from [the original documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter).
- Also, since this is my first contribution to this project, feel free to let me know if I missed something. Thank you for reviewing!

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.googlecloud/
+description: Learn about otelcol.exporter.googlecloud
+aliases:
+  - ../otelcol.exporter.googlecloud/ # /docs/alloy/latest/reference/components/otelcol.exporter.googlecloud/
```
clayton-cornell, line 5:
```suggestion
description: Learn about otelcol.exporter.googlecloud
```

This is a new topic. It never existed at the alias location, so we shouldn't add an alias to that location.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+title: otelcol.exporter.googlecloud
+---
+
+
+<span class="badge docs-labels__stage docs-labels__item">Community</span>
```
clayton-cornell, line 10:
```suggestion
---

<span class="badge docs-labels__stage docs-labels__item">Community</span>
```

Removed extra CR/LF

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+# otelcol.exporter.googlecloud
```
clayton-cornell, line 12:
```suggestion
# `otelcol.exporter.googlecloud`
```

Style change that is in-progress for the rest of the `otelcol` topics.  We can get this topic ready so it won't need as much of an edit later.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+`otelcol.exporter.googlecloud` is a wrapper over the upstream OpenTelemetry Collector `googlecloud` exporter from the `otelcol-contrib`  distribution.
+Bug reports or feature requests will be redirected to the upstream repository, if necessary.
+{{< /admonition >}}
+
+You can specify multiple `otelcol.exporter.googlecloud` components by giving them different labels.
+
+## Usage
+
+```alloy
+otelcol.exporter.googlecloud "LABEL" {
```
clayton-cornell, line 28:
```suggestion
otelcol.exporter.googlecloud "<LABEL>" {
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+### Authenticating
+
+Refer to the original [Google Cloud Exporter][] document.
+
+[Google Cloud Exporter]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter/README.md
+
+## Arguments
+
+If there are any discrepancies between the argument descriptions here and those in the original [Google Cloud Exporter][] documentation,
+the original documentation takes precedence.
```
clayton-cornell, line 41:
```suggestion
If there are any discrepancies between the argument descriptions here and those in the original [Google Cloud Exporter][] documentation, the original documentation takes precedence.
```

We can keep this as one line

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+Argument descriptions are excerpted directly from the original documentation.
```
clayton-cornell, line 42: Do we need to state this?

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+|-----------------------------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------|----------|
+| `project`                   | `string` | GCP project identifier.                                                                                                                                                                                                           | Fetch from credentials                        | no       |
+| `destination_project_quota` | `bool`   | Counts quota for traces and metrics against the project to which the data is sent (as opposed to the project associated with the Collector's service account. For example, when setting project_id or using multi-project export. | `false`                                       | no       |
```
clayton-cornell, line 49:
```suggestion
| `destination_project_quota` | `bool`   | Counts quota for traces and metrics against the project to which the data is sent as opposed to the project associated with the Collector's service account. For example, when setting `project_id` or using multi-project export. | `false`                                       | no       |
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+[Google Cloud Exporter]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter/README.md
+
+## Arguments
+
+If there are any discrepancies between the argument descriptions here and those in the original [Google Cloud Exporter][] documentation,
+the original documentation takes precedence.
+Argument descriptions are excerpted directly from the original documentation.
+
+The following arguments are supported:
```
clayton-cornell, line 44:
```suggestion
You can use the following arguments with `otelcol.exporter.googlecloud`:
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| `project`                   | `string` | GCP project identifier.                                                                                                                                                                                                           | Fetch from credentials                        | no       |
+| `destination_project_quota` | `bool`   | Counts quota for traces and metrics against the project to which the data is sent (as opposed to the project associated with the Collector's service account. For example, when setting project_id or using multi-project export. | `false`                                       | no       |
+| `user_agent`                | `string` | Override the user agent string sent on requests to Cloud Monitoring (currently only applies to metrics). Specify `{{version}}` to include the application version number.                                                         | `opentelemetry-collector-contrib {{version}}` | no       |
+
+## Blocks
+
+The following blocks are supported inside the definition of `otelcol.exporter.googlecloud`:
```
clayton-cornell, line 54:
```suggestion
You can use the following blocks with `otelcol.exporter.googlecloud`:
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+| Hierarchy     | Block             | Description                                                                | Required |
+|---------------|-------------------|----------------------------------------------------------------------------|----------|
+| impersonate   | [impersonate][]   | Configuration for service account impersonation                            | no       |
+| metric        | [metric][]        | Configuration for sending metrics to Cloud Monitoring.                     | no       |
+| trace         | [trace][]         | Configuration for sending traces to Cloud Trace.                           | no       |
+| log           | [log][]           | Configuration for sending metrics to Cloud Logging.                        | no       |
+| sending_queue | [sending_queue][] | Configures batching of data before sending.                                | no       |
+| debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no       |
```
clayton-cornell, line 63:
```suggestion
| Block                            | Description                                                                | Required |
| -------------------------------- | -------------------------------------------------------------------------- | -------- |
| [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
| [`impersonate`][impersonate]     | Configuration for service account impersonation                            | no       |
| [`log`][log]                     | Configuration for sending metrics to Cloud Logging.                        | no       |
| [`metric`][metric]               | Configuration for sending metrics to Cloud Monitoring.                     | no       |
| [`sending_queue`][sending_queue] | Configures batching of data before sending.                                | no       |
| [`trace`][trace]                 | Configuration for sending traces to Cloud Trace.                           | no       |
```

Remove Hierarchy column. Add in-line code for block names and fix linking. Sort rows alphabetically.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+[impersonate]: #impersonate-block
+[metric]: #metric-block
+[trace]: #trace-block
+[log]: #log-block
+[sending_queue]: #sending_queue-block
+[debug_metrics]: #debug_metrics-block
```
clayton-cornell, line 70:
```suggestion
[impersonate]: #impersonate
[metric]: #metric
[trace]: #trace
[log]: #log
[sending_queue]: #sending_queue
[debug_metrics]: #debug_metrics
```

Remove `block` from link - headings are updated

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+### impersonate block
```
clayton-cornell, line 72:
```suggestion
### `impersonate`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+|--------------------|----------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|----------|
+| `target_principal` | `string`       | TargetPrincipal is the email address of the service account to impersonate.                                                                                                                  |         | yes      |
+| `subject`          | `string`       | Subject is the sub field of a JWT. This field should only be set if you wish to impersonate as a user. This feature is useful when using domain wide delegation.                             | `""`    | no       |
+| `delegates`        | `list(string)` | Delegates are the service account email addresses in a delegation chain. Each service account must be granted roles/iam.serviceAccountTokenCreator on the next service account in the chain. | `[]`    | no       |
```
clayton-cornell, line 80:
```suggestion
| `delegates`        | `list(string)` | Delegates are the service account email addresses in a delegation chain. Each service account must be granted roles/iam.serviceAccountTokenCreator on the next service account in the chain. | `[]`    | no       |
| `subject`          | `string`       | Subject is the sub field of a JWT. This field should only be set if you wish to impersonate as a user. This feature is useful when using domain wide delegation.                             | `""`    | no       |
```

Required first, then optional. Alphabetic sorting.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+### metric block
```
clayton-cornell, line 82:
```suggestion
### `metric`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| `experimental_wal` > `directory`       | `string`       | Path to local directory for WAL file.                                                                                                                                                                                                                 | `./`                                                     | no       |
+| `experimental_wal` > `max_backoff`     | `string`       | Max duration to retry requests on network errors (`UNAVAILABLE` or `DEADLINE_EXCEEDED`).                                                                                                                                                              | `1h`                                                     | no       |
+
+The `>` symbol indicates deeper levels of nesting. For example, `experimental_wal > directory` refers to a directory argument defined inside a experimental_wal block.
```
clayton-cornell, line 108:
```suggestion
The > symbol indicates deeper levels of nesting.
For example, `experimental_wal` > `directory` refers to a `directory` argument defined inside a `experimental_wal `block.
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| `experimental_wal`                     | `list(object)` | If provided, enables use of a write ahead log for time series requests. Each object must contain `directory`                                                                                                                                          | `[]`                                                     | no       |
+| `experimental_wal` > `directory`       | `string`       | Path to local directory for WAL file.                                                                                                                                                                                                                 | `./`                                                     | no       |
+| `experimental_wal` > `max_backoff`     | `string`       | Max duration to retry requests on network errors (`UNAVAILABLE` or `DEADLINE_EXCEEDED`).                                                                                                                                                              | `1h`                                                     | no       |
```
clayton-cornell, line 106:
```suggestion
| Name                                   | Type           | Description                                                                                                                                                                                                                 | Default                                                  | Required |
| -------------------------------------- | -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- | -------- |
| `compression`                          | `string`       | Compression format for Metrics gRPC requests. Supported values: [`gzip`].                                                                                                                                                   | `""` (no compression)                                    | no       |
| `create_metric_descriptor_buffer_size` | `number`       | Buffer size for the  channel which asynchronously calls CreateMetricDescriptor.                                                                                                                                             | `10`                                                     | no       |
| `create_service_timeseries`            | `bool`         | If true, this sends all timeseries using `CreateServiceTimeSeries`. Implicitly, this sets `skip_create_descriptor` to true.                                                                                                 | `false`                                                  | no       |
| `cumulative_normalization`             | `bool`         | If true, normalizes cumulative metrics without start times or with explicit reset points by subtracting subsequent points from the initial point. Since it caches starting points, it may result in increased memory usage. | `true`                                                   | no       |
| `endpoint`                             | `string`       | Endpoint where metric data is sent to.                                                                                                                                                                                      | `monitoring.googleapis.com:443`                          | no       |
| `experimental_wal`                     | `list(object)` | If provided, enables use of a write ahead log for time series requests. Each object must contain `directory`                                                                                                                | `[]`                                                     | no       |
| `experimental_wal` > `directory`       | `string`       | Path to local directory for the WAL file.                                                                                                                                                                                   | `./`                                                     | no       |
| `experimental_wal` > `max_backoff`     | `string`       | Max duration to retry requests on network errors (`UNAVAILABLE` or `DEADLINE_EXCEEDED`).                                                                                                                                    | `1h`                                                     | no       |
| `grpc_pool_size`                       | `number`       | Sets the size of the connection pool in the GCP client.                                                                                                                                                                     | `1`                                                      | no       |
| `instrumentation_library_labels`       | `bool`         | If true, set the `instrumentation_source` and `instrumentation_version` labels.                                                                                                                                             | `true`                                                   | no       |
| `known_domains`                        | `list(string)` | If a metric belongs to one of these domains it doesn't get a prefix.                                                                                                                                                        | `[googleapis.com, kubernetes.io, istio.io, knative.dev]` | no       |
| `prefix`                               | `string`       | The prefix to add to metrics.                                                                                                                                                                                               | `workload.googleapis.com`                                | no       |
| `resource_filters.prefix`              | `string`       | Match resource keys by prefix.                                                                                                                                                                                              | `""`                                                     | no       |
| `resource_filters.regex`               | `string`       | Match resource keys by regular expression.                                                                                                                                                                                  | `""`                                                     | no       |
| `resource_filters`                     | `list(object)` | If provided, resource attributes matching any filter is included in metric labels. Can be defined by `prefix`, `regex`, or `prefix` AND `regex`. Each object must contain one of `prefix` or `regex` or both.               | `[]`                                                     | no       |
| `service_resource_labels`              | `bool`         | If true, the exporter copies the OTel service.name, service.namespace, and service.instance.id resource attributes into the GCM timeseries metric labels.                                                                   | `true`                                                   | no       |
| `skip_create_descriptor`               | `bool`         | If set to true, don't send metric descriptors to GCM.                                                                                                                                                                       | `false`                                                  | no       |
| `sum_of_squared_deviation`             | `bool`         | If true, enables calculation of an estimated sum of squared deviation. It's an estimate, and isn't exact.                                                                                                                   | `false`                                                  | no       |
| `use_insecure`                         | `bool`         | If true, disables gRPC client transport security. Only has effect if Endpoint isn't `""`.                                                                                                                                   | `false`                                                  | no       |
```

Minor style edit. Sort rows alphabetically. Removed one redundant sentence about the default value.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+The `>` symbol indicates deeper levels of nesting. For example, `experimental_wal > directory` refers to a directory argument defined inside a experimental_wal block.
+
+### trace block
```
clayton-cornell, line 110:
```suggestion
### `trace`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| `attribute_mappings`                 | `list(object)` | AttributeMappings determines how to map from OpenTelemetry attribute keys to Google Cloud Trace keys.  By default, it changes http and service keys so that they appear more prominently in the UI. | `[]`                            | no       |
+| `attribute_mappings` > `key`         | `string`       | Key is the OpenTelemetry attribute key                                                                                                                                                              | `""`                            | no       |
+| `attribute_mappings` > `replacement` | `string`       | Replacement is the attribute sent to Google Cloud Trace                                                                                                                                             | `""`                            | no       |
```
clayton-cornell, line 119:
```suggestion
| Name                                 | Type           | Description                                                                                                                                                                      | Default                         | Required |
| ------------------------------------ | -------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------- | -------- |
| `attribute_mappings`                 | `list(object)` | Determines how to map from OpenTelemetry attribute keys to Google Cloud Trace keys. By default, it changes HTTP and service keys so that they appear more prominently in the UI. | `[]`                            | no       |
| `attribute_mappings` > `key`         | `string`       | The OpenTelemetry attribute key.                                                                                                                                                 | `""`                            | no       |
| `attribute_mappings` > `replacement` | `string`       | The attribute sent to Google Cloud Trace.                                                                                                                                        | `""`                            | no       |
| `endpoint`                           | `string`       | Endpoint where trace data is sent.                                                                                                                                               | `cloudtrace.googleapis.com:443` | no       |
| `grpc_pool_size`                     | `int`          | Sets the size of the connection pool in the GCP client. Defaults to a single connection.                                                                                         | `1`                             | no       |
| `use_insecure`                       | `bool`         | If true, disables gRPC client transport security. Only has effect if Endpoint isn't `""`.                                                                                        | `false`                         | no       |
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+### log block
```
clayton-cornell, line 121:
```suggestion
### `log`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| `resource_filters` > `regex`  | `string`       | Match resource keys by regex.                                                                                                                                                                                   | `""`                         | no       |
+| `service_resource_labels`     | `bool`         | If true, the exporter will copy OTel's service.name, service.namespace, and service.instance.id resource attributes into the GCM timeseries metric labels.                                                      | `true`                       | no       |
+| `error_reporting_type`        | `bool`         | ErrorReportingType enables automatically parsing error logs to a json payload containing the type value for GCP Error Reporting.                                                                                | `false`                      | no       |
```
clayton-cornell, line 136:
```suggestion
| Name                          | Type           | Description                                                                                                                                                                                                | Default                      | Required |
| ----------------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------- | -------- |
| `compression`                 | `string`       | Compression format for Log gRPC requests. Supported values: [`gzip`].                                                                                                                                      | `""` (no compression)        | no       |
| `default_log_name`            | `string`       | Defines a default name for log entries. If left unset, and a log entry doesn't have the `gcp.log_name` attribute set, the exporter returns an error processing that entry.                                 | `""`                         | no       |
| `endpoint`                    | `string`       | Endpoint where log data is sent.                                                                                                                                                                           | `logging.googleapis.com:443` | no       |
| `error_reporting_type`        | `bool`         | Enables automatically parsing error logs to a JSON payload containing the type value for GCP Error Reporting.                                                                                              | `false`                      | no       |
| `grpc_pool_size`              | `number`       | Sets the size of the connection pool in the GCP client.                                                                                                                                                    | `1`                          | no       |
| `resource_filters`            | `list(object)` | If provided, resource attributes matching any filter is included in log labels. Can be defined by `prefix`, `regex`, or `prefix` AND `regex`. Each object must contain one of `prefix` or `regex` or both. | `[]`                         | no       |
| `resource_filters` > `prefix` | `string`       | Match resource keys by prefix.                                                                                                                                                                             | `""`                         | no       |
| `resource_filters` > `regex`  | `string`       | Match resource keys by regular expression.                                                                                                                                                                 | `""`                         | no       |
| `service_resource_labels`     | `bool`         | If true, the exporter copies the OTel service.name, service.namespace, and service.instance.id resource attributes into the GCM timeseries metric labels.                                                  | `true`                       | no       |
| `use_insecure`                | `bool`         | If true, disables gRPC client transport security. Only has effect if Endpoint is not `""`.                                                                                                                 | `false`                      | no       |
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+### sending_queue block
```
clayton-cornell, line 138:
```suggestion
### `sending_queue`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+The `sending_queue` block configures an in-memory buffer of batches before data is sent to the HTTP server.
+
+{{< docs/shared lookup="reference/components/otelcol-queue-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+### debug_metrics block
```
clayton-cornell, line 144:
```suggestion
### `debug_metrics`
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+| Name    | Type               | Description                                                 |
+|---------|--------------------|-------------------------------------------------------------|
+| `input` | `otelcol.Consumer` | A value other components can use to send telemetry data to. |
+
+`input` accepts `otelcol.Consumer` data for any telemetry signal (metrics, logs, or traces).
+
+## Component health
+
+`otelcol.exporter.googlecloud` is only reported as unhealthy if given an invalid
+configuration.
```
clayton-cornell, line 161:
```suggestion
`otelcol.exporter.googlecloud` is only reported as unhealthy if given an invalid configuration.
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Debug information
+
+`otelcol.exporter.googlecloud` does not expose any component-specific debug
+information.
```
clayton-cornell, line 166:
```suggestion
`otelcol.exporter.googlecloud` doesn't expose any component-specific debug information.
```

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Example
+
+### Logging
```
clayton-cornell, line 170: Do we need this extra heading layer?  I'd suggest we remove it. If, in the future, we add other examples we can add the H3 headings.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,238 @@
...
+
+This example scrapes logs from local files through a receiver for conversion to OpenTelemetry format before finally sending them to Cloud Logging.
+
+Note that this configuration includes the recommended `memory_limiter` and `batch` plugins, which avoid high latency for reporting telemetry,
+and ensure that the collector itself will stay stable (not run out of memory) by dropping telemetry if needed.
```
clayton-cornell, line 175:
```suggestion
This configuration includes the recommended `memory_limiter` and `batch` plugins, which avoid high latency for reporting telemetry, and ensure that the collector itself will stay stable (not run out of memory) by dropping telemetry if needed.
```

clayton-cornell (commented):
Suggestions to align with the incoming style and layout changes.

One suggestion I didn't make here, and it can be done in a future PR, is to sort the block descriptions to the same order as they are presented in the Blocks table.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
motoki317: I wanted to be extra sure to note where I sourced the documentation from, because otherwise they could be seen as "original" work - but the words themselves are not actually my work, but of the original googlecloudexporter repo.

But I guess that's obvious enough - since this is a wrapper of the contrib library and the original doc link is already stated - and quite verbose for the readers. I'll remove this for now, but please let me know if you have other ideas.

clayton-cornell: I think we are probably OK here with leaving it out since it's just the arg descriptions... and you've made it clear to refer to the original GCP docs.

We have other components that are wrappers with similar situations (common args and common descriptions).

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,227 @@
...
+| `sum_of_squared_deviation`             | `bool`         | If true, enables calculation of an estimated sum of squared deviation. It's an estimate, and isn't exact.                                                                                                                   | `false`                                                  | no       |
+| `use_insecure`                         | `bool`         | If true, disables gRPC client transport security. Only has effect if Endpoint isn't `""`.                                                                                                                                   | `false`                                                  | no       |
+
+The > symbol indicates deeper levels of nesting.
+For example, `experimental_wal` > `directory` refers to a `directory` argument defined inside a `experimental_wal` block.
```
dehaansa, line 123: I believe we only use this kind of detail when describing block nesting, not when showing arguments of a block. The `experimental_wal` block should have its own section in the docs, should be shown in the `blocks` section, and its attributes should not be listed under the `metric` block.

CC  @clayton-cornell correct me if I'm wrong here.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,227 @@
...
+| `grpc_pool_size`              | `number`       | Sets the size of the connection pool in the GCP client.                                                                                                                                                    | `1`                          | no       |
+| `resource_filters`            | `list(object)` | If provided, resource attributes matching any filter is included in log labels. Can be defined by `prefix`, `regex`, or `prefix` AND `regex`. Each object must contain one of `prefix` or `regex` or both. | `[]`                         | no       |
+| `resource_filters` > `prefix` | `string`       | Match resource keys by prefix.                                                                                                                                                                             | `""`                         | no       |
+| `resource_filters` > `regex`  | `string`       | Match resource keys by regular expression.                                                                                                                                                                 | `""`                         | no       |
```
dehaansa, line 92: Unless the type is a block, we typically do not list out its attributes, and the top level type is a `list(map(string))`. I don't love this convention and think something like this is valuable, but am looking for consistency. 

CC @clayton-cornell let me know if this is a new convention we're moving toward.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Main (unreleased)
 - Add string concatenation for secrets type (@ravishankar15)
 - Add support for environment variables to OpenTelemetry Collector config. (@jharvey10)
 - Replace graph in Alloy UI with a new version that supports modules and data flow visualization. (@wildum)
+- Add `otelcol.exporter.googlecloud` component to export metrics, traces, and logs to Google Cloud. (@motoki317)
```
dehaansa, line 20:
```suggestion
- Add `otelcol.exporter.googlecloud` community component to export metrics, traces, and logs to Google Cloud. (@motoki317)
```

dehaansa: Want to be sure that users who go to use this know what they need to do to enable it.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
clayton-cornell: If the information is valuable to the users... we could/should definitely include it. The documentation is evolving all the time so new/better ways to explain things should always be considered.  If we do add it here I wonder how many other areas could use a refresh to add in this extra level of information?  I'm not sure how to identify this other than a search for `list(map(string))` and decide from there.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
clayton-cornell: This (nested info) looks like the same situation here (specifically lines 108 and 109)  as in the previous comment (lines 91 and 92) which lines up with why the nesting explanation is included here.

So.. do we include the extended attribute descriptions? or drop back to the current top-level only?

And the `experimental_wal` needing its own section... I didn't know it was its own block so missed that in the review.  If its a standalone block, then yes it definitely needs to be split into its own block description section.

motoki317: Thanks for pointing this out, I moved `metric` > `experimental_wal` block to its own section.
I wasn't 100% sure if I should make its own section for this little nesting block, but I'll go with the docs standards for now.

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
motoki317: I wasn't quite sure how to go about documenting this - it definitely is valuable for users rather than entirely omitting the `prefix` and `regex` documentation, but I wasn't sure how to format this.

internal/component/otelcol/exporter/googlecloud/googlecloud.go
```diff
@@ -0,0 +1,109 @@
...
+	result.QueueSettings = *args.Queue.Convert()
+
+	result.ProjectID = args.Project
+	result.DestinationProjectQuota = args.DestinationProjectQuota
+	result.UserAgent = args.UserAgent
+
+	result.ImpersonateConfig = args.Impersonate.Convert(result.ImpersonateConfig)
+	result.MetricConfig = args.Metric.Convert(result.MetricConfig)
+	result.TraceConfig = args.Trace.Convert(result.TraceConfig)
+	result.LogConfig = args.Log.Convert(result.LogConfig)
```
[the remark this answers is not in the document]
motoki317: Originally I followed the same pattern, but I realized that doing that will not work because the googlecloudexporter populates struct fields with unexported functions with `collector.DefaultConfig()` call. So I needed to assign to each fields inside a specific field of the entire config struct - which I intended to document the reason for doing this in this comment:
https://github.com/grafana/alloy/pull/3140/files/2743b0d12bd4b4db2d35017f504e1c41652c0579#diff-ba586afa4f1c004c8e6916d315cec77f855465792860c139b91a4d00813d5750R68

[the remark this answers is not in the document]
wildum: aah that makes sense, thanks for the explanation

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloud.md
```diff
@@ -0,0 +1,235 @@
...
+| `use_insecure`                         | `bool`         | If true, disables gRPC client transport security. Only has effect if Endpoint isn't `""`.                                                                                                                                   | `false`                                                  | no       |
+
+### `experimental_wal`
+
+The following arguments are supported:
+
+| Name          | Type     | Description                                                                              | Default | Required |
+|---------------|----------|------------------------------------------------------------------------------------------|---------|----------|
+| `directory`   | `string` | Path to local directory for the WAL file.                                                | `./`    | yes      |
```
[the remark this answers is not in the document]
motoki317: Perhaps it would be a good idea, but at the same time this block is marked as "experimental" and specifying config value seems to enable this "experimental" WAL feature.
So since this is an optional feature anyway, my suggestion is: Leave this as is and let users decide if they want to enable the experimental wal feature, and let them decide where to store the directory.

> if you have two googlecloud exporters, they will point to the same directory?

By default this experimental_wal feature is disabled, so if the user decides to use 2 of googlecloud components and specify the same directory for WAL for some reason, I guess they will, but that's unlikely.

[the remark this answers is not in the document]
wildum: I was thinking about just getting rid of the directory attribute and keeping the block with just the attribute `max_backoff`. This way it's still disabled by default but if they want to enable it, it will be automatically stored in a proper location like the other components. As a user, I would prefer that all the data files are automatically structured and that I don't have to bother configuring the paths myself. 
The only value I see to let the users choose the directory is if they are migrating from the Otel collector to Alloy and they want to keep their WAL. They could still workaround it by moving their existing WAL to the proper location though.
But I'm fine leaving it like this if you want to match with upstream.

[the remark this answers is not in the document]
motoki317: > As a user, I would prefer that all the data files are automatically structured and that I don't have to bother configuring the paths myself.

Ah I see. I agree in that I would rather like an opinionated configuration values that work out of the box rather than configuring everything by myself.
But in this case, I think it'd be better to keep this a simple wrapper of the upstream, both from a maintenance and user perspective. Keeping it consistent with the upstream should make this component's behavior more predictable, helping users understand it more easily. If this were a more "core" configuration value then maybe we could provide a more sensible default. However, the component works perfectly fine without this `experimental_wal` config, and it is named experimental after all.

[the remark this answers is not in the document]
wildum: Fine by me, could you just add a comment in the code above the directory config saying that we considered using the storage path but decided to keep it as close as possible with upstream please?

wildum:
we just merged an update of our otel dependencies from v0.119 to v0.122 so you will need to rebase and update the component to the same version. Sorry for the trouble, we only do this every 6 weeks

motoki317:
@wildum 
> Could you please add a config converter for the component?

I've added a converter for googlecloudexporter.
Could you take a look at this again? Thanks!

wildum (approved):
Awesome work, thanks a lot!

PR #3144, opened 2025-03-31
Otel 122

wildum:
#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.processor.resourcedetection.md
```diff
@@ -824,6 +827,26 @@ The `resource_attributes` block supports the following blocks:
 | Block                            | Description                                                                             | Required |
 | -------------------------------- | --------------------------------------------------------------------------------------- | -------- |
 | [k8s.cluster.name][res-attr-cfg] | Toggles the `k8s.cluster.name` resource attribute. Sets `enabled` to `true` by default. | no       |
+| [k8s.cluster.uid][res-attr-cfg]  | Toggles the `k8s.cluster.uid` resource attribute. Sets `enabled` to `true` by default.  | no       |
+### dynatrace
```
clayton-cornell, line 831:
```suggestion
| [k8s.cluster.uid][res-attr-cfg]  | Toggles the `k8s.cluster.uid` resource attribute. Sets `enabled` to `true` by default.  | no       |

### dynatrace
```

docs/sources/reference/components/otelcol/otelcol.processor.resourcedetection.md
```diff
@@ -824,6 +827,26 @@ The `resource_attributes` block supports the following blocks:
...
+[resource_attributes](#dynatrace--resource_attributes) | Configures which resource attributes to add. | no
+
+#### dynatrace > resource_attributes
+
+The `resource_attributes` block supports the following blocks:
+
+Block                                             | Description                                                                                                   | Required
+--------------------------------------------------|---------------------------------------------------------------------------------------------------------------|---------
+[host.name][res-attr-cfg]                        | Toggles the `host.name` resource attribute. <br> Sets `enabled` to `true` by default.                        | no
+[dt.entity.host][res-attr-cfg]                   | Toggles the `dt.entity.host` resource attribute. <br> Sets `enabled` to `true` by default.                   | no
```
clayton-cornell, line 849:
```suggestion
[host.name][res-attr-cfg]                        | Toggles the `host.name` resource attribute.  Sets `enabled` to `true` by default.                        | no
[dt.entity.host][res-attr-cfg]                   | Toggles the `dt.entity.host` resource attribute. Sets `enabled` to `true` by default.                   | no
```

We are not supposed to use `<br>` elements in tables. If we need paragraphs/newlines, then we should use `<p>`.  Looking at these rows, I don't think we really need a linebreak.

clayton-cornell (commented):
Docs are OK for this PR. There are a few minor style issues that I'll catch in an upcoming PR (not worth the effort within this PR).

docs/sources/reference/components/otelcol/otelcol.processor.transform.md
```diff
@@ -186,6 +187,47 @@ The supported values for `context` are:
 
 Refer to [OTTL Context][] for more information about how to use contexts.
 
+### statements block
```
dehaansa, line 190: Thanks for pulling this in, I skipped it in the last update due to the other complexities but it's a really nice improvement to the processor.

CHANGELOG.md
```diff
@@ -88,6 +88,20 @@ Main (unreleased)
   - Added a new `http_headers` argument to many `discovery` and `prometheus` components.
   - Added a new `scrape_failure_log_file` argument to `prometheus.scrape`.
 
+- Upgrading to OpenTelemetry Collector v0.122
+  - `otelcol.processor.batch` telemetry is no longer emitted at "basic" level. If you want to keep the batch processor telemetry, you can switch to "normal" level.
```
dehaansa, line 92: We should add that this is no longer configurable in debug_metrics per component.

docs/sources/shared/reference/components/otelcol-debug-metrics-block.md
```diff
@@ -11,7 +11,6 @@ The following arguments are supported:
 Name                               | Type      | Description                                          | Default | Required
 -----------------------------------|-----------|------------------------------------------------------|---------|---------
 `disable_high_cardinality_metrics` | `boolean` | Whether to disable certain high cardinality metrics. | `true`  | no
-`level` | `string` |  Controls the level of detail for metrics emitted by the wrapped collector. | `"detailed"`  | no
```
dehaansa, line 14: Now that this is no longer configurable per component, how is it configured across all components?

wildum: It's a bit of a rabbit hole and I'm not sure I understood the full picture but what I got so far is that the new approach would be to enable instruments on the Meterprovider (https://github.com/open-telemetry/opentelemetry-collector/issues/11061). This would need some work on SDK side (https://github.com/open-telemetry/opentelemetry-collector/issues/11754) but the collector couldn't wait since they want this for 1.0 so they just enable the instrument in the service based on the overall level set by the telemetry config (https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/service.go#L522). If this functionality is added to the SDK, then we should be able to keep specifying the metrics level per component if we want.
The explanation might be completely inaccurate, I don't understand how it works with contrib. I feel like this metrics level is actually not so useful? I did not have much time to dig deeper

CHANGELOG.md
wildum: thanks good catch, there is actually no point to mention this about the batch processor since the metric level is not configurable anymore. I updated the changelog entry

docs/sources/reference/components/otelcol/otelcol.processor.transform.md
wildum: (credits go to @ptodev for this change)

docs/sources/shared/reference/components/otelcol-debug-metrics-block.md
thampiotr: Our deployments may be using this and will suffer from breaking change.

CHANGELOG.md
```diff
@@ -88,6 +88,20 @@ Main (unreleased)
   - Added a new `http_headers` argument to many `discovery` and `prometheus` components.
   - Added a new `scrape_failure_log_file` argument to `prometheus.scrape`.
 
+- Upgrading to OpenTelemetry Collector v0.122
+  - The telemetry level of Otel components is no longer configurable. The `level` argument in the `debug_metrics` block is kept to avoid breaking changes but it is not used anymore.
```
thampiotr, line 92: Call this (and all other breaking changes) out in the breaking changes section? IMO people need to see this so we should make some noise about it.

docs/sources/reference/components/otelcol/otelcol.exporter.splunkhec.md
```diff
@@ -120,8 +120,9 @@ Name                       | Type                  | Description
...
 `enabled`                  | `bool`                | Whether to not enqueue batches before sending to the consumerSender.                            |   `false`                     | no
 `flush_timeout`             | `time.Duration`       | The time after which a batch will be sent regardless of its size.    | `200ms`  | no
-`min_size_items`   | `uint`   |  The number of items at which the batch is sent regardless of the timeout.          |  `8192`              | no
-`max_size_items`   | `uint`   | Maximum number of batch items, if the batch exceeds this value, it will be broken up into smaller batches. Must be greater than or equal to min_size_items. Setting this value to zero disables the maximum size limit.          |  `0`           |  no
+`min_size`   | `uint`   | The minimum size of a batch.          |  `8192`              | no
+`max_size`   | `uint`   | The maximum size of a batch. If the batch exceeds this value, it will be broken up into smaller batches. Must be greater than or equal to min_size. Setting this value to zero disables the maximum size limit.          |  `0`           |  no
```
thampiotr, line 124: These two are breaking changes, do we call them out?

internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/internal/source/remotesource/remote_strategy_cache_test.go
```diff
@@ -247,11 +247,10 @@ func Test_serviceStrategyCache_Concurrency(t *testing.T) {
 	wg := sync.WaitGroup{}
 	wg.Add(numThreads)
 	for i := 0; i < numThreads; i++ {
-		ii := i
 		go func() {
 			for j := 0; j < numIterationsPerThread; j++ {
 				for _, svcName := range []string{
-					fmt.Sprintf("thread-specific-service-%d", ii),
+					fmt.Sprintf("thread-specific-service-%d", i),
```
thampiotr, line 253: This looks like re-introducing a bug?

thampiotr (commented):
Can we deploy this to a few clusters to verify? I understand this is the only way we can get this done right now and you've done great getting this done! But this is huge and very easy to miss something. So for this PR let's try to test it a bit more. And for the longer term let's continue discussion about what to do with these updates.

docs/sources/reference/components/otelcol/otelcol.exporter.splunkhec.md
wildum: thanks, that was missed

PR #3145, opened 2025-03-31
docs: Add note for using proxy_from_environment

dehaansa:
#### PR Description
A user was confused why Alloy was not utilizing their proxy environment variables, and the Alloy docs for environment variables do not point out the requirement to set `proxy_from_environment` for prometheus-styled components.

### DIFF
--- a/docs/sources/reference/cli/environment-variables.md
+++ b/docs/sources/reference/cli/environment-variables.md
@@ -40,10 +40,10 @@ Argument               | Description                                            
 ## HTTP_PROXY, HTTPS_PROXY, NO_PROXY
 
 {{< admonition type="note" >}}
-For many {{< param "PRODUCT_NAME" >}} components, there is an attribute named `proxy_from_environment` that must be set to `true` for the component's
-HTTP client to use the proxy related environment variables. For example, in the `prometheus.remote_write` component this attribute is found within the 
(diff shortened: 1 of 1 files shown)


docs/sources/reference/cli/environment-variables.md
```diff
@@ -39,6 +39,13 @@ Argument               | Description
 
 ## HTTP_PROXY, HTTPS_PROXY, NO_PROXY
 
+{{< admonition type="note" >}}
+For many {{< param "PRODUCT_NAME" >}} components, there is an attribute named `proxy_from_environment` that must be set to `true` for the component's
+HTTP client to use the proxy related environment variables. For example, in the `prometheus.remote_write` component this attribute is found within the 
+`endpoint` block. If {{< param "PRODUCT_NAME" >}} does not appear to be respecting your proxy configuration, please ensure the component is
+configured properly to utilize proxy environment variables.
```
clayton-cornell, line 46:
```suggestion
For many {{< param "PRODUCT_NAME" >}} components, there is an attribute named `proxy_from_environment` that must be set to `true` for the component's HTTP client to use the proxy-related environment variables.
For example, in the `prometheus.remote_write` component, this attribute is found within the `endpoint` block.
If {{< param "PRODUCT_NAME" >}} does not appear to be respecting your proxy configuration, make sure you configure the component to use proxy environment variables.
```

clayton-cornell:
Removed the backport to v1.5.  I set one too many versions back.

PR #3168, opened 2025-04-01
feat: http server authentication 

### LINKED ISSUE #637: Add authentication to Alloy UI
### Request

Would like to have authentication set for the Alloy UI.
Even a basic AUTH similar to Prometheus UI might be a good start.

### Use case

Since the UI exposes all the details around pod ips and other details it might be good to protect these UI and also any API access through authentication.

kalleep:
#### PR Description
Add support for basic authentication for alloy http server. 

I implement the functionallity suggested in https://github.com/grafana/alloy/issues/637#issuecomment-2602834802.

To configure auth there is a new block under the http block and looks like the following
```
http {
	auth {
		basic {
			username = sys.env("BASIC_AUTH_USERNAME")
			password = sys.env("BASIC_AUTH_PASSWORD")
		}

		filter {
			paths         = []   // Default value
			auth_if_match = true // Default value
		}
	}
}
```

To configure basic authentication both username and password need to be configured. In addition to this we have the filter block. Filter is used to determine what api paths should be protected by authentication using prefix matching. If `auth_if_match` is set to true (default) we will perform authentication on all matching paths. Setting `auth_if_match` to false will reverse it, meaning we will not perform authentication listed paths.

Fixes #637

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/config-blocks/http.md
+++ b/docs/sources/reference/config-blocks/http.md
@@ -208,6 +208,40 @@ The filter block is used to configure which API paths should be protected by aut
 
 | Name                  | Type           | Description                                                                                                            | Default | Required |
 | ----------------------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
-| `path`                              | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
+| `paths`                             | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
 | `authenticate_matching_paths`       | `bool`         | If true, authentication is required for all matching paths. If false, authentication is excluded for these paths.      | `true`  | no       |
 
+
+Example of enforcing authentication on `/metrics` and every enpoint that has `/v1` as prefix:
+```alloy
+http {
+  auth {
+    basic {
+      username = sys.env("BASIC_AUTH_USERNAME")
+      password = sys.env("BASIC_AUTH_PASSWORD")
+    }
+
+    filter {
+      paths                       = ["/metrics", "/v1"]
+      authenticate_matching_paths = true
+    }
+  }
+}
+```
+
+Example enforcing authentication on all endpoints except `/metrics`:
+```alloy
+http {
+  auth {
+    basic {
+      username = sys.env("BASIC_AUTH_USERNAME")
+      password = sys.env("BASIC_AUTH_PASSWORD")
+    }
+
+    filter {
+      paths                       = ["/metrics"]
+      authenticate_matching_paths = false
+    }
+  }
+}
+```
--- a/docs/sources/reference/config-blocks/http.md
+++ b/docs/sources/reference/config-blocks/http.md
@@ -26,8 +26,8 @@ http {
     }
 
     filter {
-      paths         = ["/"]
-      auth_if_match = true
+      paths                       = ["/"]
+      authenticate_matching_paths = true
     }
   }
 }
@@ -207,7 +207,7 @@ The basic block enables basic HTTP authentication by requiring both a username a
 The filter block is used to configure which API paths should be protected by authentication. It allows you to specify a list of paths, using prefix matching, that will require authentication.
 
 | Name                  | Type           | Description                                                                                                            | Default | Required |
-| --------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
-| `path`                | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
-| `auth_if_match`       | `bool`         | If true, authentication is required for all matching paths. If false, authentication is excluded for these paths.      | `true`  | no       |
+| ----------------------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `path`                              | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
+| `authenticate_matching_paths`       | `bool`         | If true, authentication is required for all matching paths. If false, authentication is excluded for these paths.      | `true`  | no       |
 
--- a/docs/sources/reference/config-blocks/http.md
+++ b/docs/sources/reference/config-blocks/http.md
@@ -18,6 +18,18 @@ http {
     cert_file = sys.env("TLS_CERT_FILE_PATH")
     key_file  = sys.env("TLS_KEY_FILE_PATH")
   }
+
+  http {
+    basic {
+      username = sys.env("BASIC_AUTH_USERNAME")
+      password = sys.env("BASIC_AUTH_PASSWORD")
+    }
+
+    filter {
+      paths         = ["/"]
+      auth_if_match = true
+    }
+  }
 }
 ```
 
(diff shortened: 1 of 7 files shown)


internal/service/http/auth.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/grafana/alloy/syntax/alloytypes"
+)
+
+type AuthArguments struct {
+	Basic *BasicAuthArguments `alloy:"basic,block,optional"`
+	// Filter is used to apply authentication to matching api endpoints.
+	Filter []string `alloy:"filter,attr,optional"`
```
kalleep, line 16: In https://github.com/grafana/alloy/issues/637#issuecomment-2602834802 I noticed that this kind of syntax was used for filter: `['/debug/pprof', '/component/*']`

But I figured it was easier to implement as prefix matching instead of having some kind of wildcard symbol

internal/service/http/http.go
```diff
@@ -188,6 +195,21 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		otelmux.WithTracerProvider(s.tracer),
 	))
 
+	// apply authenticator middleware, if non is configured allowAuthenticator is used and no authentication is required
```
ptodev, line 198:
```suggestion
	// Apply authenticator middleware.
	// If none is configured allowAuthenticator is used and no authentication is required.
```

internal/service/http/auth.go
ptodev: Yes, this is probably sufficient and will be easier for most people to configure than regex/wildcards.

ptodev, line 16: Some folks might want to enable auth for everything but a select number of endpoints. For example everything other than `/metrics` may need auth. This is a more secure approach too. WDYT about making `filter` a block with `include` and `exclude` arguments? Then if `include = "/"` and `exclude="/metrics"`, everything would require auth except for metrics?

kalleep: Either that or we do the reverse, instead of using filter to check if the endpoint should be protected by auth we check if the endpoint should be excluded. Then we can have sane defaults.

So `excludes = []` will protect every endpoint

ptodev: I still think it's good to have both an `include` and an `exclude` option. It'd be useful for folks who want to do either one of those for all but a select number of endpoints.

For example, you might want to lock down only endpoints that change Alloy's state like this:
```
include = [`/reload`]
exclude = [`/`]
```

I'm just not sure how we'd decide which one takes precedence. Maybe it's easier if we just use a regex and have only one filter list?

kalleep: Discussed in slack:

1. Having only one field (either `include` or `exclude`) is simple but makes it hard to express certain configs
2. Having both of them is possible and we could follow what is done in https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.windows/#network. But with this you could also have nonsensical config like this 
```
include = [`/reload`]
exclude = [`/metrics`]
```
3. Having one field called paths and another field to toggle the behavior between `include` and `exclude`, similar to what is done here https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.tail_sampling/#numeric_attribute-block.
```
filter {
  paths = [...] // [] is the default
  auth_if_match = true // true is the default
}
```

We both agreed that doing nr 3 is the simplest one for configuration

docs/sources/reference/config-blocks/http.md
```diff
@@ -175,3 +190,24 @@ The `client` block is used to check the certificate presented to the server.
...
+| `password`            | `string`       | The password to use for basic authentication.                     |   N/A   | yes      |
+
+
+# filter block
+The filter block is used to configure which API paths should be protected by authentication. It allows you to specify a list of paths, using prefix matching, that will require authentication.
+
+| Name                  | Type           | Description                                                                                                            | Default | Required |
+| --------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `path`                | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
+| `auth_if_match`       | `bool`         | If true, authentication is required for all matching paths. If false, authentication is excluded for these paths.      | `true`  | no       |
```
dehaansa, line 212: IMO this isn't a very clear name, what do you think of `authenticate_matching`?

I also considered `requires_authentication`/`authentication_required`, but they felt a little less direct.

kalleep: Yes we discussed this name a bit but have a hard time to come up with a good name and this was the most explicit one. 

The idea with this flag is to invert the behavior

Require auth on only `/metrics`:
```
filter {
  paht = ["/metrics"]
  auth_if_match = true
}
```

Require auth on _all_ endpoints except `/metrics`:
```
filter {
  paht = ["/metrics"]
  auth_if_match = false
}
```

ptodev: `authenticate_matching` sounds better to me too! WDYT about `authenticate_paths`? I think it might be even better since it's clear that we authenticate the paths above. Or `authenticate_matching_paths`?

kalleep: Yeah I like `authenticate_matching_paths` too.

docs/sources/reference/config-blocks/http.md
```diff
@@ -175,3 +190,24 @@ The `client` block is used to check the certificate presented to the server.
...
+### auth block
+The auth block configures server authentication for the http block. This can be used to enable basic authentication and to set authentication filters for specified API paths.
+
+### basic block
+The basic block enables basic HTTP authentication by requiring both a username and password for access.
+
+| Name                  | Type           | Description                                                       | Default | Required |
+| --------------------- | -------------- | ----------------------------------------------------------------- | ------- | -------- |
+| `username`            | `string`       | The username to use for basic authentication.                     |   N/A   | yes      |
+| `password`            | `string`       | The password to use for basic authentication.                     |   N/A   | yes      |
```
ptodev, line 203:
```suggestion
| `username`            | `string`       | The username to use for basic authentication.                     |     | yes      |
| `password`            | `string`       | The password to use for basic authentication.                     |     | yes      |
```

If there's no default we'd just leave it empty.

docs/sources/reference/config-blocks/http.md
```diff
@@ -175,3 +190,24 @@ The `client` block is used to check the certificate presented to the server.
...
+
+
+# filter block
+The filter block is used to configure which API paths should be protected by authentication. It allows you to specify a list of paths, using prefix matching, that will require authentication.
+
+| Name                  | Type           | Description                                                                                                            | Default | Required |
+| ----------------------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `path`                              | `list(string)` | List of API paths to be protected by authentication. The paths are matched using prefix matching.                      | `[]`    | no       |
+| `authenticate_matching_paths`       | `bool`         | If true, authentication is required for all matching paths. If false, authentication is excluded for these paths.      | `true`  | no       |
```
ptodev, line 212: It'd be nice to have more examples of different ways people can use those two arguments.

docs/sources/reference/config-blocks/http.md
dehaansa: Sounds good to me!

ptodev (approved):

clayton-cornell (commented):
The doc input is OK as-is.  There are some style changes that I'll make in a future PR (there's no point making them in this PR when the style changes will need to be done on the whole doc file).

clayton-cornell (approved):

PR #3179, opened 2025-04-02
azure: Ensure we will always set a top for subscription scope if it is not set

### LINKED ISSUE #2735: prometheus.exporter.azure returns only 10 resources
### What's wrong?

prometheus.exporter.azure returns a maximum of 10 resources when it uses Azure Metrics Batch API. It looks like it uses a default value for **top** parameter, which 10: https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch?view=rest-monitor-2023-10-01&tabs=HTTP

here are several examples of different resources:

Image

Image

here is an example of how many disks I have:

Image

and postgres servers:

Image

I think it will be good to have this parameter configurable on the exporter side

### Steps to reproduce

configure prometheus.exporter.azure for any resource type with more than 10 resources. Observe the number of metrics returned

### System information

_No response_

### Software version

Alloy v1.6.1

### Configuration

```text
prometheus.exporter.azure "disks_metrics" {
  subscriptions    = [
      "11111111111111111",
  ]
  resource_type    = "microsoft.compute/disks"
  regions          = [
      "eastus",
	]
	metrics          = [
    "Composite Disk Read Bytes/sec",
    "Composite Disk Read Operations/sec",
    "Composite Disk Write Bytes/sec",
    "Composite Disk Write Operations/sec",
    "DiskPaidBurstIOPS",
	]
	metric_aggregations = [
        "average",
	]
  timespan                    = "PT5M"
}

prometheus.exporter.azure "postgres_metrics" {
  subscriptions    = [
      "11111111111111111",
  ]
  resource_type    = "microsoft.dbforPostgreSQL/flexibleServers"
  regions          = [
      "eastus",
	]
	metrics          = [
    "cpu_percent",
    "cpu_credits_remaining",
    "cpu_credits_consumed",
    "read_iops",
    "write_iops",
    "disk_queue_depth",
    "memory_percent",
    "storage_free",
    "storage_percent",
    "txlogs_storage_used",
    "backup_storage_used",
    "active_connections",
    "max_connections",
    "physical_replication_delay_in_seconds",
    "physical_replication_delay_in_bytes",
    "logical_replication_delay_in_bytes",
    "disk_iops_consumed_percentage",
    "disk_bandwidth_consumed_percentage",
    "client_connections_active",
    "client_connections_waiting",
    "server_connections_active",
    "server_connections_idle",
    "total_pooled_connections",
    "num_pools",
    "is_db_alive",
	]
	metric_aggregations = [
        "average",
        "maximum",
	]
  timespan                    = "PT5M"
}
```

### Logs

```text

```

kgeckhart:
#### PR Description
Azure's metrics APIs require a [top parameter be used with filters](https://learn.microsoft.com/en-us/rest/api/monitor/metrics/list-at-subscription-scope?view=rest-monitor-2023-10-01&tabs=HTTP#uri-parameters)
image

We were only explicitly setting a filter when dimensions are used. When using the exporter with the `regions` config it uses the subscription scope API which requires a filter that the embedded exporter sets. Since we weren't setting top in this case the default of 10 was always being used causing missing data.

Fixes #2735

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,8 @@ Main (unreleased)
 
 - Allow kafka exporter to attempt to connect even if TLS enabled but cert & key are not specified (@dehaansa)
 
+- Fixed bug where all resources were not being collected from `prometheus.exporter.azure` when using `regions` (@kgeckhart)
+
 ### Other changes
 
 - Upgrading to Prometheus v2.55.1. (@ptodev)
--- a/internal/static/integrations/azure_exporter/config.go
+++ b/internal/static/integrations/azure_exporter/config.go
@@ -213,6 +213,12 @@ func (c *Config) ToScrapeSettings() (*metrics.RequestMetricSettings, error) {
 		settings.MetricTop = to.Ptr[int32](100_000_000)
 		settings.MetricOrderBy = "" // Order is only relevant if top won't return all the results our high value should prevent this
 	}
+
+	// Using regions is at subscription scope which adds a default filter that respects top
+	if len(settings.Regions) > 0 && settings.MetricTop == nil {
+		settings.MetricTop = to.Ptr[int32](100_000_000)
+	}
+
 	return &settings, nil
 }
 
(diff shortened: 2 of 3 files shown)


wildum (approved):
thanks!

PR #3184, opened 2025-04-02
disable labelstore by default

mattdurham:
Long time coming, labelstore exists to only support remote_write, but if your not running it then no need to use the memory and cpu. This resolves that problem.


internal/service/labelstore/service.go
```diff
@@ -122,8 +127,16 @@ func (s *service) Data() any {
 	return s
 }
 
+func (s *service) Enable() {
+	s.enabled.Store(true)
+}
+
 // GetOrAddLink is called by a remote_write endpoint component to add mapping and get back the global id.
 func (s *service) GetOrAddLink(componentID string, localRefID uint64, lbls labels.Labels) uint64 {
+	if !s.enabled.Load() {
```
mattdurham, line 136: Technically its a slightly different behavior. If you started without a remote write, and just a prometheus write queue. Then changed the config to remove the write queue and add a remote write the ref id's would be a bit odd. That being said I doubt that will ever happen. If it does then it is unlikely to be a problem since the hash is likely to be a much higher number than the incrementing ref id will get and the scraper only uses it to cache data quicker. Additionally since remote_write and prometheus.write.queue cant coexist so we are never in a mixed environment.

mattdurham: I could probably add globalref id, but then would need to place it in the lock or make it an atomic which doesnt feel great given how incredibly unlikely the scenario is.

internal/component/prometheus/remotewrite/remote_write.go
```diff
@@ -94,6 +94,8 @@ func New(o component.Options, c Arguments) (*Component, error) {
 		return nil, err
 	}
 	ls := service.(labelstore.LabelStore)
+	// Enable the labelstore since it only exists to serve remote_write.
+	ls.Enable()
```
dehaansa, line 98: If a config reload removes all remote_write components should we disable? I don't think this is a super important use case to consider, but wanted to discuss.

internal/service/labelstore/service.go
```diff
@@ -155,6 +168,10 @@ func (s *service) GetOrAddLink(componentID string, localRefID uint64, lbls label
 
 // GetOrAddGlobalRefID is used to create a global refid for a labelset
 func (s *service) GetOrAddGlobalRefID(l labels.Labels) uint64 {
+	if !s.enabled.Load() {
+		return l.Hash()
```
dehaansa, line 172: Should we guard against bad input like line 161?

internal/component/prometheus/remotewrite/remote_write.go
mattdurham: We would have to somehow track the usage which feels  not worth it. Labelstore would need to track calls to enable and disable but I feel there could be some sharp edges there. IMO no one will be flip flopping without restarting.

internal/service/labelstore/service.go
mattdurham: Probably doesnt hurt, will add.

internal/component/prometheus/remotewrite/remote_write.go
dehaansa: I wasn't considering this at the time, but fleet management is probably a use case where a remote_write could disappear and reappear without restart relatively easily.

mattdurham: If it started with remote_write it’s fine it is already enabled. The specific case is going from no remote writes to one. Might be better to return 0 in all cases since both the scrape and remote write will see that as new values. Would need to double check a few things though

internal/service/labelstore/service.go
mattdurham: On thinking about having the value be consistent is more important so Hash is the best approach.

internal/service/labelstore/service.go
```diff
@@ -26,7 +27,9 @@ type service struct {
 	totalIDs            *prometheus.Desc
 	idsInRemoteWrapping *prometheus.Desc
 	lastStaleCheck      prometheus.Gauge
+	enabled             *atomic.Bool
```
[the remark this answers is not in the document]
ptodev: A bit like a shared pointer in C++.

internal/component/prometheus/remotewrite/remote_write.go
thampiotr: I would be in favour of tracking usage. The call to .Enable() could be replaced with .Register() and increment a number of users of the labelstore. When number of users is/reaches zero, we keep labelstore emptied and disabled. When it's > 0, we keep it working. Doesn't seem like too complex and would avoid edge cases that can be hard to debug.

PR #3189, opened 2025-04-03
Fix tests that are broken at main or very flaky

thampiotr:


internal/component/loki/source/file/file_test.go
```diff
@@ -359,9 +360,11 @@ func TestDeleteRecreateFile(t *testing.T) {
 	require.NoError(t, f.Close())
 	require.NoError(t, os.Remove(f.Name()))
 
-	// Create a file with the same name
-	f, err = os.Create(filename)
-	require.NoError(t, err)
+	// Create a file with the same name. Use eventually because of Windows FS can deny access if this test runs too fast.
```
ptodev, line 363: Are you sure it's because it runs too fast? That would be really odd - surely there's a way to create files quickly on Windows.

In the past I've seen tests fail on Windows due to the files not being closed properly. Could that be the case here?

internal/component/remote/vault/vault_test.go
```diff
@@ -190,7 +192,10 @@ func getTestVaultServer(t *testing.T) *vaultapi.Client {
 	require.NoError(t, err)
 
 	t.Cleanup(func() {
-		require.NoError(t, container.Terminate(ctx))
+		terminateErr := container.Terminate(ctx)
+		if !strings.Contains(terminateErr.Error(), "context canceled") { // avoid race condition on test context cancelled
```
ptodev, line 196: Why is this necessary? Is it because the container terminates async? Do we need to wait after the Terminate function? Even if it's async I don't see how it can cause an issue, since the context should only cancel at the end of the test, after Terminate has already completed.

ptodev (approved):
Rubber stamping, but I wrote a few comments because I'm not sure if that's the best way to fix some of the tests.

internal/component/remote/vault/vault_test.go
kalleep: functions registered with `t.Cleanup` are called in registration order. Context is created from [TestContext](https://github.com/grafana/alloy/blob/main/internal/runtime/componenttest/context.go#L9). This context will be canceled when the test if finished and it happens before this cleanup hook is called. We chould just pass a next context here instead of reusing the existing one

thampiotr: Because it is failing here: https://drone.grafana.net/grafana/alloy/7212/1/2 
You may need to ask the author of the test, right now I would like to fix tests in main.

kalleep: Never mind not sure this is true

internal/component/loki/source/file/file_test.go
thampiotr: I'm not sure, but on Windows files can be locked and return errors when they were recently accessed. @mattdurham I think has more context.

PR #3191, opened 2025-04-03
Fix panic in loki source file

wildum:
#### PR Description

The panic would happen when the runner would need to stop before it has time to start running

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/tailer.go
+++ b/internal/component/loki/source/file/tailer.go
@@ -348,14 +348,14 @@ func (t *tailer) Stop() {
 		t.stopping = false
 	}()
 
-	// Save the current position before shutting down tailer
-	err := t.markPositionAndSize()
-	if err != nil {
-		level.Error(t.logger).Log("msg", "error marking file position when stopping tailer", "path", t.path, "error", err)
-	}
-
+	var err error
 	// Stop the underlying tailer to prevent resource leak.
 	if t.tail != nil {
+		// Save the current position before shutting down tailer
+		err = t.markPositionAndSize()
+		if err != nil {
+			level.Error(t.logger).Log("msg", "error marking file position when stopping tailer", "path", t.path, "error", err)
+		}
 		err = t.tail.Stop()
 	}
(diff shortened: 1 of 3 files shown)


internal/component/loki/source/file/tailer.go
```diff
@@ -348,14 +348,14 @@ func (t *tailer) Stop() {
...
-	// Save the current position before shutting down tailer
-	err := t.markPositionAndSize()
-	if err != nil {
-		level.Error(t.logger).Log("msg", "error marking file position when stopping tailer", "path", t.path, "error", err)
-	}
-
+	var err error
 	// Stop the underlying tailer to prevent resource leak.
 	if t.tail != nil {
+		// Save the current position before shutting down tailer
```
ptodev, line 354: Would be nice to have a comment on why this snippet is here.

ptodev (approved):

internal/component/loki/source/file/tailer.go
wildum: I updated the comment

wildum:
> do any of these changes need to be ported over to the decompresser for compressed files?

Good catch, I took a similar approach for the decompresser and added a test.

> have we tested with file rotation?

I ran locally on Windows and Mac stress scenarios where thousands of files are created, deleted, rotated, and written at a very high frequency for a few minutes. Then I made sure that Alloy would still correctly tail all files with a script.

dehaansa (approved):

aglees:
I think there is still an issue with this. We've had to rollback `v1.8.1` to `v1.6.1` in our churning DEV environment because logs on short-lived nodes are being missed. We're seeing something like the initial Daemonset pods getting logs captured, but it appears that we're still getting a panic and no further container logs being recorded.

dehaansa:
@aglees I'm sorry to hear that you're still encountering issues. Can you open a new issue with information about your environment, your configuration, and the panic that you are seeing?

PR #3209, opened 2025-04-04
feat: Add `otelcol.storage.file` support

dehaansa:
#### PR Description
Added support for the [uptream](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage) `filestorage` extension. This component stores data on disk using [bbolt](https://github.com/etcd-io/bbolt) and currently is supported in exporters' `sending_queue` configuration to turn the in-memory queue into a persistent disk queue and in `otelcol.receiver.filelog` for tracking file offsets between Alloy restart.

Currently this is the only `otelcol.storage` component being added, but there are others that could be added in the future to add additional options for users to leverage other storage clients.

#### Which issue(s) this PR fixes
Fixes https://github.com/grafana/alloy/issues/323

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated

PR #3211, opened 2025-04-04
fix: Use names in otel component IDs

dehaansa:
#### PR Description
The existing otel component IDs do not use names. When using something like `storage.file` https://github.com/grafana/alloy/pull/3209 the component namespaced file requires a fully unique name per component, and the current IDs do not provide that.

For example, for the following config, the storage.file creates an object named `receiver_filelog_` which would conflict with any other filelog receivers instantiated.

```
otelcol.storage.file "default" {}

otelcol.receiver.filelog "default" {
    storage = otelcol.storage.file.default.handler
    ...
}
```


wildum (approved):
Thanks, that was missed during the Otel update. Are other components affected?  Should we update the release candidate with it?

PR #3213, opened 2025-04-04
feat: enhance zap adapter to log objects and arrays

dehaansa:
#### PR Description
While testing `otelcol.storage.file` https://github.com/grafana/alloy/pull/3209 I observed issues where the logs were not logging an array, just the text "<array>". This PR resolves this issue by enhancing the zapadapter to log arrays and objects as json encoded strings.

```
#Before
ts=2025-04-04T20:29:31.073666Z level=debug msg="Consuming files" component_path=/ component_id=otelcol.receiver.filelog.default component=fileconsumer paths=<array>

#After
ts=2025-04-04T20:30:50.275364Z level=debug msg="Consuming files" component_path=/ component_id=otelcol.receiver.filelog.default component=fileconsumer paths="[\"/Users/dehaansa/go/src/github.com/grafana/alloy/local/sam.log\",\"/Users/dehaansa/go/src/github.com/grafana/alloy/local/ts3.log\"]"
```


wildum (approved):
Thanks for making this change, I think that it would be good to add a changelog entry because users might be affected. (not sure how common it is for otelc components to log entire objects/arrays and if there are any component that could start logging huge objects)

Issue #3217, opened 2025-04-07
Configuration validation


towolf:
### Request

Grafan Agent had configuration validation that was added with `agentctl` [0], which is very useful to avoid rolling out broken configs.

But this tool was later removed in Alloy, and I can't find anything that replaced it.[1]

[0] https://github.com/grafana/agent/pull/336
[1] https://github.com/grafana/alloy/pull/15

### Use case

Test configuration syntax for validity.

towolf:
@rfratto you removed this tool, do you have context on this?

kalleep:
Hey. alloy has its own syntax for configuration so the command from agent could not be re-used. I have started working on adding a validation sub command but it will require some effort to get it working properly.

kalleep:
Closing this issue. I have merge initial functionality for validation and expanding on it in https://github.com/grafana/alloy/pull/3536. I have also created https://github.com/grafana/alloy/issues/3315 to track the work

PR #3220, opened 2025-04-07
feat: Add command for config validation

kalleep:
#### PR Description
Initial work for `alloy validate` command. In this pr I have implemented the easy part of config validation and plan to expand what can be validated in future pr:s. 

Currently it can check:
* Syntax errors
* Components exists and that they are unique.

Future work:
* Support type checking, that arguments exist and are of correct type
* Investigate if we can have a flag to resolve modules

Relates to: https://github.com/grafana/alloy/issues/3217
Part of: https://github.com/grafana/alloy/issues/3315

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


wildum (commented):
This change is a bit tricky because it's easy to miss checking for the source.HasErrors(). I would assume that if err is not nil, then everything is fine. Do you think that returning the errors from all files vs returning errors from the first parsed file has a lot of value?

kalleep:
> This change is a bit tricky because it's easy to miss checking for the source.HasErrors(). I would assume that if err is not nil, then everything is fine.

Yes I agree, `ParseSource(s)` function no longer returns error at all so we must remember to check for them like you said.

The biggest problem is `loadAlloySources` where we need to check both returned error and if there is any errors collected on sources. 

> Do you think that returning the errors from all files vs returning errors from the first parsed file has a lot of value?
Yes mostly for validation. I just made some changes but the idea is to run validation as far as possible for each source. So some files could pass while others won't. If we stop at first error we will not be able to do this. 

Also currently error reporting when using `alloy run` cannot "pretty" print because source is always nil, but that is fixed in this pr.

internal/service/http/http.go
```diff
@@ -52,7 +51,7 @@ type Options struct {
 	Gatherer prometheus.Gatherer  // Where to collect metrics from.
 
 	ReadyFunc  func() bool
-	ReloadFunc func() (*alloy_runtime.Source, error)
```
kalleep, line 55: Was ignored so decided to clean it up

internal/validator/validate_test.go
```diff
@@ -0,0 +1,81 @@
...
+					sources[f.Name] = f.Data
+				}
+
+				validateErr := Validate(Options{
+					Sources:           sources,
+					ComponentRegistry: component.NewDefaultRegistry(minStability, enableCommunityComps),
+				})
+
+				diagsFile := strings.TrimSuffix(path, txtarSuffix) + diagsSuffix
+				if !fileExists(diagsFile) {
```
kalleep, line 54: Currently the test will create a snapshot file if it don't exists. So to update one we just have to delete it and run tests again.

We can look into other approaches to this. https://github.com/gkampitakis/go-snaps?tab=readme-ov-file#update-snapshots could be an option. We create / update files if  `UPDATE_SNAPSHOT=true` env is set

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,40 @@
...
+menuTitle: validate
+title: The validate command
+weight: 200
+---
+
+# The `validate` command
+
+The `validate` command validates {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+## Usage
```
thampiotr, line 13: Can you describe how to use it in CI programmatically, e.g. the non-zero exit code = error, zero exit code = success.

internal/alloycli/cmd_validate.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/grafana/alloy/internal/service/remotecfg"
+	"github.com/grafana/alloy/internal/service/ui"
+	"github.com/grafana/alloy/internal/validator"
+	"github.com/spf13/cobra"
+)
+
+func validateCommand() *cobra.Command {
+	v := &alloyValidate{
+		configFormat: "alloy",
+		minStability: featuregate.StabilityGenerallyAvailable,
```
thampiotr, line 24: Should we start with public preview at most so we can make changes if we need to?

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,40 @@
...
+* _`<PATH_NAME>`_: Required. The {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+If you give the _`<PATH_NAME>`_ argument a directory path, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files (ignoring nested directories) and loads them as a single configuration source.
+
+The following flags are supported:
+
+* `--config.format`: The format of the source file. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, `static` (default `"alloy"`).
+* `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
+* `--config.extra-args`: Extra arguments from the original format used by the converter.
+* `--stability.level`: The minimum permitted stability level of functionality. Supported values: `experimental`, `public-preview`, `generally-available` (default `"generally-available"`).
```
thampiotr, line 31: Maybe should say that this should be set to the same value you intend to run Alloy with?

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,40 @@
...
+* `--feature.community-components.enabled`: Enable community components (default `false`).
```
thampiotr, line 32: & similarly to above

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Limitations
+
+Validation performed is limited in what it can check. Currently it will check:
+
+* Syntax errors.
+* Components exist.
+* Components are **unique** across all {{< param "PRODUCT_NAME" >}} configuration file and are not repeated.
```
thampiotr, line 40:
```suggestion
Validation performed is limited in what it can check. Currently it will check for:

* Syntax errors.
* Missing components.
* Components name conflicts.
```

internal/alloycli/cmd_validate.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:          "validate [flags] file",
+		Short:        "Validate a configuration file",
+		Long:         ``,
+		Args:         cobra.RangeArgs(0, 1),
```
thampiotr, line 31: I don't think we documented that there can be 0 arguments?

internal/service/labelstore/service.go
```diff
@@ -35,13 +35,13 @@ type staleMarker struct {
 
 type Arguments struct{}
 
-var _ alloy_service.Service = (*service)(nil)
+var _ alloy_service.Service = (*Service)(nil)
 
-func New(l log.Logger, r prometheus.Registerer) *service {
+func New(l log.Logger, r prometheus.Registerer) *Service {
```
thampiotr, line 40: Maybe it'd be better to just expose something like this:
```
func ServiceDefinition() alloy_service.Definition {
  return service{}.Definition()
}
```
Then we don't need to expose the entire Service, we just expose its Definition.

internal/validator/component_registry.go
```diff
@@ -0,0 +1,45 @@
...
+// custom components (declare blocks).
+type componentRegistry struct {
+	parent component.Registry
+	custom map[string]component.Registration
+}
+
+func (cr *componentRegistry) Get(name string) (component.Registration, error) {
+	parts := strings.Split(name, ".")
+	// Custom component registered with their label and should be the first part of the name.
+	// This does hower.
```
thampiotr, line 29: Not sure what you mean?

internal/validator/component_registry.go
```diff
@@ -0,0 +1,45 @@
...
+	// FIXME(kalleep): right now we register modules as a custom component and only validate the
+	// the namespace part. We cant really know what components is contained within a module without
+	// importing in first. Maybe we could have a option for validation to resolve modules so we could
+	// validate them correctly but by default we don't resolve and validate them.
```
thampiotr, line 33:
```suggestion
	// FIXME(kalleep): right now we register modules as custom components and only validate the
	// namespace part. We can't really know what components are contained within a module without
	// importing it first. Maybe we could have an option for validation to resolve modules so we could
	// validate them correctly but by default we don't resolve and validate them.
```

I agree with the idea of the option for modules!

internal/validator/validate_test.go
thampiotr: That works, but relies a bit on a convention that may be hard to find / understand. 

I like the approach we used here, where test has a flag: https://github.com/grafana/alloy/blob/main/internal/tools/docs_generator/docs_updated_test.go#L23 

...if it's possible to adopt it here would be great, but I don't mind too much.

thampiotr (commented):
Looking very good! Minor comments

internal/service/labelstore/service.go
thampiotr: I see that it's just the labelstore service though, so maybe it's okay. I'll let you make the call.

internal/alloycli/cmd_validate.go
kalleep: Yeah it should not be possible to have 0 args. I think I based the command from `fmt` first where you could pass config from stdin. Will fix that

internal/validator/component_registry.go
kalleep: Yeah I don't know what happened when I wrote this comment :see_no_evil: .

Custom components, declares and modules, are only registered by its label. This is how its done for runtime as well https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/custom_component_registry.go#L55-L69. Not sure this is correct for when I add support for _resolving_ modules but I will fix that when I get to that.

Declare:
```alloy
declare "self_collect" {
  argument "metrics_output" {
    optional = false
    comment  = "Where to send collected metrics."
  }

  prometheus.scrape "selfmonitor" {
    targets = [{
      __address__ = "127.0.0.1:12345",
    }]

    forward_to = [argument.metrics_output.value]
  }
}

self_collect "selfmonitor" {}
```
The label of `declare` will be component name. So for these components we will always get `self_collect` as name to this function so no split needed to make it work.

But for imports it is a bit different
```
import.file "math" {
  filename = "module.alloy"
}

math.add "default" {}
```
Here we get `math.add` as the name to lookup but without resolving the module we don't know if `add` exists. So currently we only check if `math` is registered and assume its fine if it does.

Btw how modules and declares works currently it is possible to do something like this when calling `alloy run`
```
declare "prometheus" {}
```
This would prevent usage of standard prometheus components because we will try to resolve it from either a module or declare block

internal/alloycli/cmd_validate.go
kalleep: I used the same default that we do for [run](https://github.com/grafana/alloy/blob/main/internal/alloycli/cmd_run.go#L67). I think its better to have the same default here

internal/validator/validate_test.go
kalleep: I think its better to be consistent. So I will add it for these tests too

docs/sources/reference/cli/validate.md
kalleep: Is [7979237](https://github.com/grafana/alloy/pull/3220/commits/79792375b1b2f85e8ad2710be0c5f8d99e1e598c) good enough? Or do you prefer to have this documented on the flag themself?

docs/sources/reference/cli/validate.md
kalleep: [2a1128d](https://github.com/grafana/alloy/pull/3220/commits/2a1128d1ffc9cb3b6494ca8c01acc9fb1a4dbcd0)

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+The following flags are supported:
+
+* `--config.format`: The format of the source file. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, `static` (default `"alloy"`).
+* `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
+* `--config.extra-args`: Extra arguments from the original format used by the converter.
+* `--stability.level`: The minimum permitted stability level of functionality. Supported values: `experimental`, `public-preview`, `generally-available` (default `"generally-available"`).
+* `--feature.community-components.enabled`: Enable community components (default `false`).
+
+{{< admonition type="note" >}}
+`--stability.level` and `--feature.community-components.enabled` should be set too the same values that is used when running alloy.
```
thampiotr, line 37:
```suggestion
`--stability.level` and `--feature.community-components.enabled` should be set to the same value that is used when running alloy.
```

internal/alloycli/cmd_validate.go
thampiotr: Ah, this is default. Sorry, I thought about this being the stability level of the `validate` command.

internal/validator/component_registry.go
thampiotr: Thanks. All sounds good. Re: overriding `prometheus` namespace - I think it was a design choice for people that want to customise stuff, so I think it's fine.

thampiotr (approved):
LGTM % the test change for re-generating the outputs. 
Also, would be great to get @clayton-cornell to take a quick look now or even after merge.

internal/service/labelstore/service.go
kalleep: For this use case it would be useful to be able to enumerate all `service.Definition` somehow, similar to how we can to it for components.  I can try to address / clean this up in another pr if its ok

internal/validator/validate_test.go
kalleep: Something like this [81a50d3](https://github.com/grafana/alloy/pull/3220/commits/81a50d3d01d7a45438db4be8addbd9baa477992a)

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+* _`<PATH_NAME>`_: Required. The {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+If you give the _`<PATH_NAME>`_ argument a directory path, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files (ignoring nested directories) and loads them as a single configuration source.
+
+The following flags are supported:
+
+* `--config.format`: The format of the source file. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, `static` (default `"alloy"`).
+* `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
+* `--config.extra-args`: Extra arguments from the original format used by the converter.
+* `--stability.level`: The minimum permitted stability level of functionality. Supported values: `experimental`, `public-preview`, `generally-available` (default `"generally-available"`).
```
clayton-cornell, line 33:
```suggestion
* `--config.format`: Specifies the source file format. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, and `static` (default `"alloy"`).
* `--config.bypass-conversion-errors`: Enable bypassing errors during conversion (default `false`).
* `--config.extra-args`: Extra arguments from the original format used by the converter.
* `--stability.level`: The minimum permitted stability level of functionality. Supported values: `experimental`, `public-preview`, and `generally-available` (default `"generally-available"`).
```

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+* `--feature.community-components.enabled`: Enable community components (default `false`).
+
+{{< admonition type="note" >}}
+`--stability.level` and `--feature.community-components.enabled` should be set to the same value that is used when running alloy.
```
clayton-cornell, line 37:
```suggestion
Set `--stability.level` and `--feature.community-components.enabled` to the same value used when running {{< param "PRODUCT_NAME" >}}.
```

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+{{< /admonition >}}
+
+
+## Limitations
+
+Validation performed is limited in what it can check. Currently it will check for:
```
clayton-cornell, line 43:
```suggestion
Validation is limited in scope. It currently checks for:
```

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+
+* Syntax errors.
+* Missing components.
+* Components name conflicts.
```
clayton-cornell, line 47:
```suggestion
* Component name conflicts.
```

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+
+The `validate` command validates {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+## Usage
+
+```shell
+alloy validate [<FLAG> ...] <PATH_NAME>
+```
+
+This command will fail with a non-zero exit code and print diagnostics to stderr if validation fails.
```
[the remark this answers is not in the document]
clayton-cornell: Yeah I struggled with this one - what does it do if the configuration is OK. We could tweak it even more here... focus furst on a success condition, then explain what happens if it fails validation.

```suggestion
If the configuration file is valid, the `validate` command returns a zero exit code.
If the configuration file is invalid, the command  returns a non-zero exit code and prints diagnostics generated during validation to stderr.
```

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+canonical: https://grafana.com/docs/alloy/latest/reference/cli/validate/
+description: Learn about the validate command
+menuTitle: validate
+title: The validate command
+weight: 200
+---
+
+# The `validate` command
+
+The `validate` command validates {{< param "PRODUCT_NAME" >}} configuration file/directory path.
```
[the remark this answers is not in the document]
kalleep: This was phrased in the same way we have for [run](https://github.com/grafana/alloy/blob/main/docs/sources/reference/cli/run.md?plain=1#L22). If its is a path to a file we only load that file. But if its a directory we try to load all alloy files in that directory

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+This command will fail with a non-zero exit code and print diagnostics to stderr if validation fails.
+
+Replace the following:
+
+* _`<FLAG>`_: One or more flags that define the input and output of the command.
+* _`<PATH_NAME>`_: Required. The {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+If you give the _`<PATH_NAME>`_ argument a directory path, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files (ignoring nested directories) and loads them as a single configuration source.
```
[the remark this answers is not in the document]
kalleep: This was taken from https://github.com/grafana/alloy/blob/main/docs/sources/reference/cli/run.md?plain=1#L26. Should we update there too?

docs/sources/reference/cli/validate.md
kalleep: [d036be6](https://github.com/grafana/alloy/pull/3220/commits/d036be64b91cad728f0376b37bd68e0e828ea9a4)

internal/validator/component_registry.go
```diff
@@ -0,0 +1,43 @@
...
+	if reg, ok := cr.custom[parts[0]]; ok {
+		return reg, nil
+	}
+
+	return cr.parent.Get(name)
+}
+
+func (cr *componentRegistry) registerCustomComponent(c *ast.BlockStmt) {
+	// FIXME(kalleep): Figure out how to resolve args and exports for declares and how we could
```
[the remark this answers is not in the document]
kalleep: Good point, it most definitely should!

docs/sources/reference/cli/validate.md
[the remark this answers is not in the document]
kalleep: Did it here [224c91d](https://github.com/grafana/alloy/pull/3220/commits/224c91d8c6d9eadd195b16be6d2ff8bd8bda0d1e)

internal/validator/report.go
```diff
@@ -0,0 +1,28 @@
...
+			ContextLinesAfter:  1,
+		})
+		_ = p.Fprint(w, sources, diags)
+
+		// Print newline after the diagnostics.
+		fmt.Println()
+		return
+	}
+
+	_, _ = fmt.Fprintf(w, "validation failed: %s", err)
```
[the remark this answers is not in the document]
kalleep: There are two cases where diagnostic won't be printed currently. Either if we cant load the files from disk or when calling [encoder.EnsureUTF8](https://github.com/grafana/alloy/blob/main/internal/runtime/source.go#L32). All other cases should print diagnostic errors. Would you prefer to update the docs to mention this?

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,48 @@
...
+The following flags are supported:
+
+* `--config.format`: Specifies the source file format. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, and `static` (default `"alloy"`).
+* `--config.bypass-conversion-errors`: Enable bypassing errors during conversion (default `false`).
+* `--config.extra-args`: Extra arguments from the original format used by the converter.
+* `--stability.level`: The minimum permitted stability level of functionality. Supported values: `experimental`, `public-preview`, and `generally-available` (default `"generally-available"`).
+* `--feature.community-components.enabled`: Enable community components (default `false`).
+
+{{< admonition type="note" >}}
+Set `--stability.level` and `--feature.community-components.enabled` to the same value used when running {{< param "PRODUCT_NAME" >}}.
```
clayton-cornell, line 38:
```suggestion
When you validate the {{< param "PRODUCT_NAME" >}} configuration, you must set the `--stability.level` and `--feature.community-components.enabled` arguments to the same values you want to use when you run {{< param "PRODUCT_NAME" >}}.
```

This one is a bit confusing on first read. Maybe something like this small change will make it more clear?

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,48 @@
...
+{{< /admonition >}}
+
+
+## Limitations
```
clayton-cornell, line 42:
```suggestion
{{< /admonition >}}

## Limitations
```

Remove extra empty line

docs/sources/reference/cli/validate.md
kalleep: Yes I agree, this is better

docs/sources/reference/cli/validate.md
```diff
@@ -0,0 +1,47 @@
...
+Replace the following:
+
+* _`<FLAG>`_: One or more flags that define the input and output of the command.
+* _`<PATH_NAME>`_: Required. The {{< param "PRODUCT_NAME" >}} configuration file/directory path.
+
+If you give the _`<PATH_NAME>`_ argument a directory path, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files (ignoring nested directories) and loads them as a single configuration source.
+
+The following flags are supported:
+
+* `--config.format`: The format of the source file. Supported formats: `alloy`, `otelcol`, `prometheus`, `promtail`, `static` (default `"alloy"`).
```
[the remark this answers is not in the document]
dehaansa: Maybe we should have some kind of clear message added to the log in this case, otherwise I think this could be very confusing should a user ever attempt it.

clayton-cornell (approved):

PR #3221, opened 2025-04-07
Allow setting trafficDistribution in service spec

nosammai:
This allows setting [trafficDistribution](https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution) on the service spec on K8s version >=1.32


eranikid:
I've stumbled across lack of support too. Grafana Loki chart does support `trafficDistribution` for `distributor` component

fabian-woelk-el:
Please add this feature. It works without sessionAffinity.

Issue #3181, opened 2025-04-02
prometheus.exporter.postgres component fails with minimal DSN


K900:
### What's wrong?

Alloy fails to start with "panic: unexpected keypair  from pq" when the DSN is simply `postgres://`.

### Steps to reproduce

Run Alloy with attached config file.

### Software version

Alloy v1.7.5

### Configuration

```text
prometheus.exporter.postgres "default" {
        data_source_names = [
                "postgres://",
        ]
}
```

### Logs

```text
panic: unexpected keypair  from pq

goroutine 1 [running]:
github.com/grafana/alloy/internal/static/integrations/postgres_exporter.parsePostgresURL({0xc000d8aaa1?, 0x0?})
        github.com/grafana/alloy/internal/static/integrations/postgres_exporter/postgres_exporter.go:103 +0x3b4
github.com/grafana/alloy/internal/static/integrations/postgres_exporter.(*Config).InstanceKey(0xc000d4e630, {0x41c825?, 0x90?})
        github.com/grafana/alloy/internal/static/integrations/postgres_exporter/postgres_exporter.go:69 +0xb3
github.com/grafana/alloy/internal/static/integrations.NewIntegrationWithInstanceKey({0xc948e80, 0xc0010fd130}, {0xc9e0940, 0xc000d4e630}, {0xc000d8ac00?, 0x0?})
        github.com/grafana/alloy/internal/static/integrations/integration.go:55 +0x3a
github.com/grafana/alloy/internal/component/prometheus/exporter/postgres.createExporter({{0x77a3503942f0, 0xc000b40660}, {0xc00102f0e0, 0x24}, {0xc948e80, 0xc0010fd130}, {0xc00102f200, 0x2f}, 0xc001c5ba10, {0xc9dec48, ...}, ...}, ...)
        github.com/grafana/alloy/internal/component/prometheus/exporter/postgres/postgres.go:30 +0x2bb
github.com/grafana/alloy/internal/component/prometheus/exporter.(*Component).Update(0xc0013525b0, {0xad8ff40, 0xc000ba2b80})
        github.com/grafana/alloy/internal/component/prometheus/exporter/exporter.go:89 +0xd2
[...]

PR #3226, opened 2025-04-08
fix: panic when using minimal url for `prometheus.postgres.exporter`

kalleep:
#### PR Description
When using minimal postgres url `postgres://` or `postgresql://` we encounter a panic. This function is only used when computing [InstanceKey](https://github.com/grafana/alloy/blob/main/internal/static/integrations/postgres_exporter/postgres_exporter.go#L56) where we add _default_ values for `host` and `post` if they are missing.

When reading the description of `pg.ParseURL` I can see:
```
// A minimal example:
//
//	"postgres://"
//
// This will be blank, causing driver.Open to use all of the defaults

```

So while it should work to connect using something like this the upstream exporter prevent us from doing so
https://github.com/prometheus-community/postgres_exporter/blob/master/cmd/postgres_exporter/util.go#L184.

With this fix we at least won't panic and alloy will start. But it will not be able to connect to a postgres db

Fixes: https://github.com/grafana/alloy/issues/3181

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/exporter/postgres/postgres.go
+++ b/internal/component/prometheus/exporter/postgres/postgres.go
@@ -2,7 +2,6 @@ package postgres
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/prometheus/exporter"
@@ -10,7 +9,6 @@ import (
 	"github.com/grafana/alloy/internal/static/integrations"
 	"github.com/grafana/alloy/internal/static/integrations/postgres_exporter"
 	"github.com/grafana/alloy/syntax/alloytypes"
-	"github.com/lib/pq"
 	config_util "github.com/prometheus/common/config"
 )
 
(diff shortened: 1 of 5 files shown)


internal/component/prometheus/exporter/postgres/postgres.go
```diff
@@ -30,37 +28,6 @@ func createExporter(opts component.Options, args component.Arguments, defaultIns
 	return integrations.NewIntegrationWithInstanceKey(opts.Logger, a.convert(opts.ID), defaultInstanceKey)
 }
 
-func parsePostgresURL(url string) (map[string]string, error) {
```
kalleep, line 33: This was not used and a duplicate of https://github.com/grafana/alloy/blob/main/internal/static/integrations/postgres_exporter/postgres_exporter.go#L90

thampiotr (approved):
Thanks!

PR #3232, opened 2025-04-08
fix: Ensure the remoteCfg service doesn't immediately stop if not configured

dehaansa:
#### PR Description
After recent linter inspired changes to ensure errors to registerCollector were not silently ignored, a regression was introduced that would cause the `remoteCfg` service to immediately fail if remote configuration was not configured.

This is unlikely to be an issue, but if a customer starts an Alloy instance without remote configuration, adds it and attempts to reload config it would not appropriately start remote configuration without an Alloy instance restart.

```
{"ts":"2025-04-08T14:17:18.765072464Z","level":"error","msg":"node exited with error","node":"remotecfg","err":"noop client"}
```

#### PR Checklist

- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ Main (unreleased)
 
 - Fix panic in `prometheus.exporter.postgres` when using minimal url as data source name. (@kalleep)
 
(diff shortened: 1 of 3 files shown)


jharvey10 (approved):
This looks pretty straightforward to me 👍. Definitely up to you if you'd like one other review on it for good measure though.

PR #3238, opened 2025-04-09
fix: reduce flakiness of remotecfg tests

dehaansa:
#### PR Description
Example of a flake [here](https://github.com/grafana/alloy/actions/runs/14340371103/job/40197887768#step:6:1318).

By adding a WaitGroup to each test, we ensure the component has stopped running before the test ends. There was a race condition where `getLastLoadedCfgHash` would update, and then inside the component it would not have cached the cfg to disk yet. The operation to cache to disk would then race with the t.TempDir() cleanup action and cause flaky test behavior.

I also took the opportunity to improve the TestAPIResponseNotModified test by ensuring that getConfig had been called at least once additionally after modifying the response, ensuring that we are actually testing the notmodified behavior and not just running through the test quickly enough to see the cfg hasn't changed without actually having a new Get operation.


thampiotr (approved):
probably better as table tests, but I'll leave this for future refactors

PR #3259, opened 2025-04-09
fix: Add validation that the right number of messages have processed in relabel test

dehaansa:
#### PR Description
Added validation to the relabel test that at each point we evaluate the cache we have seen the right number of messages. Also updated the end of the test so similar failures in the future will provide more detail (compare the whole cache content, rather than each key independently, so you can understand if the order is wrong or if the issue is different).

```
--- FAIL: TestCache (0.20s)
    relabel_test.go:291: 
        	Error Trace:	/home/runner/work/alloy/alloy/internal/component/loki/relabel/relabel_test.go:291
        	Error:      	Not equal: 
        	            	expected: 0x73734b052292def9
        	            	actual  : 0x87303522d898b13e
        	Test:       	TestCache
```

### DIFF
--- a/internal/component/loki/relabel/relabel_test.go
+++ b/internal/component/loki/relabel/relabel_test.go
@@ -11,7 +11,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/relabel"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/common/loki"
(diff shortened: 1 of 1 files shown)


internal/component/loki/relabel/relabel_test.go
```diff
@@ -220,7 +224,10 @@ func TestCache(t *testing.T) {
 	e.Labels = lsets[2]
 	c.receiver.Chan() <- e
 
-	time.Sleep(100 * time.Millisecond)
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		require.EqualValues(t, 3, receivedMessages.Load())
+	}, 250*time.Millisecond, 25*time.Millisecond)
```
thampiotr, line 229: Some GC pauses or other hangs in CI can take multiple seconds, so I would always go with at least 3 seconds limit.

PR #3295, opened 2025-04-11
fix: Pretty print diagnostic errors when using run command

kalleep:
#### PR Description
Extracted from https://github.com/grafana/alloy/pull/3220.

This change is related to that work but is useful on its own. 

Before this change if we had any Diagnostic errors when calling `alloy run` there where not pretty printed

Before:

After:

Other notable change are if config are from several files we continue to parse other files if on fails so we can report diagnostic from multiple files.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -2,6 +2,7 @@ package alloycli
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -409,14 +410,13 @@ func (fr *alloyRun) Run(cmd *cobra.Command, configPath string) error {
 		}
 
 		alloySource, err := alloy_runtime.ParseSources(sources)
+		defer instrumentation.InstrumentConfig(err == nil, hashSourceFiles(sources), fr.clusterName)
 		if err != nil {
-			instrumentation.InstrumentConfig(false, [32]byte{}, fr.clusterName)
 			return sources, fmt.Errorf("reading config path %q: %w", configPath, err)
 		}
 
 		httpService.SetSources(alloySource.SourceFiles())
 		if err := f.LoadSource(alloySource, nil, configPath); err != nil {
-			instrumentation.InstrumentConfig(false, [32]byte{}, fr.clusterName)
 			return sources, fmt.Errorf("error during the initial load: %w", err)
 		}
 
@@ -586,6 +586,20 @@ func loadSourceFiles(path string, converterSourceFormat string, converterBypassE
 	}
 
 	return map[string][]byte{path: bb}, nil
+}
+
+func hashSourceFiles(sources map[string][]byte) [sha256.Size]byte {
+	// Combined hash of all the sources.
+	hash := sha256.New()
+
+	// Sort keys so they are always added in the same order.
+	keys := maps.Keys(sources)
+	slices.Sort(keys)
+	for _, key := range keys {
+		hash.Write(sources[key])
+	}
+
+	return [32]byte(hash.Sum(nil))
 }
 
 // addDeprecatedFlags adds flags that are deprecated, but we keep them for backwards compatibility.
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -297,7 +297,7 @@ func (fr *alloyRun) Run(cmd *cobra.Command, configPath string) error {
 	// To work around this, we lazily create variables for the functions the HTTP
 	// service needs and set them after the Alloy controller exists.
 	var (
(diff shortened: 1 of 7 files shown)


internal/alloycli/cmd_run.go
```diff
@@ -398,19 +401,26 @@ func (fr *alloyRun) Run(cmd *cobra.Command, configPath string) error {
...
 
 	ready = f.Ready
-	reload = func() (*alloy_runtime.Source, error) {
-		alloySource, err := loadAlloySource(configPath, fr.configFormat, fr.configBypassConversionErrors, fr.configExtraArgs)
-		defer instrumentation.InstrumentConfig(err == nil, alloySource.SHA256(), fr.clusterName)
-
+	reload = func() (map[string][]byte, error) {
+		sources, err := loadSourceFiles(configPath, fr.configFormat, fr.configBypassConversionErrors, fr.configExtraArgs)
 		if err != nil {
+			instrumentation.InstrumentConfig(false, [32]byte{}, fr.clusterName)
```
dehaansa, line 407: This seems like a regression, we're losing config metrics & the label on them. And we are no longer reporting it on success. Can we refactor the SHA256 function to work on raw sources, and make sure we're still calling this on non-failure case?

internal/runtime/source.go
```diff
@@ -135,6 +138,13 @@ func ParseSources(sources map[string][]byte) (*Source, error) {
 
 		sourceFragment, err := ParseSource(namedSource.Name, namedSource.Content)
 		if err != nil {
+			// If we encounter diagnostic errors we combine them and
+			// later return all of them
+			var diags diag.Diagnostics
+			if errors.As(err, &diags) {
+				mergedDiags = append(mergedDiags, diags...)
+				continue
```
dehaansa, line 146: This seems like we should continue whether or not we received diags, as we might not receive diags from the first source but could receive them from later sources. What is the case where we want to exit early on an error?

dehaansa (commented):
Initial review looks good, with a couple questions. Also, we should probably have some validation tests.

internal/alloycli/cmd_run.go
kalleep: Oh right missed success case :see_no_evil: 

>Can we refactor the SHA256 function to work on raw sources, and make sure we're still calling this on non-failure case?

Yeah I took a look at usage of `source.SHA256()` and this is the only place we use it so we could remove it from `Sources` and just do it here instead when we have gotten the raw sources from disk

kalleep: Fixed in [c1189ec](https://github.com/grafana/alloy/pull/3295/commits/c1189ecb29a45472a72f225b754ca3f90ff5a671)

I decided to not calculate the hash during parsing. The problem with having it there (and we had that problem before) is that if we ever got an error this would have been empty because we return `nil, error`.  The function was protected with (nil check)[https://github.com/grafana/alloy/blob/main/internal/runtime/source.go#L173-L175]

internal/runtime/source.go
kalleep: Yes but I don't really now a good way to combine diagnostic and non diagnostic errors. Currently the only other error source would be from [EnsureUTF8](https://github.com/grafana/alloy/blob/8d5b1b4ef2f47a3db2f1116af4988c98766920c0/internal/runtime/source.go#L33). We could try to build a diagnostic error from this one but we don't have an parsed AST because this happens before we can try to parse it so I am not sure what we would set the start and end positions to.

kalleep:
>Initial review looks good, with a couple questions. Also, we should probably have some validation tests.

Do you mean that you want some test for `ParseSources` that validates that we continue if and collect more errors?

internal/runtime/source.go
dehaansa: Fair enough, once you fix UTF-8 errors you can get all diags, good enough. 👍

dehaansa:
> > Initial review looks good, with a couple questions. Also, we should probably have some validation tests.
> 
> Do you mean that you want some test for `ParseSources` that validates that we collect parse diagnostic for multiple sources?

That would be nice, I'm not actually sure what test is missing, but when you refactored out the success case and no tests failed my brain said "there's definitely a test missing there".

kalleep:
>I'm not actually sure what test is missing, but when you refactored out the success case and no tests failed my brain said "there's definitely a test missing there".

I agree but I also don't think its easy to add a test for this as this is all happening when calling `alloy run`. The only way I can think of is to run this command as a separate process with valid config, scrape metrics and check values. Update config and send signal to reload and scrape again. Not sure its worth it.

PR #3324, opened 2025-04-14
added windows event_data XML process

fabbaena:
#### PR Description
Windows event log source creates a JSON object that contains the raw slice of bytes from EventData XML tag which contains information that can be useful. 
The format of EventData is clearly defined by:
https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-datafieldtype-complextype
and can easily converted into a JSON key-value object.

This code modification reads the byte slice from EventData and converts it into a JSON key-value object which is then stored in a new key (event_data_map) in the exported data.

Also, added a unit test for the modification.

### DIFF
--- a/internal/component/loki/source/windowsevent/format.go
+++ b/internal/component/loki/source/windowsevent/format.go
@@ -10,12 +10,31 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"encoding/xml"
+	"slices"
+
 	jsoniter "github.com/json-iterator/go"
 	"golang.org/x/sys/windows"
 
 	"github.com/grafana/loki/v3/clients/pkg/promtail/scrapeconfig"
 	"github.com/grafana/loki/v3/clients/pkg/promtail/targets/windows/win_eventlog"
 )
+
+// This type is used to unmarshal the event_data map from the XML tags <Data>...</Data>
+// Definition of this type is based on Microsoft's documentation
+// https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-datafieldtype-complextype
+type EventDataMap map[string]string
+func (e *EventDataMap) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
+	var dataStruct struct {
+		Text *xml.CharData    `xml:",chardata"`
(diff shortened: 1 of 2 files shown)


internal/component/loki/source/windowsevent/format.go
```diff
@@ -80,6 +100,20 @@ func formatLine(cfg *scrapeconfig.WindowsEventsTargetConfig, event win_eventlog.
...
+		// TODO: win_eventlog library creates the event struct with event_data as a []byte that
+		// only contains the XML tags <Data>...</Data> unwrapped from the <EventData>...</EventData> tags.
+		// This hack wraps the list of <Data>...</Data> tags in <d>...</d> tags so that the
+		// xml.Unmarshal will work.
+		fullxml := slices.Concat([]byte("<d>"), event.EventData.InnerXML, []byte("</d>"))
+		err := xml.Unmarshal(fullxml, &temp)
+		if err != nil {
+			return "", fmt.Errorf("failed to unmarshal event data: %w", err)
+		}
+		structuredEvent.EventDataMap = temp.Data
```
ptodev, line 116: Is there any reason why a user would want to have both `event_data_map` and `event_data`? If not, should we have a config option such as `event_data` that can be set to either `json` or `xml`?

fabbaena: I just added a config parameter `include_event_data_map` that enables this feature. This way, both options are available. XML in `event_data` and JSON in `event_data_map`, each can be turned on or off separately.

fabbaena:
Also, added the entry in the changelog.

fabbaena:
> In addition to Paulin's comments, you'll need a changelog entry.

Hello @dehaansa I already created a new commit with the proposed changes. Did you have a chance to look at the proposed changes?

Thank you

dehaansa:
> > In addition to Paulin's comments, you'll need a changelog entry.
> 
> Hello @dehaansa I already created a new commit with the proposed changes. Did you have a chance to look at the proposed changes?
> 
> Thank you

Apologies for the slow responses - now that we've added a new configuration attribute, it will need to be added to the documentation.

PR #3326, opened 2025-04-14
ci: pin hashes for workflow dependencies

jharvey10:
As per recommended security guidelines, this PR pins all workflow dependencies to specific hashes. Updates will be handled by renovatebot.

Also updates some .yaml workflow files to .yml like the rest.

Big "thank you" to this tool for easing this transition:
https://github.com/mheap/pin-github-action

### DIFF
--- a/.github/workflows/publish-technical-documentation-release.yml
+++ b/.github/workflows/publish-technical-documentation-release.yml
@@ -17,7 +17,7 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4.2.2
         with:
           fetch-depth: 0
       - uses: grafana/writers-toolkit/publish-technical-documentation-release@publish-technical-documentation-release/v1
--- a/.github/workflows/publish-technical-documentation-release.yml
+++ b/.github/workflows/publish-technical-documentation-release.yml
@@ -17,10 +17,10 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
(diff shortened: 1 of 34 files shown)


.github/workflows/publish-technical-documentation-release.yml
```diff
@@ -17,10 +17,10 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
-      - uses: grafana/writers-toolkit/publish-technical-documentation-release@publish-technical-documentation-release/v1
+      - uses: grafana/writers-toolkit/publish-technical-documentation-release@d83ba5389fb8de1458b12bcc35ad4a4059883029 # publish-technical-documentation-release/v1
```
jdbaldry, line 23: Let me know if you have any trouble with this. I have a memory that Renovate doesn't understand tags like `<PREFIX>/<SEMANTIC VERSION>` and only supports dash separated prefix and version.

I'd prefer to not change the versioning format so if this turns out to be the case, I'll look to submit a patch to support this format but absolutely re-tag Writers' Toolkit if we need to.

jdbaldry (approved):
`publish-technical-documentation-.+` and `update-make-docs` workflows look good to me.

thampiotr (approved):

.github/workflows/publish-technical-documentation-release.yml
jharvey10: 👍 I'll keep an eye on it! I know within renovate we can do custom versioning syntax based on regular expressions, so I think there's a way to make it work; it'll just depend on whether or not this is supported out of the box.

jdbaldry: Yeah, should clarify that I think my memory pertains specifically to the GitHub Actions version logic that you get out of the box. Let's see how it goes :)

dehaansa (approved):

PR #3331, opened 2025-04-15
docs: improve docs for `loki.secretfilter` component

romain-gaillard:
#### PR Description
This PR improves the documentation of the `loki.secretfilter` component by making clearer which version of the embedded gitleaks configuration file is included in each version. Also provides advice on using a specific file to ensure consistency when updating Alloy.

None

#### PR Checklist

- [ ] CHANGELOG.md updated
- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.secretfilter.md
+++ b/docs/sources/reference/components/loki/loki.secretfilter.md
@@ -10,9 +10,9 @@ labels:
 
 {{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
-`loki.secretfilter` receives log entries and redacts detected secrets from them.
-The detection is based on regular expression patterns, defined in the [Gitleaks configuration file](#arguments) embedded within the component.
-`loki.secretfilter` can also use a custom configuration file based on the Gitleaks configuration file structure.
+`loki.secretfilter` receives log entries and redacts detected secrets from the log lines.
+The detection is based on regular expression patterns, defined in a Gitleaks configuration file embedded within the component.
+`loki.secretfilter` can also use a [custom configuration file](#arguments) based on the [Gitleaks configuration file structure][gitleaks-config].
 
 {{< admonition type="caution" >}}
 Personally Identifiable Information (PII) isn't currently in scope and some secrets could remain undetected.
@@ -23,6 +23,8 @@ Don't rely solely on this component to redact sensitive information.
 {{< admonition type="note" >}}
 This component operates on log lines and doesn't scan labels or other metadata.
 {{< /admonition >}}
+
+[gitleaks-config]: https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml
 
 ## Usage
 
@@ -50,7 +52,7 @@ The `gitleaks_config` argument is the path to the custom `gitleaks.toml` file.
 The Gitleaks configuration file [embedded in the component][embedded-config] is used if you don't provide the path to a custom configuration file.
 
 {{< admonition type="note" >}}
-This component doesn't support all the features of the [Gitleaks configuration file][gitleaks-config].
+This component doesn't support all the features of the Gitleaks configuration file.
 It only supports regular expression-based rules, `secretGroup`, and allowlist regular expressions. `regexTarget` only supports the default value `secret`.
 Other features such as `keywords`, `entropy`, `paths`, and `stopwords` aren't supported.
 The `extend` feature isn't supported.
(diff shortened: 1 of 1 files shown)


romain-gaillard:
@clayton-cornell Your expertise would very much be appreciated here. The point is that because the embedded config file will change from one version to another, it's useful for users checking the docs of their current version to have a link showing a snapshot of the file as it is for them (so that they can check what's included or not).

This approach seems to work, but only if there exists a GitHub tag for the Alloy version (so for instance here, `ALLOY_RELEASE` points to `v1.9.0` but because it hasn't been released yet, the link to GitHub doesn't work. I suppose this shouldn't be an issue as when users check the docs, the corresponding Alloy version is released? Not sure if I'm missing a scenario).

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -10,9 +10,9 @@ labels:
 
 {{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
-`loki.secretfilter` receives log entries and redacts sensitive information from them, such as secrets.
-The detection is based on regular expression patterns, defined in the [Gitleaks configuration file][gitleaks] embedded within the component.
-`loki.secretfilter` can also use a custom configuration file based on the Gitleaks configuration file structure.
+`loki.secretfilter` receives log entries and redacts detected secrets from the log lines.
+The detection is based on regular expression patterns, defined in the Gitleaks configuration file embedded within the component.
```
clayton-cornell, line 14:
```suggestion
The detection relies on regular expression patterns, defined in the Gitleaks configuration file embedded within the component.
```

Is `relies` more accurate here? `based on` suggests that there's an extra bit that happens... that it takes the regex patterns as a starting point and builds on top of that.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -60,6 +60,10 @@ If you use a custom configuration file, you must include all the rules you want
 Unsupported fields and values in the configuration file are ignored.
 {{< /admonition >}}
 
+{{< admonition type="note" >}}
+The embedded configuration file can change from one version of Alloy to another. It's advised to use an external configuration file to ensure consistency.
```
clayton-cornell, line 64:
```suggestion
The embedded configuration file may change between Alloy versions.
To ensure consistency, use an external configuration file.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -49,7 +49,7 @@ loki.secretfilter "<LABEL>" {
 | `types`           | `map(string)`        | Types of secret to look for.                               | All types                        | no       |
 
 The `gitleaks_config` argument is the path to the custom `gitleaks.toml` file.
-The Gitleaks configuration file embedded in the component is used if you don't provide the path to a custom configuration file.
+The Gitleaks configuration file [embedded in the component][embedded-config] is used if you don't provide the path to a custom configuration file.
```
clayton-cornell, line 52:
```suggestion
If you don't provide the path to a custom configuration file, the Gitleaks configuration file [embedded in the component][embedded-config] is used.
```

clayton-cornell (dismissed):

clayton-cornell:
Can we backport this? To v1.6, v1.7, and v1.8?

docs/sources/reference/components/loki/loki.secretfilter.md
romain-gaillard: For now, it's indeed solely regexes (both for identification and allow-listing). In the future, we might add extra checks like entropy of the detected string. Either way, I think you're right and `relies` is probably more appropriate.

romain-gaillard:
> Can we backport this? To v1.6, v1.7, and v1.8?

Indeed, the clarifications made here apply to every version. I believe the first Alloy version to have this component was `1.5.0`.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -60,6 +60,11 @@ If you use a custom configuration file, you must include all the rules you want
 Unsupported fields and values in the configuration file are ignored.
 {{< /admonition >}}
 
+{{< admonition type="note" >}}
+The embedded configuration file may change between Alloy versions.
```
clayton-cornell, line 64:
```suggestion
The embedded configuration file may change between {{< param "PRODUCT_NAME" >}} versions.
```

clayton-cornell (approved):

clayton-cornell (approved):

PR #3355, opened 2025-04-16
database_observability: extend `query_tables` with prepared statements

cristiangreco:
#### PR Description
Extend the `query_tables` collector to also parse prepared statements from the `prepared_statements_instances` table.

Follow similar logic as the one used for the `query_samples` table, but we calculate a digest locally as mysql does not provide one for prepared statements (this also requires redacting the statement first, and making sure no duplicates are processed after redaction).

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/parser/parser_tidb.go
+++ b/internal/component/database_observability/mysql/collector/parser/parser_tidb.go
@@ -3,6 +3,7 @@ package parser
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"golang.org/x/exp/maps"
@@ -75,7 +76,9 @@ func (p *TiDBSqlParser) ExtractTableNames(_ log.Logger, _ string, stmt any) []st
 		tables: map[string]struct{}{},
 	}
 	(*stmt.(*ast.StmtNode)).Accept(v)
-	return maps.Keys(v.tables)
+	keys := maps.Keys(v.tables)
+	slices.Sort(keys)
+	return keys
 }
 
 func (p *TiDBSqlParser) ParseTableName(t any) string {
--- a/internal/component/database_observability/mysql/collector/query_tables.go
+++ b/internal/component/database_observability/mysql/collector/query_tables.go
@@ -103,7 +103,7 @@ func (c *QueryTables) Start(ctx context.Context) error {
 		ticker := time.NewTicker(c.collectInterval)
 
 		for {
-			if err := c.tablesFromQuerySamples(c.ctx); err != nil {
+			if err := c.tablesFromEventsStatements(c.ctx); err != nil {
 				level.Error(c.logger).Log("msg", "collector error", "err", err)
 			}
 			if err := c.tablesFromPreparedStatements(c.ctx); err != nil {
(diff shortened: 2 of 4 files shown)


internal/component/database_observability/mysql/collector/parser/parser_tidb.go
```diff
@@ -75,7 +76,9 @@ func (p *TiDBSqlParser) ExtractTableNames(_ log.Logger, _ string, stmt any) []st
 		tables: map[string]struct{}{},
 	}
 	(*stmt.(*ast.StmtNode)).Accept(v)
-	return maps.Keys(v.tables)
+	keys := maps.Keys(v.tables)
+	slices.Sort(keys)
+	return keys
```
cristiangreco, line 81: Drive-by change, makes testing easier.

internal/component/database_observability/mysql/collector/query_tables.go
```diff
@@ -157,30 +170,87 @@ func (c *QueryTables) fetchQueryTables(ctx context.Context) error {
...
+
+	if err := rs.Err(); err != nil {
+		level.Error(c.logger).Log("msg", "error during iterating over prepared statements result set", "err", err)
+		return err
+	}
+
+	return nil
+}
+
+func (c *QueryTables) tryParse(schema, digest, sqlText, fallbackSqlText string) (any, error) {
```
cristiangreco, line 234: Have swapped params order and renamed them, I think it's more readable now.

internal/component/database_observability/mysql/collector/query_tables_test.go
```diff
@@ -291,6 +365,7 @@ func TestQueryTables(t *testing.T) {
 				CollectInterval: time.Second,
 				EntryHandler:    lokiClient,
 				Logger:          log.NewLogfmtLogger(os.Stderr),
+				UseTiDBParser:   true,
```
cristiangreco, line 368: Drive-by: enabled the parser in tests given that it's the defalt parser now.

internal/component/database_observability/mysql/collector/query_tables.go
```diff
@@ -117,7 +131,7 @@ func (c *QueryTables) Stop() {
 	c.cancel()
 }
 
-func (c *QueryTables) fetchQueryTables(ctx context.Context) error {
+func (c *QueryTables) tablesFromQuerySamples(ctx context.Context) error {
```
matthewnolf, line 134: What is the rationale for this new name? It doesn't seem related to query samples as we have modelled it in the component.

internal/component/database_observability/mysql/collector/query_tables_test.go
```diff
@@ -22,13 +22,14 @@ func TestQueryTables(t *testing.T) {
 
 	testcases := []struct {
 		name       string
-		rows       [][]driver.Value
+		qRows      [][]driver.Value
```
matthewnolf, line 25: can we expand `qRows` and `psRows` - the naming is quite vague.
It's not intuitive qRows means rows from `events_statements_summary_by_digest` table

internal/component/database_observability/mysql/collector/query_tables.go
cristiangreco: Well these are actualy samples though, as they're coming from `events_statements_summary_by_digest.query_sample_text`.

Anyway I wanted to make this clearly distinct from the other methods that deals with prepared statements rather than regular queries.

Happy to use a name recommendation.

internal/component/database_observability/mysql/collector/query_tables_test.go
cristiangreco: They stand for `query` rows and `prepared statement` rows. What (possibly short) name do you suggest?

matthewnolf: `eventStatementsRows` and `preparedStatementRows`?
or `esRows` and `psRows`

PR #3356, opened 2025-04-16
fix instance key in exporters

thampiotr:
#### PR Description

- Changes `instance` label to be the component name instead of the hostname of Alloy for the following exporters:
  - SNMP
  - StatsD
  - Blackbox

- [x] Document that process/windows/unix exporters should not be used with clustering.
- [x] Call out the change and document how to get the old behaviour.
- [x] Add a warning when clustering is enabled for process/windows/unix exporters.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
@@ -17,11 +17,11 @@ The `prometheus.exporter.cadvisor` component collects container metrics using [c
 {{< admonition type="note" >}}
 Take care when using the `prometheus.exporter.cadvisor` component with [clustering](../../../../get-started/clustering/) enabled.
 
-The default instance label set by this exporter is the hostname of the machine running Alloy. Alloy clustering uses consistent hashing to distribute targets across instances, and the discovered targets must be the same (and have the same labels) across all cluster instances.
-
-Therefore, when using `prometheus.exporter.cadvisor` within a [cluster](../../../../get-started/clustering/) of Alloy instances,
-it is recommended to use a dedicated `prometheus.scrape` component that is used to scrape `prometheus.exporter.cadvisor`
-and does not have clustering enabled.
+The default `instance` label set by this exporter is the hostname of the machine running Alloy. Alloy clustering uses consistent
+hashing to distribute targets across the instances. This approach requires the discovered targets to be the same and have the same
+labels across all cluster instances. When using `prometheus.exporter.cadvisor` within a [cluster](../../../../get-started/clustering/)
+of Alloy instances, it is recommended to use a dedicated `prometheus.scrape` component that is used to
+scrape `prometheus.exporter.cadvisor` and does not have clustering enabled.
 {{< /admonition >}}
 
 ## Usage
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
@@ -14,17 +14,7 @@ title: prometheus.exporter.cadvisor
 
 The `prometheus.exporter.cadvisor` component collects container metrics using [cAdvisor](https://github.com/google/cadvisor).
 
-{{< admonition type="note" >}}
-Take care when using the `prometheus.exporter.cadvisor` component with [clustering](../../../../get-started/clustering/) enabled.
-
-The default `instance` label set by this exporter is the hostname of the machine running {{< param "PRODUCT_NAME" >}}.
-{{< param "PRODUCT_NAME" >}} clustering uses consistent hashing to distribute targets across the instances.
-This approach requires the discovered targets to be the same and have the same labels across all cluster instances.
-
-When you use `prometheus.exporter.cadvisor` within a [cluster](../../../../get-started/clustering/)
-of {{< param "PRODUCT_NAME" >}} instances, use a dedicated `prometheus.scrape` component that's used to scrape
-`prometheus.exporter.cadvisor` and doesn't have clustering enabled.
-{{< /admonition >}}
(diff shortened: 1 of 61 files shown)


internal/component/prometheus/exporter/exporter.go
```diff
@@ -149,7 +148,6 @@ func newExporter(creator Creator, name string, targetBuilderFunc func(discovery.
 			model.AddressLabel:      httpData.MemoryListenAddr,
 			model.SchemeLabel:       "http",
 			model.MetricsPathLabel:  path.Join(httpData.HTTPPathForComponent(opts.ID), "metrics"),
-			"instance":              instance,
```
thampiotr, line 152: We always set `instance` in `Update`, so no need to call set it here any more.

internal/component/prometheus/exporter/tests/instance_key_test.go
thampiotr, line 1: This test summarises all the actual changes that this PR does.

ptodev (commented):
The change looks ok to me, as long as it's documented. My main worry is that even if we make the `instance` label more sensible for some components, the experience of using clustering together with exporters will still not be user friendly. I wish we didn't have to rely on users reading about such edge cases in the docs.

docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
```diff
@@ -14,6 +14,16 @@ title: prometheus.exporter.cadvisor
...
 The `prometheus.exporter.cadvisor` component collects container metrics using [cAdvisor](https://github.com/google/cadvisor).
 
+{{< admonition type="note" >}}
+Take care when using the `prometheus.exporter.cadvisor` component with [clustering](../../../../get-started/clustering/) enabled.
+
+The default `instance` label set by this exporter is the hostname of the machine running Alloy. Alloy clustering uses consistent
+hashing to distribute targets across the instances. This approach requires the discovered targets to be the same and have the same
+labels across all cluster instances. When using `prometheus.exporter.cadvisor` within a [cluster](../../../../get-started/clustering/)
+of Alloy instances, it is recommended to use a dedicated `prometheus.scrape` component that is used to
+scrape `prometheus.exporter.cadvisor` and does not have clustering enabled.
```
clayton-cornell:
```suggestion
The default `instance` label set by this exporter is the hostname of the machine running {{< param "PRODUCT_NAME" >}}.
{{< param "PRODUCT_NAME" >}} clustering uses consistent hashing to distribute targets across the instances.
This approach requires the discovered targets to be the same and have the same labels across all cluster instances.
When you use `prometheus.exporter.process` within a [cluster](../../../../get-started/clustering/) of {{< param "PRODUCT_NAME" >}} instances, use a dedicated `prometheus.scrape` component that's used to scrape `prometheus.exporter.process` and doesn't have clustering enabled.
```

Could be more explicit in the recommendation and just state "use" the dedicated component?

ptodev: I agree that the recommendation should be more prominent. I'd even put it in as the first sentence so that it's clear to the users what they should do. Users who don't want to follow the recommendation are free to read more about it further on.

clayton-cornell:
The docs look OK. I can see it being something that could hit users, but... it's documented. I can't think of another way to communicate the info.

docs/sources/reference/components/prometheus/prometheus.exporter.process.md
```diff
@@ -14,6 +14,18 @@ title: prometheus.exporter.process
 
 The `prometheus.exporter.process` component embeds the [`process_exporter`](https://github.com/ncabatoff/process-exporter) for collecting process stats from `/proc`.
 
+{{< admonition type="note" >}}
+Take care when using the `prometheus.exporter.process` component with [clustering](../../../../get-started/clustering/) enabled.
```
ptodev: @clayton-cornell would it it more clear to say "It is not recommended to use `prometheus.exporter.process` with clustering"? We're not asking the users to take care, we're asking them to not do it. But it only becomes apparent in the final paragraph at the moment.

It would be nice to have a generic md snippet that we can include in each component doc so that it doesn't have to be edited everywhere?

internal/component/prometheus/exporter/github/github.go
```diff
@@ -21,8 +21,9 @@ func init() {
 	})
 }
 
-func createExporter(opts component.Options, args component.Arguments, defaultInstanceKey string) (integrations.Integration, string, error) {
+func createExporter(opts component.Options, args component.Arguments) (integrations.Integration, string, error) {
 	a := args.(Arguments)
+	defaultInstanceKey := opts.ID // if cannot resolve instance key, use the component ID
```
ptodev, line 26: What is the reason we don't do `WarningIfUsedInCluster` here? Is it to avoid spamming the logs?

thampiotr: Because github exporter does not use any host-specific instance label value, like the HOSTNAME of the machine that Alloy runs on. Instead it uses the host name of the github URL that is being exported. So it will work fine with clustering. 

This is covered in tests, btw. Given the URL of `https://api.github.com:8080`, the instance label becomes `api.github.com:8080`.

ptodev: I think I saw a comment in the test that the repo URL may not be enough and there coult still be issues. That's why I wasn't sure if we should still have a warning. Don't worry about it, this PR is a huge improvement over the current situation already and there's no need to make everything gold plated straight away :)

awoimbee:
The docs could be way clearer by explaining the reachability (local and shared) of the exporters.
Something like:
> the targets of this exporter are specific to each node, meaning the connected prometheus.scrape cannot use clustering [...]

Issue #1009, opened 2024-06-10
When using clustering, exporters may not work correctly due to `instance` label


thampiotr:
### What's wrong?

Most embedded Prometheus exporters set the `instance` label to [the hostname where Alloy runs](https://github.com/grafana/agent/blob/main/internal/component/prometheus/exporter/exporter.go#L179).

This breaks in a subtle, but significant way, the fundamental clustering assumption that all instances [have the same configuration](https://grafana.com/docs/alloy/next/concepts/clustering/#clustering). The exporters implicitly inject the hostname as an instance label, but instances usually have different hostnames. This leads to either no scraping of metrics at all, or duplicate scraping with different instance labels (unnecessary).

### Steps to reproduce

1. Run any exporter in a clustered mode in a cluster of 2+ instances, each running on a different host. Have scraping set up with clustering and a remote write to a metrics DB.
2. Observe that some targets will not be scraped at all, some will be scraped multiple times, with different instance labels.
3. Observe in the UI that instance label is different in exporters' targets between instances, indicating different series.

The issue was discussed [in this PR](https://github.com/grafana/agent/pull/6792), but decided to move the conversation here for better tracking and to provide a place to refer to for workarounds.

thampiotr:
There is a workaround for now: set the `instance` label to a common value for all instances in the cluster, using `discovery.relabel` component. For example, this component sets it to "alloy-cluster":

```
discovery.relabel "replace_instance" {
  targets = discovery.file.targets.targets
  rule {
    action        = "replace"
    source_labels = ["instance"]
    target_label  = "instance"
	replacement   = "alloy-cluster"
  }  
}
```

You'd add the above component between your exporters and the prometheus.scrape.

Longer term fix can be also achieved via https://github.com/grafana/alloy/issues/399. Regardless, we should have good documentation to ensure users don't fall into this pit.

rgarrigue:
Affected our 30+ blackbox probes over ~7 alloy deployed via Helm, we were missing out on random targets, triggering DatasourceNoData in our alerting. The workaround fixed it.

st-akorotkov:
TBH this proposed rule is not a workaround. Using it breaks multiple dashboard and alert since we can't distinguish nodes running node-exporter anymore.

xdr34m:
Omg, it nearly drove me insane, until i found this issue.
[...]

PR #3358, opened 2025-04-16
add tests for exporters instance label

thampiotr:
#### PR Description

As pointed out in https://github.com/grafana/alloy/issues/1009 
the instance label is not always set to something useful in exporters and it can particularly be harmful when using clustering.

In this PR I add a test that makes it clear what exporter uses what for its instance label and add TODOs to fix those that don't make sense. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/exporter/tests/instance_key_test.go
+++ b/internal/component/prometheus/exporter/tests/instance_key_test.go
@@ -3,8 +3,11 @@ package exporter_test
 import (
 	"fmt"
 	"net"
+	"os"
 	"testing"
+	"time"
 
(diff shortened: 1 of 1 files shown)


thampiotr:
> LGTM. It would be good to make sure we test every exporter we have, even in the future when we add new ones.
> 
> Is there a easy way to enumerate all exporters? Then the test could keep track of which one we are checking and fail if we missed some

Yeah, I think there is! Good point, I'll add this.

kalleep (approved):

PR #3381, opened 2025-04-17
fix: Keep ebpf component running

simonswine:
If the eBPF encounter an error, it stopped the main processing loops,
which then caused issues because any config reload for alloy would block
the whole agent.


internal/component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -104,48 +103,74 @@ type Component struct {
...
+				continue
+			}
+			c.appendable.UpdateChildren(newArgs.ForwardTo)
+			if c.args.CollectInterval != collectInterval {
+				t.Reset(c.args.CollectInterval)
+				collectInterval = c.args.CollectInterval
+			}
+		case <-t.C:
+			if !started {
+				err := c.session.Start()
```
simonswine, line 142: I imagine main of those errors are permanent (e.g. not enough permissions, we probably should handle those by stopping the ticker. Wdyt @korniltsev

simonswine:
@JohanLindvall wonder if this fixes your problem and/or gives you a better error message?

JohanLindvall:
> @JohanLindvall wonder if this fixes your problem and/or gives you a better error message?

@simonswine With these changes our Alloy Pyroscope scrapers have been running for a few hours without any issues.

I have not seen any of the new error messages logged.

simonswine:
I have built an image with the changes from this PR: In case someone wants to try: `docker.io/simonswine/alloy:retry-ebpf-loop-v2@sha256:d914e10e780f1ecb60f94134389a26b6565dea3456e13ca3e83293c2502bb5a8`

internal/component/pyroscope/ebpf/ebpf_linux.go
simonswine: Now implemented https://github.com/grafana/alloy/pull/3381/commits/19591e05769f961573d2888daae544a4bcd7c509

aleks-p (approved):
Looks good to me. I noticed the added error logging which is great. I would consider adding Info and Debug logs in a few select places to better understand what happens during important state transitions.

PR #3384, opened 2025-04-17
docs: fix link to embedded config file in `loki.secretfilter` component

romain-gaillard:
#### PR Description

This PR fixes the link introduced in #3331 which doesn't seem to work quite right for `v1.8` (latest) as it points to the `next` version.

It introduces a variable that will have to be manually updated when a change is made to the embedded config file. It should point to the last Alloy version that had a change of the embedded config file (meaning that when this file is updated, the variable should point to the next release).

#### PR Checklist

- [ ] CHANGELOG.md updated
- [X] Documentation added
- [ ] Tests updated


romain-gaillard:
Not needed. The issue with `v1.8` was unrelated and has been fixed since then.

PR #3397, opened 2025-04-20
foreach: Add id field to use collection item fields for child component identification

### LINKED ISSUE #3267: foreach: add a way to specify an ID instead of the generated uid for objects
The children created by the foreach will get a UID associated with their instance when the collection contains objects. When the metrics are enabled, the component metrics will have a component path that looks like this:

`component_path="/foreach.retention/foreach_<some_uid>/aws_log_retention.default"`

The problem is that it does not give meaningful information about the foreach child.

One option to solve this could be to add an optional "id" argument to be used instead of a UID:

```
foreach "default" {
  collection           = []
  var                      = "each"
  enable_metrics = true
  id                        = "namespace" // or each["namespace"]?
  template {}
}
```

In this scenario, the namespace field of the objects will be used instead of the UID and the component path in the metrics will look like this:

`component_path="/foreach.retention/foreach_<namespace>/aws_log_retention.default"`

Collision can be handled by simply adding a suffix with an increment:

`component_path="/foreach.retention/foreach_<namespace>_1/aws_log_retention.default"`

harshrai654:
#### PR Description

Add a new optional `id` field to the `foreach` block which allows specifying a field name from collection items to be used for child component identification. When set and the collection contains objects, the value of this field will be used to generate more meaningful component paths in metrics, making it easier to identify and debug components in large deployments.

For example:
```alloy
foreach "pods" {
    collection = [
        { namespace = "prod", name = "app-1" },
        { namespace = "dev", name = "app-1" }
    ]
    var = "each"
    id = "name"
    enable_metrics = true
    template { ... }
}
```

Will result in component paths like:
```
/foreach.pods/foreach_app-1_1/component.default
/foreach.pods/foreach_app-1_2/component.default
```

Fixes #3267 

#### Notes to the Reviewer

- The id field is optional and defaults to using the entire collection item for identification
- If the collection item is not a map or the specified field doesn't exist, it falls back to using the entire item
- Added tests to verify the behaviour with use of collection object's field as `id` along with collision scenario.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/config-blocks/foreach.md
+++ b/docs/sources/reference/config-blocks/foreach.md
@@ -7,7 +7,6 @@ menuTitle: foreach
 title: foreach
 ---
 
-
 # foreach
 
 {{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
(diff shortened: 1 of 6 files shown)


docs/sources/reference/config-blocks/foreach.md
```diff
@@ -30,27 +29,56 @@ foreach "<LABEL>" {
...
+| `id`             | `string`    | Name of the field to use from collection items for child component's identification.     | `""`    | no       |
 
 The items in the `collection` list can be of any type [type][types], such as a bool, a string, a list, or a map.
 
 {{< admonition type="warning" >}}
 Setting `enable_metrics` to `true` when `collection` has lots of elements may cause a large number of metrics to appear on the {{< param "PRODUCT_NAME" >}} `/metric` endpoint.
 {{< /admonition >}}
 
+{{< admonition type="note" >}}
+When `id` is set and `enable_metrics` is true, component paths in metrics will use the value of the specified field from collection items instead of a hash of the entire item. For example:
```
clayton-cornell, line 46:
```suggestion
{{< admonition type="note" >}}
When `id` is set and `enable_metrics` is `true`, component paths in metrics use the value of the specified field from collection items instead of a hash of the entire item.
For example:
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -30,27 +29,56 @@ foreach "<LABEL>" {
...
+        { namespace = "prod", name = "app-2" }
+    ]
+    var = "each"
+    id = "name"
+    enable_metrics = true
+    template { ... }
+}
+```
+
+This will result in component paths like:
```
clayton-cornell, line 62:
```suggestion
This results in component paths like the following:
```

docs/sources/reference/config-blocks/foreach.md
```diff
@@ -30,27 +29,56 @@ foreach "<LABEL>" {
...
+
+```
```
clayton-cornell, line 64:
````suggestion
```text
````

**Review by wildum — APPROVED** [review]
Great work, thanks!

**Comment by wildum** [comment]
Could you rebase with main please?

**Comment by harshrai654** [comment]
Rebased with [grafana:main](https://github.com/grafana/alloy/tree/main).

File: internal/runtime/testdata/foreach_metrics/foreach_4.txtar [hunk 4]
```diff
@@ -0,0 +1,52 @@
...
+# TYPE alloy_component_evaluation_queue_size gauge
+alloy_component_evaluation_queue_size{controller_id="",controller_path="/"} 1
+alloy_component_evaluation_queue_size{controller_id="foreach_dev_1",controller_path="/foreach.testForeach"} 0
+alloy_component_evaluation_queue_size{controller_id="foreach_dev_2",controller_path="/foreach.testForeach"} 0
+alloy_component_evaluation_queue_size{controller_id="foreach_prod_1",controller_path="/foreach.testForeach"} 0
+# HELP pulse_count
+# TYPE pulse_count counter
+pulse_count{component_id="testcomponents.pulse.pt",component_path="/foreach.testForeach/foreach_dev_1"} 3
+pulse_count{component_id="testcomponents.pulse.pt",component_path="/foreach.testForeach/foreach_dev_2"} 4
+pulse_count{component_id="testcomponents.pulse.pt",component_path="/foreach.testForeach/foreach_prod_1"} 3
```
wildum, line 47: looks like the pulse_count is a bit flaky, probably a race condition in the tests. This is not important for your two tests because it should only validate the component_path. I would recommend just checking one metric to simplify it ("alloy_component_controller_running_components") in both tests

PR #3404, opened 2025-04-22
Add Beyla integration test

marctc:
#### PR Description

_This PR is a continuation of #1699._ 

Add integration tests for Beyla component. This needed some refactor as Beyla needs to run as privileged container,
so the only way to achieve that is using github.com/testcontainers/testcontainers-go.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


.golangci.yml
```diff
@@ -59,6 +59,7 @@ linters:
       - examples$
       - component/otelcol/extension/jaeger_remote_sampling/internal # Copied with attribution from OTel Collector
       - component/otelcol/receiver/prometheus/internal # Copied with attribution from OTel Collector
+      - internal/cmd/integration-tests # Exclude all integration tests
```
wildum, line 62: do we need to exclude them from the lint?

internal/cmd/integration-tests/utils.go
```diff
@@ -22,29 +32,32 @@ type TestLog struct {
...
-
-func setupEnvironment() {
-	executeCommand("docker", []string{"compose", "up", "-d"}, "Setting up environment with Docker Compose")
-	fmt.Println("Sleep for 45 seconds to ensure that the env has time to initialize...")
-	time.Sleep(45 * time.Second)
+	executeCommand("make", []string{"-C", "../../..", "ALLOY_IMAGE=" + alloyImageName, "alloy-image"}, "Building Alloy")
 }
 
-func runSingleTest(testDir string, port int) {
+func runSingleTest(ctx context.Context, testDir string, port int) {
```
wildum, line 60: this function is a bit big now, maybe it could be split?

internal/cmd/integration-tests/utils.go
```diff
@@ -55,46 +68,137 @@ func runSingleTest(testDir string, port int) {
...
+			}
+			return
+		}
+
+		fmt.Printf("Loki API should be available at http://%s:%s/loki/api/v1/push\n",
+			host, mappedPort.Port())
+
+		testCmd.Env = append(os.Environ(),
+			fmt.Sprintf("ALLOY_HOST=%s", host),
+			fmt.Sprintf("ALLOY_PORT=%s", mappedPort.Port()))
```
wildum, line 184: it's a bit annoying to have test specific logic at this level, I think that's ok for now but could you just add a TODO comment to somehow improve this?

wildum (commented):
thanks a lot for picking this up!

.golangci.yml
marctc: it was complaining about not checking the defers in the code of tests, so i'd say so

.github/workflows/integration-tests.yml
```diff
@@ -13,17 +13,15 @@ permissions:
...
-    runs-on: ubuntu-latest
+    runs-on: github-hosted-ubuntu-x64-large
     steps:
     - name: Checkout code
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         persist-credentials: false
     - name: Setup Go
-      uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
+      uses: actions/setup-go@v5
```
dehaansa, line 23: Why was this changed? We should keep actions pinned to sha for security reasons.

.github/workflows/integration-tests.yml
```diff
@@ -13,17 +13,15 @@ permissions:
...
       with:
-        go-version-file: go.mod
-    - name: Set OTEL Exporter Endpoint
-      run: echo "OTEL_EXPORTER_ENDPOINT=172.17.0.1:4318" >> $GITHUB_ENV
+        go-version: "1.22"
```
dehaansa, line 25: Why was this changed?

internal/cmd/integration-tests/docker-compose.yaml
```diff
@@ -1,6 +1,4 @@
-version: "3"
```
dehaansa, line 1: Why was this changed? If test-containers requires a different version we should be explicit.

go.mod
```diff
@@ -939,8 +980,11 @@ require (
 	k8s.io/kube-openapi v0.0.0-20241105132330-32ad38e42d3f // indirect
 	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.5.0 // indirect
+	tags.cncf.io/container-device-interface v0.8.0 // indirect
 )
 
+require github.com/testcontainers/testcontainers-go/modules/compose v0.33.0
```
wildum, line 986: nit: from a comment in the initial version, it would be better to have this dependency part of the main block instead of being isolated

wildum (approved):
just a nit, else lgtm :) It takes quite some time to build the alloy docker image, making it longer to run the tests but the beyla integration test is definitely needed and we can maybe make some optimizations in the future

internal/cmd/integration-tests/docker-compose.yaml
wildum: `version` in docker compose files is obsolete

.github/workflows/integration-tests.yml
marctc: I think @wildum changed it the original PR. @wildum Could you confirm?

internal/cmd/integration-tests/docker-compose.yaml
marctc: I think @wildum changed it the original PR. @wildum Could you confirm?

.github/workflows/integration-tests.yml
wildum: yeah but that was before paulin's change, now it should be `go-version-file: go.mod`

PR #3422, opened 2025-04-23
Update Alloy data collection topic to align with other topics

clayton-cornell:
The Alloy data collection topic needs to be updated to fix a few minor typos, clean up the language used, and align with other topics example:  https://grafana.com/docs/mimir/latest/configure/about-anonymous-usage-statistics-reporting/


dehaansa (approved):
LGTM.

PR #3448, opened 2025-04-25
Add a new mimir.alerts.kubernetes component

### LINKED ISSUE #504: Listen for AlertManagerConfig CRDs and configure Alertmanager in Grafana Mimir/Cortex
Similarly to https://github.com/grafana/agent/issues/1544, the agent could discover `AlertManagerConfig` CRDs to configure mimir's alert manager via its API

ptodev:

Fixes #504

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -715,6 +715,7 @@ require (
 	github.com/mistifyio/go-zfs v2.1.2-0.20190413222219-f784269be439+incompatible // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
+	github.com/mitchellh/hashstructure v1.1.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/mna/redisc v1.3.2 // indirect
 	github.com/moby/docker-image-spec v1.3.1 // indirect
@@ -792,7 +793,6 @@ require (
 	github.com/power-devops/perfstat v0.0.0-20240221224432-82ca36839d55 // indirect
 	github.com/prometheus-community/go-runit v0.1.0 // indirect
 	github.com/prometheus-community/prom-label-proxy v0.7.0 // indirect
-	github.com/prometheus/alertmanager v0.27.0 // indirect
 	github.com/prometheus/exporter-toolkit v0.14.0 // indirect
 	github.com/puzpuzpuz/xsync/v3 v3.5.1 // indirect
 	github.com/rcrowley/go-metrics v0.0.0-20201227073835-cf1acfcdf475 // indirect
@@ -941,6 +941,7 @@ require (
 require (
 	github.com/grafana/beyla/v2 v2.1.0-alloy-1
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/filestorage v0.122.0
+	github.com/prometheus/alertmanager v0.27.0
 	go.opentelemetry.io/collector/extension/xextension v0.122.1
 )
 
@@ -1026,3 +1027,6 @@ exclude (
 )
 
 replace github.com/prometheus/procfs => github.com/prometheus/procfs v0.12.0
+
+// TODO: Upstream this change prior to merging the mimir.alerts.kubernetes component to main.
+replace github.com/prometheus-operator/prometheus-operator => github.com/ptodev/prometheus-operator v0.0.0-20250425071154-e2ee3c49b414
--- a/internal/component/mimir/alerts/kubernetes/alerts.go
+++ b/internal/component/mimir/alerts/kubernetes/alerts.go
@@ -1,0 +1,398 @@
+package alerts
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/backoff"
+	alertmgr_cfg "github.com/prometheus/alertmanager/config"
+	coreListers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/workqueue"
+	_ "k8s.io/component-base/metrics/prometheus/workqueue"
+	controller "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/yaml"
+
+	"github.com/grafana/alloy/internal/component"
+	"github.com/grafana/alloy/internal/component/mimir/util"
+	"github.com/grafana/alloy/internal/featuregate"
+	mimirClient "github.com/grafana/alloy/internal/mimir/client"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/grafana/alloy/internal/service/cluster"
+
+	commonK8s "github.com/grafana/alloy/internal/component/common/kubernetes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+
+	promExternalVersions "github.com/prometheus-operator/prometheus-operator/pkg/client/informers/externalversions"
+	promListers_v1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/client/listers/monitoring/v1alpha1"
+	promVersioned "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+const (
+	configurationUpdate = "configuration-update"
+	clusterUpdate       = "cluster-update"
+)
+
+var (
+	errShutdown = errors.New("component is shutting down")
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:      "mimir.alerts.kubernetes",
+		Stability: featuregate.StabilityExperimental,
+		Args:      Arguments{},
+		Exports:   nil,
+		Build: func(o component.Options, c component.Arguments) (component.Component, error) {
+			return New(o, c.(Arguments))
+		},
+	})
+}
+
+type Component struct {
+	log  log.Logger
+	opts component.Options
+	args Arguments
+
+	mimirClient mimirClient.Interface
+
+	configUpdates  chan ConfigUpdate
+	clusterUpdates chan struct{}
+	leader         util.Leadership
+
+	healthMut sync.RWMutex
+	health    component.Health
+	metrics   *metrics
+
+	k8sClient            kubernetes.Interface
+	namespaceSelector    labels.Selector
+	cfgSelector          labels.Selector
+	alertmanagerSelector labels.Selector
+	eventProcessor       *eventProcessor
+	promClient           promVersioned.Interface
+}
+
+type ConfigUpdate struct {
+	args Arguments
+}
+
+var _ component.Component = (*Component)(nil)
+var _ component.DebugComponent = (*Component)(nil)
+var _ component.HealthComponent = (*Component)(nil)
+var _ cluster.Component = (*Component)(nil)
+
+// New creates a new Component and initializes required clients based on the provided configuration.
+func New(o component.Options, args Arguments) (*Component, error) {
+	c, err := newNoInit(o, args)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.init()
+	if err != nil {
+		return nil, fmt.Errorf("initializing component failed: %w", err)
+	}
+
+	return c, nil
+}
+
+func newNoInit(o component.Options, args Arguments) (*Component, error) {
+	m := newMetrics()
+	if err := m.register(o.Registerer); err != nil {
+		return nil, fmt.Errorf("registering metrics failed: %w", err)
+	}
+
+	clusterSvc, err := o.GetServiceData(cluster.ServiceName)
+	if err != nil {
+		return nil, fmt.Errorf("getting cluster service failed: %w", err)
+	}
+
+	c := &Component{
+		log:            o.Logger,
+		opts:           o,
+		args:           args,
+		leader:         util.NewComponentLeadership(o.ID, o.Logger, clusterSvc.(cluster.Cluster)),
+		configUpdates:  make(chan ConfigUpdate),
+		clusterUpdates: make(chan struct{}, 1),
+		metrics:        m,
+	}
+
+	return c, nil
+}
+
+func (c *Component) Run(ctx context.Context) error {
+	c.startupWithRetries(ctx, c.leader, c, c)
+
+	for {
+		// iteration only returns a sentinel error to indicate shutdown, otherwise it handles
+		// any errors encountered itself by logging and marking the component as unhealthy.
+		err := c.iteration(ctx, c.leader, c, c)
+		if errors.Is(err, errShutdown) {
+			break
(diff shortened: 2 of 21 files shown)


go.mod
```diff
@@ -1026,3 +1027,6 @@ exclude (
 )
 
 replace github.com/prometheus/procfs => github.com/prometheus/procfs v0.12.0
+
+// TODO: Upstream this change prior to merging the mimir.alerts.kubernetes component to main.
+replace github.com/prometheus-operator/prometheus-operator => github.com/ptodev/prometheus-operator v0.0.0-20250425071154-e2ee3c49b414
```
ptodev, line 1032: I will need to merge [a PR upstream](https://github.com/prometheus-operator/prometheus-operator/pull/7486) before I merge this feature.

vladst3f:
hi! we would interested in using this - what's the plan to merge it ?
thanks!

ptodev:
Hi @vladst3f! I'm still hopeful that this can go into the next Alloy release. However, I had a lot of trouble updating the go.mod file last time I tried to update the PR. There was a dependency which another one of Alloy's dependencies wanted on the latest version but other dependencies weren't compatible with the latest version. I might need to fork some dependencies in order to merge the PR.

ptodev:
I hope we will be able to merge this PR in August, shortly after Go 1.25 is released.

The reason why it has been delayed is due to "diamond dependencies". Alloy has three dependencies which depend on `k8s.io/` Go modules - Prometheus Operator, OTel Collector Contrib, and Prometheus itself. Prometheus Operator upgraded to v0.33 of `k8s.io/`, which also requires Go 1.24. However, OTel Collector Contrib and Prometheus can't upgrade to that version of k8s because they can't upgrade their Go version because they're commietd to supporting the last 2 versions of Go. This requires them to always be on the next to last version of Go. It seems that Prometheus was upgraded for a bit (probably accidentally) and then [reverted](https://github.com/prometheus/prometheus/commit/b2e0d3f16c70b4e6d0e29a6c05da792238d9b7bf). For OTel Collector there is [a PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39725) for the k8s upgrade and [an issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38970) for the Go upgrade.

The alternative to waiting is for me to make a fork of Prometheus Operator under the Grafana GitHub org, but I'm reluctant to do this due to the extra maintenance cost.

sebt3:
An august merge sound optimistic by now :(
We're waiting for : https://github.com/prometheus/prometheus/pull/16890
which itself wait for : https://github.com/prometheus/prometheus/pull/17090
After these 2 are merged, then the OTel MR can be merged, and after this one.

vladst3f:
@ptodev, @sebt3 , https://github.com/prometheus/prometheus/pull/17200 got merged last week. So can this progress now ? 

thanks !

sebt3:
@vladst3f I'm just a user willing to finish a migration (just like you probably). @ptodev mentionned https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39725 which is still not merged. So the waiting game is still outside of alloy. I'm pretty sure once all the external dependencies are finally merged will see this PR merged

ptodev:
Unfortunately, we do need [OTel to be upgraded](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39725) first. Otherwise there'd be compilation errors. For example, the OTel k8s attributes processor depends on `openshift/client-go` which also [needs to be updated](https://github.com/openshift/client-go/pull/340):
```
└─▪ go mod why -m github.com/openshift/client-go
# github.com/openshift/client-go
github.com/grafana/alloy/internal/component/otelcol/processor/k8sattributes
github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor
github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sconfig
github.com/openshift/client-go/quota/clientset/versioned
```

sebt3:
Finally : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43890 and https://github.com/openshift/client-go/pull/343

imho, all is set now :)

ron1:
Hi @ptodev , are you able to confirm all dependencies are now in place for this to be merged?

ptodev:
Hi folks, thank you for the reminders. I was able to rebase the branch and to build it successfully. I'll try to polish it and to open it for review by Tuesday/Wednesday next week. I'm optimistic that it'll make it to Alloy v1.12.0.

internal/component/mimir/util/util.go
```diff
@@ -0,0 +1,36 @@
+package util
```
kalleep, line 1: Any reason these live in a separate package? As far as I can tell they are only used by `mimir.rules.kubernetes`

internal/component/mimir/alerts/kubernetes/alerts.go
```diff
@@ -0,0 +1,349 @@
...
+	eventProcessor    *eventProcessor
+	promClient        promVersioned.Interface
+}
+
+type ConfigUpdate struct {
+	args Arguments
+}
+
+var _ component.Component = (*Component)(nil)
+var _ component.DebugComponent = (*Component)(nil)
```
kalleep: Not necessary because it's not implemented yet

internal/component/mimir/alerts/kubernetes/alerts.go
```diff
@@ -0,0 +1,349 @@
...
+func (c *Component) Shutdown() {
+	if c.eventProcessor != nil {
+		c.eventProcessor.stop()
+		c.eventProcessor = nil
+	}
+}
+
+// syncState asks the eventProcessor to sync rule state from the Mimir Ruler. It does
+// not block waiting for state to be synced.
+func (c *Component) SyncState() {}
```
kalleep: Is this needed?

internal/component/mimir/alerts/kubernetes/events.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		e.queue.Forget(evt)
+	}
+}
+
+// stop stops adding new Kubernetes events to the queue and blocks until all existing
+// events have been processed by the run loop.
+func (e *eventProcessor) stop() {
+	close(e.stopChan)
```
kalleep, line 156: A bit confusing and hard to follow but this will stop namespaceInformer and configInformer right?

I think it would be better to move the creation of those two into this component. Then it's more clear that this is the one owning the life cycle of those

kalleep (commented):
Good work! From an initial review I have some questions / suggestions

internal/component/mimir/alerts/kubernetes/alerts.go
```diff
@@ -0,0 +1,349 @@
...
+		Stability: featuregate.StabilityExperimental,
+		Args:      Arguments{},
+		Exports:   nil,
+		Build: func(o component.Options, c component.Arguments) (component.Component, error) {
+			return New(o, c.(Arguments))
+		},
+	})
+}
+
+type Component struct {
```
kalleep, line 57: Can you add some comments around how this component works and what is allowed to mutate properties?

I get that `Lifecycle` interface is used to add some tests but it also makes it really hard to read this component.

As far as I can tell there is only every one goroutine accessing  things like args and that update just sends on a channel to communicate things like a restart and re-initialize the component.

internal/component/mimir/util/util.go
ptodev: It is currently only used by `mimir.alerts.kubernetes`, but in future PRs I intend to get `mimir.rules.kubernetes` to reuse the same code and interfaces.

internal/component/mimir/alerts/kubernetes/events.go
ptodev: > this will stop namespaceInformer and configInformer right?

Yes. It's a valid point, but for now do you mind if I just add a TODO to work on it later? I want to simplify it for both `mimir.*.kubernetes` components at the same time.

internal/component/mimir/alerts/kubernetes/alerts.go
ptodev: I added a few more comments in this struct to indicate what some fields do. The whole component is WIP and very experimental. I agree that it's hard to read, but at the moment I'd like to keep it this way so that it's more consistent with `mimir.rules.kubernetes`. In subsequent PRs I could refactor them both at the same time so that they have simpler code that's shared between the two.

internal/component/mimir/alerts/kubernetes/alerts.go
ptodev: I changed this function to be used for a ticker which syncs mimir periodically. This is similar to `mimir.rules.kubernetes`. It's more useful in that other component, but for this one it also has benefits.

internal/component/mimir/alerts/kubernetes/alerts.go
ptodev: True, I removed this.

ptodev:
@kalleep I addressed some of the comments and also added an extra test. I agree that the PR could use more polishing, but I'd prefer to do big refactoring in subsequent PRs if you don't mind :)

kalleep (commented):
Thanks. I agree we can leave some work for future prs. Just one more comment

internal/component/mimir/alerts/kubernetes/alerts.go
```diff
@@ -0,0 +1,370 @@
...
+		configUpdates: make(chan ConfigUpdate),
+		metrics:       m,
+		ticker:        time.NewTicker(args.SyncInterval),
+	}
+
+	return c, nil
+}
+
+func (c *Component) Run(ctx context.Context) error {
+	c.startupWithRetries(ctx, c, c)
```
kalleep, line 133: Could this block forever?

If you have a bad config this would never start because we are doing infinite retries. We would also not get any args updates that would fix the issue because we are blocked here right?

ron1:
Any chance this might still make it into  the 1.12 release? https://github.com/grafana/alloy/issues/4699

internal/component/mimir/alerts/kubernetes/alerts.go
ptodev: That's true! I added another TODO. Thank you for spotting this. It's also a potential bug with `mimir.rules.kubernetes`. We could think of a solution that works for both codebases.

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Debug information
+
+`mimir.alerts.kubernetes` does not expose debug information.
+
+## Debug metrics
+
+| Metric Name                                    | Type        | Description                                                              |
+| ---------------------------------------------- | ----------- | ------------------------------------------------------------------------ |
+| `mimir_alerts_client_request_duration_seconds` | `histogram` | Duration of requests to the Mimir API.                                   |
```
clayton-cornell: Is this correct? or is it `mimir_alerts_mimir_client_request_duration_seconds`

The related source in events.go says: 

```
		mimirClientTiming: prometheus.NewHistogramVec(prometheus.HistogramOpts{
			Subsystem: "mimir_alerts",
			Name:      "mimir_client_request_duration_seconds",
```

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
```diff
@@ -0,0 +1,495 @@
...
+---
+
+# `mimir.alerts.kubernetes`
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`mimir.alerts.kubernetes` discovers `AlertmanagerConfig` Kubernetes resources and loads them into a Mimir instance.
+
+* You can specify multiple `mimir.alerts.kubernetes` components by giving them different labels.
+* [Kubernetes label selectors][] can be used to limit the `Namespace` and `AlertmanagerConfig` resources considered during reconciliation.
```
clayton-cornell:
```suggestion
* You can use [Kubernetes label selectors][] to limit the `Namespace` and `AlertmanagerConfig` resources considered during reconciliation.
```

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
```diff
@@ -0,0 +1,495 @@
...
+* Compatible with the `AlertmanagerConfig` CRD from the [`prometheus-operator`][prometheus-operator].
+* This component accesses the Kubernetes REST API from [within a Pod][].
+
+{{< admonition type="note" >}}
+This component requires [Role-based access control (RBAC)][] to be set up in Kubernetes in order for {{< param "PRODUCT_NAME" >}} to access it via the Kubernetes REST API.
+
+[Role-based access control (RBAC)]: https://kubernetes.io/docs/reference/access-authn-authz/rbac/
+{{< /admonition >}}
+
+`mimir.alerts.kubernetes` does not support [clustering][clustered mode].
```
clayton-cornell:
```suggestion
`mimir.alerts.kubernetes` doesn't support [clustering][clustered mode].
```

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
```diff
@@ -0,0 +1,495 @@
...
+
+`mimir.alerts.kubernetes` doesn't export any fields.
+
+## Component health
+
+`mimir.alerts.kubernetes` is reported as unhealthy if given an invalid configuration or an error occurs during reconciliation.
+
+## Debug information
+
+`mimir.alerts.kubernetes` does not expose debug information.
```
clayton-cornell:
```suggestion
`mimir.alerts.kubernetes` doesn't expose debug information.
```

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
```diff
@@ -0,0 +1,495 @@
...
+      - name: service
+        value: webapp
+  receivers:
+  - name: "null"
+  - name: "database-pager"
+```
+
+{{< /collapse >}}
+
+The Kubernetes configuration above will create the Alertmanager configuration below and send it to Mimir:
```
clayton-cornell:
```suggestion
The Kubernetes configuration above creates the Alertmanager configuration below and sends it to Mimir:
```

docs/sources/reference/components/mimir/mimir.alerts.kubernetes.md
ptodev: You are correct! Well spotted. I double checked it on the `/metrics` endpoint just in case:
```
mimir_alerts_mimir_client_request_duration_seconds_bucket{component_id="mimir.alerts.kubernetes.default",component_path="/",operation="POST /api/v1/alerts",status_code="201",le="0.005"} 0
mimir_alerts_mimir_client_request_duration_seconds_bucket{component_id="mimir.alerts.kubernetes.default",component_path="/",operation="POST /api/v1/alerts",status_code="201",le="0.01"} 46
```
I fixed the doc for `mimir.rules.kubernetes` too.

ptodev:
Hi @ron1, yes, this will be in v1.12. It wasn't in the first release candidate, but it will go in the second one.

ron1:
@ptodev Thanks much for your efforts to get this feature merged!

PR #3452, opened 2025-04-25
fix: panic that happens when a target gets deleted when using decompression

### LINKED ISSUE #3415: Alloy 1.8.1 - loki.source.file / local.file_match - decompressor gz - Panic
### What's wrong?

Hello,
I use local.file_match to detect .gz files. loki.source.file and decompressor to decode .gz files.
If I delete a gz file which is on the disk I get a panic.

### Steps to reproduce

Delete a .gz file which is monitored by local.file_match and loki.source.file

### System information

RHEL 7 - Appliance

### Software version

Grafana Alloy 1.8.1

### Configuration

```text
This is the part of my alloy config

local.file_match "filename" {

            // Dateien, die erkannt werden sollen
  path_targets = [
    {__path__ = "/var/log/filename.1.gz"},
    {__path__ = "/var/log/filename.2.gz"},
  ]

  sync_period = "10s"
}

//
// ==========================================================================================================================
//

loki.source.file "filename" {

  targets = local.file_match.filename.targets

  file_watch {
    min_poll_frequency = "1s"
    max_poll_frequency = "16s"
  }

  decompression {
    enabled       = true
    initial_delay = "10s"
    format        = "gz"
  }

  forward_to = [loki.relabel.filename.receiver]
}
```

### Logs

```text
panic: close of nil channel

goroutine 26318 [running]:
github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).updatePosition.func1()
        /drone/src/internal/component/loki/source/file/decompresser.go:190 +0x105
github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).updatePosition(0xc00cb0d800)
        /drone/src/internal/component/loki/source/file/decompresser.go:201 +0x267
created by github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).Run in goroutine 26315
        /drone/src/internal/component/loki/source/file/decompresser.go:176 +0x4d1
panic: close of nil channel

goroutine 38433 [running]:
github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).updatePosition.func1()
        /drone/src/internal/component/loki/source/file/decompresser.go:190 +0x105
github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).updatePosition(0xc00cb0d900)
        /drone/src/internal/component/loki/source/file/decompresser.go:201 +0x267
created by github.com/grafana/alloy/internal/component/loki/source/file.(*decompressor).Run in goroutine 26314
        /drone/src/internal/component/loki/source/file/decompresser.go:176 +0x4d1
```

kalleep:
#### PR Description
As reported in #3415 we get a panic when a file is remove if decompression is enabled, I also managed to get it sometime by when canceling the job so there is a race condition in here.

This happen after changes done in https://github.com/grafana/alloy/pull/3191 and https://github.com/grafana/alloy/pull/2428

`Run` is called in a [loop](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/runner.go#L63-L69) and every time its triggered we ended up spawning a goroutine for `updatePosition`. So when we deleted the file or quick alloy the task is canceled and they all tried to close `posdone`. Because we also used to set these to `nil` we could get a panic for closing nil channel or panic for closing already closed channel.

I have only managed to reproduce this when running with decompression. The decompression job reads and decompress the whole file and is then considered done, so it will be triggered again in the loop.

When running without decompression we tail the file instead so `Run` does not return until the job is canceled so we are not going to call run several times.

With these changes its no longer possible to Start, Stop and Start again. I was kinda possible to do that before but you would leak go-routines if you did and it is never used like that by the component.

Fixes: #3415

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/file/decompresser.go
+++ b/internal/component/loki/source/file/decompresser.go
@@ -317,7 +317,7 @@ func (d *decompressor) markPositionAndSize() error {
 
 func (d *decompressor) Stop() {
 	if !d.stopped.CompareAndSwap(false, true) {
-		// We halve already called Stop
+		// We have already called Stop
 		return
 	}
 
--- a/internal/component/loki/source/file/decompresser.go
+++ b/internal/component/loki/source/file/decompresser.go
@@ -61,7 +61,7 @@ type decompressor struct {
 	size     int64
 	cfg      DecompressionConfig
 
-	// initOnce must be called from `Run` and mut most be locked.
+	// initOnce must be called from `Run` and mut must be locked.
 	initOnce          func()
 	componentStopping func() bool
 
--- a/internal/component/loki/source/file/decompresser.go
+++ b/internal/component/loki/source/file/decompresser.go
@@ -52,6 +52,7 @@ type decompressor struct {
 
 	posAndSizeMtx sync.RWMutex
 
+	stopped *atomic.Bool
 	running *atomic.Bool
 
 	decoder *encoding.Decoder
@@ -62,11 +63,10 @@ type decompressor struct {
 
 	componentStopping func() bool
 
-	mut      sync.RWMutex
-	stopping bool
-	posquit  chan struct{} // used by the readLine method to tell the updatePosition method to stop
-	posdone  chan struct{} // used by the updatePosition method to notify when it stopped
-	done     chan struct{} // used by the readLine method to notify when it stopped
+	mut     sync.RWMutex
+	posquit chan struct{} // used by the readLine method to tell the updatePosition method to stop
+	posdone chan struct{} // used by the updatePosition method to notify when it stopped
+	done    chan struct{} // used by the readLine method to notify when it stopped
 }
 
 func newDecompressor(
@@ -109,11 +109,17 @@ func newDecompressor(
 		labels:            labels,
 		labelsStr:         labelsStr,
 		running:           atomic.NewBool(false),
+		stopped:           atomic.NewBool(false),
 		position:          pos,
+		posquit:           make(chan struct{}),
+		posdone:           make(chan struct{}),
 		decoder:           decoder,
 		cfg:               cfg,
 		componentStopping: componentStopping,
 	}
+
+	// updatePosition closes the channel t.posdone on exit
+	go decompressor.updatePosition()
 
 	return decompressor, nil
 }
@@ -156,24 +162,18 @@ func mountReader(f *os.File, logger log.Logger, format CompressionFormat) (reade
 }
 
 func (d *decompressor) Run() {
-	d.mut.Lock()
-
 	// Check if the stop function was called between two Run.
-	if d.stopping {
-		d.mut.Unlock()
+	if d.stopped.Load() {
 		return
 	}
 
 	labelsMiddleware := d.labels.Merge(model.LabelSet{filenameLabel: model.LabelValue(d.path)})
 	handler := loki.AddLabelsMiddleware(labelsMiddleware).Wrap(loki.NewEntryHandler(d.receiver.Chan(), func() {}))
 	defer handler.Stop()
-	d.posquit = make(chan struct{})
-	d.posdone = make(chan struct{})
+
+	d.mut.Lock()
 	d.done = make(chan struct{})
 	d.mut.Unlock()
-
-	// updatePosition closes the channel t.posdone on exit
-	go d.updatePosition()
 
 	d.metrics.filesActive.Add(1.)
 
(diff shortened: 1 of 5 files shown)


kalleep:
Even though we are not experiencing the same thing when running without decompression I should probably make similar changes there

internal/component/loki/source/file/decompresser.go
```diff
@@ -310,26 +309,22 @@ func (d *decompressor) markPositionAndSize() error {
 }
 
 func (d *decompressor) Stop() {
-	d.mut.Lock()
-	d.stopping = true
-	defer func() {
-		d.stopping = false
-	}()
-	d.mut.Unlock()
```
kalleep, line 318: I dont get why we did it like this here. If this gets called it it is from [here](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/runner.go#L60). So the task is done anyway but maybe I am missing something

internal/component/loki/source/file/decompresser.go
```diff
@@ -109,12 +109,18 @@ func newDecompressor(
...
 		position:          pos,
+		posquit:           make(chan struct{}),
+		posdone:           make(chan struct{}),
 		decoder:           decoder,
 		cfg:               cfg,
 		componentStopping: componentStopping,
 	}
 
+	// updatePosition closes the channel t.posdone on exit
+	go decompressor.updatePosition()
```
kalleep, line 122: Should not start it here because it a decompressor will be recreated on on every update to the component and then we check if the task already exists. Working on a fix to that

kalleep:
Props to @wildum for all the test, really helpful to see if changes leaks goroutines or causes deadlocks!

kalleep:
I will try to come up with a test that could reliable trigger the panic we get without the fix and add it to the pr

internal/component/loki/source/file/tailer.go
```diff
@@ -44,15 +44,17 @@ type tailer struct {
 	posAndSizeMtx sync.Mutex
 
 	running *atomic.Bool
+	stopped *atomic.Bool
```
ptodev, line 47: This struct has so many fields for controlling its lifetime. Could it be simplified so that we don't need all of them? If the code is simpler then there's less chance for bugs too...

internal/component/loki/source/file/decompresser.go
kalleep: Yes it should have locked the mutext when chaning the state. 

>Yes, it looks like in internal/component/loki/source/file/runner.go in the for loop r.reader.Run() could keep getting called until the backoff stops.

While this is true it never happens after stop is called because then the task exits.

internal/component/loki/source/file/decompresser.go
```diff
@@ -62,11 +63,10 @@ type decompressor struct {
 
 	componentStopping func() bool
 
-	mut      sync.RWMutex
-	stopping bool
-	posquit  chan struct{} // used by the readLine method to tell the updatePosition method to stop
-	posdone  chan struct{} // used by the updatePosition method to notify when it stopped
-	done     chan struct{} // used by the readLine method to notify when it stopped
+	mut     sync.RWMutex
```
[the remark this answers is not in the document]
kalleep: Yes `done` is still created on every call to `Run`

internal/component/loki/source/file/tailer.go
kalleep: We could merged these to into an `atomic.Int32` and call it state. Then have something like

```
const (
  stateRunning int32 = 1
  stateStopped int32 = 2
)
```

And transition the state instead if you think that would be better

internal/component/loki/source/file/decompresser.go
kalleep: Just a bit more context. `tailer`  and `decompressor` behaves a bit different.

When calling `Run` on a `tailer` it will start tailing the file and only exit when the job quits. The loop you mention will only happen if we encounter any errors during setup.

When calling `Run` on a `decompressor` we read and decompress the file and go through all lines from the position we have saved and then exit, but then that job will be retried again even though we have finished the file. 

So we don't want to spawn a go routine for `updatePosition` ever time we call run, just the first time. All other calls just makes sure we have a done channel set and read / tails the file. 

Stopped is only called when the task is canceled so we are never calling stop after that.

internal/component/loki/source/file/tailer.go
kalleep: Check [a2a5f5e](https://github.com/grafana/alloy/pull/3452/commits/a2a5f5e1a161bfbf61ce78cfc9ca2aade7053038). I changed to use what I proposed. 

We now have tree states, idle, running and stopped. And you can only transition from idle->running->stopped.

With this I am pretty sure we don't need the initOnce function, will check that. And now a task can only be triggered once so this is more like this component behaved before https://github.com/grafana/alloy/pull/2428 but we still "retry" for missing files.

PR #3454, opened 2025-04-25
Update to Prometheus 3.3 (& Loki 3.5)

### LINKED ISSUE #2306: Upgrade to Prometheus 3.0
[Prometheus 3.0](https://github.com/prometheus/prometheus/releases/tag/v3.0.0) was released in November 2024. Given that Alloy includes native Prometheus pipelines, it would be good to upgrade it sooner rather than later.
* There are breaking changes in the v3 release. If we want to upgrade without making an Alloy v2 release, we should find a way to not incorporate breaking changes.
* I believe there have been changes to the remote write specification. We need to take special care that the prometheus.remote_write component works in a backwards compatible way.
* We need to incorporate new config arguments. For example an Alloy user noted that `match_first_network` in `docker_sd_config` would be useful.

dehaansa:
#### PR Description
PR to update Prometheus to a fork of 3.3 with the staleness disabling feature we are still working to upstream.

Along the way, this required an update to the walqueue for prometheus 3.x support and an update to loki 3.5.

I'm opening this PR very early in the process to get CI rolling rather than running everything locally.

Fixes #2306

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -7,7 +7,7 @@ platform:
 steps:
 - commands:
   - make GO_TAGS="nodocker" test
-  image: grafana/alloy-build-image:v0.1.17
+  image: grafana/alloy-build-image:v0.1.18
   name: Run Go tests
 trigger:
   event:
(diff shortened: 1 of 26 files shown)


dehaansa:
Closing, as Piotr finished this work in another PR.

PR #3458, opened 2025-04-27
loki.process: Add labels_from_groups attribute to regex stage for automatic label creation

### LINKED ISSUE #3146: loki.process stage.regex autolabels
### Request

I have a `stage.match` block containing several `stage.regex` blocks, each of which has an `expression` containing one or more `?P<...>` named capture groups. I want all of those as labels, and for that I need to repeat myself by explicitly defining a `stage.labels` block which lists all of those capture group names. Instead I propose to introduce namespacing for capture groups which themselves signal to be exported as a label.

### Use case

For example:

```
stage.match {
  stage.regex {
    expression = "Hello (?P<english>\w+)"
  }
  stage.regex {
    expression = "Hallo (?P<german>\w+)"
  }
  stage.labels = {
    values = {
      english = "",
      german = "",
    }
  }
}
```
may become
```
stage.match {
  stage.regex {
    expression = "Hello (?P<autolabel:english>\w+)"
  }
  stage.regex {
    expression = "Hallo (?P<autolabel:german>\w+)"
  }
}
```
I would also be happy with some other mechanism which removes the burden to repeat oneself in a non-trivial way.

harshrai654:
#### PR Description
Add `labels_from_groups` attribute to `stage.regex` in `loki.process` to automatically add named capture groups as labels. This provides a convenient way to convert extracted regex groups into labels without additional configuration.

Fixes #3146 

#### Notes to the Reviewer
- Added `labels_from_groups` bool attribute to RegexConfig
- Updated regex stage processing to add matched groups as labels when enabled
- Added tests for the new functionality 
- Updated documentation with examples showing usage
- Added changelog entry under Enhancements

#### PR Checklist
- [x] CHANGELOG.md updated
- [x] Documentation added 
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1140,7 +1140,8 @@ The name of the capture group is then used as the key in the extracted map for t
 
 Because of how {{< param "PRODUCT_NAME" >}} syntax strings work, any backslashes in `expression` must be escaped with a double backslash, for example, `"\\w"` or `"\\S+"`.
 
-When `labels_from_groups` is set to `true`, any named capture groups from the regular expression are automatically added as labels in addition to being added to the extracted map.
+When `labels_from_groups` is set to true, any named capture groups from the regex expression are automatically added as labels in addition to being added to the extracted map.
+If a capture group name matches an existing label name, the existing label's value will be overridden by the extracted value.
 
 If the `source` is empty or missing, then the stage parses the log line itself.
 If it's set, the stage parses a previously extracted value with the same name.
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1127,16 +1127,20 @@ The `stage.regex` inner block configures a processing stage that parses log line
 
 The following arguments are supported:
 
-| Name         | Type     | Description                                                        | Default | Required |
-| ------------ | -------- | ------------------------------------------------------------------ | ------- | -------- |
-| `expression` | `string` | A valid RE2 regular expression. Each capture group must be named.  |         | yes      |
-| `source`     | `string` | Name from extracted data to parse. If empty, uses the log message. | `""`    | no       |
+| Name                 | Type     | Description                                                        | Default | Required |
+| -------------------- | -------- | ------------------------------------------------------------------ | ------- | -------- |
+| `expression`         | `string` | A valid RE2 regular expression. Each capture group must be named.  |         | yes      |
+| `source`             | `string` | Name from extracted data to parse. If empty, uses the log message. | `""`    | no       |
+| `labels_from_groups` | `bool`   | Whether to automatically add named capture groups as labels.       | false   | no       |
+
 
 The `expression` field needs to be a RE2 regular expression string.
 Every matched capture group is added to the extracted map, so it must be named like: `(?P<name>re)`.
 The name of the capture group is then used as the key in the extracted map for the matched value.
 
 Because of how {{< param "PRODUCT_NAME" >}} syntax strings work, any backslashes in `expression` must be escaped with a double backslash, for example, `"\\w"` or `"\\S+"`.
+
+When `labels_from_groups` is set to true, any named capture groups from the regex expression are automatically added as labels in addition to being added to the extracted map.
 
 If the `source` is empty or missing, then the stage parses the log line itself.
 If it's set, the stage parses a previously extracted value with the same name.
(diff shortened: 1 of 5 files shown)


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1125,17 +1125,21 @@ The `stage.regex` inner block configures a processing stage that parses log line
...
+| `labels_from_groups` | `bool`   | Whether to automatically add named capture groups as labels.       | false   | no       |
+
 
 The `expression` field needs to be a RE2 regular expression string.
 Every matched capture group is added to the extracted map, so it must be named like: `(?P<name>re)`.
 The name of the capture group is then used as the key in the extracted map for the matched value.
 
 Because of how {{< param "PRODUCT_NAME" >}} syntax strings work, any backslashes in `expression` must be escaped with a double backslash, for example, `"\\w"` or `"\\S+"`.
 
+When `labels_from_groups` is set to true, any named capture groups from the regex expression are automatically added as labels in addition to being added to the extracted map.
```
clayton-cornell, line 1141:
```suggestion
When `labels_from_groups` is set to `true`, any named capture groups from the regular expression are automatically added as labels in addition to being added to the extracted map.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1125,17 +1125,21 @@ The `stage.regex` inner block configures a processing stage that parses log line
...
 
-| Name         | Type     | Description                                                        | Default | Required |
-| ------------ | -------- | ------------------------------------------------------------------ | ------- | -------- |
-| `expression` | `string` | A valid RE2 regular expression. Each capture group must be named.  |         | yes      |
-| `source`     | `string` | Name from extracted data to parse. If empty, uses the log message. | `""`    | no       |
+| Name                 | Type     | Description                                                        | Default | Required |
+| -------------------- | -------- | ------------------------------------------------------------------ | ------- | -------- |
+| `expression`         | `string` | A valid RE2 regular expression. Each capture group must be named.  |         | yes      |
+| `source`             | `string` | Name from extracted data to parse. If empty, uses the log message. | `""`    | no       |
+| `labels_from_groups` | `bool`   | Whether to automatically add named capture groups as labels.       | false   | no       |
```
clayton-cornell, line 1132:
```suggestion
| `labels_from_groups` | `bool`   | Whether to automatically add named capture groups as labels.       | `false` | no       |
```

internal/component/loki/process/stages/regex.go
```diff
@@ -118,6 +119,9 @@ func (r *regexStage) Process(labels model.LabelSet, extracted map[string]interfa
 	for i, name := range r.expression.SubexpNames() {
 		if i != 0 && name != "" {
 			extracted[name] = match[i]
+			if r.config.LabelsFromGroups {
+				labels[model.LabelName(name)] = model.LabelValue(match[i])
```
wildum, line 123: to make this more robust I would suggest calling the IsValid function on both the labelName and the labelValue before adding it to the labels. In case one of this two is not valid, you can just have a debug log + continue

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1125,17 +1125,21 @@ The `stage.regex` inner block configures a processing stage that parses log line
...
+| `labels_from_groups` | `bool`   | Whether to automatically add named capture groups as labels.       | `false` | no       |
+
 
 The `expression` field needs to be a RE2 regular expression string.
 Every matched capture group is added to the extracted map, so it must be named like: `(?P<name>re)`.
 The name of the capture group is then used as the key in the extracted map for the matched value.
 
 Because of how {{< param "PRODUCT_NAME" >}} syntax strings work, any backslashes in `expression` must be escaped with a double backslash, for example, `"\\w"` or `"\\S+"`.
 
+When `labels_from_groups` is set to `true`, any named capture groups from the regular expression are automatically added as labels in addition to being added to the extracted map.
```
wildum, line 1141: On collision between label names, the existing label will be overridden by the extracted one. I think that's ok but we should document it and maybe have a debug log in case this happens. What do you think?

wildum (commented):
Thanks for picking this up!

docs/sources/reference/components/loki/loki.process.md
harshrai654: Oh yes, forgot about the collision case. Yes we should document and log such case.

wildum (approved):
Perfect, thanks!

PR #3463, opened 2025-04-28
fix github actions

kalleep:
Fix findings reported by https://github.com/woodruffw/zizmor

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,6 +1,9 @@
 name: Backport PR Creator
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers] In order to backport PRs from external
+    # contributors, this workflow needs to run against the base repo. The checkout action below is
+    # pinned to a specific commit to prevent arbitrary code execution by a fork that updates the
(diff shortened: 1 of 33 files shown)


ptodev (approved):

jharvey10:
The updates I'm least confident about are the `permissions:` ones. As a default, `contents: read` *should be fine*, but please double check me on that.

ptodev:
> The updates I'm least confident about are the `permissions:` ones.

@jharvey10 Are you worried that not all workflows are ran as part of PR checks, and it might break something without us noticing? I'd agree, but it's still worth trying out. We could change it later if it breaks such a workflow and there's no other way around it.

jharvey10:
> > The updates I'm least confident about are the `permissions:` ones.
> 
> @jharvey10 Are you worried that not all workflows are ran as part of PR checks, and it might break something without us noticing? I'd agree, but it's still worth trying out. We could change it later if it breaks such a workflow and there's no other way around it.

@ptodev Yeah, pretty much, haha. Worst case scenario would be trying to cut a release and having to battle against actions that were previously working. On the bright side, it’s pretty straightforward to adjust.

PR #3466, opened 2025-04-29
Fix streams limit error check so that metrics are correctly labeled as `ReasonStreamLimited`

maratkhv:
#### PR Description
This PR fixes an issue where such error check
`if err.Error() == errMaxStreamsLimitExceeded { reason = ReasonStreamLimited  }`
would always fail because batch.add() returns a formatted error, so the metrics are not labeled properly.
#### Notes to the Reviewer
- I believe using a regexp here would be overkill; a simple `HasPrefix` check is sufficient.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/client/batch.go
+++ b/internal/component/common/loki/client/batch.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/component/common/loki/client/batch.go
```diff
@@ -20,6 +20,14 @@ const (
 	errMaxStreamsLimitExceeded = "streams limit exceeded, streams: %d exceeds limit: %d, stream: '%s'"
 )
 
+func isErrMaxStreamsLimitExceeded(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return strings.HasPrefix(err.Error(), "streams limit exceeded")
```
kalleep, line 28: Instead of `errMaxStreamsLimitExceeded` being a string and us having to check a string prefix we should turn that into a proper error `errors.New("treams limit exceeded")`. Then in every place we return it we change to 
`fmt.Erorrf("%w, streams: %d exceeds limit: %d, stream: '%s'", errMaxStreamsLimitExceeded, streams, maxStream, lables)`.

Then the check for this error is just `errors.Is(err, errMaxStreamsLimitExceeded)`

kalleep (commented):
LGTM, we are sorting out some issues with out ci. but once those are resolved you will need to rebase and we can merge after that. Will ping you when everything is ready

PR #3475, opened 2025-04-30
fix: panic that happens when a target gets deleted when using decompression

kalleep:
#### PR Description
Alternative to https://github.com/grafana/alloy/pull/3452. In this pr I am taking a different approach.

Instead of having to call a `Stop` function on readers we instead accept a `context.Context` in `Run`. Run will now block until context is canceled or if it stopped for any other reason, for decompressor this happens when the whole file is read.
Run is also now responsible for cleanups and we don't have to use any locks or check for nil channels. This is all done as an attempt to simplify the logic a bit.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/runner.go
+++ b/internal/component/loki/source/file/runner.go
@@ -2,7 +2,6 @@ package file
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/grafana/alloy/internal/runner"
@@ -51,23 +50,11 @@ func (r *runnerReader) Run(ctx context.Context) {
 		},
 	)
 
-	var wg sync.WaitGroup
-	wg.Add(1)
-
-	go func() {
-		defer wg.Done()
-		<-ctx.Done()
-		r.reader.Stop()
-	}()
-
 	for {
-		r.reader.Run()
+		r.reader.Run(ctx)
 		backoff.Wait()
 		if !backoff.Ongoing() {
 			break
 		}
 	}
-
-	// Wait for the stop function to exit to ensure that the reader was properly stopped.
-	wg.Wait()
 }
--- a/internal/component/loki/source/file/tailer.go
+++ b/internal/component/loki/source/file/tailer.go
@@ -49,10 +49,6 @@ type tailer struct {
 	componentStopping func() bool
 
 	tail    *tail.Tail
-	posquit chan struct{} // used by the readLine method to tell the updatePosition method to stop
-	posdone chan struct{} // used by the updatePosition method to notify when it stopped
-	done    chan struct{} // used by the readLine method to notify when it stopped
-
 	decoder *encoding.Decoder
 }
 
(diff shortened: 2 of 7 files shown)


internal/component/loki/source/file/runner.go
```diff
@@ -51,23 +50,11 @@ func (r *runnerReader) Run(ctx context.Context) {
...
-
-	go func() {
-		defer wg.Done()
-		<-ctx.Done()
-		r.reader.Stop()
-	}()
-
 	for {
-		r.reader.Run()
+		r.reader.Run(ctx)
```
kalleep, line 54: This part is a bit problematic when using `decompressor` and potentially when using `tailer`. We will retry the until the target disappear. But when using `decompressor` we don't tail a file but read it fully until its finished. I wonder if we should quit them there then so we don't keep them around because its not necessary. 

I can think of a couple  of ways to do this, either set `MaxRetries` so we don't do this forever or return some kind of result from `Run` that can indicate if we should keep going or stop

internal/component/loki/source/file/tailer.go
```diff
@@ -47,12 +48,10 @@ type tailer struct {
...
-	mut      sync.RWMutex
-	stopping bool
-	tail     *tail.Tail
-	posquit  chan struct{} // used by the readLine method to tell the updatePosition method to stop
-	posdone  chan struct{} // used by the updatePosition method to notify when it stopped
-	done     chan struct{} // used by the readLine method to notify when it stopped
+	tail    *tail.Tail
+	posquit chan struct{} // used by the readLine method to tell the updatePosition method to stop
+	posdone chan struct{} // used by the updatePosition method to notify when it stopped
+	done    chan struct{} // used by the readLine method to notify when it stopped
```
ptodev, line 54: I wonder if all these channels have to be struct fields and whether all the goroutines have to be started in Run... it is harder to see what depends on what this way. WDYT about e.g. `readLines` starting the `updatePosition` goroutine and closing it before it closes?

ptodev (commented):
Thank you, this is definitely a step in the right direction!

internal/component/loki/source/file/tailer.go
```diff
@@ -47,12 +48,10 @@ type tailer struct {
 
 	componentStopping func() bool
```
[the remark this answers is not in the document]
kalleep: This one is passed by `loki.source.file` component:
* https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/file.go#L293
* https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/file.go#L315

It is then used by both readers to determine what do do when stopping:
https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/tailer.go#L392

So yes this till works

internal/component/loki/source/file/tailer.go
kalleep: I was experimenting a bit with this before, not having them as struct fields but we have to pass them around.

I made it so that both readers works the same.

Run starts both `updatePosition` and `readlines`. When it stops either by it self or by context being canceled we stop `readlines` and wait for it. That one will close `posquit` that in turn. 

I can try to think if there is a better way to structure this

kalleep: [8e5e4b4](https://github.com/grafana/alloy/pull/3475/commits/8e5e4b415bc1826c5039e631c2ee507c16506d44) is an attempt to do this, wdyt?

wildum (approved):

PR #3483, opened 2025-05-01
Update prometheus-metrics.md

hppycoder:
Found that the endpoint is not available in traditional prometheus, it instead exists in Mimir which is on port 9009 (not 9090)

#### PR Description

When working through the documentation I found that the `/api/prom/push` does not exist on the traditional Prometheus service. Instead it exists on the Mimir one which is on the default port of 9009.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/collect/prometheus-metrics.md
+++ b/docs/sources/collect/prometheus-metrics.md
@@ -76,7 +76,7 @@ The following example demonstrates configuring `prometheus.remote_write` with mu
 ```alloy
 prometheus.remote_write "default" {
   endpoint {
-    url = "http://localhost:9090/api/prom/push"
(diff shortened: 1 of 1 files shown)


clayton-cornell (approved):
Thanks for catching this.  I'll get it packaged up and merged right away.

hppycoder:
Thanks Clayton! It wasn't all for not, learned quite a bit more about Prometheus in the process.

clayton-cornell:
@hppycoder Looks like you haven't signed the CLA yet. If you can click through on that then we can merge the PR :-)

hppycoder:
@clayton-cornell Signed it now, missed it in the threads earlier

PR #3484, opened 2025-05-01
ci: group go.opentelemetry.io deps together with other otel deps

jharvey10:
Group go.opentelemetry.io updates into the same PR group as github.com/open-telemetry/opentelemetry-collector-contrib

This should reduce a lot of the size of the non-open telemetry minor/patch updates.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -61,7 +61,7 @@
       "matchPackageNames": [
         // OpenTelemetry needs special handling due to a temporary fork
         "github.com/open-telemetry/opentelemetry-collector-contrib/**",
-        "go.opentelemetry.io/collector"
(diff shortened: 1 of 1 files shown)


ptodev (commented):
Do you think we should exclude otel dependencies from renovate altogether? We only update them manually anyway. Certainly OTel component updates need to be manual.

jharvey10:
@ptodev That's effectively what this PR will do :) The `"dependencyDashboardApproval": true` line will not allow PRs for these deps to get auto-created. They'll still show up on the dependency dashboard grouped together and show us what versions are *available*, but it won't take any action on them unless we initiate it.

dehaansa (approved):

Issue #3456, opened 2025-04-26
Log collection stuck, http request to /api/v0/web/components blocked, no logs ingested


thoro:
### What's wrong?

Hello,

I'm running alloy 1.8.2 (and previous versions to debug), in 2 of 8 nodes alloy gets stuck after a few minutes. It also does not recover after a few hours of runtime.

There are no error messages in the logs. But logs are still generated, for example:

```
ts=2025-04-26T14:25:41.171363493Z level=info msg="skipping update of position for a file which does not currently exist" component_path=/ component_id=loki.source.file.pods component=tailer
```

Creation of the support package hangs indefinitly.
Requests to some endpoints hang indefinitly, for example /api/v0/web/components. Edit: Config reload hangs aswell

Attached is a goroutines and profile pprof extract of the affected node. [profiles.zip](https://github.com/user-attachments/files/19922924/profiles.zip)

Alloy is running as daemonset, collecting local logs from files directly.

Please advise for further debug steps.

### Steps to reproduce

unclear, happens on 2 of 8 nodes, probably some environmental side effect.

### System information

Rocky Linux 9.4 5.14.0-427.24.1.el9_4.x86_64

### Software version

Grafana Alloy 1.8.2

### Configuration

```text
discovery.kubernetes "pods" {
      role = "pod"

      // Filter only pods running on the current node (when run as daemonset)
      selectors {
        role  = "pod"
        field = "spec.nodeName=" + coalesce(env("HOSTNAME"), constants.hostname)
      }
    }

    discovery.relabel "pods" {
      targets = discovery.kubernetes.pods.targets

      // Basic namespace/pod/container rules
      // Adding the node is useful to track restarting daemonsets on a specific node
      // otherwise it's not possible to group logs by node, and this might be an issue when debugging node related issues.
      rule {
        action = "replace"
        source_labels = ["__meta_kubernetes_pod_node_name"]
        target_label = "node"
      }

      rule {
        action = "replace"
        source_labels = ["__meta_kubernetes_namespace"]
        target_label = "namespace"
      }

      rule {
        action = "replace"
        source_labels = ["__meta_kubernetes_pod_name"]
        target_label = "pod"
      }

      rule {
        action = "replace"
        source_labels = ["__meta_kubernetes_pod_container_name"]
        target_label = "container"
      }

      // Extended app.kubernetes.io rules (instance, role, component, app)
      rule {
        action = "replace"
[...]

Issue #3472, opened 2025-04-29
Missing Pod Logs Not Ingested by Alloy


nickv2002:
### What's wrong?

We use the [k8s-monitoring-helm chart](https://github.com/grafana/k8s-monitoring-helm) with the in the config pated below, chart version is `1.6.33` when we noticed some pod logs were missing from our Grafana Cloud instance. We were able to get the with `kubectl logs` as well as by looking for the log file from the host node. Investigation revealed many messages in the Grafana Alloy logs similar to the one below pointing at `skipping update of position for a file which does not currently exist` for pods that no longer existed. We didn't find any messages about processing logs for the pod that was missing from Grafana Cloud (Loki).

```log
ts=2025-04-29T19:47:35.077367632Z level=info msg="skipping update of position for a file which does not currently exist" component_path=/ component_id=loki.source.file.pod_logs component=tailer path=/var/log/pods/b2b_b2b-api-56896ff857-lsvdh_10cd50b7-9f34-44b9-b60c-8ae24816932a/flyway-migrate/0.log
```

We updated our Helm chart to point at `1.6.34` and then did a rolling restart of all the Alloy pods (I guess it didn't happen automatically because Helm didn't calculate a difference). This fixed the problem and we were able to see our pod logs in Grafana cloud again.

This kind of shakes our trust in Alloy since it was failing to send logs but not in a visible/alert-able way. It wasn't all Alloy instances that were broken, just on one of dozen or so nodes. We wouldn't have noticed if not for a dev hadn't been asking where their logs were.

Questions:

1. Any tips for how to detect/avoid this going forward?
2. Any configuration changes we could make to Alloy to improve things?

### Steps to reproduce

Unknown

### System information

EKS Kubernetes running docker.io/grafana/alloy:v1.8.1

### Software version

 docker.io/grafana/alloy:v1.8.1

### Configuration

```text
cluster:
  name: qa
externalServices:
  prometheus:
    host: https://prometheus-prod-13-prod-us-east-0.grafana.net
    basicAuth:
      username: <redated>
      password: &grafanaAPIpw <redated>
  loki:
    host: https://logs-prod-006.grafana.net
    basicAuth:
      username: <redated>
      password: *grafanaAPIpw
  tempo:
    host: https://tempo-prod-04-prod-us-east-0.grafana.net
    basicAuth:
      username: <redated>
      password: *grafanaAPIpw
kube-state-metrics:
  metricLabelsAllowlist:
    - pods=[*]
[...]

PR #3488, opened 2025-05-02
fix: deadlock in loki.source.file when target is removed

kalleep:
#### PR Description
https://github.com/grafana/alloy/issues/3456 reports that some alloy nodes stopped sending logs. Alloy itself still produced some logs looking like this:
```
ts=2025-04-26T14:25:41.171363493Z level=info msg="skipping update of position for a file which does not currently exist" component_path=/ component_id=loki.source.file.pods component=tailer
```

Trying to access `/api/v0/web/components` hangs and the profile shows that most go routines are parked. This is most likely symptoms of a deadlock.

While going through the changes we have made to this component I noticed [this](https://github.com/grafana/alloy/pull/2428/files#diff-0123a6c4b122c3cd9498a53ae22da4195fa5f122009e30a38fbcad5700246925L165-L185). 

So my current hypothesis is that when we are stopping tasks we wait for them to finish, that includes flushing any logs that have not been sent over the handler channel yet. Because nothing is reading from that channel when we are performing this update we get stuck waiting there indefinitely. 

To fix this we need to keep on reading while we perform the update and stop that when we are done

Potentially fix:
https://github.com/grafana/alloy/issues/3456
https://github.com/grafana/alloy/issues/3472

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/loki/source/file/file.go
```diff
@@ -81,8 +81,6 @@ type Component struct {
 	opts    component.Options
 	metrics *metrics
 
-	updateMut sync.Mutex
```
kalleep, line 84: This was only used in `Update` so we don't need it

wildum (approved):

Issue #3045, opened 2025-03-20
Dedup exporters and connectors in the otel converter


wildum:
When the same exporter/connector is used in different pipelines, we currently duplicate it instead of reusing it:

```
receivers:
  otlp:
    protocols:
      grpc:
      http:

processors:
  batch:

exporters:
  otlp:
    endpoint: database:4317

service:
  pipelines:
    metrics: # Group <empty>
      receivers: [otlp]
      processors: []
      exporters: [otlp]
    metrics/2: # Group 2
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
```

In this case, there should only be one otlp exporter generated in Alloy.

This is what the Otel collector does: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/graph/graph.go#L236

PR #3489, opened 2025-05-02
Upgrade OpenTelemetry components from v0.122.0 to v0.125.0

ptodev:
#### PR Description

Routine update for the next Alloy release.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/converter/internal/otelcolconvert/testdata/spanmetrics_full.alloy
ptodev, line 1: I don't know why this test file is failing. If I only remove the `.alloy` file, the test suite spits out the `.alloy` file which it expects. But if I include this alloy file, it panics. 

In any case, the `.alloy` file we currently have doesn't make any sense. There is a `otelcol.connector.spanmetrics "_2_default"` which isn't receiving any traces from any other component.

internal/converter/internal/otelcolconvert/testdata/loadbalancing.alloy
ptodev, line 1: I don't know why the `sizer` and `retry_on_failure` differences occur. Those parameters do have default values in the upstream collector.

internal/converter/internal/otelcolconvert/testdata/kafka.alloy
```diff
@@ -59,11 +62,25 @@ otelcol.exporter.kafka "default" {
 	resolve_canonical_bootstrap_servers_only = true
 	client_id                                = "otelcol"
 	topic_from_attribute                     = "my_topic"
-	encoding                                 = "otlp_json"
 	partition_traces_by_id                   = true
 	partition_metrics_by_resource_attributes = true
 	timeout                                  = "11s"
 
+	logs {
```
ptodev, line 69: I'm not sure why the `logs`, `metrics`, and `traces` blocks have to be set explicitly. They do have default values in the collector.

internal/component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -93,6 +93,8 @@ func (args *Arguments) SetToDefault() {
 	}
 	args.DebugMetrics.SetToDefault()
 	args.Protocol.OTLP.SetToDefault()
+	args.Retry.SetToDefault()
+	args.Queue.SetToDefault()
```
ptodev, line 97: The fact that no default values were set must be an oversight. According to the Alloy documentation there should be default values for those two blocks.

go.mod
```diff
@@ -1026,3 +1030,13 @@ replace github.com/prometheus/procfs => github.com/prometheus/procfs v0.12.0
...
+
+replace github.com/grafana/beyla/v2 => github.com/ptodev/beyla/v2 v2.0.0-20250507083427-93b2af475c8d
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib => github.com/ptodev/opentelemetry-collector-contrib v0.0.0-20250506154750-58322eed65ce
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter => github.com/ptodev/opentelemetry-collector-contrib/exporter/kafkaexporter v0.0.0-20250506154750-58322eed65ce
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver => github.com/ptodev/opentelemetry-collector-contrib/receiver/kafkareceiver v0.0.0-20250506154750-58322eed65ce
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib/internal/kafka => github.com/ptodev/opentelemetry-collector-contrib/internal/kafka v0.0.0-20250506154750-58322eed65ce
```
ptodev, line 1042: I'll open PRs upstream to get those issues resolved.

internal/converter/internal/otelcolconvert/testdata/kafka.alloy
wildum: This happens because the logs, metrics, and traces blocks are actually nil by default in Alloy. The defaults are set in the convert function. This should also be the case for the receiver but the code is missing in the converter. 
I think that this is ok, I can make a commit to fix the converter for the kafka receiver and IMO we can keep the conversion like this. Once the topic arg is gone it will go away

internal/converter/internal/otelcolconvert/testdata/loadbalancing.alloy
wildum: The reason is that the default are not set upstream at this level, they are only set within the protocol. See: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/loadbalancingexporter/factory.go#L42

internal/converter/internal/otelcolconvert/testdata/spanmetrics_full.alloy
wildum: there was a problem in the validation function of the component somehow leading to the component not being correctly built, leading somehow to a panic. I updated the code.

But as you say, the alloy file is weird and this looks probably like a bug in the converter? Crazy that it's been like this since the beginning almost. The receiver is only connecting to one connector. I will create a bug ticket for it, no need to solve it in this PR

wildum: here is the ticket: https://github.com/grafana/alloy/issues/3537 I will investigate

go.mod
ptodev: We'll need to wait for a [Beyla](https://github.com/grafana/beyla/pull/1904) and an [OTel](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39946) PRs to be merged.

internal/converter/internal/otelcolconvert/testdata/kafka.alloy
ptodev: Thank you! Makes me wonder if the otel converter tests should instead rely in a `Convert` function for each `Arguments` struct, and then they could use the actual OTel config produced by the component.

internal/component/otelcol/exporter/loadbalancing/loadbalancing.go
ptodev: As @wildum mentioned in [another comment](https://github.com/grafana/alloy/pull/3489#discussion_r2079717638), this is actually intentional.

internal/converter/internal/otelcolconvert/testdata/loadbalancing.alloy
ptodev: But the LB exporter doesn't set defaults for `sending_queue`. So I can't explain to myself why the converter thinks Alloy doesn't use the 0 values.

internal/converter/internal/otelcolconvert/testdata/spanmetrics_full.alloy
wildum: I investigated a bit and I think that deduplicating the connectors will solve the problem: https://github.com/grafana/alloy/issues/3045

wildum: now that the converter does not panic anymore, I would suggest to keep these files in actually

docs/sources/reference/components/otelcol/otelcol.exporter.awss3.md
```diff
@@ -42,24 +42,30 @@ Replace the following:
 
 ## Arguments
 
-`otelcol.exporter.awss3` supports no arguments and is configured completely through inner blocks.
+`otelcol.exporter.otlp` supports the following arguments:
```
dehaansa, line 45:
```suggestion
`otelcol.exporter.awss3` supports the following arguments:
```

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -107,6 +93,33 @@ For example, `authentication > tls` refers to a `tls` block defined inside an `a
...
+The `logs` block configures how to send logs to Kafka brokers.
+
+Name       | Type     | Description                                                                | Default        | Required
+---------- | -------- | -------------------------------------------------------------------------- | -------------- | --------
+`topic`    | `string` | The name of the Kafka topic to which logs will be exported.                | `"otlp_logs"`  | no
+`encoding` | `string` | The encoding for logs. See [Supported encodings](#supported-encodings).    | `"otlp_proto"` | no
+
+### metrics block
+
+The `logs` block configures how to send metrics to Kafka brokers.
```
dehaansa, line 107:
```suggestion
The `metrics` block configures how to send metrics to Kafka brokers.
```

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -107,6 +93,33 @@ For example, `authentication > tls` refers to a `tls` block defined inside an `a
...
+
+Name       | Type     | Description                                                                | Default           | Required
+---------- | -------- | -------------------------------------------------------------------------- | ----------------- | --------
+`topic`    | `string` | The name of the Kafka topic to which metrics will be exported.             | `"otlp_metrics"`  | no
+`encoding` | `string` | The encoding for logs. See [Supported encodings](#supported-encodings).    | `"otlp_proto"`    | no
+
+### traces block
+
+The `logs` block configures how to send traces to Kafka brokers.
```
dehaansa, line 116:
```suggestion
The `traces` block configures how to send traces to Kafka brokers.
```

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -42,8 +42,8 @@ Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `brokers` | `array(string)` | Kafka brokers to connect to. | | yes
 `protocol_version` | `string` | Kafka protocol version to use. | | yes
-`topic` | `string` | Kafka topic to read from. | _See below_ | no
-`encoding` | `string` | Encoding of payload read from Kafka. | `"otlp_proto"` | no
+`topic` | `string` | (Deprecated) Kafka topic to read from. | `""` | no
+`encoding` | `string` | (Deprecated) Encoding of payload read from Kafka. | `""` | no
```
dehaansa, line 46: Deprecated in favor of what?

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -128,6 +130,34 @@ The `>` symbol indicates deeper levels of nesting. For example,
...
+The `logs` block configures how to receive logs from Kafka brokers.
+
+Name       | Type     | Description                                                                | Default        | Required
+---------- | -------- | -------------------------------------------------------------------------- | -------------- | --------
+`topic`    | `string` | The name of the Kafka topic on which logs will be received.                | `"otlp_logs"`  | no
+`encoding` | `string` | The encoding for logs. See [Supported encodings](#supported-encodings).    | `"otlp_proto"` | no
+
+### metrics block
+
+The `logs` block configures how to receive metrics from Kafka brokers.
```
dehaansa, line 145:
```suggestion
The `metrics` block configures how to receive metrics from Kafka brokers.
```

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -128,6 +130,34 @@ The `>` symbol indicates deeper levels of nesting. For example,
...
+
+Name       | Type     | Description                                                                | Default           | Required
+---------- | -------- | -------------------------------------------------------------------------- | ----------------- | --------
+`topic`    | `string` | The name of the Kafka topic on which metrics will be received.             | `"otlp_metrics"`  | no
+`encoding` | `string` | The encoding for logs. See [Supported encodings](#supported-encodings).    | `"otlp_proto"`    | no
+
+### traces block
+
+The `logs` block configures how to receive traces from Kafka brokers.
```
dehaansa, line 154:
```suggestion
The `traces` block configures how to receive traces from Kafka brokers.
```

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -253,6 +283,35 @@ If `max_elapsed_time` is set to `"0s"`, failed requests are retried forever unti
...
+
+- `raw`: the payload's bytes are inserted as the body of a log record.
+- `text`: the payload are decoded as text and inserted as the body of a log record. By default, it uses UTF-8 to decode. You can use `text_<ENCODING>`, like `text_utf-8`, `text_shift_jis`, etc., to customize this behavior.
+- `json`: the payload is decoded as JSON and inserted as the body of a log record.
+- `azure_resource_logs`: the payload is converted from Azure Resource Logs format to OTel format.
+
+## Message header propagation
+
+`otelcol.receiver.kafka` will extract Kafka message headers and include them as request metadata (context).
+This metadata can then be used throughout the pipeline, for example to set attributes using `otelcol.processor.attributes`.
```
dehaansa, line 313: We should have an example of this.

internal/component/otelcol/exporter/kafka/kafka.go
```diff
@@ -49,23 +66,68 @@ type Arguments struct {
...
+func (c *KafkaExporterSignalConfig) convert(topic, encoding deprecatedAttr) kafkaexporter.SignalConfig {
+	result := kafkaexporter.SignalConfig{}
+
+	if c != nil { // Use values from the new block if set.
+		if len(c.Topic) > 0 {
+			result.Topic = c.Topic
+		}
+		if len(c.Encoding) > 0 {
+			result.Encoding = c.Encoding
+		}
```
dehaansa, line 110: If the block is not nil, should we pass through and potentially use the deprecated field defaults?

ptodev: No. The intent here is to always use the block's explicit or default values if the block is explicitly defined. That way we could document it in a simple way - if the block exists, use the block values. It shouldn't matter if the defaults or if the explicit values are used.

CHANGELOG.md
```diff
@@ -88,6 +91,32 @@ Main (unreleased)
...
+  - [`otelcol.connector.spanmetrics`] A new `include_instrumentation_scope` configuration argument.
+  - [`otelcol.connector.spanmetrics`] Initialise new `calls_total` metrics at 0.
+  - [`otelcol.connector.spanmetrics`] A new `aggregation_cardinality_limit` configuration argument 
+    to limit the number of unique combinations of dimensions that will be tracked for metrics aggregation.
+  - [`otelcol.connector.spanmetrics`] Deprecate the unused argument `dimensions_cache_size`.
+  - [`otelcol.connector.spanmetrics`] Moving the start timestamp (and last seen timestamp) from the resourceMetrics level to the individual metrics level. 
+  - [`otelcol.processor.k8sattributes`] Add option to configure automatic resource attributes - with annotation prefix.
+    Implements [Specify resource attributes using Kubernetes annotations](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#specify-resource-attributes-using-kubernetes-annotations).
+    This will ensure that each metric has its own accurate start and last seen timestamps, regardless of its relationship to other spans.
```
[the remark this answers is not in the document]
ptodev: I think so. Why would it not be?

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
ptodev: I added a deprecation note under the table to say it's deprecated in favour of the `logs`, `metrics`, and `traces` arguments.

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
ptodev: Is it really necessary? Even the upstream kafka doc doesn't have such an example. If I do add it, it'd have to look something [like this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/attributesprocessor/testdata/config.yaml#L344). But I don't feel good about adding an example without testing it, and it'd take me a while to test it.

docs/sources/reference/components/otelcol/otelcol.processor.resourcedetection.md
```diff
@@ -667,10 +667,11 @@ Block                                  | Description
...
 [host.id][res-attr-cfg]                | Toggles the `host.id` resource attribute. <br> Sets `enabled` to `false` by default.                | no
 [host.ip][res-attr-cfg]                | Toggles the `host.ip` resource attribute. <br> Sets `enabled` to `false` by default.                | no
-[host.mac][res-attr-cfg]               | Toggles the `host.mac` resource attribute. <br> Sets `enabled` to `false` by default.                | no
+[host.mac][res-attr-cfg]               | Toggles the `host.mac` resource attribute. <br> Sets `enabled` to `false` by default.               | no
 [host.name][res-attr-cfg]              | Toggles the `host.name` resource attribute. <br> Sets `enabled` to `true` by default.               | no
 [os.description][res-attr-cfg]         | Toggles the `os.description` resource attribute. <br> Sets `enabled` to `false` by default.         | no
 [os.type][res-attr-cfg]                | Toggles the `os.type` resource attribute. <br> Sets `enabled` to `true` by default.                 | no
+[os.version][res-attr-cfg]             | Toggles the `os.version` resource attribute. <br> Sets `enabled` to `false` by default.             | no
```
clayton-cornell, line 674: This table has been reformatted in https://github.com/grafana/alloy/pull/3583 

the HTML `<br>` should not be used. The table has been modified in the previous PR to remove the `<br>` tag, and format the table style

The updated table would look like this (with the new `os.version` row):

```
| Block                                    | Description                                                                                    | Required |
| ---------------------------------------- | ---------------------------------------------------------------------------------------------- | -------- |
| [`host.arch`][res-attr-cfg]              | Toggles the `host.arch` resource attribute. Sets `enabled` to `false` by default.              | no       |
| [`host.cpu.cache.l2.size`][res-attr-cfg] | Toggles the `host.cpu.cache.l2.size` resource attribute. Sets `enabled` to `false` by default. | no       |
| [`host.cpu.family`][res-attr-cfg]        | Toggles the `host.cpu.family` resource attribute. Sets `enabled` to `false` by default.        | no       |
| [`host.cpu.model.id`][res-attr-cfg]      | Toggles the `host.cpu.model.id` resource attribute. Sets `enabled` to `false` by default.      | no       |
| [`host.cpu.model.name`][res-attr-cfg]    | Toggles the `host.cpu.model.name` resource attribute. Sets `enabled` to `false` by default.    | no       |
| [`host.cpu.stepping`][res-attr-cfg]      | Toggles the `host.cpu.stepping` resource attribute. Sets `enabled` to `false` by default.      | no       |
| [`host.cpu.vendor.id`][res-attr-cfg]     | Toggles the `host.cpu.vendor.id` resource attribute. Sets `enabled` to `false` by default.     | no       |
| [`host.id`][res-attr-cfg]                | Toggles the `host.id` resource attribute. Sets `enabled` to `false` by default.                | no       |
| [`host.ip`][res-attr-cfg]                | Toggles the `host.ip` resource attribute. Sets `enabled` to `false` by default.                | no       |
| [`host.mac`][res-attr-cfg]               | Toggles the `host.mac` resource attribute. Sets `enabled` to `false` by default.               | no       |
| [`host.name`][res-attr-cfg]              | Toggles the `host.name` resource attribute. Sets `enabled` to `true` by default.               | no       |
| [`os.description`][res-attr-cfg]         | Toggles the `os.description` resource attribute. Sets `enabled` to `false` by default.         | no       |
| [`os.type`][res-attr-cfg]                | Toggles the `os.type` resource attribute. Sets `enabled` to `true` by default.                 | no       |
| [`os.version`][res-attr-cfg]             | Toggles the `os.version` resource attribute. Sets `enabled` to `false` by default.             | no       |
```

CHANGELOG.md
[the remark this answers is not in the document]
dehaansa: Doesn't it belong to the spanmetrics changelog on 107?

internal/component/otelcol/exporter/kafka/kafka.go
dehaansa: The current behavior appears to fall through and use the default value of the deprecated fields if the provided values are len 0

docs/sources/shared/reference/components/otelcol-kafka-metadata.md
```diff
@@ -7,12 +7,12 @@ The `metadata` block configures how to retrieve and store metadata from the Kafk
 
 The following arguments are supported:
 
-Name                 | Type   | Description                                   | Default | Required
----------------------|--------|-----------------------------------------------|---------|---------
-`include_all_topics` | `bool` | When true, maintains metadata for all topics. | `true`  | no
+Name               | Type       | Description                                           | Default  | Required
+-------------------|------------|-------------------------------------------------------|----------|---------
+`full`             | `bool`     | Whether to maintain a full set of metadata.           | `true`   | no
+`refresh_interval` | `duration` | The frequency at which cluster metadata is refreshed. | `"10m"`  | no
```
clayton-cornell, line 13:
```suggestion
| Name               | Type       | Description                                           | Default | Required |
| ------------------ | ---------- | ----------------------------------------------------- | ------- | -------- |
| `full`             | `bool`     | Whether to maintain a full set of metadata.           | `true`  | no       |
| `refresh_interval` | `duration` | The frequency at which cluster metadata is refreshed. | `"10m"` | no       |
```

docs/sources/shared/reference/components/otelcol-queue-block.md
```diff
@@ -6,13 +6,19 @@ headless: true
...
+| `block_on_overflow` | `boolean`                  | The behavior when the component's `TotalSize` limit is reached.                            | `false`       | no       |
+| `enabled`           | `boolean`                  | Enables a buffer before sending data to the client.                                        | `true`        | no       |
+| `num_consumers`     | `number`                   | Number of readers to send batches written to the queue in parallel.                        | `10`          | no       |
+| `queue_size`        | `number`                   | Maximum number of unwritten batches allowed in the queue at the same time.                 | `1000`        | no       |
+| `sizer`             | `string`                   | How the queue and batching is measured.                                                    | `"requests"`  | no       |
+| `storage`           | `capsule(otelcol.Handler)` | Handler from an `otelcol.storage` component to use to enable a persistent queue mechanism. |               | no       |
```
clayton-cornell, line 17:
```suggestion
| Name                | Type                       | Description                                                                                | Default      | Required |
| ------------------- | -------------------------- | ------------------------------------------------------------------------------------------ | ------------ | -------- |
| `block_on_overflow` | `boolean`                  | The behavior when the component's `TotalSize` limit is reached.                            | `false`      | no       |
| `blocking`          | `boolean`                  | (Deprecated) If `true`, blocks until the queue has room for a new request.                 | `false`      | no       |
| `enabled`           | `boolean`                  | Enables a buffer before sending data to the client.                                        | `true`       | no       |
| `num_consumers`     | `number`                   | Number of readers to send batches written to the queue in parallel.                        | `10`         | no       |
| `queue_size`        | `number`                   | Maximum number of unwritten batches allowed in the queue at the same time.                 | `1000`       | no       |
| `sizer`             | `string`                   | How the queue and batching is measured.                                                    | `"requests"` | no       |
| `storage`           | `capsule(otelcol.Handler)` | Handler from an `otelcol.storage` component to use to enable a persistent queue mechanism. |              | no       |
```

docs/sources/shared/reference/components/otelcol-queue-block.md
```diff
@@ -6,13 +6,19 @@ headless: true
...
+
+The `blocking` argument is deprecated in favour of the `block_on_overflow` argument.
```
clayton-cornell, line 19:
```suggestion
The `blocking` argument is deprecated in favor of the `block_on_overflow` argument.
```

docs/sources/shared/reference/components/otelcol-queue-block.md
```diff
@@ -21,6 +27,11 @@ Batches sent to the component's `input` exported field are added to the buffer a
 Assuming 100 requests/second, the default queue size `1000` provides about 10 seconds of outage tolerance.
 To calculate the correct value for `queue_size`, multiply the average number of outgoing requests per second by the time in seconds that outages are tolerated. A very high value can cause Out Of Memory (OOM) kills.
 
+The `sizer` argument could be set to:
+  - `requests`: number of incoming batches of metrics, logs, traces (the most performant option).
+  - `items`: number of the smallest parts of each signal (spans, metric data points, log records).
+  - `bytes`: the size of serialized data in bytes (the least performant option).
```
clayton-cornell, line 33:
```suggestion
The `sizer` argument could be set to:

* `requests`: number of incoming batches of metrics, logs, traces (the most performant option).
* `items`: number of the smallest parts of each signal (spans, metric data points, log records).
* `bytes`: the size of serialized data in bytes (the least performant option).
```

clayton-cornell (commented):
A few suggestions and comments - most are formatting related to align with the formatting changes in https://github.com/grafana/alloy/pull/3583

internal/component/otelcol/exporter/kafka/kafka.go
ptodev: Ah I see! Those are not the deprecated field defaults. Those are the defaults to use if neither the deprecated nor the "new" field has a non-empty value. I added comments to make this more clear. The confusion comes from the name of the `deprecatedArg` struct... but I don't know what else to call it?

docs/sources/reference/components/otelcol/otelcol.processor.k8sattributes.md
```diff
@@ -150,6 +151,10 @@ By default, if `metadata` is not specified, the following fields are extracted a
 * `container.image.name` (requires one of the following additional attributes to be set: `container.id` or `k8s.container.name`)
 * `container.image.tag` (requires one of the following additional attributes to be set: `container.id` or `k8s.container.name`)
 
+When `otel_annotations` is set to `true`, annotations such as `resource.opentelemetry.io/foo` will be translated to the `foo` resource attribute, etc.
```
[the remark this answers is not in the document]
ptodev: Good point - I replaced `foo` with `exampleResource`.

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -316,6 +316,30 @@ For more information about this topic, refer to [Distributed traces with Beyla][
...
+
+You can disable the TCP/IP level encoding and TC programs by setting `context_propagation` to `"headers"`. 
+This context propagation support is fully compatible with any OpenTelemetry distributed tracing library.
+
+`context_propagation` can be set to either one of the following values:
+
+* `all`:	Enable both HTTP and IP options context propagation.
+* `headers`:	Enable context propagation via the HTTP headers only.
+* `ip`:	Enable context propagation via the IP options field only.
+* `disabled`:	Disable trace context propagation.
```
clayton-cornell, line 339:
```suggestion
* `all`: Enable both HTTP and IP options context propagation.
* `headers`: Enable context propagation via the HTTP headers only.
* `ip`: Enable context propagation via the IP options field only.
* `disabled`: Disable trace context propagation.
```

Removed tabs. We try not to use them in the markdown.

clayton-cornell (approved):
Doc look OK.

clayton-cornell (approved):
Doc look OK.  Approving for docs.

CHANGELOG.md
```diff
@@ -107,6 +110,32 @@ Main (unreleased)
...
```
dehaansa, line 129:
```suggestion
  - [`otelcol.connector.spanmetrics`] Moving the start timestamp (and last seen timestamp) from the resourceMetrics level to the individual metrics level. 
    This will ensure that each metric has its own accurate start and last seen timestamps, regardless of its relationship to other spans.
  - [`otelcol.processor.k8sattributes`] Add option to configure automatic resource attributes - with annotation prefix.
    Implements [Specify resource attributes using Kubernetes annotations](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#specify-resource-attributes-using-kubernetes-annotations).
```

CHANGELOG.md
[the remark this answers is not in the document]
dehaansa: I meant this: https://github.com/grafana/alloy/pull/3489/files#r2101054293

dehaansa (approved):
Approved with 1 attempted clarification on the changelog comment I made previously.

CHANGELOG.md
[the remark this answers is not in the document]
ptodev: Aha, makes sense, thanks for spotting it!

PR #3503, opened 2025-05-05
Update to go 1.24.2 and bump build container

kalleep:
#### PR Description
Update to go 1.24.2 and bump build container used.


jharvey10:
Should we also use this PR to update the one over in [syntax/go.mod](https://github.com/grafana/alloy/blob/main/syntax/go.mod)?

kalleep:
> Should we also use this PR to update the one over in [syntax/go.mod](https://github.com/grafana/alloy/blob/main/syntax/go.mod)?

Ah good catch

thampiotr (approved):

PR #3504, opened 2025-05-05
fix homebrew and helm release actions

kalleep:
#### PR Description
Both of these actions fails to generate a token.

Not sure what other permissions we need to make them work but pretty sure we need this one so we can create a token to begin with.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -70,6 +70,9 @@ jobs:
     needs: [setup]
     runs-on: ubuntu-latest
     if: needs.setup.outputs.changed == 'true'
+    permissions:
+      contents: write
+      id-token: write
     steps:
       - uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1.12.0
         id: app-token
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -12,7 +12,6 @@ env:
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/helm-release.yml
```diff
@@ -69,6 +69,9 @@ jobs:
     needs: [setup]
     runs-on: ubuntu-latest
     if: needs.setup.outputs.changed == 'true'
+    permissions:
+      contents: write
+      id-token: write
```
jharvey10, line 74: Looking at the GH docs, create-github-app-token doesn't seem to need `id-token: write`, I believe because they're using private key to do it instead. My read on this is that the actions token will govern permissions for all of the steps in which it's used here, so we might not need to specify _any_ additional permissions in the workflow itself.

As a side note, the gh app token can have it's permissions limited too when it's created; by default it will inherit all of the permissions that the app installation has against the repo(s).

```suggestion
```

kalleep: Maybe but currently this fails on this step https://github.com/grafana/alloy/actions/runs/14836822313/job/41649919010

I was looking at what other things we have that uses this action and found https://github.com/grafana/writers-toolkit/blob/main/publish-technical-documentation/action.yml#L3-L11 so I assumed we needed it

kalleep: Reading a bit closer I can see that it is for another action

PR #3513, opened 2025-05-06
fix(helm): check autoscaling values

### LINKED ISSUE #3474: Helm chart: Deployment/StatefulSet replica count not ignored on autoscaling
### What's wrong?

The helm chart values currently state that the key `controller.autoscaling` should not be used, but instead `controller.autoscaling.horizontal` should be used. See the comment here: https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/values.yaml#L252

I just tried that and failed because of this condition here: https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/controllers/deployment.yaml#L14

This only checks for the deprecated way of enabling autoscaling.

I'll probably create a MR next week to solve this.

### Steps to reproduce

Use `controller.autoscaling.horizontal` instead of `controller.autoscaling` and if `replicas` is set to 1 for example and `controller.autoscaling.horizontal.minReplicas` is set to 2, one pod comes up and gets terminated right away.

### System information

_No response_

### Software version

_No response_

### Configuration

```text

```

### Logs

```text

```

tobimichael96:
This adds `controller.autoscaling.horizontal.enabled` to `deployment.yaml` and `statefulset.yaml` check in order to ignore `replicas` if set to true

#### PR Checklist

- [ ] CHANGELOG.md updated
- [X] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/templates/controllers/deployment.yaml
+++ b/operations/helm/charts/alloy/templates/controllers/deployment.yaml
@@ -11,7 +11,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if not .Values.controller.autoscaling.enabled }}
+  {{- if (and (not .Values.controller.autoscaling.enabled) (not .Values.controller.autoscaling.horizontal.enabled)) }}
   replicas: {{ .Values.controller.replicas }}
   {{- end }}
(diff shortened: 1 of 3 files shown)


tobimichael96:
Closed in favor of https://github.com/grafana/alloy/pull/3514

PR #3523, opened 2025-05-07
Add Stringer implementation for OptionalSecret type

### LINKED ISSUE #1470: Formatting of local.file in format function does not return the string contents
### What's wrong?

When you try to use the format function to crate a new string and one of the arguments is the contents of the local.file you get something like this back **{%!s(bool=false) contents_of_the_file}:15692** when you use **format("%s:%s", local.file.file.content, "15692")**. This does not happen when you use for example join to combine both strings.

### Steps to reproduce

Crate a alloy config which locks something like this as added below.

### System information

_No response_

### Software version

Grafana Alloy 1.3.0

### Configuration

```text
local.file "instance_id" {
  filename  = "/etc/alloy/instance_id.txt"
}

prometheus.scrape "default" {
  targets    = [
    {"__address__" = "127.0.0.1:15692", "instance" = format("%s:%s", local.file.instance_id.content, "15692")},
  ]
  forward_to = [prometheus.remote_write.default.receiver]
  scrape_interval = "15s"
}
```

### Logs

_No response_

stuart-warren:
#### PR Description

Add Stringer implementation for OptionalSecret type

Trying to format an OptionalSecret type to a string currently returns `{%!s(bool=false) foobar}` rather than a string

Example config to show issue:
```
local.file "iloip" {
  filename = file.path_join(sys.env("SNAP_DATA"), "config", "iloip.txt")
}

// https://github.com/FlxPeters/redfish_exporter
prometheus.scrape "redfish_exporter" {
  targets = [
    {"__address__" = "localhost:9610",
     "__metrics_path__" = "/redfish",
     "__param_target" = coalesce(local.file.iloip.content,"notfound")},
  ]

}
```

Fixes #1470

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Main (unreleased)
 
 - Fix [#3386](https://github.com/grafana/alloy/issues/3386) lower casing scheme in `prometheus.operator.scrapeconfigs`. (@alex-berger)
 
+- Fix [#1470](https://github.com/grafana/alloy/issues/1470) allow formatting of an OptionalSecret (from `local.file`) as a string. (@stuart-warren)
+
 ### Other changes
 
 - Update the zap logging adapter used by `otelcol` components to log arrays and objects. (@dehaansa)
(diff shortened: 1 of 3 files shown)


wildum:
hi, thanks for contributing :) I think that the problem can already be solved by using the function [convert.non_sensitive](https://grafana.com/docs/alloy/latest/reference/stdlib/convert/#nonsensitive) or do you see a problem with this approach?

stuart-warren:
@wildum thanks, that does appear to work, was just a bit non-obvious :)

PR #3525, opened 2025-05-07
ci: update backport workflow permissions

jharvey10:
synchronizing permissions with ones listed in the upstream workflow we're calling.

NOTE: This "looks" right, but remains untested until it's integrated into main and used.


dehaansa (approved):
LGTM, thanks!

PR #3530, opened 2025-05-07
fix(ui): make component graph links follow `--server.http.ui-path-prefix`

### LINKED ISSUE #3437: Alloy UI - Graph React flow UI does not follow ui-path-prefix
### What's wrong?

When setting --server.http.ui-path-prefix and clicking on one element below (see capture), the path is not added to the URL, it is only in the "Graph" React flow. 

Elsewhere it seems to be working.

Image

### Steps to reproduce

Setup allow with --server.http.ui-path-prefix=alloy and click on one element of the graph

### System information

Linux

### Software version

Grafana Alloy v1.8.1

### Configuration

```text

```

### Logs

```text

```

solidcellaMoon:
#### PR Description

This updates `ComponentGraph.tsx` to insert the correct path prefix in `onNodeClick`.

When Alloy runs with `--server.http.ui-path-prefix` (e.g. `/alloy`), clicking a node in the graph incorrectly opens new-tab to `/component/<id>` instead of `/alloy/component/<id>`, resulting in a 404 error (#3437). Users have to manually add the prefix in the address bar to reach the page—an unnecessary extra step.

With this change, clicking a node in the graph opens new-tab to `/<prefix>/component/<id>` when a prefix is set, or `/component/<id>` when no prefix is configured.

Fixes #3437 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Main (unreleased)
 
 - Fix [#3386](https://github.com/grafana/alloy/issues/3386) lower casing scheme in `prometheus.operator.scrapeconfigs`. (@alex-berger)
 
+- Fix [#3437](https://github.com/grafana/alloy/issues/3437) Component Graph links now follow `--server.http.ui-path-prefix`. (@solidcellaMoon)
+
 ### Other changes
 
 - Update the zap logging adapter used by `otelcol` components to log arrays and objects. (@dehaansa)
--- a/internal/web/ui/src/features/graph/ComponentGraph.tsx
+++ b/internal/web/ui/src/features/graph/ComponentGraph.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useMemo, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
 import { Edge, Node, ReactFlow, useEdgesState, useNodesState } from '@xyflow/react';
 
 import { useGraph } from '../../hooks/graph';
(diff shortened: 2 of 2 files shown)


jharvey10:
Thanks for opening this!

I do see this changes the behavior of the UI slightly: new window/tab versus same window.

Accessibility-wise, same window is usually preferable (as your implementation does), but the problem with our existing code is that it's a react-flow node (i.e. role="button") and not an anchor tag, so users don't have the option to do something like cmd+click or middle click to open a new tab.

@wildum Perhaps we accept this PR as-is and make a follow on issue to better accommodate accessibility of the node graph?

Food for thought:
image
where details is a link to the component page.

wildum:
hi, thanks for catching and fixing this bug :) It was actually changed upon request to open the component page in a different tab instead of using the same page: https://github.com/grafana/alloy/pull/3123/files. This is also my preferred way when I debug an issue. I often move the nodes around to organize the graph better, and if I exit the page, I would need to reorganize them again.

Is there a way to keep the existing behavior of opening the component page in a new page, but with the correct prefix?

@jharvey10 the problem with the details is that we don't have a lot of room in the node. Keeping the nodes small makes it easier to have an overview of the whole graph. Maybe a small icon would be ok

solidcellaMoon:
I updated the PR description to note that the fix inserts the correct path prefix while still opening the component page in a new tab.

solidcellaMoon:
I have a quick question about CI.

The only failing check is `zizmor GitHub Actions static analysis`, and the log shows
```
fatal: no audit was performed
error: failed to load config: No such file or directory (os error 2)
 = note: check your configuration file for errors
```

I can’t find any zizmor setup in this repo, and none of the recently merged PRs seem to modify its settings either. Is there something I should commit for this job?

kalleep:
@solidcellaMoon it does not work from forks, but its not required anyway

PR #3534, opened 2025-05-08
Improve string format doc

wildum:
Some users encountered issues when trying to format secrets. This doc update explains that the secrets can either be converted to string via the convert.nonsensitive function or users can use string concatenation

Related to https://github.com/grafana/alloy/pull/3523


clayton-cornell (approved):
LGTM

PR #3542, opened 2025-05-09
fix(pyroscope.scrape): godeltaprof scraping

korniltsev:
#### PR Description
Pyroscope targets hide scraping godeltaprof_ profiles, by replacing the name from `godeltaprof_memory` to `memory`

The problem is that the scrape pool decides to use `deltaAppender` based on the name and now it does not know it's godeltaprof profile and assumes regular memory profile and tries to compute delta second time on the godeltaprof profiles.

In this PR I pass  a flag to never use `deltaAppender` for godeltaprof profiles 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Main (unreleased)
 
 - Fix `otelcol.receiver.filelog` documentation's default value for `start_at`. (@petewall)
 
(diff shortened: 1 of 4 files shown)


marcsanmi (approved):
LGTM

PR #3564, opened 2025-05-12
Stop signing RPM artifacts

ptodev:
This rollbacks [a change](https://github.com/grafana/agent/pull/1879) made three years ago. In that change, it seems that the Grafana Platform squad introduced RPM signing so that the central repository used by yum can also have signed binaries. @julienduchesne, I suppose this is no longer needed?

The reason I'm opening this PR is that the current RPM signing process is blocking the Alloy migration from Drone to GitHub Actions. The GPG keys are stored in a way which can't be retrieved by the [get-vault-secrets](https://github.com/grafana/shared-workflows/tree/main/actions/get-vault-secrets) action. The platform team is reluctant to move or copy them, so we would like to see if we can just stop signing the binaries altogether.

The platform team has a separate process for signing RPMs destined for `https://rpm.grafana.com`, the repo used by `yum`. So `yum` will continue to have signed binaries.


.drone/drone.yml
```diff
@@ -250,6 +250,10 @@ platform:
   arch: amd64
   os: linux
 steps:
+- commands:
+  - git fetch --tags
+  image: grafana/alloy-build-image:v0.1.18
+  name: Fetch tags
```
ptodev, line 256: @kalleep This change came up after I ran `make drone`. It's because `make drone` wasn't ran in an [earlier PR](https://github.com/grafana/alloy/pull/3543). There's a [doc](https://github.com/grafana/alloy/blob/main/.drone/README.md) on how to do this.

I'd also fogotten to run `make drone` on a change I did a few months ago. It's no big deal, but the drone builds might refuse to run automatically on PRs. They might need to be triggered explicitly until `make drone` is ran on the `main` branch.

Issue #3315, opened 2025-04-14
Add validate command for alloy config


kalleep:
### Background

Currently there is no way to perform any sort of validation of alloy config files like [agent have](https://github.com/grafana/agent/blob/main/cmd/grafana-agentctl/main.go#L116). 

They only way to perform any sort of validation is to either deploy config and check logs or try to run alloy locally with the config. 

### Proposal

Add a validate command to alloy. This command should:

1. Parse config and report any syntax errors.
2. Check if components exists and can be used with proper flags (stability and community components enabled).
3. Perform some sort of type checking of arguments.

The first two are pretty simple and already implemented in https://github.com/grafana/alloy/pull/3220.
The third point is most likely going to be pretty tricky. We cannot really reuse the runtime flow because this will have side effect, like expanding environment variables. Instead I think we want to perform type checking and some type inference e.g. we infer that `sys.env("VARIABLE")` returns a string. This will have some limitations and cannot catch all possible errors but we can work towards making this as robust a possible.

I prefer to split this work over several pr:s because I will need to experiment a lot on point 3 and we can get some value of the easy parts faster.

Other things to explore is how we can expose this for external use like fleet management

xh3b4sd:
I could have really used this today.

kalleep:
Currently validation can validate components, services, configs, declares, foreach and custom components from declare blocks.

TODO:
- [ ] Complete the graph with edges so we can validate circular references and that component we reference exists
- [ ] Flag for the command that would allow us to "resolve" imports
- [ ] Start performing type check of attributes

kalleep:
Closing this issue because validation command has already been implemented and will create sub-issues for improvements we can do.

PR #3565, opened 2025-05-12
feat: validate foreach blocks

kalleep:
#### PR Description
Continue to expand capabilities of validation by adding support for `foreach` blocks.

Part of: #3315

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell (commented):
Doc change is small . Good as-is

internal/dynamic/foreach/foreach.go
```diff
@@ -0,0 +1,23 @@
+package foreach
```
wildum, line 1: is the purpose of the dynamic folder to expose configs of internal nodes? why is it called dynamic?

internal/validator/testdata/default/valid.txtar
```diff
@@ -1,74 +1,153 @@
...
+			}
+
+			rule {
+				replacement  = each["__meta_kubernetes_pod_container_name"]
+				target_label = "k8s_pod_container_name"
+				action       = "replace"
+			}
+
+			forward_to = [prometheus.remote_write.mimir.receiver]
+		}
```
wildum, line 122: you could put the math.default in the template too to make sure that instantiating a custom component in a foreach is valid

internal/validator/validate.go
```diff
@@ -138,12 +145,100 @@ func (v *validator) validateConfigs(configs []*ast.BlockStmt) diag.Diagnostics {
...
+		diags.Add(diag.Diagnostic{
+			Severity: diag.SeverityLevelError,
+			StartPos: ast.StartPos(block).Position(),
+			EndPos:   ast.EndPos(block).Position(),
+			Message:  fmt.Sprintf("missing required block %q", foreach.TypeTemplate),
+		})
+		return diags
+	}
+
+	// We extarct all blocks from template body and evaludate them as as components.
```
wildum, line 207:
```suggestion
	// We extract all blocks from template body and evaluate them as components.
```

internal/validator/validate.go
```diff
@@ -138,12 +145,100 @@ func (v *validator) validateConfigs(configs []*ast.BlockStmt) diag.Diagnostics {
...
 
 	return diags
 }
 
+func (v *validator) validateForeach(block *ast.BlockStmt) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	name := block.GetBlockName()
+	// Check required stability level.
+	if err := featuregate.CheckAllowed(foreach.StabilityLevel, v.minStability, fmt.Sprintf("foreach block %q", name)); err != nil {
```
wildum, line 160: would be cool to have a test for the stability level. It's probably the most common error that the user will run to

internal/dynamic/foreach/foreach.go
wildum: I think that this scenario will happen for other blocks (like the argument block in declares if you add validation support) so I would scope it in a pkg. Maybe something like internal/nodeconf/foreach?

kalleep: Yeah I think so too. The same will happen for declares and imports. Sure nodeconf sounds good, we can always change that if we find a more fitting name

kalleep: Done in [2729e32](https://github.com/grafana/alloy/pull/3565/commits/2729e3274a25e5d1157f1c69be2f9166eb2f9d5c)

wildum (approved):

PR #3579, opened 2025-05-13
Upgrade prometheus-operator from v0.69.1 to v0.82.2

ptodev:
This is a requirement for the [upcoming mimir.alerts.kubernetes component](https://github.com/grafana/alloy/pull/3448). An upcoming version of Prometheus Operator will contain [a change](https://github.com/prometheus-operator/prometheus-operator/pull/7486) which will need to be pulled in. Since we are running on a much older version of the operator, I'm updating the dependencies now to reduce the blast radius over time.

I idd not add a changelog entry because there shouldn't be any user-facing changes.


kalleep (approved):
LGTM

PR #3585, opened 2025-05-14
ci: update renovate.json5

jharvey10:
Several Go dependencies were getting ignored because they did not fall under the update categories of "minor" and "patch".

I reworked the config to not rely on those as matching criteria while still separating out major version updates.

I noticed this was an issue because one CVE resolution was not being opened as a PR despite having a fix available.


wildum (approved):

jharvey10:
Fixes go dependency ignoring for certain dep types

PR #3586, opened 2025-05-14
docs: fix typo in stage.match example and description

tobiabocchi:
#### PR Description

I believe there is a subtle error in the docs regarding the stage.match example. I think I fixed it using the correct values for the selector and related documentation.

I could not find an issue related to this.

#### PR Checklist

- [x] Documentation added


clayton-cornell (approved):
Thanks for catching that one :-)

PR #3604, opened 2025-05-14
feat: listen http server on unix socket

### LINKED ISSUE #1509: Is there any way to disable the Alloy HTTP server
When alloy runs, it launches a local HTTP server for debugging, defaulting to binding to `127.0.0.1:1234` (see https://grafana.com/docs/alloy/latest/troubleshoot/debug and https://grafana.com/docs/alloy/latest/reference/cli/run/ ).

Going through the documentation (config file docs and command line interface docs), there appears to be no way to actually disable this local HTTP server. Is there any way to disable it (clearly not documented) short of patching Alloy's source code and building it from source? This would be much much much better (and more reliable) than having to use a firewall rule to block the port on localhost.

For reference, this is for running Alloy on multi-user nodes where exposing internal interfaces to local users is something to be avoided.

ebuildy:
#### PR Description

Make possible alloy http server to bind unix socket.

Fixes #1509

#### Notes to the Reviewer

Didnt test clustering, I guess this cannot work, I could check "if unix AND cluster -> error"?

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -178,7 +178,17 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		}
 	}()
 
-	netLis, err := net.Listen("tcp", s.opts.HTTPListenAddr)
+	var netLis net.Listener
+	var err error
+
+	if strings.HasPrefix(s.opts.HTTPListenAddr, "unix:") {
+		listenAddr := strings.TrimPrefix(strings.TrimLeft(s.opts.HTTPListenAddr, "unix:"), "//")
+		level.Error(s.log).Log("msg", listenAddr)
+		netLis, err = net.Listen("unix", listenAddr)
+	} else {
+		netLis, err = net.Listen("tcp", s.opts.HTTPListenAddr)
+	}
+
 	if err != nil {
 		// There is no recovering from failing to listen on the port.
 		level.Error(s.log).Log("msg", fmt.Sprintf("failed to listen on %s", s.opts.HTTPListenAddr), "err", err)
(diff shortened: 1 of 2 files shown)


internal/service/http/http.go
```diff
@@ -178,7 +178,17 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 		}
 	}()
 
-	netLis, err := net.Listen("tcp", s.opts.HTTPListenAddr)
+	var netLis net.Listener
+	var err error
+
+	if strings.HasPrefix(s.opts.HTTPListenAddr, "unix:") {
```
ptodev, line 184: Wouldn't this theoretically be a breaking change if you have a domain called `unix`?

ebuildy: well, ":" is only used to split protocol right? (I am not sure)

ebuildy:
> Hi, thank you for the PR! Do you mind sharing more information about the use case please? I'm not very familiar with Unix sockets.
> 
> > you run Alloy inside a container, inside a pod with host network
> 
> Is this so that Alloy's HTTP server is only visible within the node? Are Unix sockets a standard practice to avoid exposing something outside of such a node? This feels like a workaround rather than a general solution to the problem.
> 
> > you run Alloy inside a PC and you dont want the user see Alloy web UI
> 
> If both the user and Alloy are on the same PC, wouldn't the user be able to still access the Alloy UI?

Yes unix socket is a common way to control service exposition, Docker is the best example, but I agress this is more common between CLI and API (such Docker).

PR #3606, opened 2025-05-15
Migrate the Drone release steps to GitHub Actions

ptodev:
I tested this on my fork as much as I could. For example, the "publish github release" step [worked](https://github.com/ptodev/alloy/actions/runs/15065212788/job/42348401314) and generated a [draft release](https://github.com/ptodev/alloy/releases). I couldn't test other steps for generating containers all the way because the free Actions runners I use on my fork run out of disk space... but I have high confidence in them since they went far enough in the workflow.


kalleep (approved):
LGTM, nice work  :+1:

PR #3607, opened 2025-05-15
feat: validate declare blocks

kalleep:
#### PR Description
This pr add support to validate blocks inside declare blocks, this includes arguments, exports and component blocks.

We don't validate the usage of declares yet, for that we need to store information about arguments and I plan to expand on that in future pr.

I also changed to store all nodes in `dag.Graph` and change how errors are recorded. This is something else I also want to expand on by creating the edges between nodes and perform similar validation we do for runtime.

Part of: https://github.com/grafana/alloy/issues/3315

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/validator/testdata/default/invalid_declare.diags
```diff
@@ -0,0 +1,313 @@
...
+43 |     logging { }
+   |     ^^^^^^^
+44 | 
+
+Error: main.alloy:43:2: cannot find the definition of component name "logging"
+
+42 |     // Service configuration.
+43 |     logging { }
+   |     ^^^^^^^
+44 |
```
wildum, line 251: why are the errors duplicated? the errors about the logging block are misleading, it should be "logging block not allowed inside a module"

internal/validator/validate.go
```diff
@@ -85,100 +119,132 @@ func (v *validator) run(cr *componentRegistry) error {
...
-	return diags
+		configs, declares, components := extractBlocks(node, node.block.Body)
+		node.diags.Merge(v.validate(&state{
+			root:       false,
+			graph:      newGraph(),
+			declares:   declares,
+			configs:    configs,
+			components: components,
+			cr:         newComponentRegistry(s.cr),
+		}))
```
wildum, line 157: It's possible to use a custom component within a declare:
```
declare "a" { // content }
declare "b" { a "test" {} // content}
```
I think that the validation will fail depending on the order of the declare, no?

kalleep: Yes you are correct, if you swap those it will fail

wildum (approved):

internal/validator/validate.go
```diff
@@ -61,125 +63,196 @@ func (v *validator) run(cr *componentRegistry) error {
...
+			EndPos:   node.block.NamePos.Add(len(name) - 1).Position(),
 			Message:  err.Error(),
 		})
 	}
 
-	// Require label for all foreach blocks.
-	if block.Label == "" {
-		diags.Add(diag.Diagnostic{
+	// Require label for foreach block.
+	if node.block.Label == "" {
```
wildum, line 250: foreach dont actually need label right now, I just discovered it recently but I think that they should. So we can keep this as it is if you want

kalleep: Yeah so I am pretty sure we have the same problem for argument and export too.

PR #3618, opened 2025-05-16
update oracle db exporter

### LINKED ISSUE #2740: Please update prometheus.exporter.oracledb to use exporter fork supported by Oracle
### Request

The `prometheus.exporter.oracledb` component in Alloy imports an [OracleDB exporter](https://github.com/iamseth/oracledb_exporter) which was archived and is in read-only mode since Feb 13th.

It would be great if Alloy starts using the [OracleDB exporter fork from Oracle](https://github.com/oracle/oracle-db-appdev-monitoring).

### Use case

The fork from Oracle implements bug fixes and is actively supported. That would avoid the need of running and maintaining a second container for using the new OracleDB exporter fork.

wildum:
#### PR Description

With this update, we are changing the underlying exporter from the archived exporter https://github.com/iamseth/oracledb_exporter to maintained exporter https://github.com/oracle/oracle-db-appdev-monitoring.

One important change is that the credentials are not part of the connection string anymore. To avoid breaking existing users, the component now checks and extract the crendentials from the URL if needed.

Unavoidable breaking changes:
- two metrics are gone - the rest have the same names according to both documentations
- the undocumented custom_metrics argument takes file paths instead of raw custom metrics

Fixes #2740

#### Notes to the Reviewer

I tested it locally with the oracle docker image: container-registry.oracle.com/database/express:21.3.0-xe

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/prometheus/prometheus.exporter.oracledb.md
```diff
@@ -27,17 +27,28 @@ prometheus.exporter.oracledb "<LABEL>" {
...
-The [`oracledb_exporter` running](https://github.com/iamseth/oracledb_exporter/tree/master#running) documentation shows the format and provides examples of the `connection_string` argument:
+For backward compatibility, the `username` and `password` arguments can still be provided in the `connection_string` argument:
 
 ```conn
-oracle://user:pass@server/service_name[?OPTION1=VALUE1[&OPTIONn=VALUEn]...]
+oracle://user:pass@host:port/service_name[?OPTION1=VALUE1[&OPTIONn=VALUEn]...]
+```
+
+If the `connection_string` argument doesn't contain the `username` and `password`, the `username` and `password` arguments must be provided.
+In this case, the url must have the format:
```
clayton-cornell, line 48:
```suggestion
In this case, the URL must have the format:
```

docs/sources/reference/components/prometheus/prometheus.exporter.oracledb.md
```diff
@@ -27,17 +27,28 @@ prometheus.exporter.oracledb "<LABEL>" {
...
+| ------------------- | ---------- | -------------------------------------------------------------- | ------- | -------- |
+| `connection_string` | `secret`   | The connection string used to connect to an Oracle Database.   |         | yes      |
+| `username`          | `string`   | The username to use for authentication to the Oracle Database. |         | no       |
+| `password`          | `secret`   | The password to use for authentication to the Oracle Database. |         | no       |
+| `max_idle_conns`    | `int`      | Number of maximum idle connections in the connection pool.     | `0`     | no       |
+| `max_open_conns`    | `int`      | Number of maximum open connections in the connection pool.     | `10`    | no       |
+| `query_timeout`     | `int`      | The query timeout in seconds.                                  | `5`     | no       |
+| `default_metrics`   | `string`   | The path to the default metrics file.                          |         | no       |
+| `custom_metrics`    | `[]string` | The paths to the custom metrics files.                         |         | no       |
```
kalleep, line 39: Should we document that these files need to be toml format?

internal/component/prometheus/exporter/oracledb/oracledb.go
```diff
@@ -60,19 +64,22 @@ func (a *Arguments) Validate() error {
...
-		return errNoHostname
+	if strings.HasPrefix(string(a.ConnectionString), "oracle://") {
+		u, err := url.Parse(string(a.ConnectionString))
+		if err != nil {
+			return err
+		}
+		_, set := u.User.Password()
+		if !set {
+			return errNoPassword
+		}
```
kalleep, line 80: Should it not be possible to connect without any username password? I imagine that would be rare but I am pretty sure it was possible before.

kalleep (commented):
Nice work.

I wonder if we can make the connection string, username and password validation / parsing a bit simpler because now we kinda need to validate in both places.


What if we did all of the work in the component and just make sure we pass whatever is expected when we convert?

1. Always parse the url
2. Validate the username and password is not set in both url and attributes but allow both to be empty.
3. Create the connection string in the expected format in the convert function (without scheme, password and username)

docs/sources/reference/components/prometheus/prometheus.exporter.oracledb.md
wildum: done + added a line to point out to examples in the upstream directory

internal/component/prometheus/exporter/oracledb/oracledb.go
wildum: I made the changes that you suggested, the logic is now much simpler and only in the component. I allowed connecting without credentials in the backward compatible way, assuming that the format would then be `oracle://host:port/service_name[?OPTION1=VALUE1[&OPTIONn=VALUEn]...]`. I'm not sure that's ever a valid scenario but better be flexible and let the exporter complain rather than potentially blocking a valid scenario.
I tested the different scenarios again using the oracledb docker image. lmk what you think about the changes

kalleep (approved):
LGTM

PR #3630, opened 2025-05-19
fix(pyroscope.java): Ensure the jfr is deleted on shutdown

simonswine:


alsoba13 (approved):

internal/component/pyroscope/java/loop.go
```diff
@@ -114,15 +121,22 @@ func (p *profilingLoop) loop(ctx context.Context) {
 	}
 }
 
+func (p *profilingLoop) cleanup() {
+	jfrFile := asprof.ProcessPath(p.jfrFile, p.pid)
+
+	err := os.Remove(jfrFile)
+	if err != nil {
+		_ = level.Warn(p.logger).Log("msg", "failed to delete jfr file", "err", err)
+	}
```
aleks-p, line 130: I have found that this is not always enough. When running on Linux and you stop the profiled java process during profiling, the jfr file is no longer in `/proc/<pid>` (this gets cleaned up by the OS), but the jfr file is still in `/tmp` on the host. This is what is reported in https://grafana.slack.com/archives/C049PLMV8TB/p1747153446309779 as well.

Something like this should do it:

```java
        jfrFile := asprof.ProcessPath(p.jfrFile, p.pid)
	_ = level.Debug(p.logger).Log("msg", "deleting jfr file from target", "jfr_file", jfrFile)
	if err := os.Remove(jfrFile); err != nil {
		if errors.Is(err, os.ErrNotExist) {
			_ = level.Debug(p.logger).Log("msg", "jfr file already deleted on target", "jfr_file", jfrFile)
			// if the asprof process stopped, the file could still be on the host
			_ = level.Debug(p.logger).Log("msg", "deleting jfr file from host", "jfr_file", p.jfrFile)
			if err = os.Remove(p.jfrFile); err != nil {
				if errors.Is(err, os.ErrNotExist) {
					_ = level.Debug(p.logger).Log("msg", "jfr file already deleted on host", "jfr_file", p.jfrFile)
					return
				}
				_ = level.Warn(p.logger).Log("msg", "failed to delete jfr file", "jfr_file", p.jfrFile, "err", err)
				return
			}
		}
		_ = level.Warn(p.logger).Log("msg", "failed to delete jfr file", "jfr_file", p.jfrFile, "err", err)
	}
```

internal/component/pyroscope/java/loop.go
```diff
@@ -102,6 +108,7 @@ func (p *profilingLoop) loop(ctx context.Context) {
 		}
 		sleep()
 		if ctx.Err() != nil {
+			p.cleanup()
```
aleks-p, line 111: this could also go in `Close`, not sure if it makes a big difference or not

simonswine: Aggreed, not came across `Close` before I wrote the test.

This should cover more cases

internal/component/pyroscope/java/loop.go
simonswine: We can't fully fix it, because it a process in another mount namespace goes away, we still leak the jfr files.

I think with taking care of the process and the host at least, we catch two very common cases.

Your logging was a bit to verbose, have a look what you think about my change...

aleks-p: Looks good to me and works the same in my tests.

> Your logging was a bit to verbose, have a look what you think about my change...

It was mostly debug logging. I think we want to at least say that we are deleting a file (or have deleted a file) so that next time an issue happens we can ask the user to enable debug logging and make sure we hit this area of the code.

Issue #3629, opened 2025-05-19
/-/ready and /-/healthy http endpoints are spamming logs


niklass08:
### What's wrong?

When requesting the /-/ready and /-/healthy http endpoints they are producing the following log line
`2025/05/19 17:08:54 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux.getRRW.func2.1 (mux.go:114)`

The log line does not seem to have a level attached to it, and is produced each time one of these endpoints is called. It's an issue especially whe using these endpoints for automated probing.

The bug is reproduced on at least alloy 1.8.3 and 1.7.5

### Steps to reproduce

- Retrieve the alloy package from the github release page and install it.
- run `alloy run .`
- call `localhost:12345/-/healthy` or `localhost:12345/-/ready`
- observe the spam log line in stdout

### System information

ubuntu-22.04 x86_64

### Software version

alloy 1.8.3 and 1.7.5

### Configuration

```text
no configuration file needed
```

### Logs

```text
2025/05/19 17:08:54 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux.getRRW.func2.1 (mux.go:114)
```

kalleep:
Thanks for reporting, I can only reproduce this for `localhost:12345/-/healthy` and its dues to us writing response body before writing the header here https://github.com/grafana/alloy/blob/c79a8ce3e0f56ce570fb4509897f678414fea5f6/internal/service/http/http.go#L236-L237.

Looking at the ready endpoint everything seems correct and I don't get the error. I will make a pr with the fix

niklass08:
Thanks for your quick answer and fix ! Sorry for reporting this on the ready endpoint as well, little oversight on my end

PR #3639, opened 2025-05-21
fix: write header before writing the response

kalleep:
#### PR Description
We should write header before writing response body.

Fixes: https://github.com/grafana/alloy/issues/3629

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Main (unreleased)
 
 - Fix a bug in the `foreach` preventing the UI from showing the components in the template when the block was re-evaluated. (@wildum)
 
+- Fix alloy health handler so header is written before response body. (@kalleep)
+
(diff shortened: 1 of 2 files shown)


wildum (approved):

tcurdt:
Thanks for the patch. So this will be available with an upcoming 1.8.4 release?

wildum:
hi @tcurdt we just released v1.9.0 but this fix did not make the cut. This means that it will be either be available in v1.10 which is planned in 6 weeks from now, or in a v1.9.x patch if we release one in between

wildum:
ah nevermind I think that the fix did make the cut @tcurdt

tcurdt:
Even better :)

PR #3642, opened 2025-05-21
database_observability: collect thread_id, end_event_id to query sample collector

fridgepoet:
#### PR Description

This PR adds the selection of THREAD_ID and END_EVENT_ID and logging of these to the query sample collector. This will make it clearer that rows with the same EVENT_ID are distinct due to different THREAD_IDs. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/query_sample.go
+++ b/internal/component/database_observability/mysql/collector/query_sample.go
@@ -54,6 +54,7 @@ SELECT
 	statements.MAX_CONTROLLED_MEMORY,
 	statements.MAX_TOTAL_MEMORY,
 	waits.event_id as WAIT_EVENT_ID,
+	waits.end_event_id as WAIT_END_EVENT_ID,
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -39,7 +39,9 @@ WHERE variable_name = 'UPTIME'`
 const selectQuerySamples = `
 SELECT
 	statements.CURRENT_SCHEMA,
+	statements.THREAD_ID,
 	statements.EVENT_ID,
+	statements.END_EVENT_ID,
```
cristiangreco, line 44: The PR looks good! I'm wondering if we could add WAIT_END_EVENT_ID while at it?

cristiangreco (approved):

PR #3643, opened 2025-05-21
fix faro tests windows

wildum:


t00mas (approved):

internal/component/faro/receiver/sourcemaps.go
```diff
@@ -46,7 +46,7 @@ type (
 type osFileService struct{}
 
 func (fs osFileService) ValidateFilePath(name string) (string, error) {
```
wildum, line 48: @t00mas is this function useful? It's broken on Windows, and with this change the Test_sourceMapsStoreImpl_InvalidPathMetrics test is still broken

t00mas: Function itself doesn't matter that much, all this is about CodeQL's https://codeql.github.com/codeql-query-help/go/go-path-injection/ so it shouldn't contain path separators nor escape sequences.

t00mas: I don't see any issues with that functionality wise, hopefully CodeQL doesn't complain too much.

wildum: mmh I see, the problem is that this is blocking our release. To avoid any risks, I'm just gonna remove this part from the validation"strings.Contains(name, "\\")" so that it works on windows + fix the path on this other test. Can you address this in another PR when you have time?

PR #3645, opened 2025-05-21
Fixed minor typo

monsdar:
This is nothing more than a minor syntax fix. It is shown when calling the help of alloy CLI, so it's placing is quite prominent. Should've no impact on anything, but looks more professional if new users aren't greeted with a typo right away.


kalleep (approved):
LGTM, thanks!

PR #3657, opened 2025-05-22
Update component secret-filter gitleaks.toml

andrejshapal:
#### PR Description
Hi, 
The file was gitleaks.toml not updated for last 6 months and there was lot of fixes for different false positives and other improvements.
We faced falsepositive where ip address in traefik logs was masked by generic-api-key:
(?i)(?:key|api|token|secret|client|passwd|password|auth|access)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)(?:'|\"|\s|=|\x60){0,5}([0-9a-z\-_.=]{10,150})(?:['|\"|\n|\r|\s|\x60|;]|$)

I think it the file should be kept up to date and we should rely on gitleaks changes.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/secretfilter/gitleaks.toml
+++ b/internal/component/loki/secretfilter/gitleaks.toml
@@ -25,13 +25,16 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 #
-# Copied from https://github.com/gitleaks/gitleaks/blob/3be7faa1ffc80f1aa071192a09141d5e829e1af3/config/gitleaks.toml
+# Copied from https://github.com/gitleaks/gitleaks/blob/v8.26.0/config/gitleaks.toml
 # 
 # This file has been auto-generated. Do not edit manually.
 # If you would like to contribute new rules, please use
 # cmd/generate/config/main.go and follow the contributing guidelines
-# at https://github.com/zricethezav/gitleaks/blob/master/CONTRIBUTING.md
-
+# at https://github.com/gitleaks/gitleaks/blob/master/CONTRIBUTING.md
+#
+# How the hell does secret scanning work? Read this:
+#   https://lookingatcomputer.substack.com/p/regex-is-almost-all-you-need
+#
 # This is the default gitleaks configuration file.
 # Rules and allowlists are defined within this file.
 # Rules instruct gitleaks on what should be considered a secret.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -82,6 +82,9 @@ Main (unreleased)
   - [0.27.1] Support dogstatsd extended aggregation
   - [0.27.2] Fix panic on certain invalid lines
 
+- Update secret-filter gitleaks.toml from v8.19.0 to v8.26.0
+
+
```
kalleep, line 87:
```suggestion
```

kalleep (commented):
Thanks, just one minor nit but then we are good to go.

>I think it the file should be kept up to date and we should rely on gitleaks changes.

We could have a github action for this that run once a week. If there is a new release it could fetch the latest generated file and make a pr

kalleep (approved):

kalleep (commented):
Hey there are a test failing due to changes in the global allow list, specifically this regex
`'''^(?i:a+|b+|c+|d+|e+|f+|g+|h+|i+|j+|k+|l+|m+|n+|o+|p+|q+|r+|s+|t+|u+|v+|w+|x+|y+|z+|\*+|\.+)$''',`.

We have a test that generates a generic api key that looks like this `token:AAAAAAAAAAAAAAA`. So we need to find a pattern that would need to be caught.
 
The you need to do the changes [here](https://github.com/grafana/alloy/blob/8942fb21932a0ad29d6e092558c562be8e30b4d3/internal/component/loki/secretfilter/secretfilter_test.go#L185)

andrejshapal:
> Hey there are a test failing due to changes in the global allow list, specifically this regex `'''^(?i:a+|b+|c+|d+|e+|f+|g+|h+|i+|j+|k+|l+|m+|n+|o+|p+|q+|r+|s+|t+|u+|v+|w+|x+|y+|z+|\*+|\.+)$''',`.
> 
> We have a test that generates a generic api key that looks like this `token:AAAAAAAAAAAAAAA`. So we need to find a pattern that would need to be caught. I pretty sure just adding a numeric character would make it to pass all allow filters and be detected as a token
> 
> Then you need to do the changes [here](https://github.com/grafana/alloy/blob/8942fb21932a0ad29d6e092558c562be8e30b4d3/internal/component/loki/secretfilter/secretfilter_test.go#L185)

Sure thing, will do soon.

andrejshapal:
@kalleep Need some approval.

kalleep:
Hey everything looks good except that our fuzzing is failing. I fetched the seed and tried it locally but that passes.

But I was able to generate need seeds locally by checking out this pr and run `go test -v -fuzz="FuzzProcessEntry\$" -run="FuzzProcessEntry\$" --fuzztime="5m" -parallel=16 .` I am pretty sure we are hitting https://github.com/golang/go/issues/56238. Internally there is a hardcoded timeout for each test of [10s](https://github.com/golang/go/blob/master/src/internal/fuzz/worker.go#L492). With this new files we have more regexps so I am pretty sure this is happening. 

I will look into rewriting this test a bit because we are calling proccessEntry for 8 different components within a single run, and now with more regexes after the update this can clearly take longer when testing different configurations within a single test run

andrejshapal:
@kalleep Thank you for your help. I am not familiar with buzz at this point. Locally it was passing as you said. 
Regarding secret-filter parsing time, this is not just buzz issue and I would like to discuss this:
https://github.com/grafana/alloy/issues/3694
I found the secret-filter handful, but I am wondering how to make it safe for alloy. Regex operations are super greedy.

kelnage (approved):
LGTM! In general, end-users can use their own `gitleaks.toml` file using [the `gitleaks_config` component argument](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.secretfilter/#arguments), but updating to reflect improvements to the regular expressions is not a bad idea generally.

andrejshapal:
@kelnage Will you merge it later as for me "Merging is blocked"?

dehaansa:
While this says it can merge automatically, it looks like the changelog may be merged incorrectly. Can you merge main/rebase to ensure the changelog entry goes into the unreleased section?

andrejshapal:
@dehaansa done

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ Main (unreleased)
 
 - (_Experimental_) `prometheus.write.queue` initialize queue metrics that are seconds values as time.Now, not 0. (@dehaansa)
 
+- Update secret-filter gitleaks.toml from v8.19.0 to v8.26.0
```
dehaansa:
```suggestion
- Update secret-filter gitleaks.toml from v8.19.0 to v8.26.0 (@andrejshapal)
```

PR #3685, opened 2025-05-23
General updates and corrections to the Alloy Component docs part 3

clayton-cornell:
This PR is part of a series of PRs that aim to catch a few minor typos and errors that have crept into the docs through all the recent refactoring and updates.

Most corrections are minor:

- fix occasional typos
- sort out some table formatting that was missed or has shown up from other unrelated edits
- remove some extra spaces at the start/end of lines

This PR also includes validating the args etc. against the source files (as a double check to make sure everything matches). Some corrections include correcting data types to match what's documented and removing quotes from default values on int and duration types.


dehaansa (approved):
LGTM

PR #3700, opened 2025-05-26
fix: datadog url env var

ampc:
#### PR Description
Fix Datadog URL environment variable

#### PR Checklist

- [x] Documentation added


clayton-cornell (approved):
Thanks for catching and fixing this!

PR #3706, opened 2025-05-26
General updates and corrections to the Alloy Component docs part 7

clayton-cornell:
This PR is part of a series of PRs that aim to catch a few minor typos and errors that have crept into the docs through all the recent refactoring and updates.

Most corrections are minor:

- fix occasional typos
- sort out some table formatting that was missed or has shown up from other unrelated edits
- remove some extra spaces at the start/end of lines

This PR also includes validating the args etc. against the source files (as a double check to make sure everything matches). Some corrections include correcting data types to match what's documented and removing quotes from default values on int and duration types.


docs/sources/reference/components/prometheus/prometheus.relabel.md
```diff
@@ -48,14 +48,14 @@ prometheus.relabel "<LABEL>" {
...
 
-| Name             | Type                    | Description                                                             | Default | Required |
-| ---------------- | ----------------------- | ----------------------------------------------------------------------- | ------- | -------- |
-| `forward_to`     | `list(MetricsReceiver)` | Where the metrics should be forwarded to, after relabeling takes place. |         | yes      |
-| `max_cache_size` | `int`                   | The maximum number of elements to hold in the relabeling cache.         | 100,000 | no       |
+| Name             | Type                    | Description                                                             | Default  | Required |
+| ---------------- | ----------------------- | ----------------------------------------------------------------------- | -------- | -------- |
+| `forward_to`     | `list(MetricsReceiver)` | Where the metrics should be forwarded to, after relabeling takes place. |          | yes      |
+| `max_cache_size` | `int`                   | The maximum number of elements to hold in the relabeling cache.         | `100000` | no       |
```
clayton-cornell, line 54: Is this correct to drop the comma in the number?

docs/sources/reference/components/prometheus/prometheus.relabel.md
```diff
@@ -48,14 +48,14 @@ prometheus.relabel "<LABEL>" {
...
 
 ## Blocks
 
-You can use the following blocks with `prometheus.relabel`:
+You can use the following block with `prometheus.relabel`:
```
dehaansa, line 58: Should this be pluralized, as you can provide multiple `rule` blocks? Either way, we may want to consider updating all the block tables to show when you can use multiples 🙃

dehaansa (approved):

docs/sources/reference/components/prometheus/prometheus.relabel.md
clayton-cornell: I thought about that... but even if you're giving multiple `rule` blocks, you're still using a single block type/name... it's a 50/50 style/language choice.

PR #3707, opened 2025-05-27
chore: add fix tests flag converters

kalleep:
#### PR Description
Add flag `-fix-tests` flag to our converter tests. When running e.g. `go test ./internal/converter/internal/otelcolconvert/... -fix-tests` we will re-generate snapshot files based on diagnostic and alloy output.

Part of the tests are loading any generated alloy config stored on disk, so for now we won't generate any alloy output if we have a critical error, that would prevent test from passing because the component won't be able to load.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/converter/internal/otelcolconvert/otelcolconvert_test.go
```diff
@@ -3,12 +3,16 @@
...
 import (
+	"flag"
 	"testing"
 
 	"github.com/grafana/alloy/internal/converter/internal/otelcolconvert"
 	"github.com/grafana/alloy/internal/converter/internal/test_common"
 )
 
+// Set this flag to update snapshots e.g. `go test -v ./interal/converter/internal/otelcolconverter/...` -fix-tests
+var fixTestsFlag = flag.Bool("fix-tests", false, "update the test files with the current generated content")
```
dehaansa, line 14: What led you to choose -fix-tests over the original commented -update? Upstream OTel does -update, I'm not sure if this is a wider convention or if you're aware of a convention, IMO would like to follow convention if there is one.

kalleep: From here https://github.com/grafana/alloy/pull/3220#discussion_r2048680571. When I added a this for validation it was pointed out that we already had a place that used -fix-tests so I added it to be consistent. We can change that flag but then we should do so in all places

wildum (approved):

Issue #3696, opened 2025-05-26
Github actions: Test action does not fail if tests in syntax package fails


kalleep:
### What's wrong?

https://github.com/grafana/alloy/actions/runs/15253262308/job/42894791778

PR #3708, opened 2025-05-27
fix syntax tests in ci

kalleep:
#### PR Description
If any tests in syntax module failed in GHA or Drone the pipeline would still succeed. 

Not really sure why status code from `-execdir` is not propagated correctly.

Fixes: #3696

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


Makefile
```diff
@@ -161,7 +161,7 @@ run-alloylint: alloylint
 test:
```
wildum, line 161: the comment above should be updated because it's not dynamically finding submodules to run the tests anymore.

wildum (approved):
LGTM, thanks for catching this, just one comment update

PR #3722, opened 2025-05-29
General updates and validation of config blocks documentation

clayton-cornell:
This PR is a basic update to the config-block documentation. The focus is a verification against the source as well as a general cleanup of small typos and errors as well as alignment with the component doc styles.

### DIFF
--- a/docs/sources/reference/config-blocks/logging.md
+++ b/docs/sources/reference/config-blocks/logging.md
@@ -1,14 +1,69 @@
 ---
 canonical: https://grafana.com/docs/alloy/latest/reference/config-blocks/logging/
 description: Learn about the logging configuration block
-menuTitle: logging
-title: logging block
+labels:
+  stage: general-availability
+  products:
+    - oss
+title: logging
 ---
 
-# logging block
+# `logging`
 
 `logging` is an optional configuration block used to customize how {{< param "PRODUCT_NAME" >}} produces log messages.
 `logging` is specified without a label and can only be provided once per configuration file.
+
+## Usage
+
+```alloy
+logging {
+
+}
+```
+
+## Arguments
+
+You can use the following arguments with `logging`:
+
+| Name       | Type                 | Description                                | Default    | Required |
+| ---------- | -------------------- | ------------------------------------------ | ---------- | -------- |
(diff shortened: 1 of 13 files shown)


docs/sources/reference/config-blocks/logging.md
```diff
@@ -1,66 +1,78 @@
...
+| ---------- | -------------------- | ------------------------------------------ | ---------- | -------- |
+| `format`   | `string`             | Format to use for writing log lines        | `"logfmt"` | no       |
+| `level`    | `string`             | Level at which log lines should be written | `"info"`   | no       |
+| `write_to` | `list(LogsReceiver)` | List of receivers to send log entries to   | `[]`       | no       |
 
 ### Log level
 
 The following strings are recognized as valid log levels:
 
+* `"debug"`: Write all logs, including _debug_ level logs.
```
dehaansa: Alpha-sorting this list doesn't quite feel right, as it was previously hierarchically ordered. If you want to enforce alpha everywhere it's fine, but IMO it's not better.

docs/sources/reference/config-blocks/remotecfg.md
```diff
@@ -70,48 +62,64 @@ The `poll_frequency` must be set to at least `"10s"`.
...
 
-* [`bearer_token` argument][arguments].
-* [`bearer_token_file` argument][arguments].
-* [`basic_auth` block][basic_auth].
-* [`authorization` block][authorization].
-* [`oauth2` block][oauth2].
+* [`authorization`][authorization] block
+* [`basic_auth`][basic_auth] block
+* [`bearer_token_file`][arguments] argument
+* [`bearer_token`][arguments]argument
```
dehaansa:
```suggestion
* [`bearer_token`][arguments] argument
```

docs/sources/reference/config-blocks/logging.md
clayton-cornell: We should do what makes the most sense. Alpha sorting was more to do with consistency, but it should never be a hard unbreakable rule. There are definitely times where other ways of presenting make more sense.  This is easy to revert to the original sorting.

clayton-cornell: Reverted this change in a new commit

PR #3725, opened 2025-05-30
Update aws_firehose collector to use CW event timestamps

### LINKED ISSUE #1460: AWS Firehose CloudWatchLogs not using internal timestamp from the event
### What's wrong?

The `loki.source.awsfirehose` has `use_incoming_timestamp` boolean attribute documented as:
> Whether or not to use the timestamp received from the request.

I was trying to have the original CloudWatch Log event `timestamp` used - which is what I would expect by setting the `use_incoming_timestamp` attribute to `true` but that's simply not the case based on my understanding of the source code.

In [aws_filrehose handler](https://github.com/grafana/alloy/blob/v1.3.0/internal/component/loki/source/aws_firehose/internal/handler.go#L280) I see the timestamp is always passed through to the `handleCloudwatchLogsRecord` function and it is either the current timestamp or, in case of `use_incoming_timestamp:true`, the timestamp in the Firehose request.

Is there any reason why the actual _event_ timestamp is never used, even though it is declared in the [CloudwatchLogEvent structure](https://github.com/grafana/alloy/blob/v1.3.0/internal/component/loki/source/aws_firehose/internal/model.go#L47)?

Or maybe it would also make sense to have another configuration attribute which would imply using the source event timestamp?

### Steps to reproduce

In any configuration of Grafana Alloy I don't think it's possible to get the original CloudWatch Log event `timestamp` when using the AWS Firehose as source because it's ignored - see linked code in the description.

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

michaelPotter:
#### PR Description
Currently, if the `useIncomingTS` flag is set to true, the timestamps of all incoming log entries are set to the timestamp provided in the firehose envelope. This timestamp represents the moment when AWS Firehose shipped a batch of log entries rather than the timestamp of the individual entries themselves.

This change updates the loki.aws_firehose collector to now use the timestamps attached to specific log entries (when processing cloudwatch logs-subscription data) instead. For direct put data, where entry-specific timestamps are not guaranteed, the timestamp attached to the firehose wrapper is still used.

Fixes #1460 

#### Notes to the Reviewer

In my opinion this is a more expected behavior of the useIncomingTs flag, but I am open to feedback if there's a preferred way to attach the event timestamps (e.g. another config flag or use of labels as suggested in the issue).

Thanks for your time!

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ Main (unreleased)
 
 - Add `hash_string_id` argument to `foreach` block to hash the string representation of the pipeline id instead of using the string itself. (@wildum)
 
+### Bugfixes
+
+- Fix `loki.source.firehose` to propagate specific cloudwatch event timestamps when useIncomingTs is set to true. (@michaelPotter)
+
 v1.9.0
 -----------------
 
--- a/internal/component/loki/source/aws_firehose/internal/handler.go
+++ b/internal/component/loki/source/aws_firehose/internal/handler.go
@@ -275,7 +275,7 @@ func (h *Handler) handleCloudwatchLogsRecord(ctx context.Context, data []byte, c
 
 	for _, event := range cwRecord.LogEvents {
 		if h.useIncomingTs {
-			timestamp = time.Unix(event.Timestamp/millisecondsPerSecond, 0)
+			timestamp = time.UnixMilli(event.Timestamp)
 		}
 		h.sender.Send(ctx, loki.Entry{
 			Labels: h.postProcessLabels(cwLogsLabels.Labels()),
--- a/internal/component/loki/source/aws_firehose/internal/handler.go
+++ b/internal/component/loki/source/aws_firehose/internal/handler.go
@@ -274,6 +274,9 @@ func (h *Handler) handleCloudwatchLogsRecord(ctx context.Context, data []byte, c
 	cwLogsLabels.Set("__aws_cw_msg_type", cwRecord.MessageType)
 
 	for _, event := range cwRecord.LogEvents {
+		if h.useIncomingTs {
+			timestamp = time.Unix(event.Timestamp/millisecondsPerSecond, 0)
+		}
 		h.sender.Send(ctx, loki.Entry{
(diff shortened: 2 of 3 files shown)


michaelPotter:
I've rebased my changes to the tip of main to resolve the pending merge conflict.

dehaansa (approved):
LGTM, thanks for the fix. Assuming CI passes, we'll merge.

michaelPotter:
Thanks for the review dehaansa.

I think the test failure here may be unrelated? The firehose tests look like they passed:
```
ok  	github.com/grafana/alloy/internal/component/loki/source/aws_firehose	4.688s
ok  	github.com/grafana/alloy/internal/component/loki/source/aws_firehose/internal	1.151s
```

The error seems to be with a cloudflare component:
```
level=warn msg="failed iterating over logs" err=error start=1970-01-01T00:30:00Z end=1970-01-01T01:00:00Z retries=0 lineRead=1
level=error msg="failed to pull logs" err="5 errors: no logs; no logs; no logs; no logs; no logs" start=1970-01-01T00:59:00Z end=1970-01-01T01:00:00Z
--- FAIL: Test_CloudflareTargetError (0.00s)
    target_test.go:227: 
        	Error Trace:	/home/runner/work/alloy/alloy/internal/component/loki/source/cloudflare/internal/cloudflaretarget/target_test.go:227
        	Error:      	Should be true
        	Test:       	Test_CloudflareTargetError
level=warn msg="failed iterating over logs, out of cloudflare range, not retrying" err="HTTP status 400: bad query: error parsing time: invalid time range: too early: logs older than 168h0m0s are not available" start=1970-01-01T00:59:00Z end=1970-01-01T00:59:20Z retries=0
```

Is there anything I can do to help? Should I rebase again?

dehaansa:
We'll try re-running them, otherwise it will have to be fixed in another PR to get this merged. I'll follow up on this if that's a persistent failure on the cloudflare target.

PR #3727, opened 2025-06-02
ci: stop using windows-2019 runners

zerok:
The windows-2019 runners are deprecated as of 2025-06-01 and will be fully unsupported by 2025-06-30. There will be multiple brown-outs over the course of this month which might affect builds and releases unless we remove those runners.

See https://github.com/actions/runner-images/issues/12045 for details.

### DIFF
--- a/.github/workflows/check-windows-build-image.yml
+++ b/.github/workflows/check-windows-build-image.yml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   check-windows-build-image:
-    runs-on: windows-2019
+    runs-on: windows-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/publish-alloy-windows.yml
+++ b/.github/workflows/publish-alloy-windows.yml
@@ -24,7 +24,7 @@ jobs:
     name: Publish Alloy Windows container
     strategy:
       matrix:
-        os: [windows-2022, windows-2019]
+        os: [windows-2022, windows-2025]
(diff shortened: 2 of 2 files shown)


zerok:
Unfortunately I have no idea if this affects the releases themselves but since the binaries are built using just the Go toolchain I'm hoping this is fine 🙂

ptodev:
Hi @zerok, thank you for the PR! I think it would be ok to merge it as long as it has a changelog entry. Would you mind adding one please?

It's not ideal remove that container since normally Alloy supports all OS versions officially supported by the OS vendor. Windows 2019 will still be supported until 2029. That said, it's probably not a huge issue because if a user is using containers then they are probably good at updating their OS 😅  Alloy itself will still work on Windows 2019 even though we won't distribute a container image.

ptodev:
@zerok Oh wait I just realised that `tools/ci/docker-containers-windows` uses the `nanoserver:ltsc2019` base image when running on the Windows 2025 build server. Now your comment makes sense. It'd be good to make sure we can really run this container on Windows 2019... I'll see if I can test this somehow.

ptodev:
Unfortunately, it's not possible to build images for Windows 2019 on new versions of Windows like this. Docker needs the host OS to match the base OS of the image we are building. To get around this, we would need to run Docker with HyperV. I've been experimenting with this on #3913, but I hit a Docker issue.

zerok:
Ah, too bad 🙁 The only other option I see is to have a self-hosted runner somewhere that runs Windows 2019. Keeping that one secure would be a bit of an undertaking, though 😒

ptodev:
We will fix this as part of #3971

PR #3766, opened 2025-06-03
Accept optional strings as number arguments

ptodev:
#### PR Description

Currently, if you try to assign a non-secret optional secret to a `number` argument, you'll get an error `should be number, got capsule`. For example, this generates an error:

```
local.file "test" {
  filename  = "/Users/paulintodev/Desktop/test_input.txt"
  is_secret = false
}

  endpoint {
    ...
    queue_config {
        max_samples_per_send = local.file.samples.content
    }
  }
}
```

On the other hand, this wouldn't generate an error:
```
  endpoint {
    ...
    queue_config {
        max_samples_per_send = "1000"
    }
  }
}
```

The behaviour of a non-secret optional secret and that of a string needs to be consistent, so this PR will fix it.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/syntax/internal/value/value.go
+++ b/syntax/internal/value/value.go
@@ -545,16 +545,12 @@ func convertValue(val Value, toType Type) (Value, error) {
 	case TypeCapsule:
 		// Some capsules, such as optional secrects, may be convertible to a string.
 		// Try to convert them to a string and then rerun convertValue.
-		cc, ok := val.Interface().(ConvertibleIntoCapsule)
-		if ok {
-			into := reflect.New(reflect.TypeOf(string(""))).Elem()
-
-			err := cc.ConvertInto(into.Addr().Interface())
+		into := reflect.New(reflect.TypeOf(string(""))).Elem()
+		ok, err := tryCapsuleConvert(val, into, TypeString)
+		if ok && err == nil {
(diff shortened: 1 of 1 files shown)


syntax/internal/value/value.go
```diff
@@ -541,6 +541,22 @@ func convertValue(val Value, toType Type) (Value, error) {
 				return Uint(parsed), nil
 			}
 		}
+
+	case TypeCapsule:
+		// Some capsules, such as optional secrects, may be convertible to a string.
+		// Try to convert them to a string and then rerun convertValue.
+		cc, ok := val.Interface().(ConvertibleIntoCapsule)
```
dehaansa: Should this reuse the `tryCapsuleConvert` function from `syntax/internal/value/decode.go`?

thampiotr: +1 I think that would be the right code path - question is why it doesn't work currently, perhaps some reflection code subtle issue

ptodev: I updated the code to use `tryCapsuleConvert `. I'd actually tried doing this originally, but I felt like I was going through too many hoops. After making the change now I feel like it looks ok though.

@thampiotr the reason why it doesn't work currently is because `convertValue` doesn't handle capsules.

PR #3770, opened 2025-06-04
Add unit tests for remote.kubernetes components

ptodev:
#### PR Description

Adding unit tests for `remote.kubernetes` components which use a fake K8s to test whether the correct exports are produced.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/remote/kubernetes/internal/test_util/util.go
+++ b/internal/component/remote/kubernetes/internal/test_util/util.go
@@ -1,0 +1,111 @@
+package test_util
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/go-kit/log"
+	kubernetes_common "github.com/grafana/alloy/internal/component/common/kubernetes"
+	"github.com/grafana/alloy/internal/component/remote/kubernetes"
+	"github.com/grafana/alloy/internal/runtime/componenttest"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/grafana/alloy/internal/util"
+	"github.com/grafana/alloy/syntax"
+	"github.com/grafana/dskit/backoff"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/runtime"
+	client_go "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+type FakeClientBuidler struct {
+	KubeObjects []runtime.Object
+}
+
+func (b FakeClientBuidler) GetKubernetesClient(_ log.Logger,
+	_ *kubernetes_common.ClientArguments) (client_go.Interface, error) {
+
+	fakeClientset := fake.NewClientset(b.KubeObjects...)
+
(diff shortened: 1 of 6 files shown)


internal/component/remote/kubernetes/internal/test_util/util.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (b FakeClientBuidler) GetKubernetesClient(_ log.Logger,
+	_ *kubernetes_common.ClientArguments) (client_go.Interface, error) {
+
+	fakeClientset := fake.NewClientset(b.KubeObjects...)
+
+	return fakeClientset, nil
+}
+
+func Eventually(t *testing.T, min, max time.Duration, retries int, f func() error) {
```
thampiotr, line 35: Any reason why not using the other implementations we have available of similar function?

internal/component/remote/kubernetes/internal/test_util/util.go
```diff
@@ -0,0 +1,111 @@
...
+		Eventually(t.T, 10*time.Millisecond, 10*time.Second, 5, func() error {
+			actual := ctrl.Exports().(kubernetes.Exports)
+
+			if len(actual.Data) != len(expect.Data) {
+				return fmt.Errorf("expected %#v piece of data, got %#v", len(expect.Data), len(actual.Data))
+			}
+
+			if !reflect.DeepEqual(expect.Data, actual.Data) {
+				return fmt.Errorf("expected %#v, got %#v", expect, actual)
+			}
```
thampiotr, line 105: Nit: there's a slight advantage of using `require` or `assert` packages because they will print a nice diff for many data types when they are not equal that helps to spot where the issue is.

internal/component/remote/kubernetes/kubernetes.go
```diff
@@ -99,15 +102,19 @@ func New(opts component.Options, args Arguments, rType ResourceType) (*Component
...
-	if err := c.Update(args); err != nil {
-		return nil, err
-	}
 	return c, nil
 }
 
 // Run starts the remote.kubernetes.* component.
 func (c *Component) Run(ctx context.Context) error {
+	// Do the initial Update() in Run() instead of in New().
+	// That way New() is stateless and has no side effects.
```
thampiotr, line 111: Is it because of the initial poll or does the `GetKubernetesClient` try to talk to k8s on creation and can return errors if misconfigured? Not sure how `New()` is stateful, seems to me that multiple calls to `New()` will not modify internal state, but maybe I'm missing something. 

By moving this from `New` to `Run`, we change the way errors will be handled, no? Errors in `New` I think generally give a more detailed information back to the user and can stop applying a new config?

ptodev: Sorry, "stateful" is the wrong word. What I meant is that `New()` won't have side effects. It makes it easier to substitute parts of the component with fake ones. I think it also aligns better with how Alloy components are meant to be designed. Normally you wouldn't want a `New` function to connect to external services, to create files, etc.

PR #3779, opened 2025-06-05
`loki.secretfilter`: Add support for entropy

romain-gaillard:
#### PR Description

This PR adds support for entropy-based filtering in the `loki.secretfilter` component. When enabled, and if an entropy threshold is defined in the config file for a specific rule, it will compare the entropy of the identified potential secret with the threshold and only redact it if the entropy is high enough. This feature works in a similar way as in Gitleaks.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -50,14 +50,15 @@ You can use the following arguments with `loki.secretfilter`:
...
 
 The `gitleaks_config` argument is the path to the custom `gitleaks.toml` file.
 If you don't provide the path to a custom configuration file, the Gitleaks configuration file [embedded in the component][embedded-config] is used.
 
 {{< admonition type="note" >}}
 This component doesn't support all the features of the Gitleaks configuration file.
 It only supports regular expression-based rules, `secretGroup`, and allowlist regular expressions. `regexTarget` only supports the default value `secret`.
-Other features such as `keywords`, `entropy`, `paths`, and `stopwords` aren't supported.
+Other features such as `keywords`, `paths`, and `stopwords` aren't supported.
 The `extend` feature isn't supported.
```
kelnage, line 63: Why is this separate from the other features that aren't supported?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -50,14 +50,15 @@ You can use the following arguments with `loki.secretfilter`:
 | `partial_mask`    | `int`                | Show the first N characters of the secret.                     | `0`                                | no       |
 | `redact_with`     | `string`             | String to use to redact secrets.                               | `"<REDACTED-SECRET:$SECRET_NAME>"` | no       |
 | `types`           | `map(string)`        | Types of secret to look for.                                   | All types                          | no       |
+| `enable_entropy`  | `bool`               | Enable entropy-based filtering.                                | `false`                            | no       |
```
clayton-cornell: To be consistent, can you alphabetize this? So, that means moving it up between `allowlist` and `gitleaks_config`.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -100,6 +101,8 @@ If set to `0`, the entire secret is redacted.
 If a secret isn't at least 6 characters long, it's entirely redacted.
 For short secrets, at most half of the secret is shown.
 
+The `enable_entropy` argument enables entropy-based filtering. When set to `true`, the entropy of the detected potential secret is calculated and compared against the threshold provided for the matching rule in the configuration file. The secret is then redacted only if the entropy is above the threshold. This can help reduce false positives.
```
```suggestion
The `enable_entropy` argument enables entropy-based filtering.
When you set this to `true`, the entropy of the detected potential secret is calculated and compared against the threshold provided for the matching rule in the configuration file.
The secret is then redacted only if the entropy is above the threshold.
This can help reduce false positives.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -89,8 +90,8 @@ Currently, the secret types known to have this behavior are: `aws-access-token`.
 
 The `redact_with` argument is a string that can use variables such as `$SECRET_NAME`, replaced with the matching secret type, and `$SECRET_HASH`, replaced with the SHA1 hash of the secret.
 
-The `include_generic` argument is a boolean that includes the generic API key rule in the Gitleaks configuration file if set to `true`.
-It's disabled by default because it can generate false positives.
+The `include_generic` argument is a boolean that enables the generic API key rule in the Gitleaks configuration file if set to `true`.
+It's disabled by default because it can generate false positives. Consider enabling entropy-based filtering if you enable this rule.
```
```suggestion
It's disabled by default because it can generate false positives.
Consider enabling entropy-based filtering if you enable this rule.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -125,13 +128,14 @@ The following fields are exported and can be referenced by other components:
...
+| `loki_secretfilter_secrets_redacted_total`                | Counter | Total number of secrets that have been redacted.                                                              |
+| `loki_secretfilter_secrets_redacted_by_rule_total`        | Counter | Number of secrets redacted, partitioned by rule name.                                                         |
+| `loki_secretfilter_secrets_redacted_by_origin`            | Counter | Number of secrets redacted, partitioned by origin label value.                                                |
+| `loki_secretfilter_secrets_allowlisted_total`             | Counter | Number of secrets that matched a rule but were in an allowlist, partitioned by source.                        |
+| `loki_secretfilter_processing_duration_seconds`           | Summary | Summary of the time taken to process and redact logs in seconds.                                              |
+| `loki_secretfilter_secrets_skipped_entropy_by_rule_total` | Counter | Number of secrets that matched a rule but whose entropy was too low to be redacted, partitioned by rule name. |
```
```suggestion
| `loki_secretfilter_processing_duration_seconds`           | Summary | Summary of the time taken to process and redact logs in seconds.                                              |
| `loki_secretfilter_secrets_allowlisted_total`             | Counter | Number of secrets that matched a rule but were in an allowlist, partitioned by source.                        |
| `loki_secretfilter_secrets_redacted_by_origin`            | Counter | Number of secrets redacted, partitioned by origin label value.                                                |
| `loki_secretfilter_secrets_redacted_by_rule_total`        | Counter | Number of secrets redacted, partitioned by rule name.                                                         |
| `loki_secretfilter_secrets_redacted_total`                | Counter | Total number of secrets that have been redacted.                                                              |
| `loki_secretfilter_secrets_skipped_entropy_by_rule_total` | Counter | Number of secrets that matched a rule but whose entropy was too low to be redacted, partitioned by rule name. |
```

Just sorting rows alphabetically

docs/sources/reference/components/loki/loki.secretfilter.md
romain-gaillard: Thanks, done in 4656b70450c483bf9fa2fb22b0d0a046fa939a6e

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -516,4 +543,24 @@ func (c *Component) Update(args component.Arguments) error {
...
+	frequences := make(map[rune]int)
+
+	// Calculate the frequency of each rune in the string
+	for _, char := range str {
+		frequences[char]++
+	}
+
+	// Calculate the entropy using the frequency of each rune
+	for _, freq := range frequences {
+		probability := float64(freq) / float64(len(str))
```
kelnage: It's a very minor detail, but the [gitleaks implementation](https://github.com/gitleaks/gitleaks/blob/958f55ac629b5351bb7b9c80f00c10823fef64d3/detect/utils.go#L127) divides 1 by `len(str)` and multiplies each frequency by it. I think this is for performance (FP multiplication is significantly faster than division) - but I suppose it _could_ also have an affect on the FP calculation too, so might be worthwhile doing either way?

kelnage (commented):
Couple of minor points - one regarding docs, the other about floating point arithmetic :fearful: :sweat_smile:

docs/sources/reference/components/loki/loki.secretfilter.md
romain-gaillard: You mean `extend`? Because the other elements are rule-specific and this one is configuration-wide.

Basically the docs explain that it supports regular expression-based rules and gives what is supported (to a certain extent for `regexTarget` which only supports the default value). Then it lists what it doesn't support for rules, and then other non-rule related settings (like `extend`).

internal/component/loki/secretfilter/secretfilter.go
romain-gaillard: Thanks! Implemented in 10df07f48beb7d80b298941863680ee4f4250d5a

clayton-cornell:
Docs look good to me :-)

romain-gaillard:
> LGTM! Flakey e2e tests causing the fails I think?

Thanks! Yep, looks to be happening on main too.

clayton-cornell (approved):

thampiotr (approved):

thampiotr:
> > LGTM! Flakey e2e tests causing the fails I think?
> 
> Thanks! Yep, looks to be happening on main too.

Integration tests are fixed in main.

PR #3801, opened 2025-06-09
database_observability: Add explain plan collector

rgeyer:
#### PR Description

Adds a new explain plan collector to the dbo11 component.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Main (unreleased)
 
 - Add `hash_string_id` argument to `foreach` block to hash the string representation of the pipeline id instead of using the string itself. (@wildum)
 - Update `async-profiler` binaries for `pyroscope.java` to 4.0-87b7b42 (@github-hamza-bouqal)
+- (_Experimental_) Add `explain_plan` collector to `database_observability.mysql` component. (@rgeyer)
```
matthewnolf: Can you put this under a header for the db o11y component. [example](https://github.com/grafana/alloy/pull/3801/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL89-L99)

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
+	"github.com/grafana/alloy/internal/runtime/logging"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const (
+	OP_EXPLAIN_PLAN        = "explain_plan"
+	OP_EXPLAIN_PLAN_OUTPUT = "explain_plan_output"
+	ExplainPlanName        = "explain_plan" // This could be OP_EXPLAIN_PLAN?
```
matthewnolf: nit: Let's remove this comment, having them separate means we can independently change the logging and config keywords

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+}
+
+type PlanNode struct {
+	Operation ExplainPlanOutputOperation `json:"operation"`
+	Details   NodeDetails                `json:"details"`
+	Children  []PlanNode                 `json:"children,omitempty"`
+}
+
+type NodeDetails struct {
+	EstimatedRows int64                     `json:"estimatedRows"`
```
matthewnolf: We're mixing snake case and camel case - is that expected?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		}
+		planNode.Children = append(planNode.Children, childNode)
+	})
+	if err != nil {
+		return planNode, err
+	}
+	return planNode, nil
+}
+
+func NewExplainPlan(args ExplainPlanArguments) (*ExplainPlan, error) {
```
matthewnolf: Can you move this next to the type definition https://github.com/grafana/alloy/pull/3801/files#diff-cc02483faf7ed483c3a3a921ade523b48e22702b5609e4cee5461cf02fb362fdR64

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		defer rs.Close()
+
+		// Populate cache
+		for rs.Next() {
+			if err := rs.Err(); err != nil {
+				level.Error(c.logger).Log("msg", "failed to iterate rs digests for explain plans", "err", err)
+				return err
+			}
+
+			var qi queryInfo
```
matthewnolf: Is `queryInfo` used elsewhere? Can we move the definition closer to the usage

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+			if err = rs.Scan(&qi.schemaName, &qi.digest, &qi.queryText); err != nil {
+				level.Error(c.logger).Log("msg", "failed to scan digest for explain plans", "err", err)
+				return err
+			}
+			c.queryCache = append(c.queryCache, qi)
+		}
+		// Calculate batch size based on current cache size
+		c.currentBatchSize = int(float64(len(c.queryCache)) * c.perScrapeRatio)
+		if c.currentBatchSize < 1 {
+			c.currentBatchSize = 1 // Ensure we process at least one query
```
matthewnolf: Could we add 1 in the previous calculation?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+}
+
+func (c *ExplainPlan) Stop() {
+	c.cancel()
+}
+
+func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
+	// If cache is empty, fetch all available queries
+	if len(c.queryCache) == 0 {
+		rs, err := c.dbConnection.QueryContext(ctx, selectDigestsForExplainPlan)
```
matthewnolf: It might be worth breaking this into a separate function like `populateCache`

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		c.currentBatchSize = int(float64(len(c.queryCache)) * c.perScrapeRatio)
+		if c.currentBatchSize < 1 {
+			c.currentBatchSize = 1 // Ensure we process at least one query
+		}
+		level.Info(c.logger).Log("msg", "fetched digests", "count", len(c.queryCache), "batch_size", c.currentBatchSize)
+	}
+
+	// Process up to batchSize queries from cache
+	processedCount := 0
+	for i, qi := range c.queryCache {
```
matthewnolf: Do we need `i`?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		logger := log.With(c.logger, "digest", qi.digest)
+		if processedCount >= c.currentBatchSize {
+			break
+		}
+
+		processedCount++
+
+		// Defer deletion of current item from cache and increment processed count
+		defer func(index int) {
```
matthewnolf: Based on the comment, should L596 be within the deferred func?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+
+		// Skip non-select queries
+		if !strings.HasPrefix(strings.ToLower(qi.queryText), "select") {
+			parser := parser.NewTiDBSqlParser()
+			redacted, err := parser.Redact(qi.queryText)
+			if err != nil {
+				level.Error(logger).Log("msg", "failed to redact sql", "err", err)
+				continue
+			}
+			level.Debug(logger).Log("msg", "skipping non-select query", "query_text", redacted)
```
matthewnolf: What's the reason to log the redact and then log the query here?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+			if _, err := c.dbConnection.ExecContext(ctx, "USE "+*qi.schemaName); err != nil {
+				level.Error(logger).Log("msg", "failed to set schema", "err", err)
+				continue
+			}
+		}
+
+		rsExplain := c.dbConnection.QueryRowContext(ctx, selectExplainPlansPrefix+qi.queryText)
+
+		if err := rsExplain.Err(); err != nil {
+			level.Error(logger).Log("msg", "failed to iterate rs for explain plan", "err", err)
```
matthewnolf:
```suggestion
			level.Error(logger).Log("msg", "failed to run explain plan", "err", err)

```

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+			continue
+		}
+
+		var byteExplainPlanJSON []byte
+		if err := rsExplain.Scan(&byteExplainPlanJSON); err != nil {
+			level.Error(logger).Log("msg", "failed to scan explain plan json", "err", err)
+			continue
+		}
+
+		generatedAt := time.Now().Format(time.RFC3339)
```
matthewnolf: Maybe this should happen after error checking?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+			loggedSchemaName = *qi.schemaName
+		}
+
+		redactedByteExplainPlanJSON, _, err := RedactAttachedConditions(byteExplainPlanJSON)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to redact explain plan json", "err", err)
+			continue
+		}
+
+		level.Debug(logger).Log("msg", "explain plan output",
```
matthewnolf: Do we need this?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+	}
+
+	planNode, err := parseTopLevelPlanNode(logger, qblock)
+	if err != nil {
+		return nil, err
+	}
+	output.Plan = planNode
+
+	return output, nil
+}
```
matthewnolf: Do these types need to be exposed? Can't we make them package private

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		}
+
+		logMessage := fmt.Sprintf(
+			`schema="%s" digest="%s" explain_plan_output="%s"`,
+			loggedSchemaName,
+			qi.digest,
+			base64.StdEncoding.EncodeToString(explainPlanOutputJSON),
+		)
+
+		c.entryHandler.Chan() <- buildLokiEntryWithTimestamp(
```
matthewnolf: You can use `buildLokiEntry` if you're not concerned with setting a custom timestamp for the entry

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+			logging.LevelInfo,
+			OP_EXPLAIN_PLAN_OUTPUT,
+			c.instanceKey,
+			logMessage,
+			int64(time.Now().UnixNano()),
+		)
+		// TODO: Add context to logging when errors occur so the original node can be found.
```
matthewnolf: Is this still needed?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		// I.E. query_block->nested_loop[1]->table etc..
+	}
+
+	return nil
+}
+
+// This is presently only used in the tests. Later we'll probably use it to redact attached conditions before
```
matthewnolf: Is this comment still true?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+// debug logging the entire explain plan in alloy.
+func RedactAttachedConditions(explainPlanJSON []byte) ([]byte, int, error) {
```
matthewnolf: Can we also make this private?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+	c.ctx = ctx
+	c.cancel = cancel
+
+	go func() {
+		defer func() {
+			c.Stop()
+			c.running.Store(false)
+		}()
+
+		ticker := time.NewTicker(c.scrapeInterval)
```
matthewnolf: Looking at the config added - is there a specific config value for explain plans?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,754 @@
...
+		Metadata: MetadataInfo{
+			DatabaseEngine:  "MySQL",
+			DatabaseVersion: dbVersion,
+			QueryIdentifier: digest,
+			GeneratedAt:     generatedAt,
+		},
+	}
+
+	qblock, qblockType, _, err := jsonparser.Get(explainJson, "query_block")
+	if qblockType == jsonparser.NotExist {
```
matthewnolf: Is `qblockType` going to be set here? We're not handling the error first and instead only if `qblockType == jsonparser.NotExist`

CHANGELOG.md
rgeyer: Sure.. I suppose we're planning additional changes? Given this was only a single improvement I figured a single line was adequate, and could be updated when/if we add more features.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Good question.. This was based on the original design doc, and it is what @fabienne-m is using now. If we're going to make a change, it should probably be now.

Which do you prefer fab? Camel case seems to be the majority.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: I'm not sure I follow? This is the only place that the redacted query is logged.

The intent here was to keep track of what types of queries (besides SELECT) are being skipped.

The error log on 614 is only in the case that it's not possible to redact the query, in which case an error is logged and this query is skipped.

matthewnolf: Sorry my message was poorly formed. My question was why we're keeping a track of these? Is that something that is still needed?

rgeyer: Yeah, fair.. This is vestigial from early work. I think when/if we wish to track this, we'll create a metric for it. I.E. `database_o11y_explain_skipped_query_total{verb="INSERT, DELETE, ETC"}`

I'll clear this out.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: For now, yes.. For troubleshooting, since we're taking the raw explain plan, and refactoring it into our own schema, if we don't have a record of the raw explain plan, and our generated schema doesn't look right, there's no real way to compare and fix the issue.

It does require turning on debug logging for alloy. We *could* add a config option to only debug log certain digests, given that we'd probably be trying to target certain ones in a real-life scenario.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yes, for the same reason that we're debug logging the raw explain plan json.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yes, it's passed into the defer function to find the current index in the slice to delete.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yes, it's stored in the `ExplainPlan` struct, as the [queue which is burned down.](https://github.com/grafana/alloy/blob/d72f93e0cb2da7365fbbca6cb31cb434a7bb0b0c/internal/component/database_observability/mysql/collector/explain_plan.go#L70)

I suppose I could put the definition next to the `ExplainPlan` struct 🤷‍♂️

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yes, it is passed in [here](https://github.com/grafana/alloy/blob/d72f93e0cb2da7365fbbca6cb31cb434a7bb0b0c/internal/component/database_observability/mysql/component.go#L342), and assigned to the `ExplainPlan` struct [here](https://github.com/grafana/alloy/blob/d72f93e0cb2da7365fbbca6cb31cb434a7bb0b0c/internal/component/database_observability/mysql/collector/explain_plan.go#L503).

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Hmmn.. Good call.. The [interface](https://pkg.go.dev/github.com/buger/jsonparser#Get) *does* say that it will always be set. However, I'm not considering the case where the `explainJson` simply isn't a json string, or is otherwise unparsable.

I'll check `err` first. Good eye. Thanks!

rgeyer: Hmmn.. Actually, looks like it'll be fine. I'll add a couple tests with random strings etc.

https://go.dev/play/p/qhtlULqbWnS

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: I moved the structs down to where the methods for the ExplainPlan struct are defined. Not sure if that's idiomatic, possibly this is am implication that they should be in different files.

internal/component/database_observability/mysql/collector/explain_plan.go
matthewnolf, line 1: Can we add a respective test file `explain_plan_test` with tests for the collector?

rgeyer: Yes.. I have tests in the db-o11y-collector repo, tho they depend upon testdata files, and I'm not sure the precedent for that in the alloy repo.

I *can* refactor those to include the source test data in the test file itself as a string, tho it'd take a bit of doing. @cristiangreco and I had discussed getting this merged sooner than later for functionality, and backfilling that, which I fully intend to do.

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+		CURRENT_SCHEMA,
+		DIGEST,
+		SQL_TEXT
+	FROM performance_schema.events_statements_history
+	WHERE TIMER_END > DATE_SUB(NOW(), INTERVAL 1 DAY)
+	AND SQL_TEXT IS NOT NULL
+	AND DIGEST IS NOT NULL
+	AND CURRENT_SCHEMA NOT IN ('mysql', 'performance_schema', 'sys', 'information_schema');`
+
+const selectExplainPlansPrefix = `EXPLAIN FORMAT=JSON `
```
matthewnolf:
```suggestion
const selectExplainPlanPrefix = `EXPLAIN FORMAT=JSON `
```

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+
+const selectDbSchemaVersion = `SELECT VERSION()`
```
matthewnolf:
```suggestion
const selectDBSchemaVersion = `SELECT VERSION()`
```

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+		c.queryCache = append(c.queryCache, qi)
+	}
+	// Calculate batch size based on current cache size
+	c.currentBatchSize = int(math.Ceil(float64(len(c.queryCache)) * c.perScrapeRatio))
+	level.Info(c.logger).Log("msg", "fetched digests", "count", len(c.queryCache), "batch_size", c.currentBatchSize)
+	return nil
+}
+
+func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
+	// If cache is empty, fetch all available queries
```
matthewnolf: Some of the comments seem to be attached to code that is quite self describing. Could you take out all but the ones for cases where you think the code is not self describing?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+		if err := c.populateQueryCache(ctx); err != nil {
+			return err
+		}
+	}
+
+	// Process up to batchSize queries from cache
+	processedCount := 0
+	for i, qi := range c.queryCache {
+		logger := log.With(c.logger, "digest", qi.digest)
+		if processedCount >= c.currentBatchSize {
```
matthewnolf, line 587: We should do this before updating the logger?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+	KeyUsed       *string                   `json:"keyUsed,omitempty"`
+	JoinType      *string                   `json:"joinType,omitempty"`
+	JoinAlgorithm *explainPlanJoinAlgorithm `json:"joinAlgorithm,omitempty"`
+	Condition     *string                   `json:"condition,omitempty"`
+	GroupByKeys   []string                  `json:"groupByKeys,omitempty"`
+	SortKeys      []string                  `json:"sortKeys,omitempty"`
+	Warning       *string                   `json:"warning,omitempty"`
+}
+
+func NewExplainPlanOutput(logger log.Logger, dbVersion string, digest string, explainJson []byte, generatedAt string) (*explainPlanOutput, error) {
```
matthewnolf: Can we make this package private?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+	planNode, err := parseTopLevelPlanNode(logger, qblock)
+	if err != nil {
+		return nil, err
+	}
+	output.Plan = planNode
+
+	return output, nil
+}
+
+func parseTopLevelPlanNode(logger log.Logger, topLevelPlanNode []byte) (planNode, error) {
```
matthewnolf, line 139: Did you consider making the return type a pointer, so that we can return `nil, err` in case of an error?

internal/component/database_observability/mysql/collector/explain_plan.go
matthewnolf: While we may be okay now, there's no protection in case this behaviour changes in the future. I think we should still check the error first as a safety measure.

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+
+	if unionResult, _, _, err := jsonparser.Get(topLevelPlanNode, "union_result"); err == nil {
+		pnode, err := parseUnionResultNode(logger, unionResult)
+		if err != nil {
+			return planNode{}, err
+		}
+		return pnode, nil
+	}
+
+	return planNode{}, nil
```
matthewnolf: What happens here? We didn't understand what the top level node was, should we throw an error?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+	pnode := planNode{
+		Operation: explainPlanOutputOperationTableScan,
+		Details:   nodeDetails{},
+	}
+
+	// Check for join algorithm. Nested loop would be set in parseNestedLoopJoinNode, since not all table nodes are children of nested loop joins.
+	if joinAlgorithm, _, _, err := jsonparser.Get(tableNode, "using_join_buffer"); err == nil {
+		if string(joinAlgorithm) == "hash join" {
+			joinAlgorithmConst := explainPlanJoinAlgorithmHash
+			pnode.Details.JoinAlgorithm = &joinAlgorithmConst
```
matthewnolf, line 205:
```suggestion
			pnode.Details.JoinAlgorithm = &explainPlanJoinAlgorithmHash
```

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,750 @@
...
+			// In theory, this could be okay? Are there other things that could be in nested loop?
+			return
+		}
+		childDetails, err := parseTableNode(logger, tableNode)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to parse table node in nested loop join", "error", err)
+			return
+		}
+		if childDetails.Details.JoinAlgorithm == nil {
+			childDetails.Details.JoinAlgorithm = &algo // TODO:This is a duplicate from the parent node. Is that necessary?
```
matthewnolf: Are these questions (+L299) something we need to figure out now

internal/component/database_observability/mysql/collector/explain_plan.go
matthewnolf: This comment applies to all the parseTYPE functions actually

matthewnolf (commented):
Nice, looking good!

internal/component/database_observability/mysql/collector/explain_plan.go
fabienne-m: Yeah, please change it all to camelCase. In the frontend I already have it look like this: 

```
interface ExplainPlanMetadata {
  databaseEngine: string;
  databaseVersion: string;
  queryIdentifier: string;
  generatedAt: string;
}

interface ExplainPlanDetails {
  estimatedRows: number;
  estimatedCost?: number;
  tableName?: string;
  alias?: string;
  accessType?: string;
  keyUsed?: string;
  joinType?: string;
  joinAlgorithm?: string;
  condition?: string;
  groupByKeys?: string[];
  sortKeys?: string[];
  warning?: string;
}

interface ExplainPlanNode {
  operation: string;
  details: ExplainPlanDetails;
  children?: ExplainPlanNode[];
}

export interface ExplainPlan {
  metadata: ExplainPlanMetadata;
  plan: ExplainPlanNode;
}
```

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: There is a test for it, and we control when/if the dependency gets updated. The test would fail if the interface/contract of the library changed.

In any case.. The other invocations of this library do check the error, so having this one be different does stand out as apparently wrong. I'll make the change. :)

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: `explainPlanJoinAlgorithmHash` is an enum/const, and therefore you can not take the address of it. Thus why it gets assigned to a variable here, so that the assigned memory address can be fetched and assigned to the pnode struct.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Oops, thought I had in the previous batch.. Done.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yeah, fair.. Pretty sure those are vestigial from when I was first writing this, and left them as notes to myself on what to implement 😂

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: So, a couple of things are at play here.

First, "top level node" is perhaps misleading. We're actually looking for what type of operation this is, usually encapsulated in a top level query_block. However this get's called for other node/operation types which have nested operations, and those operations could, theoretically be any of the "top level" nodes.

See `parseGroupingOperationNode` through `parseUnionResultNode` where [this method is called](https://github.com/grafana/alloy/blob/b3bb8a3ae50376db9a115de313d5f0f1046ee1cb/internal/component/database_observability/mysql/collector/explain_plan.go#L345).

Second, we're only parsing parts of the explain plan which we *do* know about. It's entirely likely that there are properties which we *don't* know about. The result would be an empty node in the ExplainPlanOutput struct, and is a good reason why we still need the debug output of the raw explain plan, and the proposed `TODO` of providing context of [where errors occur](https://github.com/grafana/alloy/blob/b3bb8a3ae50376db9a115de313d5f0f1046ee1cb/internal/component/database_observability/mysql/collector/explain_plan.go#L679).

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: 🤔 Yeah, good question. As I went to answer this, I realized some refactoring was in order, but not necessarily returning a nil planNode.

Each of these parseTYPE methods are called only after we've discovered the json object which contains properties we want to extract. So we *know* there is a node there.

By returning the incomplete node, and the incomplete explain output, and logging that, it should actually go a long way toward solving the [TODO](https://github.com/grafana/alloy/blob/b3bb8a3ae50376db9a115de313d5f0f1046ee1cb/internal/component/database_observability/mysql/collector/explain_plan.go#L679 )I mentioned in another comment.

I put all of these changes in a [single commit](https://github.com/grafana/alloy/pull/3801/commits/2e20c95020b6735dc03c1319700cc284c9676a16). Let me know if this makes sense.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Right.. This was actually a question for @fabienne-m. I think we both agreed in the design document that the joinalgorithm should be assigned to both the parent, and child of a join. However this seems redundant to me now that it's been implemented.

Is the join algorithm required for both the parent and child? @fabienne-m ?

fabienne-m: I think only the parent node that joins children nodes, should have the join algorithm. A child of a join should only have a join algorithm if it itself is also a parent node that joins further children.

So no it's not required for both, just for the parent.

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,735 @@
...
+
+	"github.com/buger/jsonparser"
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
+	"github.com/grafana/alloy/internal/runtime/logging"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const (
+	OP_EXPLAIN_PLAN        = "explain_plan"
```
cristiangreco: Seems we don't use this const anymore?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,735 @@
...
+const selectDigestsForExplainPlan = `
+	SELECT
+		CURRENT_SCHEMA,
+		DIGEST,
+		SQL_TEXT
+	FROM performance_schema.events_statements_history
+	WHERE TIMER_END > DATE_SUB(NOW(), INTERVAL 1 DAY)
+	AND SQL_TEXT IS NOT NULL
+	AND DIGEST IS NOT NULL
+	AND CURRENT_SCHEMA NOT IN ('mysql', 'performance_schema', 'sys', 'information_schema');`
```
cristiangreco: nit: I think we can drop the trailing `;`?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,735 @@
...
+		}
+
+		redactedByteExplainPlanJSON, _, err := redactAttachedConditions(byteExplainPlanJSON)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to redact explain plan json", "err", err)
+			continue
+		}
+
+		level.Debug(logger).Log("msg", "db native explain plan",
+			"op", OP_EXPLAIN_PLAN_OUTPUT,
```
cristiangreco: I think we should use a separate `op` here, to avoid issues on the frontend when fetching log lines. (I know we talked this is for on-demand debug purpose, but technically it still need to be a different op)

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,735 @@
...
+
+		if !utf8.Valid(byteExplainPlanJSON) {
+			level.Error(logger).Log("msg", "explain plan json bytes is not valid UTF-8")
+			continue
+		}
+
+		loggedSchemaName := "<nil>"
+		if qi.schemaName != nil {
+			loggedSchemaName = *qi.schemaName
+		}
```
cristiangreco: When does this happen? I think the frontend doesn't handle this, it always queries for a specific schema. Wondering if we should just drop the plan here instead?

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Happy to change the op as well, but.. This also get's logged as alloy log output, so it would have a different job as well, assuming it was even being scraped and shipped to loki. 🤔

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: I believe this happens for select queries that are exclusively function calls, like `SELECT CURDATE()` etc. There's no schema since it wasn't executed against any table(s).

You're suggesting that if the schema is unavailable, maybe we should log a debug/warning and move on?

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Did do.. It's now `OP_NATIVE_EXPLAIN_PLAN`.

internal/component/database_observability/mysql/collector/explain_plan.go
matthewnolf: Yeah I think that makes sense, as digest is only specific within a schema and we currently scope everything else to a schema now

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -0,0 +1,735 @@
...
+
+		redactedByteExplainPlanJSON, _, err := redactAttachedConditions(byteExplainPlanJSON)
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to redact explain plan json", "err", err)
+			continue
+		}
+
+		level.Debug(logger).Log("msg", "db native explain plan",
+			"op", OP_NATIVE_EXPLAIN_PLAN,
```
matthewnolf: Can we make the op just a generic debug statement - rather than specific to the debug entry

internal/component/database_observability/mysql/collector/explain_plan.go
matthewnolf: So perhaps the final `planNode` here should represent something unknown then in this case?
Like
`planNode{Operation: "unknown"}`

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Yeah, actually.. I'm not entirely certain why there even is an `op` here. I think it's because when I first created this, I copy/pasted it from the *actual* log output which is sent to the loki collector.

I'm not sure this needs an op at all, I could simply delete the constant, and this param from the log? 🤔

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Ahh yes, I like this. I've added it [here](https://github.com/grafana/alloy/pull/3801/commits/74be1ac7f4bab9417aeff72cf52eae4c9a47a414).

internal/component/database_observability/mysql/collector/explain_plan.go
cristiangreco: > This also get's logged as alloy log output

Ah right, sorry I didn't realise. Just to avoid any possible confusion I think is still good to use a different op (as you've done now).

internal/component/database_observability/mysql/collector/explain_plan.go
cristiangreco: > You're suggesting that if the schema is unavailable, maybe we should log a debug/warning and move on?

Yes, basically the plugin won't make use of that log line if there's no match with a schema. But you're right, rather than dropping the information let's just use an Alloy debug log.

internal/component/database_observability/mysql/collector/explain_plan.go
cristiangreco: > I'm not sure this needs an op at all, I could simply delete the constant, and this param from the log? 🤔

In theory yes, given it's an Alloy log we don't really need the op at all.

rgeyer: Yeah, I already dropped it. I also added the digest id so we can actually correlate it with errors or questions when debugging. :)

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Realistically, these sorts of selects don't have a useful explain plan either.

I went ahead and just dropped them. We can re-add later or put in a more detailed log if we think we need it later.

matthewnolf (approved):

PR #3823, opened 2025-06-11
Fix bad link anchors that were missed in the previous refactoring PRs

clayton-cornell:
Fix bad link anchors that were missed in the previous refactoring PRs


clayton-cornell:
Remove backports to v1.7 and v1.7. The workflows are not working for backports to these releases. It's enough to backport to v1.9 and v1.8

PR #3828, opened 2025-06-12
database_observability: add locks collector

fridgepoet:
#### PR Description

This additional collector to the database_observability component (for MySQL) allows the user to identify which query is locking another query. This is a very common cause of performance degradation, as it is usual that concurrent threads run transactions against the same table/data, but only one can be executed at once, causing a cascade of locked threads that go all the way to the application level.

This particular collector focuses on data locks, which can be very granular, at the table level or even row and column - i.e. the command `UPDATE x.Name where Name = "Joe"` is gonna lock only the columns and rows being filtered.

Information is gathered from:
* `data_locks`: contains the current data locks being held and requested;
* `data_locks_waits`: implements a many-to-many relationship showing which data lock requests are being blocked by another lock being held;

Contributes to https://github.com/grafana/grafana-dbo11y-app/issues/795 

#### Notes to the Reviewer

Co-authored by @gaantunes 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -331,7 +331,6 @@ v1.8.0
   - `query_tables`: rename collector from `query_sample` to better reflect responsibility (@matthewnolf)
   - `query_sample`: add new collector that replaces previous implementation to collect more detailed sample information (@matthewnolf)
   - `query_sample`: refactor parsing of truncated queries (@cristiangreco)
-  - `locks`: addition of data locks collector (@gaantunes @fridgepoet)
 
 - Add labels validation in `pyroscope.write` to prevent duplicate labels and invalid label names/values. (@marcsanmi)
 


internal/component/database_observability/mysql/collector/locks.go
```diff
@@ -0,0 +1,188 @@
...
+	CollectInterval   time.Duration
+	LockWaitThreshold time.Duration
+	EntryHandler      loki.EntryHandler
+
+	running *atomic.Bool
+	cancel  context.CancelFunc
+	Logger  log.Logger
+}
+
+type Lock struct {
```
matthewnolf:
```suggestion
type LockCollector struct {
```

I don't think it goes with convention, but this seems more clear (and we can update previous collectors afterwards)

internal/component/database_observability/mysql/collector/locks.go
```diff
@@ -0,0 +1,188 @@
...
+			break
+		}
+
+		dataLocks = append(dataLocks, dataLock{
+			WaitingTimerWait: waitingTimerWait, WaitingLockTime: waitingLockTime, WaitingDigest: waitingDigest, WaitingDigestText: waitingDigestText,
+			BlockingTimerWait: blockingTimerWait, BlockingDigest: blockingDigest, BlockingDigestText: blockingDigestText,
+		})
+	}
+
+	for _, lock := range dataLocks {
```
matthewnolf: Why not log within the scan loop? Instead of adding all the locks to a slice and then iterating over them to log

internal/component/database_observability/mysql/collector/locks_test.go
matthewnolf, line 1: I think we're just missing a couple test cases for error handling

fridgepoet: Done!

internal/component/database_observability/mysql/component.go
```diff
@@ -63,6 +65,8 @@ type Arguments struct {
 var DefaultArguments = Arguments{
 	CollectInterval:               1 * time.Minute,
 	SetupConsumersCollectInterval: 1 * time.Hour,
+	LocksCollectInterval:          1 * time.Second,
+	LocksThreshold:                500 * time.Millisecond,
```
fridgepoet: highlighting these default values:
* Is 1 second too frequent? It seems like this would be the problematic lock duration threshold....
* locksThreshold is still set to 500ms to be a little bit more "aware" if there is a problem.

However both of these values may be too low. What do folks think?

matthewnolf: Given these are just defaults that we expect to will be overriden by configuration, I think we should lean more on them being safe/sensible than aggressively capturing.

I'd change `LocksCollectInterval` to >30s and `LocksThreshold` to >1s.

We can always lower this later, if it's not causing any issues but not collecting as much as we want.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -25,24 +25,27 @@ database_observability.mysql "<LABEL>" {
...
+| `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |           | no       |
+| `disable_query_redaction`          | `bool`               | Collect unredacted SQL query text including parameters.                                        | `false`   | no       |
+| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |           | no       |
+| `setup_consumers_collect_interval` | `duration`           | How frequently to collect performance_schema.setup_consumers information from database.        | `"1h"`    | no       |
+| `locks_collect_interval`           | `duration`           | How frequently to collect locks information from database.                                     | `"1s"`    | no       |
+| `locks_threshold`                  | `duration`           | Threshold for locks to be considered slow. If a lock exceeds this duration, it will be logged. | `"500ms"` | no       |
```
```suggestion
| `locks_collect_interval`           | `duration`           | How frequently to collect locks information from database.                                     | `"1s"`    | no       |
| `locks_threshold`                  | `duration`           | Threshold for locks to be considered slow. If a lock exceeds this duration, it will be logged. | `"500ms"` | no       |
| `setup_consumers_collect_interval` | `duration`           | How frequently to collect performance_schema.setup_consumers information from the database.    | `"1h"`    | no       |
```

Alpha order for rows.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -25,24 +25,27 @@ database_observability.mysql "<LABEL>" {
...
-| `schema_table`    | Collect schemas and tables from `information_schema`. | yes                |
-| `query_sample`    | Collect query samples.                                | no                 |
-| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`. | yes                |
+| Name              | Description                                              | Enabled by default |
+|-------------------|----------------------------------------------------------|--------------------|
+| `query_tables`    | Collect query table information.                         | yes                |
+| `schema_table`    | Collect schemas and tables from `information_schema`.    | yes                |
+| `query_sample`    | Collect query samples.                                   | no                 |
+| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.    | yes                |
+| `locks`           | Collect queries that are waiting/blocking other queries. | no                 |
```
clayton-cornell, line 48:
```suggestion
| Name              | Description                                              | Enabled by default |
| ----------------- | -------------------------------------------------------- | ------------------ |
| `locks`           | Collect queries that are waiting/blocking other queries. | no                 |
| `query_sample`    | Collect query samples.                                   | no                 |
| `query_tables`    | Collect query table information.                         | yes                |
| `schema_table`    | Collect schemas and tables from `information_schema`.    | yes                |
| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.    | yes                |
```

Alpha order for rows. Minor table formatting changes.

fridgepoet: Thank you Clayton!

internal/component/database_observability/mysql/component.go
fridgepoet: Done, I set it on the lower limit still. I put LocksCollectInterval at 30s and LocksThreshold at 1s. Let me know what you think.

clayton-cornell:
Docs look ok now :-)

CHANGELOG.md
```diff
@@ -319,6 +319,7 @@ v1.8.0
   - `query_tables`: rename collector from `query_sample` to better reflect responsibility (@matthewnolf)
   - `query_sample`: add new collector that replaces previous implementation to collect more detailed sample information (@matthewnolf)
   - `query_sample`: refactor parsing of truncated queries (@cristiangreco)
+  - `locks`: addition of data locks collector (@gaantunes @fridgepoet)
```
cristiangreco: nit: this should go in the `unreleased` section I believe

matthewnolf (approved):

CHANGELOG.md
fridgepoet: oops thank you for updating!

PR #3830, opened 2025-06-12
fix loki.source.file docs re legacy positions file

thampiotr:
#### PR Description

The legacy positions file is not being deleted, contrary to what the docs say. 

I think it's fine - it does not need to be deleted, because if the new positions file exists, we won't be attempting the conversion. I think it's safer to leave it where it is, in case someone wants to attempt the migration again or had misconfigured the location where the new positions file lives. 

So I'm fixing the docs here. 

Also adding some logging to make things more clear.


thampiotr:
Can ignore integration test failure as it's addressed here: https://github.com/grafana/alloy/pull/3832

tpaschalis (approved):
LGTM

PR #3835, opened 2025-06-13
fix(dockerfile): build UI as separate docker target

korniltsev:
#### PR Description
This PR builds UI as a separate docker `ui-build` target and then copies over to the `build` target.
This change fixes thee following issues:
1. Every time a go file changes, the UI is rebuilt, including fetching modules(due to invalid cache path)
2. Every time something in .git changes, the the UI is rebuilt, including fetching modules(due to invalid cache path)
3. The UI is built for each platform if building multiplatform image.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,14 @@
 # default when running `docker buildx build` or when DOCKER_BUILDKIT=1 is set
 # in environment variables.
 
+FROM --platform=$BUILDPLATFORM grafana/alloy-build-image:v0.1.19 AS ui-build
+ARG BUILDPLATFORM
+COPY ./internal/web/ui /ui
+WORKDIR /ui
+RUN --mount=type=cache,target=/ui/node_modules,sharing=locked \
+    yarn --network-timeout=1200000                            \
+    && yarn run build
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,6 +4,14 @@
 # default when running `docker buildx build` or when DOCKER_BUILDKIT=1 is set
 # in environment variables.
 
+FROM --platform=$BUILDPLATFORM grafana/alloy-build-image:v0.1.19 AS ui-build
```
simonswine, line 7: Nit: Maybe the build image tag could be a 
```
ARG BUILD_IMAGE_TAG="v0.1.19"
```

simonswine (approved):
LGTM

Works for me as designed. Maybe also adding a `.git` to [`.dockerignore`](https://github.com/grafana/alloy/blob/main/.dockerignore)

korniltsev:
> Works for me as designed. Maybe also adding a .git to [.dockerignore](https://github.com/grafana/alloy/blob/main/.dockerignore)

I suspect we can't do this because Makefile or some of the invoked scripts are relying on git being present. This is also the reason why I do not use make in ui-build stage in this PR

PR #3841, opened 2025-06-16
chore: bump otel to 127

korniltsev:
#### PR Description
Bump otel dependencies from 126 to 127
#### Which issue(s) this PR fixes
The new otel ebpf profiler PR https://github.com/grafana/alloy/pull/2920 requires upgrading to 128. I intend to submit another bump in a followup to keep the changes smaller and atomic

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #2920, opened 2025-03-06
pyroscope.ebpf: otel profiler implementation

korniltsev:
#### PR Description
This PR changes the implementation of `pyroscope.ebpf` compoenent from [grafana/pyroscope/ebpf](https://github.com/grafana/pyroscope/tree/main/ebpf) to [opentelemetry-ebpf-profiler](https://github.com/grafana/opentelemetry-ebpf-profiler?tab=readme-ov-file.) - a fork of opentelemetry-ebpf-profiler.
The notable changes of the fork: native frame symbolization, dynamic profiling policy, target discovery labels.

The fork uses a rust library for symbol extraction. See https://github.com/open-telemetry/opentelemetry-ebpf-profiler/tree/main/rust-crates.

In this PR I put all `pyroscope.ebpf` code behind a `pyroscope_ebpf` go tag.

Some component arguments became no-op as there are no corresponding knobs in the new implementations. These args a deprecated
```
// deprecated
	PidCacheSize int `alloy:"pid_cache_size,attr,optional"`
	// deprecated
	BuildIDCacheSize int `alloy:"build_id_cache_size,attr,optional"`
	// deprecated
	SameFileCacheSize int `alloy:"same_file_cache_size,attr,optional"`
	// deprecated
	CacheRounds int `alloy:"cache_rounds,attr,optional"`
	// deprecated
	GoTableFallback bool `alloy:"go_table_fallback,attr,optional"`
	// deprecated
	SymbolsMapSize int `alloy:"symbols_map_size,attr,optional"`
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [ ] Config converters updated

PR #3843, opened 2025-06-16
chore: bump otel to 128

korniltsev:
#### PR Description
Continuation of https://github.com/grafana/alloy/pull/3841

Bump otel dependencies from 127 to 128
The new otel ebpf profiler PR https://github.com/grafana/alloy/pull/2920 requires upgrading to 128. 

This PR is only upgrading the dependency and fixes compilation to unblock the above profiling PR. The proper upgrade should be handled here https://github.com/grafana/alloy/issues/3844
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


clayton-cornell:
Is this a version change that should also be reflected in the main `_index.md` and `_index.md.t` files in the documentation?

korniltsev:
> Is this a version change that should also be reflected in the main _index.md and _index.md.t files in the documentation?

Updated the markdowns. 

This PR is only upgrading the dependency and fixes compilation to unblock the profiling PR. The proper upgrade should be handled here https://github.com/grafana/alloy/issues/3844 . I don't have capacity to do the proper upgrade.

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ Main (unreleased)
 
 - Upgrade `otelcol` components from OpenTelemetry v0.126.0 to v0.127.0 (@korniltsev)
 
+- Upgrade `otelcol` components from OpenTelemetry v0.127.0 to v0.128.0 (@korniltsev)
```
thampiotr: Since 127 was never released in Alloy, you can collapse it all to one changelog entry.

internal/component/otelcol/auth/oauth2/oauth2.go
```diff
@@ -62,7 +62,7 @@ func (args Arguments) ConvertClient() (otelcomponent.Config, error) {
 		TokenURL:         args.TokenURL,
 		EndpointParams:   args.EndpointParams,
 		Scopes:           args.Scopes,
-		TLSSetting:       *args.TLSSetting.Convert(),
+		TLS:              *args.TLSSetting.Convert(),
```
thampiotr: Can you rename `args.TLSSetting` also to `args.TLS`?

internal/static/traces/config_test.go
```diff
@@ -366,36 +400,6 @@ service:
   extensions: ["jaegerremotesampling/0", "jaegerremotesampling/1"]
 `,
 		},
-		{
-			name: "push_config and remote_write",
-			cfg: `
```
thampiotr, line 371: Why removing this test case?

korniltsev: it was fragile and incorrect and was testing for wrong thing. 

It was supposed to test for an error when push_config and remote_write are configured together. 

But in fact (before upgrade to v128) it was receiving a completely different error "empty config for Jaeger receiver".

And the test did check for the presence of error, not the value/type of the error and the test was incorrectly green. 

Before v128 the nil map was unmarshaled an empty map and failed validation leading to an error and now after v128  it is unmarshaled as default configuration with no error.

Anyway the tests should be removed because  `push_config` was removed long ago

internal/component/otelcol/auth/oauth2/oauth2.go
korniltsev: renamed

CHANGELOG.md
korniltsev: Collapsed

PR #3855, opened 2025-06-16
ci: port main version of .github dir to release branch

jharvey10:
#### PR Description

Synchronizes the github workflows with the `main` branch.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,9 +1,6 @@
 name: Backport PR Creator
 on:
-  pull_request_target: # zizmor: ignore[dangerous-triggers] In order to backport PRs from external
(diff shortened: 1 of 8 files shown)


jharvey10:
It makes sense that this PR fails because of the integration tests, as this was a problem on `main` that required both a code change, a workflow change, and a dependency update. These workflow checks aren't all idempotent, so this is bound to happen over time.

I suppose the more tactical approach for these older release branches is to just fix the zizmor issues and move on ignoring the other failing checks, though it's a little unfortunate that we're subjecting docs changes (which are self-isolated) to a much broader process that is prone to more failure over time than the docs process itself.

Wondering if a potential solution here is to add more hierarchy to the github workflows that get run on a PR so we can skip any checks that are irrelevant to the changes in question.

PR #3857, opened 2025-06-17
fix(pyroscope.ebpf): send profiles concurrently

korniltsev:
#### PR Description
We noticed an issue when pyrsocope added an artificial delay . If response takes 500ms and we have 128 pprofs, it takes 128*500ms which is suboptimal.
This pr sends ebpf profiles asynchronously in a pool of gorutines with a cap of 32 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/ebpf/pool.go
+++ b/internal/component/pyroscope/ebpf/pool.go
@@ -1,0 +1,29 @@
+package ebpf
+
+import "sync"
+
+// copypaste from https://github.com/grafana/pyroscope/blob/2c7e5971dc682cb442da4a5e234d01a2366a42c5/pkg/experiment/ingester/segment.go#L698
+// with a slight modification - removed `do` func
+type workerPool struct {
+	workers sync.WaitGroup
+	// jobs must not be used after stop
+	jobs chan func()
+}
+
+func (p *workerPool) run(workers int) {
+	p.jobs = make(chan func())
+	p.workers.Add(workers)
+	for i := 0; i < workers; i++ {
+		go func() {
+			defer p.workers.Done()
+			for job := range p.jobs {
+				job()
+			}
+		}()
+	}
+}
+
+func (p *workerPool) stop() {
+	close(p.jobs)
+	p.workers.Wait()
+}
--- a/internal/component/pyroscope/ebpf/send.go
+++ b/internal/component/pyroscope/ebpf/send.go
@@ -1,0 +1,64 @@
+//go:build (linux && arm64) || (linux && amd64)
+
+package ebpf
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-kit/log/level"
+	"github.com/grafana/alloy/internal/component/pyroscope"
+	"github.com/grafana/pyroscope/ebpf/pprof"
+)
+
+const maxSendConcurrency = 32
+
+func (c *Component) sendProfiles(ctx context.Context, builders *pprof.ProfileBuilders) {
+	start := time.Now()
+	pool := workerPool{}
+	pool.run(min(maxSendConcurrency, len(builders.Builders)))
+	queued := 0
+	ctx, cancel := context.WithTimeout(ctx, c.args.CollectInterval)
+	defer func() {
+		pool.stop()
+		cancel()
+		level.Debug(c.options.Logger).Log("msg", "sent profiles", "duration", time.Since(start), "queued", queued)
+	}()
+	j := 0
+	for _, builder := range builders.Builders {
+		serviceName := builder.Labels.Get("service_name")
+		c.metrics.pprofsTotal.WithLabelValues(serviceName).Inc()
+		c.metrics.pprofSamplesTotal.WithLabelValues(serviceName).Add(float64(len(builder.Profile.Sample)))
+
+		buf := bytes.NewBuffer(nil)
+		_, err := builder.Write(buf)
+		if err != nil {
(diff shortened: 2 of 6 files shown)


internal/component/pyroscope/ebpf/send.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const maxSendConcurrency = 32
+
+func (c *Component) sendProfiles(ctx context.Context, builders *pprof.ProfileBuilders) {
+	pool := workerPool{}
+	pool.run(min(maxSendConcurrency, len(builders.Builders)), len(builders.Builders))
+	defer func() {
+		pool.stop()
+	}()
```
kolesnikovae: I think this is where `runtime.NumCPU` really shines: the more CPUs a _node_ has, the more likely it is that we'll need a larger queue. But `min(maxSendConcurrency, len(builders.Builders))` also looks good.

A hard cap of 32 seems reasonable to me but I'm wondering if having a configurable parameter here would be safer

internal/component/pyroscope/ebpf/pool.go
```diff
@@ -0,0 +1,36 @@
...
+	p.jobs = make(chan func(), jobs)
+	p.workers.Add(workers)
+	for i := 0; i < workers; i++ {
+		go func() {
+			defer p.workers.Done()
+			for job := range p.jobs {
+				job()
+			}
+		}()
+	}
```
kolesnikovae: Not sure I understand the `p.jobs != nil` bit

internal/component/pyroscope/ebpf/pool.go
```diff
@@ -0,0 +1,36 @@
+package ebpf
+
+import "sync"
+
+// copypaste from https://github.com/grafana/pyroscope/blob/2c7e5971dc682cb442da4a5e234d01a2366a42c5/pkg/experiment/ingester/segment.go#L698
+// with a slight change - added `jobs` parameter to fully buffer all jobs in the chan
+type workerPool struct {
+	workers sync.WaitGroup
+	jobs    chan func()
+}
```
kolesnikovae: Just a note: we may want to replace it with something more elaborate in the future

internal/component/pyroscope/ebpf/send.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, builder := range builders.Builders {
+		if ctx.Err() != nil {
+			return
+		}
+
+		serviceName := builder.Labels.Get("service_name")
+		c.metrics.pprofsTotal.WithLabelValues(serviceName).Inc()
```
kolesnikovae, line 32: For the future us: `service_name` may have arbitrary cardinality. Unless absolutely necessary, let's remove it. Otherwise, it will be very painful if we hit problems here.

internal/component/pyroscope/ebpf/send.go
```diff
@@ -0,0 +1,51 @@
...
+		appender := c.appendable.Appender()
+		c.metrics.pprofBytesTotal.WithLabelValues(serviceName).Add(float64(len(rawProfile)))
+
+		pool.do(func() {
+			samples := []*pyroscope.RawSample{{RawProfile: rawProfile}}
+			err = appender.Append(ctx, builder.Labels, samples)
+			if err != nil {
+				level.Error(c.options.Logger).Log("msg", "ebpf pprof write", "err", err)
+			}
+		})
```
kolesnikovae: `pool.do` is blocking and `appender.Append` may take a really long time in practice. Should we maybe do something like:
```go
// The timeout should be aligned with the collection interval:
// if we didn't manage to send everything, we won't block profiler (if I get it right)
ctx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()

for (something) {
   // IDK if we're allowed to call it concurrently
   c.appendable.Appender()
   job := func() {
      err = appender.Append(ctx, builder.Labels, samples)
      // Log
   }
   select {
   case pool.jobs <- job:
   case <-ctx.Done():
      // Log
      return
   }
}

```

internal/component/pyroscope/ebpf/send.go
korniltsev: I would make it configurable if I did not have to update documentations for this. I'd prefer to hardcode it for now

internal/component/pyroscope/ebpf/pool.go
korniltsev: same :) a blind copy-paste. removed.

internal/component/pyroscope/ebpf/send.go
korniltsev: I think there may be issue if I just do `defer cancel()` it will cancel the context for the last 32 profiles batch.

I think I better do 
```go
	defer func() {
		pool.stop() // wait for profiles to be sent and goroutines to finish. This should be quick if the timeout deadline already reached assuming the fanoutClient respects the ctx
		cancel() // release the timer if we sent profiles faster then timeout
	}()
```

internal/component/pyroscope/ebpf/send.go
korniltsev: Let's do this separately https://github.com/grafana/alloy/issues/3864

internal/component/pyroscope/ebpf/send.go
kolesnikovae: It's up to you

internal/component/pyroscope/ebpf/send.go
kolesnikovae: Yeah, I didn't test that, just wanted to avoid blocking on `pool.do()`

aleks-p (approved):

PR #3863, opened 2025-06-17
Add info about Windows permissions

### LINKED ISSUE #3619: Document Windows permissions required to run Alloy on Windows
Documentation is required to detail what permissions and settings are required for the Windows account that runs Alloy.

We don't have established documentation for this, in part as it depends on their configuration and what they are seeking to monitor. The following should be a reasonable list of permissions, but it's a mix of groups and general advice depending on the requirement.
As this isn't documented and thoroughly tested there may be iteration required.

- Alloy needs "Event Log Readers" group to use the windows event log components.
- Alloy needs "Performance Monitor Users" to use many of the features of the windows exporter.
- Alloy may need "Performance Log Users" for some features of the windows exporter
- If they want to use the process & services collectors of the windows exporter they'll need access to list all processes/services, I don't know off hand how to configure that without Local System.
- Alloy needs appropriate permissions to manage files in the Alloy Storage directory (default %PROGRAMDATA%\GrafanaLabs\Alloy\data).
- Depending on the components configured, Alloy may also require permissions to request/manage temporary directory files.
- Alloy needs appropriate permissions to read any on-disk log files for applications being monitored. This will require whatever groups define the ACL permissions to read the logs/directories.
- Alloy needs appropriate permissions & proxy configuration to access any network endpoints (prometheus scrape, grafana cloud write destinations, etc) used to collector & deliver telemetry
- Alloy needs access to the relevant registry values used to configure Alloy.
- Alloy needs permission to TCP listen on the configured UI port (default 12345)
- Alloy user requires the permission to 'run as a service'

clayton-cornell:
#### PR Description

Add info about Windows permissions and security requirements for running Alloy

Fixes #3619 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/set-up/install/windows.md
+++ b/docs/sources/set-up/install/windows.md
@@ -12,6 +12,45 @@ weight: 500
 
 You can install {{< param "PRODUCT_NAME" >}} on Windows as a standard graphical install, or as a silent install.
 
+## Configure the Windows permissions
+
+To effectively monitor Windows telemetry with {{< param "PRODUCT_NAME" >}}, the user account you use to run {{< param "PRODUCT_NAME" >}} requires specific access permissions.
+These permissions ensure {{< param "PRODUCT_NAME" >}} can collect the necessary data, manage its local storage, and communicate with other services.
+The exact permissions depend on your system's security configuration and the specific telemetry you need to collect.
+
+### Windows security groups
+
+To collect common Windows telemetry, for example event logs and performance counters, the user account you use to run {{< param "PRODUCT_NAME" >}} should be a member of the following Windows security groups.
+These groups provide the necessary read access without granting excessive privileges.
+
+* `Event Log Readers`: This group allows members to read data from local event logs, including Application, System, Security, and other custom logs.
+  This is essential for any Alloy configuration that collects Windows Event Log data.
+* `Performance Monitor Users`: This group allows non-administrator users to access performance counter data.
+  This group is important for {{< param "PRODUCT_NAME" >}} components that collect Windows performance metrics, for example CPU, memory, disk I/O, and network usage.
+* `Performance Log Users`: This group is used to schedule logging of performance counter data and manage performance alerts.
+   Performance Log Users is necessary for advanced or historical data collection scenarios, particularly those that involve the Windows Data Collector Sets.
+
+### File system and network permissions
+
+Beyond the standard Windows groups, {{< param "PRODUCT_NAME" >}} requires some specific permissions for its operational functions:
+
+* Storage directory permissions: {{< param "PRODUCT_NAME" >}} needs read, write, and modify permissions to manage files and directories within its data storage location.
+  Default Location: %PROGRAMDATA%\GrafanaLabs\Alloy\data
+* Application log file read permissions: If you configure {{< param "PRODUCT_NAME" >}} to read application log files directly from disk, the user account you use to run {{< param "PRODUCT_NAME" >}} must have read access to those log files and their containing directories.
+  You may need to modify the Access Control Lists for these resources or add the {{< param "PRODUCT_NAME" >}} service account to a custom group that has these permissions.
+* Network access for telemetry destinations: {{< param "PRODUCT_NAME" >}} needs network connectivity and, if applicable, proxy configuration, to communicate with its configured telemetry endpoints.
+  This includes:
+  * Source endpoints: For scraping metrics from Prometheus exporters and pulling logs from remote APIs.
+  * Destination Endpoints: For writing metrics to Prometheus or Grafana Cloud, and sending logs to Loki.
+  Make sure your firewall rules allow outbound connections from the {{< param "PRODUCT_NAME" >}} host to these destinations on the necessary ports.
+* Registry access: The user account you use to run {{< param "PRODUCT_NAME" >}} may need access to the Windows Registry to configure things like [environment variables](https://grafana.com/docs/alloy/latest/configure/windows/#change-environment-variable-values).
+* UI port listening permission: If you want to enable the {{< param "PRODUCT_NAME" >}} UI, the user account you use to run {{< param "PRODUCT_NAME" >}} must have permission to listen on the configured UI port.
+  The default port is `12345`.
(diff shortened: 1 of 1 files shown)


jharvey10:
I'm definitely not an expert by any means on Alloy on Windows, but the docs contents read well and make sense to me. The only thing that might help me close the loop might be a link out to a place in the Windows ecosystem that talks about *how* you assign permissions to users or add members to groups, etc. Is this all done via Active Directory?

clayton-cornell:
I added some links back to the MS docs.

docs/sources/configure/windows.md
```diff
@@ -115,5 +115,53 @@ To expose the UI to other machines, complete the following steps:
 
      To listen on all interfaces, replace _`<LISTEN_ADDR>`_ with `0.0.0.0`.
 
+## Configure Windows permissions
```
thampiotr, line 118: Some things that I think need clarifying: Do I need to perform all these steps? How do I perform them? If I install following the [set-up guide](https://github.com/grafana/alloy/blob/a75614246dc234d60d7c8122827b16b231f52715/docs/sources/set-up/install/windows.md) do I still need to give these permissions explicitly for things to work (I don't think so?).

I don't know for sure, but I'm guessing that our set-up procedures already set these permissions automatically, but we want them documented for more security-conscious users who want to have tighter control over what is given? Or so that they can do a threat level assessment?

jeremy-bouchet:
Hello Grafana Labs,

We uses your Grafana Cloud solution for the observability of our systems and applications.

And we deployed recently Alloy agent (v1.9.0) on our MSSQL servers (Windows Server 2029), we need to configure a AD service account on the "Alloy" Windows service to be capable to connect to the MSSQL instance and to get mssql metrics.

We did some tests with a lot of several connection strings in the Alloy mssql config file but since I configured my AD service account to run the "Alloy" Windows service, this one don't start.

I want to stay my AD service account on my MSSQL server in the "Users" local group, I don't want to define this AD service account in the "Administrators" group.

**Please may you define the documentation to explain to everyone how to configure a Windows account / AD Windows service account with the most limited permissions on the Windows server ?**

**And also please may you do your tests on your side to validate that all your documentation points works fine in the practice.**

At D-EDGE we really need of these ones, and our RSSI want to decrease at maximum the permissions of the windows account which run the "Alloy" agent.

And for the moment we are really blocked about that.

cc @annelaurefroment @ahadjidj 

regards,

docs/sources/configure/windows.md
ptodev: I agree that specifying the exacts steps users should take would be nice... I just don't know if it's very realistic? I don't know how complicated these things can get. But if we can at least point users in the right direction or including links to MS docs with information how to configure aspects of Windows then that'd be nice.

ptodev:
Hi, @jbouchet-dedge 👋  I'm honestly not sure if the first iteration of this doc could include a lot of detailed steps. For example, I don't know if there are multiple ways to configure security settings - maybe not everyone uses AD? Also, it'd be hard to find the minimum requirements for each Alloy component. The page looks ok as a starting point, but I hope our solutions engineers could extend it over time with more concrete steps that they would normally do when setting up Alloy for customers.

Regarding your specific issue - maybe it's worth running [Process Monitor](https://learn.microsoft.com/en-us/sysinternals/downloads/procmon) to see if Alloy failed to retrieve something like a registry key or a file?

docs/sources/configure/windows.md
thampiotr: I don't think you need to do these steps if you used an installer, right? 
I think that's the thing I want to clarify more so than 'how' to do it.

ptodev: I haven't checked, but I'd be very surprised if the installer sets up all the things listed below. I doubt that it updates firewall and proxy settings for example.

dehaansa (approved):
This looks good to me as an improvement over the existing docs. I think it would be nice to have more of this configurable via the installer etc as noted by Piotr but I think this improves the status quo 👍

docs/sources/configure/windows.md
```diff
@@ -115,5 +115,53 @@ To expose the UI to other machines, complete the following steps:
...
+
+To effectively monitor Windows telemetry with {{< param "PRODUCT_NAME" >}}, the user account you use to run {{< param "PRODUCT_NAME" >}} requires specific access permissions.
+These permissions ensure {{< param "PRODUCT_NAME" >}} can collect the necessary data, manage its local storage, and communicate with other services.
+The exact permissions depend on your system's security configuration and the specific telemetry you need to collect.
+
+### Windows security groups
+
+To collect common Windows telemetry, for example event logs and performance counters, the user account you use to run {{< param "PRODUCT_NAME" >}} should be a member of the following [Windows security groups](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-groups).
```
martincostello: Links to `learn.microsoft.com` should be de-localized so that _if_ content exists for another language that the user has an `Accept-Language` preference for, then that's the content they get served.

```suggestion
To collect common Windows telemetry, for example event logs and performance counters, the user account you use to run {{< param "PRODUCT_NAME" >}} should be a member of the following [Windows security groups](https://learn.microsoft.com/windows-server/identity/ad-ds/manage/understand-security-groups).
```

docs/sources/configure/windows.md
```diff
@@ -115,5 +115,53 @@ To expose the UI to other machines, complete the following steps:
...
+These groups provide the necessary read access without granting excessive privileges.
```
martincostello: Maybe describe as "minimal required" rather than "excessive"?

docs/sources/configure/windows.md
clayton-cornell: Fixed in b129b642e582321fa4752fce1e1081483b99e310

docs/sources/configure/windows.md
clayton-cornell: Fixed in [`b129b64` (#3863)](https://github.com/grafana/alloy/pull/3863/commits/b129b642e582321fa4752fce1e1081483b99e310)

clayton-cornell:
Merging this PR. We can look at expanding the content (within scope) in a future iteration.

PR #3869, opened 2025-06-18
Greater control over RBAC objects and rules

### LINKED ISSUE #196: Support configuring agent clusterRole permission in values.yaml

### LINKED ISSUE #681: Feature: More granular control over RBAC creation
### Request

When deploying via the Helm chart and using `rbac.create = true`, then ClusterRole and ClusterRoleBinding objects are created. Part of those contains this rule:
```yaml
  # needed for remote.kubernetes.*
  - apiGroups: [""]
    resources:
      - "configmaps"
      - "secrets"
    verbs:
      - get
      - list
      - watch
```

This is a very permissive: "Alloy's in yer cluster, readin' yer secrets!"

My feature request: Allow finer grained control over RBAC rules, along the lines of either:

```yaml
rbac:
  create: true
  rules:
...
    - apiGroups: [""]
      resources:
        - "configmaps"
        - "secrets"
      verbs:
        - get
        - list
        - watch
...
```
or provide an option to namespace the rules:
```yaml
rbac:
  create: true
  secrets:
    namespace: []   # Empty namespace means all (e.g. its in the ClusterRole), otherwise, create a Role and RoleBinding for each namespace
```

### Use case

Minimizing the security footprint of Alloy. Allow for reduction or removal of secret access if it's not required.

### LINKED ISSUE #2224: Alloy chart should not give itself cluster-wide secret read permissions by default
### What's wrong?

This is a stronger variant of [this issue](https://github.com/grafana/alloy/issues/681), defining the problem as a bug rather than a feature request.

My assertion is that a helm chart should not blithely give itself permission to read all secrets across a cluster. As far as I can tell, this access level exists in order to make the [remote.kubernetes.secret component](https://grafana.com/docs/alloy/latest/reference/components/remote/remote.kubernetes.secret/) "just work," but that is a component that in my view *shouldn't* "just work" without an explicit opt-in to exposing secrets. The documentation for this component should be updated to reflect that RBAC permissions must be given to the alloy service account to read any secrets that are to be exposed with this component. The [feature enhancement](https://github.com/grafana/alloy/issues/681) requested above, if provided, should meet the opt-in need rather than enable opting out.

Unfortunately, the feature provided by https://github.com/grafana/k8s-monitoring-helm/pull/224 in the k8s-monitoring chart relies on these extremely elevated default permissions, so fixing this will require breaking that chart. I will file an issue there and cross-reference it here.

### Steps to reproduce

Install the chart configured with a remote.kubernetes.secret component for any secret in the cluster. Alloy will be able to read the secret without any explicit opt-in by the secret owner.

### System information

_No response_

### Software version

_No response_

### Configuration

```text

```

### Logs

```text

```

### LINKED ISSUE #3839: [Helm] Customize rbac creation in helm
### Request

https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/rbac.yaml
Currently, if `rbac.create: true` is used in helm, a bunch of rules are applied. 
Can we add another setting to customize the rules here?

### Use case

Same reason of use case in https://github.com/grafana/alloy/issues/3838
Due to security considerations, we don't want created role has so many permissions while pod list and watch would be enough.

petewall:
#### PR Description

This change allows for the RBAC object creation to create namespaced-Role and RoleBinding objects instead of ClusterRoles and ClusterRoleBindings.

It also allows for the specific rules attached to the Roles/ClusterRoles to be modified.

Fixes #196
Fixes #3839
Fixes #2224
Fixes #681

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -10,8 +10,6 @@ internal API changes are not present.
 Unreleased
 ----------
 
-- Add support for configuring initialDelaySeconds and timeoutSeconds in Helm chart for readiness probe. (@peter-meltcafe)
-
 - Allow for creating Roles and RoleBindings instead of ClusterRoles and ClusterRoleBindings. (@petewall)
 
 - Allow for customizing the specific RBAC rules being created. (@petewall & @kun98-liu)


thampiotr (approved):
Looks good, we're onboard with adding this, thanks!

petewall:
One thing we'll have to know is the impact of losing `nonResourceURLs` from ClusterRoles.

petewall:
I did some more testing recently. Without cluster roles, we lose access to non-namespaced objects. This seems obvious in hindsight. What this means is we lose the ability to run something like:
```
  role = "node"
}
```

jacobcrawford:
It would be really nice if this could be released! 😄 @petewall any reason for delaying?

operations/helm/charts/alloy/templates/rbac.yaml
```diff
@@ -7,92 +40,8 @@ metadata:
...
-  # needed for remote.kubernetes.*
-  - apiGroups: [""]
-    resources:
-      - "configmaps"
-      - "secrets"
-    verbs:
-      - get
-      - list
-      - watch
-  # needed for otelcol.processor.k8sattributes
```
dehaansa, line 88: Did we lose these permissions?

dehaansa (commented):
Question about permissions for k8sattributes processor, otherwise looks good, requires a changelog merge conflict resolution too.

Gugu7264:
Hi! Is there any plan to merge&release this anytime soon?

operations/helm/charts/alloy/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ internal API changes are not present.
 Unreleased
 ----------
 
+- Add support for configuring initialDelaySeconds and timeoutSeconds in Helm chart for readiness probe. (@peter-meltcafe)
+
```
dehaansa:
```suggestion
```

Changelog merge resolution issue.

PR #3873, opened 2025-06-20
[Helm] Add alloy.enableHttpServerPort

### LINKED ISSUE #3838: [Helm] Disable containerPort in helm chart
### Request

https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/containers/_agent.yaml#L42-L59
current helm chart by default adds one port for readiness probe. 
can we provide another setting like `addContainerPort` and also exec command for readiness probe?

### Use case

My team is recently considering to adopt grafana-alloy to replace pyroscope java agent. We would like to run grafana-alloy as daemonset and with hostPID to scrape java processes' profiles.

However, since it requires both hostPID and root user, it is better to forbid grafana-alloy pods from exposing any endpoint through http. so we don't want to add any `containerPort` to the pod template.

kun98-liu:
#### PR Description
In some cases, it is preferred to not expose container port to other pods. This PR is to add a new option alloy.enableContainerPort to disable it.

Fixes #3838 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


operations/helm/charts/alloy/templates/containers/_agent.yaml
```diff
@@ -39,6 +39,7 @@
   envFrom:
     {{- toYaml $values.envFrom | nindent 4 }}
   {{- end }}
+  {{- if $values.enableContainerPort }}
```
dehaansa: This should only encompass the first container port, and still allow users to configure extraPorts when using this configuration.

kun98-liu: Thanks for taking look at this PR! 🙇 

> still allow users to configure extraPorts when using this configuration.

To be honest, i feel it weird to have `extraPorts` when user has explicitly set `enableContainerPort: false`. 🤔 May I ask why?

dehaansa: I would think there would be a case where the user does not want the UI/etc http server to be exposed, but would still want external clients to send in data to a prometheus.receive_http, otelcol.receiver.otlp, etc.

kun98-liu: ah, i see. then maybe `enableContainerPort` is actually not a good name. I think something like  `enableHttpServerPort` could be better. What do you think?

dehaansa: Sure, that sounds good. `exposeAlloyWebServer` or `exposeAlloyUIPort` or something are also good options if you're weighing options, I think any of them are acceptable.

kun98-liu: Updated to enableHttpServerPort 🙇

dehaansa (approved):
LGTM, thanks for the contribution!

PR #3899, opened 2025-06-25
[v1.9 fixes] Upgrade a dependency, add windows exporter no-op params

ptodev:
A dependency was updated to resolve a CVE (see #3875).

The windows exporter change is because users have been complaining about difficulty with upgrades. If the change is approved do a similar change to the main branch.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
@@ -71,6 +71,12 @@ You can use the following blocks with `prometheus.exporter.windows`:
 | [`tcp`][tcp]                               | Configures the `tcp` collector.                | no       |
 | [`text_file`][text_file]                   | Configures the `text_file` collector.          | no       |
 
+{{< admonition type="note" >}}
+In previous versions of Alloy there used to be a `msmq` block.
+It is still allowed in Alloy confguration files but its usage now is a no-op.
+Users are advised to remove it from their confguration files because it will be removed completely in upcoming Alloy releases.
+{{< /admonition >}}
+
 [dfsr]: #dfsr
 [dns]: #dns
 [exchange]: #exchange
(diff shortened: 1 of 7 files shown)


dehaansa (approved):
LGTM, thanks!

Edit: looks like this needs a merge before CI will be happy.

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
```diff
@@ -71,6 +71,12 @@ You can use the following blocks with `prometheus.exporter.windows`:
 | [`tcp`][tcp]                               | Configures the `tcp` collector.                | no       |
 | [`text_file`][text_file]                   | Configures the `text_file` collector.          | no       |
 
+{{< admonition type="note" >}}
+In previous versions of Alloy there used to be a `msmq` block.
+It is still allowed in Alloy confguration files but its usage now is a no-op.
+Users are advised to remove it from their confguration files because it will be removed completely in upcoming Alloy releases.
```
clayton-cornell:
```suggestion
Starting with release 1.9.0, the `msmq` block is deprecated.
It will be removed in a future release.

You can still include this block in your configuration files. However, its usage is now a no-op.
```

Aligning with the guidance here: https://grafana.com/docs/writers-toolkit/write/deprecate-remove/  If this works the same text changes should be applied to the second note as well.

We deleted the msmq block docs between v1.8 and v1.9 releases..but didn't mark it as deprecated?  Same with the attributes... deleted but we didn't mark them as deprecated?

ptodev: Thank you, I applied the same change to the other snippet. Yes, they were not marked as deprecated. They were just removed from the docs. I'd rather not add them to the tables so that we don't draw attention to no-op fields.

ptodev:
I tested this on a windows machine and I indeed see the warnings:
```
ts=2025-06-25T19:29:28.804019Z level=warn msg="the `msmq` block is deprecated - its usage is a no-op and it will be removed in the future" component_path=/ component_id=prometheus.exporter.windows.windows_exporter
ts=2025-06-25T19:29:28.804019Z level=warn msg="the `use_api` attribute inside the `service` block is deprecated - its usage is a no-op and it will be removed in the future" component_path=/ component_id=prometheus.exporter.windows.windows_exporter
ts=2025-06-25T19:29:28.804019Z level=warn msg="the `where_clause` attribute inside the `service` block is deprecated - its usage is a no-op and it will be removed in the future" component_path=/ component_id=prometheus.exporter.windows.windows_exporter
ts=2025-06-25T19:29:28.804019Z level=warn msg="the `enable_v2_collector` attribute inside the `service` block is deprecated - its usage is a no-op and it will be removed in the future" component_path=/ component_id=prometheus.exporter.windows.windows_exporter
```

This is my config:
```hcl
prometheus.exporter.windows "windows_exporter" {
    msmq {}
    service {
        where_clause = ""
        use_api = "false"
        enable_v2_collector = "false"
    }
}
```

ptodev:
PR for the main branch: #3900

clayton-cornell (approved):

PR #3901, opened 2025-06-25
dbo11y: Change the table used to fetch queries for explain plans

rgeyer:
#### PR Description
Changes the performance_schema table used to fetch queries, so that more (all?) digests should be able to be fetched.

I didn't change the Changelog, given that this is a minor improvement to an unreleased feature. I could add a bullet point if we think it's important.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/explain_plan.go
+++ b/internal/component/database_observability/mysql/collector/explain_plan.go
@@ -454,12 +454,12 @@ type queryInfo struct {
 }
 
 type ExplainPlanArguments struct {
-	DB             *sql.DB
-	InstanceKey    string
-	ScrapeInterval time.Duration
-	PerScrapeRatio float64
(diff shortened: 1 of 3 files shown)


internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -458,6 +459,7 @@ type ExplainPlanArguments struct {
 	ScrapeInterval time.Duration
 	PerScrapeRatio float64
 	EntryHandler   loki.EntryHandler
+	LastSeen       time.Time
```
cristiangreco: Nit: given it's an arg, I'd name it `InitialLookback` like the option itself. The `lastseen` in the struct at line 476 instead is fine as is.

PR #3916, opened 2025-06-29
feat(otelcol): add support for htpasswd file authentication

pkarakal:
#### PR Description
This adds support for using `htpasswd` for server authentication. Since `otelcol.auth.basic` is a wrapper around `basicauthextension`, which already supports both inline and file based server authentication, this adds such support for alloy as well. This would be really useful when we want to have multiple users trying to send data to Alloy and we don't want to a reverse proxy for authenticating calls.

This also makes the `username` and `password` fields optional since we should be able to just use `htpasswd` files for server side authentication. They must be filled in when the handler is used for client authentication. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -31,17 +31,21 @@ You can specify multiple `otelcol.auth.basic` components by giving them differen
 otelcol.auth.basic "<LABEL>" {
   username = "<USERNAME>"
   password = "<PASSWORD>"
+  
+  htpasswd_file = "/etc/alloy/.htpasswd"
```
ptodev:
```suggestion
```

The usage section only lists required attributes and blocks. You could add an example to the example section though.

internal/component/otelcol/auth/basic/basic.go
```diff
@@ -57,11 +83,17 @@ func (args Arguments) ConvertClient() (otelcomponent.Config, error) {
 
 // ConvertServer implements auth.Arguments.
 func (args Arguments) ConvertServer() (otelcomponent.Config, error) {
-	return &basicauthextension.Config{
-		Htpasswd: &basicauthextension.HtpasswdSettings{
-			Inline: fmt.Sprintf("%s:%s", args.Username, args.Password),
-		},
-	}, nil
+	c := &basicauthextension.Config{
+		Htpasswd: &basicauthextension.HtpasswdSettings{},
```
ptodev, line 123: It'd be better if this is only created if `args.HtpasswdFile` is not empty.

internal/component/otelcol/auth/basic/basic.go
```diff
@@ -31,8 +37,10 @@ func init() {
 
 // Arguments configures the otelcol.auth.basic component.
 type Arguments struct {
-	Username string            `alloy:"username,attr"`
-	Password alloytypes.Secret `alloy:"password,attr"`
+	Username string            `alloy:"username,attr,optional"`
+	Password alloytypes.Secret `alloy:"password,attr,optional"`
+
+	HtpasswdFile string `alloy:"htpasswd_file,attr,optional"`
```
ptodev:
```suggestion
	HtpasswdFile string `alloy:"htpasswd_file,block,optional"`
```

To match OTel, you'll need to create an htpasswd block which has two attributes - `file` and `inline`.

internal/component/otelcol/auth/basic/basic.go
```diff
@@ -57,11 +83,17 @@ func (args Arguments) ConvertClient() (otelcomponent.Config, error) {
...
-		Htpasswd: &basicauthextension.HtpasswdSettings{
-			Inline: fmt.Sprintf("%s:%s", args.Username, args.Password),
-		},
-	}, nil
+	c := &basicauthextension.Config{
+		Htpasswd: &basicauthextension.HtpasswdSettings{},
+	}
+	if args.HtpasswdFile != "" {
+		c.Htpasswd.File = args.HtpasswdFile
+	}
```
ptodev: The best way to do the conversion is to make `htpasswd` a struct and make a convert function for it, similarly to `KafkaExporterSignalConfig`.

internal/component/otelcol/auth/basic/basic.go
pkarakal: so in case the `htpasswd` block is provided do you want to ignore the username password that may be provided? Asking mainly cause the extension supports having both file and inline, and inline is currently created by just combining username/password, which is a behavior I tried to keep.

pkarakal: @ptodev feel free to let me know if that's what you had in mind.

internal/component/otelcol/auth/basic/basic.go
```diff
@@ -29,10 +35,28 @@ func init() {
...
+}
+
+func (c HtpasswdConfig) convert() *basicauthextension.HtpasswdSettings {
+	settings := &basicauthextension.HtpasswdSettings{}
+	if c.File != "" {
+		settings.File = c.File
+	}
+	if c.Inline != "" {
+		settings.Inline = c.Inline
+	}
```
ptodev: There's no need to check if they're empty. It's ok to pass them as they are.

internal/component/otelcol/auth/basic/basic.go
ptodev: Hi, apologies for the late reply! The first time reviewed the PR I didn't quite understand some aspects of the upstream component (that one config block is for server whereas the other is for client). I also didn't realise that Alloy sets the client config directly, without having a client block in Alloy syntax. This makes it more complicated...

Ideally we should match the OTel config 1:1. But we also don't want to break existing users, at least not until Alloy v2 is released. So I think the best thing for you to do now is add both a `htpasswd` and a `client_auth` block, and to mention in the docs that the current arguments are deprecated. Then the code should ignore them if there is a `client_auth` block. I hope this makes sense? I'm sorry that it has to be so complicated :/

If we do this then I'll also open an issue with a `v2.0-breaking-change` label to remove the current arguments in Alloy v2.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,17 +36,23 @@ otelcol.auth.basic "<LABEL>" {
...
+  
+  client_auth {
+    username = "<USERNAME>"
+    password = "<PASSWORD>"
+  }
 }
 ```
 
 ## Arguments
+Deprecated in favor of the [`client_auth`][client_auth] and [`htpasswd`][htpasswd] blocks.
```
clayton-cornell:
```suggestion
## Arguments

Deprecated in favor of the [`client_auth`][client_auth] and [`htpasswd`][htpasswd] blocks.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -55,12 +61,25 @@ You can use the following block with `otelcol.auth.basic`:
...
+| [`client_auth`][client_auth]     | Configures the service authentication for an exporter                      | no       |
 | [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
 | [`htpasswd`][htpasswd]           | Configures the service authentication for a receiver                       | no       |
 
+[client_auth]: #client_auth
 [debug_metrics]: #debug_metrics
 [htpasswd]: #htpasswd
 
+### `client_auth`
+The `client_auth` block configures how the client extensions will authenticate to servers.
```
clayton-cornell:
```suggestion
### `client_auth`

The `client_auth` block configures how the client extensions will authenticate to servers.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -55,12 +61,25 @@ You can use the following block with `otelcol.auth.basic`:
...
+
+| Name       | Type     | Description                                       | Default | Required |
+|------------|----------|---------------------------------------------------|---------|----------|
+| `password` | `string` | Password to use for basic authentication requests |         | yes      |
+| `inline`   | `string` | Username to use for basic authentication requests |         | yes      |
```
clayton-cornell:
```suggestion
| `inline`   | `string` | Username to use for basic authentication requests |         | yes      |
| `password` | `string` | Password to use for basic authentication requests |         | yes      |
```

Rows in alphabetic order

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -55,12 +61,25 @@ You can use the following block with `otelcol.auth.basic`:
...
+
+If both the `client_auth` block and the `username`/`password` attributes are specified, to more closely match the
```
clayton-cornell:
```suggestion
If both the `client_auth` block and the `username` and `password` attributes are specified, to more closely match the 
```

We should avoid using a slash in body text.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -74,6 +93,10 @@ The `htpasswd` block configures how the server extensions will authenticate call
 | `file`   | `string` | Path to the htpasswd file to use for basic authentication requests | `""`    | no       |
 | `inline` | `string` | The htpasswd file inline content                                   | `""`    | no       |
 
+If both the `htpasswd` block and the `username`/`password` attributes are specified, to not break existing functionality
+and to more closely match the upstream extension's behavior, the `username` and `password` are appended to the `inline`
+attribute of this block.
```
clayton-cornell:
```suggestion
If both the `htpasswd` block and the `username` and `password` attributes are specified, the `username` and password` are appended to the `inline` attribute of this block.
This is done to make sure that existing functionality continues to work, and to more closely match the behavior of the upstream extension.
```

Attempting to simplify the sentences. Split into two sentences, and reorder the info/idea flow a little.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +136,95 @@ otelcol.auth.basic "creds" {
 }
 ```
 
-[otelcol.exporter.otlp]: ../otelcol.exporter.otlp/
+
+### Authenticating requests for receivers
+
+#### Use Username/Password
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a single
```
clayton-cornell:
```suggestion
#### Use username and password

This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a single
```

We should have a blank line between a heading and body text. 

Removed the `/` from the heading

Use sentence case in headings instead of initial caps.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +136,95 @@ otelcol.auth.basic "creds" {
...
+otelcol.exporter.debug "default" {}
+
+otelcol.auth.basic "creds" {
+  username = "demo"
+  password = sys.env("API_KEY")
+}
+```
+
+#### Use htpasswd file
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using an htpasswd
```
clayton-cornell:
```suggestion
#### Use htpasswd file

This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using an htpasswd 
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +136,95 @@ otelcol.auth.basic "creds" {
...
+
+otelcol.auth.basic "creds" {
+  htpasswd {
+    file = "/etc/alloy/.htpasswd"
+  }
+}
+```
+
+#### Combination of both
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a combination
```
clayton-cornell:
```suggestion
#### Combination of both

This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a combination
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +136,95 @@ otelcol.auth.basic "creds" {
...
+  password = sys.env("API_KEY")
+  
+  htpasswd {
+    file = "/etc/alloy/.htpasswd"
+  }
+}
+```
+
+
+[otelcol.receiver.otlp]: ../otelcol.receiver.otlp/
```
clayton-cornell, line 268:
````suggestion
```

[otelcol.receiver.otlp]: ../otelcol.receiver.otlp/
````

Removed extra blank line

File: docs/sources/reference/components/otelcol/otelcol.auth.basic.md [hunk 15]
```diff
@@ -91,4 +136,95 @@ otelcol.auth.basic "creds" {
 }
 ```
 
-[otelcol.exporter.otlp]: ../otelcol.exporter.otlp/
+
+### Authenticating requests for receivers
+
```
clayton-cornell, line 140: Style-wise we should never have one heading follow another without some descriptive text between.

Maybe something like

`The following examples demonstrate how you can use username and password or an htpasswd file for basic authentication.`

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +137,135 @@ otelcol.auth.basic "creds" {
 }
 ```
 
+
+### Authenticating requests for receivers
+
+The following examples demonstrate how to perform basic authentication using either a username and password combination,
+the client authentication block, an htpasswd file or a combination of them.
```
clayton-cornell:
```suggestion
The following examples demonstrate how to perform basic authentication using either a username and password, the `client_auth` block, an htpasswd file. or a combination of these authentication methods.
```

Name the block, add a missing comma, and make it more explicit what "them" means in this context.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +137,135 @@ otelcol.auth.basic "creds" {
...
+    file = "/etc/alloy/.htpasswd"
+  }
+}
+```
+
+#### Combination of both
+
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a combination
+of both an htpasswd file and username/password. Note that if the username provided also exists in the htpasswd file, it 
+takes precedence over the one in the htpasswd file:
```
clayton-cornell:
```suggestion
This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a combination of both an htpasswd file, and username and password.
If you specify the `username` argument and include the username in the htpasswd file, the `username` argument takes precedence.
```

I need some clarity on the second sentence. I tried to clarify the intent a but bit... I'm not sure I've got it right. 

I read this as... If the user provides a password arg and sets password in the htpasswd file, the password arg take precedence. Does this imply they can be different usernames? The original wording indicates (to me anyway) that they would be identical... and if they are the same, does it matter to the user that the arg take precedence? I can only see this being important if they would be different usernames... 

Also... is this sentence in the right place? It looks like it's important and maybe should be included in the `htpasswd` block information vs down in an example.

pkarakal: This is basically the behavior of the upstream component.

 If for some reason you specify **the same user** both in the `.htpasswd` file and `inline`, the one that is `inline` takes precedence. Since in the case of specifying both the deprecated username and password combination and the htpasswd, we append the username and password to the inline block, it takes precedence over the rest. 

I think that as you suggested, this is an important enough information that should be moved to the htpasswd block instead.

clayton-cornell:
Docs look of for this iteration :-) I'll wait on code reviews before clicking approve for docs.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -31,32 +31,88 @@ You can specify multiple `otelcol.auth.basic` components by giving them differen
...
+  
+  htpasswd {
+    file = "/etc/alloy/.htpasswd"
+    inline = "<USERNAME>:<PASSWORD>"
+  }
+  
+  client_auth {
+    username = "<USERNAME>"
+    password = "<PASSWORD>"
+  }
```
clayton-cornell: I would remove this section. the `htpasswd` and `client_auth` blocks are optional. We should keep the **Usage** section to just the required elements.

pkarakal:
Can I have an update on this, or an ETA on when you expect this to get reviewed/released?

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +138,160 @@ otelcol.auth.basic "creds" {
...
+
+otelcol.exporter.debug "default" {}
+
+otelcol.auth.basic "creds" {
+  username = "demo"
+  password = sys.env("API_KEY")
+}
+```
+
+#### Use client authentication
```
dehaansa, line 143: What's the difference between this example and the `Forward signals to exporters` example?

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +138,160 @@ otelcol.auth.basic "creds" {
...
+otelcol.auth.basic "creds" {
+  htpasswd {
+    inline = "user1:password1\nuser2:password2"
+  }
+}
+```
+
+#### Combination of htpasswd and the username and password attributes
+
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using a combination
```
dehaansa: As this is not a recommended path, we probably shouldn't include it in the docs.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
pkarakal: The difference is that the first (`forward signals to exporters` uses the deprecated username and password top-level arguments whereas this uses the new client authentication block. Since I have split this into client and server authentication examples, I think I should just move this a below `Forward signals to exporters`.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
pkarakal: I think that this could be a good migration path for users to client and htpasswd blocks, but if you don't agree, I have no problem removing this.

They could first add a new user to the htpasswd file, start using the new one in their systems and once the old user is not used anymore, they could just drop it. WDYT?

I can explicitly add this as a migration path if you'd like.

dehaansa: My opinion is that if we clearly state fields are deprecated, we don't want to show them in configuration examples. CC @clayton-cornell any thoughts on this?

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
dehaansa: I think pairing them together would be better, but similar to the other thread I'm not a fan of showing deprecated fields in config examples in most cases. Let's see what our docs team says.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
clayton-cornell: I missed that in my earlier review. 

I agree, if args/fields are clearly marked as deprecated, we should not be providing examples that use these args. The examples should show current state only. If someone needs to see the previous implementation, the docs showing the args/fields before deprecation are available in a previous doc version. 

For migration path... maybe add as a note admonition? with text something like "To migrate an existing configuration to use `client_auth` and `htpasswd` blocks, you can..."

pkarakal: @clayton-cornell what do you think about this? I don't have any strong feelings regardless but I would really like for this to be in a future release since I would like to reduce the components I have to support in production (and the actual hops the requests take).

clayton-cornell: And I forgot to commit my pending comments on this PR. Sorry.

pkarakal: Thank you for the insights. I have removed the example of the deprecated attributes and moved the migration guides to a note admonition as you suggested. Feel free to take a look and let me know. 

(I have also rebased on top of the latest main as the pr was pretty out of date)

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,73 @@ otelcol.auth.basic "<LABEL>" {
...
 | [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
+| [`htpasswd`][htpasswd]           | Configures server authentication using htpasswd format for receivers       | no       |
 
+[client_auth]: #client_auth
 [debug_metrics]: #debug_metrics
+[htpasswd]: #htpasswd
+
+### `client_auth`
+
+The `client_auth` block configures credentials that client extensions (such as exporters) will use to authenticate to servers.
```
clayton-cornell:
```suggestion
The `client_auth` block configures credentials that client extensions (such as exporters) use to authenticate to servers.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,73 @@ otelcol.auth.basic "<LABEL>" {
...
+
+| Name       | Type     | Description                                         | Default | Required |
+|------------|----------|-----------------------------------------------------|---------|----------|
+| `password` | `string` | Password to use for basic authentication requests   |         | yes      |
+| `username` | `string` | Username to use for basic authentication requests   |         | yes      |
```
clayton-cornell:
```suggestion
| Name       | Type     | Description                                        | Default | Required |
| ---------- | -------- | -------------------------------------------------- | ------- | -------- |
| `password` | `string` | Password to use for basic authentication requests. |         | yes      |
| `username` | `string` | Username to use for basic authentication requests. |         | yes      |
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,73 @@ otelcol.auth.basic "<LABEL>" {
...
 ## Blocks
 
 You can use the following block with `otelcol.auth.basic`:
 
 | Block                            | Description                                                                | Required |
-| -------------------------------- | -------------------------------------------------------------------------- | -------- |
+|----------------------------------|----------------------------------------------------------------------------|----------|
+| [`client_auth`][client_auth]     | Configures client authentication credentials for exporters                 | no       |
 | [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
+| [`htpasswd`][htpasswd]           | Configures server authentication using htpasswd format for receivers       | no       |
```
clayton-cornell:
```suggestion
|----------------------------------|----------------------------------------------------------------------------|----------|
| [`client_auth`][client_auth]     | Configures client authentication credentials for exporters.                | no       |
| [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state. | no       |
| [`htpasswd`][htpasswd]           | Configures server authentication using htpasswd format for receivers.      | no       |
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,73 @@ otelcol.auth.basic "<LABEL>" {
...
 
+### `htpasswd`
+
+The `htpasswd` block configures how the server extensions (such as receivers) will authenticate incoming requests using
+the htpasswd format.
+
+| Name     | Type     | Description                                                        | Default | Required |
+|----------|----------|--------------------------------------------------------------------|---------|----------|
+| `file`   | `string` | Path to the htpasswd file to use for basic authentication requests | `""`    | no       |
+| `inline` | `string` | The htpasswd file content in inline format                         | `""`    | no       |
```
clayton-cornell:
```suggestion
| Name     | Type     | Description                                                           | Default | Required |
| -------- | -------- | --------------------------------------------------------------------- | ------- | -------- |
| `file`   | `string` | Path to the `htpasswd` file to use for basic authentication requests. | `""`    | no       |
| `inline` | `string` | The `htpasswd` file content in inline format.                         | `""`    | no       |
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,73 @@ otelcol.auth.basic "<LABEL>" {
...
+
+You can specify either `file`, `inline`, or both. When using `inline`, the format should be `username:password` with 
+each user on a new line.
```
clayton-cornell:
```suggestion
You can specify either `file`, `inline`, or both.
When you use `inline`, the format should be `username:password` with each user on a new line.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +138,139 @@ otelcol.auth.basic "creds" {
...
+  }
+}
+```
+
+{{< admonition type="note" >}}
+To migrate from the deprecated `username` and `password` attributes, move them into the `client_auth` block for client 
+authentication.
+{{< /admonition >}}
+
+
```
clayton-cornell:
```suggestion
To migrate from the deprecated `username` and `password` attributes, move them into the `client_auth` block for client authentication.
{{< /admonition >}}

```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +138,139 @@ otelcol.auth.basic "creds" {
...
+#### Use htpasswd file
+
+This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using an htpasswd 
+file containing the users to use for basic auth:
```
clayton-cornell:
```suggestion
This example configures [`otelcol.receiver.otlp`][otelcol.receiver.otlp] to use basic authentication using an `htpasswd` file containing the users to use for basic authentication:
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +138,139 @@ otelcol.auth.basic "creds" {
...
+otelcol.auth.basic "creds" {
+  htpasswd {
+    file = "/etc/alloy/.htpasswd"
+  }
+}
+```
+
+#### Use htpasswd inline content
+
+This example shows how to specify htpasswd content directly in the configuration:
```
clayton-cornell:
```suggestion
This example shows how to specify `htpasswd` content directly in the configuration:
```

clayton-cornell (commented):
Some minor tidy suggestions on the docs markdown. There's still passive voice, but I'll leave for this PR.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,68 @@ otelcol.auth.basic "<LABEL>" {
...
+
+The `client_auth` block configures credentials that client extensions (such as exporters) use to authenticate to servers.
+
+| Name       | Type     | Description                                        | Default | Required |
+| ---------- | -------- | -------------------------------------------------- | ------- | -------- |
+| `password` | `string` | Password to use for basic authentication requests. |         | yes      |
+| `username` | `string` | Username to use for basic authentication requests. |         | yes      |
+
+{{< admonition type="note" >}}
+When you specify both the `client_auth` block and the deprecated top-level `username` and `password` attributes, the `client_auth` block takes precedence and Alloy ignores the top-level attributes for client authentication.
```
clayton-cornell:
```suggestion
When you specify both the `client_auth` block and the deprecated top-level `username` and `password` attributes, the `client_auth` block takes precedence and {{< param "PRODUCT_NAME" >}} ignores the top-level attributes for client authentication.
```

Missed this in my previous reviews. We use a product variable for the Alloy product name. The result is the same in the published docs.. just good to update this to the variable for consistency.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,68 @@ otelcol.auth.basic "<LABEL>" {
...
+| Name     | Type     | Description                                                           | Default | Required |
+| -------- | -------- | --------------------------------------------------------------------- | ------- | -------- |
+| `file`   | `string` | Path to the `htpasswd` file to use for basic authentication requests. | `""`    | no       |
+| `inline` | `string` | The `htpasswd` file content in inline format.                         | `""`    | no       |
+
+You can specify either `file`, `inline`, or both.
+When you use `inline`, the format should be `username:password` with each user on a new line.
+
+{{< admonition type="note" >}}
+When you specify both the `htpasswd` block and the deprecated top-level `username` and `password` attributes, Alloy automatically appends the deprecated credentials to the `inline` content. This allows authentication using credentials from both the htpasswd configuration and the deprecated attributes.
```
clayton-cornell:
```suggestion
When you specify both the `htpasswd` block and the deprecated top-level `username` and `password` attributes, {{< param "PRODUCT_NAME" >}} automatically appends the deprecated credentials to the `inline` content.
This allows authentication using credentials from both the `htpasswd` configuration and the deprecated attributes.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -91,4 +136,133 @@ otelcol.auth.basic "creds" {
...
+
+otelcol.auth.basic "creds" {
+  htpasswd {
+    inline = "user1:password1\nuser2:password2"
+  }
+}
+```
+
+{{< admonition type="note" >}}
+To make the migration from the deprecated `username` and `password` attributes easier, you can specify both the deprecated attributes and the `htpasswd` block in the same configuration. Alloy appends the deprecated attributes to the `htpasswd` content.
```
clayton-cornell:
```suggestion
To make the migration from the deprecated `username` and `password` attributes easier, you can specify both the deprecated attributes and the `htpasswd` block in the same configuration.
{{< param "PRODUCT_NAME" >}} appends the deprecated attributes to the `htpasswd` content.
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -36,27 +36,68 @@ otelcol.auth.basic "<LABEL>" {
...
+When you specify both the `client_auth` block and the deprecated top-level `username` and `password` attributes, the `client_auth` block takes precedence and Alloy ignores the top-level attributes for client authentication.
+{{< /admonition >}}
 
 ### `debug_metrics`
 
 {{< docs/shared lookup="reference/components/otelcol-debug-metrics-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
+### `htpasswd`
+
+The `htpasswd` block configures how server extensions (such as receivers) authenticate incoming requests using the htpasswd format.
```
clayton-cornell:
```suggestion
The `htpasswd` block configures how server extensions (such as receivers) authenticate incoming requests using the `htpasswd` format.
```

clayton-cornell (approved):

PR #3922, opened 2025-06-30
database_observability: add auto-enable setup consumers to query_sample

cristiangreco:
#### PR Description
Introduce a check in the `query_sample` collector to automatically enable the `events_statements_cpu` consumer in the `performance_schema.setup_consumers` table.

This is guarded by two config options:
- `allow_update_performance_schema_settings`: to enable "updates" to performance_schema settings altogether.
- `auto_enable_setup_consumers`: to start auto-enabling of the setting specific to the `query_sample` collector.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/query_sample.go
+++ b/internal/component/database_observability/mysql/collector/query_sample.go
@@ -77,24 +77,26 @@ const updateSetupConsumers = `
 		WHERE name = 'events_statements_cpu'`
 
 type QuerySampleArguments struct {
-	DB                       *sql.DB
-	InstanceKey              string
-	CollectInterval          time.Duration
-	EntryHandler             loki.EntryHandler
-	DisableQueryRedaction    bool
-	AutoEnableSetupConsumers bool
+	DB                          *sql.DB
+	InstanceKey                 string
+	CollectInterval             time.Duration
+	EntryHandler                loki.EntryHandler
+	DisableQueryRedaction       bool
+	AutoEnableSetupConsumers    bool
+	SetupConsumersCheckInterval time.Duration
 
 	Logger log.Logger
 }
 
 type QuerySample struct {
-	dbConnection             *sql.DB
-	instanceKey              string
-	collectInterval          time.Duration
-	entryHandler             loki.EntryHandler
-	sqlParser                parser.Parser
-	disableQueryRedaction    bool
-	autoEnableSetupConsumers bool
+	dbConnection                *sql.DB
+	instanceKey                 string
+	collectInterval             time.Duration
+	entryHandler                loki.EntryHandler
+	sqlParser                   parser.Parser
+	disableQueryRedaction       bool
+	autoEnableSetupConsumers    bool
+	setupConsumersCheckInterval time.Duration
 
 	logger  log.Logger
 	running *atomic.Bool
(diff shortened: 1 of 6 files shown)


internal/component/database_observability/mysql/component.go
```diff
@@ -50,11 +50,12 @@ var (
...
 type Arguments struct {
-	DataSourceName    alloytypes.Secret   `alloy:"data_source_name,attr"`
-	CollectInterval   time.Duration       `alloy:"collect_interval,attr,optional"`
-	ForwardTo         []loki.LogsReceiver `alloy:"forward_to,attr"`
-	EnableCollectors  []string            `alloy:"enable_collectors,attr,optional"`
-	DisableCollectors []string            `alloy:"disable_collectors,attr,optional"`
+	DataSourceName                alloytypes.Secret   `alloy:"data_source_name,attr"`
+	CollectInterval               time.Duration       `alloy:"collect_interval,attr,optional"`
+	ForwardTo                     []loki.LogsReceiver `alloy:"forward_to,attr"`
+	EnableCollectors              []string            `alloy:"enable_collectors,attr,optional"`
```
fridgepoet, line 56: I wonder, (not sure if it's possible), in the future when we have several different setup_consumers we want to configure in a fine-grained manner, will it be easier provide a list to enable/disable similar to EnableCollectors/DisableCollectors? 

Might be overengineered, though, so maybe it's better to start with a simple boolean for now that is at the lowest level.

fridgepoet:
Should we distinguish because something that "collects" and something that will "write"?

fridgepoet (approved):

internal/component/database_observability/mysql/component.go
cristiangreco: Hmm something to think about, but agree to start simple for now as we don't know how widespread this functionality will be.

gaantunes:
I wonder if we should create a separate collector that is responsible for maintaining all write operations based on enabled colletors array (taking into account defaults of course). Detaching it would allow a longer scrape period for these operations, instead of checking/writing every time we do a scrape in each collector.

cristiangreco:
> I wonder if we should create a separate collector that is responsible for maintaining all write operations based on enabled colletors array (taking into account defaults of course). Detaching it would allow a longer scrape period for these operations, instead of checking/writing every time we do a scrape in each collector.

Yes I agree the the two things should not be necessarily tied together. I'd prefer to postpone that for a followup PR though if that's ok (so here we first verify this is working and proves valuable, then optimise it).

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -184,6 +192,12 @@ func (c *QuerySample) initializeBookmark(ctx context.Context) error {
 }
 
 func (c *QuerySample) fetchQuerySamples(ctx context.Context) error {
+	if c.autoEnableSetupConsumers {
+		if err := c.updateSetupConsumersSettings(ctx); err != nil {
+			return err
+		}
+	}
+
```
cristiangreco: As commented by @gaantunes we could decouple the update from the regular collector loop. I'm keen to postpone that work for a followup PR, here I'm more interested in the correctness of the approach.

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -381,6 +395,23 @@ func (c *QuerySample) fetchQuerySamples(ctx context.Context) error {
...
+		level.Error(c.logger).Log("msg", "failed to update performance_schema.setup_consumers", "err", err)
+		return err
+	}
+
+	rowsAffected, err := rs.RowsAffected()
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to get rows affected from performance_schema.setup_consumers", "err", err)
+		return err
+	}
+	level.Debug(c.logger).Log("msg", "updated performance_schema.setup_consumers", "rows_affected", rowsAffected)
```
matthewnolf, line 427: Shouldn't we check rows affected = 1? If it's something else, the operation didn't work as expected

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -381,6 +395,23 @@ func (c *QuerySample) fetchQuerySamples(ctx context.Context) error {
...
+func (c *QuerySample) updateSetupConsumersSettings(ctx context.Context) error {
+	rs, err := c.dbConnection.ExecContext(ctx, updateSetupConsumers)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to update performance_schema.setup_consumers", "err", err)
+		return err
+	}
+
+	rowsAffected, err := rs.RowsAffected()
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to get rows affected from performance_schema.setup_consumers", "err", err)
```
matthewnolf: Are we going to end up double logging this error? Here and in the ticker loop

internal/component/database_observability/mysql/collector/query_sample.go
matthewnolf: I'd be inclined to include this now, or atleast tie it to `setup_consumers_collect_interval`. Running this every minute seems too often

internal/component/database_observability/mysql/collector/query_sample.go
cristiangreco: Good point, I'll change these to be `fmt.Errorf` rather than to log here.

internal/component/database_observability/mysql/collector/query_sample.go
cristiangreco: `rowsAffected == 0` is also fine, it just means it was already enabled (that's why this is just a debug log)

gaantunes (approved):

PR #3923, opened 2025-06-30
database_observability: move time-related functions to a separate file

cristiangreco:
#### PR Description
Move stuff related to time, uptime and overflows to a separate file to make it easier to test and reuse.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/time.go
+++ b/internal/component/database_observability/mysql/collector/time.go
@@ -2,7 +2,7 @@ package collector
 
 import "math"
 
-var picosecondsOverflowInSeconds = PicosecondsToSeconds(float64(math.MaxUint64))
+var picosecondsOverflowInSeconds = picosecondsToSeconds(float64(math.MaxUint64))
 
 const (
 	picosecondsPerSecond      float64 = 1e12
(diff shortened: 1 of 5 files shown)


internal/component/database_observability/mysql/collector/time.go
```diff
@@ -0,0 +1,62 @@
...
+	millisecondsPerSecond     float64 = 1e3
+	millisecondsPerPicosecond float64 = 1e9
+	nanosecondsPerMillisecond float64 = 1e6
+)
+
+// CalculateWallTime calculates the wall-clock timestamp for an event.
+// The timerPicoseconds indicates event timing since server startup.
+// Since this value can overflow after approximately ~213 days (column type of bigint unsigned),
+// this function accounts for overflows by calculating the number of previous overflows and
+// compensating accordingly. Returns the timestamp in milliseconds when the event occurred.
```
cristiangreco: This is the same code as before, I've just extracted a function comment and appended time-related suffix to parameters.

internal/component/database_observability/mysql/collector/time.go
matthewnolf, line 1: Do the functions here need to be exported? I'm assuming it's just for the tests, which previously were not in a separate test package

cristiangreco: Yes they're exported for testing. Are you suggesting we avoid using a separate test package to keep things consistent?

matthewnolf: Yep, for consistency as you mention and so that we don't then need to export them

cristiangreco: Reverted the change.

Issue #3926, opened 2025-07-01
prometheus.remote_write: WAL can hold inactive series forever


kgeckhart:
### What's wrong?

It's possible to trigger a case where the prometheus WAL contains multiple RefIDs for the same series. When this happens the WAL can only GC one of the series leaving all the rest in memory forever. This happens because of the disconnect between segment checkpointing and series gc'ing.

The problem,
* The WAL assumes a RefID is unique across the WAL
* When a series is deleted via the GC process [the most recent WAL segment is tracked with the SeriesID](https://github.com/grafana/alloy/blob/905a1648c6129b7372a8d1fb48cf79b356c909d3/internal/static/metrics/wal/wal.go#L570-L575) to ensure we don't delete a series that still exists on the WAL. 
* We use this map as an [input to a keep function for checkpointing](https://github.com/grafana/alloy/blob/905a1648c6129b7372a8d1fb48cf79b356c909d3/internal/static/metrics/wal/wal.go#L523-L530), when the series exists in the deleted map with a segment that's newer than the one we are checkpointing we will keep it

This growth is rather slow but in our case we have accumulated tens of millions of active series that cannot be gc'ed from the WAL active series.

### Steps to reproduce

1. Truncate the WAL GC'ing a series (RefID1) but making sure it stays in the checkpoint
2. Recreate the series in the WAL (RefID2)
3. Before RefID1 is removed from the checkpoint alloy is restarted (this can take multiple rounds of checkpointing over many hours so very liable to happen)
4. On startup alloy will now have RefID1 and RefID2 as active series breaking the "RefIDs must be unique rule" (RefID1 still exists in the checkpoint and RefID 2 exists in the segments so both are loaded)
5. RefID1 cannot be GC'ed because GC'ing goes based on hash and a hash can only point to a singular RefID which in this case will be RefID2
6. If RefID2 is GC'ed fully, a restart of alloy can allow RefID1 to be freed but we are more than likely to see this repeat over and over slowly accumulating active series

Test which shows it in action: https://github.com/grafana/alloy/commit/01f6937d21f9bf594238e10d6184985caa1db1f4

### Visual representation of the issue

Consider a series `foo` with label `bar` which received a few samples stored in the WAL at 10:15, then stopped. The samples for `foo` are in WAL segment C with 1 currently active series. Series in memory contains SeriesRef 10 and hashes contains the label hash of `bar` mapped to SeriesRef 10.

```
           10:00        12:00
[...]

PR #3927, opened 2025-07-01
prometheus.remote_write: prevent unbounded growth of inactive series in WAL

kgeckhart:
#### PR Description
This PR fixes an issue that can cause active series in the WAL to grow without bound described in, https://github.com/grafana/alloy/issues/3926, by refactoring the `loadWAL` implementation to more closely mirror the prometheus tsdb implementation it was likely based upon (https://github.com/prometheus/prometheus/blob/e83dc66bdb67a01ff06a96e81faf66b0095d9948/tsdb/head_wal.go#L78). Notes to the reviewer has a breakdown of the changesets + rationale.

Fixes https://github.com/grafana/alloy/issues/3926

#### Notes to the Reviewer

* The first changeset is the test that shows the issue fully described in the linked issue
* https://github.com/grafana/alloy/pull/3927/commits/901772eab6551da064d9e08257df8c5f89caf73f - Ensure WAL replay considers the fact that the same labels can have duplicate SeriesRefs
  * These changes are all inspired by how [tsdb head](https://github.com/prometheus/prometheus/blob/e83dc66bdb67a01ff06a96e81faf66b0095d9948/tsdb/head_wal.go#L255-L275) decodes series during WAL replay
  * New series are created only if they represent a unique label hash through the new `stripeSeries.GetOrSet`
  * We only track series that were not created (aka duplicates) in `multiRef`
  * Duplicate series are tracked in the WAL as deleted with the lastSegment allowing them to kept in checkpoints until they are gone from the WAL
* https://github.com/grafana/alloy/pull/3927/commits/abd5e6392499525e00ff58a535c1e51980ec64b1 - Rename multiRef to duplicateRefToValidRef and use it rewrite duplicate series as expected
  * Alloy was using `multiRef` to track all series created and the change to only track duplicates series breaks the logic in all sample decode steps resulting in most samples being considered `nonExistentSeriesRefs`
  * `multiRef` is used to rewrite the sample SeriesRef if needed
  * A SeriesRef is considered a nonExistentSeriesRef only if it does not exist in series
  * Since I struggled to realize the intent of `multiRef` based on the short name I renamed it to `duplicateRefToValidRef` which IMO helps illustrate its purpose a lot more clearly
* https://github.com/grafana/alloy/pull/3927/commits/1575ac5440753c599fde5ccb139f3611a4988414 - Keep duplicated SeriesRefs until the last segment they are read from
  * As we are adding the duplicate SeriesRefs to deleted we were using the last segment but we are replaying so we can know exactly where we are at
  * If we track the last segment we found the SeriesRef in deleted we should be able to clear these duplicate refs from the checkpoint earlier
  * This is potentially upstreamable to tsbd head as well

#### PR Checklist

- [ ] CHANGELOG.md updated (will add when closer to ready to merge to prevent ongoing conflicts)
- [x] Tests updated


bboreham:
Is this relevant to the upstream Prometheus (Agent) WAL too?

kgeckhart:
> Is this relevant to the upstream Prometheus (Agent) WAL too?

@bboreham great question, I believe so as the code is largely the same between the two but I haven't done the same test upstream yet. If this fix makes sense and works as I expect I'm happy to upstream it as well.

bboreham:
I didn’t really follow the code in this PR. 

There is an upstream issue which might be related: [WAL Checkpoint holds deleted series for 1 extra compaction cycle #12286 ](https://github.com/prometheus/prometheus/issues/12286)

internal/static/metrics/wal/series.go
```diff
@@ -115,9 +122,63 @@ func newStripeSeries(stripeSize int) *stripeSeries {
...
+	if s.initialized.Load() {
+		return 0, false
+	}
+
+	inactiveSeries := 0
+	// Start with hashes first because it's easier to get to a series from the hash than a hash from a series.
+	for _, hashSeries := range s.hashes {
+		for hash, seriesForHash := range hashSeries {
+			for _, series := range seriesForHash {
+				if series.lastTs == 0 {
```
ptodev: Why is a value of 0 a good criteria for deleting the series? Is it a special value?

internal/static/metrics/wal/series.go
```diff
@@ -115,9 +122,63 @@ func newStripeSeries(stripeSize int) *stripeSeries {
 	return s
 }
 
+// MarkInitialized marks the stripeSeries initialized, allowing usage of stripeSeries.gc(). Returns
+// true if the stripeSeries was not initialized before, false otherwise.
+func (s *stripeSeries) MarkInitialized() bool {
+	return s.initialized.CompareAndSwap(false, true)
```
ptodev: It's possible that `RemoveInactiveSeries` is still executing when this returns. Would that be a problem?

kgeckhart:
> I didn’t really follow the code in this PR.
> 
> There is an upstream issue which might be related: [WAL Checkpoint holds deleted series for 1 extra compaction cycle #12286 ](https://github.com/prometheus/prometheus/issues/12286)

Thanks for pointing out the issue @bboreham it's absolutely related but of a different flavor. I'll adapt the flow you originally described in https://github.com/prometheus/prometheus/issues/12286#issuecomment-1519086491 with modifications for this problem.

edit: Visualization was cleaned up and moved to the issue https://github.com/grafana/alloy/issues/3926#issue-3192214499

internal/static/metrics/wal/series.go
kgeckhart: Good point, it's the default value given to a series when it's loaded during WAL replay https://github.com/grafana/alloy/blob/52e92ad437735d7946c86c2036ff3fa786f28a49/internal/static/metrics/wal/wal.go#L375-L380 If the series ends up with a sample we will set the lastTs to the value of the sample https://github.com/grafana/alloy/blob/52e92ad437735d7946c86c2036ff3fa786f28a49/internal/static/metrics/wal/wal.go#L395-L407

I could refactor so the `lastTs` to consider is passed in to the function and this way it's managed between a shared constant. WDYT?

internal/static/metrics/wal/series.go
kgeckhart: Very much so, the `RemoveActiveSeries` doesn't try to take locks because it assumes it's the only thing accessing the data. 

The hard part is I don't have a great way to control the "no active reads and writes" side of this. We use the same functions when first loading the data and when managing the data after it's been loaded. I could update `RemoveInactiveSeries` to take the appropriate locks and consider something could be actively writing. I would need to think through all the semantics of a new entry being Set https://github.com/grafana/alloy/blob/52e92ad437735d7946c86c2036ff3fa786f28a49/internal/static/metrics/wal/series.go#L246-L264 while iterating both data structures and accurately set the removed series count.

internal/static/metrics/wal/series.go
ptodev: I'm just worried that if we add more complexity there could be additional bugs down the line. I'm also not sure I understand the problem. Is 0 not a good value to use, given that the timestamp will be updated for series which have at least one sample?

internal/static/metrics/wal/wal_test.go
ptodev, line 1: Is it necessary to do any additional tests, e.g. in a real environment?

internal/static/metrics/wal/series.go
kgeckhart: > I'm just worried that if we add more complexity there could be additional bugs down the line. I'm also not sure I understand the problem. 

Does https://github.com/grafana/alloy/pull/3927#issuecomment-3028244307 help with illustrating the problem? We need to purge any series that had no samples after replaying the checkpoint + segments. A series that had no samples after replay is an inactive series and keeping them around is a waste of space. The bug which causes this to happen is a bit more complex but the end result is an ever growing checkpoint file filled with series that have no samples.

> Is 0 not a good value to use, given that the timestamp will be updated for series which have at least one sample?

0 just happens to be the magic number we as the timestamp for a series before we read samples. It's the first time some other code has relied upon that magic number so refactoring it out to a constant that can be shared should help prevent further bugs due to accidental changes of that init value.

internal/static/metrics/wal/wal_test.go
kgeckhart: Is there a good place to try to add an integration / acceptance test? Perhaps adding another one similar to https://github.com/grafana/alloy/blob/2619183bb521678f8e133b0f7f88b1a28256c763/internal/component/prometheus/remotewrite/remote_write_test.go#L20-L89? 

I did run it against a real WAL that had accumulated ~14 million inactive series in the checkpoint and it pruned all them keeping the ~500k I expected to be active.

internal/static/metrics/wal/series.go
bboreham: FWIW 0 is considered a valid timestamp in Prometheus, and comes up quite a lot in tests (as do negative timestamps).

It's certainly plausible for Alloy to document it as not a valid timestamp for remote-write.

kgeckhart:
@bboreham I went down the rabbit hole to figure out the potential impact in upstream between prometheus agent and the prometheus tsdb itself (it has very similar WAL replay code). 

The prometheus agent is impacted because it makes the same assumption during WAL replay as alloy, a series that does not exist by its RefID is a series to be created https://github.com/prometheus/prometheus/blob/e83dc66bdb67a01ff06a96e81faf66b0095d9948/tsdb/agent/db.go#L522-L533.

The tsdb itself is not impacted because it does not trust the Series RefID to indicate if a series should be created or not it ends up using the hash. 
https://github.com/prometheus/prometheus/blob/e83dc66bdb67a01ff06a96e81faf66b0095d9948/tsdb/head_wal.go#L255-L261 calls https://github.com/prometheus/prometheus/blob/a11772234de54be257fe60e07b313d974058d4a9/tsdb/head.go#L1752-L1753 which is going to decide based on the existence of the hash if the series should exists or should be created. Edit: Further, if it wasn't created it's going to add it back to be tracked as deleted 

I'll see what it looks like to mirror the tsdb behavior into alloy as I think it should be a simpler fix than what I proposed here.

PR #3937, opened 2025-07-02
database_observability: review and simplify parser methods

cristiangreco:
#### PR Description
Now that we have a single parser implementation, we can simplify it by:
- Removing unused methods from interface
- Avoiding using `any` type

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/parser/parser.go
+++ b/internal/component/database_observability/mysql/collector/parser/parser.go
@@ -1,23 +1,16 @@
 package parser
 
-import "github.com/go-kit/log"
+import (
+	"github.com/pingcap/tidb/pkg/parser/ast"
+)
 
 type Parser interface {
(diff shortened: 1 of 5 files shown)


internal/component/database_observability/mysql/collector/parser/parser.go
```diff
@@ -1,23 +1,16 @@
 package parser
 
-import "github.com/go-kit/log"
+import (
+	"github.com/pingcap/tidb/pkg/parser/ast"
+)
 
 type Parser interface {
```
matthewnolf, line 7: Do we still need the interface?

cristiangreco: It's useful because of the mock parser we have in some tests.

PR #3945, opened 2025-07-04
database_observability: Add skeleton postgres component

matthewnolf:
#### PR Description
This adds a new database observability component for postgres. The purpose of this component will be to collect postgres query analysis data; replicating that of the mysql component.

The required collectors for that functionality will be built over time. This change introduces the initial component so that the collectors can subsequently be added.

The most basic `connection_info` collector is added, so we can begin sending the `database_observability_connection_info` metric for pg databases.

A docs page is added for the new component but the component README is not updated to reference postgres things, while we work out the small details on configuring the DB etc. 

The changelog isn't updated to include this new addition, as we don't expect adoption of this component yet - however happy to add an entry if it makes sense to do so.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/compatibility/_index.md
+++ b/docs/sources/reference/compatibility/_index.md
@@ -46,6 +46,7 @@ The following components, grouped by namespace, _export_ Targets.
 
 {{< collapse title="database_observability" >}}
 - [database_observability.mysql](../components/database_observability/database_observability.mysql)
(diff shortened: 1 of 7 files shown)


cristiangreco (approved):
🚀

PR #3957, opened 2025-07-08
dbo11y: in `explain_plan` collector grab connection while fetching plan

cristiangreco:
#### PR Description
The collector issues an `USE` statement to set the schema before running the `EXPLAIN` statement. This requires grabbing a dedicated db connection from the pool to avoid issues with concurrent queries from other collectors that may be running on the same connection.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/explain_plan.go
+++ b/internal/component/database_observability/mysql/collector/explain_plan.go
@@ -616,23 +616,9 @@ func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
 		}
 		logger = log.With(logger, "schema_name", *qi.schemaName)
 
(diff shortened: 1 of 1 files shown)


cristiangreco:
> By the way, is there any test we can update for this?

Apparently this path is not covered by tests. I'll look into that as a followup if that's ok.

fridgepoet:
Happy to help out with any followup tasks if needed, Cristian, just lemme know!

cristiangreco:
> Happy to help out with any followup tasks if needed, Cristian, just lemme know!

Sure, if you want to go ahead with adding some specific tests here I'll leave it with you.

PR #3965, opened 2025-07-08
Fix usage for loki enrich documentation

dehaansa:
#### PR Description
A user in community slack noted that the usage docs had incorrect attribute names in the loki enrich docs.


mattdurham (approved):
lgtm

PR #3968, opened 2025-07-09
database_observability: use tokenizer to extract table names as fallback

cristiangreco:
#### PR Description
Improve parsing of table names from sql statements by using go-sqllexer as a fallback when the primary parser fails.

This means that there are chances we log a partial list of table names when the sql statement is truncated, but we're heading towards making alloy more resilient and mark statements as deemed unparseable.

n.a.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,8 @@ v1.10.0-rc.0
 - (_Experimental_) Additions to experimental `database_observability.mysql` component:
   - Add `explain_plan` collector to `database_observability.mysql` component. (@rgeyer)
   - `locks`: addition of data locks collector (@gaantunes @fridgepoet)
-  - Query sample collector is now enabled by default (@matthewnolf)
+  - `query_sample` collector is now enabled by default (@matthewnolf)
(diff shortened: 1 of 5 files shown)


fridgepoet (approved):

gaantunes:
I wonder if this new parser has the same issue that lists CTE tables as actual tables?

cristiangreco:
> I wonder if this new parser has the same issue that lists CTE tables as actual tables?

As chatted, CTE tables appear only in the list of tables touched by a particular query, not in the global list of tables. 

But for a more direct answer, yes CTE tables should still be detected by the lexer.

gaantunes:
> > I wonder if this new parser has the same issue that lists CTE tables as actual tables?
> 
> As chatted, CTE tables appear only in the list of tables touched by a particular query, not in the global list of tables.
> 
> But for a more direct answer, yes CTE tables should still be detected by the lexer.

We can seek classifying these as temp tables in a future iteration.

gaantunes (approved):

PR #3973, opened 2025-07-09
database_observability: Allow fetching explain plans for queries beginning in 'with'

rgeyer:
#### PR Description

While uncommon, it is acceptable to begin a select query using the `WITH` verb. See [docs](https://dev.mysql.com/doc/refman/8.4/en/with.html).

As such, we should be fetching explain plans for those queries as well.


fridgepoet:
nice! btw, can we have a test protecting this?

rgeyer:
> nice! btw, can we have a test protecting this?

Added some for this functionality, and some other low hanging fruit surrounding these queries. Will follow up with more test coverage.

https://github.com/grafana/alloy/pull/3973/commits/87afe13e1ca492b0921b1fca0f798390f6301337

cristiangreco (approved):
LGTM

PR #3982, opened 2025-07-11
database_observability: add initial steps for Postgres setup

cristiangreco:
#### PR Description
This is just a starting point, will extend it later.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/README.md
+++ b/internal/component/database_observability/README.md
@@ -339,3 +339,23 @@ prometheus.scrape "database_observability_mysql_example_db_2" {
   forward_to = [prometheus.remote_write.metrics_service.receiver]
 }
 ```
+
+## Setting up the Postgres database
+
+1. Your Postgres DB should be version 16.
+
+2. Create a dedicated DB user and grant permissions.
+
+```sql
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/README.md
```diff
@@ -339,3 +339,23 @@ prometheus.scrape "database_observability_mysql_example_db_2" {
   forward_to = [prometheus.remote_write.metrics_service.receiver]
 }
 ```
+
+## Setting up the Postgres database
+
+1. Your Postgres DB should be version 16.
```
fridgepoet, line 345: is it greater than or equal to version 16?

cristiangreco: Yeah I think we can claim support for 17 as well but we're working on updating the exporter first.

matthewnolf: We've not tested on 14 or 15 so this is fine for now. But agree we'll want to revise this in future

matthewnolf (approved):

PR #3993, opened 2025-07-15
database_observability: fix scan type for nullable digests and digest texts

fridgepoet:
#### PR Description

This PR fixes the type that waiting/block Digest and Digest Text scan to as they are nullable fields.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/locks.go
+++ b/internal/component/database_observability/mysql/collector/locks.go
@@ -144,7 +144,7 @@ func (c *LockCollector) fetchLocks(ctx context.Context) error {
 
 	for rsdl.Next() {
 		var waitingTimerWait, waitingLockTime, blockingTimerWait, blockingLockTime float64
-		var waitingDigest, waitingDigestText, blockingDigest, blockingDigestText string
+		var waitingDigest, waitingDigestText, blockingDigest, blockingDigestText sql.NullString
 
 		err := rsdl.Scan(&waitingTimerWait, &waitingLockTime, &waitingDigest, &waitingDigestText,
 			&blockingTimerWait, &blockingLockTime, &blockingDigest, &blockingDigestText)
@@ -157,10 +157,10 @@ func (c *LockCollector) fetchLocks(ctx context.Context) error {
 		if waitingLockTime > secondsToPicoseconds(c.lockTimeThreshold.Seconds()) {
 			lockMsg := fmt.Sprintf(
 				`waiting_digest="%s" waiting_digest_text="%s" blocking_digest="%s" blocking_digest_text="%s" waiting_timer_wait="%fms" waiting_lock_time="%fms" blocking_timer_wait="%fms" blocking_lock_time="%fms"`,
-				waitingDigest,
-				waitingDigestText,
-				blockingDigest,
-				blockingDigestText,
+				waitingDigest.String,
+				waitingDigestText.String,
+				blockingDigest.String,
+				blockingDigestText.String,
 				picosecondsToMilliseconds(waitingTimerWait),
 				picosecondsToMilliseconds(waitingLockTime),
 				picosecondsToMilliseconds(blockingTimerWait),
--- a/internal/component/database_observability/mysql/collector/locks_test.go
+++ b/internal/component/database_observability/mysql/collector/locks_test.go
@@ -188,6 +188,62 @@ func Test_QueryLocks(t *testing.T) {
 		assert.Equal(t, `level="info" waiting_digest="xyz789" waiting_digest_text="SELECT * FROM orders WHERE user_id = ?" blocking_digest="ghi012" blocking_digest_text="DELETE FROM sessions WHERE expired = ?" waiting_timer_wait="2500.000000ms" waiting_lock_time="2000.000000ms" blocking_timer_wait="3000.000000ms" blocking_lock_time="2700.000000ms"`, lokiEntries[1].Line)
 	})
 
+	t.Run("data lock with null digests and digest texts", func(t *testing.T) {
+		db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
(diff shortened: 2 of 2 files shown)


internal/component/database_observability/mysql/collector/locks.go
```diff
@@ -144,7 +144,7 @@ func (c *LockCollector) fetchLocks(ctx context.Context) error {
 
 	for rsdl.Next() {
 		var waitingTimerWait, waitingLockTime, blockingTimerWait, blockingLockTime float64
-		var waitingDigest, waitingDigestText, blockingDigest, blockingDigestText string
+		var waitingDigest, waitingDigestText, blockingDigest, blockingDigestText sql.NullString
```
cristiangreco, line 147: Hmm if these are null we can't really do much with the data, right? Do you know when it happens? Are all 4 columns null at the same time or do we get at least info about one of the two sides (waiting vs blocking)?

fridgepoet: So in our test DB, I captured a null waiting_digest while the blocking_digest was **not** null.
The blocking_digest was an UPDATE operation taking >5s. What I think happened is that the waiting statement (likely a much faster SELECT or simple operation) completed and was removed from the events_statements_current table before our locks collector ran, while the blocking statement was still executing due to its >5s duration.

Even though that is an artificial situation, I think that's pretty interesting information. What do you think?

gaantunes: The mysql documentation talks about that [here](https://dev.mysql.com/doc/refman/8.4/en/innodb-information-schema-examples.html), specifically in Identifying a Blocking Query After the Issuing Session Becomes Idle. It says we shoud use `events_statements_history` in this scenario. Maybe we could add a coalesce to the query that fetches it from history when, and only when, current is null.

cristiangreco: Great work! 

Stepping back for a moment, at this stage we're interested in showing in the UI the information about the duration, count and list of statements waiting on a blocker. I think it's acceptable that the waiter is null (i.e. it's just a manifestation of the scenario you described where we couldn't capture who is the waiter but we know there's been one). 
So I'm leaning towards modifying the query so that `blockingDigest != null`. This should fix the error while at the same time providing somewhat useful information, right?

Gabriel's suggestion to fallback to `_history` could be nice to explore, but I'd ticket that as separate work. wdyt?

fridgepoet: Thank you both for having a look and your input!

> Gabriel's suggestion to fallback to _history could be nice to explore, but I'd ticket that as separate work.

Agreed, let's reach a resolution here and then we can work on `_history` improvements. [Ticket created](https://github.com/grafana/grafana-dbo11y-app/issues/1359).

>So I'm leaning towards modifying the query so that blockingDigest != null.

So in the meantime I have also found the reverse case (more rare in our test scenarios): blocking_digest is null and the waiting_digest is present.

It's a similar scenario to what we've discussed above: a blocking statement might cause a lock, complete and be cleared, while a "long-running" waiting statement is still executing _while_ our collector runs.

Although always presenting blocking_digest _is_ useful information, I think the reverse is also true. For example, what if we always see that certain waiting statements are repeatedly getting locked? 

Knowing about blocking_digest provides insight into what's causing a problem (good candidates for optimization).
Knowing about waiting_digest provides insight into what's being affected (impact on the data retrieval). 

For the moment, I would soft vote to take the conservative approach and keep both sides of the story until we learn a little more about how we can interpret this data. What do you think?
At most, I think we can add some additional debug info (thread IDs? event IDs?) to confirm the sequence of events.

(Another challenge, however, is how we present this type of scenario to the DB O11y users so that it is understandable.)

cristiangreco: > For the moment, I would soft vote to take the conservative approach and keep both sides of the story until we learn a little more about how we can interpret this data. What do you think?

Hmm I'm fine with that. So we're saying we're going to capture cases when blocker is null or waiter is null. But the case where both are null is useless I believe, we should just skip that?

fridgepoet: I'm not 100% sure it will be useless. But I was unable to see this kind of case.
If that's ok with you, let's try this for a while to see how it runs on some other databases.

cristiangreco: I don't intend to block the PR, but I'm struggling to follow how having all of the following null is meaningful?
```
waitingDigest,
waitingDigestText,
blockingDigest,
blockingDigestText
```
Am I missing or misinterpreting something here? AFAICS the output would tell us that there's some lock time but we cannot really account it to anything.

fridgepoet: I am not entirely sure in my previous explanations and during what I think can be a learning phase, I think it might be better not to eliminate this case entirely. 

If there is a timing pattern related to locks (but for some reason waiting+blocking are null), this could be interesting to raise to the user (a graphic visualization). 

At the very least, if we observe many waiting+blocking nulls, we'll know there is a problem with our monitoring coverage.

cristiangreco: >At the very least, if we observe many waiting+blocking nulls, we'll know there is a problem with our monitoring coverage.

I find it difficult to understand how we can link the wait time to any other piece of information, but don't want to block this PR any further. Let's circle back when we have more data then please.

fridgepoet: (Thanks, all I meant with the above quote was that if we have too many waiting+blocking which are null, then maybe we need to revisit how we are collecting the lock events because we might be missing too many.)

I'll make a ticket to circle back to revisit this later: https://github.com/grafana/grafana-dbo11y-app/issues/1370

PR #3994, opened 2025-07-15
database_observability: Revert explain plans for 'with' queries

rgeyer:
#### PR Description

We need more rigor to insure that we're not running explain plans for queries which can write to (or delete from) a schema.


gaantunes (approved):
LGTM

PR #4006, opened 2025-07-17
Ensure cluster notifications are sent to remotecfg components

dehaansa:
#### PR Description
Components loaded through remotecfg that are clustering aware were not getting their cluster alert notifications.

While I was fixing this, I also fixed the only other use of GetAllComponents, used for statistics reporting.

I tested this by configuring the following remotecfg pipeline in GC Fleet Management and running two alloy instances locally, 

```
prometheus.scrape "alloy" {
	clustering {
		enabled = true
	}
	targets = [
		{
			__address__ = "localhost:12345",
			job         = "alloy1",
		},
		{
			__address__ = "localhost:12346",
			job         = "alloy2",
		},
	]

}

prometheus.remote_write "metrics_service" {
	endpoint {
		url = "<prom_endpoint>"

		basic_auth {
			username = "<user>"
			password = "<password>"
		}
	}
}
```

I used a local remotecfg block with the collector ID coming from an environment variable (`id = sys.env("FLEET_ID")`) and started the two nodes manually in terminal.

```
FLEET_ID=alloy1 ./build/alloy run --server.http.listen-addr=0.0.0.0:12345 --cluster.join-addresses=127.0.0.1:12346 --server.http.ui-path-prefix=/ --cluster.enabled=true --cluster.name=alloy-cluster --cluster.node-name=alloy1 ./local/cluster_test.alloy

FLEET_ID=alloy2 ./build/alloy run --server.http.listen-addr=0.0.0.0:12346 --cluster.join-addresses=127.0.0.1:12345 --server.http.ui-path-prefix=/ --cluster.enabled=true --cluster.name=alloy-cluster --cluster.node-name=alloy2 ./local/cluster_test.alloy
```

I confirmed that each node only had one target using the local UI `http://localhost:12345/remotecfg/component/remotecfg/cluster_failover_test.default/prometheus.scrape.alloy` and then stopped one instance, and confirmed that the still running instance got both targets. Before the change the targets list would not change based on peers joining or leaving the cluster.

#### Notes to the Reviewer

I'm open to any recommendations for adding unit tests here, but did not see a good way to create them offhand. We need to revisit unit tests for remotecfg.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -592,7 +592,7 @@ func (i *agentInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFun
 	return next
 }
 
-func GetRemoteCfgHost(host service.Host) (service.Host, error) {
+func GetHost(host service.Host) (service.Host, error) {
 	svc, found := host.GetService(ServiceName)
 	if !found {
 		return nil, fmt.Errorf("remote config service not available")
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -591,3 +591,16 @@ func (i *agentInterceptor) WrapStreamingClient(next connect.StreamingClientFunc)
 func (i *agentInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return next
 }
+
+func GetRemoteCfgHost(host service.Host) (service.Host, error) {
+	svc, found := host.GetService(ServiceName)
+	if !found {
+		return nil, fmt.Errorf("remote config service not available")
+	}
+
+	data := svc.Data().(Data)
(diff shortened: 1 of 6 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -591,3 +591,16 @@ func (i *agentInterceptor) WrapStreamingClient(next connect.StreamingClientFunc)
 func (i *agentInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
 	return next
 }
+
+func GetRemoteCfgHost(host service.Host) (service.Host, error) {
```
tpaschalis: nit: consider naming `GetHost` as the remotecfg part is implicit from the package.

tpaschalis (approved):
LGTM, good find!

PR #4011, opened 2025-07-18
Print extra debug information in case of inotify issues.

ptodev:
This is especially useful when debugging the "too many open files" error.

I compared the output with [another utility](https://github.com/mikesart/inotify-info/tree/master):
Screenshot 2025-07-18 at 15 12 53

The config I used to reproduce the issue is:

```hcl
discovery.file "example" {
  files = ["/home/ubuntu/alloy_test/file_targets.yaml"]
}

foreach "file" {
  collection = discovery.file.example.targets
  var        = "each"
  template {
    local.file "testing" {
      filename = each["__address__"]
    }
  }
}
```

The contents of `file_targets.yaml` contain several thousand lines referencing the same targets file:
```yaml
- targets:
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  - /home/ubuntu/alloy_test/test_target_1.txt
  ...
```

`test_target_1.txt` is just an empty file on the filesystem.

### DIFF
--- a/internal/filedetector/internal/inotifyinfo/inotifyinfo.go
+++ b/internal/filedetector/internal/inotifyinfo/inotifyinfo.go
@@ -29,16 +29,16 @@ func DiagnosticsJson(logger log.Logger) string {
 	return diagnostics(logger).json(logger)
 }
 
-func diagnostics(logger log.Logger) *inotifyInfo {
+func diagnostics(logger log.Logger) inotifyInfo {
 	fs, err := procfs.NewFS("/proc")
 	if err != nil {
 		level.Error(logger).Log("msg", "inotify diagnostics: failed to create procfs", "err", err)
-		return nil
+		return inotifyInfo{}
 	}
 	procs, err := fs.AllProcs()
 	if err != nil {
 		level.Error(logger).Log("msg", "inotify diagnostics: failed to get all procs", "err", err)
-		return nil
+		return inotifyInfo{}
 	}
 
 	res := inotifyInfo{
@@ -49,6 +49,7 @@ func diagnostics(logger log.Logger) *inotifyInfo {
 		fds, err := proc.FileDescriptorsInfo()
 		if err != nil {
 			level.Error(logger).Log("msg", "inotify diagnostics: failed to get file descriptors", "err", err)
+			continue
 		}
 
 		instances := 0
@@ -83,7 +84,7 @@ func diagnostics(logger log.Logger) *inotifyInfo {
 	res.MaxUserInstances = readInotifyLimits(logger, "max_user_instances")
 	res.MaxUserWatches = readInotifyLimits(logger, "max_user_watches")
 
-	return &res
+	return res
 }
 
 func readInotifyLimits(logger log.Logger, filename string) int {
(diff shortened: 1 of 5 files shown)


internal/filedetector/internal/inotifyinfo/inotifyinfo.go
```diff
@@ -0,0 +1,111 @@
...
+
+	res.MaxQueuedEvents = readInotifyLimits(logger, "max_queued_events")
+	res.MaxUserInstances = readInotifyLimits(logger, "max_user_instances")
+	res.MaxUserWatches = readInotifyLimits(logger, "max_user_watches")
+
+	return &res
+}
+
+func readInotifyLimits(logger log.Logger, filename string) int {
+	filename = "/proc/sys/fs/inotify/" + filename
```
thampiotr: Is that path guaranteed to exist?

internal/filedetector/internal/inotifyinfo/inotifyinfo.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"encoding/json"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/prometheus/procfs"
```
thampiotr: Nice, didn't know about this!

internal/filedetector/internal/inotifyinfo/inotifyinfo.go
```diff
@@ -0,0 +1,111 @@
...
+
+type procInotifyInfo struct {
+	PID              int    `json:"pid"`
+	ProcessName      string `json:"process_name"`
+	InotifyWatchers  int    `json:"inotify_watchers"`
+	InotifyInstances int    `json:"inotify_instances"`
+}
+
+func DiagnosticsJson(logger log.Logger) string {
+	return diagnostics(logger).json(logger)
```
thampiotr: This is destined to panic in case of error in `diagnostics`.

thampiotr (commented):
We need to fix that panic possibility. 
And probably a test would be a good idea, since there is a scope for panics / issues.

internal/filedetector/internal/inotifyinfo/inotifyinfo.go
ptodev: Good question. I think it won't exist in old Linux kernels. I'll add a check!

internal/filedetector/internal/inotifyinfo/inotifyinfo.go
ptodev: Actually it doesn't panic if diagnostics returns a nil. But I'll add some tests, I agree it's important to make it clear.

ptodev: I didn't add tests but I refactored the code to not use pointers.

internal/filedetector/internal/inotifyinfo/inotifyinfo.go
ptodev: I think there' s no need for a check actually. There's currently an error message and the number will be set to -1 if the file doesn't exist.

PR #4014, opened 2025-07-18
database_observability: Add backoff and denylisting for explain plan errors

rgeyer:
#### PR Description
This adds a backoff mechanism for explain queries which are experiencing errors.

When an error occurs while trying to execute an explain plan, that query is added to a backoff list. That query will wait `explain_plan_failure_backoff_rounds` (default 5) scrape intervals (rounds) before trying again.

If `explain_plan_max_failure_count` (default 2) failures occur, the query is deny listed for the duration of the lifespan of the current alloy process.

While queries are on backoff, other queries can be fetched from the performance_schema, and have explain plans executed for them.

#### PR Checklist
- [ ] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -36,6 +36,8 @@ You can use the following arguments with `database_observability.mysql`:
 | `explain_plan_collect_interval`    | `duration`           | How frequently to collect explain plan information from database.                              | `"1m"`  | no       |
 | `explain_plan_per_collect_ratio`   | `float`              | Ratio of explain plan queries to collect per collect interval.                                 | `1.0`   | no       |
 | `explain_plan_initial_lookback`    | `duration`           | How far back to look for explain plan queries on the first collection interval.                | `"24h"` | no       |
+| `explain_plan_failure_backoff_rounds` | `int`              | Number of rounds to skip after a query fails.                                                 | `5`     | no       |
+| `explain_plan_max_failure_count`   | `int`              | Number of failures before a query is denylisted.                                              | `2`     | no       |
 | `locks_collect_interval`           | `duration`           | How frequently to collect locks information from database.                                     | `"30s"` | no       |
 | `locks_threshold`                  | `duration`           | Threshold for locks to be considered slow. If a lock exceeds this duration, it will be logged. | `"1s"`  | no       |
 | `setup_consumers_collect_interval` | `duration`           | How frequently to collect performance_schema.setup_consumers information from the database.    | `"1h"`  | no       |
--- a/internal/component/database_observability/mysql/collector/explain_plan.go
+++ b/internal/component/database_observability/mysql/collector/explain_plan.go
@@ -451,7 +451,6 @@ type queryInfo struct {
 	schemaName             string
 	digest                 string
 	queryText              string
-	failureBackoff         bool
 	failureCount           int
 	roundsSinceLastFailure int
 }
@@ -479,6 +478,7 @@ type ExplainPlan struct {
 	dbVersion            string
 	scrapeInterval       time.Duration
 	queryCache           map[string]*queryInfo
+	queryBackoffList     map[string]*queryInfo
 	queryDenylist        map[string]*queryInfo
 	perScrapeRatio       float64
 	currentBatchSize     int
@@ -510,6 +510,7 @@ func NewExplainPlan(args ExplainPlanArguments) (*ExplainPlan, error) {
 		scrapeInterval:       args.ScrapeInterval,
 		queryCache:           make(map[string]*queryInfo),
 		queryDenylist:        make(map[string]*queryInfo),
+		queryBackoffList:     make(map[string]*queryInfo),
 		perScrapeRatio:       args.PerScrapeRatio,
 		failureBackoffRounds: args.FailureBackoffRounds,
 		maxFailureCount:      args.MaxFailureCount,
@@ -573,6 +574,15 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 	}
 	defer rs.Close()
 
+	// Add queries with expired backoff first
+	for _, qi := range c.queryBackoffList {
+		if qi.roundsSinceLastFailure >= c.failureBackoffRounds {
+			qi.roundsSinceLastFailure = 0
+			c.queryCache[qi.key()] = qi
+			delete(c.queryBackoffList, qi.key())
+		}
+	}
+
 	// Populate cache
 	for rs.Next() {
 		if err := rs.Err(); err != nil {
@@ -581,7 +591,6 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 		}
 
 		qi := &queryInfo{
-			failureBackoff:         false,
 			failureCount:           0,
 			roundsSinceLastFailure: 0,
 		}
@@ -610,35 +619,29 @@ func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
 		}
 	}
 
+	for _, bqi := range c.queryBackoffList {
+		bqi.roundsSinceLastFailure++
+	}
+
 	processedCount := 0
 	for _, qi := range c.queryCache {
 		failed := false
 		if processedCount >= c.currentBatchSize {
 			break
 		}
-		if qi.failureBackoff {
-			if qi.roundsSinceLastFailure >= c.failureBackoffRounds {
-				qi.failureBackoff = false
-				qi.roundsSinceLastFailure = 0
-			} else {
-				qi.roundsSinceLastFailure++
-				continue
-			}
-		}
 		logger := log.With(c.logger, "digest", qi.digest)
 
 		defer func(failed *bool) {
 			if *failed {
 				qi.failureCount++
-				qi.failureBackoff = true
 				if qi.failureCount >= c.maxFailureCount {
-					delete(c.queryCache, qi.key())
 					c.queryDenylist[qi.key()] = qi
 					level.Info(c.logger).Log("msg", "query denylisted", "digest", qi.digest)
+				} else {
+					c.queryBackoffList[qi.key()] = qi
 				}
-			} else {
-				delete(c.queryCache, qi.key())
 			}
+			delete(c.queryCache, qi.key())
 			processedCount++
 		}(&failed)
 
--- a/internal/component/database_observability/mysql/collector/explain_plan.go
+++ b/internal/component/database_observability/mysql/collector/explain_plan.go
@@ -3,12 +3,12 @@ package collector
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"database/sql"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"math"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -448,37 +448,48 @@ func parseUnionResultNode(logger log.Logger, unionResultNode []byte) (planNode, 
 }
 
 type queryInfo struct {
-	schemaName string
-	digest     string
-	queryText  string
+	schemaName             string
+	digest                 string
+	queryText              string
+	failureBackoff         bool
+	failureCount           int
+	roundsSinceLastFailure int
+}
+
+func (qi *queryInfo) key() string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(qi.schemaName+"|"+qi.digest)))
 }
 
 type ExplainPlanArguments struct {
-	DB              *sql.DB
-	InstanceKey     string
-	ScrapeInterval  time.Duration
-	PerScrapeRatio  float64
-	EntryHandler    loki.EntryHandler
-	InitialLookback time.Time
+	DB                   *sql.DB
+	InstanceKey          string
+	ScrapeInterval       time.Duration
+	PerScrapeRatio       float64
+	EntryHandler         loki.EntryHandler
+	InitialLookback      time.Time
+	FailureBackoffRounds int
+	MaxFailureCount      int
 
 	Logger log.Logger
 }
 
 type ExplainPlan struct {
-	dbConnection     *sql.DB
-	instanceKey      string
-	dbVersion        string
-	scrapeInterval   time.Duration
-	queryCache       []queryInfo
-	perScrapeRatio   float64
-	currentBatchSize int
-	entryHandler     loki.EntryHandler
-	lastSeen         time.Time
-
-	logger  log.Logger
-	running *atomic.Bool
-	ctx     context.Context
-	cancel  context.CancelFunc
+	dbConnection         *sql.DB
+	instanceKey          string
+	dbVersion            string
+	scrapeInterval       time.Duration
+	queryCache           map[string]*queryInfo
+	queryDenylist        map[string]*queryInfo
+	perScrapeRatio       float64
+	currentBatchSize     int
+	entryHandler         loki.EntryHandler
+	lastSeen             time.Time
+	failureBackoffRounds int
+	maxFailureCount      int
+	logger               log.Logger
+	running              *atomic.Bool
+	ctx                  context.Context
+	cancel               context.CancelFunc
 }
 
 func NewExplainPlan(args ExplainPlanArguments) (*ExplainPlan, error) {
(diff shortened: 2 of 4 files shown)


fridgepoet:
The type of errors the explain plans experience, it's something that benefits from retries? Just wondering if it fails the first time if it will almost always fail afterwards or not.

cristiangreco (commented):
This looks functionally good. I'm interested in @fridgepoet 's question. I'd _think_ the specific sample we explain might (very rarely?) influence the failure/success, but I'm curious to hear if Ryan has other ideas.

matthewnolf:
> The type of errors the explain plans experience, it's something that benefits from retries? Just wondering if it fails the first time if it will almost always fail afterwards or not.

> This looks functionally good. I'm interested in @fridgepoet 's question. I'd think the specific sample we explain might (very rarely?) influence the failure/success, but I'm curious to hear if Ryan has other ideas.

I'm also curious here. I'd imagine retrying the same query would result in the same error. However a different query text may give a different result. In that case, we should only retry the explain when the query_text has changed. The number of rounds before retying seems less effective, as you're having to hope that the query_text has changed in that time without guaranteeing it.
Can we cache the query_text (or it's de-normalised hash), and only retry at that point?

rgeyer:
> The type of errors the explain plans experience, it's something that benefits from retries? Just wondering if it fails the first time if it will almost always fail afterwards or not.

> This looks functionally good. I'm interested in @fridgepoet 's question. I'd think the specific sample we explain might (very rarely?) influence the failure/success, but I'm curious to hear if Ryan has other ideas.

Right, fair point(s). The only error(s) that would occur would either be DB issues, or failure to parse the JSON that is returned. Given that we're creating a new DB connection, so that we can change the current schema with a `USE` statement, there's a very small chance that you could hit a recoverable error like too many connections.

Most everything else tho, would probably be non-recoverable, and wouldn't benefit from retries. For example, the reason for denylisting queries arose from explain queries that were having permissions errors. That won't resolve itself without intervention.

> Can we cache the query_text (or it's de-normalised hash), and only retry at that point?

It looks like the query_sample_text can change for a digest over time. Either because another version of the digest query has a longer wait time, or because the previous sample is "too old". We can certainly change the logic to check for this, tho I wonder if it would realistically get retried in a reasonable amount of time?

Especially in a case where the error(s) are non recoverable in nature anyway (permissions issue, etc) we could end up with a huge backlog of cached queries waiting for a chance to retry (when the sample changes), which would have to be checked every round, or at least every time we check for new queries. 🤔 

I guess the reason for the wait/retry was in an attempt to provide a softer mechanism than simply denylisting a digest after a single failure, tho perhaps that's the best approach. 🤔

matthewnolf:
> > The type of errors the explain plans experience, it's something that benefits from retries? Just wondering if it fails the first time if it will almost always fail afterwards or not.
> 
> > This looks functionally good. I'm interested in @fridgepoet 's question. I'd think the specific sample we explain might (very rarely?) influence the failure/success, but I'm curious to hear if Ryan has other ideas.
> 
> Right, fair point(s). The only error(s) that would occur would either be DB issues, or failure to parse the JSON that is returned. Given that we're creating a new DB connection, so that we can change the current schema with a `USE` statement, there's a very small chance that you could hit a recoverable error like too many connections.
> 
> Most everything else tho, would probably be non-recoverable, and wouldn't benefit from retries. For example, the reason for denylisting queries arose from explain queries that were having permissions errors. That won't resolve itself without intervention.
> 
> > Can we cache the query_text (or it's de-normalised hash), and only retry at that point?
> 
> It looks like the query_sample_text can change for a digest over time. Either because another version of the digest query has a longer wait time, or because the previous sample is "too old". We can certainly change the logic to check for this, tho I wonder if it would realistically get retried in a reasonable amount of time?
> 
> Especially in a case where the error(s) are non recoverable in nature anyway (permissions issue, etc) we could end up with a huge backlog of cached queries waiting for a chance to retry (when the sample changes), which would have to be checked every round, or at least every time we check for new queries. 🤔
> 
> I guess the reason for the wait/retry was in an attempt to provide a softer mechanism than simply denylisting a digest after a single failure, tho perhaps that's the best approach. 🤔

Thanks for the reply. I'm not clear on what errors we're trying to solve, and which not. Could you explain that? My understanding so far is:

1. Transient errors such as the network
2. Semantic query errors with the specific query text
3. Setup ('non recoverable') errors



For 1. would resolve itself, and also effect any query indiscriminately so I'm not sure we need such error handling for those cases other than retrying on the next run.
For 2. This would maybe be resolved by trying a different query_text as it is updated by the database. It seems unlikely to me that one query_text would work but another not under the same digest however happy to try in that situation.
For 3. Which setup errors would we run into? The DB user alloy is using should have permission to run the explain - if auth is failing, that's a configuration issue that we wouldn't want to suppress. Are there others?

rgeyer:
> Thanks for the reply. I'm not clear on what errors we're trying to solve, and which not. Could you explain that? My understanding so far is:
> 
> 1. Transient errors such as the network
> 2. Semantic query errors with the specific query text
> 3. Setup ('non recoverable') errors
> 
> For 1. would resolve itself, and also effect any query indiscriminately so I'm not sure we need such error handling for those cases other than retrying on the next run. For 2. This would maybe be resolved by trying a different query_text as it is updated by the database. It seems unlikely to me that one query_text would work but another not under the same digest however happy to try in that situation. For 3. Which setup errors would we run into? The DB user alloy is using should have permission to run the explain - if auth is failing, that's a configuration issue that we wouldn't want to suppress. Are there others?

1. wouldn't necessarily resolve itself, in as much as, without the retry logic that's been added in this PR, a rarely executed query could appear in the performance schema, experience an error while running an explain plan, and never get executed again. This is because (again, without this PR) a given digest is only tried once, then removed from the query cache whether it succeeds or fails.

2. wasn't really a consideration when initially building this out, tho it's a reasonable scenario to address. It would require different logic as we've discussed.

Primarily, the reason for building this out in the first place was due to 3, but a slight variation on that. It's not necessarily a case where the service user has *no* permissions and is thus a hard failure of config, but rather a scenario where the service user has permissions, but does not have `SELECT` permissions on the schema being operated on.

I.E. You can make select queries against `performance_schema` and `schema1` but not `schema2`. When the collector attempts to run explain queries against `schema2`, you would encounter permission errors. This could be an intentional configuration. Maybe a config option to exclude certain schemas would be a better option here.

matthewnolf:
> Primarily, the reason for building this out in the first place was due to 3, but a slight variation on that. It's not necessarily a case where the service user has _no_ permissions and is thus a hard failure of config, but rather a scenario where the service user has permissions, but does not have `SELECT` permissions on the schema being operated on.
> 
> I.E. You can make select queries against `performance_schema` and `schema1` but not `schema2`. When the collector attempts to run explain queries against `schema2`, you would encounter permission errors. This could be an intentional configuration. Maybe a config option to exclude certain schemas would be a better option here.

Yeah I like the idea of a schemas list which we run the explains on, and ignore the rest (with a default of everything). And I think if we go that route, 1. is an acceptable edge case for now

cristiangreco:
Hey wanted to check what is the state of this change? I thought we wanted to remove the backoff policy and resort to an error log or an error caching strategy?

rgeyer:
> Hey wanted to check what is the state of this change? I thought we wanted to remove the backoff policy and resort to an error log or an error caching strategy?

I don't know that we came to an official decision here.

I think the unofficial consensus is that we should denylist unrecoverable errors (no permissions for schema, unable to parse the output), and simply "pass" others after logging an error, allowing the same query to be retried the next time it appears in the `performance_schema.events_statements_summary_by_digest`.

Ohh, and perhaps a config param for "schema_[deny|allow]list" so that a deliberate configuration with a service account that can only access a few schema(s) can be accommodated without having to denylist queries after failure.

I can go ahead and implement this and update the PR.

cristiangreco:
That sounds good to me, let's go ahead with that

gaantunes:
I don't want to show up late and add another layer of complexity, but if a query is failing due to permission lacking and it was supposed to work, will it be ever retried? Say someone is configuring the collector for the first time, and get their permissions wrong on the db side. They realize that, and set the permissions afterwards. Will the collector pick it up at some point? Or would they need to restart it?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -562,19 +581,27 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
...
+		qi := &queryInfo{failureCount: 0}
 		var ls time.Time
 		if err = rs.Scan(&qi.schemaName, &qi.digest, &qi.queryText, &ls); err != nil {
 			level.Error(c.logger).Log("msg", "failed to scan digest for explain plans", "err", err)
 			return err
 		}
-		c.queryCache = append(c.queryCache, qi)
+		for _, schema := range c.schemaDenyList {
+			if strings.EqualFold(qi.schemaName, schema) {
+				continue SCHEMA_DENYLIST
```
matthewnolf: Can we find an approach that doesn't use goto, as they make the control flow harder to follow and as such we generally avoid their use

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -36,6 +36,7 @@ You can use the following arguments with `database_observability.mysql`:
 | `explain_plan_collect_interval`    | `duration`           | How frequently to collect explain plan information from database.                              | `"1m"`  | no       |
 | `explain_plan_per_collect_ratio`   | `float`              | Ratio of explain plan queries to collect per collect interval.                                 | `1.0`   | no       |
 | `explain_plan_initial_lookback`    | `duration`           | How far back to look for explain plan queries on the first collection interval.                | `"24h"` | no       |
+| `explain_plan_schema_denylist`     | `list(string)`       | List of schemas to denylist from explain plan collection.                                      |         | no       |
```
matthewnolf: nit: `exclude-schemas` inspired by the PG exporter config. I think it sounds more self explanatory, thoughts?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -450,37 +450,51 @@ func parseUnionResultNode(logger log.Logger, unionResultNode []byte) (planNode,
...
-	ctx     context.Context
-	cancel  context.CancelFunc
+	dbConnection          *sql.DB
+	instanceKey           string
+	dbVersion             string
+	scrapeInterval        time.Duration
+	queryCache            map[string]*queryInfo
+	queryDenylist         map[string]*queryInfo
+	schemaDenyList        []string
+	unrecoverableSQLCodes []knownSQLCodes
```
matthewnolf: Do these need to be attached to the struct? Can't they just be package variables?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -450,37 +450,51 @@ func parseUnionResultNode(logger log.Logger, unionResultNode []byte) (planNode,
...
+	digest       string
+	queryText    string
+	failureCount int
 }
 
+func (qi *queryInfo) key() string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(qi.schemaName+"|"+qi.digest)))
+}
+
+type knownSQLCodes string
```
matthewnolf, line 467: nit: I'm unsure if the type alias adds anything here?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -562,19 +581,27 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
...
 
-		var qi queryInfo
+		qi := &queryInfo{failureCount: 0}
 		var ls time.Time
 		if err = rs.Scan(&qi.schemaName, &qi.digest, &qi.queryText, &ls); err != nil {
 			level.Error(c.logger).Log("msg", "failed to scan digest for explain plans", "err", err)
 			return err
 		}
-		c.queryCache = append(c.queryCache, qi)
+		for _, schema := range c.schemaDenyList {
```
matthewnolf: I think you can use something like https://pkg.go.dev/slices#Contains

rgeyer:
> I don't want to show up late and add another layer of complexity, but if a query is failing due to permission lacking and it was supposed to work, will it be ever retried? Say someone is configuring the collector for the first time, and get their permissions wrong on the db side. They realize that, and set the permissions afterwards. Will the collector pick it up at some point? Or would they need to restart it?

Not a problem at all, that's what reviews are for.

The way this is designed, an unrecoverable error (permissions in this case) is immediately denylisted. Alloy would have to be restarted to clear that. I suppose technically if you made a configuration change to alloy, the collector would get restarted which would cause the same result.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: There's no type safety or other utility. It was added mostly for readability/clarity.

While I went through the [MySQL error code list](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html) and nothing else obviously stood out to me, I suspect we may add a few additional "known" codes that we'd want to exclude.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Right, of course 🤦‍♂️

gaantunes:
>Not a problem at all, that's what reviews are for.
>
>The way this is designed, an unrecoverable error (permissions in this case) is immediately denylisted. Alloy would have to >be restarted to clear that. I suppose technically if you made a configuration change to alloy, the collector would get >restarted which would cause the same result.

Ok, no problem with this, as long as we make it very clear on the docs. otherwise customers doing this first configuration might give a missing permission on the db and still not get data and not understand why.

docs/sources/reference/components/database_observability/database_observability.mysql.md
rgeyer: Yeah, I like this, made the change. Thanks!

clayton-cornell:
Doc change is small. No input from me here.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Right, fair point. Moved this up to the package.

internal/component/database_observability/mysql/collector/explain_plan.go
rgeyer: Unless there's a strong objection, I'm going to leave this as-is. Tho I did move the unrecoverable codes out of the struct and into the package.

clayton-cornell (commented):
Doc change is OK

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -448,16 +449,32 @@ func parseUnionResultNode(logger log.Logger, unionResultNode []byte) (planNode,
...
-	schemaName *string
-	digest     string
-	queryText  string
+	schemaName   string
+	digest       string
+	queryText    string
+	failureCount int
+}
+
+func (qi *queryInfo) key() string {
```
cristiangreco: nit, could this be named something like `cacheKey()` for clarity?

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -557,13 +577,21 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 			return err
 		}
 
-		var qi queryInfo
+		qi := &queryInfo{failureCount: 0}
```
cristiangreco: another nit: no need to init the counter to zero

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -448,16 +449,32 @@ func parseUnionResultNode(logger log.Logger, unionResultNode []byte) (planNode,
...
-	digest     string
-	queryText  string
+	schemaName   string
+	digest       string
+	queryText    string
+	failureCount int
+}
+
+func (qi *queryInfo) key() string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(qi.schemaName+"|"+qi.digest)))
```
cristiangreco: This method is called multiple times in subsequent code, maybe we could save the cache key in `queryInfo` when it's computed for the first time.

Generally speaking, `sprintf` and concatenations cause a bunch of small allocations that might unnecessarily keep the GC busy. I'm wondering if we need to compute the md5 hash at all... could we just use `qi.schemaName+qi.digest` as cache key? (possibly don't even need the `|` separator, as this is not a piece of debugging information we expose somewhere)

cristiangreco (approved):

PR #4021, opened 2025-07-22
database_observability: fix postgres user setup commands

cristiangreco:
#### PR Description
Fix commands syntax.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/README.md
+++ b/internal/component/database_observability/README.md
@@ -351,9 +351,9 @@ prometheus.scrape "database_observability_mysql_example_db_2" {
 2. Create a dedicated DB user and grant permissions.
 
 ```sql
-CREATE USER 'db-o11y'@'%' IDENTIFIED by '<password>';
-GRANT pg_monitor TO 'db-o11y';
-GRANT pg_read_all_stats TO 'db-o11y';
+CREATE USER "db-o11y" WITH PASSWORD '<password>';
+GRANT pg_monitor TO "db-o11y";
+GRANT pg_read_all_stats TO "db-o11y";
 ```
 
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/README.md
```diff
@@ -348,21 +348,23 @@ prometheus.scrape "database_observability_mysql_example_db_2" {
...
-2. Create a dedicated DB user and grant permissions.
+2. Add the `pg_stat_statements` module to `shared_preload_libraries` in `postgresql.conf`. This requires a restart of the Postgres DB.
+
+3. Create the `pg_stat_statements` extension in every database you want to monitor.
 
 ```sql
-CREATE USER 'db-o11y'@'%' IDENTIFIED by '<password>';
-GRANT pg_monitor TO 'db-o11y';
-GRANT pg_read_all_stats TO 'db-o11y';
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements;
```
matthewnolf, line 356: I think it's worth a step to check it is enabled:
Something like
```
Check the extension is enabled: 
SELECT * 
FROM pg_extension WHERE extname='pg_stat_statements'
```

cristiangreco: It's kinda implicit in the last step, where we ask to verify

```sql
SELECT * FROM pg_stat_statements LIMIT 1;
```

Do you think it's better to have an explicit check on the extension?

matthewnolf: > Do you think it's better to have an explicit check on the extension?

I do, because that step is actually verifying the user is setup correctly and don't indicate the issue could be related to this step enabling the extension

PR #4034, opened 2025-07-22
move metrics and agentInterceptor out of remotecfg.go

erikbaranowski:
#### PR Description

Organize the remotecfg.go for maintainability. Increase unit test coverage which was missing prior due to too much in one file.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/service/remotecfg/metrics.go
+++ b/internal/service/remotecfg/metrics.go
@@ -1,0 +1,65 @@
+package remotecfg
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+type metrics struct {
+	lastLoadSuccess      prometheus.Gauge
+	lastFetchNotModified prometheus.Gauge
+	totalFailures        prometheus.Counter
+	configHash           *prometheus.GaugeVec
+	lastFetchSuccessTime prometheus.Gauge
+	totalAttempts        prometheus.Counter
+	getConfigTime        prometheus.Histogram
+}
+
+func registerMetrics(reg prometheus.Registerer) *metrics {
+	prom := promauto.With(reg)
+	return &metrics{
+		configHash: prom.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "remotecfg_hash",
+				Help: "Hash of the currently active remote configuration.",
+			},
+			[]string{"hash"},
+		),
+		lastLoadSuccess: prom.NewGauge(
+			prometheus.GaugeOpts{
+				Name: "remotecfg_last_load_successful",
+				Help: "Remote config loaded successfully",
+			},
+		),
+		lastFetchNotModified: prom.NewGauge(
+			prometheus.GaugeOpts{
+				Name: "remotecfg_last_load_not_modified",
+				Help: "Remote config not modified since last fetch",
+			},
+		),
+		totalFailures: prom.NewCounter(
+			prometheus.CounterOpts{
+				Name: "remotecfg_load_failures_total",
+				Help: "Remote configuration load failures",
+			},
+		),
+		totalAttempts: prom.NewCounter(
+			prometheus.CounterOpts{
+				Name: "remotecfg_load_attempts_total",
+				Help: "Attempts to load remote configuration",
+			},
+		),
+		lastFetchSuccessTime: prom.NewGauge(
+			prometheus.GaugeOpts{
+				Name: "remotecfg_last_load_success_timestamp_seconds",
+				Help: "Timestamp of the last successful remote configuration load",
+			},
+		),
+		getConfigTime: prom.NewHistogram(
+			prometheus.HistogramOpts{
+				Name: "remotecfg_request_duration_seconds",
+				Help: "Duration of remote configuration requests.",
+			},
+		),
+	}
+}
--- a/internal/service/remotecfg/remotecfg.go
+++ b/internal/service/remotecfg/remotecfg.go
@@ -31,7 +31,6 @@ import (
 	"github.com/grafana/alloy/syntax"
 	"github.com/grafana/alloy/syntax/ast"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	commonconfig "github.com/prometheus/common/config"
 )
 
@@ -82,16 +81,6 @@ type Service struct {
 	// This is the AST file parsed from the configuration. This is used
 	// for the support bundle
 	astFile *ast.File
-}
-
-type metrics struct {
-	lastLoadSuccess      prometheus.Gauge
-	lastFetchNotModified prometheus.Gauge
-	totalFailures        prometheus.Counter
-	configHash           *prometheus.GaugeVec
-	lastFetchSuccessTime prometheus.Gauge
-	totalAttempts        prometheus.Counter
-	getConfigTime        prometheus.Histogram
 }
 
 // ServiceName defines the name used for the remotecfg service.
@@ -199,56 +188,6 @@ func getSystemAttributes() map[string]string {
 	}
 }
 
-func (s *Service) registerMetrics() {
-	prom := promauto.With(s.opts.Metrics)
-	mets := &metrics{
-		configHash: prom.NewGaugeVec(
-			prometheus.GaugeOpts{
-				Name: "remotecfg_hash",
-				Help: "Hash of the currently active remote configuration.",
-			},
-			[]string{"hash"},
-		),
-		lastLoadSuccess: prom.NewGauge(
-			prometheus.GaugeOpts{
-				Name: "remotecfg_last_load_successful",
-				Help: "Remote config loaded successfully",
-			},
-		),
-		lastFetchNotModified: prom.NewGauge(
-			prometheus.GaugeOpts{
-				Name: "remotecfg_last_load_not_modified",
-				Help: "Remote config not modified since last fetch",
-			},
-		),
-		totalFailures: prom.NewCounter(
-			prometheus.CounterOpts{
-				Name: "remotecfg_load_failures_total",
-				Help: "Remote configuration load failures",
-			},
-		),
-		totalAttempts: prom.NewCounter(
-			prometheus.CounterOpts{
-				Name: "remotecfg_load_attempts_total",
-				Help: "Attempts to load remote configuration",
-			},
-		),
-		lastFetchSuccessTime: prom.NewGauge(
-			prometheus.GaugeOpts{
-				Name: "remotecfg_last_load_success_timestamp_seconds",
-				Help: "Timestamp of the last successful remote configuration load",
-			},
-		),
-		getConfigTime: prom.NewHistogram(
-			prometheus.HistogramOpts{
-				Name: "remotecfg_request_duration_seconds",
-				Help: "Duration of remote configuration requests.",
-			},
-		),
-	}
-	s.metrics = mets
-}
-
 // Data returns an instance of [Data]. Calls to Data are cachable by the
 // caller.
 // Data must only be called after Run.
@@ -291,6 +230,7 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 	// We need to use the mtx here as the cluster service will
 	// start in parallel and attempt to access the controller via Data()
 	s.mut.Lock()
+	s.metrics = registerMetrics(s.opts.Metrics)
 	s.ctrl = host.NewController(ServiceName)
 	s.mut.Unlock()
 
@@ -373,9 +313,6 @@ func (s *Service) Update(newConfig any) error {
 		return err
 	}
 
-	if s.metrics == nil {
-		s.registerMetrics()
-	}
 	s.mut.Unlock()
 
 	// If we've already called Run, then immediately trigger an API call with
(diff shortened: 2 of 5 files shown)


internal/service/remotecfg/remotecfg.go
```diff
@@ -288,39 +119,38 @@ var _ service.Service = (*Service)(nil)
...
 	// Run the service's own controller.
 	go func() {
-		s.ctrl.Run(ctx)
+		s.cm.getController().Run(ctx)
 	}()
 
 	for {
 		select {
-		case <-s.ticker.C:
-			err := s.fetchRemote()
```
erikbaranowski, line 318: I'm unclear why we only fetchRemote here. The new code handles fetching remote with the fallback to local included.

internal/service/remotecfg/api_client.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// newAPIClientWithClient creates a metrics-wrapped apiClient from an existing CollectorServiceClient.
+// This is primarily used for testing with mock clients.
+func newAPIClientWithClient(client collectorv1connect.CollectorServiceClient, metrics *metrics) *apiClient {
+	return &apiClient{
+		client:  client,
+		metrics: metrics,
+	}
+}
```
tpaschalis, line 45: small nit: can't we simply define a package-level mock client that tests will share? Is there an issue with reuse? Or a `newMockClient()`?

```
var mockClient = &apiClient{
  client: &mockCollectorClient{},
  metrics: registerMetrics(prometheus.NewRegistry())
}
```

internal/service/remotecfg/external.go
```diff
@@ -0,0 +1,32 @@
+// This file contains functionality that is used by custom code external to the
+// remotecfg package. These functions provide public APIs for accessing remotecfg
+// service data and are intended for use by other packages and components.
+package remotecfg
```
tpaschalis: nit: I'm on the fence, but should we pull `Data()` here as well?

Also, let's make it consistent `remotecfg` on the logs as well.

internal/service/remotecfg/remotecfg.go
dehaansa: Because cached local config is checked on service startup, there's no need to check for cached config every time you query the API because if it's changed from what it was on startup it only changed due to a successful fetchRemote?

internal/service/remotecfg/external.go
erikbaranowski: I wasn't a big fan of this file for a final version so I moved code back to remotecfg.

internal/service/remotecfg/api_client.go
erikbaranowski: I created newMockAPIClient

internal/service/remotecfg/remotecfg.go
erikbaranowski: ya that's right. I've added a boolean to control this behavior for performance reasons

internal/service/remotecfg/api_client_test.go
```diff
@@ -0,0 +1,299 @@
...
+// mockCollectorClient is a mock implementation of CollectorServiceClient for testing
+type mockCollectorClient struct {
+	mut                     sync.RWMutex
+	getConfigFunc           func(context.Context, *connect.Request[collectorv1.GetConfigRequest]) (*connect.Response[collectorv1.GetConfigResponse], error)
+	registerCollectorFunc   func(context.Context, *connect.Request[collectorv1.RegisterCollectorRequest]) (*connect.Response[collectorv1.RegisterCollectorResponse], error)
+	unregisterCollectorFunc func(context.Context, *connect.Request[collectorv1.UnregisterCollectorRequest]) (*connect.Response[collectorv1.UnregisterCollectorResponse], error)
+	getConfigCallCount      atomic.Int32
+	registerCallCount       atomic.Int32
+	unregisterCallCount     atomic.Int32
+}
```
tpaschalis, line 30: This looks pretty similar to what we have at the very bottom of remotecfg_test.go with `collectorClient`.

Is the idea to delete those once we merge these new tests?

internal/service/remotecfg/api_client_test.go
```diff
@@ -0,0 +1,299 @@
...
+		URL:              "https://example.com/api",
+		PollFrequency:    30 * time.Second,
+		HTTPClientConfig: config.CloneDefaultHTTPClientConfig(),
+	}
+
+	client, err := newAPIClient(args, metrics)
+	require.NoError(t, err)
+
+	// Type assertion to verify interface compliance
+	_ = client
```
tpaschalis: nit: Did we mean to assert something else here?

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+func (cm *configManager) getCachedConfig() ([]byte, error) {
+	p := cm.getCachedConfigPath()
+	return os.ReadFile(p)
+}
+
+func (cm *configManager) setCachedConfig(b []byte) {
+	p := cm.getCachedConfigPath()
+	err := os.WriteFile(p, b, 0750)
+	if err != nil {
+		cm.logger.Log("level", "error", "msg", "failed to flush remote configuration contents the on-disk cache", "err", err)
```
tpaschalis: Why do we change the logging format/convention here from the following?

```
level.Error(cm.logger).Log("msg", "failed to flush remote configuration contents the on-disk cache", "err", err)
```

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+	// This is the base frequency at which we poll the API for configuration changes. A jitter is applied to this value.
+	pollFrequency time.Duration
+
+	// This is the hash of the last successfully loaded and running configuration.
+	// Used to track what config is actually active in the controller.
+	lastLoadedConfigHash string
+
+	// This is the hash of the last configuration received from the remote API.
+	// Used to avoid re-fetching the same config, regardless of whether it loaded successfully.
+	lastReceivedConfigHash string
```
tpaschalis, line 66: Do you think it's worth exposing this as a metric as well? Could it be helpful to check against the server to verify what is being sent? It can be in a follow-up PR

internal/service/remotecfg/api_client.go
```diff
@@ -0,0 +1,119 @@
...
+// Package-level function for creating API clients - can be replaced in tests to
+// use a mock client which doesn't make any API calls.
+var createAPIClient = newAPIClient
+
+// apiClient is a wrapper around the collectorv1connect.CollectorServiceClient that
+// provides metrics and error handling.
+type apiClient struct {
+	client  collectorv1connect.CollectorServiceClient
+	metrics *metrics
+}
```
tpaschalis, line 25: So do I understand correctly that the purpose of this is to 'just' call to `getConfigTime.Observe` in the GetConfig call?

I haven't read the core part of remotecfg.go, but I'm trying to understand the abstraction where we

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+		"config_hash", newConfigHash, "config_size", len(b))
+
+	// If successful, flush to disk and keep a copy.
+	cm.setCachedConfig(b)
+	return nil
+}
+
+func (cm *configManager) fetchLoadLocalConfig() {
+	cachePath := cm.getCachedConfigPath()
+	b, err := cm.getCachedConfig()
```
tpaschalis: Why do we need the cachePath here, is it just for the logging line? Then maybe we can move it inline with the logger just for clarity.

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+
+	// Handle other errors
+	if err != nil {
+		cm.logger.Log("level", "error", "msg", "failed to fetch remote config", "err", err)
+		cm.metrics.totalFailures.Add(1)
+		cm.metrics.lastLoadSuccess.Set(0)
+		return err
+	}
+
+	// Success case - we have valid config data from the API
```
tpaschalis: Let's remove this comment -- we don't now yet that the data is valid

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+			"config_hash", newConfigHash, "already_loaded", alreadyLoaded)
+		return nil
+	}
+
+	// Record that we received this config from remote (regardless of parse success)
+	cm.setLastReceivedCfgHash(newConfigHash)
+
+	err = cm.parseAndLoad(b)
+	if err != nil {
+		// Failed to parse/load the configuration - received hash is recorded, but loaded hash unchanged
```
tpaschalis, line 256: We don't record the received hash here as well. See other comment, maybe it's a good idea to do so?

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// fetchContext contains the dependencies needed for fetching configuration
+type fetchContext struct {
+	getAPIConfig func() (*collectorv1.GetConfigResponse, error)
+}
+
+// fetchLoadConfig attempts to read configuration from the API and the local cache
+// and then parse/load their contents in order of preference.
+// If allowCacheFallback is false, it will not attempt to load from cache on remote failure.
```
tpaschalis: Can we document here what the parameter is about? I had a harder time understanding and then had to go look it up.

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,373 @@
...
+func (cm *configManager) getAstFile() *ast.File {
+	cm.mut.RLock()
+	defer cm.mut.RUnlock()
+	return cm.astFile
+}
+
+// fetchContext contains the dependencies needed for fetching configuration
+type fetchContext struct {
+	getAPIConfig func() (*collectorv1.GetConfigResponse, error)
+}
```
tpaschalis: Why do we introduce this abstraction layer here? Is there a reason we don't want to expose the service here?

Not that it's a big deal, but I'm trying to understand the benefit against the mental cost we're introducing

internal/service/remotecfg/config_manager.go
erikbaranowski: I renamed and updated the comment phrasing

internal/service/remotecfg/config_manager.go
erikbaranowski: removed stale comment

internal/service/remotecfg/config_manager.go
erikbaranowski: updated in all places in the file

internal/service/remotecfg/api_client_test.go
erikbaranowski: I removed the test, not useful IMO

internal/service/remotecfg/config_manager.go
erikbaranowski: added

internal/service/remotecfg/config_manager.go
erikbaranowski: we record it just above `parseAndLoad` via `cm.setLastReceivedCfgHash(newConfigHash)`. I think this makes sense so that we record it prior and regardless of whether the parseAndLoad works.

internal/service/remotecfg/api_client.go
erikbaranowski: apiClient is the client for interacting with the fleet management API. The metrics are used only for GetConfig timings but could be extended to time other requests if we'd like. The apiClient also includes the interceptor for setting the User-Agent header on requests.

internal/service/remotecfg/api_client_test.go
erikbaranowski: nice catch, swapped to all mockCollectorClient

internal/service/remotecfg/config_manager.go
erikbaranowski: good catch, this went through some variations that had more than 1 property to pass through as context

internal/service/remotecfg/api_client.go
tpaschalis: Ok, fine with me then at the moment.

internal/service/remotecfg/config_manager.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/grafana/alloy/internal/service"
+	"github.com/grafana/alloy/internal/util/jitter"
+	"github.com/grafana/alloy/syntax/ast"
+)
+
+const baseJitter = 100 * time.Millisecond
+
+// This value is used when we want to disable polling. We use a value that is
+// slightly less than MaxInt to avoid overflowing
+const disablePollingFrequency = math.MaxInt64 - baseJitter
```
thampiotr, line 25: that's interesting - when could this overflow?

internal/service/remotecfg/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+type metrics struct {
+	lastLoadSuccess        prometheus.Gauge
+	lastFetchNotModified   prometheus.Gauge
+	totalFailures          prometheus.Counter
+	configHash             *prometheus.GaugeVec
+	lastReceivedConfigHash *prometheus.GaugeVec
+	lastFetchSuccessTime   prometheus.Gauge
+	totalAttempts          prometheus.Counter
+	getConfigTime          prometheus.Histogram
```
thampiotr, line 17: unrelated: I would be happy to add a section on Controller dashboard for these metrics to help in future :)

thampiotr (approved):
LGTM!

internal/service/remotecfg/config_manager.go
erikbaranowski: we risk overflowing the math.MaxInt64 when adding a jitter to it if we don't subtract the baseJitter.

PR #4041, opened 2025-07-23
Readiness probe

### LINKED ISSUE #3194: Make initialDelaySeconds and timeoutSeconds configurable for readiness probe in Helm char
In the current Grafana Alloy Helm chart, the initialDelaySeconds and timeoutSeconds values for the readiness probe are hardcoded and not configurable via values.yaml.

In our setup, Alloy often becomes ready too quickly—before some of its components are fully initialized. This occasionally leads to errors shortly after startup, but by that point, Kubernetes already considers the pod healthy and doesn't trigger automatic rollbacks.

Being able to fine-tune the readiness probe by adjusting initialDelaySeconds and timeoutSeconds would help us avoid these false positives and improve deployment stability.

I'm happy to open a PR for this change, but wanted to check if there’s any reason these fields weren’t exposed initially.

peter-meltcafe:
#### PR Description

This PR changes two hard coded values in the helm chart so they are configurable. The two values are initialDelaySeconds and timeoutSeconds which are used in the readinessProbe.

#3194 

#### Notes to the Reviewer
As values.listenPort and values.listenScheme were already configurable items I have added values.timeoutSeconds and values.livenessProbe as top level items. However, values.livenessProbe is an object that takes a full liveness probe config. I believe that adding values.readinessProbe would be a better option for this change as it would give better support for configuring the readinessProbe. Unfortunately, it would make this a breaking change for users of the helm chart who have configured listenPort or listenScheme. I'm very happy to go down this route, if you feel this is a more appropriate way of adding this configuration.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


dehaansa (approved):
LGTM - I fixed the typo in the test case name (rediness -> readiness) and otherwise 👍

PR #4044, opened 2025-07-23
Fixing prom v3 upgrade issues

thampiotr:
#### PR Description

Fixes several issues that the upgrade to prom v3 has introduced:

- Updated to new Prometheus version with a fix for memory corruption. https://github.com/grafana/prometheus/pull/43
- Updated Prometheus/common version with a fix to delimited proto not escaped correctly: https://github.com/prometheus/common/pull/809 
- Change the default metric validation scheme back to 'legacy' because most of backends that Alloy users depend on do not support yet UTF-8.
- Change `scrape_native_histograms` default to `false` as it was incorrectly set to `true` in the upgrade.
- Fixed few documentation issues in prometheus.scrape
- Update integration test to use the new WAL. More work will be needed to test other configurations. 
- Added missing tests to scrape config defaults
- Added a timeout when shutting down the worker pool to prevent Alloy from handing when initial config load fails. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


internal/component/prometheus/scrape/scrape.go
```diff
@@ -154,8 +154,8 @@ func (arg *Arguments) SetToDefault() {
 		ScrapeInterval:           1 * time.Minute,  // From config.DefaultGlobalConfig
 		ScrapeTimeout:            10 * time.Second, // From config.DefaultGlobalConfig
 		ScrapeProtocols:          slices.Clone(defaultScrapeProtocols),
-		ScrapeFallbackProtocol:   string(config.PrometheusText0_0_4), // Use same fallback protocol as Prometheus v2
-		ScrapeNativeHistograms:   true,
+		ScrapeFallbackProtocol:   string(config.PrometheusText0_0_4), // Use the same fallback protocol as Prometheus v2
+		ScrapeNativeHistograms:   false,
```
thampiotr, line 158: Turns out it's still false. That was my misreading of defaults that are changing automatically.

internal/runtime/internal/controller/loader.go
```diff
@@ -291,7 +291,11 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 // Cleanup unregisters any existing metrics and optionally stops the worker pool.
 func (l *Loader) Cleanup(stopWorkerPool bool) {
 	if stopWorkerPool {
-		l.workerPool.Stop()
+		// Wait at most 5 seconds for currently evaluating components to finish.
+		err := l.workerPool.Stop(time.Second * 5)
+		if err != nil {
+			level.Warn(l.log).Log("msg", "timed out stopping worker pool", "err", err)
+		}
```
thampiotr, line 298: Sneaking this in here, but I think it can go in separate PR if I have some more time to do it.

internal/cmd/integration-tests/tests/scrape-prom-metrics/config.alloy
```diff
@@ -1,28 +1,29 @@
...
   scrape_timeout = "500ms"
 }
 
-prometheus.remote_write "scrape_prom_metrics" {
-  endpoint {
-    url = "http://mimir:9009/api/v1/push"
-    send_native_histograms = true
-    metadata_config {
-        send_interval = "1s"
+prometheus.write.queue "scrape_prom_metrics" {
```
thampiotr, line 18: As a follow up, I want to expand our integration tests to cover more cases and protocols and use old remote_write component too. But for now we should switch to the one we use most commonly: new WAL with (mostly) default settings. The only settings we're giving here are to make the test go faster.

internal/runtime/internal/controller/loader.go
thampiotr: This wasn't strictly introduced with Prom v3 change, but when we had "too many open files" issues on a cluster, the instances were stuck forever while terminating, because some tasks in the worker pool were blocked (in OpenTelemetry components). The worker pool contains only evaluation tasks, which will call the `Update`, so I think it's safe to give these tasks a deadline of 5 seconds to complete.

internal/component/prometheus/scrape/scrape.go
ptodev: I think we should change it to "true" soon, but given that it's "false" in Alloy 1.10 it's ok to keep it that way for now.

internal/runtime/internal/worker/worker_pool.go
```diff
@@ -69,9 +72,20 @@ func (w *fixedWorkerPool) QueueSize() int {
 	return w.workQueue.queueSize()
 }
 
-func (w *fixedWorkerPool) Stop() {
+func (w *fixedWorkerPool) Stop(timeout time.Duration) error {
 	close(w.quit)
-	w.allStopped.Wait()
+	done := make(chan struct{})
+	go func() {
```
ptodev: Isn't this goroutine going to keep running even after `case <-time.After(timeout)`?

internal/runtime/internal/controller/loader.go
ptodev: Isn't 5 seconds too short? I'm also worried about corrupted WALs, etc.

thampiotr: > The worker pool contains only evaluation tasks, which will call the Update, so I think it's safe to give these tasks a deadline of 5 seconds to complete.

This is deadline for exiting from `Update` function call, not draining WAL. So I think it's fine. You can see the typical evaluation times on this chart (from one of big internal deployments):

Image

internal/runtime/internal/worker/worker_pool.go
thampiotr: Yes, and this is fine. `Stop` can only be called once. This is part of a shutdown sequence.

go.mod
```diff
@@ -1056,7 +1056,10 @@ replace github.com/fsnotify/fsnotify v1.8.0 => github.com/fsnotify/fsnotify v1.7
 // TODO: remove replace directive once:
 // * There is a release of Prometheus which addresses https://github.com/prometheus/prometheus/issues/14049,
 // for example, via this implementation: https://github.com/grafana/prometheus/pull/34
-replace github.com/prometheus/prometheus => github.com/grafana/prometheus v1.8.2-0.20250709144109-5551df55271b
+replace github.com/prometheus/prometheus => github.com/thampiotr/prometheus v0.0.0-20250731131506-a28a7d9c5b3b // staleness_disabling_v3.4.2 branch
+
+// TODO: remove this replace directive once the upstream issue is fixed: https://github.com/prometheus/common/pull/809
+replace github.com/prometheus/common => github.com/thampiotr/common v0.0.0-20250801132732-61b386b68d79
```
thampiotr: We're almost there! My latest tests have resolved all the issues I know about. The only work remaining is to get the fixes merged upstream: 
- Corrupted metrics names: pending https://github.com/prometheus/prometheus/pull/16957 getting merged upstream, after which this will need cherry-picking to our branch here: https://github.com/grafana/prometheus/pull/44 
- Incorrect metric escaping in protobuf protocol: pending https://github.com/prometheus/common/pull/809 getting merged in common. 

@ptodev it would be great if we can review the rest and I will just replace these dependencies as soon as upstreams are merged 🤞

internal/converter/internal/prometheusconvert/testdata/kubernetes.alloy
```diff
@@ -13,16 +13,14 @@ prometheus.scrape "prometheus1" {
...
-	metric_name_escaping_scheme = "allow-utf-8"
+	forward_to = [prometheus.remote_write.default.receiver]
+	job_name   = "prometheus1"
 }
 
 prometheus.scrape "prometheus2" {
-	targets                     = discovery.kubernetes.prometheus2.targets
-	forward_to                  = [prometheus.remote_write.default.receiver]
-	job_name                    = "prometheus2"
-	metric_name_escaping_scheme = "allow-utf-8"
```
ptodev, line 25: We use a different default for `metric_name_escaping_scheme` from upstream?

internal/component/prometheus/scrape/scrape.go
```diff
@@ -290,10 +306,12 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 	alloyAppendable := prometheus.NewFanout(args.ForwardTo, o.ID, o.Registerer, ls)
 	scrapeOptions := &scrape.Options{
+		// NOTE: This is not Update()-able.
```
ptodev, line 309: It's very weird to have non-updateable arguments. Does it have to be this way? It'd be worth at least documenting it so that it doesn't trip someone up.

go.mod
```diff
@@ -1056,7 +1056,10 @@ replace github.com/fsnotify/fsnotify v1.8.0 => github.com/fsnotify/fsnotify v1.7
 // TODO: remove replace directive once:
 // * There is a release of Prometheus which addresses https://github.com/prometheus/prometheus/issues/14049,
 // for example, via this implementation: https://github.com/grafana/prometheus/pull/34
-replace github.com/prometheus/prometheus => github.com/grafana/prometheus v1.8.2-0.20250709144109-5551df55271b
+replace github.com/prometheus/prometheus => github.com/thampiotr/prometheus v0.0.0-20250731131506-a28a7d9c5b3b // staleness_disabling_v3.4.2 branch
```
kalleep, line 1059: As far as I can tell we use `fix-corrupted-name-label` in your fork, that includes the staleness fix. Are we planning to use your fork here or is this only temporary?

internal/runtime/internal/worker/worker_pool.go
```diff
@@ -4,16 +4,19 @@ import (
...
 )
 
 type Pool interface {
 	// Stop stops the worker pool. It does not wait to drain any internal queues, but it does wait for the currently
-	// running tasks to complete. It must only be called once.
-	Stop()
+	// running tasks to complete up to the specified timeout. It must only be called once.
+	// Returns an error if the timeout is exceeded before all workers have stopped, indicating tasks that
+	// take longer than timeout to complete.
+	Stop(timeout time.Duration) error
```
kalleep: Maybe split out this part into another pr? Then it won't be blocked by Prometheus changes. It would also make this change easier to track not being bundled up with "unrelated" changes

internal/converter/internal/prometheusconvert/testdata/kubernetes.alloy
thampiotr: Yes, this is deliberate, because many of backends that are used with Alloy do not support UTF-8 yet, for example, Mimir.

internal/component/prometheus/scrape/scrape.go
thampiotr: I believe this was the case already previously. If we want to fix it I'd like to do it outside of the scope of this PR. I added it to the docs.

go.mod
thampiotr: This is temporary while we get this fix https://github.com/prometheus/prometheus/pull/16957 merged upstream. I mentioned on Slack, sorry if you missed this.

internal/runtime/internal/worker/worker_pool.go
thampiotr: Ok, was hoping it's not too much to review. Will do separate PR.

kalleep (approved):
LGTM, nice work on this

go.mod
ptodev: I'm happy to merge the PR as long as we don't release the next version with this fork.

ptodev (approved):

thampiotr:
> I'm happy to merge the PR as long as we don't release the next version with this fork.

Thanks! I'll merge now to avoid further toil with merge conflicts and so I can get the dev deployment pipeline working again. I am watching for my upstream merge fixed and will follow-up with fixing the go.mod soon.

PR #4046, opened 2025-07-23
database_observability: add Postgres table schema collector

fridgepoet:
#### PR Description

This PR makes use of Postgres system tables to collect and log information about:

* database name and schema names
* table names in each schema
* and the columns and their attributes

Index and foreign keys will be added in an upcoming PR.
Analogous to existing MySQL collector's table schema collector.

Contributes to https://github.com/grafana/grafana-dbo11y-app/issues/1388

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -52,7 +53,7 @@ The `database_observability.postgres` component doesn't support any blocks. You
 database_observability.postgres "orders_db" {
   data_source_name = "postgres://user:pass@localhost:5432/mydb"
   forward_to = [loki.write.logs_service.receiver]
-  enable_collectors = ["activity", "query_tables"]
```
fridgepoet, line 55: looks like "activity" might've been left accidentally before

internal/component/database_observability/postgres/component.go
```diff
@@ -297,8 +298,27 @@ func (c *Component) startCollectors() error {
...
 		return err
 	}
+
 	c.collectors = append(c.collectors, ciCollector)
 
+	if collectors[collector.SchemaTableName] {
+		stCollector, err := collector.NewSchemaTable(collector.SchemaTableArguments{
+			DB:              dbConnection,
+			InstanceKey:     c.instanceKey,
+			CollectInterval: c.args.CollectInterval,
```
fridgepoet: I can set a configurable collect interval as 1 min might be too frequent before caching. What do y'all think?

internal/component/database_observability/postgres/collector/schema_table.go
```diff
@@ -0,0 +1,340 @@
...
+	    AND nspname NOT LIKE 'pg_temp_%'
+	    AND nspname NOT LIKE 'pg_toast_%'`
+
+	selectTableNames = `
+	SELECT 
+		pg_class.relname as table_name
+	FROM pg_catalog.pg_class pg_class
+	JOIN pg_catalog.pg_namespace pg_namespace ON pg_class.relnamespace = pg_namespace.oid
+	WHERE pg_namespace.nspname = $1 
+		AND pg_class.relkind IN ('r', 'v', 'm', 'f', 'p')  -- filter for application-facing objects
```
cristiangreco: nit: here and in the other queries, I'd use code comments rather than sql comments (because the latter are uselessly sent to the database)

internal/component/database_observability/postgres/component.go
cristiangreco: We could define the collect interval at the collector level, set it to something higher than 1m, and put a TODO to revisit later if/when we add caching.

internal/component/database_observability/postgres/collector/schema_table.go
```diff
@@ -0,0 +1,340 @@
...
+
+const (
+	OP_SCHEMA_DETECTION = "schema_detection"
+	OP_TABLE_DETECTION  = "table_detection"
+	OP_CREATE_STATEMENT = "create_statement"
+	SchemaTableName     = "schema_table"
+)
+
+const (
+	selectDatabaseName = `SELECT current_database()`
```
cristiangreco, line 29: Just for my understanding, I think in a followup we will want to list/iterate over all databases, right?

clayton-cornell:
Doc changes look OK

internal/component/database_observability/postgres/collector/schema_table.go
fridgepoet: I believe we can at most list the databases, but we won't be able to iterate over them to get more information.
During connection, a database already has to be specified and cross-database queries are not supported, so we won't be able to list schemas/tables/etc for other databases.

What do you think?
Is that something we want to do, just list them anyway? That would be `SELECT datname FROM pg_database;`

cristiangreco: >  That would be `SELECT datname FROM pg_database;`

Yes, I was thinking something similar to what postgres_exporter does [here](https://github.com/prometheus-community/postgres_exporter/blob/master/cmd/postgres_exporter/queries.go#L275).

Let's take this offline anyway, it's out of scope from this PR I think.

cristiangreco (approved):

clayton-cornell (approved):

PR #4050, opened 2025-07-24
fix: stage.multiline now passes through structured metadata in loki.process

jan-mrm:
#### PR Description

This PR updates the `stage.multiline` of `loki.process` so that structured meatadata, that is already present for a log line, is passed through and not dropped anymore.

#### Notes to the Reviewer

I didn't open a full bug (GitHub issue) since the fix itself seemed small. I added a test for the fixed behavior.
I'm not sure if the `StructuredMetadata` should be passed like this or should be cloned.
Also I'm a bit irritated because of the multiple `Entry` in line `s.startLineEntry.Entry.Entry.Timestamp` compared to what I'm currently adding in this PR `s.startLineEntry.StructuredMetadata`. (I'm not a go developer 🙂)

##### Bug Description:

When using the `stage.multiline` every log line (independent of it being an actual multiline that is combined or a single line which just matches the `firstline`) that is processed by the stage loses previously added structured metadata since it is not passed to the output `Entry`.

This is especially noticeable since in the k8s-monitoring chart the build-in [structuredMetadata](https://github.com/grafana/k8s-monitoring-helm/blob/ae61aa31f3019f7b95c8f56b3574d23053ae4271/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_log_processing.alloy.tpl#L44-L53) section is added before the [extraLogProcessingStages](https://github.com/grafana/k8s-monitoring-helm/blob/ae61aa31f3019f7b95c8f56b3574d23053ae4271/charts/k8s-monitoring/charts/feature-pod-logs/templates/_common_log_processing.alloy.tpl#L68-L70) section where a custom multiline stage would be added, so that the structured metadata is lost.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Main (unreleased)
 
 - (_Experimental_) Additions to experimental `database_observability.mysql` component:
   - `query_sample` collector now supports auto-enabling the necessary `setup_consumers` settings (@cristiangreco)
+  - add `query_tables` collector for postgres (@matthewnolf)
 
 ### Enhancements
 
@@ -54,6 +55,8 @@ Main (unreleased)
 - Add `application_host` and `network_inter_zone` features to `beyla.ebpf` component. (@marctc)
 
 - Set the publisher name in the Windows installer to "Grafana Labs". (@martincostello)
+
+- Switch to the community maintained fork of `go-jmespath` that has more features. (@dehaansa)
 
 ### Bugfixes
 
@@ -72,6 +75,8 @@ Main (unreleased)
 - Fixed an issue where certain `otelcol.*` components could prevent Alloy from shutting down when provided invalid configuration. (@thampiotr)
 
 - Fix URLs in the Windows installer being wrapped in quotes. (@martincostello)
+
+- Fixed a bug in `prometheus.write.queue` which caused labelling issues when providing more than one label in `external_labels`. (@dehaansa)
 
 - Fix issues in `loki.process` where `stage.multiline` did not pass through structured metadata. (@jan-mrm)
 
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -27,8 +27,16 @@ You can use the following arguments with `database_observability.postgres`:
 
 | Name                               | Type                 | Description                                                                                    | Default | Required |
 |------------------------------------|----------------------|------------------------------------------------------------------------------------------------|---------|----------|
-| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                       |         | yes      |
+| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                    |         | yes      |
 | `forward_to`                       | `list(LogsReceiver)` | Where to forward log entries after processing.                                                 |         | yes      |
+| `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |         | no       |
+| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |
+
+The following collectors are configurable:
+
+| Name              | Description                                              | Enabled by default |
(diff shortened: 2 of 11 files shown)


internal/component/loki/process/stages/multiline.go
```diff
@@ -190,8 +190,9 @@ func (m *multilineStage) flush(out chan Entry, s *multilineState) {
 		Entry: loki.Entry{
 			Labels: s.startLineEntry.Entry.Labels.Clone(),
 			Entry: logproto.Entry{
-				Timestamp: s.startLineEntry.Entry.Entry.Timestamp,
-				Line:      s.buffer.String(),
+				Timestamp:          s.startLineEntry.Entry.Entry.Timestamp,
+				Line:               s.buffer.String(),
+				StructuredMetadata: s.startLineEntry.StructuredMetadata,
```
ptodev:
```suggestion
				StructuredMetadata: s.startLineEntry.Entry.Entry.StructuredMetadata,
```

I agree that `Entry.Entry` looks weird, but I'd prefer to keep it this way. It's more consistent with the rest of the code and it can also prevent ambiguity about what struct field is being used.

Could you also clone it please? The labels are being cloned, so it'd be good to clone this too.

ptodev (commented):
Thank you! Just one minor comment :)

Structured metadata is a feature which was added later on, so there could be places like this where it's not yet supported.

jan-mrm:
Much appreciated! 🙂

> Thank you! Just one minor comment :)

I implemented your comment.

> Structured metadata is a feature which was added later on, so there could be places like this where it's not yet supported.

True, yes, makes sense.

jan-mrm:
@ptodev sorry I missed the second part of your comment about cloning. Adding it.

jan-mrm:
@ptodev StructuredMetadata (of type LabelsAdapter) has no clone and its from a dependency so it probably would have to be added as a loop above (like for the extracted data). I seem to struggle with that :D do you have an implementation for that by any chance?

internal/component/loki/process/stages/multiline.go
```diff
@@ -190,8 +190,9 @@ func (m *multilineStage) flush(out chan Entry, s *multilineState) {
 		Entry: loki.Entry{
 			Labels: s.startLineEntry.Entry.Labels.Clone(),
 			Entry: logproto.Entry{
-				Timestamp: s.startLineEntry.Entry.Entry.Timestamp,
-				Line:      s.buffer.String(),
+				Timestamp:          s.startLineEntry.Entry.Entry.Timestamp,
+				Line:               s.buffer.String(),
+				StructuredMetadata: s.startLineEntry.Entry.Entry.StructuredMetadata,
```
ptodev:
```suggestion
				StructuredMetadata: slices.Clone(s.startLineEntry.Entry.Entry.StructuredMetadata),
```

I think this should be sufficient for cloning purposes.

jan-mrm: thanks, that looks good. I added the needed import as well

jan-mrm:
Hey @ptodev is there anything missing or anything I can do to get this fixed? 
I do see a merge conflict in the changelog. If i update it you probably have to run the checks again. Do you resolve that when merging or should I do that?
Would really appreciate this fix 🙂

ptodev:
Hi @jan-mrm, apologies for the delay! I'm happy to merge it if the PR is rebased and the changelog conflict is resolved. Thank you!

jan-mrm:
@ptodev great, thank you so much. I updated the branch

jan-mrm:
Hey @ptodev , I'm guessing I update the branch again to resolve the new conflict? When would you be ok to merge so I don't have to update the branch again and again? 🙃

ptodev (approved):

ptodev:
@jan-mrm Apologies! The team is looking into ways to change the changelog so that we don't have changelog conflits :)

jan-mrm:
thank you @ptodev for merging this. yes, I'm a big fan of a meaningful changelog but somehow without those conflicts everyone needs to resolve all the time  👍

PR #4051, opened 2025-07-24
fix faro.receiver cors not allowing x-scope-orgid, traceparent headers

mar4uk:
#### PR Description
For multi-tenancy support, users should set the `X-Scope-OrgID` header to requests to `faro.receiver` Alloy. But currently `faro.receiver` CORS configuration doesn't allow `X-Scope-OrgID` header.

Users may want to pass `Traceparent` header for distributed tracing support. This header should be allowed by `faro.receiver` as well

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/faro/faro.receiver.md
+++ b/docs/sources/reference/components/faro/faro.receiver.md
@@ -103,7 +103,7 @@ The default value, `[]`, disables CORS support.
 To support requests from all origins, set `cors_allowed_origins` to `["*"]`.
 The `*` character indicates a wildcard.
 
-The allowed headers for cross-domain requests are hardcoded as: `Content-Type`, `Traceparent`, `X-API-Key`, `X-Faro-Session-Id`, `X-Scope-OrgID`.
+You can use the following headers for cross-domain requests: `Content-Type`, `Traceparent`, `X-API-Key`, `X-Faro-Session-Id`, or `X-Scope-OrgID`.
 
 When the `api_key` argument is non-empty, client requests must have an HTTP header called `X-API-Key` matching the value of the `api_key` argument.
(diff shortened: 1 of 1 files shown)


internal/component/faro/receiver/handler.go
```diff
@@ -19,6 +19,8 @@ import (
 
 const apiKeyHeader = "x-api-key"
 
+var defaultAllowedHeaders = []string{"content-type", "traceparent", apiKeyHeader, "x-faro-session-id", "x-scope-orgid"}
```
ptodev, line 22: Should this be documented on the component's doc page?

mar4uk: makes sense! included the list on the doc page

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -103,6 +103,8 @@ The default value, `[]`, disables CORS support.
 To support requests from all origins, set `cors_allowed_origins` to `["*"]`.
 The `*` character indicates a wildcard.
 
+The allowed headers for cross-domain requests are hardcoded as: `Content-Type`, `Traceparent`, `X-API-Key`, `X-Faro-Session-Id`, `X-Scope-OrgID`.
```
clayton-cornell:
```suggestion
You can use the following headers for cross-domain requests: `Content-Type`, `Traceparent`, `X-API-Key`, `X-Faro-Session-Id`, and `X-Scope-OrgID`.
```

I don't think we need to state they are hardcoded... just say "use these headers".

mar4uk: This sounds better, thank you! 👍

clayton-cornell:
Docs are ok.

dehaansa (approved):

PR #4076, opened 2025-07-29
Update Alloy docs to use new badge shortcode

clayton-cornell:
#### PR Description

Update Alloy docs to use new badge shortcode. https://grafana.com/docs/writers-toolkit/write/shortcodes/#badge

Prettier changes for tables were included in this PR.


dehaansa:
Is the badge styling expected to be different?
new:
image

old:
image

clayton-cornell:
> Is the badge styling expected to be different? 

Yes. The old way was using an HTML span (not really approved/susainable). The badge styling is more focused on in-line, and as a result is a bit smaller.  We can look at a different location within the text - for example at the start of the opening paragraph instead of on its own line.

PR #4096, opened 2025-07-31
database_observability: accept targets and apply relabeling

cristiangreco:
#### PR Description
POC code for internal relabeling in the `database_observability.mysql` component.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/gatherer.go
+++ b/internal/component/database_observability/mysql/gatherer.go
@@ -1,0 +1,85 @@
+package mysql
+
+import (
+	alloy_relabel "github.com/grafana/alloy/internal/component/common/relabel"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+type RelabelingGatherer struct {
+	gatherer prometheus.Gatherer
+	rules    []*alloy_relabel.Config
+}
+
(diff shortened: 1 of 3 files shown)


internal/component/database_observability/mysql/gatherer.go
```diff
@@ -0,0 +1,85 @@
+package mysql
```
cristiangreco, line 1: This can be shared with `database_observability.postgres`.

internal/component/database_observability/mysql/gatherer.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+
+	return metricFamilies, nil
+}
+
+// labelBuilder implements the alloy_relabel.LabelBuilder interface for dto.LabelPair slices
+type labelBuilder struct {
+	labels []*dto.LabelPair
+}
```
cristiangreco, line 34: If we end up with just one single `replace` rule then this can be simplified (the only methods that matter are `Set` and `Del`).

internal/component/database_observability/mysql/gatherer.go
matthewnolf: I was just thinking that, this can be shared between the two components

PR #4097, opened 2025-07-31
database_observability: Add mysql engine version to connection_info

rgeyer:
#### PR Description
Adds the MySQL engine version to the labelset for the connection_info collector.

### DIFF
--- a/internal/component/database_observability/mysql/collector/connection_info.go
+++ b/internal/component/database_observability/mysql/collector/connection_info.go
@@ -19,13 +19,15 @@ var (
 )
 
 type ConnectionInfoArguments struct {
-	DSN      string
-	Registry *prometheus.Registry
+	DSN       string
+	Registry  *prometheus.Registry
+	DBVersion string
 }
 
 type ConnectionInfo struct {
 	DSN        string
 	Registry   *prometheus.Registry
+	DBVersion  string
 	InfoMetric *prometheus.GaugeVec
 
 	running *atomic.Bool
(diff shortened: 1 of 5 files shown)


internal/component/database_observability/mysql/collector/connection_info.go
```diff
@@ -36,13 +38,14 @@ func NewConnectionInfo(args ConnectionInfoArguments) (*ConnectionInfo, error) {
 		Namespace: "database_observability",
 		Name:      "connection_info",
 		Help:      "Information about the connection",
-	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine"})
+	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine", "db_version"})
```
cristiangreco: Change seems ok to me, I'd just check for opinions on the label name. Maybe given that we're using `engine` terminology already, this new one could be `engine_version`?

And maybe we should also rename the `engine` to `engine_name`?

rgeyer: Renamed the new label to `engine_version`. I think `engine` is fine, and self explanatory, and changing it now would have non trivial downstream impacts. Happy to revisit if we think it needs the additional clarity,

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -462,6 +460,7 @@ type ExplainPlanArguments struct {
 	PerScrapeRatio  float64
 	EntryHandler    loki.EntryHandler
 	InitialLookback time.Time
+	DBVersion       string
```
gaantunes, line 463: should we use the same `engine` terminology?

rgeyer: Fair play. Updated that as well. We might consider renaming that in the explain plan schema as well, tho that will require frontend changes.

gaantunes: Might be worth it to keep consistency as we are in the early days yet, if the effort is not huge ofc.

matthewnolf: +1

PR #4099, opened 2025-07-31
database_observability: Add pgsql engine version to connection_info

rgeyer:
#### PR Description
More-or-less a mirror of #4097 to implement this same functionality for pgsql.

#### Notes to the Reviewer
The `SHOW server_version` is still fairly verbose, and provides not only the version, but also the OS flavor and build details (I.E. `17.2 (Debian 17.2-1.pgdg120+1)`).

Do we care about the fine-grained detail, or do we only want the major version (17.2)?

My gut is to just parse this and drop everything after the major version number.

### DIFF
--- a/internal/component/database_observability/postgres/collector/connection_info.go
+++ b/internal/component/database_observability/postgres/collector/connection_info.go
@@ -38,7 +38,7 @@ func NewConnectionInfo(args ConnectionInfoArguments) (*ConnectionInfo, error) {
 		Namespace: "database_observability",
 		Name:      "connection_info",
 		Help:      "Information about the connection",
-	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine", "engine_version", "engine_version_suffix"})
+	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine", "engine_version"})
 
 	args.Registry.MustRegister(infoMetric)
 
@@ -64,7 +64,6 @@ func (c *ConnectionInfo) Start(ctx context.Context) error {
 		dbInstanceIdentifier = "unknown"
 		engine               = "postgres"
 		engineVersion        = "unknown"
-		engineVersionSuffix  = "none"
 	)
 
 	parts, err := ParseURL(c.DSN)
@@ -93,11 +92,8 @@ func (c *ConnectionInfo) Start(ctx context.Context) error {
 	if len(matches) > 1 {
 		engineVersion = matches[1]
 	}
-	if len(matches) > 2 && matches[2] != "" {
-		engineVersionSuffix = strings.TrimSpace(matches[2])
-	}
 
-	c.InfoMetric.WithLabelValues(providerName, providerRegion, dbInstanceIdentifier, engine, engineVersion, engineVersionSuffix).Set(1)
+	c.InfoMetric.WithLabelValues(providerName, providerRegion, dbInstanceIdentifier, engine, engineVersion).Set(1)
 	return nil
 }
 
--- a/internal/component/database_observability/postgres/collector/connection_info_test.go
+++ b/internal/component/database_observability/postgres/collector/connection_info_test.go
@@ -17,40 +17,32 @@ func TestConnectionInfo(t *testing.T) {
 	const baseExpectedMetrics = `
 	# HELP database_observability_connection_info Information about the connection
 	# TYPE database_observability_connection_info gauge
-	database_observability_connection_info{db_instance_identifier="%s",engine="%s",engine_version="%s",engine_version_suffix="%s",provider_name="%s",provider_region="%s"} 1
+	database_observability_connection_info{db_instance_identifier="%s",engine="%s",engine_version="%s",provider_name="%s",provider_region="%s"} 1
 `
 
 	testCases := []struct {
-		name                string
-		dsn                 string
-		engineVersion       string
-		engineVersionSuffix string
-		expectedMetrics     string
+		name            string
+		dsn             string
+		engineVersion   string
+		expectedMetrics string
 	}{
 		{
 			name:            "generic dsn",
 			dsn:             "postgres://user:pass@localhost:5432/mydb",
 			engineVersion:   "15.4",
-			expectedMetrics: fmt.Sprintf(baseExpectedMetrics, "unknown", "postgres", "15.4", "none", "unknown", "unknown"),
+			expectedMetrics: fmt.Sprintf(baseExpectedMetrics, "unknown", "postgres", "15.4", "unknown", "unknown"),
 		},
 		{
 			name:            "AWS/RDS dsn",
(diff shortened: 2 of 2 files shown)


gaantunes:
I wonder why the version information includes build and os info... It might be important for debugging edge cases. Can we put it in a additional field.

rgeyer:
> I wonder why the version information includes build and os info... It might be important for debugging edge cases. Can we put it in a additional field.

Did do, it now produces a result like this..

`database_observability_connection_info{db_instance_identifier="unknown", db_version="17.2", db_version_suffix="(Debian 17.2-1.pgdg120+1)", engine="postgres", instance="dbo11y", job="integrations/db-o11y", provider_name="unknown", provider_region="unknown"}`

internal/component/database_observability/postgres/collector/connection_info.go
```diff
@@ -34,13 +38,14 @@ func NewConnectionInfo(args ConnectionInfoArguments) (*ConnectionInfo, error) {
...
 		Help:      "Information about the connection",
-	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine"})
+	}, []string{"provider_name", "provider_region", "db_instance_identifier", "engine", "engine_version", "engine_version_suffix"})
 
 	args.Registry.MustRegister(infoMetric)
 
 	return &ConnectionInfo{
 		DSN:        args.DSN,
 		Registry:   args.Registry,
+		DBVersion:  args.DBVersion,
```
gaantunes: should we use the same `engine terminology`?

rgeyer: Fair play. Updated that as well. In MySQL we might consider renaming that in the explain plan schema as well.

internal/component/database_observability/postgres/collector/connection_info.go
```diff
@@ -12,19 +13,22 @@ import (
 const ConnectionInfoName = "connection_info"
 
 var (
-	rdsRegex   = regexp.MustCompile(`(?P<identifier>[^\.]+)\.([^\.]+)\.(?P<region>[^\.]+)\.rds\.amazonaws\.com`)
-	azureRegex = regexp.MustCompile(`(?P<identifier>[^\.]+)\.postgres\.database\.azure\.com`)
+	rdsRegex     = regexp.MustCompile(`(?P<identifier>[^\.]+)\.([^\.]+)\.(?P<region>[^\.]+)\.rds\.amazonaws\.com`)
+	azureRegex   = regexp.MustCompile(`(?P<identifier>[^\.]+)\.postgres\.database\.azure\.com`)
+	versionRegex = regexp.MustCompile(`(?P<version>^[1-9]+\.[1-9]+)(?P<suffix>.*)?$`)
```
matthewnolf: nit: I'd put this in a seperate var block as cloud provider info is not the same thing as engine info

also update to engineVersionRegex
```suggestion
	engineVersionRegex = regexp.MustCompile(`(?P<version>^[1-9]+\.[1-9]+)(?P<suffix>.*)?$`)
```

internal/component/database_observability/postgres/collector/connection_info.go
```diff
@@ -58,6 +63,8 @@ func (c *ConnectionInfo) Start(ctx context.Context) error {
 		providerRegion       = "unknown"
 		dbInstanceIdentifier = "unknown"
 		engine               = "postgres"
+		engineVersion        = "unknown"
+		engineVersionSuffix  = "none"
```
matthewnolf: Do we need the suffix for anything? If not I'd remove it for now (easier to add later when we need it)

internal/component/database_observability/postgres/collector/connection_info.go
```diff
@@ -81,7 +88,17 @@ func (c *ConnectionInfo) Start(ctx context.Context) error {
 			}
 		}
 	}
-	c.InfoMetric.WithLabelValues(providerName, providerRegion, dbInstanceIdentifier, engine).Set(1)
+
+	matches := versionRegex.FindStringSubmatch(c.EngineVersion)
+	fmt.Println(matches)
```
matthewnolf:
```suggestion
```

internal/component/database_observability/postgres/collector/connection_info.go
```diff
@@ -81,7 +88,17 @@ func (c *ConnectionInfo) Start(ctx context.Context) error {
...
+	if len(matches) > 1 {
```
matthewnolf, line 92: Should these be inclusive?

internal/component/database_observability/postgres/collector/connection_info.go
gaantunes: I asked Ryan to put this up since it won't hurt and this info might be useful for debugging in the future.

matthewnolf: I feel the opposite - let's not collect data we don't need. We can always add it later once we have a use for it.

internal/component/database_observability/postgres/collector/connection_info.go
rgeyer: ooo yikes, you caught me leaving debugging stuff in, embarassing! Thanks 😂

internal/component/database_observability/postgres/collector/connection_info.go
rgeyer: @cristiangreco and/or @fridgepoet, any opinions here? Looking for a tie breaker vote.

I'm not particular to either solution (collecting and forwarding the "extended version, or not), but it's currently implemented to discover, and send the full version string in two different labels.

internal/component/database_observability/postgres/collector/connection_info.go
rgeyer: Negative.. The first match is the whole regex match, then the subsequent two elements of the slice would be the named matches.

Maybe I could use regex a bit more appropriately? `regexp.Regexp.SubexpIndex(name string)` allows us to find the index of a sub match, tho we know it's location, so we'd still need to check the length of the matches slice to ensure it's in there. 🤔

internal/component/database_observability/postgres/collector/connection_info.go
fridgepoet: Not coming with a strong opinion here, but I am inclined to agree that we can add it later once we have a use for it.

Now just need Cristian to go with the opposite :p

cristiangreco: I'd also vote for not adding it now, especially for the sake of trying to keep the same set of labels across mysql/postgres as much as possible.

rgeyer: Sounds good, I'll pull it for now. I'm going to leave the version parsing regex in place, and allow it to still match the suffix, we just won't extract and forward it as a label.

Thanks all for the engagement!

PR #4104, opened 2025-08-01
Change to the community maintained fork of go-jmespath

### LINKED ISSUE #3333: [loki.process] Ability to remove keys from JSON
### Request

The `stage.json` Loki processing stage uses JMESPath to extract values from an incoming JSON message. JMESPath does not support any way to remove keys by name in its output. I'm proposing to develop a new process stage, `stage.json_drop` with the ability to drop keys from a JSON object.

Given the following JSON entry  and `loki.process`, the output would be as follows:

```json
{"message": "Hello world !", "id": "0xdeadbeef", "kubernetes": {"namespace": "default"}}
```

```alloy
loki.process "process" {
  stage.json_drop {
     drop = ["id", "kubernetes"] // Remove selected top-level keys before forwarding the entry
  }
}
```

```txt
Log:  {"message": "Hello world !"}
```

### Use case

I'm using Grafana Alloy to read messages coming from a Kafka topic, select the relevant informations in the message to craft a log entry, set correct labels and structured metadata, then forward to Loki. 

The Kafka messages are produced by Fluentbit running in Kubernetes clusters that collects pod logs in the ECS format. The ECS format adds data in the JSON object that varies for all services emitting logs in that format. 

<details><summary>Shape of such message: </summary>

```json
{
  "@metadata": { //filebeat metadata },
  "@timestamp": "//UTC ISO Timestamp",
  "agent": { //filebeat metadata },
  "ecs": { //ECS Reference },
  "host": { // Host where filebeat runs  },
  "fields": { // extra fields data from filebeat  },
  "input": { // Filebeat metadata },
  "kubernetes": { // Kubernetes metadata, with container/deployment/podLabels/nodeLabels/Namespace...  },
  "log": { // Logger metadata },
  // 
  "message": "// Actual log message without structure or data",
  "span": { // Span ID structured
    "id": "0000000000"
  },
  "trace": { // Trace ID structured
    "id": "0000000000"
  },
  "http": { // Actual log informations  },
  "foo": { // Many various keys could be here }
}
```

</details>

For my use case, if the `stage.json` could drop keys, I would be able to forward the Kafka message directly to Loki in JSON format after removing all the noise. Since I don't know all the keys that could be added to the JSON, I am stuck selecting a finite set of keys to create a log entry via template.

I am also sure that there are other uses beyond my case for dropping a JSON key during a pipeline step. 

**Additional**
[Grafana Forums Thread](https://community.grafana.com/t/how-to-remove-items-from-extracted-json-map-in-grafana-alloy-before-forwarding-to-loki/119993)

dehaansa:
#### PR Description
Change to the community maintained for of go-jmespath, allowing for more functionality in `stage.json`.

Fixes #3333

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -484,7 +484,7 @@ Then the extracted `ip` value is given as source to `geoip` stage. The `geoip` s
 
 The `stage.json` inner block configures a JSON processing stage that parses incoming log lines or previously extracted values as JSON and uses [JMESPath expressions][] to extract new values from them.
 
-[JMESPath expressions]: https://jmespath.org/tutorial.html
+[JMESPath expressions]: https://jmespath.site/tutorial.html
 
 The following arguments are supported:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ Main (unreleased)
 
 - Add `application_host` and `network_inter_zone` features to `beyla.ebpf` component. (@marctc)
 
+- Switch to the community maintained fork of `go-jmespath` that has more features. (@dehaansa)
+
 ### Bugfixes
 
 - Fix issues with propagating cluster peers change notifications to components configured with remotecfg. (@dehaansa)
--- a/go.mod
+++ b/go.mod
@@ -99,7 +99,6 @@ require (
 	github.com/influxdata/influxdb1-client v0.0.0-20220302092344-a9ab5670611c
 	github.com/jaegertracing/jaeger-idl v0.5.0
 	github.com/jaswdr/faker/v2 v2.3.2
-	github.com/jmespath/go-jmespath v0.4.0
 	github.com/jonboulle/clockwork v0.5.0
 	github.com/json-iterator/go v1.1.12
 	github.com/klauspost/compress v1.18.0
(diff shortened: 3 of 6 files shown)


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -484,7 +484,7 @@ Then the extracted `ip` value is given as source to `geoip` stage. The `geoip` s
 
 The `stage.json` inner block configures a JSON processing stage that parses incoming log lines or previously extracted values as JSON and uses [JMESPath expressions][] to extract new values from them.
 
-[JMESPath expressions]: https://jmespath.org/tutorial.html
+[JMESPath expressions]: https://jmespath.site/tutorial.html
```
clayton-cornell: Is this change intentional? The URL https://jmespath.site/tutorial.html ends in a 404, but the original link goes to a live webpage.

dehaansa: Ah, looks like the community site is `https://jmespath.site/main/#tutorial`

dehaansa:
```suggestion
[JMESPath expressions]: https://jmespath.site/main/#tutorial
```

clayton-cornell: Ah I didn't dig deeper when I saw the original URL was working. :-P

PR #4037, opened 2025-07-22
Update windows exporter dependency to v0.31.1

dehaansa:
#### PR Description
Updating the windows exporter dependency, again. A few more breaking changes, a few upgrades, and a few panic fixes.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [x] Tests updated
- [X] Config converters updated

PR #4108, opened 2025-08-04
 fix: propagate update arguments to windows exporter

kalleep:
#### PR Description
Follow up to https://github.com/grafana/alloy/pull/3943. We did not propagate `update` arguments to windows exporter.
 
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,8 @@ Main (unreleased)
 
 - Fix issue with `faro.receiver` cors not allowing X-Scope-OrgID and traceparent headers. (@mar4uk)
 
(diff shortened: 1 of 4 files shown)


dehaansa:
I believe I've already got this covered in https://github.com/grafana/alloy/pull/4037 Edit: looks like I may have missed the defaulting in the static structs.

kalleep:
> I believe I've already got this covered in #4037 Edit: looks like I may have missed the defaulting in the static structs.

Ah did not see that one. That one is better because we are also updating the exporter. Lets close this one. I just remembered that I was waiting for this fixed to be release before I went away and did a quick check in the code and noticed it was not supported yet. I did not check any prs :(

PR #4110, opened 2025-08-04
fix: gcplog push message to support both message_id and messageId

### LINKED ISSUE #4081: `loki.source.gcplog` expecting `message_id` but GCP only sends `messageId`
### What's wrong?

When using the `loki.source.gcplog` module in Alloy, we are seeing repeated warning logs like the following:

```
{"ts":"2025-07-30T06:46:45.948897014Z","level":"warn","msg":"invalid gcp push request","component_path":"/","component_id":"loki.source.gcplog.local","component":"gcp_push","err":"push message has no ID"}
```

Upon inspecting the HTTP request payload, we see the following structure being sent by Google Cloud Pub/Sub:

```json
{
  "message": {
    "attributes": {
      "logging.googleapis.com/timestamp": "2025-07-30T01:06:37.680525692Z"
    },
    "data": "XXX",
    "messageId": "15758686253422883",
    "publishTime": "2025-07-30T01:06:39Z"
  }
}
```

It appears Alloy is expecting the field `message_id`, but Google is sending `messageId`. Similarly, `publishTime` is used instead of `publish_time`.

From what we can tell, these fields are managed by Google and are not configurable by users. If Alloy is strictly expecting snake_case field names, this may cause compatibility issues with the standard GCP payload format.

### Steps to reproduce

Follow the Modules Push examples - https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.gcplog/#example

### System information

K8s

### Software version

v1.7.5

### Configuration

```text
loki.source.gcplog "local" {
      push {
        labels = {
          job = "gcp",
        }
      }
    }
    
    loki.write "loki" {
      endpoint {
        url = "http://loki-distributor.observability:3100/loki/api/v1/push"
      }
    }
```

### Logs

```text
{"ts":"2025-07-30T06:46:45.948897014Z","level":"warn","msg":"invalid gcp push request","component_path":"/","component_id":"loki.source.gcplog.local","component":"gcp_push","err":"push message has no ID"}
```

kalleep:
#### PR Description
According to https://cloud.google.com/pubsub/docs/push#receive_push request from gcp pub sub should include both `message_id` and `messageId`. But as reported in https://github.com/grafana/alloy/issues/4081 this not always seem to be the case. We have the same problem for `publish_time` and `publishTime`.

So to make this more resilient we can extract both and check both values. 

 Fixes #4081

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
+++ b/internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
@@ -30,12 +30,10 @@ type PushMessageBody struct {
 }
 
 type PushMessage struct {
-	Attributes                 map[string]string `json:"attributes"`
-	Data                       string            `json:"data"`
-	MessageID                  string            `json:"messageId"`
-	DeprecatedMessageID        string            `json:"message_id"`
-	PublishTimestamp           string            `json:"publishTime"`
-	DeprecatedPublishTimestamp string            `json:"publish_time"`
+	Attributes          map[string]string `json:"attributes"`
+	Data                string            `json:"data"`
+	MessageID           string            `json:"messageId"`
+	DeprecatedMessageID string            `json:"message_id"`
 }
 
 // Validate checks that the required fields of a PushMessage are set.
@@ -57,13 +55,6 @@ func (pm PushMessageBody) ID() string {
 		return pm.Message.MessageID
 	}
 	return pm.Message.DeprecatedMessageID
-}
-
-func (pm PushMessageBody) PT() string {
-	if pm.Message.PublishTimestamp != "" {
-		return pm.Message.PublishTimestamp
-	}
-	return pm.Message.DeprecatedPublishTimestamp
 }
 
 // translate converts a GCP PushMessage into a loki.Entry. It parses the


internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
```diff
@@ -22,24 +22,28 @@ import (
...
-		ID               string            `json:"message_id"`
-		PublishTimestamp string            `json:"publish_time"`
-	} `json:"message"`
-	Subscription string `json:"subscription"`
+	Attributes                 map[string]string `json:"attributes"`
+	Data                       string            `json:"data"`
+	MessageID                  string            `json:"messageId"`
+	DeprecatedMessageID        string            `json:"message_id"`
+	PublishTimestamp           string            `json:"publishTime"`
+	DeprecatedPublishTimestamp string            `json:"publish_time"`
```
kalleep: We are not using these fields. Should we just remove them or keep in case we want to use them in the future?

kgeckhart: Perhaps remove them with a comment that they are available? We could potentially use it to expose a latency metric for publish -> receive time but not sure that's really needed since GCP exposes this latency as well (https://cloud.google.com/monitoring/api/metrics_gcp_p_z#gcp-pubsub), 
image

internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
```diff
@@ -48,14 +52,28 @@ func (pm PushMessage) Validate() error {
...
 }
 
+func (pm PushMessageBody) ID() string {
+	if pm.Message.MessageID != "" {
+		return pm.Message.MessageID
+	}
+	return pm.Message.DeprecatedMessageID
+}
+
+func (pm PushMessageBody) PT() string {
```
kgeckhart:
```suggestion
func (pm PushMessageBody) PublishTimestamp() string {
```

nit: I would have to go to the implementation to know that PT in this case is `PublishTimestamp`

kgeckhart (approved):

internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
kalleep: I will drop them and add a comment that links to the structure they use

internal/component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
kalleep: Yeah that is true. But I will remove PublishTime for now anyway so we dont need to have this one

PR #3394, opened 2025-04-17
Document an issue triage workflow for the repository

dehaansa:
#### PR Description
In order to better manage the landscape of issues raised against the Alloy repository, I'd like to propose an issue triage workflow to help define the triage process. A more consistent triage process will make it easier for team members to step in and do a small amount of triage and leave the issue landscape in a better place than they found it. It should also enable us to identify long-term trends on issues.

PR #4117, opened 2025-08-04
Attempt to fix generate script for labels

dehaansa:
#### PR Description
Followup to #3394 and #4116


dehaansa:
Confirmed that this fixes the [workflow](https://github.com/grafana/alloy/actions/runs/16733994946/job/47368565436).

Issue #3711, opened 2025-05-27
validation: complete graph by configuring edges


kalleep:
For validation we use `dag.Graph` but we currently do not setup edges between nodes similar to how its done for [runtime](https://github.com/grafana/alloy/blob/66b0c353ce62c7951280070b3586be59f2b5e9bf/internal/runtime/internal/controller/loader.go#L600).

With edges between nodes we can validate that there are no circular references and that nodes we reference exists.

We should add the same references that we do for runtime https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/loader.go#L600 and perform similar lookup as we do here https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/component_references.go#L44C6-L44C25

Issue #3712, opened 2025-05-27
validation: perform type check of arguments


kalleep:
During validation we only check that properties exists and that required properties are set.

It would be good if we could also check that the the type is correct for the argument.

PR #4127, opened 2025-08-05
feat: connect and validate graph when using validate command

kalleep:
#### PR Description

This pr adds support for validating the graph edges, this includes checking that references exist (arguments, other components and stdlib functions).

I extracted code into ast util package that we can share between runtime and validation. `findReferences` is inspired from `ComponentReferences` but emits diagnostics instead of warning logs.

fixes: #3711

#### Notes to the Reviewer

Even though we can check for references to other components we don't (yet!) the that properties are exported from the target component. That will be implemented as part of https://github.com/grafana/alloy/issues/3712

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/validator/validate.go
+++ b/internal/validator/validate.go
@@ -141,7 +141,8 @@ func (v *validator) validateDeclares(s *state) {
 		}
 
 		configs, declares, services, components := extractBlocks(node, node.block.Body, v.sm)
-		// We set an empty body here because they are evalutated in a sub graph.
+		// We need to empty the body of the declare block so that later when we call findReferences on nodes
+		// we don't find references that are only added to the "sub" graph and validated seperatly.
 		node.block.Body = ast.Body{}
 
 		moduleState := &state{


internal/dag/tarjan.go
```diff
@@ -9,7 +11,30 @@ func StronglyConnectedComponents(g *Graph) [][]Node {
 		lowLink: make(map[Node]int, len(nodes)),
 	}
 
-	for _, n := range g.Nodes() {
+	for _, n := range nodes {
+		// Calculate strong connect components for non-visited nodes
```
thampiotr:
```suggestion
		// Calculate strongly connected components for non-visited nodes
```

internal/validator/testdata/default/invalid_foreach.diags
```diff
@@ -106,3 +106,10 @@ Error: main.alloy:60:1: cannot find the definition of component name "mod.myModu
 60 | mod.myModule "mod" { }
    | ^^^^^^^^^^^^
 61 | 
+
+Error: main.alloy:119:15: component "prometheus.exporter.redis.default.targets" does not exist or is out of scope
```
thampiotr, line 110:
```suggestion
Error: main.alloy:119:15: component "prometheus.exporter.redis.default" does not exist or is out of scope
```

To be more precise with what the component is.

internal/validator/testdata/default/invalid_refs.diags
```diff
@@ -0,0 +1,27 @@
+Error: main.alloy:20:17: component "prometheus.relabel.default.receiver" does not exist or is out of scope
+
+19 | 
+20 |         forward_to = [prometheus.relabel.default.receiver]
+   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
thampiotr:
```suggestion
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Nit :)

internal/validator/graph.go
```diff
@@ -1,32 +1,60 @@
...
+	}
+
+	if g.parent != nil {
+		return g.parent.GetByID(id)
+	}
+
+	return nil
+}
+
+func (g *graph) FromParent(n dag.Node) bool {
```
thampiotr:
```suggestion
func (g *graph) IsFromParent(n dag.Node) bool {
```

internal/validator/graph.go
```diff
@@ -1,32 +1,60 @@
...
+	return g.Graph.GetByID(n.NodeID()) == nil && g.parent != nil
```
thampiotr: we only check if it doesn't exist in the current graph and that parent is not nil - what if the node doesn't exist in the parent for some obscure reason? 

I would imagine this function should check `GetByID` on `g.parent`?

internal/validator/graph.go
```diff
@@ -66,35 +94,110 @@ func validateGraph(s *state) diag.Diagnostics {
...
+			if node.n.args != nil {
+				diags.Merge(typecheck.Block(node.n.block, node.n.args))
+			}
+			diags.Merge(validateGraph(node.state, minStability))
+		}
+
+		refs, refDiags := findReferences(n, s.graph, s.scope, minStability)
+		diags.Merge(refDiags)
+		for _, ref := range refs {
+			// Nodes within a foreach block can reference nodes from outside of the block
```
thampiotr, line 114: Do we have a test for this? I can see we have a test for "the other way around" in invalid_foreach.txtar.

internal/validator/graph.go
```diff
@@ -66,35 +94,110 @@ func validateGraph(s *state) diag.Diagnostics {
...
+			// but the otherway around is not possible. So we use a "sub" graph of the parent so
```
thampiotr:
```suggestion
			// but the other way around is not possible. So we use a "sub" graph of the parent so
```

internal/validator/references.go
```diff
@@ -0,0 +1,68 @@
...
+		componentRefMatch := !resolveDiags.HasErrors()
+
+		_, scopeMatch := scope.Lookup(t[0].Name)
+		if !componentRefMatch && !scopeMatch {
+			diags.Merge(resolveDiags)
+			continue
+		}
+
+		if componentRefMatch {
+			if scope.IsStdlibIdentifiers(t[0].Name) {
```
thampiotr, line 38: I think we should add tests for this and the following deprecated and stability level checks.

internal/validator/validate.go
```diff
@@ -135,10 +140,9 @@ func (v *validator) validateDeclares(s *state) {
 			node.id = node.id + "-" + strconv.Itoa(i)
 		}
 
-		// Add declare to graph
-		s.graph.Add(node)
-
 		configs, declares, services, components := extractBlocks(node, node.block.Body, v.sm)
+		// We set an empty body here because they are evalutated in a sub graph.
+		node.block.Body = ast.Body{}
```
thampiotr: This is kinda hard to follow why... would it panic if we forgot this? is there a way to not have to remember about this?

syntax/typecheck/unwrap.go
```diff
@@ -1,24 +1,36 @@
...
-	aw := &attrWalker{target: name, value: defaultValue}
+func UnwrapBlockAttr(b *ast.BlockStmt, name string, defaultValue syntax.Value) syntax.Value {
+	aw := &attrWalker{target: name, targetKind: defaultValue.Reflect().Kind(), value: defaultValue}
 	ast.Walk(aw, b)
 
 	return aw.value
 }
 
+// TryUnwrapBlockAttr tries to unwrap the attribute value.
+// If the attribute are not found or cannot be unwraped
```
thampiotr:
```suggestion
// TryUnwrapBlockAttr tries to unwrap the attribute value.
// If the attribute are not found or cannot be unwrapped
// the second return argument is set to false.
```

syntax/typecheck/unwrap_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/grafana/alloy/syntax/parser"
 )
 
-func TestTryUnwrapBlockAttr(t *testing.T) {
+func TestUnwrapBlockAttr(t *testing.T) {
```
thampiotr, line 15: Would be good to add tests for the helper function you added.

internal/validator/testdata/default/invalid_foreach.diags
kalleep: This is the same message we get with `alloy run`. But I agree that it is pretty confusing. Will think of a better message for it

internal/validator/graph.go
kalleep: Can't remember why i did not perform that check. But yes it should most likely call `GetByID` on the parent

internal/validator/graph.go
kalleep: Yes as part of valid.txtar https://github.com/grafana/alloy/blob/main/internal/validator/testdata/default/valid.txtar#L131C29-L131C41.

It was already there before this change

internal/validator/validate.go
kalleep: It would not panic but would produce errors that should not be reported. 

The body part is a different "scope". If we keep it there it would give report errors that components local to the declare don't exists because its not part of that same graph but the "sub" graph instead, we need to perform the validation recursively

internal/validator/testdata/default/invalid_foreach.diags
kalleep: This one is pretty tricky because we don't know what part of all the identifiers is a component. To solve that we would probably need to pass in `component.Registry` to and try to find a matching component. Right now I changed the reporting bellow to highlight the full string. But I can try this approach is you think we should. Then the fallback would be to print the full thing

thampiotr: I think whatever is easier as long as it's clear is good. Maybe we can say "component or export ... does not exist".

internal/validator/validate.go
thampiotr: Maybe a bit better comment if it's not possible to avoid? I still  don't fully follow this 😅

kalleep: Yeah my explanation was not that great 

If we have a declare that looks like this:

```
declare "pods_and_nodes" {
	discovery.kubernetes "pods" {
		role = "pod"
	}

		role = "node"
	}

	export "kubernetes_resources" {
		value = array.concat(
			discovery.kubernetes.pods.targets,
			discovery.kubernetes.nodes.targets,
		)
	}
}

pods_and_nodes "pan" {}
```
We do a couple of things during validation like generate arguments, register a custom component and register it as a node in the graph. After that we recursively call validate with its body and create a new graph (that is linked to the parent graph) because the components inside are not reachable from the outside. The only reason we need to set an empty body is so that when we later validate the graph itself we call findReferences for each node and if we have the body still it will find e.g. `discovery.kubernetes.pods.targets` but that one is not validated at the same level.

PR #4131, opened 2025-08-05
Use new Community badge label in Alloy docs

clayton-cornell:
Update community supported component docs:

 - Remove the HTML span that adds a Community badge
 - Add label.tags frontmatter with Community label and tooltip.

Update developer docs with new badge requirements.


clayton-cornell:
image

PR #4135, opened 2025-08-05
DB O11y: PG query samples

gaantunes:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -27,8 +27,17 @@ You can use the following arguments with `database_observability.postgres`:
 
 | Name                               | Type                 | Description                                                                                    | Default | Required |
 |------------------------------------|----------------------|------------------------------------------------------------------------------------------------|---------|----------|
-| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                       |         | yes      |
+| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                    |         | yes      |
 | `forward_to`                       | `list(LogsReceiver)` | Where to forward log entries after processing.                                                 |         | yes      |
+| `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |         | no       |
+| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |
+
+
+The following collectors are configurable:
+
+| Name              | Description                                              | Enabled by default |
+|-------------------|----------------------------------------------------------|--------------------|
+| `query_tables`    | Collect query table information.                         | no                 |
 
 ## Blocks
 
--- a/internal/component/database_observability/postgres/collector/pg_activity.go
+++ b/internal/component/database_observability/postgres/collector/pg_activity.go
@@ -1,0 +1,366 @@
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/lib/pq"
+	"go.uber.org/atomic"
+
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/component/database_observability"
+	"github.com/grafana/alloy/internal/runtime/logging"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const (
+	OP_QUERY_SAMPLE = "query_sample"
+	OP_WAIT_EVENT   = "wait_event"
+)
+
+const selectPgStatActivity = `
+	SELECT 
+		d.datname,
+		s.datid,
+		s.pid,
+		s.leader_pid,
+		s.usesysid,
+		s.usename,		
+		s.application_name,
+		s.client_addr,
+		s.client_port,
+		s.state_change,
+		clock_timestamp() as now,
+		s.backend_start,
+		s.xact_start,
+		s.query_start,		
+		s.wait_event_type,
+		s.wait_event,
+		s.state,
+		s.backend_type,
+		s.backend_xid,
+		s.backend_xmin,
+		s.query_id,
+		s.query,
+		pg_blocking_pids(s.pid) as blocked_by_pids
+	FROM pg_stat_activity s 
+		JOIN pg_database d ON s.datid = d.oid AND NOT d.datistemplate AND d.datallowconn
+	WHERE 
+		s.pid <> pg_backend_pid() AND
+		(
+			s.backend_type != 'client backend' OR
+			(coalesce(TRIM(s.query), '') != '' AND s.query_start IS NOT NULL AND s.state != 'idle' and coalesce(TRIM(s.state), '') != '')
+		)
+`
+
+type ActivityInfo struct {
+	DatabaseName    sql.NullString
+	DatabaseID      int
+	PID             int
+	LeaderPID       sql.NullInt64
+	UserSysID       int
+	Username        sql.NullString
+	ApplicationName sql.NullString
+	ClientAddr      sql.NullString
+	ClientPort      sql.NullInt32
+	StateChange     sql.NullTime
+	Now             time.Time
+	BackendStart    sql.NullTime
+	XactStart       sql.NullTime
+	QueryStart      sql.NullTime
+	WaitEventType   sql.NullString
+	WaitEvent       sql.NullString
+	State           sql.NullString
+	BackendType     sql.NullString
+	BackendXID      sql.NullInt32
+	BackendXmin     sql.NullInt32
+	QueryID         sql.NullInt64
+	Query           sql.NullString
+	BlockedByPids   pq.Int64Array
+}
+
+const ActivityName = "activity"
+
+type ActivityArguments struct {
+	DB                    *sql.DB
+	InstanceKey           string
+	CollectInterval       time.Duration
+	DisableQueryRedaction bool
+	EntryHandler          loki.EntryHandler
+	Logger                log.Logger
+}
+
+type Activity struct {
+	dbConnection          *sql.DB
+	instanceKey           string
+	collectInterval       time.Duration
+	disableQueryRedaction bool
+	entryHandler          loki.EntryHandler
+
+	logger  log.Logger
+	running *atomic.Bool
+	ctx     context.Context
+	cancel  context.CancelFunc
+}
+
+func NewActivity(args ActivityArguments) (*Activity, error) {
+	return &Activity{
+		dbConnection:          args.DB,
+		instanceKey:           args.InstanceKey,
+		collectInterval:       args.CollectInterval,
+		disableQueryRedaction: args.DisableQueryRedaction,
+		entryHandler:          args.EntryHandler,
+		logger:                log.With(args.Logger, "collector", ActivityName),
+		running:               &atomic.Bool{},
+	}, nil
+}
+
+func (c *Activity) Name() string {
+	return ActivityName
+}
+
+func (c *Activity) Start(ctx context.Context) error {
+	level.Debug(c.logger).Log("msg", ActivityName+" collector started")
+
+	c.running.Store(true)
+	ctx, cancel := context.WithCancel(ctx)
+	c.ctx = ctx
+	c.cancel = cancel
+
+	go func() {
+		defer func() {
+			c.Stop()
+			c.running.Store(false)
+		}()
+
+		ticker := time.NewTicker(c.collectInterval)
+
+		for {
+			if err := c.fetchActivity(c.ctx); err != nil {
+				level.Error(c.logger).Log("msg", "collector error", "err", err)
+			}
+
+			select {
+			case <-c.ctx.Done():
+				return
+			case <-ticker.C:
+				// continue loop
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (c *Activity) Stopped() bool {
+	return !c.running.Load()
+}
+
+// Stop should be kept idempotent
+func (c *Activity) Stop() {
+	c.cancel()
+}
+
+// calculateDuration returns a formatted duration string between a nullable time and current time
+func calculateDuration(nullableTime sql.NullTime, currentTime time.Time) string {
+	if nullableTime.Valid {
+		return currentTime.Sub(nullableTime.Time).Round(time.Millisecond).String()
+	}
+	return ""
+}
+
+func (c *Activity) fetchActivity(ctx context.Context) error {
+	rows, err := c.dbConnection.QueryContext(ctx, selectPgStatActivity)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to query pg_stat_activity", "err", err)
+		return err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		activity := ActivityInfo{}
+		err := rows.Scan(
+			&activity.DatabaseName,
+			&activity.DatabaseID,
+			&activity.PID,
+			&activity.LeaderPID,
+			&activity.UserSysID,
+			&activity.Username,
+			&activity.ApplicationName,
+			&activity.ClientAddr,
+			&activity.ClientPort,
+			&activity.StateChange,
+			&activity.Now,
+			&activity.BackendStart,
+			&activity.XactStart,
+			&activity.QueryStart,
+			&activity.WaitEventType,
+			&activity.WaitEvent,
+			&activity.State,
+			&activity.BackendType,
+			&activity.BackendXID,
+			&activity.BackendXmin,
+			&activity.QueryID,
+			&activity.Query,
+			&activity.BlockedByPids,
+		)
+		if err != nil {
+			level.Info(c.logger).Log("msg", "failed to scan pg_stat_activity set", "err", err)
+			continue
+		}
+
+		_, err = c.validateActivity(activity)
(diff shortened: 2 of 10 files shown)


docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -27,8 +27,17 @@ You can use the following arguments with `database_observability.postgres`:
...
 |------------------------------------|----------------------|------------------------------------------------------------------------------------------------|---------|----------|
-| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                       |         | yes      |
+| `data_source_name`                 | `secret`             | [Data Source Name][] for the Postgres server to connect to.                                    |         | yes      |
 | `forward_to`                       | `list(LogsReceiver)` | Where to forward log entries after processing.                                                 |         | yes      |
+| `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |         | no       |
+| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |
+
+
+The following collectors are configurable:
```
```suggestion
| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |

The following collectors are configurable:
```

Minor: just removing an extra line.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -34,9 +34,10 @@ You can use the following arguments with `database_observability.postgres`:
 
 The following collectors are configurable:
 
-| Name              | Description                                              | Enabled by default |
-|-------------------|----------------------------------------------------------|--------------------|
-| `query_tables`    | Collect query table information.                         | no                 |
+| Name              | Description                                                                                               | Enabled by default |
+|-------------------|-----------------------------------------------------------------------------------------------------------|--------------------|
+| `activity`        | Collect PostgreSQL activity information from pg_stat_activity, including query samples and wait events.   | no                 |
```
clayton-cornell, line 39:
```suggestion
| `activity`        | Collect PostgreSQL activity information from `pg_stat_activity`, including query samples and wait events. | no                 |
```

Use code tags here

clayton-cornell:
Docs are fine. One minor non-blocking suggestion

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+	QueryStart      sql.NullTime
+	WaitEventType   sql.NullString
+	WaitEvent       sql.NullString
+	State           sql.NullString
+	BackendType     sql.NullString
+	BackendXID      sql.NullInt32
+	BackendXmin     sql.NullInt32
+	QueryID         sql.NullInt64
+	Query           sql.NullString
+	BlockedByPids   pq.Int64Array
```
matthewnolf:
```suggestion
	BlockedByPIDs   pq.Int64Array
```

internal/component/database_observability/postgres/collector/pg_activity.go
matthewnolf, line 1: I don't think we need the `pg_` prefix on the filename given this is already postgres specific package.

Did you consider calling it something like `query_sample`? The convention is that we named things after what they collect, not where they get it from

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+// calculateDuration returns a formatted duration string between a nullable time and current time
+func calculateDuration(nullableTime sql.NullTime, currentTime time.Time) string {
+	if nullableTime.Valid {
+		return currentTime.Sub(nullableTime.Time).Round(time.Millisecond).String()
+	}
+	return ""
+}
+
+func (c *Activity) fetchActivity(ctx context.Context) error {
+	slog.Info("Fetching activity")
```
matthewnolf: Do we need this line?

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+			&activity.State,
+			&activity.BackendType,
+			&activity.BackendXID,
+			&activity.BackendXmin,
+			&activity.QueryID,
+			&activity.Query,
+			&activity.BlockedByPids,
+		)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to scan pg_stat_activity set", "err", err)
```
matthewnolf:
```suggestion
			level.Error(c.logger).Log("msg", "failed to scan pg_stat_activity", "err", err)
```

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+			level.Debug(c.logger).Log("msg", "invalid pg_stat_activity set", "err", err)
+			continue
+		}
+
+		query := ""
+		if activity.Query.Valid {
+			query = activity.Query.String
+		}
+
+		leaderPID := ""
```
matthewnolf: Why not make leaderPID an int?

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+			&activity.BlockedByPids,
+		)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to scan pg_stat_activity set", "err", err)
+			continue
+		}
+
+		_, err = c.validateActivity(activity)
+		if err != nil {
+			level.Debug(c.logger).Log("msg", "invalid pg_stat_activity set", "err", err)
```
matthewnolf: Is it worth including the queryID here?

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+		xactDuration := calculateDuration(activity.XactStart, activity.Now)
+		backendDuration := calculateDuration(activity.BackendStart, activity.Now)
+
+		databaseName := ""
+		if activity.DatabaseName.Valid {
+			databaseName = activity.DatabaseName.String
+		}
+
+		userName := ""
+		if activity.Username.Valid {
```
matthewnolf: `activity.Username.String` will return either the value, or an empty string if not valid. I don't think we're gaining anything from putting these `.Valid` checks in

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+		if activity.WaitEventType.Valid && activity.WaitEvent.Valid {
+			waitEventFullName = fmt.Sprintf("%s:%s", activity.WaitEventType.String, activity.WaitEvent.String)
+			waitEvent = activity.WaitEvent.String
+			waitEventType = activity.WaitEventType.String
+		}
+
+		// Build query sample entry
+		sampleLabels := fmt.Sprintf(
+			`clock_timestamp="%s" instance="%s" app="%s" client="%s" backend_type="%s" backend_time="%s" state="%s" pid="%d" leader_pid="%s" user="%s" userid="%d" datname="%s" datid="%d" xact_time="%s" xid="%d" xmin="%d" query_time="%s" queryid="%d" query="%s" engine="postgres"`,
+			activity.Now.Format(time.RFC3339Nano),
```
matthewnolf: What is `clock_timestamp` for? We can use `BuildLokiEntryWithTimestamp` to set the timestamp of the log to match when the event took place

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+		return err
+	}
+
+	// Update last scrape time after successful scrape
+	c.lastScrape = scrapeTime
+
+	return nil
+}
+
+func (c Activity) validateActivity(activity ActivityInfo) (bool, error) {
```
matthewnolf: Do we need to return a `bool` here too? Can the lack of an `err` indicate the validation was successful

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+
+const (
+	OP_QUERY_SAMPLE = "query_sample"
+	OP_WAIT_EVENT   = "wait_event"
+	ActivityName    = "activity"
+)
+
+const selectPgStatActivity = `
+	SELECT 
+		d.datname,
```
matthewnolf: nit: How are we ordering these fields?

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+`
+
+type ActivityInfo struct {
+	DatabaseName    sql.NullString
+	DatabaseID      int
+	PID             int
+	LeaderPID       sql.NullInt64
+	UserSysID       int
+	Username        sql.NullString
+	ApplicationName sql.NullString
```
matthewnolf: Are we using all of this data?

internal/component/database_observability/postgres/collector/pg_activity.go
```diff
@@ -0,0 +1,373 @@
...
+		state := ""
+		if activity.State.Valid {
+			state = activity.State.String
+		}
+
+		clientAddr := ""
+		if activity.ClientAddr.Valid {
+			clientAddr = activity.ClientAddr.String
+			if activity.ClientPort.Valid {
+				clientAddr = fmt.Sprintf("%s:%d", clientAddr, activity.ClientPort.Int32)
```
matthewnolf: Similarly, I think this can just be:
`clientAddr := fmt.Sprintf("%s:%d", activity.ClientAddr.String, clientAddr, activity.ClientPort.Int32`

matthewnolf (commented):
Looks good, left a few comments to discuss

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: Yes I can make this change.

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: I think it is useful for debugging. We do the same for the query association collector. Putting it under `debug`  makes most sense though.

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: It can be null, so I think just leaving it as string is easier. Or would you rather log it as 0 in this case?

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: Huum ok, will look into that.

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: Yep I can make it that way.

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: WDYM?

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: We might not be using a few of them, will clean that up, good catch.

internal/component/database_observability/postgres/collector/pg_activity.go
gaantunes: Can you elaborate a bit? Not sure how this is better than the current code.

internal/component/database_observability/postgres/collector/pg_activity.go
matthewnolf: Okay, as you prefer

internal/component/database_observability/postgres/collector/pg_activity.go
matthewnolf: Just was wondering if there was any order on the fields we were selecting so it can be organised and more understandable

internal/component/database_observability/postgres/collector/pg_activity.go
matthewnolf: Yep, I was mentioning a similar idea to this https://github.com/grafana/alloy/pull/4135/files#r2272708550

Effectively, doing the `.Valid` checks are adding a fair bit of branching that I think we can get away without.
The suggestion was that `clientAddr = fmt.Sprintf("%s:%d", clientAddr, activity.ClientPort.Int32)` could replace all of the lines L265-L271
```
if activity.ClientAddr.Valid {
			clientAddr = activity.ClientAddr.String
			if activity.ClientPort.Valid {
				clientAddr = fmt.Sprintf("%s:%d", clientAddr, activity.ClientPort.Int32)
             }
}
```

gaantunes: These fields are very often null, so that is the reason for the checks. does `fmt.Sprintf` handles that internally?

matthewnolf: `.String` or `.Int32` will return an empty string/int if it is null - which is the same thing as you are already doing with the valid checks

gaantunes: Yeah I know, but if they return null we will log ":" as the address field wont we? I want to log only the address if port is not available, address:port if both are available, or empty otherwise.

matthewnolf: Yeah, originally I thought that's probably okay but will leave it up to you

internal/component/database_observability/postgres/collector/query_sample.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+
+		waitEventFullName := ""
+		waitEvent := ""
+		waitEventType := ""
+		if activity.WaitEventType.Valid && activity.WaitEvent.Valid {
+			waitEventFullName = fmt.Sprintf("%s:%s", activity.WaitEventType.String, activity.WaitEvent.String)
+			waitEvent = activity.WaitEvent.String
+			waitEventType = activity.WaitEventType.String
+		}
```
matthewnolf:
```suggestion
		waitEventFullName := ""
		waitEvent := activity.WaitEvent.String
		waitEventType := activity.WaitEventType.String
		if activity.WaitEventType.Valid && activity.WaitEvent.Valid {
			waitEventFullName = fmt.Sprintf("%s:%s", activity.WaitEventType.String, activity.WaitEvent.String)
		}
```

internal/component/database_observability/postgres/collector/query_sample.go
```diff
@@ -0,0 +1,334 @@
...
+
+		// Build query sample entry
+		sampleLabels := fmt.Sprintf(
+			`instance="%s" datname="%s" pid="%d" leader_pid="%s" user="%s" app="%s" client="%s" backend_type="%s" backend_time="%s" xid="%d" xmin="%d" xact_time="%s" state="%s" query_time="%s" queryid="%d" query="%s" engine="postgres"`,
+			c.instanceKey,
+			activity.DatabaseName.String,
+			activity.PID,
+			leaderPID,
+			activity.Username.String,
+			activity.ApplicationName.String,
```
matthewnolf, line 258: Might have asked this before, sorry, but are we showing this in the app?

internal/component/database_observability/postgres/collector/query_sample_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"database/sql"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/go-kit/log"
```
matthewnolf, line 13: nit: can we format these imports

internal/component/database_observability/postgres/collector/query_sample_test.go
```diff
@@ -0,0 +1,257 @@
...
+					))
+			},
+			expectedError: false,
+
+			expectedLabels: []model.LabelSet{
+				{"job": database_observability.JobName, "op": OP_QUERY_SAMPLE, "instance": "test"},
+			},
+			expectedLines: []string{
+				fmt.Sprintf(`level="info" instance="test" datname="testdb" pid="100" leader_pid="" user="testuser" app="testapp" client="127.0.0.1:5432" backend_type="client backend" backend_time="%s" xid="500" xmin="400" xact_time="%s" state="active" query_time="%s" queryid="123" query="SELECT * FROM users" engine="postgres" cpu_time="%s"`,
+					time.Duration(0).String(),
```
matthewnolf: Can we test that the timestamp calculations are working

internal/component/database_observability/postgres/collector/query_sample.go
gaantunes: We do, these are configured to only be shown when available on the log line, right @joannaWebDev ?

lstng: These are the [keys](https://github.com/grafana/grafana-dbo11y-app/blob/7192c2c63a56016b22bc3b1bc1e56f0c49142a83/src/features/postgres-query-samples/utils/tooltipTransformation.ts#L10-L9) we are currently showing. If they are not present in the BE, they will not be displayed in the FE. Also, I understand we are going to change ` clock_timestamp`, so will update that aprt once you guys are good to go

gaantunes: Yes, clock_timestamp is going away.

Issue #4052, opened 2025-07-25
Functionality seems out of sync with documentation


326068345:
### What's wrong?

When I use the import.git module for authentication, I find that I cannot use the password_file parameter

### Steps to reproduce

The error can be reproduced by launching Alloy with the provided configuration

### System information

kubernetes 1.32.3

### Software version

helm chart version: alloy-1.2.0,alloy version:  v1.10.0

### Configuration

```text
import.git "alloy_conf" {
  repository = "https://aaaa/bbbb/cccc.git"
  path = "/"
  pull_frequency = "60s"
  revision = "main"
  basic_auth {
    username = "user1"
    password_file = "/etc/secrets/ALLOY_GIT_PASS"

    }
}
```

### Logs

```text
ts=2025-07-24T18:56:08.540714155Z level=error msg="failed to evaluate config" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node=import.git.alloy_conf err="decoding configuration: /etc/alloy/config.alloy:9:5: unrecognized attribute name \"password_file\""
ts=2025-07-24T18:56:08.540744963Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node_id=import.git.alloy_conf duration=110.488µs
ts=2025-07-24T18:56:08.540834142Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node_id=remotecfg duration=70.914µs
ts=2025-07-24T18:56:08.540892772Z level=info msg="applying non-TLS config to HTTP server" service=http
ts=2025-07-24T18:56:08.54091262Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node_id=http duration=29.747µs
ts=2025-07-24T18:56:08.540924372Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node_id=ui duration=952ns
ts=2025-07-24T18:56:08.540932758Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=e459979baf3ab5950469831b6fa703b7 node_id=cluster duration=731ns
[...]

PR #4149, opened 2025-08-07
docs: import.git basic_auth block

kalleep:
#### PR Description
Auth block in `import.git` only supports setting username and password directly. We used a common block that we have for shared http client code.

Fixes: #4052

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


clayton-cornell:
@kalleep I'm assuming this can be backported to v1.9 and v1.10?

kalleep:
@clayton-cornell Yes we should backport this!

PR #4152, opened 2025-08-07
fix:  prometheus.operator.podmonitors recognizes portNumber from crd

### LINKED ISSUE #3949: portNumber field in Prometheus Operator PodMonitor CRD is ignored
### What's wrong?

The Prometheus Operator PodMonitor CRD (or rather the PodMetricsEndpoint CRD therein) has a portNumber field to allow setting the port number of the metrics endpoint to scrape: https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PodMetricsEndpoint

Alloy seems to ignore this configuration and only recognizes `port` (which can only reference named ports) and the deprecated `targetPort` (which can be used instead of `portNumber` but is deprecated):
https://github.com/grafana/alloy/blob/b09bc3106f3592f4d0151fc75c2864a5b6a7569e/internal/component/prometheus/operator/configgen/config_gen_podmonitor.go#L172

Probably because `portNumber` is somewhat new.

### Steps to reproduce

Have a pod with two ports, e.g. 8080 and 9091, where 9091 should be used for metrics.

Create a PodMonitor resource something like this:

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: my-metrics
  namespace: default
spec:
  podMetricsEndpoints:
  - path: /metrics
    portNumber: 9091
```

Observe that Alloy tries to scrape metrics via port 8080.

### System information

_No response_

### Software version

Grafana Alloy 1.7.4

### Configuration

```text

```

### Logs

```text

```

kalleep:
#### PR Description
This pr adds support for `portNumber` from PodMonitor CRD https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PodMetricsEndpoint
 
 I also updated the reference link we had to point to the current commit we use inside alloy
 
Fixes #3949 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ Main (unreleased)
 
 - Fix ebpf profiler metrics `pyroscope_ebpf_active_targets`, `pyroscope_ebpf_profiling_sessions_total`, `pyroscope_ebpf_profiling_sessions_failing_total` not being updated. (luweglarz)
 
+- Fix `prometheus.operator.podmonitors` so it now handle portNumber from PodMonitor CRD. (@kalleep)
+
 v1.10.1
 -----------------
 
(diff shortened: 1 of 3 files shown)


dehaansa (approved):
LGTM

Issue #1185, opened 2024-07-03
High memory usage with low active time series count


travisgroth:
### What's wrong?

I just deployed alloy via the k8s-monitoring helm chart and noticed the memory usage was fairly high for a single node cluster with only a handful of applications: 700M or so.

While investigating, I discovered the default api-server metrics have extremely high cardinality and I had about 50k active time series.  By dropping a few metrics, I was able to get this down to around 10K but memory usage only came down to about 500M.  I then got more aggressive and simply disabled the apiserver integration.  This gives about 2k active time series and memory is down to 350M.  My last step was to remove all relabels (I had one per job to change the name) which got me down to about 300M steady state.

This seems quite high for the baseline of a small single node cluster.  I understand there's going to be some flat minimum usage for running alloy but I wanted to check - **is this expected for such a small workload**?

By way of comparison, the backend is victoria-metrics and the memory usage is always lower than alloy's steady state.

Here's a visual.  You can see the step function as I restart with each successively smaller active time series count:

<img width="1581" alt="image" src="https://github.com/grafana/alloy/assets/6198562/c1b6a822-7c1b-4114-a149-c7a0e89e3aee">

Configuration is very straight forward (see below).

Installed via [helm](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring#install-chart)

### Steps to reproduce

- Run alloy with ~2k active time series scraping prometheus and writing to prometheus remote storage
- Observe outsize memory usage

### System information

Kubernetes v1.30.1+k3s1 / 

### Software version

v1.2

### Configuration

Note - this is being deployed as a subchart from Argocd so the values are under `k8s-monitoring`.

Helm values:

```yaml

k8s-monitoring:
  cluster:
    name: mycluster

  externalServices:
    prometheus:
      host: http://victoria-metrics-single-server.victoria-metrics.svc.cluster.local:8428
      writeEndpoint: /api/v1/write
    loki:
      host: localhost
  configValidator:
    enabled: false
  configAnalysis:
    enabled: false
  opencost:
    enabled: false
  metrics:
    kubelet:
      enabled: true
      # extraMetricRelabelingRules: |-
      #   rule {
      #     action        = "replace"
      #     replacement   = "kubelet"
      #     target_label  = "job"
      #   }

    cadvisor:
      enabled: true
      # extraMetricRelabelingRules: |-
      #   rule {
      #     action        = "replace"
      #     replacement   = "kubelet"
[...]

PR #4160, opened 2025-08-07
Update logs-in-kubernetes docs to warn users about daemonsets

dehaansa:
#### PR Description
Add a tip for ensuring that you use the appropriate configuration to not collect duplicate logs when deploying Alloy as a daemonset.

See https://github.com/grafana/alloy/issues/1185#issuecomment-3162752392

### DIFF
--- a/docs/sources/collect/logs-in-kubernetes.md
+++ b/docs/sources/collect/logs-in-kubernetes.md
@@ -166,7 +166,7 @@ Replace the following values:
 You can get pods logs through the log files on each node. In this guide, you get the logs through the Kubernetes API because it doesn't require system privileges for {{< param "PRODUCT_NAME" >}}.
 {{< /admonition >}}
 
-{{< admonition type="tip" >}}
+{{< admonition type="note" >}}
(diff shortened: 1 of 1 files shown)


dehaansa:
Not sure if the recommendation for docs would be stacked tips or all in one admonition, let me know. 
image

clayton-cornell:
> Not sure if the recommendation for docs would be stacked tips or all in one admonition, let me know.

Yeah... we try to avoid that wherever possible. Sometimes it's the only way to get the point across, sometimes we can rework the doc structure to combine admonitions... or write the info into body text.

There's no explicit guidance for this in the Writer's Toolkit.

In this case though, the new test addition looks more like a note to me. A tip is helpful extra info that be skipped... in this case that looks like it's important information.

docs/sources/collect/logs-in-kubernetes.md
```diff
@@ -166,6 +166,12 @@ Replace the following values:
 You can get pods logs through the log files on each node. In this guide, you get the logs through the Kubernetes API because it doesn't require system privileges for {{< param "PRODUCT_NAME" >}}.
 {{< /admonition >}}
 
+{{< admonition type="tip" >}}
```
clayton-cornell:
```suggestion
{{< admonition type="note" >}}
```

I think this could be better as a note. It's not something the user can ignore/skip if they have deployed things as a daemonset.

Issue #1637, opened 2024-09-09
prometheus.exporter.static


petewall:
### Request

I've had a few times I've wanted to inject a static metric. Currently, in Kubernetes, the "easiest" option is:
1. Write that static metric in the format:
```
# HELP my_product_build_info A description
# TYPE my_product_build_info gauge
my_product_build_info{features="lowpower"} 1
```
2. Save that file as `my_metric.prom` in a ConfigMap
3. VolumeMount that config map to the alloy pods
4. Use `prometheus.exporter.unix`'s `textfile` system to read the `*.prom` files inside of the mounted directory

This works, but it's cumbersome, requires volumemounts, would not work with other text components (`local.file`, `env(...)`) or just supplying a static string. It also loads the whole unix exporter subsystem, which feels heavyweight for just reading a file.

Here's my proposal:

```
prometheus.exporter.static "build_info" {
  metric {
    name = "my_product_build_info"
    labels = {
      version = "1.2.3"
      features = "turbo-mode"
    }
    value = 1
  }

  metric {
    file = "/path/to/my_metric.prom"
  }

  metric {
    string = remote.kubernetes.configmap.metric["my_metric.prom"]
  }

  metric {
    string = env("MY_METRIC")
  }

  scrape_interval = "1h"
  forward_to = [ prometheus.remote_write.prom.receiver ]
}


prometheus.remote_write "prom" { ... }
```

### Use case

Much easier time creating static metrics without having to do a lot of extra work.

mattdurham:
Looks reasonable, some nits, we could move name to the label. Trying to figure out how loading from a string works, ie is it only a single metric? Would need to add metadata support.

```
metric “test_1” {
  labels = {
    cluster = “prod”
  }
}
```

wildum:
I like the idea, it would also be a great tool to test a pipeline (you could use it to try different scenarios and make sure that your prometheus.relabel components behave as expected with live debugging)

mattdurham:
Also adding a `metrics string` that behaves much like `textfile` in node_exporter so that you can bulk loadl

smaddock:
Possible duplicate of #331?

dehaansa:
We discussed this proposal briefly in our team proposal meeting today and are open to moving forward with it, however we are unlikely to prioritize it internally. If anyone is interested in working on it, please comment and we can discuss a design!

PR #4173, opened 2025-08-11
feat: prometheus.exporter.static

kalleep:
#### PR Description
Add `prometheus.exporter.static` component that is capable of exposing prometheus metrics loaded from text in prometheus exposition format.
 

Fixes: #1637

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/exporter/static/static.go
+++ b/internal/component/prometheus/exporter/static/static.go
@@ -10,6 +10,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
 
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/prometheus/exporter"
@@ -47,7 +48,8 @@ func (a *Arguments) SetToDefault() {
 
 // Validate implements syntax.Validator.
 func (a *Arguments) Validate() error {
-	var p expfmt.TextParser
+	// TODO: add support for choosing validation scheme: https://github.com/grafana/alloy/issues/4122
+	p := expfmt.NewTextParser(model.LegacyValidation)
 	_, err := p.TextToMetricFamilies(strings.NewReader(a.Text))
 	if err != nil {
 		return fmt.Errorf("failed to parse prom text: %w", err)


internal/component/prometheus/exporter/static/static.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+type Gauge struct {
+	Name   string            `alloy:"name,attr"`
+	Help   string            `alloy:"help,attr"`
+	Value  float64           `alloy:"value,attr"`
+	Labels map[string]string `alloy:"labels,attr"`
+}
+
+type Histogram struct {
```
ptodev: I wonder how both native and classic histograms would work. Is `prometheus.NewHistogram` used for both?

kalleep: Not sure. But as mentioned in slack we probably will start with only a text format that we want to expose. Context: https://raintank-corp.slack.com/archives/C02GSU8SHBN/p1754905022466809

kalleep:
@ptodev @thampiotr 

A couple of questions:
1. Are we ok to start with this approach. Currently I have implemented loading in text in prometheus expositon format. It is not as strict as textfile collector in node exporter like not allowing timestamp. I skipped that part because it could be useful for testing to include it.
2. What should the instance key be for this exporter? Maybe configurable?

thampiotr:
> 1. Are we ok to start with this approach.

LGTM

> 2. What should the instance key be for this exporter? Maybe configurable?

I think by default it should be the componentID (component name + name label, e.g. prometheus.static.exporter.foo).
This way if someone sets it up in a cluster and uses clustered scraper, only one instance of the exporter will be getting scraped.

internal/component/prometheus/exporter/static/static.go
```diff
@@ -0,0 +1,108 @@
...
+	})
+}
+
+func createExporter(opts component.Options, args component.Arguments, defaultInstanceKey string) (integrations.Integration, string, error) {
+	a := args.(Arguments)
+	return integrations.NewIntegrationWithInstanceKey(opts.Logger, a.Convert(), opts.ID)
+}
+
+type Arguments struct {
+	Text string `alloy:"text,attr"`
```
kalleep, line 39: I have not implement the full proposal made in https://github.com/grafana/alloy/issues/1637 as we discussed in slack a bit. But I do wonder if we want to make this an array instead so you would specify something like:
```
prometheus.exporter.static "demo" {
  text {
    value = ""
  }
}
```

The reason for that would be if we want to extend this to include alloy syntax for metrics, like the proposal have, it would play better.

A block with the same structure would also be sufficient because that I am pretty sure it would make it backward compatible to convert it to an array in the future.

docs/sources/reference/components/prometheus/prometheus.exporter.static.md
```diff
@@ -0,0 +1,116 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.static/
+aliases:
+  - ../prometheus.exporter.static/ # /docs/alloy/latest/reference/components/prometheus.exporter.static/
```
clayton-cornell: We haven't been adding aliases for new Alloy topics. This doesn't hurt anything but typically we've been dropping these until they are needed (when a previously published topic is moved or renamed).

docs/sources/reference/components/prometheus/prometheus.exporter.static.md
```diff
@@ -0,0 +1,116 @@
...
+description: Learn about prometheus.exporter.static
+labels:
+  stage: experimental
+  products:
+    - oss
+title: prometheus.exporter.static
+---
+
+# `prometheus.exporter.static`
+
```
clayton-cornell, line 12:
```suggestion

{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}

```

docs/sources/reference/components/prometheus/prometheus.exporter.static.md
```diff
@@ -0,0 +1,116 @@
...
+`prometheus.exporter.static` loads metrics from text specified in [`Prometheus exposition format`](https://prometheus.io/docs/instrumenting/exposition_formats/) and expose them for scraping.
```
clayton-cornell:
```suggestion
`prometheus.exporter.static` loads metrics from text specified in [`Prometheus exposition format`](https://prometheus.io/docs/instrumenting/exposition_formats/) and exposes them for scraping.
```

docs/sources/reference/components/prometheus/prometheus.exporter.static.md
```diff
@@ -0,0 +1,116 @@
...
+## Usage
+
+```alloy
+prometheus.exporter.static "<LABEL>" {
+}
+```
+
+## Arguments
+
+You can use the following arguments with `prometheus.exporter.static`:
```
clayton-cornell:
```suggestion
You can use the following argument with `prometheus.exporter.static`:
```

docs/sources/reference/components/prometheus/prometheus.exporter.static.md
```diff
@@ -0,0 +1,116 @@
...
+
+| Name   | Type     | Description                           | Default | Required |
+| ------ | -------- | ------------------------------------- | ------- | -------- |
+| `text` | `string` | Text in prometheus exposition format. | `""`    | no       |
```
clayton-cornell:
```suggestion
| `text` | `string` | Text in Prometheus exposition format. | `""`    | no       |
```

clayton-cornell (approved):

PR #4180, opened 2025-08-12
database_observability: Add support for supplying CSP information

matthewnolf:
#### PR Description
This change enables `database_observability` users to optionally supply AWS cloud provider information for the database they are monitoring. In doing so, we can more reliably extract AWS specific information such as the region and instance name, as well as identify the AWS account info.
These variables are attached to the `database_observability_connection_info` metric and can be used for filtering and grouping of instances for correlation.

We start with only AWS, with the option to add more in future.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -55,7 +55,33 @@ The following collectors are configurable:
...
+[aws]: #aws
+
+
+### `cloud_provider`
+
+The `cloud_provider`block has no attributes.
+The `cloudprovider` block enables you to supply information related to the cloud provider that hosts the database under observation. This information will be appended to metrics collected as labels for easier filtering and grouping.
+It contains an [`aws`][aws] block.
+
+### `aws`
```
```suggestion
[aws]: #aws

### `cloud_provider`

The `cloud_provider` block has no attributes.
It contains zero or more [`aws`][aws] blocks.
You use the `cloud_provider` block to provide information related to the cloud provider that hosts the database under observation.
This information is appended as labels to the collected metrics.
The labels make it easier for you to filter and group your metrics.

### `aws`
```

Suggestions to make this block description consistent with how we describe similar blocks in `loki.rules.kubernetes` and `mimir.rules.kubernetes`.

Is it correct that you can have zero `aws` blocks? I assumed so because `aws` is not required.

matthewnolf: That's right. In future we will have other blocks nested under `cloud_provider` so it's possible that there would be no `aws` config

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -55,7 +55,33 @@ The following collectors are configurable:
...
+The > symbol indicates deeper levels of nesting.
+For example, `cloud_provider` > `aws` refers to a `aws` block defined inside an `cloud_provider` block.
+
+[cloud_provider]: #cloud_provider
+[aws]: #aws
+
+
+### `cloud_provider`
+
+The `cloud_provider`block has no attributes.
```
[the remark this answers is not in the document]
clayton-cornell: Covered in a previous suggestion

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -55,7 +55,33 @@ The following collectors are configurable:
...
+The `cloudprovider` block enables you to supply information related to the cloud provider that hosts the database under observation. This information will be appended to metrics collected as labels for easier filtering and grouping.
```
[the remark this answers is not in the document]
clayton-cornell: Covered in a previous suggestion

internal/component/database_observability/mysql/collector/connection_info.go
```diff
@@ -56,39 +61,53 @@ func (c *ConnectionInfo) Name() string {
...
+		providerAccount      = "unknown"
 		dbInstanceIdentifier = "unknown"
 		engine               = "mysql"
 	)
 
-	host, _, err := net.SplitHostPort(cfg.Addr)
-	if err == nil && host != "" {
-		if strings.HasSuffix(host, "rds.amazonaws.com") {
+	if c.CloudProvider != nil {
+		if c.CloudProvider.AWS != nil {
```
cristiangreco, line 73: Hmm why would we accept a nil `CloudProvider.AWS`? I think that if you configure a `CloudProvider` block then we should validate that you're also configuring one of the possible sub-blocks (e.g. eventually one of `aws`, `azure`, etc). We should also validate that it's exactly one sub-block (no zero, no more than one). wdyt?

internal/component/database_observability/mysql/component_test.go
```diff
@@ -60,6 +60,44 @@ func Test_collectSQLText(t *testing.T) {
 	})
 }
 
+func Test_parseCloudProvider(t *testing.T) {
```
cristiangreco, line 68: Maybe for completeness we could add a case where the arn is invalid?

matthewnolf: The ARN is validated before it is passed to this collector. What would we be testing here beyond a nil check?

internal/component/database_observability/mysql/collector/connection_info.go
matthewnolf: Discussed offline and we agreed there could be a usecase here where CloudProvider is empty in a config to indicate the component's target DB is not run on any cloud provider

cristiangreco (approved):

cristiangreco (approved):

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -56,7 +56,34 @@ The following collectors are configurable:
...
+
+### `aws`
+
+The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html) identifier for the database being monitored.
+
+| Name  | Type     | Description                                             | Default | Required |
+|-------|----------|---------------------------------------------------------|---------|----------|
+| `arn` | `string` | The ARN associated with the database under observation. |         | yes      |
 
 ## Example
```
clayton-cornell, line 88: Should we expand the example a bit... or add a second example to show the new blocks in action?

clayton-cornell (commented):
No specific suggestions on the new doc content.

docs/sources/reference/components/database_observability/database_observability.mysql.md
matthewnolf: Done in https://github.com/grafana/alloy/pull/4180/commits/eacb6064c3babb811648c00e34b30f634a8f3816

PR #4184, opened 2025-08-13
pyroscope.receive_http: fix shutdown condition

korniltsev:
#### PR Description
The current check does `reflect.DeepEqual(c.server, *newArgs.Server.HTTP)` and the provided arguments are of different type, so the check always fails
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/receive_http/receive_http.go
+++ b/internal/component/pyroscope/receive_http/receive_http.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"connectrpc.com/connect"
+
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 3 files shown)


internal/component/pyroscope/receive_http/receive_http.go
```diff
@@ -98,6 +100,13 @@ func (c *Component) Run(ctx context.Context) error {
 }
 
 func (c *Component) Update(args component.Arguments) error {
+	_, err := c.update(args)
+	return err
+}
+
+// returns true if the server was shutdown
+func (c *Component) update(args component.Arguments) (bool, error) {
+	shutdown := false
```
kalleep, line 109: nit: Leave a comment that this is only used for tests. Also this will always be the same value as `serverNeedsRestarting` so maybe we can just use that one to signal that the server was shutdown (even though in some cases it is also started back up)

kalleep (approved):

marcsanmi (approved):
LGTM

Issue #3918, opened 2025-06-30
Context timeout warnings logged when forwarding Pyroscope data after upgrading to 1.9.2


JohanLindvall:
### What's wrong?

After upgrading to Alloy v1.9.2, we are seeing context timeout warnings logged. This is logged by the scraping daemonsets. The forwarding alloy instances also log similar warnings.

The warnings appear at least an hour after the new version was deployed. After that, I revered to 1.9.1.

Does this indicate a real problem? Our config is unchanged compared to 1.9.1

### Steps to reproduce

Install Alloy v1.9.2 using Helm charts and use Pyroscope forwarding

### System information

AKS v1.32.4, ARM64

### Software version

Alloy v1.9.2

### Configuration

```text
===config.alloy===

logging {
	level  = "debug"
	format = "json"
}

import.file "components" {
	filename = "/etc/alloy-components"
}

components.pyroscope_ebpf_discovery "pyroscope_ebpf_discovery" { }

components.output_pyroscope "output_pyroscope" { }

pyroscope.ebpf "local_pods" {
	targets    = components.pyroscope_ebpf_discovery.pyroscope_ebpf_discovery.targets
	forward_to = [components.output_pyroscope.output_pyroscope.receiver]
	demangle   = "full"
}
=== output_pyroscope.alloy===
declare "output_pyroscope" {
	export "receiver" {
		value = pyroscope.write.endpoint.receiver
	}

	pyroscope.write "endpoint" {
		endpoint {
			url = "http://cmb-alloy.monitoring-alloy.svc.cluster.local.:4040"
		}
	}
}
===pyroscope_ebpf_discovery.alloy===
declare "pyroscope_ebpf_discovery" {
	export "targets" {
		value = discovery.relabel.local_pods.output
	}

	discovery.kubernetes "all_pods" {
		role = "pod"

		selectors {
			field = "spec.nodeName=" + sys.env("HOSTNAME")
			role  = "pod"
		}
	}

	discovery.relabel "local_pods" {
		targets = discovery.kubernetes.all_pods.targets

		rule {
			action        = "drop"
			regex         = "Succeeded|Failed"
			source_labels = ["__meta_kubernetes_pod_phase"]
		}

		rule {
			action        = "replace"
			regex         = "(.*)@(.*)"
			replacement   = "ebpf/${1}/${2}"
			separator     = "@"
			source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_container_name"]
[...]

PR #4185, opened 2025-08-13
fix(pyroscope.receive_http): increase default connection limit to 16k

korniltsev:
#### PR Description
100 is too small

Fixes https://github.com/grafana/alloy/issues/3918

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
@@ -92,6 +92,7 @@ To diagnose this issue:
 2. If the connection limit is being reached, you have several options:
 
    **Option A: Increase the connection limit**
+   
    ```alloy
    pyroscope.receive_http "default" {
      http {
--- a/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
@@ -60,7 +60,7 @@ You can use the following arguments to configure the `http` block. Any omitted f
 
 | Name                   | Type       | Description                                                                                                      | Default  | Required |
 |------------------------|------------|------------------------------------------------------------------------------------------------------------------|----------|----------|
-| `conn_limit`           | `int`      | Maximum number of simultaneous HTTP connections. Defaults to 0 (unlimited).                                      | `0`      | no       |
+| `conn_limit`           | `int`      | Maximum number of simultaneous HTTP connections. Defaults to 16384.                                              | `16384`  | no       |
 | `listen_address`       | `string`   | Network address on which the server listens for new connections. Defaults to accepting all incoming connections. | `""`     | no       |
 | `listen_port`          | `int`      | Port number on which the server listens for new connections.                                                     | `8080`   | no       |
 | `server_idle_timeout`  | `duration` | Idle timeout for the HTTP server.                                                                                | `"120s"` | no       |
@@ -74,6 +74,37 @@ You can use the following arguments to configure the `http` block. Any omitted f
 ## Component health
 
 `pyroscope.receive_http` is reported as unhealthy if it's given an invalid configuration.
+
+## Debug metrics
+
+`pyroscope_receive_http_tcp_connections` (gauge): Current number of accepted TCP connections.
+`pyroscope_receive_http_tcp_connections_limit` (gauge): The max number of TCP connections that can be accepted (0 means no limit).
+
+## Troubleshooting
+
+### Connection limit errors
+
+If you see errors like `"failed to push to endpoint" err="deadline_exceeded: context deadline exceeded"` in components that push to `pyroscope.receive_http` (such as `pyroscope.write`), this may be caused by reaching the TCP connection limit of `pyroscope.receive_http`.
+
+To diagnose this issue:
+
+1. Check the `pyroscope_receive_http_tcp_connections` metric to see if it's approaching or at the `pyroscope_receive_http_tcp_connections_limit`.
+2. If the connection limit is being reached, you have several options:
+
+   **Option A: Increase the connection limit**
+   ```alloy
+   pyroscope.receive_http "default" {
+     http {
+       conn_limit = 32768  // Increase from default 16384
+       // ... other settings
+     }
+     // ... rest of configuration
+   }
+   ```
+
+   **Option B: Horizontal scaling**
+   
+   Deploy multiple instances of `pyroscope.receive_http` behind a load balancer to distribute the connection load across multiple receivers. This approach provides better scalability and fault tolerance.
 
 ## Example
 
--- a/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
@@ -80,9 +80,8 @@ You can use the following arguments to configure the `http` block. Any omitted f
 `pyroscope_receive_http_tcp_connections` (gauge): Current number of accepted TCP connections.
 `pyroscope_receive_http_tcp_connections_limit` (gauge): The max number of TCP connections that can be accepted (0 means no limit).
 
-## Troubleshoot
 
-### Connection limit errors
+## Connection limit errors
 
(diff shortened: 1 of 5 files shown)


internal/component/pyroscope/receive_http/receive_http.go
```diff
@@ -29,11 +29,6 @@ import (
 	"github.com/grafana/pyroscope/api/model/labelset"
 )
 
-const (
-	// defaultMaxConnLimit defines the maximum number of simultaneous HTTP connections
-	defaultMaxConnLimit = 100
```
marcsanmi, line 34: I remember being told in the past by one of the Alloy's engineer to keep a limit. I initially set it to no limit by default 🤔  
https://github.com/grafana/alloy/pull/1886#discussion_r1808277827

korniltsev: Our dev cluster is over 1k nodes. We scrape them with 2 alloy replicas. And the default configuration just does not work. We should either increase the limit drastically, or remove it all together.

korniltsev: Prods are even bigger.

Also this is probably aggravated by the v2 latency.

korniltsev: Cc @mattdurham as the original reviewer author

korniltsev: nginx has 512 per worker by default. if we keep the limit it should be something similar,  and proportional to number of cores,  for instance 4096 * {NUM_CORES} .

 I still think we better remove the limit and hurt those users who extremely misuse the proxy, rather then keep the limit and hurt legit users because we chose the limit incorrectly.

korniltsev: Screenshot_20250813_211446
This is from dev-us-central-0 and two scrapers

simonswine: I would like you to reconsider this and set a higher default limit.

I guess each of those connection will take one open file on linux for example. I would hate for this to come back and impact other signal's collection. 

Not too sure let's say your latency was 20ms on v1 (probably even less) and is now on v2 500ms, we would want to allow 2500 connections to achieve the same bandwidth.

korniltsev: if we keep the limit, we should also make sure it is easier detected. (now it manifests as "context deadline exceeded" which gives 0 hints)

clayton-cornell:
No input from the doc side... changes are prettier related and a small description change.

internal/component/pyroscope/receive_http/receive_http.go
korniltsev: @simonswine thank you for your review. I agree, removing the limit is not the right choice. Thanks for bringing me back to earth :see_no_evil:

korniltsev: Screenshot_20250814_125308
Configured `dev-us-central-0` ebpf daemonset to use the proxy

korniltsev:
@clayton-cornell  I've updated the docs as the contents of the PR changed

simonswine (approved):
Thanks for working on this! 

❤️  the trouble shooting addition

LGTM!

docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
```diff
@@ -75,6 +75,37 @@ You can use the following arguments to configure the `http` block. Any omitted f
 
 `pyroscope.receive_http` is reported as unhealthy if it's given an invalid configuration.
 
+## Debug metrics
+
+`pyroscope_receive_http_tcp_connections` (gauge): Current number of accepted TCP connections.
+`pyroscope_receive_http_tcp_connections_limit` (gauge): The max number of TCP connections that can be accepted (0 means no limit).
+
+## Troubleshooting
```
clayton-cornell:
```suggestion
## Troubleshoot
```

General guidance is to avoid gerunds (words ending in "ing").

docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
```diff
@@ -75,6 +75,37 @@ You can use the following arguments to configure the `http` block. Any omitted f
...
+
```
clayton-cornell, line 83: We should avoid headings following headings without text in between.

Do we expect more troubleshooting areas? Or is this (connection limit errors) the only one (for now) that we will have here?

If this is the only expected troubleshooting step, then I'd suggest we drop the `### Connection limit errors` heading. Otherwise we will need to dream up some generic intro text to stick in between the headings

docs/sources/reference/components/pyroscope/pyroscope.write.md
```diff
@@ -135,6 +135,30 @@ In those cases, exported fields are kept at their last healthy values.
 
 `pyroscope.write` doesn't expose any component-specific debug information.
 
+## Troubleshooting
```
clayton-cornell:
```suggestion
## Troubleshoot
```

docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
```diff
@@ -75,6 +75,37 @@ You can use the following arguments to configure the `http` block. Any omitted f
...
+### Connection limit errors
+
+If you see errors like `"failed to push to endpoint" err="deadline_exceeded: context deadline exceeded"` in components that push to `pyroscope.receive_http` (such as `pyroscope.write`), this may be caused by reaching the TCP connection limit of `pyroscope.receive_http`.
+
+To diagnose this issue:
+
+1. Check the `pyroscope_receive_http_tcp_connections` metric to see if it's approaching or at the `pyroscope_receive_http_tcp_connections_limit`.
+2. If the connection limit is being reached, you have several options:
```
clayton-cornell, line 91:
```suggestion
1. If the connection limit is being reached, you have several options:
```

docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
```diff
@@ -75,6 +75,37 @@ You can use the following arguments to configure the `http` block. Any omitted f
...
+
+   **Option A: Increase the connection limit**
+   ```alloy
```
clayton-cornell:
````suggestion
   **Option A: Increase the connection limit**
   
   ```alloy
````

File: docs/sources/reference/components/pyroscope/pyroscope.write.md [hunk 7]
```diff
@@ -135,6 +135,30 @@ In those cases, exported fields are kept at their last healthy values.
...
+
+### Deadline exceeded errors
+
+If you encounter errors like `"failed to push to endpoint" err="deadline_exceeded: context deadline exceeded"` when `pyroscope.write` is pushing to a `pyroscope.receive_http` component, this may indicate that the receiving component has reached its TCP connection limit.
+
+To resolve this issue:
+
+1. Check the `pyroscope_receive_http_tcp_connections` and `pyroscope_receive_http_tcp_connections_limit` metrics on the `pyroscope.receive_http` component.
+2. If the connection limit is being reached, you have several options:
```
clayton-cornell:
```suggestion
1. If the connection limit is being reached, you have several options:
```

docs/sources/reference/components/pyroscope/pyroscope.write.md
```diff
@@ -135,6 +135,30 @@ In those cases, exported fields are kept at their last healthy values.
...
+
+   **Option A: Increase the connection limit** in the `pyroscope.receive_http` configuration:
+   ```alloy
```
clayton-cornell:
````suggestion
   **Option A: Increase the connection limit** in the `pyroscope.receive_http` configuration:
   
   ```alloy
````

docs/sources/reference/components/pyroscope/pyroscope.receive_http.md
clayton-cornell: This is a style thing that we ask for as documented here: https://grafana.com/docs/writers-toolkit/write/markdown-guide/#numbered-lists

docs/sources/reference/components/pyroscope/pyroscope.write.md
clayton-cornell: This is a style thing that we ask for as documented here: [grafana.com/docs/writers-toolkit/write/markdown-guide#numbered-lists](https://grafana.com/docs/writers-toolkit/write/markdown-guide/#numbered-lists)

PR #4187, opened 2025-08-13
chore: pre-compute static label names and values and pack into a string slice

kalleep:
#### PR Description
When checking the source code of `stage.static_labels` component I noticed that we do some unnecessary work during processing such as validation of label name and converting and converting the value into a string.

This can all be done before we run the stage and just loop over label name and value pairs.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/loki/process/stages/static_labels.go
```diff
@@ -20,16 +17,31 @@ type StaticLabelsConfig struct {
...
-	}), nil
+	values := make([]string, 0, len(config.Values)*2)
+	for n, v := range config.Values {
+		if v == nil || *v == "" {
+			continue
+		}
+
+		s, err := getString(*v)
+		if err != nil {
+			return nil, fmt.Errorf("faild to convert static label value: %w", err)
```
kalleep: Not sure if we should error out here or log and continue as we did before

internal/component/loki/process/stages/static_labels.go
```diff
@@ -20,16 +17,31 @@ type StaticLabelsConfig struct {
...
+		}
+
+		if !model.LabelValue(s).IsValid() {
+			return nil, fmt.Errorf("invalid label value: %s", s)
```
kalleep: Same here

thampiotr (approved):
nice, thanks!

ptodev (approved):

internal/component/loki/process/stages/static_labels.go
ptodev: I like it this way when `newStaticLabelsStage` returns an error, since this is guaranteed to error every time and it's better to fix it when it first gets loaded. It's different from the previous behaviour but I think it's an improvement.

PR #4196, opened 2025-08-14
database_observability: Extract traceparent from query samples

matthewnolf:
#### PR Description
Updates query sample collection for MySQL to always attempt to extract the `traceparent` from the sample. To achieve that, we now always read the `sql_text` field, but only log it if redaction is disabled.

If the traceparent can't be parsed, for instance if it's truncated, then it will be empty.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


matthewnolf:
Closing this as it is stale. The changes will be useful in future when trace linkage is prioritised, so can be used as prior-art

PR #4197, opened 2025-08-14
add query redaction

gaantunes:
#### PR Description

This is adding query redaction capacity to the Postgres `query_sample` collector.
Also moving the code to extract tables from the `query_tables` collector to a new `lexer.go` that holds all lexer related code.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/lexer.go
+++ b/internal/component/database_observability/postgres/collector/lexer.go
@@ -1,0 +1,27 @@
+package collector
+
+import (
+	"fmt"
+
+	"github.com/DataDog/go-sqllexer"
+)
+
+// extractTableNames extracts the table names from a SQL query
+func extractTableNames(sql string) ([]string, error) {
+	normalizer := sqllexer.NewNormalizer(
+		sqllexer.WithCollectTables(true),
+	)
+	_, metadata, err := normalizer.Normalize(sql, sqllexer.WithDBMS(sqllexer.DBMSPostgres))
+	if err != nil {
+		return nil, fmt.Errorf("failed to normalize SQL: %w", err)
+	}
+
+	// Return all table names, including those that end with "..." for truncated queries, as we can't know if the table name was truncated or not
+	return metadata.Tables, nil
+}
+
+// redact obfuscates a SQL query by replacing literals with ? placeholders
+func redact(sql string) string {
+	obfuscatedSql := sqllexer.NewObfuscator().Obfuscate(sql)
+	return obfuscatedSql
+}
--- a/internal/component/database_observability/postgres/collector/query_sample.go
+++ b/internal/component/database_observability/postgres/collector/query_sample.go
@@ -91,18 +91,20 @@ type ActivityInfo struct {
 }
 
 type ActivityArguments struct {
-	DB              *sql.DB
-	InstanceKey     string
-	CollectInterval time.Duration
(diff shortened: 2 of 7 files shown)


internal/component/database_observability/postgres/collector/query_sample.go
```diff
@@ -187,88 +190,94 @@ func (c *Activity) fetchActivity(ctx context.Context) error {
...
-		if activity.WaitEventType.Valid && activity.WaitEvent.Valid {
-			waitEventFullName = fmt.Sprintf("%s:%s", activity.WaitEventType.String, activity.WaitEvent.String)
+		waitEvent := sample.WaitEvent.String
+		waitEventType := sample.WaitEventType.String
+		if sample.WaitEventType.Valid && sample.WaitEvent.Valid {
+			waitEventFullName = fmt.Sprintf("%s:%s", sample.WaitEventType.String, sample.WaitEvent.String)
+		}
+
+		// Get query string and redact if needed
+		queryStr := sample.Query.String
```
matthewnolf:
```suggestion
		queryText := sample.Query.String
```

internal/component/database_observability/postgres/collector/lexer.go
matthewnolf, line 1: I'm wondering if we need this dedicated file, seeing as the two things within it are only used in one place respectively.

Generally keeping things close to their consumers makes the code simpler to follow, wondering if that wasn't the case here?

gaantunes: The goal was to keep all lexer operations in a single place to organize the code, instead of referencing it from 2 different files. We might need other lexer operations in the future, like collect comments to get the tracespans. Or we might need to change the lexer/parser again.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -29,14 +29,17 @@ You can use the following arguments with `database_observability.postgres`:
...
 | `forward_to`                       | `list(LogsReceiver)` | Where to forward log entries after processing.                                                 |         | yes      |
+| `collect_interval`                 | `duration`           | How frequently to collect information from database.                                           | `"1m"`  | no       |
+| `query_sample_collect_interval`    | `duration`           | How frequently to collect query samples from database.                                         | `"15s"` | no       |
 | `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |         | no       |
+| `disable_query_redaction`          | `bool`               | Collect unredacted SQL query text including parameters.                                        | `false` | no       |
 | `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |
```
```suggestion
| `disable_collectors`               | `list(string)`       | A list of collectors to disable from the default set.                                          |         | no       |
| `disable_query_redaction`          | `bool`               | Collect unredacted SQL query text including parameters.                                        | `false` | no       |
| `enable_collectors`                | `list(string)`       | A list of collectors to enable on top of the default set.                                      |         | no       |
| `query_sample_collect_interval`    | `duration`           | How frequently to collect query samples from database.                                         | `"15s"` | no       |
```

Alpha sort the rows (required first, then optional)

clayton-cornell (approved):
Approving for docs.

cristiangreco (approved):

PR #4213, opened 2025-08-18
integration-tests: Always run down and use stable network name

kgeckhart:
#### PR Description

Adjusts the way the integration tests are run so that we will still run down when a test fails, ensures that we will always use a stable docker compose network name, and adds some guards to the unix test so that it can still pass on mac + windows.

#### Notes to the Reviewer

Locally, the unix integration test failed because mac does not have huge page support. As soon as a test run fails you are required to take manual steps to be able to run the tests again because of the hard os.Exit on failure. Even if you cleanup the remaining docker containers you must manually remove the network or risk the network logic accidentally picking the old unused network resulting in really confusing errors like,

```
ts=2025-08-18T19:39:30.165204143Z level=warn msg="Failed to send batch, retrying" component_path=/ component_id=prometheus.remote_write.node_exporter subcomponent=rw remote_name=aa1415 url=http://mimir:9009/api/v1/push err="Post \"http://mimir:9009/api/v1/push\": dial tcp: lookup mimir on 127.0.0.11:53: no such host"
ts=2025-08-18T19:40:32.883688331Z level=warn msg="Failed to send batch, retrying" component_path=/ component_id=prometheus.remote_write.node_exporter subcomponent=rw remote_name=aa1415 url=http://mimir:9009/api/v1/push err="Post \"http://mimir:9009/api/v1/push\": dial tcp: lookup mimir on 127.0.0.11:53: no such host"
```

Making sure we always run down and use a stable network name should hopefully make any other problems a lot more obvious.

### DIFF
--- a/internal/cmd/integration-tests/main.go
+++ b/internal/cmd/integration-tests/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


kalleep (approved):

kgeckhart:
I ended up adding one more change to the unix test so that it could pass on darwin. The huge page metrics being missing is what started this chain of fixes.

PR #4012, opened 2025-07-18
Implement Alloy REPL

jharvey10:
### PR Description

Adds two new sub-commands to Alloy: `alloy repl` and `alloy gql`

Both of these provide diagnostic information about a variety of things across Alloy with the potential for adding more in the future.

Most notably, something like the following will enable ECS (and similar) readiness checks from the command line:
```
alloy gql 'alloy{ isReady }'
```
The same can be done through the REPL via:
```
alloy{ isReady }
```

The repl provides full autocomplete for all of the queries and fields that can be used on the GraphQL endpoint. In the future, it may autocomplete the argument names as well. In the meantime for those, [the schema](https://github.com/grafana/alloy/tree/0fdce84e4970c681276df73da2859391f3d4ad6a/internal/service/graphql/graph/schema) is the best place to discover what the args are. 

### Notes for reviewers

There's thre main parts of this pr.
1. The graphql infrastructure which is located at `internal/service/graphql/`
   - `graph` contains the main resolver implementations for queries and some fields, as well as some other folders
      - `model` contains individual pieces of queries, such as a component, an endpoint, or a health object.
      - schema contains the actual graphql schema files that are used to generate the stubs when using `make generate-graphql-stubs`
   - `client` contains the code to invoke the graphql endpoint and get a response back for a query
   - the top-level files define the gqlgen config, asset generation deps, and associated server code
2. `internal/alloycli/repl` contains the implementation of the repl using the `go-prompt` library. It uses the graphql client code to assist with generating suggestions and running actual queries
3. some components now implement an optional TestConnection interface, which allows them to provide a response for what ultimately ends up in the `component{connectionStatus}` field.

### Which issue(s) this PR fixes

Closes #477 sort of. It provides a more strategic and useful option for accessing the most common use cases reported by users of Alloy who want curl/wget/etc.

Closes #1891 

### PR Checklist

- [ ] Documentation added

PR #4214, opened 2025-08-18
REPL: GraphQL Schema First Pass

jharvey10:
#### PR Description

The purpose of this PR is to review the schema that will back the GraphQL API included as part of the REPL.

Full REPL PR: #4012

NOTE: The full PR will be updated to reflect the results of this PR after it is reviewed/approved.

## Paired down, first-pass schema

```graphql
Query {
  # basic info about alloy
  alloy {
    branch
    buildDate
    buildUser
    isReady
    revision
    version
  }

  # single "get" of a component
  component(id) {
    id
    name
    health {
      message
      lastUpdated
    }
  }
  # multi-get of all components
  components
}
```

## Future ideas for schema layout

```graphql
Query {
  # basic info about alloy
  alloy {
    branch
    buildDate
    buildUser
    isReady
    revision
    version
  }

  # single "get" of a component
  component(id) {
    id
    name
    health {
      message
      lastUpdated
    }
    # component config info, as json-encoded strings
    arguments
    debugInfo
    exports

    # structured data about targets and endpoints
    targets # array of "target"
    endpoints # array of "endpoint"
  }
  # multi-get of all components
  components

  # single "get" of a discovery/scrape target
  target(address) {
    address
    hostname
    ip
    port
    sourceComponent
    health {
      message
      lastUpdated
    }
  }
  # multi-get of all targets
  targets

  # single "get" of an exporter endpoint or remote write endpoint
  endpoint(address) {
    address
    hostname
    ip
    port
    sourceComponent
    health {
      message
      lastUpdated
    }
  }
  # multi-get of all endpoints
  endpoints

  # receivers? (LogsReceiver?, Appendable?) # multi-get of all receivers
}
```

### Is Alloy up and running?

The `alloy` query provides a mix of static fields and a dynamic `isReady` field, which uses the same readiness health checks within alloy that already exist. The advantage of having this in GraphQL (and by extension the REPL) is that it can be accessed via command line script. This will solve ECS-type use cases which require a script to be invoked inside of a container to determine its readiness.

### Is DNS working?

The `target` and `endpoint` queries allow a user to specify any address and the API will resolve a hostname to an IP (and vice versa), both of which are accessible in the resulting data object.

### What things outside of Alloy are connected to my components?

Components can have network connections to a variety of things; discovery targets, remote write endpoints, etc. The `targets` and `endpoints` queries/fields manifest these on the API for end users in a more meaningful way than a raw dump of a component's `exports` field.

### Can I reach a target, exporter endpoint, or remote write endpoint from Alloy?

Since the `target` and `endpoint` queries let you specify any arbitrary address, it can be used as a simple connectivity test to a target/endpoint. The result is manifested in the `health` field.

**Note:** In the draft REPL PR, the implementation returns something like "TCP connection successful" or a low-level error message given back by the Go networking code. This could be enhanced in the future to cover more use cases, such as cert-aware TLS connections, etc.

### What components are running in Alloy?

The `components` and `component` queries provide a way to get details about each component running in Alloy, similar to what is viewable in the Alloy UI today (both remote and locally configured ones).

### Is a component running in a meaningful way?

The `component` query manifests the component's health, along with all of its targets/endpoints and their health. This provides a comprehensive view of the status/health of the component.


internal/service/graphql/graph/schema/components.graphqls
```diff
@@ -0,0 +1,48 @@
...
+  """
+  component(id: ID!): Component
+}
+
+type Component {
+  "Arguments of the component."
+  arguments: String!
+
+  "Connection status of the component."
+  connectionStatus: String!
```
ptodev: What is "connection status" referring to?

internal/service/graphql/graph/schema/endpoints.graphqls
```diff
@@ -0,0 +1,49 @@
...
+  """
+  Endpoint for a given host and port.
+  """
+  endpoint(host: String!, port: Int!): Endpoint!
+}
+
+"""
+Type of endpoint used by Alloy for different communication purposes.
+"""
+enum EndpointType {
```
ptodev: "Endpoint" isn't a term that we normally use in Alloy, unless it's in the context of an HTTP server starting up. Normally in Alloy everything is an argument, block or export. Is there any way to conform to just that terminology?

internal/service/graphql/graph/schema/alloy.graphqls
ptodev, line 1: Will the UI also use the GraphQL schema one day? Sorry for the silly question, know next to nothing about GraphQL and UIs 😅

internal/service/graphql/graph/schema/components.graphqls
jharvey10: @dehaansa Can provide more context for this one, but my understanding is that it is a component-specific way to perform an end-to-end connection test, which may mean something like writing a test metric to a remote endpoint.

What is challenging here is that a connection test doesn't necessarily make sense for all components, and even for ones that it does, they may or may not implement the necessary interface. I believe right now in the REPL draft PR, it is implemented for prometheus remote write and otelcol exporters.

I do wonder if there's a way we can better structure this to account for the fact that not all components have a meaningful "connection" to make.

We also tried hard to make these fields all read-oriented (i.e. "give me the status", as opposed to "test the connection").

internal/service/graphql/graph/schema/endpoints.graphqls
jharvey10: The challenge here is that an export (by virtue of the `Component` interface) can mean basically anything (i.e. it's defined as `interface{}`. My intent with `endpoint` is truly to have it mean any server outside of Alloy that Alloy connects to. Some `exports` fall into this category, and some do not.

From a debugging perspective, I imagine that the things that are most important to see are the things that came from *outside* of Alloy, like discovery targets, so I wrote this particular API catering to that use case.

I'm not opposed to rethinking this or shuffling it around, but there's a subtle challenge (that's also an opportunity) with the GraphQL API: It needs to be very structured and well-defined to be useful (i.e. a component always has a `name` field). But it also doesn't necessarily have to follow the programmatic conventions internal to Alloy. I'd argue that it's more important for this API to cater to end users than to mirror Alloy's own internal structure, even if that means a departure from our internal nomenclature. Then the GraphQL layer acts as the bridge between those two things.

internal/service/graphql/graph/schema/alloy.graphqls
jharvey10: Yeah! I wanted to include that in the hackathon, but ran out of time. In a future state, I can invision the REPL, the Alloy UI, and Fleet Management all using this API as their primary source of information out of Alloy.

It's a pretty easy lift to query this API from the UI once the endpoint is up and running.

jharvey10:
I updated the schema in the PR description (not in the source files yet!). Please take a look at the updated version for a complete future state of the GraphQL API as well as a paired down first-pass version of the schema (suggested by @kalleep) which is the primary focus of this PR.

thampiotr:
I think the paired down version looks solid. The targets / endpoints may need some more discussion around naming etc, but they would unlock a bunch of nice use cases. It's also experimental so we should be able to change our mind if needed.

jharvey10:
Thanks, @thampiotr! I'll update the schema files to reflect that so we can get this PR merged soon

internal/service/graphql/graph/schema/alloy.graphqls
ptodev: Nice, that'd be great!

PR #4216, opened 2025-08-19
database_observability: extend auto-enable to wait consumers

cristiangreco:
#### PR Description
In the mysql collector `query_sample`, extend the auto-enable functionality to also turn on `events_waits_{current,history}` consumers.

Followup of https://github.com/grafana/alloy/pull/3922

#### Notes to the Reviewer
No need to update changelog as the entry from previous PR is generic enough.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/query_sample.go
+++ b/internal/component/database_observability/mysql/collector/query_sample.go
@@ -74,7 +74,7 @@ WHERE
 const updateSetupConsumers = `
 	UPDATE performance_schema.setup_consumers
 		SET enabled = 'yes'
-		WHERE name = 'events_statements_cpu'`
+		WHERE name in ('events_statements_cpu', 'events_waits_current', 'events_waits_history')`
 
 type QuerySampleArguments struct {
 	DB                          *sql.DB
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -74,7 +74,7 @@ WHERE
 const updateSetupConsumers = `
 	UPDATE performance_schema.setup_consumers
 		SET enabled = 'yes'
-		WHERE name = 'events_statements_cpu'`
+		WHERE name in ('events_statements_cpu', 'events_waits_current', 'events_waits_history')`
```
matthewnolf, line 77: Is there a test which should be updated here

cristiangreco: There's an existing test but it can stay unchanged as the change in the query doesn't really affect the logic of the collector. However, I've just pushed a minor update that makes the test reports "3 rows affected" for the sake of consistency.

PR #4217, opened 2025-08-19
database_observability: move common log labels to middleware

matthewnolf:
#### PR Description
This change modifies how common labels are attached to log entries throughout the `database_observability.mysql`  and `database_observability.postgres` components. Previously key data was passed to each collector, which then in turn appended those to the log entries.

With this change, that data is added within a middleware, removing the need from collectors to ensure it is logged. This makes it easier for us to add other common labels to all collectors in future - such as a unique server ID.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/component.go
+++ b/internal/component/database_observability/postgres/component.go
@@ -243,8 +243,7 @@ func (c *Component) startCollectors() error {
 		return err
 	}
 	c.dbConnection = dbConnection
-	entryHandler := loki.NewEntryHandler(c.handler.Chan(), func() {})
-	entryHandler = addLokiLabels(entryHandler, c.instanceKey)
+	entryHandler := addLokiLabels(loki.NewEntryHandler(c.handler.Chan(), func() {}), c.instanceKey)
 
 	collectors := enableOrDisableCollectors(c.args)
 
--- a/internal/component/database_observability/mysql/component.go
+++ b/internal/component/database_observability/mysql/component.go
@@ -295,8 +295,7 @@ func (c *Component) startCollectors() error {
 		return err
 	}
 
-	entryHandler := loki.NewEntryHandler(c.handler.Chan(), func() {})
-	entryHandler = addLokiLabels(entryHandler, c.instanceKey)
+	entryHandler := addLokiLabels(loki.NewEntryHandler(c.handler.Chan(), func() {}), c.instanceKey)
 
 	collectors := enableOrDisableCollectors(c.args)
 


matthewnolf:
> The idea is neat, the only things that concerns me a bit is the fact that tests don't reflect anymore the actual precise format of a log line, which is a bit misleading to read.
> 
> The only workaround I can think about is wrapping the `loki_fake.NewClient` itself with a middleware in tests. wdyt?

I do tend to agree, and had that thought initially. However if we consider that the collectors are no longer responsible for adding these labels, then including those labels in the tests is somewhat out of scope. Still agree we should test those labels somewhere, but as the collectors are no longer in control of the additional labels - if we make external changes it will break these tests which is undesirable. 
Is there somewhere else you think we could test that? Possibly some kind of e2e test

cristiangreco:
> > The idea is neat, the only things that concerns me a bit is the fact that tests don't reflect anymore the actual precise format of a log line, which is a bit misleading to read.
> > The only workaround I can think about is wrapping the `loki_fake.NewClient` itself with a middleware in tests. wdyt?
> 
> I do tend to agree, and had that thought initially. However if we consider that the collectors are no longer responsible for adding these labels, then including those labels in the tests is somewhat out of scope. Still agree we should test those labels somewhere, but as the collectors are no longer in control of the additional labels - if we make external changes it will break these tests which is undesirable. Is there somewhere else you think we could test that? Possibly some kind of e2e test

I see what you mean, but yeah as you say I feel we should now test that the common labels are there somehow. Given current code structure, the responsibility is now on `mysql/component.go` and `postgres/component.go`. We could extend their corresponding tests to make sure that any collector enabled in the component emits logs that include the common labels. Ideally we would use some sort of mock collector, although that's not feasible with the current code as collector names and their initialisation is hardcoded.

internal/component/database_observability/postgres/component.go
```diff
@@ -244,13 +244,13 @@ func (c *Component) startCollectors() error {
 	}
 	c.dbConnection = dbConnection
 	entryHandler := loki.NewEntryHandler(c.handler.Chan(), func() {})
+	entryHandler = addLokiLabels(entryHandler, c.instanceKey)
```
cristiangreco: Maybe we can merge these two lines?

matthewnolf: Sure

cristiangreco (approved):

PR #4222, opened 2025-08-19
fix: remove orphaned wal directories when endpoint is removed from prometheus.write.queue

### LINKED ISSUE #4095: `prometheus.write.queue` should delete WALs of deleted endpoints
### What's wrong?

If an endpoint is removed from the `prometheus.write.queue` config, it still stays on disk even after the config is reloaded or Alloy is restarted. This is not desirable. For example, a user might want to remove the endpoint if the WAL for that endpoint takes too much space.

### Steps to reproduce

Remove an endpoint. Check the `wal` folder.

### System information

_No response_

### Software version

_No response_

### Configuration

```text

```

### Logs

```text

```

kalleep:
#### PR Description
When endpoints are removed from the configuration we should remove the directories associated with them. This pr also handles scanning for orphaned wal dirs if config is changed during downtime of alloy.

fixes: #4095

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,9 +119,13 @@ Main (unreleased)
 
 - Fix issue in prometheus remote_write WAL which could allow it to hold an active series forever. (@kgeckhart)
 
+<<<<<<< HEAD
 - Fix issue in static and promtail converter where metrics type was not properly handled. (@kalleep)
 
 - Fix `prometheus.operator.*` components to allow them to scrape correctly Prometheus Operator CRDs. (@thomas-gouveia)
+=======
+- Fix issue in `prometheus.write.queue` where wal directories left orphaned if endpoint was removed. (@kalleep)
+>>>>>>> cf2d1574e (update changelog)
 
 v1.10.2
 -----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,13 +119,9 @@ Main (unreleased)
 
 - Fix issue in prometheus remote_write WAL which could allow it to hold an active series forever. (@kgeckhart)
 
-<<<<<<< HEAD
 - Fix issue in static and promtail converter where metrics type was not properly handled. (@kalleep)
 
 - Fix `prometheus.operator.*` components to allow them to scrape correctly Prometheus Operator CRDs. (@thomas-gouveia)
-=======
-- Fix issue in `prometheus.write.queue` where wal directories left orphaned if endpoint was removed. (@kalleep)
->>>>>>> cf2d1574e (update changelog)
 
 v1.10.2
 -----------------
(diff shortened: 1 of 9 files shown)


kalleep:
This warrants an integration test. Where would be the best place to that that up?

jharvey10:
Is it worth considering any sort of "cooldown" period before deleting the wal data? Like if an endpoint gets removed (maybe a configuration error) and comes back within a few minutes, but we've already thrown away the wal data.

kalleep:
> Is it worth considering any sort of "cooldown" period before deleting the wal data? Like if an endpoint gets removed (maybe a configuration error) and comes back within a few minutes, but we've already thrown away the wal data.

Good question. By configuration error you mean like accidentally remove some endpoint? @ptodev do you have any opinion on this?

kgeckhart:
> > Is it worth considering any sort of "cooldown" period before deleting the wal data? Like if an endpoint gets removed (maybe a configuration error) and comes back within a few minutes, but we've already thrown away the wal data.
> 
> Good question. By configuration error you mean like accidentally remove some endpoint? @ptodev do you have any opinion on this?

I think the answer to this largely depends on the behavior of `write.queue` on startup. IIRC it attempts to read any data left on the WAL and send it. If so, does it also have a "don't send data older than x?" setting? We could use that to define the period we should wait before cleaning up because after that point the data won't be sent anyway. Having this sort of delayed cleanup is much more complicated from a race condition perspective though.

kalleep:
>I think the answer to this largely depends on the behavior of write.queue on startup. IIRC it attempts to read any data left on the WAL and send it. If so, does it also have a "don't send data older than x?" setting? We could use that to define the period we should wait before cleaning up because after that point the data won't be sent anyway. Having this sort of delayed cleanup is much more complicated from a race condition perspective though.

Yes that is something I am a bit worried about. The solution I can think of for that is to run a cleanup job in background every so often. Check for wal dirs that are not connected to any endpoint and stat them. If it has been modified within the configured "don't send older than" window we keep them otherwise we clean them up. But yeah this feels pretty racy

jharvey10:
> Good question. By configuration error you mean like accidentally remove some endpoint?

Yeah, though to be totally honest, I don't know if this is a point worth designing for or not. 

IIUC, the risk here would be: User deletes endpoint from their config while a bunch of data is still unwritten to the remote endpoint. User then wants to undo this change in a way that preserves the data that hadn't yet been written to the remote. I have no pulse on whether or not a user would care about this, since *technically* they did remove the endpoint (whether on purpose or by accident).

For this very niche edge case, the new behavior is slightly different than before, where config changes could be made back and forth with no loss of the unwritten data (I think?), but I don't know if that's a useful behavior given the potential benefits of this PR as it stands.

PR #4230, opened 2025-08-20
Do not bail if Beyla terminates with an error

rafaelroquetto:
#### PR Description

Beyla can sometimes terminate with an error. If that happens, we shouldn't bail, but instead allow the Alloy component to retry (since it's got a new config anyway). Oftentimes the error is just informative, such as "the instrumenter took too long to finalise and was forcefully aborted". 

If we allow an error to be returned here, Beyla will never ever be reloaded again, as the entire Component Run loop will go away.

Fixes https://github.com/grafana/fleet-management/issues/815

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rafaelroquetto:
@ptodev hi - it looks like a few integration tests are failing with seemingly unrelated causes. Looks like the runners are flaky - any idea of how to go about this?

PR #4240, opened 2025-08-21
integration-tests: Use docker compose directly and introduce a stateful mode

kgeckhart:
#### PR Description
This PR improves the integration-test feedback loop by,
* Replacing the `github.com/testcontainers/testcontainers-go/modules/compose` dependency with `os.Cmd` calls to `docker compose`
    * The compose project created by the testcontainers compose module could not be managed directly by the docker compose CLI
    * This creates problems such as, https://github.com/grafana/alloy/pull/4213, where containers/networks had to be manually deleted 
    * Replacing it also drops a lot of dependencies and makes it easier to manage. 
* Replace the 10 second initialization wait for `docker compose` healthchecks on services which are required to be healthy for the tests to start
    * The list of services which require healthchecks might need further tuning later
    * The alloy container would hard stop if kafka wasn't available and logs pipelines were flaky if Loki wasn't ready
* Introduces a stateful mode for fast local iteration
    * This will not run a `docker compose down` at the end of a test run leaving everything in place for the next run
    * Test safety is maintained through adding the start parameter to Mimir/Loki range queries
    * The start parameter is set to the start time of the corresponding alloy test container through environment variables

Runtime improvements with `time`
testcontainers compose:
`go run . --skip-build  4.54s user 4.92s system 19% cpu 49.768 total`

direct compose using healthchecks instead of 10s sleep:
`go run . --skip-build  4.47s user 4.24s system 25% cpu 34.773 total`

stateful second run
`go run . --skip-build --stateful  3.91s user 3.76s system 43% cpu 17.591 total`

- [x] Documentation added
- [x] Tests updated


internal/cmd/integration-tests/tests/loki-enrich/enrich_test.go
kgeckhart, line 1: There's two tests in this file which were previously using the same `test_name` = `network_device_enriched` in the alloy.config. Since they are two unique tests producing different data it felt safer to ensure isolation of their data by having them use different `test_name`s

internal/cmd/integration-tests/main.go
```diff
@@ -31,46 +38,26 @@ func main() {
...
-		fmt.Println("Stop test containers with docker compose config")
-		err := compose.Down(ctx, tc.RemoveImagesAll)
-		if err != nil {
-			panic(fmt.Errorf("could not remove the docker compose: %v", err))
-		}
-	}()
-
-	fmt.Println("Sleep for 10 seconds to ensure that the env has time to initialize...")
-	time.Sleep(10 * time.Second)
+	executeCommand("docker", []string{"compose", "up", "kafka", "loki", "--wait"}, "Waiting for necessary compose services to be healthy")
```
kalleep: How come we are only waiting on kafka and loki? Do we hope that everything else is started if the health checks are fine for them or should we also wait for other services such as mimir, tempo etc?

kgeckhart: When I was running things locally the tests were running without issues consistently waiting on only these two to be available. I was going to wait to see what happened in CI before marking it ready but it's not running (I might need to resolve the merge conflict?).

kalleep: Yeah so one of them is possibly the slowest to start, feels a bit bad to rely on that but it is probably fine for now :+1: .

kgeckhart: Yeah I'll keep the 10 sec wait when not running stateful since we know that's going to be fine. Then can look to remove it when I add a health check for the rest of the services. This way CI is less likely to be unstable

PR #4269, opened 2025-08-26
database observabiilty: include support for Aurora MySQL v3.x / MySQL v8.0.x

augustobott:
#### PR Description

When turning on db-011y with MySQL, we end up on a situation with the queries for query sampling on db-o11y only work with MySQL versions 8.4.0 or later. This is incompatile with services such as AWS RDS Aurora MySQL (a couple of columns on the "regular" query don't exist yet on MySQL version 8.0.x).

Patches db-o11y to allow sampling queries from older versions of MySQL. In this specific use-case, for AWS Aurora version v3.x (which is compatible with MySQL v8.0.x). The code has been patched to check once if the MySQL server version is lower than version 8.4.0 and sets a flag, used on the code to select which flavor to use (legacy or modern).

#### Notes to the Reviewer

This has been tested with Aurora MySQL v3.04 (compatible with MySQL Community v8.0.28) and v3.10 (compatible with MySQL Community v8.0.42). Both Aurora MySQL versions tested are advertised as LTS by AWS (long term support). 

This has also been tested with another RDS MySQL instance running version 8.4.6 successfully.

#### PR Checklist

Not entirely sure about this next section, so leaving as-is for the time being.

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -36,7 +38,45 @@ SELECT unix_timestamp() AS now,
 FROM performance_schema.global_status
 WHERE variable_name = 'UPTIME'`
 
-const selectQuerySamples = `
+// selectQuerySamplesLegacy is used for MySQL versions older than 8.4.x
+// i.e. Aurora MySQL v3.x is compatible with MySQL v8.0.x
+const selectQuerySamplesLegacy = `
```
cristiangreco: It's a nitpick, but rather than calling this "legacy" could we add a suffix that refers to the version? e.g. `selectQuerySamplesMysql80` or something similar.

augustobott: Sure thing. I'll get those updated. Thanks for your feedback.

internal/component/database_observability/mysql/collector/query_sample.go
```diff
@@ -138,6 +185,12 @@ func (c *QuerySample) Start(ctx context.Context) error {
 	c.ctx = ctx
 	c.cancel = cancel
 
+	if err := c.checkMySQLVersion(ctx); err != nil {
+		level.Error(c.logger).Log("msg", "failed to check mysql version, defaulting to legacy query format", "err", err)
+		// Default to false in case of an error
+		c.isMySQL84OrGreater.Store(false)
+	}
+
```
cristiangreco: Just a note that ideally we would run this check in `internal/component/database_observability/mysql/component.go` and pass the value to the collector. However I don't want to suggest overcomplicating your change, for the sake of helping you get started asap. I think it's something that can be updated later in a followup PR maybe.

augustobott: A followup PR sounds good if that's reasonable. Perhaps this check should be inside function `startCollectors`?

matthewnolf:
@augustobott Thanks very much for this! Would you be able to include the error message you're currently running into? I'm curious so that we can build understanding on which columns have changed between 8.x versions

internal/component/database_observability/mysql/collector/query_sample.go
cristiangreco: Yes, correct, in that file we're already running a `SELECT VERSION()` for other purposes - so that could be a good place to parse the version string and pass it down to this file.

augustobott:
> @augustobott Thanks very much for this! Would you be able to include the error message you're currently running into? I'm curious so that we can build understanding on which columns have changed between 8.x versions

Sure thing. This is the error message when running the _current version_ in a Docker container (I just rebuilt the whole thing a couple of minutes before running this)

```
ts=2025-08-27T09:02:20.994927135Z level=error msg="collector error" component_path=/ component_id=database_observability.mysql.REDACTED collector=query_sample err="failed to fetch query samples: Error 1054 (42S22): Unknown column 'statements.MAX_CONTROLLED_MEMORY' in 'field list'"
```

Steps to reproduce? Just ran `docker build -t alloy-mysql-monitoring .` with the following Dockerfile:

```
# Use the official Grafana Alloy image
FROM grafana/alloy:latest

# Copy the configuration file into the container
COPY local.file /etc/alloy/config.alloy

# The command to run when the container starts.
# This now correctly includes the "run" subcommand.
CMD ["run", "/etc/alloy/config.alloy", "--stability.level=experimental"]
```

File `config.alloy` follows instructions found [here](https://github.com/grafana/alloy/tree/main/internal/component/database_observability).

matthewnolf:
> > @augustobott Thanks very much for this! Would you be able to include the error message you're currently running into? I'm curious so that we can build understanding on which columns have changed between 8.x versions
> 
> Sure thing. This is the error message when running the _current version_ in a Docker container (I just rebuilt the whole thing a couple of minutes before running this)
> ...


Thanks, so from looking at the [relevant docs](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-events-statements-current-table.html) - the following columns differ between 8.0 -> 8.4

* `CPU_TIME`: Added in 8.0.28
* `MAX_CONTROLLED_MEMORY`: Added in 8.0.31
* `MAX_TOTAL_MEMORY`: Added in 8.0.31

In the `selectQuerySamplesLegacy`, we should remove `CPU_TIME` additionally in that case.

augustobott:
> Thanks, so from looking at the [relevant docs](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-events-statements-current-table.html) - the following columns differ between 8.0 -> 8.4
> 
> * `CPU_TIME`: Added in 8.0.28
> * `MAX_CONTROLLED_MEMORY`: Added in 8.0.31
> * `MAX_TOTAL_MEMORY`: Added in 8.0.31
> 
> In the `selectQuerySamplesLegacy`, we should remove `CPU_TIME` additionally in that case.

Fair enough. I'll make adjustments to the code to factor those in.

matthewnolf:
@augustobott It looks like some of the tests are failing and need updating with the additional logic added. We should pass the `DBVersion` as an argument in the `NewQuerySample` constructor [here](https://github.com/grafana/alloy/blob/d9d197af9ffca20270ffd88226f0cd71a0a46173/internal/component/database_observability/mysql/collector/query_sample_test.go#L395-L399) so that when the version is interrogated there is a value.

Can we also add a test to verify each code path is reached (versions before `8.0.28`, `8.0.31`, `8.4.0`).

Let me know if I can help at all!

matthewnolf:
Hey @augustobott let me know if I can help at all getting this pushed through and fixing this issue

augustobott:
We decided to take a different approach for this. Hence, closing up this PR.

PR #4275, opened 2025-08-27
fix: loki source file read legacy position

kalleep:
#### PR Description
When using `legacy_positions_file` alloy will translate the old position format to what we use in alloy. in the old format labels was never stored so we generate position with empty label set.

In our docs we have this note:
>The legacy_positions_file argument is used when you are transitioning from legacy. The legacy positions file is converted to the new format. This operation only occurs if the new positions file doesn’t exist and the legacy_positions_file is valid. If you add any labels before loki.source.file, then the positions file conversion won’t work. The legacy positions file didn’t have a concept of labels in the positions file, so the conversion assumes no labels.

This limitation is hard to deal with when doing a migration in e.g. a k8s cluster where discovery is used. 

This is an attempt to remove this limitation. If we have configured `legacy_positions_file` we first try to see if we have any existing position entry for path + label combination. Otherwise we try with an empty labels `{}`. This would allow us to pick up recorded path from a legacy collector e.g. promtail or agent-static.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.file.md
+++ b/docs/sources/reference/components/loki/loki.source.file.md
@@ -54,8 +54,7 @@ When set to true, only new logs are read, ignoring the existing ones.
 The `legacy_positions_file` argument is used when you are transitioning from Grafana Agent Static Mode to Grafana Alloy. 
 The format of the positions file is different in Grafana Alloy, so this will convert it to the new format.
 This operation only occurs if the new positions file doesn't exist and the `legacy_positions_file` is valid.
-When `legacy_positions_file` is set Alloy will first try to find the position identified by path and labels, if not found
-it will try to find the positon using path and empty labels to use as a start offset.
+When `legacy_positions_file` is set, Alloy will try to find previous positions for a given file by matching the path and labels, falling back to matching on path only if no match is found.
 
 ## Blocks
 


docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -51,12 +51,8 @@ If not set, it defaults to UTF-8.
 You can use the `tail_from_end` argument when you want to tail a large file without reading its entire content.
 When set to true, only new logs are read, ignoring the existing ones.
 
-{{< admonition type="note" >}}
 The `legacy_positions_file` argument is used when you are transitioning from legacy. The legacy positions file is converted to the new format.
```
ptodev:
```suggestion
The `legacy_positions_file` argument is used when you are transitioning from Grafana Agent Static Mode to Grafana Alloy. 
The format of the position files is different in Grafana Alloy, so this will convert convert it to the new format.
```
I think this is what it's trying to say?

internal/component/loki/source/file/file.go
```diff
@@ -329,6 +329,7 @@ func (c *Component) createReader(path string, labels model.LabelSet) (reader, er
 			c.args.Encoding,
 			pollOptions,
 			c.args.TailFromEnd,
+			c.args.LegacyPositionsFile != "",
```
ptodev, line 332: I suppose there are no unintended side effects if `LegacyPositionsFile != ""`, but there is also a non-legacy positions file present?

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -51,12 +51,8 @@ If not set, it defaults to UTF-8.
...
 This operation only occurs if the new positions file doesn't exist and the `legacy_positions_file` is valid.
-If you add any labels before `loki.source.file`, then the positions file conversion won't work.
-The legacy positions file didn't have a concept of labels in the positions file, so the conversion assumes no labels.
-{{< /admonition >}}
```
ptodev, line 59: IIUC, ignoring the labels is assuming that a file is only present once in `targets`? And the reason for tracking labels in Alloy's loki.source.file is that a file may be present more than once, with different labels? If that's the case should we check if a file is present more than once and return a warning?

There must be a reason for tracking the labels, so I'm just wondering what could possibly break if we don't.

internal/component/loki/source/file/file.go
kalleep: Then it reads from non legacy file.

docs/sources/reference/components/loki/loki.source.file.md
kalleep: Yes you are correct. With how it currently works it is possible to tail the same file if we have several targets with different labels but the same __path__ and record different position for them

internal/component/loki/source/file/file.go
kalleep: Thinking a bit about it there is one side effect.

If we set legacy_file_position and translation happens we now have a new file with new permissions format. Then if we ever get new targets that with new labels it will always start to read from the offset that was translated. This would only ever happen if legacy_position_file attribute is set. If we set it, perform the migration and then later remove the attribute we will have the translation but we will not pick up the "migrated" position

ptodev: If the only bad thing that happens is that something might be read twice, that's ok. As long as no logs are lost then it's all fine.

kalleep: Just to try to clarify a bit because when I read my own explanation its not clear what I meant.

When we translate legacy position we get a record looking like this 
```
  ? path: /var/log/pods/somepod/0.log
    labels: '{}'
  : "30290"
```

The problem I am addressing in this pr is that if we have any [non internal labels](https://github.com/grafana/alloy/blob/e45b01c9a17d4576459b956610e5105f5b1ee896/internal/component/loki/source/file/file.go#L221) we could not pick up the previous recorded position. With the fix in this pr we will check if we have recorded any position with empty labels i.e. the migrated one if no position is found for path + labels. This fallback will only happen as long as `legacy_file_position` is set and is only used as the start offset. Then all updates to the position will be recorded with the combination of path + labels so we are not touching the old one.

I guess there could be a weird case if we also have an active tailer for path + empty labels. Then it would pick up the position that was last recorded for that one. 
 
 But in general the attribute should be set when rolling out the new deployment, let it pick up the previous position and then the attribute should be removed.

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -51,12 +51,11 @@ If not set, it defaults to UTF-8.
 You can use the `tail_from_end` argument when you want to tail a large file without reading its entire content.
 When set to true, only new logs are read, ignoring the existing ones.
 
-{{< admonition type="note" >}}
-The `legacy_positions_file` argument is used when you are transitioning from legacy. The legacy positions file is converted to the new format.
+The `legacy_positions_file` argument is used when you are transitioning from Grafana Agent Static Mode to Grafana Alloy. 
+The format of the position files is different in Grafana Alloy, so this will convert convert it to the new format.
```
thampiotr:
```suggestion
The format of the positions file is different in Grafana Alloy, so this will convert it to the new format.
```

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -51,12 +51,11 @@ If not set, it defaults to UTF-8.
...
 This operation only occurs if the new positions file doesn't exist and the `legacy_positions_file` is valid.
-If you add any labels before `loki.source.file`, then the positions file conversion won't work.
-The legacy positions file didn't have a concept of labels in the positions file, so the conversion assumes no labels.
-{{< /admonition >}}
+When `legacy_positions_file` is set Alloy will first try to find the position identified by path and labels, if not found
+it will try to find the positon using path and empty labels to use as a start offset.
```
thampiotr:
```suggestion
When `legacy_positions_file` is set, Alloy will try to find previous positions for a given file by matching the path and labels, falling back to matching on path only if no match is found.
```

CHANGELOG.md
```diff
@@ -219,6 +219,8 @@ v1.10.0
 
 - Add `tail_from_end` argument to `loki.source.podlogs` to optionally start reading from the end of a log stream for newly discovered pods. (@harshrai654)
 
+- Remove limitation in `loki.source.file` when `legacy_position_file` is uset. Can now recover position even if labels are added before. (@kalleep)
```
```suggestion
- Remove limitation in `loki.source.file` when `legacy_position_file` is unset. Alloy can now recover legacy positions even if labels are added. (@kalleep)
```

internal/component/loki/source/file/tailer.go
```diff
@@ -181,7 +185,16 @@ func (t *tailer) initRun() (loki.EntryHandler, error) {
 		return nil, fmt.Errorf("failed to get file position: %w", err)
 	}
 
-	// NOTE: The code assumes that if a position is available and that the file is bigger than the position, then
+	// If we translated legacy positions we should try to get position offset without labels
+	// when no other position was matched.
+	if pos == 0 && t.legacyPositionUsed {
+		pos, err = t.positions.Get(t.path, "{}")
+		if err != nil {
+			return nil, fmt.Errorf("failed to get file position: %w", err)
```
thampiotr:
```suggestion
			return nil, fmt.Errorf("failed to get file position with empty labels: %w", err)
```

thampiotr (approved):
some minor comments, but LGTM

PR #4304, opened 2025-08-29
fix: convert journal legacy position

kalleep:
#### PR Description
When transitioning from either promtail or agent static there is no way to pick up the old position file in legacy format.

This pr address this by adding a new function that will read the legacy position file and try to find the journal cursor belonging to the old job and translate it to the new cursor used by alloy.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/positions/positions.go
+++ b/internal/component/common/loki/positions/positions.go
@@ -134,6 +134,47 @@ func ConvertLegacyPositionsFile(legacyPath, newPath string, l log.Logger) {
 	level.Info(l).Log("msg", "successfully converted legacy positions file to the new format", "path", newPath, "legacy_path", legacyPath)
 }
 
+// ConvertLegacyPositionsFileJournal will convert the legacy positions file to the new format for a journal job if:
+// 1. There is no file at the newpath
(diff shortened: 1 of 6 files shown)


internal/component/common/loki/positions/positions.go
```diff
@@ -134,6 +134,47 @@ func ConvertLegacyPositionsFile(legacyPath, newPath string, l log.Logger) {
 	level.Info(l).Log("msg", "successfully converted legacy positions file to the new format", "path", newPath, "legacy_path", legacyPath)
 }
 
+// ConvertLegacyPositionsFileJournal will convert the legacy positions file to the new format for a journal job if:
+// 1. There is no file at the newpath
+// 2. There is a file at the legacy path and that it is valid yaml
```
ptodev, line 139: It'd be good to mention this in the doc too

ptodev:
It'd be nice to have a changelog entry too. It could help users who are migrating logs pipelines from agent now.

kalleep:
> It'd be nice to have a changelog entry too. It could help users who are migrating logs pipelines from agent now.

ops

PR #4305, opened 2025-08-29
remote_write: Only use a slice when we have a hash conflict

kgeckhart:
#### PR Description

Syncs code from the upstream memSeries implemetation (https://github.com/prometheus/prometheus/blob/main/tsdb/agent/series.go) to our version to further reduce the memory footprint of the WAL. The idea is that we do not often have hash conflicts for series and having a lot single entry slices is a lot of unnecessary overhead. Only using a slice when there's a conflict should reduce this by a fair amount.

bench results for reference

```
benchstat series-baseline.txt series-new.txt
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/static/metrics/wal
cpu: Apple M3 Pro
                │ series-baseline.txt │           series-new.txt            │
                │       sec/op        │    sec/op     vs base               │
CreateSeries-11           539.8n ± 9%   455.0n ± 11%  -15.70% (p=0.004 n=6)

                │ series-baseline.txt │           series-new.txt           │
                │        B/op         │    B/op      vs base               │
CreateSeries-11            264.0 ± 9%   149.5 ± 33%  -43.37% (p=0.002 n=6)

                │ series-baseline.txt │          series-new.txt           │
                │      allocs/op      │ allocs/op   vs base               │
CreateSeries-11            2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.002 n=6)

```

#### Notes to the Reviewer
After this our there's only a few differences between our implementation and upstream which I will attempt to sync (the WAL bugfix being the most important one). We might be able to switch to using upstream directly here instead but it's not as high of a priority. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,8 @@ Main (unreleased)
 - (_Experimental_) Add an extra parameter to the `array.combine_maps` standard library function
   to enable preserving the first input list even if there is no match. (@ptodev)
 
+- Reduce memory overhead of `prometheus.remote_write`'s WAL by bringing in an upstream change to only track series in a slice if there's a hash conflict. (@kgeckhart)
+
 ### Bugfixes
 
 - Update `webdevops/go-common` dependency to resolve concurrent map write panic. (@dehaansa)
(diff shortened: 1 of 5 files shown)


kalleep:
Is something missing in the pr, or are you only adding tests and benchmarks?

kgeckhart:
> Is something missing in the pr, or are you only adding tests and benchmarks?

🤦 thanks, should be fixed now

dehaansa (approved):

PR #4306, opened 2025-08-29
remote_write: Reduce stripe series size by 4x

kgeckhart:
#### PR Description
Reduce the size of the stripe series to be 4x less than the default prometheus value.

benchstat
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/static/metrics/wal
cpu: Apple M3 Pro
                    │ size-baseline.txt │            size-mid.txt            │
                    │      sec/op       │   sec/op     vs base               │
StripeSeriesSize-11       1523.7µ ± 33%   265.6µ ± 3%  -82.57% (p=0.002 n=6)

                    │ size-baseline.txt │            size-mid.txt             │
                    │       B/op        │     B/op      vs base               │
StripeSeriesSize-11       3712.1Ki ± 0%   928.0Ki ± 0%  -75.00% (p=0.002 n=6)

                    │ size-baseline.txt │            size-mid.txt            │
                    │     allocs/op     │  allocs/op   vs base               │
StripeSeriesSize-11         49.16k ± 0%   12.29k ± 0%  -74.99% (p=0.002 n=6)
```

#### Notes to the Reviewer

When analyzing profiles from the prometheus v3 upgrade @bboreham kindly pointed out that an incredibly large amount of of inuse memory is coming from the `stripeSeries` and is directly influenced by the `stripeSeries` size.
image

Line 104 is where we allocate all the locks for the stripe series. We allocate this once per remote write component in use which for certain use cases can be a lot. As Bryan shared, the size of the `stripeSeries` is to help ensure there is minimal lock contention when using the WAL but it's likely even prometheus's default value is too high (https://github.com/prometheus/prometheus/issues/17100). 

His PR to reduce it even lower to 256 https://github.com/prometheus/prometheus/pull/17101 showed some increase in contention so I chose to keep up much higher where we should still see a dramatic drop off while an ideal upstream value is determined.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


internal/static/metrics/wal/wal_test.go
```diff
@@ -568,6 +568,24 @@ func BenchmarkAppendExemplar(b *testing.B) {
...
+// goos: darwin
+// goarch: arm64
+// pkg: github.com/grafana/alloy/internal/static/metrics/wal
+// cpu: Apple M3 Pro
+// BenchmarkStripeSeriesSize-11                4521            264358 ns/op          950286 B/op      12292 allocs/op
+// BenchmarkStripeSeriesSize-11                4574            272891 ns/op          950287 B/op      12292 allocs/op
+// BenchmarkStripeSeriesSize-11                4414            272839 ns/op          950287 B/op      12292 allocs/op
+// BenchmarkStripeSeriesSize-11                4282            264746 ns/op          950286 B/op      12292 allocs/op
+// BenchmarkStripeSeriesSize-11                4428            265390 ns/op          950287 B/op      12292 allocs/op
+// BenchmarkStripeSeriesSize-11                4327            265847 ns/op          950286 B/op      12292 allocs/op
```
bboreham: Personally I wouldn't put this in the code; I'm not clear what someone reading the code is supposed to do with them.  The before/after comparison in the PR description is fine to justify the change.

PR #4098, opened 2025-07-31
Fix metadata propagation from scrape to otel components

ptodev:
#### PR Description

Fixes propagation of metric metadata, but only for pipelines which start with a prometheus.scrape and transition into otelcol components. The metrics need to be sent by an otelcol exporter. 

I can see the metadata in live debugging. Before:

Screenshot 2025-07-31 at 18 26 58

After:

Screenshot 2025-07-31 at 18 39 12

#### Which issue(s) this PR fixes

Related to [#547](https://github.com/grafana/alloy/issues/547)

#### Notes to the Reviewer

I need to assess the performance impact of this feature before writing the docs and before opening the PR for review.
For now it'd be good to keep it as an experimental opt-in feature until we're sure it's stable and without consuming much more resources.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated
- [ ] Config converters updated

PR #4309, opened 2025-08-29
Support a histogram which is both classic and native

ptodev:
There are several changes in this PR, applied in this order:
1. Porting of the latest prometheus receiver code from upstream. The code in the `internal` folder is mostly just copy pasted from contrib. The most important change this brings in is a [bugfix](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39451) which supports mixed histograms. Those are histograms which are both native and classic.
2. @krajorama fixed the translation code in the `internal` folder so that it doesn't rely on metadata so much. The current upstream code doesn't work for Alloy because it assumes metadata is always available. In Alloy this is usually not the case because metadata support is still experimental. Krajo also opened a [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43833) to apply the same fix upstream.
3. Partially roll back #4098, because apparently we don't need `prometheus.scrape` to set its own metadata store for the context. Apparently, the scrape loop in the Prometheus code can add a metadata store for the context if we set `PassMetadataInContext` to `true` in the scrape settings. This is how Contrib does it. I didn't realise this back when I worked on #4098.

### DIFF
--- a/internal/component/prometheus/scrape/scrape_test.go
+++ b/internal/component/prometheus/scrape/scrape_test.go
@@ -572,9 +572,8 @@ func setupTestNativeHistogram() prometheus_client.Histogram {
 // setupTestMixedHistogram creates and initializes a test histogram with both classic and native buckets
 func setupTestMixedHistogram() prometheus_client.Histogram {
 	mixedHistogram := prometheus_client.NewHistogram(prometheus_client.HistogramOpts{
-		Name: "test_mixed_histogram",
-		Help: "A test histogram metric with both classic and native buckets",
-		// TODO: This is not a classic histogram?
+		Name:    "test_mixed_histogram",
+		Help:    "A test histogram metric with both classic and native buckets",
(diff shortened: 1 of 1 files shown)


ptodev:
Hi @krajorama, do you mind signing the CLA please?

internal/component/prometheus/scrape/scrape_test.go
```diff
@@ -747,6 +569,27 @@ func setupTestNativeHistogram() prometheus_client.Histogram {
 	return nativeHistogram
 }
 
+// setupTestMixedHistogram creates and initializes a test histogram with both classic and native buckets
+func setupTestMixedHistogram() prometheus_client.Histogram {
+	mixedHistogram := prometheus_client.NewHistogram(prometheus_client.HistogramOpts{
+		Name: "test_mixed_histogram",
+		Help: "A test histogram metric with both classic and native buckets",
+		// TODO: This is not a classic histogram?
```
krajorama: Yes it is (also) a classic histogram, as soon as you've added buckets.

```suggestion
```

kalleep (approved):
LGTM

PR #4311, opened 2025-08-29
Add remote write integration test

kgeckhart:
#### PR Description
This PR adds remote write to the existing prom scrape tests, adds the ability to set a test timeout when running integration tests (who has 90s to wait for a failure 😆 ), and some minor changes to assert error messages for more troubleshooting context. 

#### PR Checklist

- [x] Tests updated


internal/cmd/integration-tests/common/common.go
```diff
@@ -14,8 +14,10 @@ type Unmarshaler interface {
 	Unmarshal([]byte) error
 }
 
-const DefaultRetryInterval = 100 * time.Millisecond
-const DefaultTimeout = 90 * time.Second
+const (
+	DefaultRetryInterval = 100 * time.Millisecond
+	DefaultTimeout       = 90 * time.Second
+)
```
jharvey10, line 20: Just curious: Is this type of formatting adjustment something you did/do by hand? Is it a goland feature? I'm all for linting/formatting and would love to auto-correct even more stuff based on a set of conventions we have.

jharvey10:
Am I correct that in this PR, the new configurable timeout isn't yet utilized? If so, is there any place where it would make sense to leverage it as part of this PR?

kgeckhart:
> Am I correct that in this PR, the new configurable timeout isn't yet utilized? If so, is there any place where it would make sense to leverage it as part of this PR?

The configurable timeout is really helpful for fast local iteration with `--stateful`. Locally I can get away with a ~10s timeout as I'm writing new tests. I don't know that I really want to override the default anywhere right now since that will drive CI where resources are more constrained/unreliable.

internal/cmd/integration-tests/common/common.go
kgeckhart: I did this one by hand but other repos I have worked in used https://github.com/mvdan/gofumpt which is a bit more strict and would have done this. I think most find golangci-lint to be enough.

internal/cmd/integration-tests/common/env.go
```diff
@@ -5,4 +5,5 @@ var (
 	AlloyStartTimeEnv = "ALLOY_START_TIME_UNIX"
 	AlloyHostEnv      = "ALLOY_HOST"
 	AlloyPortEnv      = "ALLOY_PORT"
+	TestTimeout       = "TEST_TIMEOUT"
```
kalleep: Should we try to be consistent with naming here?

We could prefix all envs with alloy, so `ALLOY_STATEFUL_TEST` and `ALLOY_TEST_TIMEOUT`

kalleep (approved):

internal/cmd/integration-tests/common/env.go
kgeckhart: I ended up doing this because the ones prefixed with `ALLOY_` is info about the alloy container the test will use while `TEST_TIMEOUT` and `STATEFUL_TEST` is test config. I'll change the stateful env so it's prefixed with `TEST_` and add a comment about the prefixing structure.

PR #4314, opened 2025-08-30
fix: normalize attr key in logfmt logger

zry98:
#### PR Description

This fixes invalid logfmt key names (mostly containing space) in Alloy's self logs, causing the "failed to decode logfmt" error while ingesting self logs and parsing them with the `stage.logfmt` stage in `loki.process` component.

Some examples are: `"boringcrypto enabled"=true` from [here](https://github.com/grafana/alloy/blob/e45b01c9a17d4576459b956610e5105f5b1ee896/internal/alloycli/cmd_run.go#L250), `"extracted data"=` from [here](https://github.com/grafana/alloy/blob/e45b01c9a17d4576459b956610e5105f5b1ee896/internal/component/loki/process/stages/regex.go#L154), etc. There are still many cases present in the code, despite the efforts put into fixing some of those in https://github.com/grafana/alloy/pull/3495.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,8 @@ Main (unreleased)
 
 - `loki.source.journal` now supports `legacy_positon` block that can be used to translate Static Agent or Promtail position files. (@kalleep)
 
+- Normalize attr key name in logfmt logger. (@zry98)
+
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


kalleep (approved):
Thank you!

PR #4319, opened 2025-09-01
fix(loki.source.docker): issues around start/stop that can cause panics 

### LINKED ISSUE #4246: panic: WaitGroup reused in loki.source.docker when following logs via docker-socket-proxy
### Component(s)

loki.source.docker

### What's wrong?

Alloy tails Docker logs via a socket proxy (HAProxy). Long-lived .../logs?follow=1 streams intermittently end with unexpected EOF, and eventually Alloy panics with:
`panic: sync: WaitGroup is reused before previous Wait has returned`
The panic originates in internal/component/loki/source/docker/.../dockertarget/target.go while tearing down/restarting a docker log target.

### Steps to reproduce

1. Deploy docker-socket-proxy (HAProxy-based) with at least NETWORKS=1, CONTAINERS=1, INFO=1.
2. Point Alloy’s loki.source.docker and discovery.docker at tcp://docker-socket-proxy:2375.
3. Start a few containers that produce bursts of logs followed by long idle periods.
4. Observe Alloy logs for repeated warnings:
`level=warn msg="could not transfer logs" ... err="unexpected EOF"`
5. After some reconnect churn, Alloy panics and exits with the stack below.

### System information

Linux 6.13.3-1.el9.elrepo.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Feb 17 11:41:29 EST 2025 x86_64 x86_64 x86_64 GNU/Linux

### Software version

registry-1.docker.io/grafana/alloy:v1.10.2

### Configuration

```text
HAProxy configuration - `https://github.com/Tecnativa/docker-socket-proxy/blob/master/haproxy.cfg`

Minimal docker-compose

version: "3.8"

services:
  docker-socket-proxy:
    image: tecnativa/docker-socket-proxy:latest
    environment:
      CONTAINERS: "1"
      NETWORKS: "1"
      INFO: "1"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    ports:
      - "127.0.0.1:2375:2375"
    restart: unless-stopped

  alloy:
    image: registry-1.docker.io/grafana/alloy:v1.10.2
    command: ["run", "--server.http.listen-addr=0.0.0.0:12345", "/etc/alloy/config.alloy"]
    volumes:
      - ./config.alloy:/etc/alloy/config.alloy:ro
    depends_on:
      - docker-socket-proxy
    restart: unless-stopped


Alloy configuration


discovery.docker "targets" {
  host             = "tcp://docker-socket-proxy:2375"
  refresh_interval = "30s"
}

loki.source.docker "logs" {
  host       = "tcp://docker-socket-proxy:2375"
  targets    = discovery.docker.targets.targets
}

loki.write "default" {
  endpoint {
    url = "http://loki:3100/loki/api/v1/push"
  }
}
```

### Logs

```text
ts=2025-08-24T04:33:49.460814809Z level=warn msg="could not transfer logs" component_path=/ component_id=loki.source.docker.docker target=docker/9e6368ba17d8427784065e996ab056ed75d06072261ac6a9341c82e57aca5fcb written=0 container=9e6368ba17d8427784065e996ab056ed75d06072261ac6a9341c82e57aca5fcb err="unexpected EOF" panic: sync: WaitGroup is reused before previous Wait has returned goroutine 1351878 [running]: sync.(*WaitGroup).Wait(0xc001fdde48?) /usr/local/go/src/sync/waitgroup.go:120 +0x74 github.com/grafana/alloy/internal/component/loki/source/docker/internal/dockertarget.(*Target).Stop(0xc002339e10) /src/alloy/internal/component/loki/source/docker/internal/dockertarget/target.go:245 +0x4a github.com/grafana/alloy/internal/component/loki/source/docker/internal/dockertarget.(*Target).processLoop.func1.1() /src/alloy/internal/component/loki/source/docker/internal/dockertarget/target.go:130 +0x6f github.com/grafana/alloy/internal/component/loki/source/docker/internal/dockertarget.(*Target).processLoop.func1() /src/alloy/internal/component/loki/source/docker/internal/dockertarget/target.go:144 +0x3e7 created by github.com/grafana/alloy/internal/component/loki/source/docker/internal/dockertarget.(*Target).processLoop in goroutine 1351892 /src/alloy/internal/component/loki/source/docker/internal/dockertarget/target.go:125 +0x8ad
```

### Tip

kalleep:
#### PR Description
From (*WaitGroup).Add docs
```
// Note that calls with a positive delta that occur when the counter is zero
// must happen before a Wait. Calls with a negative delta, or calls with a
// positive delta that start when the counter is greater than zero, may happen
// at any time.
```

We have two cases where stop can be called. From the [scheduled task](https://github.com/grafana/alloy/blob/60a0b54ae311fb9a7cdd2f6618957079ac04cc03/internal/component/loki/source/docker/runner.go#L123) or if we get an error when (including EOF) [tailing logs via api ](https://github.com/grafana/alloy/blob/60a0b54ae311fb9a7cdd2f6618957079ac04cc03/internal/component/loki/source/docker/internal/dockertarget/target.go#L130).

We also periodically check if container is available and call [StartIfNotRunning](https://github.com/grafana/alloy/blob/60a0b54ae311fb9a7cdd2f6618957079ac04cc03/internal/component/loki/source/docker/runner.go#L119-L121). So the panic can happen if the tailing unexpectedly exists at the same time we run our periodic check, i.e. the waitgroup gets reused before wait call has finished. 

I added a test to quickly call both start and stop that will panic on main branch.

Fixes #4246

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/internal/dockertarget/target.go
+++ b/internal/component/loki/source/docker/internal/dockertarget/target.go
@@ -96,7 +96,9 @@ func (t *Target) processLoop(ctx context.Context) {
 	inspectInfo, err := t.client.ContainerInspect(ctx, t.containerName)
 	if err != nil {
 		level.Error(t.logger).Log("msg", "could not inspect container info", "container", t.containerName, "err", err)
-		t.stopWithError(err)
+		t.mu.Lock()
+		t.err = err
+		t.mu.Unlock()
 		return
 	}
 
@@ -109,7 +111,9 @@ func (t *Target) processLoop(ctx context.Context) {
 	})
 	if err != nil {
 		level.Error(t.logger).Log("msg", "could not fetch logs for container", "container", t.containerName, "err", err)
-		t.stopWithError(err)
+		t.mu.Lock()
+		t.err = err
+		t.mu.Unlock()
 		return
 	}
 	defer logs.Close()
@@ -240,17 +244,8 @@ func (t *Target) StartIfNotRunning() {
 
 // Stop shuts down the target.
 func (t *Target) Stop() {
-	t.stopWithError(nil)
-}
-
-func (t *Target) stopWithError(err error) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
-
-	if err != nil {
-		t.err = err
-	}
-
 	if t.running {
 		t.running = false
 		if t.cancel != nil {


internal/component/loki/source/docker/internal/dockertarget/target.go
```diff
@@ -92,31 +93,26 @@ func NewTarget(metrics *Metrics, logger log.Logger, handler loki.EntryHandler, p
 }
 
 func (t *Target) processLoop(ctx context.Context) {
-	t.running.Store(true)
-	defer t.running.Store(false)
-
-	t.wg.Add(1)
-	defer t.wg.Done()
```
kgeckhart, line 99: Did this get moved elsewhere because I can't find it. I think it was guarding against the waitgroup having no tasks if the processLoop is stopped while waiting for the `t.client.` calls to finish. But even that has a bit of a race condition because the call to processLoop is done from a goroutine 🤔. 

I think we might need the wg.Add but perhaps it should be at the call site vs here. What are your thoughts?

kalleep: I remove it because it's not really needed. proccessLoop is waiting on context to be canceled, that happens whenever Stop is called. After we have canceled the context stop will wait for wg to finish. That can only finish after processLoop have existed because that's when the log tail connection is closed and triggering the inner tasks to exit.

kalleep: yeah you are right in that the previous code had the issue and that add should have been called before `go proccessLoop(ctx)`. Still I don't believe it's need because we pass that same context to the client call, that should cancel the call and we should get an error. Adding the proccessLoop to wg as a task right now could cause a deadlock because call to stopWithError would also try to get the lock and the first call to Stop would already hold the lock and wait for the wg to finish. 

I really don't like this scenario and will try to think of a better way to solve this

kalleep: Yeah if we add proccessLoop to wg we can get a deadlock currently but as I stated before it is not necessary to add it there. We only need to add the "inner" tasks

kgeckhart: Yeah I agree it's not needed and was really odd. The process loop spawns the tasks for the waitgroup it certainly doesn't need to be a participant in it itself.

kgeckhart (approved):
Nice work!

PR #4320, opened 2025-09-01
Add an extra parameter to the `array.combine_maps` standard library function

ptodev:
#### PR Description

This would make `array.combine_maps` more useful when enriching Prometheus Exporter targets with service discovery labels. In many cases everything should still be scraped even if the targets can't be enriched.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/syntax/internal/stdlib/stdlib.go
+++ b/syntax/internal/stdlib/stdlib.go
@@ -318,9 +318,10 @@ func concatMaps(left, right value.Value) (value.Value, error) {
 // args[0]: []map[string]string: lhs array
 // args[1]: []map[string]string: rhs array
 // args[2]: []string:            merge conditions
+// args[3]: bool:                (optional) retain unmatched elements from the lhs array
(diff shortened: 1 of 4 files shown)


syntax/internal/stdlib/stdlib.go
```diff
@@ -376,6 +377,23 @@ var combineMaps = value.RawFunction(func(funcValue value.Value, args ...value.Va
 		}
 	}
 
+	// Validate args[3]
+	passthrough_lhs := false
```
kalleep:
```suggestion
	passthroughLHS := false
```
Underscores are usually not used in variable names in go

kalleep (approved):
LGTM

syntax/internal/stdlib/stdlib.go
ptodev: Oof sorry, my old habit strikes back 😆

PR #4322, opened 2025-09-01
[WIP] feat: Sign Windows executables

martincostello:
#### PR Description

Initial changes to support signing Windows executables before they are attached to the GitHub release.

Because Windows signing has to take place on a Windows machine, this requires the build to be split into different phases. First the binaries are built, then the Windows executables are Authenticode signed, then the signed files are attached to the GitHub release.

This may require further modularisation to allow the files that go inside the installer to also be signed, depending on how flexible the `sign` tool is to recurse into the installer to sign the contents (or not).

https://github.com/grafana/app-o11y/issues/1056

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] ~~Tests updated~~
- [ ] ~~Config converters updated~~

### DIFF
--- a/tools/release
+++ b/tools/release
@@ -3,10 +3,10 @@
 # This script should be run from the root of the repository.
 set -ex
 
-# Zip up all the binaries to reduce the download size. DEBs and RPMs
-# aren't included to be easier to work with.
+# Zip up all the binaries to reduce the download size. DEBs, RPMs and
+# the Windows installer aren't included to be easier to work with.
 find dist/ -type f \
-  -name 'alloy*' -not -name '*.deb' -not -name '*.rpm' \
+  -name 'alloy*' -not -name '*.deb' -not -name '*.rpm' -not -name 'alloy-installer-windows-*.exe' \
   -exec zip -j -m "{}.zip" "{}" \;
 
 # Get the SHA256SUMS before continuing.
--- a/.github/workflows/publish-alloy-release.yml
+++ b/.github/workflows/publish-alloy-release.yml
@@ -1,48 +1,189 @@
 name: Publish alloy release containers
 on:
-  push:
-    tags:
-      - v*
+  # TODO Remove once tested
+  pull_request:
+    branches: [main]
+  #push:
+  #  tags:
+  #    - v*
 
 permissions:
   contents: read
 
 jobs:
-  publish_linux_container:
-    uses: ./.github/workflows/publish-alloy-linux.yml
-    permissions:
-      contents: read
-      id-token: write
-    with:
-      img-name: alloy
+  # TODO Restore once tested
+  #publish_linux_container:
+  #  uses: ./.github/workflows/publish-alloy-linux.yml
+  #  permissions:
+  #    contents: read
+  #    id-token: write
+  #  with:
+  #    img-name: alloy
   
-  publish_linux_boringcrypto_container:
-    uses: ./.github/workflows/publish-alloy-linux.yml
-    permissions:
-      contents: read
-      id-token: write
-    with:
-      img-name: alloy-boringcrypto
+  #publish_linux_boringcrypto_container:
+  #  uses: ./.github/workflows/publish-alloy-linux.yml
+  #  permissions:
+  #    contents: read
+  #    id-token: write
+  #  with:
+  #    img-name: alloy-boringcrypto
 
-  publish_windows_container:
-    uses: ./.github/workflows/publish-alloy-windows.yml
+  #publish_windows_container:
+  #  uses: ./.github/workflows/publish-alloy-windows.yml
+  #  permissions:
+  #    contents: read
+  #    id-token: write
+  #  with:
+  #    img-name: alloy
+
+  build_alloy:
+    name: Build Alloy
+    container: grafana/alloy-build-image:v0.1.21@sha256:36b1e621f32f46aa1de58843b885f0885faabe7b29468462d36a60bea1245e33
+    runs-on:
+      labels: github-hosted-ubuntu-x64-large
+    # TODO Restore once tested
+    #needs:
+    #- publish_linux_container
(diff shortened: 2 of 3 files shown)


packaging/windows/install_script.nsis
```diff
@@ -289,3 +289,11 @@ Section "uninstall"
...
   DeleteRegKey HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\${APPNAME}"
 SectionEnd
+
+!if '{{ SIGNING_COMMAND }}' != ''
+    # Signing for installer and uninstaller; nsis 3.08 required for uninstfinalize!
+    # "= 0" comparison required to prevent both tasks running in parallel, which would cause signing to fail
+    # %1 is replaced by the installer and uninstaller paths, respectively
+    !finalize '{{ SIGNING_COMMAND }} "%1"' = 0
+    !uninstfinalize '{{ SIGNING_COMMAND }} "%1"' = 0
+!endif
```
martincostello: To be able to use this, the installer would have to be built on Windows instead of the custom Linux container.

Until then, we cannot sign the uninstaller as we need to sign it before it is embedded in the installer.

martincostello:
```suggestion
#!if '{{ SIGNING_COMMAND }}' != ''
#    # Signing for installer and uninstaller; nsis 3.08 required for uninstfinalize!
#    # "= 0" comparison required to prevent both tasks running in parallel, which would cause signing to fail
#    # %1 is replaced by the installer and uninstaller paths, respectively
#    !finalize '{{ SIGNING_COMMAND }} "%1"' = 0
#    !uninstfinalize '{{ SIGNING_COMMAND }} "%1"' = 0
#!endif
```

tools/signing/authenticode.ps1
```diff
@@ -0,0 +1,55 @@
...
+    [Parameter(Mandatory = $false)][string] $DescriptionUrl,
+    [Parameter(Mandatory = $false)][string] $Publisher = "Grafana Labs"
+)
+
+$ErrorActionPreference = "Stop"
+$InformationPreference = "Continue"
+$ProgressPreference = "SilentlyContinue"
+
+if ([string]::IsNullOrEmpty($Description)) {
+    $Description = "Grafana Alloy is an OpenTelemetry Collector distribution with programmable pipelines."
```
martincostello: This is quite wordy, so I think just `Grafana Alloy` would look better:

image

martincostello:
Unfortunately the latest approach doesn't work as I thought it might.

Building the installer rebuilds the `.exe` files, which causes the signed versions that are downloaded from the previous job to be overwritten.

martincostello:
Confirmed with [this build](https://github.com/grafana/alloy/actions/runs/17433171676?pr=4322) that all the `*-signed` artifacts are signed, and that the installer itself contains the same signed assets (except the uninstaller).

martincostello:
One of the fuzz tests [failed](https://github.com/grafana/alloy/actions/runs/17435215316/attempts/1?pr=4322), but that can't be caused by these changes.

tools/release
```diff
@@ -3,10 +3,10 @@
 # This script should be run from the root of the repository.
 set -ex
 
-# Zip up all the binaries to reduce the download size. DEBs and RPMs
-# aren't included to be easier to work with.
+# Zip up all the binaries to reduce the download size. DEBs, RPMs and
+# the Windows installer aren't included to be easier to work with.
 find dist/ -type f \
-  -name 'alloy*' -not -name '*.deb' -not -name '*.rpm' \
+  -name 'alloy*' -not -name '*.deb' -not -name '*.rpm' -not -name 'alloy-installer-windows-*.exe' \
```
ptodev: This [was changed in the past](https://github.com/grafana/agent/issues/2887) and people complained. I'd prefer to continue zipping up the installer. It'd help prevent breaking changes in case someone has a workflow somewhere to pull it as a zip.

We could consider not zipping it, but it'd be best to do it in a major version or in a way that we still have the zip for a version or two so that people can migrate. Anyway, I think it's fine to just continue with a zipped installer.

martincostello:
Temporarily moved back to draft to refactor to use the shared action.

kalleep:
> One of the fuzz tests [failed](https://github.com/grafana/alloy/actions/runs/17435215316/attempts/1?pr=4322), but that can't be caused by these changes.

This test is a bit flaky, I looked into that one a bit here https://github.com/grafana/alloy/pull/3657#issuecomment-2909084528 but I don't think we have done a proper workaround for the limitations yet :(

martincostello:
Confirmed with [this build](https://github.com/grafana/alloy/actions/runs/17464903665?pr=4322) that the `*-signed` artifacts are all signed, and that the Windows installer contains the signed Alloy executables.

tools/release
martincostello: I can revert this change if necessary, but I was lead to believe that the only reason it was zipped was to avoid warnings from downloading an unsigned `.exe` file.

As the file is now signed, that shouldn't be a problem and it's then one less thing for users to do to run the installer when downloading it from the release.

It also provides a simple download URL, meaning once the first signed release is shipped, I can look to create a [winget](https://learn.microsoft.com/windows/package-manager/winget/) installer package to make it easier for users to acquire alloy on modern Windows systems.

If the ZIP is still wanted, having _both_ would be best.

martincostello: Also worth noting that this is only for the installer, not the service as well.

The installer is already highly compressed and the saving from ZIP-ing it is negligible using v1.10.2 for comparison:

- Installer ZIP - 83,694 KB
- Installer - 85,519 KB

thampiotr: Could we publish both? 
Unfortunately we are committed to not make breaking changes and users may have written scripts that will rely on .zip. We don't want to break them.

PR #4324, opened 2025-09-01
database_observability: minor updates to README

cristiangreco:
#### PR Description

- improve docs about mysql "setup_consumers", avoid referencing metrics when Alloy is not yet up and running
- add `cloud_providers` block in the examples
- other minor changes

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/README.md
+++ b/internal/component/database_observability/README.md
@@ -74,7 +74,7 @@ SHOW VARIABLES LIKE 'performance_schema_max_digest_length';
 +--------------------------------------+-------+
 ```
 
-6. [OPTIONAL] Enable the `events_statements_cpu` consumer if you want to capture query samples. Verify the current setting with a sql query:
+6. [OPTIONAL] Enable the `events_statements_cpu` consumer if you want to capture CPU activity and time on query samples. Verify the current setting with a sql query:
 
 ```sql
 SELECT * FROM performance_schema.setup_consumers WHERE NAME = 'events_statements_cpu';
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/README.md
```diff
@@ -430,6 +428,24 @@ prometheus.exporter.postgres "integrations_postgres_exporter_<your_DB_name>" {
...
+    disable_query_redaction = true
+  }
+
+  // OPTIONAL: provide additional information specific to the Cloud Provider
+  // that hosts the database to enable certain infrastructure observability features.
+  cloud_provider {
+    aws {
+      arn = "your-rds-db-arn"
+    }
+  }
```
cristiangreco, line 450: This is not yet implemented for postgres, but a followup in our todo list so I thought we might just add it now so we don't forget.

internal/component/database_observability/README.md
```diff
@@ -72,13 +74,7 @@ SHOW VARIABLES LIKE 'performance_schema_max_digest_length';
...
 ```
 
-6. Optionally enable the `events_statements_cpu` consumer if you want to capture query samples. Verify the current setting:
-
-```promql
-database_observability_setup_consumers_enabled{job="integrations/db-o11y", consumer_name="events_statements_cpu"}
-```
-
-or with a sql query:
+6. [OPTIONAL] Enable the `events_statements_cpu` consumer if you want to capture query samples. Verify the current setting with a sql query:
```
matthewnolf: This consumer gives us CPU time / activity on query samples - but query samples themselves are still available without it enabled. Perhaps we can be more clear here and suggest something like:

```
Enable the `events_statements_cpu` consumer if you want to capture CPU activity and time on query samples
```

cristiangreco: Updated

PR #4335, opened 2025-09-03
database_observability: improve error handling for MySQL and Postgres

gaantunes:

This is improving error handling for both `database_observability.mysql` and `database_observability.postgres` components, by not returning errors arising when the collectors are started, but rather logging it and reporting unhealthy.

Additionally, a best-effort approach was implemented, so that if one collector fails it does not stop other collectors to start, if possible.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/component.go
+++ b/internal/component/database_observability/mysql/component.go
@@ -3,7 +3,6 @@ package mysql
 import (
 	"context"
 	"database/sql"
-	"errors"
 	"fmt"
 	"net/http"
 	"path"
@@ -51,6 +50,11 @@ func init() {
 var (
 	_ syntax.Defaulter = (*Arguments)(nil)
 	_ syntax.Validator = (*Arguments)(nil)
+)
+
+// test-time injection points (overridden in tests for mocking)
+var (
+	mysqlOpenSQL = sql.Open
 )
 
 type Arguments struct {
@@ -274,42 +278,10 @@ func (c *Component) Update(args component.Arguments) error {
 
 	c.args = args.(Arguments)
 
-	dbConnection, err := sql.Open("mysql", formatDSN(string(c.args.DataSourceName), "parseTime=true"))
-	if err != nil {
-		return err
-	}
-
-	if dbConnection == nil {
-		return errors.New("nil DB connection")
-	}
-	if err = dbConnection.Ping(); err != nil {
-		return err
-	}
-	c.dbConnection = dbConnection
-
-	rs := c.dbConnection.QueryRowContext(context.Background(), selectServerInfo)
-	if err = rs.Err(); err != nil {
-		level.Error(c.opts.Logger).Log("msg", "failed to query engine version", "err", err)
-		return err
-	}
-
-	var serverUUID, engineVersion string
-	if err := rs.Scan(&serverUUID, &engineVersion); err != nil {
-		level.Error(c.opts.Logger).Log("msg", "failed to scan engine version", "err", err)
-		return err
-	}
-
-	c.args.Targets = append([]discovery.Target{c.baseTarget}, c.args.Targets...)
-	targets := make([]discovery.Target, 0, len(c.args.Targets)+1)
-	for _, t := range c.args.Targets {
-		builder := discovery.NewTargetBuilderFrom(t)
-		if relabel.ProcessBuilder(builder, database_observability.GetRelabelingRules(serverUUID)...) {
-			targets = append(targets, builder.Target())
-		}
-	}
-
+	// Always export at least the base target so the component can expose metrics
+	// even if the database connection is currently unavailable.
 	c.opts.OnStateChange(Exports{
-		Targets: targets,
+		Targets: []discovery.Target{c.baseTarget},
 	})
 
 	for _, collector := range c.collectors {
@@ -317,9 +289,9 @@ func (c *Component) Update(args component.Arguments) error {
 	}
 	c.collectors = nil
 
-	if err := c.startCollectors(serverUUID, engineVersion); err != nil {
+	if err := c.startCollectors(); err != nil {
 		c.healthErr.Store(err.Error())
-		return err
+		return nil
 	}
 
 	c.healthErr.Store("")
@@ -351,12 +323,58 @@ func enableOrDisableCollectors(a Arguments) map[string]bool {
 	return collectors
 }
 
-func (c *Component) startCollectors(serverUUID string, engineVersion string) error {
+func (c *Component) startCollectors() error {
+	// Establish DB connection and fetch server info here so Update remains thin.
+	dbConnection, err := mysqlOpenSQL("mysql", formatDSN(string(c.args.DataSourceName), "parseTime=true"))
+	if err != nil {
+		err = fmt.Errorf("failed to start collectors: failed to open MySQL connection: %w", err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		return err
+	}
+	if dbConnection == nil {
+		err = fmt.Errorf("failed to start collectors: nil DB connection")
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		return err
+	}
+	if err = dbConnection.Ping(); err != nil {
+		err = fmt.Errorf("failed to start collectors: failed to ping MySQL: %w", err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		return err
+	}
+	c.dbConnection = dbConnection
+
+	rs := c.dbConnection.QueryRowContext(context.Background(), selectServerInfo)
+	if err = rs.Err(); err != nil {
+		err = fmt.Errorf("failed to query engine version: %w", err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		return err
+	}
+
+	var serverUUID, engineVersion string
+	if err := rs.Scan(&serverUUID, &engineVersion); err != nil {
+		err = fmt.Errorf("failed to scan engine version: %w", err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		return err
+	}
+
+	// Update exported targets based on server UUID relabeling
+	c.args.Targets = append([]discovery.Target{c.baseTarget}, c.args.Targets...)
+	targets := make([]discovery.Target, 0, len(c.args.Targets)+1)
+	for _, t := range c.args.Targets {
+		builder := discovery.NewTargetBuilderFrom(t)
+		if relabel.ProcessBuilder(builder, database_observability.GetRelabelingRules(serverUUID)...) {
+			targets = append(targets, builder.Target())
+		}
+	}
+	c.opts.OnStateChange(Exports{
+		Targets: targets,
+	})
 	var cloudProviderInfo *database_observability.CloudProvider
 	if c.args.CloudProvider != nil && c.args.CloudProvider.AWS != nil {
 		arn, err := arn.Parse(c.args.CloudProvider.AWS.ARN)
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to parse AWS cloud provider ARN", "err", err)
+			err = fmt.Errorf("failed to parse AWS cloud provider ARN: %w", err)
+			level.Error(c.opts.Logger).Log("msg", err.Error())
 			return err
 		}
 		cloudProviderInfo = &database_observability.CloudProvider{
@@ -370,6 +388,9 @@ func (c *Component) startCollectors(serverUUID string, engineVersion string) err
 
 	collectors := enableOrDisableCollectors(c.args)
 
+	// Best-effort start: try building/starting every enabled collector and aggregate errors.
+	var startErrors []string
+
 	if collectors[collector.QueryTablesName] {
 		qtCollector, err := collector.NewQueryTables(collector.QueryTablesArguments{
 			DB:              c.dbConnection,
@@ -378,14 +399,18 @@ func (c *Component) startCollectors(serverUUID string, engineVersion string) err
 			Logger:          c.opts.Logger,
 		})
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to create QueryTable collector", "err", err)
-			return err
+			err = fmt.Errorf("failed to create %s collector: %w", collector.QueryTablesName, err)
+			level.Error(c.opts.Logger).Log("msg", err.Error())
+			startErrors = append(startErrors, err.Error())
+		} else {
+			if err := qtCollector.Start(context.Background()); err != nil {
+				err = fmt.Errorf("failed to start %s collector: %w", collector.QueryTablesName, err)
+				level.Error(c.opts.Logger).Log("msg", err.Error())
+				startErrors = append(startErrors, err.Error())
+			} else {
+				c.collectors = append(c.collectors, qtCollector)
+			}
 		}
-		if err := qtCollector.Start(context.Background()); err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to start QueryTable collector", "err", err)
-			return err
-		}
-		c.collectors = append(c.collectors, qtCollector)
 	}
 
 	if collectors[collector.SchemaTableName] {
@@ -400,14 +425,18 @@ func (c *Component) startCollectors(serverUUID string, engineVersion string) err
 			Logger:       c.opts.Logger,
 		})
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to create SchemaTable collector", "err", err)
-			return err
+			err = fmt.Errorf("failed to create %s collector: %w", collector.SchemaTableName, err)
+			level.Error(c.opts.Logger).Log("msg", err.Error())
+			startErrors = append(startErrors, err.Error())
+		} else {
+			if err := stCollector.Start(context.Background()); err != nil {
+				err = fmt.Errorf("failed to start %s collector: %w", collector.SchemaTableName, err)
+				level.Error(c.opts.Logger).Log("msg", err.Error())
+				startErrors = append(startErrors, err.Error())
+			} else {
+				c.collectors = append(c.collectors, stCollector)
+			}
 		}
-		if err := stCollector.Start(context.Background()); err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to start SchemaTable collector", "err", err)
-			return err
-		}
-		c.collectors = append(c.collectors, stCollector)
 	}
 
 	if collectors[collector.QuerySampleName] {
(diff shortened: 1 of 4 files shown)


internal/component/database_observability/mysql/component.go
```diff
@@ -274,52 +278,20 @@ func (c *Component) Update(args component.Arguments) error {
 
 	c.args = args.(Arguments)
 
-	dbConnection, err := sql.Open("mysql", formatDSN(string(c.args.DataSourceName), "parseTime=true"))
```
gaantunes, line 284: moving this chunk of code into the startCollectors func.

internal/component/database_observability/mysql/component.go
```diff
@@ -53,6 +52,11 @@ var (
 	_ syntax.Validator = (*Arguments)(nil)
 )
 
+// test-time injection points (overridden in tests for mocking)
+var (
```
matthewnolf: having this global state means tests can interfere with each other. I'd suggest that we make this a parameter to whatever needs it and then inject it that way

internal/component/database_observability/mysql/component.go
matthewnolf: What is the reason though? I'm not sure it adds anything by moving this other than having to update Targets twice

internal/component/database_observability/mysql/component.go
```diff
@@ -351,12 +323,58 @@ func enableOrDisableCollectors(a Arguments) map[string]bool {
 	return collectors
 }
 
-func (c *Component) startCollectors(serverUUID string, engineVersion string) error {
+func (c *Component) startCollectors() error {
+	// Establish DB connection and fetch server info here so Update remains thin.
+	dbConnection, err := mysqlOpenSQL("mysql", formatDSN(string(c.args.DataSourceName), "parseTime=true"))
+	if err != nil {
+		err = fmt.Errorf("failed to start collectors: failed to open MySQL connection: %w", err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
```
matthewnolf: Do we need to log here? Is the error not logged after it is propagated?

internal/component/database_observability/mysql/component.go
```diff
@@ -378,14 +399,18 @@ func (c *Component) startCollectors(serverUUID string, engineVersion string) err
 			Logger:          c.opts.Logger,
 		})
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to create QueryTable collector", "err", err)
-			return err
-		}
-		if err := qtCollector.Start(context.Background()); err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to start QueryTable collector", "err", err)
-			return err
+			err = fmt.Errorf("failed to create %s collector: %w", collector.QueryTablesName, err)
```
matthewnolf: There is a fair bit of duplication here with the error handling between collectors, can we add a small closure that captures that logic and re-use for each collector

internal/component/database_observability/mysql/component_test.go
```diff
@@ -276,3 +283,95 @@ func Test_addLokiLabels(t *testing.T) {
...
+			return http_service.Data{MemoryListenAddr: "127.0.0.1:0", BaseHTTPPath: "/component"}, nil
+		},
+		OnStateChange: func(e cmp.Exports) { gotExports = e },
+	}
+	c, err := New(opts, args)
+	require.NoError(t, err)
+	exported, ok := gotExports.(Exports)
+	require.True(t, ok)
+	require.Len(t, exported.Targets, 1)
+	time.Sleep(10 * time.Millisecond)
```
matthewnolf: Why sleeping here? What is not deterministic?

internal/component/database_observability/mysql/component.go
gaantunes: Code organization and readability. We were starting collectors out of the startCollectors function, which did not start any collector.

internal/component/database_observability/mysql/component.go
gaantunes: Yes, I see this issue in other places as well, checking all logs to make sure we do it cleanly.

internal/component/database_observability/mysql/component_test.go
gaantunes: Good catch, that is code left behind from another approach I was taking.

internal/component/database_observability/mysql/component.go
gaantunes: Will do.

internal/component/database_observability/mysql/collector/explain_plan.go
```diff
@@ -473,7 +473,7 @@ func (c *ExplainPlan) Start(ctx context.Context) error {
 
 		for {
 			if err := c.fetchExplainPlans(c.ctx); err != nil {
-				level.Error(c.logger).Log("msg", "collector error", "err", err)
+				level.Error(c.logger).Log("msg", ExplainPlanName+" collector error", "err", err)
```
matthewnolf: The logger already has this as a label https://github.com/grafana/alloy/pull/4335/files#diff-cc02483faf7ed483c3a3a921ade523b48e22702b5609e4cee5461cf02fb362fdL449

internal/component/database_observability/mysql/component.go
matthewnolf: This function doesn't start any collectors, it just opens the SQL connection. The concern I have here is that `Targets` will not have the `server_id` attached as a label through the relabel rules anymore which could cause issues on the query side.

internal/component/database_observability/mysql/collector/explain_plan.go
gaantunes: Ugh completely missed it. Thanks for the head up.

internal/component/database_observability/mysql/component.go
gaantunes: Ok got it.
So, the idea of this initial target is to spin up just the connection info collector with the gauge set to `0` when we can't establish the connection. I believe I have seen this pattern elsewhere, and it should be a good target for an alert.
I changed the code to replace this initial target with the one passing through the relabeling rules.
Let me know what you think of this.

gaantunes: Changed it to be a separate metric, called `connection_up`.
Also only exporting the baseTarget when an error is returned by the startCollectors function, which only happens when the connection was unsuccessful or we could not get the engine version.

matthewnolf:
Could we move the `connection_up` changes to a separate PR? I think it could use some seperate discussion outside of the goal here which is to stop the db_o11y component's crashing alloy

gaantunes:
> Could we move the `connection_up` changes to a separate PR? I think it could use some seperate discussion outside of the goal here which is to stop the db_o11y component's crashing alloy

Simplified it to just ping the connection from time to time and report healthy/unhealthy.
Opening another PR with the other code for discussion after this is merged.

internal/component/database_observability/mysql/component.go
```diff
@@ -274,52 +287,19 @@ func (c *Component) Update(args component.Arguments) error {
...
 	c.collectors = nil
 
-	if err := c.startCollectors(serverUUID, engineVersion); err != nil {
+	if err := c.startCollectors(); err != nil {
 		c.healthErr.Store(err.Error())
-		return err
+		// Export at least the base target so the component can expose metrics
+		// even if the database connection is currently unavailable.
+		c.opts.OnStateChange(Exports{
+			Targets: []discovery.Target{c.baseTarget},
```
matthewnolf: Thanks for moving stuff with the metric. I think we will still end up here with baseTarget not having the `server_id` metric that we're adding with relabels. 
This is a regression as it will mean we produce multiple AS for this DB (one with the server_id, one without).

matthewnolf: Unless while this function is running the targets are not available?

gaantunes: I will remove the baseTarget initialization, and just expose targets if the component is able to connect to the database and start collectors successfully, as it is today.

internal/component/database_observability/mysql/collector/setup_consumers.go
```diff
@@ -78,7 +78,7 @@ func (c *SetupConsumers) Start(ctx context.Context) error {
 
 		for {
 			if err := c.getSetupConsumers(c.ctx); err != nil {
-				level.Error(c.logger).Log("msg", "setupConsumers collector error", "err", err)
+				level.Error(c.logger).Log("msg", "collector error", "err", err)
```
cristiangreco, line 81: nit: while at it, can we change the message to something more specific?

internal/component/database_observability/postgres/collector/schema_table.go
```diff
@@ -185,13 +185,12 @@ func (c *SchemaTable) extractNames(ctx context.Context) error {
...
 	if err := rs.Scan(&dbName); err != nil {
-		level.Error(c.logger).Log("msg", "failed to scan database name", "err", err)
-		return err
+		return fmt.Errorf("failed to scan database name: %w", err)
 	}
 
 	schemaRs, err := c.dbConnection.QueryContext(ctx, selectSchemaNames)
 	if err != nil {
-		level.Error(c.logger).Log("msg", "failed to query pg_namespace", "database", dbName, "err", err)
+		err = fmt.Errorf("failed to query pg_namespace for database %s: %w", dbName, err)
```
cristiangreco: Can we just `return fmt.Errorf(...)`?

internal/component/database_observability/postgres/collector/schema_table.go
```diff
@@ -213,7 +212,7 @@ func (c *SchemaTable) extractNames(ctx context.Context) error {
 	}
 
 	if err := schemaRs.Err(); err != nil {
-		level.Error(c.logger).Log("msg", "error during iterating over pg_namespace result set", "database", dbName, "err", err)
+		err = fmt.Errorf("error during iterating over pg_namespace result set for database %s: %w", dbName, err)
```
cristiangreco: Same as above.

internal/component/database_observability/mysql/component.go
```diff
@@ -272,6 +277,12 @@ func (c *Component) getBaseTarget() (discovery.Target, error) {
 var versionRegex = regexp.MustCompile(`^((\d+)(\.\d+)(\.\d+))`)
 
 func (c *Component) Update(args component.Arguments) error {
+	reportError := func(errorWrapper string, err error) {
+		err = fmt.Errorf("%s: %w", errorWrapper, err)
+		level.Error(c.opts.Logger).Log("msg", err.Error())
+		c.healthErr.Store(err.Error())
+	}
+
```
cristiangreco: I think this could be it's own method, rather than living inside `Update`.

internal/component/database_observability/mysql/component.go
```diff
@@ -272,6 +277,12 @@ func (c *Component) getBaseTarget() (discovery.Target, error) {
 var versionRegex = regexp.MustCompile(`^((\d+)(\.\d+)(\.\d+))`)
 
 func (c *Component) Update(args component.Arguments) error {
+	reportError := func(errorWrapper string, err error) {
```
cristiangreco: nit: maybe `errorMsg` instead `errorWrapper`?

internal/component/database_observability/mysql/component.go
```diff
@@ -369,12 +384,20 @@ func enableOrDisableCollectors(a Arguments) map[string]bool {
 }
 
 func (c *Component) startCollectors(serverUUID string, engineVersion string, parsedEngineVersion semver.Version) error {
+	// Best-effort start: try building/starting every enabled collector and aggregate errors.
+	var startErrors []string
+
+	logStartError := func(collectorName, action string, err error) {
+		wrapped := fmt.Errorf("failed to %s %s collector: %w", action, collectorName, err)
```
cristiangreco: Could use `fmt.Sprintf` instead of creating an error. It seems to have a limitation that you can't use `%w` but it can be replaced with `%+v` I believe.

internal/component/database_observability/mysql/collector/setup_consumers.go
gaantunes: You mean in all collectors or just this one? All collectors are following this pattern.
Do you have a suggestion for a more specific message?

internal/component/database_observability/mysql/component.go
```diff
@@ -369,12 +384,20 @@ func enableOrDisableCollectors(a Arguments) map[string]bool {
...
+		level.Error(c.opts.Logger).Log("msg", wrapped.Error())
+		startErrors = append(startErrors, wrapped.Error())
+	}
+
 	var cloudProviderInfo *database_observability.CloudProvider
 	if c.args.CloudProvider != nil && c.args.CloudProvider.AWS != nil {
 		arn, err := arn.Parse(c.args.CloudProvider.AWS.ARN)
 		if err != nil {
 			level.Error(c.opts.Logger).Log("msg", "failed to parse AWS cloud provider ARN", "err", err)
-			return err
```
cristiangreco, line 377: Hmm wondering if this could be added to the list of errors as well, although shouldn't prevent startup.

internal/component/database_observability/mysql/component_test.go
```diff
@@ -276,3 +284,86 @@ func Test_addLokiLabels(t *testing.T) {
...
+
+// TestMySQL_StartCollectors_ReportsUnhealthy_StackedErrors tests that the component tries to start collectors on a best effort basis,
+// reports unhealthy stacking errors for the collectors that failed to start and generate metrics for the collectors that started successfully.
+func TestMySQL_StartCollectors_ReportsUnhealthy_StackedErrors(t *testing.T) {
+	t.Parallel()
+
+	args := Arguments{
+		DataSourceName:    "user:pass@tcp(127.0.0.1:3306)/db",
+		DisableCollectors: []string{"query_details", "schema_details", "setup_consumers", "explain_plans"},
+		EnableCollectors:  []string{"query_samples", "locks"},
```
cristiangreco, line 316: Ideally we should have some sort of mock collector for these tests (it came up in a previous conversation with Matt as well). Fine for now however.

internal/component/database_observability/mysql/component.go
gaantunes: Initially StartErrors are limited to those that prevent the collector from starting. Other errors are just logged and we move forward. Perhaps this should be a `Warning` level log, since it does not stop the collector from starting successfully.

internal/component/database_observability/mysql/component_test.go
gaantunes: Noted. I want to work on it further afterwards, implementing this loop that keeps trying to start failing collectors/db connections. We can implement these mocks then.

internal/component/database_observability/mysql/collector/setup_consumers.go
cristiangreco: Actually, now I noticed that `"collector error"` would be aligned with other collectors, so ignore me.

internal/component/database_observability/mysql/component.go
cristiangreco: Don't have a strong opinion atm, maybe let's leave it unchanged for now?

internal/component/database_observability/mysql/collector/schema_table.go
```diff
@@ -251,8 +250,7 @@ func (c *SchemaTable) extractSchema(ctx context.Context) error {
 	}
 
 	if err := rs.Err(); err != nil {
-		level.Error(c.logger).Log("msg", "error during iterating over schemas result set", "err", err)
-		return err
+		return fmt.Errorf("error during iterating over schemas result set: %w", err)
```
matthewnolf: nit: given these will be concatanated now upstream (presumably something like: `"error running collector: error during iterating over schemas result set:..."`) we don't need to prefix these with "error" and it makes the string concatenation harder to read

matthewnolf: But it can be for a followup

internal/component/database_observability/mysql/component.go
```diff
@@ -192,16 +191,22 @@ type Component struct {
...
 	dbConnection *sql.DB
 	healthErr    *atomic.String
+	openSQL      func(driverName, dataSourceName string) (*sql.DB, error)
 }
 
 func New(opts component.Options, args Arguments) (*Component, error) {
+	return newWithOpen(opts, args, sql.Open)
+}
+
+func newWithOpen(opts component.Options, args Arguments, openFn func(driverName, dataSourceName string) (*sql.DB, error)) (*Component, error) {
```
matthewnolf: nit: I think `new(...)` is fine and follows convention to lowercase the internal version of a public func

internal/component/database_observability/mysql/component.go
```diff
@@ -271,6 +276,12 @@ func (c *Component) getBaseTarget() (discovery.Target, error) {
 // for MySQL: "8.0.36-28.1"
 var versionRegex = regexp.MustCompile(`^((\d+)(\.\d+)(\.\d+))`)
 
+func (c *Component) reportError(errorMsg string, err error) {
+	err = fmt.Errorf("%s: %w", errorMsg, err)
+	level.Error(c.opts.Logger).Log("msg", err.Error())
```
matthewnolf: Can't this just be `level.Error(c.opts.Logger).Log("msg", fmt.Sprintf("%s: %w", errorMsg, err))`

internal/component/database_observability/mysql/component.go
```diff
@@ -336,7 +351,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 	if err := c.startCollectors(serverUUID, engineVersion, parsedEngineVersion); err != nil {
 		c.healthErr.Store(err.Error())
-		return err
+		return nil
```
matthewnolf, line 353: Should this error not be logged?

internal/component/database_observability/mysql/component.go
```diff
@@ -369,12 +384,20 @@ func enableOrDisableCollectors(a Arguments) map[string]bool {
 }
 
 func (c *Component) startCollectors(serverUUID string, engineVersion string, parsedEngineVersion semver.Version) error {
+	// Best-effort start: try building/starting every enabled collector and aggregate errors.
```
matthewnolf: Can we move this to a function level comment.
```
// startCollectors attempts to start all of the enabled collectors. If one or more collectors fail to start, their errors are reported 
```

internal/component/database_observability/mysql/component.go
```diff
@@ -395,14 +418,13 @@ func (c *Component) startCollectors(serverUUID string, engineVersion string, par
...
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to create QueryTable collector", "err", err)
-			return err
-		}
-		if err := qtCollector.Start(context.Background()); err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to start QueryTable collector", "err", err)
-			return err
+			logStartError(collector.QueryTablesName, "create", err)
+		} else {
+			if err := qtCollector.Start(context.Background()); err != nil {
```
matthewnolf, line 422: Why don't we just start all of the collectors at the end, instead of all of the if/else blocks for each collector

internal/component/database_observability/mysql/component.go
gaantunes: Good catch, yes.

internal/component/database_observability/mysql/component.go
gaantunes: Can do it this way, but not sure it would be much cleaner.
Would need another map of collectors to start from in the end, and still one if else to append it or log creation errors.

matthewnolf: Think it'd _something_ like:

```
collector, err := NewXCollector()
if err != nil {
   //handle err by adding to a slice
} else {
  c.collectors = append(c.collectors, collect)
}
// Do this for all collectors

for _, collector := range c.collectors {
  err := collector.Start()
  if err != nil {
    // handle err
    slices.delete(c.collectors, collector)
  }
}
```

internal/component/database_observability/mysql/component_test.go
matthewnolf, line 1: nit: There's a lot of comments in here which are redundant (explaining the obvious)

matthewnolf:
LGTM. Please add an entry to the changelog documenting the fix.

internal/component/database_observability/mysql/component.go
gaantunes: I think you need to provide an index to this `delete` function. I will move along with the current approach, we can discuss it further on next iterations.

PR #4345, opened 2025-09-04
pyroscope: add test-pyroscope make & gh routines

korniltsev-grafanista:
#### PR Description
- Add  `make pyroscope-test` target. The motivation is an faster alternative to `make test` to run locally. 
- Add github actions step to triger the new make target. The motivation is faster feedback loop
 
image

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.github/workflows/test_pr.yml
+++ b/.github/workflows/test_pr.yml
@@ -26,19 +26,3 @@ jobs:
 
     - run: make GO_TAGS="nodocker" test
 
-  test_pyroscope:
-    name: Test pyroscope packages on Linux
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-        with:
-          persist-credentials: false
-
-      - name: Set up Go
-        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
-        with:
-          go-version-file: go.mod
-          cache: false
-
-      - run: make GO_TAGS="nodocker" test-pyroscope
--- a/.github/workflows/test_pyroscope_pr.yml
+++ b/.github/workflows/test_pyroscope_pr.yml
@@ -1,0 +1,29 @@
+name: Test Pyroscope PR
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  test_pyroscope:
+    name: Test pyroscope packages on Linux
+    runs-on: ubuntu-latest
+    if: >
+      github.actor == 'korniltsev-grafanista' ||
+      github.actor == 'simonswine' ||
+      github.actor == 'aleks-p' ||
+      github.actor == 'marcsanmi'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - run: make GO_TAGS="nodocker" test-pyroscope


kalleep (commented):
I can imagine the motivation for adding a make target to make it more convenient locally but why would we need it as a separate step in the ci? We still run pyroscope tests with our other action

korniltsev-grafanista:
> but why would we need it as a separate step in the ci?

Same as running the step locally - faster feedback. Do not wait for 25 minutes to learn about a test failure.

korniltsev-grafanista:
Here's a fresh example https://github.com/grafana/alloy/actions/runs/17486793977/job/49667324957 
I could have been notified in 4 minutes instead of 19

kalleep:
>Here's a fresh example https://github.com/grafana/alloy/actions/runs/17486793977/job/49667324957
I could have been notified in 4 minutes instead of 19

I can see the value for faster feedback loops but this could set a bad precedence where we add separate steps for whatever components we focus on at the moment. 

But maybe there is a solution here we could add if the job only triggers in prs _and_ only if changes are made to any pyroscope files, WDYT?

korniltsev-grafanista:
> I can see the value for faster feedback loops but this could set a bad precedence where we add separate steps for whatever components we focus on at the moment.

To be fair I don't see a problem. I can totally see how beyla/loki/temp/pyroscope have separate smoke-test targets for our developer happiness.

What is your concern?

> But maybe there is a solution here we could add if the job only triggers in prs and only if changes are made to any pyroscope files, WDYT?

I am ok with this, until I am not (until I break something without changing the whitelisted files).
it is also tricky and will have false positive tirggers - if we add this trigger whitelist I will be insisting on including go.mod go.sum and Makefile into the list

korniltsev-grafanista:
```diff
+      - name: Check for pyroscope changes
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            pyroscope:
+              - 'internal/component/pyroscope/**'
+              - 'go.mod'
+              - 'go.sum'
+              - 'Makefile'
+              - '.github/workflows/test_pr.yml'
```

If we add go.mod (which we have to, since pyroscope have depdendencies)  to the whitelist, then I don't see no value in the whitelist at all - it will run on pretty much every PR

kalleep:
Yep I can see that trying to solve with files is indeed pretty error prone. Another solution would be to use labels.

So one option would be to trigger this on a component basis, we do generate labels for all components. This could work but if you wanted to test all pyroscope related test and not just a specific component it would be to limited.

Another options is to check area labels. We do have `area/pyroscope` and we could add similar ones for others like area/loki, area/prometheus etc.

So my motivation behind having some kind of logic for running specific pyroscope tests in pr:s is that those tests are just unnecessary for a majority of them.

korniltsev-grafanista:
> So one option would be to trigger this on a component basis, we do generate labels for all components. This could work but if you wanted to test all pyroscope related test and not just a specific component it would be to limited.

Do you mean we would need to manually apply labels to the PR? 

https://github.com/grafana/alloy/pull/4354 My this PR has no labels. This one has no labels as well https://github.com/grafana/alloy/pull/4348. 

If the suggestion is to apply labels manually in web UI, then please no. Can we have something that is automated instead of tedious extra manual step or a step that I forget to perform?

> just unnecessary for a majority of them.

I agree they are unnecesary, but at the same time they do not hurt, do they? For example, we have Mac tests that run for each ebpf  change I submit and it does not hurt/bother me (actually it does LOL, because MacOS seem to be flaky and I have to retry, but that's offtopic)

korniltsev-grafanista:
I've come up with a user-based whitelist, PTAL.

Still don't understand the reason to push back on this for everyone. Why don't you want to learn about a test failure faster?

kalleep:
>Still don't understand the reason to push back on this for everyone. Why don't you want to learn about a test failure faster?

So the push back is because what I stated earlier, it is unnecessary for the majority of pr:s. But I think you are maybe misunderstanding me. I really like the idea and I would like to have a solution that would work for other things than just pyroscope because we suffer the same problem in other areas as well. 

The idea with using labels for this is not optimal but I can at least see how we could make this solution generic by using it. And we could look into auto label pr:s with area / component that are touched and run the specialized test with quicker feedback look for them.

korniltsev-grafanista:
I strongly against manually applying labels. The user whitelist should fix the unnecessary "issue". PTAL

Issue #4126, opened 2025-08-05
Expand Prometheus related integration tests


thampiotr:
### Add a description

- Cover different scraping protocols
- Cover different validation schemes
- Use both new and old WAL components
- Use receive_http component

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #4347, opened 2025-09-04
Add integration test for remote write v2 with remote_write

kgeckhart:
#### PR Description
Adds an integration test for prometheus remote write v2 support with `remote_write`

Related to: https://github.com/grafana/alloy/issues/4126

### DIFF
--- a/internal/cmd/integration-tests/common/metric.go
+++ b/internal/cmd/integration-tests/common/metric.go
@@ -21,7 +21,7 @@ type MetricData struct {
 }
 
 type HistogramRawData struct {
-	Timestamp int64
+	Timestamp float64
 	Data      HistogramData
 }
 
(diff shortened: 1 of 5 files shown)


internal/cmd/integration-tests/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 name: alloy-integration-tests
 services:
   mimir:
-    image: grafana/mimir:2.10.4
+    image: grafana/mimir:2.17.0
```
kgeckhart, line 4: This is the latest release and the minimum mimir version to support PRW v2

internal/cmd/integration-tests/common/metric.go
```diff
@@ -21,7 +21,7 @@ type MetricData struct {
 }
 
 type HistogramRawData struct {
-	Timestamp int64
+	Timestamp float64
```
kgeckhart, line 24: This was a side effect of upgrading Mimir but I did not look at why

PR #4348, opened 2025-09-04
fix: demote noisy log in loki client

kalleep:
#### PR Description
This log is really noisy and will happen when we encounter errors that we can retry, e.g. 429 and 500 errors.

Common scenario is when we are getting rate limited and performs retries. We already have a metric that covers number of retries performed and will emit an error log if we cannot send logs after all retries are exhausted.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/client/client.go
+++ b/internal/component/common/loki/client/client.go
@@ -377,7 +377,7 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
 			break
 		}
 
-		level.Warn(c.logger).Log("msg", "error sending batch, will retry", "status", status, "tenant", tenantID, "error", err)
+		level.Debug(c.logger).Log("msg", "error sending batch, will retry", "status", status, "tenant", tenantID, "error", err)
 		c.metrics.batchRetries.WithLabelValues(c.cfg.URL.Host, tenantID).Inc()
 		backoff.Wait()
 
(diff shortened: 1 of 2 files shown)


internal/component/common/loki/client/client.go
```diff
@@ -387,17 +387,15 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
 		}
 	}
 
-	if err != nil {
```
kalleep, line 390: This condition is always true. We only get here if we have a non-nil error https://github.com/grafana/alloy/blob/0966e34d3adbe12cb3d283e69406150873ba2321/internal/component/common/loki/client/client.go#L368-L373

internal/component/common/loki/client/client.go
```diff
@@ -387,17 +387,15 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
...
-		level.Error(c.logger).Log("msg", "final error sending batch", "status", status, "tenant", tenantID, "error", err)
-		// If the reason for the last retry error was rate limiting, count the drops as such, even if the previous errors
-		// were for a different reason
-		dropReason := ReasonGeneric
-		if batchIsRateLimited(status) {
-			dropReason = ReasonRateLimited
-		}
-		c.metrics.droppedBytes.WithLabelValues(c.cfg.URL.Host, tenantID, dropReason).Add(bufBytes)
-		c.metrics.droppedEntries.WithLabelValues(c.cfg.URL.Host, tenantID, dropReason).Add(float64(entriesCount))
+	level.Error(c.logger).Log("msg", "final error sending batch", "status", status, "tenant", tenantID, "error", err)
```
ptodev:
```suggestion
	level.Error(c.logger).Log("msg", "final error sending batch, no retries left, dropping data", "status", status, "tenant", tenantID, "error", err)
```
This makes it more clear?

ptodev (approved):

PR #4352, opened 2025-09-04
Fix syslog shutdown race conditions

thampiotr:
#### PR Description

loki.source.syslog had a few race conditions:
- on config reload the following happens: new components are created, old components shut down and new components are run. syslog was binding port in the New function, so this led to port conflict because the old component didn't shut down yet. So what I do here is make sure we don't try to bind ports when there's a call to New, but on call to Run.
- on shutdown, calling Stop on targets drains the pending messages - this won't work if there is nothing reading from the handler channel - so I add a similar approach to loki.source.file, where we start a temporary goroutine that will continue draining when we Stop the targets.

Internal escalation https://github.com/grafana/support-escalations/issues/18202

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
+++ b/internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
@@ -29,7 +29,7 @@ import (
 var (
 	DefaultIdleTimeout      = 120 * time.Second
 	DefaultMaxMessageLength = 8192
-	DefaultProtocol         = protocolTCP
+	DefaultProtocol         = ProtocolTCP
 )
 
 // SyslogTarget listens to syslog messages.
@@ -72,14 +72,14 @@ func NewSyslogTarget(
 	}
 
 	switch t.transportProtocol() {
-	case protocolTCP:
+	case ProtocolTCP:
 		t.transport = NewSyslogTCPTransport(
 			config,
 			t.handleMessage,
 			t.handleMessageError,
 			logger,
 		)
-	case protocolUDP:
+	case ProtocolUDP:
 		t.transport = NewSyslogUDPTransport(
 			config,
 			t.handleMessage,
--- a/internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget_test.go
+++ b/internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget_test.go
@@ -19,13 +19,14 @@ import (
 	"github.com/grafana/alloy/internal/component/loki/source/syslog/internal/syslogtarget/syslogparser"
 
 	"github.com/go-kit/log"
-	scrapeconfig "github.com/grafana/alloy/internal/component/loki/source/syslog/config"
 	"github.com/leodido/go-syslog/v4"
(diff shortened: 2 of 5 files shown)


internal/component/loki/source/syslog/syslog.go
```diff
@@ -98,46 +109,93 @@ func (c *Component) Update(args component.Arguments) error {
...
-			if err != nil {
-				level.Error(c.opts.Logger).Log("msg", "failed to create syslog listener with provided config", "err", err)
-				continue
+	return nil
+}
+
+func (c *Component) startDrainingRoutine() func() {
+	readCtx, cancel := context.WithCancel(context.Background())
+	c.mut.RLock()
+	defer c.mut.RUnlock()
```
kalleep, line 131: Should we remove holding the lock here and make that a condition for calling the function?

As far as I can tell we need to take the lock at all call sites anyway

kalleep (approved):
Nice! Just left one comment

internal/component/loki/source/syslog/syslog.go
kgeckhart: My two cents is that since it's a read lock it's safer to leave here. I don't know of any way to enforce a caller has taken the read lock before calling so keeping it here makes sure it will always happen and any number of read locks can be taken at a time.

internal/component/loki/source/syslog/syslog.go
```diff
@@ -98,46 +109,93 @@ func (c *Component) Update(args component.Arguments) error {
...
+		}
+	}
+	c.mut.RUnlock()
+
+	// Stop draining routine
+	cancel()
+
+	// Create new targets
+	c.mut.Lock()
+	defer c.mut.Unlock()
```
kgeckhart, line 175: 🤔 is there a potential race condition / panic here if a reload happens at the same time as shutdown? Shutdown is going to trigger https://github.com/grafana/alloy/blob/c2f9bbb4b9e9d7aea00eb775df3c24ec0311a1cc/internal/component/loki/source/syslog/syslog.go#L73-L88 where both only take a readlock so they could simultaneously try to stop targets right? The target stop function has the potential to close an already close channel from what I can tell (hopefully something else would error ahead of it). 

I don't know that it's a realistic/possible scenario but I do wonder if it's safer to take a write lock before we attempt to mutate the state of the targets.

thampiotr: Thanks for taking a closer look. 
> if a reload happens at the same time as shutdown

if I understood right and by "reload" you mean call to `reloadTargets` - then this is not possible to happen at the same time as shutdown deferred function, because `reloadTargets` happens on the same thread as the `Run` function uses.

Let me know if you were thinking of a different scenario

internal/component/loki/source/syslog/syslog.go
thampiotr: +1 I prefer it to be inside the function so it's less likely to be misused by accident

PR #4353, opened 2025-09-04
database_observability: ensure compatibility with MySQL 8.0.28 or lower

cristiangreco:
#### PR Description
Ensure that the query that fetches query samples is compatible with MySQL 8.0.28 or lower.

Specifically:
- CPU_TIME: Added in 8.0.28
- MAX_CONTROLLED_MEMORY: Added in 8.0.31
- MAX_TOTAL_MEMORY: Added in 8.0.31

Used https://github.com/grafana/alloy/pull/4269 as a starter.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/query_sample_test.go
+++ b/internal/component/database_observability/mysql/collector/query_sample_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
 )
 
-var recentMySQLVersion = semver.MustParse("8.0.32")
+var latestCompatibleVersion = semver.MustParse("8.0.32")
 
 func TestQuerySample(t *testing.T) {
 	defer goleak.VerifyNone(t)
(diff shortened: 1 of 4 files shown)


internal/component/database_observability/mysql/collector/query_sample_test.go
```diff
@@ -19,6 +20,8 @@ import (
 	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector/parser"
 )
 
+var recentMySQLVersion = semver.MustParse("8.0.32")
```
matthewnolf: nit: not sure what 'recent' means here

cristiangreco: I was looking for something like "safe" or "latest". wdyt?

cristiangreco: @matthewnolf do you have any suggestion?

matthewnolf: Latest would be `8.4.x` though right? I'm assuming this is the version after which nothing else changes?

PR #4354, opened 2025-09-05
pyroscope: trace_id propagation and loging

korniltsev-grafanista:
#### PR Description
- include  `trace_id` in log entries
- propagate the trace downstream to pyroscope servers

```
ts=2025-09-05T08:53:07.344977577Z level=debug msg=AppendIngest component_path=/ component_id=pyroscope.write.w trace_id=c882d8a1df9054c28db7ab3b6ac89ff5 sz=11 n=1 dl=none st=2025-09-05T15:53:06.279+07:00
...
ts=2025-09-05T08:53:07.207394191Z caller=http.go:346 level=debug traceID=c882d8a1df9054c28db7ab3b6ac89ff5 tenant=678013 method=POST uri="/ingest?name=qwe%7Bservice_name%3Dqwe%7D" status=200 duration=380.127989ms request_header_Accept-Encoding=gzip request_header_Content-Type=application/x-www-form-urlencoded request_header_User-Agent=Go-http-client/1.1 request_header_X-Alloy-Id=5825b981-996d-45d5-aa2f-7c7edb7521fe request_body_size=11B request_body_read_duration=7.31µs msg="http request processed"
```
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -29,7 +29,6 @@ import (
 	"github.com/grafana/pyroscope/api/gen/proto/go/push/v1/pushv1connect"
 	typesv1 "github.com/grafana/pyroscope/api/gen/proto/go/types/v1"
 	"github.com/grafana/pyroscope/api/model/labelset"
-	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	commonconfig "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"


internal/component/pyroscope/write/write.go
```diff
@@ -590,3 +683,24 @@ func validateLabels(lbls labels.Labels) error {
 
 	return nil
 }
+
+func configureTracing(config Arguments, httpClient *http.Client) error {
+	if opentracing.IsGlobalTracerRegistered() {
+		return fmt.Errorf("opentracing was configured, but is not expected to be")
```
simonswine:
```suggestion
		return fmt.Errorf("opentracing was configured, but is not expected to be")
```

Not super clear to me if this is a path users might face, but the action to be taken when seeing should be clearer. Maybe if it is impossible to configure alloy with opentracing we should just panic?

Maybe we can explain how this is related to those settings: https://grafana.com/docs/alloy/latest/reference/config-blocks/tracing/

simonswine (approved):
LGTM

Great addition, thanks for helping us debug some of the harder issues there are.

PR #4367, opened 2025-09-05
Remotecfg status

erikbaranowski:
#### PR Description

TODO
- Changelog
- Tests
- Docs

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


syntax/diag/diag_test.go
```diff
@@ -0,0 +1,426 @@
+package diag
```
erikbaranowski, line 1: we've got more tests than from the PR additions since diag.go was missing unit tests. I can cut back to just what was added on the PR but it seems good to have the tests.

internal/service/remotecfg/remotecfg.go
```diff
@@ -276,9 +276,10 @@ func (s *Service) getConfig() (*collectorv1.GetConfigResponse, error) {
 
 	response, err := s.apiClient.GetConfig(s.getContext(), &connect.Request[collectorv1.GetConfigRequest]{
 		Msg: &collectorv1.GetConfigRequest{
-			Id:              s.args.ID,
-			LocalAttributes: s.attrs,
-			Hash:            s.cm.getRemoteHash(),
+			Id:                 s.args.ID,
+			LocalAttributes:    s.attrs,
+			Hash:               s.cm.getRemoteHash(),
+			RemoteConfigStatus: s.cm.getRemoteConfigStatus(),
```
dehaansa: According to the alloy-remote-config proto
```
	// The status of the remote config that was previously received from the Server.
	// This field SHOULD be unset if the remote config status is unchanged since the
	// last GetConfigRequest message.
```

This doesn't feel like it matches up to the implementation here, is this still the expectation?

erikbaranowski: good call, I forgot to implement that detail for efficiency on the server side.

erikbaranowski: added https://github.com/grafana/alloy/pull/4367/commits/5d72086c3113f9357933295faf186c18fdf83387

internal/service/remotecfg/config_manager.go
```diff
@@ -391,6 +437,75 @@ func (cm *configManager) getRemoteHash() string {
...
+	return cm.copyRemoteConfigStatus()
+}
+
+// getRemoteConfigStatusForRequest returns the remote config status only if it should be sent
+// (first call or status has changed since last sent).
+func (cm *configManager) getRemoteConfigStatusForRequest() *collectorv1.RemoteConfigStatus {
+	cm.mut.Lock()
+	defer cm.mut.Unlock()
+
+	// Send status if we've never sent one before (first call) or if it has changed
```
dehaansa, line 466: This comment doesn't accurately describe the other state, where we reverted to nil because of a failed request.

dehaansa (approved):

internal/service/remotecfg/config_manager.go
erikbaranowski: good call, I updated the comment to describe what it does rather than where it is used so we aren't after a moving target.

PR #4368, opened 2025-09-06
pyroscope.write: rework to accept useragent and seed in constructor

korniltsev-grafanista:

- do not use `alloyseed.Get()` inside the component - pass as an argument to constructor
- do not use `useragent.Get()` inside the component - pass as an argument to constructor
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/pyroscope/write/write.go
+++ b/internal/component/pyroscope/write/write.go
@@ -153,7 +153,14 @@ type Exports struct {
 }
 
 // New creates a new pyroscope.write component.
-func New(logger log.Logger, tracer trace.Tracer, reg prometheus.Registerer, onStateChange func(Exports), userAgent string, uid string, c Arguments) (*Component, error) {
+func New(
+	logger log.Logger,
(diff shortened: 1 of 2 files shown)


marcsanmi:
Why this is needed? Do we might want to set them up differently at some point?

marcsanmi (approved):
LGTM

korniltsev-grafanista:
> Why this is needed? Do we might want to set them up differently at some point?

I am working on decoupling pyroscope components from alloy. (https://github.com/grafana/alloy/pull/4345 https://github.com/grafana/alloy/pull/4369 )
This will allow pyroscope components
- have less transitive depdendencies 
- faster tests compilation and run time
- better code reuse ( at some point we may want to build a pyroscope agent oustide alloy ( now we dont))

it is also just a good design principle in general to not use global variables but accept them from the outside, makes the code more flexible https://en.wikipedia.org/wiki/Dependency_injection

PR #4376, opened 2025-09-08
fix: slow updates to loki.source.file

kalleep:
#### PR Description
There are situations where updating `loki.source.file` can take a unreasonable amount of time.

We observed this happening if update happens when pipeline is blocked, e.g. slow writes to external system.
This will not fix the underlying issue but it will help the majority of cases when receivers are not updated but only targets.

I removed `arg` property on component and passes all parameters to createReader instead. So now we have one RWLock protecting tasks and one RWLock protecting recievers.

#### Notes to the Reviewer
Another drive by fix was to only take read lock when `DebugInfo` is called.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,8 @@ Main (unreleased)
 
 - Reduce log level from warning for `loki.write` when request fails and will be retried. (@kalleep)
 
+- Fix slow updates to `loki.source.file` when only targets have changed and pipeline is blocked on writes. (@kalleep):W
+
 ### Bugfixes
 
 - Update `webdevops/go-common` dependency to resolve concurrent map write panic. (@dehaansa)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,7 +115,7 @@ Main (unreleased)
 
 - Reduce log level from warning for `loki.write` when request fails and will be retried. (@kalleep)
 
-- Fix slow updates to `loki.source.file` when only targets have changed and pipeline is blocked on writes. (@kalleep):W
+- Fix slow updates to `loki.source.file` when only targets have changed and pipeline is blocked on writes. (@kalleep)
 
 ### Bugfixes
 
--- a/internal/component/loki/source/file/file.go
+++ b/internal/component/loki/source/file/file.go
@@ -276,9 +276,9 @@ func (c *Component) Update(args component.Arguments) error {
 // DebugInfo returns information about the status of tailed targets.
 // TODO(@tpaschalis) Decorate with more debug information once it's made
 // available, such as the last time a log line was read.
-func (c *Component) DebugInfo() interface{} {
-	c.tasksMut.Lock()
-	defer c.tasksMut.Unlock()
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -115,6 +115,8 @@ Main (unreleased)
 
 - Reduce log level from warning for `loki.write` when request fails and will be retried. (@kalleep)
 
+- Fix slow updates to `loki.source.file` when only targets have changed and pipeline is blocked on writes. (@kalleep):W
```
```suggestion
- Fix slow updates to `loki.source.file` when only targets have changed and pipeline is blocked on writes. (@kalleep)
```

internal/component/loki/source/file/file.go
```diff
@@ -142,26 +145,27 @@ func (c *Component) Run(ctx context.Context) error {
...
+			c.receiversMut.RLock()
 			for _, receiver := range c.receivers {
 				receiver.Chan() <- entry
 			}
-			c.mut.RUnlock()
+			c.receiversMut.RUnlock()
 		case <-c.updateReaders:
-			c.mut.Lock()
+			c.tasksMut.Lock()
+			c.receiversMut.RLock()
```
thampiotr, line 170: I think you can grab this lock for each individual entry in the goroutine?

internal/component/loki/source/file/file.go
```diff
@@ -204,13 +209,21 @@ func (c *Component) Run(ctx context.Context) error {
 func (c *Component) Update(args component.Arguments) error {
 	newArgs := args.(Arguments)
 
-	c.mut.Lock()
-	defer c.mut.Unlock()
-	c.args = newArgs
```
thampiotr, line 209: We seem to use `taskMut` also for c.args, so maybe let's just stick to calling it `mut`?

internal/component/loki/source/file/file.go
```diff
@@ -81,12 +82,14 @@ type Component struct {
 	opts    component.Options
 	metrics *metrics
 
-	mut       sync.RWMutex
-	args      Arguments
```
thampiotr, line 85: args are still protected by the mutex, so I would expect them to be declared next to their mutex

internal/component/loki/source/file/file.go
kalleep: No, we no longer store args. Just tasks built from args.Targets and recievers from args.ForwardTo

internal/component/loki/source/file/file.go
kalleep: Sure I can but would that not just be unnecessary locking / unlocking?

thampiotr: yeah, thinking about it, this should not be contended - I was just thinking about symmetry of this and the other loop above

thampiotr (approved):

korniltsev-grafanista:
Do we have a mutex/block profile demonstrating this problem?

PR #4377, opened 2025-09-08
fix: attempt to reduce allocations when using external labels in loki.write

kalleep:
#### PR Description
This shows up on profiles a lot. This is a similar fix to what was done in https://github.com/grafana/alloy/pull/3984.

This can still end up doing allocations if the map don't have enough capacity. But it should reduce allocation in case we have capacity because we are no longer always allocating a new map and copy over labels.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,8 @@ Main (unreleased)
 
 - Reduce log level from warning for `loki.write` when request fails and will be retried. (@kalleep)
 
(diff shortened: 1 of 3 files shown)


kalleep:
Ok, this will not be as simple. Entries are fanned out to multiple endpoints so we would be doing concurrent writes to a map.

PR #4382, opened 2025-09-09
fix: reduce allocations in loki.write when using external labels

kalleep:
#### PR Description
`external_labels` is configured for a loki.write component but the mutation was done per endpoint. So if we had two endpoints we would apply external labels twice. 

This pr moved out the mutation so it happens before client code. This shows up in profiles a lot and this is one way to at least reduce it a bit.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/manager_test.go
+++ b/internal/component/common/loki/client/manager_test.go
@@ -23,15 +23,13 @@ import (
 	"github.com/grafana/alloy/internal/component/common/loki/wal"
 )
 
-var testLimitsConfig = limit.Config{
-	MaxLineSizeTruncate: false,
-	MaxStreams:          0,
-	MaxLineSize:         0,
-}
-
 var (
+	testLimitsConfig = limit.Config{
+		MaxLineSizeTruncate: false,
+		MaxStreams:          0,
+		MaxLineSize:         0,
+	}
 	nilMetrics = NewMetrics(nil)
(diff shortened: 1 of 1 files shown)


kalleep:
I also did a lot of clean up for loki client code, things that was just copied over from promtail but was dead code in alloy. I can break that out in a separate pr if that clutters it too much

internal/component/loki/write/write.go
```diff
@@ -205,12 +207,23 @@ func (c *Component) Update(args component.Arguments) error {
...
-	return err
+	return nil
+}
+
+func newEntryHandler(handler loki.EntryHandler, externalLabels model.LabelSet) loki.EntryHandler {
+	if len(externalLabels) == 0 {
+		return handler
+	}
+	return loki.NewEntryMutatorHandler(handler, func(e loki.Entry) loki.Entry {
+		e.Labels = externalLabels.Merge(e.Labels)
```
thampiotr, line 226: Is this better than `e.Labels.Merge(externalLabels)`?

kalleep: It is important to use `externalLabels.Merge` to keep the same behavior as we had before, if a label is found it both maps we prefer the one from from `e.Labels` over `externalLabels`

thampiotr: gotcha! thanks for checking

PR #4388, opened 2025-09-10
feat: add `autoscaling.external`

KimMJ:
#### PR Description

Currently there is no option to disable `replicas` in Statefulset and Deployment controller type, when I use KEDA to autoscale alloy based on number of metrics that alloy cluster pulls. I tried `replicas: null` but it also reset replica to 1 when upgrade helm. This could make some downtime (less than 30s).

This PR is to add `autoscaling.external` variable to remove `replicas` when external autoscaler is used.

No issue exists.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


kalleep:
Hey sorry we missed this, a fix have been merged in https://github.com/grafana/alloy/pull/6311 so I will close this one

PR #4395, opened 2025-09-10
Write queue prwv2

dehaansa:
#### PR Description
Update prometheus.write.queue component to have prometheus remote write v2 support.

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/prometheus/write/queue/types.go
+++ b/internal/component/prometheus/write/queue/types.go
@@ -7,6 +7,7 @@ import (
 	"github.com/grafana/alloy/syntax/alloytypes"
 	"github.com/grafana/walqueue/types"
 	"github.com/prometheus/common/version"
+	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/storage"
 )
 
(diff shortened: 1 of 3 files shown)


internal/component/prometheus/write/queue/types.go
```diff
@@ -92,6 +95,11 @@ func (r *Arguments) Validate() error {
 		if conn.Parallelism.AllowedNetworkErrorFraction < 0 || conn.Parallelism.AllowedNetworkErrorFraction > 1 {
 			return fmt.Errorf("allowed_network_error_percent must be between 0.00 and 1.00")
 		}
+		if conn.ProtobufMessage == RemoteWriteProtoMsg(config.RemoteWriteProtoMsgV2) {
+			if conn.MetadataCacheSize <= 0 {
```
kalleep, line 99: Should we require experimental feature flag for remote write v2 like we do for [prometheus.remotewrite](https://github.com/grafana/alloy/blob/main/internal/component/prometheus/remotewrite/remote_write.go#L353)?

kalleep: The component itself is experimental, so it's not required

kalleep (approved):

internal/component/prometheus/write/queue/types.go
```diff
@@ -126,6 +134,33 @@ type EndpointConfig struct {
...
+	case string(config.RemoteWriteProtoMsgV1):
+		*s = RemoteWriteProtoMsg(config.RemoteWriteProtoMsgV1)
+	case string(config.RemoteWriteProtoMsgV2):
+		*s = RemoteWriteProtoMsg(config.RemoteWriteProtoMsgV2)
+	default:
+		return fmt.Errorf("unknown remote write proto message: %s", str)
+	}
+
+	return nil
 }
```
jharvey10, line 164: This might be my lack of Go knowledge showing through, but a couple questions about this:

- Is this called frequently? The name suggests it might, but I'm not well versed in how this works from end to end.
- If not, my next question is mostly irrelevant :)
- If so, are we paying a performance penalty here by converting the text byte slice (and potentially the two  config options) to strings? Is there a more direct equality comparison that could be used instead to avoid that?

PR #4410, opened 2025-09-12
chore: fix flaky runtime test

kalleep:
#### PR Description
This fixes a really flaky test in runtime.

The problem was that in a lot of tests we re creating a controller, starting the controller and then asserting some kind of result. If this assert was too fast we would close down the controller before any components and services was scheduled to run, so we did not stop them.

This is mostly fine except when a service starts a go routine when `New` is called. When clustering service is created we call [ckit.NewNode](https://github.com/grafana/alloy/blob/d0690dd5ac1a184a1f684f6a2bab9f1b92dbbc6a/internal/service/cluster/cluster.go#L187), this ends up calling [gossiphttp.NewTransport](https://github.com/grafana/ckit/blob/c91f07842b3440e5129cbe7dea33fcd787656403/node.go#L165) and this will spawn a [go routine](https://github.com/grafana/ckit/blob/c91f07842b3440e5129cbe7dea33fcd787656403/internal/gossiphttp/transport.go#L211). So when a test that starts a controller finishes too fast we have not scheduled this service so we never end up stopping it either leaking the go routine and tripping the leak verification we have.  

To solve this we expose `LoadComplete` that will return true once services and components have been configured and scheduled to run.

Fixes: https://github.com/grafana/alloy/pull/4410

#### Notes to the Reviewer
I also passed io.Discard for all logging to these test. It is really hard to see what is going on when a test fails because we are printing a bunch of logs, I can revert that change if we want to keep this behaviour.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/runtime/alloy.go
+++ b/internal/runtime/alloy.go
@@ -129,12 +129,9 @@ type Runtime struct {
 
 	loadFinished chan struct{}
 
-	loadMut    sync.RWMutex
-	loadedOnce atomic.Bool
-	// scheduledOnce is used to indicate that services and components have been scheduled at least once.
-	scheduledOnce atomic.Bool
-	// reloaded is used in test to make sure a reload have happened.
-	reloaded atomic.Bool
+	loadMut      sync.RWMutex
+	loadedOnce   atomic.Bool
+	loadComplete atomic.Bool
 }
 
 // New creates a new, unstarted Alloy controller. Call Run to run the controller.
@@ -305,8 +302,7 @@ func (f *Runtime) Run(ctx context.Context) {
 				level.Error(f.log).Log("msg", "failed to load components and services", "err", err)
 			}
 
-			f.scheduledOnce.Store(true)
-			f.reloaded.Store(true)
+			f.loadComplete.Store(true)
 		}
 	}
 }
@@ -350,14 +346,14 @@ func (f *Runtime) applyLoaderConfig(applyOptions controller.ApplyOptions) error 
 	f.loadMut.Lock()
 	defer f.loadMut.Unlock()
 
+	f.loadComplete.Store(false)
+
 	diags := f.loader.Apply(applyOptions)
 	if !f.loadedOnce.Load() && diags.HasErrors() {
 		// The first call to Load should not run any components if there were
 		// errors in the configuration file.
 		return diags
 	}
-
-	f.reloaded.Store(false)
 	f.loadedOnce.Store(true)
 
 	select {
@@ -368,11 +364,12 @@ func (f *Runtime) applyLoaderConfig(applyOptions controller.ApplyOptions) error 
 	return diags.ErrorOrNil()
 }
 
-// Ready returns whether the Alloy controller has finished its initial load and scheduled all components and services.
+// Ready returns whether the Alloy controller has finished its initial load.
 func (f *Runtime) Ready() bool {
-	return f.scheduledOnce.Load()
+	return f.loadedOnce.Load()
 }
 
-func (f *Runtime) Reloaded() bool {
-	return f.reloaded.Load()
+// LoadComplete returns true when the components loaded via LoadSource have been scheduled and are running. Used in testing.
+func (f *Runtime) LoadComplete() bool {
+	return f.loadComplete.Load()
 }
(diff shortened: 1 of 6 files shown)


internal/runtime/import_test.go
```diff
@@ -348,6 +354,10 @@ func testConfig(t *testing.T, config string, reloadConfig string, update func())
 			return export.LastAdded <= -10
 		}, 3*time.Second, 10*time.Millisecond)
 	}
+
+	// This is a hack to make sure we have some time for runtime to scheduled all components
+	// so they can be properly exited
+	time.Sleep(3 * time.Second)
```
kalleep: This is not optimal. The fix applied else where do not work when we reload config.

This have the same problem where we created a new component and extract the result before it has been scheduled to run. So if the test is too fast we will leak.

I will look into a better fix

internal/runtime/alloy.go
```diff
@@ -131,6 +131,10 @@ type Runtime struct {
 
 	loadMut    sync.RWMutex
 	loadedOnce atomic.Bool
+	// scheduledOnce is used to indicate that services and components have been scheduled at least once.
+	scheduledOnce atomic.Bool
+	// reloaded is used in test to make sure a reload have happened.
```
thampiotr:
```suggestion
	// scheduledOnce is used to indicate that services and components have been scheduled at least once.
	scheduledOnce atomic.Bool
	// reloaded is used in test to make sure a reload has happened.
	reloaded atomic.Bool
```

internal/runtime/alloy.go
```diff
@@ -131,6 +131,10 @@ type Runtime struct {
 
 	loadMut    sync.RWMutex
 	loadedOnce atomic.Bool
+	// scheduledOnce is used to indicate that services and components have been scheduled at least once.
+	scheduledOnce atomic.Bool
```
kalleep: We could remove this one and only use reloaded in tests

internal/runtime/alloy.go
```diff
@@ -360,7 +368,11 @@ func (f *Runtime) applyLoaderConfig(applyOptions controller.ApplyOptions) error
...
 }
 
-// Ready returns whether the Alloy controller has finished its initial load.
+// Ready returns whether the Alloy controller has finished its initial load and scheduled all components and services.
 func (f *Runtime) Ready() bool {
-	return f.loadedOnce.Load()
+	return f.scheduledOnce.Load()
+}
+
+func (f *Runtime) Reloaded() bool {
```
thampiotr: ### 
```suggestion
// LoadComplete returns true when the components loaded via LoadSource have been scheduled and are running. Used in testing.
func (f *Runtime) LoadComplete() bool {
```

I think this is better name, because it is used in conjunction with `LoadSource`. After we run `LoadSource` we wait for `LoadComplete`.

internal/runtime/alloy.go
```diff
@@ -360,7 +368,11 @@ func (f *Runtime) applyLoaderConfig(applyOptions controller.ApplyOptions) error
...
+	return f.reloaded.Load()
```
thampiotr:
```suggestion
	return f.schedulerSynchronized.Load()
```

I know this is longer name, but I think it better captures the internal meaning of this variable too. So I think we can change it.

internal/runtime/alloy.go
```diff
@@ -350,6 +356,8 @@ func (f *Runtime) applyLoaderConfig(applyOptions controller.ApplyOptions) error
 		// errors in the configuration file.
 		return diags
 	}
+
+	f.reloaded.Store(false)
```
thampiotr: I think technically this should be set to false even if we go into the `if !f.loadedOnce.Load() && diags.HasErrors() {` condition. So perhaps move it before the `if`?

It probably doesn't have practical consequences tho.

internal/runtime/alloy.go
thampiotr: Isn't `Ready()` called somewhere else outside tests?

internal/runtime/alloy.go
kalleep: Yeah I think this is better and we don't have to use `scheduledOnce` anymore and can revert back to the old code for that

kalleep:
I reverted back to the old behavior for Ready and changed to have loadComplete that we use in all tests

internal/runtime/alloy.go
kalleep: It is used by `/~/ready` endpoint. This works out because this endpoint is only available after http_service have been scheduled to run.

PR #4413, opened 2025-09-12
chore: enable module cache for pr tests

kalleep:
#### PR Description
Run lint job on its own and enable cache for lint_go and test_linux.

Instead of having two different test for linux I opted to run the one we run on main in pr:s as well.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


kalleep:
cc @ptodev I saw that you disabled these before. What is stopping us from using the cache currently?

PR #4416, opened 2025-09-12
db-o11y: add indexes to PG schema_details collector

fridgepoet:
#### PR Description

Following up on [work](https://github.com/grafana/alloy/pull/4046) with the Postgres Schema Details collector, this PR
extends the existing schema_details collector to collect table index information, both column-based and expression-based.

Note that nullability is assumed true if the index involves an expression.

Contributes to: https://github.com/grafana/grafana-dbo11y-app/issues/1388

#### Notes to the Reviewer

Not adding to the changelog because the old entry already mentions the Postgres schema_details collector.

Here is a quick snippet of how it should look based on a local frontend WIP changes:

https://github.com/user-attachments/assets/b048dc09-76a0-42a2-bf2f-5b52a71db2ee

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/schema_details_test.go
+++ b/internal/component/database_observability/postgres/collector/schema_details_test.go
@@ -86,7 +86,8 @@ func TestSchemaTable(t *testing.T) {
 					"unique",
 					"column_names",
 					"expressions",
-				}).AddRow("authors_pkey", "btree", true, pq.StringArray{"id"}, pq.StringArray{}),
+					"has_nullable_column",
+				}).AddRow("authors_pkey", "btree", true, pq.StringArray{"id"}, pq.StringArray{}, true),
 			)
 
 		err = collector.Start(t.Context())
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -341,5 +392,38 @@ func (c *SchemaDetails) fetchColumnsDefinitions(ctx context.Context, databaseNam
...
+
+		// TODO: Expression-based indexes nullability difficult to determine
+
+		tblSpec.Indexes = append(tblSpec.Indexes, indexSpec{
+			Name:        indexName,
+			Type:        indexType,
+			Unique:      unique,
+			Columns:     columns,
+			Expressions: expressions,
+		})
```
fridgepoet: Nullability a bit difficult to determine when there is an expression.

PostgreSQL doesn't have an equivalent to MySQL's `information_schema.statistics.nullable` field (where we are getting nullable info for MySQL).

Options:
* make "nullable" able to be absent (nil) and deal in the frontend
* remove nullable from Postgres schema_details collector
* assume some value if there is an expression

fridgepoet: after discussion, we’ll do “-” when it involves an expression

fridgepoet: After another discussion, we'll keep the following behavior: we’ll do “-” when it involves an expression
But it'll be done in the frontend. In the backend, we'll hardcode some value (nullable = true when there is an expression)

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -341,5 +395,40 @@ func (c *SchemaDetails) fetchColumnsDefinitions(ctx context.Context, databaseNam
 		return nil, err
 	}
 
+	indexesRS, err := c.dbConnection.QueryContext(ctx, selectIndexes, schemaName, tableName)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to query indexes", "schema", schemaName, "table", tableName, "err", err)
```
cristiangreco: Could we additionally log the database name like we do in other places?

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -341,5 +395,40 @@ func (c *SchemaDetails) fetchColumnsDefinitions(ctx context.Context, databaseNam
...
+		var unique, hasNullableColumn bool
+		var columns, expressions pq.StringArray
+
+		if err := indexesRS.Scan(&indexName, &indexType, &unique, &columns, &expressions, &hasNullableColumn); err != nil {
+			level.Error(c.logger).Log("msg", "failed to scan indexes", "schema", schemaName, "table", tableName, "err", err)
+			return nil, err
+		}
+
+		// nullable if has nullable columns or has expressions
+		nullable := hasNullableColumn || len(expressions) > 0 // TODO: assume that indexes with any expressions are nullable
```
cristiangreco: nit: more than a TODO this is what we're doing now, correct? Maybe we could leave as TODO to followup investigation into nullability of expressions.

cristiangreco (commented):
LGTM, left a couple of nitpick comments

matthewnolf (approved):

Issue #547, opened 2022-01-13
Manually stored metrics do not have metadata


rfratto:
Samples which are manually written to the WAL (i.e., written outside of the normal scraping process of a metrics instance) do not support metadata being sent over remote write. 

Metadata is used to display metric type and help info from Prometheus  query frontends like Grafana.

Remote write currently [expects](https://github.com/prometheus/prometheus/blob/01c55822169a9123367e333dc407edd8a263f75d/storage/remote/metadata_watcher.go#L137) that metadata comes from an instance of scrape.Manager, which is not always the case for code relying on remote write to send samples. 

This impacts at least the following: 

* spanmetrics 
* integrations-next (which uses an external scrape manager whose metadata is not exposed directly to remote_write)
* Metrics from integrations when using Grafana Agent Operator (since it uses `integrations-next`; see above) 
* Flow components

This went mostly unnoticed because most metrics still have metadata. I first became aware of this when prototyping grafana/agent#1261, where scraping and remote_write was initially fully decoupled, causing all metadata to disappear. 

It's not immediately obvious what the best solution here is. Some way of being able to customize what metadata gets collected by remote_write would be ideal though.

srclosson:
Just ran into this in another POV. Would love to add a +1 to this issue

wildum:
this is also a problem for converting Prometheus metrics to Oltp metrics in alloy because the otelcol.receiver.prometheus is using the metadata to retrieve the type

mattdurham:
We likely can enable the scrape setting for opentelemetry metadata that adds it to the appender context.

EduardoCusihuaman:
+1 on this. We're hitting the same wall.

We're processing metrics from `remote_write` and our downstream logic needs to differentiate between Counters and Gauges to work properly. The missing type metadata is a blocker for us right now.

Looks like the real fix is tied to the `remote_write` protocol v2.0 [implementation](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/) . Just wanted to add our use case to the list of reasons to get that prioritized.

Thanks for all the work on Alloy 👨‍💼

kgeckhart:
Just wanted to document the known use cases and where things are at,

- [x] `prometheus.scrape` -> `otelcol.receiver.prometheus`: resolved by https://github.com/grafana/alloy/pull/4098
- [ ] `promtheus.write.queue` https://github.com/grafana/alloy/issues/1905: resolved by https://github.com/grafana/alloy/pull/4098
[...]

PR #4418, opened 2025-09-12
Add metadata integration tests for wal.queue

kgeckhart:
#### PR Description

Sets up all the necessary infrastructure to write metadata integration tests starting with write.queue

The biggest challenges to writing a metadata integration test comes from limitations in the metadata api (https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata),
1. It can only be filtered by metric names
    - We usually use labels for test isolation and that cannot be done here
    - I modified prom-gen so that it could expose metrics with a prefix so we can query unique metric names per test
2. There's no timestamp filtering - so at the end if the asserts pass I still fail it when running in stateful

Related to: https://github.com/grafana/alloy/issues/547

#### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/cmd/integration-tests/common/metadata_asserts.go
+++ b/internal/cmd/integration-tests/common/metadata_asserts.go
@@ -38,7 +38,7 @@ func MimirMetadataTest(t *testing.T, expectedMetadata map[string]Metadata) {
 	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		metricMetadata, err = GetMetadata("")
 		assert.NoError(c, err)
-		assert.Subset(t, maps.Keys(metricMetadata.Data), expectedMetricsWithMetadata, "did not find metadata for the expected metrics")
+		assert.Subset(c, maps.Keys(metricMetadata.Data), expectedMetricsWithMetadata, "did not find metadata for the expected metrics")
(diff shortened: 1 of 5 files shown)


kgeckhart:
Draft again while I figure out what CI doesn't like that works okay locally

internal/cmd/integration-tests/common/metadata_asserts.go
```diff
@@ -0,0 +1,82 @@
...
+
+	expectedMetricsWithMetadata := make([]string, 0, len(expectedMetadata))
+	for metricName := range expectedMetadata {
+		expectedMetricsWithMetadata = append(expectedMetricsWithMetadata, metricName)
+	}
+
+	var metricMetadata MetadataResponse
+	var err error
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		metricMetadata, err = GetMetadata("")
```
ptodev: Do we need the `optionalMetricName` param for `GetMetadata`? This is the only time `GetMetadata` is used, and it doesn't use it.

kgeckhart: Ahh good catch, I had plans that involved both that ended up being unnecessary. I'll remove it.

PR #4422, opened 2025-09-15
WIP: Introduce Prometheus StringLabel Compatability

blewis12:
Currently includes an update to the Loki dependency, as well as the `grafana/prometheus` pin version and removes the `prometheus/otlptranslator` version pin (details in [commit message](https://github.com/grafana/alloy/commit/19e47deaab0b27a3efd76d5802eabf71917591f8))

These changes so far do not enable string labels at all, but fix dependency issues from pulling in string label compatibility changes from more recent versions of Loki


blewis12:
It was decided that first we can use the labels.Labels API in Alloy - which is backwards compatible, and would enable us to switch to string labels with no compilation errors. Once this is done and merged, we can revisit this update to dependencies to officially enable string labels and measure any performance changes (hopefully improvements)

Issue #3265, opened 2025-04-10
Alloy graph: handle long component names/labels


wildum:
For some component names and for long labels, the text will go outside the box. This is not very pretty, we should find a way to improve it

Cropping the names/ids is probably not a good option because if you cant identify the nodes it defeats the purpose of using the graph of debugging

maxlemieux:
Example here from a k8s-monitoring alloy-metrics graph. It looks like we set the value here: https://github.com/grafana/alloy/blob/main/internal/web/ui/src/features/graph/buildGraph.ts#L11

<img width="739" height="1004" alt="Image" src="https://github.com/user-attachments/assets/f0811a90-a2b2-4dca-b9ee-f66680ae6e75" />

blewis12:
Fiddling around with this locally to see if I can get something pretty  - I can submit a draft PR tomorrow for feedback :)

PR #4426, opened 2025-09-15
Ensure Long Component Names in Alloy UI Don't Overflow Nodes in Graph

blewis12:
#### PR Description

First pass at addressing an issue where long component names overflow in the graph UI due to fixed sizing of the nodes. This change introduces a helper function which makes an educated guess at how wide the node should be based on the length of the name/label values. A before and after:

Screenshot 2025-09-15 at 13 05 31
Screenshot 2025-09-15 at 13 02 36

https://github.com/grafana/alloy/issues/3265

#### Notes to the Reviewer
I couldn't figure out a way to get an accurate font size to use in the calculation, so I used an "average" value of 7.5. It seems to work from what I can see, but of course would be better to have a more accurate/flexible value there.

#### PR Checklist
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,8 @@ Main (unreleased)
 
 ### Enhancements
 
+- Ensure text in the UI does not overflow node boundaries in the graph. (@blewis12)
+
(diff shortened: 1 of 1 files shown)


jharvey10:
Nice! I think this will work for now. I'd love sometime to talk through some of the ideas I have to enhance the nodes on this view and increase the information density a bit.

blewis12:
@jharvey10 definitely! I would love to hear your ideas, I feel there's certainly a lot of room to expand on what we've currently got

PR #4431, opened 2025-09-15
Ensure integration test failures fail build and remove broken tests

kgeckhart:
#### PR Description

Ensure we will fail when an integration test fails and removes the set of integration tests that are currently broken (looks like it was the the otel upgrade that broke them).

This also drops the usage of a buffered channel for test logs. It assumes that each test will only have 1 log but if the test fails and the alloy container cannot be cleaned up it will cause 2 error logs. Seemed safer to switch to a slice with a mutex.

#### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/cmd/integration-tests/tests/beyla/beyla_test.go
+++ b/internal/cmd/integration-tests/tests/beyla/beyla_test.go
@@ -2,28 +2,22 @@ package main
 
 import (
 	"testing"
-
-	"github.com/grafana/alloy/internal/cmd/integration-tests/common"
 )
 
 func TestBeylaMetrics(t *testing.T) {
-	var beylaMetrics = []string{
-		"beyla_internal_build_info",                // check that internal Beyla metrics are reported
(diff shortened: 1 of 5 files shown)


internal/cmd/integration-tests/tests/beyla/beyla_test.go
```diff
@@ -2,28 +2,22 @@ package main
...
-
-func TestBeylaMetricsAndTraces(t *testing.T) {
-	// Run both metrics and traces tests together
-	t.Run("metrics", TestBeylaMetrics)
-	t.Run("traces", TestBeylaTraces)
+	//tags := map[string]string{
+	//	"service.name": "main", // This should match the instrumented app
+	//}
+	// common.TracesTest(t, tags, "beyla")
 }
```
kgeckhart, line 29: Remove because it increases test run time without testing anything different. All these tests run with the same backend storage so if Metrics and Traces pass individually this will pass too and if one fails this just makes it take longer for the run to fail.

PR #4432, opened 2025-09-15
database_observability: switch schema after fetching explain plan

cristiangreco:
WIP

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -652,7 +652,10 @@ func (c *ExplainPlans) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) (
 	if err != nil {
 		return nil, fmt.Errorf("failed to get connection: %w", err)
 	}
-	defer conn.Close()
+	defer func() {
```
matthewnolf: nit: Can we put this after the the `USE qi.schemaName` so it's clear the operation + undo operation

cristiangreco: Do you mean having two separate `defer`? (one for `conn.Close()` and one for `USE performance_schema`)

I'm fine with that, although `USE performance_schema` is harmless even if `USE qi.schemaName` fails.

matthewnolf: No, just a single defer, but placing it after the thing it undoes for readability

rgeyer: Yeah, I would agree, having it after the execution of the use statement makes it a little more clear what it's doing.

A comment could do the same thing,.. Both might be warranted. so that it doesn't get undone some time in the future. 🤔

rgeyer (approved):
Aside from the aforementioned nit, LGTM. Thanks for this!

internal/component/database_observability/mysql/collector/explain_plans.go
cristiangreco: The problem I see with a single defer is that if `USE qi.schemaName` fails, then we return an error from this method without having called `conn.Close()`.

So I was thinking that either we have two `defer`, like:

```go
func (c *ExplainPlans) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([]byte, error) {
	conn, err := c.dbConnection.Conn(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to get connection: %w", err)
	}
	defer conn.Close()

	useStatement := fmt.Sprintf("USE `%s`", qi.schemaName)
	if _, err := conn.ExecContext(ctx, useStatement); err != nil {
		return nil, fmt.Errorf("failed to set schema: %w", err)
	}
	defer func() {
		_, _ = conn.ExecContext(ctx, "USE `performance_schema`")
	}()

	// ...
```

or we add an additional `conn.Close()` in the error handling of  the `USE` statement:

```go
func (c *ExplainPlans) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([]byte, error) {
	conn, err := c.dbConnection.Conn(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to get connection: %w", err)
	}

	useStatement := fmt.Sprintf("USE `%s`", qi.schemaName)
	if _, err := conn.ExecContext(ctx, useStatement); err != nil {
		conn.Close()
		return nil, fmt.Errorf("failed to set schema: %w", err)
	}
	defer func() {
		_, _ = conn.ExecContext(ctx, "USE `performance_schema`")
		conn.Close()
	}()

	// ...
}
```

matthewnolf: I think the former is fine. Multiple defers are okay, but I think for readability having them after the original statement is most clear.

cristiangreco: I pushed a second commit, could you both PTAL again?

matthewnolf (approved):

PR #4433, opened 2025-09-15
database_observability: drop digest_text from query samples

cristiangreco:
WIP

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/query_samples.go
+++ b/internal/component/database_observability/mysql/collector/query_samples.go
@@ -47,7 +47,6 @@ SELECT
 	statements.EVENT_ID,
 	statements.END_EVENT_ID,
 	statements.DIGEST,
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/mysql/collector/query_samples_test.go
```diff
@@ -34,14 +34,13 @@ func TestQuerySamples(t *testing.T) {
 		errorLine  string
 	}{
 		{
-			name: "select query without wait event",
+			name: "select query",
```
cristiangreco, line 37: I've removed most of the test cases here because we're not parsing the `digest_text` anymore. There's parsing logic and similar test cases in the `query_tables` collector though.

matthewnolf (approved):

PR #4439, opened 2025-09-16
chore: reduce allocations for `stage.template`

kalleep:
#### PR Description
This showed up in profiles a lot for places that allocates. The most significant change should be that we pre-allocate space in the map, but we can also reuse buffer between calls.

#### Notes to the Reviewer
Benchmark old
```
BenchmarkTemplateStage
BenchmarkTemplateStage-16    	  233258	      4986 ns/op	    2057 B/op	      31 allocs/op
BenchmarkTemplateStage-16    	  213933	      4943 ns/op	    2057 B/op	      31 allocs/op
BenchmarkTemplateStage-16    	  221922	      4956 ns/op	    2057 B/op	      31 allocs/op
BenchmarkTemplateStage-16    	  208196	      4935 ns/op	    2057 B/op	      31 allocs/op
BenchmarkTemplateStage-16    	  233574	      4901 ns/op	    2057 B/op	      31 allocs/op
```

Benchmark new
```
BenchmarkTemplateStage
BenchmarkTemplateStage-16    	  240601	      4681 ns/op	    1658 B/op	      28 allocs/op
BenchmarkTemplateStage-16    	  223374	      4714 ns/op	    1658 B/op	      28 allocs/op
BenchmarkTemplateStage-16    	  246976	      4597 ns/op	    1658 B/op	      28 allocs/op
BenchmarkTemplateStage-16    	  242415	      4685 ns/op	    1658 B/op	      28 allocs/op
BenchmarkTemplateStage-16    	  248100	      4652 ns/op	    1657 B/op	      28 allocs/op
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/loki/process/stages/template.go
```diff
@@ -95,11 +101,13 @@ type templateStage struct {
 	cfgs     TemplateConfig
 	logger   log.Logger
 	template *template.Template
+	bufPool  sync.Pool
 }
 
 // Process implements Stage
-func (o *templateStage) Process(labels model.LabelSet, extracted map[string]interface{}, t *time.Time, entry *string) {
-	td := make(map[string]interface{})
+func (o *templateStage) Process(labels model.LabelSet, extracted map[string]any, t *time.Time, entry *string) {
```
blewis12, line 108: perhaps I missed this, but what is the motivation behind switching `interface` for `any`?

kalleep: any is just a type alias for interface{} https://github.com/golang/go/blob/ca0e03560df7279bc307da08db7237beb32b0d99/src/builtin/builtin.go#L97 usually you get warnings in editors (and lint if you enable). 

We should look into enable that lint and replace interface{} with any everywhere

kalleep:
>LGTM - I believe the Process() function cannot be called concurrently anyway.

I don't think so either, we could move out the pool to be global so it would be shared if you have several template stages.

CHANGELOG.md
```diff
@@ -143,6 +143,8 @@ Main (unreleased)
 
 - Reduce compressed request size in `prometheus.write.queue` by ensuring append order is maintained when sending metrics to the WAL. (@kgeckhart)
 
+- Reduce allocations for `loki.process` when `stage.temaple` is used. (@kalleep)
```
dehaansa: Typo in the stage name

PR #4440, opened 2025-09-16
pyroscope.java: Fix java log level parameter

simonswine:
The version bundled of the async profiler has no loglevel parameter:

```
ts=2025-09-16T08:16:50.898924708Z level=error component_path=/profiling.feature component_id=pyroscope.java.java_pods pid=1184752 err="failed to start: asprof failed to run: asprof failed to run /tmp/alloy-asprof-ae0261b1093f2bc4df44a87300fef98dcdebccb5/bin/asprof: exit status 1  Unrecognized option: --loglevel\n"
```

### DIFF
--- a/internal/component/pyroscope/java/loop.go
+++ b/internal/component/pyroscope/java/loop.go
@@ -245,7 +245,7 @@ func (p *profilingLoop) start() error {
 		argv = append(argv, "--lock", cfg.Lock)
 	}
 	if cfg.LogLevel != "" {
-		argv = append(argv, "--loglevel", cfg.LogLevel)
+		argv = append(argv, "-L", cfg.LogLevel)
 	}
 	if cfg.Quiet {
 		argv = append(argv, "--quiet")


simonswine:
Update: 

Seems like those arguments were confused with java agent parameters:

https://github.com/async-profiler/async-profiler/blob/70a13bcd0340f40edaeaf66378861771f19df7e7/src/main/main.cpp#L485

https://github.com/async-profiler/async-profiler/blob/70a13bcd0340f40edaeaf66378861771f19df7e7/src/arguments.cpp#L90-L91

I do still think we don't call the agent with quiet, as this only seems to be set for "collect"

https://github.com/async-profiler/async-profiler/blob/70a13bcd0340f40edaeaf66378861771f19df7e7/src/main/main.cpp#L583

Should we use collect instead of start?

korniltsev-grafanista:
> Should we use collect instead of start?

It would be switched to `start` anyway
https://github.com/async-profiler/async-profiler/blob/70a13bcd0340f40edaeaf66378861771f19df7e7/src/main/main.cpp#L534

~I don't understand what `collect` is at all. It seems like it is never handled https://github.com/search?q=repo%3Aasync-profiler%2Fasync-profiler+%22%5C%22collect%5C%22%22&type=code~ ignore this, let me take a closer look

korniltsev-grafanista:
At the moment of writing I did not know about `collect`
I added `--timeout` in hope to handle a case when alloy invokes `start` and the dies and then java program is profiled forever, likely OOM killed later.
Looks like `collect` can't be used with `--timeout`

korniltsev-grafanista:
Hm, maybe we can exploit and inject `quiet` like this :imp: 
```
argv = append(argv, "-L", cfg.LogLevel+",quiet")
```

korniltsev-grafanista:
We don't need quiet at all. 

Asprof binary launcher invokes start/stop commands with `log=/tmp/asprof-log.26077.25508` 
Thanks to that, the "Annoying"  start/stop messages are printed into the log file instead of  `STDOUT_FILENO`

 https://github.com/async-profiler/async-profiler/blob/f071146d1176e029bd2a0691ee368ac890847be4/src/profiler.cpp#L1848

We can safely ignore and do not pass quiet at all.
We can also open upstream PR/issue to add a `--quiet`  to the launcher to save 2 `write` syscalls writing the messages to the temporary log file (which I think is not very important)

korniltsev-grafanista (approved):
Maybe update docs, mentioning `quiet` is not used at the moment (unless we add the flag upstream)

PR #4452, opened 2025-09-17
pyroscope.java: allow custom asprof distributions instead of embeded one.

korniltsev-grafanista:
- Decouple `ExtractDistributions` from `asprof.Profiler`
- Demove `asprof.Profiler` alltogether - after decoupling archive extraction logic, turns out it does nothing - use `asprof.Distribution` instead everywhere.
- Add an undocumented argument to use a custom `asprof` distribution - useful for cases when there is a fix for some problem upstream, but alloy has not been updated yet, so users could have a workaround without waiting for pyroscope to update the asprof within the alloy. Also useful for development - clone asprof, run make, point alloy to the build dir.
- Update playground to allow Java profiling

This is for the next release, we should wait before RC is cut


internal/component/pyroscope/java/args.go
```diff
@@ -14,6 +14,9 @@ type Arguments struct {
 
 	TmpDir          string          `alloy:"tmp_dir,attr,optional"`
 	ProfilingConfig ProfilingConfig `alloy:"profiling_config,block,optional"`
+
+	// undocumented
```
aleks-p, line 18: is the plan to document this at a later time?

internal/component/pyroscope/java/asprof/asprof.go
```diff
@@ -20,49 +20,47 @@ var fsMutex sync.Mutex
...
 
 type Archive struct {
-	data    []byte
-	version int
-	format  int
+	data   []byte
+	format int
+}
+
+func (a *Archive) SHA1() string {
```
aleks-p: should we make this private?

internal/component/pyroscope/java/java.go
```diff
@@ -30,30 +32,42 @@ func init() {
...
 
+func New(logger log.Logger, reg prometheus.Registerer, id string, a Arguments) (*Component, error) {
+	if os.Getuid() != 0 {
+		return nil, fmt.Errorf("java profiler: must be run as root")
+	}
+	var (
+		dist asprof.Distribution
+		err  error
+	)
+	if a.Dist != "" {
```
aleks-p, line 48: Should we add some logging here to know if we are running with an embedded or a custom distribution?

internal/component/pyroscope/java/args.go
korniltsev-grafanista: it would be nice if there was a notion of experimental configuration. I am just scared to commiting to maintain this.  We can document it /  make it public separately.

PR #4454, opened 2025-09-17
pyroscope.java: add integration tests

korniltsev-grafanista:
https://github.com/grafana/alloy/pull/4440 should have been caught by a test

- ~add an integration test to the docker-compose test suite~ edit: could not make it work, will do in a followup
- add an integration test by spawning containers manually

I could not run the test locally though, containers don't build :shrug:  :(

```
 > [ui-build 4/4] RUN --mount=type=cache,target=/ui/node_modules,sharing=locked     yarn --network-timeout=1200000                                && yarn run build:
0.276 yarn install v1.22.22
0.327 [1/4] Resolving packages...
0.587 [2/4] Fetching packages...
20.66 error Error: getaddrinfo EAI_AGAIN registry.yarnpkg.com
20.66     at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26)
20.66 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
depends on https://github.com/grafana/alloy/pull/4452

### DIFF
--- a/internal/component/pyroscope/java/integration/integration_test.go
+++ b/internal/component/pyroscope/java/integration/integration_test.go
@@ -28,7 +28,6 @@ func TestPyroscopeJavaIntegration(t *testing.T) {
 	if os.Getenv("GITHUB_ACTIONS") == "true" && os.Getenv("GITHUB_JOB") != "test_pyroscope" {
 		t.Skip("Skipping Pyroscope Java integration test in GitHub Actions (job name is not test_pyroscope)")
 	}
-	t.FailNow()
 	wg := sync.WaitGroup{}
 	defer func() {
 		wg.Wait()


.github/workflows/test_pyroscope_pr.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           go-version-file: go.mod
           cache: false
 
-      - run: make GO_TAGS="nodocker" test-pyroscope
\ No newline at end of file
+      - run: sudo make test-pyroscope
```
oleg-kozlyuk-grafana, line 29: What is `sudo` needed for? A couple of concerns here:
* Not sure if GitHub runners are truly ephemeral
* It becomes easy to merge a change that only allows to run these tests as sudo later

korniltsev-grafanista: `pyroscope.java` component requires root to instrument container's FS and send signals / interact with jvm

I've added a test that requires root.

What do you suggest?

.dockerignore
```diff
@@ -7,3 +7,4 @@ internal/web/ui/build/
 packaging/windows/LICENSE
 packaging/windows/agent-windows-amd64.exe
 cmd/grafana-agent/Dockerfile
+alloy
```
simonswine, line 10: Only checked Dockerfile and Dockerfile.windows, which are not using it.

Issue #4323, opened 2025-09-01
Adopt Prometheus stringlabels


ptodev:
### Component(s)

`prometheus.*`, `discovery.*`, and `loki.*` components.

### Request

Prometheus has migrated from [regular labels](https://github.com/prometheus/prometheus/blob/2cbeef6d9582d7dee2bbe493a7cb42bda8f37553/model/labels/labels_common.go) to "[stringlabels](https://github.com/prometheus/prometheus/blob/2cbeef6d9582d7dee2bbe493a7cb42bda8f37553/model/labels/labels_stringlabels.go)". Which labels implementation to use is governed by a Go build tag. The new implementation is supposed to be more efficient, but we would need to verify this by running benchmarks for Alloy specifically.

### Use case

Improvement in efficiency.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

kalleep:
To do this we need to use Prometheus label apis everywhere, for example we cannot use a regular for loop over labels like we can for slice labels so for that case we need to use [Range](https://github.com/prometheus/prometheus/blob/c5743037ef55f29445397f26bcf9ef1c32a27642/model/labels/labels_stringlabels.go#L400).

blewis12:
As a brief note that addresses [this comment](https://github.com/grafana/alloy/pull/4455#discussion_r2359227383), I did a couple runs of the `BenchmarkRelabelComponent` test to compare a before/after of that PR being merged:

```
➜  alloy git:(fc96e92d3) ✗ benchstat baseline.txt new.txt 
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/component/loki/relabel
cpu: Apple M4
                    │ baseline.txt │              new.txt              │
                    │    sec/op    │   sec/op     vs base              │
RelabelComponent-10    3.521µ ± 0%   3.482µ ± 1%  -1.12% (p=0.002 n=6)

                    │ baseline.txt │              new.txt               │
                    │     B/op     │     B/op      vs base              │
RelabelComponent-10   1.992Ki ± 0%   2.021Ki ± 0%  +1.47% (p=0.002 n=6)

                    │ baseline.txt │             new.txt              │
                    │  allocs/op   │ allocs/op   vs base              │
RelabelComponent-10     35.00 ± 0%   34.00 ± 0%  -2.86% (p=0.002 n=6)
```

[...]

PR #4455, opened 2025-09-17
Use labels.Labels API For String Label Compatability

blewis12:
#### PR Description
Implements the Labels API from prometheus, this should _not_ change any behavior - instead this leaves us in a place that we can enable string labels (done using go tags) and in theory not encounter any compilation issues.

I have broken the changes down by commit, some have a small description as well for more information. 

**Note on sorting/builders:**
Once interesting thing I noticed was that, in the new label.Labels API, the order of labels is implied to always be correct - because of this I have removed some manual calls to `sort.Sort()` that were there before (this was also done in the[ Loki implementation](https://github.com/grafana/loki/pull/18490/files#r2216436742)). This was also caught in [this test failure](https://github.com/grafana/alloy/pull/4455/commits/b19afa44eb6442661b0994439dce19a393651e44). The main exception to this is the `labels.ScratchBuilder`, in which `.Sort()` must be called to guarantee ordering - I used the scratch builder in cases where we were building...from scratch 😄  but there's also the plain builder which is supposed to modify existing labels. 

There are some cases where we're using the plain builder (`labels.NewBuilder`) and passing through nil, essentially doing what a scratch builder is meant to do - i've kept those the same for now so that this PR is only addressing string label compatibility, but maybe it makes sense to revisit those and see if the scratch builder could be more efficient

#### Which issue this PR relates to
https://github.com/grafana/alloy/issues/4323

#### PR Checklist
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/service/labelstore/service.go
+++ b/internal/service/labelstore/service.go
@@ -155,6 +155,11 @@ func (s *Service) GetOrAddLink(componentID string, localRefID uint64, lbls label
 
 // GetOrAddGlobalRefID is used to create a global refid for a labelset
 func (s *Service) GetOrAddGlobalRefID(l labels.Labels) uint64 {
+	// Guard against bad input.
+	if l.IsEmpty() {
+		return 0
+	}
+
 	s.mut.Lock()
 	defer s.mut.Unlock()
 


blewis12:
Playing around with some profiling, scraping every 15s for about 30s against `avalanche`. We see higher consumption in `(*ScratchBuilder).UnsafeAddBytes` on this branch it seems. This branch replaces a lot of instances where we manually construct a slice to use the Builder helpers, in particular the ScratchBuilder - additionally, i can see [here](https://github.com/grafana/alloy/blob/8977cc84dea3db82deae23a995ef4d1474b4d34c/internal/component/discovery/target.go#L89) that benchmarking was already done, and that in fact the ScratchBuilder/Builder perform _less_ optimal than the slice construction (without string labels enabled of course). I did a couple runs of the profiling and this seemed to be a consistent thing

Screenshot 2025-09-18 at 12 45 32
Screenshot 2025-09-18 at 12 44 54


Wondering if this is expected, and if we can eventually expect to see this offset by the gains we see from string labels being enabled

thampiotr:
> Wondering if this is expected, and if we can eventually expect to see this offset by the gains we see from string labels being enabled

I think the difference is pretty small - thanks for checking!

internal/component/loki/relabel/relabel.go
```diff
@@ -239,19 +239,18 @@ func (c *Component) relabel(e loki.Entry) model.LabelSet {
 }
 
 func (c *Component) process(e loki.Entry) model.LabelSet {
-	var lbls labels.Labels
+	builder := labels.NewScratchBuilder(len(e.Labels))
```
thampiotr, line 242: The previous implementation was creating new `labels.Labels` here, so this should be similar, but I think there is scope for optimising this path (and I think it's a hot path) with reusing the same builder and calling [Reset()](https://github.com/prometheus/prometheus/blob/d5cc5e2738e2d0bd80c9ea31ba48fae80613917c/model/labels/labels_stringlabels.go#L467) on it. 

But if there is no regression, it's probably better to do that in a separate PR to keep changes smaller and independent.

internal/component/pyroscope/write/write.go
```diff
@@ -661,31 +660,38 @@ func validateLabels(lbls labels.Labels) error {
...
+			err = fmt.Errorf("invalid label value for %s: %s", l.Name, l.Value)
+			return
 		}
 
 		// Skip label name validation for pyroscope reserved labels
 		if l.Name != pyroscope.LabelName {
 			// Validate label name
-			if err := labelset.ValidateLabelName(l.Name); err != nil {
-				return fmt.Errorf("invalid label name: %w", err)
+			// two errors is ok?
```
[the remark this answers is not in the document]
thampiotr: You were wondering if it's okay to wrap an error? I think that's fine, gives some context 👍

internal/service/labelstore/service.go
```diff
@@ -158,10 +158,7 @@ func (s *Service) GetOrAddGlobalRefID(l labels.Labels) uint64 {
 	s.mut.Lock()
 	defer s.mut.Unlock()
 
-	// Guard against bad input.
-	if l == nil {
-		return 0
-	}
+	// think: can we actually remove this? seems like yes
```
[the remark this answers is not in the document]
thampiotr: I don't mind. If we keep it, let's put this before the mutex tho 😅

thampiotr (commented):
LGTM so far!

internal/service/labelstore/service.go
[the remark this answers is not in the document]
blewis12: lol woops, I can remove - that was just a note to myself

internal/component/pyroscope/write/write.go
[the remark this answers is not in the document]
blewis12: ah this was another note to myself that I thought I removed

internal/service/labelstore/service.go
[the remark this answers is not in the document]
blewis12: https://github.com/grafana/alloy/pull/4455/commits/848cd72eea5b2e44d629bfe447420e27d591d8b5

internal/component/pyroscope/write/write.go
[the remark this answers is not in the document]
blewis12: https://github.com/grafana/alloy/pull/4455/commits/848cd72eea5b2e44d629bfe447420e27d591d8b5

internal/component/loki/relabel/relabel.go
blewis12: More of a clarifying question - but how do we know if something is a hot path? I ask mostly because when I did my profiling I couldn't see this method in the top consumers anywhere, but maybe there's a different way to tell this?

blewis12: And cool, that pattern makes sense - I can see it done in the [prometheus codebase as well](https://github.com/prometheus/prometheus/blob/d5cc5e2738e2d0bd80c9ea31ba48fae80613917c/tsdb/record/record.go#L294)

I can make those optimizations in a separate PR to keep things independent 👍

internal/service/labelstore/service.go
```diff
@@ -158,11 +158,6 @@ func (s *Service) GetOrAddGlobalRefID(l labels.Labels) uint64 {
 	s.mut.Lock()
 	defer s.mut.Unlock()
 
-	// Guard against bad input.
-	if l == nil {
-		return 0
-	}
-
```
kalleep, line 165: Are we sure this won't cause any issues? 

If we keep to be on the safe side we can just move it before the lock

blewis12: You're right, it's probably best to keep that in - I had removed since now `nil` is not a possible value anymore (and so therefore `l.Hash()` won't cause a panic, which I assume is why this check was here before). I will add it back just to be safe

blewis12: https://github.com/grafana/alloy/pull/4455/commits/915e8059486ee93234f202d4f8bd91159f3e0cc2

internal/component/loki/relabel/relabel.go
kgeckhart: As far as what to look for to know it's a hot path, this one is by feeling. Relabeling is a really popular use case as is shipping loki logs with alloy. It would only pop up in profiles if you were using a the component.

blewis12: I see 👍 I had enabled the component to see what popped up profile-wise - and saw it when poking around in pyroscope. But makes sense in general feel-wise

kalleep (approved):

Issue #4409, opened 2025-09-12
macos CI tests are flaky


korniltsev-grafanista:
### Add a description
Example: https://github.com/grafana/alloy/actions/runs/17663109418/job/50199871112?pr=4408

Consider retrying them automatically (instead of forcing users to hit restart button manually)

For example with the help of https://github.com/gotestyourself/gotestsum?tab=readme-ov-file#re-running-failed-tests

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

kalleep:
Yes I have a fix for it here https://github.com/grafana/alloy/pull/4410

kalleep:
Both https://github.com/grafana/alloy/pull/4414 and https://github.com/grafana/alloy/pull/4410 addressed flaky tests.

Everything seems to work correctly now

korniltsev-grafanista:
https://github.com/grafana/alloy/actions/runs/17761287107/attempts/1?pr=4436

kalleep:
@korniltsev-grafanista yeah i started to notice this too. Now the build job is really flaky...

korniltsev-grafanista:
another example https://github.com/grafana/alloy/actions/runs/17817446933/job/50653415948?pr=4454

kalleep:
Not limited to macos https://github.com/grafana/alloy/actions/runs/17821058195/job/50663452349?pr=4458

But yes this is a flaky test, will try do reproduce this one locally.

We have had some problems with flaky tests from time to time. I wonder if we should keep a more general open issue for them that we can update when we discover new ones 🤔

kalleep:
I will close this one. I opened a tracking issue for flake tests that we should update when we discover them https://github.com/grafana/alloy/issues/5142

PR #4460, opened 2025-09-18
chore: fix flaky test in loki.source.api

kalleep:
#### PR Description
https://github.com/grafana/alloy/actions/runs/17817446933/job/50653415948?pr=4454
https://github.com/grafana/alloy/actions/runs/17821058195/job/50663452349?pr=4458

This test could fail in two ways:
1. getFreePort returned the same port for grpc and http
2. getFreePort returned the same port for http when port change test ran, this would not trigger a restart

two calls to getFreePort can return the same port so I created an extra function that would exclude already "taken" ports for the test.

I also changed to use `log.NewNopLogger` to reduce all the noise around these tests. 

Reported here https://github.com/grafana/alloy/issues/4409

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


internal/component/loki/source/api/api_test.go
```diff
@@ -609,8 +612,15 @@ func testArgsWithPorts(httpPort int, grpcPort int) Arguments {
...
 }
 
-func getFreePort(t *testing.T) int {
-	port, err := freeport.GetFreePort()
-	require.NoError(t, err)
-	return port
+func getFreePort(t *testing.T, exclude ...int) int {
+	for range 10 {
+		port, err := freeport.GetFreePort()
+		require.NoError(t, err)
```
blewis12:
```suggestion
    const maxRetries = 10
	for range maxRetries {
		port, err := freeport.GetFreePort()
		require.NoError(t, err)
```

nit: extra variable but slightly more idiomatic?

internal/component/loki/source/api/api_test.go
```diff
@@ -609,8 +612,16 @@ func testArgsWithPorts(httpPort int, grpcPort int) Arguments {
 	}
 }
 
-func getFreePort(t *testing.T) int {
-	port, err := freeport.GetFreePort()
-	require.NoError(t, err)
-	return port
+func getFreePort(t *testing.T, exclude ...int) int {
+	const maxRetries = 10
+	for range maxRetries {
```
kgeckhart, line 617: Looking at the implementation here I think this has the same race condition I've seen else where you ask for an unused port it will lock it but as soon as the function executes it unlocks it. Another test could call and lock the same port before the server which request the port could use it causing an error.

I found switching to use random ports in tests to be a more effective solution.

kgeckhart (approved):
As discussed, 🤞 this patches it and if not we can look at refactoring to random ports

PR #4464, opened 2025-09-18
Add docs note about converting promtail daemonset configs

dehaansa:
#### PR Description
We see concerns about the resource consumption of Alloy in community slack/github somewhat often for users coming from promtail. Often these users are experiencing a limitation of the converter not converting with the context of the deployment environment (k8s daemonset) leading to Alloy doing a lot of duplicate work vs Promtail.

While it would be great to improve the converter, an initial step is to call out this limitation on the docs for converting from Promtail.


kalleep (approved):

docs/sources/set-up/migrate/from-promtail.md
```diff
@@ -179,6 +179,7 @@ The following list is specific to the convert command and not {{< param "PRODUCT
   Make sure that you use the new metric names, for example, in your alerts and dashboards queries.
 * The logs produced by {{< param "PRODUCT_NAME" >}} differ from those produced by Promtail.
 * {{< param "PRODUCT_NAME" >}} exposes the {{< param "PRODUCT_NAME" >}} [UI][], which differs from the Promtail Web UI.
+* If you are converting a promtail configuration for a kubernetes daemonset, you will likely want to [modify the generated configuration][single-node-discovery] to ensure `discovery.kubernetes` only discovers pods residing on the same node as the {{< param "PRODUCT_NAME" >}} pod.
```
clayton-cornell:
```suggestion
* If you are converting a Promtail configuration for a Kubernetes daemonset, [modify the generated configuration][single-node-discovery] to ensure `discovery.kubernetes` only discovers Pods residing on the same node as the {{< param "PRODUCT_NAME" >}} Pod.
```

Fix capitalization and make it more explicit

PR #4471, opened 2025-09-19
fix flaky test

kalleep:
#### PR Description
Yet another flaky test last observed https://github.com/grafana/alloy/actions/runs/17852325325/job/50763467503.

I have seen this a few times.

I reworked the test a bit and it's also faster now that we don't sleep for 5 seconds.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/prometheus/operator/common/component_test.go
+++ b/internal/component/prometheus/operator/common/component_test.go
@@ -8,48 +8,43 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/prometheus/operator"
 	"github.com/grafana/alloy/internal/service/cluster"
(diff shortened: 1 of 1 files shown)


internal/component/prometheus/operator/common/component_test.go
```diff
@@ -8,48 +8,42 @@ import (
...
-	}
+func (m crdManagerFactoryHungRun) New(_ component.Options, _ cluster.Cluster, _ log.Logger, _ *operator.Arguments, _ string, _ labelstore.LabelStore) crdManagerInterface {
+	return &crdManagerHungRun{m.running, m.contextCenceled, m.stopRun}
 }
 
 type crdManagerHungRun struct {
-	stopRun chan struct{}
-	onRun   chan struct{}
+	running         *atomic.Bool
+	contextCenceled *atomic.Bool
```
kgeckhart, line 37:
```suggestion
	contextCancelled *atomic.Bool
```

kgeckhart (commented):
Minor misspelling but otherwise makes sense

internal/component/prometheus/operator/common/component_test.go
kalleep: So you can spell it either _Cancelled_ or _Canceled_. In golang context package Canceled is used so I went with that.

kgeckhart (approved):

PR #4474, opened 2025-09-19
database_observability: Add postgres explain plan collector

rgeyer:
#### PR Description
Adds a collector and config for explain plans in postgres.

#### Notes to the Reviewer
Due to the way that postgres handles inbound db connections, the explain plans needs to actually create a new connection, specifying the target DB/schema name. As a result the original DSN is being passed into the collector, and parsed to replace the db name.

There was some question (at least in my own mind) as to whether we should use `pg_stat_activity` or `pg_stat_statements` to source queries which we will run explain plans on. The assertion was that `pg_stat_statements` may be more complete. However, the explain plan collector depends upon fetching only queries which have run since the last iteration of the explain plan collector. `pg_stat_statements` does not have a timestamp for when the last execution of the query happened.

My (very unscientific) test reveals that `pg_stat_activity` does seem to have the superset of queries, so we'll be using that.
```
postgres=> select query_id from pg_stat_activity where query_id not in (select queryid from pg_stat_statements);
      query_id
---------------------
 1450693418307658559
 3373908757839370106
(2 rows)

postgres=> select queryid from pg_stat_statements where queryid not in (select query_id from pg_stat_activity);
 queryid
---------
(0 rows)
```

#### PR Checklist

- [X] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


rgeyer:
~Oppsies, needs docs.. I'll write those.~

Added

gaantunes:
>My (very unscientific) test reveals that pg_stat_activity does seem to have the superset of queries, so we'll be using that.

I am not sure that is entirely true... but your test is indeed interesting.
My understanding is `pg_stat_activity` will only contain currently running queries, so in theory any query that does not run often would very probably be missed. My gut feeling is we should use `pg_stat_statements`, which tracks statistics of all queries executed in the database, up to a set limit.

rgeyer:
> > My (very unscientific) test reveals that pg_stat_activity does seem to have the superset of queries, so we'll be using that.
> 
> I am not sure that is entirely true... but your test is indeed interesting. My understanding is `pg_stat_activity` will only contain currently running queries, so in theory any query that does not run often would very probably be missed. My gut feeling is we should use `pg_stat_statements`, which tracks statistics of all queries executed in the database, up to a set limit.

Right.. Okay, that's (somewhat) what I was afraid of. `pg_stat_statements` does not seem to have any "last executed" field for queries, so the logic I have in the explain_plans to only grab "queries executed since the last update interval" isn't really feasible.

I suppose for postgres, I can just fetch and run explain plans for *everything* in `pg_stat_statements`.

I can for sure change the query to use that table/extension. Just not sure how that might impact performance 🤔

fridgepoet:
I wonder if `pg_stat_activity` is a good first iteration for this?

> I suppose for postgres, I can just fetch and run explain plans for everything in pg_stat_statements.

What about tracking which statements were already processed in-memory? 
We could create a hash from each statement and store that in a map. If that entry already exists, then we don't fetch and run the explain plan again.
There might already be a field which is a hash of the normalized query.
This would reduce overhead compared to running explain plans for every statement on every collection interval.
Just an idea, might be for a second/third iteration of this.

gaantunes:
While we do not have a last executed field, we do have the `calls` field that increases on each query execution. We can use this field along with a hashmap as @fridgepoet suggested, to only get plans for queries that got executed within each scrape interval.

Since stats can be reset, we can use `stats_since` (available from v17 onward) or `stats_reset` from `pg_stat_statements_info` (available from v14 onward) to check track it if `calls` went down between scrapes.

gaantunes:
> I wonder if pg_stat_activity is a good first iteration for this?

It can be a first iteration, yes. But will be quite faulty, as is our sampling information right now. We just sample the most executed queries, leaving a lot behind.

rgeyer:
Brilliant! Thank you for the conversation on this.

> We could create a hash from each statement and store that in a map. If that entry already exists, then we don't fetch and run the explain plan again.

The problem with this is that each query will *literally* only ever have an explain plan run once. Even if the same query id/hash has been executed more recently. So, effectively you'd get a batch of explain plans when alloy starts up, then they'd grow stale pretty quickly.

I'll go ahead and implement the suggestion from @gaantunes using `calls` which I hadn't noticed in the schema. This should provide the desired behavior.

cristiangreco:
I personally agree with using the proposed approach to track queryId and calls. However have approved the PR meanwhile because I think that change can potentially be worked in a second PR while we get this in meanwhile.

rgeyer:
> I personally agree with using the proposed approach to track queryId and calls. However have approved the PR meanwhile because I think that change can potentially be worked in a second PR while we get this in meanwhile.

Yeah, agreed. With all the activities going on, probably good to get these flowing to spot any other issues. I'll merge it and get the follow-up PR ASAP.

Issue #4020, opened 2025-07-22
Memory leak in GCP Stackriver Exporter component(for Dataflow metrics) in Alloy - Suspected memory leak due to new set of values for job_id label each day


zaidqureshi10:
### What's wrong?

Description:

We’re facing high and gradually increasing memory usage in an Alloy-based pod deployed for scraping Google Cloud Dataflow job metrics via the Stackdriver Prometheus exporter. After a detailed investigation, we believe the issue is caused by the job_id label in the scraped metrics. Over time, Alloy is not clearing the old job_id values from memory, leading to accumulation

Observations:

The job_id label is included in all metrics scraped from Dataflow.
In our use case, each new Dataflow job gets a new unique job_id, and jobs are constantly created throughout the day.
As a result, the job_id label value set continuously grows without bound, introducing very high cardinality over time.
We observed that memory usage in the Alloy pod continuously increases and does not stabilize even after hours of runtime.

Troubleshooting & Analysis:

1. Comparative Testing With and Without job_id

To verify our suspicion, we deployed a second Alloy pod with the same configuration but modified it to exclude the job_id label using relabeling rules.

Original pod (with job_id): Memory usage increased steadily over time, reaching to the poing of consuming 100% of allocated memory.
Test pod (without job_id): Memory usage remained stable and low (10-15%).

This strongly suggests that high cardinality from job_id is the primary contributor to memory pressure.

2. Heap Profile (pprof) Analysis

We collected a heap profile (heap.pprof) from the high-memory pod and analyzed it with go tool pprof. Top memory consumers were:

FunctionMemorylabelstore.GetOrAddGlobalRefID, labelstore.GetOrAddLink, and labels.Builder.Labels

These functions are heavily involved in label management, further confirming that label churn — especially due to unique and continuously growing job_ids — is creating a memory footprint that does not get cleaned up.

Conclusion

The job_id label, which reflects a unique ID for every Dataflow job, continuously introduces new label combinations in the metric stream. Over time leading to high memory usage and potential memory leaks.

Once we removed this label via relabeling, the memory usage dropped and stabilized, clearly confirming the cause.

Request for Feedback:

Please let us know of any solutions we can implement.

<img width="1026" height="473" alt="Image" src="https://github.com/user-attachments/assets/f65fe7db-6982-42f3-b926-42db3b48beb3" />
[...]

PR #4485, opened 2025-09-23
Remove labelstore usage from relabel

kgeckhart:
#### PR Description
Drops direct labelstore.LabelStore usage from `prometheus.relabel`. 

This component is wrapped in a `prometheus.Fanout` and `prometheus.Interceptor` which will ensure any series coming in/leaving the component will be properly managed in the labelstore.Store. Those two components first check if it's necessary to use labelstore.LabelStore while relabel will always call it. I updated an existing test to show this is the case. 

Removing the relabel usage of labelstore should reduce mutex contentions by ~10% (25% if go internal mutex contention from `runtime._LostContendedRuntimeLock` is excluded),
image

At first I thought relabeling using `labelstore.LabelStore` in this manner was the root of https://github.com/grafana/alloy/issues/4020. But it seems as thought the amount of labels they are seeing is valid due to the high cardinality of the label they are using. These changes should have a positive impact on their use case but there does not appear to be a "fix".

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/relabel/relabel.go
+++ b/internal/component/prometheus/relabel/relabel.go
@@ -112,10 +112,10 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	}
 
 	data, err := o.GetServiceData(labelstore.ServiceName)
-	ls := data.(labelstore.LabelStore)
 	if err != nil {
 		return nil, err
 	}
+	ls := data.(labelstore.LabelStore)
 	c := &Component{
 		opts:               o,
 		cache:              cache,


internal/component/prometheus/relabel/relabel.go
```diff
@@ -112,13 +112,13 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	}
 
 	data, err := o.GetServiceData(labelstore.ServiceName)
+	ls := data.(labelstore.LabelStore)
```
kalleep: This probably infallible but lets move the type asset after `if err != nil` check

kalleep (approved):

PR #4494, opened 2025-09-25
chore: ignore canceled context to prevent flaky test

kalleep:
#### PR Description
Really flaky test. We just wait for a initial request to be made and when it's done we abort.

This will often lead to [this call failing](https://github.com/grafana/alloy/blob/c0c7d32b0ec1c41d1e10aa43d23e1538fbe9c8ad/internal/service/remotecfg/remotecfg.go#L139) and return `context.Canceled` and we should not care about that here.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


blewis12 (approved):
nice - I was just looking at this test failure that popped up in a separate PR 😄

PR #4511, opened 2025-09-29
feat(Helm Chart): Make service.externalTrafficPolicy avaliable

### LINKED ISSUE #4510: Helm chart - service.externalTrafficPolicy be added as a valid values option
### Component(s)

_No response_

### Request

I'd like the following helm value be valid and respected in a similar manner to internalTrafficPolicy.
```
service:
  externalTrafficPolicy: local
```

### Use case

When deploying Alloy to receive TCP logs from sources outside the Kubernetes cluster, it is important for Alloy to be able to resolve the hostname or IP of the log source. Currently, with the default externalTrafficPolicy: cluster, the source of incoming logs is obscured due to the behavior of Kubernetes service routing.

By supporting externalTrafficPolicy: local, Alloy could preserve the client source IP, enabling accurate source resolution for TCP log ingestion. This is particularly useful for environments that rely on client IPs for logging, auditing, or filtering.

Similar behavior is already supported for internalTrafficPolicy, so extending support to externalTrafficPolicy would be consistent.

### Tip

at-blacknight:
#### PR Description
Minor changes in order to be able to set alloy service externalTrafficPolicy through the helm chart.

Fixes #4510

#### Notes to the Reviewer
Very green to OSS PR's in general so please be kind, let me know what I should do differently.

#### PR Checklist

- [ x] CHANGELOG.md updated
- [ x] Documentation added


kalleep (dismissed):
LGTM

kalleep:
Pretty sure you need to re-generate helm test files `make generate-helm-test` and possibly docs `make generate-helm-docs`

at-blacknight:
@kalleep Thanks for getting the CI going. 

Seems to be passing checks now. 

I'm unsure if we should be silently swallowing the external traffic policy when configured for clusterIP (how I currently have it)
OR
error when attempting to configure externalTrafficPolicy when using clusterIP

not sure which way to lean.

kalleep (approved):

ptodev (approved):

PR #4516, opened 2025-09-30
chore: point out breaking change in prometheus.scrape for v1.11

kalleep:
#### PR Description
In https://github.com/grafana/alloy/pull/4044  the default value for `scrape_native_histograms` was changed from true to false. This setting alone was not enough to enable native histogram scraping but you would also have to configure `scrape_protocols` to start with `PrometheusProto`. 

In the mentioned pr we changed this behavior so that just configuring scrape_protocols is not enough by default anymore e.g. 
```
prometheus.scrape "test" {
  scrape_protocols = ["PrometheusProto"]
}
``` 

But now you need to explicitly enable it. We also added validation for `scrape_protocols` when this is configured and automatically change the [default protocols](https://github.com/grafana/alloy/blob/777fa641a1a1ac960782e33ca7ef64937ff4e995/internal/component/prometheus/scrape/scrape.go#L191-L195).
```
prometheus.scrape "test" {
  scrape_native_histograms = true
}
``` 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/release-notes.md
+++ b/docs/sources/release-notes.md
@@ -36,8 +36,7 @@ See the upstream [Prometheus v3 migration guide](https://prometheus.io/docs/prom
 
 ### Breaking changes in `prometheus.scrape`
 
-`scrape_native_histograms` attribute for `prometheus.scrape` is now set to `false`,
-whereas in previous versions of Alloy it would default to `true`. 
+`scrape_native_histograms` attribute for `prometheus.scrape` is now set to `false`, whereas in previous versions of Alloy it would default to `true`. 
 This means that it is no longer enough to just configure `scrape_protocols` to start with `PrometheusProto` to scrape native histograms - `scrape_native_histograms` has to be enabled. 
 If `scrape_native_histograms` is enabled, `scrape_protocols` will automatically be configured correctly for you to include `PrometheusProto`.
 If you configure it explicitly, Alloy will validate that `PrometheusProto` is in the `scrape_protocols` list.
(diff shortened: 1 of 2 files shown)


docs/sources/release-notes.md
```diff
@@ -34,6 +34,24 @@ Prometheus dependency had a major version upgrade from v2.55.1 to v3.4.2.
 
 See the upstream [Prometheus v3 migration guide](https://prometheus.io/docs/prometheus/3.4/migration/) for more details.
 
+### Breaking changes in `prometheus.scrape`
+
+`scrape_native_histograms` attribute for `prometheus.scrape` is now set to false previously default was true. This means that it is no longer enough to just configure `scrape_protocols` to start with `PrometheusProto` to scrape native histograms but it has to be enabled. If `scrape_native_histograms` is enabled without `scrape_protocols` set it will be configured correctly for you, otherwise Alloy will validate that `PrometheusProto` is in the `scrape_protocols` list.
```
ptodev:
```suggestion
`scrape_native_histograms` attribute for `prometheus.scrape` is now set to `false`,
whereas in previous versions of Alloy it would default to `true`. 
This means that it is no longer enough to just configure `scrape_protocols` to start with `PrometheusProto` to scrape native histograms - `scrape_native_histograms` has to be enabled. 
If `scrape_native_histograms` is enabled, `scrape_protocols` will automatically be configured correctly for you to include `PrometheusProto`.
If you configure it explicitly, Alloy will validate that `PrometheusProto` is in the `scrape_protocols` list.
```

PR #4523, opened 2025-09-30
Db-o11y: Cache pg sample and wait events and send finalized entries

gaantunes:
#### PR Description

Shifting from sending entries on every scrape, to maintain the entries in memory and only send it when the query ends the execution.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/query_samples.go
+++ b/internal/component/database_observability/postgres/collector/query_samples.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/go-kit/log"
@@ -21,6 +22,34 @@ const (
 	OP_QUERY_SAMPLE       = "query_sample"
 	OP_WAIT_EVENT         = "wait_event"
 )
+
+// SampleKey uniquely identifies a running sample (query execution instance)
+// while it is in a non-idle state.
+type SampleKey struct {
+	PID     int
+	QueryID int64
+	XID     int32
+}
+
+// WaitEventOccurrence tracks a continuous occurrence of the same wait event
+// with the same blocked_by_pids set.
+type WaitEventOccurrence struct {
+	WaitEventType string
+	WaitEvent     string
+	BlockedByPIDs []int64 // normalized set (sorted, unique)
+	LastWaitTime  string  // last stateDuration seen for this occurrence
+	LastState     string
+	LastTimestamp time.Time
+}
+
+// SampleState holds the in-memory buffered state for a running sample.
+type SampleState struct {
+	LastRow        QuerySamplesInfo
+	LastSeenAt     time.Time
+	LastCpuTime    string // last cpu_time observed under CPU condition
+	WaitEvents     []WaitEventOccurrence
+	OpenOccurrence *int // index into WaitEvents slice; nil if none open
+}
 
 const selectPgStatActivity = `
 	SELECT
@@ -101,6 +130,9 @@ type QuerySamples struct {
 	running *atomic.Bool
 	ctx     context.Context
 	cancel  context.CancelFunc
+
+	// in-memory state of running samples
+	samples map[SampleKey]*SampleState
 }
 
 func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
@@ -111,6 +143,7 @@ func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
 		disableQueryRedaction: args.DisableQueryRedaction,
 		logger:                log.With(args.Logger, "collector", QuerySamplesCollector),
 		running:               &atomic.Bool{},
+		samples:               map[SampleKey]*SampleState{},
 	}, nil
 }
 
@@ -175,6 +208,9 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
 	}
 	defer rows.Close()
 
+	activeKeys := map[SampleKey]struct{}{}
+	idleKeys := map[SampleKey]struct{}{}
+
 	for rows.Next() {
 		sample := QuerySamplesInfo{}
 		err := rows.Scan(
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
+		level.Error(c.logger).Log("msg", "failed to iterate pg_stat_activity rows", "err", err)
+		return err
+	}
+
+	c.finalizeSamples(activeKeys, idleKeys)
+	return nil
+}
+
+// queryPgStatActivity executes the query against pg_stat_activity
+func (c *QuerySamples) queryPgStatActivity(ctx context.Context) (*sql.Rows, error) {
```
matthewnolf: Why does this warrant being its own function?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
-	if !sample.DatabaseName.Valid {
-		return fmt.Errorf("database name is not valid. sample set: %+v", sample)
+// normalizePIDs returns a sorted unique slice for stable set comparisons/printing.
+func normalizePIDs(pids pq.Int64Array) []int64 {
+	set := map[int64]struct{}{}
+	for _, pid := range pids {
+		set[pid] = struct{}{}
+	}
+	out := make([]int64, 0, len(set))
+	for pid := range set {
```
matthewnolf: Why are we looping twice?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
+		waitEventFullName,
+		we.BlockedByPIDs,
+		state.LastRow.QueryID.Int64,
+		queryText,
+	)
+}
+
+// calculateDuration returns a formatted duration string between a nullable time and current time
+func calculateDuration(nullableTime sql.NullTime, currentTime time.Time) string {
+	if nullableTime.Valid {
```
matthewnolf: Nit: Can we error early, and otherwise return the current time. This keeps all errors indented (within an if) and the normal/happy flow easier to follow

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
+	}
+	state.LastRow = sample
+	state.LastSeenAt = sample.Now
+	// Close any open wait event
+	state.tracker.CloseOpen()
+}
+
+// emitAndDeleteSample builds final entries for a sample and removes it from memory.
+func (c *QuerySamples) emitAndDeleteSample(key SampleKey, state *SampleState) {
+	// Build and emit OP_QUERY_SAMPLE
```
matthewnolf: I feel like a lot of the godoc and inline comments are superflous

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
+		c.samples[key] = state
+	}
+	state.LastRow = sample
+	state.LastSeenAt = sample.Now
+	// Close any open wait event
+	state.tracker.CloseOpen()
+}
+
+// emitAndDeleteSample builds final entries for a sample and removes it from memory.
+func (c *QuerySamples) emitAndDeleteSample(key SampleKey, state *SampleState) {
```
matthewnolf: nit: Why have both `key` and `state` as args? Doesn't `c.samples[key]` give you `state` anyway?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -160,162 +233,341 @@ func (c *QuerySamples) Stop() {
...
 func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
-	rows, err := c.dbConnection.QueryContext(ctx, selectPgStatActivity)
+	rows, err := c.queryPgStatActivity(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to query pg_stat_activity: %w", err)
+		return err
 	}
 	defer rows.Close()
 
+	activeKeys := map[SampleKey]struct{}{}
```
matthewnolf, line 237: Is it worth having a single map keys = `map[SampleKey]SampleState`?

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: It doesn't really, will move it back to the main function.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Changing implementation to a single loop for better readibility.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: I like documenting the functions behavior, it makes the code easier to read. I can remove it if you are strongly against though.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: They are separated to make finalize logic straightforward, differentiating between samples that turned idle from samples that disappeared. If we put it all together more logic will be needed to classify samples within the same map.

internal/component/database_observability/postgres/collector/query_samples.go
matthewnolf: I think there is a balance, a lot of the comments are saying what the code is doing. Their usefulness comes when they explain **why** it is doing it

gaantunes: Removed those that I judged redundant. Let me know if it looks better.

PR #4528, opened 2025-10-01
Update README with better user grants guidance

cristiangreco:
#### PR Description
Added notes on grants for the `db-o11y` to access specific schemas.

n.a.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/README.md
+++ b/internal/component/database_observability/README.md
@@ -406,9 +406,24 @@ show track_activity_query_size;
 CREATE USER "db-o11y" WITH PASSWORD '<password>';
 GRANT pg_monitor TO "db-o11y";
 GRANT pg_read_all_stats TO "db-o11y";
+GRANT pg_read_all_data TO "db-o11y"; /* see note */
 ```
 
-7. Verify that the user has been properly created.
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/README.md
```diff
@@ -406,9 +406,24 @@ show track_activity_query_size;
...
 ```
 
-7. Verify that the user has been properly created.
+Please note: Regarding `GRANT pg_read_all_data TO "db-o11y"`, it is possible to restrict permissions, if necessary. Instead, grant the `db-o11y` user privileges access only to the objects (databases and schemas) for which you want information. For example, to restrict permissions only to specific schemas in a database named `payments`:
+
+```sql
+CREATE USER "db-o11y" WITH PASSWORD '<password>';
+GRANT pg_monitor TO "db-o11y";
+
+-- connect to the 'payments' database and grant access to specific schemas
```
matthewnolf: I think we should prefer that users are deliberate with permissions. Would therefore suggest this first. As an alternative, if users do want **all access**, they can grant that with `pg_read_all_data`.

My point here is mainly about the preference/ordering of the suggestion. We should encourage more fine grained access

PR #4542, opened 2025-10-02
database_observability: Postgres explain plan improvements

rgeyer:
#### PR Description
Fast-follow to the initial implementation of postgres explain plans.

This switches to using the pg_stat_statements table for fetching queries for explain plans. Previously pg_stat_activity was used, and it was determined that this might not produce all queries which had been executed.

Using a different table necessitated a different strategy for determining which queries had been executed, since the last scrape interval, which is why the settings for "InitialLookback" was removed.

Now *all* queries will be explained in the first batch. Then only queries which have been executed since the last scrape interval going forward.

Finally, this change *also* uses a temporary prepared statement to execute the explain plan. This is necessary because the queries are parameterized. In a prepared statement, you may provide `null` parameters to the statement, and set the `plan_cache_mode` to `force_generic_plan` in order to get a usable explain plan without real parameter values.

#### Notes to the Reviewer
I did not change the CHANGELOG, since this is a fast-follow update to unreleased functionality.

#### PR Checklist
- [X] Documentation added

### DIFF
--- a/internal/component/database_observability/postgres/collector/explain_plan.go
+++ b/internal/component/database_observability/postgres/collector/explain_plan.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -27,18 +28,35 @@ const (
 	OP_EXPLAIN_PLAN_OUTPUT = "explain_plan_output"
 )
 
-const selectQueriesForExplainPlan = `
+const selectQueriesForExplainPlanPre17 = `
 	SELECT
 		d.datname,
-		s.query_id,
+		s.queryid,
 		s.query,
-		s.query_start
-	FROM pg_stat_activity s
-		JOIN pg_database d ON s.datid = d.oid AND NOT d.datistemplate AND d.datallowconn
-	WHERE s.query_id is not null and s.query_start is not null and s.query_start > $1
+		s.calls,
+		NOW() AT TIME ZONE 'UTC' AS stats_since
+	FROM pg_stat_statements s
+		JOIN pg_database d ON s.dbid = d.oid AND NOT d.datistemplate AND d.datallowconn
+	WHERE s.queryid IS NOT NULL AND s.query IS NOT NULL
 `
 
-const selectExplainPlanPrefix = `EXPLAIN (FORMAT JSON) `
+const selectQueriesCallResetTimePre17 = `
+	SELECT stats_reset FROM pg_stat_statements_info
+`
+
+const selectQueriesForExplainPlan17Plus = `
+	SELECT
+		d.datname,
+		s.queryid,
+		s.query,
+		s.calls,
+		s.stats_since
+	FROM pg_stat_statements s
+		JOIN pg_database d ON s.dbid = d.oid AND NOT d.datistemplate AND d.datallowconn
+	WHERE s.queryid IS NOT NULL AND s.query IS NOT NULL	
+`
+
+const selectExplainPlanPrefix = `EXPLAIN (FORMAT JSON) EXECUTE `
 
 var unrecoverablePostgresSQLErrors = []string{
 	"pq: permission denied for table",
@@ -195,14 +213,18 @@ type queryInfo struct {
 	queryText    string
 	failureCount int
 	uniqueKey    string
+	calls        int64
+	callsReset   time.Time
 }
 
-func newQueryInfo(datname, queryId, queryText string) *queryInfo {
+func newQueryInfo(datname, queryId, queryText string, calls int64, callsReset time.Time) *queryInfo {
 	return &queryInfo{
-		datname:   datname,
-		queryId:   queryId,
-		queryText: queryText,
-		uniqueKey: datname + queryId,
+		datname:    datname,
+		queryId:    queryId,
+		queryText:  queryText,
+		uniqueKey:  datname + queryId,
+		calls:      calls,
+		callsReset: callsReset,
 	}
 }
 
@@ -220,37 +242,37 @@ type ExplainPlanArguments struct {
 }
 
 type ExplainPlan struct {
-	dbConnection     *sql.DB
-	dbDSN            string
-	dbVersion        string
-	scrapeInterval   time.Duration
-	queryCache       map[string]*queryInfo
-	queryDenylist    map[string]*queryInfo
-	excludeSchemas   []string
-	perScrapeRatio   float64
-	currentBatchSize int
-	entryHandler     loki.EntryHandler
-	lastSeen         time.Time
-	logger           log.Logger
-	running          *atomic.Bool
-	ctx              context.Context
-	cancel           context.CancelFunc
+	dbConnection       *sql.DB
+	dbDSN              string
+	dbVersion          string
+	scrapeInterval     time.Duration
+	queryCache         map[string]*queryInfo
+	queryDenylist      map[string]*queryInfo
+	finishedQueryCache map[string]*queryInfo
+	excludeSchemas     []string
+	perScrapeRatio     float64
+	currentBatchSize   int
+	entryHandler       loki.EntryHandler
+	logger             log.Logger
+	running            *atomic.Bool
+	ctx                context.Context
+	cancel             context.CancelFunc
 }
 
 func NewExplainPlan(args ExplainPlanArguments) (*ExplainPlan, error) {
 	return &ExplainPlan{
-		dbConnection:   args.DB,
-		dbDSN:          args.DSN,
-		dbVersion:      args.DBVersion,
-		scrapeInterval: args.ScrapeInterval,
-		queryCache:     make(map[string]*queryInfo),
-		queryDenylist:  make(map[string]*queryInfo),
-		excludeSchemas: args.ExcludeSchemas,
-		perScrapeRatio: args.PerScrapeRatio,
-		entryHandler:   args.EntryHandler,
-		lastSeen:       args.InitialLookback,
-		logger:         log.With(args.Logger, "collector", ExplainPlanCollector),
-		running:        atomic.NewBool(false),
+		dbConnection:       args.DB,
+		dbDSN:              args.DSN,
+		dbVersion:          args.DBVersion,
+		scrapeInterval:     args.ScrapeInterval,
+		queryCache:         make(map[string]*queryInfo),
+		queryDenylist:      make(map[string]*queryInfo),
+		finishedQueryCache: make(map[string]*queryInfo),
+		excludeSchemas:     args.ExcludeSchemas,
+		perScrapeRatio:     args.PerScrapeRatio,
+		entryHandler:       args.EntryHandler,
+		logger:             log.With(args.Logger, "collector", ExplainPlanCollector),
+		running:            atomic.NewBool(false),
 	}, nil
 }
 
@@ -300,7 +322,29 @@ func (c *ExplainPlan) Stop() {
 }
 
 func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
-	rs, err := c.dbConnection.QueryContext(ctx, selectQueriesForExplainPlan, c.lastSeen)
+	var selectStatement string
+	var resetTS time.Time
+	if f, err := strconv.ParseFloat(c.dbVersion, 64); err == nil && f >= 17.0 {
+		selectStatement = selectQueriesForExplainPlan17Plus
+	} else {
+		stat_reset, err := c.dbConnection.QueryContext(ctx, selectQueriesCallResetTimePre17)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to fetch stats reset time for explain plans", "err", err)
+			return err
+		}
+		defer stat_reset.Close()
+		if stat_reset.Next() {
+			var ls time.Time
+			if err := stat_reset.Scan(&ls); err != nil {
+				level.Error(c.logger).Log("msg", "failed to scan stats reset time for explain plans", "err", err)
+				return err
+			}
+			resetTS = ls
+		}
+		selectStatement = selectQueriesForExplainPlanPre17
+	}
+
+	rs, err := c.dbConnection.QueryContext(ctx, selectStatement)
 	if err != nil {
 		level.Error(c.logger).Log("msg", "failed to fetch digests for explain plans", "err", err)
 		return err
@@ -309,8 +353,9 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 
 	for rs.Next() {
 		var datname, queryId, query string
+		var calls int64
 		var ls time.Time
-		if err := rs.Scan(&datname, &queryId, &query, &ls); err != nil {
+		if err := rs.Scan(&datname, &queryId, &query, &calls, &ls); err != nil {
 			level.Error(c.logger).Log("msg", "failed to scan query for explain plan", "err", err)
 			return err
 		}
@@ -322,12 +367,22 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 			continue
 		}
 
-		qi := newQueryInfo(datname, queryId, query)
-		if _, ok := c.queryDenylist[qi.uniqueKey]; !ok {
-			c.queryCache[qi.uniqueKey] = qi
+		statsReset := resetTS
+		if f, err := strconv.ParseFloat(c.dbVersion, 64); err == nil && f >= 17.0 {
+			statsReset = ls
 		}
-		if ls.After(c.lastSeen) {
-			c.lastSeen = ls
+		qi := newQueryInfo(datname, queryId, query, calls, statsReset)
+		if _, ok := c.queryDenylist[qi.uniqueKey]; !ok {
+			if previous, ok := c.finishedQueryCache[qi.uniqueKey]; ok {
+				if calls == previous.calls {
+					continue
+				}
+				if calls > previous.calls && (statsReset.Equal(previous.callsReset) || statsReset.Before(previous.callsReset)) {
+					continue
+				}
+				delete(c.finishedQueryCache, qi.uniqueKey)
+			}
+			c.queryCache[qi.uniqueKey] = qi
 		}
 	}
 
@@ -356,6 +411,8 @@ func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
 				qi.failureCount++
 				c.queryDenylist[qi.uniqueKey] = qi
 				level.Info(c.logger).Log("msg", "query denylisted", "query_id", qi.queryId)
+			} else {
+				c.finishedQueryCache[qi.uniqueKey] = qi
 			}
 			delete(c.queryCache, qi.uniqueKey)
 			processedCount++
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
 	}
 	defer conn.Close()
 
+	prepared_statement_name := strings.ReplaceAll(fmt.Sprintf("explain_plan_%s", qi.queryId), "-", "_")
+	if _, err := conn.ExecContext(ctx, fmt.Sprintf("PREPARE %s AS %s", prepared_statement_name, qi.queryText)); err != nil {
+		return nil, fmt.Errorf("failed to prepare explain plan: %w", err)
+	}
+
+	defer func() {
+		if _, err := conn.ExecContext(ctx, fmt.Sprintf("DEALLOCATE %s", prepared_statement_name)); err != nil {
+			level.Error(c.logger).Log("msg", "failed to deallocate explain plan", "err", err)
+		}
+	}()
+
 	setSearchPathStatement := fmt.Sprintf("SET search_path TO %s, public", qi.datname)
 	if _, err := conn.ExecContext(ctx, setSearchPathStatement); err != nil {
 		return nil, fmt.Errorf("failed to set search path: %w", err)
 	}
 
-	rsExplain := conn.QueryRowContext(ctx, selectExplainPlanPrefix+qi.queryText)
+	if _, err := conn.ExecContext(ctx, "SET plan_cache_mode = force_generic_plan"); err != nil {
+		return nil, fmt.Errorf("failed to set plan cache mode: %w", err)
+	}
+
+	paramCountRegex := regexp.MustCompile(`\$\d+`)
+	paramCount := len(paramCountRegex.FindAllString(qi.queryText, -1))
+
+	explainQuery := selectExplainPlanPrefix
+	explainQuery += prepared_statement_name
+	explainQuery += "("
+	for range paramCount {
+		explainQuery += "null,"
+	}
+	if paramCount > 0 {
+		explainQuery = explainQuery[:len(explainQuery)-1]
+	}
+	explainQuery += ")"
+
+	level.Debug(c.logger).Log("msg", "running explain plan", "query", explainQuery)
+
+	rsExplain := conn.QueryRowContext(ctx, explainQuery)
 	if err := rsExplain.Err(); err != nil {
 		return nil, fmt.Errorf("failed to run explain plan: %w", err)
 	}
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -81,7 +81,6 @@ This collector has no config options.
 |--------------------------------|----------------|------------------------------------------------------|---------|----------|
 | `collect_interval`             | `duration`     | How frequently to collect information from database. | `"1m"`  | no       |
 | `per_collect_ratio`            | `float64`      | The ratio of queries to collect explain plans for.   | `1.0`   | no       |
-| `initial_lookback`             | `duration`     | The amount of time to look back for explain plans.   | `24h`   | no       |
 | `explain_plan_exclude_schemas` | `list(string)` | Schemas to exclude from explain plans.               | `[]`    | no       |
 
 ## Example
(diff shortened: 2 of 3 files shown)


fridgepoet:
Would you mind adding more tests for this collector and also some tests highlighting/documenting this change?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -27,18 +28,35 @@ const (
...
-	WHERE s.query_id is not null and s.query_start is not null and s.query_start > $1
+		s.calls,
+		NOW() AT TIME ZONE 'UTC' AS stats_since
+	FROM pg_stat_statements s
+		JOIN pg_database d ON s.dbid = d.oid AND NOT d.datistemplate AND d.datallowconn
+	WHERE s.queryid IS NOT NULL AND s.query IS NOT NULL
 `
 
-const selectExplainPlanPrefix = `EXPLAIN (FORMAT JSON) `
+const selectQueriesCallResetTimePre17 = `
```
matthewnolf: Is the difference between `selectQueriesForExplainPlanPre17` and `selectQueriesCallResetTimePre17` the 
`s.stats_since` column?

Perhaps we can follow the pattern we use elsewhere of templating the query and adding columns with `fmt.Sprintf` dependant on version
https://github.com/grafana/alloy/blob/main/internal/component/database_observability/mysql/collector/query_samples.go#L61

matthewnolf:
> Now all queries will be explained in the first batch. Then only queries which have been executed since the last scrape interval going forward.

What is the point in re-explaining a query if it was since executed? Is there a possibility the explain plan would change over time?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -27,18 +28,35 @@ const (
...
+		s.queryid,
+		s.query,
+		s.calls,
+		s.stats_since
+	FROM pg_stat_statements s
+		JOIN pg_database d ON s.dbid = d.oid AND NOT d.datistemplate AND d.datallowconn
+	WHERE s.queryid IS NOT NULL AND s.query IS NOT NULL	
+`
+
+const selectExplainPlanPrefix = `EXPLAIN (FORMAT JSON) EXECUTE `
```
matthewnolf, line 43: nit: Given this is only used once, can we move it to it's consumer for readability

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
+	explainQuery := selectExplainPlanPrefix
+	explainQuery += prepared_statement_name
+	explainQuery += "("
+	for range paramCount {
+		explainQuery += "null,"
+	}
+	if paramCount > 0 {
+		explainQuery = explainQuery[:len(explainQuery)-1]
+	}
+	explainQuery += ")"
```
matthewnolf: I think we can probably simplify this string building with something like:
`fmt.Sprintf("%s%s(%s)", selectExplainPlanPrefix, prepared_statement_name, slices.Repeat([]string{"null,"}, len(paramCount)))`

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
 	}
 	defer conn.Close()
 
+	prepared_statement_name := strings.ReplaceAll(fmt.Sprintf("explain_plan_%s", qi.queryId), "-", "_")
```
matthewnolf: nit: mixing camel and snake case

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -300,7 +322,29 @@ func (c *ExplainPlan) Stop() {
...
+	} else {
+		stat_reset, err := c.dbConnection.QueryContext(ctx, selectQueriesCallResetTimePre17)
+		if err != nil {
+			level.Error(c.logger).Log("msg", "failed to fetch stats reset time for explain plans", "err", err)
+			return err
+		}
+		defer stat_reset.Close()
+		if stat_reset.Next() {
+			var ls time.Time
+			if err := stat_reset.Scan(&ls); err != nil {
```
matthewnolf: Why not just scan into `resetTS` straight away?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
 	defer conn.Close()
 
+	prepared_statement_name := strings.ReplaceAll(fmt.Sprintf("explain_plan_%s", qi.queryId), "-", "_")
+	if _, err := conn.ExecContext(ctx, fmt.Sprintf("PREPARE %s AS %s", prepared_statement_name, qi.queryText)); err != nil {
+		return nil, fmt.Errorf("failed to prepare explain plan: %w", err)
+	}
+
+	defer func() {
+		if _, err := conn.ExecContext(ctx, fmt.Sprintf("DEALLOCATE %s", prepared_statement_name)); err != nil {
+			level.Error(c.logger).Log("msg", "failed to deallocate explain plan", "err", err)
```
matthewnolf: We should log which prepared statement failed to deallocate

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
+		}
+	}()
+
 	setSearchPathStatement := fmt.Sprintf("SET search_path TO %s, public", qi.datname)
 	if _, err := conn.ExecContext(ctx, setSearchPathStatement); err != nil {
 		return nil, fmt.Errorf("failed to set search path: %w", err)
 	}
 
-	rsExplain := conn.QueryRowContext(ctx, selectExplainPlanPrefix+qi.queryText)
+	if _, err := conn.ExecContext(ctx, "SET plan_cache_mode = force_generic_plan"); err != nil {
```
matthewnolf, line 536: Do we need to undo these values after explaining?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -466,12 +523,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
+	explainQuery += "("
+	for range paramCount {
+		explainQuery += "null,"
+	}
+	if paramCount > 0 {
+		explainQuery = explainQuery[:len(explainQuery)-1]
+	}
+	explainQuery += ")"
+
+	level.Debug(c.logger).Log("msg", "running explain plan", "query", explainQuery)
```
matthewnolf: Are we at risk of logging unredacted queries?

rgeyer:
> > Now all queries will be explained in the first batch. Then only queries which have been executed since the last scrape interval going forward.
> 
> What is the point in re-explaining a query if it was since executed? Is there a possibility the explain plan would change over time?

Mostly it's just a ham-fisted way to make sure that we have a "current" explain for a given query. If we only ran an explain once, we'd only have one log entry, and the frontend would have to query a very large timerange to find it.

By re-running the explain every time the query is executed, we ensure we have "fresh" data in the backend. There is also a very small chance that the explain could change, but that's a secondary concern.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Yes, that's the only difference. Great feedback, I'll implement it, thx!

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Negative. The original queries are parameterized, and the `explainQuery` here is the prepared statement with null params.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: No. For postgres we have to (re)connect to the DB specifying the schema we're trying to run the explain query in.

Each explain query is using a bespoke connection, including the schema/db name, then closing the new connection.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Yeah, the issue is the trailing comma in the null params. I did simplify this a bit tho, thanks for pointing it out.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Also... I think that was left in accidentally. There's no real need to log that data. I've removed it. :)

matthewnolf:
>Mostly it's just a ham-fisted way to make sure that we have a "current" explain for a given query. If we only ran an explain once, we'd only have one log entry, and the frontend would have to query a very large timerange to find it.

By re-running the explain every time the query is executed, we ensure we have "fresh" data in the backend. There is also a very small chance that the explain could change, but that's a secondary concern.

Thanks for explaining. For queries that aren't executed in a while, the query to find the explain output would have to extend to a large timerange, or are we periodically re-running all queries?

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,414 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+	// Invalid JSON input
+	explainJSON := []byte(`invalid json`)
+
+	output, err := newExplainPlanOutput(dbVersion, queryId, explainJSON, generatedAt)
+
+	require.Error(t, err)
+	assert.Nil(t, output)
+}
+
+func TestExplainPlan_PopulateQueryCache_Version17Plus(t *testing.T) {
```
matthewnolf: nit: Can we group some of these tests together that are testing the same functionality? E.g. `TestExplainPlan_PopulateQueryCache_Version17Plus` + `TestExplainPlan_PopulateQueryCache_VersionPre17`

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,414 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+		AddRow("testdb2", "345678", "SELECT * FROM orders", int64(20), time.Now())
+
+	mock.ExpectQuery("SELECT.*FROM pg_stat_statements s.*").WillReturnRows(rows)
+
+	ctx := context.Background()
+	err = explainPlan.populateQueryCache(ctx)
+
+	require.NoError(t, err)
+	assert.Len(t, explainPlan.queryCache, 2)         // Should exclude information_schema
+	assert.Equal(t, 1, explainPlan.currentBatchSize) // 50% of 2 queries
```
matthewnolf: nit: Some of these comments seem a bit unnecessary

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,414 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+
+	mock.ExpectQuery("SELECT.*FROM pg_stat_statements s.*").WillReturnRows(rows)
+
+	ctx := context.Background()
+	err = explainPlan.fetchExplainPlans(ctx)
+
+	// Should succeed but not process any queries since they require actual DB connections
+	require.NoError(t, err)
+
+	assert.NoError(t, mock.ExpectationsWereMet())
```
matthewnolf: We should validate the cache is correctly populated

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,414 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+	// Mock the populateQueryCache call
+	rows := sqlmock.NewRows([]string{"datname", "queryid", "query", "calls", "stats_since"}).
+		AddRow("testdb", "123456", "SELECT * FROM users WHERE id = $1", int64(10), time.Now())
+
+	mock.ExpectQuery("SELECT.*FROM pg_stat_statements s.*").WillReturnRows(rows)
+
+	ctx := context.Background()
+	err = explainPlan.fetchExplainPlans(ctx)
+
+	// Should succeed but not process any queries since they require actual DB connections
```
matthewnolf: I'm guessing this is because of the new sql connection for the explain. 
It's unfortunate there is no way to test that, as it is quite critical to the flow. Did you investigate any alternatives?

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,414 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+
+	// Mock the stats reset query for pre-17
+	resetRows := sqlmock.NewRows([]string{"stats_reset"}).AddRow(resetTime)
+	mock.ExpectQuery("SELECT stats_reset FROM pg_stat_statements_info").WillReturnRows(resetRows)
+
+	// Mock the main query for pre-17
+	rows := sqlmock.NewRows([]string{"datname", "queryid", "query", "calls", "stats_since"}).
+		AddRow("testdb", "123456", "SELECT * FROM users WHERE id = $1", int64(10), time.Now())
+
+	mock.ExpectQuery("SELECT.*FROM pg_stat_statements s.*").WillReturnRows(rows)
```
matthewnolf: Why are we not using the full query here? Which has logic in it for the difference between pre-17 and post right?

rgeyer: I changed this in most places, but for some reason, on the "pre 17" query, it chokes on matching the query in the mock, so this broader and more encapsulating query regex is used.

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: Fair.. Sorta just stream-of-consciousness.. Happens sometimes. I cleaned it up.

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: This will populate the cache and "process" the contents of the cache in one go. I can (and will) verify that the queries went into the "finished" cache tho.

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: Correct. Because the function creates a new DB connection we don't have access to mock it. I *could* change the function signature to allow me to pass in an (optional) db connection, but that would really only be used during the tests.

The only other option is to try ty reuse the db connection with the supplied DSN, but that simply doesn't work in postgres. You can't just `USE schema_name` like you can in MySQL, you have to connect to the schema you wish to operate on.

rgeyer:
> > Mostly it's just a ham-fisted way to make sure that we have a "current" explain for a given query. If we only ran an explain once, we'd only have one log entry, and the frontend would have to query a very large timerange to find it.
> 
> By re-running the explain every time the query is executed, we ensure we have "fresh" data in the backend. There is also a very small chance that the explain could change, but that's a secondary concern.
> 
> Thanks for explaining. For queries that aren't executed in a while, the query to find the explain output would have to extend to a large timerange, or are we periodically re-running all queries?

Yeah, good question. We don't really have a strategy in place for periodically grabbing "stale" queries. I would imagine that we'd have this problem/consideration for query samples as well?

At the moment, I think it's "okay" if mostly inactive queries do eventually lose out on having explain plans.

internal/component/database_observability/postgres/collector/explain_plan_test.go
fridgepoet: (not saying we have to do this in this PR)
Isn't there a way we could spin up this scenario in docker? make this a CI test?

internal/component/database_observability/postgres/collector/explain_plan_test.go
fridgepoet: (just my 2 cents feel free to ignore: gosh, that's a shame to have this "just because" of the test choking... maybe we can help look into this more to understand what's going on with the mock....?)

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: Yes, we could create an integration test for this. To some degree that's covered by the dogfooding, and k6 tests we're already running. We could setup a specific k6 suite that exercises this, along with the requisite backend.

Agree tho that would be for a follow-up PR. Hoping to get this across the line for public preview.

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: I also agree here. I'd like for this test to be "clean". That said, I spent about 60min troubleshooting this to no avail. Unsure how much more time we want to spend on figuring out what's happening inside of the mock library 🤔

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,409 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+
+		assert.Contains(t, explainPlan.queryCache, "testdb123456")
+		assert.Contains(t, explainPlan.queryCache, "testdb2345678")
+		assert.NotContains(t, explainPlan.queryCache, "information_schema789012")
+
+		assert.NoError(t, mock.ExpectationsWereMet())
+	})
+
+	t.Run("populate query cache for PostgreSQL < 17", func(t *testing.T) {
+		db, mock, err := sqlmock.New()
```
fridgepoet:
```suggestion
		db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
```

internal/component/database_observability/postgres/collector/explain_plan_test.go
```diff
@@ -2274,3 +2280,409 @@ func TestReplaceDatabaseNameInDSN(t *testing.T) {
...
+		resetRows := sqlmock.NewRows([]string{"stats_reset"}).AddRow(resetTime)
+		mock.ExpectQuery("SELECT stats_reset FROM pg_stat_statements_info").WillReturnRows(resetRows)
+
+		rows := sqlmock.NewRows([]string{"datname", "queryid", "query", "calls", "stats_since"}).
+			AddRow("testdb", "123456", "SELECT * FROM users WHERE id = $1", int64(10), time.Now())
+
+		mock.ExpectQuery("SELECT.*FROM pg_stat_statements.*").WillReturnRows(rows)
+		// For some reason, this fails on the `.` character in the where clause "WHERE s.queryid...".
+		// So we're using a broader regex to match the query on the mock.
+		// mock.ExpectQuery(fmt.Sprintf(selectQueriesForExplainPlanTemplate, "NOW() AT TIME ZONE 'UTC' AS stats_since")).WillReturnRows(rows)
```
fridgepoet:
```suggestion
		mock.ExpectQuery(fmt.Sprintf(selectQueriesForExplainPlanTemplate, "NOW() AT TIME ZONE 'UTC' AS stats_since")).WillReturnRows(rows)
```

internal/component/database_observability/postgres/collector/explain_plan_test.go
fridgepoet: apparently it matching as regex by default but what we can do here is tell sqlmock to do exact string matching instead.

(not sure if this resolves all of the concern)

internal/component/database_observability/postgres/collector/explain_plan_test.go
fridgepoet: Not sure if it is correct, but I added some suggestions.

internal/component/database_observability/postgres/collector/explain_plan_test.go
matthewnolf: > To some degree that's covered by the dogfooding, and k6 tests we're already running

I wouldn't trust that to verify things work or break unexpectedly 😅

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: You're 100% right, that was an oversight on my part 🤦‍♂️

internal/component/database_observability/postgres/collector/explain_plan_test.go
rgeyer: Okay, added some functionality, and a test which does the full "happy path" of executing an explain. Feeling much more confident in it now.

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -195,17 +197,32 @@ type queryInfo struct {
...
+		datname:    datname,
+		queryId:    queryId,
+		queryText:  queryText,
+		uniqueKey:  datname + queryId,
+		calls:      calls,
+		callsReset: callsReset,
 	}
 }
 
+type DatabaseConnectionFactory interface {
```
matthewnolf: Does this make more sense to be a function? 
```
type DatabaseConnectionFactory func(dsn string) (*sql.DB, error)
```

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -220,37 +237,39 @@ type ExplainPlanArguments struct {
...
-		excludeSchemas: args.ExcludeSchemas,
-		perScrapeRatio: args.PerScrapeRatio,
-		entryHandler:   args.EntryHandler,
-		lastSeen:       args.InitialLookback,
-		logger:         log.With(args.Logger, "collector", ExplainPlanCollector),
-		running:        atomic.NewBool(false),
+		dbConnection:        args.DB,
+		dbDSN:               args.DSN,
+		dbVersion:           args.DBVersion,
+		dbConnectionFactory: &databaseConnectionFactory{},
```
matthewnolf: Following my last comment - this can then just be `sql.Open`

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -460,18 +512,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
+	logger := log.With(c.logger, "query_id", qi.queryId, "datname", qi.datname, "preparedStatementName", preparedStatementName)
+	if _, err := conn.ExecContext(ctx, fmt.Sprintf("PREPARE %s AS %s", preparedStatementName, qi.queryText)); err != nil {
+		return nil, fmt.Errorf("failed to prepare explain plan: %w", err)
+	}
+
+	defer func() {
+		if _, err := conn.ExecContext(ctx, fmt.Sprintf("DEALLOCATE %s", preparedStatementName)); err != nil {
+			level.Error(logger).Log("msg", "failed to deallocate explain plan", "err", err)
+		}
+	}()
```
cristiangreco, line 529: Question related to `SET search_path` below: do we need to switch back to `public` schema only after the explain? I think for postgres is not a big deal, but just wondering if you have noticed any side effect of the changed `search_path`?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -322,13 +362,23 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 			continue
 		}
 
-		qi := newQueryInfo(datname, queryId, query)
+		statsReset := resetTS
+		if f, err := strconv.ParseFloat(c.dbVersion, 64); err == nil && f >= 17.0 {
+			statsReset = ls
+		}
```
cristiangreco, line 367: Could we parse the version once instead that in a loop?

Btw nitpick maybe for a followup, in the mysql component we use the semver module which makes it quite handy to check versions.

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -460,18 +512,43 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
...
 	if _, err := conn.ExecContext(ctx, setSearchPathStatement); err != nil {
 		return nil, fmt.Errorf("failed to set search path: %w", err)
 	}
 
-	rsExplain := conn.QueryRowContext(ctx, selectExplainPlanPrefix+qi.queryText)
+	if _, err := conn.ExecContext(ctx, "SET plan_cache_mode = force_generic_plan"); err != nil {
+		return nil, fmt.Errorf("failed to set plan cache mode: %w", err)
+	}
+
+	paramCountRegex := regexp.MustCompile(`\$\d+`)
```
cristiangreco: This can be a var at the top of the file, to compile it only once for performance optimisation.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: For postgres, we must create a new connection which specifies the schema/db. As such setting the search_path is probably redundant, but I'm doing it "for safety". @gaantunes may have some insight here.

That said, the DB connection is closed at the end of each explain. Even if it were to get reused (not sure how the underlying sql library manages the connection pool) I would expect it would be connecting again to the same schema, and as such the search_path would still be appropriate.

I haven't seen any negative impacts from this, but my testing has been pretty low-volume locally.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Yeah, good call.. I'm fine to make both changes. This got implemented this way because it was necessary to do some regex shenanigans to parse apart the complex postgres version string. "Muscle memory" dictated that I just convert the actual version number to a float and carry on. Using semver is the right approach tho, so I'll adjust.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Yeah, kinda 6 of one, half dozen of another.

Since it's only a single function, I suppose your proposal is the better approach. Adjusted accordingly in [this commit](https://github.com/grafana/alloy/pull/4542/commits/85d5c7bbffc6af91949c0e8339b0fbd85521f9eb).

cristiangreco (approved):

PR #4543, opened 2025-10-02
Update Beyla component to 2.7.2

grcevski:
#### PR Description

This PR updates the Beyla component to 2.7.2 which includes number of improvements, especially around the stability of the produced target_info and survey_info metrics.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated


CHANGELOG.md
```diff
@@ -7,15 +7,16 @@ This document contains a historical list of changes between releases. Only
 changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
-Main (unreleased)
------------------
+## Main (unreleased)
```
kalleep: Why did all headers for each release change?

grcevski: Good catch, no idea, probably some automatic formatter in my editor. I'll revert them.

grcevski: Fixed

CHANGELOG.md
kgeckhart, line 1: Looks like this file might have gone through a bad merge

grcevski: Ah man, let me fix that, I didn't check after I merged main.

grcevski: Fixed, thanks for catching it

kalleep:
To update go version we need to do that in two steps, first a pr to update build image and then later all alloy related part.

I have already done that so no need to have those changes in this pr:
https://github.com/grafana/alloy/pull/4546
https://github.com/grafana/alloy/pull/4637

grcevski:
> To update go version we need to do that in two steps, first a pr to update build image and then later all alloy related part.
> 
> I have already done that so no need to have those changes in this pr: #4546 #4637

Thanks! I've updated the PR now and it only has the minimal changes to the go mod, the semconv update for traces and the minor code updates that makes build correctly.

go.mod
```diff
@@ -223,88 +223,88 @@ require (
...
+	go.opentelemetry.io/collector/component/componenttest v0.136.0
+	go.opentelemetry.io/collector/config/configauth v0.136.0
+	go.opentelemetry.io/collector/config/configcompression v1.42.0
+	go.opentelemetry.io/collector/config/configgrpc v0.136.0
+	go.opentelemetry.io/collector/config/confighttp v0.136.0
+	go.opentelemetry.io/collector/config/confignet v1.42.0
+	go.opentelemetry.io/collector/config/configopaque v1.42.0
+	go.opentelemetry.io/collector/config/configoptional v0.136.0
+	go.opentelemetry.io/collector/config/configretry v1.42.0
+	go.opentelemetry.io/collector/config/configtelemetry v0.136.0
```
ptodev, line 238: Is it necessary to update these packages? Ideally we shouldn't update them outside of the routine OTel upgrades, because they could include changes to the config structs that also need to be documented.

Perhaps you can reset the entire go mod file back to match the one in `main`, and then `go mod tidy` to see if this is necessary?

grcevski: I think it was necessary, the Beyla version has dependency on them, so unless I drop it there I think it was required. I can take a look to see if it will allow me to revert those only.

grcevski: I tried but it reverts it because of the Beyla dependency. I think OTel has moved to 0.138 already.

If there were config issues with fields, wouldn't we expect that we'll get compilation errors because of missing fields? In my experience, they don't change the behaviour of the existing options, they make new ones and remove the old ones, so we should've had a compile problem.

ptodev: There are a couple of reasons why it's best not to update those:
* There could be breaking changes to already existing config arguments which do not lead to compilation issues.
* When main OTel upgrade is being done, it'd be more clear to the person doing it what has been documented. E.g. they'd know that from this version to that version there could be undocumented arguments that they need to add in.

It's not the end of the world if such dependencies are upgraded outside of the OTel release cycle, but it'd be worth it to eyeball the OTel changelog for breaking changes that could impact them.

ptodev: I had a look at the [v0.135](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.135.0) and [v0.136](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.136.0) releases and I don't see major changes. Potentially the only thing we need to change on our side is to document two new metrics in places like [this one](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.exporter.otlp/#debug-metrics). It's not critical to update this now since I don't think the list of metrics in our docs is very up to date anyway... we will need automation to keep this in sync.

PR #4546, opened 2025-10-03
Update build image to use golang v1.25.1

kalleep:
To prepare to use go 1.25.1 for alloy

### DIFF
--- a/.github/workflows/check-linux-build-image.yml
+++ b/.github/workflows/check-linux-build-image.yml
@@ -14,8 +14,8 @@ jobs:
     strategy:
       matrix:
         build:
-          - runtime: golang:1.24.6-alpine3.22
-          - runtime: mcr.microsoft.com/oss/go/microsoft/golang:1.24.6-bookworm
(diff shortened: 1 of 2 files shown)


kalleep:
> Should we do the windows build image as well?
> 
> over at:
> 
> `tools/build-image/windows/Dockerfile`
> 
> Looks like it's lagging behind at the moment at 1.24.0
> 
> Edit: Actually I'm not sure I understand how (if at all) windows build images are used. I see that dockerfile is used for testing a windows build image, but I wasn't finding any publication of a windows build image. 🤷‍♂️

Well windows build image is currently not used at all, but we should bring it back. At the moment everything from that images is included in Dockerfile.windows https://github.com/grafana/alloy/blob/main/Dockerfile.windows#L10-L19. This happened when we migrated to GHA but we should bring it back. Building windows release images takes 1 hour in our ci atm...

jharvey10:
👍 Okay, I created issue #4584 for that and I'll mark this one as approved in the meantime.

Issue #4262, opened 2025-08-26
Feature Request: Add file-based LogsReceiver with rotation and compression support


korniltsev:
## Summary

Currently, the `logging` block's `write_to` parameter only supports Loki-based LogsReceiver components (like `loki.write`, `loki.relabel`, etc.). There's no built-in way to write Alloy's internal logs directly to files with log rotation and compression capabilities.

## Problem

Users who want to:
1. Write Alloy's internal logs to local files (not just stderr)
2. Implement log rotation to manage disk usage
3. Compress rotated logs to save space
4. Avoid the complexity of setting up a full Loki instance just for internal logging

Currently have limited options and must rely on external tools or complex setups.

## Proposed Solution

Add a new component (e.g., `local.file_write`) that implements the LogsReceiver interface and provides:

- File-based log output with configurable path
- Built-in log rotation based on size, age, or both
- Compression of rotated files
- Configurable retention policies

This could leverage existing libraries like [lumberjack](https://github.com/natefinch/lumberjack) which provides these features out of the box.

## Example Usage

```alloy
local.file_write "internal_logs" {
  filename     = "/var/log/alloy/alloy.log"
  max_size     = 100  // MB
  max_backups  = 3
  max_age      = 28   // days
  compress     = true
}

logging {
  level    = "info"
  format   = "json"
  write_to = [local.file_write.internal_logs.receiver]
}
```

## Benefits

- Simplified log management for users who prefer file-based logging
- Built-in rotation and compression without external dependencies
- Consistent with Alloy's component-based architecture
- Reduces operational complexity for users not using Loki

## Alternative Workarounds

Currently, users must:
- Use system-level log rotation tools with stderr redirection
- Set up a local Loki instance just for internal logs
- Implement custom log forwarding solutions

A built-in file-based LogsReceiver would eliminate these complexities.

kalleep:
I like the idea but the component should be named `loki.file_writer`. This would be limited to loki pipelines.  Will bring it up during our next proposal review meeting

korniltsev:
I am not very familiar with otel , but worth double checking if they have anything like this, to not reinvent the wheel

kalleep:
For otel this would be [fileexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/fileexporter).

[...]

PR #4551, opened 2025-10-04
Add json field ops

achrefbensaad:
#### PR Description

Introduced a new `stage.json_field` component that enables modifying JSON-formatted
log lines by updating or deleting specific fields.

- Supports `update` and `delete` operations
- Added `drop_malformed` flag to skip invalid JSON logs
- `update` creates the field if it does not exist
- Values for `source` are resolved exclusively from the `.Extracted` map
- Added documentation and usage examples, including a rename pattern using
  `stage.json` + `stage.json_field`

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


achrefbensaad:
@clayton-cornell I am not sure if I should update converters, can you please advise.

kalleep:
Hey, thanks for making a pr but what I would really like to see before we consider this is a proposal and describe what you want too achieve and how it can be done, an example can be found here https://github.com/grafana/alloy/issues/4262.

This gives us (Alloy maintainers) a chance to discuss the proposal both internally and externally. Maybe there are other ways to already solve what you are trying to achieve or alternative solutions that could work better

clayton-cornell:
> I am not sure if I should update converters, can you please advise.

I'll wait on the results of the proposal/discussion that Karl asked for before diving into docs.

Issue #4490, opened 2025-09-24
Add retries for podmonitors/servicemonitors/etc


dehaansa:
### Component(s)

prometheus.operator.podmonitors

### Request

When Alloy starts, if the k8s network is not yet available it may fail to find ServiceMonitors/PodMonitors/etc and does not appropriately retry.

### Use case

It should retry!

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

dehaansa:
Similar recent fix: https://github.com/grafana/alloy/pull/3688

james-callahan:
> Similar recent fix: [#3688](https://github.com/grafana/alloy/pull/3688)

The fix in there only retries 3 times.

I would expect it to retry forever

or alternatively: fail entirely (and exit; possibly via the readiness endpoint returning an error?)

PR #4568, opened 2025-10-07
Use sync timeout to decide how long to retry waiting for k8s api for prometheus.operator.* components

dehaansa:
#### PR Description
Addressing https://github.com/grafana/alloy/issues/4490 by enhancing the existing retries behavior to honor the configured timeout, to allow users to increase the timeout if appropriate for their environment.


internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -364,18 +364,21 @@ func (c *crdManager) configureInformers(ctx context.Context, informers cache.Inf
...
 			MaxBackoff: 10 * time.Second,
-			MaxRetries: 3, // retry up to 3 times
+			MaxRetries: 0, // Will retry until InformerSyncTimeout is reached
 		},
 	)
 	for backoff.Ongoing() {
 		// Retry to get the informer in case of a timeout.
 		informer, err = getInformer(ctx, informers, prototype, c.args.InformerSyncTimeout)
-		if err == nil {
+		if err == nil || time.Now().After(done) {
```
kalleep: Not sure we need to do this check here. We create a context that will cancel backoff after some time. When that happens `backoff.Ongoing()` will be false.

Then I am pretty sure we need to check `backoff.Errs()`. In this setup this function will only return an error if it was canceled.

dehaansa: The intent here was to not log that we were going to retry if we had already exceeded the timeout before the next iteration checks ongoing, but I suppose we can just check Errs instead 👍

dehaansa: Updated the logic (a few times) to be clearer & accurate, let me know if you see other opportunities for improvement.

Issue #4486, opened 2025-09-24
Loki Write Component should support configuring the max line size to truncate very large entries


spenczar:
### Component(s)

loki.write

### Request

Permit configuration of the "MaxLineSize" and "MaxLineSizeTruncate" options used by the Loki client so that excessively long log entries get truncated.

That is, I'd like to set the `github.com/grafana/alloy/internal/component/common/loki/limit.Config.MaxLineSize` and `github.com/grafana/alloy/internal/component/common/loki/limit.Config.MaxLineSizeTruncate` configuration options and pass them in to the Loki client in `github.com/grafana/alloy/internal/component/loki/write`.

### Use case

When Alloy is sending log entries to Loki, it is very important that none of the logs pass a critical size threshold. Loki will [reject a write](https://grafana.com/docs/loki/latest/operations/request-validation-rate-limits/#line_too_long) if it contains an entry which is too long.

Users who are operating Loki themselves might be able to configure Loki to handle this, but if you're using Grafana Cloud, that is not so simple.

So, it's important to be able to truncate log entries when writing to Loki, or else you risk failing entire batches of writes.

The Loki Write component already supports this feature. It will truncate the log entry, if configure appropriately; it just needs a `maxLineSize` to be set: https://github.com/grafana/alloy/blob/main/internal/component/common/loki/client/client.go#L254-L263.

This feature is tested and seems to work... but it can't actually be enabled currently, because the only constructor for the client does not set `limit.Config.MaxLineSize`, and does not allow it to be set from any configuration from the user.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

spenczar:
Related: #3039

kalleep:
Thanks, yes these cannot currently be configured and I don't t hink we should add them to `loki.write` either. We should clean up the code there.

From the [limit settings](https://github.com/grafana/alloy/blob/627997b8a9c3e9b37358bf93840e6e0d11a09d15/internal/component/common/loki/limit/config.go#L7-L15) everything is already covered except for `MaxLineSizeTruncate`.

The limits can be set by using [stage.limit](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagelimit) and `MaxLineSize` is covered by [stage.drop ](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagedrop). That just leaves us with the last setting. For that one we could add `stage.truncate` if its not already covered by something else

PR #4576, opened 2025-10-09
Remove some loki/v3 dependencies

ptodev:
It would be good to remove the `loki/v3` dependency, because it complicates Alloy's dependency tree and makes OTel and Prometheus upgrades harder. This would have to be done over multiple PRs so that it's easier to review.

This initial PR contains 3 changes:

1. Using `loki/pkg/push` instead of `loki/v3/pkg/logproto`
2. ~~Vendoring the `flagext` package from the loki repo~~
3. A few tests use a noop logger directly rather than importing it from `loki/v3`


internal/component/common/loki/limit/config.go
```diff
@@ -1,7 +1,7 @@
 package limit
 
 import (
-	"github.com/grafana/loki/v3/pkg/util/flagext"
+	"github.com/grafana/alloy/internal/loki/util/flagext"
```
kalleep: There are a lot we can clean-up here. MaxLineSize (uses flagext.ByteSize) is not configurable in alloy. There are other stuff but we could remove that one and then we don't have to vendor flagext

kalleep: Then only thing in limit config we can can configure is MaxStreams, everything else is dead code.

See https://github.com/grafana/alloy/issues/4486#issuecomment-3327092968 for more info

ptodev: Thank you, I didn't realise! I reverted the `flagext` change. I'll take a look in a subsequent PR.

PR #4578, opened 2025-10-09
Refactor: Reuse The Same Label Builder Instance in Relabel Component

blewis12:
#### PR Description
Implements a small performance improvement wrt memory allocations, mentioned [here](https://github.com/grafana/alloy/pull/4455#discussion_r2359227383). There didn't seem to be a major regression with the original change (some benchmarking results [here](https://github.com/grafana/alloy/issues/4323#issuecomment-3371105410)) - but since this is a hot path it's definitely worth ensuring it's as performant as possible.

I did some more benchmarking here and there is a slight improvement using this approach wrt memory allocations 

Screenshot 2025-10-09 at 11 59 32

-

#### Notes to the Reviewer
My understanding of concurrency in the alloy context is still a work in progress, but from what I can see this `process()` method (which is called in the relabel method, called [here](https://github.com/grafana/alloy/blob/d9a06863ea7a66b95080604800c0de4daa97e2ac/internal/component/loki/relabel/relabel.go#L126) in Run) should be thread safe - I am running on the assumption that only one loki.relabel component can run at a time, and on reload components are shut down before new ones are loaded (which seems to be dealt with [here](https://github.com/grafana/alloy/blob/d9a06863ea7a66b95080604800c0de4daa97e2ac/internal/runtime/internal/controller/scheduler.go#L53)?), so I did not implement concurrency control here. Please let me know if this assumption is wrong!

### DIFF
--- a/internal/component/loki/relabel/relabel.go
+++ b/internal/component/loki/relabel/relabel.go
@@ -74,6 +74,8 @@ type Component struct {
 	maxCacheSize int
 
 	debugDataPublisher livedebugging.DebugDataPublisher
+
+	builder labels.ScratchBuilder
 }
 
 var (
@@ -99,6 +101,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		cache:              cache,
 		maxCacheSize:       args.MaxCacheSize,
 		debugDataPublisher: debugDataPublisher.(livedebugging.DebugDataPublisher),
+		builder:            labels.NewScratchBuilder(0),
 	}
 
 	// Create and immediately export the receiver which remains the same for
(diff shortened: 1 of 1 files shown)


ptodev:
> I am running on the assumption that only one loki.relabel component can run at a time

There can certainly be more than one relabel component running at a given time. I'd approved the PR but then I saw your comment. I think your PR would work fine if there is more than one relabel component, since each component will have its own `builder`?

blewis12:
@ptodev ah yes I see, so there can be more than one instance of a component (which is fine) but that single component is only run once and not used across different threads

thampiotr:
> @ptodev ah yes I see, so there can be more than one instance of a component (which is fine) but that single component is only run once and not used across different threads

That's correct. And there is only one thread in `Run` that does the processing and uses the builder, so this PR looks correct to me. 

Not the biggest gain in performance, but I think this change won't hurt.

blewis12:
Sounds good - could I get ✅ from either of you in that case?

PR #4586, opened 2025-10-09
loki: safe apis for send and recv between components  

kalleep:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/types.go
+++ b/internal/component/common/loki/types.go
@@ -25,8 +25,8 @@ const finalEntryTimeout = 5 * time.Second
 // communication.
 type LogsReceiver interface {
 	Chan() chan Entry
-	Send(context.Context, Entry) error
-	Recv(context.Context) (Entry, error)
+	Send(context.Context, Entry) bool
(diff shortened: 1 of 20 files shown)


internal/component/common/loki/types.go
```diff
@@ -26,14 +26,8 @@ const finalEntryTimeout = 5 * time.Second
...
-}
-
-type logsReceiver struct {
-	entries chan Entry
-}
-
-func (l *logsReceiver) Chan() chan Entry {
-	return l.entries
+	Send(context.Context, Entry) error
+	Recv(context.Context) (Entry, error)
```
thampiotr: I think the only possible error you can get here is when context is cancelled. So maybe we should return `bool` and treat is as continue sending / receiving when true or stream is closed / no more data when false.

kalleep: Yes we can go with bool. I was thinking that in the future this could change but currently we only need a signal that we should stop sending / receiving so bool is fine

Issue #4580, opened 2025-10-09
Add stage.truncate to loki.process


kalleep:
### Component(s)

loki.process

### Background

In promtail and static agent you could configure `ReadlineRate`, `ReadlineBurst`, `ReadlineRateEnabled`, `ReadlineRateDrop`, `MaxLineSize` and `MaxLineSizeTruncate` for the client.

All rate limiting config can be performed using `stage.limit`, MaxLine is covered by `stage.drop`. So the only thing you cannot do currently in alloy is truncating lines.

### Proposal

Add a `stage.truncate` that will truncate log lines over a certain threshold 

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

Issue #3694, opened 2025-05-26
Secretfilter component is too heavy for big logs and don't have fallback degrading logging


andrejshapal:
### Background

Hello, 

We have enabled secret filter on our alloy setup and started to see degraded logging performance. 
What was happening:
1. One of the services produced logs every few minutes: The log line contains 130961 chars. 131 KB. This is not good to have such logs, but not something we really control.
2. The secret-filter with turned on generic-api-token was parsing these logs for seconds which was blocking the logs flow. 
3. While the flow was blocked, the log rotation on the node occurred and the logs were moved to new file.
4. After secret-filter parsed big log, alloy resumed log streaming, but the file was new therefore part of the logs was not consumed.
Such components as secret-filter are very dangerous for logging consistency for loaded services. I would like to discuss how to handle this, I could make necessary adjustments via PR in case if we will find the direction of improvement. 

### Proposal

1. I am wondering why secret-filter is standalone component and not included in loki.processor. Loki processor has several advantages:
a) As far as I remember, it allows to see in metrics the duration of each stage in processor. Handful for debugging and alerting.
b) It is possible to pass to secretfilter just some logs avoiding specific which could be not needed to filter or to avoid big logs to be parsed.

2. Loki.processor could have stage to truncate big logs to limited size. The loki config has the following:
```
# Maximum line size on ingestion path. Example: 256kb. Any log line exceeding
# this limit will be discarded unless `distributor.max-line-size-truncate` is
# set which in case it is truncated instead of discarding it completely. There
# is no limit when unset or set to 0.
# CLI flag: -distributor.max-line-size
[max_line_size: <int> | default = 256KB]

# Whether to truncate lines that exceed max_line_size.
# CLI flag: -distributor.max-line-size-truncate
[max_line_size_truncate: <boolean> | default = false]
```
But why should we even send/parse/process huge payloads if we can reduce size somewhere in the beginning. More than this, Ifaik this truncating is cheap operation. Caveats here what to do with structured logs... JSON, cri, fmt? Maybe it is possible to truncate keeping the structure? But to truncate without keeping the format is still better than processing big payloads putting the logging under risk of being overloaded and start skipping data.

3.  If loki.processor can not be used, we can keep these components standalone and chain them with more complicated structure of components (several loki.processor).

4. Since secretfilter is very heavy parsing component, I would propose the following:a) Create configurable timeout for one log procession (default 0s?). Can be achieved with context I think. 
b) Create configurable action on timeout: keep or drop
[...]

Issue #3039, opened 2025-03-20
Default Values and Configuration for limits_config in Alloy Missing After Conversion from Promtail


atayfour:
### What's wrong?

I am migrating from Promtail to Grafana Alloy, and I encountered an issue where limits_config settings are not carried over when using the alloy convert command.

### Steps to reproduce

1. In Promtail, I have the following limits_config settings:

```yaml
limits_config:
  max_line_size: 1000kb
  max_line_size_truncate: true
  readline_rate_enabled: true
  readline_rate: 2500
  readline_burst: 2500
```

2. I ran the conversion command:
`alloy convert --source-format=promtail --output=<OUTPUT_CONFIG_PATH> <INPUT_CONFIG_PATH>
`
3. However, the converted Alloy config did not include the limits_config part.

**Questions!**
1) What are the default values for the following fields in Alloy?

```
type Config struct {
	ReadlineRate        float64          `mapstructure:"readline_rate" yaml:"readline_rate" json:"readline_rate"`
	ReadlineBurst       int              `mapstructure:"readline_burst" yaml:"readline_burst" json:"readline_burst"`
	ReadlineRateEnabled bool             `mapstructure:"readline_rate_enabled,omitempty" yaml:"readline_rate_enabled,omitempty"  json:"readline_rate_enabled"`
	ReadlineRateDrop    bool             `mapstructure:"readline_rate_drop,omitempty" yaml:"readline_rate_drop,omitempty"  json:"readline_rate_drop"`
	MaxStreams          int              `mapstructure:"max_streams" yaml:"max_streams" json:"max_streams"`
	MaxLineSize         flagext.ByteSize `mapstructure:"max_line_size" yaml:"max_line_size" json:"max_line_size"`
	MaxLineSizeTruncate bool             `mapstructure:"max_line_size_truncate" yaml:"max_line_size_truncate" json:"max_line_size_truncate"`
}

```
2) How can I configure these settings in the Alloy configuration file? I am deploying Alloy via Helm charts
3) Should the conversion tool (alloy convert) include the limits_config section automatically, or is this expected behavior?

harshrai654:
Can I check this issue as well?

harshrai654:
Hello, I've been looking into this issue and wanted to share a summary of my findings.

First, my investigation confirms that the `alloy convert` command fails with an error when a Promtail config contains `limits_config`, which seems to be the intended behaviour. 
```
Error: (Error) limits_config is not yet supported in Alloy
```
[...]

PR #4587, opened 2025-10-09
Implement a truncate stage for loki.process

dehaansa:
Fixes https://github.com/grafana/alloy/issues/4580
Fixes https://github.com/grafana/alloy/issues/4486
Relevant to https://github.com/grafana/alloy/issues/3694
Relevant to https://github.com/grafana/alloy/issues/3039

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1902,7 +1902,7 @@ The following arguments are supported:
 | `suffix`      | `string` | Suffix to append to truncated values.                        | ``       | no       |
 
 The `limit` attribute should be expressed in logical units, for example `"1KiB"`.
-The stage checks the length of the log line, label values, or structured metadata values against the configured limit and truncates if it exceeds the limit.
+The stage checks the byte length of the log line, label values, or structured metadata values against the configured limit and truncates if it exceeds the limit.
 If you provide a `suffix`, the limit is reduced by the length of the `suffix`, and the `suffix` is appended to the truncated value.
 
 The `source_type` attribute must be one of `"line"`, `"label"`, `"structured_metadata"`, or `"extracted"`.


internal/component/loki/process/stages/truncate.go
```diff
@@ -0,0 +1,169 @@
...
+// Cleanup implements Stage.
+func (*truncateStage) Cleanup() {
+	// no-op
+}
+
+func getTruncateCountMetric(registerer prometheus.Registerer) *prometheus.CounterVec {
+	truncateCount := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "loki_process_truncated_fields_total",
+		Help: "A count of all log lines, labels, or structured_metadata truncated as a result of a pipeline stage",
+	}, []string{"field"})
```
thampiotr, line 236: Maybe we shouldn't make it per-field?

internal/component/loki/process/stages/truncate_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/grafana/alloy/internal/util"
+)
+
+func Test_TruncateStage_Process(t *testing.T) {
+	// Enable debug logging
+	cfg := &dskit.Config{}
+	require.Nil(t, cfg.LogLevel.Set("debug"))
+	Debug = true
+
+	tests := []struct {
```
thampiotr, line 22: Maybe add tests for defaults and pass-through.

internal/component/loki/process/stages/truncate.go
dehaansa: It's not like per label Name, it's either line, label, or structured_metadata, so just 3 max cardinality. Would love a better name than "field" for that pattern though.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
 }
 ```
 
+### `stage.truncate`
+
+The `stage.truncate` inner block configures a processing stage that will truncate log entries, labels, or structured_metadata that exceed configured limits.
```
```suggestion
The `stage.truncate` inner block configures a processing stage that truncates log entries, labels, or structured metadata that exceed configured limits.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+
+| Name                        | Type     | Description                                                      | Default | Required |
+|-----------------------------|----------|------------------------------------------------------------------|---------|----------|
+| `line_limit`                | `string` | Maximum length of a log line before truncating.                  | `""`    | no       |
+| `label_limit`               | `string` | Maximum length of a label value before truncating.               | `""`    | no       |
+| `structured_metadata_limit` | `string` | Maximum length of a structured metadata value before truncating. | `""`    | no       |
+| `suffix`                    | `string` | Suffix to append to truncate fields.                             | ``      | no       |
+
+A `truncate` stage must have at least one `limit` field configured.
+Each of the `limit` fields should be expressed in logical units (e.g. `"1KiB"`)
```
```suggestion
Each of the `limit` fields should be expressed in logical units, for example `"1KiB"`.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+The stage will check the length of the log line, label values, or structured metadata values against the configured limit and truncate if it exceeds the limit
```
```suggestion
The stage checks the length of the log line, label values, or structured metadata values against the configured limit and truncate if it exceeds the limit.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+If a `suffix` is provided, the limit will be reduced by the length of the `suffix`, and the `suffix` will be appended to the truncated field.
```
```suggestion
If you provide a `suffix`, the limit is reduced by the length of the `suffix`, and the `suffix` is appended to the truncated field.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+
+Whenever a line, label value, or structured_metadata value is truncated, the metric `loki_process_truncated_fields_total` is incremented.
+The `field` label will either be `line`, `label`, or `structured_metadata`.
+
+If anything has been truncated, the extracted fields for the entry will now contain a `"truncated"` field containing a comma delimited list of field types that have been truncated.
```
```suggestion
If anything has been truncated, the extracted fields for the entry contain a `"truncated"` field with a comma delimited list of field types that have been truncated.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+  forward_to = [loki.write.default.receiver]
+  stage.truncate {
+    entry_limit = "12B"
+    label_limit = "8B"
+    suffix = "..."
+  }
+}
+```
+
+Given the following input entry, the stage will truncate the entry `"I'm a log message!"` into `"I'm a log..."`, and the value of the `"label2"` label `"hello world"` into `"hello..."`.
```
```suggestion
Given the following input entry, the stage truncates the entry `"I'm a log message!"` into `"I'm a log..."`, and the value of the `"label2"` label `"hello world"` into `"hello..."`.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+As there is no `structured_metadata_limit` configured, the structured_metadata will not be evaluated for truncation.
```
```suggestion
As there is no `structured_metadata_limit` configured, the structured metadata isn't evaluated for truncation.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,59 @@ stage.timestamp {
...
+
+```text
+entry: "I'm a log message!"
+labels: { "label1": "hi", "label2": "hello world" }
+structured_metadata: { "metadata1": "here is some metadata", "metadata2": "and here is some more"}
+```
+
+The extracted fields map for the entry will now contain the `"truncated"` field,
+which can be used to add a label (in a `stage.labels`) or structured_metadata (in a `stage.structured_metadata`) to the entry
+so that the logs can be identified as truncated after processing.
```
```suggestion
The extracted fields map for the entry contain the `"truncated"` field.
You can use this map to add a label in `stage.labels` or structured metadata in `stage.structured_metadata` to the entry so you can identify the logs as truncated after processing.
```

docs/sources/reference/components/loki/loki.process.md
dehaansa: This one doesn't quite read right to me, should it be contains?

clayton-cornell: The hard part is removing the passive "will now" from the sentence.  If we have the plural `fields` then it'll be singular `contain`.  The fields (more than one field) contain... or the field (just one) contains.  We could say `...each extracted field contains a "truncated" field with a comma delimited list...`

dehaansa: Looks like we use "extracted map" elsewhere rather than "extracted fields", will change it and use contains.

erikfinnman:
This looks very promising for a use case we have been struggling with to handle with Alloy (we have some software writing extremely large log lines that are rejected when Alloy tries to upload the to Grafana Cloud). But will there not be a risk that this produces e.g. invalid JSON if the log format is JSON and some field is truncated in "half"?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,58 @@ stage.timestamp {
...
+Given the following input entry, the stage truncates the entry `"I'm a log message!"` into `"I'm a log..."`, and the value of the `"label2"` label `"hello world"` into `"hello..."`.
+As there is no `structured_metadata_limit` configured, the structured metadata isn't evaluated for truncation.
+
+```text
+entry: "I'm a log message!"
+labels: { "label1": "hi", "label2": "hello world" }
+structured_metadata: { "metadata1": "here is some metadata", "metadata2": "and here is some more"}
+```
+
+The extracted map for the entry contain the `"truncated"` field.
```
```suggestion
The extracted map for the entry contains the `"truncated"` field.
```

dehaansa:
> This looks very promising for a use case we have been struggling with to handle with Alloy (we have some software writing extremely large log lines that are rejected when Alloy tries to upload the to Grafana Cloud). But will there not be a risk that this produces e.g. invalid JSON if the log format is JSON and some field is truncated in "half"?

That's a valid concern, as it is this wouldn't be safe to use to truncate very large JSON entries. I don't have a great solution for that problem at the moment, and it'll require some additional thought. It seems to me that either way it could be a future enhancement to this stage (or one of the json specific stages), and wouldn't necessarily be a blocking concern.

erikfinnman:
> > This looks very promising for a use case we have been struggling with to handle with Alloy (we have some software writing extremely large log lines that are rejected when Alloy tries to upload the to Grafana Cloud). But will there not be a risk that this produces e.g. invalid JSON if the log format is JSON and some field is truncated in "half"?
> 
> That's a valid concern, as it is this wouldn't be safe to use to truncate very large JSON entries. I don't have a great solution for that problem at the moment, and it'll require some additional thought. It seems to me that either way it could be a future enhancement to this stage (or one of the json specific stages), and wouldn't necessarily be a blocking concern.

Fair enough - we solved it it the meantime we some clever use of the `stage.template` along with Go template strings `printf` statements to truncate the fields. But we did this on a field-by-field basis  (to make sure that we don't cut a valid JSON in half), so we should be able to replace that logic with this truncate stage once available.

Have thought about have fields in the extracted maps as arguments to this stage? So the truncation could be applied to individual fields only instead of the entire log line?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,58 @@ stage.timestamp {
...
+The `stage.truncate` inner block configures a processing stage that truncates log entries, labels, or structured metadata that exceed configured limits.
+
+The following arguments are supported:
+
+| Name                        | Type     | Description                                                      | Default | Required |
+|-----------------------------|----------|------------------------------------------------------------------|---------|----------|
+| `line_limit`                | `string` | Maximum length of a log line before truncating.                  | `""`    | no       |
+| `label_limit`               | `string` | Maximum length of a label value before truncating.               | `""`    | no       |
+| `structured_metadata_limit` | `string` | Maximum length of a structured metadata value before truncating. | `""`    | no       |
+| `suffix`                    | `string` | Suffix to append to truncate fields.                             | ``      | no       |
```
thampiotr:
```suggestion
| `suffix`                    | `string` | Suffix to append to truncated fields.                             | ``      | no       |
```

internal/component/loki/process/stages/truncate.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/alecthomas/units"
+	"github.com/go-kit/log"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/grafana/alloy/internal/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+)
+
+const (
+	ErrTruncateStageEmptyConfig = "truncate stage config must contain at least one of `entry_limit`, `label_limit`, or `structured_metadata_limit`"
```
thampiotr: Nit: could be private

internal/component/loki/process/stages/truncate.go
thampiotr: Probably okay then, since it's an internal name anyway.

thampiotr (approved):
LGTM, only some nits!

dehaansa:
@erikfinnman based on your feedback I refactored the stage & added support for truncating fields in the extracted map, let me know if it looks like it would be useful for you in this form or if I've missed something!

erikfinnman:
> @erikfinnman based on your feedback I refactored the stage & added support for truncating fields in the extracted map, let me know if it looks like it would be useful for you in this form or if I've missed something!

This looks really promising! We will definitely give this a spin once it's released and available.

Just to give you an idea, this is what we currently do for our JSON-based logs. As you can see, once we have trimmed the log entry, we use `stage.output` to add it back for further processing:

```

            // Initializing line content
            stage.template {
                source   = "line_content"
                template = "{{`{{ .Entry }}`}}"
            }
    
            // Cutting long message value
            stage.json {
                expressions = {message = ""}
            }
            stage.template {
                source   = "line_content"
                template = "{{`{{ if eq .message nil }}{{ else }}{{ (set (.line_content | fromJson) \"message\" (printf \"%.15000s\" .message)) | toJson }}{{ end }}`}}"
            }
    
            // Cutting long error.stack_trace value
            stage.json {
                expressions = {error_stack_trace = "\"error.stack_trace\""}
            }
            stage.template {
                source   = "line_content"
                template = "{{`{{ if eq .error_stack_trace nil }}{{ else }}{{ (set (.line_content | fromJson) \"error.stack_trace\" (printf \"%.15000s\" .error_stack_trace)) | toJson }}{{ end }}`}}"
            }
    
            // Cutting long error.stack_trace value
            stage.json {
                expressions = {error_message = "\"error.message\""}
            }
            stage.template {
                source   = "line_content"
                template = "{{`{{ if eq .error_message nil }}{{ else }}{{ (set (.line_content | fromJson) \"error.message\" (printf \"%.15000s\" .error_message)) | toJson }}{{ end }}`}}"
            }
    
            // Setting the log line content
            stage.output {
                source = "line_content"
            }

```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,82 @@ stage.timestamp {
...
+| `source_type` | `string` | Source location of the data to truncate.                     | `"line"` | no       |
+| `suffix`      | `string` | Suffix to append to truncated values.                        | ``       | no       |
+
+The `limit` attribute should be expressed in logical units, for example `"1KiB"`.
+The stage checks the length of the log line, label values, or structured metadata values against the configured limit and truncates if it exceeds the limit.
+If you provide a `suffix`, the limit is reduced by the length of the `suffix`, and the `suffix` is appended to the truncated value.
+
+The `source_type` attribute must be one of `"line"`, `"label"`, `"structured_metadata"`, or `"extracted"`.
+If the `source` attribute is specified, the stage will only truncate a label, structured_metadata, or extracted field of the same name.
+If `source` is empty, all labels, structured_metadata, or extracted fields will be truncated if they exceed the limit.
```
thampiotr, line 1910: Maybe we could call it a `key`? or `name`?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,82 @@ stage.timestamp {
...
+The following arguments are supported:
+
+| Name          | Type     | Description                                                  | Default  | Required |
+|---------------|----------|--------------------------------------------------------------|----------|----------|
+| `limit`       | `string` | Maximum length before truncating.                            | `""`     | yes      |
+| `source`      | `string` | Source of the data to truncate. If empty, will truncate all. | `""`     | no       |
+| `source_type` | `string` | Source location of the data to truncate.                     | `"line"` | no       |
+| `suffix`      | `string` | Suffix to append to truncated values.                        | ``       | no       |
+
+The `limit` attribute should be expressed in logical units, for example `"1KiB"`.
```
thampiotr, line 1904:
```suggestion
The `limit` attribute should be expressed in data size units, for example `"1KiB"`.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1848,6 +1850,82 @@ stage.timestamp {
...
+The stage checks the length of the log line, label values, or structured metadata values against the configured limit and truncates if it exceeds the limit.
```
thampiotr: len(s) will return the size in bytes, not the number of characters... and some characters will take multiple bytes. So this is accurate, but I thought maybe we mention somewhat that this is length of encoded bytes we are talking about here.

docs/sources/reference/components/loki/loki.process.md
dehaansa: I would be fine with that, but I'm following the pattern established in the other loki.process stages, `source` appears to be the standard.

PR #4602, opened 2025-10-10
database_observability: use query_start as a key instead of xid for query samples, removing idle state entries

gaantunes:
#### PR Description

In Postgres, the xid is lazily assigned, so while a query may show up in `pg_stat_activity`, it will only be assigned a xid when it affectively touches or locks data. Plus, not all queries are transactions.

This is migrating to use query_start instead of xid as the sample event keys.

Also remove `idle` state events altogether, since we need to track specifically `idle in transaction` and `idle in transaction (aborted)`, which is covered.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/query_samples.go
+++ b/internal/component/database_observability/postgres/collector/query_samples.go
@@ -57,8 +57,7 @@ const selectPgStatActivity = `
 		s.state != 'idle' AND
 		(
 			s.backend_type != 'client backend' OR
-			(
-				s.pid != pg_backend_pid() AND
+			(				
 				coalesce(TRIM(s.query), '') != '' AND
 				s.query_id != 0
 			)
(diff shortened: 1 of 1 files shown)


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -297,33 +298,15 @@ func (c *QuerySamples) scanRow(rows *sql.Rows) (QuerySamplesInfo, error) {
...
-
-	for key := range c.samples {
-		if _, stillActive := activeKeys[key]; stillActive {
-			continue
-		}
-		if _, wasIdle := idleKeys[key]; wasIdle {
-			continue
-		}
-		c.emitAndDeleteSample(key)
+	key := SampleKey{PID: sample.PID, QueryID: sample.QueryID.Int64, QueryStartNs: 0}
```
cristiangreco: Can we use a `NewSampleKey` builder method like we do for the tracker struct?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -54,12 +53,15 @@ const selectPgStatActivity = `
 	FROM pg_stat_activity s
 		JOIN pg_database d ON s.datid = d.oid AND NOT d.datistemplate AND d.datallowconn
 	WHERE
-		s.backend_type != 'client backend' OR
+		s.pid != pg_backend_pid() AND
```
cristiangreco, line 56: I _think_ that this first `s.pid != pg_backend_pid()` is redundant?

PR #4611, opened 2025-10-14
fix: faster update of file tasks for loki.source.file

kalleep:
#### PR Description
`loki.source.file` have a main loop that is responsible to read and forward log entries from tailer tasks. When we get an update to add new tasks we store new tasks and add a signal to this main loop, this is done to keep calls to Update cheap.

The problem is that we will not update running tasks if we are blocked on sending an entry. This pr splits up the consumer work and task scheduling  into two different go routines. With this change we no longer need to start a drain job while we update tasks.

The other issue we have is that `local.file_match` is only evaluating paths every 10s (default). In this pr I added a mechanism so that it will evaluate all paths and export them every time we get new discovery data.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


kalleep:
I created two pr:s for changes done in this https://github.com/grafana/alloy/pull/4634 and https://github.com/grafana/alloy/pull/4635

Issue #4623, opened 2025-10-17
Panic in otelcol.receiver.syslog during Arguments.Convert() function


slash909uk:
### Component(s)

otelcol.receiver.syslog

### What's wrong?

Upgraded from 1.10.2 -> 1.11.2 on Ubuntu Server. Restarted and immediately Alloy panics and core dumps.
Log shows the fault is in Arguments.Convert() function call of otelcol.receiver.syslog component.

Config file has not changed since 1.10.2, was working fine before.
config file and log extract attached.

[vmui_logs_2025-10-17_160754.csv](https://github.com/user-attachments/files/22975642/vmui_logs_2025-10-17_160754.csv)

[config.alloy.txt](https://github.com/user-attachments/files/22975722/config.alloy.txt)

### Steps to reproduce

start alloy with this config file. You will need --stability.level=experimental (I have use the debug output components for tracing.)

### Configuration

```text

```

### Logs

```text

```

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

slash909uk:
Have just quickly reviewed the changes in otelcol.receiver.syslog.syslog.go included in 1.11 vs 1.10 and I can see one edit at line #161 that attempts to dereference a config item (args.TCP) without checking it exists. This code was inside a check before. I don't define TCP parameters in my config, only UDP. Maybe this is the problem? Link below:

[Sorry I don't have a build environment for alloy so can't provide a PR to test it... HTH]

https://github.com/grafana/alloy/blob/2482d9d4ac8312ddf51d6ec10c3cbbb82de8d53e/internal/component/otelcol/receiver/syslog/syslog.go#L161

slash909uk:
Ok did a quick test; I introduced a TCP child entry in the config file for otelcol.receiver.syslog entity and it no longer panics/core dumps. I think line #161 is definitely at fault 👍 Log is now as below, which is as expected given there is no TCP config:

Oct 17 21:27:07 mort alloy[2582060]: 257 |   otelcol.receiver.syslog "default" {
Oct 17 21:27:07 mort alloy[2582060]:     |  _^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Oct 17 21:27:07 mort alloy[2582060]: 258 | |     protocol = "rfc3164"
Oct 17 21:27:07 mort alloy[2582060]: 259 | |     on_error = "drop"
Oct 17 21:27:07 mort alloy[2582060]: 260 | |     udp {
[...]

PR #4628, opened 2025-10-20
fix: panic when no tcp block is used for otelcol.receiver.syslog

kalleep:
#### PR Description
tcp block is optional so we just need to handle that properly when converting into otel config.

Fixes: https://github.com/grafana/alloy/issues/4623

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


blewis12 (approved):

slash909uk:
Looks good, thank you!

PR #4629, opened 2025-10-20
feat(loki.process): add regex field to structured_metadata stage

### LINKED ISSUE #2156: loki.process --> stage.json, stage.logfmt - extract all label:value pairs without explicit expression
### Request

At the moment in Grafana alloy 1.5.0 and loki.process -> stage.json needs an "expression" and stage.logfmt needs a "mapping".
all these stages make it mandatory to always know all of the label:value pairs which may appear in a log line.

On the other hand in Grafana Explore we have line filter "JSON" and "LOGFMT" and with one click all json formatted parts are correctly listed in label:value.

I would like to see a similar option in loki.process stages:

1.) I would like to be able to configures a specific set of labels I want to extract (as it is now)
2.) I would like to be able to extract all possible labels except the list I define (opposite of now.)
3.) I would like to be able to extract all possible labels.

I would like to see this for JSON and LOGFMT - maybe others are possible, too.

For following stages like "stage.labels" or "stage.struchtured_metadata" I would like to be able to do similar thing:

1.) create struchtured_metadata for all previous extracted labels
2.) create struchtured_metadata for all previous extracted labels except the list I configure
3.) create structured_metadata for a set of previous extracted labels

### Use case

This would simplify and shorten the alloy config files. If I want to extract 10 labels of a json logline I have to add 10 stage.json label:value pairs.
Next step is to add 8 out of 10 of these values to stage.structured_metadata.

If I have a stage.match and I want to match the log lines based on the process/service, then I have to do these stages and labels for all services.

In my example I have the following applications:
alloy, mimir, loki, tempo, grafana, prometheus, blackbox_exporter, smokeping_prober, snmp_exporter
All these applications log to /var/log/messages,.

I create a stage.regex and parse the process and the log message which is logfmt.

I create several stage.match for
process=alloy
process=mimir
process=loki
process=tempo
...

I create stage.logfmt for every process to extract always the same parts (ts, msg, level, component, component_id, trace_id, ..) and there are several more fields I do not know because they may only appear in e.g. debug or info level mode.

after I extracted all these 10.20 fields I create stage.structured_metadata for 17 out of 20 labels.

Every stage.match gets it's own service_name label.

If I could use:
```
stage.json {
    expressions = { ".*" }
}
```
I would extract everything which is formatted as json in the logline. if new parts appear or new software version adds new field I do not have to modify it.
```
// This would use als stage.json label_value pairs except "ts" and "msg"
stage.structured_metadata {
    values_exclude = {
      ts  = "",
      msg = "",
    }
}

// This would only use the selected labels "ts" and "msg"
stage.structured_metadata {
    values = {
      ts  = "",
      msg = "",
    }
}

// this will include all stage.json labels in struchtured_metadata
stage.structured_metadata {
    values = {
      .*  = ".*",
    }
}
```

I don't know if this is usefull for "stage.labels" as we should avoid high labels count - however this may be up to the user.

### LINKED ISSUE #4696: regex label match for loki.process/stage.structured_metadata
### Component(s)

loki.process

### Request

Processing log messages, we would like to be able to put all labels matching a specific regex into the structured_metadata. 
Currently it is only possible to explicitly select the labels by static `values` entries in `stage.structured_metadata`. 
I propose adding a parameter `map` which allows for a regexp with a group match denoting the structured_metadata label key name.  This corresponds to the logic in of the `labelmap` action in the `relabel` resources.

I will attach a PR.

### Use case

Ingesting RFC5424 syslog we have structured data assigned to incoming logs. This metadata is "dynamic" (quasi abritrary depending on the sender). To avoid explosion of cardinality due to new labels, we want to be able to "store" all incoming structured data as corresponding structured metadata.

Therfore we need a stage like:
```
loki.process "syslog" {
    stage.structured_metadata {
        values = {}
        map = "syslog_message_(.*)"
    }
}
```

### Tip

timonegk:
#### PR Description

This PR adds a `regex` field to the `structured_metadata` stage in `loki.process`. It follows the comments on #2578.

Fixes #4696 and fixes #2156.

#### Notes to the Reviewer

In the test, I had to convert the `LabelsAdapter` to a map to be able to compare them without requiring them to be in the same order.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/structured_metadata.go
+++ b/internal/component/loki/process/stages/structured_metadata.go
@@ -16,15 +16,15 @@ type StructuredMetadataConfig struct {
 }
 
 func newStructuredMetadataStage(logger log.Logger, configs StructuredMetadataConfig) (Stage, error) {
-  var validatedLabelsConfig map[string]string
-  var err error
+	var validatedLabelsConfig map[string]string
+	var err error
 
-  if len(configs.Values) > 0 {
-    validatedLabelsConfig, err = validateLabelsConfig(configs.Values)
-    if err != nil {
-      return nil, err
-    }
-  }
+	if len(configs.Values) > 0 {
+		validatedLabelsConfig, err = validateLabelsConfig(configs.Values)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	re, err := regexp.Compile(configs.Regex)
 	if err != nil {


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1483,14 +1483,16 @@ stage.static_labels {
...
-| -------- | ------------- | --------------------------------------------------------------------------- | ------- | -------- |
-| `values` | `map(string)` | Specifies the list of labels to add from extracted values map to log entry. | `{}`    | no       |
+| Name     | Type          | Description                                                                                 | Default | Required |
+| -------- | ------------- | ------------------------------------------------------------------------------------------- | ------- | -------- |
+| `values` | `map(string)` | Specifies the list of labels to add from extracted values map to log entry.                 | `{}`    | no       |
+| `regex`  | `string`      | Specifies a regular expression, matching extracted values are added to structured metadata. | `""`    | no       |
 
-In a `structured_metadata` stage, the map's keys define the label to set and the values are how to look them up.
+In a `structured_metadata` stage, the `values` map's keys define the label to set and the values are how to look them up.
```
clayton-cornell: This sentence doesn't read right with the addition of `values`. As I read the original text, it's a key:value pair. The key defines the label and the value is the lookup. What has changed here?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1483,14 +1483,16 @@ stage.static_labels {
 ### `stage.structured_metadata`
 
 The `stage.structured_metadata` inner block configures a stage that can read data from the extracted values map and add them to log entries as structured metadata.
+If any labels are added to structured metadata, they are removed from the label map.
```
```suggestion
If you add labels to structured metadata, the labels are removed from the label map.
```

docs/sources/reference/components/loki/loki.process.md
timonegk: The name of the attribute is `values`, and it is a map of keys and values, which is a bit confusing. I added `values` because it is now no longer the only possible attribute.
If you have a suggestion on how to change the line, it would be welcome.

clayton-cornell:
```suggestion
In a `structured_metadata` stage, the keys for the `values` map define the label to set and the values define how to look them up.
```

Maybe like this?  This way it's a bit more clear that we are referring to the map by its attribute name.

timonegk: I added your suggestion but changed "label" to "labels".

internal/component/loki/process/stages/structured_metadata_test.go
ptodev, line 1: In the current examples all the values we are adding are strings. It'd be nice to add a test for an extracted value which is a map itself. E.g. someone might use [stage.json](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagejson) and get an extracted map which contains nested values.

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -33,9 +60,33 @@ func (*structuredMetadataStage) Cleanup() {
...
 			e.StructuredMetadata = append(e.StructuredMetadata, push.LabelAdapter{Name: string(labelName), Value: string(labelValue)})
 		})
+		// Handle extracted values matching the regex
+		if s.regex.String() != "" {
+			for lName, lValue := range e.Extracted {
+				if s.regex.MatchString(string(lName)) {
+					str, err := getString(lValue)
+					if err != nil {
+						if Debug {
+							level.Debug(s.logger).Log("msg", "failed to convert extracted label value to string", "err", err, "type", reflect.TypeOf(lValue))
```
ptodev, line 68: In what case would the value be invalid? Is it worth documenting it and having a test for it?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1523,6 +1501,7 @@ stage.structured_metadata {
       env  = "",         // Sets up an 'env' property to structured metadata, based on the 'env' extracted value.
       user = "username", // Sets up a 'user' property to structured metadata, based on the 'username' extracted value.
     }
+    regex = "label_.*"   // Adds all extracted values starting with 'label_' to structured metadata.
```
ptodev:
```suggestion
    regex = "__pod_.*"   // Adds all extracted values starting with '__pod_' to structured metadata.
```

A more realistic example would be nice.

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -1,24 +1,51 @@
...
-func newStructuredMetadataStage(logger log.Logger, configs LabelsConfig) (Stage, error) {
-	labelsConfig, err := validateLabelsConfig(configs)
+type StructuredMetadataConfig struct {
+	Values map[string]*string `alloy:"values,attr,optional"`
+	Regex  string             `alloy:"regex,attr,optional"`
+}
+
+func newStructuredMetadataStage(logger log.Logger, configs StructuredMetadataConfig) (Stage, error) {
+	labelsConfig := map[string]string{}
+	// same as in labels.go, but allow empty list
```
ptodev: Why not check if Values has a length greater than 0, and then call `validateLabelsConfig` only in that case? Alternatively, you can call `validateLabelsConfig` for every case and then ignore `ErrEmptyLabelStageConfig`, but that wouldn't be as clean.

internal/component/loki/process/stages/structured_metadata.go
timonegk: I don't exactly know. I used the same safeguards as in `labels.go`. I assume that it's for invalid UTF8 strings. Since `labels_test.go` does not test that case, I don't think it's necessary here.

internal/component/loki/process/stages/structured_metadata_test.go
timonegk: Sorry, I don't understand the use case. How should a map be converted to a label?

docs/sources/reference/components/loki/loki.process.md
timonegk: This is actually exactly the use case for which I was implementing the feature. I want to use a `discovery.relabel` component to rename all `__meta_kubernetes_pod_label_(.*)` to `label_$1` and then use `structured_metadata` in `loki.process` to move them to structured metadata.
But I see that the example `label_.*` is not very helpful because it is not clear that it refers to kubernetes pod labels instead of some generic label. But I think the `__pod_.*` example is not better because it is probably not best practice to add custom double-underscore metadata to Loki, or is it?

internal/component/loki/process/stages/structured_metadata.go
timonegk: Good suggestion, I changed it accordingly.

internal/component/loki/process/stages/structured_metadata_test.go
ptodev: The value is just a string. I pushed a commit to your branch with the test :) I hope you don't mind.

The regex does have a limitation that it can only act on the top level keys of the extracted map, but I don't think that will be a problem. If you need to extract keys and values from nested values then I think it's possible to pull them to the top level of the extracted map in an earlier stage.

internal/component/loki/process/stages/structured_metadata.go
ptodev: Yes, let's not do this. I tried to write a test for this but then it dawned on me that if you have an invalid label, the chances are it'd get caught by an earlier process step for parsing the log line.

docs/sources/reference/components/loki/loki.process.md
ptodev: Interesting question. I don't know. Let's use `pod_.*` in the example, just to be safe?

internal/component/loki/process/stages/structured_metadata_test.go
timonegk: Thank you!

docs/sources/reference/components/loki/loki.process.md
timonegk: Alright, I changed the example.

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -1,24 +1,46 @@
...
+	Regex  string             `alloy:"regex,attr,optional"`
+}
+
+func newStructuredMetadataStage(logger log.Logger, configs StructuredMetadataConfig) (Stage, error) {
+  var validatedLabelsConfig map[string]string
+  var err error
+
+  if len(configs.Values) > 0 {
+    labelsConfig := LabelsConfig{Values: configs.Values}
+    validatedLabelsConfig, err = validateLabelsConfig(labelsConfig)
```
ptodev:
```suggestion
    validatedLabelsConfig, err = validateLabelsConfig(configs.Values)
```

It'd be cleaner to change the function signature to accept `map[string]*string`.

ptodev: I'm not sure why the build for the PR is not starting... @timonegk would you mind rebasing the PR to see if it helps run the CI? If you could also change this bit of code here it would be great please.

timonegk: I added the change and rebased the PR.

timonegk:
When I run the tests, the new test `expected_structured_metadata_from_nested_values` seems to be flaky due to the order of the nested object:
```
--- FAIL: Test_StructuredMetadataStage (0.00s)
    --- FAIL: Test_StructuredMetadataStage/expected_structured_metadata_from_nested_values (0.00s)
        structured_metadata_test.go:218: 
            	Error Trace:	/home/timon/Projekte/repositories/alloy/internal/component/loki/process/stages/structured_metadata_test.go:218
            	Error:      	Not equal: 
            	            	expected: map[string]string{"component_nested":"{\"name\":\"ingester\",\"props\":{\"n1\":\"v1\",\"n2\":\"v2\"}}", "component_non_nested":"non_nested_val"}
            	            	actual  : map[string]string{"component_nested":"{\"name\":\"ingester\",\"props\":{\"n2\":\"v2\",\"n1\":\"v1\"}}", "component_non_nested":"non_nested_val"}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,3 +1,3 @@
            	            	 (map[string]string) (len=2) {
            	            	- (string) (len=16) "component_nested": (string) (len=49) "{\"name\":\"ingester\",\"props\":{\"n1\":\"v1\",\"n2\":\"v2\"}}",
            	            	+ (string) (len=16) "component_nested": (string) (len=49) "{\"name\":\"ingester\",\"props\":{\"n2\":\"v2\",\"n1\":\"v1\"}}",
            	            	  (string) (len=20) "component_non_nested": (string) (len=14) "non_nested_val"
            	Test:       	Test_StructuredMetadataStage/expected_structured_metadata_from_nested_values
FAIL
FAIL	github.com/grafana/alloy/internal/component/loki/process/stages	0.024s
FAIL
```

PR #4633, opened 2025-10-21
chore: use doublestar v4

### LINKED ISSUE #4423: Alloy local.file_match supporting include/exclude multiple path
### Component(s)

_No response_

### Background

I often has the need to tail file from several path pattern `a`,`b`,`c` or exclude file with multiple pattern such as `.*`, `*.gz`

### Proposal

Currently from the [official doc](https://grafana.com/docs/alloy/latest/reference/components/local/local.file_match/) we can only configure a single path 
The underlying file match lib that alloy use is [doublestar](https://github.com/bmatcuk/doublestar), for newer version (v4) it supports pattern like {a,b,c} to enable multiple match.

shall upgrade the lib version seems it is a low hang fruit.
https://github.com/grafana/alloy/blob/main/internal/component/local/file_match/watch.go#L26

### Tip

kalleep:
#### PR Description
Update to use v4 of doublestar for file globbing. Not sure why we did not use it before because promtail also uses v4 

https://github.com/grafana/loki/blob/main/clients/pkg/promtail/targets/file/filetarget.go#L10
https://github.com/grafana/loki/blob/main/clients/pkg/promtail/targets/file/filetarget.go#L249

Fixes: https://github.com/grafana/alloy/issues/4423

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/local/file_match/watch.go
+++ b/internal/component/local/file_match/watch.go
@@ -5,7 +5,7 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/bmatcuk/doublestar"
+	"github.com/bmatcuk/doublestar/v4"
 	"github.com/go-kit/log"
 
 	"github.com/grafana/alloy/internal/component/common/loki/utils"
@@ -23,7 +23,7 @@ type watch struct {
 func (w *watch) getPaths() ([]discovery.Target, error) {
 	allMatchingPaths := make([]discovery.Target, 0)
 
-	matches, err := doublestar.Glob(w.getPath())
+	matches, err := doublestar.FilepathGlob(w.getPath())
 	if err != nil {
 		return nil, err
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@ Main (unreleased)
 
 - `prometheus.exporter.snowflake` dependency has been updated to 20251016132346-6d442402afb2, which updates data ownership queries to use `last_over_time` for a 24 hour period. (@dasomeone)
 
+- `local.file_match` now useses github.com/bmatcuk/doublestar/v4. (@kalleep)
+
 ### Bugfixes
 
 - Stop `loki.source.kubernetes` discarding log lines with duplicate timestamps. (@ciaranj)
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,6 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.87.3
 	github.com/aws/aws-sdk-go-v2/service/servicediscovery v1.37.0
 	github.com/blang/semver/v4 v4.0.0
-	github.com/bmatcuk/doublestar v1.3.4
 	github.com/boynux/squid-exporter v1.10.5-0.20230618153315-c1fae094e18e
 	github.com/buger/jsonparser v1.1.1
 	github.com/burningalchemist/sql_exporter v0.0.0-20240103092044-466b38b6abc4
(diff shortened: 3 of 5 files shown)


internal/component/local/file_match/watch.go
```diff
@@ -23,7 +23,7 @@ type watch struct {
 func (w *watch) getPaths() ([]discovery.Target, error) {
 	allMatchingPaths := make([]discovery.Target, 0)
 
-	matches, err := doublestar.Glob(w.getPath())
```
blewis12, line 26: Do we understand the differences between the v1 Glob() method vs v4 FilepathGlob()? I'm wondering if there could be subtle differences in behavior (looking [here](https://github.com/bmatcuk/doublestar/blob/v4/UPGRADING.md))

kalleep: yep there could be some small differences. Will take a look at what those can be and document it. Depending on what those changes are it could prevent us from updating this lib

blewis12: Sounds good to me 👍 if the differences are doable then we could add more tests, so long as it's documented in some way

kalleep:
Using this branch to test a couple of things

jharvey10:
@kalleep Before you close this one out, could I request that it stay open for a couple days? This is a great candidate for testing out AI analysis of a dep update and calling out code changes we may need to make.

jharvey10:
## 🔍 Dependency Review

<details>
<summary>github.com/bmatcuk/doublestar v1.3.4 → github.com/bmatcuk/doublestar/v4 v4.9.1 — ❌ Changes Needed</summary>

Reasoning:
- This is a major-version jump with a module path change (v4 module).
- v4 introduces a split between “path-style” and “filepath/OS-style” APIs. Code using filesystem paths must switch to the Filepath* variants.
- Import path changes to include /v4.

What to change:
- Update imports from `github.com/bmatcuk/doublestar` to `github.com/bmatcuk/doublestar/v4`.
- Replace any filesystem globbing calls that previously used `doublestar.Glob(...)` with `doublestar.FilepathGlob(...)`.
- Replace any filesystem matching calls that previously used `doublestar.Match(...)` with `doublestar.FilepathMatch(...)` (if present in your codebase).
- Keep using the non-Filepath variants only for forward-slash, non-OS-specific “virtual path” matching.

Concrete code updates observed/required:

- internal/component/local/file_match/watch.go
```diff
- import "github.com/bmatcuk/doublestar"
+ import "github.com/bmatcuk/doublestar/v4"

- matches, err := doublestar.Glob(w.getPath())
+ matches, err := doublestar.FilepathGlob(w.getPath())
```

- internal/component/loki/source/file/file.go
```diff
+ import "github.com/bmatcuk/doublestar/v4"
```

If you have other usages (search for “doublestar.”):
- Filesystem matching:
```diff
- ok, err := doublestar.Match(pattern, filename)
+ ok, err := doublestar.FilepathMatch(pattern, filename)
```
- Filesystem globbing:
```diff
- files, err := doublestar.Glob(pattern)
+ files, err := doublestar.FilepathGlob(pattern)
```

Why:
- In v4, the “path-style” APIs (`Match`, `Glob`, etc.) operate on forward-slash paths and do not interpret OS-specific separators (not suitable for Windows paths). The “Filepath*” variants are intended for real filesystem paths and honor OS-specific separators.

Relevant changelog/notes (upstream):
- v4.0.0 (breaking): Introduced v4 module path and split APIs into path-style and filepath-style variants. For filesystem use-cases, switch to `FilepathMatch`/`FilepathGlob`. Import path is now `github.com/bmatcuk/doublestar/v4`.
  - See: bmatcuk/doublestar releases v4.0.0 and README sections on “Filepath functions”.

</details>


## Notes

- No other module dependency changes require review based on the provided go.mod diff.
- This PR already implements the necessary import/API changes for doublestar in the touched files; verify there are no remaining usages of `doublestar.Glob`/`doublestar.Match` elsewhere that operate on filesystem paths. A quick grep for `doublestar.` is recommended.

PR #4634, opened 2025-10-21
fix: loki source file scheduleloop

kalleep:
#### PR Description
Extracted from https://github.com/grafana/alloy/pull/4611

Run `consumer` loop and `scheduler` loop in their own go routines. With this we no longer need to start start a `consumer` go routine while updating the tasks. This would also allow us to remove / add tasks faster because we won't be block by send calls.

Another improvement was made to `Update`, we now track paths that we have checked regardless if we could create a task or not. Before if we had several targets (likely with k8s discovery) we would stat files that failed for other targets causing two identical logs and unnecessary syscalls. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/file/file.go
+++ b/internal/component/loki/source/file/file.go
@@ -247,7 +247,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 	// There are cases where we have several targets with the same path + public labels
 	// but the path no longer exist so we cannot create a task for it. So we need to track
-	// what we have checked seperatly from the task map to prevent performing checks that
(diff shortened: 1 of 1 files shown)


kalleep:
>LGTM. Since this was pretty problematic piece of code, can we make sure we follow-up with this in an internal cluster to verify that restart/reload works fine? thanks!

I have been running this in our largest dev cluster since this morning with a couple of restarts so we should be fine

PR #4635, opened 2025-10-21
fix: trigger the sync job when targets are updated

kalleep:
#### PR Description
Extracted from https://github.com/grafana/alloy/pull/4611

Before this component would only export new targets whenever it started or if the timer was triggered. When using this component with discovery mechanism it is useful to "expand" paths and exports as soon as we have new discovery data.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,6 @@ Main (unreleased)
 
 - Schedule new path targets faster in `loki.source.file`. (@kalleep)
 
-- `local.file_match` will now trigger it's sync job when the component is updated with new targets. (@kalleep)
 
 ### Bugfixes
 
@@ -64,6 +63,8 @@ Main (unreleased)
 - Support Scrape Protocol specification in CRDS for `prometheus.operator.*` components. (@dehaansa) 
 
 - Fix `otelcol.exporter.splunkhec` arguments missing documented `otel_attrs_to_hec_metadata` block. (@dehaansa)
+
+- `local.file_match` now publish targets faster whenever targets in arguments changes. (@kalleep)
 
 v1.11.2
 -----------------


internal/component/local/file_match/file.go
```diff
@@ -37,20 +37,23 @@ var _ component.Component = (*Component)(nil)
...
 func New(o component.Options, args Arguments) (*Component, error) {
 	c := &Component{
 		opts:     o,
-		mut:      sync.RWMutex{},
+		mut:      sync.Mutex{},
 		args:     args,
 		watches:  make([]watch, 0),
 		watchDog: time.NewTicker(args.SyncPeriod),
+		// Buffered channel to avoid blocking
+		targetsChanged: make(chan struct{}, 1),
```
thampiotr, line 56: I don't think it needs to be buffered, because we have a `default:` in select too.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ Main (unreleased)
 
 - `prometheus.exporter.postgres` dependency has been updated to v0.18.1. This includes new `stat_progress_vacuum` and `buffercache_summary` collectors, as well as other bugfixes and enhancements. (@cristiangreco)
 
+- `local.file_match` will now trigger it's sync job when the component is updated with new targets. (@kalleep)
```
thampiotr: That's a bit of an implementation detail. I think you want to write a bugfix message on what this fixes.

internal/component/local/file_match/file.go
kalleep: I think a unbuffered channel is fine but we use this in all other places e.g. https://github.com/grafana/alloy/blob/main/internal/component/loki/source/syslog/syslog.go#L59

PR #4636, opened 2025-10-21
Upgrade Prometheus to v3.7.1 and resolve issues

thampiotr:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/common/loki/client/queue_client.go
```diff
@@ -261,9 +261,7 @@ func (c *queueClient) StoreSeries(series []record.RefSeries, segment int) {
 	c.seriesLock.Lock()
 	defer c.seriesLock.Unlock()
 	for _, seriesRec := range series {
-		c.seriesSegment[seriesRec.Ref] = segment
```
kalleep, line 264: Was this removed intentionally?

internal/component/common/loki/client/queue_client.go
```diff
@@ -261,9 +261,7 @@ func (c *queueClient) StoreSeries(series []record.RefSeries, segment int) {
...
-		labels := lokiutil.MapToModelLabelSet(seriesRec.Labels.Map())
-		c.series[seriesRec.Ref] = labels
```
kalleep, line 266: I guess MapToModelLabelSet is not safe to use if you are not using slicelabels? 

Maybe we should add a lint rule that bans it?

internal/component/common/loki/wal/writer.go
```diff
@@ -261,8 +261,7 @@ func newEntryWriter() *entryWriter {
 // write, it first has to be reset, and then overwritten accordingly. Therefore, WriteEntry is not thread-safe.
 func (ew *entryWriter) WriteEntry(entry loki.Entry, wl WAL, _ log.Logger) error {
 	// Reset wal record slices
-	ew.reusableWALRecord.RefEntries = ew.reusableWALRecord.RefEntries[:0]
-	ew.reusableWALRecord.Series = ew.reusableWALRecord.Series[:0]
+	ew.reusableWALRecord.Reset()
```
kalleep, line 264: 👍

internal/component/prometheus/remotewrite/remote_write.go
```diff
@@ -92,7 +92,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 			log.With(o.Logger, "subcomponent", "rw"),
 		),
 	)
-	remoteStore := remote.NewStorage(remoteLogger, o.Registerer, startTime, o.DataPath, remoteFlushDeadline, nil)
+	remoteStore := remote.NewStorage(remoteLogger, o.Registerer, startTime, o.DataPath, remoteFlushDeadline, nil, false)
```
kalleep, line 96: should we have a fixme here to look into the last option. From what I gathered if true it will try to extract unit and type from labels and if successful send metadata.

internal/static/agentctl/waltools/samples.go
```diff
@@ -75,7 +75,7 @@ func FindSamples(walDir string, selectorStr string) ([]*SampleStats, error) {
 }
 
 func collectSeries(r *wlog.Reader, selector labels.Selector, labelsByRef map[chunks.HeadSeriesRef]labels.Labels) error {
-	var dec record.Decoder
+	dec := record.NewDecoder(nil, nil)
```
kalleep: Not sure it's fine to pass nil for logger. This would panic whenever a log would be emitted, maybe pass a nop logger?

internal/static/metrics/wal/util.go
```diff
@@ -65,7 +65,7 @@ func (r walReplayer) Replay(dir string) error {
 }
 
 func (r walReplayer) replayWAL(reader *wlog.Reader) error {
-	var dec record.Decoder
+	dec := record.NewDecoder(nil, nil)
```
kalleep: Same here

go.mod
```diff
@@ -137,18 +138,18 @@ require (
...
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheus v0.134.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributesprocessor v0.134.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/cumulativetodeltaprocessor v0.134.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/processor/deltatocumulativeprocessor v0.134.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/processor/deltatocumulativeprocessor v0.136.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/filterprocessor v0.134.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/groupbyattrsprocessor v0.134.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/intervalprocessor v0.134.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor v0.134.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sattributesprocessor v0.138.0
```
kalleep, line 152: Did we have to update the otel deps?

kalleep (commented):
Some small comments. 

We have the same problem as we do for https://github.com/grafana/alloy/pull/4543. That we need to partially update otel deps

internal/component/common/loki/client/queue_client.go
thampiotr: No! 🤦

internal/component/common/loki/client/queue_client.go
thampiotr: It is safe in some other situations. And it will be safe once we move to `stringlabels` which will come soon, so I wouldn't worry about this for now. There's also a test that can catch this.

internal/component/prometheus/remotewrite/remote_write.go
thampiotr: Yes, we don't have that feature exposed yet. I'll add a TODO and a GH issue. I don't want to be adding more scope to this PR and this can wait as it's experimental.

internal/service/labelstore/service_test.go
```diff
@@ -16,11 +16,7 @@ import (
 
 func TestAddingMarker(t *testing.T) {
 	mapping := New(log.NewNopLogger(), prometheus.DefaultRegisterer)
-	l := labels.Labels{}
-	l = append(l, labels.Label{
-		Name:  "__name__",
-		Value: "test",
-	})
+	l := labels.FromStrings("__name__", "test")
```
blewis12, line 19: I must have missed these in the first iteration of the labels API implementation - thanks!

internal/component/prometheus/operator/configgen/config_gen_probe_test.go
```diff
@@ -262,16 +265,20 @@ func TestGenerateProbeConfig(t *testing.T) {
...
+				ScrapeProtocols:                []config.ScrapeProtocol{config.PrometheusProto},
+				ScrapeFallbackProtocol:         config.PrometheusText0_0_4,
+				AlwaysScrapeClassicHistograms:  falsePtr,
+				ConvertClassicHistogramsToNHCB: falsePtr,
+				EnableCompression:              true,
+				MetricsPath:                    "/probe",
+				Scheme:                         "http",
+				Params:                         url.Values{"module": []string{"http_2xx"}},
+				MetricNameValidationScheme:     model.LegacyValidation,
+				MetricNameEscapingScheme:       "underscores",
```
blewis12: Would it be considered bad practice to use `model.UnderscoreEscaping.String()` instead of hard coding the string? I understand this is how it's implemented elsewhere though

internal/static/agentctl/waltools/samples.go
thampiotr: Yes! thanks for catching this. Not sure if this code is in use, but it's best to fix it.

internal/component/prometheus/operator/configgen/config_gen_probe_test.go
thampiotr: Probably not. I can change this, but I'll arbitrarily leave the `"http"` strings, even though I'm sure there is a constant for them somewhere.

go.mod
thampiotr: Yes, some things were clashing and it was tough to resolve. In the case of `k8sattributesprocessor` we needed it to support k8s.io/client-go v0.34.1 which was pulled in by Prometheus.

thampiotr: And we definitely don't want to couple the OTel upgrade with Prometheus upgrade, the change set size is already too large. So I think this is unfortunately the way forward, at least for now. In the future we can think of coordinating what versions we want to upgrade to better, so we choose versions that generate the least pain, even if they are a bit behind 🤔

PR #4641, opened 2025-10-21
labelstore: Allow local links to be replaced

kgeckhart:
#### PR Description
This PR adjust the labelstore to allow for local links to be able to be replaced. The labelstore does not own the local series refs, the WAL does and as such the WAL can GC the local series ref used by the labelstore without the label store knowing. 

If this happens we will pass a series ref to the WAL appender that might not be valid to the WAL which is fine for use cases like [Append](https://github.com/grafana/alloy/blob/3e66dc5bde713216af155e40db3f7c938501148b/internal/static/metrics/wal/wal.go#L713-L737) which will create a new series ref if the one provided is invalid. But when calling [AppendExemplar](https://github.com/grafana/alloy/blob/3e66dc5bde713216af155e40db3f7c938501148b/internal/static/metrics/wal/wal.go#L769-L774) an invalid series ref is an error case. We have seen some reports of exemplar errors which look like, 
```
ts=2025-09-19T06:58:35.477632626Z level=error msg="failed to write exemplar for metric sample" component_path=/ component_id=otelcol.exporter.prometheus.metrics_converter <metric>=err "1 error occurred:\n\t* unknown series ref when trying to add exemplar: 75717\n\n"=(MISSING)  
ts=2025-09-19T06:58:35.477694227Z level=error msg="failed to add exemplar" component_path=/ component_id=otelcol.exporter.prometheus.metrics_converter "metric name"=<metric> bucket=+Inf err="1 error occurred:\n\t* unknown series ref when trying to add exemplar: 75721\n\n"
```

Since we logged a SeriesRef it means we had a link from a GlobalRef -> LocalRef but that LocalRef does not appear to be valid any longer. The code path which appends exemplars for our otel prometheus exporter will always send the series before the exemplar, [https://github.com/grafana/alloy/blob/fc96e92d385cf0ab04ca693ac9df5fcc7ace0a32/int[…]mponent/otelcol/exporter/prometheus/internal/convert/convert.go](https://github.com/grafana/alloy/blob/fc96e92d385cf0ab04ca693ac9df5fcc7ace0a32/internal/component/otelcol/exporter/prometheus/internal/convert/convert.go#L449-L470) -> [https://github.com/grafana/alloy/blob/fc96e92d385cf0ab04ca693ac9df5fcc7ace0a32/int[…]mponent/otelcol/exporter/prometheus/internal/convert/convert.go](https://github.com/grafana/alloy/blob/fc96e92d385cf0ab04ca693ac9df5fcc7ace0a32/internal/component/otelcol/exporter/prometheus/internal/convert/convert.go#L349-L359) -> [https://github.com/grafana/alloy/blob/946fda9c3766afda5f6af684f75f27fc201ec098/int[…]component/otelcol/exporter/prometheus/internal/convert/cache.go](https://github.com/grafana/alloy/blob/946fda9c3766afda5f6af684f75f27fc201ec098/internal/component/otelcol/exporter/prometheus/internal/convert/cache.go#L124-L133), so in this case it seems like the series exists but our local link is no longer valid and when the series was sent it should have replaced the existing link.  

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/remotewrite/remote_write.go
+++ b/internal/component/prometheus/remotewrite/remote_write.go
@@ -131,6 +131,13 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		if cachedLocalRef != 0 && cachedLocalRef == newLocalRef {
 			return
 		}
+
+		// There are some unique scenarios that can have an append end with no error but the returned localRef is zero (duplicate exemplars).
+		// We don't want to update a valid link to an invalid link
+		if cachedLocalRef != 0 && newLocalRef == 0 {
+			return
+		}
+
 		// This should never happen in a proper appender chain. Since we cannot enforce it, we are extra defensive.
 		if globalRef == 0 {
 			globalRef = ls.GetOrAddGlobalRefID(l)
@@ -231,9 +238,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 			localRef := ls.GetLocalRefID(res.opts.ID, uint64(globalRef))
 			newLocalRef, nextErr := next.AppendExemplar(storage.SeriesRef(localRef), l, e)
-
-			// We shouldn't add a local link if there was an error or the new local ref was zero which can happen for duplicate exemplars
-			if nextErr == nil && newLocalRef != 0 {
+			if nextErr == nil {
 				handleLocalLink(uint64(globalRef), l, localRef, uint64(newLocalRef))
 			}
 


internal/component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -463,7 +463,7 @@ func (conv *Converter) consumeSum(app storage.Appender, memResource *memorySerie
 		if convType == model.MetricTypeCounter {
 			for i := 0; i < dp.Exemplars().Len(); i++ {
 				if err := conv.writeExemplar(app, memSeries, dp.Exemplars().At(i)); err != nil {
-					level.Error(conv.log).Log("msg", "failed to write exemplar for metric sample", metricName, "err", err)
```
kgeckhart, line 466: Drive-by fix for the broken error log noted in the PR

internal/service/labelstore/data.go
```diff
@@ -3,15 +3,12 @@ package labelstore
...
-	// GetOrAddLink returns the global id for the values, if none found one will be created based on the lbls.
-	GetOrAddLink(componentID string, localRefID uint64, lbls labels.Labels) uint64
+	// AddLocalLink adds a mapping from local to global id for the given component.
+	AddLocalLink(componentID string, globalRefID uint64, localRefID uint64)
 
 	// GetOrAddGlobalRefID finds or adds a global id for the given label map.
 	GetOrAddGlobalRefID(l labels.Labels) uint64
 
-	// GetGlobalRefID returns the global id for a component and the local id. Returns 0 if nothing found.
-	GetGlobalRefID(componentID string, localRefID uint64) uint64
```
kgeckhart, line 13: Nothing was calling this function

internal/component/prometheus/remotewrite/remote_write.go
```diff
@@ -142,9 +161,8 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 			localID := ls.GetLocalRefID(res.opts.ID, uint64(globalRef))
 			newRef, nextErr := next.Append(storage.SeriesRef(localID), l, t, v)
-			if localID == 0 {
-				ls.GetOrAddLink(res.opts.ID, uint64(newRef), l)
-			}
+			handleLocalLink(uint64(globalRef), l, localID, uint64(newRef))
+
```
kgeckhart: The `GetOrAddLink` implementation was ignoring the incoming ref which is pretty safe to assume is a valid GlobalRef from the labelstore with how our appenders work.

dehaansa (approved):
LGTM.

kgeckhart:
I found out this is also a bug in prometheus pipelines but it's not logged as an error it's logged as debug, 

```
ts=2025-10-27T18:50:56.256627037Z level=debug msg="Error while adding exemplar in AddExemplar" component_path=/ component_id=prometheus.scrape.kubernetes_pods scrape_pool=kubernetes-pods target=<> exemplar="{Labels: <redacted>} Value:2.551167972 Ts:1761588711493 HasTs:true}" err="1 error occurred:\n\t* unknown series ref when trying to add exemplar: 1716785\n\n"
```

Running the version in this PR does not produce these errors anymore. We also had external confirmation this fixed OTLP pipelines as well.

PR #4642, opened 2025-10-22
Implement behavior in loki.source.file to handle corrupted positions file entries

dehaansa:
#### PR Description
A user ended up with `\0` nul characters in their positions file. To allow the collector to recover from this issue and continue to tail the file(s) affected, we built behavior to recover from corrupted positions file entries.

While the previous behavior would have "skipped" a file with a bad positions entry, I chose to make the default to restart a file as this is the most likely desired behavior from users. In this case there may be duplicate log entries processed, but there is a reduced chance of data loss.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/internal/component/common/loki/positions/positions.go
+++ b/internal/component/common/loki/positions/positions.go
@@ -245,7 +245,6 @@ func (p *positions) Get(path, labels string) (int64, error) {
 	if !ok {
 		return 0, nil
 	}
-
 	return strconv.ParseInt(pos, 10, 64)
 }
 
--- a/internal/component/loki/source/file/decompresser_test.go
+++ b/internal/component/loki/source/file/decompresser_test.go
@@ -359,7 +359,7 @@ func TestDecompressorPositionFileEntryDeleted(t *testing.T) {
 		l,
 		ch1,
 		positionsFile,
-		func() bool { return true },
+		func() bool { return false },
 		sourceOptions{
 			path:                 filename,
 			labels:               labels,


internal/component/loki/source/file/file.go
```diff
@@ -44,13 +44,41 @@ const (
...
+	case "skip":
+		*o = OnPositionsFileErrorSkip
+	case "restart_from_end":
+		*o = OnPositionsFileErrorRestartEnd
+	case "restart_from_start":
+		*o = OnPositionsFileErrorRestartStart
+	default:
+		return fmt.Errorf("unknown OnPositionsFileError value: %s", s)
+	}
+	return nil
```
kalleep:
```suggestion
	s := OnPositionsFileError(text)
	switch s {
	case OnPositionsFileErrorSkip, OnPositionsFileErrorRestartEnd, OnPositionsFileErrorRestartStart:
		*o = s
	default:
		return fmt.Errorf("unknown OnPositionsFileError value: %s", s)
	}
	return nil
```

internal/component/loki/source/file/file.go
```diff
@@ -280,13 +309,15 @@ func (c *Component) Update(args component.Arguments) error {
...
-			tailFromEnd:         newArgs.TailFromEnd,
-			legacyPositionUsed:  newArgs.LegacyPositionsFile != "",
+			path:                 path,
+			labels:               labels,
+			encoding:             newArgs.Encoding,
+			decompressionConfig:  newArgs.DecompressionConfig,
+			fileWatch:            newArgs.FileWatch,
+			tailFromEnd:          newArgs.TailFromEnd,
+			onPositionsFileError: newArgs.OnPositionsFileError,
+
```
kalleep, line 350:
```suggestion
```

internal/component/loki/source/file/decompresser.go
```diff
@@ -82,7 +85,21 @@ func newDecompressor(
...
+			level.Warn(logger).Log("msg", "`restart_from_end` is not supported for compressed files, defaulting to `skip`")
+			return nil, fmt.Errorf("failed to get file position: %w", err)
+		case OnPositionsFileErrorRestartStart:
+			pos = 0
+			positions.Put(path, labelsStr, pos)
+			level.Info(logger).Log("msg", "reset position to start of file after positions error")
+		default:
+			level.Debug(logger).Log("msg", "unrecognized `on_positions_file_error` option, defaulting to `skip`", "option", onPositionsFileError)
+			return nil, fmt.Errorf("failed to get file position: %w", err)
+		}
```
kalleep: This is identical in decompresser and tailer, maybe a helper function?

dehaansa: It's _almost_ identical, decompressor doesn't support `restart_from_end`. I thought about decomposing but didn't like my ideas for adding specific behavior for that behavior being different per use.

internal/component/loki/source/file/file.go
```diff
@@ -44,13 +44,41 @@ const (
...
+	LegacyPositionsFile  string               `alloy:"legacy_positions_file,attr,optional"`
+	OnPositionsFileError OnPositionsFileError `alloy:"on_positions_file_error,attr,optional"`
+}
+
+type OnPositionsFileError string
+
+const (
+	OnPositionsFileErrorSkip         OnPositionsFileError = "skip"
+	OnPositionsFileErrorRestartEnd   OnPositionsFileError = "restart_from_end"
+	OnPositionsFileErrorRestartStart OnPositionsFileError = "restart_from_start"
```
kgeckhart: WDYT about?

```suggestion
	OnPositionsFileErrorRestartBeginning OnPositionsFileError = "restart_from_beginning"
```

Something about `restart_from_start` just feels really awkward to say, not a huge deal

internal/component/loki/source/file/file.go
```diff
@@ -280,13 +309,15 @@ func (c *Component) Update(args component.Arguments) error {
...
+			encoding:             newArgs.Encoding,
+			decompressionConfig:  newArgs.DecompressionConfig,
+			fileWatch:            newArgs.FileWatch,
+			tailFromEnd:          newArgs.TailFromEnd,
+			onPositionsFileError: newArgs.OnPositionsFileError,
+
+			legacyPositionUsed: newArgs.LegacyPositionsFile != "",
 		})
 		if err != nil {
 			continue
```
kgeckhart, line 365: Somewhat relevant for this PR, the way logging is handled in this component is awkward. Anything that returns an error causes the file to be skipped but instead of logging that info here we have a bunch of lower level errors logged that don't mention that the file is skipped 🤔 . Now with your PR we add logs that mention "skip" but no other logs say anything about skipping happening that just happens implicitly when a different error is logged?

internal/component/loki/source/file/decompresser.go
kgeckhart:
```suggestion
		switch onPositionsFileError {
		case OnPositionsFileErrorSkip:
			return nil, fmt.Errorf("failed to get file position: %w", err)
		case OnPositionsFileErrorRestartEnd:
			level.Warn(logger).Log("msg", "`restart_from_end` is not supported for compressed files, defaulting to `skip`")
			return nil, fmt.Errorf("failed to get file position: %w", err)
		case OnPositionsFileErrorRestartStart:
			pos = 0
			positions.Put(path, labelsStr, pos)
			level.Info(logger).Log("msg", "reset position to start of file after positions error", "original_error", err)
		default:
			level.Warn(logger).Log("msg", "unrecognized `on_positions_file_error` option, defaulting to `skip`", "option", onPositionsFileError)
			return nil, fmt.Errorf("failed to get file position: %w", err)
		}
```

Cleaning up the logging a bit, anything that returns an error is going to log it again so no need to double log. I think unifying the restart log with the error that caused the restart will make it easier to parse instead of having to look for a separate log line for the original error.

internal/component/loki/source/file/tailer.go
```diff
@@ -182,7 +184,25 @@ func (t *tailer) initRun() (loki.EntryHandler, error) {
 
 	pos, err := t.positions.Get(t.path, t.labelsStr)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get file position: %w", err)
+		level.Error(t.logger).Log("msg", "failed to get file position", "file", t.path, "error", err)
```
kgeckhart: Same changes here for logging as in the decompressor

clayton-cornell:
No input on docs changes. Looks OK there.

PR #4644, opened 2025-10-22
fix: pass RemoveEmptyGroups to process exporter config

mhamzahkhan:
#### PR Description

I noticed that even though I had remove_empty_groups set to true, the number of empty process groups was still rising (eg using `count(namedprocess_namegroup_num_procs == 0)`), and noticed that in the code the `RemoveEmptyGroups` field was not being passed through in the `Convert()` method.

With this fix, it now appears to be working and I no longer see empty process groups.

#### PR Checklist

- [x] CHANGELOG.md updated


dehaansa (approved):
Thanks for fixing this!

PR #4653, opened 2025-10-23
Docs: Fix broken link in sources/troubleshoot/debug

tiffanyfay:
#### PR Description
[] was used instead of () for a link so it was broken

N/A


clayton-cornell (approved):
Thanks for catching this!

PR #4669, opened 2025-10-24
prometheus.exporter.unix: Expose ARP collector config options

ptodev:
#### PR Description

This is useful for customers who use non-Ethernet networking devices like Infiniband.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.unix.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.unix.md
@@ -58,6 +58,7 @@ You can use the following blocks with `prometheus.exporter.unix`:
 
 | Name                         | Description                             | Required |
 | ---------------------------- | --------------------------------------- | -------- |
+| [`arp`][arp]                 | Configures the `arp` collector.         | no       |
 | [`bcache`][bcache]           | Configures the `bcache` collector.      | no       |
 | [`cpu`][cpu]                 | Configures the `cpu` collector.         | no       |
 | [`disk`][disk]               | Configures the `diskstats` collector.   | no       |
@@ -79,6 +80,7 @@ You can use the following blocks with `prometheus.exporter.unix`:
 | [`textfile`][textfile]       | Configures the `textfile` collector.    | no       |
 | [`vmstat`][vmstat]           | Configures the `vmstat` collector.      | no       |
 
+[arp]: #arp
 [bcache]: #bcache
 [cpu]: #cpu
 [disk]: #disk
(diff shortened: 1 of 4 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.unix.md
```diff
@@ -100,6 +102,16 @@ You can use the following blocks with `prometheus.exporter.unix`:
...
 
+### `arp`
+
+| Name             | Type      | Description                                                                                             | Default | Required |
+| ---------------- | --------- | --------------------------------------------------------------------------------------------------------| ------- | -------- |
+| `device_exclude` | `string`  | Regular expression of devices to exclude for `arp` collector. Mutually exclusive with `device_include`. |         | no       |
+| `device_include` | `string`  | Regular expression of devices to include for `arp` collector. Mutually exclusive with `device_exclude`. |         | no       |
+| `netlink`        | `boolean` | Use netlink to gather ARP stats instead of `/proc/net/arp`.                                             | `false` | no       |
+
+It is recommended to set `netlink` to `true` on systems with InfiniBand or other non-Ethernet devices.
```
blewis12, line 113: Looking at the code and once again a bit confused - isn't using `netlink` instead of `/proc/net/arp` set to[ true by default](https://github.com/grafana/node_exporter/blob/612bdf9540ec087883bb11b8e36023f04a4db5b2/kingpinconfig/flags.go#L27) in the node exporter?

ptodev: In Alloy netlink has been set to `false` because [it's set to blankBool](https://github.com/grafana/alloy/pull/4669/files#diff-9e4e5650512e1032c76452efb1e8b4498263c6463ef7a45e55870bffb78f1e99L494) which equals `false`. We can leave it to `false` so that we don't break any customers.

I think we should consider setting it to `true` by default for Alloy v2 or for a subsequent Alloy minor release. I could open an issue for it. Thanks for spotting this! I didn't realise it's the default upstream.

blewis12: > In Alloy netlink has been set to false because it's set to blankBool which equals false

ahh right I see - sounds good to me as a follow up 👍

ptodev: I opened #4673 and added the [v2.0-breaking-change](https://github.com/grafana/alloy/issues?q=state%3Aopen%20label%3A%22v2.0-breaking-change%22) label to it.

PR #4674, opened 2025-10-24
Fix issues from textparser breaking change

kgeckhart:
#### PR Description

Fixes the components impacted by the breaking change in prometheus/common exfmt.TextParser included in v0.66.0
image

I also added integration tests for the static and unix exporter. The use case for these is not often used in a default setup but in the case of  prometheus.exporter.unix it's used by the k8s helm chart so expanding the test to cover seemed useful.

I didn't want to take on the burden of adding windows integration tests ATM but https://github.com/prometheus-community/windows_exporter/blob/v0.31.3/internal/collector/textfile/textfile.go#L387 v0.31.3 is what we need. I didn't see anything particularly concerning in the changelog https://github.com/prometheus-community/windows_exporter/releases. We might want to pull https://github.com/prometheus-community/windows_exporter/pull/2165 in but def for a separate PR.

All other use cases I found were either tests, utilities for tests, or utilities we don't call
image

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,8 @@ Main (unreleased)
 
 - `local.file_match` now publish targets faster whenever targets in arguments changes. (@kalleep)
 
+- Fix breaking changes in the texfile collector for `prometheus.exporter.windows`, and `prometheus.exporter.unix`, when prometheus/common was upgraded. (@kgeckhart)
+
 v1.11.2
 -----------------
 


internal/component/prometheus/exporter/static/static.go
```diff
@@ -85,7 +85,8 @@ type Integration struct {
 }
 
 func (i *Integration) MetricsHandler() (http.Handler, error) {
-	var p expfmt.TextParser
+	// TODO: add support for choosing validation scheme: https://github.com/grafana/alloy/issues/4122
+	p := expfmt.NewTextParser(model.LegacyValidation)
```
kalleep, line 89: Ah sorry for missing this one. I caught this issue in Validate because I added a test for it.. Thanks for adding a integration test!

PR #4683, opened 2025-10-24
docs(prometheus.scrape): Augment prometheus.scrape 'scheme' argument strengthening link to protocol.

lewismc:
#### Augment prometheus.scrape 'scheme' argument strengthening link to protocol.

During a recently customer call, the `scheme` argument was discussed and the customer seemed surprised/confused
- why use `scheme` and not `protocol`?
- what are the valid values for `scheme`?
This small PR hopefully offers more clarity in a bid to answer those questions. 
Thanks

N/A

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added


docs/sources/reference/components/prometheus/prometheus.scrape.md
```diff
@@ -33,7 +33,15 @@ The list of arguments that can be used to configure the block is presented below
...
-If conflicting attributes are passed, for example, defining both a BearerToken and BearerTokenFile or configuring both Basic Authorization and OAuth2 at the same time, the component reports an error.
+One of the following can be provided:
+
+* [`authorization`][authorization] block
+* [`basic_auth`][basic_auth] block
+* [`bearer_token_file`](#arguments) argument
+* [`bearer_token`](#arguments) argument
+* [`oauth2`][oauth2] block
+
+If conflicting attributes are passed, e.g., defining both a `bearer_token` and `bearer_token_file` or configuring both `basic_auth` and `oauth2` at the same time, the component reports an error.
```
clayton-cornell:
```suggestion
If conflicting attributes are passed, for example, defining both a `bearer_token` and `bearer_token_file` or configuring both `basic_auth` and `oauth2` at the same time, the component reports an error.
```

docs/sources/reference/components/prometheus/prometheus.scrape.md
```diff
@@ -69,7 +77,7 @@ You can use the following arguments with `prometheus.scrape`:
...
 | `sample_limit`                       | `uint`                  | More than this many samples post metric-relabeling causes the scrape to fail                                             |                                                                                                  | no       |
-| `scheme`                             | `string`                | The URL scheme with which to fetch metrics from targets.                                                                 |                                                                                                  | no       |
+| `scheme`                             | `string`                | The URL protocol scheme with which to fetch metrics from targets.                                                        |                                                                                                  | no       |
```
clayton-cornell:
```suggestion
| `scheme`                             | `string`                | The URL protocol scheme used to fetch metrics from targets.                                                        |                                                                                                  | no       |
```

We could probably use `used` here and get rid of the `with which`.  It's not wrong but no one speaks like that anymore.

clayton-cornell (commented):
A couple minor tweaks we can include in this clarification/cleanup.

lewismc:
Thanks for the review @clayton-cornell 👍

PR #4686, opened 2025-10-25
Add flags to the `run` command for configuring cAdvisor housekeeping

### LINKED ISSUE #4685: cAdvisor exporter high CPU usage
### Component(s)

prometheus.exporter.cadvisor

### Add a description

I've set up grafana alloy inside a docker container with:
- prometheus unix exporter
- prometheus cadvisor exporter
- prometheus remote_write

It does not have many time series and only scaps every 60 seconds. Despite that, it constantly has 5% CPU usage on a 2 vCPU VM. Which is well above what I expected for it (~1%) especially while idle.

After profiling it seems, that the prometheus cadvisor exporter is responsible for a good chunk of that - specifically github.com/google/cadvisor/manager.(*containerData).housekeeping

See [cAdvisor housekeeping docs](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md#housekeeping) and [related cAdvisor issue](https://github.com/google/cadvisor/issues/2523).

Basically cAdvisor runs its housekeeping (read metric collection) **every second** by default. This can be configured via a cAdvisor flag. However in grafana alloy it cannot be configured. So we're stuck with the 1s default.

It would be nice to be able to increase the housekeeping interval to 30s or 60s for my use case to reduce unnecessary stats polling.

```console
$ alloy --version
alloy, version v1.11.2 (branch: HEAD, revision: 4a8f93c)
  build user:       root@buildkitsandbox
  build date:       2025-10-10T13:12:19Z
  go version:       go1.24.6
  platform:         linux/amd64
  tags:             netgo,builtinassets,promtail_journal_enabled
```

```console
$ docker --version6.16.8-200.fc42.x86_64
Docker version 28.4.0, build 1.fc42
```

```console
$ uname -r
6.16.8-200.fc42.x86_64
```

```console
$ cat config.alloy
prometheus.exporter.unix "node" {
  rootfs_path = "/host/root"
  procfs_path = "/host/proc"
  sysfs_path  = "/host/sys"
  enable_collectors = ["systemd", "mountstats"]

  systemd {
    enable_restarts = true
    start_time = true
    task_metrics = true
  }
}

prometheus.scrape "node_scraper" {
  targets         = prometheus.exporter.unix.node.targets
  forward_to      = [ prometheus.relabel.node.receiver ]
  scrape_interval = "60s"
}

prometheus.relabel "node" {
  forward_to = [ prometheus.remote_write.main.receiver ]

  rule {
    target_label  = "instance"
    replacement   = "hostname"
  }

  rule {
    target_label  = "nodename"
    replacement   = "hostname"
  }
}

prometheus.exporter.cadvisor "container" {
  store_container_labels = false
  allowlisted_container_labels = ["com.docker.compose.project", "com.docker.compose.service", "org.opencontainers.image.name", "org.opencontainers.image.version"]
}

prometheus.scrape "scrape_metrics" {
  targets         = prometheus.exporter.cadvisor.container.targets
  forward_to      = [ prometheus.relabel.container.receiver ]
  scrape_interval = "60s"
}

prometheus.relabel "container" {
  forward_to = [ prometheus.remote_write.main.receiver ]

  rule {
    target_label = "instance"
    replacement  = "hostname"
  }
}

prometheus.remote_write "main" {
  endpoint {
    url = "http://prometheus:9090/api/v1/write"

    metadata_config {
      send_interval = "60s"
    }
  }
}
```

### Tip

into-the-v0id:
#### PR Description

Goal of this PR is to make cAdvisor housekeeping configurable in grafana alloy (see #4685).

cAdvisor handles the configuration of its housekeeping exclusively via flags. See [cAdvisor housekeeping docs](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md#housekeeping) and [cAdvisor source code](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/container.go#L50). The variable in the source code is also global - so we cannot configure this on an alloy config block level. We would need to keep this setting also global in grafana alloy. Following options:
- global options in alloy config file
- environment variables
- pass through flags

I've went with passing through the flags since most grafana alloy users already pass custom flags to alloy and it's less hassle than a new config option in the config files.

The `run` cli command has now the following flags:

```
--cadvisor.allow-dynamic-housekeeping            Whether to allow the cadvisor housekeeping interval to be dynamic (default true)
--cadvisor.housekeeping-interval duration        Interval between cadvisor container housekeepings (default 1s)
--cadvisor.max-housekeeping-interval duration    Largest interval to allow between cadvisor container housekeepings (default 1m0s)
```

I've also made dynamic housekeeping configurable while I'm at it ([cAdvisor source code](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/manager.go#L69)).

The defaults match the defaults of cAdvisor - so this shouldn't be a breaking change.

Fixes #4685

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ Main (unreleased)
 
 - `prometheus.exporter.unix`: Add an `arp` config block to configure the ARP collector. (@ptodev)
 
+- Make cAdvisor housekeeping configurable via flags for the `run` command (@into-the-v0id)
+
 ### Bugfixes
 
 - Stop `loki.source.kubernetes` discarding log lines with duplicate timestamps. (@ciaranj)
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/go-kit/log"
+	"github.com/google/cadvisor/manager"
 	"github.com/grafana/ckit/advertise"
 	"github.com/grafana/ckit/peer"
 	"github.com/prometheus/client_golang/prometheus"
@@ -55,20 +56,23 @@ import (
 
 func runCommand() *cobra.Command {
 	r := &alloyRun{
-		inMemoryAddr:          "alloy.internal:12345",
-		httpListenAddr:        "127.0.0.1:12345",
-		storagePath:           "data-alloy/",
-		minStability:          featuregate.StabilityGenerallyAvailable,
-		uiPrefix:              "/",
-		disableReporting:      false,
-		enablePprof:           true,
-		configFormat:          "alloy",
-		clusterAdvInterfaces:  advertise.DefaultInterfaces,
-		clusterMaxJoinPeers:   5,
-		clusterRejoinInterval: 60 * time.Second,
-		disableSupportBundle:  false,
-		windowsPriority:       windowspriority.PriorityNormal,
-		taskShutdownDeadline:  10 * time.Minute,
+		inMemoryAddr:                     "alloy.internal:12345",
+		httpListenAddr:                   "127.0.0.1:12345",
+		storagePath:                      "data-alloy/",
+		minStability:                     featuregate.StabilityGenerallyAvailable,
+		uiPrefix:                         "/",
+		disableReporting:                 false,
+		enablePprof:                      true,
+		configFormat:                     "alloy",
+		clusterAdvInterfaces:             advertise.DefaultInterfaces,
+		clusterMaxJoinPeers:              5,
+		clusterRejoinInterval:            60 * time.Second,
+		disableSupportBundle:             false,
+		windowsPriority:                  windowspriority.PriorityNormal,
+		taskShutdownDeadline:             10 * time.Minute,
+		cadvisorHousekeepingInterval:     1 * time.Second,
+		cadvisorAllowDynamicHousekeeping: true,
+		cadvisorMaxHousekeepingInterval:  60 * time.Second,
 	}
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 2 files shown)


into-the-v0id:
Hi @kalleep, thanks for the reply.

The `manager.HousekeepingConfigFlags` (and the `HousekeepingConfig` param to `manager.New`) only control dynamic housekeeping. So the following cAdvisor CLI Flags:

- `--allow_dynamic_housekeeping`
- `--max_housekeeping_interval`

Those could be configured within a grafana alloy config block.

However, what I care most about is the `--housekeeping_interval` Flag which cannot be configured when calling `manager.New`. This Flag is read [here](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/container.go#L50) into the `HousekeepingInterval` global variable. [Here](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/container.go#L440) this variable is passed to the container controller. As far as I can see there is no way to make this Flag configurable in a grafana alloy `prometheus.exporter.cadvisor` config block. Only as a new global config block.

kalleep:
> Hi @kalleep, thanks for the reply.
> 
> The `manager.HousekeepingConfigFlags` (and the `HousekeepingConfig` param to `manager.New`) only control dynamic housekeeping. So the following cAdvisor CLI Flags:
> 
> * `--allow_dynamic_housekeeping`
> * `--max_housekeeping_interval`
> 
> Those could be configured within a grafana alloy config block.
> 
> However, what I care most about is the `--housekeeping_interval` Flag which cannot be configured when calling `manager.New`. This Flag is read [here](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/container.go#L50) into the `HousekeepingInterval` global variable. [Here](https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/container.go#L440) this variable is passed to the container controller. As far as I can see there is no way to make this Flag configurable in a grafana alloy `prometheus.exporter.cadvisor` config block. Only as a new global config block.

Gotcha, still we don't want to add a global flag for it in alloy. We do use a fork https://github.com/grafana/cadvisor to be able to run cadvisor. So if we want this we could update the fork first and then add this config in alloy

mikepayne02:
@into-the-v0id I am currently running across this issue on my alloy instance (high cpu usage from cadvisor housekeeping). Looking at the code, I came to the same conclusion that the configuration struct couldn't be used to change the housekeeping interval. 

I agree with @kalleep that it would be better to make changes to the fork than pass the global flags. I'll look into what changes can be made.

PR #4688, opened 2025-10-26
feat(faro): add per-app rate limiting for receiver

### LINKED ISSUE #4687: Add per-application rate limiting for Faro receiver
### Component(s)

faro.receiver

### Background

Currently, the Faro receiver applies rate limiting globally across all applications. When one application generates excessive telemetry data, it consumes the entire rate limit quota and affects all other applications sending data through the same Grafana Alloy instance.

Excessive telemetry data can result from misconfigured sampling rates in the Faro Web SDK configuration. Because this configuration is set in each application, you can't control the data flow at ingestion without affecting all applications sending data through the centralized gateway. This creates a cascading failure where one application impacts the availability of the Faro receiver (or Loki/Tempo) for all other applications.

This is particularly problematic in multi-tenant environments where multiple frontend applications share the same Faro receiver gateway (in a company with multiple applications for example).

This feature improves segregation between applications and increases the reliability of the receiver.

### Proposal

Add per-application rate limiting to the Faro receiver component. Each unique combination of application and environment would get its own rate limiter with configurable limits.

The choice of the `app/env` tag is driven by the configuration that needs to be defined when setting up the Faro SDK.

Configuration example:

```alloy
faro.receiver "default" {
    server {
        rate_limiting {
            // existing
            enabled = true
            rate = 100
            burst_size = 200

            // proposal
            per_app_enabled = true
            per_app_rate = 50
            per_app_burst_size = 100
        }
    }
    
    output {
        logs = [loki.write.default.receiver]
    }
}
```

The application and environment would be extracted from the Faro payload metadata (app/env fields).

### Tip

hhertout:
#### PR Description

Adds per-application rate limiting to the Faro receiver component to prevent 
one application from consuming the rate limit quota of other applications.

Currently, rate limiting is **global across all applications**. When one application 
generates excessive telemetry (due to misconfigured sampling or bugs), it affects 
all other applications sending data through the same Grafana Alloy instance.

This is particularly problematic in multi-tenant environments where multiple 
frontend applications share the same Faro receiver gateway.

When `per_app_enabled` is `true`, each unique combination of application and 
environment gets its own rate limiter with the configured limits. The application 
and environment are extracted from the Faro payload metadata (`app.name` and 
`app.environment` fields).

If these fields are missing, values default to "unknown" to ensure isolation.

**configuration example**

```alloy
faro.receiver "default" {
    server {
        rate_limiting {
            per_app_enabled = true
            per_app_rate = 50
            per_app_burst_size = 100
        }
    }
    
    output {
        logs = [loki.write.default.receiver]
    }
}
```

Fixes #4687

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/faro/receiver/handler.go
+++ b/internal/component/faro/receiver/handler.go
@@ -199,10 +199,9 @@ func (h *handler) processRequest(rw http.ResponseWriter, req *http.Request, p pa
 }
 
 // extractAppEnv extracts the app and environment from the Faro payload metadata.
-// Returns "unknown" for missing values to ensure isolation.
+// Returns empty strings for missing values to ensure isolation without collision with valid app names.
 func (h *handler) extractAppEnv(p payload.Payload) (string, string) {
-	app := "unknown"
-	env := "unknown"
+	app, env := "", ""
 
 	if p.Meta.App.Name != "" {
 		app = p.Meta.App.Name
--- a/internal/component/faro/receiver/rate_limiter.go
+++ b/internal/component/faro/receiver/rate_limiter.go
@@ -24,16 +24,6 @@ const (
 // Segregates rate limiting configurations per application and environment.
 type AppRateLimitingConfigKey string
 
-// AppRateLimiter wraps a rate limiter with metadata for cleanup purposes.
-// The lastUsed field tracks when the limiter was last accessed, allowing
-// the cleanup routine to remove limiters that haven't been used within
-// DEFAULT_LIMITER_EXPIRY duration (10 minutes by default).
-// This prevents memory leaks from applications that stop sending requests.
-type AppRateLimiter struct {
-	limiter  *rate.Limiter
-	lastUsed time.Time
-}
-
 // String returns the string representation of the AppRateLimitingConfigKey.
 func (k AppRateLimitingConfigKey) String() string {
 	return string(k)
@@ -42,6 +32,16 @@ func (k AppRateLimitingConfigKey) String() string {
 // ParseAppRateLimitingConfigKey creates a key from app and environment values.
 func ParseAppRateLimitingConfigKey(app, env string) AppRateLimitingConfigKey {
 	return AppRateLimitingConfigKey(fmt.Sprintf("%s:%s", app, env))
+}
+
+// AppRateLimiter wraps a rate limiter with metadata for cleanup purposes.
+// The lastUsed field tracks when the limiter was last accessed, allowing
+// the cleanup routine to remove limiters that haven't been used within
+// DEFAULT_LIMITER_EXPIRY duration (10 minutes by default).
+// This prevents memory leaks from applications that stop sending requests.
+type AppRateLimiter struct {
+	limiter  *rate.Limiter
+	lastUsed time.Time
 }
 
 // AppRateLimitingConfig manages rate limiters per application/environment combination.
--- a/internal/component/faro/receiver/handler_test.go
+++ b/internal/component/faro/receiver/handler_test.go
@@ -505,7 +505,7 @@ func TestHandler_RateLimitingPerApp_EmptyMetadata(t *testing.T) {
 				"exceptions": [],
 				"measurements": []
 			}`,
-			description:    "Apps without meta field share unknown:unknown rate limiter",
+			description:    "Apps without meta field share ':' rate limiter (empty app and env)",
 			requests:       3,
 			expectedStatus: []int{http.StatusAccepted, http.StatusAccepted, http.StatusTooManyRequests},
 		},
@@ -523,7 +523,7 @@ func TestHandler_RateLimitingPerApp_EmptyMetadata(t *testing.T) {
 					}
 				}
 			}`,
-			description:    "Empty app name uses unknown:production key",
+			description:    "Empty app name uses ':production' key",
 			requests:       3,
 			expectedStatus: []int{http.StatusAccepted, http.StatusAccepted, http.StatusTooManyRequests},
 		},
@@ -541,7 +541,7 @@ func TestHandler_RateLimitingPerApp_EmptyMetadata(t *testing.T) {
 					}
 				}
 			}`,
-			description:    "Empty environment uses myapp:unknown key",
+			description:    "Empty environment uses 'myapp:' key",
 			requests:       3,
 			expectedStatus: []int{http.StatusAccepted, http.StatusAccepted, http.StatusTooManyRequests},
 		},
@@ -559,7 +559,7 @@ func TestHandler_RateLimitingPerApp_EmptyMetadata(t *testing.T) {
 					}
 				}
 			}`,
-			description:    "Both empty falls back to unknown:unknown",
+			description:    "Both empty uses ':' key (empty app and env)",
 			requests:       3,
 			expectedStatus: []int{http.StatusAccepted, http.StatusAccepted, http.StatusTooManyRequests},
 		},
@@ -594,7 +594,7 @@ func TestHandler_RateLimitingPerApp_EmptyMetadata(t *testing.T) {
 }
 
 func TestHandler_RateLimitingPerApp_MultipleAppsWithoutMetadata(t *testing.T) {
-	// This test verifies that multiple apps without metadata share the same rate limiter (unknown:unknown)
+	// This test verifies that multiple apps without metadata share the same rate limiter (':' - empty app and env)
 	handler := newHandler(util.TestLogger(t), prometheus.NewRegistry(), []exporter{})
 
 	args := ServerArguments{
@@ -602,7 +602,7 @@ func TestHandler_RateLimitingPerApp_MultipleAppsWithoutMetadata(t *testing.T) {
 			Enabled:   true,
 			Strategy:  "per_app",
 			Rate:      1.0,
-			BurstSize: 2.0, // burst of 2 requests total for all unknown apps
+			BurstSize: 2.0, // burst of 2 requests total for all apps without metadata
 		},
 	}
 	handler.Update(args)
@@ -626,8 +626,8 @@ func TestHandler_RateLimitingPerApp_MultipleAppsWithoutMetadata(t *testing.T) {
 		}
 	}
 
-	// Only 2 requests should succeed (burst size = 2) since all share unknown:unknown
-	assert.Equal(t, 2, successCount, "All apps without metadata should share the same rate limiter (unknown:unknown)")
+	// Only 2 requests should succeed (burst size = 2) since all share ':' rate limiter
+	assert.Equal(t, 2, successCount, "All apps without metadata should share the same rate limiter (':' - empty strings)")
 }
 
 func TestHandler_RateLimitingPerApp_ProperMetadataNotAffectedByUnknown(t *testing.T) {
@@ -644,28 +644,28 @@ func TestHandler_RateLimitingPerApp_ProperMetadataNotAffectedByUnknown(t *testin
 	}
 	handler.Update(args)
 
-	// First, exhaust the unknown:unknown quota with 2 requests (burst = 2)
-	unknownPayloads := []string{
+	// First, exhaust the ':' quota (empty app/env) with 2 requests (burst = 2)
+	emptyMetadataPayloads := []string{
 		`{"logs": [], "meta": {}}`,
 		`{"logs": []}`,
 	}
 
-	for _, payloadStr := range unknownPayloads {
+	for _, payloadStr := range emptyMetadataPayloads {
 		req := httptest.NewRequest("POST", "/", strings.NewReader(payloadStr))
 		req.Header.Set("Content-Type", "application/json")
 
 		rr := httptest.NewRecorder()
 		handler.handleRequest(rr, req)
 
-		assert.Equal(t, http.StatusAccepted, rr.Code, "unknown apps should succeed within burst limit")
+		assert.Equal(t, http.StatusAccepted, rr.Code, "apps without metadata should succeed within burst limit")
 	}
 
-	// Verify that a third unknown request is rejected
+	// Verify that a third request without metadata is rejected
 	req := httptest.NewRequest("POST", "/", strings.NewReader(`{"logs": []}`))
 	req.Header.Set("Content-Type", "application/json")
 	rr := httptest.NewRecorder()
 	handler.handleRequest(rr, req)
-	assert.Equal(t, http.StatusTooManyRequests, rr.Code, "third unknown request should be rate limited")
+	assert.Equal(t, http.StatusTooManyRequests, rr.Code, "third request without metadata should be rate limited")
 
 	// Now send requests from an app with proper metadata - should have its own quota
 	properAppPayload := `{
@@ -678,8 +678,8 @@ func TestHandler_RateLimitingPerApp_ProperMetadataNotAffectedByUnknown(t *testin
 		}
 	}`
 
-	// Should be able to make 2 requests (burst = 2) even though unknown:unknown is exhausted
-	// But not affected by previous unknown requests
+	// Should be able to make 2 requests (burst = 2) even though ':' quota is exhausted
+	// Not affected by previous requests without metadata
 	for i := range 2 {
 		req := httptest.NewRequest("POST", "/", strings.NewReader(properAppPayload))
 		req.Header.Set("Content-Type", "application/json")
@@ -690,7 +690,7 @@ func TestHandler_RateLimitingPerApp_ProperMetadataNotAffectedByUnknown(t *testin
 		assert.Equal(t, http.StatusAccepted, rr.Code, "myapp:production should have its own quota, request %d", i+1)
 	}
 
-	// Third request for myapp:production should be rate limited (its own limit, not affected by unknown:unknown)
+	// Third request for myapp:production should be rate limited (its own limit, not affected by ':' limiter)
 	req = httptest.NewRequest("POST", "/", strings.NewReader(properAppPayload))
 	req.Header.Set("Content-Type", "application/json")
 	rr = httptest.NewRecorder()
@@ -729,7 +729,7 @@ func TestHandler_ExtractAppEnv(t *testing.T) {
 					},
 				},
 			},
-			expectedApp: "unknown",
+			expectedApp: "",
 			expectedEnv: "production",
 		},
 		{
@@ -742,13 +742,13 @@ func TestHandler_ExtractAppEnv(t *testing.T) {
 				},
 			},
 			expectedApp: "myapp",
-			expectedEnv: "unknown",
+			expectedEnv: "",
 		},
 		{
 			name:        "empty payload",
 			payload:     payload.Payload{},
-			expectedApp: "unknown",
-			expectedEnv: "unknown",
+			expectedApp: "",
+			expectedEnv: "",
 		},
 	}
 


docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -115,18 +115,25 @@ If the `api_key` argument is empty, no authentication checks are performed, and
...
+| `enabled`            | `bool`   | Whether to enable rate limiting.                     | `true`  | no       |
+| `rate`               | `number` | Rate of allowed requests per second.                 | `50`    | no       |
+| `burst_size`         | `number` | Allowed burst size of requests.                      | `100`   | no       |
+| `per_app_enabled`    | `bool`   | Whether to enable per-application rate limiting.     | `false` | no       |
+| `per_app_rate`       | `number` | Rate of allowed requests per second per application. | `50`    | no       |
+| `per_app_burst_size` | `number` | Allowed burst size of requests per application.      | `100`   | no       |
 
 Rate limiting functions as a [token bucket algorithm][token-bucket], where a bucket has a maximum capacity for up to `burst_size` requests and refills at a rate of `rate` per second.
 
 Each HTTP request drains the capacity of the bucket by one. After the bucket is empty, HTTP requests are rejected with an `HTTP 429 Too Many Requests` status code until the bucket has more available capacity.
```
clayton-cornell:
```suggestion
Each HTTP request drains the capacity of the bucket by one.
After the bucket is empty, HTTP requests are rejected with an `HTTP 429 Too Many Requests` status code until the bucket has more available capacity.
```

Just adding a line break

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -115,18 +115,25 @@ If the `api_key` argument is empty, no authentication checks are performed, and
...
 
 Configuring the `rate` argument determines how fast the bucket refills, and configuring the `burst_size` argument determines how many requests can be received in a burst before the bucket is empty and starts rejecting requests.
 
+When `per_app_enabled` is `true`, rate limiting is applied per application and environment combination. Each unique combination of application and environment gets its own rate limiter with the limits specified by `per_app_rate` and `per_app_burst_size`. This prevents one application from affecting the rate limits of other applications sending their data on the same gateway.
```
clayton-cornell:
```suggestion
When `per_app_enabled` is `true`, rate limiting is applied per application and environment combination.
Each unique combination of application and environment gets its own rate limiter with the limits specified by `per_app_rate` and `per_app_burst_size`.
This prevents one application from affecting the rate limits of other applications sending their data on the same gateway.
```

Just adding semantic line breaks

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -115,18 +115,31 @@ If the `api_key` argument is empty, no authentication checks are performed, and
...
-| `rate`       | `number` | Rate of allowed requests per second. | `50`    | no       |
-| `burst_size` | `number` | Allowed burst size of requests.      | `100`   | no       |
+| Name                 | Type     | Description                                          | Default | Required |
+| -------------------- | -------- | ---------------------------------------------------- | ------- | -------- |
+| `enabled`            | `bool`   | Whether to enable rate limiting.                     | `true`  | no       |
+| `rate`               | `number` | Rate of allowed requests per second.                 | `50`    | no       |
+| `burst_size`         | `number` | Allowed burst size of requests.                      | `100`   | no       |
+| `per_app_enabled`    | `bool`   | Whether to enable per-application rate limiting.     | `false` | no       |
+| `per_app_rate`       | `number` | Rate of allowed requests per second per application. | `50`    | no       |
+| `per_app_burst_size` | `number` | Allowed burst size of requests per application.      | `100`   | no       |
```
ptodev: WDYT about retaining the original 3 attributes and make only one new attribute called `strategy` that could default to `global`, but it could also be set to `per_app`? It wouldn't make sense to have both types of limit enabled, and it looks like the implementation in the PR will just ignore the global limits if the per app limits are enabled?

I also wonder if this could be made even more configurable by configuring which part of the json payload we could load balance by, so that users can choose something other than app name. But maybe that's overkill and not needed in practice?

hhertout: You’re definitely right! That makes sense and reduces ambiguity for the user in the configuration file.
Following your recommendation, I imagine something like this:

```
faro.receiver "default" {
    server {
        rate_limiting {
            strategy = “per_application” # default stay to “global” to avoid breaking change
            per_app_rate = 50
            per_app_burst_size = 100
        }
    }
    
    output {
        logs = [loki.write.default.receiver]
    }
}  
```

Are we aligned on this?

That's true, I agree it doesn’t make sense to enable both, which is why I chose to ignore the global limits when the per-application limits are enabled.

Regarding the configuration of specific attributes that a user can set, I also had this idea in mind. However, I believe the main filter should be based on the “mandatory” labels defined by the Faro SDK, which are also relevant for the segregation in the limits.
From my point of view, and based on my experience with Faro in a centralized environment, it would be interesting to add more fine-tuning capabilities for the limits in the future. Hence for now, I think the first layer of filtering should be based on the app/env name, which is why I’ve focused on this for the initial PR.
Moreover, I believe that enabling it is sufficient to solve most of the issues.

Thank you for your review ! I’ll make the changes accordingly to your recommendations !

internal/component/faro/receiver/handler.go
```diff
@@ -99,11 +109,41 @@ func (h *handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	} else {
+		if !h.rateLimiter.Allow() {
+			http.Error(rw, http.StatusText(http.StatusTooManyRequests), http.StatusTooManyRequests)
+			return
+		}
+
+		if err := json.NewDecoder(req.Body).Decode(&p); err != nil {
+			http.Error(rw, err.Error(), http.StatusBadRequest)
+			return
+		}
```
blewis12: Feel free to address this once you've resolved the previous comments related to the structure and naming of the arguments, as there may be some conflict :) but regarding this particular block, I feel we could avoid two things:

1. Doing a nil check against the rate limiter instances, when instead we could check against the strategy name (which also helps us if we want to introduce more strategies in the future)
2. Having nested if/else statements

I feel we could separate the rate limiting checks into a separate method, and instead of doing a nil comparison we could use a switch statement. Something like this:

```
func (h *handler) handleRequest(rw http.ResponseWriter, req *http.Request) {
	var p payload.Payload

	// TODO: can we cleanly include this within isWithinRateLimits?
	if err := json.NewDecoder(req.Body).Decode(&p); err != nil {
		http.Error(rw, err.Error(), http.StatusBadRequest)
		return
	}

	if !h.isWithinRateLimits(p) {
		http.Error(rw, http.StatusText(http.StatusTooManyRequests), http.StatusTooManyRequests)
		return
	}

	h.processRequest(rw, req, p)
}

func (h *handler) isWithinRateLimits(p payload.Payload) bool {
	if !h.args.RateLimiting.Enabled {
		return true
	}

	switch h.args.RateLimiting.Strategy {
	case RateLimitingStrategyGlobal:
		return h.rateLimiter.Allow()
	case RateLimitingStrategyPerApp:
		app, env := h.extractAppEnv(p)
		allowed := h.appRateLimiter.Allow(app, env)
		if !allowed {
			level.Debug(h.log).Log(
				"msg", "per-app rate limit exceeded",
				"app", app,
				"env", env,
			)
		}
		return allowed
	default:
		return true
	}
}
```

wdyt?

internal/component/faro/receiver/handler.go
```diff
@@ -143,3 +177,20 @@ func (h *handler) handleRequest(rw http.ResponseWriter, req *http.Request) {
...
 }
+
+// extractAppEnv extracts the app and environment from the Faro payload metadata.
+// Returns "unknown" for missing values to ensure isolation.
+func (h *handler) extractAppEnv(p payload.Payload) (string, string) {
+	app := "unknown"
+	env := "unknown"
+
+	if p.Meta.App.Name != "" {
+		app = p.Meta.App.Name
```
blewis12, line 207: If there are multiple apps that are not setting `p.Meta` for whatever reason, these apps will be grouped under `unknown` and rate limited as one - is that intended?

Could we add a test for empty values? so the case that `p.Meta` is entirely unset, just so we can test that flow

internal/component/faro/receiver/handler.go
hhertout: Yes you're right.

I was trying to optimize the process by avoiding unnecessary JSON decode if not needed... but it adds too much complexity... and also impact the readability. My bad !
I've simplified it as you suggested, the function now always decodes first, then checks rate limits.

To answer your question: // TODO: can we cleanly include this within isWithinRateLimits?

I don't think we can. The main issue is that `isWithinRateLimits()` would need to handle JSON decoding, which mixes two different responsibilities and we'd fall into the same situation as before. It would require passing `req.Body` and `payload.Payload` as reference and should returning `(bool, error)`, and at the end of the day it doen't help so much... I tried a few approaches but couldn't find a way to avoid this do it cleanly.

internal/component/faro/receiver/handler.go
hhertout: Correct me if I'm wrong, but here's what I'm aiming for:

If an app doesn't provide metadata, it falls under the "unknown:unknown" key, meaning all apps without metadata share the same rate limiter (like strategy set to 'global'). However, if they properly set their metadata attributes, they get isolated with their own dedicated rate limiter.

By doing this, I believe it discourages bad practices and encourages users to properly configure their metadata if they want proper isolation when sharing infrastructure with other apps.

I'll add a test to verify this behavior with empty metadata values.

hhertout: I've added three new tests to cover the empty metadata cases:

1. `TestHandler_RateLimitingPerApp_EmptyMetadata`: Tests different scenarios where metadata is missing or empty (entirely missing meta, empty app name, empty environment, both empty). Each case verifies that requests are properly rate limited.

2. `TestHandler_RateLimitingPerApp_MultipleAppsWithoutMetadata`: Verifies that multiple apps without metadata share the same rate limiter under the `unknown:unknown` key.

3. `TestHandler_RateLimitingPerApp_ProperMetadataNotAffectedByUnknown`: Ensures that apps with proper metadata are isolated and not affected by apps without metadata exhausting the `unknown:unknown` quota.

blewis12: That makes sense to me as a strategy 👍 I think it would be best to add this to the documentation so that users are aware that all apps that are not configuring their metadata will be grouped under the same rate limiting group. Other than that, thanks for adding tests!

internal/component/faro/receiver/handler.go
blewis12: > To answer your question: // TODO: can we cleanly include this within isWithinRateLimits? I don't think we can.

Agreed, I previously thought this might change behavior but now I see that before we are already attempting to decode the payload regardless of rate limiting enablement. LGTM 👍

internal/component/faro/receiver/rate_limiter.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// ParseAppRateLimitingConfigKey creates a key from app and environment values.
+func ParseAppRateLimitingConfigKey(app, env string) AppRateLimitingConfigKey {
+	return AppRateLimitingConfigKey(fmt.Sprintf("%s:%s", app, env))
+}
+
+// AppRateLimitingConfig manages rate limiters per application/environment combination.
+type AppRateLimitingConfig struct {
+	pool  map[AppRateLimitingConfigKey]*rate.Limiter
```
blewis12: I'm wondering if we wind up in a situation where an alloy instance remains running for a while, and the applications it monitors come and go - we'd wind up with an ever growing map of key -> limiter.

Could we have some sort of eviction policy here if an app/env combo has not been seen in X amount of time?

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -115,15 +115,29 @@ If the `api_key` argument is empty, no authentication checks are performed, and
...
 
 Rate limiting functions as a [token bucket algorithm][token-bucket], where a bucket has a maximum capacity for up to `burst_size` requests and refills at a rate of `rate` per second.
 
-Each HTTP request drains the capacity of the bucket by one. After the bucket is empty, HTTP requests are rejected with an `HTTP 429 Too Many Requests` status code until the bucket has more available capacity.
+Each HTTP request drains the capacity of the bucket by one. 
+
+After the bucket is empty, HTTP requests are rejected with an `HTTP 429 Too Many Requests` status code until the bucket has more available capacity.
+
+When `strategy` is `per_app`, rate limiting is applied per application and environment combination instead of globally.
+
```
blewis12, line 137: Can we add an extra sentence clarifying that any application not exposing metadata will be grouped under the `unknown` rate limited app?

hhertout: Sure !

internal/component/faro/receiver/rate_limiter.go
hhertout: Makes sense! I'll implement this. However, I'm not sure what the best eviction timeout (X) should be. Do you have a recommendation? I think something like 10min could be ok.

I'm also wondering if it would be valuable to add observability metrics as well. We could expose metrics like the number of Allow() success/failure, the pool size, or the available token number. This would give alloy administrators some visibility around the rate limiter configuration. WDYT?

hhertout: I've implemented internal monitoring with Prometheus metrics that I believe will be useful for observability.

I identified a potential issue where we could flood the pool by sending requests with unlimited unique app/env combinations, leading to memory exhaustion. To address this, I've added a self-rate limiter that restricts the creation of new rate limiters. The protection uses the same `rate` and `burst` parameters configured by the user, so it scales naturally with the expected traffic volume.

Let me know if you have any concerns with this approach!

blewis12: My idea here would be to wrap our rate limiters in a struct, like

```
type AppLimiter struct {
    Limiter     *rate.Limiter
    LastAccess  time.Time
}
```

whenever we get a new request from that app we can do

```
lim.LastAccess = time.Now()
```

and have a periodic cleanup job that runs every 10 minutes as you say and checks for app identifiers that have not have a request sent in 30 minutes. I reckon that could work as a start, it's important we document this as well if you're in agreement there

What do you think of that approach?

And very much in agreement with exposing meta metrics! It would help with debugging any potential errors and ensure that users understanding how this rate limiter is behaving

internal/component/faro/receiver/rate_limiter.go
```diff
@@ -141,6 +146,11 @@ func (r *AppRateLimitingConfig) Allow(app, env string) bool {
 	limiter, exists := r.GetPoolLimiter(key)
 
 	if !exists {
+		// Self rate limiting to avoid memory overload on app creation
+		if !r.selfLimiter.Allow() {
+			return false
+		}
```
blewis12: hm i'm not so sure about this - it means that if a bunch of new apps spin up to the point that this rate limiter is activated, then new apps will effectively be blocked - despite they themselves not sending enough requests to be rate limited. It feels almost like falling back to global rate limiting, which isn't super intuitive 🤔

if anything I feel like the eviction policy we talked about before should handle us piling too many newly created app-rate-limiters in memory. If a user has so many apps that we're seeing issues with memory due to so many entries in the map, then that's another issue - but i'd say unrelated to a self rate limiter

internal/component/faro/receiver/rate_limiter.go
hhertout: We are aligned on how to implement it I suppose, this is what I did here :

```go
// rate_limiter.go:25

type AppRateLimiter struct {
    limiter  *rate.Limiter
    lastUsed time.Time  // Same as your LastAccess
}
```
I've just set up a 10-minute job to clean up apps that haven't received any requests in the last 10 minutes. But I can change it if you prefer! Let me know if there's anything wrong with the implementation I made.

I have one more question: when you say "it's important we document this as well if you're in agreement there", do you mean in the alloy documentation or just in the codebase?
In my opinion, the cleanup mechanism is seamless and transparent to the end user (it happens automatically in the background without any configuration or impact on behavior), so I'm not sure it's relevant to add it to the user documentation. However, I've documented it in the code.

internal/component/faro/receiver/rate_limiter.go
hhertout: You're right, I see the issue now. 

The `selfLimiter` only triggers when creating a new app entry, not for existing apps. So it's not exactly a global fallback in my opinion. Initially, in my mind the priority would be on protecting the gateway infrastructure in case a large number of new apps arrive concurrently in a short time window. That's why I based the `selfLimiter` on the user's configured rate limiting parameters, so it would scale up accordingly with their expected traffic (higher configured rate = more tolerance for new app creation).

However, I fully understand your point now, it's counter-intuitive to set up per-app rate limiting for segregation, only to end up being rate limited at pool entry creation before even sending any requests. The user experience would be very confusing... I'm aligned with your thoughts.

It adds unnecessary complexity, sorry for the misunderstanding. I removed the `selfLimiter` implementation and rely on the cleanup mechanism instead.

Thanks for the feedback, it really helped clarify the design!

internal/component/faro/receiver/rate_limiter.go
blewis12: ah! I got confused when looking at the code and missed that you already basically had what I suggested 🤦 then yes, that sounds good to me 

I see your point documentation-wise as well, the more I think about it the more I don't think it's necessary either. If it causes any confusion down the line we can address that then - but for now documenting via code works well

blewis12:
LGTM - i've pinged folks from the faro side to see if we can get their eyes on this to make sure i'm not missing something. Thanks @hhertout for your contribution!

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -115,15 +115,29 @@ If the `api_key` argument is empty, no authentication checks are performed, and
...
+This prevents one application from affecting the rate limits of other applications sending their data on the same gateway.
+
+The application and environment are extracted from the Faro payload metadata using the `app.name` and `app.environment` fields. If fields are missing, the rate limiting falls back to a global strategy for that request and shares the rate limiter with other applications.
```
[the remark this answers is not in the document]
blewis12: Arguably this is more clear - though I think we can remove that last sentence and leave it as "The application and environment are extracted from the Faro payload metadata using the `app.name` and `app.environment` fields. If these fields are missing, the request shares a rate limiter with other requests that also lack this metadata, using an `unknown:unknown` key" - though @clayton-cornell your opinion would be valuable there

internal/component/faro/receiver/handler.go
```diff
@@ -62,12 +65,21 @@ func (h *handler) Update(args ServerArguments) {
 		// burst to be filled.
 		t := time.Now().Add(-time.Duration(float64(time.Second) * args.RateLimiting.Rate * args.RateLimiting.BurstSize))
 
+		// Always set the global rate limiter for fallback protection
 		h.rateLimiter.SetLimitAt(t, rate.Limit(args.RateLimiting.Rate))
 		h.rateLimiter.SetBurstAt(t, int(args.RateLimiting.BurstSize))
+
+		// Initialize or update the per-app rate limiter if strategy is per_app
+		if args.RateLimiting.Strategy == RateLimitingStrategyPerApp {
+			h.appRateLimiter = NewAppRateLimitingConfig(h.args.RateLimiting.Rate, int(h.args.RateLimiting.BurstSize), h.reg)
```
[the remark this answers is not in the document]
blewis12: Ah yes. I would prefer to check for both `args.RateLimiting.Strategy` and `h.appRateLimiter != nil` in this case - it feels a bit more clear than just checking for `nil`. @hhertout what do you think of adding that?

internal/component/faro/receiver/receiver.go
```diff
@@ -154,6 +154,11 @@ func (c *Component) Update(args component.Arguments) error {
 		)
 		c.argsMut.RUnlock()
 
+		// Start cleanup routine if per-app rate limiting is enabled
+		if c.handler.appRateLimiter != nil {
+			go c.handler.appRateLimiter.CleanupRoutine(ctx, DEFAULT_CLEANUP_INTERVAL)
+		}
```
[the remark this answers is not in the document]
blewis12: hmm ok, so the issue is that the actor context could be cancelled before the cleanup goroutine finishes. Basically we'd have a situation where [cancelCurrentActor](https://github.com/grafana/alloy/blob/119402ddcf526755ea53617c05696b58bd855904/internal/component/faro/receiver/receiver.go#L106) is called, but then it's a race between our cleanup function returning and our actor function returning. 

So we'd need to ensure that the actor function return waits the cleanup goroutine to terminate properly. We could implement that with a wg - like

```var cleanupWg sync.WaitGroup
		if c.handler.appRateLimiter != nil {
			cleanupWg.Add(1)
			go func() {
				defer cleanupWg.Done()
				c.handler.appRateLimiter.CleanupRoutine(ctx, DEFAULT_CLEANUP_INTERVAL)
			}()
		}


...

cleanupWg.Wait()
}
```

internal/component/faro/receiver/handler.go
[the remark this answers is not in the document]
hhertout: Fair enough ! I made the change.
I added a test to ensure it behave correctly.

docs/sources/reference/components/faro/faro.receiver.md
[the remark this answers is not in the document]
hhertout: If I may give my opinion on this suggestion, from my user point of view, "using an `unknown:unknown` key" might be omitted, as the name of the key used isn't relevant. 
For the rest of the sentence, I'm completely aligned.

internal/component/faro/receiver/receiver.go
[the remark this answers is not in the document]
hhertout: I made the change. I didn't noticed that, my bad.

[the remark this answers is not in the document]
blewis12: neither did I :) cheers

internal/component/faro/receiver/receiver.go
```diff
@@ -154,6 +154,15 @@ func (c *Component) Update(args component.Arguments) error {
 		)
 		c.argsMut.RUnlock()
 
+		var cleanupWg sync.WaitGroup
+
+		// Start cleanup routine if per-app rate limiting is enabled
+		if c.handler.appRateLimiter != nil {
+			cleanupWg.Go(func() {
+				c.handler.appRateLimiter.CleanupRoutine(ctx, DEFAULT_CLEANUP_INTERVAL)
+			})
```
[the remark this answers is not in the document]
blewis12: not true, we use that in our codebase [already](https://github.com/grafana/alloy/blob/5abd05985f5fded6e0c196697846d9ecfe1ee40c/internal/component/loki/source/file/scheduler.go#L49)

internal/component/faro/receiver/handler.go
```diff
@@ -62,12 +65,21 @@ func (h *handler) Update(args ServerArguments) {
...
 
+		// Always set the global rate limiter for fallback protection
 		h.rateLimiter.SetLimitAt(t, rate.Limit(args.RateLimiting.Rate))
 		h.rateLimiter.SetBurstAt(t, int(args.RateLimiting.BurstSize))
+
+		// Initialize or update the per-app rate limiter if strategy is per_app
+		if args.RateLimiting.Strategy != RateLimitingStrategyPerApp {
+			h.appRateLimiter = nil
+		} else if h.appRateLimiter == nil {
+			h.appRateLimiter = NewAppRateLimitingConfig(h.args.RateLimiting.Rate, int(h.args.RateLimiting.BurstSize), h.reg)
```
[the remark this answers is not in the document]
blewis12: @hhertout We already have `h.args = args` above this block so I don't believe this should make a difference - but it's probably best practice to be reading from `args` itself and not the existing handler values

internal/component/faro/receiver/handler.go
```diff
@@ -99,11 +111,45 @@ func (h *handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	var p payload.Payload
+
+	if err := json.NewDecoder(req.Body).Decode(&p); err != nil {
+		http.Error(rw, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	if !h.isWithinRateLimits(p) {
 		http.Error(rw, http.StatusText(http.StatusTooManyRequests), http.StatusTooManyRequests)
 		return
```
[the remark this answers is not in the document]
blewis12: @hhertout This is indeed a change that affects current global rate limiting behavior, so we should address this. We need to ensure that the payload is valid before going into application rate limiting so that we don't panic when attempting to read metadata - but we should also check for global rate limits first to ensure those that have global rate limiting enabled don't decode rate limited payloads

Perhaps in the end we need to 

1. Check global rate limiting
2. Decode the payload
3. Check app rate limiting

internal/component/faro/receiver/rate_limiter.go
```diff
@@ -0,0 +1,171 @@
...
+// Allow checks if a request is allowed for the given app/environment combination.
+// Creates a new rate limiter if one doesn't exist for the app/env key.
+func (r *AppRateLimitingConfig) Allow(app, env string) bool {
+	key := ParseAppRateLimitingConfigKey(app, env)
+
+	limiter, exists := r.GetPoolLimiter(key)
+
+	if !exists {
+		// Create new rate limiter with pre-filled bucket (similar to handler.Update)
+		t := time.Now().Add(-time.Duration(float64(time.Second) * float64(r.rate) * float64(r.burst)))
```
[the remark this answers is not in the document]
blewis12: the change is not valid from looking at this locally

docs/sources/reference/components/faro/faro.receiver.md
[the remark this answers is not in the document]
blewis12: @hhertout agreed - feel free to add this and we can wait for Clayton to review that

internal/component/faro/receiver/rate_limiter.go
```diff
@@ -0,0 +1,171 @@
...
+		newLimiter := rate.NewLimiter(r.rate, r.burst)
+		newLimiter.SetLimitAt(t, r.rate)
+		newLimiter.SetBurstAt(t, r.burst)
+
+		limiter = r.SetPoolLimiter(key, newLimiter)
+	} else {
+		// Update last used time
+		r.mu.Lock()
+		r.pool[key].lastUsed = time.Now()
```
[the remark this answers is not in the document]
blewis12: @hhertout I believe this is pointing to the fact that our cleanup goroutine might clean up this key between us fetching the key at `GetPoolLimiter` (in which we grab the lock and then release it after retuning the key) and then attempting to access it later in this method. 

We could either wrap this entire method of reading and writing from the pool in a lock - but I feel for now that checking that the key has not been removed so that we don't cause a `panic` here would suffice. Let me know what you reckon

internal/component/faro/receiver/handler.go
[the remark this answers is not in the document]
hhertout: I found a way to improve the readability in the handler function, by creating function for each type of limiter.
Tell me if it looks good to you ?

internal/component/faro/receiver/rate_limiter.go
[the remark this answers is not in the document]
hhertout: I've modified the approach based on your feedback.

Initially, I considered keeping the RLock() in GetPoolLimiter() to avoid read bottlenecks and checking if the key still exists after acquiring the write lock:

```go
if appLimiter, exists := r.pool[key]; exists {
    appLimiter.lastUsed = time.Now()
}
```

However, in the case the limiter doesn't exist, this would require duplicating the limiter creation logic in the else branch, not ideal...

After reviewing the code, for now GetPoolLimiter() is currently only called from Allow(), which already needs to acquire a write lock in all cases (either to create a new limiter or update. So I simplified the implementation by acquiring the write lock at the start of Allow(), making the entire method thread-safe without the complexity of double-checking.

Deep down I reckon that it has some benefits:
It allow to use the function GetPoolLimiter() with the read lock but also with the read/write lock as it is no longer defined in the body of the function (even for the SetLimiter btw).
The lock protects the entire operation in the Allow method.

The only potential concern is a slight bottleneck during cleanup (every 10 minutes) if there is others function using the GetPoolLimiter() in the future, but given the infrequent cleanup interval, the impact should be minimal.

WDYT ?

docs/sources/reference/components/faro/faro.receiver.md
```diff
@@ -212,6 +226,9 @@ The template value is replaced with the release value provided by the [Faro Web
 * `faro_receiver_sourcemap_cache_size` (counter): Number of items in sourcemap cache per origin.
 * `faro_receiver_sourcemap_downloads_total` (counter): Total number of sourcemap downloads performed per origin and status.
 * `faro_receiver_sourcemap_file_reads_total` (counter): Total number of sourcemap retrievals using the filesystem per origin and status.
+* `faro_receiver_rate_limiter_active_app` (gauge): Number of active applications with rate limiters. Inactive limiters are cleaned up every 10 minutes.
+* `faro_receiver_rate_limiter_requests_total` (counter): Total number of requests processed by the rate limiter per app/environment.
+
```
clayton-cornell, line 231:
```suggestion
* `faro_receiver_rate_limiter_requests_total` (counter): Total number of requests processed by the rate limiter per app/environment.
```

Just removing extra blank line

internal/component/faro/receiver/rate_limiter.go
```diff
@@ -0,0 +1,174 @@
...
+
+	r.mu.Unlock()
+
+	allowed := limiter.Allow()
+	r.rateLimitDecisions.WithLabelValues(app, env, fmt.Sprintf("%t", allowed)).Inc()
+	return allowed
+}
+
+func (r *AppRateLimitingConfig) getNewLimiter() *rate.Limiter {
+	t := time.Now().Add(-time.Duration(float64(time.Second) * float64(r.rate) * float64(r.burst)))
```
mar4uk, line 172: @hhertout could you please explain a bit why time is calculated like this? why not just time.Now()? Maybe adding a comment above would be great

internal/component/faro/receiver/rate_limiter_test.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+func TestNewAppRateLimitingConfig(t *testing.T) {
+	tests := []struct {
+		name      string
+		rateLimit float64
+		burst     int
+	}{
+		{
+			name: "standard rate", rateLimit: 100.0,
```
mar4uk: nit: rateLimit should be next line same as other tests

internal/component/faro/receiver/rate_limiter_test.go
```diff
@@ -0,0 +1,432 @@
...
+	assert.True(t, exists, "app3:dev should still exist")
+}
+
+func TestAppRateLimitingConfig_CleanupRoutine_RemovesOldLimiters(t *testing.T) {
+	// Use very short intervals for faster test
+	originalExpiry := DEFAULT_LIMITER_EXPIRY
+	defer func() {
+		// Note: Can't actually reset the const, but documenting intent
+		_ = originalExpiry
+	}()
```
mar4uk: does this code affect the test? why does it needed?

internal/component/faro/receiver/rate_limiter.go
mar4uk: probably the comment should be the same as handler has https://github.com/grafana/alloy/blob/main/internal/component/faro/receiver/handler.go#L59-L62

internal/component/faro/receiver/handler.go
```diff
@@ -143,3 +197,20 @@ func (h *handler) handleRequest(rw http.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(http.StatusAccepted)
 	_, _ = rw.Write([]byte("ok"))
 }
+
+// extractAppEnv extracts the app and environment from the Faro payload metadata.
+// Returns "unknown" for missing values to ensure isolation.
+func (h *handler) extractAppEnv(p payload.Payload) (string, string) {
+	app := "unknown"
+	env := "unknown"
```
mar4uk: what if customer names his app "unknown"? Basically unknown is valid name for an app... As well as environmane could be set to "unknown" as well. Then his app will be rate limited by the same limiter as all other apps where name and/or environment are not set. Shall we use empty string instead of "unknown" to handle this case?

hhertout: You're right. Could we imagine something like "undefined:undefined" or just ":" in the case that both app/env are missing ?

Setting "" as the fallback value could avoid any confusion in the end, so I think it might be the best option.

I tried to find another example to know what fallback value I could have used, but I didn’t find any. For example, `loki.process` doesn’t apply the rate limit if the key is missing.
https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/limit.go#L106

In the end, I trust your opinion, let me know what you think I must set in this case. You're definitely in a better position than I am to decide which key makes the most sense.

internal/component/faro/receiver/rate_limiter_test.go
hhertout: I forgot to remove it after a previous attempt.
My mistake sorry for that.

internal/component/faro/receiver/handler.go
mar4uk: I would set empty strings in case app or environment are not set. It also aligns with the logic in rate_limiter: it works with keys `"":""`, `"myapp":""`, `"":"production"`

hhertout: Alright! Thanks a lot for your advice, I’ve changed it.

mar4uk:
Looks good! 👍  Thank you for your contribution!

PR #4690, opened 2025-10-27
Use labels.Labels API For String Label Compatibility Part 2

blewis12:
#### PR Description
A follow up to [this PR](https://github.com/grafana/alloy/pull/4455) - there are a bunch of place I missed before when compiling against the stringlabels tag. This PR doesn't update Loki, which would be the final step, but this is being done separately in the prometheus upgrade. 

This PR also bumps to a more recent commit of walqueue that contains [these changes ](https://github.com/grafana/walqueue/pull/56)

doesn't fix yet, but part of https://github.com/grafana/alloy/issues/4323

#### PR Checklist
- [ ] CHANGELOG.md updated // Question to reviewer: should I be updating the changelog here? The change is not user facing so I assume not, but want to be sure

### DIFF
--- a/internal/component/prometheus/remotewrite/types.go
+++ b/internal/component/prometheus/remotewrite/types.go
@@ -265,14 +265,10 @@ func convertConfigs(cfg Arguments) (*config.Config, error) {
 
 	return &config.Config{
 		GlobalConfig: config.GlobalConfig{
-			ExternalLabels: toLabels(cfg.ExternalLabels),
+			ExternalLabels: labels.FromMap(cfg.ExternalLabels),
 		},
 		RemoteWriteConfigs: rwConfigs,
 	}, nil
-}
-
-func toLabels(in map[string]string) labels.Labels {
-	return labels.FromMap(in)
 }
 
 // ManagedIdentityConfig is used to store managed identity config values


internal/component/prometheus/remotewrite/types.go
```diff
@@ -273,12 +272,7 @@ func convertConfigs(cfg Arguments) (*config.Config, error) {
 }
 
 func toLabels(in map[string]string) labels.Labels {
-	res := make(labels.Labels, 0, len(in))
-	for k, v := range in {
-		res = append(res, labels.Label{Name: k, Value: v})
-	}
-	sort.Sort(res)
-	return res
+	return labels.FromMap(in)
```
thampiotr: Nit: I think we can get rid of this helper function.

internal/component/pyroscope/scrape/target.go
```diff
@@ -61,30 +60,40 @@ type Target struct {
...
+					lblsBuilder.Set(l.Name, pprofMutex)
 					godeltaprof = true
 				}
 			}
-			continue
+			return
 		}
-		publicLabels = lbls[i:]
-		break
-	}
```
thampiotr, line 87: Seems like it was taking a shortcut here. If there is some benchmark, maybe we should run it? But I'd suspect this won't make a big difference... and performance will soon change again when we switch to stringlabels.

thampiotr:
> Question to reviewer: should I be updating the changelog here? The change is not user facing so I assume not, but want to be sure

I don't think we need changelog update here.

thampiotr (approved):
LGTM! feel free to fix the nits or resolve the comments

internal/component/pyroscope/scrape/target.go
blewis12: yeah that's what I thought, but since we're talking about pyroscope scrape targets I figured it wouldn't make the biggest difference. that said it's definitely worth benchmarking, i'll have another look here just to be sure

internal/component/prometheus/remotewrite/types.go
blewis12: true 👍 will do

internal/component/pyroscope/scrape/target.go
blewis12: From what I can see:

1. This code is called when we `reload` a component, aka not a hot path from what I understand
2. We are looping through all public labels instead of `lbls[i:]` which is more inefficient, but adding to the unsorted builder is a linear O(1) operation. We have to sort public labels before retrieving them from the builder, but that occurs just once
3. In practice, I don't think the number of Periskop labels would be high enough that it would cause significant impact (this is my newcomer opinion though, it could be wrong)

We have a benchmark test for this, `BenchmarkPopulateLabels`, running that I can see that we do indeed have a regression in memory allocation due to the builder. I would think though that with stringlabels enabled this would offset that slight increase

For now I will leave this in - but noting in case we happen to see any significant regressions

PR #4703, opened 2025-10-28
(code pool - do not merge) database_observability: consider idle states to calculate query sample times 

gaantunes:
#### PR Description

Including idle states processing, encompassing two scenarios:
- Previously captured query actively running becomes idle, emit the sample calculating the duration from the stateChange coming from the idle state row.
- Not yet captured query shows up as idle, emit a one-off sample with the information available in the idle row.

Given one-off idle samples generate a lot of entries, I am adding a throttling mechanism with a default of 30s per queryid. Exceptions are samples in `Idle in Transaction` and `Idle in Transaction (Aborted)` state, containing wait events and cpu processing time info.

Also adding a fix to the schema details tests which are intermittently failing.

~~Remove the state field from samples as it does not add much valuable information. In case of `idle in transaction (aborted)`, report a new `finished_with_error` field.~~
State is actually important to keep, as `idle in transaction` indicates long running transactions and `idle in transaction (aborted)` errored transactions, so I am adding it back.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -258,22 +265,45 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
+		} else if _, already := c.emitted[key]; !already {
+			dummy := &SampleState{LastRow: sample, LastSeenAt: sample.Now, tracker: newWaitEventTracker()}
+			if sample.StateChange.Valid {
+				dummy.EndOverride = sample.StateChange
+				dummy.LastSeenAt = sample.StateChange.Time
+			}
+			c.samples[key] = dummy
+		}
+		// refresh emitted last seen for idle keys (dedupe & TTL)
+		c.emitted[key] = sample.Now
```
cristiangreco: Hmm probably this update needs to happen within the two IFs above. If we update unconditionally, there's a risk of edge case where an entry is idle in every single scrape and we continue refreshing the timestamp and thus never deleteing it from `c.emitted`. Does this resonate?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -258,22 +265,45 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
-	for key := range c.samples {
-		if _, stillActive := activeKeys[key]; stillActive {
+	// finalize samples that are no longer active or have EndOverride set (idle finalized or one off idle sample)
+	for key, st := range c.samples {
+		if _, stillActive := activeKeys[key]; stillActive && !st.EndOverride.Valid {
 			continue
 		}
 		c.emitAndDeleteSample(key)
 	}
+	c.cleanupEmitted(time.Now())
```
cristiangreco, line 401: We're mixing time source here. Might be safer to use the latest timestamp from scrape?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -258,22 +265,45 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
+			activeKeys[key] = struct{}{}
+			continue
+		}
+
+		if st, hadActive := c.samples[key]; hadActive {
+			end := sample.StateChange
+			if end.Valid {
+				st.EndOverride = end
+				st.LastSeenAt = end.Time
+			}
```
cristiangreco: What happens if `end` is not `Valid`? I think that with the new logic on line 301 there's a risk that we track the sample as active forever. Maybe `c.samples` need an explicit cleanup every X minutes like `c.emitted`?

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Yes good catch, placing it inside the ifs.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: I have placed a fallback case to use clock_timestamp (`now()`) if state_change is not available.

Active rows will either turn into `idle`, `idle in transaction (aborted)`, `idle in transaction` or just disappear because it was replaced (in which case it is emitted).
First 2 cases are covered in the ifs below. A session in an `idle in transaction` state will gain a new query_id when it leaves this state, so it will be emitted.

I think there is no case where a sample lives forever.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: We use Time.Now to compare with the lastSeen, if it was emitted 10 minutes ago we clean it up.

cristiangreco: Hmm what I meant is we should probably use `sample.Now` instead of `time.Now()`?

Anyway for the cleanup it might be better to have autoexpiry (like with the `expirable.LRU` map we use elsewhere in the collector) rather than relying on timestamps diff. But ignore this for now, I think we can do it in a followup. Let's get the collection logic right first.

PR #4706, opened 2025-10-28
database_observability: Pass queries for explain plans more permissively.

rgeyer:
#### PR Description
In fact, pass everything, besides what is on the denylist which (hopefully) covers all of the possible write reserved words

Fixes grafana/grafana-dbo11y-app#1734

#### Notes to the Reviewer
Take a look at the denylist. I'm reasonably confident in it, but I am not an absolute expert on either MySQL or Postgres SQL.

~I could not find a satisfactory solution with a lexer, or asking the db engine for metadata about the created stored procedure. Everything I could find pretty much implied that "you should know what the query does already".~

Edit: A lexer was ultimately the correct choice, if only because it gave us more ability to examine each token, including it's type. In retrospect, I have no idea why I thought a "contains" would work without more examination. My mistake!

The good news is, because we are not doing `EXPLAIN ANALYZE` nothing is actually executed, so even if some write queries slip by, it will not have any impact, and we can hopefully improve detection over time.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -568,7 +568,17 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 			continue
 		}
 
-		if !strings.HasPrefix(strings.ToLower(qi.queryText), "select") {
+		uppercaseQueryText := strings.ToUpper(qi.queryText)
+		containsReservedWord := false
+		for _, word := range database_observability.ExplainReservedWordDenyList {
+			if strings.Contains(uppercaseQueryText, word) {
```
cristiangreco: I'm wondering if it makes sense to use "starts with" rather than "contains"? Some of the reserved words listed above could appear in the middle of a select (e.g. as column name) and still be legit. wdyt?

internal/component/database_observability/explain_plan_output.go
```diff
@@ -35,6 +35,39 @@ const (
 	ExplainPlanJoinAlgorithmNestedLoop ExplainPlanJoinAlgorithm = "nested_loop"
 )
 
+// ExplainReservedWordDenyList contains SQL reserved words that indicate write operations
+// to the database. These are primarily DML (Data Manipulation Language) and DDL
+// (Data Definition Language) commands that modify database state.
+// This was extracted from the MySQL and PostgreSQL documentation by Claude Sonnet 4 on Oct 28, 2025
+// and audited by @rgeyer and others in the dbo11y team.
+var ExplainReservedWordDenyList = []string{
```
cristiangreco: Should "EXPLAIN" itself be listed here?

internal/component/database_observability/mysql/collector/explain_plans.go
rgeyer: The main use-case that we're aware of, are queries beginning in `WITH`. Unfortunately, any query could follow a `WITH` including a destructive or write query.

While queries *could* contain a reserved word, it would have to be single quoted to reference a column name or other keyword. Or be double quoted if it were in a blob of text. This is where a lexer *might* give us more confidence, but at a non-trivial cost of complexity and additional processing of each query. 🤔 

Having a shared query analysis component that publishes results to downstream components may be useful in the future. I'm attempting to balance complexity here rather than find the ideal cure-all, but that may be a flawed strategy.

internal/component/database_observability/explain_plan_output.go
rgeyer: It absolutely should! I'll add it, thanks. :)

internal/component/database_observability/mysql/collector/explain_plans.go
rgeyer: Actually... Upon closer inspection, you're right in a way.. Even very simple queries in our dev environment are being rejected because they contain these keywords as part of other sequences.

I'll have to actually invoke a lexer and better pick apart the query. 😩

rgeyer: Using a lexer now, including tests to assure that we do not drop valid read-only queries. 😄

internal/component/database_observability/explain_plan_output.go
```diff
@@ -35,6 +35,42 @@ const (
...
+
+	// Index operations
+	"REINDEX", "ANALYZE", "OPTIMIZE",
+
+	// User/Permission management
+	"GRANT", "REVOKE", "CREATE USER", "DROP USER", "ALTER USER",
+
+	// MySQL specific write operations
+	"LOAD", "REPLACE", "DELAYED", "IGNORE", "ON DUPLICATE KEY",
+	"LOW_PRIORITY", "HIGH_PRIORITY", "QUICK", "LOCK", "UNLOCK",
```
gaantunes: I believe in older versions of MySQL one could use the `LOCK IN SHARE MODE`. We are targeting only more recent MySQL versions though, so I believe it is fine.

internal/component/database_observability/explain_plan_output.go
```diff
@@ -35,6 +35,42 @@ const (
...
 )
 
+// ExplainReservedWordDenyList contains SQL reserved words that indicate write operations
+// to the database. These are primarily DML (Data Manipulation Language) and DDL
+// (Data Definition Language) commands that modify database state.
+// This was extracted from the MySQL and PostgreSQL documentation by Claude Sonnet 4 on Oct 28, 2025
+// and audited by @rgeyer and others in the dbo11y team.
+var ExplainReservedWordDenyList = []string{
+	// Data Manipulation Language (DML) - Write operations
+	"INSERT", "UPDATE", "DELETE", "REPLACE", "MERGE", "UPSERT",
```
gaantunes: Will this exclude the `FOR UPDATE` clause used in selects to lock the data?

internal/component/database_observability/lexer.go
```diff
@@ -0,0 +1,69 @@
...
+// that indicate write operations, excluding those in string literals or comments
+func ContainsReservedKeywords(query string, reservedWords []string) bool {
+	// Create a map for faster lookup
+	reservedMap := make(map[string]bool)
+	for _, word := range reservedWords {
+		reservedMap[strings.ToUpper(word)] = true
+	}
+
+	// Use the lexer to tokenize the query
+	lexer := sqllexer.New(query, sqllexer.WithDBMS(sqllexer.DBMSPostgres))
```
cristiangreco: `sqllexer.DBMSPostgres` -> this should probably a param if we're using this method for both mysql and postgres.

internal/component/database_observability/lexer.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// ContainsReservedKeywords checks if the SQL query contains any reserved keywords
+// that indicate write operations, excluding those in string literals or comments
+func ContainsReservedKeywords(query string, reservedWords []string) bool {
+	// Create a map for faster lookup
+	reservedMap := make(map[string]bool)
+	for _, word := range reservedWords {
+		reservedMap[strings.ToUpper(word)] = true
+	}
```
cristiangreco: Might be worth doing a static init of a map given how frequently this method is called?

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -417,7 +417,10 @@ func (c *ExplainPlan) fetchExplainPlans(ctx context.Context) error {
 			continue
 		}
 
-		if !strings.HasPrefix(strings.ToLower(qi.queryText), "select") {
+		containsReservedWord := database_observability.ContainsReservedKeywords(qi.queryText, database_observability.ExplainReservedWordDenyList)
+
+		if containsReservedWord {
+			level.Debug(logger).Log("msg", "skipping query containing reserved word", "query", qi.queryText)
```
cristiangreco: Could just the `queryid` be enough? Or if we want to log the query text explicitly maybe we should do the same on mysql?

internal/component/database_observability/explain_plan_output.go
rgeyer: Are you suggesting that a query with `LOCK`, or more specifically `LOCK IN SHARE MODE` is actually not a write query, and should be permitted?

I looked at the docs for `LOCK IN SHARE MODE` and it doesn't really "write" to the DB. It does lock the retrieved rows, so it potentially *blocks* writes for other queries. Not sure how that behaves when in "explain mode", it's possible that it doesn't actually lock anything.

Based on this, I'm inclined to remove `LOCK` and `UNLOCK` from this list.

internal/component/database_observability/explain_plan_output.go
rgeyer: Same analysis as the previous comment. I'm not sure we *need* to exclude those.

That said, the lexer *does* consider multi word commands, and I added this one, and `LOCK IN SHARE MODE`, including tests for these.

LMK if we think we ought to drop those tho, they seem harmless to me?

internal/component/database_observability/lexer.go
rgeyer: 🤦‍♂️ Of course.. Copy/paste oversight. Thank you for catching this!

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Reverted that for now. It was actually vestigial from some debug work I was doing locally.

Might revisit this in a follow-up PR, but it doesn't belong here.

internal/component/database_observability/explain_plan_output.go
gaantunes: Yes, what I meant was that LOCK can be used in SELECTS as well. I also think we should remove it from the denylist.

internal/component/database_observability/lexer_test.go
```diff
@@ -309,3 +311,177 @@ func TestPgSqlParser_ExtractTableNames(t *testing.T) {
...
+			query:    "SELECT * FROM users AS insert_table",
+			expected: false,
+		},
+		{
+			name:     "SELECT with legacy LOCK IN SHARE MODE",
+			query:    "SELECT name FROM users LOCK IN SHARE MODE",
+			expected: false,
+		},
+		{
+			name:     "SELECT with FOR UPDATE",
```
gaantunes, line 467: My understanding is this will be dropped? I think it shouldn't, if that is possible.
We can always iterate of course.

rgeyer: Yes, you're correct, this will currently be ignored. I'm looking into it now. I was under the impression from the tests that a compound command is properly identified, but that's not the case so.

For now, it may be best to continue to drop it, not because `FOR UPDATE` is a problem but because it contains `UPDATE` and we can not (easily) tell the difference.

gaantunes: Yes lets work iteratively.
The selects containing FOR UPDATE can easily cause contention, though. So adding these to the list is important as a next step.

gaantunes (approved):

Issue #3490, opened 2025-05-02
loki file reader tasks lives until target goes away when using decompression


kalleep:
### What's wrong?

When we are running lokie file reader task and have decompression enabled we never "un-schedule" the task even after we have finished reading the whole file.

I would expect the task to be removed after its done because I don't see the point in continue reading them when we are done.

This was the behavior before https://github.com/grafana/alloy/pull/2428

### Steps to reproduce

Run `loki.source.file` with decompression enabled.

### Configuration

```text

```

### Logs

```text

```

PR #4711, opened 2025-10-29
fix: use new scheduler for loki source file

kalleep:
#### PR Description
Before this we where using the shared [runner](https://github.com/grafana/alloy/blob/main/internal/runner/runner.go#L39) we have.

While this works okay it don't really suits the scheduling we need for `loki.source.file`. 

The runner Only have `ApplyTasks` that will accept a list of tasks and perform the book keeping for you around stopping tasks no longer in the list and starting new tasks. 

This forced us to perform the scheduling in `Run`. While this works it could still be to slow to pick up some targets resulting in `failed to tail file: stat` errors. 

I instead implemented a scheduler (that we can make more generic for upcoming work) that will not perform the book keeping on what to start and what to stop for you.

`ScheduleSource` will start provided source in a go routine.
`StopSource` will stop a running source without waiting for confirmation that it has stopped.
`Stop` will stop all running sources and wait for them to finish.

This gives us more flexibility and run sources as soon as we detect a new one. We will also stop running sources that we should no longer tail without blocking.

Fixes: https://github.com/grafana/alloy/issues/3490

#### Notes to the Reviewer

* We no longer log "error stopping tailer" when file is deleted
* We only log  "failed to run tailer" if it happens the first time we try. 
  *  tailer is wrapped with retries that only stops when source is no longer needed, so this could be a bit spammy. 
* I aslo added SourceWithRetry so we now only wrap tailers and not decompressors, they should run to completion and then stop
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/scheduler.go
+++ b/internal/component/loki/source/file/scheduler.go
@@ -8,8 +8,9 @@ import (
 	"github.com/grafana/dskit/backoff"
 )
 
-// Scheduler is used to manage sources.
-// So it is up to the users to ensure proper synchronization.
+// Scheduler manages the lifecycle of sources.
+// It is not safe for concurrent use: callers must ensure proper synchronization
+// when accessing or modifying Scheduler and its sources from multiple goroutines.
 type Scheduler[K comparable] struct {
 	ctx     context.Context
 	cancel  context.CancelFunc


internal/component/loki/source/file/scheduler.go
```diff
@@ -0,0 +1,138 @@
...
+		ctx:    ctx,
+		cancel: cancel,
+		source: source,
+	}
+
+	s.sources[k] = st
+
+	s.running.Go(func() {
+		st.source.Run(st.ctx)
+	})
```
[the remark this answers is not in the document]
kalleep: It does actually have a `Go` since 1.25 https://pkg.go.dev/sync#WaitGroup.Go

internal/component/loki/source/file/file.go
```diff
@@ -160,67 +152,29 @@ func (c *Component) Run(ctx context.Context) error {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case entry := <-c.handler.Chan():
+			c.receiversMut.RLock()
+			for _, receiver := range c.receivers {
+				select {
+				case <-ctx.Done():
+					return nil
```
thampiotr, line 171: I don't think it's mandatory, but just for the sake of clean code - can we unlock the c.receiversMut before returning?

internal/component/loki/source/file/file.go
```diff
@@ -279,71 +242,63 @@ func (c *Component) Update(args component.Arguments) error {
...
 	}
 
-	if len(newArgs.Targets) == 0 {
-		level.Debug(c.opts.Logger).Log("msg", "no files targets were passed, nothing will be tailed")
+	// Stop all sources that we no longer should consume.
+	for source := range c.scheduler.Sources() {
+		if _, ok := shouldRun[source.Key()]; ok {
+			continue
+		}
+		c.scheduler.StopSource(source)
```
thampiotr: Is this blocking?

internal/component/loki/source/file/file.go
```diff
@@ -370,36 +323,33 @@ func (c *Component) createReader(opts readerOptions) (reader, error) {
 			c.IsStopping,
 		)
 		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to create decompressor", "error", err, "filename", opts.path)
```
thampiotr, line 373: Are we sure we want to delete this log line? Will we get the filename printed somewhere else?

internal/component/loki/source/file/file.go
```diff
@@ -279,71 +242,63 @@ func (c *Component) Update(args component.Arguments) error {
...
-			tailFromEnd:         newArgs.TailFromEnd,
-			legacyPositionUsed:  newArgs.LegacyPositionsFile != "",
+			encoding:            args.Encoding,
+			decompressionConfig: args.DecompressionConfig,
+			fileWatch:           args.FileWatch,
+			tailFromEnd:         args.TailFromEnd,
+			legacyPositionUsed:  args.LegacyPositionsFile != "",
 		})
 		if err != nil {
+			level.Error(c.opts.Logger).Log("msg", "failed to create file source", "error", err, "filename", path)
```
thampiotr, line 255: Maybe we should be clear that these logs won't be read as a consequence.

internal/component/loki/source/file/file.go
```diff
@@ -370,36 +323,33 @@ func (c *Component) createReader(opts readerOptions) (reader, error) {
...
-			level.Error(c.opts.Logger).Log("msg", "failed to create decompressor", "error", err, "filename", opts.path)
-			return nil, fmt.Errorf("failed to create decompressor %s", err)
+			return nil, fmt.Errorf("failed to create decompressor %w", err)
 		}
-		reader = decompressor
-	} else {
-		pollOptions := watch.PollingFileWatcherOptions{
+		return decompressor, nil
+	}
+	tailer, err := newTailer(
```
thampiotr, line 335: Should we just pass opts?

internal/component/loki/source/file/scheduler.go
```diff
@@ -0,0 +1,138 @@
...
+	ctx, cancel := context.WithCancel(s.ctx)
+	st := scheduledSource[K]{
+		ctx:    ctx,
+		cancel: cancel,
+		source: source,
+	}
+
+	s.sources[k] = st
+
+	s.running.Go(func() {
```
thampiotr, line 49: Nice! new API :D

internal/component/loki/source/file/file.go
thampiotr:
```suggestion
		c.scheduler.StopSource(source) // stops without blocking
```

internal/component/loki/source/file/scheduler.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return &Scheduler[K]{
+		ctx:     ctx,
+		cancel:  cancel,
+		sources: make(map[K]scheduledSource[K]),
+	}
+}
+
+// ApplySource will register and start run provided source.
+// If a source with the same key already exists it will do nothing.
```
thampiotr:
```suggestion
// ApplySource will register and run the provided source in a goroutine.
// If a source with the same key already exists it will do nothing.
```
Just want to highlight that this is non-blocking too.

internal/component/loki/source/file/scheduler.go
```diff
@@ -0,0 +1,138 @@
...
+func (s *Scheduler[K]) ApplySource(source Source[K]) {
```
thampiotr: Can we rename `ApplySource`? I think that name is just inherited from the old runner. Maybe we call it `ScheduleSource`?

internal/component/loki/source/file/scheduler.go
```diff
@@ -0,0 +1,138 @@
...
+	k := source.Key()
+	scheduledTask, ok := s.sources[k]
+	if !ok {
+		return
+	}
+	delete(s.sources, k)
+	scheduledTask.cancel()
+}
+
+// Sources returns a iterator of all scheduled sources.
```
thampiotr:
```suggestion
// Sources returns an iterator of all scheduled sources.
```

internal/component/loki/source/file/file.go
```diff
@@ -279,71 +242,63 @@ func (c *Component) Update(args component.Arguments) error {
...
 func (c *Component) DebugInfo() any {
-	c.tasksMut.RLock()
-	defer c.tasksMut.RUnlock()
-	var res readerDebugInfo
-	for e, task := range c.tasks {
-		offset, _ := c.posFile.Get(e.Path, e.Labels)
+	c.schedulerMut.RLock()
+	defer c.schedulerMut.RUnlock()
+	var res debugInfo
+	for s := range c.scheduler.Sources() {
```
thampiotr, line 297: Did you consider making the `schedulerMut` a field of the `scheduler`.

The pros are I think that it will be easier to read, more encapsulated and less error-prone.

The cons are that it would mean you don't have a 'transaction' where you make a whole bunch of changes to the scheduler in one long operation, like during `Update`. This may be okay and actually even better, but it could also be not desired and lead to some race conditions.

CHANGELOG.md
```diff
@@ -60,6 +60,8 @@ Main (unreleased)
 
 - `prometheus.exporter.snowflake` dependency has been updated to 20251016132346-6d442402afb2, which updates data ownership queries to use `last_over_time` for a 24 hour period. (@dasomeone)
 
+- Schedule new path targets faster in `loki.source.file`. (@kalleep)
```
thampiotr: I'd say that it would be more of a bugfix where we fix the issue where there can be a big lag when scheduling / rescheduling a large number of files.

internal/component/loki/source/file/scheduler.go
kalleep: Yeah, just a convenient api added in go 1.25

internal/component/loki/source/file/file.go
kalleep: Ah yeah for sure. We did not have that before but I agree we should have it

internal/component/loki/source/file/file.go
kalleep: Yes these are now logged by
```
level.Error(c.opts.Logger).Log("msg", "failed to create file source", "error", err, "filename", path)
```

Before we logged here and returned an error and on caller side we had
```
if err != nil {
  continue
}
```
So I figured it was better to log in one place

internal/component/loki/source/file/file.go
kalleep: Yep lets do it

internal/component/loki/source/file/file.go
```diff
@@ -160,67 +152,30 @@ func (c *Component) Run(ctx context.Context) error {
 				}
 			}
 		}()
-
-		c.tasksMut.RLock()
+		c.schedulerMut.RLock()
```
thampiotr: the scheduler.Stop() is changing its inner state, so I think you need a write lock for this...

internal/component/loki/source/file/file.go
```diff
@@ -279,71 +243,63 @@ func (c *Component) Update(args component.Arguments) error {
...
 
-	if len(newArgs.Targets) == 0 {
-		level.Debug(c.opts.Logger).Log("msg", "no files targets were passed, nothing will be tailed")
+	// Stop all sources that we no longer should consume.
+	for source := range c.scheduler.Sources() {
+		if _, ok := shouldRun[source.Key()]; ok {
+			continue
+		}
+		c.scheduler.StopSource(source) // stops without blocking
 	}
```
thampiotr: Nit: Here we are iterating over the sources and deleting while iterating. It is safe because the scheduler uses a map internally, but if we ever changes it to a slice or something else, this would no longer be safe. I'm generally always suspicious when I see something mutating a collection while iterating over it. Maybe we can collect the keys to stop and then stop them in a separate loop.

internal/component/loki/source/file/file.go
```diff
@@ -353,10 +309,8 @@ type readerOptions struct {
 	legacyPositionUsed  bool
 }
 
-// For most files, createReader returns a tailer implementation. If the file suffix alludes to it being
-// a compressed file, then a decompressor will be created instead.
-func (c *Component) createReader(opts readerOptions) (reader, error) {
-	var reader reader
+// newSource will return return a decompressor source if enabeld, otherwise a tailer source.
```
thampiotr:
```suggestion
// newSource will return a decompressor source if enabled, otherwise a tailer source.
```

internal/component/loki/source/file/file.go
kalleep: Ah yeah that is a really bad pattern I agree, yep should do what you suggested for sure!

internal/component/loki/source/file/file.go
kalleep: So what you mentioned as a con is why I did not add a mutex to the scheduler but handle it from the outside. 

There may be some api that would allow us to handle transactions with inner locks but would have to think about it for a little bit

thampiotr: Yeah, you can have some `func (*Scheduler) DoInTransaction(action func())`. 
But we should go with whatever is easier to follow and less error-prone, so if you try it and conclude it's not the best way, I don't mind.

kalleep: Would it be okay to leave this as a follow up after I have merged this and the other work I am adding to this component?

I think it would be clearer then if this is something we should do or not.

thampiotr (approved):

PR #4714, opened 2025-10-29
implement the remotecfg effective config feature

erikbaranowski:
#### PR Description

Also fix 1 scenario for remote config status which is detailed [here](https://github.com/grafana/alloy/pull/4714#discussion_r2475717073)

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ Main (unreleased)
 
 - Send remote config status to the remote server for the remotecfg service. (@erikbaranowski)
 
+- Send effective config to the remote server for the remotecfg service. (@erikbaranowski)
+
 - Add a `stat_statements` configuration block to the `prometheus.exporter.postgres` component to enable selecting both the query ID and the full SQL statement. The new block includes one option to enable statement selection, and another to configure the maximum length of the statement text. (@SimonSerrano) 
 
 ### Enhancements


internal/service/remotecfg/config_manager.go
```diff
@@ -200,23 +214,30 @@ func (cm *configManager) fetchLoadConfig(getAPIConfig func() (*collectorv1.GetCo
...
+
+	cm.notifyStatusUpdate(getAPIConfig)
 }
 
 // notifyStatusUpdate makes an immediate GetConfig call to notify the server of status changes.
 // This is used when we want to immediately report status changes without waiting for the next poll cycle.
 func (cm *configManager) notifyStatusUpdate(getAPIConfig func() (*collectorv1.GetConfigResponse, error)) {
-	level.Debug(cm.logger).Log("msg", "making immediate GetConfig call to report status update")
+	// Avoid unnecessary immediate calls if there's nothing new to report.
+	if !cm.hasPendingUpdates() {
```
erikbaranowski, line 229: this let's us call notifyStatusUpdate during fetch by detecting if the API call is actually needed without having to  analyze and inject it numerous places at a lower level.

internal/service/remotecfg/config_manager.go
```diff
@@ -200,23 +214,30 @@ func (cm *configManager) fetchLoadConfig(getAPIConfig func() (*collectorv1.GetCo
...
 		if useCacheAsFallback {
 			level.Error(cm.logger).Log("msg", "failed to fetch remote config, falling back to cache", "err", err)
-			cm.fetchLoadLocalConfig()
+			cm.fetchLoadLocalConfig(getAPIConfig)
 		} else {
 			level.Error(cm.logger).Log("msg", "failed to fetch remote config, continuing with current config", "err", err)
 		}
 	}
+
+	cm.notifyStatusUpdate(getAPIConfig)
```
erikbaranowski, line 222: just call `notifyStatusUpdate` once here

https://github.com/grafana/alloy/pull/4714/files#r2475710117

internal/service/remotecfg/config_manager.go
```diff
@@ -233,6 +254,18 @@ func (cm *configManager) fetchLoadRemoteConfig(getAPIConfig func() (*collectorv1
 	if err == errNotModified {
 		level.Debug(cm.logger).Log("msg", "skipping over API response since it has not been modified since last fetch")
 		cm.metrics.lastFetchNotModified.Set(1)
+
+		// Only mark APPLIED if the last received remote config matches the currently
```
erikbaranowski, line 257: this is a fix for how remote config status works which I stumbled upon in my testing. here's the scenario:

- we have a working config running
- the remote config is updated and an API error returns from GetConfig because of (for example) an invalid export injection reference
- the original config is used as a fallback from the cache
- the remote config gets reverted to match the original config which is running from cache
- the remote config status is never reset to Applied with no error message despite the server now sending a working config

internal/service/remotecfg/remotecfg_test.go
```diff
@@ -608,37 +605,6 @@ func TestRemoteConfigStatus_InitialState(t *testing.T) {
 	assertRemoteConfigStatus(t, env, collectorv1.RemoteConfigStatuses_RemoteConfigStatuses_UNSET, false)
 }
 
-func TestGetErrorMessage_DiagnosticErrors(t *testing.T) {
```
erikbaranowski, line 611: moved these lower level tests to config_manager_test.go

internal/service/remotecfg/config_manager.go
```diff
@@ -496,6 +527,163 @@ func (cm *configManager) copyRemoteConfigStatus() *collectorv1.RemoteConfigStatu
 	}
 }
 
+// setEffectiveConfig updates the effective configuration that is currently running.
+func (cm *configManager) setEffectiveConfig(config []byte) {
```
tpaschalis, line 531: I'm trying to verify what we mean by effective config here. Do we include the local configuration or no?

The description of the OpAMP message is: 

```
    // The current effective configuration of the Agent. The effective configuration is
    // the one that is currently used by the Agent. The effective configuration may be
    // different from the remote configuration received from the Server earlier, e.g.
    // because the Agent uses a local configuration instead (or in addition).
```

As far as I can tell, we only ever call this method using the bytes from the remote server as the argument -- I can't see if we somehow include the 'local' configuration as well.

At the same time, the only use I can see for the 'root' configPath is when we call `ctrl.LoadSource(b, nil, configPath)` and following the code I see it's only ever used to set the arg scope.

erikbaranowski: it's a good point, let me lay out how things work in this implementation

### Definitions
**root config**: config stored locally to start up alloy, Includes the `remotecfg` block config for the `remotecfg` service.

**cached remote config**: the last working remote config received and stored locally in alloy's storage path. Used as a fallback if the latest remote config loads with error.

**remote config**: latest config sent by the server

**effective config**: the remote config or cached remote config if being used as a fallback

### Implementation Details
- root config
  - not currently included in effective config
  - maybe this is something to consider in the future but alloy doesn't merge remote/local config yamls so the context is a bit different than typical opamp.

I can apply some clearer wording in the comment [upstream](https://github.com/grafana/alloy-remote-config/blob/47414cf6afa8fbbd35a53c4a8e60f677ad02dd95/api/collector/v1/collector.proto#L29-L35) and pull it down on this PR if we have consensus on leaving off root config for now in the effective config response.

tpaschalis: I'm OK with it. The distinction between us and upstream that the 'local' and 'remote' configs are not merged but run separately makes us different enough. I think it's potentially worth including the local configuration in the future as well.

tpaschalis (approved):
LGTM!

PR #4734, opened 2025-10-30
Temp -> OTel Engine POC: Introduce Workspace With Shared Replace Directives

blewis12:
This is a draft just to see how this behaves in pipelines and gather pros/cons/blockers


blewis12:
hmm no - this needs further thought 😆

PR #4059, opened 2025-07-28
Fix races reported in Alloy tests 

bboreham:
#### PR Description

Fix races reported in Alloy tests 
Adjusted make rule and clarified comment about this.

Also Run packaging tests with `-race`, because I couldn't see why not.

#### Notes to the Reviewer

Nothing here would result in bad data corruption; it's mostly tests and some ancilliary data.

One race remains in the `node_exporter` package. I created https://github.com/grafana/node_exporter/pull/35, although Alloy is using a different branch.

Note tests in `internal/component/faro/receiver/receiver_otelcol_test.go` were not being run in CI.

#### PR Checklist

- [x] Tests updated

PR #4743, opened 2025-10-31
feat: loki.source.file file_match block

kalleep:
#### PR Description

This PR adds a new `file_match` block to `loki.source.file` that enables built-in file discovery using glob patterns. This  allows users to discover files directly within `loki.source.file` without needing a separate `local.file_match` component.

If `file_match` block is not enabled component behaves like it did before. 

If this block enabled  `local.file_match` is no longer needed and we avoid the overhead of waiting for alloy graph updates before we can schedule new files to tail.

#### Notes to the Reviewer

I added "resolver" interface that will change depending on configuration. The static one behaves exactly like before, requiring absolute paths and the glob implementation will perform globing with doublestar.

The watcher will do nothing when file_match is not configured.

I also changed to use nop logger in test, logging in tests adds additional noise that makes it hard to see test output. Maybe we should be able to control logging for test with an env variable instead? 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/file.go
+++ b/internal/component/loki/source/file/file.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/grafana/dskit/backoff"
 	"github.com/prometheus/common/model"
-
 	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component"
--- a/internal/component/loki/source/file/tailer.go
+++ b/internal/component/loki/source/file/tailer.go
@@ -14,9 +14,6 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/grafana/alloy/internal/component/loki/source/file/internal/tail"
-	"github.com/grafana/alloy/internal/component/loki/source/file/internal/tail/watch"
-	"github.com/grafana/alloy/internal/loki/util"
 	"github.com/grafana/loki/pkg/push"
 	"github.com/prometheus/common/model"
 	"go.uber.org/atomic"
@@ -27,6 +24,9 @@ import (
 	"github.com/grafana/alloy/internal/component/common/loki"
 	"github.com/grafana/alloy/internal/component/common/loki/positions"
 	"github.com/grafana/alloy/internal/component/common/loki/utils"
+	"github.com/grafana/alloy/internal/component/loki/source/file/internal/tail"
+	"github.com/grafana/alloy/internal/component/loki/source/file/internal/tail/watch"
+	"github.com/grafana/alloy/internal/loki/util"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
 )
 


docs/sources/reference/components/local/local.file_match.md
```diff
@@ -14,6 +14,13 @@ title: local.file_match
 
 `local.file_match` discovers files on the local filesystem using glob patterns and the [doublestar][] library.
 
+{{< admonition type="note" >}}
+For simple use cases where you only need to discover files for a single `loki.source.file` component, consider using the built-in [`file_match`](../../loki/loki.source.file/#file_match) block in `loki.source.file` instead.
+This provides better performance and simpler configuration by integrating file discovery directly into the component.
+
+Use `local.file_match` when you need to share discovered targets across multiple components, reuse file discovery results in different parts of your configuration, or accept targets from other discovery components (e.g., `discovery.relabel`) before file matching.
```
thampiotr:
```suggestion
Use `local.file_match` component when you need to share discovered targets across multiple components, reuse file discovery results in different parts of your configuration, or accept targets from other discovery components (e.g., `discovery.relabel`) before file matching.
```

It probably should be used if there's a need to do some more processing of targets it outputs, right?

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -14,6 +14,13 @@ title: local.file_match
 
 `local.file_match` discovers files on the local filesystem using glob patterns and the [doublestar][] library.
 
+{{< admonition type="note" >}}
+For simple use cases where you only need to discover files for a single `loki.source.file` component, consider using the built-in [`file_match`](../../loki/loki.source.file/#file_match) block in `loki.source.file` instead.
```
thampiotr: I think it's also better option for very large numbers of files that are frequently changing? 

Perhaps we should recommend to always use the built-in block and only use local.file_match for things that are not possible with the built-in block. That way we ensure people get maximum correctness & performance. 

Should we try create a list of use-cases that are not possible with built-in block?

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
...
+| `enabled`           | `bool`              | Enable file discovery using glob patterns.                  | `false` | no       |
+| `ignore_older_than` | `duration`          | Ignores files with modification times before this duration. | `"0s"`  | no       |
+| `sync_period`       | `duration`          | How often to sync filesystem and targets.                   | `"10s"` | no       |
+
+Benefits of using `file_match` over `local.file_match`:
+- Simplified configuration: No need for a separate `local.file_match` component
+- Better performance: File discovery is integrated directly into `loki.source.file`, eliminating the overhead of component-to-component communication
+- Same functionality: Supports the same glob patterns and features, including `__path_exclude__` for excluding files
+
+When `enabled` is set to `true`, you can use glob patterns (e.g., `/tmp/*.log` or `/var/log/**/*.log`) directly in the `targets` argument's `__path__` label.
```
thampiotr: What happens when enabled is false?

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
...
+The component periodically scans the filesystem based on `sync_period` and automatically discovers new files, removes deleted files, and ignores files older than `ignore_older_than` if specified.
```
thampiotr, line 132: Didn't we have fsnotify-based implementation where it's supported?

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
 
 If file changes are detected, the poll frequency is reset to `min_poll_frequency`.
 
+### `file_match`
+
+The `file_match` block enables built-in file discovery directly within `loki.source.file`.
+When enabled, the component discovers files on the local filesystem using glob patterns and the [doublestar][] library.
```
thampiotr, line 115: What happens when someone provides file_match and also sends targets in using the previous pattern? 

I think that should ideally be an error, unless there's a simple way to make both work.

kalleep: It will work but you get some overhead because now you are globing and syncing twice. Because all paths will match directly to one file and that is fine to pass to `Glob`

docs/sources/reference/components/loki/loki.source.file.md
kalleep: I don't think so. The only usage I can see of fsnotify is https://github.com/grafana/alloy/blob/main/internal/filedetector/detector.go#L68 and this is used by `local.file` and not `local.file_match`.

docs/sources/reference/components/loki/loki.source.file.md
kalleep: Then we require absolute file paths, like this component behaved before this pr

docs/sources/reference/components/local/local.file_match.md
kalleep: Yes!

docs/sources/reference/components/local/local.file_match.md
kalleep: Yes there is examples with file globing using either builtin or `local.file_match`. There I added this header 
`### File globbing with `file_match` (recommended)`

Do you think we should make it clearer here?

internal/component/loki/source/file/file.go
```diff
@@ -197,25 +232,43 @@ func (c *Component) Update(args component.Arguments) error {
...
+		c.resolver = newGlobResolver()
+		// It is safe to take the lock in the callback function.
+		// We never invoke it directly and it's only called by the watcher.
+		c.watcher.Update(newArgs.FileMatch.SyncPeriod, func() {
+			c.mut.Lock()
+			defer c.mut.Unlock()
+			c.scheduleSources()
+		})
+	} else if !newArgs.FileMatch.Enabled && c.args.FileMatch.Enabled {
+		c.resolver = newStaticResolver()
```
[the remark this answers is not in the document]
kalleep: This should not be a problem.

In the first case where Update is called and and watcher ticks before `watcher.Update` is called the tick will be a no op. 

If Update is called and watcher ticks after we have have updated the watcher but we are still holding mut in components Update then watcher will wait until it can take the lock and perform the sync.

internal/converter/internal/promtailconvert/testdata/azure.alloy
```diff
@@ -17,7 +17,9 @@ local.file_match "fun" {
 }
 
 loki.source.file "fun" {
-	targets               = local.file_match.fun.targets
-	forward_to            = []
+	targets    = local.file_match.fun.targets
+	forward_to = []
+
+	file_match { }
```
thampiotr: Hmm, something is not right with the defaults I think. Ideally there should be no change needed for the converters.

kalleep: True, we don't see this for e.g. FileWatch. Will check why this happens

kalleep: Okay it's fixed now

docs/sources/reference/components/local/local.file_match.md
thampiotr: Yes, I think we need to make it more clear when to use which option. And probably the logic should be to use the built-in `file_match` whenever you can and a separate component for cases that are not possible with built-in (we need to list a few of them)

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
...
+
+The following arguments are supported:
+
+| Name                | Type                | Description                                                 | Default | Required |
+|---------------------|---------------------|-------------------------------------------------------------|---------|----------|
+| `enabled`           | `bool`              | Enable file discovery using glob patterns.                  | `false` | no       |
+| `ignore_older_than` | `duration`          | Ignores files with modification times before this duration. | `"0s"`  | no       |
+| `sync_period`       | `duration`          | How often to sync filesystem and targets.                   | `"10s"` | no       |
```
thampiotr, line 123: is that a good default? we have a chance to change it

docs/sources/reference/components/loki/loki.source.file.md
thampiotr: Can we write that in docs?

kalleep: We already have it here https://github.com/grafana/alloy/pull/4743/files#diff-4b8c6b16f7e932bea249731a9f49d947024699eb129a37f0846aed9c1086e945R21-R23.

Do you think we need it in both places?

docs/sources/reference/components/loki/loki.source.file.md
kalleep: I think the defaults are good, this is the same defaults that are using in promtail too.

docs/sources/reference/components/loki/loki.source.file.md
kalleep: Changed this in https://github.com/grafana/alloy/pull/4743/commits/61c3e8ce75dc9a2b373be3995975dce54df97bae 

I listed the use cases I could think of for `local.file_match`

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -14,6 +14,14 @@ title: local.file_match
 
 `local.file_match` discovers files on the local filesystem using glob patterns and the [doublestar][] library.
 
+{{< admonition type="note" >}}
+You should always prefer using the built-in [`file_match`](../../loki/loki.source.file/#file_match) block in `loki.source.file` when possible.
```
clayton-cornell:
```suggestion
Use the built-in [`file_match`](../../loki/loki.source.file/#file_match) block in `loki.source.file` whenever possible.
```

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -14,6 +14,14 @@ title: local.file_match
...
+This provides better performance by integrating file discovery directly into the component, eliminating the overhead of a separate component.
+The performance benefit is especially noticeable in environments with many files to watch.
+
+Use `local.file_match` component only when you need to share discovered targets across multiple components, reuse file discovery results in different parts of your configuration, or accept targets from other discovery components (e.g., `discovery.relabel`) before file matching.
```
clayton-cornell:
```suggestion
Use the `local.file_match` component only when you need to share discovered targets across multiple components, reuse file discovery results in different parts of your configuration, or accept targets from other discovery components, for example,  `discovery.relabel` before file matching.
```

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -60,13 +62,15 @@ When `legacy_positions_file` is set, Alloy will try to find previous positions f
...
 
-| Name                             | Description                                                       | Required |
-| -------------------------------- | ----------------------------------------------------------------- | -------- |
-| [`decompression`][decompression] | Configure reading logs from compressed files.                     | no       |
-| [`file_watch`][file_watch]       | Configure how often files should be polled from disk for changes. | no       |
+| Name                             | Description                                                                  | Required |
+| -------------------------------- | ---------------------------------------------------------------------------- | -------- |
+| [`decompression`][decompression] | Configure reading logs from compressed files.                                | no       |
+| [`file_watch`][file_watch]       | Configure how often files should be polled from disk for changes.            | no       |
+| [`file_match`][file_match]       | Configure file discovery using glob patterns for automatic target discovery. | no       |
```
```suggestion
| [`file_match`][file_match]       | Configure file discovery using glob patterns for automatic target discovery. | no       |
| [`file_watch`][file_watch]       | Configure how often files should be polled from disk for changes.            | no       |
```

Alpha sorting

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
...
+The following arguments are supported:
+
+| Name                | Type                | Description                                                 | Default | Required |
+|---------------------|---------------------|-------------------------------------------------------------|---------|----------|
+| `enabled`           | `bool`              | Enable file discovery using glob patterns.                  | `false` | no       |
+| `ignore_older_than` | `duration`          | Ignores files with modification times before this duration. | `"0s"`  | no       |
+| `sync_period`       | `duration`          | How often to sync filesystem and targets.                   | `"10s"` | no       |
+
+Benefits of using `file_match` over `local.file_match`:
+- Simplified configuration: No need for a separate `local.file_match` component
```
clayton-cornell, line 127:
```suggestion
Benefits of using `file_match` over `local.file_match`:

- Simplified configuration: No need for a separate `local.file_match` component
```

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -105,6 +109,27 @@ If no file changes are detected, the poll frequency doubles until a file change
...
+- Reduced overhead: File discovery is integrated directly into `loki.source.file`, eliminating the need for component-to-component communication
+- Same functionality: Supports the same glob patterns and features, including `__path_exclude__` for excluding files
+
+When `enabled` is set to `true`, you can use glob patterns (e.g., `/tmp/*.log` or `/var/log/**/*.log`) directly in the `targets` argument's `__path__` label.
```
```suggestion
When `enabled` is set to `true`, you can use glob patterns, for example, `/tmp/*.log` or `/var/log/**/*.log`, directly in the `targets` argument's `__path__` label.
```

We should avoid brackets and e.g. in the docs.

internal/component/loki/source/file/file_test.go
thampiotr, line 1: There is a `//go:build !race` on top of this file. 
- Are we sure this test runs in CI? it's not obvious to me right now
- Could we get rid of this and fix the race conditions, since we done some improvements recently?

internal/component/loki/source/file/file.go
```diff
@@ -197,25 +254,39 @@ func (c *Component) Update(args component.Arguments) error {
...
 
-	c.scheduleSources(newArgs)
+	// Choose resolver on FileMatch.
+	if newArgs.FileMatch.Enabled {
+		c.resolver = newGlobResolver()
+	} else {
+		c.resolver = newStaticResolver()
+	}
+
+	if newArgs.FileMatch.SyncPeriod != c.args.FileMatch.SyncPeriod && newArgs.FileMatch.SyncPeriod < 0 {
```
thampiotr:
```suggestion
	if newArgs.FileMatch.SyncPeriod != c.args.FileMatch.SyncPeriod && newArgs.FileMatch.SyncPeriod > 0 {
```
Although I'm not sure if that can even happen? Because default is not `0`, right? 
maybe we should think what to do with c.watcher when we disable FileMatch separately.

internal/component/loki/source/file/file.go
```diff
@@ -114,14 +144,18 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, err
 	}
 
-	c := &Component{
-		opts:    o,
-		metrics: newMetrics(o.Registerer),
+	if args.FileMatch.SyncPeriod < 1 {
+		args.FileMatch.SyncPeriod = 10 * time.Second
+	}
```
thampiotr: I don't think you need that for default. In case someone sets it to negative number, can we have a Validate() function instead for this? 

We want to use the existing default/validation mechanisms so that New can be more focused.

internal/component/loki/source/file/file.go
```diff
@@ -151,30 +185,53 @@ func (c *Component) Run(ctx context.Context) error {
...
+	wg.Go(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-c.watcher.C:
+				c.mut.Lock()
+				if !c.args.FileMatch.Enabled {
+					return
+				}
```
thampiotr, line 234: So if we disable FileMatch and then enable FileMatch, I think the goroutine will not be restarted... can we add a test for this and fix the issue? 

We have two choices: always keep the goroutines and watcher ticker on and check for FileMatch enabled - a bit more expensive but a simpler code. Or we schedule them only when FileMatch is enabled - a bit more complex, but fewer resources used. I don't have a strong opinion, I would like this code to be easier to follow though...

docs/sources/reference/components/local/local.file_match.md
```diff
@@ -14,6 +14,14 @@ title: local.file_match
 
 `local.file_match` discovers files on the local filesystem using glob patterns and the [doublestar][] library.
 
+{{< admonition type="note" >}}
+Use the built-in [`file_match`](../../loki/loki.source.file/#file_match) block in `loki.source.file` whenever possible.
```
thampiotr, line 18: We might want to reach out to k8s helm charts maintainers to change their defaults if they have them.

internal/component/loki/source/file/file.go
kalleep: Good catch! My intention was to keep it running regardless, so we should not return here.

And only because it's way more simple to do that then having to manage it in between Updates

internal/component/loki/source/file/file.go
kalleep: Agreed, that is better. I added this one and the other check later because a bunch of tests failed but will fix those up

kalleep:
@thampiotr Fixed watcher and added Validate to arguments. 

I also removed the tag and applied fix from https://github.com/grafana/alloy/pull/4059

thampiotr (approved):
Nice work! thanks

PR #4752, opened 2025-11-02
fix: WAL and matches conversion for prometheus.remote_write and loki.source.journal

fgouteroux:
#### PR Description

Fix static converter missing `wal` in `prometheus.remote_write`, and `matches` in `loki.source.journal`.

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


fgouteroux:
Hi team, no one to check this PR ?

blewis12:
Hiya @fgouteroux! Sorry about the delay in review, this one seems to have slipped through the cracks somehow. I'm assigning myself to this to review, but if other maintainer folks also want to jump in feel free

internal/converter/internal/staticconvert/internal/build/builder_integrations.go
```diff
@@ -193,7 +193,14 @@ func (b *ConfigBuilder) appendExporter(commonConfig *int_config.Common, name str
 		return b.jobNameToCompLabel(jobName)
 	}
 
-	b.diags.AddAll(prometheusconvert.AppendAllNested(b.f, promConfig, jobNameToCompLabelsFunc, extraTargets, b.globalCtx.IntegrationsRemoteWriteExports))
+	// Extract WAL settings from common config
+	walOptions := &remotewrite.WALOptions{
+		TruncateFrequency: commonConfig.WALTruncateFrequency,
```
blewis12: `commonConfig.WALTruncateFrequency` has a default/empty value of `0`, essentially disabling WAL truncation. I think on an empty value we should be setting `WALTruncateFrequency` to `remotewrite.DefaultWALOptions.WALTruncateFrequency`. This means that if we're always falling back to the default if a value is unset

I can see our integrations.yaml conversion test is failing for this reason

internal/converter/internal/staticconvert/internal/build/builder_integrations.go
```diff
@@ -327,6 +335,13 @@ func (b *ConfigBuilder) appendExporterV2(commonConfig *common_v2.MetricsConfig,
 			remoteWriteExports = &remotewrite.Exports{
 				Receiver: common.ConvertAppendable{Expr: "prometheus.remote_write." + label + ".receiver"},
 			}
+
+			// Extract WAL settings from metrics instance config
+			walOptions = &remotewrite.WALOptions{
```
blewis12, line 358: I can see that we have tests failing (not sure why they're not running in GH but you can run them in `staticconvert_test.go` locally). I think the issue is that we're using two sets of defaults in our conversion. In v1 we're using `remotewrite.DefaultWALOptions` which are

```
DefaultWALOptions = WALOptions{
		TruncateFrequency: 2 * time.Hour,
		MinKeepaliveTime:  5 * time.Minute,
		MaxKeepaliveTime:  8 * time.Hour,
	}
```

But in v2 we're using [instance defaults](https://github.com/grafana/alloy/blob/main/internal/static/metrics/instance/instance.go#L35-L42) which are different. 

In our builder, we check against the _alloy_ defaults, and [omit if they match](https://github.com/grafana/alloy/blob/923d127c50949f88a37ac808154240de0649df09/syntax/token/builder/builder.go#L194). So the tests fail because we're unexpectedly including a `wal` block with non-default values

My instinct is to align these two defaults, but I need to think on that a little more 🤔

blewis12:
(Just changed the title of the PR to match our title validation action - this is a new feature introduced into the codebase)

internal/converter/internal/staticconvert/internal/build/builder_integrations.go
blewis12: > My instinct is to align these two defaults

Digging into it a bit more I realize this is not the way: the core issue is that instance defaults are different than Alloy defaults. In `appendStaticPrometheus` we are settings values to the incoming `instance` (`staticConfig.Metrics.Configs[..].WALTruncateFrequency`) values, but if these are omitted from the input config they are set to the instance defaults in the [unmarshaller](https://github.com/grafana/alloy/blob/923d127c50949f88a37ac808154240de0649df09/internal/static/metrics/instance/instance.go#L70).  Likewise in `appendExporterV2` we are settings the values to the incoming `metrics` values, which are also derived from `staticConfig.Metrics.Configs[..].WALTruncateFrequency` so we have the same issue.

I think that we should be falling back to Alloy defaults, that is the current behavior of the conversion if a user omits WAL configuration in their incoming config - and these defaults have likely been set for a good reason. My suggestion therefore is:

* If an incoming config has either a zero value or an instance default value, which would have been applied automatically by the unmarshaller, then fallback to Alloy defaults
* Else, use the incoming values

The other option is to keep the instance defaults and update the tests, this would also change current behavior for users who are not referencing WAL settings at all in their incoming config - so I suggest against this.

blewis12: Discussed this a bit internally, and there's one use case that my above point doesn't address: if the incoming config has had it's WAL values explicitly set to the default. Since we cannot differentiate between an empty and a default value, that makes knowing when to fallback to Alloy defaults tricky.

In the end, it might actually make more sense to use the `instance` defaults - and modify the tests to reflect this. The current behavior will also be changed, but one could argue that it's also been "fixed"

Anyway, sorry for the back and forth there, I hope that trail of thought makes sense. Happy also to hear your opinion on that one @fgouteroux

fgouteroux: Thanks @blewis12! Agreed — using instance defaults makes sense since we can't distinguish unset from explicitly-set-to-default. Updated the code and tests accordingly.

fgouteroux:
hi @blewis12, we forgot to progress on this topic, could we move forward ?

PR #4671, opened 2025-10-24
Reapply Prometheus upgrade and fix ckit & upstream issue

thampiotr:
#### PR Description

* Bring back the Prometheus upgrade reverted here: https://github.com/grafana/alloy/pull/4667 
* Add ckit fix: https://github.com/grafana/ckit/pull/109 
* Add the https://github.com/prometheus/prometheus/pull/17412 fix by rebasing to the new, updated Prom fork branch.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated
- [ ] Config converters updated

PR #4756, opened 2025-11-03
Remove Slicelabels Tag to Enable Stringlabels 

blewis12:
#### PR Description
Enables stringlabels in alloy. Since [prometheus was upgraded](https://github.com/grafana/alloy/pull/4671), stringlabels are the default and so we had to include the `slicelabels` tag in order to have things compile whilst there were still stringlabel incompatible code - this PR removes those tags

https://github.com/grafana/alloy/issues/4323

#### PR Checklist
- [ ] CHANGELOG.md updated <- shall I add this to the changelog? it should not be user facing - but it could be argued that changes in memory are user facing if they are monitoring this 🤷

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
         # TODO: Try enabling caching later. It might use up too much disk space on runners so needs extra testing.
         cache: false
     - run: make generate-ui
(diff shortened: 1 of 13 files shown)


ptodev:
@blewis12 Since the PR is still WIP, would you like to change the status to "draft"?

blewis12:
@ptodev good point, done

thampiotr (approved):

PR #4757, opened 2025-11-03
test(beyla): add missing trace attribute

skl:
#### PR Description

`TestBeylaTraces` integration test fails to find traces matching:

- service.name: "main"
- test_name: "beyla"

This PR fixes this by adding the `test_name: "beyla"` attribute to traces using a `otelcol.processor.attributes` component.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/cmd/integration-tests/tests/beyla/config.alloy
+++ b/internal/cmd/integration-tests/tests/beyla/config.alloy
@@ -11,6 +11,18 @@ beyla.ebpf "default" {
       ]
   }
   output {
+    traces = [otelcol.processor.attributes.beyla.input]
+  }
+}
+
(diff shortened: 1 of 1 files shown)


ptodev:
I had a chat with Cursor regarding this integration test, and this fix is also what it suggested. I don't mind adding it in. But I don't understand why the test is flaky? If the missing attribute is the cause for failures, why would it ever succeed?

skl:
@ptodev I don't entirely understand how this could've passed, either. It has been skipped/unskipped a few times in the past... all I can say is, see how it goes with this attribute being set, if it's still flaky we can look at it another way and at least rule this out?

PR #4762, opened 2025-11-04
fix(metrics/wal): missing NHCBs failure

krajorama:
#### PR Description

First commit: The metrics/wal does not write out native histograms with custom buckets (NHCB). Update the unit test TestStorage to verify write/read of histograms and trigger the bug.

Second commit: fix by correctly encoding NHCBs to WAL.

#### PR Checklist

- [x] CHANGELOG.md updated
- [N/A] Documentation added
- [X] Tests updated
- [N/A] Config converters updated


internal/static/metrics/wal/wal_test.go
```diff
@@ -776,6 +814,54 @@ func buildSeries(nameSlice []string) seriesList {
...
+		{
+			name:    "integer histogram",
+			samples: []sample{{2, -1, tsdbutil.GenerateTestHistogram(1), nil}, {20, -1, tsdbutil.GenerateTestHistogram(100), nil}},
+			exemplars: []exemplar.Exemplar{
+				{Labels: labels.Labels{{Name: "foobar", Value: "barfoo"}}, Value: float64(10.0), Ts: int64(2), HasTs: true},
+				{Labels: labels.Labels{{Name: "lorem", Value: "ipsum"}}, Value: float64(100.0), Ts: int64(20), HasTs: true},
+			},
+		},
+		{
+			name:    "float histogram",
```
thampiotr, line 839: Is it worth testing an edge case where both float and integer histograms are set? Or is there something that will reliably prevent it from happening earlier on?

internal/static/metrics/wal/wal_test.go
```diff
@@ -787,6 +873,28 @@ func (b byRefSample) Less(i, j int) bool {
...
+
+func (b byRefHistogramSample) Len() int      { return len(b) }
+func (b byRefHistogramSample) Swap(i, j int) { b[i], b[j] = b[j], b[i] }
+func (b byRefHistogramSample) Less(i, j int) bool {
+	if b[i].Ref == b[j].Ref {
+		return b[i].T < b[j].T
+	}
+	return b[i].Ref < b[j].Ref
+}
+
```
thampiotr, line 886: Maybe we can reduce all the three sorting structs code by using https://pkg.go.dev/sort#Slice ?

ptodev (approved):

internal/static/metrics/wal/wal_test.go
krajorama: The current WAL code doesn't much care about commit order for mixed typed series. We could write a unit test to demonstrate it, but maybe it would be better to have a separate PR and take inspiration from https://github.com/prometheus/prometheus/pull/17071.

internal/static/metrics/wal/wal_test.go
krajorama: Not sure what you mean. Currently wal order is type first then ref then time. So you have to separate items by type anyway.
And sort.Slice seems complicated to use in general, I'd probably have to write a function to create the `less`
function from input?

thampiotr: I thought of using:
```
func sortSamples(samples []record.RefSample) {
	sort.Slice(samples, func(i, j int) bool {
		if samples[i].Ref == samples[j].Ref {
			return samples[i].T < samples[j].T
		}
		return samples[i].Ref < samples[j].Ref
	})
}
```
and similar functions instead of defining slices with 3 different method each... but I know the slices is the 'classic' way of doing this sort in Go and the difference would be minor, so we can skip this suggestion 👍

PR #4769, opened 2025-11-04
database_observability: Reduce info level logging noise for explain plans

rgeyer:
#### PR Description
We were logging a few things at `info` that belonged at `debug`.

Also caught a missing error check, and sync'd up the log messages a bit better.

Fixes grafana/grafana-dbo11y-app#1789

### DIFF
--- a/internal/component/database_observability/postgres/collector/explain_plan.go
+++ b/internal/component/database_observability/postgres/collector/explain_plan.go
@@ -339,6 +339,11 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 	defer rs.Close()
 
 	for rs.Next() {
+		if err := rs.Err(); err != nil {
+			level.Error(c.logger).Log("msg", "failed to iterate rs digests for explain plans", "err", err)
+			return err
+		}
+
 		var datname, queryId, query string
 		var calls int64
 		var ls time.Time
@@ -374,7 +379,7 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 	}
 
 	c.currentBatchSize = int(math.Ceil(float64(len(c.queryCache)) * c.perScrapeRatio))
-	level.Info(c.logger).Log("msg", "populated query cache", "count", len(c.queryCache), "batch_size", c.currentBatchSize)
+	level.Debug(c.logger).Log("msg", "populated query cache", "count", len(c.queryCache), "batch_size", c.currentBatchSize)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -339,6 +339,11 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
 	defer rs.Close()
 
 	for rs.Next() {
+		if err := rs.Err(); err != nil {
```
matthewnolf: I think this should be called outside of the for loop? If `rs.Next()` returns true there won't have been an error on iteration

rgeyer: Yes, you're quite right.

I did this wrong in the mysql loop, and simply copied it here when I was writing this PR. Just updated both to the "correct" usage of the `Rows.Next()` API. I also aligned the error messaging/logging between the two collectors.

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -316,25 +316,20 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
...
+		if err := statReset.Err(); err != nil {
+			return fmt.Errorf("failed to fetch stats reset time for explain plans: %w", err)
 		}
-		defer statReset.Close()
-		if statReset.Next() {
-			if err := statReset.Scan(&resetTS); err != nil {
-				level.Error(c.logger).Log("msg", "failed to scan stats reset time for explain plans", "err", err)
-				return err
-			}
+		if err := statReset.Scan(&resetTS); err != nil {
```
matthewnolf, line 323: I think we're now missing a `statReset.Next()` here
Note from the docs:
> Every call to [Rows.Scan](https://pkg.go.dev/database/sql#Rows.Scan), even the first one, must be preceded by a call to [Rows.Next](https://pkg.go.dev/database/sql#Rows.Next).

rgeyer: I changed the context from a `QueryContext` to a `QueryRowContext`, since this will only ever return one row.

Sorry for the weird compound change.

internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -316,25 +316,20 @@ func (c *ExplainPlan) populateQueryCache(ctx context.Context) error {
...
+			return fmt.Errorf("failed to scan stats reset time for explain plans: %w", err)
 		}
 		selectStatement = fmt.Sprintf(selectQueriesForExplainPlanTemplate, "NOW() AT TIME ZONE 'UTC' AS stats_since")
 	}
 
 	rs, err := c.dbConnection.QueryContext(ctx, selectStatement)
 	if err != nil {
 		level.Error(c.logger).Log("msg", "failed to fetch digests for explain plans", "err", err)
```
matthewnolf: Will we end up double logging this here and where the error gets propagated?

rgeyer: Yeah, quite right. I thought I caught all of those. Fixed this one as well.

matthewnolf (approved):

PR #4771, opened 2025-11-04
database_observability: consider idle states to calculate Postgres query sample times

gaantunes:
#### PR Description

Including idle states processing, encompassing two scenarios:

Previously captured query actively running becomes idle, emit the sample calculating the duration from the stateChange coming from the idle state row.
Not yet captured query shows up as idle, emit a one-off sample with the information available in the idle row.

#### Notes to the Reviewer

This will generate a lot more samples for our postgres environments. https://github.com/grafana/alloy/pull/4772 will handle it.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/query_samples.go
+++ b/internal/component/database_observability/postgres/collector/query_samples.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
+	"github.com/hashicorp/golang-lru/v2/expirable"
 	"github.com/lib/pq"
 	"go.uber.org/atomic"
 
@@ -115,7 +116,7 @@ type QuerySamples struct {
 	// in-memory state of running samples
 	samples map[SampleKey]*SampleState
 	// keep track of keys that were already emitted to avoid duplicates
-	emitted map[SampleKey]time.Time
+	emitted *expirable.LRU[SampleKey, struct{}]
 }
 
 // SampleKey uses (PID, QueryID, QueryStartNs) so concurrent executions of the same
@@ -203,6 +204,9 @@ func (w WaitEventIdentity) Equal(other WaitEventIdentity) bool {
 }
 
 func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
+	const emittedCacheSize = 1000 //pg_stat_statements default max number of statements to track
+	const emittedCacheTTL = 10 * time.Minute
+
 	return &QuerySamples{
 		dbConnection:          args.DB,
 		collectInterval:       args.CollectInterval,
@@ -211,7 +215,7 @@ func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
 		logger:                log.With(args.Logger, "collector", QuerySamplesCollector),
 		running:               &atomic.Bool{},
 		samples:               map[SampleKey]*SampleState{},
-		emitted:               map[SampleKey]time.Time{},
+		emitted:               expirable.NewLRU[SampleKey, struct{}](emittedCacheSize, nil, emittedCacheTTL),
 	}, nil
 }
 
@@ -305,14 +309,14 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
 			if st, hadActive := c.samples[key]; hadActive {
 				st.markEndedAt(sample)
 				st.LastRow.State = sample.State
-				c.emitted[key] = sample.Now //is actually emitted at the end of the loop
-			} else if _, already := c.emitted[key]; !already {
+				c.emitted.Add(key, struct{}{}) // is actually emitted at the end of the loop
+			} else if _, already := c.emitted.Get(key); !already {
 				// new idle sample not yet seen -> create a new sample state to track and emit it
 				newIdleState := &SampleState{LastRow: sample, tracker: newWaitEventTracker()}
 				newIdleState.markEndedAt(sample)
 				newIdleState.LastRow.State = sample.State
 				c.samples[key] = newIdleState
-				c.emitted[key] = sample.Now //is actually emitted at the end of the loop
+				c.emitted.Add(key, struct{}{}) // is actually emitted at the end of the loop
 			}
 			continue
 		}
@@ -333,7 +337,6 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
 		}
 		c.emitAndDeleteSample(key)
 	}
-	c.cleanupEmitted(time.Now())
 	return nil
 }
 
@@ -581,15 +584,6 @@ func equalPIDSets(a, b []int64) bool {
 		}
 	}
 	return true
-}
-
-func (c *QuerySamples) cleanupEmitted(now time.Time) {
-	const ttl = 10 * time.Minute
-	for k, lastSeen := range c.emitted {
-		if now.Sub(lastSeen) > ttl {
-			delete(c.emitted, k)
-		}
-	}
 }
 
 func isIdleState(state string) bool {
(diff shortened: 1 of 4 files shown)


rgeyer (approved):

internal/component/database_observability/postgres/collector/schema_details_test.go
```diff
@@ -698,6 +700,7 @@ func Test_Postgres_SchemaDetails(t *testing.T) {
 		}, 2*time.Second, 100*time.Millisecond)
 
 		collector.Stop()
+		require.Eventually(t, func() bool { return collector.Stopped() }, 5*time.Second, 100*time.Millisecond)
```
matthewnolf: Do we need to test that `collector.Stop` works in each of these tests?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -137,6 +142,26 @@ type SampleState struct {
...
+}
+
+func (s *SampleState) updateCpuTimeIfActive(sample QuerySamplesInfo) {
+	if !sample.WaitEventType.Valid && !sample.WaitEvent.Valid && sample.State.String == stateActive {
+		s.LastCpuTime = calculateDuration(sample.StateChange, sample.Now)
+	}
+}
+
+// markEndedAt sets EndOverride and LastSeenAt based on the sample's state change or clock_timestamp if not available
+func (s *SampleState) markEndedAt(sample QuerySamplesInfo) {
```
matthewnolf:
```suggestion
func (s *SampleState) setEndedAt(sample QuerySamplesInfo) {
```

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -179,6 +204,9 @@ func (w WaitEventIdentity) Equal(other WaitEventIdentity) bool {
 }
 
 func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
+	const emittedCacheSize = 1000 //pg_stat_statements default max number of statements to track
```
matthewnolf, line 207: Should we not get this setting from the DB?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -179,6 +204,9 @@ func (w WaitEventIdentity) Equal(other WaitEventIdentity) bool {
...
+	const emittedCacheTTL = 10 * time.Minute
```
matthewnolf, line 208: Should this be configurable?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -112,6 +115,8 @@ type QuerySamples struct {
 
 	// in-memory state of running samples
 	samples map[SampleKey]*SampleState
+	// keep track of keys that were already emitted to avoid duplicates
```
matthewnolf: Why do we need this and `samples`. Can't we determine what we need from the presence of the QuerySample in `samples`

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -137,6 +142,26 @@ type SampleState struct {
 	LastSeenAt  time.Time
 	LastCpuTime string // last cpu_time observed under CPU condition
 	tracker     WaitEventTracker
+	// EndOverride is used to compute durations and timestamps when a query
+	// transitioned to idle or was only observed as idle.
+	EndOverride sql.NullTime
```
matthewnolf: This name is a bit confusing to me - it's not an override but when we determined the sample was idle?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -137,6 +142,26 @@ type SampleState struct {
...
+}
+
+func (s *SampleState) updateCpuTimeIfActive(sample QuerySamplesInfo) {
```
matthewnolf:
```suggestion
func (s *SampleState) updateCPUTimeIfActive(sample QuerySamplesInfo) {
```

matthewnolf: I think we could just call this `updateCPUTime`, the fact we use the state is kind of an implementation detail

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -269,6 +298,29 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
 		}
 
+		if !isIdleState(sample.State.String) {
+			c.upsertActiveSample(key, sample)
+			activeKeys[key] = struct{}{}
+			continue
+		}
+
+		// Handle idle states specially: emit finalized sample once
+		if isIdleState(sample.State.String) {
```
matthewnolf, line 302: Do we need this if block? It's preceded by the opposite which will `continue` if true

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -269,6 +298,29 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
+			if st, hadActive := c.samples[key]; hadActive {
+				st.markEndedAt(sample)
+				st.LastRow.State = sample.State
+				c.emitted.Add(key, struct{}{}) // is actually emitted at the end of the loop
```
matthewnolf: Why can't we just set `c.samples[key]` to indicate it has been emitted? We could have some process to regularly remove

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: This is just a reference value, since TTL is 10 minutes and we actually track pg_stat_activity, which is ephemeral, I don't expect we will ever use the actual pg_stat_statements max.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: I think this is more of an internal mechanism we don't need to expose to customers to reduce configuration complexity.

internal/component/database_observability/postgres/collector/schema_details_test.go
gaantunes: I think it adds value, checks if Stop() works as expected.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: We remove the sample from the `samples` map once it is emitted to free up memory. The `emitted` map is a simpler structure that tracks which `idle` rows where already emitted, since they will reappear at every scrape and we don't want to emit it again.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Good point, will rename for more clarity.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Good catch, will simplify the logic to contain only 1 if.

internal/component/database_observability/postgres/collector/schema_details_test.go
matthewnolf: I'm not against having a test for that. But unit tests should test one thing

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: We could, but I believe this would increase code complexity on the form of additional state checks, as the `samples` map is designed to keep in-fligh not yet finalized states only.

internal/component/database_observability/postgres/collector/schema_details_test.go
gaantunes: Fair. will remove.

matthewnolf (approved):

PR #4772, opened 2025-11-04
database_observability: add throttle and burst mechanisms to Postgres query samples

gaantunes:
#### PR Description

Adds burst and throttling mechanisms to the Postgres query samples collector.

The first is designed to capture the active query execution span and wait events more precisely. Whenever we detect a active row, a scrape burst window is opened. The goal is to try to capture the moment when the execution ends as precisely as possible.

The query was changed for improved performance, as `pg_stat_activity` is very lightweight (all data is kept by Postgres in memory), but `pg_blocking_pids` can add some overhead. We will only call the second if there are any `Lock` wait types in first scrape.

The second is designed to avoid a massive amount of samples for queries with a high execution rate. We have diminishing returns for every added sample after we have collected a good number of them. 

The throttling is configurable via the `base_throttle_interval`, that defaults to 1m. This value is then adjusted by a logarithmic factor based on the per minute rate of executions captured for a given queryid in a 5m window. That means  for a `base_throttle_interval` of 1m a query executed 2 to 10 times per minute will have 1 sample sent every `base_throttle_interval*2=2m`, one executed 11 to 100 times every `base_throttle_interval*3=3m`, 101 to 1000 times `base_throttle_interval*4=4m`, and so on.

The throttle exempts samples if it ended with `Idle In Transaction` or `Idle In Transaction (aborted)` states, has any wait events attached, or captured cpu load, given the insightful information they carry.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -94,6 +94,9 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
 |---------------------------|------------|---------------------------------------------------------------|---------|----------|
 | `collect_interval`        | `duration` | How frequently to collect information from database.          | `"15s"` | no       |
 | `disable_query_redaction` | `bool`     | Collect unredacted SQL query text (might include parameters). | `false` | no       |
+| `base_throttle_interval`  | `duration` | Base interval for adaptive throttling of query sample emits. Set to `"0s"` to disable. | `"1m"`  | no       |
```
```suggestion
| Name                      | Type       | Description                                                                            | Default | Required |
| ------------------------- | ---------- | -------------------------------------------------------------------------------------- | ------- | -------- |
| `base_throttle_interval`  | `duration` | Base interval for adaptive throttling of query sample emits. Set to `"0s"` to disable. | `"1m"`  | no       |
| `collect_interval`        | `duration` | How frequently to collect information from database.                                   | `"15s"` | no       |
| `disable_query_redaction` | `bool`     | Collect unredacted SQL query text, This might include parameters.                      | `false` | no       |
```

Alpha sort rows, pretty print table, tweak sentence to remove brackets.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -94,6 +94,9 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+
+`base_throttle_interval` uses a logarithmic backoff based on the per‑minute rate of finalized samples for the same query over a fixed 5‑minute window. This prevents an overly high sampling rate for hot queries. Exempt samples (containing wait events, finalized with idle in transaction states, or with observed CPU load) always emit and do not advance the throttle window.
```
```suggestion
`base_throttle_interval` uses a logarithmic backoff based on the per‑minute rate of finalized samples for the same query over a fixed 5‑minute window.
This prevents an overly high sampling rate for hot queries.
Exempt samples containing wait events, finalized with idle in transaction states, or with observed CPU load always emit and do not advance the throttle window.
```

Removing brackets, add semantic line breaks.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -94,6 +94,15 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
 |---------------------------|------------|---------------------------------------------------------------|---------|----------|
 | `collect_interval`        | `duration` | How frequently to collect information from database.          | `"15s"` | no       |
 | `disable_query_redaction` | `bool`     | Collect unredacted SQL query text (might include parameters). | `false` | no       |
+| Name                      | Type       | Description                                                                            | Default | Required |
```
clayton-cornell, line 93: Something hiccuped in this update...

gaantunes: Fixed

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -249,43 +299,79 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
 		}
+		if sample.WaitEventType.Valid && sample.WaitEventType.String == waitEventTypeLock {
+			hasLockWait = true
+		}
+		buffered = append(buffered, sample)
+	}
+
+	if err := rows.Err(); err != nil {
+		level.Error(c.logger).Log("msg", "failed to iterate pg_stat_activity rows", "err", err)
+		return false, err
```
rgeyer: Is it useful to log this, and return the error? I was doing this a bunch in Explain Plans until I realized that the caller would be logging the error that's returned.

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -249,43 +299,79 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
...
+	}
 
+	// Enrich blocked_by_pids only when there are lock waits
+	blockedByByPID := map[int]pq.Int64Array{}
```
rgeyer: Is `ByBy` a typo, or is there a nuance I'm missing?

rgeyer (approved):
Couple of nit/questions, but approving with the assumption that you'll resolve those or answer that they're by design.

I am not super familiar with the query sample logic, but the code changes here LGTM.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Oh yes good catch, I will change it to return fmt.Errorf

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Good catch.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -90,10 +90,15 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+| ------------------------- | ---------- | -------------------------------------------------------------------------------------- | ------- | -------- |
+| `base_throttle_interval`  | `duration` | Base interval for adaptive throttling of query sample emits. Set to `"0s"` to disable. | `"1m"`  | no       |
+| `collect_interval`        | `duration` | How frequently to collect information from database.                                   | `"15s"` | no       |
+| `disable_query_redaction` | `bool`     | Collect unredacted SQL query text, This might include parameters.                      | `false` | no       |
+
+`base_throttle_interval` uses a logarithmic backoff based on the per‑minute rate of finalized samples for the same query over a fixed 5‑minute window.
+This prevents an overly high sampling rate for hot queries.
+Exempt samples containing wait events, finalized with idle in transaction states, or with observed CPU load, which always emit and do not advance the throttle window.
```
```suggestion
Samples with wait events, idle-in-transaction states, or observed CPU load are exempt from throttling and always emit immediately without advancing the throttle window.
```

This sentence doesn't make sense as written. Is this suggestion what you were trying to say?

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -90,10 +90,15 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
-|---------------------------|------------|---------------------------------------------------------------|---------|----------|
-| `collect_interval`        | `duration` | How frequently to collect information from database.          | `"15s"` | no       |
-| `disable_query_redaction` | `bool`     | Collect unredacted SQL query text (might include parameters). | `false` | no       |
+| Name                      | Type       | Description                                                                            | Default | Required |
+| ------------------------- | ---------- | -------------------------------------------------------------------------------------- | ------- | -------- |
+| `base_throttle_interval`  | `duration` | Base interval for adaptive throttling of query sample emits. Set to `"0s"` to disable. | `"1m"`  | no       |
+| `collect_interval`        | `duration` | How frequently to collect information from database.                                   | `"15s"` | no       |
+| `disable_query_redaction` | `bool`     | Collect unredacted SQL query text, This might include parameters.                      | `false` | no       |
```
```suggestion
| `disable_query_redaction` | `bool`     | Collect unredacted SQL query text. This might include parameters.                      | `false` | no       |
```

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -204,18 +221,21 @@ func (w WaitEventIdentity) Equal(other WaitEventIdentity) bool {
...
+		collectInterval:              args.CollectInterval,
+		entryHandler:                 args.EntryHandler,
+		disableQueryRedaction:        args.DisableQueryRedaction,
+		logger:                       log.With(args.Logger, "collector", QuerySamplesCollector),
+		running:                      &atomic.Bool{},
+		samples:                      map[SampleKey]*SampleState{},
+		idleEmitted:                  expirable.NewLRU[SampleKey, struct{}](emittedCacheSize, nil, emittedCacheTTL),
+		baseThrottleInterval:         args.BaseThrottleInterval,
+		lastEmittedByQueryID:         map[int64]time.Time{},
+		recentFinalizationsByQueryID: map[int64][]time.Time{},
```
fridgepoet: any concern about these growing too much over time?

gaantunes: The first only keeps the last time per queryId, and the seconds is kept as a 5m moving window.
We do have the problem of queryids being not consistent though, which can end up generating multiple entries for the same query and growing this larger than expected. I guess a LRU would not hurt.

clayton-cornell (approved):

PR #4594, opened 2025-10-10
Remove more loki/v3 dependencies

ptodev:
Simplifying Alloy's dependencies. The less modules in go.mod, the better. This PR is a follow up on #4579 and #4576.

I'm pretty sure this will introduce bugfixes to loki.source.file too, but I'm not entirely sure which ones and it's probably not worth it to try document them. It's especially nice that the tail package will now use a more recent fsnotify version.

PR #4773, opened 2025-11-05
chore: cleanup loki wal code

kalleep:
Cleanup after #4594

We just need to keep the old promtail config struct for converters. Rest was either duplicated or could be moved to the wal package we had


ptodev (approved):
Nice, thank you!

PR #4780, opened 2025-11-05
chore: more loki cleanup

kalleep:
Changes:

1. Change to always use `push` package instead of `logproto`
2. Move code in `promtail/util` to `loki/util` and reuse code we have there instead of importing from loki.
3. Move code from `loki/logentry/logql` to `loki/logql`.
4. Only keep promtail config for `consulagent`, we already have the code ported [here](https://github.com/grafana/alloy/tree/main/internal/component/discovery/consulagent)
5. Cleanup util package and only keep what we use
6. Fixed windows test missed in last pr
7. Vendor promtail metrics config so we can use that in converters
8. Vendor flagext.LabelSet so we can remove loki/v3 imports

After this there are 7 imports of `github.com/grafana/loki/v3`. These take a bit more effort but we are close

### DIFF
--- a/internal/static/logs/logs.go
+++ b/internal/static/logs/logs.go
@@ -9,15 +9,9 @@ import (
 	"github.com/grafana/alloy/internal/loki/promtail/server"
 	"github.com/grafana/alloy/internal/loki/promtail/tracing"
 	"github.com/grafana/alloy/internal/loki/promtail/wal"
-	"github.com/grafana/loki/v3/clients/pkg/promtail/client"
 
-	"github.com/grafana/alloy/internal/useragent"
 	_ "github.com/grafana/alloy/internal/util/otelfeaturegatefix" // Gracefully handle duplicate OTEL feature gates
 )
-
-func init() {
-	client.UserAgent = useragent.Get()
-}
 
 // DefaultConfig returns a default config for a Logs instance.
 func DefaultConfig() config.Config {


kalleep:
Have a diff in converter test for `discovery.consulagent`

internal/converter/internal/promtailconvert/internal/build/stages.go
```diff
@@ -488,7 +488,7 @@ func toAlloyMetricsProcessStage(name string, pMetric promtailstages.MetricConfig
 	// Create metric according to type
 	switch strings.ToLower(pMetric.MetricType) {
 	case promtailstages.MetricTypeCounter:
-		pCounter, err := promtailmetric.NewCounters(name, pMetric.Description, pMetric.Config, int64(maxIdle.Seconds()))
+		cfg, err := promtailmetric.ParseCounterConfig(pMetric.Config)
```
ptodev, line 491: Are we still taking `maxIdle` into account?

internal/loki/logql/parser.go
```diff
@@ -45,7 +44,7 @@ func ParseMatchers(input string) ([]*labels.Matcher, error) {
 	}
 	matcherExpr, ok := expr.(*matchersExpr)
 	if !ok {
-		return nil, logqlmodel.ErrParseMatchers
+		return nil, errors.New("only label matchers are supported")
```
ptodev, line 47: Could there be a function somewhere which checks for `ErrParseMatchers`? E.g. if it sees this error it'd ignore it?

internal/converter/internal/promtailconvert/internal/build/stages.go
kalleep: That one did not make any difference. It was not validated but used to construct a [metricsvec](https://github.com/grafana/loki/blob/main/clients/pkg/logentry/metric/metricvec.go#L26)

internal/loki/logql/parser.go
kalleep: No, this function is not even used in our codebase. I keept it here because we could have a usecase for it to remove a loki import but want to test that out in another pr

PR #4781, opened 2025-11-05
feat(alloy): add cloudflare receiver

x1unix:

This PR adds a component wrapping around the Otel Cloudflare receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/cloudflarereceiver

#### Notes to the Reviewer

CC @dehaansa @jharvey10 for a review

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated
- [x] Documentation added


internal/component/otelcol/receiver/cloudflare/cloudflare.go
```diff
@@ -0,0 +1,114 @@
...
+		Stability: featuregate.StabilityExperimental,
+		Args:      Arguments{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			fact := cloudflarereceiver.NewFactory()
+			return receiver.New(opts, fact, args.(Arguments))
+		},
+	})
+}
+
+type Arguments struct {
```
dehaansa, line 35: Missing the `output` argument that defines the next pipeline step.

```
// Output configures where to send processed data. Required.
    Output *otelcol.ConsumerArguments `alloy:"output,block"`
```

internal/component/otelcol/receiver/cloudflare/cloudflare.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Extensions implements receiver.Arguments.
+func (args Arguments) Extensions() map[otelcomponent.ID]otelcomponent.Component {
+	return nil
+}
+
+// NextConsumers implements receiver.Arguments.
+func (args Arguments) NextConsumers() *otelcol.ConsumerArguments {
+	return nil
```
dehaansa: Should map to the `output` argument.

docs/sources/reference/components/otelcol/otelcol.receiver.cloudflare.md
```diff
@@ -0,0 +1,93 @@
...
+    "ClientIP" = "http_request.client_ip"
+    "ClientRequestURI" = "http_request.uri"
+  }
+
+  tls {
+    cert_file = "/path/to/cert.pem"
+    key_file = "/path/to/key.pem"
+  }
+}
+```
```
dehaansa, line 128: There should be a generated section at the bottom talking about compatible components for the `output` argument.

docs/sources/reference/components/otelcol/otelcol.receiver.cloudflare.md
```diff
@@ -0,0 +1,93 @@
...
+| ------------------ | ------------------- | --------------------------------------------------------------------------------------------------------- | ---------------------- | -------- |
+| `endpoint`         | `string`            | The `<HOST:PORT>` endpoint address on which the receiver will await requests from Cloudflare.             |                        | yes      |
+| `secret`           | `string`            | If this value is set, the receiver expects to see it in any valid requests under the `X-CF-Secret` header |                        | no       |
+| `timestamp_field`  | `string`            | Log field name that contains timestamp.                                                                   | `"EdgeStartTimestamp"` | no       |
+| `timestamp_format` | `string`            | One of `unix`, `unixnano`, or `rfc3339`, matching how your LogPush job encodes the timestamp field.       | `"unixnano"`           | no       |
+| `attributes`       | `map[string]string` | Sets log attributes from message fields. Only string, boolean, integer or float fields can be mapped.     |                        | no       |
```
dehaansa: Should add a note underneath that if empty, all fields from the log message will be applied as attributes. It's in the upstream docs.

docs/sources/reference/components/otelcol/otelcol.receiver.cloudflare.md
```diff
@@ -0,0 +1,143 @@
...
```
clayton-cornell:
```suggestion
| Name               | Type                | Description                                                                                                | Default                | Required |
| ------------------ | ------------------- | ---------------------------------------------------------------------------------------------------------- | ---------------------- | -------- |
| `endpoint`         | `string`            | The `<HOST:PORT>` endpoint address on which the receiver awaits requests from Cloudflare.                  |                        | yes      |
| `attributes`       | `map[string]string` | Sets log attributes from message fields. Only string, boolean, integer, or float fields can be mapped.     |                        | no       |
| `secret`           | `string`            | If this value is set, the receiver expects to see it in any valid requests under the `X-CF-Secret` header. |                        | no       |
| `timestamp_field`  | `string`            | Log field name that contains timestamp.                                                                    | `"EdgeStartTimestamp"` | no       |
| `timestamp_format` | `string`            | One of `unix`, `unixnano`, or `rfc3339`, matching how your LogPush job encodes the timestamp field.        | `"unixnano"`           | no       |
```

Minor punctuation, alpha sort rows, and pretty print table

.github/ISSUE_TEMPLATE/blank.yaml
```diff
@@ -120,6 +120,7 @@ body:
       - otelcol.processor.transform
       - otelcol.receiver.awscloudwatch
       - otelcol.receiver.awsecscontainermetrics
+      - otelcol.receiver.cloudflare
```
jharvey10, line 123: Maybe more of a question for @dehaansa

Does the automation you wrote update these automatically? Or maybe that's what was used here already to include the new component in the templates?

internal/component/otelcol/receiver/cloudflare/cloudflare.go
```diff
@@ -0,0 +1,117 @@
...
+	Output *otelcol.ConsumerArguments `alloy:"output,block"`
+}
+
+// SetToDefault implements syntax.Defaulter.
+func (args *Arguments) SetToDefault() {
+	// Although otel's receiver already initializes defaults of downstream config,
+	// let's do it as well to avoid breaking changes if defaults are changed in upstream.
+	if args.TimestampField == "" {
+		args.TimestampField = "EdgeStartTimestamp"
+	}
```
jharvey10: I could go either way on this/these. Generally, (since this is experimental stability and since the upstream dep is a 0.x release) even if OTel includes a breaking change, we will "blindly" carry that through to maintain the behavior that upstream has (with a note in the changelog).

@dehaansa @ptodev any thoughts on this?

.github/ISSUE_TEMPLATE/blank.yaml
x1unix: @jharvey10 I had to manually add this as one of CI jobs checks presence of each component.

internal/component/otelcol/receiver/cloudflare/cloudflare.go
x1unix: @jharvey10 makes sense, then I'll let upstream to deal with defaults.

I have only one concern - component docs mention defaults for each attribute.\
We should track any update the docs every time upstream changes defaults.

.github/ISSUE_TEMPLATE/blank.yaml
jharvey10: Ahh okay. Was that using the `make generate-gh-issue-templates` command?

internal/component/otelcol/receiver/cloudflare/cloudflare.go
jharvey10: Definitely agree that we should track those changes. Typically when we upgrade otel each Alloy minor release, there's an exhaustive look through all of the changelogs and we pull out any deviations/breaking changes and call those out as needed in the changelog and docs). Our BAU process should cover what's needed for what you suggested to happen.

x1unix: @jharvey10 addressed in https://github.com/grafana/alloy/pull/4781/commits/c56e42e0902f85b9afe0fdd2404fbcebf5d947dc

PR #4782, opened 2025-11-05
database_observability: handle null server_ip/port in server_id generation

matthewnolf:
#### PR Description
To generate the `server_id` for a postgres instance, we rely the following internal functions:
* `pg_control_system`
* `inet_server_addr`
* `inet_server_port`

`inet_server_addr` and `inet_server_port` will return NULL when a UDP socket is used for the DB connection, as stated in [PG docs](https://www.postgresql.org/docs/current/functions-info.html)

> Returns the IP address on which the server accepted the current connection, or NULL if the current connection is via a Unix-domain socket.

As a safety fallback, instead of erroring, we safely handle the NULL value and instead rely solely on the `pg_control_system` to generate the server_id.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


egovova11 (approved):

internal/component/database_observability/postgres/component.go
```diff
@@ -284,19 +284,19 @@ func (c *Component) Update(args component.Arguments) error {
...
 
-	var systemID, systemIP, systemPort, engineVersion string
+	var systemID, systemIP, systemPort, engineVersion sql.NullString
 	if err := rs.Scan(&systemID, &systemIP, &systemPort, &engineVersion); err != nil {
 		c.reportError("failed to scan engine version", err)
 		return nil
 	}
 
-	systemID = fmt.Sprintf("%x", sha256.Sum256([]byte(fmt.Sprintf("%s:%s:%s", systemID, systemIP, systemPort))))
+	generatedSystemID := fmt.Sprintf("%x", sha256.Sum256([]byte(fmt.Sprintf("%s:%s:%s", systemID.String, systemIP.String, systemPort.String))))
```
fridgepoet, line 293: didn't really read into it, but will everything depending on this generated system ID be OK when the ID and Port are both null (empty string)? we won't have issues with the same generated ID?

matthewnolf: The reason we use `systemID` + `IP` + `port` is that clustered postgres instances share the `systemID` (not unique per instance within a cluster), so the `IP` and `port` add uniqueness. 

Taking this away means that clustered postgres instances may share a `server_id` - with all the issues that causes around aggregation. 
However this only will happen when a unix domain socket is used - whereas today the component fails.

I don't think this is a long term solution, but it stops the worst thing happening where all UDP connections fail

fridgepoet (approved):

Issue #4728, opened 2025-10-30
Loki.Write: implement queue_config


kalleep:
### Component(s)

loki.write

### Add a description

`loki.write` only have one “sender” per endpoint, meaning that the full log pipeline will be blocked when when we try to send a batch. This can be problematic in cases where we have backpressure  e.g. https://github.com/grafana/alloy/issues/4508#issuecomment-3360638788. 

We should implement sharding of batches and have multiple writers similar to what we have in [prometheus.remote_write](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#queue_config)

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

martensson:
Just discovered this issue, and it is pretty critical for resilience with multiple loki endpoints. 

In our case we have two Loki clusters in Production, and when one of them was unavailable we started to loose all logs on the other cluster as well. Alloy should split the pipeline between the different endpoints to avoid blocking and loosing all logs from being forwarded.

Current workaround if anyone sees this is to split the sources as well, and let each source have its own loki.write. Not the prettiest of config, but it works!

PR #4788, opened 2025-11-06
loki: implement queue_config for loki.write

kalleep:
#### PR Description
This PR implements `queue_config` for the `loki.write` component, enabling users to configure queue-based batching and parallel processing. The implementation introduces a new sharding architecture that distributes log entries across multiple parallel queues based on label fingerprints that is based on [Prometheus rw sharding](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L1235).

The shards implementation is used with both "normal" client and `queue_client`. So we no longer need to duplicate the same logic in both implementations.

Before this pr we had a `queue_config` block that was only used when `WAL` was enabled. It is now always used.

Currently no automatic "resharding" is implemented. Implementing this without the WAL will most likely be pretty primitive. So for now `min_shards` is the only configurable value until we address this.

Ideally we would move a couple of attributes from `endpoint` block to `queue_config` block to closer match `prometheus.remote_write`. But we can't do that without doing a breaking change. These attributes are:

* `retry_on_http_429`
* `max_backoff_period`
* `min_backoff_period`
* `batch_size`
* `batch_wait`

Part of: https://github.com/grafana/alloy/issues/4728

#### Notes to the Reviewer

I moved wal writer ownership into client.Manager. No need to expose it to `loki.write` component.
I plan to work on resharding in followup pr

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.write.md
+++ b/docs/sources/reference/components/loki/loki.write.md
@@ -128,7 +128,7 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
 
 ### `queue_config`
 
-The optional `queue_config` block configures, how the underlying client queues batches of logs sent to Loki.
+The optional `queue_config` block configures, how the endpoint queues batches of logs sent to Loki.
 
 The following arguments are supported:
 


docs/sources/reference/components/loki/loki.write.md
```diff
@@ -104,8 +104,9 @@ The following arguments are supported:
 If no `tenant_id` is provided, the component assumes that the Loki instance at `endpoint` is running in single-tenant mode and no X-Scope-OrgID header is sent.
 
 When multiple `endpoint` blocks are provided, the `loki.write` component creates a client for each.
-Received log entries are fanned-out to these clients in succession.
-That means that if one client is bottlenecked, it may impact the rest.
+Received log entries are fanned-out to these clients in succession. That means that if one client is bottlenecked, it may impact the rest.
```
thampiotr, line 107: We introduce the word 'client' here, while it's an unnecessary implementation detail I think - we can just talk about 'endpoints' and queues which each has maybe a 'worker' that sends the logs to their destination.

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
 
 ### `queue_config`
 
-{{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
-
-The optional `queue_config` block configures, when WAL is enabled, how the underlying client queues batches of logs sent to Loki.
-Refer to [Write-Ahead block](#wal) for more information.
+The optional `queue_config` block configures, how the underlying client queues batches of logs sent to Loki.
```
```suggestion
The optional `queue_config` block configures, how the endpoint queues batches of logs sent to Loki.
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
 
 The following arguments are supported:
 
 | Name            | Type       | Description                                                                                                                                                                   | Default | Required |
 | --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
 | `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered a worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB` | no       |
 | `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it enqueues pending batches and drains the send queue sending each.  | `"1m"`  | no       |
```
thampiotr, line 138:
```suggestion
| `drain_timeout` | `duration` | Configures the maximum time the endpoint can take to drain its send queue upon shutdown. | `"1m"`  | no       |
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
+| `min_shards`    | `number`   | Minimum amount of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
```
thampiotr, line 139:
```suggestion
| `min_shards`    | `number`   | Minimum number of concurrent workers sending logs to the endpoint.                                                                                                          | `1`      | no       |
```
Maybe we should also stop using term 'shard' and use 'worker'?

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
+
+Each queue manages a number of concurrent _shards_ which is responsible for sending a fraction of batches to their respective endpoints.
```
thampiotr, line 141:
```suggestion
Each queue manages a number of concurrent _workers_ which are responsible for sending batches to their respective endpoints.
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
+Each shard has a buffer of batches it keeps in memory, controlled with the `capacity` argument.
```
thampiotr, line 142: - each queue manages concurrent shards
- each shard has buffer in memory controlled by capacity
- capacity is defined as the size of queue buffer

so what is the queue? is it the thing that manages the shards that have buffers? or is it the buffer inside of each shard and there is some kind of manager above them?

We need to document it in a way that will make the answers to above questions clear :) 

This doc already had some of these issues, but I think we need to clean this up and make it easier to understand with this PR.

docs/sources/reference/components/loki/loki.write.md
kalleep: I used the same wording that we do in [prometheus.remote_write](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#queue_config)

docs/sources/reference/components/loki/loki.write.md
kalleep: Ah yeah this is wrong, each shard has it's own queue

docs/sources/reference/components/loki/loki.write.md
kalleep: Same as [this](https://github.com/grafana/alloy/pull/4788#discussion_r2527199661)

Issue #4744, opened 2025-10-31
Converters: change to use file_match block in loki.source.file


kalleep:
### Component(s)

loki.source.file

### Add a description

When converting code from either static or promtail we will use `local.file_match`. We should change to use `file_match` block instead after https://github.com/grafana/alloy/pull/4743 is merged

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #4791, opened 2025-11-06
feat: update converter to use file_match block

kalleep:
#### PR Description
Now that https://github.com/grafana/alloy/pull/4743 is merged we can update converters to use `file_match` block instead of going through `local.file_match`.

fixes: #4744

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Enhancements
+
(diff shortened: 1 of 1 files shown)


kalleep:
> I know it's not considered experimental, but do we need to do more dogfooding before rolling it out as a default more widely?

Sure we can target next release with this specific change if you are more comfortable with it?

ptodev:
Targeting next release would be nice. Especially if for some reason users are translating the config on the latest version, but still expect the previous Alloy version to work with the translated config.

PR #4794, opened 2025-11-06
Better support for tailing non-UTF-8 files

### LINKED ISSUE #217: loki.source.file component - automatically determine file encoding
### Request

loki.source.file could automatically determine the file encoding instead of requiring it to be set as one of the arguments.

UTF-16 files have a BOM, byte order mark, as the first two bytes of a file that could be used to automatically determine the encoding.  Linux and Windows do have other ways to determine the file encoding.

### Use case

We don't always know what the encoding of some files are, and users can change the encoding without notifying us.  
SSRS in Windows create some log files with UTF-16 LE BOM instead of the normal UTF-8 encoded files in the same directory.
In our environment, different machines have different encodings for some of the log files.

ptodev:
#### PR Description

The tailer used to convert the encoding of each line individually. This doesn't work for files with a BOM, since the BOM is only on the very first line. This PR changes the tailers to open the file as a whole with the correct encoding. It also fixes an issue with Windows logs which caused a carriage return to still be visible in the final log line. This is now stripped, just like the standard new line character.

Fixes #217

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/file_test.go
+++ b/internal/component/loki/source/file/file_test.go
@@ -415,10 +415,11 @@ func TestEncoding(t *testing.T) {
 					DataPath:      t.TempDir(),
 				}
 
-				filePath := filepath.Join("testdata", "encoding", tc.filename)
+				filePath, err := filepath.Abs(filepath.Join("testdata", "encoding", tc.filename))
+				require.NoError(t, err)
 
 				// Verify the test data file exists
-				_, err := os.Stat(filePath)
+				_, err = os.Stat(filePath)
 				require.NoError(t, err, fmt.Sprintf("%s test file should exist in testdata/encoding/", tc.filename))
 
 				ch1 := loki.NewLogsReceiver()


internal/component/loki/source/file/internal/tail/tail.go
```diff
@@ -468,11 +478,20 @@ func (tail *Tail) finishDelete() error {
...
+			tail.reader = bufio.NewReader(tail.Decoder.Reader(tail.file))
+		}
 	} else {
-		tail.reader = bufio.NewReader(tail.file)
+		if tail.MaxLineSize > 0 {
+			// add 2 to account for newline characters
+			tail.reader = bufio.NewReaderSize(tail.file, tail.MaxLineSize+2)
+		} else {
+			tail.reader = bufio.NewReader(tail.file)
+		}
```
kalleep:
```suggestion
	var reader io.Reader
	if tail.Decoder != nil {
		reader = tail.Decoder.Reader(tail.file)
	} else {
		reader = tail.file
	}

	if tail.MaxLineSize > 0 {
		// add 2 to account for newline characters
		tail.reader = bufio.NewReaderSize(reader, tail.MaxLineSize+2)
	} else {
		tail.reader = bufio.NewReader(reader)
	}

```

internal/component/loki/source/file/file_test.go
```diff
@@ -352,70 +351,154 @@ func TestTwoTargets(t *testing.T) {
...
+		{"LF UTF-16 BE (gzipped)", "/LF/UTF-16_BE.txt.gz", "UTF-16BE", gzDecompress},
+		{"LF UTF-16 LE with BOM (gzipped)", "/LF/UTF-16_LE_BOM.txt.gz", "UTF-16", gzDecompress},
+		{"LF UTF-16 BE with BOM (gzipped)", "/LF/UTF-16_BE_BOM.txt.gz", "UTF-16", gzDecompress},
+	}
 
-		// Create and run the component.
-		c, err := New(opts, args)
-		require.NoError(t, err)
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
kalleep: Can we add back `runTests` here? This will run all tests with `file_match` enabled and disabled, just to verify that everything works correctly using both.

kalleep (commented):
Nice, just two comments

PR #4799, opened 2025-11-07
database_observability: add table registry cache to schema_details and add "validated" to parsed table name logs

fridgepoet:
#### PR Description
This PR introduces:
* a table registry which caches the databases > schemas > tables from the Schema Details collector
* parsed table names from the query_details collector are validated against this table registry
  * logs in op=query_parsed_table_name now emitted with an additional boolean field "validated"
  * this additional field will be used in the frontend

The Schema Details collector must be enabled in order for table names to be "validated".

Sample logs:
```
2025-11-07 10:56:56.419infolevel="info" queryid="2541281651416271417" datname="books_store" table="book_mix" engine="postgres" validated="false"

2025-11-07 10:56:56.419infolevel="info" queryid="2541281651416271417" datname="books_store" table="reading_list_books" engine="postgres" validated="true"
```

Contributes to https://github.com/grafana/grafana-dbo11y-app/issues/1574

#### Notes to the Reviewer
See also https://github.com/grafana/grafana-dbo11y-app/issues/1838

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/schema_details.go
+++ b/internal/component/database_observability/postgres/collector/schema_details.go
@@ -168,9 +168,9 @@ const (
 )
 
 type tableInfo struct {
-	database     string
-	schema       string
-	tableName    string
+	database     database
+	schema       schema
+	tableName    table
 	updateTime   time.Time
 	b64TableSpec string
 }
@@ -206,37 +206,41 @@ type foreignKey struct {
 	ReferencedColumnName string `json:"referenced_column_name"`
 }
 
+type database string
+type schema string
+type table string
+
 // TableRegistry is a source-of-truth cache that keeps track of databases, schemas, tables
 type TableRegistry struct {
 	mu     sync.RWMutex
-	tables map[string]map[string]map[string]bool // map[database]map[schema]map[table]bool
+	tables map[database]map[schema]map[table]bool
 }
 
 func NewTableRegistry() *TableRegistry {
 	return &TableRegistry{
-		tables: make(map[string]map[string]map[string]bool),
+		tables: make(map[database]map[schema]map[table]bool),
 	}
 }
 
-func (tr *TableRegistry) SetTablesForDatabase(database string, tablesInfo []*tableInfo) {
+func (tr *TableRegistry) SetTablesForDatabase(database database, tablesInfo []*tableInfo) {
 	tr.mu.Lock()
 	defer tr.mu.Unlock()
 
 	delete(tr.tables, database)
 
 	if len(tablesInfo) > 0 {
-		tr.tables[database] = make(map[string]map[string]bool)
-		for _, table := range tablesInfo {
-			if tr.tables[database][table.schema] == nil {
-				tr.tables[database][table.schema] = make(map[string]bool)
+		tr.tables[database] = make(map[schema]map[table]bool)
+		for _, tblInfo := range tablesInfo {
+			if tr.tables[database][tblInfo.schema] == nil {
+				tr.tables[database][tblInfo.schema] = make(map[table]bool)
 			}
-			tr.tables[database][table.schema][table.tableName] = true
+			tr.tables[database][tblInfo.schema][tblInfo.tableName] = true
 		}
 	}
 }
 
 // IsValid returns whether or not a given database and parsed table name exists in the source-of-truth table registry
-func (tr *TableRegistry) IsValid(database, parsedTableName string) bool {
+func (tr *TableRegistry) IsValid(database database, parsedTableName string) bool {
 	tr.mu.RLock()
 	defer tr.mu.RUnlock()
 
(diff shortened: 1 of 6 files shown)


internal/component/database_observability/postgres/component.go
```diff
@@ -368,20 +368,47 @@ func (c *Component) startCollectors(systemID string, engineVersion string) error
...
 			EntryHandler:    entryHandler,
+			TableRegistry:   tableRegistry,
 			Logger:          c.opts.Logger,
 		})
 		if err != nil {
 			logStartError(collector.QueryDetailsCollector, "create", err)
 		}
-		if err := qCollector.Start(context.Background()); err != nil {
-			logStartError(collector.QueryDetailsCollector, "start", err)
+		if qCollector != nil {
```
fridgepoet: Added nil checks for all the collectors before `.Start` is called on each one.
Happy to move this to a separate PR, let me know.

internal/component/database_observability/postgres/component.go
```diff
@@ -366,20 +366,47 @@ func (c *Component) startCollectors(systemID string, engineVersion string) error
...
+			CacheSize:       c.args.SchemaDetailsArguments.CacheSize,
+			CacheTTL:        c.args.SchemaDetailsArguments.CacheTTL,
+			EntryHandler:    entryHandler,
+			Logger:          c.opts.Logger,
+		})
+		if err != nil {
+			logStartError(collector.SchemaDetailsCollector, "create", err)
+		}
+		if stCollector != nil {
+			tableRegistry = stCollector.GetTableRegistry()
```
fridgepoet: tableRegistry needs to be reassigned here by the Schema Details collector in the case that it is enabled so that the Query Details collector receives an updated reference.

Tried to add a test to protect this order, but could not successfully inject a mocked DB (have run into this problem before). Happy to take suggestions or look into this as a follow-up.

internal/component/database_observability/postgres/component.go
matthewnolf: Why don't we just if/else on if the New[Collector] fails.
If err != nil {
  log and continue
} else {
  start and add to map
}

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -205,6 +206,72 @@ type foreignKey struct {
 	ReferencedColumnName string `json:"referenced_column_name"`
 }
 
+// TableRegistry is a source-of-truth cache that keeps track of databases, schemas, tables
+type TableRegistry struct {
```
matthewnolf, line 214: nit: If it's keeping hold of more than just tables, why do we call it a TableRegistry?

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -205,6 +206,72 @@ type foreignKey struct {
...
+	mu     sync.RWMutex
+	tables map[string]map[string]map[string]bool // map[database]map[schema]map[table]bool
```
matthewnolf: Can you create alias types if it would help for readability on what this represents

matthewnolf: You can just use empty struct `struct{}` instead of `bool`, if we're only testing for presence

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -205,6 +206,72 @@ type foreignKey struct {
...
+
+	schemas, ok := tr.tables[database]
+	if !ok {
+		return false
+	}
+
+	schemaName, tableName := parseSchemaQualifiedIfAny(parsedTableName)
+	switch schemaName {
+	case "": // parsedTableName isn't schema-qualified, e.g. SELECT * FROM table_name.
+		// table name can only be validated as "exists somewhere in the database", see limitation: https://github.com/grafana/grafana-dbo11y-app/issues/1838
```
matthewnolf: This is a private project issue, is there somewhere else this can be tracked

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -205,6 +206,72 @@ type foreignKey struct {
...
+		for _, tables := range schemas {
+			if tables[tableName] {
```
matthewnolf: This is slightly misleading though? A table could exist in one schema (A), but not (B). We could return that a table exists in (B) because we don't validate the schema.

Can we not get that from the original query/connection?

fridgepoet: I don't think so, could I be missing something?
(But yes, I ran into this and considered this a limitation but acceptable trade-off which is why I created the [issue](https://github.com/grafana/alloy/issues/4815) above. Off the top of my head, I don't think we'll be able to to solve this, but this is something we can explore more in-depth in the future.)

internal/component/database_observability/postgres/collector/schema_details.go
fridgepoet: DatabaseSchemaTableRegistry?

internal/component/database_observability/postgres/collector/schema_details.go
fridgepoet: Sure, I can move that to `alloy` if this is what we decide on (and once it's merged)

internal/component/database_observability/postgres/component.go
fridgepoet: I think I will revert the suggestion and we can defer these changes to another PR.
Matching what is existing for now.
I would like to avoid introducing unintentional changes to the collectors slice at the moment.

internal/component/database_observability/postgres/collector/schema_details.go
matthewnolf: Discussed this, and it is what's highlighted in the GH Issue

matthewnolf (approved):
LGTM - please add an entry to the CHANGELOG

PR #4802, opened 2025-11-07
chore: cleanup tail and fix locks

kalleep:
#### PR Description
After we moved https://github.com/grafana/tail into alloy I did a quick review of the code.

I did not want to change too much of how it functions in this pr, maybe we can do that later.

But I wanted to remove functionality we do not use:
* Removed logger interface and take go-kig logger and change to debug level
* Remove rate limit implementation - this is not something we configure ever
* Remove inotify implementation of watcher - we always use poll
* Remove MaxLineSize - nothing we use
* Remove ReOpen - we always configure this so this is now the only behavior 
* Remove Follow - we always configure this so this is now the only behavior 
* Remove MustExist - we always configure this so this is now the only behavior
* Remove Pipe - we never use this

I renamed lk to readerMut, this is the mutex used for accessing reader and updated how locks are done in `Tell` and `Size`, these locks needs to be held while we want to use these resources. There are more stuff to look into here regarding locking but it's easier to do that with less changes so it can be done in a followup pr.

I also rewrote the tests to avoid races there.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/tail.go
+++ b/internal/component/loki/source/file/internal/tail/tail.go
@@ -391,7 +391,6 @@ func (tail *Tail) finishDelete() error {
 func (tail *Tail) openReader() {
 	tail.readerMut.Lock()
 	defer tail.readerMut.Unlock()
-	tail.reader = bufio.NewReader(tail.file)
 	if tail.Decoder != nil {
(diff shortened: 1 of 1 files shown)


ptodev (approved):
Thank you for the detailed description! It really helps me when I'm reviewing it.

PR #4805, opened 2025-11-07
database_observability: Fix a few small things discovered in dogfooding

rgeyer:
##### Explain Plan prepared statements for queries without params
When a query has no parameters, I need to execute the prepared statement I'm creating WITHOUT parens. I.E. `EXPLAIN (FORMAT JSON) EXECUTE explain_plan_123` vs `EXPLAIN (FORMAT JSON) EXECUTE explain_plan_123()`

##### Missed reserved words which can not appear in an explain request
Added a few more keywords to the denylist which were not caught previously

##### Broader denylisting of queries when errors are encoutered
Specifically *any* permissions error, and syntax errors.

This will result in any query with a syntax error to be logged as an error only once the first time it is tried, then the query won't be attempted again unless alloy is restarted.

Fixes grafana/grafana-dbo11y-app#1787

### DIFF
--- a/internal/component/database_observability/postgres/collector/explain_plan.go
+++ b/internal/component/database_observability/postgres/collector/explain_plan.go
@@ -521,8 +521,7 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
 
 	explainQuery := fmt.Sprintf("%s%s", selectExplainPlanPrefix, preparedStatementName)
 	paramCount := len(paramCountRegex.FindAllString(qi.queryText, -1))
-	if paramCount == 0 {
-	} else {
+	if paramCount > 0 {
 		nullParams := strings.Repeat("null,", paramCount)
 		if paramCount > 0 {
 			nullParams = nullParams[:len(nullParams)-1]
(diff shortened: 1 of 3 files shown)


internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -518,14 +519,17 @@ func (c *ExplainPlan) fetchExplainPlanJSON(ctx context.Context, qi queryInfo) ([
 		return nil, fmt.Errorf("failed to set plan cache mode: %w", err)
 	}
 
+	explainQuery := fmt.Sprintf("%s%s", selectExplainPlanPrefix, preparedStatementName)
 	paramCount := len(paramCountRegex.FindAllString(qi.queryText, -1))
+	if paramCount == 0 {
```
gaantunes: Not sure I got the reason for this if else here. wouldn't it be better to just do `if paramcount > 0`?

rgeyer: Uggh.. yes.. I at first thought I'd need an if/else, but then defaulted to just an if logic, but forgot to remove it. 🤦‍♂️

gaantunes (approved):

PR #4810, opened 2025-11-10
Fix: Address Remaining Stringlabels Issues Before Enabling

blewis12:
#### PR Description
This PR addresses the remaining stringlabel issues seen in alloy. They are fixed already in[ this PR](https://github.com/grafana/alloy/pull/4756) but I am cherry picking them to this separate PR so that they can be reviewed and merged separately from the stringlabel enablement itself. 

This PR includes the following changes

* Remaining compilation errors from incompatible stringlabel struct/slice types
* Hashing fun: We currently [copy over](https://github.com/grafana/alloy/blob/5bf2d4fd22563a2cb16aab95cca6483d075c67e6/internal/component/discovery/target.go#L315) the prometheus labels hashing function (modifying its input data type slightly to fit our data structures) and use that to hash prometheus labels for our Targets. However, stringlabels changes the the way the labels are represented in memory and thus the output of the labels.Hash() function, which causes some tests that that are checking the output of this function against a specific hash to fail. The reason this breaks is because, in slicelabels, we store labels as key/value pairs (Label struct) in a slice. When this is converted into a byte string to be passed into the xxHash function, it is represented as the label values separated by the 0xFF char. For stringlabels, we store the labels already as an entire string within the Labels struct, containing label values preceded/separated by length encodings. This means that the byte slice we eventually pass through to xxHash is different when we enable stringlabels. We now use the `StableHash()` function, which was [introduced](https://github.com/prometheus/prometheus/commit/501bc6419ee54ee5617d75c3bc16ecf36d94c552) to address this backwards incompatibility, in cases where we are computing target hashes to ensure we don't break clustering behavior when rolling out changes. 

The remaining places we use the backwards incompatible `Hash()` function are in the WAL and LabelStore - but these are cleared on restart, and so we shouldn't have issues with mixed states

Screenshot 2025-11-10 at 09 10 50

Part of https://github.com/grafana/alloy/issues/4323

#### PR Checklist
- [x] Tests updated

### DIFF
--- a/internal/component/pyroscope/scrape/target.go
+++ b/internal/component/pyroscope/scrape/target.go
@@ -97,7 +97,7 @@ func NewTarget(lbls labels.Labels, params url.Values) *Target {
 	url := urlFromTarget(lbls, params)
 
 	h := fnv.New64a()
-	_, _ = h.Write([]byte(strconv.FormatUint(publicLabels.Hash(), 16)))
+	_, _ = h.Write([]byte(strconv.FormatUint(labels.StableHash(publicLabels), 16)))
 	_, _ = h.Write([]byte(url))
 
 	return &Target{


internal/component/discovery/label_hasher_slicelabels.go
```diff
@@ -0,0 +1,47 @@
...
+
+var (
+	seps = []byte{'\xff'}
+)
+
+// NOTE 1: This function is copied from Prometheus codebase and adapted to work correctly with Alloy types.
+// NOTE 2: It is important to keep the hashing function consistent between Alloy versions in order to have
+//
+//	smooth rollouts without duplicated or missing data. There are tests to verify this behaviour. Do not change it.
+func HashLabelsInOrder(t Target, order []string) uint64 {
```
kalleep: Nit: these functions are internal to this package right? So we should not export them

blewis12: ah yes, missed that: https://github.com/grafana/alloy/pull/4810/commits/30519f47c2a0567633a7851f20a3975bf6a77230

blewis12:
> I don't think hashing for labelstore or the wal will be impacted if it's unstable between versions

That's been my major question for the day, and was pondering about how this might impact WAL. Seems like not though, in which case I will switch those back to the previous `.Hash()` call 👍 

There is definitely a performance implication using StableHash, as we need to [decode the label string](https://github.com/prometheus/prometheus/blob/e47c7e2f96576ade4005d6cc5e60b7c9a32af4a4/model/labels/sharding_stringlabels.go#L30-L31) first before we can hash it

PR #4814, opened 2025-11-10
database_observability: Keep evaluating queries for reserved words after finding one

rgeyer:
#### PR Description
Also add `TRANSACTION` as a keyword.

### DIFF
--- a/internal/component/database_observability/lexer.go
+++ b/internal/component/database_observability/lexer.go
@@ -34,7 +34,6 @@ func isCommandKeywordOrIdentifier(token sqllexer.Token) bool {
 // ContainsReservedKeywords checks if the SQL query contains any reserved keywords
 // that indicate write operations, excluding those in string literals or comments
 func ContainsReservedKeywords(query string, reservedWords map[string]ExplainReservedWordMetadata, dbms sqllexer.DBMSType) (bool, error) {
-	result := false
 	// Use the lexer to tokenize the query
 	lexer := sqllexer.New(query, sqllexer.WithDBMS(dbms))
 	tokenBuffer := make([]sqllexer.Token, 0)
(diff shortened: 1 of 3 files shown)


internal/component/database_observability/lexer_test.go
```diff
@@ -482,6 +482,18 @@ func TestContainsReservedKeywords(t *testing.T) {
 			reservedWords: ExplainReservedWordDenyList,
 			expected:      false,
 		},
+		{
+			name:          "FOR UPDATE preceds UPDATE",
```
matthewnolf:
```suggestion
			name:          "FOR UPDATE precedes UPDATE",
```

fridgepoet: precedes?

fridgepoet (approved):

internal/component/database_observability/lexer.go
```diff
@@ -72,16 +73,15 @@ func ContainsReservedKeywords(query string, reservedWords map[string]ExplainRese
 							}
 						}
 					}
-					if matchedTokens == lookbackCount {
-						return false, nil
+					if matchedTokens < lookbackCount {
```
matthewnolf, line 75: Result is always set to true here, why not return early

rgeyer: In the case of true, I suppose this is fine. I was just avoiding that since returning false early was the cause of this issue. 🤔

internal/component/database_observability/lexer_test.go
rgeyer: I R SPEL GUD

internal/component/database_observability/lexer.go
rgeyer: Made the change, saves some compute cycles not evaluating the rest of the query if a denied word has been found. :)

PR #4818, opened 2025-11-11
loki.source.api: cancel in-flight requests when graceful shutdown period has expired

kalleep:
#### PR Description
`loki.source.api` was never terminating requests when shutting down. It performed graceful shutdown of the server and that includes stopping new request from being accepted and closing idle connections. But in-flight request would not be terminated. That would only happen after all entries have been pushed down the pipeline.

So instead of that we can respect the graceful timeout and then terminate all running request with 503. Causing request retries on caller side. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/api/internal/lokipush/push_api_server.go
+++ b/internal/component/loki/source/api/internal/lokipush/push_api_server.go
@@ -117,11 +117,15 @@ func (s *PushAPIServer) ServerConfig() fnet.ServerConfig {
 
 func (s *PushAPIServer) Shutdown() {
 	level.Info(s.logger).Log("msg", "stopping push API server")
-	// StopAndShutdown tries to gracefully shutdown server.
-	// No in-flight requests will be canceled.
+	// StopAndShutdown tries to gracefully shutdown.
+	// It will stop idle and incoming connections
+	// and try to wait for all in-flight connections
+	// to finish. If configured timeout `ServerGracefulShutdownTimeout`
+	// expired this call will be unblocked.
 	s.server.StopAndShutdown()
 
-	// After we have tried a graceful shutdown we cancel all in-flight requests.
+	// After we have tried a graceful shutdown we force all remaining in-flight
+	// requests to exit.
 	close(s.forceShutdown)
 }
 


docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -50,13 +50,14 @@ The component starts an HTTP server on the configured port and address with the
...
+| --------------------------- | -------------------- | ---------------------------------------------------------------------------------- | ---------- | -------- |
+| `forward_to`                | `list(LogsReceiver)` | List of receivers to send log entries to.                                          |            | yes      |
+| `labels`                    | `map(string)`        | The labels to associate with each received logs record.                            | `{}`       | no       |
+| `relabel_rules`             | `RelabelRules`       | Relabeling rules to apply on log entries.                                          | `{}`       | no       |
+| `use_incoming_timestamp`    | `bool`               | Whether to use the timestamp received from request.                                | `false`    | no       |
+| `max_send_message_size`     | `size`               | Maximum size of a request to the push API.                                         | `"100MiB"` | no       |
+| `graceful_shutdown_timeout` | `duration`           | Timeout for servers graceful shutdown. If configured, should be greater than zero. | `"30s"`    | no       |
```
thampiotr:
```suggestion
| `graceful_shutdown_timeout` | `duration`           | Timeout for server's graceful shutdown. If configured, should be greater than zero. | `"30s"`    | no       |
```

internal/component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -33,7 +33,8 @@ type PushAPIServer struct {
 	logger       log.Logger
 	serverConfig *fnet.ServerConfig
 	server       *fnet.TargetServer
-	handler      loki.EntryHandler
+	handler      loki.LogsReceiver
+	hardClose    chan struct{}
```
thampiotr: Do you like the name `forceShutdown`? I think it would work a bit better.

internal/component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -115,7 +117,13 @@ func (s *PushAPIServer) ServerConfig() fnet.ServerConfig {
 
 func (s *PushAPIServer) Shutdown() {
 	level.Info(s.logger).Log("msg", "stopping push API server")
+	// StopAndShutdown tries to gracefully shutdown server.
+	// No inflight requests will be canceled.
 	s.server.StopAndShutdown()
+
+	// After timeout has expired we cancel all inflight requests.
+	<-time.After(s.serverConfig.GracefulShutdownTimeout)
```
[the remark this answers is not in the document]
thampiotr: I think this one has a point :)

CHANGELOG.md
```diff
@@ -89,6 +89,8 @@ Main (unreleased)
 
 - Use POST requests for remote config requests to avoid hitting http2 header limits. (@tpaschalis)
 
+- `loki.source.api` now cancels all inflight requests with 503 after `graceful_shutdown_timeout` has expired. (@kalleep)
```
```suggestion
- `loki.source.api` during component shutdown will now reject all the inflight requests with status code 503 after `graceful_shutdown_timeout` has expired. (@kalleep)
```

internal/component/loki/source/api/internal/lokipush/push_api_server.go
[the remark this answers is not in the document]
kalleep: Yes for sure, did not think of that. But the suggested fix do not work. Need to think of something else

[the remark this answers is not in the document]
kalleep: Ah the wait is not needed. [s.server.StopAndShutdown()](https://github.com/grafana/dskit/blob/main/server/server.go#L704) calls Shutdown on the http server with a context using `ServerGracefulShutdownTimeout`. 

From docs:
> Shutdown gracefully shuts down the server without interrupting any
active connections. Shutdown works by first closing all open
listeners, then closing all idle connections, and then waiting
indefinitely for connections to return to idle and then shut down.
If the provided context expires before the shutdown is complete,
Shutdown returns the context's error, otherwise it returns any

So we can always force close after this call have retruned.

internal/component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -115,7 +117,12 @@ func (s *PushAPIServer) ServerConfig() fnet.ServerConfig {
 
 func (s *PushAPIServer) Shutdown() {
 	level.Info(s.logger).Log("msg", "stopping push API server")
+	// StopAndShutdown tries to gracefully shutdown server.
+	// No in-flight requests will be canceled.
 	s.server.StopAndShutdown()
```
thampiotr, line 125: Is this a blocking call? What is the timeout for this? Can you add a comment?

PR #4819, opened 2025-11-11
database_observability: add database exclude list for Postgres

fridgepoet:
#### PR Description

This PR adds an exclusion list for databases in the Postgres schema_details collector excluding "azure_maintenance".

Observed logs like the following on AKS:

```
2025-11-03 18:10:48.844ERRORts=2025-11-03T18:10:48.844511627Z level=error msg="failed to analyze table spec" component_path=/ component_id=database_observability.postgres.postgres_xxxx_db collector=schema_details datname=azure_maintenance schema=pgavailability table=lsnmover err="pq: permission denied for schema pgavailability"
```
After examining the connection permissions for the user used to scrape, we decided to start a simple exclusion list similar to what already exists for the MySQL collector.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/postgres/collector/constants.go
```diff
@@ -0,0 +1,3 @@
+package collector
```
matthewnolf: Why don't we put this where it will be used (schema details collector)

fridgepoet: This copies the same pattern as in the MySQL collector. 
Happy to do that, though, want me to change it?

matthewnolf: In MySQL I think it's consumed in multiple collectors. If we need that in the future I think we can extract it but for now I'd keep it close to the single consumer

matthewnolf (approved):

fridgepoet:
Thanks, 
As discussed with Matt, this could become a configurable exclusion list in the future.

PR #4821, opened 2025-11-12
chore: tail fixes

kalleep:
#### PR Description
Some more simplification / cleanup of tail:
1. Remove unused Err property on Line
2. Set location and reader in TailFile, with this we don't have to call "reopen" at the start of `tailFileSync`. This was done before because some settings would allow to start a tail without an existing file, but this is no longer true.
3. We now select on sending a line to Lines channel and Dying instead to just cheking Dying at the end of the loop. This allows us to call TailFile and close it without having another gorouting reading at the same time. This will help with other changes I plan to make for `loki.source.file`

### DIFF
--- a/internal/component/loki/source/file/internal/tail/tail.go
+++ b/internal/component/loki/source/file/internal/tail/tail.go
@@ -94,9 +94,17 @@ func TailFile(filename string, config Config) (*Tail, error) {
 		return nil, err
 	}
 
+	// Seek to requested location.
+	if t.Location != nil {
+		_, err := t.file.Seek(t.Location.Offset, t.Location.Whence)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	t.watcher.SetFile(t.file)
 
-	t.openReader()
+	t.reader = t.getReader()
 
 	go t.tailFileSync()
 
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/file/internal/tail/tail.go
```diff
@@ -95,8 +94,18 @@ func TailFile(filename string, config Config) (*Tail, error) {
 		return nil, err
 	}
 
+	// Seek to requested location.
+	if t.Location != nil {
+		_, err := t.file.Seek(t.Location.Offset, t.Location.Whence)
+		if err != nil {
+			return nil, err
+		}
+	}
```
ptodev, line 103: Ideally, creation of Alloy components shouldn't have side effects such as starting goroutines or file and network IO. Those side effects are something that should only happen when Run is called. In this case I'm ok with this change because this function is already trying to do stuff like open a file and even throw an error if it's not there... but this is an antipattern.

internal/component/loki/source/file/tailer.go
```diff
@@ -323,12 +323,6 @@ func (t *tailer) readLines(handler loki.EntryHandler, done chan struct{}) {
 			return
 		}
 
-		// Note currently the tail implementation hardcodes Err to nil, this should never hit.
-		if line.Err != nil {
-			level.Error(t.logger).Log("msg", "tail routine: error reading line", "path", t.key.Path, "error", line.Err)
-			continue
-		}
-
```
ptodev, line 331: Should we keep this just in case the implementation changes to make Err more useful?

kalleep: I don't see why. If we need it later for some reason we can add it back

internal/component/loki/source/file/internal/tail/tail.go
kalleep: I agree to some extent. This one already had side effects like you said and also starting a goroutine. 

This is not a component and is only used within one. This is not called until a tail is scheduled and that can only happen after the component is running.

But I am trying to simplify this package a bit and plan to do even more simplifications. Not even sure we need to have a channel here and a background job, maybe we can just call a function like `Line` that would either give us the next line or block until we have more changes (using the watcher). But I want to do that in another pr

ptodev (approved):

PR #4834, opened 2025-11-13
fix: no longer drop request if stream is dropped in loki.source.api

kalleep:
#### PR Description
This pr solves several issues in `loki.source.api`.

* No longer cancel request if stream is dropped by relabel rules, this is clearly a return instead of continue bug.

* Add function `ForceShutdown` that will cancel all request before shutting down server.
  * When component is shutting down we are not doing any draining like in [loki.source.file](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/file.go#L209-L219). And IMO we should _not_ do it here. It's better to just cancel all in-flight requests so that caller can retry.
* Send batches instead of individual entries from request handler to component. This makes it a bit more transaction safe where we don't partially ingest entries if request is either canceled or a shutdown happens.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.api.md
+++ b/docs/sources/reference/components/loki/loki.source.api.md
@@ -105,7 +105,6 @@ The metrics include labels such as `status_code` where relevant, which can be us
 * `loki_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
 * `loki_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
 * `loki_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
-* `loki_source_api_entries_processed` (counter): Total number of log entries processed.
 * `loki_source_api_entries_written` (counter): Total number of log entries forwarded.
 
 ## Example
--- a/internal/component/loki/source/api/internal/lokipush/push_api_server.go
+++ b/internal/component/loki/source/api/internal/lokipush/push_api_server.go
@@ -215,8 +215,6 @@ func (s *PushAPIServer) handleLoki(w http.ResponseWriter, r *http.Request) {
 		lastErr error
 	)
 	for _, stream := range req.Streams {
-		s.metrics.entriesProcessed.Add(float64(len(stream.Entries)))
-
 		ls, err := promql_parser.ParseMetric(stream.Labels)
 		if err != nil {
 			lastErr = err
@@ -317,9 +315,7 @@ func (s *PushAPIServer) handlePlaintext(w http.ResponseWriter, r *http.Request) 
 			continue
 		}
 
-		s.metrics.entriesProcessed.Inc()
 		entries = append(entries, loki.Entry{Labels: addLabels, Entry: lokipush.Entry{Timestamp: time.Now(), Line: line}})
-
 		if err == io.EOF {
 			break
 		}
--- a/internal/component/loki/source/api/internal/lokipush/metrics.go
+++ b/internal/component/loki/source/api/internal/lokipush/metrics.go
@@ -11,17 +11,11 @@ func newMetircs(reg prometheus.Registerer) *metrics {
 			Name: "loki_source_api_entries_written",
 			Help: "Total number of entries written.",
 		}),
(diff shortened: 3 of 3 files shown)


internal/component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -250,22 +262,26 @@ func (s *PushAPIServer) handleLoki(w http.ResponseWriter, r *http.Request) {
...
+		case <-r.Context().Done():
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		case <-s.forceShutdown:
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
+
+		if lastErr != nil {
+			level.Warn(s.logger).Log("msg", "at least one entry in the push request failed to process", "err", lastErr.Error())
```
thampiotr, line 285: I think it would be cheap and easy to add a counter of how many entries failed out of how many in the batch. And that would be useful.

internal/component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -250,22 +262,26 @@ func (s *PushAPIServer) handleLoki(w http.ResponseWriter, r *http.Request) {
...
-			select {
-			case s.handler.Chan() <- e:
-			case <-r.Context().Done():
-				w.WriteHeader(http.StatusServiceUnavailable)
-				return
-			case <-s.forceShutdown:
-				w.WriteHeader(http.StatusServiceUnavailable)
-				return
-			}
+			entries = append(entries, e)
```
thampiotr, line 266: Might want to reuse the slices in a pool if this comes up in allocations.

kalleep: Agree, do you think we should do that now or wait and see if it shows up?

internal/component/loki/source/api/internal/lokipush/push_api_server.go
kalleep: Hm after checking what kind of metrics other exposes we could add two metrics:
* loki_source_api_entries_processed - this would be the total number of entries in a request
* loki_source_api_entries_written - this would be the total number of entries that we have forwarded

With that you could derive how many are dropped either by relabeling or by invalid labels, WDYT?

kalleep: I went with my suggestion. Let me know what you think :)

internal/component/loki/source/api/internal/lokipush/push_api_server.go
kgeckhart: If we go down the pooling path could we try to avoid sending it through a channel? It's fine today because we know the implementation is an unbuffered channel but it feels risky to depend upon that given the server can't enforce it.

thampiotr: Let's wait to see if this comes up in profiles.

docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -105,6 +105,8 @@ The metrics include labels such as `status_code` where relevant, which can be us
 * `loki_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
 * `loki_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
 * `loki_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
+* `loki_source_api_entries_processed` (counter): Total number of log entries processed.
+* `loki_source_api_entries_written` (counter): Total number of log entries forwarded.
```
thampiotr: I think we do want some metrics, but we also don't want to have too many metrics. Should we settle on only one off them? I think `loki_source_api_entries_written` would be better option to keep. The idea is to be able to spot some major issues and then continue debugging using other means like logging, profiles or local repro.

kalleep: Sure, only kept `loki_source_api_entries_written`

PR #4840, opened 2025-11-14
chore: refactor loki client code

kalleep:
#### PR Description
Extracted from https://github.com/grafana/alloy/pull/4788

1. Remove unused function StopNow - only used in tests
2. Remove unused function Name - only used in tests
3. Move metrics to it's own file
4. Move maxStreams to config
5. Move walWriter into manager
6. Update a bunch of sync.WaitGroup to use new api
7. Remove addFromWal in batch and add segmentNum to add. For non wal usage this will always be 0.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/batch.go
+++ b/internal/component/common/loki/client/batch.go
@@ -60,8 +60,8 @@ func newBatch(maxStreams int, entries ...loki.Entry) *batch {
 	return b
 }
 
-// add an entry to the batch, tracking that the data being added
-// comes from segment segmentNum read from the WAL.
+// add an entry to the batch. segmentNum is used to associate batch with a segment from WAL.
+// If entry is added from non backed WAL client it can be anything and is unused.
 func (b *batch) add(entry loki.Entry, segmentNum int) error {
 	b.totalBytes += entrySize(entry.Entry)
 


internal/component/common/loki/client/batch.go
```diff
@@ -54,44 +54,21 @@ func newBatch(maxStreams int, entries ...loki.Entry) *batch {
...
+		_ = b.add(entry, 0)
 	}
 
 	return b
 }
 
-// add an entry to the batch
-func (b *batch) add(entry loki.Entry) error {
+// add an entry to the batch, tracking that the data being added
+// comes from segment segmentNum read from the WAL.
```
thampiotr: It won't always be wal now, right?
can you comment what segmentNum should be if there is no wal? 

are we sure we need it? it seems confusing when wal is disabled

internal/component/common/loki/client/client_test.go
```diff
@@ -464,160 +463,11 @@ func TestClient_Handle(t *testing.T) {
...
+			assert.ElementsMatch(t, testData.expectedReqs, receivedReqs)
 
 			expectedMetrics := strings.ReplaceAll(testData.expectedMetrics, "__HOST__", serverURL.Host)
 			err = testutil.GatherAndCompare(reg, strings.NewReader(expectedMetrics), "loki_write_sent_entries_total", "loki_write_dropped_entries_total", "loki_write_mutated_entries_total", "loki_write_mutated_bytes_total")
 			assert.NoError(t, err)
 		})
 	}
 }
-
-func TestClient_StopNow(t *testing.T) {
```
thampiotr, line 481: Are we not reducing coverage by deleting these tests? can't we keep them?

kalleep: They only tested `StopNow`. This function is never called in code so I don't see any value in keeping it around.

internal/component/common/loki/client/batch.go
kalleep: Yeah I know. In non wal implementation we always call with zero but the number won't matter.
This only matters when you call [reportDataAsSent](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/client/batch.go#L201C17-L201C33) with a MarkerHandler and we never do in non wal scenarios. 

I agree it is a bit confusing but it's needed for https://github.com/grafana/alloy/pull/4788.

But I can revert the change here if you prefer we make that change in that pr instead?

kalleep: Updated the comment in [1a250b4](https://github.com/grafana/alloy/pull/4840/commits/1a250b4f362f7aea6172de431b6e8a9f6f41082e)

internal/component/common/loki/client/manager_test.go
```diff
@@ -36,9 +35,9 @@ func TestManager_NoDuplicateMetricsPanic(t *testing.T) {
 
 	require.NotPanics(t, func() {
 		for i := 0; i < 2; i++ {
-			_, err := NewManager(metrics, log.NewLogfmtLogger(os.Stdout), 0, reg, wal.Config{
+			_, err := NewManager(metrics, log.NewNopLogger(), reg, wal.Config{
```
thampiotr, line 38: Nit: keeping a logger that prints to output can be useful and doesn't have much cost. I think it makes sense to keep it debug-ready so in CI you can see the logs if something fails.

thampiotr (approved):

internal/component/common/loki/client/manager_test.go
kalleep: While I agree that it could be useful, most of the time it's not and just adds a lot of noise. I think I suggested this somewhere else it would be nice if we could control what logger was used for _all_ test with and env and default should be off.

PR #4861, opened 2025-11-16
chore: Add support for mounting Alloy config from Kubernetes Secrets

### LINKED ISSUE #443: Allow mounting the config via a secret instead of a configmap or provide an easier way to override default
### Request

I would like an easier way to manage the config and secrets containing the authentication. Currently, we are forced into (the only sensible pattern) of setting env vars from secrets with `agent.extraEnv` and hope that the keys defined in the secret matches up to the substitution variable defined in the config.  

Ideally we would use remote config but the functionality is still a little bit limited at the time of writing.  So we need something that is similar enough to a remote config but safer.  Hence, I would prefer to have the option of mounting a secret (we can RBAC safely) which contains the config with authentication details.

### Use case

Considering that the use of authentication for remote write is common, it is not out of the ordinary that we may want to store the config as a secret with the authentication baked in.  I am well aware of the usage of env vars which can be substituted in easily (via say `${SOME_PASSWORD}`), but the reality is that if we are pulling in secrets with custom mounting + extra env setting then it is actually simpler to combine the config and the secret, mount once and skip the extra env stuff!

Obviously we can indeed just mount an arbitrary k8s secrets, but the way the helm chart is structured means that we cannot easily point the binary to run against the mounted volume. We can "cheat" by setting `.Values.agent.mode` as something that is not "flow" or "static" which will free us from the mandated setting of `{{ include "grafana-agent.config-map.key" . }}`.  Then all we need is to pass in `.Values.agent.extraArgs` pointing to the custom mounted volume.  Evidently this is not ideal as we taking multiple steps to bypass the deliberate construction of the helm chart. A more user friendly approach should let us
1. Mount the config as secret directly
2. Allow an alternative location for the config file in `static` mode, i.e. the full file path; the flow mode has the ability to declare the location of the credentials so not required.

naptalie:
## Summary

This enhancement addresses issue #443 by allowing users to mount Grafana Alloy configuration from Kubernetes Secrets instead of only from ConfigMaps. This provides better security for configurations containing sensitive credentials.

### New Features
- Added `alloy.secret` configuration option in [values.yaml](operations/helm/charts/alloy/values.yaml:43-52) with the following fields:
  - `alloy.secret.create`: Create a new Secret for the config file
  - `alloy.secret.content`: Content to assign to the Secret (supports templating via `tpl`)
  - `alloy.secret.name`: Name of existing Secret to use
  - `alloy.secret.key`: Key in Secret to get config from

### Implementation Details
- Secret takes precedence over ConfigMap when both are configured
- ConfigMap is not created when using Secret (prevents resource duplication)
- Support for both creating new Secrets and referencing existing ones
- Added helper templates in [_config.tpl](operations/helm/charts/alloy/templates/_config.tpl) for Secret name and key resolution
- Updated [pod template](operations/helm/charts/alloy/templates/controllers/_pod.yaml:72-80) to conditionally mount Secret or ConfigMap
- Added checksum annotation support for Secret content to trigger pod restarts on config changes

### Testing
- Added CI test configuration for creating new Secret: [custom-secret-config-values.yaml](operations/helm/charts/alloy/ci/custom-secret-config-values.yaml)
- Added CI test configuration for using existing Secret: [existing-secret-values.yaml](operations/helm/charts/alloy/ci/existing-secret-values.yaml)
- All existing tests continue to pass with ConfigMap (backward compatible)

### Documentation
- Updated README.md with new `alloy.secret` section
- Updated README.md.gotmpl template
- Added values table entries for all new secret configuration options
- Updated CHANGELOG.md

## Backward Compatibility

The implementation maintains full backward compatibility with existing ConfigMap-based configurations. Users can continue using ConfigMaps without any changes.

## Test Plan

1. Default behavior (ConfigMap) continues to work:
   ```bash
   helm template test . --values ci/default-values-values.yaml
   ```

2. New Secret creation works:
   ```bash
   helm template test . --values ci/custom-secret-config-values.yaml
   ```

3. Referencing existing Secret works:
   ```bash
   helm template test . --values ci/existing-secret-values.yaml
   ```

## Usage Example

To use a Secret for your Alloy configuration:

```yaml
alloy:
  secret:
    create: true
    content: |
      prometheus.remote_write "default" {
        endpoint {
          url = "https://prometheus.example.com/api/v1/write"
          basic_auth {
            username = "my-username"
            password = "my-secret-password"
          }
        }
      }
```

Or to reference an existing Secret:

```yaml
alloy:
  secret:
    create: false
    name: my-existing-secret
    key: config.alloy
```

Fixes #443

Co-Authored-By: Claude <noreply@anthropic.com>


operations/helm/charts/alloy/templates/rbac.yaml
```diff
@@ -11,7 +11,11 @@ metadata:
     {{- include "alloy.labels" $ | nindent 4 }}
     app.kubernetes.io/component: rbac
 rules:
+  {{- if $.Values.rbac.rules }}
```
dehaansa: I think this was leaked over from your other PR.

dehaansa (commented):
Broad strokes of the implementation seems fine, couple issues and CI.

operations/helm/charts/alloy/templates/configmap.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $values := (mustMergeOverwrite .Values.alloy (or .Values.agent dict)) -}}
-{{- if $values.configMap.create }}
+{{- if and $values.configMap.create (not (or $values.secret.create $values.secret.name)) }}
```
dehaansa: Can this leverage alloy.config-source.use-secret?

PR #4863, opened 2025-11-17
chore: refactor loki.write manager

kalleep:
#### PR Description
Followup to https://github.com/grafana/alloy/pull/4840

Manager was really messy and depending if you ran with WAL or not certain fields is used.

So instead the manager was split up in two different implementations, one for WAL and another one when wal is disabled.

Also renamed the interface / structure from manager to consumer.

1. FanoutConsumer
  * This is the one that is used when WAL is disabled. 
2. WALConsumer
  * This is used when WAL is enabled and all entries are written into the wal before client's consumes them.

These two uses different clients so I moved their respective client code into the manager file. I also renamed QueueClient to WALClient. This name fits better because it is a client that consumes from the WAL and uses a queue for writing entries.

Later when we add https://github.com/grafana/alloy/pull/4788 both client will use queues and shards. But how they get their data is the thing that is different between them. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/write/write.go
+++ b/internal/component/loki/write/write.go
@@ -85,11 +85,11 @@ type Component struct {
 	args     Arguments
 	receiver loki.LogsReceiver
 
-	// remote write components
-	clientManger client.Consumer
+	// remote write consumer
+	consumer client.Consumer
 
-	// sink is the place where log entries received by this component should be written to. If WAL
-	// is enabled, this will be the WAL Writer, otherwise, the client manager
+	// sink is the place where log entries received by this component should be written to.
+	// It will in turn write to client.Consumer.
 	sink loki.EntryHandler
 }
 
@@ -120,12 +120,12 @@ func (c *Component) Run(ctx context.Context) error {
 			c.sink.Stop()
 		}
 
-		if c.clientManger != nil {
-			if d, ok := c.clientManger.(client.DrainableConsumer); ok {
+		if c.consumer != nil {
+			if d, ok := c.consumer.(client.DrainableConsumer); ok {
 				// stop and drain since component is shutting down.
 				d.StopAndDrain()
 			} else {
-				c.clientManger.Stop()
+				c.consumer.Stop()
 			}
 		}
 	}()
@@ -159,9 +159,9 @@ func (c *Component) Update(args component.Arguments) error {
 		c.sink.Stop()
 	}
 
-	if c.clientManger != nil {
+	if c.consumer != nil {
 		// only drain on component shutdown
-		c.clientManger.Stop()
+		c.consumer.Stop()
 	}
 
 	cfgs := newArgs.convertClientConfigs()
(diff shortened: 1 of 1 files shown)


internal/component/common/loki/client/internal/marker_handler.go
```diff
@@ -212,3 +210,19 @@ func FindMarkableSegment(segmentDataCount map[int]*countDataItem, tooOldThreshol
...
+
+type NopMarkerHandler struct{}
+
+func (n *NopMarkerHandler) LastMarkedSegment() int { return 0 }
+
+func (n *NopMarkerHandler) UpdateReceivedData(segmentId int, dataCount int) {}
+
+func (n *NopMarkerHandler) UpdateSentData(segmentId int, dataCount int) {}
+
+func (n *NopMarkerHandler) Stop() {}
```
thampiotr, line 234: This could go into a test file, no?

internal/component/common/loki/client/manager.go
```diff
@@ -1,246 +1,9 @@
...
-	"github.com/prometheus/client_golang/prometheus"
-
-	"github.com/grafana/alloy/internal/component/common/loki"
-	"github.com/grafana/alloy/internal/component/common/loki/client/internal"
-	"github.com/grafana/alloy/internal/component/common/loki/wal"
-	"github.com/grafana/alloy/internal/runtime/logging/level"
-)
-
-type StoppableWatcher interface {
+type Manager interface {
```
thampiotr: Is manager a good name? Isn't `EntryConsumer` more reflective of the interface? 
We could also make this implement a new `Stoppable` and `Drainable` and compose these interfaces, more idiomatic way in Go.

internal/component/common/loki/client/manager_memory.go
```diff
@@ -13,17 +14,111 @@ import (
...
+		m.clients = append(m.clients, client)
+	}
+
+	m.wg.Go(m.run)
+	return m, nil
+}
+
+var _ Manager = (*MemoryManager)(nil)
+
+type MemoryManager struct {
```
thampiotr: It doesn't manage any memory, so this is not the best name IMO. This seems to fan out to multiple clients, right?

internal/component/common/loki/client/manager_memory.go
```diff
@@ -13,17 +14,111 @@ import (
...
+	clients []Client
```
thampiotr: The `Client` interface is very similar to `Manager` and probably we could use one for both for some kind of general consumer that has a Stop function.

internal/component/common/loki/client/consumer_fanout.go
thampiotr, line 1: There's a function called `New` here, which should have a more specific name of what it creates.

internal/component/common/loki/client/manager_wal.go
```diff
@@ -16,132 +16,148 @@ import (
...
 }
 
-// queue wraps a buffered channel and a routine that reads from it, sending batches of entries.
-type queue struct {
-	client *queueClient
-	q      chan queuedBatch
-	quit   chan struct{}
-	wg     sync.WaitGroup
-	logger log.Logger
+type pair struct {
```
thampiotr: We need a better name for this. This looks more like a placeholder name, it's not helping with readability, but I think it could

internal/component/common/loki/client/internal/marker_handler.go
kalleep: Currently it's only used in tests yes. But with https://github.com/grafana/alloy/pull/4788 I need it in non test code

internal/component/common/loki/client/manager_wal.go
kalleep: Ah yes this was a placeholder name

kalleep:
@thampiotr thanks for the suggestions. I agree that a lot of these names are not great.

I went with the interfaces you suggested and only changed names for the managers.

So now we have 
1. WALConsumer implementation
2. FanoutConsumer implementation

Everything is internal to these components and I got rid of `Client` interface, it was not used. I think with this and implementing the sharding I am not sure how much more we need to do. I need to think about that a bit :)

internal/component/common/loki/client/manager.go
kalleep: I just went with consumer. We could also reuse EntryHandler if we want and have DrainableEntryHandler just for WAL implementation. WDYT?

kalleep:
Removed Stoppable and Drainable. A client.Consumer will always need to implement Stop and Chan and only WAL implementation needs to have StopAndDrain so it simplifies it a lot.

PR #4865, opened 2025-11-17
ci: fix publish-alloy-release workflow

jharvey10:
gh command is was not found. Replace it with the way we do this elsewhere in the repo

### DIFF
--- a/.github/workflows/publish-alloy-release.yml
+++ b/.github/workflows/publish-alloy-release.yml
@@ -253,34 +253,29 @@ jobs:
           ALLOYBOT_APP_ID=alloybot:app_id
           ALLOYBOT_PRIVATE_KEY=alloybot:private_key
 
-    - name: Get GitHub app token
-      uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
-      id: get-token
+    - name: Create GitHub app token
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-alloy-release.yml
```diff
@@ -253,34 +253,29 @@ jobs:
...
+        persist-credentials: false
 
     - name: Configure Git
-      env:
-        APP_ID: ${{ steps.get-app.outputs.app-id }}
-        APP_LOGIN: ${{ steps.get-app.outputs.app-login }}
       run: |
-        git config --global user.name "${APP_LOGIN}"
-        git config --global user.email "${APP_ID}+${APP_LOGIN}@users.noreply.github.com"
+        git config user.name "$GITHUB_ACTOR"
```
blewis12, line 274: Would you mind explaining the change with the environment variables? Just want to make sure i'm understanding any differences (if there are any) between what we're setting before with `APP_ID`/`APP_LOGIN` vs now with `GITHUB_ACTOR` - I can see that we do this in the helm-release chart but it seems like most other places we still use this `APP_ID`/`APP_LOGIN` combo 🤔

blewis12: aware this is merged hehe, just out of curiosity

PR #4872, opened 2025-11-18
database_observability: fix schema_details collector to fetch column definitions with case sensitive table names

fridgepoet:
#### PR Description

This PR modifies the select column query to look up table name from pg_class and schema from pg_namespace, preserving any case sensitivity of the table name.

Previously, the lookup was done with a cast `::regclass`. `::regclass` lowercases unquoted identifiers, so if the table was created with mixed case, the lookup failed.

For example:
If a table like `public.SomeMixedCaseName` exists, then the previous code would fail with an error message like:
`level=error msg="failed to get table definitions" component_path=/ component_id=database_observability.postgres.postgres_postgres_db collector=schema_details database=some_db schema=public err="pq: relation \"public.somemixedcasename\" does not exist"`

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ Main (unreleased)
 
 - `loki.source.api` no longer drops request when relabel rules drops a specific stream. (@kalleep)
 
-- (_Experimental_) Additions to experimental `database_observability.postgres` component:
+- (_Public Preview_) Additions to `database_observability.postgres` component:
     - fixes collection of Postgres schema details for mixed case table names (@fridgepoet)
 
 v1.12.0-rc.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ Main (unreleased)
 
 - `loki.source.api` no longer drops request when relabel rules drops a specific stream. (@kalleep)
 
+- (_Experimental_) Additions to experimental `database_observability.postgres` component:
+    - fixes collection of Postgres schema details for mixed case table names (@fridgepoet)
+
 v1.12.0-rc.0
 -----------------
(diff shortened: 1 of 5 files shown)


internal/component/database_observability/postgres/collector/connection_info_test.go
```diff
@@ -15,7 +15,9 @@ import (
 )
 
 func TestConnectionInfo(t *testing.T) {
-	defer goleak.VerifyNone(t)
```
fridgepoet, line 18: Three tests were failing locally if I ran them with `-count=2` because of this. Just a drive-by improvement but can remove this from the PR if we prefer.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Main (unreleased)
 
 - `loki.source.api` no longer drops request when relabel rules drops a specific stream. (@kalleep)
 
+- (_Experimental_) Additions to experimental `database_observability.postgres` component:
```
fridgepoet: Put this under Next >  Bugfixes and with an Experimental tag. Is that right?

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Main (unreleased)
 
 - `loki.source.api` no longer drops request when relabel rules drops a specific stream. (@kalleep)
 
+- (_Public Preview_) Additions to `database_observability.postgres` component:
```
fridgepoet, line 21: Put this under Next > Bugfixes and with an Public Preview tag. Is that right?

matthewnolf (approved):

PR #4878, opened 2025-11-18
chore: fix tests that fail on windows

dehaansa:


dehaansa:
The mimir client test is still expected to fail, Paulin's working on that separately.

PR #4879, opened 2025-11-18
Remove experimental flag from stage.windowsevent

kgeckhart:
#### PR Description

Removes the experimental requirement from stage.windowsevent which hasn't changed since it was originally introduced in Feb. No doc changes required because the docs didn't mention it was experimental.

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Main (unreleased)
 ### Features
 
 - A new `mimir.alerts.kubernetes` component which discovers `AlertmanagerConfig` Kubernetes resources and loads them into a Mimir instance. (@ptodev)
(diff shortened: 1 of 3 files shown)


ptodev:
I had a quick look at the [docs](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagewindowsevent) to see if there's anything weird about this feature that we might need to change. The only thing that stood out to me was:

> When overwrite_existing is set to true, the stage overwrites existing extracted data fields with the same name. If set to false, the _extracted suffix is appended to an existing field name.

I wonder what happens if the field in the event log is called "testField", but there is already both a "testField" and a "testField_extracted" in the extracted map. This is the only thing I could think of that we might need to change in the future... but I don't mind if we mark this stable and look into it more later on.

PR #4881, opened 2025-11-19
database_observability: remove README

fridgepoet:
#### PR Description

As we've moved into Public Preview, we ask that users refer to documentation at https://grafana.com/docs/grafana-cloud/monitor-applications/database-observability/get-started/ 
This PR removes duplicate instructions.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


matthewnolf:
I don't think we should remove the README. A lot of the content is duplicated across docs, so it makes sense to instead link there - but that itself is a useful redirect. Additionally, I'd imagine we may want to document the technical details of how the component works here in the near future.

cristiangreco:
> I don't think we should remove the README. A lot of the content is duplicated across docs, so it makes sense to instead link there - but that itself is a useful redirect. Additionally, I'd imagine we may want to document the technical details of how the component works here in the near future.

Why is the PR closed? I think we can keep the file but just change its content to point to https://grafana.com/docs/grafana-cloud/monitor-applications/database-observability/get-started/

fridgepoet:
Oh sorry, Cristian! Missed seeing this comment, but thanks for bringing it to my attention. Made a proposition here: https://github.com/grafana/alloy/pull/4918

Issue #4838, opened 2025-11-14
loki.write: queue_config metrics


kalleep:
### Component(s)

loki.write

### Add a description

We should add metrics for shards added in https://github.com/grafana/alloy/pull/4788.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #4882, opened 2025-11-19
feat: Add sharding for loki.write

kalleep:
#### PR Description
This PR implements `queue_config` for the `loki.write` component, enabling users to configure queue-based batching and parallel processing. The implementation introduces a new sharding architecture that distributes log entries across multiple parallel queues based on label fingerprints. This implementation is based on [Prometheus rw sharding](https://github.com/prometheus/prometheus/blob/main/storage/remote/queue_manager.go#L1235).

The shards implementation is used with both "normal" clients and "WAL" clients. So we get rid of a lot of duplicated logic.

Before this pr we had a `queue_config` block that was only used when `WAL` was enabled. It is now always used and will affect clients regardless. 

Currently no automatic "resharding" is implemented. Implementing this without the WAL will most likely be pretty primitive. So for now `min_shards` is the only configurable value until we address this.

Ideally we would move a couple of attributes from `endpoint` block to `queue_config` block to closer match `prometheus.remote_write`. But we can't do that without doing a breaking change. These attributes are:

* `retry_on_http_429`
* `max_backoff_period`
* `min_backoff_period`
* `batch_size`
* `batch_wait`

Part of: https://github.com/grafana/alloy/issues/4728

#### Notes to the Reviewer

I moved wal writer ownership into client.Manager. No need to expose it to `loki.write` component.
I plan to work on resharding in followup pr

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.write.md
+++ b/docs/sources/reference/components/loki/loki.write.md
@@ -143,6 +143,9 @@ The following arguments are supported:
 Each endpoint is divided into a number of concurrent _shards_ which are responsible for sending a fraction of batches. The number of shards is controlled with `min_shards` argument.
 Each shard has a queue of batches it keeps in memory, controlled with the `capacity` argument.
 
+Queue size is calculated using `batch_size` and `capacity` for each shard. So if `batch_size` is 1MiB and `capacity` is 10MiB each shard would be able to queue up 10 batches.
+The maximum amount of memory required for all configured shards can be calculated using `capacity` * `min_shards`. 
+
 ### `tls_config`
 
 {{< docs/shared lookup="reference/components/tls-config-block.md" source="alloy" version="<ALLOY_VERSION>" >}}


docs/sources/reference/components/loki/loki.write.md
```diff
@@ -46,7 +46,7 @@ You can use the following blocks with `loki.write`:
 | `endpoint` > [`basic_auth`][basic_auth]            | Configure `basic_auth` for authenticating to the endpoint. | no       |
 | `endpoint` > [`oauth2`][oauth2]                    | Configure OAuth 2.0 for authenticating to the endpoint.    | no       |
 | `endpoint` > `oauth2` > [`tls_config`][tls_config] | Configure TLS settings for connecting to the endpoint.     | no       |
-| `endpoint` > [`queue_config`][queue_config]        | When WAL is enabled, configures the queue client.          | no       |
+| `endpoint` > [`queue_config`][queue_config]        | Configure the queue used for endpoint.                     | no       |
```
```suggestion
| `endpoint` > [`queue_config`][queue_config]        | Configure the queue used for the endpoint.                     | no       |
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
 
 ### `queue_config`
 
-{{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
thampiotr, line 130: Shouldn't this continue to be experimental?

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
 
 The following arguments are supported:
 
 | Name            | Type       | Description                                                                                                                                                                   | Default | Required |
 | --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
 | `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered a worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB` | no       |
 | `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it enqueues pending batches and drains the send queue sending each.  | `"1m"`  | no       |
+| `min_shards`    | `number`   | Minimum amount of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
```
```suggestion
| `min_shards`    | `number`   | Minimum number of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
-
-The optional `queue_config` block configures, when WAL is enabled, how the underlying client queues batches of logs sent to Loki.
-Refer to [Write-Ahead block](#wal) for more information.
+The optional `queue_config` block configures how the endpoint queues batches of logs sent to Loki.
 
 The following arguments are supported:
 
 | Name            | Type       | Description                                                                                                                                                                   | Default | Required |
 | --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
 | `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered a worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB` | no       |
```
thampiotr, line 139:
```suggestion
| `capacity`      | `string`   | Controls the size of the underlying send queue buffer of each shard. Consider this setting as the worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB` | no       |
```

What does it even mean 'all enqueued batches are full'? Shouldn't it say that it's the total size of all the enqueued batches instead?

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
 | `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it enqueues pending batches and drains the send queue sending each.  | `"1m"`  | no       |
+| `min_shards`    | `number`   | Minimum amount of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
+
+Each endpoint manages a number of concurrent _shards_ which is responsible for sending a fraction of batches, number of shards are controlled with `min_shards` argument.
```
```suggestion
Each endpoint is divided into a number of concurrent _shards_ which are responsible for sending a fraction of batches. The number of shards is controlled with `min_shards` argument.
```

internal/component/common/loki/client/config.go
```diff
@@ -47,6 +47,9 @@ type QueueConfig struct {
 	// is the 1 MiB default, and a capacity of 100 MiB, the underlying buffered channel would buffer up to 100 batches.
 	Capacity int
 
+	// MinShards is the minimum amount of concurrent shards sending batches to the endpoint.
```
thampiotr:
```suggestion
	// MinShards is the minimum number of concurrent shards sending batches to the endpoint.
```

docs/sources/reference/components/loki/loki.write.md
thampiotr: Should we consider calling it something more intuitive than shards? I feel this term is just inherited from Prometheus 🤔

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -127,17 +128,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
 
 ### `queue_config`
 
-{{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
-
-The optional `queue_config` block configures, when WAL is enabled, how the underlying client queues batches of logs sent to Loki.
-Refer to [Write-Ahead block](#wal) for more information.
+The optional `queue_config` block configures how the endpoint queues batches of logs sent to Loki.
```
thampiotr, line 133: We call it queue_config, but we are actually configuring both sharding and queuing in one block.

Maybe this should be called something like [parallelism](https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.write.queue/#parallelism) in write.queue?

docs/sources/reference/components/loki/loki.write.md
kalleep: Yeah we could keep this as experimental. But we would always use this config after this pr.

What would be considered experimental with it would be naming and changing defaults I guess.

docs/sources/reference/components/loki/loki.write.md
kalleep: Yeah this was there before and I did not check / alter it.

But essentially whenever the capacity is full that means that the queue of batches is full and we cannot enqueue another one so we would block here until we get more capacity

docs/sources/reference/components/loki/loki.write.md
kalleep: Discussed this in slack, naming is fine because it would match what we use in `prometheus.remote_write`

docs/sources/reference/components/loki/loki.write.md
kalleep: Discussed this in slack, naming is fine because it would match what we use in `prometheus.remote_write`

kalleep:
In [ffb2bec](https://github.com/grafana/alloy/pull/4882/commits/ffb2bec2270a34a3cfa0575f15ccddd00d33f457) I renamed the two clients implementation we have to endpoint and walEndpoint. 

I will look into the other ways we discussed structuring this but will have this as a fallback if that don't work out

kalleep:
@thampiotr I update the pr with an attempt to share stuff between non WAL and WAL implementation.

So we now have one endpoint struct that will handle shards. This implemntation has one method `enqueue` that will try to forward entry to shards for queuing with retries. If endpoint is stopped any attempts will fail end return false.

We can use this directly in Fanout and we no longer need channels between fanout and endpoint.

For WAL implementation I renamed it walEndpointAdapter. This implements the necessary interface that watcher expected and will just call enqueue on endpoint, that internally handles retries etc.

Naming could be a bit off but this would be an option we can go with. Let me know what you think :)

clayton-cornell:
Nothing really stands out for the small doc changes. It's good as-is.

docs/sources/reference/components/loki/loki.write.md
thampiotr: This setting is per-shard right? we should clarify this here.

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -129,15 +130,18 @@ When `retry_on_http_429` is enabled, the retry mechanism is governed by the back
...
 | `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered a worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB` | no       |
 | `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it enqueues pending batches and drains the send queue sending each.  | `"1m"`  | no       |
+| `min_shards`    | `number`   | Minimum number of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
+
+Each endpoint is divided into a number of concurrent _shards_ which are responsible for sending a fraction of batches. The number of shards is controlled with `min_shards` argument.
+Each shard has a queue of batches it keeps in memory, controlled with the `capacity` argument.
```
thampiotr, line 144: I wonder if we shold provide some example about calculating memory requirement. e.g. 10 shards x 10MiB = 100 MiB of headroom required by Alloy process.

clayton-cornell: Is that something we'd place here in this topic or in https://grafana.com/docs/alloy/latest/introduction/estimate-resource-usage/ with a link from `loki.write` to the resource usage topic?

internal/component/loki/write/types.go
```diff
@@ -70,22 +70,25 @@ func (r *EndpointOptions) Validate() error {
...
 type QueueConfig struct {
 	Capacity     units.Base2Bytes `alloy:"capacity,attr,optional"`
+	MinShards    int              `alloy:"min_shards,attr,optional"`
 	DrainTimeout time.Duration    `alloy:"drain_timeout,attr,optional"`
 }
 
+var defaultQueueConfig = QueueConfig{
+	Capacity:     10 * units.MiB, // considering the default BatchSize of 1MiB, this gives us a default buffered channel of size 10
+	MinShards:    1,
+	DrainTimeout: 15 * time.Second,
```
thampiotr, line 83: In docs this is 1m

internal/component/loki/write/write.go
```diff
@@ -99,6 +100,10 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		opts: o,
 	}
 
+	if err := validateConfigStabilityLevel(o, args); err != nil {
+		return nil, err
+	}
```
thampiotr: Don't we call Update anyway where this check will happen?

docs/sources/reference/components/loki/loki.write.md
thampiotr: This one would be a bit too low-level. Also it's still experimental feature. I would keep it here only.

internal/component/common/loki/client/endpoint.go
```diff
@@ -0,0 +1,80 @@
...
+		MinBackoff: 5 * time.Millisecond,
+		MaxBackoff: 50 * time.Millisecond,
+	})
+
+	for !c.shards.enqueue(entry, segmentNum) {
+		backoff.Wait()
+		if !backoff.Ongoing() {
+			return false
+		}
+	}
```
thampiotr, line 61: If this blocks forever due to some issues in enqueue or other functions downstream from here, what would be the symptoms? Would we have any logs or metrics or some other way of debugging this?

internal/component/common/loki/client/shards.go
```diff
@@ -0,0 +1,526 @@
...
+	close(s.softShutdown)
+
+	for _, q := range s.queues {
+		go q.flushAndShutdown(s.done)
+	}
+
+	select {
+	case <-s.done:
+		return
+	case <-time.After(s.cfg.QueueConfig.DrainTimeout):
```
thampiotr, line 306: Should we log something or signal in some way this timeout happened?

internal/component/common/loki/client/shards.go
```diff
@@ -0,0 +1,526 @@
...
+	}
+
+	fingerprint := entry.Labels.FastFingerprint()
+	shard := uint64(fingerprint) % uint64(len(s.queues))
+
+	select {
+	case <-s.softShutdown:
+		return false
+	default:
+		return s.queues[shard].append(tenantID, entry, segmentNum)
```
thampiotr, line 376: Do we still need to hold the lock when calling append here? Or it doesn't really matter?

internal/component/loki/write/write.go
kalleep: That is true :+1:

internal/component/loki/write/types.go
kalleep: No BatchSize is configured to 1MB but capacity is 10MB in docs

internal/component/common/loki/client/shards.go
kalleep: We don't because has it's own locking but do not matter right now because we can only queue one item at a time currently, thanks to the limitations of the pipeline.

I think it's safest to hold the lock for the full duration for now. I need to revisit this if we change the pipeline

internal/component/common/loki/client/endpoint.go
kalleep: Not more than we have today. If we get stuck here for whatever reason `loki_write_sent_bytes_total` would go down to 0 and `loki_write_dropped_bytes_total` be 0.

I have a issue for adding more useful metrics for shardings here https://github.com/grafana/alloy/issues/4838 I could also go through add ad useful logs in the same pr for that one.

docs/sources/reference/components/loki/loki.write.md
kalleep: Added a small section for this. Let me know what you think

PR #4890, opened 2025-11-19
remove labelstore from prometheus interceptor

kgeckhart:
#### PR Description

Removes unnecessary usage of labelstore from the prometheus.interceptor used in prometheus pipelines. 

Beyond simplifying the code there's a performance gain from the reduction in duplicate staleness tracking through the labelstore,

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/component/prometheus
cpu: Apple M3 Pro
                                  │   baseline   │                new                 │
                                  │    sec/op    │   sec/op     vs base               │
Pipelines/default/2-metrics-11      5.638µ ±  7%   5.297µ ± 3%   -6.05% (p=0.009 n=6)
Pipelines/relabel/2-metrics-11      6.267µ ±  1%   5.819µ ± 2%   -7.14% (p=0.002 n=6)
Pipelines/default/10-metrics-11     27.78µ ±  6%   25.79µ ± 1%   -7.17% (p=0.002 n=6)
Pipelines/relabel/10-metrics-11     32.76µ ±  7%   29.71µ ± 2%   -9.29% (p=0.002 n=6)
Pipelines/default/100-metrics-11    309.6µ ± 10%   271.5µ ± 3%  -12.31% (p=0.002 n=6)
Pipelines/relabel/100-metrics-11    350.0µ ±  2%   308.3µ ± 5%  -11.92% (p=0.002 n=6)
Pipelines/default/1000-metrics-11   3.099m ±  8%   2.772m ± 3%  -10.57% (p=0.002 n=6)
Pipelines/relabel/1000-metrics-11   3.644m ±  4%   3.228m ± 3%  -11.44% (p=0.002 n=6)
geomean                             118.7µ         107.4µ        -9.52%

                                  │   baseline   │                 new                 │
                                  │     B/op     │     B/op      vs base               │
Pipelines/default/2-metrics-11      1.926Ki ± 0%   1.644Ki ± 0%  -14.66% (p=0.002 n=6)
Pipelines/relabel/2-metrics-11      2.505Ki ± 0%   2.082Ki ± 0%  -16.90% (p=0.002 n=6)
Pipelines/default/10-metrics-11     9.629Ki ± 0%   8.217Ki ± 0%  -14.66% (p=0.002 n=6)
Pipelines/relabel/10-metrics-11     12.52Ki ± 0%   10.41Ki ± 0%  -16.88% (p=0.002 n=6)
Pipelines/default/100-metrics-11    98.36Ki ± 0%   84.27Ki ± 0%  -14.33% (p=0.002 n=6)
Pipelines/relabel/100-metrics-11    126.6Ki ± 0%   105.5Ki ± 0%  -16.68% (p=0.002 n=6)
Pipelines/default/1000-metrics-11   997.1Ki ± 0%   856.4Ki ± 0%  -14.11% (p=0.002 n=6)
Pipelines/relabel/1000-metrics-11   1.249Mi ± 0%   1.043Mi ± 0%  -16.50% (p=0.002 n=6)
geomean                             41.76Ki        35.24Ki       -15.60%

                                  │  baseline   │                new                 │
                                  │  allocs/op  │  allocs/op   vs base               │
Pipelines/default/2-metrics-11       46.00 ± 0%    42.00 ± 0%   -8.70% (p=0.002 n=6)
Pipelines/relabel/2-metrics-11       58.00 ± 0%    52.00 ± 0%  -10.34% (p=0.002 n=6)
Pipelines/default/10-metrics-11      230.0 ± 0%    210.0 ± 0%   -8.70% (p=0.002 n=6)
Pipelines/relabel/10-metrics-11      290.0 ± 0%    260.0 ± 0%  -10.34% (p=0.002 n=6)
Pipelines/default/100-metrics-11    2.300k ± 0%   2.100k ± 0%   -8.70% (p=0.002 n=6)
Pipelines/relabel/100-metrics-11    2.900k ± 0%   2.600k ± 0%  -10.34% (p=0.002 n=6)
Pipelines/default/1000-metrics-11   24.49k ± 0%   22.49k ± 0%   -8.17% (p=0.002 n=6)
Pipelines/relabel/1000-metrics-11   30.49k ± 0%   27.49k ± 0%   -9.84% (p=0.002 n=6)
geomean                              985.0         892.5        -9.40%
```

The results were generated through the "headless" prometheus pipeline tests I added to track end-to-end scenarios related to WAL functionality that will be important as the labelstore continues to evolve.

#### Notes to the Reviewer

Fanout is the appender that must be used to emit metrics from any alloy component as it's what ensures we can multi-cast metrics to downstream consumers. The interceptor is used by most components that emit metrics but it is not a requirement, enrich + relabel use it to run their functionality, and scrape uses it to support livedebugging but it's not used by receive_http, otelcol.exporter.prometheus, and the operator which all use Fanout directly. Since any component that needs to emit metrics uses Fanout it makes the most sense to put logic there to enforce emitted metrics have a valid global seriesRef / are properly tracked for staleness.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/fanout.go
+++ b/internal/component/prometheus/fanout.go
@@ -27,7 +27,9 @@ var _ storage.Appendable = (*Fanout)(nil)
 type Fanout struct {
 	mut sync.RWMutex
 	// children is where to fan out.
-	children       []storage.Appendable
+	children []storage.Appendable
+	// ComponentID is what component this belongs to.
+	componentID    string
 	writeLatency   prometheus.Histogram
 	samplesCounter prometheus.Counter
 	ls             labelstore.LabelStore
@@ -38,7 +40,7 @@ type Fanout struct {
 }
 
 // NewFanout creates a fanout appendable.
-func NewFanout(children []storage.Appendable, register prometheus.Registerer, ls labelstore.LabelStore) *Fanout {
+func NewFanout(children []storage.Appendable, componentID string, register prometheus.Registerer, ls labelstore.LabelStore) *Fanout {
 	wl := prometheus.NewHistogram(prometheus.HistogramOpts{
 		Name:    "prometheus_fanout_latency",
 		Help:    "Write latency for sending to direct and indirect components",
@@ -54,6 +56,7 @@ func NewFanout(children []storage.Appendable, register prometheus.Registerer, ls
 
 	return &Fanout{
 		children:       children,
+		componentID:    componentID,
 		writeLatency:   wl,
 		samplesCounter: s,
 		ls:             ls,
--- a/internal/component/prometheus/scrape/scrape.go
+++ b/internal/component/prometheus/scrape/scrape.go
@@ -307,7 +307,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, fmt.Errorf("honor_metadata is an experimental feature, and must be enabled by setting the stability.level flag to experimental")
 	}
 
-	alloyAppendable := prometheus.NewFanout(args.ForwardTo, o.Registerer, ls)
+	alloyAppendable := prometheus.NewFanout(args.ForwardTo, o.ID, o.Registerer, ls)
 	scrapeOptions := &scrape.Options{
 		// NOTE: This is not Update()-able.
 		ExtraMetrics: args.ExtraMetrics,
--- a/internal/component/otelcol/exporter/prometheus/prometheus.go
+++ b/internal/component/otelcol/exporter/prometheus/prometheus.go
@@ -8,8 +8,6 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/prometheus/prometheus/storage"
-
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/otelcol"
 	"github.com/grafana/alloy/internal/component/otelcol/exporter/prometheus/internal/convert"
@@ -17,6 +15,7 @@ import (
 	"github.com/grafana/alloy/internal/component/prometheus"
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/service/labelstore"
+	"github.com/prometheus/prometheus/storage"
 )
 
 func init() {
@@ -88,7 +87,7 @@ func New(o component.Options, c Arguments) (*Component, error) {
 		return nil, err
 	}
 	ls := service.(labelstore.LabelStore)
-	fanout := prometheus.NewFanout(nil, o.Registerer, ls)
+	fanout := prometheus.NewFanout(nil, o.ID, o.Registerer, ls)
 
 	converter := convert.New(o.Logger, fanout, convertArgumentsToConvertOptions(c))
 
--- a/internal/component/prometheus/enrich/enrich.go
+++ b/internal/component/prometheus/enrich/enrich.go
@@ -97,7 +97,7 @@ func New(opts component.Options, args Arguments) (*Component, error) {
 		}
 	}
 
-	c.fanout = prometheus.NewFanout(args.ForwardTo, opts.Registerer, ls)
+	c.fanout = prometheus.NewFanout(args.ForwardTo, opts.ID, opts.Registerer, ls)
 	c.receiver = prometheus.NewInterceptor(
 		c.fanout,
 		prometheus.WithComponentID(c.opts.ID),
--- a/internal/component/prometheus/enrich/enrich_test.go
+++ b/internal/component/prometheus/enrich/enrich_test.go
@@ -5,16 +5,15 @@ import (
 	"testing"
 	"time"
 
-	prom "github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/prometheus/model/labels"
-	"github.com/prometheus/prometheus/storage"
-	"github.com/stretchr/testify/require"
-
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/discovery"
 	"github.com/grafana/alloy/internal/component/prometheus"
 	"github.com/grafana/alloy/internal/service/labelstore"
 	"github.com/grafana/alloy/internal/util"
+	prom "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/stretchr/testify/require"
 )
 
 func TestEnricher(t *testing.T) {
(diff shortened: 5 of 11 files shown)


internal/component/prometheus/scrape/scrape.go
```diff
@@ -648,11 +648,12 @@ func (c *Component) populatePromLabels(targets []discovery.Target, jobName strin
 	return allTargets
 }
 
-func (c *Component) newInterceptor(ls labelstore.LabelStore) *prometheus.Interceptor {
+// TODO add a test that series creation still works through interceptor + fanout
```
kgeckhart: This was the only place where the interceptor was providing value as it was injecting the ref that scrape would use and cache. I'll push a change to add this test tomorrow.

internal/component/prometheus/fanout.go
```diff
@@ -55,7 +53,6 @@ func NewFanout(children []storage.Appendable, componentID string, register prome
 
 	return &Fanout{
 		children:       children,
-		componentID:    componentID,
```
kgeckhart, line 58: Drive-by: This wasn't referenced anywhere

internal/component/prometheus/relabel/relabel_test.go
```diff
@@ -25,30 +25,6 @@ import (
 	"github.com/grafana/alloy/syntax"
 )
 
-func TestRelabelThroughAppend(t *testing.T) {
```
kgeckhart, line 28: This is covered through the relabel pipeline test

internal/component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -80,11 +81,20 @@ func Test(t *testing.T) {
...
-		require.Equal(t, expect, res.Timeseries)
+		if len(res.Timeseries) == 1 {
+			results = append(results, res.Timeseries[0])
+		} else if len(res.Timeseries) == 2 {
+			results = res.Timeseries
+		}
+		// When we have two results make sure they match what we expect
+		if len(res.Timeseries) == 2 {
+			require.Equal(t, expect, res.Timeseries)
+		}
```
kgeckhart: Drive-by: This test failed so I thought I broke it but it's just flaky
```
go test ./internal/component/prometheus/remotewrite/... -run '^Test$' -count 20
```
used to break it locally

internal/component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -80,11 +81,20 @@ func Test(t *testing.T) {
...
-		require.Equal(t, expect, res.Timeseries)
+		if len(res.Timeseries) == 1 {
+			results = append(results, res.Timeseries[0])
+		} else if len(res.Timeseries) == 2 {
+			results = res.Timeseries
+		}
+		// When we have two results make sure they match what we expect
+		if len(results) == 2 {
+			require.Equal(t, expect, results)
+		}
```
kgeckhart: Drive-by: This test failed so I thought I broke it but it's just flaky since it's possible one metric is remote written without the other.

go test ./internal/component/prometheus/remotewrite/... -run '^Test$' -count 20

used to break it locally

ptodev: Don't we want to have a for loop to ensure we eventually write both of the time series?

ptodev: Or we can change the remote write config to make sure both metrics are remote written at the same time? I'm not sure if that's possible though. Maybe we can set max_samples_per_send to 2 and batch_send_deadline to an hour? 😆

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ Main (unreleased)
 
 - Add `meta_cache_address` to `beyla.ebpf` component. (@skl)
 
+- Remove labelstore interactions from the prometheus interceptor simplifying prometheus pipelines. (@kgeckhart)
```
ptodev: Does this change really need a changelog entry? It is more like an internal refinment that doesn't impact users.

internal/component/prometheus/scrape/scrape.go
```diff
@@ -353,7 +350,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		unregisterer:        unregisterer,
 	}
 
-	interceptor := c.newInterceptor(ls)
+	interceptor := NewInterceptor(livedebugging.ComponentID(o.ID), c.debugDataPublisher, alloyAppendable)
```
ptodev, line 353: It's interesting that components such as `scrape` and `receive_http` also have an interceptor 🤔 It's not just for downstream components that have metrics streamed to them by upstream components.

internal/component/prometheus/remotewrite/remote_write_test.go
kgeckhart: Ugh yeah you're right this still isn't right. I'll fix it properly this time.

internal/component/prometheus/scrape/scrape.go
kgeckhart: I'll comment on this in regards to why fanout is makes more sense for the labelstore interaction vs the interceptor. `receive_http` does not actually use the interceptor because it has no further logic to inject it only uses `fanout`.

CHANGELOG.md
kgeckhart: 🤷 it's an optimization on a highly used pipeline

kgeckhart:
> I'm just not sure why the Fanout is a better place for a cache than Interceptor? In theory should the Interceptor be better, since it's the entry point to a component?

It's a good question that I didn't expand upon in the PR. `Fanout` is the appender that must be used to emit metrics from any alloy component as it's what ensures we can multi-cast metrics to downstream consumers. The interceptor is used by most components that emit metrics but it is not a requirement, `enrich` + `relabel` use it to run their functionality, and `scrape` uses it to support livedebugging but it's not used by `receive_http`, `otelcol.exporter.prometheus`, and the `operator` which all use `Fanout` directly. Since any component that needs to emit metrics uses `Fanout` it makes the most sense to put logic there to enforce emitted metrics have a valid global seriesRef.

I added a comment to Fanout to indicate it's role with global series refs and updated the PR description with the further context about why Fanout.

internal/component/prometheus/remotewrite/remote_write_test.go
kgeckhart: I went with the max samples + batch send that matches the test timeout, thanks for the idea!

internal/component/prometheus/fanout.go
kgeckhart: I'm putting it back because I think I'll need to for future refactoring of labelstore

CHANGELOG.md
ptodev: The issue is that the entry does say how the change impacts end users. If it's an optimisation I'd expect it to say something like "improved the CPU performance of prometheus.* components".

PR #4901, opened 2025-11-21
chore: move position from common/loki to an internal package for loki source

kalleep:
#### PR Description
The position package is only used by `loki.source.*`. No need to have it in a common package so I moved it under internal in loki sources.

#### Notes to the Reviewer
Some drive by import ordering fixes.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


blewis12 (approved):
Granted i'm not overly familiar with these components, but on a high level LGTM

PR #4905, opened 2025-11-21
Vendor Alertmanager config

ptodev:
#3448 would take the Alertmanager (AM) config from the AM repo:
```
github.com/prometheus/alertmanager/config
```

This PR instead copies AM config from the prometheus-operator repo. The benefit in doing this is that there are less side effects during marshal/unmarshal. The main driver is that the official struct from `prometheus/alertmanager` doesn't even behave consistently across oeprating systems. For example, a test currently fails on Windows:

```
Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -248,3 +248,3 @@
        	            	     templates:
        	            	-    - /etc/alertmanager/template/*.tmpl
        	            	+    - testdata\alertmanager\etc\alertmanager\template\*.tmpl
``` 

This is because the code checks if `templates` is an absolute path and modifies it if it's not. On windows, paths starting with `/` do not look like absolute paths. This is a problem for Alloy since Alloy could be running on a different OS from Alertmanager.

### DIFF
--- a/internal/mimir/alertmanager/types.go
+++ b/internal/mimir/alertmanager/types.go
@@ -49,7 +49,7 @@ func (c Config) String() (string, error) {
 	return string(confBytes), nil
 }
 
-// This is a standard Unmarshal functioion.
+// This is a standard Unmarshal function.
 // The benefit of having it here is that it's guaranteed to use the correct "gopkg.in/yaml.v2" package.
 func Unmarshal(data []byte) (*Config, error) {
 	var c Config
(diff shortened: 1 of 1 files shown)


internal/mimir/client/testdata/alertmanager/response.good.yml
ptodev, line 1: The "expected" output has changed in some tests, because `prometheus/alertmanager` would take the global settings and apply them to each route. It would also always print some attributes like `continue: false`, even thought it's already set to `false`. I was always aware of this but since it didn't break anything I thought the benefit of using `prometheus/alertmanager` structs is still worth the downsides.

What changed my mind is when I saw that the file paths get modified, in an OS-inconsistent way no less. I'd rather be safe and use structs which get marshalled and unmarshalled with no side effects.

PR #4610, opened 2025-10-14
feat(component/otelcol/connector): add count connector

yann-soubeyrand:
#### PR Description

This pull request adds the OpenTelemetry Collector [count connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/countconnector). This connector can be used to count spans, span events, metrics, data points, and log records.

#### Notes to the Reviewer

This PR depends on #4550.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

PR #4913, opened 2025-11-23
feat(otelcol.connector): adding connector count

hhertout:
#### PR Description

The purpose of this pool request is to add the implementation of the `connector.count` of open telemetry in alloy.

The count connector generates metrics by counting telemetry signals (traces, logs, metrics) passing through the pipeline. This provides visibility into data ingestion volumes and allows tracking of data patterns across the collection layer.

OpenTelemetry connector count implementation: [https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/countconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/countconnector)

##### Features

The count connector provides visibility into your telemetry pipeline:

- **Monitor ingestion volumes**: Track the number of spans, logs, and metrics that Alloy processes
- **Add pipeline observability**: Instrument the collector itself to detect bottlenecks or data loss
- **Filter and segment data**: Count specific signals using OTTL conditions, for example HTTP requests or error logs
- **Group by attributes**: Generate dimensional metrics grouped by service, environment, or other attributes

You can configure the count connector with custom metric definitions that use conditions to filter telemetry and attributes to group counts.

Default values are taken from the OpenTelemetry Collector repository to align with its default behavior. When you don't specify custom metrics, the connector generates standard count metrics for all telemetry types.

**Example of alloy configuration:**

```alloy
otelcol.connector.count "default" {
  // Count only HTTP GET spans
  spans {
    name        = "http_get_requests"
    description = "Number of HTTP GET requests"
    conditions  = [
      "attributes[\"http.method\"] == \"GET\"",
    ]
  }

  // Count spans grouped by service and environment
  spans {
    name        = "spans_by_service"
    description = "Spans per service and environment"
    attributes {
      key           = "service.name"
      default_value = "unknown"
    }
    attributes {
      key           = "env"
      default_value = "production"
    }
  }

  // Count error and fatal logs only
  logs {
    name        = "error_logs"
    description = "Error and fatal log records"
    conditions  = [
      "severity_number >= 17",
    ]
  }

  // Count logs by environment
  logs {
    name        = "logs_by_env"
    description = "Log records per environment"
    attributes {
      key = "env"
    }
  }

  output {
    metrics = [otelcol.exporter.otlp.default.input]
  }
}
```

Default values are taken from the OpenTelemetry Collector repository to align with its default behavior.

#### Testing

- Unit tests
- Testdata validation for OTel→Alloy converter
- Manual end-to-end testing with Prometheus export

#### Notes to the Reviewer

During development, I encountered a build failure in the vendored dependencies related to OpenTelemetry Collector's experimental profiles support. The error manifested in the debugexporter component.

It takes me a while to understand where the problem would be, but from my understanding: 
The root cause was a version mismatch between OpenTelemetry Collector components. The project uses OpenTelemetry Collector v0.139.0, but the profiles API (go.opentelemetry.io/collector/pdata/pprofile) had been upgraded to v0.140.0 in the vendor directory. The debugexporter v0.139.0 expected the older profiles API, creating an incompatibility.

I didn't want to upgrade the entire OpenTelemetry Collector dependency tree to v0.140.x for several reasons:

- It would introduce unrelated breaking changes
- The scope exceeded the PR's goal of adding the count connector
- It risked introducing additional compatibility issues with other Grafana Alloy components

As a workaround, I wrote integration tests to verify the behavior is correct.
I tested the complete behavior locally by fixing the vendor files, and running Alloy with the `connector.count`, to ensure everything runs correctly.

**Impact**: This workaround doesn't affect the count connector functionality. The profiles functionality in `debugexporter` isn't used by the count connector implementation.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


kalleep:
Hey, thanks for the pr. This is a duplicate of https://github.com/grafana/alloy/pull/4610 and https://github.com/grafana/alloy/pull/4550.

We have just not gotten around to review those

hhertout:
Hello, 

Sorry, I hadn't seen the second one you mentioned. 
For the first one, after review it, my implementation is different, with tests, and the converter has also been implemented.

Let me know if you're still interested. Otherwise, I'll close it. There's no need for another duplicate if it's not useful.

go.mod
```diff
@@ -975,6 +975,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 )
 
+require github.com/open-telemetry/opentelemetry-collector-contrib/connector/countconnector v0.139.0
```
kalleep: Lets group this with our other otel dependencies.

kalleep:
Yes I see that. I merged the pr to add support for these kind of connectors so you could rebase on that and remove your changes to connector.go.

There have not been any reactions to docs feeback in the other pr and I prefer the config structure added in this one so I am fine to go with this one instead.

internal/component/otelcol/connector/count/count_integration_test.go
```diff
@@ -0,0 +1,318 @@
+package count_test
```
kalleep: What makes these "integration" tests? 

These are just test to convert the config from alloy and verifying them like we already do in `TestArguments_UnmarshalAlloy`. So lets remove these tests and just make sure these cases are covered there

yann-soubeyrand:
Hello, I’m the author of the other PR. I’m sorry, I didn’t have the time (and still don’t have for the moment) to move it forward. I chose this (somewhat weird) config structure, because it allows disabling the default counts (for example if you’re only interested in counting logs). I’m not sure it improves performances (I guess the unused counts do not run if there are no consumer for them, right?), so maybe it’s not worth it.

kalleep:
>Hello, I’m the author of the other PR. I’m sorry, I didn’t have the time (and still don’t have for the moment) to move it forward. I chose this (somewhat weird) config structure, because it allows disabling the default counts (for example if you’re only interested in counting logs). I’m not sure it improves performances (I guess the unused counts do not run if there are no consumer for them, right?), so maybe it’s not worth it.

Gotcha, yes it won't make any difference and that I why I prefer the config structure in this pr. I merged your changes to support these kind of connectors. So lets move ahead with this pr instead

internal/component/otelcol/connector/count/count_integration_test.go
hhertout: You're right. These tests were added to verify behavior during development when I encountered the vendor build issue mentioned in the PR description. I couldn't run Alloy locally before applying the workaround, so I used these tests to ensure correctness.

I'll remove it.

hhertout:
> Hello, I’m the author of the other PR. I’m sorry, I didn’t have the time (and still don’t have for the moment) to move it forward. I chose this (somewhat weird) config structure, because it allows disabling the default counts (for example if you’re only interested in counting logs). I’m not sure it improves performances (I guess the unused counts do not run if there are no consumer for them, right?), so maybe it’s not worth it.

That makes sense. I can add that functionality if you think it would be valuable. It shouldn't be too complex...

hhertout:
> Yes I see that. I merged the pr to add support for these kind of connectors so you could rebase on that and remove your changes to connector.go.
> 
> There have not been any reactions to docs feeback in the other pr and I prefer the config structure added in this one so I am fine to go with this one instead.

I rebased, but I have to keep the if structure instead of the switch statements... due to this:

```go
// ConnectorType() int implements connector.Arguments.
func (Arguments) ConnectorType() int {
	return connector.ConnectorLogsToMetrics | connector.ConnectorTracesToMetrics | connector.ConnectorMetricsToMetrics
}
```

As I understand, ConnectorType must return the type of connector used. In this case, it returns the three types of connectors, because it allows to generate metrics based on metrics, traces and logs inputs. This is slightly different from the `span metrics` for example, where there is only one type.

In the mentioned PR, the implementation have the following code:

```go
func (args Arguments) ConnectorType() int {
	return connector.ConnectorLogsToMetrics
}
```

which is, from my point of view, not enough ? This will generate only metrics based on the log input right ?

Moreover, a switch statement doesn't work with combined flags since it checks for exact equality.
Here it works for both single-flag connectors (servicegraph, spanmetrics) and multi-flag connectors (count).

All existing connector tests pass with this approach.

WDYT ?

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.connector.count/
+aliases:
+  - ../otelcol.connector.count/ # /docs/alloy/latest/reference/components/otelcol.connector.count/
+description: Learn about otelcol.connector.count
```
clayton-cornell:
```suggestion
description: Learn about otelcol.connector.count
```

We can skip the alias on new topics. We need the alias if we move or rename a topic that's been published.

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+labels:
+  stage: experimental
+  products:
+    - oss
+title: otelcol.connector.count
+---
+
+# `otelcol.connector.count`
+
+`otelcol.connector.count` accepts spans, span events, metrics, data points, and log records from other `otelcol` components and generates metrics that count the received telemetry data.
```
clayton-cornell, line 15:
```suggestion
# `otelcol.connector.count`

{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}

`otelcol.connector.count` accepts spans, span events, metrics, data points, and log records from other `otelcol` components and generates metrics that count the received telemetry data.
```

If this is an experimental component, we need to add the shared element for experimental.

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+otelcol.connector.count "<LABEL>" {
+  output {
+    metrics = [...]
+  }
+}
+```
+
+## Arguments
+
+`otelcol.connector.count` doesn't support any arguments and can be configured using blocks only.
```
clayton-cornell:
```suggestion
The `otelcol.connector.count` component doesn't support any arguments. You can configure this component with blocks.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+| [`metrics`][metrics]             | Configures custom metrics for counting metrics.                                          | no       |
+| [`metrics` > `attributes`][attributes] | Groups metric counts by attribute values.                                          | no       |
+| [`datapoints`][datapoints]       | Configures custom metrics for counting data points.                                      | no       |
+| [`datapoints` > `attributes`][attributes] | Groups data point counts by attribute values.                                     | no       |
+| [`logs`][logs]                   | Configures custom metrics for counting log records.                                      | no       |
+| [`logs` > `attributes`][attributes] | Groups log counts by attribute values.                                                | no       |
+| [`output`][output]               | Configures where to send received telemetry data.                                        | yes      |
+| [`debug_metrics`][debug_metrics] | Configures the metrics that this component generates to monitor its state.               | no       |
```
clayton-cornell:
```suggestion
| Block                                     | Description                                                                | Required |
| ----------------------------------------- | -------------------------------------------------------------------------- | -------- |
| [`output`][output]                        | Configures where to send received telemetry data.                          | yes      |
| [`datapoints`][datapoints]                | Configures custom metrics for counting data points.                        | no       |
| `datapoints` > [`attributes`][attributes] | Groups data point counts by attribute values.                              | no       |
| [`debug_metrics`][debug_metrics]          | Configures the metrics that this component generates to monitor its state. | no       |
| [`logs`][logs]                            | Configures custom metrics for counting log records.                        | no       |
| `logs` > [`attributes`][attributes]       | Groups log counts by attribute values.                                     | no       |
| [`metrics`][metrics]                      | Configures custom metrics for counting metrics.                            | no       |
| `metrics` > [`attributes`][attributes]    | Groups metric counts by attribute values.                                  | no       |
| [`spanevents`][spanevents]                | Configures custom metrics for counting span events.                        | no       |
| `spanevents` > [`attributes`][attributes] | Groups span event counts by attribute values.                              | no       |
| [`spans`][spans]                          | Configures custom metrics for counting spans.                              | no       |
| `spans` > [`attributes`][attributes]      | Groups span counts by attribute values.                                    | no       |
```

Fix the link syntax. Pretty print. alpha sort, required first, then optional

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`name` | `string` | Metric name. | Uses default name based on telemetry type. | no
+`description` | `string` | Metric description. | `""` | no
+`conditions` | `list(string)` | OTTL expressions for filtering spans. | `[]` | no
+
+The `conditions` argument accepts a list of [OTTL][] expressions that filter which spans to count.
+If any condition matches, the span is counted.
+If no conditions are specified, all spans are counted.
```
clayton-cornell:
```suggestion
If any condition matches, the connector counts the span.
If you don't specify conditions, the connector counts all spans.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+
+You can specify multiple `spans` blocks to define different metrics.
+
+If no `spans` blocks are defined, the connector emits a default metric named `trace.span.count`.
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`name` | `string` | Metric name. | Uses default name based on telemetry type. | no
+`description` | `string` | Metric description. | `""` | no
+`conditions` | `list(string)` | OTTL expressions for filtering spans. | `[]` | no
```
clayton-cornell:
```suggestion
| Name          | Type           | Description                           | Default                                    | Required |
| ------------- | -------------- | ------------------------------------- | ------------------------------------------ | -------- |
| `conditions`  | `list(string)` | OTTL expressions for filtering spans. | `[]`                                       | no       |
| `description` | `string`       | Metric description.                   | `""`                                       | no       |
| `name`        | `string`       | Metric name.                          | Uses default name based on telemetry type. | no       |
```

Formatting, and sort

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+
+[OTTL]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl
+
+### `spanevents`
+
+The `spanevents` block configures a custom metric for counting span events.
+
+You can specify multiple `spanevents` blocks to define different metrics.
+
+If no `spanevents` blocks are defined, the connector emits a default metric named `trace.span.event.count`.
```
clayton-cornell:
```suggestion
If you don't define any `spanevents` blocks, the connector emits a default metric named `trace.span.event.count`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+This block shares the same configuration structure as [`spans`][spans].
+Refer to the [`spans`][spans] block documentation for the complete list of supported arguments and blocks.
+
+### `metrics`
+
+The `metrics` block configures a custom metric for counting metrics.
+
+You can specify multiple `metrics` blocks to define different metrics.
+
+If no `metrics` blocks are defined, the connector emits a default metric named `metric.count`.
```
clayton-cornell:
```suggestion
If you don't define any `metrics` blocks, the connector emits a default metric named `metric.count`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+This block shares the same configuration structure as [`spans`][spans].
+Refer to the [`spans`][spans] block documentation for the complete list of supported arguments and blocks.
+
+### `datapoints`
+
+The `datapoints` block configures a custom metric for counting data points.
+
+You can specify multiple `datapoints` blocks to define different metrics.
+
+If no `datapoints` blocks are defined, the connector emits a default metric named `metric.datapoint.count`.
```
clayton-cornell:
```suggestion
If you don't define any `datapoints` blocks, the connector emits a default metric named `metric.datapoint.count`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+This block shares the same configuration structure as [`spans`][spans].
+Refer to the [`spans`][spans] block documentation for the complete list of supported arguments and blocks.
+
+### `logs`
+
+The `logs` block configures a custom metric for counting log records.
+
+You can specify multiple `logs` blocks to define different metrics.
+
+If no `logs` blocks are defined, the connector emits a default metric named `log.record.count`.
```
clayton-cornell:
```suggestion
If you don't define any `logs` blocks, the connector emits a default metric named `log.record.count`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+Each unique combination of attribute values generates a separate data point on the metric.
+
+You can specify multiple `attributes` blocks within `spans`, `spanevents`, `metrics`, `datapoints`, or `logs` blocks to group by multiple attributes.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`key` | `string` | Attribute key name. | | yes
+`default_value` | `any` | Default value if the attribute doesn't exist. | | no
```
clayton-cornell:
```suggestion
| Name            | Type     | Description                                   | Default | Required |
| --------------- | -------- | --------------------------------------------- | ------- | -------- |
| `key`           | `string` | Attribute key name.                           |         | yes      |
| `default_value` | `any`    | Default value if the attribute doesn't exist. |         | no       |
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+
+The following fields are exported and can be referenced by other components:
+
+| Name    | Type               | Description                                                      |
+| ------- | ------------------ | ---------------------------------------------------------------- |
+| `input` | `otelcol.Consumer` | A value that other components can use to send telemetry data to. |
+
+`input` accepts `otelcol.Consumer` data for traces, metrics, and logs.
+
+All telemetry received through `input` is counted according to the configured blocks and emitted as metrics to the components specified in `output`.
```
clayton-cornell:
```suggestion
The component counts all telemetry received through `input` according to the configured blocks and emits metrics to the components specified in `output`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`key` | `string` | Attribute key name. | | yes
+`default_value` | `any` | Default value if the attribute doesn't exist. | | no
+
+Attribute precedence: span/log/metric attributes > scope attributes > resource attributes.
+
+### `output`
+
+{{< docs/shared lookup="reference/components/output-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
clayton-cornell, line 154:
```suggestion
### `output`

{{< badge text="Required" >}}

{{< docs/shared lookup="reference/components/output-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
```

clayton-cornell (commented):
Some suggestions.

It's not critical (not blocking), but it'd be nice for consistency if the block sections were in the same order as given in the block table (after the suggested sorting is applied).

docs/sources/reference/components/otelcol/otelcol.connector.count.md
hhertout: Thank you a lot for you review ! I've made the changes.

Though I've a question about this comment:

I don't know if I have to do it tbh... what's the default recommendation for this ? Should I mark it as experimental ?

clayton-cornell: There are a couple of things here we should verify.

1. Is this `experimental`? If yes, then we need the lable.stage in the metadata (already there) and the shared text block include (suggested). If no, then we need to remove or change the label.stage in the metadata and then skip this suggestion.
1.  Is this a Community supported component? If yes, then we need to add the community label in the metadata, and add the shared community text block around line 12-13.

I'll defer to @grafana/grafana-alloy-maintainers for the decision on which way to go, and I can provide the doc suggestions to align.

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -0,0 +1,370 @@
...
+[output]: #output
+[debug_metrics]: #debug_metrics
+
+### `spans`
+
+The `spans` block configures a custom metric for counting spans.
+
+You can specify multiple `spans` blocks to define different metrics.
+
+If no `spans` blocks are defined, the connector emits a default metric named `trace.span.count`.
```
clayton-cornell:
```suggestion
If you don't define any `spans` blocks, the connector emits a default metric named `trace.span.count`.
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
hhertout: Thank you for your answer !

kalleep:
Yes you are right in that the other change to `connector.go` is not enough and while you change would work for count it would not work for other connectors if we decide to add them.

E.g. [failoverconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/failoverconnector/factory.go#L23-L25) would not work without significant changes to this. This can connect to more than just metrics so it would fail [this check](https://github.com/hhertout/alloy/blob/feature/otel-connector-count/internal/component/otelcol/connector/connector.go#L205-L207).

So I think we should change that check a bit 

```go
if !outputsToMetrics(connectorType) && len(next.Metrics) > 0 {
	return errors.New("this connector cannot output metrics")
}

if !outputsToLogs(connectorType) && len(next.Logs) > 0 {
	return errors.New("this connector cannot output logs")
}

if !outputsToTraces(connectorType) && len(next.Traces) > 0 {
	return errors.New("this connector cannot output traces")
}
```

And add some helper functions 

```go
func outputsToMetrics(connectorType int) bool {
	return (connectorType&ConnectorTracesToMetrics) != 0 ||
		(connectorType&ConnectorMetricsToMetrics) != 0 ||
		(connectorType&ConnectorLogsToMetrics) != 0
}

func outputsToLogs(connectorType int) bool {
	return (connectorType&ConnectorTracesToLogs) != 0 ||
		(connectorType&ConnectorMetricsToLogs) != 0 ||
		(connectorType&ConnectorLogsToLogs) != 0
}

func outputsToTraces(connectorType int) bool {
	return (connectorType&ConnectorTracesToTraces) != 0 ||
		(connectorType&ConnectorMetricsToTraces) != 0 ||
		(connectorType&ConnectorLogsToTraces) != 0
}
```

For this to work we also need to update iota enum

```go
const (
	ConnectorTracesToTraces = iota + 1 // thin is the change we need
	ConnectorTracesToMetrics
	ConnectorTracesToLogs
	ConnectorMetricsToTraces
	ConnectorMetricsToMetrics
	ConnectorMetricsToLogs
	ConnectorLogsToTraces
	ConnectorLogsToMetrics
	ConnectorLogsToLogs
)
```

And because we have not implemented any support for `*ToLogs` and `*ToTraces` yet we could return error that they are not supported yet.

hhertout:
> Yes you are right in that the other change to `connector.go` is not enough and while you change would work for count it would not work for other connectors if we decide to add them.
> 
> E.g. [failoverconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/failoverconnector/factory.go#L23-L25) would not work without significant changes to this. This can connect to more than just metrics so it would fail [this check](https://github.com/hhertout/alloy/blob/feature/otel-connector-count/internal/component/otelcol/connector/connector.go#L205-L207).
> 
> So I think we should change that check a bit
> 
> ```go
> if !outputsToMetrics(connectorType) && len(next.Metrics) > 0 {
> 	return errors.New("this connector cannot output metrics")
> }
> 
> if !outputsToLogs(connectorType) && len(next.Logs) > 0 {
> 	return errors.New("this connector cannot output logs")
> }
> 
> if !outputsToTraces(connectorType) && len(next.Traces) > 0 {
> 	return errors.New("this connector cannot output traces")
> }
> ```
> 
> And add some helper functions
> 
> ```go
> func outputsToMetrics(connectorType int) bool {
> 	return (connectorType&ConnectorTracesToMetrics) != 0 ||
> 		(connectorType&ConnectorMetricsToMetrics) != 0 ||
> 		(connectorType&ConnectorLogsToMetrics) != 0
> }
> 
> func outputsToLogs(connectorType int) bool {
> 	return (connectorType&ConnectorTracesToLogs) != 0 ||
> 		(connectorType&ConnectorMetricsToLogs) != 0 ||
> 		(connectorType&ConnectorLogsToLogs) != 0
> }
> 
> func outputsToTraces(connectorType int) bool {
> 	return (connectorType&ConnectorTracesToTraces) != 0 ||
> 		(connectorType&ConnectorMetricsToTraces) != 0 ||
> 		(connectorType&ConnectorLogsToTraces) != 0
> }
> ```
> 
> For this to work we also need to update iota enum
> 
> ```go
> const (
> 	ConnectorTracesToTraces = iota + 1 // thin is the change we need
> 	ConnectorTracesToMetrics
> 	ConnectorTracesToLogs
> 	ConnectorMetricsToTraces
> 	ConnectorMetricsToMetrics
> 	ConnectorMetricsToLogs
> 	ConnectorLogsToTraces
> 	ConnectorLogsToMetrics
> 	ConnectorLogsToLogs
> )
> ```
> 
> And because we have not implemented any support for `*ToLogs` and `*ToTraces` yet we could return error that they are not supported yet.

Just to make sure I understand, the goal of these changes is to open the `connector.go` to any connectorType, and open the function to other connector features (present in OpenTelemetry but not in Alloy), even if it's not already implemented right ? So in summary, take advantage of this PR to add more genericity for future implementation ?

hhertout:
I also took this opportunity to add documentation on the helper functions about the bitwise checks.
Let me know if it's understandable and if you'd like any other modifications.

All tests are passing.

internal/component/otelcol/connector/connector.go
```diff
@@ -255,3 +269,69 @@ func (p *Connector) CurrentHealth() component.Health {
...
+// Examples:
+//
+//   - spanmetrics connector: connectorType = ConnectorTracesToMetrics (value 2)
+//     Test: (2 & ConnectorTracesToMetrics) = (2 & 2) = 2 != 0 ✓ outputs metrics
+//
+//   - count connector: connectorType = ConnectorTracesToMetrics | ConnectorLogsToMetrics (values 2 | 8)
+//     Test: (10 & ConnectorTracesToMetrics) = (10 & 2) = 2 != 0 ✓ outputs metrics
+//
+//   - traces-only connector: connectorType = ConnectorTracesToTraces (value 1)
+//     Test: (1 & ConnectorTracesToMetrics) = (1 & 2) = 0 ✗ does NOT output metrics
```
kalleep:
```suggestion
// outputsToMetrics checks if the connector can output metrics by testing if any of the
// *ToMetrics flags are set in the connectorType.
```

internal/component/otelcol/connector/connector.go
```diff
@@ -255,3 +269,69 @@ func (p *Connector) CurrentHealth() component.Health {
...
+// Examples:
+//
+//   - logs connector: connectorType = ConnectorTracesToLogs (value 3)
+//     Test: (3 & ConnectorTracesToLogs) = (3 & 3) = 3 != 0 ✓ outputs logs
+//
+//   - multi-signal: connectorType = ConnectorTracesToLogs | ConnectorLogsToLogs (values 3 | 9)
+//     Test: (11 & ConnectorTracesToLogs) = (11 & 3) = 3 != 0 ✓ outputs logs
+//
+//   - metrics-only connector: connectorType = ConnectorTracesToMetrics (value 2)
+//     Test: (2 & ConnectorTracesToLogs) = (2 & 3) = 2 ✗ does NOT output logs
```
kalleep:
```suggestion
// outputsToLogs checks if the connector can output logs by testing if any of the
// *ToLogs flags are set in the connectorType.
```

internal/component/otelcol/connector/connector.go
```diff
@@ -255,3 +269,69 @@ func (p *Connector) CurrentHealth() component.Health {
...
+// Examples:
+//
+//   - traces passthrough: connectorType = ConnectorTracesToTraces (value 1)
+//     Test: (1 & ConnectorTracesToTraces) = (1 & 1) = 1 != 0 ✓ outputs traces
+//
+//   - multi-signal: connectorType = ConnectorTracesToTraces | ConnectorLogsToTraces (values 1 | 7)
+//     Test: (7 & ConnectorTracesToTraces) = (7 & 1) = 1 != 0 ✓ outputs traces
+//
+//   - metrics-only connector: connectorType = ConnectorTracesToMetrics (value 2)
+//     Test: (2 & ConnectorTracesToTraces) = (2 & 1) = 0 ✗ does NOT output traces
```
kalleep:
```suggestion
// outputsToTraces checks if the connector can output traces by testing if any of the
// *ToTraces flags are set in the connectorType.
```

kalleep (commented):
Its fine to add comments but these a bit excessive, its fine to just describe what the function(s) is used for.

clayton-cornell:
@kalleep Before you merge, there's an open question in the docs... is this experimental? Community? Either/or will require a tweak to the docs before we merge.

kalleep:
@clayton-cornell good catch. Its [alpha](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/countconnector) so I think we should have it as experimental .

docs/sources/reference/components/otelcol/otelcol.connector.count.md
hhertout: I’ve added the experimental note to the documentation, in line with the latest comment.

internal/component/otelcol/connector/count/count.go
```diff
@@ -0,0 +1,224 @@
...
+	kindSpanEvents = "spanevents"
+	kindMetrics    = "metrics"
+	kindDataPoints = "datapoints"
+	kindLogs       = "logs"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:      "otelcol.connector.count",
+		Stability: featuregate.StabilityGenerallyAvailable,
```
kalleep: You also need to change here and set it to experimental

hhertout: Forgot it... sorry

PR #4816, opened 2025-11-10
OTel v0.139.0

dehaansa:
#### PR Description

From the upstream changelogs, the following changes are included.

```

- `pkg/exporterhelper`: Add default values for `sending_queue::batch` configuration. (#13766)
  Setting `sending_queue::batch` to an empty value now results in the same setup as the default batch processor configuration.
- `exporterhelper`: Add new `exporter_queue_batch_send_size` and `exporter_queue_batch_send_size_bytes` metrics, showing the size of telemetry batches from the exporter. (#12894)

- `all`: Add `keep_alives_enabled` option to ServerConfig to control HTTP keep-alives for all components that create an HTTP server. (#13783)
- `all`: Add `keep_alives_enabled` option to ServerConfig to control HTTP keep-alives for all components that create an HTTP server. (#13783)

- `processor/k8sattributes`: Support extracting deployment name purely from the owner reference (#42530)
- `processor/k8sattributesprocessor`: Add support for k8s.cronjob.uid attribute in k8sattributesprocessor (#42557)- `k8sattributesprocessor`: 
- `processor/k8sattributes`: Support extracting labels and annotations from k8s DaemonSets (#37957)
- `processor/k8sattributes`: Support extracting labels and annotations from k8s Jobs (#37957)

- `processor/resourcedetection`: Add the `dt.smartscape.host` resource attribute to data enriched with the Dynatrace detector (#43650)
- `processor/resourcedetection`: Add Openstack Nova resource detector to gather Openstack instance metadata as resource attributes (#39117)
- `processor/resourcedetection`: Add Azure availability zone to resourcedetectionprocessor (#40983)
- `resourcedetectionprocessor`: Added Oracle Cloud resource detection support to resourcedetectionprocessor, enabling automatic population of Oracle Cloud-specific resource attributes. (#35091)
- `processor/resourcedetection`: Add support for akamai cloud in resourcedetectionprocessor (#42543)
- `processor/resourcedetectionprocessor`: Add support for Scaleway host image attributes in the resourcedetectionprocessor (#42682)
- `processor/resourcedetectionprocessor`: Add support for Vultr cloud provider in the resourcedetectionprocessor (#42569)
- `processor/resourcedetection`: Add support for DigitalOcean in resourcedetectionprocessor (#42803)
- `processor/resourcedetection`: Add support for upcloud in resourcedetectionprocessor (#42801)
[...]

PR #4920, opened 2025-11-24
fix(pyroscope.ebpf): Use meter noop in otel-ebpf-profiler

simonswine:
With a recent change, the default meter provider was no longer
initialized. This change disables the otel-ebpf-profiler internal
metrics to avoid warnings like this to appear:

```
WARN[0121] Invalid metric id 102, skipping
WARN[0121] Invalid metric id 272, skipping
```


simonswine:
@bobrik thanks for noticing and reporting:

https://github.com/grafana/alloy/pull/4816#issuecomment-3567163773

simonswine:
@jharvey10 it would be great, if you could include this into the next rc for v1.12

Issue #398, opened 2023-07-26
Expose more configuration options for `prometheus.exporter.cloudwatch` component


Jeinhaus:
### Background

We were previously using the [yace cloudwatch-exporter](https://github.com/nerdswords/yet-another-cloudwatch-exporter) with the [`-labels-snake-case` option](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/master/docs/configuration.md#configuration) to keep exported metrics consistent with other prometheus metrics.
As far as I can see, the Grafana Agent's `prometheus.exporter.cloudwatch` component does not expose this configuration option.

Did I overlook these and/or are there any plans to implement them?
It would ease our migration to Grafana Agent, because we wouldn't have to change existing dashboards that use these metrics.

### Proposal

Expose existing configuration options in the `prometheus.exporter.cloudwatch` component

pyo-seyeol:
I'd like to add my strong support for this feature request. The lack of a `delay` setting is a critical issue for reliable production monitoring.

We've observed the exact problem described in this issue: when querying aggregated CloudWatch statistics (e.g., Average over a 5-minute period), the exporter fetches an incomplete, "in-progress" data point because it queries up to now().

This results in two major problems:
1. Unstable Metrics: The `Average` value for the most recent data point fluctuates for several minutes until the `SampleCount` for the period is complete.
2. False Alerts: Alerts based on these unstable values are prone to firing incorrectly.

A `delay` setting is the standard, proven solution for this. It ensures the exporter queries for metric data that has had time to fully aggregate. For example, by setting delay = "5m" for a metric with a period = "5m", we can ensure we only fetch stable, complete data points.

This functionality is a key feature in yet-another-cloudwatch-exporter (YACE), and bringing this capability to the Alloy `prometheus.exporter.cloudwatch` component would make it significantly more robust and reliable for production use.

Thank you for your work on Alloy and for considering this important improvement.

pyo-counting:
I'd like to add my strong support for this feature request. The lack of a `delay` setting is a critical issue for reliable production monitoring.

When querying aggregated CloudWatch statistics (e.g., Average over a 5-minute period), the exporter fetches an incomplete, "in-progress" data point because it queries up to `now()`.

This results in two major problems:
1. Unstable Metrics: The `Average` value for the most recent data point fluctuates for several minutes until the `SampleCount` for the period is complete.
2. False Alerts: Alerts based on these unstable values are prone to firing incorrectly.

[...]

PR #4936, opened 2025-11-25
feat: add delay option to CW exporter

### LINKED ISSUE #3929: Cloudwatch metrics are too low
### What's wrong?

We've been migrating from YACE to using Alloy with the embedded cloudwatch exporter. We are seeing an issue with the exported metrics: the values for the latest datapoint are not the same as the metrics in Cloudwatch ([bug in YACE](https://github.com/prometheus-community/yet-another-cloudwatch-exporter/issues/1549)). In our YACE config, we had a delay for each metric for a few minutes, which would allow the metric datapoint value to stabilize in Cloudwatch, and we would then store the stable value to Mimir.

As such, I do not think setting the delay to 0 in the [exporter config](https://github.com/grafana/alloy/blob/main/internal/static/integrations/cloudwatch_exporter/config.go#L246) is the correct thing to do - the latest datapoint (and often the second-to-last one) are missing values, which makes the averages and sums be too low, regardless of the length and the delay parameters used. Having a delay is the only way we have so far found that allow us to fetch accurate metrics, so I believe this is a bug in Alloy.

### Steps to reproduce

1. Fetch some volatile metric from Cloudwatch using Alloy's Cloudwatch exporter. I've used Lambda invocations for a function that is invoked ~3k times per second, with period of 1 minute, length of 5 minutes and scrape interval of 1 minute.
2. Observe both the actual Cloudwatch metric and the one imported using the exporter, e.g. in a same dashboard. Notice that the imported metric is never quite the same than the one from Cloudwatch. The imported metric in the screenshot is offset by -2 minutes to more closely align with the shape of the Cloudwatch metric.

Image

3. The latest value from Cloudwatch is low, but it grows for more than a minute, until it reaches a stable value that no longer changes.

4. This could be solved with changing the delay parameter of the metric to 2 minutes and using Cloudwatch timestamps. For some other metrics, 5 minute delay could be more suitable.

### System information

Linux amd64 and arm64

### Software version

Grafana Alloy 1.7.1

### Configuration

```text
prometheus.exporter.cloudwatch "default" {
	sts_region = "eu-central-1"

	discovery {
		type    = "AWS/Lambda"
		regions = ["eu-central-1"]

		metric {
			name       = "Invocations"
			statistics = ["Sum"]
			period     = "1m"
			length     = "5m"
		}
	}
}

prometheus.scrape "cloudwatch_exporter" {
	clustering {
		enabled = true
	}

	targets         = prometheus.exporter.cloudwatch.default.targets
	scrape_interval = "1m"
	forward_to      = [prometheus.relabel.cloudwatch_exporter_relabel.receiver]
}
```

### Logs

```text

```

tmeijn:
#### PR Description

- Add `delay` option to `prometheus.exporter.cloudwatch` component to delay scraping of metrics to account for CloudWatch ingestion latency.

- Export `yace_.*` metrics from the underlying YACE Exporter to `prometheus.exporter.cloudwatch`.

Fixes #3929, Partially #398

#### Notes to the Reviewer

> [!WARNING] 
> This is vibe coded, so it might not be the best. However testing a local build it works as I'd expect.

Validation of `yace_.*` metrics:

image

`delay` option present in alloy ui:

image

CloudWatch vs Prometheus

image

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.fluentforward.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.fluentforward.md
@@ -18,7 +18,7 @@ title: otelcol.receiver.fluentforward
 `otelcol.receiver.fluentforward` is a wrapper over the upstream OpenTelemetry Collector [`fluentforward`][] receiver.
 Bug reports or feature requests will be redirected to the upstream repository, if necessary.
 
-[`fluentforward`]: <https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/{{>< param "OTEL_VERSION" >}}/receiver/fluentforwardreceiver
+[`fluentforward`]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/{{< param "OTEL_VERSION" >}}/receiver/fluentforwardreceiver
 {{< /admonition >}}
 
 You can specify multiple `otelcol.receiver.fluentforward` components by giving them different labels.


internal/static/integrations/cloudwatch_exporter/config.go
```diff
@@ -231,28 +232,10 @@ func toYACEConfig(c *Config) (yaceModel.JobsConfig, bool, error) {
 	if err != nil {
 		return yaceModel.JobsConfig{}, fipsEnabled, err
 	}
-	PatchYACEDefaults(&modelConf)
 
 	return modelConf, fipsEnabled, nil
 }
 
-// PatchYACEDefaults overrides some default values YACE applies after validation.
-func PatchYACEDefaults(yc *yaceModel.JobsConfig) {
```
kalleep, line 240: Not sure I am reading this correctly but would it not be impossible to have 0 delay now, so the defaults would not work anymore? 

If that is correct this would be a breaking change

kalleep (commented):
Changes looks mostly fine to me but I left one question, also I can see that @kgeckhart was involved with the original pr so maybe he have more context around this

internal/static/integrations/cloudwatch_exporter/config.go
tmeijn: Hey @kalleep if I look at the current main code of YACE here: https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/0c9677d91836f0a4150a55172a0ce5081574b407/pkg/config/config.go#L364-L373

If I interpret it correctly, it basically ignores the `delay` setting on the metric level and only respects the `delay` at the job level. This is also why I did not set the configuration on the metric level as is done with `period`, but instead on the job level.

If a delay is not configured by the user, it will be 0 if I am correct.

tmeijn: I also think this test asserts there is no regression in this regard: https://github.com/grafana/alloy/blob/298d5c97c652075be61c6e73515678382c89dacc/internal/component/prometheus/exporter/cloudwatch/config_test.go#L337-L366

tmeijn:
Hey @kalleep, @kgeckhart anything I can do to help this move forwards?

internal/static/integrations/cloudwatch_exporter/config.go
kgeckhart: You're correct @tmeijn support for it was removed upstream because it was never implemented and properly supporting delay is really annoying as it changes how you query cloudwatch.

PR #4940, opened 2025-11-26
proposal: Change architecture for loki pipelines

kalleep:
This pr is used to add a proposal to change how loki pipeline works to address some issues we see today.


docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,153 @@
...
+Processing components would:
+Implement `Appendable` to return an `Appender` that runs processing for each entry and fan-out to all downstream components.
+
+Sink components would:
+Implement `Appendable` to return an `Appender` that buffers entries until either `Commit` or `Rollback` is called.
+
+Pros:
+* Increase throughput of log pipelines.
+* A way to signal back to the source
+* Iterator-like pipeline - one entry at a time
```
thampiotr: Why is this a pro?

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,153 @@
...
+* Transaction semantics provide better error handling
+* Aligns with Prometheus patterns already in use
```
thampiotr: I'm also not sure this is a pro 😅 Prometheus patterns come from a database world, we are a telemetry collector.

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,153 @@
...
+We could adopt this pattern for loki pipelines by having:
+```go
+type Appendable interface {
+    Appender(ctx context.Context) Appender
+}
+
+type Appender interface {
+    Append(entry Entry) error
+    Commit() error
+    Rollback() error
```
thampiotr: There's no requirement for us to have Rollback as part of the problem statement. Isn't this complicating things without good reason?

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,153 @@
...
+loki.write "loki" {}
+```
+
+`loki.source.file` will tail all files from targets and compete to send on the channel exposed by `loki.process`. Only one entry will be processed by each stage configured in `loki.process`. If a reload happens or if Alloy is shutting down, logs could be silently dropped.
+
+## Proposal 0: Do nothing
+This architecture works in most cases, it will be hard to use slow components such as `secretfilter` because a lot of the time it's too slow.
+It's also hard to use Alloy as a gateway for loki pipelines with e.g. `loki.source.api` due to the limitations listed above.
+
+## Proposal 1: Chain function calls
```
thampiotr: Could you outline for both proposal how the transition would look? 
I do think Proposal 1 looks more attractive, but I could be convinced to do something in-between Proposal 1 and 2. 

What I'm not sure about is whether we flip a switch here or make this a gradual transition. It could affect the behaviour of the pipelines, especially the error propagation. But if it is better in every way, a flip-a-switch approach is viable. If it has some trade-offs we may need to make it an opt-in, gradual transition.

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,153 @@
...
+Solving the issues listed above.
+
+A batch of entries should be considered successfully consumed when they are queued up for sending. We could try to extend this to when it was successfully sent over the wire, but that could be considered an improvement at a later stage.
+
+Pros:
+* Increase throughput of log pipelines.
+* A way to signal back to the source
+Cons:
+* We need to rewrite all loki components with this new pattern and make them safe to call in parallel.
+* We go from an iterator-like pipeline to passing slices around. Every component would have to iterate over this slice and we need to make sure it's safe to mutate because of fan-out.
```
thampiotr: Ah yeah, that's true for mutating components. I think we will need to copy data to prevent mutating (other options like overlays are not really a thing we have in Go). So if we need to copy, we have two main options:

1. Fan-out always copies and components are free to mutate as they get a safe copy
2. Fan-out does not copy, components need to be careful not to mutate original entries and make copies where needed. 

Option 2 is much more performant as majority of entries won't need to be mutated I'd guess. So when we get a slice and requires no changes, we forward it as-is. But if it needs changes, we need to create a new slice, reuse all the entries in it that don't need mutating, but create copies for those that need mutating. I think that can work, but we will need to be careful.

docs/design/4940-reliable-loki-pipelines.md
kalleep: Well it should have better performance characteristics and you spotted one of then in this comment https://github.com/grafana/alloy/pull/4940#discussion_r2571430720

The other one would be that every component would have to perform a loop on all entries passed individually. In proposal 2 it would always be one loop, starting from the source.

docs/design/4940-reliable-loki-pipelines.md
kalleep: That's true, I can remove that as a pro

docs/design/4940-reliable-loki-pipelines.md
kalleep: Maybe but I can think of one case where it would be useful to have `Rollback`. We can name these differently but I just took what we could use from prometheus interfaces.

For a source like `loki.source.file` I don't think that Rollback is all that useful but for sources like `loki.source.api` it could be.

Imagine we get a api request. We start a "transaction" by calling Appender with a context. A api request could be aborted in two scenarios, client aborts the request or we are restarting / shutting down.

In both these cases we should not commit the batch and return a proper error response. If client aborted due to e.g. timeout they won't care about the response but we would return it either way.

The implementation could be context aware (we pass a context when we acquire Appender)  but here we have two signal we need to handle and I have not figured out a good solution to that other than having some action that would abort the batch. 

With channels this is easier because we can just [listen for both signals](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/api/internal/lokipush/push_api_server.go#L264-L272)

docs/design/4940-reliable-loki-pipelines.md
kalleep: Yes exactly. The second option you is most likely better for performance but we also always need to handle it properly.

The first should be the "safest". You could optimize this a bit.

1. If we only fan-out to one component we can skip the clone.
2. If we fan-out to multiple components we copy to all but one that gets the original.

docs/design/4940-reliable-loki-pipelines.md
kalleep: Both proposal 1 and 2 are similar in that we change from channel based pipeline to function based pipelines. So I don't think the transition would be any different between them. I also think that the transition could be a separate discussion, like do we want to be able to still use the current pipeline code for a while.

Would you like me to add a section to discuss the difference between proposal 1 and 2?

kalleep: >I do think Proposal 1 looks more attractive, but I could be convinced to do something in-between Proposal 1 and 2.

That was my preferred solution first too before I though some more about the actual implementation. In proposal 2 we don't need to handle slices. We don't have to loop in each component individually but it all just becomes one loop. So it's closer to what we have today but we still get the benefits what this proposal aims to solve with the additional benefit that sources now also have control to abort if they need to.

docs/design/4940-reliable-loki-pipelines.md
kalleep: Btw I think all of our components have potential to mutate. The only one that don't is `loki.echo`.

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,223 @@
...
+
+Pipeline components like loki.process, loki.relabel, loki.write, and fan-out all implement Consumer. These components are stream-agnostic — they process whatever entries they receive and do not perform sharding. The entries they receive will already be grouped by stream.
+
+**Note on label mutation**: loki.process and loki.relabel can change an entry's labels, which changes its stream identity. This is safe because sharding happens before these components run, based on the original stream labels. All entries from the same original stream are on the same worker and processed in order — this ordering is preserved through mutations. When entries reach loki.write, it reshards internally based on final labels. If entries from different original streams end up in the same target stream after relabeling, their interleaving is fine — they had no ordering relationship before relabeling.
```
kgeckhart: I think the placement of this section about sharding is a bit confusing as the reader needs to understand a bit about the sharding strategy to really understand why this matters. But the sharding strategy for pipelines is introduced later.

I believe if I read more about the sharding strategy below my question will be answered but perhaps this comment should be after the sharding strategy is introduced or the sharding consumer should be introduced first.

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Error handling and backpressure
+
+Errors propagate synchronously back to the source through Consume() return values. The source has not committed yet (no 200 sent, no positions advanced), so it can always handle errors safely.
+
+| Error                   | What happens                                     | Source behavior                                                                                  |
+| ----------------------- | ------------------------------------------------ | ------------------------------------------------------------------------------------------------ |
+| WAL I/O failure         | loki.write.Consume() returns retryable error     | API: respond 429/503. File: don't advance position. Both retry.                                  |
+| Queue full (congestion) | loki.write.Consume() blocks                      | API: request eventually times out → retryable error. File: reader blocks, positions not updated. |
```
kgeckhart, line 186: Should we expose a similar `block_on_overflow` behavior as the OTel sending_queue?

docs/design/4940-reliable-loki-pipelines.md
kgeckhart: Reading further didn't quite answer my question, is loki.write a "fan-in" step? 

> When entries reach loki.write, it reshards internally based on final labels.

Reads as though it could be that loki.write is sharding again from the original shard. But I believe it's suggesting that it's going to fan-in from the original shards to fan-out again. Right? 

If that's right, do we need to do hash based sharding up front or can we do a basic pool with a fan-in at the end? 

If that's not right, won't we get out of order writes by nature of having shards of shards?

docs/design/4940-reliable-loki-pipelines.md
```diff
@@ -0,0 +1,223 @@
...
+
+**Note on label mutation**: loki.process and loki.relabel can change an entry's labels, which changes its stream identity. This is safe because sharding happens before these components run, based on the original stream labels. All entries from the same original stream are on the same worker and processed in order — this ordering is preserved through mutations. When entries reach loki.write, it reshards internally based on final labels. If entries from different original streams end up in the same target stream after relabeling, their interleaving is fine — they had no ordering relationship before relabeling.
+
+The existing loki.Entry already carries `model.LabelSet` which identifies the stream, so no changes to the entry type are needed.
+
+### ShardingConsumer
+
+Source components (loki.source.api, loki.source.file) receive entries that may belong to multiple streams. A ShardingConsumer sits at the boundary between source and pipeline. It groups entries by stream, dispatches each group to a worker goroutine (by stream hash), and waits for all workers to complete. Each worker calls a plain Consumer chain (e.g. loki.process → loki.write) with entries from a single stream.
```
kgeckhart, line 100: Could we use a name like `LogSource` here instead of `ShardingConsumer`? If I'm writing a component which is a log source I'm only going to know to use `ShardingConsumer` because every other implementation does vs it being the obvious thing needed to source logs for a log pipeline.

docs/design/4940-reliable-loki-pipelines.md
kalleep: We already do :)

docs/design/4940-reliable-loki-pipelines.md
kalleep: Yes I see how this can be a bit confusing but loki.write it self is agnostic to streams but each endpoint already do [internal sharding based on the hash](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/client/shards.go#L377-L378)

docs/design/4940-reliable-loki-pipelines.md
kalleep: Sure I am open to different naming.

We have this `This is for illustration purposes only. The actual naming may be different.` over the Consumer interface definition but it could be applied to all of them.

docs/design/4940-reliable-loki-pipelines.md
kgeckhart: I think that's fine but don't we end up with ordering issues in this case?

> If entries from different original streams end up in the same target stream after relabeling, their interleaving is fine — they had no ordering relationship before relabeling.

Like we relabel to drop the pod label, which could have us merging multiple streams that were originally separate. loki.write will re-queue them as expected but their queue'ing order could be incorrect now, right?

kalleep: Yes we do but I am not really sure what we can do in that specific case, we could theoretically try to fix ordering but that would required us to mutate queued batches and would only work as long when batches are not forwarded. 

I will think about this a bit, maybe we could at least have some kind priority when adding to non-queued batch based on timestamp but not too sure about that.

thampiotr: wouldn't this already be an issue right now? I think if you don't have OOO enabled you need to be careful about your labels not to make two separate streams look like one... would this design change things?

kalleep: Yes this would already be the case. If you start to relabel so that all logs have the same labels then we will have ordering issues both currently and with these changes. ShardingConsumer do not change that

kalleep: Yes you are right. Will move it under the sharding section

kgeckhart: I was mostly confused by the note which suggested there's no ordering problems when there can be order problems.

Could we include a section on ordering after architecture before, `### Error handling and backpressure` collecting this note + the "not all backends support out of order", acknowledge mutating components can break ordering today and that it's not going to be fixed with this?

PR #4941, opened 2025-11-26
feat(loki.process): Add regex field to logfmt and json stages

### LINKED ISSUE #2156: loki.process --> stage.json, stage.logfmt - extract all label:value pairs without explicit expression

timonegk:
#### PR Description

After I added the `regex` field to the `structured_metadata` stage in #4629, this PR adds the field to `logfmt` and `json` as well. As pointed out by @bmundt6 in https://github.com/grafana/alloy/issues/2156#issuecomment-3539729140, this is required to fully close #2156.

Continues the fix for #2156.

#### Notes to the Reviewer

In the documentation, I removed the section about the logfmt-parser preserving the type of booleans and numbers because these types are actually always converted to strings.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -500,12 +500,15 @@ The following arguments are supported:
 
 | Name             | Type          | Description                                           | Default | Required |
 | ---------------- | ------------- | ----------------------------------------------------- | ------- | -------- |
-| `expressions`    | `map(string)` | Key-value pairs of JMESPath expressions.              |         | yes      |
+| `expressions`    | `map(string)` | Key-value pairs of JMESPath expressions.              |         | no       |
+| `regex`          | `string`      | Regular expression matched against JSON keys.         |         | no       |
 | `drop_malformed` | `bool`        | Drop lines whose input can't be parsed as valid JSON. | `false` | no       |
 | `source`         | `string`      | Source of the data to parse as JSON.                  | `""`    | no       |
 
 The `expressions` field is the set of key-value pairs of JMESPath expressions to run.
-The map key defines the name with which the data is extracted, while the map value is the expression used to populate the value.
+The map key defines the name used to extract the data, while the map value is the expression used to populate the value.
+
+The `regex` field is a regular expression. All keys in the JSON source matching the regular expression are extracted.
 
 When configuring a JSON stage, the `source` field defines the source of data to parse as JSON.
 By default, this is the log line itself, but it can also be a previously extracted value.
@@ -673,19 +676,19 @@ The `stage.logfmt` inner block configures a processing stage that reads incoming
 
 The following arguments are supported:
 
-| Name      | Type          | Description                                    | Default | Required |
-| --------- | ------------- | ---------------------------------------------- | ------- | -------- |
-| `mapping` | `map(string)` | Key-value pairs of `logmft` fields to extract. |         | yes      |
-| `source`  | `string`      | Source of the data to parse as `logfmt`.       | `""`    | no       |
+| Name      | Type          | Description                                     | Default | Required |
+| --------- | ------------- | ----------------------------------------------- | ------- | -------- |
+| `mapping` | `map(string)` | Key-value pairs of `logmft` fields to extract.  |         | no       |
+| `regex`   | `string`      | Regular expression matched against logfmt keys. |         | no       |
+| `source`  | `string`      | Source of the data to parse as `logfmt`.        | `""`    | no       |
+
+The `mapping` field is the set of key-value pairs.
+The map key defines the name used to extract the data, while the map value is the logfmt field used to populate the value.
+
+The `regex` field is a regular expression. All logfmt fields matching the regular expression are extracted.
 
 The `source` field defines the source of data to parse as `logfmt`.
 When `source` is missing or empty, the stage parses the log line itself, but it can also be used to parse a previously extracted value.
-
-This stage uses the [go-logfmt][] unmarshaler, so that numeric or boolean types are unmarshalled into their correct form.
-The stage doesn't perform any other type conversions.
-If the extracted value is a complex type, it's treated as a string.
-
-[go-logfmt]: https://github.com/go-logfmt/logfmt
 
 The following log line and stages demonstrates how this works.
 
--- a/internal/component/loki/process/stages/json.go
+++ b/internal/component/loki/process/stages/json.go
@@ -143,58 +143,25 @@ func (j *jsonStage) processEntry(extracted map[string]any, entry *string) error 
 		return errors.New(ErrMalformedJSON)
 	}
 
-	for n, e := range j.expressions {
-		r, err := e.Search(data)
+	for name, expr := range j.expressions {
+		rawResult, err := expr.Search(data)
 		if err != nil {
 			if Debug {
 				level.Debug(j.logger).Log("msg", "failed to search JMES expression", "err", err)
 			}
 			continue
 		}
-
-		switch r.(type) {
-		case float64:
-			// All numbers in JSON are unmarshaled to float64.
-			extracted[n] = r
-		case string:
-			extracted[n] = r
-		case bool:
-			extracted[n] = r
-		case nil:
-			extracted[n] = nil
-		default:
-			// If the value wasn't a string or a number, marshal it back to json
-			jm, err := json.Marshal(r)
-			if err != nil {
-				if Debug {
-					level.Debug(j.logger).Log("msg", "failed to marshal complex type back to string", "err", err)
-				}
-				continue
-			}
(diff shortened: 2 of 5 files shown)


internal/component/loki/process/stages/json.go
```diff
@@ -164,6 +174,31 @@ func (j *jsonStage) processEntry(extracted map[string]interface{}, entry *string
...
+					extracted[key] = nil
+				default:
+					jm, err := json.Marshal(value)
+					if err != nil {
+						if Debug {
+							level.Debug(j.logger).Log("msg", "failed to marshal complex type back to string", "err", err)
+						}
+						continue
+					}
+					extracted[key] = string(jm)
```
kalleep: This is exactly the same as the branch when using expressions, please put this logic in a function.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -653,19 +656,19 @@ The `stage.logfmt` inner block configures a processing stage that reads incoming
...
+| Name      | Type          | Description                                     | Default | Required |
+| --------- | ------------- | ----------------------------------------------- | ------- | -------- |
+| `mapping` | `map(string)` | Key-value pairs of `logmft` fields to extract.  |         | no       |
+| `regex`   | `string`      | Regular expression matched against logfmt keys. |         | no       |
+| `source`  | `string`      | Source of the data to parse as `logfmt`.        | `""`    | no       |
 
-The `source` field defines the source of data to parse as `logfmt`.
-When `source` is missing or empty, the stage parses the log line itself, but it can also be used to parse a previously extracted value.
+The `mapping` field is the set of key-value pairs.
+The map key defines the name with which the data is extracted, while the map value is the logfmt field used to populate the value.
```
```suggestion
The map key defines the name used to extract the data, while the map value is the logfmt field used to populate the value.
```

We could simplify here and remove the "with which".  It's not wrong, but if we can simplify the language used, we should.

clayton-cornell (commented):
Really minor suggestion.  Otherwise the docs look fine. Over to the dev team for code review.

docs/sources/reference/components/loki/loki.process.md
timonegk: I copied that section from the json stage, so I made the change there as well.

timonegk:
@ptodev since you reviewed #4629, could you also review this?

pan3793:
@clayton-cornell @ptodev hi maintainers, can we move this forward? https://github.com/grafana/alloy/issues/2156 receives dozens of 👍, this is quite a useful feature.

kalleep:
Sorry this one slipped through. I will put this in my backlog to review

internal/component/loki/process/stages/logfmt.go
```diff
@@ -98,13 +107,22 @@ func (j *logfmtStage) Process(labels model.LabelSet, extracted map[string]interf
...
+	regexExtractedEntriesCount := 0
 	for decoder.ScanRecord() {
 		for decoder.ScanKeyval() {
+			// handle "mapping"
 			mapKey, ok := j.inverseMapping[string(decoder.Key())]
 			if ok {
 				extracted[mapKey] = string(decoder.Value())
-				extractedEntriesCount++
+				mappingExtractedEntriesCount++
+			} else if j.regex.String() != "" {
```
kalleep: What I don't like here is that now logfmt behaves differently to json.

In json we try both expressions and regex if they are configured but here we prefer mapping, why?

internal/component/loki/process/stages/logfmt.go
```diff
@@ -98,13 +107,22 @@ func (j *logfmtStage) Process(labels model.LabelSet, extracted map[string]interf
...
+				// handle "regex"
+				fmt.Println(j.regex.String(), string(decoder.Key()))
```
kalleep: Remove

timonegk: done

internal/component/loki/process/stages/logfmt.go
timonegk: Must have been a mistake, I changed it so that both have the JSON behavior.

internal/component/loki/process/stages/json.go
timonegk: I extracted the logic to a function and also renamed some variables to avoid single-letter names.

internal/component/loki/process/stages/json.go
```diff
@@ -170,6 +172,31 @@ func (j *jsonStage) processEntry(extracted map[string]interface{}, entry *string
 	return nil
 }
 
+// extractWithType returns the value if it's a simple type (string, number, bool),
```
kalleep:
```suggestion
// simplifyType returns the value if it's a simple type (string, number, bool),
```

internal/component/loki/process/stages/json.go
```diff
@@ -170,6 +172,31 @@ func (j *jsonStage) processEntry(extracted map[string]interface{}, entry *string
...
+// otherwise, it returns it as a JSON string
+func (j *jsonStage) simplifyType(value any) (any, error) {
```
kalleep: We never to anything with the error at call sites, we just use it as a signal, could we change to return (any, bool) instead

internal/component/loki/process/stages/logfmt.go
```diff
@@ -98,13 +107,22 @@ func (j *logfmtStage) Process(labels model.LabelSet, extracted map[string]interf
...
+			// handle "mapping"
 			mapKey, ok := j.inverseMapping[string(decoder.Key())]
 			if ok {
 				extracted[mapKey] = string(decoder.Value())
-				extractedEntriesCount++
+				mappingExtractedEntriesCount++
+			}
+			// handle "regex"
+			if j.regex.String() != "" {
+				if j.regex.MatchString(string(decoder.Key())) {
```
kalleep: nit: We are doing `string(decoder.Key())` in three places. Instead do it once in the beginning of the loop
`key := string(decoder.Key())` and use that

timonegk:
@kalleep I have integrated your comments from the second review and rebased onto main.

kalleep:
Hey some linting issues left to fix, after that we should be good to go

timonegk:
@kalleep should be fixed now

kalleep (approved):

PR #4947, opened 2025-11-27
fix cluster drift alert

### LINKED ISSUE #4945: alloy mixin alerting for cluster drift broken
### Component(s)

_No response_

### What's wrong?

We've deployed Alloy-Mixin into our cluster and noticed that it reports ClusterConfigurationDrift. We use a mostly unmodified default configuration from k8s-monitoring and found the culprit to be about alloy-logs and alloy-singleton having different configuration hashes but tried to be combined.

# Solution

One possible solution would be to include the `app` in the alert which would group only the right values together. Maybe it's even considered a bug in the k8s-monitoring deployment where the jobs should have different names.

### Steps to reproduce

https://github.com/grafana/alloy/blob/35d1f1c5d567e219b439bfa66dccede7f6ab813f/operations/alloy-mixin/alerts/clustering.libsonnet#L81-L84

Here the metric is split up by cluster, namespace, job, cluster_name but alloy logs is not clustered and therefore both metrics have no cluster_name. Cluster, Namespace and Job are equal which results in the alert being fired. I'm unsure if the job label should be different but in the k8s-monitoring installation it's always set up to be `integrations/alloy`.

### System information

_No response_

### Software version

Grafana Allox Mixin v1.11.3

### Configuration

```text

```

### Logs

```text

```

### Tip

tobiaskohlbau:
In case the job name is equal (for e.g. k8s-monitoring installation) multiple instances of alloy get deployed into a cluster alloy-logs, alloy-singleton which results in an alert firing for configuration drift. This CL addresses the issue by grouping together on the app label which matches alloy-logs and alloy-metrics in the case of k8s-monitoring.

Fixes #4945

#### Notes to the Reviewer

This is only a proposed change, lets discuss the details in #4945 first and settle on the right solution.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/alloy-mixin/alerts/clustering.libsonnet
+++ b/operations/alloy-mixin/alerts/clustering.libsonnet
@@ -80,11 +80,11 @@ local alert = import './utils/alert.jsonnet';
           'ClusterConfigurationDrift',
           if enableK8sCluster then |||
             count without (sha256) (
-                max by (cluster, namespace, sha256, job, cluster_name) (alloy_config_hash and on(cluster, namespace, job, cluster_name) cluster_node_info)
+                max by (cluster, namespace, sha256, job, cluster_name, app) (alloy_config_hash and on(cluster, namespace, job, cluster_name) cluster_node_info)
             ) > 1
           ||| else |||
             count without (sha256) (
-                max by (sha256, job) (alloy_config_hash and on(job) cluster_node_info)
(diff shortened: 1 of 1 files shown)


tobiaskohlbau:
Superseded by #5181

PR #4948, opened 2025-11-27
feat: Add parcas symbols upload to pyroscope.ebpf

korniltsev-grafanista:
#### PR Description
This PR introduces debug information upload using parca's API implemented in https://github.com/grafana/pyroscope/pull/4648 on the pyroscope

This feature is implemented in two places: the profiler and the proxy

```go
type debuginfo.Appender interface {
    Upload(j UploadJob) // for the profiler: fanout to all the nested children 
    Client() debuginfogrpc.DebuginfoServiceClient // for the proxy - retrieve the first grpc client available and proxy data to the client
}
```
The profile uploader actually lives in the `pyroscope.write` component to support per-write component uploads(ugh who needs this? no-one!) , but this is an implementation detail and it's hidden from the users.
The proxy only can proxy data to the one client - it can not fan-out.

The `parca` package is copied from the parca's repo with some adjustments (`logrus` -> `logkit`, the grpc client is passed as an argument to the upload routine instead of creating and owning the grpc client inside the uploader)

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/pyroscope/util/internal/configurations/ebpf.txt
+++ b/internal/component/pyroscope/util/internal/configurations/ebpf.txt
@@ -22,11 +22,13 @@ pyroscope.write "local_pyroscope" {
 }
 
 pyroscope.ebpf "default" {
+    sample_rate = 19
 	targets      = discovery.relabel.all_process_targets.output
 	targets_only = false
 	forward_to   = [pyroscope.write.local_pyroscope.receiver]
 	demangle     = "full"
 	go_enabled   = false
+	reporter_unsymbolized_stubs = false
 
 	debug_info {
 		upload                  = true


internal/component/pyroscope/enrich/enrich.go
```diff
@@ -107,6 +109,10 @@ func (e *enrichAppendable) AppendIngest(ctx context.Context, profile *pyroscope.
 	return e.component.fanout.Appender().AppendIngest(ctx, enrichedProfile)
 }
 
+func (e *enrichAppendable) UploadDebugInfo(ctx context.Context, fileID libpf.FileID, fileName string, buildID string, open func() (process.ReadAtCloser, error)) {
+	e.component.fanout.Appender().UploadDebugInfo(ctx, fileID, fileName, buildID, open)
```
korniltsev-grafanista: remove this, this is not needed

go.mod
```diff
@@ -271,7 +273,7 @@ require (
 	go.opentelemetry.io/collector/scraper/scraperhelper v0.139.0
 	go.opentelemetry.io/collector/service v0.139.0
 	go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux v0.45.0
-	go.opentelemetry.io/ebpf-profiler v0.0.0-00010101000000-000000000000
+	go.opentelemetry.io/ebpf-profiler v0.0.0-20250519164423-009a07f3803c
```
korniltsev-grafanista: make it 00000

go.mod
```diff
@@ -968,8 +970,14 @@ require (
 
 require github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awss3receiver v0.139.0
 
+require (
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awss3receiver v0.139.0
```
korniltsev-grafanista: check why did this change?

internal/component/pyroscope/write/write.go
```diff
@@ -145,42 +155,119 @@ func New(
...
+		dataPath:      dataPath,
+		receiver:      receiver,
 	}, nil
 }
 
 // Run implements Component.
 func (c *Component) Run(ctx context.Context) error {
+	var receiverCtx context.Context
+
+	c.mu.Lock()
```
korniltsev-grafanista, line 168: this looks fishy, what if context is canceled at the time of Update

internal/component/pyroscope/ebpf/args.go
```diff
@@ -28,6 +29,8 @@ type Arguments struct {
 	LazyMode            bool                   `alloy:"lazy_mode,attr,optional"`
 	DeprecatedArguments DeprecatedArguments    `alloy:",squash"`
 
+	DebugInfoOptions debuginfo.Arguments `alloy:"debug_info,block,optional"`
```
marcsanmi, line 32: Do you think the `debug_info` configuration could be moved from `pyroscope.ebpf` to `pyroscope.write` for better coherence? The debuginfo uploader is created per-endpoint in `pyroscope.write` but configured in `pyroscope.ebpf`. What would happen if multiple ebpf components share the same pyroscope.write? I think one config would be silently ignored

korniltsev-grafanista: It makes sense from the engineering perspective, but not from the user perspective. In fact it initially was there and then I moved it to the ebpf component.
I think having this configuration in the `write` component is confusing for users - imagine users configuring scraping/proxy/java profiling - they would all stumble into configuration knobs that has nothing to do with the sending data and are only relevant if youre using ebpf component.

PR #4951, opened 2025-11-28
feat(prometheus.exporter.azure): expose concurrency settings

akselleirv:
#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
@@ -85,7 +85,6 @@ You can use the following arguments with `prometheus.exporter.azure`:
 | `validate_dimensions`               | `bool`         | Enable dimension validation in the azure SDK.                                                                                                            | `false`                                                                       | no       |
 | `concurrency_subscription`          | `int`          | Number of subscriptions that can concurrently send metric requests.                                                                                      | `5`                                                                           | no       |
 | `concurrency_subscription_resource` | `int`          | Number of concurrent metric requests per resource within a subscription.                                                                                 | `10`                                                                          | no       |
-| `enable_caching`                    | `bool`         | Enable internal caching to reduce redundant API calls.                                                                                                   | `false`                                                                       | no       |
 
 The list of available `resource_type` values and their corresponding `metrics` can be found in [Azure Monitor essentials][].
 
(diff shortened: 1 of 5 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
```diff
@@ -109,10 +112,15 @@ Valid values for `azure_cloud_environment` are `azurecloud`, `azurechinacloud`,
 `validate_dimensions` is disabled by default to reduce the number of Azure exporter instances required when a `resource_type` has metrics with varying dimensions.
 When `validate_dimensions` is enabled you will need one exporter instance per metric + dimension combination which is more tedious to maintain.
 
-`timespan` and `interval` are used to control how metrics are queried from Azure Monitor. 
-The exporter queries metrics over the `timespan` and returns the most recent datapoint at the specified `interval`. 
+`timespan` and `interval` are used to control how metrics are queried from Azure Monitor.
```
clayton-cornell:
```suggestion
Use `timespan` and `interval` to control how the exporter queries metrics from Azure Monitor.
```

Small change to make the statement active instead of passive.

internal/static/integrations/azure_exporter/azure_exporter.go
```diff
@@ -72,7 +71,13 @@ func (e Exporter) MetricsHandler() (http.Handler, error) {
 			return
 		}
 
-		prober := metrics.NewMetricProber(ctx, logEntry, nil, settings, e.ConcurrencyConfig)
+		// Use concurrency settings from merged config to support query parameter overrides
+		concurrencyConfig := azure_config.Opts{}
+		concurrencyConfig.Prober.ConcurrencySubscription = mergedConfig.ConcurrencySubscription
+		concurrencyConfig.Prober.ConcurrencySubscriptionResource = mergedConfig.ConcurrencySubscriptionResource
+		concurrencyConfig.Prober.Cache = mergedConfig.EnableCaching
```
kgeckhart: Can we keep this PR to just the concurrency settings and remove the cache flag support? 

This isn't the only code that would have to change to enable caching and it's a bit convoluted to support properly. I can see resource graph response caching being valuable but the metric cache feels unnecessary unless you have misconfiguration alloy to scrape more often than the timespan config.

akselleirv: @kgeckhart apologies for the long responds time, but I removed the cache flag in https://github.com/grafana/alloy/pull/4951/commits/f78ef9d5e951977d5f1dc9853b8cd4a5fba51442

Issue #4793, opened 2025-11-06
Add admonitions in  discovery.kubernetes  to clarify 


clayton-cornell:
> We can add admonitions to those topics... I would convert the first suggested note into a Caution... stands out more and suitable because there's a bigger impact if things are missed.  The second suggested note... I think would be OK as a note... location... probably the same general area as suggested. 

 _Originally posted by @clayton-cornell in [#4787](https://github.com/grafana/alloy/issues/4787#issuecomment-3498371980)_

Issue #4787, opened 2025-11-06
Docs feedback: It should be specified that each daemonset pod watches ALL pods logs in the cluster


adrian-salas:
### URL

https://grafana.com/docs/alloy/latest/configure/kubernetes/

### Feedback

For the context, we have had issues lately on our Azure cluster where it seemed that we reached AKS Inflight limits.

Azure Inflight determines if the Kube API is throttled.
Using Alloy on its default configuration on 4 cluster saturated the Kube API in a way that new objects couldn't even be created.

Since each alloy pod (we could have 20 nodes, so 20 pods) would watch all logs of the cluster, the path containerLogs was definitely saturated, and Azure Inflight too.

In the end, we could add the following rule for Alloy on our pod configuration:

```
        rule {
          source_labels = ["__meta_kubernetes_pod_node_name"]
          action        = "keep"
          regex         = env("K8S_NODE_NAME")
        }
```

with `K8S_NODE_NAME` being and extraEnv given to alloy 

```
  extraEnv:
    - name: K8S_NODE_NAME
      valueFrom:
        fieldRef:
          fieldPath: spec.nodeName
```

That configuration allows retrieval of pods logs that are on the same node as the alloy pods that watches

Firstly, there was a huge memory decrease, each pod was consuming ~3Go of memory per pod.
Now Alloy uses around 300Mo per pod.

The path for containerLogs had a night an day difference too:

<img width="1901" height="721" alt="Image" src="https://github.com/user-attachments/assets/74ffbc8f-1a6f-41a8-a28b-d8577b241a63" />

And also, specific for Azure on our case, a decrease in Inlfight requests:

<img width="1901" height="721" alt="Image" src="https://github.com/user-attachments/assets/ebaedef5-5ddd-441f-a75e-bfdda5fab60a" />

--
Consequences:

Since Loki deduplicates entries, we hadn't noticed that alloy was actually watching all pods each at first.

We have had 4 saturations on our Kubernetes API, that affected our pipelines (runners were throttled)
Also, our monitoring has been affected for a while, since we have had alloy disabled on multiple times.

Of course, there was time spend from our team as well as increased costs for our clusters since we tried changing AKS Inflight tier (around 300$ per cluster)

To summary, i think for the potential cost of that behavior, i think it should be better specified in the documentation that this happens.
Of course, using clustering mode for Alloy would mitigate the issue too.
Either way, all of what I described wasn't really clear for us when reading the documentation

### Tip

[...]

PR #4955, opened 2025-11-28
chore(doc): admonition for API saturation risk

adrian-salas:
Added caution about Kubernetes API saturation risk when running Alloy as a DaemonSet, along with recommended solutions

Proposition for #4787 #4793 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


docs/sources/reference/components/discovery/discovery.kubernetes.md
```diff
@@ -15,6 +15,24 @@ title: discovery.kubernetes
...
+  See the example in the [Limit to only Pods on the same node](#limit-to-only-pods-on-the-same-node) section below for a configuration snippet that uses label selectors and environment variables to achieve this.
+- **Clustering mode:** For larger deployments, consider setting up Alloy in clustering mode.
+- **Monitor resource consumption:** Regularly check API server throttling, memory usage, and inflight requests, especially on cloud-managed clusters (e.g., Azure AKS).
+
+Failure to properly configure Alloy can result in degraded cluster performance, increased cloud costs, and operational risk.  
+Please review your configuration carefully and consult the examples below.
+{{< /admonition >}}
+
 If you supply no connection information, this component defaults to an in-cluster configuration.
 A kubeconfig file or manual connection settings can be used to override the defaults.
```
clayton-cornell:
```suggestion
If you supply no connection information, this component defaults to an in-cluster configuration.
A kubeconfig file or manual connection settings can be used to override the defaults.

## Performance considerations

By default, `discovery.kubernetes` discovers resources across all namespaces in your cluster.
In DaemonSet deployments, this means every {{< param "PRODUCT_NAME" >}} Pod watches all resources, which can increase API server load.

For better performance and reduced API load:

- Use the [`namespaces`](#namespaces) block to limit discovery to specific namespaces.
- Use [`selectors`](#selectors) to filter resources by labels or fields.  
- Consider the node-local example in [Limit to only Pods on the same node](#limit-to-only-pods-on-the-same-node).
- Use clustering mode for larger deployments to distribute the discovery load.
- Monitor API server metrics like request rate, throttling, and memory usage, especially on managed clusters.
```

How about this? We simplify the information and give the reader some really clear things they can do to handle the performance issues. This is active (gives specific and clear steps) and I think says the same thing as the Caution did.

adrian-salas: That actually way clearer indeed.

IMO with that message, the "problem" will happen less

PR #4969, opened 2025-12-01
database_observability: Always log an explain_plan_output for each query, showing status and reason.

rgeyer:
#### PR Description
This ensures that there is always a log message emitted for each query from the explain plan collector, no matter what the outcome of that explain plan collection was.

This can be used to inform the UI when a query was skipped because it contained a reserved word, for example.

Resolves grafana/grafana-dbo11y-app#1735

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -46,13 +46,47 @@ const selectDigestsForExplainPlan = `
 
 const selectExplainPlanPrefix = `EXPLAIN FORMAT=JSON `
 
+func sendPremptedExplainPlansOutput(schemaName string, dbVersion string, digest string, generatedAt string, entryHandler loki.EntryHandler, result database_observability.ExplainProcessingResult, reason string) error {
```
cristiangreco: Any chance that this method can be reused even for the success case? We do almost similar stuff between success failure (except parsing the plan).

Could this be a collector method, to avoid passing the entryHandler?

internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -517,12 +551,22 @@ func (c *ExplainPlans) populateQueryCache(ctx context.Context) error {
 			return strings.EqualFold(schema, schemaName)
 		}) {
 
+			err := sendPremptedExplainPlansOutput(schemaName, c.dbVersion, digest, time.Now().Format(time.RFC3339), c.entryHandler, database_observability.ExplainProcessingResultSkipped, "query belongs to excluded schema")
```
cristiangreco: nit: we use `time.Now().Format(time.RFC3339)` repeatedly, maybe worth just setting a value at the top of the loop?

internal/component/database_observability/explain_plan_output.go
```diff
@@ -143,3 +164,28 @@ type ExplainPlanNodeDetails struct {
 	SortKeys      []string                  `json:"sortKeys,omitempty"`
 	Warning       *string                   `json:"warning,omitempty"`
 }
+
+func ExtractExplainPlanOutputFromLogMsg(lokiEntry loki.Entry) (ExplainPlanOutput, error) {
```
cristiangreco: If this is used only for testing, maybe could be moved to a shared `_test.go` file, or we could add a comment that says it's here just for testing.

internal/component/database_observability/mysql/collector/explain_plans.go
rgeyer: Yeah, good call, That improves this implementation quite a bit.

Implemented the change. :)

internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -46,28 +46,18 @@ const selectDigestsForExplainPlan = `
...
 
 	planNode, err := parseTopLevelPlanNode(logger, qblock)
-	output.Plan = planNode
 	if err != nil {
-		return output, err
+		return nil, err
 	}
 
-	return output, nil
+	return &planNode, nil
```
cristiangreco: Now that this method has been cut down a lot, perhaps we could just `return parseTopLevelPlanNode(logger, qblock)`?

cristiangreco (approved):

PR #4978, opened 2025-12-02
dbo11y: stop tracking alloy's own queries in mysql

cristiangreco:
#### PR Description
This PR introduces a change for excluding "own queries" in mysql: we add a new `setup_actors` collector that checks (and optionally updates) settings for the currently connected user. By default it just checks that `enabled=NO` and logs out if that's not the case. Optionally, the component can auto-update this setting itself. This behaviour is disabled by default, though: it follows the same pattern of `setup_consumers` where two boolean settings are required to enable the auto-update. Additionally, auto-update requires `INSERT` and `UPDATE` permissions (docs on the website).

Relates to https://github.com/grafana/grafana-dbo11y-app/issues/1307

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -140,10 +140,10 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
 
 ### `setup_actors`
 
-| Name                       | Type       | Description                                                                    | Default | Required |
-|----------------------------|------------|--------------------------------------------------------------------------------|---------|----------|
-| `collect_interval`         | `duration` | How frequently to check if `setup_actors` are configured correctly.    | `"1h"`  | no       |
+| Name                       | Type       | Description                                                            | Default | Required |
+| -------------------------- | ---------- | ---------------------------------------------------------------------- | ------- | -------- |
 | `auto_update_setup_actors` | `boolean`  | Whether to enable updating `performance_schema.setup_actors` settings. | `false` | no       |
+| `collect_interval`         | `duration` | How frequently to check if `setup_actors` are configured correctly.    | `"1h"`  | no       |
 
 
 ## Example
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -37,14 +37,15 @@ You can use the following arguments with `database_observability.mysql`:
 
 The following collectors are configurable:
 
-| Name              | Description                                              | Enabled by default |
-|-------------------|----------------------------------------------------------|--------------------|
-| `query_details`   | Collect queries information.                             | yes                |
-| `schema_details`  | Collect schemas and tables from `information_schema`.    | yes                |
-| `query_samples`   | Collect query samples.                                   | yes                |
-| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.    | yes                |
-| `locks`           | Collect queries that are waiting/blocking other queries. | no                 |
-| `explain_plans`   | Collect explain plans information.                       | yes                |
+| Name              | Description                                                  | Enabled by default |
+|-------------------|--------------------------------------------------------------|--------------------|
+| `query_details`   | Collect queries information.                                 | yes                |
+| `schema_details`  | Collect schemas and tables from `information_schema`.        | yes                |
+| `query_samples`   | Collect query samples.                                       | yes                |
+| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.        | yes                |
+| `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
+| `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
+| `explain_plans`   | Collect explain plans information.                           | yes                |
 
 ## Blocks
 
@@ -55,6 +56,7 @@ You can use the following blocks with `database_observability.mysql`:
 | [`cloud_provider`][cloud_provider]   | Provide Cloud Provider information.               | no       |
 | `cloud_provider` > [`aws`][aws]      | Provide AWS database host information.            | no       |
 | [`setup_consumers`][setup_consumers] | Configure the `setup_consumers` collector.        | no       |
+| [`setup_actors`][setup_actors]       | Configure the `setup_actors` collector.           | no       |
 | [`query_details`][query_details]     | Configure the queries collector.                  | no       |
 | [`schema_details`][schema_details]   | Configure the schema and table details collector. | no       |
 | [`explain_plans`][explain_plans]     | Configure the explain plans collector.            | no       |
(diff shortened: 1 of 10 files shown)


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -275,7 +286,11 @@ func (c *QuerySamples) fetchQuerySample(ctx context.Context) error {
 		queryTextField = queryTextClause
 	}
 
-	query := fmt.Sprintf(selectPgStatActivity, queryTextField)
+	ignoreAlloyQueriesClauseField := ""
+	if c.ignoreAlloyQueries {
+		ignoreAlloyQueriesClauseField = ignoreAlloyQueriesClause
+	}
+	query := fmt.Sprintf(selectPgStatActivity, queryTextField, ignoreAlloyQueriesClauseField)
```
ghost: <strike>String-formatted SQL query detected. This could lead to SQL injection if the string is not sanitized properly. Audit this call to ensure the SQL is not manipulable by external data.</strike>

🧹 _Removed in commit b39850a8aa6e106d9d14fcf7b60d57ad078cbe90_ 🧹

fridgepoet:
Neat!
This other PR https://github.com/grafana/alloy/pull/4991 is similar, but in Postgres, right? I am wondering if the collector names should reflect the outcome rather than the underlying settings names.
If in both cases we're not tracking alloy's queries, should the collectors be named after their purpose?

cristiangreco:
> Neat! This other PR #4991 is similar, but in Postgres, right? 

Yes, more details in the ticket

> I am wondering if the collector names should reflect the outcome rather than the underlying settings names. If in both cases we're not tracking alloy's queries, should the collectors be named after their purpose?

Yeah I see the issue, I'm not sure what a good name could be though. I guess we might have more of these kind of "check-adjust-report" functionality in the future, and maybe they shouldn't be collectors (or maybe it's ok, if they also report data?).

fridgepoet:
Agreed, calling this a "collector" is also a little disorienting. 
At least for Postgres #4991 `exclude_current_user` is somewhat self-explanatory. 
Mostly raised my comment just to highlight how "user-friendly" we are being for this. I don't know if users will know of the top of their head what "setup_actors" is and even I will likely not remember forever.

Also I linked #4991 to question if the two collectors should have the same name if they share the same end goal?

(Maybe something like `exclude_own_queries` or `exclude_self_queries`?)

cristiangreco:
> I don't know if users will know of the top of their head what "setup_actors" is and even I will likely not remember forever.
> Also I linked #4991 to question if the two collectors should have the same name if they share the same end goal?

Don't think should have the same name, they do different things, and trying to make things too homogeneous across different engines is counter-productive imo. TBH there's a risk we might be overthinking this: as long as our docs explain what the collector is used for, it's not that much important trying to find an alternative to the name "collector" (at this stage, at least).

internal/component/database_observability/mysql/collector/setup_actors.go
```diff
@@ -0,0 +1,172 @@
...
+	var user string
+	err := c.dbConnection.QueryRowContext(ctx, selectUserQuery).Scan(&user)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to get current user", "err", err)
+		return err
+	}
+
+	var enabled, history string
+	err = c.dbConnection.QueryRowContext(ctx, selectQuery, user).Scan(&enabled, &history)
+	if err == sql.ErrNoRows {
```
fridgepoet: this could be an errors.Is()

fridgepoet (approved):

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -37,14 +37,15 @@ You can use the following arguments with `database_observability.mysql`:
...
-| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.    | yes                |
-| `locks`           | Collect queries that are waiting/blocking other queries. | no                 |
-| `explain_plans`   | Collect explain plans information.                       | yes                |
+| Name              | Description                                                  | Enabled by default |
+|-------------------|--------------------------------------------------------------|--------------------|
+| `query_details`   | Collect queries information.                                 | yes                |
+| `schema_details`  | Collect schemas and tables from `information_schema`.        | yes                |
+| `query_samples`   | Collect query samples.                                       | yes                |
+| `setup_consumers` | Collect enabled `performance_schema.setup_consumers`.        | yes                |
+| `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
```
matthewnolf, line 46: It doesn't look like it is enabled by default?

internal/component/database_observability/mysql/collector/setup_actors.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (c *SetupActors) Stop() {
+	c.cancel()
+	c.running.Store(false)
+}
+
+func (c *SetupActors) checkSetupActors(ctx context.Context) error {
+	var user string
+	err := c.dbConnection.QueryRowContext(ctx, selectUserQuery).Scan(&user)
```
matthewnolf: Can this be done on startup? Is there a need to get the current user every run?

matthewnolf (approved):

docs/sources/reference/components/database_observability/database_observability.mysql.md
cristiangreco: It's in `component.go`: 

```go
collector.SetupActorsCollector: true
```

internal/component/database_observability/mysql/collector/setup_actors.go
cristiangreco: Good call, I can move this operation to before entering the loop.

As an edge case, you could change the user in the DSN string and reload the Alloy config - but that should restart the component and all collectors, so should be ok.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -135,6 +138,14 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
 | `auto_enable_setup_consumers`    | `boolean`  | Whether to enable some specific `performance_schema.setup_consumers` settings. | `false` | no       |
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
 
+### `setup_actors`
+
+| Name                       | Type       | Description                                                                    | Default | Required |
+|----------------------------|------------|--------------------------------------------------------------------------------|---------|----------|
+| `collect_interval`         | `duration` | How frequently to check if `setup_actors` are configured correctly.    | `"1h"`  | no       |
+| `auto_update_setup_actors` | `boolean`  | Whether to enable updating `performance_schema.setup_actors` settings. | `false` | no       |
```
```suggestion
| Name                       | Type       | Description                                                            | Default | Required |
| -------------------------- | ---------- | ---------------------------------------------------------------------- | ------- | -------- |
| `auto_update_setup_actors` | `boolean`  | Whether to enable updating `performance_schema.setup_actors` settings. | `false` | no       |
| `collect_interval`         | `duration` | How frequently to check if `setup_actors` are configured correctly.    | `"1h"`  | no       |
```

Just a minor tidy and alpha sort

matthewnolf (approved):

PR #4994, opened 2025-12-03
loki.write: always read and close response body

kalleep:
While investigate issues I noticed that we don't always fully read the response body.

From go source code docs:
```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body. The default HTTP client's Transport may not
// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```

### DIFF
--- a/internal/component/common/loki/client/consumer_fanout.go
+++ b/internal/component/common/loki/client/consumer_fanout.go
@@ -364,9 +364,11 @@ func (c *client) send(ctx context.Context, tenantID string, buf []byte) (int, er
 
 	// NOTE: it is important in go to fully read the body and
 	// close it so that the connection can be reused.
-	// We only partailly read the body if we encounter a non 2xx error
+	// We only partially read the body if we encounter a non 2xx error
 	// so we should always consume whats left.
 	// https://github.com/golang/go/blob/32a9804c7ba3f4a0e0bd26cc24b9204860a49ec8/src/net/http/response.go#L59-L64
+	// It is unclear that we always need to drain the body but
+	// https://github.com/golang/go/issues/60240#issuecomment-1551060433 seems to indicate that we should.
 	defer func() {
 		_, _ = io.Copy(io.Discard, resp.Body)
 		_ = resp.Body.Close()
--- a/internal/component/common/loki/client/consumer_wal.go
+++ b/internal/component/common/loki/client/consumer_wal.go
@@ -393,7 +393,7 @@ func (c *walClient) runSendOldBatches() {
 				c.sendQueue.enqueue(qb)
 			}
 
-			batchesToFlush = batchesToFlush[:0] // renew slide
+			batchesToFlush = batchesToFlush[:0] // renew slice
 		}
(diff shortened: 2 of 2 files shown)


internal/component/common/loki/client/consumer_wal.go
```diff
@@ -464,17 +462,15 @@ func (c *walClient) sendBatch(ctx context.Context, tenantID string, batch *batch
...
-		level.Error(c.logger).Log("msg", "final error sending batch", "status", status, "tenant", tenantID, "error", err)
-		// If the reason for the last retry error was rate limiting, count the drops as such, even if the previous errors
-		// were for a different reason
-		dropReason := ReasonGeneric
-		if batchIsRateLimited(status) {
-			dropReason = ReasonRateLimited
-		}
-		c.metrics.droppedBytes.WithLabelValues(c.cfg.URL.Host, tenantID, dropReason).Add(bufBytes)
-		c.metrics.droppedEntries.WithLabelValues(c.cfg.URL.Host, tenantID, dropReason).Add(float64(entriesCount))
+	level.Error(c.logger).Log("msg", "final error sending batch, no retries left, dropping data", "status", status, "tenant", tenantID, "error", err)
```
kalleep, line 465: If we get here the error is always non-nil. Fixed this for fanout before but forgot to fix it here too

internal/component/common/loki/client/consumer_fanout.go
```diff
@@ -363,7 +361,16 @@ func (c *client) send(ctx context.Context, tenantID string, buf []byte) (int, er
...
 	}
-	defer lokiutil.LogError(c.logger, "closing response body", resp.Body.Close)
+
+	// NOTE: it is important in go to fully read the body and
+	// close it so that the connection can be reused.
+	// We only partailly read the body if we encounter a non 2xx error
+	// so we should always consume whats left.
+	// https://github.com/golang/go/blob/32a9804c7ba3f4a0e0bd26cc24b9204860a49ec8/src/net/http/response.go#L59-L64
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
```
ptodev, line 373:
```suggestion
```
Why do you think it's necessary to read the response?

kalleep: Well without fully reading the body a connection cannot be reused.
`if the Body is not read to completion and closed`. 

I looked at [remote_write](https://github.com/prometheus/prometheus/blob/main/storage/remote/client.go#L299-L302) and they are doing the same there too.

ptodev: Do you mind adding a comment to [here](https://github.com/golang/go/issues/60240#issuecomment-1551060433), as well as a note that remote_write does it too? It's very weird and I don't understand why it's not documented/ I'm also not even sure if it's necessary, but better to be safe - I'm happy to add it in.

PR #5014, opened 2025-12-05
update go to v1.25.5 in build image

kalleep:
update our build image to use go v1.25.5 with various fixes and security fixes


kalleep:
> should we also update this [elsewhere in the code](https://github.com/search?q=repo%3Agrafana%2Falloy%20%22golang%3A1.25.1%22&type=code) so that the go versions are aligned?

We need to do that in another pr. First we need to update and build our build image

blewis12 (approved):

Issue #4729, opened 2025-10-30
Shared library for loki.source components.


kalleep:
### Add a description

Almost all Loki source files implement common patterns but we are still going back and fixing common issues around deadlocks and race conditions.

We should implement a shared library that abstracts away the lifecycle of sources/targets and implement safe common patterns for concurrency handling.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5026, opened 2025-12-08
fix: faster target scheduling for loki.source.docker

kalleep:
#### PR Description
`loki.source.docker` suffered from the same issue that `loki.source.file` did where scheduling could take some time for new targets.

In this pr I move `Scheduler` to source package so I can be reused by both components. With this I also moved "target" into docker package and rename to `tailer`, this tailer now implements `Source` interface.

I also fixed a issue where stopping component could deadlock if nothing was reading from handler chan.

Related to:  https://github.com/grafana/alloy/issues/4729

#### Notes to the Reviewer
* Moved target and metrics from internal package to `tailer.go` and `metrics.go` and they are no longer exported.

* Created a shared structure `Fanout`, this implements the common pattern to abort send operation if context is canceled.

* Create shared function `Consume`  to run the consume loop that will abort if context is canceled.

* Create shared function `Drain` that can be used when component is stopped. 
 
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


internal/component/loki/source/drain.go
```diff
@@ -0,0 +1,32 @@
...
+func Drain(recv loki.LogsReceiver, f func()) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-recv.Chan():
+				// Consume and discard entries to prevent channel blocking
```
thampiotr: I think we should check if the channel is closed, because we would be running a hot loop here?
```
			case _, ok := <-recv.Chan():
				if !ok { // channel closed
					return
				}
```

internal/component/common/loki/fanout.go
```diff
@@ -0,0 +1,63 @@
...
+// Send forwards a log entry to all registered receivers. It returns an error
+// if the context is cancelled while sending.
+func (f *Fanout) Send(ctx context.Context, entry Entry) error {
+	f.mut.RLock()
+	defer f.mut.RUnlock()
+	for _, recv := range f.children {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case recv.Chan() <- entry:
```
thampiotr, line 44: Is it possible that we will be stuck on this `recv.Chan() <- entry` because the receiver is too busy or dead, which would mean we are holding on to `f.mut.RLock()` and thus we are blocking all the other operations such as `UpdateChildren`?

I'm a bit unsure about correctness of this.

kalleep: This is what we where doing in `loki.source.file` and have not seen any problems there but we could test it out.

CHANGELOG.md
```diff
@@ -52,6 +52,10 @@ Main (unreleased)
 
 - Fix the `prometheus.operator.*` components internal scrape manager now having a way to enable ingesting native histograms. (@dehaansa)
 
+- Fix issue in `loki.source.docker` where scheduling containers to tail could take too long. (@kalleep)
```
thampiotr: Let's describe what would be the impact to the user.

internal/component/common/loki/fanout.go
```diff
@@ -0,0 +1,63 @@
...
+		f.mut.RUnlock()
+	}
+}
+
+func receiversChanged(prev, next []LogsReceiver) bool {
+	if len(prev) != len(next) {
+		return true
+	}
+	for i := range prev {
+		if !reflect.DeepEqual(prev[i], next[i]) {
```
dehaansa, line 69: Can the order change? If so, should that be considered different?

internal/component/loki/source/docker/docker.go
```diff
@@ -47,9 +48,11 @@ func init() {
 var userAgent = useragent.Get()
 
 const (
+	// See github.com/prometheus/prometheus/discovery/moby
```
dehaansa: This comment isn't very helpful

internal/component/loki/source/drain_test.go
```diff
@@ -0,0 +1,46 @@
...
+		time.Sleep(100 * time.Millisecond)
+		completed = true
+	})
+
+	wg.Wait()
+	require.True(t, completed, "Drain should complete without deadlock")
+
+	// Verify channel is empty.
+	select {
+	case <-recv.Chan():
```
dehaansa: Some concerns that this could be flaky. It _shouldn't_ be, but there's no guarantee that drain empties the channel from what I read, it just drains as much as possible before the passed func() finishes executing.

internal/component/common/loki/fanout.go
kalleep: I don't think so but would have to verify that :+1:

internal/component/loki/source/drain_test.go
kalleep: Yeah that is true, i guess it would be enough just to test that this does not hang because that is the real use of it.

We use this because some sources needs to drain what ever they are trying to send on the channel while they are stopping

internal/component/loki/source/docker/docker.go
kalleep: Sure, I just copied the comment that existed but I can remove it

internal/component/common/loki/fanout.go
kalleep: So from my testing the order won't change, only if it's actually changed in the config

CHANGELOG.md
```diff
@@ -77,6 +77,10 @@ Main (unreleased)
 
 - Fix default values for relabel rules, this caused issues in e.g. `prometheus.operator.servicemonitors` when using labeldrop. (@kalleep)
 
+- Fix issue in `loki.source.docker` where scheduling containers to tail could take too long, this would increase the chance of Alloy missing shortlived jobs. (@kalleep)
```
```suggestion
- Fix issue in `loki.source.docker` where scheduling containers to tail could take too long, this would increase the chance of Alloy missing short-lived jobs. (@kalleep)
```

internal/component/loki/source/scheduler.go
```diff
@@ -86,20 +86,33 @@ func (s *Scheduler[K]) Len() int {
...
 }
 
+// Reset will stop all running sources and wait for them to finish and reset
+// Scheduler to a usable state.
+func (s *Scheduler[K]) Reset() {
+	s.cancel()
+	s.running.Wait()
+	s.sources = make(map[K]scheduledSource[K])
+	s.ctx, s.cancel = context.WithCancel(context.Background())
+}
```
thampiotr, line 103: These two are quite similar... Can we reduce to have just one?

internal/component/loki/source/file/decompresser.go
```diff
@@ -339,8 +339,14 @@ func (d *decompressor) Key() positions.Entry {
 	return d.key
 }
 
-func (d *decompressor) IsRunning() bool {
-	return d.running.Load()
+func (d *decompressor) DebugInfo() any {
```
thampiotr, line 342: Another gotcha with DebugInfo functions in the past was that the lock order was not consistent and we had a deadlock. Happened at least 2x. I looked and I don't see this issue here, but the fact that we need to depend on manual review worries me here. Maybe we can refactor in the future to avoid the multiple locks sequencing issues.

kalleep: yes that is a issue for sure

internal/component/loki/source/scheduler.go
kalleep: Maybe I can think of a better way to do it. but we don't want to create a new context when we stop a component, that would leak resources..

kalleep:
>It looks correct, but I do feel like it's a bit of a risky change. I know you've done manual testing, so I'm approving this, but another pair of eyes will be a good idea.

Sure I can wait for @dehaansa do to a review too.

dehaansa (approved):

PR #5045, opened 2025-12-09
feat: support metadata in Remote Write V2

ptodev:
Adding metadata support to `prometheus.remote_write` component, but only if Remote Write v2 has been configured.

In order for `prometheus.remote_write` to receive metadata, `prometheus.scrape` must be configured with `honor_metadata = true`.

BEGIN_COMMIT_OVERRIDE
feat: add metadata support to `prometheus.remote_write` component, but only if Remote Write v2 has been configured. In order for `prometheus.remote_write` to receive metadata, `prometheus.scrape` must be configured with `honor_metadata = true`.
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/component/prometheus/remotewrite/remote_write_test.go
+++ b/internal/component/prometheus/remotewrite/remote_write_test.go
@@ -47,13 +47,15 @@ func TestSend(t *testing.T) {
 	sampleTimestamp := time.Now().Add(time.Hour).UnixMilli()
 
 	tests := []struct {
-		name      string
-		rwVersion RemoteWriteVersion
-		metrics   []Appendable
+		name            string
+		rwVersion       RemoteWriteVersion
+		metrics         []Appendable
+		expectedSamples int
 	}{
 		{
-			name:      "Remote write v1",
-			rwVersion: RemoteWriteVersionV1,
+			name:            "Remote write v1",
+			rwVersion:       RemoteWriteVersionV1,
+			expectedSamples: 2,
 			metrics: []Appendable{
 				&Sample{
 					Labels: labels.FromStrings("foo", "bar"),
@@ -68,8 +70,9 @@ func TestSend(t *testing.T) {
 			},
 		},
 		{
-			name:      "Remote write v2 with metadata",
-			rwVersion: RemoteWriteVersionV2,
+			name:            "Remote write v2 with metadata",
+			expectedSamples: 6,
+			rwVersion:       RemoteWriteVersionV2,
 			metrics: []Appendable{
 				// TODO: Add exemplars
 				&Sample{
@@ -169,6 +172,7 @@ func TestSend(t *testing.T) {
 			expectedMetricsBytes, err := os.ReadFile(filepath.Join(testdataDir, "expected_metrics.txt"))
 			require.NoError(t, err)
 			expectedMetrics := strings.ReplaceAll(string(expectedMetricsBytes), "__MIMIR_RW_URL__", srv.URL)
+			expectedMetrics = strings.ReplaceAll(expectedMetrics, "__EXPECTED_SAMPLES__", strconv.Itoa(test.expectedSamples))
 
 			cfgProtobufMessageAttr := ""
 			if test.rwVersion == RemoteWriteVersionV2 {
@@ -180,13 +184,22 @@ func TestSend(t *testing.T) {
 						cluster = "local",
 					}
 					endpoint {
-						name           = "test-url"
-						url            = "%s/api/v1/write"
-				        send_native_histograms = true
+						name                   = "test-url"
+						url                    = "%s/api/v1/write"
+						send_native_histograms = true
 						%s
+
+						queue_config {
+							// The WAL watcher should send the expected number of samples as soon as it gets them.
+							// That way all samples are sent in one batch and the RW request can be tested for all samples.
+							// Also, the test is not slowed down by the WAL watcher sending samples due to a timeout
+							// caused by waiting for an unnecessarily large number of samples.
+							max_samples_per_send = %d
+							batch_send_deadline = "1m"
+						}
 					}
 				`,
-				srv.URL, cfgProtobufMessageAttr)
+				srv.URL, cfgProtobufMessageAttr, test.expectedSamples)
 
 			// Create our component and wait for it to start running, so we can write
 			// metrics to the WAL.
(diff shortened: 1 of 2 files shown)


internal/component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -18,101 +30,396 @@ import (
...
+			require.NoError(t, tc.WaitRunning(5*time.Second))
+
+			sendMetrics(t, tc, test.metrics)
+
+			select {
+			case <-time.After(120 * time.Second):
+				require.FailNow(t, "timed out waiting for metrics")
+			case actual := <-writeResult:
+				require.JSONEq(t, expectedResponse, actual)
+			}
```
kgeckhart, line 226: This might end up with the same flakiness we fixed in, https://github.com/grafana/alloy/pull/4890/changes/58cce1d3228ae3423bc9b82033b1abec11b42cb5, since it depends on all the metrics to be appended before the default batch deadline. Might be fine but might be able to improve runtime/stability. 

I think this is true for the other usages here too.

internal/component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -18,101 +30,396 @@ import (
...
+					"prometheus_remote_write_wal_samples_appended_total",
+				)
+				require.NoError(c, err)
+			}, 10*time.Second, 100*time.Millisecond)
+		})
+	}
+}
+
+// The metadata should be sent again on every remote write HTTP request, even if it hasn't changed.
+func TestMetadataResend_V2(t *testing.T) {
```
kgeckhart, line 257: Could we follow the same strategy here with loading the responses from files? The text isn't as massive but the payloads are still a little large to have inline.

internal/component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -18,101 +30,396 @@ import (
...
+			writeResult := make(chan string)
+
+			// Create a remote_write server which forwards any received payloads to the
+			// writeResult channel.
+			responseStats := getResponseStats(test.metrics)
+			srv := newTestServer(t, writeResult, test.rwVersion, responseStats)
+			defer srv.Close()
+
+			// Load expected response and metrics from testdata
+			testdataDir := filepath.Join("testdata", test.name)
```
kgeckhart: Nit: could we do a spaces -> underscore translation on the test name? Folders with spaces tend to be a bit meh to deal with.

internal/static/metrics/wal/wal.go
```diff
@@ -91,6 +90,11 @@ func newStorageMetrics(r prometheus.Registerer) *storageMetrics {
 		Help: "Total number of exemplars appended to the WAL",
 	})
 
+	m.totalAppendedMetadata = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "prometheus_remote_write_wal_metadata_appended_total",
+		Help: "Total number of metadata entries appended to the WAL",
+	})
```
kgeckhart, line 96:
```suggestion
	m.totalAppendedMetadata = prometheus.NewCounter(prometheus.CounterOpts{
		Name: "prometheus_remote_write_wal_metadata_updates_total",
		Help: "Total number of metadata updates sent through the WAL",
	})
```

Since we aren't necessarily appending Metadata we only do it if it was updated.

internal/static/metrics/wal/wal_test.go
```diff
@@ -116,8 +118,120 @@ func TestStorage(t *testing.T) {
...
+prometheus_remote_write_wal_samples_appended_total 40
+`
+
+	err = testutil.GatherAndCompare(
+		reg,
+		strings.NewReader(expectedMetrics),
+		"prometheus_remote_write_wal_metadata_appended_total",
+		"prometheus_remote_write_wal_samples_appended_total",
+	)
+	require.NoError(t, err, "expected prometheus registry to be empty")
```
kgeckhart:
```suggestion
	require.NoError(t, err)
```

The assert error is a bit misleading and I don't think we need an extra message but depends on how detailed the err is.

internal/static/metrics/wal/series.go
```diff
@@ -20,6 +21,7 @@ type memSeries struct {
 
 	ref  chunks.HeadSeriesRef
 	lset labels.Labels
+	meta *metadata.Metadata
```
kgeckhart, line 26: One of the things that came up when discussing adding this to agent mode was memory overhead, https://github.com/prometheus/prometheus/issues/17619#issuecomment-3581661000. 

We should probably keep an eye on this when we enable it and think about using `unique` to intern the Type/Unit values.

internal/static/metrics/wal/wal_test.go
ptodev: I agree. This got autogenerated and I removed several instances of it, but missed tis one.

internal/static/metrics/wal/series.go
ptodev: Good point, I added a TODO

internal/component/prometheus/remotewrite/remote_write_test.go
ptodev: Looking at this now, I'm not really sure why it would fail. We call `Commit()` when we have finished forwarding all the samples we want to send. The tests write so few samples, that it'd be reasonable to expect the default config to send all of them at once rather than in tiny batches. I tried to reproduce the failure using the command you mentioned, and I even bumped up the counter to 200, but the tests still passed...

```
└─▪ go test ./internal/component/prometheus/remotewrite/... -run '^Test$' -count 200
ok      github.com/grafana/alloy/internal/component/prometheus/remotewrite      0.512s [no tests to run]
```

Am I missing something?

kgeckhart: I didn't test it with your changes and I did miss the adjustment to send all the metrics before commit. 

It is still susceptible to the same flaw as the commit per metric version though. The watcher is going to send a batch of decoded samples which [queue_manager is going to iterate over to enqueue in a shard](https://github.com/prometheus/prometheus/blob/0ac2221a205f06743d36d684843a568001ccbec7/storage/remote/queue_manager.go#L725-L782), [individually appending them to the current batch](https://github.com/prometheus/prometheus/blob/0ac2221a205f06743d36d684843a568001ccbec7/storage/remote/queue_manager.go#L1435-L1457), at any point while iterating the [shard timer can trigger causing the current batch to be sent](https://github.com/prometheus/prometheus/blob/0ac2221a205f06743d36d684843a568001ccbec7/storage/remote/queue_manager.go#L1640-L1647). 

Committing the whole batch will help but you're fighting a race condition with the shard timer that you could lose at any point in time. The original version didn't fail often, even with my multiple bad refactors I ran it 100 times and it didn't catch it.

ptodev: I changed the tests in the file to behave in a similar way to what we discussed in the other PR. Apart from reducing flakiness, this reduces the running time of the tests too. Previously they'd need the wait extra time until they hit a WAL timeout.

```
test=TestSend/Remote_write_v2_with_metadata ts=12:13:38.097 caller="ts=2025-12-15T12:13" msg=":38.097213Z level=info msg=\"Done replaying WAL\" subcomponent=rw remote_name=test-url url=http://127.0.0.1:53457/api/v1/write duration=502.081875ms"
test=TestSend/Remote_write_v2_with_metadata ts=12:13:53.098 caller="ts=2025-12-15T12:13" msg=":53.097595Z level=debug msg=\"Watcher is reading the WAL due to timeout, haven't received any write notifications recently\" subcomponent=rw remote_name=test-url url=http://127.0.0.1:53457/api/v1/write timeout=15s"
```

Issue #5039, opened 2025-12-09
Picking a prometheus WAL


kgeckhart:
### Component(s)

prometheus.remote_write

### Background

As of today there are two write-ahead-log (WAL) implementations for prometheus metrics in Alloy

[prometheus.remote_write](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/) (referred to as remote_write)
- Largely based on upstream prometheus concepts in the “[head_WAL](https://github.com/prometheus/prometheus/blob/main/tsdb/head_wal.go)”
- Perceived as inheriting concepts that are relevant for a database, IE caring about “active series”, which is an implementation tradeoff rather than a database concept that was inherited
- One of the most used components in Alloy

[prometheus.write.queue](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.write.queue/) (referred to as write.queue)
- Built to solve issues related to remote write such as, 
  - Reduce memory overhead as “active series” increase
  - Add ability to replay data on startup
- Custom implementation based on https://github.com/grafana/walqueue/ with no links to upstream prometheus
- Currently an experimental component with minimal usage

We need to decide which component we continue to invest in as the component of choice for remote write with Alloy.

### Proposal

This will be expanded upon in a design doc PR

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5048, opened 2025-12-09
proposal: Picking a Prometheus WAL

kgeckhart:
#### PR Description

Add proposal to decide our path forward with our two prometheus WAL implementations. 

Related to: https://github.com/grafana/alloy/issues/5039


docs/design/5039-picking-a-prometheus-WAL.md
```diff
@@ -0,0 +1,105 @@
...
+*  Having this data will allow us to write up pros and cons for choosing a “prometheus native” pipeline vs an “OTel native” pipeline with metrics
+*  It also will give us a lot of data we can use to benefit both ecosystems via improvements to the OTel collector WAL implementations and/or the prometheus remote write exporter WAL
+
+## Proposal 0: Do Nothing
+
+We continue to provide support for multiple WAL implementations increasing toil.
+
+This is not a viable options and accomplishes none of the goals.
+
+## Proposal 1: Improve remote_write to be the WAL we need
```
ptodev, line 52: It'd be good to make it more clear what the accepted solution is. One way to do it is if Proposal 1 becomes the entire "Proposal" section, and the rest could be listed in "Alternative Solutions" similarly to an [earlier doc](https://github.com/grafana/alloy/blob/main/docs/design/1161-standardize-stdlib.md).

kgeckhart: Yeah I guess I was more trying to directly translate our internal design doc process here. We do have two very valid proposals and based the discussion I would want to add a "conclusion" to have the PR merged. I can see the suggested format being useful when there's a single viable option but otherwise it feels a bit "leading" to present what is likely the authors choice as the "Proposal" and everything else as "Alternates".

jharvey10 (approved):

PR #5049, opened 2025-12-09
ci: hackathon release improvements

jharvey10:
Relates to #5078 

## TODO (pre-merge)

- [x] update all release docs
- [x] resolve zizmor vulnerabilities
- [x] update rulesets
- [x] change workflows to use vault
- [x] final pass through go code
- [x] modify changelog to be ready for release please
- [x] modify release (now publish-release-assets.sh) to upload assets to existing release, rather than create a release itself
- [x] remove make generate-versioned-files
- [x] check what it would look like to merge the release/v1.12 branch into main, and preemptively do it
- [x] change setting for squash to recommend PR title and commit message
- [x] lock PRs immediately after merge

## TODO (post-merge)

See https://github.com/grafana/alloy/issues/5079

### DIFF
--- a/tools/release/create-release-branch/main.go
+++ b/tools/release/create-release-branch/main.go
@@ -5,8 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"log"
-	"sort"
-	"strings"
 
 	gh "github.com/grafana/alloy/tools/release/internal/github"
 	"github.com/grafana/alloy/tools/release/internal/version"
@@ -15,10 +13,10 @@ import (
 const (
 	// backportLabelColor is the hex color for backport labels (without '#' prefix).
 	backportLabelColor = "63a504"
+	// releaseBranchPrefix is the prefix for release branches.
+	releaseBranchPrefix = "release/v"
 	// backportLabelPrefix is the prefix for backport labels.
 	backportLabelPrefix = "backport/v"
-	// maxBackportLabels is the maximum number of backport labels to keep.
-	maxBackportLabels = 3
 )
 
 func main() {
@@ -56,10 +54,10 @@ func main() {
 	}
 	fmt.Printf("Next minor version: %s\n", nextMinor)
 
-	branchName := fmt.Sprintf("release/v%s", nextMinor)
+	branchName := fmt.Sprintf("%s%s", releaseBranchPrefix, nextMinor)
 	fmt.Printf("Release branch: %s\n", branchName)
 
-	backportLabel := fmt.Sprintf("backport/v%s", nextMinor)
+	backportLabel := fmt.Sprintf("%s%s", backportLabelPrefix, nextMinor)
 	fmt.Printf("Backport label: %s\n", backportLabel)
 
 	// Check if branch already exists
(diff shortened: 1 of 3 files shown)


docs/developer/contributing.md
```diff
@@ -114,54 +114,6 @@ you can @-reply a reviewer asking for a review in the pull request or a
 comment, or you can ask for a review on the Slack channel
 [#alloy](https://slack.grafana.com).
 
-## Updating the changelog
```
dehaansa, line 117: We should document how naming goes into changelog here, or link to the doc if it's documented elsewhere.

docs/developer/shepherding-releases.md
```diff
@@ -0,0 +1,130 @@
...
+   Alloy version. This might mean updating the Go version, changing build tags, default config file
+   contents, etc.
+4. Merge the PR.
+
+### 8. Announce
+
+You did it! Now it's time to celebrate by announcing the Alloy release in the following Slack
+channels:
+
+[#alloy (internal slack)](https://grafanalabs.enterprise.slack.com/archives/CSN5HV0CQ)
```
dehaansa: Probably shouldn't include internal slack references here.

docs/developer/updating-otel/README.md
```diff
@@ -13,115 +20,149 @@ go.opentelemetry.io/otel/sdk
...
+- [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector)
+- [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib)
+- [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go)
 
 Unfortunately, updating Otel dependencies is not straightforward:
 
-* Some of the modules in `opentelemetry-collector` come from a [grafana/opentelemetry-collector](https://github.com/grafana/opentelemetry-collector) fork.
-  * This is mostly so that we can include metrics of Collector components with the metrics shown under Alloy's `/metrics` endpoint.
-* All Collector and Collector-Contrib dependencies should be updated at the same time, because they
+- Some of the modules in `opentelemetry-collector` come from a
```
dehaansa, line 29: Is this still true?

tools/release/backport/main.go
```diff
@@ -0,0 +1,184 @@
+package main
```
dehaansa, line 1: Is this based off of the shared github action we have for backport, or something hand rolled?

tools/release/create-release-branch/main.go
```diff
@@ -0,0 +1,174 @@
...
+	fmt.Printf("✅ Created label: %s\n", backportLabel)
+
+	// Clean up old backport labels (keep only the most recent maxBackportLabels)
+	if err := cleanupOldBackportLabels(ctx, client); err != nil {
+		log.Fatalf("Failed to clean up old backport labels: %v", err)
+	}
+}
+
+// cleanupOldBackportLabels removes backport labels older than n-2.
+func cleanupOldBackportLabels(ctx context.Context, client *gh.Client) error {
```
dehaansa: Will deleting these labels make it harder to track old backports? I'm not sure I see the value of removing them

docs/developer/shepherding-releases.md
jharvey10: There is precedence for this in several Grafana repos, but it's also easy enough for us to find, so I'll remove it.

docs/developer/updating-otel/README.md
jharvey10: We can definitely revisit this doc in a separate PR. Most of my changes were prettier formatting the file after I added the contents of what was in `docs/developer/release/00-ensure-otel-dep-updated.md` to it. (The stuff up above this section with some adjusted wording)

tools/release/backport/main.go
jharvey10: This is conceptually inspired by the shared backport GHA, but is much more simplistic and procedural (by design). Its purpose is to facilitate semantic commit-friendly cherry-picking in a very reliable way, and doesn't extend at all beyond that use case. The goal with this (and the other) automated workflow(s) is to keep the scope and flow as narrow and easy to follow as possible.

tools/release/create-release-branch/main.go
jharvey10: Deleting the labels is a subtle way to enforce that things shouldn't be backported further than n-2. It doesn't _prevent_ you from making the label yourself and using it, but helps establish that paradigm.

For each backport, we get a definitive reference in the original PR that looks like this:

image

And that can never be mutated as labels come and go.

There's two primary questions I had in mind when designing the workflow from the tracking perspective:

1. Given a PR, was it backported? And more specifically, what's its backport status? Those history entries on the issue provide that.
2. What (in total) has been backported to a given release branch? That's discoverable via the changelog (or the commit history) since the changelog corresponds 1-to-1 with feat/fix commits.

docs/developer/contributing.md
jharvey10: Good call. I don't think I mentioned semantic conventions anywhere yet other than the workflow which will point them out if they don't pass.

I expect most of the time that the maintainer who merges each PR will ensure the title is accurate, but the contributor should definitely have some guidance for getting all green checks on their PRs.

kalleep:
Just a suggestion, something we could revisit after this pr. But we could have one binary for all tools and use a command to specify what tool you want to invoke instead of having a binary for each tool

docs/developer/contributing.md
jharvey10: Done!

PR #5055, opened 2025-12-10
feat(loki.process): Support structured metadata as source type of stage.labels for loki.process

### LINKED ISSUE #4717: loki.process should support extracting data from strucutured metadata
### Component(s)

loki.process

### Request

When passing log streams to `loki.process`, it's currently not possible to use existing structured metadata. Being able to work with existing structured metadata is crucial for transform those log entries.

I think the simplest solution would be to offer adding all structured metadata into extracted values. From there, we can use extracted values to add labels and/or new structured metadata.

### Use case

We send log data to Alloy via `loki.source.api`. Received log entries contain a structured metadata field called `pod`. We need to transform it into a log label called `instance`.

### Tip

baurmatt:
#### PR Description

This enhances the `labels` stage of `loki.process` to be able to retrieve data from structured metadata.

Fixes #4717

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/converter/internal/promtailconvert/internal/build/limits_config.go
+++ b/internal/converter/internal/promtailconvert/internal/build/limits_config.go
@@ -44,7 +44,7 @@ func buildLimitsConfigStages(cfg limit.Config) []stages.StageConfig {
 					Rules: []*stages.RuleConfig{
 						{
 							Limit:      lineSizeBytes,
-							SourceType: stages.TruncateSourceLine,
+							SourceType: stages.SourceTypeLine,
 						},
 					},
 				},


berlincount:
I'd like to ping @grafana/grafana-alloy-maintainers for this ;)

ptodev:
Why not just use [stage.structured_metadata](https://grafana.com/docs/alloy/next/reference/components/loki/loki.process/#stagestructured_metadata) to place all the structured metadata into the extracted map, and then use `stage.labels` afterwards? I'm not sure if I'm missing something?

baurmatt:
@ptodev [stage.structured_metadata](https://grafana.com/docs/alloy/next/reference/components/loki/loki.process/#stagestructured_metadata) uses data from the extracted values map and creates labels from it. My usecase is, as described in https://github.com/grafana/alloy/issues/4717, that we receive log entries with `loki.source.api` which already has structured metadata attached, thus [stage.structured_metadata](https://grafana.com/docs/alloy/next/reference/components/loki/loki.process/#stagestructured_metadata) can't access it.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -604,6 +605,16 @@ stage.labels {
 }
 ```
 
+```alloy
+stage.labels {
```
dehaansa, line 609: Example should include the recommendation to strip the moved attributes with a stage.structured_metadata_drop

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Main (unreleased)
 
 - Reduced resource overhead of `prometheus.scrape`, `prometheus.relabel`, `prometheus.enrich`, and `prometheus.remote_write` by removing unnecessary usage of labelstore.LabelStore. (@kgeckhart)
 
+- Support structured metadata as source type of `stage.labels` for `loki.process`
```
dehaansa: Changelog is now generated from PR titles, read the updated contributing guide.

internal/component/loki/process/stages/labels.go
```diff
@@ -4,96 +4,150 @@ import (
...
+
+	LabelsSourceStructuredMetadata string = "structured_metadata"
+	LabelsSourceExtractedMap       string = "extracted"
 )
 
 // LabelsConfig is a set of labels to be extracted
 type LabelsConfig struct {
-	Values map[string]*string `alloy:"values,attr"`
+	Values     map[string]*string `alloy:"values,attr"`
+	SourceType string             `alloy:"source_type,attr,optional"`
```
kalleep: We have something similar for `stage.truncate`. I think we should use the same type, rename it to SourceType.

There is a issue with parsing the type that is solved by https://github.com/grafana/alloy/pull/5419.

The only issue is that `line` or `label` do not makes sense here. So ideally you could allow a subset of these. But we can leave that as a TODO and this pr should not be blocked by this.

docs/sources/reference/components/loki/loki.process.md
baurmatt: Added.

CHANGELOG.md
baurmatt: Removed, thanks! Sadly it seems I don't have permissions to edit the PR title 😭 Is this expected?

internal/component/loki/process/stages/labels.go
baurmatt: I've merge both types into one. Please have a look if I got your request right.

The labels stage already rejects line/label with this:

```go
	switch cfg.SourceType {
	case LabelsSourceExtractedMap, LabelsSourceStructuredMetadata:
	default:
		return fmt.Errorf(ErrInvalidSourceType, cfg.SourceType)
	}
```

baurmatt:
I've rebased and added changes as suggested. Sadly I'm not allowed to change the PR title, thought the Git commit title has been altered.

CHANGELOG.md
baurmatt: I was able to change it with the CLI... No idea why the webinterface doesn't provide a way for it.

baurmatt:
I was able to change the PR title with the gh CLI tool. No idea the webinterface doesn't provide a way for it.

Would appreciate another review round :)

kalleep:
It seems like you forgot to rename the type

baurmatt:
@kalleep Sorry, I forgot to commit the new types.go file -.- Fixed and rebased :)

kalleep:
Some issues in test when you rebased

baurmatt:
@kalleep Fixed! Tests as well.

internal/component/loki/process/stages/truncate.go
```diff
@@ -36,27 +36,25 @@ type TruncateConfig struct {
...
+	Sources    []string         `alloy:"sources,attr,optional"`
+	SourceType SourceType       `alloy:"source_type,attr,optional"`
 
 	effectiveLimit units.Base2Bytes
 }
 
 var (
-	_ encoding.TextMarshaler   = TruncateSourceType("")
-	_ encoding.TextUnmarshaler = (*TruncateSourceType)(nil)
+	_ encoding.TextMarshaler   = SourceType("")
```
kalleep: Not that you moved these to types.go all related methods and these assertions should also move

baurmatt: Right, sorry for not catching this! I've moved it over to types.go as well.

baurmatt: @kalleep Could we please get this merged? :)

clayton-cornell (approved):

internal/component/loki/process/stages/labels.go
```diff
@@ -4,92 +4,145 @@ import (
...
 	}
-	for labelName, labelSrc := range c {
+
+	switch cfg.SourceType {
+	case SourceTypeExtractedMap, SourceTypeStructuredMetadata:
+	default:
+		return fmt.Errorf(ErrInvalidSourceType, cfg.SourceType)
+	}
+
+	// We must not mutate the c.Values, create a copy with changes we need.
```
kalleep, line 46: We have this comment but then we still change c.Values to the new map, why was this change necessary?

internal/component/loki/process/stages/labels.go
```diff
@@ -4,92 +4,145 @@ import (
...
+	logger log.Logger
 }
 
-// Process implements Stage
-func (l *labelStage) Process(labels model.LabelSet, extracted map[string]any, _ *time.Time, _ *string) {
-	processLabelsConfigs(l.logger, extracted, l.labelsConfig, func(labelName model.LabelName, labelValue model.LabelValue) {
-		labels[labelName] = labelValue
-	})
+// Run implements Stage
+func (l *labelStage) Run(in chan Entry) chan Entry {
```
kalleep, line 89: I don't see why this change was necessary, i.e. manually implement Stage.

internal/component/loki/process/stages/labels.go
```diff
@@ -4,92 +4,145 @@ import (
...
+	out := make(chan Entry)
+	go func() {
+		defer close(out)
+		for e := range in {
+			switch l.cfg.SourceType {
+			case SourceTypeExtractedMap:
+				l.addLabelFromExtractedMap(e.Labels, e.Extracted)
+			case SourceTypeStructuredMetadata:
+				l.addLabelsFromStructuredMetadata(e.Labels, e.StructuredMetadata)
```
kalleep, line 98: We should be able to determine this during stage creation using function. So when you create this stage we should set function based on SourceType, this function would need to accept the same parameters but do different things based on what we configured.

internal/component/loki/process/stages/labels.go
baurmatt: From my understanding, this is needed to get access to the structured metadata as `Process` doesn't have them.

internal/component/loki/process/stages/labels.go
baurmatt: This was implemented to get access to `SourceType` (and it's default, set above) in `Run`.

internal/component/loki/process/stages/labels.go
baurmatt: My implementation was inspired by the implementation on the [`truncate`](https://github.com/grafana/alloy/blob/c47338fec6eb26fa9f2d040cc002f78721b93427/internal/component/loki/process/stages/truncate.go#L141) stage.

I'm not 100% sure how you want me to implement this. Could you please clarify?

internal/component/loki/process/stages/labels.go
kalleep: Sure but we should still not set new values for c.Values but return a map like we did before this change.

We can keep both in labelStage

internal/component/loki/process/stages/labels.go
kalleep: It's not too important. I can do a clean up later here and for truncate

internal/component/loki/process/stages/labels.go
baurmatt: Alright, will change it! Does this mean I'm not supposed to change `SourceType` as well within `validateLabelsConfig`? Asking because that's the reason why I changed it to the current implementation and the idea, again, was taken from the truncate stage. Sorry for bothering you with this!

kalleep: Ideally not. 
There are two interfaces that you can implement for `LabelsConfig`:

1. syntax.Validator - with this you can perform validation when config is evaluated. So when starting alloy we could catch these mistakes earlier.
2. syntax.Defaulter - with this you can set default values.

Ideally we use these but I have noticed that a lot of stage configs don't do it but I don't want to block this work further so lets just create the map like it did before and return it

baurmatt: Thanks for the clarification! :) I've reverted the changes from `validateLabelsConfig`.

baurmatt:
The failing tests don't seem to be related to my changes.

kalleep:
It's not, that was my fault. Just rebase on main and it should be good

baurmatt:
@kalleep Sure, done! :)

baurmatt:
Updated limits_config.go to match new centralized data type.

kalleep:
Hey you need to rebase in main again. We had some issues when building binaries for Mac that we just fixed. If you set "Allow edits by maintainers" I can make sure to get this pr merged so I don't have to bother you

PR #5061, opened 2025-12-10
chore: change the defaults of otelcol.exporter queue batch block

### LINKED ISSUE #4488: The default otelcol exporter queue batch sizes should work with Grafana Cloud
### Component(s)

_No response_

### Request

The default values for the batch processor [changed recently](https://github.com/grafana/alloy/pull/4363) in order to work better with Grafana Cloud. The batch processor is going to be deprecated soon in favour of a [batching config block](https://grafana.com/docs/alloy/next/reference/components/otelcol/otelcol.exporter.otlp/#batch) in the exporter components. We should make sure that this new `batch` config block also works well with Grafana Cloud by default.

Note that the first version of Alloy which will have the new `batch` block will be v1.11.0.

If we find that the default upstream values for the `batch` block don't work well with Grafana backends, then it would be useful to see if they need to be changed upstream? There is a possibility that they might be set to send unrealistically high batches like it was the case with the batch processor.

### Use case

Keeping up with the latest OTel best practices while making sure that Alloy users aren't getting errors or warnings sending to a Grafana backend.

### Tip

ptodev:
#### PR Description

Similar to #4363, but for the batch block in the exporters. This makes the defaults more friendly to Grafana Cloud. 

Fixes #4488

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
feat: Change the defaults for `sending_queue` > `batch` block inside `otelcol.exporter` components (#5061)

BREAKING CHANGE: `otelcol.exporter` > `sending_queue` > `batch` > `min_size` changed from `8192` to `2000` and `max_size` changed from `0` to `3000`
END_COMMIT_OVERRIDE

### DIFF
--- a/docs/sources/shared/reference/components/otelcol-queue-batch-block.md
+++ b/docs/sources/shared/reference/components/otelcol-queue-batch-block.md
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
 headless: true
 ---
 
+Batching is disabled by default.
+To enable it, explicitly include `batch {}` in your Alloy configuration.
+You do not need to include a `batch {}` block in your `otelcol.exporter` if you already use a `otelcol.processor.batch` component,
+although batching in the exporter is the prefered method because it is more flexible.
+
 The following arguments are supported:
 
 | Name            | Type        | Description                                                                                                | Default      | Required |
 | --------------- | ----------- | ---------------------------------------------------------------------------------------------------------- | ------------ | -------- |
 | `flush_timeout` | `duration`  | Time after which a batch will be sent regardless of its size. Must be a non-zero value.                    |  `"200ms"`   | no       |
-| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `8192`      | no       |
-| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |              | no       |
+| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `2000`      | no       |
+| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |  `3000`      | no       |
 | `sizer`         | `string`    | How the queue and batching is measured. Overrides the sizer set at the `sending_queue` level for batching. |  `"items"`   | no       |
 
 If configured, `max_size` must be greater than or equal to `min_size`.
--- a/internal/component/otelcol/config_queue.go
+++ b/internal/component/otelcol/config_queue.go
@@ -153,7 +153,19 @@ type BatchConfig struct {
 	Sizer        string        `alloy:"sizer,attr,optional"`
 }
 
+var _ syntax.Defaulter = (*BatchConfig)(nil)
+
 var defaultBatchConfig = otelexporterhelper.NewDefaultQueueConfig().Batch
+
+// SetToDefault implements syntax.Defaulter.
+func (args *BatchConfig) SetToDefault() {
+	*args = BatchConfig{
+		FlushTimeout: 200 * time.Millisecond,
+		MinSize:      2000,
+		MaxSize:      3000,
+		Sizer:        "items",
+	}
+}
 
 // Validate returns an error if args is invalid.
 func (args *BatchConfig) Validate() error {
--- a/internal/component/otelcol/exporter/loadbalancing/loadbalancing_test.go
+++ b/internal/component/otelcol/exporter/loadbalancing/loadbalancing_test.go
@@ -699,6 +699,184 @@ func TestConfigConversion(t *testing.T) {
 	}
 }
 
+func TestQueueBatchConfig(t *testing.T) {
+	tests := []struct {
+		testName string
+		alloyCfg string
+		expected otelcol.QueueArguments
+	}{
+		{
+			testName: "default",
+			alloyCfg: `
+			resolver {
+				static {
+					hostnames = ["endpoint-1"]
+				}
+			}
+			protocol {
+				otlp {
+					client {}
+				}
+			}
+			sending_queue {
+				batch {}
+			}
+			`,
+			expected: otelcol.QueueArguments{
+				Enabled:      true,
+				NumConsumers: 10,
+				QueueSize:    1000,
+				Sizer:        "requests",
+				Batch: &otelcol.BatchConfig{
+					FlushTimeout: 200 * time.Millisecond,
+					MinSize:      2000,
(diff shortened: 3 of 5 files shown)


docs/sources/reference/components/otelcol/otelcol.exporter.otlphttp.md
```diff
@@ -127,7 +127,7 @@ The `sending_queue` block configures queueing and batching for the exporter.
 {{< docs/shared lookup="reference/components/otelcol-queue-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
 ### `batch`
-
+<!-- TODO: How to mention that `batch {}` has to be specified in the config for the defaults to apply? -->
```
ptodev: @clayton-cornell I wonder what you think about this? It's because the block is a pointer and the default value is nil. I don't think we have a consistent way to document this kind of config.

kalleep: So by default batching is off? Is that because likely people have  bathing processor?

ptodev: I think so. It's [the same](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue-batch-settings) upstream:

> Batch settings are available in the sending queue. Batching is disabled, by default. To enable default batch settings, use batch: {}.

ptodev: @clayton-cornell I added a paragraph at the top of [otelcol-queue-batch-block.md](https://github.com/grafana/alloy/pull/5061/files#diff-3e93cd4016ead7d11785b8cdfee9725a597e050d5ff159698dec898bcecdf2ac). I think it's sufficient?

docs/sources/shared/reference/components/otelcol-queue-batch-block.md
```diff
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
 headless: true
 ---
 
+Batching is disabled by default.
+To enable it, explicitly include `batch {}` in your Alloy configuration.
+You do not need to include a `batch {}` block in your `otelcol.exporter` if you already use a `otelcol.processor.batch` component,
+although batching in the exporter is the prefered method because it is more flexible.
```
clayton-cornell:
```suggestion
By default, batching is disabled .
To enable it, explicitly include `batch {}` in your {{< param "PRODUCT_NAME" >}} configuration.
You don't need to include a `batch {}` block in your `otelcol.exporter` if you already use a `otelcol.processor.batch` component, although batching in the exporter is the preferred method because it's more flexible.
```

docs/sources/shared/reference/components/otelcol-queue-batch-block.md
```diff
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
...
+
 The following arguments are supported:
 
 | Name            | Type        | Description                                                                                                | Default      | Required |
 | --------------- | ----------- | ---------------------------------------------------------------------------------------------------------- | ------------ | -------- |
 | `flush_timeout` | `duration`  | Time after which a batch will be sent regardless of its size. Must be a non-zero value.                    |  `"200ms"`   | no       |
```
clayton-cornell:
```suggestion
| Name            | Type       | Description                                                                                                           | Default   | Required |
| --------------- | ---------- | --------------------------------------------------------------------------------------------------------------------- | --------- | -------- |
| `flush_timeout` | `duration` | Time after which {{< param "PRODUCT_NAME" >}} sends a batch regardless of its size. Must be a non-zero value.         | `"200ms"` | no       |
```

docs/sources/shared/reference/components/otelcol-queue-batch-block.md
```diff
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
...
-| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `8192`      | no       |
-| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |              | no       |
+| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `2000`      | no       |
+| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |  `3000`      | no       |
 | `sizer`         | `string`    | How the queue and batching is measured. Overrides the sizer set at the `sending_queue` level for batching. |  `"items"`   | no       |
```
clayton-cornell:
```suggestion
| `sizer`         | `string`   | The method used to measure the queue and batching. Overrides the sizer set at the `sending_queue` level for batching. | `"items"` | no       |
```

docs/sources/shared/reference/components/otelcol-queue-batch-block.md
```diff
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
...
+although batching in the exporter is the prefered method because it is more flexible.
+
 The following arguments are supported:
 
 | Name            | Type        | Description                                                                                                | Default      | Required |
 | --------------- | ----------- | ---------------------------------------------------------------------------------------------------------- | ------------ | -------- |
 | `flush_timeout` | `duration`  | Time after which a batch will be sent regardless of its size. Must be a non-zero value.                    |  `"200ms"`   | no       |
-| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `8192`      | no       |
-| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |              | no       |
+| `min_size`      | `number`    | The minimum size of a batch.                                                                               |  `2000`      | no       |
```
clayton-cornell:
```suggestion
| `min_size`      | `number`   | The minimum size of a batch.                                                                                          | `2000`    | no       |
```

docs/sources/shared/reference/components/otelcol-queue-batch-block.md
```diff
@@ -4,13 +4,18 @@ description: Shared content, otelcol queue batch block
...
+| `max_size`      | `number`    | The maximum size of a batch, enables batch splitting.                                                      |  `3000`      | no       |
```
clayton-cornell:
```suggestion
| `max_size`      | `number`   | The maximum size of a batch, enables batch splitting.                                                                 | `3000`    | no       |
```

Issue #5062, opened 2025-12-10
Make labelstore.LabelStore in to a SeriesRefMappingStore


kgeckhart:
### Component(s)

Indirectly impacts all prometheus components

### Background

labelstore.LabelStore is deeply integrated in to prometheus pipelines. Below is a sequence diagram of where it sits and its role in the pipelines,

```mermaid
sequenceDiagram
  prometheus.scrape->>fanout: append Ref 0
  fanout->>labelstore: GetOrAddGlobalRef
  labelstore->>fanout: GlobalRef 1
  loop for each remote_write component 0 ... n 
    fanout->>remote_write_n: append Ref 1
    remote_write_n->>labelstore: GetLocalRef for Ref 1
    labelstore->>remote_write_n: LocalRef or 0
    remote_write_n->>WAL: append
    WAL->>remote_write_n: LocalRef 1
    remote_write_n->>labelstore: Add/Update LocalLink (Ref 1 - LocalRef 1)
  end
  fanout->>prometheus.scrape: Ref 1
  Note over prometheus.scrape: Next scrape send same SeriesRef
  prometheus.scrape->>fanout: append Ref 1
```

This functionality exists because in alloy we can multi-cast appended samples to 0 or more remote_write components. Each remote_write component comes with a WAL that assumes it is responsible for allocating a globally unique SeriesRef as that is its responsibility in upstream prometheus. LabelStore ensures we have a SeriesRef that is re-usable for multiple calls to append.

The problem we have is that for each GlobalRef we add we are storing a copy of the labels in LabelStore. If the pipeline includes a `prometheus.relabel` component we have two GlobalRefs one before relabeling and one after with both label sets stored in memory. 

The drawbacks to this design are,
1. Each GlobalRef we add we are storing a copy of the labels in LabelStore further increasing memory usage
    * If the pipeline includes a `prometheus.relabel` component we have two GlobalRefs one before relabeling and one after with both label sets stored in memory
2. Knowing when a ref should be removed from the LabelStore is complicated and requires tracking the staleness of each appended sample (it's complicated enough that there's actually an indefinite retention bug when used with clustering)
3. We do all of this in many scenarios where it's not necessary like,
    * There are no remote_write components running in alloy
    * One pipeline has a remote_write component and another one does not (both still require LabelStore) 
    * A pipeline only contains a single remote_write component - we don't need LabelStore we can pass through the WAL SeriesRef instead

### Proposal

[...]

PR #5067, opened 2025-12-11
labelstore: Use rwlock

kgeckhart:
#### PR Description

This PR switches labelstore to use a RWMutex with a RLock for `labelstore.GetLocalLink`. I tried to add some concurrency test to show proof of impact but locally I can't seem to get enough concurrency or this doesn't help so I'll do some testing internally in draft. Full breakdown of rational in Notes to the Reviewer

Related to: 

#### Notes to the Reviewer

With very high append concurrency we have hit scrape timeouts which correlate with high mutex contention.

image (1)

The mutex contention is largely driven by `labelstore.GetLocalLink` which is called on every single append call. During the time period above that function was responsible for over 50% of our mutex wait,
image

Since it's a pure read operation switching to RWLock with RLock _should_ help. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/pipeline_test.go
+++ b/internal/component/prometheus/pipeline_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


kgeckhart:
Closing, internal tests showed that the reader `GetLocalRefID` appeared to starve the main writer `GetOrAddGlobalRefID` enough that there was no reduction in contention and in some cases increase. I'll focused on the improvements from https://github.com/grafana/alloy/issues/5062 instead since this isn't as quick of a win as expected.

PR #5080, opened 2025-12-12
feat: Add postgres logs error collector to database_observability.postgres component

gaantunes:
#### PR Description

Implementing a Postgres logs parser focused on error collection, with the following capabilities:

- Extracts insights from log messages, like affected tables, columns, blocker and blocked pids, among others
- Generates metrics from the error lines, categorizing errors by sqlstate_code, user, queryid, auth failures and more
- Accepts the JSON format only, given its the recommended format for parsing Postgres logs

#### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/postgres/component.go
```diff
@@ -138,7 +146,8 @@ func (a *Arguments) Validate() error {
 }
 
 type Exports struct {
-	Targets []discovery.Target `alloy:"targets,attr"`
+	Targets           []discovery.Target `alloy:"targets,attr"`
+	ErrorLogsReceiver loki.LogsReceiver  `alloy:"error_logs_receiver,attr,optional"`
```
cristiangreco, line 156: @gaantunes I've done a first pass, my feeling is that we're doing more than needed at this stage, I'll explain:
- I'd focus on producing one single metric, `postgres_errors_total`, but change the labels from `"severity", "database", "instance"` to `"database", "instance", "queryid"`. This seems to be the core of what we'd need to show an error rate (global and per-query).
- ignore anything else for now, e.g. no need to parse anything other than query error messages, no need for other metrics IMO. We also don't need to write back to EntryHandler a custom log line, we just need the metric really.

So my ask is to basically to trim it down dramatically and keep it very, very simple as first pass. We can still re-introduce later anything else that seems useful.

gaantunes: Hi @cristiangreco. 
Customers have showed interest in knowing what is the exact error in previous feedbacks. If we are already going to parse the logs to have metrics, why not provide this information already?

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -114,6 +119,28 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
 | `per_collect_ratio`            | `float64`      | The ratio of queries to collect explain plans for.   | `1.0`   | no       |
 | `explain_plan_exclude_schemas` | `list(string)` | Schemas to exclude from explain plans.               | `[]`    | no       |
 
+### `error_logs`
+
+The `error_logs` collector is always active and processes PostgreSQL error logs in JSON format (requires `log_destination = 'jsonlog'` in PostgreSQL configuration).
```
clayton-cornell, line 124:
```suggestion
The `error_logs` collector is always active and processes PostgreSQL error logs in JSON format.
This requires `log_destination = 'jsonlog'` in your PostgreSQL configuration.
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -114,6 +119,28 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+It receives log entries through the component's `forward_to` parameter, extracts error information, maps SQLSTATE codes to human-readable error names, and tracks metrics by error type and severity.
+
+Unlike other collectors, `error_logs` does not need to be enabled via `enable_collectors` - it automatically processes any PostgreSQL JSON logs that are forwarded to the component.
```
clayton-cornell, line 127:
```suggestion
Unlike other collectors, `error_logs` does not need to be enabled via `enable_collectors`.
It automatically processes any PostgreSQL JSON logs that are forwarded to the component.
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -114,6 +119,28 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+
+By default, SQL statements and Personally Identifiable Information (PII) in error log fields are redacted to protect sensitive data.
+This includes:
+- SQL literals in `statement` and `internal_query` fields
+- Data values in `detail` and `context` fields (such as constraint violations, key values, and row data)
```
clayton-cornell, line 132:
```suggestion
- Data values in `detail` and `context` fields such as constraint violations, key values, and row data
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -114,6 +119,28 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+- Parenthesized values in error messages (e.g., `Key (id)=(123)` becomes `Key (id)=(?)`)
```
clayton-cornell, line 133:
```suggestion
- Parenthesized values in error messages, for example, `Key (id)=(123)` becomes `Key (id)=(?)`
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -114,6 +119,28 @@ The `aws` block supplies the [ARN](https://docs.aws.amazon.com/IAM/latest/UserGu
...
+
+| Name                      | Type   | Description                                                                                         | Default | Required |
+|---------------------------|--------|-----------------------------------------------------------------------------------------------------|---------|----------|
+| `disable_query_redaction` | `bool` | Disable redaction of SQL queries and PII in error logs. Use with caution in production environments. | `false` | no       |
+
+{{< admonition type="warning" >}}
+Setting `disable_query_redaction` to `true` will expose SQL query parameters, constraint values, and other potentially sensitive information in error logs.
```
clayton-cornell, line 140:
```suggestion
If you set `disable_query_redaction` to `true`, it will expose SQL query parameters, constraint values, and other potentially sensitive information in error logs.
```

internal/component/database_observability/postgres/component.go
gaantunes: As discussed in our 1:1 this will be split into different PRs to approach the problem iteratively.

PR #5095, opened 2025-12-15
chore: add shared reconcile function for loki sources

kalleep:
#### PR Description
Create a shared function that can be used to reconcile scheduler for loki sources.

Part of: https://github.com/grafana/alloy/issues/4729

#### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/reconcile.go
+++ b/internal/component/loki/source/reconcile.go
@@ -54,7 +54,7 @@ func Reconcile[Key comparable, Input any](
 		source, err := sourceFactoryFn(key, i)
 		if err != nil {
 			if !errors.Is(err, ErrSkip) {
-				level.Error(logger).Log("msg", "failed to create source", "error", err)
+				level.Error(logger).Log("msg", "failed to create source, skipping", "error", err, "key", key)
 			}
 			delete(shouldRun, key)
 			continue
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -230,7 +230,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 			return newTailer(
 				c.metrics,
-				log.With(c.opts.Logger, "target", fmt.Sprintf("docker/%s", containerID)),
+				log.With(c.opts.Logger, "component", "tailer", "container", fmt.Sprintf("docker/%s", containerID)),
 				c.handler,
 				c.posFile,
 				containerID,


internal/component/loki/source/reconcile.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+// ErrSkip is used to indicate that a particular source should not be scheduled.
+var ErrSkip = errors.New("skip source")
+
+// SourceKeyFn extracts a comparable key of type T from an input value of type I.
+// The key is used to uniquely identify sources in the scheduler.
+type SourceKeyFn[T comparable, I any] func(I) T
```
thampiotr:
```suggestion
type ToKeyFn[Input any, Key comparable] func(Input) Key
```

I find it a bit easier to read this way...

internal/component/loki/source/reconcile.go
```diff
@@ -0,0 +1,80 @@
...
+
+// SourceFactoryFn creates a Source[T] from a key and input value.
+// It returns the created source (or nil if creation failed or should be skipped)
+// and an error. Return ErrSkip to indicate that the source should not be scheduled
+// without logging an error.
+type SourceFactoryFn[T comparable, I any] func(T, I) (Source[T], error)
```
thampiotr:
```suggestion
type SourceFactoryFn[Key comparable, Input any] func(Key, Input) (Source[Key], error)
```

Also a bit more specific to our use case, but still it's generic-enough.

internal/component/loki/source/reconcile.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Reconcile synchronizes the scheduler's set of running sources with a desired state.
+// It iterates over inputs, creates sources for new items, and stops sources that are
+// no longer needed.
+func Reconcile[T comparable, I any](
```
thampiotr:
```suggestion
func Reconcile[Key comparable, Input any](
```

internal/component/loki/source/reconcile.go
```diff
@@ -0,0 +1,80 @@
...
+
+		// Skip if a source with this key is already running.
+		if s.Contains(key) {
+			continue
+		}
+
+		source, err := sourceFactoryFn(key, i)
+		if err != nil {
+			if !errors.Is(err, ErrSkip) {
+				level.Error(logger).Log("msg", "failed to create source", "error", err)
```
thampiotr: We should write here that this source will not be used for logs. And also print the key in one way or another so it's possible to identify what is being missed.

internal/component/loki/source/reconcile.go
kalleep: Ah yeah this is much better!

internal/component/loki/source/reconcile.go
kalleep: Great suggestion, key should be fine to use. For docker it's container id (if one exists) and for file it's public labels + path. Need to figure out a nice way to format these

kalleep: Updated in [f695ed8](https://github.com/grafana/alloy/pull/5095/commits/f695ed895cb92aa9fb9bcb1aed4bf567569037d0).

WDYT? 

For loki.source.docker key will be container id but for loki.source.file key is path + public labels so it will look something like this:

```
ts=2025-12-15T15:11:42.032910221Z level=error msg="failed to create source, skipping" component_path=/ component_id=loki.source.file.file error="failed to tail file, is directory" key="{Path: Path Labels:{}}"
```
And that is not great :(

thampiotr: I think it's better to make sure we have all the info we need to investigate than it is to have pretty logs. But it would be a problem if this line is logged frequently. I don't assume we expect it to be logged frequently?

kalleep: No only when something have gone wrong. If it's a intentional skip then this is not logged

PR #5107, opened 2025-12-15
ci: add main branch running tally for next release

jharvey10:
release-please will now create a PR tracking the main branch (which should not ever be merged). It will provide a running list of all changelog entries that will show up for the next minor release.

This also fixes a bug on release branches where during RC cuts, it is possible to build an RC based on an old commit.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -62,7 +62,7 @@ jobs:
             echo 'When ready to cut a release:'
             echo '1. Create a release branch (e.g., `release/v1.13`)'
             echo '2. The release-please automation will take over on that branch'
-            echo '3. This PR will remain on main to track changes'
+            echo '3. This PR will remain on main to track the next release'
             echo ''
             echo '---'
             echo "EOF"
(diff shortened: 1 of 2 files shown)


release-please-config-main.json
```diff
@@ -0,0 +1,34 @@
+{
```
kalleep: is this a copy of the other file? Why do we need two?

jharvey10: It is largely a copy of the other file, yeah. Really the only reason for this is to change the text of the resulting PR against the main branch so it says "DO NOT MERGE". Mostly just to avoid someone accidentally working with the wrong release-please PR. But I think I might be able to expose more of the command line arguments to pass in these two additional config options we need in the workflow to avoid having a separate config file. I'll give that a shot and see if I can eliminate this separate file.

jharvey10: Wasn't as straightforward as I'd hoped to implement it; the args for pr title and header don't get passed the same way as other config options in release-please, but I was able to modify the PR after release-please creates it to update the title and body.

PR #5114, opened 2025-12-16
feat: Add unexposed otel engine and extension to codebase and change build structure

blewis12:
### What this PR Introduces
This PR introduces the Alloy Collector Distro, which has it's components defined in an [OCB](https://opentelemetry.io/docs/collector/extend/ocb/) file and generated using the [builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) tool . The beefy change here is a re-jig of the build graph of the project, which is why I would like to introduce this as a single PR to ensure that folks are on board with it. The `otel` subcommand itself is _not_ exposed yet, this will be introduced in a follow up PR which includes documentation changes. 

Please bare in mind a fair bit of code here is generated. I will try to provide an overview in the following sections to make it easier to break down

**OCB**
The components to include in our otel distribution are all defined in our `builder-manifest.yaml` OCB file. The file itself references the extensions/exporters/processors and so forth that we want to include, it also defines `replace` directives that we want to include in our final build. This is what will generate the main `go.mod` file in the `collector` directory that will be the build root for the project. Note that this is important to keep in sync with the replace directives we have in the `alloy` module, for which we use the new `dependency-replacements.yaml` file and generation targets ([relevant design doc](https://docs.google.com/document/d/1akSUpMksgYoazjEs4cD38YlyTzx-5AcRX0lVIH9MYd4/edit?tab=t.rox6mgp16j3c#heading=h.juzrvrygbq7v))

**General file structure** ([relevant design doc](https://docs.google.com/document/d/18B1mtoF1rOqzD262yLGnznPawrcodTu_yc6OJTlLEHU/edit?tab=t.0))
```
alloy/
│
├── collector/                     # OTel Collector build configuration
│   ├── builder-config.yaml        # OCB (OpenTelemetry Collector Builder) configuration
│   ├── generate.go                # Entry point for code generation
│   ├── generator/                 # Templating and custom generation code
│   │
│   └── [Generated Files]          # Everything else is generated by OCB + generator, including go.mod
│
├── extension/                     # OTel Collector extension
│   └── alloyengine/               # Alloy engine extension for OTel Collector
│       ├── extension.go           # Main logic for extension
│       ├── factory.go             # Factory to hook extension into collector
│
└── flowcmd/                       # command files moved from root to this directory so they can be imported by collector 
```

**Changed build graph**
Untitled - Flow

**Building locally / Makefile changes**
When building locally via `make alloy`, the entire chain of replacement and collector generation/build will run - I found this to be the smoothest experience when developing locally so I didn't have to think about manually running commands before building in my changes. The time taken to do the generation is also relatively low. However in CI, the generation currently does not run. My reasoning for this is that we are checking in the built/generated code - what is built in CI should be a reflection of the files we are checking in. If we did generation as part of CI, then the user might check in files would not work when building locally but pass in CI or vice versa

**Versioning**
The version of the manifest is currently being kept in sync with the alloy version (currently v1.12.1). IMO we should keep the versioning synced, as `otel` is a subcommand of the alloy CLI and so our distro is embedded within alloy from a user-perspective. I feel it'd make more sense for users to see a single version when they run `alloy --version` and `alloy otel --version`

**Testing**
Very minimal, since command has not been exposed yet - once this is merged, integration tests and lifecycle tests will be added 

**Change in Binary Size**
before:
```
-rwxr-xr-x@ 1 bejallewis staff 463M Dec 18 12:50 ./build/alloy 
```

after:
```
-rwxr-xr-x@ 1 bejallewis staff 464M Dec 18 12:49 ./build/alloy
```

So this increases the binary size by ~1Mb, due to the collector code being bundled in

relates to https://github.com/grafana/alloy/issues/4719

**Edit**
I've also left out versioning for the alloyextension here - from what I understand, this can be done with release-please - but I feel like this can be done in a follow up PR if we want to actually tag/release this extension for use in external OCB manifests


collector/generator/generator.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func main() {
+	log.Println("Generating Alloy OTel Collector main file...")
+	dir, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("failed to get working directory: %v", err)
+	}
+	log.Printf("working dir: %v", dir)
+	otelGeneratedMain := os.Args[2]
```
thampiotr: Nit: we can use built-in `flag` package and get a bit easier to read CLI.

collector/generator/generator.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if !replaced {
+		return fmt.Errorf("target line not found in %s", filePath)
+	}
+
+	newContent := strings.Join(lines, "\n")
+
+	fi, err := os.Stat(filePath)
+	var mode os.FileMode = 0o644
```
thampiotr, line 84: TIL about octal literals

collector/builder-config.yaml
```diff
@@ -0,0 +1,96 @@
...
+exporters:
+  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.139.0
+  - gomod: go.opentelemetry.io/collector/exporter/nopexporter v0.139.0
+  - gomod: go.opentelemetry.io/collector/exporter/otlpexporter v0.139.0
+  - gomod: go.opentelemetry.io/collector/exporter/otlphttpexporter v0.139.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.139.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter v0.139.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter v0.139.0
+    # TODO: figure out dependency issues
+    #  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter v0.139.0
```
thampiotr: there is some weirdness here too, because in Alloy default engine the `otelcol.exporter.prometheus` is not the same as the upstream one - it converts metrics in-memory to prometheus format.

collector/generator/generator.go
```diff
@@ -0,0 +1,87 @@
...
+			replaced = true
+			break
+		}
+	}
+
+	if !replaced {
+		return fmt.Errorf("target line not found in %s", filePath)
+	}
+
+	newContent := strings.Join(lines, "\n")
```
thampiotr, line 81: We could add the 'this code is generated' warning here and thus we won't have to explain the difference between .go and .tpl.

extension/alloyengine/config.go
```diff
@@ -0,0 +1,34 @@
+package alloyengine
+
+import (
+	"fmt"
+	"os"
+)
+
+// Config represents the configuration for the alloyflow extension.
+type Config struct {
+	ConfigPath string            `mapstructure:"config_path"`
```
thampiotr: Maybe we call it `config_file`, as opposed to the future `config` which would be in-line YAML.... or perhaps a raw multiline string. Just thinking how to avoid future renames.

extension/alloyengine/extension.go
```diff
@@ -0,0 +1,163 @@
...
+		previousState := e.state
+		e.state = stateTerminated
+		e.stateMutex.Unlock()
+
+		if err == nil {
+			e.settings.Logger.Debug("run command exited successfully without error")
+		} else if previousState == stateShuttingDown {
+			e.settings.Logger.Warn("run command exited with an error during shutdown", zap.Error(err))
+		} else {
+			e.settings.Logger.Error("run command exited unexpectedly with an error", zap.Error(err))
```
thampiotr: I wonder if we should keep restarting it with some backoff, or would it restart itself in this case? should be easy to test with invalid alloy default engine config file

extension/alloyengine/extension_test.go
```diff
@@ -0,0 +1,132 @@
...
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	t.Cleanup(cancel)
+	require.NoError(t, e.Shutdown(shutdownCtx))
+
+	// Verify the run goroutine has exited and state is terminated.
+	select {
+	case <-e.runExited:
+		// ok
+	case <-time.After(2 * time.Second):
+		t.Fatal("run command did not exit in time")
```
thampiotr: Can we use the Eventually pattern? or maybe we should tolerate some small delay here? in CI, we can have whole container pause for a few seconds if we get unlucky

extension/alloyengine/extension_test.go
```diff
@@ -0,0 +1,132 @@
...
+	expected := errors.New("boom")
+	e := newTestExtension(t, func() *cobra.Command { return errorCommand(expected) })
+
+	require.NoError(t, e.Start(context.Background(), componenttest.NewNopHost()))
+
+	// Wait for the command to exit and extension to transition to terminated.
+	select {
+	case <-e.runExited:
+		// ok
+	case <-time.After(2 * time.Second):
```
thampiotr: similar to above ^

extension/alloyengine/extension_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	t.Cleanup(cancel)
+	require.NoError(t, e.Shutdown(shutdownCtx))
+
+	// The internal goroutine should have transitioned to terminated even on error during shutdown.
+	select {
+	case <-e.runExited:
+		// ok
+	case <-time.After(2 * time.Second):
```
thampiotr: here too :)

extension/alloyengine/config.go
thampiotr, line 1: Should we add a README.md for the extension? There is probably some OTel standard format we could use.

flowcmd/fallback_version.go
```diff
@@ -1,20 +1,23 @@
...
-//go:embed .release-please-manifest.json
-var fallbackVersionJSON []byte
-
 // fallbackVersion returns a version string to use for when the version isn't
 // explicitly set at build time. The version string will always have -devel
 // appended to it.
 func fallbackVersion() string {
-	return fallbackVersionFromJSON(fallbackVersionJSON)
+	manifestPath := "../.release-please-manifest.json"
+	data, err := os.ReadFile(manifestPath)
```
thampiotr: oh, I'm not sure this will work anywhere outside of Alloy repo? 
I thought that the `go:embed` trick was setting the `var` at compile time, while the `fallbackVersion()` now is getting to the file at runtime

Dockerfile
```diff
@@ -20,6 +20,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG TARGETVARIANT
 ARG RELEASE_BUILD=1
+ARG GENERATE_COLLECTOR_DISTRO=0
```
thampiotr: Not sure I follow why do we need this?
Could we avoid it somehow?

thampiotr (commented):
Looking good! 
One thing I also wonder about: does the PR dependency review bot need some adjustments to its prompt to look into the right go.mod files?

Dockerfile
blewis12: I added this because I don't want to run any code generation when we build from containers when we run `make alloy` - we could remove [this from the makefile](https://github.com/grafana/alloy/pull/5114/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R199-R208) and not have `generate-otel-collector-distro` be a pre-req for building alloy locally _but_ I found it to be quite tedious when developing that way 🤔 maybe there's another way to achieve the same thing though, I can have a think

extension/alloyengine/config.go
blewis12: Good point 👍 I'll add that

flowcmd/fallback_version.go
blewis12: agh - yes you're right. Annoyingly embed cannot access files outside the directory. Previously I had `.release-please-manifest` in the `flowcmd` directory, which we can do - but it's a bit weird since the other files are at the directory root (FYI @jharvey10)

collector/generator/generator.go
blewis12: taken from [here](https://github.com/grafana/alloy/compare/main...thampiotr:alloy:thampiotr/otel-extension-poc-simpler#diff-106a7388fad5b6c4fcc998be7708a7006bee6089ef2229982ce87e99531fcd86R78) hehe, I like it it's more readable

collector/generator/generator.go
blewis12: hmm not sure I understand you - `main_alloy.tpl` already has a "code is generated" comment, unless you're talking about something else?

collector/generator/generator.go
blewis12: https://github.com/grafana/alloy/pull/5114/commits/3f4fd87939a15ea9680b48e8f8c9a12d36263ac8

extension/alloyengine/config.go
blewis12: hmm yeah good point - I think `config_file` is a ambiguous enough to include both a path or online. _or_ we can include these as two different options. Thinking how we do this in our [Alloy helm chart as inspo](https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/values.yaml#L32-L41) here with regards to config maps. We could do something like

```
config:
    inline: boolean
    # for when inline: true
    value: string|null
    # for when inline: false
    config_file: string|null
```

for now we can just keep `config_file` but have the flexibility to add the other fields later. wdyt?

collector/generator/generator.go
thampiotr: You could remove the warning from `main_alloy.tpl` and add it here by prepending it to `newContent` - this way the template does not need to contain the explanation that 'do not edit' is relevant only if you're looking at the .go file and not when you're looking at the .tpl file.

extension/alloyengine/config.go
thampiotr: Hmm, I think in some other components we have two settings that are mutually exclusive: `config` and `config_file`. Like here: https://grafana.com/docs/alloy/next/reference/components/prometheus/prometheus.exporter.snmp/

In OTEL collector there is also native mechanism to include a file, like this `exporters: ${file:exporters.yaml}`.

Not sure I like what helm chart does. The `inline` as a bool has only two states. What if we need another state one day?

I think there are really two options: 
```
# option A
config:
  format: (file|yaml|alloy)
  value: ...

# option B - multiple fields, mutually exclusive
config_file: foo.alloy
config_alloy: <str of alloy syntax>
config: <new yaml representation of alloy syntax>
```

Maybe we go with option A then? 🤔

Dockerfile
thampiotr: Maybe we have two Makefile targets - one for CI and one for developers. `make alloy-ci` and `make alloy`? I am not a big fan of build steps that act differently in different environments automatically, might end up hard to debug one day.

collector/generator/generator.go
blewis12: oh I see what you're saying now - yes that makes total sense 👍

Dockerfile
blewis12: true - adding a separate target reduces context-confusion, i'll look into implementing it this way

extension/alloyengine/extension.go
blewis12: In the case of invalid config then we don't retry starting the alloy engine but we certainly could. At first I thought maybe we could skip this since Alloy doesn't currently support that (it just shuts down), but now we might have a case where the OTel Engine is running just fine but the underlying Default Engine might have crashed due to bad config. 

At the very least restarting every X interval with backoff would make the failure more visible in the logs, and means the user doesn't have to also restart the OTel Engine after they've fixed the alloy config

extension/alloyengine/config.go
blewis12: Option A sounds like a good shout to me 👍

extension/alloyengine/extension_test.go
blewis12: resolving the other two comments so it's addressed in one place: https://github.com/grafana/alloy/pull/5114/commits/5a9a77381d16c3b4b6e56368838e3422c9ea6ca0

I reduced the timeout to 1s since that seems reasonable, we can adjust if we encounter issues I'd reckon. Let me know if I misunderstood your comment though and missed something

collector/generator/generator.go
blewis12: https://github.com/grafana/alloy/pull/5114/commits/3f78bdf2ae1144702dfabdf31a498dcabfd28ba6

extension/alloyengine/config.go
blewis12: https://github.com/grafana/alloy/pull/5114/commits/280b5d2013e602f673626314990913ec7e1f2264

jharvey10:
> The version of the manifest is currently being kept in sync with the alloy version (currently v1.12.1). This is manual for now but can be automated. 

Check out these files for how to automatically get this version kept up to date by release-please :)
`docs/sources/_index.md`
`release-please-config.json`

jharvey10:
> One thing I also wonder about: does the PR dependency review bot need some adjustments to its prompt to look into the right go.mod files?

@thampiotr It shouldn't! It's currently configured to receive the git diff as input, which it is told could contain go mod files, and it's expected to carry out investigation for each of them.

flowcmd/fallback_version.go
jharvey10: I'm still trying to get acclimated with the top level logic of how we enter into the program and begin execution, but I see that the collector folder is all generated, which I presume makes this quite a bit more challenging.

This fallbackVersion thing _used to be_ based on the `VERSION` file (which I swapped out for the release-please manifest).

There's no rule that we _can't_ move the release-please files someplace else, but their intent is not to be "owned" by any other part of the code. I'm also a bit hesitant to put them into a folder which is predominantly _generated_ files, where these are definitely not.

This fallback version file seems conceptually similar to the generated `collector/main.go` file in that it embeds version information somehow into the file. It just happens to output a complete source file, whereas this fallbackversion stuff stores the data in a byte array for later retrieval.

Maybe we have some optionality to align this more with how otel works.

Dockerfile
jharvey10: Was about to chime in with a related comment and then found this thread :)

In a general sense, I'm not a big fan of checking in any file to scm which has a deterministic output given the same inputs.

Is there a technical reason why we need to have so many of these files checked into git instead of just generating them at the moment at which they're needed?

> I added this because I don't want to run any code generation when we build from containers when we run `make alloy`

I'm curious why 🤔. If this is how we have to do it when running locally, wouldn't we want to do the exact same thing when the container/CI builds the official binary? If the intent is to rely on the already-existing-generated files and not wanting to clobber them by accident, then that hints at perhaps a more severe issue with non-determinstic build outputs.

As a counter example, here's some pieces from the graphql/repl stuff I'd worked on a while ago, which needs to generate intermediate programmatic assets prior to `make alloy`. Most of them _aren't_ checked in and therefore aren't present in this PR.

[Makefile](https://github.com/grafana/alloy/pull/4012/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)

[alloy.resolvers.go](https://github.com/grafana/alloy/pull/4012/changes#diff-816be806353c5c056fbe9371d7187f06123cc2cd9d44c23621219461928cf709)

^^ Interestingly, this file presents as checking in a generated file, but in this case, only the stub of this file is generated, and you're expected to fill in the blanks with custom logic (this is defined by nature of the gqlgen.yml file)

As a pre-req for _development_ against the gql stuff, you have to run `make generate` (or similar) so that you can use `go run` or `go test` or edit files without errors everywhere, but even if you've never run that, you can still run `make alloy` and get a 100% complete build including the generated assets without them being checked in, regardless of if you're running locally.

thampiotr:
Should we add a featuregate check that will ensure this can only be used when stability level is set to experimental?

extension/alloyengine/extension.go
blewis12: I've implemented this here: https://github.com/grafana/alloy/pull/5114/commits/3f351417360b639ec2496440262ff3257d7e68cd

I had to tweak some logic in `internal/alloycli/cmd_run.go` to get that to not panic, but I don't that should change much when it comes to how alloy itself runs. I decided to opt for infinite retries until the context is cancelled. In my opinion it's more visible in the logs, and means that the user can update said config without a time limit to have the alloy instance run properly. Open to other takes there though

extension/alloyengine/config.go
blewis12: done here: https://github.com/grafana/alloy/pull/5114/commits/c28df018b537c83e8daa5fa9c3006a166ee70ee3

There isn't a default format to use, since those in inside otel-collector and contrib are partially generated. Looking at external extension readme's, the format appears to be more freeform

flowcmd/fallback_version.go
blewis12: So i've done two things in relation to this:

1. Moved the release-please file back to the root directory: https://github.com/grafana/alloy/pull/5114/commits/5c75653e3087a2d2c59fbab7bf18cdd1958b724c which means we can use `embed` as before. I realized that the main file we really need to be inside the `flowcmd` directory is `flowcmd.go` which _imports_ the fallback logic. The fallback logic can remain under the root directory, which i've named as the `alloy` package to correspond with the directory structure
2. Attempted to sync the version of the collector with alloy here: https://github.com/grafana/alloy/pull/5114/commits/131a6e2e16124735cd92eb9b7c079f90f301ae50. @jharvey10 perhaps you can check my logic there? This is how I assume this is done, but I could be missing something

Dockerfile
blewis12: That's a great point. My reasoning behind keeping these generated files was:

* The generated files _are_ the `main` entry point into the program, I feel it could be conceptually confusing to be able to check out code that has no obvious main entry point
* Developers would be required to have the `builder` tool installed to generate the code when cloning the repo, as opposed to being able to compile the generated code as it stands
* All our docker containers (and homebrew formula?) will also need to include the ocb builder tool
* The build output should be deterministic with the `builder` tool, I don't believe this should be an issue.

However in general, looking online at least, it seems like it's generally advised to _not_ check in generated files for Go projects as you say. I'd want to think on this a bit further, maybe we can discuss/rubber duck in a call?

blewis12:
> Should we add a featuregate check that will ensure this can only be used when stability level is set to experimental?

The only issue there I can see is that our current stability.level flag is assigned to the `run` command - so we don't have access to that in `otel`. We could add an additional stability.level flag for the `otel` command to mirror `run` though - I think from a user perspective that should make enough sense: https://github.com/grafana/alloy/pull/5114/commits/70797f6ea5e81351ae796325d0d8b731c34962e6

collector/builder-config.yaml
blewis12: I'm attempting to solve this by pointing to my own [fork](https://github.com/blewis12/opentelemetry-collector-contrib/tree/patch-exporter-prometheusremotewrite-v0.139.0-to-prometheus-v0.308.0) which points to v0.139.0 of the `prometheusremotewriteexporter` and updates it's prometheus dependency to match alloy (v0.308.0). From what I can see, the errors from doing were mostly from moving types around. Things compile now, but I need to look at it a bit more to make sure i'm not missing something

blewis12: > there is some weirdness here too, because in Alloy default engine the `otelcol.exporter.prometheus` is not the same as the upstream one

hmm i'm not sure I understand, aren't `otelcol.exporter.prometheus` and the upstream `prometheusremotewriteexporter` different in terms of responsibility? 🤔 

`otelcol.exporter.prometheus` converts the metrics to the prometheus format in-memory and then forwards to the remote write component. Whereas `prometheusremotewriteexporter` does this all in one component (conversion and write). I suppose I don't have the historical context for why it was done that way, but I imagine that would be why we don't use the upstream component?

thampiotr:
> The only issue there I can see is that our current stability.level flag is assigned to the `run` command - so we don't have access to that in `otel`.

Ah, that's right! I don't think in this case it makes sense to add it. We should just make sure we document in the CLI that this is experimental OTel engine.

collector/main_alloy.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/spf13/cobra"
+	"go.opentelemetry.io/collector/otelcol"
+)
+
+func newAlloyCommand(params otelcol.CollectorSettings) *cobra.Command {
+	otelCmd := otelcol.NewCommand(params)
+	// Modify the command to fit better in Alloy
+	otelCmd.Use = "otel"
+	otelCmd.Short = "Alloy OTel Collector runtime mode"
+	otelCmd.Long = "Use Alloy with OpenTelemetry Collector runtime"
```
thampiotr: This is the part I was thinking about:
- We should add here that this is experimental. I don't personally think it makes sense to add an otel-incompatible CLI flag for stability level. Sorry for missing this before I wrote my earlier suggestion. 
- We are using the word 'runtime mode' here. This is user facing and we want to use "OTel engine".

collector/main_alloy.go
```diff
@@ -0,0 +1,42 @@
...
+				minStability.String(),
+			)
+		}
+
+		return nil
+	}
+
+	flowCmd := flowcmd.RootCommand()
+	// TODO: Officially add this command to expose in a follow up PR
+	// flowCmd.AddCommand(otelCmd)
```
thampiotr: We can add it and set to hidden. I wouldn't mind that, can make testing easier.

collector/main_alloy.go
blewis12: ah! a good reminder to rescan and check for outdated semantics, I didn't see this. I'll make those adjustments

collector/main_alloy.go
blewis12: ah cool I didn't know you could do that, just pushed a commit now

collector/main_alloy.go
blewis12: just pushed these changes now 👍

blewis12:
sorry to the ping @grafana/db-o11y-squad - had some rebase weirdness which brought in changes that automatically assigned you for review

collector/main_alloy.go
```diff
@@ -0,0 +1,23 @@
...
+
+	otelCmd.Use = "otel"
+	otelCmd.Short = "Use Alloy with OTel Engine"
+	otelCmd.Long = "[EXPERIMENTAL] Use Alloy with OpenTelemetry Collector Engine"
+
+	flowCmd := flowcmd.RootCommand()
+	flowCmd.AddCommand(otelCmd)
+
+	// TODO: Officially add this command to expose in a follow up PR
+	otelCmd.Hidden = true
```
thampiotr: Will this work when called after `AddComand`?

collector/generate.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+//go:generate builder --config ./builder-config.yaml --skip-compilation
```
thampiotr: Do we need to write somewhere in contributing.md that you want to have `builder` installed? 

Or perhaps ideally we would pull it in automatically and have it pinned to a specific version? I'm not 100% sure how to do this, but I think it should be possible 😅

extension/alloyengine/README.md
```diff
@@ -0,0 +1,55 @@
...
+| `value` | string | Yes | - | The path to the Alloy configuration file to run. |
+| `flags` | map[string]string | No | `{}` | Additional flags to pass to the `alloy run` command. Flags should be specified without the leading `--` prefix. |
+
+### Example Configuration
+
+```yaml
+extensions:
+  alloyengine:
+    format: file
+    value: ./config.alloy
```
thampiotr: I find the direct alloyengine/value a bit confusing. 

Maybe we should try something like this?

```
alloyengine:
  config_path: ...
```

```
alloyengine:
  config_inline: "inline alloy config as string"
```

```
alloyengine:
  config:
    (yaml representation of Alloy cfg)
```

Only one way can be used at a time, it's an error to set more than one.

I feel we talked about this before, sorry if I forgot!

collector/main_alloy.go
blewis12: it does! I gave it a go locally

flowcmd/fallback_version.go
blewis12: Discussed with Joe offline who affirmed that the versioning implementation makes sense, so will resolve this comment chain. From what I can see, and having a chat with Cursor, the `embed` functionality should work as expected

extension/alloyengine/README.md
blewis12: Ah yes we talked about this [here](https://github.com/grafana/alloy/pull/5114#discussion_r2631220312). Maybe it's a nit but my concern with having multiple fields that fail if you try to assign more than one is that it feels less idiomatic beyond a certain number of options. Thinking about how this is done in the OTel world, they have the concept of a [confmap](https://aws-otel.github.io/docs/components/confmap-providers#:~:text=The%20confmap%20providers%20are%20a,and%20all%20the%20other%20components.), which allows users to specify their config in the format of `scheme:value`, where scheme would be inline/file/http and the value would be what gets passed into the config resolver. I sort of like the existing set up in that it mirrors this a bit, albeit split into two separate fields - format and value.

Do we ever see ourselves doing more than just pointing to a file or using inline? If not then I think it makes sense to go with your suggestion, else maybe we'd want something more flexible 🤔

collector/generate.go
blewis12: I like the idea of having the builder auto-install to make things seamless, i've got an incoming change which implements this as well as some other build changes that i'll write a more extended comment on :)

blewis12: I'm going to resolve this comment so we can discuss in one place, perhaps in [this comment chain](https://github.com/grafana/alloy/pull/5114#discussion_r2631258025) makes sense for now

Dockerfile
blewis12: @thampiotr @jharvey10 i've pushed some changes that address the issues of the build process, and whether to check in generated files or not. I feel like these two issues are coupled in how they are addressed so I'd like to discuss them both together. I pulled in a couple of commits to both [remove generated files](https://github.com/grafana/alloy/pull/5114/commits/cda280b58787f1d527fa0bc19b028fa80a4c13e6) and [introduce a new build strategy](https://github.com/grafana/alloy/pull/5114/commits/3b501a8625f8ce844a4207b60cb07a11fd02646f)

**Generated file strategy:**
* Generated files are now not checked in. I actually noticed when playing around locally that generated files can cause a lot of noise - especially because they do appear to regenerate with the same contents but with different white spaces when a dependency is changed, even if it means that not actual generated code is modified.
To be honest I think we could go either way on this one, perhaps some folks think it's clearer to see the files in GitHub and others find it noisy. The main downside is we don't have a clear entry point without running the generation code, but on the plus side we avoid merge conflicts from generated code, reduce repository size, and ensure developers always generate with the latest tooling rather than potentially stale checked-in artifacts.

**Build strategy:**
* The builder tool is downloaded as part of the Alloy build using `go get`. The version is pinned in the BUILDER_VERSION environment variable
* If the builder tool at the specified version is already available in the path this will be used and we won't re-download. As a follow up step, I'd add to our build images logic to install the builder at the specified version, so we're not doing this on every run

**Build flow:**
* Since we have access to the builder tool in CI and we don't check in generated code, we are essentially required to run the same build command as we do locally. This means we don't need to introduce new make targets for running locally vs in CI
* I decided to remove `generate-module-dependencies` as a pre-req for generating the collector. I think it makes sense that this is run manually once replace directives are modified, and this isn't necessary for every build. If a user forgets to run this, they will be notified by the GA that explicitly checks for it

blewis12: Another alternative regarding downloading the builder that would mitigate modifying the users PATH would be to either:

* Include binaries in the Alloy repo, but we'd have to do this for each different architecture type and choose the right one at runtime
* I can't find a way to programmatically download the correct binary at runtime, but we could perhaps download and build-from-source the OCB builder binary as part of the build process. It feels a bit clunky to do it this way though 🤔 maybe there's another way i'm not thinking of

thampiotr: I'm just one opinion here among others, but many tools like even GH code navigation, text search, or even opening the fresh checkout in an IDE won't work without generated code... not to mention that it's harder to read, review and make sense of.

I think that the problem of 'generated code is noisy' can be addressed in a different way. This is probably because the file is not go fmt-ed, or there is some other way in which the noise can be reduced. I would vote on we us keeping the generated files and try to reduce the noise.

Regarding the build strategy - yep, getting the builder with `go get` sounds good. 

> As a follow up step, I'd add to our build images logic to install the builder at the specified version, so we're not doing this on every run

I think the builder image already downloads all the modules in go.mod, right? I am not sure it will save us that much time, depending how long it takes to compile the builder. 

> This means we don't need to introduce new make targets for running locally vs in CI

I don't think we must introduce different make targets, right? With pinned builder + GH action check that generated files are up-to-date, the CI makefile targets should work for local workflow too. 

> this isn't necessary for every build. If a user forgets to run this, they will be notified by the GA that explicitly checks for it

SGTM

Overall I would like to propose:
- check in the generated code
- have a pinned builder - similar to what we have now
- have one makefile target that will be used locally and in CI
- have a check in CI that the generated code is up-to-date
- reduce any generation noise by running go fmt or whatever else is necessary

tools/check-ocb-builder-is-installed.sh
thampiotr, line 1: As mentioned in our chat, I'd rather we don't have a file like this. It should be possible to just use `go install` to our `build` folder or just use `go run` to run the module directly.

Dockerfile
blewis12: Updating this thread since this was discussed externally:

There are some strong arguments for keeping the generated files in this PR - the main being that these are not just intermediary generated files, but they are the main entry point to the program. Merge conflicts themselves can be easily resolved with re-generating the code, and we can ensure that users are always generating the expected files by pinning the `builder` version as we are now. 

Unless there are major complaints there, I will add back the generated files - the build strategy can remain the same (essentially meaning we re-generate in our pipeline). I will also add a GA to check that the output of the collector generation is the same as what is checked in. I need to look a bit further as to why we're getting some weird whitespaces changes, it was suggested to use `go fmt` to help with this.

blewis12: I agree after our discussion that keeping the generated files is best practice here 👍 

> I think the builder image already downloads all the modules in go.mod, right? I am not sure it will save us that much time

Ah, I only mean that we don't have to install the builder on every run if we include it in the build image. We had some discussion as well on different ways that we can invoke the builder without needing to install it on the users PATH, which I can explore now

> I don't think we must introduce different make targets, right?

No I agree - we can just stick with the one, having the builder installed and running the generation both locally and in CI works just fine. I can just add the CI check to validate the sanity of the checked in code vs the generation output

I also realized that the noisy generated files is probably a me-problem 🤦 something funky with how i'm committing files in my IDE. For next steps then I will

* Add back the generated code
* Add a CI check to ensure the generated code is in sync
* Revise the process in which we download the builder and run it

tools/check-ocb-builder-is-installed.sh
blewis12: Go run works nicely 👍 and means we don't even need a script which is great

collector/builder-config.yaml
blewis12: Ok, after having looked into things i'm fairly sure the breaking changes are limited to types moving around - the types themselves remain unchanged. Making a note thought that this component is definitely one to dogfood just to make sure there's nothing wonky there

@thampiotr is it generally best practices to keep our forks in the grafana repo? I've seen a couple point to personal repo's as well, so wasn't sure if there's a preferred approach

Dockerfile
jharvey10: I'm definitely fine with including the generated files if we feel they are a meaningful inclusion in our repo, and it sounds like we do indeed think that they provide context, clarity, and completeness. 👍

extension/alloyengine/README.md
blewis12: After thinking on this and tinkering around and looking for other examples, it's not the greatest to have the `value` yaml key having a changing value depending on the `format`. I think something that might make sense is

```
alloyengine:
    config:
      file:
        path: ./config.alloy
    # inline: 
    #      value: ....
    # yaml: 
    #       path:  ....
```

This introduces a specific config field, which would enforce one-of values in its validation logic. A user can specify a `file`, `inline`, `yaml` etc. It should be fairly easy to validate that only one of these is input in the code, and I like the fact that the different types of config are grouped under a single field instead of just under the `alloyengine` root. As of now we only support `file`, but when we extend this we can implement a one-of check to ensure multiple config types are provided. WDYT?

collector/builder-config.yaml
thampiotr: Definitely we prefer them in grafana repo :) 
Sorry about confusion about the components. I was thinking about `otelcol.exporter.prometheus` vs. [prometheusexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter). But we don't need to worry about it here.

extension/alloyengine/README.md
thampiotr: I like this approach too. I was also thinking about making it simple to implement with standard unmarshalling of yaml into a struct, without a need for some custom logic, just using field tags. And your approach should be simple to implement. 

Only one thought, maybe your approach could be flattened still a bit:
```
alloyengine:
    config:
      file: ./config.alloy
    # inline: "..." 
    # yaml: 
    #       key:  value
    #       ...
```

.github/workflows/check-generate-otel-collector-distro.yml
```diff
@@ -0,0 +1,45 @@
...
+          cache: false
+
+      - name: Run generate-otel-collector-distro
+        run: |
+          make generate-otel-collector-distro
+
+      - name: Check for generated file changes
+        run: |
+          # List changed files in the collector directory (excluding generator source files)
+          CHANGED=$(git diff --name-only --ignore-all-space | grep -E '^collector/.*\.(go|mod)$' | grep -v '^collector/generator/' || true)
```
thampiotr: I am not sure if we need to do that grepping and not sure I follow this :) 

I have used a simpler logic before here: https://github.com/grafana/alloy/blob/main/.github/workflows/check-alloy-mixin-rendered.yml#L34

Perhaps there is a good reason to have the grepping above and I have a bug in the mixing workflow? or perhaps we can simplify this?

extension/alloyengine/README.md
```diff
@@ -0,0 +1,64 @@
...
+| `flags` | map[string]string | No | `{}` | Additional flags to pass to the `alloy run` command. Flags should be specified without the leading `--` prefix. |
+
+### Config Object
+
+The `config` object specifies the Alloy configuration source. It is a one-of type, meaning only one configuration source type can be specified.
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `file` | object | Yes | - | File-based configuration source. |
+| `file.path` | string | Yes | - | The path to the Alloy configuration file to run. |
```
thampiotr: Let's remember to update this with whatever we decide. And the example below.

flowcmd/flowcmd.go
```diff
@@ -25,12 +28,18 @@ func init() {
...
 	}
 
 	prometheus.MustRegister(build.NewCollector("alloy"))
 }
 
-func main() {
-	alloycli.Run()
+// RootCommand exposes the root Cobra command constructed by the internal alloy CLI.
+func RootCommand() *cobra.Command {
+	fmt.Println("hi")
```
thampiotr: some debugging leftover

thampiotr (commented):
I think we're almost there!

flowcmd/flowcmd.go
blewis12: woops, yep forgot to remove that

Dockerfile
blewis12: excellent, resolving this for now but if we want to further discuss I am happy to unresolve

collector/builder-config.yaml
blewis12: gotcha - i've cherry-picked my commits to the grafana fork and referenced that [here](https://github.com/grafana/alloy/pull/5114/commits/de44972132f03c89eb14371865211b7500b6fa89)

i've checked out v0.139.0 of the exporter component before applying my patches so that the version is in sync with the other OTel components

extension/alloyengine/README.md
blewis12: True, my only thought there is that adding the nested format makes it clearer that `file` is in fact a `path`. Previously we called it `file_path`, so having the nested `file.path` translates a bit of that clarity at the expense of a nested field. Then again looking at these nested fields, `config.file` already feels intuitive without the need for an explicit mention of a path

blewis12: https://github.com/grafana/alloy/pull/5114/commits/6ac8095d3717ac274724e154fc3730650f5a2bfe

I've updated the readme a bit too, it's a bit confusing to mention the one-of rule when there's only one type currently

.github/workflows/check-generate-otel-collector-distro.yml
blewis12: oh good catch, I had actually copied this over from the `generate-module-dependencies` action and didn't modify it properly. The grep there is related to narrowing down to go mod/sum files, which isn't needed here

I'll copy your logic, that applies well to this scenario where we're checking against files in a directory

blewis12: https://github.com/grafana/alloy/pull/5114/commits/eeb31f5782d1c97b59a1572ed65fda126a3791fd

extension/alloyengine/README.md
blewis12: updated 👍

.github/workflows/check-generate-otel-collector-distro.yml
```diff
@@ -0,0 +1,43 @@
...
+        run: |
+          make generate-otel-collector-distro
+
+      - name: Check for generated file changes
+        run: |
+          # List changed files in the collector directory (excluding generator source files)
+          if [ -n "$(git status --porcelain -- collector/)" ]; then
+            echo "::error::Generated collector files are out of sync with generate-otel-collector-distro."
+            echo "The following files changed:"
+            echo "$CHANGED"
```
thampiotr: There is no $CHANGED right now

Makefile
```diff
@@ -267,6 +268,13 @@ else
 	cd ./tools/generate-module-dependencies && $(GO_ENV) go generate
 endif
 
+generate-otel-collector-distro:
+ifeq ($(USE_CONTAINER),1)
+	$(RERUN_IN_CONTAINER)
+else
+	cd ./collector && GOOS= GOARCH= BUILDER_VERSION=$(BUILDER_VERSION) go generate
```
thampiotr, line 276: I'm not sure why we need to set GOOS and GOARCH to empty?

thampiotr (approved):
From my POV this is ready % couple of minor comments.
Nice work!

.github/workflows/check-generate-otel-collector-distro.yml
blewis12: ah yes, fixed now to just use your logic from the mixin action

Makefile
blewis12: So we actually get failures without that - link [here](https://github.com/grafana/alloy/actions/runs/20822313663/job/59814037988?pr=5114)

The way I understand it is:

* We set the GOOS and GOARCH variables in order to cross compile the alloy binary
* Those variables are then passed down to the generator file, so we wind up running something like 
```
 GOOS=freebsd GOARCH=amd64 go run go.opentelemetry.io/collector/cmd/builder@${BUILDER_VERSION}
```
* So now we're cross compiling the builder for freebsd, and we cannot run that on our runner which is `github-hosted-ubuntu-x64`

Clearing these variables just means we don't cross compile the builder, and instead allow `go run` to default to whatever the host machine is

blewis12: I've added a comment in the Makefile, just to be clear about the intention: https://github.com/grafana/alloy/pull/5114/commits/b978015dbcc8d1524ac9b57f9ef85693d54fc147

thampiotr: ah, cool, that makes perfect sense

extension/alloyengine/README.md
```diff
@@ -0,0 +1,63 @@
...
+The extension manages the lifecycle of the embedded default engine:
+
+- **Start**: When the extension starts, it launches the default engine in a separate goroutine, executing the specified Alloy configuration file
+- **Ready**: The extension reports ready once the default engine has successfully started
+- **Shutdown**: When the extension is shut down, it gracefully terminates the default engine and waits for it to exit
+
+## Stability
+
+This extension is currently marked as **Development** stability level. The API and behavior may change in future releases.
+
```
kgeckhart, line 63: Non-blocking question: I'm guessing this is using the OTel stability definitions for upstreaming purposes. Since it lives in alloy for now should we use the alloy stability levels until it can be upstreamed?

blewis12: That's a good point actually, I was trying to make this as OTel-ey as possible but when we talk about this feature externally we should be sticking to Alloy stability levels for the time being. I will change that now

kgeckhart (commented):
Non-blocking general comment: I found the PR description incredibly helpful for understanding the layout of the changes. I think it could be useful to find a home for some of the content in the README.md and perhaps introducing a collector/README.md

blewis12:
>  It could be useful to find a home for some of the content in the README.md and perhaps introducing a collector/README.md

True, there is going to be a follow up PR containing more documentation - but since this PR includes changes that primarily affect developers and therefore should be referenced in the contributing guide, i've added a README [here](https://github.com/grafana/alloy/pull/5114/commits/456c6607bc00cab43d9793fdfe2beaf8ac58b16c) - along with a brief summary and link in the contributing.md file

PR #5115, opened 2025-12-16
feat: database_observability: add health_check collector to validate configuration

matthewnolf:
#### PR Description
Adds a new collector within the `database_observability.mysql` component for reporting diagnostic data of the target database. This data enables us to visualise configuration issues.

Many of the healthChecks have been removed since we can access the required data from other collectors/signals. Remaining are just a few which we can not otherwise query.
#### PR Checklist

- [x] Documentation added
- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
feat(database_observability): Add health_check collector to validate configuration (#5115)
END_COMMIT_OVERRIDE


internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	r.result = true
+	r.value = "ok"
+	return r
+}
+
+// checkMySQLVersion validates that the database the MySQL version >= 8.0.
+func checkMySQLVersion(ctx context.Context, db *sql.DB) healthCheckResult {
+	r := healthCheckResult{name: "MySQLVersionSupported"}
+	const q = `SELECT VERSION()`
```
cristiangreco: Could we re-use the value from `component.go` (which already runs this query)?

internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -0,0 +1,356 @@
...
+		return r
+	}
+	r.result = true
+	return r
+}
+
+// checkSetupConsumerCPUTimeEnabled validates events_statements_cpu consumer is enabled.
+func checkSetupConsumerCPUTimeEnabled(ctx context.Context, db *sql.DB) healthCheckResult {
+	r := healthCheckResult{name: "SetupConsumerCPUTimeEnabled"}
+	const q = `SELECT enabled FROM performance_schema.setup_consumers WHERE NAME = 'events_statements_cpu'`
```
cristiangreco: Is it worth having a single check for `setup_consumers` maybe?

internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	r.result = strings.EqualFold(varValue, "ON") || varValue == "1"
+	r.value = varValue
+	return r
+}
+
+// checkDBConnectionValid validates the database connection with a short timeout.
+func checkDBConnectionValid(ctx context.Context, db *sql.DB) healthCheckResult {
+	r := healthCheckResult{name: "DBConnectionValid"}
```
cristiangreco: Wondering if `mysql_up` can do the job here?

internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// checkDigestVariablesLength ensures text/digest length variables are >= 4096.
+func checkDigestVariablesLength(ctx context.Context, db *sql.DB) healthCheckResult {
+	r := healthCheckResult{name: "DigestVariablesLengthCheck"}
+	const q = `
+SELECT
+	@@performance_schema_max_sql_text_length,
+	@@performance_schema_max_digest_length,
+	@@max_digest_length`
```
cristiangreco: Could we use perhaps these?

mysql_global_variables_performance_schema_max_digest_length
mysql_global_variables_performance_schema_max_sql_text_length
mysql_global_variables_max_digest_length

internal/component/database_observability/mysql/collector/health_check.go
cristiangreco: Or maybe we could just use `database_observability_setup_consumers_enabled`?

matthewnolf:
@cristiangreco I like the approach of using data already available instead of repeating it here. I have removed the healthChecks where the data was available in other metrics

cristiangreco (approved):

PR #5125, opened 2025-12-16
chore: use shared source library in loki.source.cloudflare

kalleep:
#### PR Description
Similar to other pr in that we should use shared functions and structure for loki source.

Like in https://github.com/grafana/alloy/pull/5026 I flattened the structure (no internal lib) and renamed traget to tailer.

But unlike e.g. loki.source.file and loki.source.docker this only ever have one tailer so no need to have a scheduler for it. 

Part of: https://github.com/grafana/alloy/issues/4729

#### Notes to the Reviewer

I remove manual validation for FieldsType, instead I implemented `encoding.TextMarshaler`  and `encoding.TextUnmarshaler`

#### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/cloudflare/tailer.go
+++ b/internal/component/loki/source/cloudflare/tailer.go
@@ -146,7 +146,6 @@ func (t *tailer) start() {
 			}
 		}
 	})
-
 }
 
 // pull pulls logs from cloudflare for a given time range.


kalleep:
We don't need a scheduler for this component but if we want to be consistent with other sources we could add it anyway.

ptodev (approved):

PR #4991, opened 2025-12-03
feat: database_observability: stop tracking own instrumentation queries

cristiangreco:
#### PR Description
This PR introduces a change for excluding "own queries" in postgres from `pg_stat_activity`: we need to drop samples (and wait events) for the currently connected user. This new behaviour is enabled by default through the setting `exclude_current_user`, as it's deemed safe / recommended.

No changes to be done on `pg_stat_statements` as that is achieved through user permissions update (docs on the website).

#### Which issue(s) this PR fixes
Relates to https://github.com/grafana/grafana-dbo11y-app/issues/1307

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [ ] Config converters updated

BEGIN_COMMIT_OVERRIDE
feat(database_observability): Stop tracking own instrumentation queries (#4991)

This PR introduces a change for excluding "own queries" in postgres from
`pg_stat_activity`: we need to drop samples (and wait events) for the
currently connected user. This new behaviour is enabled by default
through the setting `exclude_current_user`, as it's deemed safe /
recommended.

No changes to be done on `pg_stat_statements` as that is achieved
through user permissions update (docs on the website).
END_COMMIT_OVERRIDE

PR #5130, opened 2025-12-16
fix(database_observability): Solve test flakiness in MySQL and Postgres sample collectors

gaantunes:
#### PR Description

Fixing some flaky tests by:

Adding Waitgroups for better routine handlling.

#### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


internal/component/database_observability/mysql/collector/query_samples_test.go
```diff
@@ -198,14 +198,14 @@ func TestQuerySamples(t *testing.T) {
 				require.Equal(t, len(entries), len(tc.logsLines))
 
 				require.Contains(t, logBuffer.String(), tc.errorLine)
-			}, 5*time.Second, 100*time.Millisecond)
+			}, 10*time.Second, 200*time.Millisecond)
```
cristiangreco: I forgot to mention that I went through these tests in https://github.com/grafana/alloy/pull/4991, changing the code around expectations on log lines to try to fix the flakyness - don't think we need to increase timeouts here.

gaantunes: No need for this PR then?

cristiangreco: Probably not, I haven't seen other flaky occurrences. However I see you changed `query_samples.go` as well, and seems your update might fix a leak of the `runSetupConsumersCheck` goroutine? If so, maybe that specific change could be reviewed in a separate PR.

cristiangreco (approved):

PR #5131, opened 2025-12-16
fix: improved postgres version parsing for explain plans in database_observability component

rgeyer:
#### PR Description
Effectively any version string that wasn't a very traditional semver string was failing to parse. This would impact many installations but was called out specifically for a Percona DB and a postgres docker image.

Specifically, these strings had "extra" bits at the end of the version string which semver was treating as part of the minor version.

#### PR Checklist

- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
fix(database_observability): Improve postgres version parsing for explain plans in database_observability component (#5131)
END_COMMIT_OVERRIDE


internal/component/database_observability/postgres/collector/explain_plan.go
```diff
@@ -253,7 +253,16 @@ func NewExplainPlan(args ExplainPlanArguments) (*ExplainPlan, error) {
 		entryHandler:        args.EntryHandler,
 		logger:              log.With(args.Logger, "collector", ExplainPlanCollector),
 		running:             atomic.NewBool(false),
-	}, nil
+	}
+	// Pre-sanitize the version a bit before semver gets it
```
matthewnolf: nit: "a bit" is not very clear

internal/component/database_observability/postgres/collector/explain_plan_test.go
matthewnolf, line 1: Should we add a test for the exact versions that are failing today?

rgeyer: Did do, on lines 2293-2305. That said, this is just one of the failing version strings, but it's one I directly experienced with the postgres docker image.

I didn't get a chance to grab the full Percona version number that's failing, but the error is the same, semver is trying to parse spaces and special characters as a minor version. The changes (and the test) account for basically any combination of that.

internal/component/database_observability/postgres/collector/explain_plan.go
rgeyer: Fair enough, hopefully it is more informative now.

cristiangreco (approved):
Approving as I see Matt's comments have been addressed already

PR #5143, opened 2025-12-18
fix: Compute signatures from files so that loki.source.file can handle atomic writes

kalleep:
#### PR Description
Noticed when doing some manual testing. Our tailer do _not_ handle atomic writes, e.g. neovim will perform these kind of writes on every save.

Not sure how big of a problem this is in "normal" log rotation environments. What happens before this pr is that if I have a log file in neovim and save `:w` we will detect a delete event, reopen the file and consume it fully.

If this is something we want to handle we can compute a signature. The signature consist of currently hard coded number of bytes (1024). 

This pr handle incomplete signatures and recompute it once we have read past desired size.

#### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/atomic_write_posix_test.go
+++ b/internal/component/loki/source/file/internal/tail/atomic_write_posix_test.go
@@ -10,6 +10,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func atomicwrite(t *testing.T, name, newContent string) {
+func atomicWrite(t *testing.T, name, newContent string) {
 	require.NoError(t, atomic.WriteFile(name, strings.NewReader(newContent)))
 }
--- a/internal/component/loki/source/file/internal/tail/file_test.go
+++ b/internal/component/loki/source/file/internal/tail/file_test.go
@@ -389,7 +389,7 @@ func TestFile(t *testing.T) {
 		// Read first two lines
 		verify(t, file, &Line{Text: "line1", Offset: 6}, nil)
 		verify(t, file, &Line{Text: "line2", Offset: 12}, nil)
-		atomicwrite(t, name, "line1\nline2\nline3\nline4\nnewline1\n")
+		atomicWrite(t, name, "line1\nline2\nline3\nline4\nnewline1\n")
 		verify(t, file, &Line{Text: "line3", Offset: 18}, nil)
 		verify(t, file, &Line{Text: "line4", Offset: 24}, nil)
 		verify(t, file, &Line{Text: "newline1", Offset: 33}, nil)
@@ -412,7 +412,7 @@ func TestFile(t *testing.T) {
 		// Read first two lines
 		verify(t, file, &Line{Text: "line1", Offset: 6}, nil)
 		verify(t, file, &Line{Text: "line2", Offset: 12}, nil)
-		atomicwrite(t, name, "newline1\n")
+		atomicWrite(t, name, "newline1\n")
 		// Because we buffer lines when file is deleted we still get line3 and line4.
 		verify(t, file, &Line{Text: "line3", Offset: 18}, nil)
 		verify(t, file, &Line{Text: "line4", Offset: 24}, nil)
--- a/internal/component/loki/source/file/internal/tail/atomic_write_windows_test.go
+++ b/internal/component/loki/source/file/internal/tail/atomic_write_windows_test.go
@@ -13,10 +13,10 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-// atomicwrite performs an atomic write on Windows using ReplaceFileW.
+// atomicWrite performs an atomic write on Windows using ReplaceFileW.
 // This allows replacing a file when it's open with FILE_SHARE_DELETE.
 // This function is only used in tests.
-func atomicwrite(t *testing.T, name, newContent string) {
+func atomicWrite(t *testing.T, name, newContent string) {
 	var (
 		kernel32     = windows.NewLazySystemDLL("kernel32.dll")
 		replaceFileW = kernel32.NewProc("ReplaceFileW")


internal/component/loki/source/file/internal/tail/signature.go
```diff
@@ -0,0 +1,77 @@
...
+type signature struct {
+	d []byte
+}
+
+// signatureSize is the target size for a complete signature.
+const signatureSize = 512
+
+// signatureThresholds defines the byte offsets at which we should recompute the signature
+// as the file grows. This allows us to progressively build a more complete signature.
+var signatureThresholds = []int{64, 128, 256, 512}
```
thampiotr: Why add this complication? could we wait for the signatureSize and assume it's 0 at first? or unknown?

internal/component/loki/source/file/internal/tail/signature.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// signature represents the first N bytes of a file, used to detect atomic writes
+// where a file is replaced but may have the same initial content.
+type signature struct {
+	d []byte
+}
+
+// signatureSize is the target size for a complete signature.
+const signatureSize = 512
```
thampiotr: I see some other implementations use 1KiB, maybe let's switch to 1024 here

internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -116,6 +122,25 @@ read:
...
+
+		sig, err := newSignatureFromFile(f.file)
+		if err != nil {
+			return nil, err
+		}
+
+		f.signature = sig
+		if _, err := f.file.Seek(f.lastOffset, io.SeekStart); err != nil {
+			return nil, err
+		}
```
thampiotr: Should seeking back be in defer? in case we exit at line 133?

internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -116,6 +122,25 @@ read:
 
 	f.lastOffset = offset
 
+	// Recompute signature if we've crossed a threshold and haven't reached it yet.
+	// This progressively builds a more complete signature as the file grows.
+	if f.signature.shouldRecompute(f.lastOffset) {
```
thampiotr: I wonder if we could put some of this logic somewhere else, these function is already pretty big

internal/component/loki/source/file/internal/tail/signature.go
kalleep: It kinda depends on when you start tailing a file. 

If we start tailing before the file is `signatureSize` we take whatever is in there. So if it's 0 then that is our current signature.

If we only update the signature again when we reach `signatureSize` there is are some problems:

* If we started at 0 we asume it's not a match causing rereads. 
* The smaller the signature is the higher likely hood of collision causing missed logs. 

And that is what we are trying to combat with this. But this is for sure a edge case with rotation happening on really small files. In real workloads I don't think this should cause problems

internal/component/loki/source/file/internal/tail/file.go
kalleep: Where would you put it?

internal/component/loki/source/file/internal/tail/file.go
kalleep: We don't expect it to fail at 133 and if it does the tailer will stop, so it would not do anything for us.

internal/component/loki/source/file/internal/tail/signature.go
```diff
@@ -0,0 +1,77 @@
...
+
+// equal compares two signatures. For incomplete signatures, it compares only
+// the overlapping bytes. For complete signatures, both must be the same length and content.
+func (s *signature) equal(other *signature) bool {
+	len1 := len(s.d)
+	if len1 == 0 {
+		return false
+	}
+
+	len2 := len(other.d)
```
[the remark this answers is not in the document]
kalleep: It's fine to handle them as not equal in this case

internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -116,6 +122,25 @@ read:
...
+		if err != nil {
+			return nil, err
+		}
+
+		f.signature = sig
+		if _, err := f.file.Seek(f.lastOffset, io.SeekStart); err != nil {
+			return nil, err
+		}
+		f.reader.Reset(f.file)
+	}
```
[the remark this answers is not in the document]
kalleep: Yes this is a bit more expensive but will only happen when we don't have reached the "optimal" threshold. If we start tailing file that is already bigger than 1024 bytes this does not happen, and in the other case it only happens until we have reached that size

kalleep:
@thampiotr I changed to use `ReadAt` in newSignatureFromFile so we no longer need to perform the extra seek calls we did before.

kalleep:
Windows test failing is unrelated to this pr `TestSourceMapsStoreImpl_CleanOldCachedEntries`. Will take a look at fixing that test in another pr.

internal/component/loki/source/file/internal/tail/fileext/file_windows.go
```diff
@@ -113,3 +114,75 @@ func getFileStandardInfo(f *os.File) (*fileStandardInfo, error) {
 	runtime.KeepAlive(f)
 	return si, nil
 }
+
+// AtomicWrite performs an atomic write on Windows using ReplaceFileW.
+// This allows replacing a file when it's open with FILE_SHARE_DELETE.
+// This function is only used in tests.
+func AtomicWrite(name string, content []byte) error {
```
thampiotr: Getting file operations right is not easy, especially on Windows. This looks super complex, so I wouldn't want this code to be used in production. 

I would like to make it very obvious that this is only used in tests and cannot and should not be ported to production code. So can we put this inside whatever test file needs it and have a switch on `runtime.GOOS` to change implementation? 

Or if that wouldn't compile, can we create a special `testutil` package for this function only and require to pass a `testing.T` to it so it's not possible to call it outside of tests? And we should document that this should not be used in production. 

If we can do that, I will be okay for this to be merged 👍

internal/component/loki/source/file/internal/tail/file_test.go
```diff
@@ -404,3 +453,7 @@ func rotateFile(t *testing.T, name, newContent string) {
 	// Create new file with same name
 	require.NoError(t, os.WriteFile(name, []byte(newContent), 0600))
 }
+
+func atomicwrite(t *testing.T, name, newContent string) {
```
thampiotr: Let's capitalise it as AtomicWrite.

internal/component/loki/source/file/internal/tail/fileext/file_windows.go
kalleep: Yeah I agree, this one is simple enough but I will put it in a test only file :)

internal/component/loki/source/file/internal/tail/signature.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// completed returns true if the signature has reached the target size.
+func (s *signature) completed() bool {
+	return len(s.d) == signatureSize
+}
+
+// equal compares two signatures. For incomplete signatures, it compares only
+// the overlapping bytes. For complete signatures, both must be the same length and content.
+// If signature has a lenght of 0 equal will always return false.
```
thampiotr:
```suggestion
// If signature has a length of 0 equal will always return false.
```

internal/component/loki/source/file/internal/tail/fileext/file_windows.go
kalleep: fixed in [f997287](https://github.com/grafana/alloy/pull/5143/commits/f997287d2b147dea8cbd5ad753e5b4fb8dbf655b)

internal/component/loki/source/file/internal/tail/fileext/file_windows.go
```diff
@@ -113,3 +114,75 @@ func getFileStandardInfo(f *os.File) (*fileStandardInfo, error) {
...
+	ret, _, errno := replaceFileW.Call(
+		uintptr(unsafe.Pointer(replacedPath)),
+		uintptr(unsafe.Pointer(replacementPath)),
+		0, // NULL backup file
+		0, // No flags
+		0, // NULL exclude
+		0, // NULL reserved
+	)
+
+	// errno is always non-nil so we should only sett err if ret is 0.
```
thampiotr:
```suggestion
	// errno is always non-nil so we should only set err if ret is 0.
```

internal/component/loki/source/file/internal/tail/atomic_write_posix_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/natefinch/atomic"
+	"github.com/stretchr/testify/require"
+)
+
+func atomicwrite(t *testing.T, name, newContent string) {
```
thampiotr: Let's capitalise it AtomicWrite following Go convention.

PR #5144, opened 2025-12-18
docs: remove SECURITY.md

Proximyst:
We have an org-wide SECURITY.md file that will be shown by GitHub in the repository UI automatically. To avoid confusion and the files falling out of sync, we should remove the local copy.

Note that the PGP key will change; the old key that is still listed in Alloy's current local copy is expired and thus invalid.

Example of repository with no local copy of SECURITY.md:
image


Proximyst:
Nope, just by removing the old file, the new file is adopted. 🪄

PR #5146, opened 2025-12-18
perf: Avoid eager fmt.Sprintf in stage.logfmt debug logging

brainexe:
#### PR Description

Remove eager `fmt.Sprintf` calls in `logfmtStage.Process()` debug logging and use structured logging key-value pairs instead. This eliminates significant CPU overhead when debug logging is disabled.

## Motivation

CPU profiling in our production system  showed `fmt.Sprintf` consuming significant CPU cycles in the logfmt stage's `Process()` method even though debug logging was disabled. Overall ~2.3% of the whole cpu time.
The issue is that `fmt.Sprintf` was being evaluated eagerly before `level.Debug()` checked whether the log level was enabled, causing unnecessary formatting work on every log entry.

sprintf-2 3

## Changes

With structured logging, values are only formatted when the log is actually written. When debug is disabled, `level.Debug()` returns a no-op logger that discards arguments without formatting them.

## Benchmark results

Performance improvement of the single log line when debug logging is disabled (the common production case):
In my test case it went down from ~26,000 ns/op to ~0.3 ns/op without memory allocation anymore. 

#### PR Checklist

- [x] ~Documentation added~
- [x] All existing unit tests pass without modification
- [x] Build completes successfully
- [x] Debug log output verified to be correct when debug logging is enabled (the first message is slightly differnt, as the numbers as not part of the message)

### DIFF
--- a/internal/component/loki/process/stages/replace.go
+++ b/internal/component/loki/process/stages/replace.go
@@ -125,7 +125,9 @@ func (r *replaceStage) Process(labels model.LabelSet, extracted map[string]inter
 			}
 		}
 	}
-	level.Debug(r.logger).Log("msg", "extracted data debug in replace stage", "extracted_data", fmt.Sprintf("%v", extracted))
+	if Debug {
+		level.Debug(r.logger).Log("msg", "extracted data debug in replace stage", "extracted_data", extracted)
+	}
 }
 
 func (r *replaceStage) getReplacedEntry(matchAllIndex [][]int, input string, td map[string]string, templ *template.Template) (string, map[string]string, error) {


kalleep:
Thanks, the problem mostly stem from that we don't have a way to inspect log level with the current logging interface we use. So even with your changes we are doing unnecessary works when debug if off. For loki.proccess we have this https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/util.go#L13-L19.

We use this in various places exactly to combat the issue you are seeing e.g.  https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/drop.go#L140

I think we should consider using that instead because then we are skipping allocations entirely

brainexe:
ah yea, good points, I missed the `Debug` flag here :+1: 

This will also kinda eliminate the 0.7% overall usage of the `level.Debug` part, at least from our production case, so in total -~3% usage with teh fmt.Sprintf.
Bildschirmfoto vom 2025-12-19 14-42-14

kalleep:
@brainexe Not sure why [PR Title Lint / Validate PR title (pull_request)](https://github.com/grafana/alloy/actions/runs/20396587576/job/58712697547?pr=5146) fails currently. Will be out a while during the holidays but will make sure this gets merged when I am back.

Thanks for the pr!

brainexe:
@kalleep small *ping*. ready to get merged? :)

kalleep:
@brainexe Sorry went on holidays and totally forgot about this one

PR #5140, opened 2025-12-18
feat(loki.source.syslog): support raw format

x1unix:
#### PR Description

This PR adds support of `raw` format to consume non-RFC5424 compliant logs such as [CEF](https://www.splunk.com/en_us/blog/learn/common-event-format-cef.html).\
The format skips syslog message parsing, except non-transparent/octet count framing detection.

#### Which issue(s) this PR fixes

Fixes #1711

#### PR Checklist

- [x] Documentation added
- [x] Tests updated
- [x] Config converters updated

BEGIN_COMMIT_OVERRIDE
feat(loki.source.syslog): Support raw format (#5140)
END_COMMIT_OVERRIDE

PR #5165, opened 2026-01-02
feat(loki.source.syslog): support cisco-specific syslog fields

x1unix:

This is a followup PR to a [previous PR](https://github.com/grafana/alloy/pull/5140) to extend `loki.source.syslog` component to enable parsing IOS-specific syslog fields.

This PR exposes existing cisco-specific go-syslog [parser options](https://github.com/leodido/go-syslog/tree/develop/rfc3164#cisco-device-configuration) to component arguments.

#### PR Checklist

- [x] Documentation added
- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
feat(loki.source.syslog): Support cisco-specific syslog fields (#5165)
END_COMMIT_OVERRIDE

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.syslog.md
+++ b/docs/sources/reference/components/loki/loki.source.syslog.md
@@ -191,18 +191,18 @@ The `rfc3164_cisco_components` configures parsing of non-standard Cisco IOS sysl
 This block can only be used when you set `syslog_format` to `rfc3164`.
 {{< /admonition >}}
 
-The following argument is supported:
+The following arguments are supported:
 
 | Name               | Type   | Description                                     | Default | Required |
 |--------------------|--------|-------------------------------------------------|---------|----------|
 | `enable_all`       | `bool` | Enables all components below.                   | `false` | no       |
 | `message_counter`  | `bool` | Enables syslog message counter field parsing.   | `false` | no       |
 | `sequence_number`  | `bool` | Enables service sequence number field parsing.  | `false` | no       |
-| `hostname`         | `bool` | Enables origin hostname fleld parsing.          | `false` | no       |
-| `second_fractions` | `bool` | Enables miliseconds parsing in timestamp field. | `false` | no       |
+| `hostname`         | `bool` | Enables origin hostname field parsing.          | `false` | no       |
+| `second_fractions` | `bool` | Enables milliseconds parsing in timestamp field.| `false` | no       |
 
 {{< admonition type="note" >}}
-At-least one option has to be enabled if `enable_all` is set to `false`.
+At least one option has to be enabled if `enable_all` is set to `false`.
 {{< /admonition >}}
 
 {{< admonition type="caution" >}}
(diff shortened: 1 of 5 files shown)


docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -177,6 +181,65 @@ The following argument is supported:
...
+
+{{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+The `rfc3164_cisco_components` configures parsing of non-standard Cisco IOS syslog extensions. 
+
+{{< admonition type="note" >}}
+This block can only be used when you set `syslog_format` to `rfc3164`.
+{{< /admonition >}}
+
+The following argument is supported:
```
jharvey10:
```suggestion
The following arguments are supported:
```

internal/component/loki/source/syslog/types.go
```diff
@@ -25,13 +26,36 @@ type ListenerConfig struct {
...
+}
+
+func (sc *RFC3164CiscoComponents) Validate() error {
+	if sc == nil || sc.EnableAll {
+		return nil
+	}
+
+	isEmpty := !sc.Hostname && !sc.MessageCounter && !sc.SecondFractions && !sc.SequenceNumber
+	if isEmpty {
+		return errors.New("at least one option in rfc3164_cisco_components has to be enabled")
```
jharvey10, line 53: ~Could we tweak this error to better align with _either_ of the user actions that are required? They must either enable one of these options _or_ set `enable_all` to true.~

Do we have precedence in Alloy for the `enable_all` behavior like this? I know for something like the unix exporter, we have a default set of things that get used if nothing is provided, which can be overridden in a couple of different ways.

If we can, it would be great to provide a set of sensible defaults for users. I don't have the domain knowledge to know if "all" of them being on by default would be appropriate, but perhaps something to consider.

x1unix: > Do we have precedence in Alloy for the enable_all behavior like this? I know for something like the unix exporter, we have a default set of things that get used if nothing is provided, which can be overridden in a couple of different ways.

Yes, `enable_all` takes priority over all other flags:

```go
parseCfg.RFC3164CiscoComponents = &syslogparser.RFC3164CiscoComponents{
	MessageCounter:  ciscoCfg.EnableAll || ciscoCfg.MessageCounter,
	SequenceNumber:  ciscoCfg.EnableAll || ciscoCfg.SequenceNumber,
	CiscoHostname:   ciscoCfg.EnableAll || ciscoCfg.Hostname,
	SecondFractions: ciscoCfg.EnableAll || ciscoCfg.SecondFractions,
}

```


> If we can, it would be great to provide a set of sensible defaults for users. I don't have the domain knowledge to know if "all" of them being on by default would be appropriate, but perhaps something to consider.

The problem is that parser settings are very config-specific and should match the IOS configuration (this mentioned in [the parser docs](https://github.com/leodido/go-syslog/tree/develop/rfc3164#cisco-device-configuration
)):


> Important: Your parser configuration must match your Cisco device configuration. The parser cannot auto-detect which components are present because they share similar formats (mostly digits followed by colon).

jharvey10: As per our discussion on Slack, I understand this better and think this aligns well with the device configuration semantics. `enable_all` is a convenience for users who use everything in the supported set, and if this set becomes large in the future, we can always revisit the way these are specified.

PR #5166, opened 2026-01-02
fix: update syslog parser tests to account for new years logic

blewis12:
Happy new year!

Our syslogparser tests began failing, which can be traced back to some special logic handled in [this function](https://github.com/grafana/alloy/blob/923d127c50949f88a37ac808154240de0649df09/internal/util/time.go#L9). Because our tests are now running in January and are parsing the month of December, the output year will _not_ be equal to time.now().year(), instead our SetYearForLimitedTimeFormat function returns now().year()-1. Similarly, if we're attempting to parse a January time in the month of December, our SetYearForLimitedTimeFormat returns now().year()+1

This PR introduces tests so that this behavior is explicit, and adjusts the existing test to account for the logic

BEGIN_COMMIT_OVERRIDE
test: update syslog parser tests to account for new years logic (#5166)
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/component/loki/source/syslog/internal/syslogtarget/syslogparser/syslogparser_test.go
+++ b/internal/component/loki/source/syslog/internal/syslogtarget/syslogparser/syslogparser_test.go
@@ -137,13 +137,13 @@ func TestParseStream_RFC3164TimestampWithYear(t *testing.T) {
 	// Manually calculate the expected year based on the edge case logic (SetYearForLimitedTimeFormat)
 	now := time.Now()
 	expectedYear := now.Year()
-	if now.Month() == 1 {
+	if now.Month() == time.January {
 		expectedYear = now.Year() - 1
-	} else if now.Month() == 12 {
+	} else if now.Month() == time.December {
 		expectedYear = now.Year() + 1
 	}
 
-	require.Equal(t, time.Date(expectedYear, 12, 1, 0, 0, 0, 0, time.UTC), *results[0].Message.(*rfc3164.SyslogMessage).Timestamp)
+	require.Equal(t, time.Date(expectedYear, time.December, 1, 0, 0, 0, 0, time.UTC), *results[0].Message.(*rfc3164.SyslogMessage).Timestamp)
 }
 
 // Tests the edge case where a December log is parsed in January, which should use the previous year.


internal/component/loki/source/syslog/internal/syslogtarget/syslogparser/syslogparser_test.go
```diff
@@ -132,5 +133,49 @@ func TestParseStream_RFC3164TimestampWithYear(t *testing.T) {
...
+	// Manually calculate the expected year based on the edge case logic (SetYearForLimitedTimeFormat)
+	now := time.Now()
+	expectedYear := now.Year()
+	if now.Month() == 1 {
+		expectedYear = now.Year() - 1
+	} else if now.Month() == 12 {
+		expectedYear = now.Year() + 1
+	}
+
+	require.Equal(t, time.Date(expectedYear, 12, 1, 0, 0, 0, 0, time.UTC), *results[0].Message.(*rfc3164.SyslogMessage).Timestamp)
```
jharvey10: Does it make senes here to also use `time.December` like below? Also same question for the `now.Month() == 12` above, etc.

blewis12: can do! I essentially just copied what was in `SetYearForLimitedTimeFormat` but using `time.` is more idiomatic

blewis12: https://github.com/grafana/alloy/pull/5166/commits/20ca6ee9bcb721815350182fbabbf1ca59443523

jharvey10: One more here? 🙂  

```go
require.Equal(t, time.Date(expectedYear, 12, 1, 0, 0, 0, 0, time.UTC), *results[0].Message.(*rfc3164.SyslogMessage).Timestamp)
                                      // ^^
```

blewis12: done 👍 didn't notice that one

jharvey10 (approved):

PR #5169, opened 2026-01-05
docs: Add note to Prometheus tutorial

clayton-cornell:
By default, Prometheus relies on a pull mechanism to scrape metrics from targets. The Prometheus tutorail needs to be updated to add a note explaining how to configure Prometheus to accept pushed telemetry.

### DIFF
--- a/docs/sources/tutorials/send-metrics-to-prometheus.md
+++ b/docs/sources/tutorials/send-metrics-to-prometheus.md
@@ -217,9 +217,7 @@ If you choose to run Alloy in a Docker container, make sure you use the `--serve
 If you don't use this argument, the [debugging UI][debug] won't be available outside of the Docker container.
 
 [debug]: ../../troubleshoot/debug/#alloy-ui
-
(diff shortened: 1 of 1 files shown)


kgeckhart (commented):
Is that flag necessary anymore? It's not referenced in the latest docs for 3.8, https://prometheus.io/docs/prometheus/latest/feature_flags/. I can only find it referenced in the old docs for 2.55, https://prometheus.io/docs/prometheus/2.55/feature_flags/#remote-write-receiver, with a note that `--web.enable-remote-write-receiver` should be used because `--enable-feature=remote-write-receiver` is deprecated.

clayton-cornell:
> Is that flag necessary anymore? It's not referenced in the latest docs for 3.8

Oh dear.. I got nailed by bad search results and I didn't dig deep enough. After some more digging and heading in the right direction as you suggested, I've revised the text, and included a link back to the Prometheus API documentation to provide more context/information.

PR #5174, opened 2026-01-05
chore: revamp some contributor docs and pr template

jharvey10:
Add better instructions for the 90%+ use case.

### DIFF
--- a/docs/developer/contributing.md
+++ b/docs/developer/contributing.md
@@ -112,13 +112,13 @@ Slack channel [#alloy](https://slack.grafana.com).
 
 ### tl;dr:
 
-PR titles (and by extension CHANGELOG entries) should:
+#### PR titles (and by extension CHANGELOG entries) should:
 
 1. Adhere to [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) style and use one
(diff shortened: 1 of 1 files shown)


docs/developer/contributing.md
```diff
@@ -110,6 +110,33 @@ Slack channel [#alloy](https://slack.grafana.com).
...
+
+For example: `feat: increase config file read speed by 1500%`
+
+> Readers should be able to understand how a change impacts them. Default to being explicit over
+> vague.
+>
+> - Vague: `fix: fix issue with metric names`
+> - Explicit: `fix: fix 's' getting replaced by 'z' in metric names`
+
+PR "Extended descriptions" (i.e. commit bodies):
```
blewis12: it might be worth noting that this field is automatically populated with commits, and in order to follow these instructions one has to clear the generated contents? Just because personally I wasn't sure at first

blewis12 (approved):

docs/developer/contributing.md
jharvey10: Good thinking! I say this down below, but I've now put a small note about it in the tl;dr section as a quick reference.

PR #5179, opened 2026-01-06
docs(proposal): Adjust design doc process

kgeckhart:
#### PR Description
Adjust the design doc process for proposals by,
* Refactoring the template to include multiple options
* Proposal issue stays open even with a design doc for work tracking purposes
* Drops a declined reason that we haven't used from the proposal process (I do struggle with the value for the remaining declined reasons but the remaining ones have 2-4 proposals)


docs/design/README.md
```diff
@@ -209,15 +202,6 @@ Proposals are in one of the following states:
   Reconsidering a previously declined proposal with the same approach may be
   considered when there is significant new information.
 
-* **Declined as Infeasible**: If a proposal directly contradicts the core
```
blewis12, line 212: I assume that the reasons one might `Decline as Infeasible` would now be grouped under the general `Declined`? I actually wasn't aware we had multiple decline modes until now

kgeckhart: Exactly, the extra reasons for declined feel like they don't add much value. This particular one has had no proposals in it so it felt like it could get removed. I'll do a follow up to propose removing the rest of them.

PR #5196, opened 2026-01-07
feat(database_observability.postgres): Add adaptive burst window mechanism to query_samples collector

gaantunes:
### Pull Request Details

This burst mechanism is designed to capture the active query execution span and wait events more precisely. Whenever we detect an active row, a scrape burst window is opened. The goal is to try to capture the moment when the execution ends as precisely as possible.

The query was changed for improved performance, as `pg_stat_activity` is very lightweight (all data is kept by Postgres in memory), but `pg_blocking_pids` can add some overhead. We will only call the second if there are any Lock wait types in first scrape.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/dsn.go
+++ b/internal/component/database_observability/postgres/collector/dsn.go
@@ -8,10 +8,12 @@ import (
 
 type databaseConnectionFactory func(dsn string) (*sql.DB, error)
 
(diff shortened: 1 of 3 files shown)


matthewnolf:
@gaantunes Is this still needed?

gaantunes:
> @gaantunes Is this still needed?

I am closing as stale since this did not have much interest.

PR #5197, opened 2026-01-07
feat(database_observability.postgres): Add adaptive throttling mechanism to query_samples collector

gaantunes:

This throttling mechanism is designed to avoid an excessive number of samples for queries with a high execution rate. We have diminishing returns for every additional sample after a representative baseline is collected.

A StatStatementsRegistry is being created to run alongside the other collectors and query pg_stat_statements every minute, computing per-(queryid, datname) execution rates from consecutive call count deltas. It is then used by the `query_sample` collector.

This will be expanded in subsequent PRs to address the needs of `explain_plan` and `query_details` collectors, to avoid multiple queries to pg_stat_statements from each collector. Starts addressing https://github.com/grafana/grafana-dbo11y-app/issues/2126

Until two snapshots have been collected (~first minute after startup), all samples are emitted unconditionally.

The throttling is applied based on a logarithmic scale:

| Executions per minute | Effective interval |
|-----------------------|--------------------|
| ≤ 1                   | 1m (base)          |
| 2 – 10                | 2m                 |
| 11 – 100              | 3m                 |
| 101 – 1,000           | 4m                 |
| 1,001 – 10,000        | 5m                 |

Samples that are always exempt from throttling (no matter the rate):
idle in transaction state
Samples with active wait events
Samples with observed CPU time

### PR Checklist

- [x] Documentation added
- [x] Tests updated


clayton-cornell:
Docs look ok

clayton-cornell:
I wonder if it's worth adding a note somewhere in the doc stating/explaining why you'd want to disable throttling? Is this an option the users will understand just be seeing it available or is more context helpful?

gaantunes:
> I wonder if it's worth adding a note somewhere in the doc stating/explaining why you'd want to disable throttling? Is this an option the users will understand just be seeing it available or is more context helpful?

I will think about something.

matthewnolf:
@gaantunes Is this still needed?

gaantunes:
> @gaantunes Is this still needed?

I am closing as stale since this did not have much interest.

PR #5203, opened 2026-01-07
fix: Skip explain plans which lookup individual records and return no rows 

### LINKED ISSUE #4987: gracefully handle queries which do not return any rows when running explain plan
### Component(s)

database_observability.mysql

### Add a description

When a query is executed with a `WHERE` clause which does not return a row, the resulting explain plan is empty.

For example `select last_name from employees where emp_no = 1;` will not return any rows and the resulting in a native explain plan that looks like this.

```
{
  "query_block": {
    "select_id": 1,
    "message": "no matching row in const table"
  }
}
```

The same query with an employee number that exists, I.E. `select last_name from employees where emp_no = 10001;` will produce a useful explain plan.

The `QUERY_SAMPLE_TEXT` in `performance_schema.events_statements_summary_by_digest` will contain the most recent concrete query that was executed for a given digest, possibly resulting in inconsistent explain plans, sometimes containing explain details or being empty as shown above.

We need to detect if the query sample will result in any returned rows, and not send an empty explain plan in that case.

### Tip

rgeyer:
In MySQL, the query we're executing for explain plans gets an updated concrete query in `performance_schema` periodically.

For queries which look up a single row, by a primary key id, for a record which does not exist, no explain plan is created.

This created situations where a customer would sometimes see an explain plan, and sometimes not for a given query.

This PR simply refrains from logging the empty explain plan, such that the dbo11y UI should only find explain plan records with content for a given query.

Fixes #4987

### Notes to the Reviewer
This just debug logs the fact that we're skipping, and does not communicate with the user or the dbo11y UI in any way.

In the future, I'd like to add some additional telemetry which a user can opt-in to with their alloy config to have more visibility to the frequency with which individual queries are dropped for the various reasons.

### PR Checklist
- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
fix(database_observability): Skip explain plans which lookup individual records and return no rows (#5203)
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/component/database_observability/mysql/collector/explain_plans.go
+++ b/internal/component/database_observability/mysql/collector/explain_plans.go
@@ -692,8 +692,13 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 			continue
 		}
 
-		msgBlock, _, _, err := jsonparser.Get(redactedByteExplainPlanJSON, "query_block", "message")
-		if err == nil {
+		msgBlock, msgDatType, _, err := jsonparser.Get(redactedByteExplainPlanJSON, "query_block", "message")
+		if err != nil {
+			// An explain plan response typically will not have a message field if it is successful.
+			if msgDatType != jsonparser.NotExist {
+				level.Error(logger).Log("msg", "failed to parse explain plan json", "err", err)
+			}
+		} else {
 			if strings.Contains(string(msgBlock), "no matching row in const table") {
 				level.Debug(logger).Log("msg", "explain plan query resulted in no rows, skipping", "message", string(msgBlock))
 				continue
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -692,6 +692,14 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 			continue
 		}
 
+		msgBlock, _, _, err := jsonparser.Get(redactedByteExplainPlanJSON, "query_block", "message")
+		if err == nil {
```
cristiangreco: Just checking: if `err != nil` are we happy not doing anything and moving on?

cristiangreco (approved):

internal/component/database_observability/mysql/collector/explain_plans.go
rgeyer: I think yes. The most likely scenario is that it did not find the requested json node.

That said, this is probably not the most deterministic way to handle it. I should check for error, and empty string separately.

PR #5208, opened 2026-01-08
chore(pyroscope.ebpf): update ebpf-profiler to latest main

simonswine:
That should do it, let me know if you have more questions


blewis12:
heya @simonswine - there were some changes recently introduced to alloy that mean that replace directives are now generated in the go.mod file. In order to fix the currently failing `generate-module-dependencies` action, you'd need to update the replace directive [here](https://github.com/grafana/alloy/blob/main/dependency-replacements.yaml#L96-L98) and run `make generate-module-dependencies` to update the main go.mod file. Let me know if you encounter any issues there!

PR #5210, opened 2026-01-08
fix: Fix bugs when `loki.source.file` uses non-UTF-8 encoding

ptodev:
### Pull Request Details

E.g. for UTF-16, sometimes spaces would appear between characters due to encoding settings being lost after reopening a file.

Also, gauging whether to reopen a file wasn't working well - files were often reopened unnecessarily. That's because UTF-16 has a wider character set and it would confuse the tailer which thought that the file has shrunk, and therefore the tailer thought it must have rotated.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/file.go
+++ b/internal/component/loki/source/file/internal/tail/file.go
@@ -149,19 +149,30 @@ func (f *File) Stop() error {
 // wait blocks until a file event is detected (modification, truncation, or deletion).
 // Returns an error if the context is canceled or an unrecoverable error occurs.
 func (f *File) wait(partial bool) error {
-	offset, err := f.offset()
+	// Use file size instead of calculated offset for blockUntilEvent.
+	// This is important when using a decoder (e.g., UTF-16) because the offset()
+	// calculation mixes raw file bytes with decoded bytes, which are incompatible
+	// units and can cause incorrect comparisons.
+	fi, err := f.file.Stat()
 	if err != nil {
 		return err
 	}
+	fileSize := fi.Size()
 
-	event, err := blockUntilEvent(f.ctx, f.file, offset, f.cfg)
+	event, err := blockUntilEvent(f.ctx, f.file, fileSize, f.cfg)
 	switch event {
 	case eventModified:
 		level.Debug(f.logger).Log("msg", "file modified")
 		if partial {
 			// We need to reset to last successful offset because we consumed a partial line.
+			// TODO: Return error from Seek()?
 			f.file.Seek(f.lastOffset, io.SeekStart)
-			f.reader.Reset(f.file)
+			f.resetReader()
+		} else if f.cfg.Decoder != nil {
+			// Reset the reader when a decoder is configured. Once a decoder's underlying
+			// transform.Reader returns EOF, it won't read more data even if new content
+			// is appended. Resetting creates a fresh decoder that can read the new content.
+			f.resetReader()
 		}
 		return nil
 	case eventTruncated:
@@ -186,9 +197,11 @@ func (f *File) wait(partial bool) error {
 // The newline and any trailing carriage return (for Windows line endings) are stripped.
 func (f *File) readLine() (string, error) {
 	line, err := f.reader.ReadString('\n')
+
 	if err != nil {
 		return line, err
 	}
+
 	return strings.TrimRight(line, "\r\n"), err
 }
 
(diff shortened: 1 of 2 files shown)


internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -149,19 +149,30 @@ func (f *File) Stop() error {
...
+	fileSize := fi.Size()
 
-	event, err := blockUntilEvent(f.ctx, f.file, offset, f.cfg)
+	event, err := blockUntilEvent(f.ctx, f.file, fileSize, f.cfg)
 	switch event {
 	case eventModified:
 		level.Debug(f.logger).Log("msg", "file modified")
 		if partial {
 			// We need to reset to last successful offset because we consumed a partial line.
+			// TODO: Return error from Seek()?
```
blewis12, line 168: Does this TODO need to be addressed?

internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -298,13 +311,21 @@ func (f *File) reopen(truncated bool) error {
...
 
 	return backoff.Err()
 }
 
+// resetReader recreates the reader chain, preserving the decoder if one is configured.
+// This is necessary when seeking to a new position in the file, as the decoder may have
+// internal state that needs to be reset. Simply calling bufio.Reader.Reset() would
+// bypass the decoder and read raw bytes instead of decoded content.
+func (f *File) resetReader() {
+	f.reader = newReader(f.file, f.cfg)
```
blewis12, line 326: I imagine that this would add a bit of extra allocations since we're recreating the reader from new, but since this is not a hot path that runs thousands of times per second it shouldn't be an issue eh?

internal/component/loki/source/file/internal/tail/file.go
ptodev: It is not necessary to do this now. I don't want to break something in the current behaviour, since I don't understand the failure mechanism well enough and I don't have time to test it.

internal/component/loki/source/file/internal/tail/file.go
ptodev: Yes, I don't expect this to be called too frequently. Also, I don't see why memory allocations would be a big problem - it's just wrapping the reader with a decoder reader, which it should do in cases of special encoding anyway.

blewis12 (approved):
Thank you for looking into this 🙇

internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -149,19 +149,30 @@ func (f *File) Stop() error {
...
-	offset, err := f.offset()
+	// Use file size instead of calculated offset for blockUntilEvent.
+	// This is important when using a decoder (e.g., UTF-16) because the offset()
+	// calculation mixes raw file bytes with decoded bytes, which are incompatible
+	// units and can cause incorrect comparisons.
+	fi, err := f.file.Stat()
 	if err != nil {
 		return err
 	}
+	fileSize := fi.Size()
```
kalleep, line 160: I am not sure this is correct. We used offset because we are interested to get events from where we last read.

If new data has come in between us trying to read a line and calling wait we would miss that.

internal/component/loki/source/file/internal/tail/file.go
kalleep, line 168: It's fine to return an error here!

internal/component/loki/source/file/internal/tail/file.go
ptodev: Thank you, I think you're right 🤔 I'll see if I can get offset() to only use raw bytes.

williamhowf:
server_log:
‘2000’: user ‘743004’ updated: Accounts - 743004 - Company: ‘Biên Hòa122’ -> ‘Biên Hòa 1’

loki:
‘2000’: user ‘743004’ updated: Accounts - 743004 - Company: ‘Bin Ha122’ -> ‘Bin Ha 1’

Hi guys, i'm not sure whether this fix will solve the current issue that i face? server log file encoding utf-16le bom. 

version 1.12.0

kalleep:
> server_log: ‘2000’: user ‘743004’ updated: Accounts - 743004 - Company: ‘Biên Hòa122’ -> ‘Biên Hòa 1’
> 
> loki: ‘2000’: user ‘743004’ updated: Accounts - 743004 - Company: ‘Bin Ha122’ -> ‘Bin Ha 1’
> 
> Hi guys, i'm not sure whether this fix will solve the current issue that i face? server log file encoding utf-16le bom.
> 
> version 1.12.0

@williamhowf It's not fixing the issue. I am working on a alternative solution

ptodev:
Closing this in favour of #5259

PR #5213, opened 2026-01-08
fix: Fix kafka exporter instance argument

nilforoosh:

When there is only one URI in kafka-exporter, the instance parameter is ignored and there is a bug.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/static/integrations/kafka_exporter/kafka_exporter.go
+++ b/internal/static/integrations/kafka_exporter/kafka_exporter.go
@@ -154,7 +154,7 @@ func (c *Config) Name() string {
 // there is not exactly one Kafka node, the user must manually provide
 // their own value for instance key in the common config.
 func (c *Config) InstanceKey(_ string) (string, error) {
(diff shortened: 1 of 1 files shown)


internal/static/integrations/kafka_exporter/kafka_exporter.go
```diff
@@ -154,7 +154,7 @@ func (c *Config) Name() string {
 // there is not exactly one Kafka node, the user must manually provide
 // their own value for instance key in the common config.
 func (c *Config) InstanceKey(_ string) (string, error) {
-	if len(c.KafkaURIs) == 1 {
+	if c.Instance == "" && len(c.KafkaURIs) == 1 {
```
dehaansa, line 157: To make this a bit more readable, should we refactor to have a guard clause checking if Instance is set, and otherwise check the KafkaURIs?

```
if c.Instance != "" {
  return c.Instance, nil
}

//check KafkaURIs here
```

Issue #3007, opened 2025-03-15
RFC3164 syslog message without PRI are dropped / lost


gogglespisano:
### What's wrong?

loki.source.syslog drops messages in RFC3164 format that do not include the <nn> PRI field, logging it as an error.

The PRI field is not mandatory in RFC3164.  These messages should be parsed and the facility and severity level unset or set to default values.

### Steps to reproduce

Send a RFC3164 syslog message that does not include the PRI <nn> field.

### Configuration

```text
loki.source.syslog "XXX" {
  listener {
    address = "0.0.0.0:1234"
    protocol = "udp"
    syslog_format = "rfc3164"
    label_structured_data = true
    use_rfc5424_message = true
  }
}
```

### Logs

```text

```

gogglespisano:
#2275 could be a way to at least keep them from being dropped

dehaansa:
I'm interested in improving the component to support non-compliant syslog messages, but I do want to question where you are seeing that the priority is optional? In the [RFC](https://www.rfc-editor.org/rfc/rfc3164.html#section-4.1) it is not shown to be optional. Other servers may handle it better than Alloy does, but I believe "correct" spec compliance requires it.

gogglespisano:
Section 4.2 says all parts are RECOMMENDED and section 4.3.3 says what to do if PRI is not there.

Where are you seeing it is required?

dehaansa:
It's strange how the RFC defines that behavior for a relay (4.3.3), but not for a collector/syslog server. 4.2 on the other hand really just says that "hey, if it's on the right port it's syslog, I guess" so it's hard to argue with that.

gogglespisano:
I was kind of surprised when I first looked at it.  I was going to blame the syslog generator for a bad format until I read essentially everything goes.

PR #5216, opened 2026-01-08
feat(loki.source.syslog): Implement livedebugging support

x1unix:

This PR introduces live debugging support for `loki.source.syslog` component.

### Pull Request Details

Logged items:
- Error message and context.
- Incoming logs + original and mapped labels.

As as some users (including myself) [sometimes struggle](https://github.com/grafana/alloy/issues/3007) to properly configure `relabel_rules`, I've added capture of both original and mapped labels.

n/a

### Notes to the Reviewer

Besides livedebugging support, this PR also refactors some function calls to use struct instead of argument list due to a grown number of input parameters.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/syslog/livedebugging.go
+++ b/internal/component/loki/source/syslog/livedebugging.go
@@ -53,10 +53,6 @@ func (l liveDebuggingWriter) pushData(thunk func() string) {
 
 // OnNewMessage implements syslogtarget.DebugListener.
 func (l *liveDebuggingWriter) OnNewMessage(e syslogtarget.NewMessageDebugEvent) {
-	if !l.pub.IsActive(l.componentID) {
-		return
-	}
-
 	l.pushData(func() string {
 		sb := &strings.Builder{}
 		sb.Grow(1 << 11) // 2 KiB
--- a/internal/component/loki/source/syslog/syslog_test.go
+++ b/internal/component/loki/source/syslog/syslog_test.go
@@ -8,21 +8,20 @@ import (
 	"testing"
 	"time"
 
-	"github.com/grafana/alloy/internal/service/livedebugging"
 	"github.com/grafana/regexp"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	scrapeconfig "github.com/grafana/alloy/internal/component/loki/source/syslog/config"
-	"github.com/grafana/alloy/internal/featuregate"
-
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/component/common/loki"
 	alloy_relabel "github.com/grafana/alloy/internal/component/common/relabel"
+	scrapeconfig "github.com/grafana/alloy/internal/component/loki/source/syslog/config"
 	"github.com/grafana/alloy/internal/component/loki/source/syslog/internal/syslogtarget"
+	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/runtime/componenttest"
+	"github.com/grafana/alloy/internal/service/livedebugging"
 	"github.com/grafana/alloy/internal/util"
 )
 
--- a/internal/service/livedebugging/livedebugging.go
+++ b/internal/service/livedebugging/livedebugging.go
@@ -32,9 +32,6 @@ type CallbackManager interface {
 type DebugDataPublisher interface {
 	// Publish sends debugging data for a given componentID if a least one consumer is listening for debugging data for the given componentID.
 	PublishIfActive(data Data)
-
-	// IsActive returns whether debuging is enabled and there is at least one consumer listening for debugging data for the given component ID.
-	IsActive(cid ComponentID) bool
 }
 
 type liveDebugging struct {
@@ -53,18 +50,6 @@ func NewLiveDebugging() *liveDebugging {
 	return &liveDebugging{
 		callbacks: make(map[ComponentID]map[CallbackID]func(Data)),
 	}
-}
-
-func (s *liveDebugging) IsActive(cid ComponentID) bool {
-	s.loadMut.RLock()
-	defer s.loadMut.RUnlock()
-
-	if !s.enabled {
-		return false
-	}
-
-	callbacks, ok := s.callbacks[cid]
-	return ok && len(callbacks) > 0
 }
 
 func (s *liveDebugging) PublishIfActive(data Data) {


internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
```diff
@@ -173,6 +210,15 @@ func (t *SyslogTarget) handleMessageRFC5424(connLabels labels.Labels, msg *rfc54
 			m = fullMsg
 		}
 	}
+
+	t.dbgListener.OnNewMessage(NewMessageDebugEvent{
```
dehaansa, line 210: A little concerned about the potential memory ramifications of this change, have you benchmarked at all under load?

internal/component/loki/source/syslog/livedebugging.go
```diff
@@ -0,0 +1,108 @@
...
+	})
+}
+
+// OnNewMessage implements syslogtarget.DebugListener.
+func (l *liveDebuggingWriter) OnNewMessage(e syslogtarget.NewMessageDebugEvent) {
+	l.pushData(func() string {
+		sb := &strings.Builder{}
+		sb.Grow(1 << 11) // 2 KiB
+		fmt.Fprintf(
+			sb, "[IN] New Log: Format=%q DT=%q\n  Message: %q\n",
```
dehaansa:
```suggestion
			sb, "[IN] New Log: Format=%q ts=%q\n  Message: %q\n",
```

internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
```diff
@@ -96,9 +129,12 @@ func NewSyslogTarget(metrics *Metrics, logger log.Logger, handler loki.EntryHand
 func (t *SyslogTarget) handleMessageError(err error) {
 	var ne net.Error
 	if errors.As(err, &ne) && ne.Timeout() {
+		t.dbgListener.OnError("connection timed out", ne)
```
kgeckhart: Are there other instance of us pushing non-dataflow info through live debugging? I feel like most usages I've seen involved data flowing rather than events like this and the one below on line 137. 

Sticking to just data would allow you to drop the debug listener all together to use DebugDataPublisher https://github.com/grafana/alloy/blob/10b0d8c13bc9157ef4b41b4162d34db2cc551623/internal/service/livedebugging/livedebugging.go#L29-L33 with lazy evaluation of the payloads (Sam's comment).

internal/component/loki/source/syslog/livedebugging.go
```diff
@@ -0,0 +1,108 @@
...
+func (l *liveDebuggingWriter) OnNewMessage(e syslogtarget.NewMessageDebugEvent) {
+	l.pushData(func() string {
+		sb := &strings.Builder{}
+		sb.Grow(1 << 11) // 2 KiB
+		fmt.Fprintf(
+			sb, "[IN] New Log: Format=%q ts=%q\n  Message: %q\n",
+			e.Format, e.Timestamp.Format(time.RFC3339), e.Message,
+		)
+
+		// print mapped and original labels to simplify relabel configuration debugging
```
kgeckhart, line 64: I think re-using the existing format loki relabel uses for before/after would be fine here, https://github.com/grafana/alloy/blob/2a28ee4eef6535f7299ba2116a6345ed1123c258/internal/component/loki/relabel/relabel.go#L140

internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
x1unix: I've added locally a small test that sends 500 syslog messages to test memory consumption using `runtime.ReadMemStats` with both livedebugging enabled and disabled.

Also I've added an extra check to prevent creating a thunk closure if livedebug is disabled.\
This check has to be done every time as live debugging can be toggled in runtime and event subscription happens using a dedicated HTTP handler (I assume that web ui used it).

Here are memory measurement results:

<details> <summary>LiveDebugging enabled:</summary>

```
TotalAlloc: 6177496
Allocs:     2037608
Mallocs:    54746
```

</details>

<details> <summary>LiveDebugging disabled:</summary>

```
TotalAlloc: 2315864
Allocs:     2315864
Mallocs:    36194
```

</details>

Indeed, enabling live debugging increases memory usage.
When live debugging is disabled - memory usage is mostly the same as before.

internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
x1unix: @kgeckhart I've checked what else is using live debugging - it's used only by metrics, except `loki.process` which just dumps metadata as JSON:

image

The reason I've added an extra publisher is to display captured and ignored labels in human readable form to simplify relabel config debugging.

Also I've added an extra logic to avoid capturing data if live debugging is disabled globally and/or current component is not inspected.

internal/component/loki/source/syslog/livedebugging.go
x1unix: @kgeckhart I didn't know that those types implement `fmt.Stringer`, thanks for noticing that.

Fixed in https://github.com/grafana/alloy/pull/5216/changes/dfc6ce7985e839e3290b1adac1fb5bf67d6fd6b3

internal/component/loki/source/syslog/syslog_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/alloy/internal/service/livedebugging"
```
kalleep: nit: move this to other alloy import group

internal/service/livedebugging/livedebugging.go
```diff
@@ -30,15 +32,21 @@ type CallbackManager interface {
 type DebugDataPublisher interface {
 	// Publish sends debugging data for a given componentID if a least one consumer is listening for debugging data for the given componentID.
 	PublishIfActive(data Data)
+
+	// IsActive returns whether debuging is enabled and there is at least one consumer listening for debugging data for the given component ID.
+	IsActive(cid ComponentID) bool
```
kalleep: I am not convinced we need to add this function. We use DataFunc so allocation will only happen if it's enabled and we have an active listener when we call `PublishIfActive`. If we use this function we would take the lock twice when it's active right?

x1unix: @kalleep removed in https://github.com/grafana/alloy/pull/5216/changes/8d21a3e5f5253e3efaf216ab4a05ce4bc77ab716

internal/component/loki/source/syslog/syslog_test.go
x1unix: @kalleep `goimports` messes up imports when there was an import group already.

Fixed in https://github.com/grafana/alloy/pull/5216/changes/8d21a3e5f5253e3efaf216ab4a05ce4bc77ab716

PR #5222, opened 2026-01-09
feat: Database_observability: add health check collector for postgres component

matthewnolf:
This change adds a new health_check collector to the `database_observability.postgres` component, responsible for exposing diagnostic status information. This information can be used to diagnose misconfiguration issues.

This is a followup to this PR, which first introduced this for [MySQL](https://github.com/grafana/alloy/pull/5115)

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/health_check.go
+++ b/internal/component/database_observability/postgres/collector/health_check.go
@@ -1,0 +1,207 @@
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/go-kit/log"
+	"go.uber.org/atomic"
+
+	"github.com/grafana/alloy/internal/build"
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/component/database_observability"
+	"github.com/grafana/alloy/internal/runtime/logging"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const (
+	HealthCheckCollector = "health_check"
+	OP_HEALTH_STATUS     = "health_status"
+)
+
+type HealthCheckArguments struct {
+	DB              *sql.DB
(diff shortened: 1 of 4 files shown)


internal/component/database_observability/postgres/collector/health_check.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if rows.Next() {
+		r.result = true
+	}
+
+	if err := rows.Err(); err != nil {
+		r.err = fmt.Errorf("iterate pg_stat_statements: %w", err)
+	}
+
+	r.result = true
```
cristiangreco: What is the meaning of result here, given it's always set to true (while above we check for `rows.Next()`)?

matthewnolf: You're right - we don't need this line

matthewnolf: Fixed in 2bfaeb0bcd28566581aa804465dd4320e7abb4d3

internal/component/database_observability/postgres/collector/health_check.go
```diff
@@ -0,0 +1,205 @@
...
+	// Always succeeds; returns the version string embedded at build time.
+	r.result = true
+	r.value = build.Version
+	return r
+}
+
+// checkPgStatStatementsEnabled verifies that the pg_stat_statements extension is enabled.
+func checkPgStatStatementsEnabled(ctx context.Context, db *sql.DB) healthCheckResult {
+	r := healthCheckResult{name: "PgStatStatementsEnabled"}
+	const q = `SELECT * FROM pg_extension WHERE extname = 'pg_stat_statements'`
```
cristiangreco, line 142: Just to note that our docs says to enable the extension on every database, but I appreciate it might worth starting simple and just running the check on the initial connection database.

cristiangreco (approved):

PR #5233, opened 2026-01-12
docs: Add documentation about replacement generation to the contribution guide

blewis12:
This PR introduces some clearer documentation on how `replace` directives are generated in the alloy project. I realized that although I introduced a README to the tool itself, this wasn't made clear in the actual contributing guide


thampiotr (commented):
If there is something duplicated in the other README.md, maybe we can point from there to contributing.md? Just thought we want to reduce duplication.

blewis12:
> If there is something duplicated in the other README.md, maybe we can point from there to contributing.md? Just thought we want to reduce duplication.

yeah that's a fair point, let me revisit and see if I can make the contributing.md side a bit shorter and summarizing the idea - then point to the tool readme that is more of a deep dive with clear steps

Issue #5234, opened 2026-01-12
add e2e test scenarios for loki related pipelines.


kalleep:
I am working on a lot of improvements / fixes to our loki pipelines within alloy.

To increase confidence in the changes being made we should add e2e test for different scenarios. Starting with going from loki.source.file to writing them into loki. This should cover file rotations etc.

We should also extend to cover more scenarios as we go.

kalleep:
Closing this in favor of https://github.com/grafana/alloy/issues/4953

Part of that is to increase test coverage

PR #5237, opened 2026-01-12
chore: Fix broken loki integration tests

kalleep:
The integration tests we had for loki code did not work. [AssertLogsPresent](https://github.com/grafana/alloy/blob/6bd93eff3a1b1588f1a4dc32a101b1080ed16369/internal/cmd/integration-tests/common/logs_assert.go#L29C6-L29C23) and [AssertLogsMissing](https://github.com/grafana/alloy/blob/6bd93eff3a1b1588f1a4dc32a101b1080ed16369/internal/cmd/integration-tests/common/logs_assert.go#L52C6-L52C23) always returned when we had no result so we never verified that we actually got logs in loki.

On top of this there was two more issues:
1. loki-enrich - had a timing issue where logs was forwarded before targets was propagated to component, this could be fixed with a sleep in the test or hardcode the discovery data
2. loki-file - Result from loki is split up in streams so we need to be able to check different entry counts of different stream. Right now I added a function that finds the first stream containing all the labels we are interested in.

part of: #5234

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/cmd/integration-tests/tests/loki-file/loki_file_test.go
+++ b/internal/cmd/integration-tests/tests/loki-file/loki_file_test.go
@@ -23,9 +23,9 @@ func TestReadLogFile(t *testing.T) {
 		},
 		common.ExpectedLogResult{
 			Labels: map[string]string{
-				"detected_level": "test",
+				"detected_level": "error",
 			},
-			EntryCount: 1,
+			EntryCount: 2,
 		},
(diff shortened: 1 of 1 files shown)


kgeckhart (approved):
Great catch!

thampiotr (approved):

PR #5239, opened 2026-01-12
ci: Remove workflow that updates make docs procedure

jdbaldry:
This workflow was useful when the procedure changed frequently but has become a burden as it has sprawled across repositories.

It can fail without obvious notfication and requires more permissions than I'd like to perform the task.

It's replaced by a centralized script that updates the procedure in a push fashion on an ad-hoc basis.


jdbaldry:
@copilot resolve the merge conflicts in this pull request

clayton-cornell (approved):
I don't know how to test this... but it seems reasonable and an improvement in CI is always a good thing.

PR #5259, opened 2026-01-14
fix: Issues when reading files using non UTF-8 encoding in loki.source.file 

kalleep:
Previously we wrapped our file with the provided decoder and then consumed it using bufio.Reader. This had several problems:

1. We did not report correct offset into the file
2. If you used UTF-16 and we had a stored offset we could not determine the correct encoding (UTF-16LE or UTF-16BE).

### Pull Request Details
To correctly handle file positions we need to operate on the raw bytes in the reader and (optionally) perform the decoding after. This makes sure we always record the correct offset.

Because we now operate on raw bytes we needed to change how we terminated newlines and carriage return. The decoder can translate it for us so we know what we should be looking for. We now also do additional buffering of non terminated lines and do not have to perform re-seeks after modified event when we consumed a partial line.

We are now also able to automatically detect UTF-8, UTF-16LE and UTF-16BE encodings and correctly decode it if there is a BOM present for the file.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/bom.go
+++ b/internal/component/loki/source/file/internal/tail/bom.go
@@ -2,7 +2,6 @@ package tail
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"os"
 
@@ -84,7 +83,6 @@ func resolveEncodingFromBOM(bomBytes []byte, originalEnc encoding.Encoding) enco
 		// UTF-16 BE BOM detected - return encoding with IgnoreBOM since we skip it
 		return unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)
 	case bytes.HasPrefix(bomBytes, bomUTF16LE):
-		fmt.Println("Detected UTF-16LE")
 		// UTF-16 LE BOM detected - return encoding with IgnoreBOM since we skip it
 		return unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)
 	default:


internal/component/loki/source/file/internal/tail/bom.go
```diff
@@ -0,0 +1,89 @@
...
+	default:
+		return 0
+	}
+}
+
+// resolveEncodingFromBOM takes the BOM bytes and the original encoding,
+// and returns the resolved encoding. If a UTF-16 BOM is detected, it returns
+// an encoding with the correct endianness and IgnoreBOM mode.
+// Otherwise, it returns the original encoding.
+func resolveEncodingFromBOM(bomBytes []byte, originalEnc encoding.Encoding) encoding.Encoding {
```
ptodev, line 73: If I understand correctly, you don't have to set the `encoding` parameter in the config now if there is a BOM? This would be very nice, although it is technically a breaking change for users who currently set the encoding explicitly. It would be good to update the docs with a recommendation for when to use the `encoding` argument.

Also, it looks like the new code will read the BOM and adjust the encoding even when the offset is not 0, which is very good!

internal/component/loki/source/file/internal/tail/file_test.go
ptodev, line 1: I think the new code will be able to handle cases where a single component tails two files with different encodings - A UTF-8 file without, and a UTF-16 file with a BOM. This was previously unsupported? Do you think it's worth adding a test for it and documenting this?

kalleep: >I think the new code will be able to handle cases where a single component tails two files with different encodings - A UTF-8 file without, and a UTF-16 file with a BOM. This was previously unsupported? Do you think it's worth adding a test for it and documenting this?

Not really, well it would work for UTF-16 only. We could adjust this to make it work for other encodings as well e.g. if we detect UTF-8.

Would it be ok to do this in a followup?

internal/component/loki/source/file/internal/tail/bom.go
kalleep: >Also, it looks like the new code will read the BOM and adjust the encoding even when the offset is not 0, which is very good!

Yes, if u set UTF-16 we need to handle that case. But as I mentioned in the other comment the only encoding we adjust for is UTF-16. Would require a lot more work do to this automatically

internal/component/loki/source/file/internal/tail/file_test.go
ptodev: Sure! I think UTF-8 and UTF-16 are by far the most widely used encodings, so as long as we support those two we would make things easier for logs of users. Then for other encodings we could add them on a case by case basis if a user requests it.

kalleep: Great will do that in a followup then :)

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -48,8 +48,8 @@ You can use the following arguments with `loki.source.file`:
 | `on_positions_file_error` | `string`             | How to handle a corrupt positions file entry for a given file. | `"restart_from_beginning"` | no       |
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
-The `encoding` argument must be a valid [IANA encoding][] name.
-If not set, it defaults to UTF-8.
+The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} is able to automatically change
+the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
```
ptodev, line 52:
```suggestion
the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
The BOM will be taken into account even if Alloy resumes tailing a file from the middle of the file. This can happen after Alloy is restarted.
```

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -48,8 +48,9 @@ You can use the following arguments with `loki.source.file`:
 | `on_positions_file_error` | `string`             | How to handle a corrupt positions file entry for a given file. | `"restart_from_beginning"` | no       |
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
-The `encoding` argument must be a valid [IANA encoding][] name.
-If not set, it defaults to UTF-8.
+The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} is able to automatically change
```
```suggestion
The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} can automatically change
```

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -48,8 +48,9 @@ You can use the following arguments with `loki.source.file`:
...
+the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
+The BOM will be taken into account even if Alloy resumes tailing a file from the middle of the file. This can happen after Alloy is restarted.
```
```suggestion
The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
```

internal/component/loki/source/file/internal/tail/bom.go
```diff
@@ -76,10 +77,14 @@ func resolveEncodingFromBOM(bomBytes []byte, originalEnc encoding.Encoding) enco
...
 
 	switch {
+	case bytes.HasPrefix(bomBytes, bomUTF8):
+		// UTF-8 BOM detected - return encoding
+		return encoding.Nop
 	case bytes.HasPrefix(bomBytes, bomUTF16BE):
 		// UTF-16 BE BOM detected - return encoding with IgnoreBOM since we skip it
 		return unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)
 	case bytes.HasPrefix(bomBytes, bomUTF16LE):
+		fmt.Println("Detected UTF-16LE")
```
ptodev:
```suggestion
```

ptodev (approved):
LGTM, thanks! There's just a print statement we need to remove.

PR #5266, opened 2026-01-15
docs: Add explanation of `sync_period`

chu121su12:

I am dealing with old system that generates log file named `<service>_<month>`. The log file would then be replaced every year, so I needded to make sure a certain alloy log scraping behavior. I found a promtail behavior (also with simliar situation as me) in the [community discussion](https://community.grafana.com/t/a-way-of-deleting-log-files-after-promtail-has-scraped-and-forwarded-them-to-loki/60626) and I think it is relevant to be included in the docs ***IF*** the behavior is the same in alloy.
> You actually won’t need to remove the entries from the positions file since promtail is already aware of files being removed (and will cleanup the positions file every -positions.sync-period).

~The description is also moved from `loki` to the more general `monitor` for easier discovery.~

### Notes to the Reviewer

Ensure promtail's `-positions.sync-period` is the same with alloy.

### PR Checklist

- [x] Documentation moved & added

### DIFF
--- a/docs/sources/monitor/monitor-logs-from-file.md
+++ b/docs/sources/monitor/monitor-logs-from-file.md
@@ -114,7 +114,7 @@ The [`local.file_match`][local.file_match] component discovers files on the loca
 In this example, the component requires the following arguments:
 
 * `path_targets`: Targets to expand. Looks for glob patterns on the `__path__` key.
-* `sync_period`: How often to sync the filesystem and targets. When files appear or disappear, the list of targets is automatically updated based on this configuration. When disappeared files got synced, the corresponding stored position is cleaned up.
+* `sync_period`: How often to sync the filesystem and targets. When files are added or removed, the list of targets is automatically updated based on this configuration. When files are removed, the stored position is removed in the next filesystem sync.
 
 ```alloy
 local.file_match "local_files" {


docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -219,7 +219,6 @@ loki.write "local" {
 ### File globbing with `file_match` (recommended)
 
 This example collects log entries from files matching the `*.log` pattern using the built-in `file_match` block.
-When files appear or disappear, the list of targets is automatically updated based on the `sync_period` configuration.
```
kalleep, line 222: I don't see any reason to remove it from here

docs/sources/monitor/monitor-logs-from-file.md
```diff
@@ -114,7 +114,7 @@ The [`local.file_match`][local.file_match] component discovers files on the loca
 In this example, the component requires the following arguments:
 
 * `path_targets`: Targets to expand. Looks for glob patterns on the `__path__` key.
-* `sync_period`: How often to sync the filesystem and targets.
+* `sync_period`: How often to sync the filesystem and targets. When files appear or disappear, the list of targets is automatically updated based on this configuration. When disappeared files got synced, the corresponding stored position is cleaned up.
```
clayton-cornell: Could you expand on what you mean by this `When disappeared files got synced, the corresponding stored position is cleaned up.`? 

I'm trying to understand what you're trying to clarify here. Is there some extra step done in the sync when a file is removed and a new sync runs?

kalleep: I can chime in here. When we are reading files we also store the offset after the last consumed line in a position file.

If a target e.g. a file is removed and such we don't have a need to track it so the position is removed.

docs/sources/reference/components/loki/loki.source.file.md
chu121su12: reverted

docs/sources/monitor/monitor-logs-from-file.md
chu121su12: The removal can be done manually, via logrotate, container/vm restart, cron, or some script contraption (like the OP does in the linked post). If the wording needs to be elaborated, maybe something like this:

> When files disappeared (removed manually, by scheduled task, or some other means) and the sync rerun, the corresponding stored position is cleaned up.

clayton-cornell: Maybe something like this then?

```suggestion
* `sync_period`: How often to sync the filesystem and targets. When files are added or removed, the list of targets is automatically updated based on this configuration. When files are removed, the stored position is removed in the next filesystem sync.
```

Issue #515, opened 2022-11-29
Flow UI: Page not very responsive while loading large component


xiyu95:
**Description**
Component detail page with large amount of exports values takes a long time to load due to large amount of returning data (~20 - 30 seconds loading time)
Sometimes this will cause the entire page to crash
Page reaction on user action is very slow on larger data size

josh-thisisbud:
Experiencing this issue at the moment

Would appreciate some pagination of the lists as this seems to be the main thing causing us grief

Daniel-Vaz:
After digging a bit further, I believe that what's constituting the biggest chunk of time consumption in the rendering of the results is the `SyntaxHighlighter` - [Source](https://github.com/grafana/alloy/blob/bb508af1660d75fae0ac9c4a33c5d300ac87f99d/internal/web/ui/src/features/component/ComponentBody.tsx#L28)

Building a custom image while just changing this to something simpler like `<code></code>` Tags I can see a really big improvement in the total amount of time in the page rendering:

![image](https://github.com/grafana/alloy/assets/29305556/449908ab-c27b-4677-b10a-b6036b0bb194)

rfratto:
Agreed that the issue is primarily the syntax highlighting. 

A good short term fix might be disabling syntax highlighting if the payload is longer than some reasonable size (maybe 1KB). Alternatively we might be able to explore lazily loading the syntax so it doens't block the first render.

Daniel-Vaz:
Can't a similar approach be used, like `alloyStringify`, or even extend its functionality ?

rfratto:
I'm not sure what you mean, what would using `alloyStringify` mean in this context?

akevdmeer:
I believe that what @Daniel-Vaz means is that the current code highlights the result of that alloyStringify() which requires parsing as Javascript. But there is no need to parse: alloyStringify() knows the structure of what it is generating and the highlighting could be done in the same pass efficiently. Basically to fuse this highlighting with the stringifying.

rfratto:
Ah, thanks! Yeah, it seems possible that we could have a new function to produce the highlighted DOM directly. It's worth some investigation for sure.

jharvey10:
Hey there! Fully realize this is an old issue, but we did "resolve" this by disabling syntax highlighting in the UI until we can find a more scalable approach to it.

In the coming months, the underlying framework for the UI will be upgraded, which may present some opportunities to reintroduce this functionality in a more performant way.

PR #5268, opened 2026-01-15
fix(ui): Large arguments are downloaded as files instead of rendered

### LINKED ISSUE #1121: Alloy UI not displaying targets
### What's wrong?

Screenshot 2024-06-24 at 17 58 03

Alloy UI freezes and does not display the information.
I suspect that this is related to CPU/Mem. We are using the default configurations for the resource allocation and deployed Alloy as Deamonsets to EKS cluster.

The number of targets is relatively small (250 Pods). 

### Steps to reproduce

Install Grafana Helm Chart `v0.3.2` with the `values.yaml` to EKS cluster version `v1.27`. 

### System information

EKS v1.27

### Software version

Grafana Alloy v0.3.2

### Configuration

```text
# -- Overrides the chart's name. Used to change the infix in the resource names.
nameOverride: null

# -- Overrides the chart's computed fullname. Used to change the full prefix of
# resource names.
fullnameOverride: null

## Global properties for image pulling override the values defined under `image.registry` and `configReloader.image.registry`.
## If you want to override only one image registry, use the specific fields but if you want to override them all, use `global.image.registry`
global:
  image:
    # -- Global image registry to use if it needs to be overriden for some specific use cases (e.g local registries, custom images, ...)
    registry: ""

    # -- Optional set of global image pull secrets.
    pullSecrets: []

  # -- Security context to apply to the Grafana Alloy pod.
  podSecurityContext: {}

crds:
  # -- Whether to install CRDs for monitoring.
  create: true

## Various Alloy settings. For backwards compatibility with the grafana-agent
## chart, this field may also be called "agent". Naming this field "agent" is
## deprecated and will be removed in a future release.

alloy:
  configMap:
    # -- Create a new ConfigMap for the config file.
    create: true
    # -- Content to assign to the new ConfigMap.  This is passed into `tpl` allowing for templating from values.
    content: |-
      // config.alloy
      // Reference: https://grafana.com/docs/alloy/latest/reference/components




      // Meta Monitoring Logs - https://grafana.com/docs/alloy/latest/tasks/metamonitoring
      logging {
        level    = "info"
        format   = "logfmt"
        write_to = [loki.write.default.receiver]
      }




      // Meta Monitoring Metrics - https://grafana.com/docs/alloy/latest/tasks/metamonitoring
      prometheus.exporter.self "alloy_meta_monitoring" {
      }





      // Kubernetes Pod Discovery
      discovery.kubernetes "pods" {
        role = "pod"
        enable_http2 = false
      }

      // Kubernetes Pod Targets Relabel Config
      discovery.relabel "logs" {

        targets = discovery.kubernetes.pods.targets

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_namespace"]
          target_label  = "namespace"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_name"]
          target_label  = "pod"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_container_name"]
          target_label  = "container"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_label_app_kubernetes_io_name"]
          target_label  = "app"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_node_name"]
          target_label  = "node"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_container_image"]
          target_label  = "container_image"
        }

        rule {
          action = "replace"
          source_labels = ["__meta_kubernetes_pod_ready"]
          target_label  = "pod_status"
        }
      }





      // Loki Source
      loki.source.kubernetes "pods" {
        targets    = discovery.kubernetes.pods.targets
      }





      // Kubernetes Events
      loki.source.kubernetes_events "k8s_events" {
        log_format = "logfmt"
      }





      // Loki Write Config
      loki.write "default" {
        endpoint {
          name = "Loki"
          url = "http://grafana-loki-gateway/loki/api/v1/push"
          basic_auth {
            username = "grafana_cloud"
            password = "grafana_cloud_test_password"
          }
          tenant_id = "apps"
        }
      }





    # -- Name of existing ConfigMap to use. Used when create is false.
    name: null
    # -- Key in ConfigMap to get config from.
    key: null

  clustering:
    # -- Deploy Alloy in a cluster to allow for load distribution.
    enabled: true

    # -- Name for the port used for clustering, useful if running inside an Istio Mesh
    portName: http

  # -- Minimum stability level of components and behavior to enable. Must be
  # one of "experimental", "public-preview", or "generally-available".
  stabilityLevel: "public-preview"

  # -- Path to where Grafana Alloy stores data (for example, the Write-Ahead Log).
  # By default, data is lost between reboots.
  storagePath: /tmp/alloy

  # -- Address to listen for traffic on. 0.0.0.0 exposes the UI to other
  # containers.
  listenAddr: 0.0.0.0

  # -- Port to listen for traffic on.
  listenPort: 12345

  # -- Scheme is needed for readiness probes. If enabling tls in your configs, set to "HTTPS"
  listenScheme: HTTP

  # --  Base path where the UI is exposed.
  uiPathPrefix: /

  # -- Enables sending Grafana Labs anonymous usage stats to help improve Grafana
  # Alloy.
  enableReporting: true

  # -- Extra environment variables to pass to the Alloy container.
  extraEnv: []

  # -- Maps all the keys on a ConfigMap or Secret as environment variables. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#envfromsource-v1-core
  envFrom: []

  # -- Extra args to pass to `alloy run`: https://grafana.com/docs/alloy/latest/reference/cli/run/
  extraArgs: []

  # -- Extra ports to expose on the Alloy container.
  extraPorts: []
  # - name: "faro"
  #   port: 12347
  #   targetPort: 12347
  #   protocol: "TCP"

  mounts:
    # -- Mount /var/log from the host into the container for log collection.
    varlog: false
    # -- Mount /var/lib/docker/containers from the host into the container for log
    # collection.
    dockercontainers: false

    # -- Extra volume mounts to add into the Grafana Alloy container. Does not
    # affect the watch container.
    extra: []

  # -- Security context to apply to the Grafana Alloy container.
  securityContext: {}

  # -- Resource requests and limits to apply to the Grafana Alloy container.
  resources: {}

image:
  # -- Grafana Alloy image registry (defaults to docker.io)
  registry: "docker.io"
  # -- Grafana Alloy image repository.
  repository: grafana/alloy
  # -- (string) Grafana Alloy image tag. When empty, the Chart's appVersion is
  # used.
  tag: null
  # -- Grafana Alloy image's SHA256 digest (either in format "sha256:XYZ" or "XYZ"). When set, will override `image.tag`.
  digest: null
  # -- Grafana Alloy image pull policy.
  pullPolicy: IfNotPresent
  # -- Optional set of image pull secrets.
  pullSecrets: []

rbac:
  # -- Whether to create RBAC resources for Alloy.
  create: true

serviceAccount:
  # -- Whether to create a service account for the Grafana Alloy deployment.
  create: true
  # -- Additional labels to add to the created service account.
  additionalLabels: {}
  # -- Annotations to add to the created service account.
  annotations: {}
  # -- The name of the existing service account to use when
  # serviceAccount.create is false.
  name: null

# Options for the extra controller used for config reloading.
configReloader:
  # -- Enables automatically reloading when the Alloy config changes.
  enabled: true
  image:
    # -- Config reloader image registry (defaults to docker.io)
    registry: "ghcr.io"
    # -- Repository to get config reloader image from.
    repository: jimmidyson/configmap-reload
    # -- Tag of image to use for config reloading.
    tag: v0.12.0
    # -- SHA256 digest of image to use for config reloading (either in format "sha256:XYZ" or "XYZ"). When set, will override `configReloader.image.tag`
    digest: ""
  # -- Override the args passed to the container.
  customArgs: []
  # -- Resource requests and limits to apply to the config reloader container.
  resources:
    requests:
      cpu: "1m"
      memory: "5Mi"
  # -- Security context to apply to the Grafana configReloader container.
  securityContext: {}

controller:
  # -- Type of controller to use for deploying Grafana Alloy in the cluster.
  # Must be one of 'daemonset', 'deployment', or 'statefulset'.
  type: 'daemonset'

  # -- Number of pods to deploy. Ignored when controller.type is 'daemonset'.
  replicas: 1

  # -- Annotations to add to controller.
  extraAnnotations: {}

  # -- Whether to deploy pods in parallel. Only used when controller.type is
  # 'statefulset'.
  parallelRollout: true

  # -- Configures Pods to use the host network. When set to true, the ports that will be used must be specified.
  hostNetwork: false

  # -- Configures Pods to use the host PID namespace.
  hostPID: false

  # -- Configures the DNS policy for the pod. https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
  dnsPolicy: ClusterFirst

  # -- Update strategy for updating deployed Pods.
  updateStrategy: {}

  # -- nodeSelector to apply to Grafana Alloy pods.
  nodeSelector: {}

  # -- Tolerations to apply to Grafana Alloy pods.
  tolerations: []

  # -- Topology Spread Constraints to apply to Grafana Alloy pods.
  topologySpreadConstraints: []

  # -- priorityClassName to apply to Grafana Alloy pods.
  priorityClassName: ''

  # -- Extra pod annotations to add.
  podAnnotations: {}

  # -- Extra pod labels to add.
  podLabels: {}

  # -- Whether to enable automatic deletion of stale PVCs due to a scale down operation, when controller.type is 'statefulset'.
  enableStatefulSetAutoDeletePVC: false

  autoscaling:
    # -- Creates a HorizontalPodAutoscaler for controller type deployment.
    enabled: false
    # -- The lower limit for the number of replicas to which the autoscaler can scale down.
    minReplicas: 1
    # -- The upper limit for the number of replicas to which the autoscaler can scale up.
    maxReplicas: 5
    # -- Average CPU utilization across all relevant pods, a percentage of the requested value of the resource for the pods. Setting `targetCPUUtilizationPercentage` to 0 will disable CPU scaling.
    targetCPUUtilizationPercentage: 0
    # -- Average Memory utilization across all relevant pods, a percentage of the requested value of the resource for the pods. Setting `targetMemoryUtilizationPercentage` to 0 will disable Memory scaling.
    targetMemoryUtilizationPercentage: 80

    scaleDown:
      # -- List of policies to determine the scale-down behavior.
      policies: []
        # - type: Pods
        #   value: 4
        #   periodSeconds: 60
      # -- Determines which of the provided scaling-down policies to apply if multiple are specified.
      selectPolicy: Max
      # -- The duration that the autoscaling mechanism should look back on to make decisions about scaling down.
      stabilizationWindowSeconds: 300

    scaleUp:
      # -- List of policies to determine the scale-up behavior.
      policies: []
        # - type: Pods
        #   value: 4
        #   periodSeconds: 60
      # -- Determines which of the provided scaling-up policies to apply if multiple are specified.
      selectPolicy: Max
      # -- The duration that the autoscaling mechanism should look back on to make decisions about scaling up.
      stabilizationWindowSeconds: 0

  # -- Affinity configuration for pods.
  affinity: {}

  volumes:
    # -- Extra volumes to add to the Grafana Alloy pod.
    extra: []

  # -- volumeClaimTemplates to add when controller.type is 'statefulset'.
  volumeClaimTemplates: []

  ## -- Additional init containers to run.
  ## ref: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
  ##
  initContainers: []

  # -- Additional containers to run alongside the Alloy container and initContainers.
  extraContainers: []

service:
  # -- Creates a Service for the controller's pods.
  enabled: true
  # -- Service type
  type: ClusterIP
  # -- NodePort port. Only takes effect when `service.type: NodePort`
  nodePort: 31128
  # -- Cluster IP, can be set to None, empty "" or an IP address
  clusterIP: ''
  # -- Value for internal traffic policy. 'Cluster' or 'Local'
  internalTrafficPolicy: Cluster
  annotations: {}
    # cloud.google.com/load-balancer-type: Internal

serviceMonitor:
  enabled: false
  # -- Additional labels for the service monitor.
  additionalLabels: {}
  # -- Scrape interval. If not set, the Prometheus default scrape interval is used.
  interval: ""
  # -- MetricRelabelConfigs to apply to samples after scraping, but before ingestion.
  # ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#relabelconfig
  metricRelabelings: []
  # - action: keep
  #   regex: 'kube_(daemonset|deployment|pod|namespace|node|statefulset).+'
  #   sourceLabels: [__name__]

  # -- Customize tls parameters for the service monitor
  tlsConfig: {}

  # -- RelabelConfigs to apply to samples before scraping
  # ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#relabelconfig
  relabelings: []
  # - sourceLabels: [__meta_kubernetes_pod_node_name]
  #   separator: ;
  #   regex: ^(.*)$
  #   targetLabel: nodename
  #   replacement: $1
  #   action: replace
ingress:
  # -- Enables ingress for Alloy (Faro port)
  enabled: false
  # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
  # See https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress
  # ingressClassName: nginx
  # Values can be templated
  annotations:
    {}
    # kubernetes.io/ingress.class: nginx
    # kubernetes.io/tls-acme: "true"
  labels: {}
  path: /
  faroPort: 12347

  # pathType is only for k8s >= 1.1=
  pathType: Prefix

  hosts:
    - chart-example.local
  ## Extra paths to prepend to every host configuration. This is useful when working with annotation based services.
  extraPaths: []
  # - path: /*
  #   backend:
  #     serviceName: ssl-redirect
  #     servicePort: use-annotation
  ## Or for k8s > 1.19
  # - path: /*
  #   pathType: Prefix
  #   backend:
  #     service:
  #       name: ssl-redirect
  #       port:
  #         name: use-annotation

  tls: []
  #  - secretName: chart-example-tls
  #    hosts:
  #      - chart-example.local
```

### Logs

_No response_

ptodev:
### Pull Request Details

The UI is not really usable when large amounts of text are displayed. An earlier changed mitigated this partially by disabling json highlighting, but in practice if a blob of text is large enough it'd still cause the UI to freeze. It's also hard to copy paste it and it's not possible to view it in full.

This PR will change the component-specific UI pages to display a download link for arguments whose values contain more than 50 000 characters.

https://github.com/user-attachments/assets/cf58658c-f90e-4c31-820c-2cf3bbf20130

Fixes #1121
Related to #515

### Notes to the reviewer

TBH there is a minor issue with the button. It goes blue before it's actually clickable 😅 I couldn't fix that, but anyway it's a big improvement over the current behaviour.

I know very little about UI development, so I had to rely on Cursor to generate the code for me. I hope it didn't do something silly. On a high level it make sense to me.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
+++ b/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
@@ -1,0 +1,175 @@
+import { faLink } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useCallback, useEffect, useState } from 'react';
+
+import { alloyStringify } from '../alloy-syntax-js/stringify';
+import { type Value, ValueType } from '../alloy-syntax-js/types';
+
+interface AsyncStringifiedValueProps {
+  value: Value;
+  maxLength?: number;
+}
+
+type LargeValueState =
+  | { status: 'loading' } // Shows greyed-out download text
+  | { status: 'ready'; result: string }; // Shows clickable download link
+
+/**
+ * Estimates the size of a value to determine if it will likely exceed maxLength.
+ * Returns a rough count of elements (array length, object keys, string length).
+ */
+function estimateValueSize(value: Value): number {
+  switch (value.type) {
+    case ValueType.STRING:
+      return value.value.length;
+    case ValueType.ARRAY:
+      return value.value.length;
+    case ValueType.OBJECT:
+      return value.value.length; // number of key-value pairs
+    default:
+      return 0;
+  }
+}
+
+/**
+ * Returns true if the value is a simple type that can be stringified synchronously.
+ * Simple types: null, number, bool, function, capsule
+ * Complex types that need async: object, array, string (strings can be very large)
+ */
+function isSimpleValue(value: Value): boolean {
+  switch (value.type) {
+    case ValueType.NULL:
+    case ValueType.NUMBER:
+    case ValueType.BOOL:
+    case ValueType.FUNCTION:
+    case ValueType.CAPSULE:
+      return true;
+    case ValueType.STRING:
+    case ValueType.ARRAY:
+    case ValueType.OBJECT:
+      return false;
+    default:
+      return false;
+  }
+}
+
+/**
+ * Renders a stringified value.
+ * Simple values (numbers, bools, etc.) are rendered synchronously.
+ * Small complex values are also rendered synchronously.
+ * Large complex values (that will exceed maxLength) are rendered asynchronously with a download link.
+ */
+const AsyncStringifiedValue = ({ value, maxLength = 50000 }: AsyncStringifiedValueProps) => {
+  // Simple values can be rendered synchronously - they're fast and small
+  if (isSimpleValue(value)) {
+    return <>{alloyStringify(value)}</>;
+  }
+
+  // Check if this value is likely to exceed maxLength
+  const isLargeValue = estimateValueSize(value) >= maxLength;
+
+  if (isLargeValue) {
+    // Large values use async rendering with download button
+    return <AsyncLargeValue value={value} />;
+  }
+
+  // Small complex values can be rendered synchronously
+  const result = alloyStringify(value);
+  if (result.length > maxLength) {
+    // Unexpectedly large after stringification - use download button
+    return <DownloadButton result={result} />;
+  }
+  return <>{result}</>;
+};
+
+/**
+ * Renders large values asynchronously with a download button.
+ * Shows greyed-out download text immediately, becomes clickable when ready.
+ */
+const AsyncLargeValue = ({ value }: { value: Value }) => {
+  const [state, setState] = useState<LargeValueState>({ status: 'loading' });
+
+  useEffect(() => {
+    let cancelled = false;
+
+    // Use double-RAF to ensure the loading state is painted before blocking work starts
+    requestAnimationFrame(() => {
+      if (cancelled) return;
+      requestAnimationFrame(() => {
+        if (cancelled) return;
+
+        try {
+          const result = alloyStringify(value);
+          if (!cancelled) {
+            setState({ status: 'ready', result });
+          }
+        } catch {
+          if (!cancelled) {
+            setState({ status: 'ready', result: '[Error stringifying value]' });
+          }
+        }
+      });
+    });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [value]);
+
+  if (state.status === 'loading') {
+    return (
+      <span
+        style={{
+          display: 'inline-block',
+          fontSize: '10px',
+          padding: '5px',
+          color: '#ffffff',
+          backgroundColor: '#888',
+          border: '1px solid #888',
+          borderRadius: '3px',
+        }}
+      >
+        <FontAwesomeIcon icon={faLink} /> Download the value contents
+      </span>
+    );
+  }
+
+  return <DownloadButton result={state.result} />;
+};
+
+/**
+ * A button that downloads the given result as a file.
+ */
+const DownloadButton = ({ result }: { result: string }) => {
+  const handleDownload = useCallback(() => {
+    const blob = new Blob([result], { type: 'text/plain' });
+    const url = URL.createObjectURL(blob);
+    const link = document.createElement('a');
+    link.href = url;
+    link.download = 'value.json';
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    URL.revokeObjectURL(url);
+  }, [result]);
+
+  return (
+    <button
+      onClick={handleDownload}
+      style={{
+        display: 'inline-block',
+        fontSize: '10px',
+        padding: '5px',
+        color: '#ffffff',
+        backgroundColor: 'rgb(56, 133, 220)',
+        border: '1px solid rgb(56, 133, 220)',
+        borderRadius: '3px',
+        cursor: 'pointer',
+      }}
+    >
+      <FontAwesomeIcon icon={faLink} /> Download the value contents
+    </button>
+  );
+};
+
+export default AsyncStringifiedValue;
--- a/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
+++ b/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
@@ -92,27 +92,26 @@ const AsyncLargeValue = ({ value }: { value: Value }) => {
   useEffect(() => {
     let cancelled = false;
 
-    // Use double-RAF to ensure the loading state is painted before blocking work starts
-    requestAnimationFrame(() => {
+    // Use setTimeout to defer blocking work to a macrotask, allowing React to
+    // commit and paint the loading state first
+    const timeoutId = setTimeout(() => {
       if (cancelled) return;
-      requestAnimationFrame(() => {
-        if (cancelled) return;
 
-        try {
-          const result = alloyStringify(value);
-          if (!cancelled) {
-            setState({ status: 'ready', result });
-          }
-        } catch {
-          if (!cancelled) {
-            setState({ status: 'ready', result: '[Error stringifying value]' });
-          }
+      try {
+        const result = alloyStringify(value);
+        if (!cancelled) {
+          setState({ status: 'ready', result });
         }
-      });
-    });
+      } catch {
+        if (!cancelled) {
+          setState({ status: 'ready', result: '[Error stringifying value]' });
+        }
+      }
+    }, 0);
 
     return () => {
       cancelled = true;
+      clearTimeout(timeoutId);
     };
   }, [value]);
 
--- a/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
+++ b/internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
@@ -84,7 +84,7 @@ const AsyncStringifiedValue = ({ value, maxLength = 50000 }: AsyncStringifiedVal
 
 /**
  * Renders large values asynchronously with a download button.
- * Shows greyed-out download text immediately, becomes clickable when ready.
+ * Shows a disabled download button immediately, becomes clickable when ready.
  */
 const AsyncLargeValue = ({ value }: { value: Value }) => {
   const [state, setState] = useState<LargeValueState>({ status: 'loading' });
@@ -115,31 +115,14 @@ const AsyncLargeValue = ({ value }: { value: Value }) => {
     };
   }, [value]);
 
-  if (state.status === 'loading') {
-    return (
-      <span
-        style={{
-          display: 'inline-block',
-          fontSize: '10px',
-          padding: '5px',
-          color: '#ffffff',
-          backgroundColor: '#888',
-          border: '1px solid #888',
-          borderRadius: '3px',
-        }}
-      >
-        <FontAwesomeIcon icon={faLink} /> Download the value contents
-      </span>
-    );
-  }
-
-  return <DownloadButton result={state.result} />;
+  const result = state.status === 'ready' ? state.result : '';
+  return <DownloadButton result={result} disabled={state.status === 'loading'} />;
 };
 
 /**
  * A button that downloads the given result as a file.
  */
-const DownloadButton = ({ result }: { result: string }) => {
+const DownloadButton = ({ result, disabled }: { result: string; disabled?: boolean }) => {
   const handleDownload = useCallback(() => {
     const blob = new Blob([result], { type: 'text/plain' });
     const url = URL.createObjectURL(blob);
@@ -154,6 +137,7 @@ const DownloadButton = ({ result }: { result: string }) => {
 
   return (
     <button
+      disabled={disabled}
       onClick={handleDownload}
       style={{
         display: 'inline-block',
--- a/internal/web/ui/src/pages/ComponentDetailPage.tsx
+++ b/internal/web/ui/src/pages/ComponentDetailPage.tsx
@@ -6,6 +6,26 @@ import { type ComponentDetail, type ComponentInfo, componentInfoByID } from '../
 import { useComponentInfo } from '../hooks/componentInfo';
 import { parseID } from '../utils/id';
 
+const contentStyle: React.CSSProperties = {
+  maxWidth: '1440px',
+  marginLeft: 'auto',
+  marginRight: 'auto',
+  padding: '20px',
+};
+
+const messageStyle: React.CSSProperties = {
+  color: '#666',
+  fontStyle: 'italic',
+  margin: '0 0 0.5rem 0',
+};
+
+const endpointStyle: React.CSSProperties = {
+  fontSize: '0.875rem',
+  fontFamily: 'monospace',
+  color: '#888',
+  margin: 0,
+};
+
 const ComponentDetailPage: FC = () => {
   const { '*': id } = useParams();
   const { moduleID } = parseID(id || '');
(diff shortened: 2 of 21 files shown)


kalleep:
What is taking time here, rendering or making the request to download everything.

If its rendering taking time maybe we can paginate it on the front-end so we only ever render a subset of the data

ptodev:
> What is taking time here, rendering or making the request to download everything.

The request to download is what takes the most amount of time.

> If its rendering taking time maybe we can paginate it on the front-end so we only ever render a subset of the data

The problem with rendering is that displaying huge amounts of information on the screen is not very useful. I think pagination also wouldn't be super useful, since users are often are looking for a very specific target in their list of targets - they wouldn't want to go through many pages until they find it. Being able to just download the whole contents in a file gives you the most amount of flexibility.

kalleep:
>The request to download is what takes the most amount of time.

Okay then pagination would not help.

>The problem with rendering is that displaying huge amounts of information on the screen is not very useful. I think pagination also wouldn't be super useful, since users are often are looking for a very specific target in their list of targets

I mean as soon as we have the data and do pagination etc you could always support filtering, like a fuzzy search.
But if the problem is the response time from alloy because the request is too big then yeah it would not help either way.

Still IMHO we should still do pagination with filtering and if requests are too big we could only fetch the data when you click a button like you have here.

internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
```diff
@@ -0,0 +1,174 @@
...
+        const result = alloyStringify(value);
+        if (!cancelled) {
+          setState({ status: 'ready', result });
+        }
+      } catch {
+        if (!cancelled) {
+          setState({ status: 'ready', result: '[Error stringifying value]' });
+        }
+      }
+    }, 0);
```
jharvey10, line 110: This will definitely work, but if you like, there's actually a react built-in that solves for this exact use case: `useDeferredValue`. (Also combined with `useMemo` for expensive value calculation caching). Its purpose is to treat a particular value as lower priority (in terms of rendering) to have react automatically process it in the background. So here, you'd instead have something like:

```tsx
const AsyncLargeValue = ({ value }: { value: Value }) => {
  const deferredValue = useDeferredValue(value);

  const isPending = value !== deferredValue;

  const result = useMemo(() => {
    try {
      return alloyStringify(deferredValue);
    } catch {
      return '[Error stringifying value]';
    }
  }, [deferredValue]);

  return <DownloadButton result={result} disabled={isPending} />;
};
```

internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
```diff
@@ -0,0 +1,174 @@
...
+          border: '1px solid #888',
+          borderRadius: '3px',
+        }}
+      >
+        <FontAwesomeIcon icon={faLink} /> Download the value contents
+      </span>
+    );
+  }
+
+  return <DownloadButton result={state.result} />;
```
jharvey10: This definitely works as well. You could potentially use the loading status in `state` to send a `disabled` prop down through to the `DownloadButton` and put that on the `button` element. Something like:

```tsx
return <DownloadButton result={state.result} disabled={state.status === 'loading'} />
```

and in the download button
```tsx
return (
    <button
      disabled={disabled}
      onClick={handleDownload}
      style={{
        display: 'inline-block',
        fontSize: '10px',
        padding: '5px',
        color: '#ffffff',
        backgroundColor: 'rgb(56, 133, 220)',
        border: '1px solid rgb(56, 133, 220)',
        borderRadius: '3px',
        cursor: 'pointer',
      }}
    >
      <FontAwesomeIcon icon={faLink} /> Download the value contents
    </button>
```

Then you wouldn't need a separate span while the download button isn't ready to click yet.

internal/web/ui/src/pages/ComponentDetailPage.tsx
```diff
@@ -6,26 +6,51 @@ import { type ComponentDetail, type ComponentInfo, componentInfoByID } from '../
...
+  margin: '0 0 0.5rem 0',
+};
+
+const endpointStyle: React.CSSProperties = {
+  fontSize: '0.875rem',
+  fontFamily: 'monospace',
+  color: '#888',
+  margin: 0,
+};
+
```
jharvey10: Suggestion to use css classes instead of inline styles as a more idiomatic (and usually more performant) approach.

in react:
```tsx
import styles from './ComponentDetailPage.module.css'

// then use with:
<p className={styles.endpoint}>...</p>
```

in css:
```css
.endpoint {
  font-size: 0.875rem;
  font-family: 'Fira Code', monospace;
  color: #888;
  margin: 0;
}
```

internal/web/ui/src/pages/ComponentDetailPage.tsx
```diff
@@ -35,21 +60,42 @@ const ComponentDetailPage: FC = () => {
...
-  return component ? <ComponentView component={component} info={infoByID} /> : <div></div>;
+  if (component) {
+    return <ComponentView component={component} info={infoByID} />;
+  }
+
+  return (
+    <main style={contentStyle}>
+      <p style={messageStyle}>Loading component data...</p>
+      {loadingEndpoint && <p style={endpointStyle}>Fetching: {loadingEndpoint}</p>}
+    </main>
```
jharvey10: I don't think this issue is specific to this PR, but I think this adds to the issue. There should only be one `main` element per html page, and I believe this is already covered by the main layout component (Page.tsx).

internal/web/ui/tsconfig.app.json
```diff
@@ -14,6 +14,7 @@
     "moduleDetection": "force",
     "noEmit": true,
     "jsx": "react-jsx",
+    "resolveJsonModule": true,
```
jharvey10: I may be overlooking this, but is there a particular component importing a json module that makes this required? I didn't see any off-hand, but I could be missing it.

jharvey10 (approved):
Just a few optional suggestions and questions, but overall looks good to me!

jharvey10:
@kalleep Agree that there could be better on-demand requesting, and probably a revamp of the entire UX of this page is prudent.

I think there's some pre-existing page load logic that makes doing smarter loading at the moment a bit more involved of a change, but I'm certain that we can refactor those parts as well to make even more improvements.

internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
ptodev: I'm going to give up on implementing this change because Cursor disagrees with it, and after researching it for a bit I'm still not sure how exactly to make it better. I could probably iterate on it a bit and test it, but if it ultimately looks the same to the end user there's probably not enough benefit to change it.

Screenshot 2026-03-27 at 10 51 41

internal/web/ui/tsconfig.app.json
ptodev: I think it's a remnant from earlier code, I'll delete it.

internal/web/ui/src/pages/ComponentDetailPage.tsx
ptodev: Thanks, done!

internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
ptodev: Thanks, done!

internal/web/ui/src/features/component/AsyncStringifiedValue.tsx
jharvey10: Interesting! I guess there's some subtleties with useDeferredValue I don't know about. Practically speaking, when I've used it the UI tends to "feel" more responsive. The non-deferred render pieces happen immediately, which makes the UI "feel" snappy, and then when things are idle, the deferred value recomputes and is typically "ready" by the time it's actually needed (or an intermediate fallback state is put in to bridge the gap). 

The point about the equality check is a fair one though. That's probably just my pseudo logic breaking down from not properly thinking through the initial render case.

Either way, I don't think there's anything wrong with your approach!

PR #5270, opened 2026-01-15
fix: Allow loki.source.file to read renaming lines of a deleted file before it tries to re open a new one 

kalleep:
### Pull Request Details
When we detect that a file is deleted we drain anything left from the opened fd. But then we tried to reopen then new file created.

I am working better integration tests for this component and discovered that we could miss these buffered lines if the file was deleted and no new file would was created. What we want to do is to be able to consume these even if a new file never is created and the target disappear eventually.

So before this pr if a file was deleted and we successfully drained the rest of the file these would not be returned before the new filed appeared because we would be stuck waiting on this new file.

It is safe to return here because we will return buffered lines and then get the same event again.

The test I added fails on main.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/file.go
+++ b/internal/component/loki/source/file/internal/tail/file.go
@@ -166,7 +166,7 @@ func (f *File) wait() error {
 		// If a file is deleted we want to make sure we drain what's remaining in the open file.
 		f.drain()
 		// If we have any buffered lines after drain we can return here to make sure they are consumed and
-		// we are not blocking on a new file to aprea.
+		// we are not blocking on reopening the new file.
 		if len(f.bufferedLines) > 0 {
 			level.Debug(f.logger).Log("msg", "finish reading deleted file before reopen")
 			return nil


internal/component/loki/source/file/internal/tail/file.go
```diff
@@ -163,8 +163,14 @@ func (f *File) wait() error {
...
 		level.Debug(f.logger).Log("msg", "file deleted")
-		// if a file is deleted we want to make sure we drain what's remaining in the open file.
+		// If a file is deleted we want to make sure we drain what's remaining in the open file.
 		f.drain()
+		// If we have any buffered lines after drain we can return here to make sure they are consumed and
+		// we are not blocking on reopening the new file.
+		if len(f.bufferedLines) > 0 {
+			level.Debug(f.logger).Log("msg", "finish reading deleted file before reopen")
+			return nil
+		}
```
ptodev, line 173: Do truncated files have the same problem? And would it make sense for `reopen()` to make sure there's nothing in the buffer before reopening?

kalleep: If a truncation have happened it would mean that the file is smaller than it was before so it is impossible to buffer anything.

ptodev: Is it possible that we buffered something prior to the truncation but we still haven't processed it?

If it's possible to buffer something prior to deletion, I figured it must be possible to buffer something prior to truncation too.

kalleep: No we only buffer lines when we get a delete event so that is not possible

PR #5271, opened 2026-01-16
chore: Add integration tests for loki.source.file

kalleep:
Add integration tests for loki.source.file testing various rotation strategies.

### Pull Request Details
This pr adds new integration tests for loki.soure.file that will test:
1. Rename rotation
2. Delete rotation
3. Copy truncate rotation
4. New file rotation

This is to make sure we test common rotation setups.

By far the worst strategy is Copy truncate. I had to add a small sleep in that rotation before truncation happens so that alloy have time to ingest everything before file gets reset.

To make this work I also had to add support for mounting directories into alloy container. So if there is a `mount` directory in the test directory it will be mounter under `/etc/alloy/mount`. Alloy starts file discovery on this dir and all writes and rotation happens from the test itself.

Part of: https://github.com/grafana/alloy/issues/5234

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
+++ b/integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
@@ -247,7 +247,6 @@ type writer struct {
 }
 
 func (w *writer) run() error {
-	// We need an initial sleep so that alloy has time to discover the first set of files.
 	ticker := time.NewTicker(50 * time.Millisecond)
 
 	defer func() {


kalleep:
Test is failing in ci but not locally so will investigate that, we get a panic because t.ErorrF is called in a goroutine and not the test code..

Edit: Fixed and test are passing

integration-tests/docker/utils.go
```diff
@@ -182,8 +186,41 @@ func runSingleTest(ctx context.Context, testDir string, port int, stateful bool,
 		}
 	}()
 
+	// FIXME(@kalleep): All these checks for special setups are
+	// annoying. We should figure out a way for a test to describe
+	// special setups it needs like mounts, exported ports etc.
+	if dirName == "loki-file-rotation" {
```
kalleep, line 189: This part is really annoying.

What I wanted to achieve is that is there is a mount directory we automatically mount it into the container. But the mount fails is there is anything in there.

Git will exclude the directory because it's empty so test fails.

We have these special setups for loki-file-rotation, loki-enrish and beyla tests. I wan to look into some way to describe special setup we need like mounts, exposing ports. Having e.g. a test.yaml file that will describe these things so it can be reused for other tests without having to hardcode these things here.

kalleep: Added a issue for this https://github.com/grafana/alloy/issues/5272

thampiotr: I think you can keep empty directories in git by adding a `.gitignore` file inside of them? And maybe we can have pattern matching that will not try to tail that file?

integration-tests/docker/tests/loki-file-rotation/config.alloy
```diff
@@ -0,0 +1,29 @@
...
+}
+
+loki.process "test" {
+  stage.logfmt {
+  	mapping = { "test_name" = "" }
+  }
+
+  stage.labels {
+  	values = { "test_name" = "" }
+  }
```
thampiotr, line 20: Do we need both stages?

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+		if err := copyTo.Close(); err != nil {
+			return nil, fmt.Errorf("failed to close new file: %w", err)
+		}
+
+		// CopyTruncate is an unforgiving strategy. If the last write is too close
+		// to the time we truncate the file it is likely that we miss some logs.
+		// This sleep is here to ensure that we pass the test because without it we will
+		// miss some logs. We ingest around 4950 out of 5000 without this sleep.
+		time.Sleep(2 * time.Second)
```
thampiotr, line 99: Let's add an TODO and issue for this, or link to existing one?

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
thampiotr, line 1: That's a really nice test!

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
```diff
@@ -0,0 +1,323 @@
...
+	totalWritten int
+
+	rotateFn rotateFn
+
+	results []common.ExpectedLogResult
+}
+
+func (w *writer) run() error {
+	// We need an initial sleep so that alloy has time to discover the first set of files.
+	time.Sleep(5 * time.Second)
```
thampiotr: Hmmm, I am not sure about this. In real world scenario files can be written to as soon as they are created. What happens if we skip this sleep?

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
```diff
@@ -0,0 +1,323 @@
...
+type rotateFn func(f *os.File, generation int) (*os.File, error)
+
+type writer struct {
+	id         int
+	generation int
+	testName   string
+
+	file *os.File
+
+	fileWritten  int
```
thampiotr: The name of this is not very intuitive. Can you improve it? fileWritten I would expect to be a file handle or file name, but it is some kind of counter. Is this number of files written? or number of rotations?

integration-tests/docker/utils.go
thampiotr: Maybe not having a test.yaml but some kind of test config you set up on the test in Go code before its run. The test could be identified by its name or its directory path or some other form of ID.

thampiotr (commented):
These look like very valuable tests! thanks for adding

integration-tests/docker/tests/loki-file-rotation/config.alloy
kalleep: Yes, this is a bit hacky I set the test name from log lines because we have several test with the same config.

First block extracts `test_name` to shared map and second block adds it as a label

integration-tests/docker/utils.go
kalleep: >I think you can keep empty directories in git by adding a .gitignore file inside of them? And maybe we can have pattern matching that will not try to tail that file?

Well I tested with having something like `.gitkeep` (convention for keeping empty directories). But the problem is then that volume mount fails.

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
kalleep: We don't have an exists one but copy truncate is always going to be problematic for log collectors. In all other cases we can still consume to EOF, even when files get deleted. But here we don't have any guarantees.

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
kalleep: And that is true but this is more of a issue with the test itself / rotations being to fast.

When we run our integration tests we start up alloy before we run test code. So when alloy starts our directory is empty. Then we start running our test code e.i. writing log files. The problem is that we are doing kinda fast rotations so without the sleep we could already have written and rotated the first set of files before alloy have had time to discover them.

In real world rotations are not (usually) done this fast as we are doing in these tests.

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
kalleep: Its the number of log lines written for current generation, I can try to come up with a better name

integration-tests/docker/utils.go
kalleep: >Maybe not having a test.yaml but some kind of test config you set up on the test in Go code before its run. The test could be identified by its name or its directory path or some other form of ID.

I was thinking this first but it's kinda tricky with how we run our integration tests. Each test is it's own sub procces called from [here](https://github.com/grafana/alloy/blob/main/internal/cmd/integration-tests/utils.go#L215-L230). Essential we run the integration tests binary and it will perform all setups. Once that is done it will trigger `go test` on your test directory. Because of how we run our tests it's not really possible to describe "extra" things you need. That's why I suggested having something like a `test.yaml` for this. But open to other suggestions as well

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
kalleep: Updated in this [commit](https://github.com/grafana/alloy/pull/5271/commits/2fcacd99eb589c4edb382e399e6bb495d291cef3)

integration-tests/docker/utils.go
kalleep: So having some kind of config file for tests are by far the easiest solution if we don't look into refactoring how we run these test, that could also be a valid option.

I wrote some POC for this here https://github.com/grafana/alloy/compare/kalleep/loki-file-rotation-integration-test...kalleep/integration-test-setup?expand=1

thampiotr: cool, I think that yaml is okay, maybe a follow-up PR

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
thampiotr: That's still okay to have an issue even if it will never be solved 😓 
Maybe one day there will be some standard way to handle it. Or we can detect it and log a warning recommending against it. Or document that we don't recommend it.

internal/cmd/integration-tests/tests/loki-file-rotation/loki_file_rotation_test.go
thampiotr: Gotcha! Any way we can have some kind of probe to wait for Alloy to be ready? Thinkin that 5s may be sometimes too much and sometimes too little making it flaky. Generally Sleep is often leading to flaky tests.

kalleep: I don't know if there is a easy way to query what targets have been discovered by a component in alloy through the api. That would be By far the best option. Then we could perform an initial wait until we have discovered the initial files.

kalleep: But yeah the sleep is not optimal

thampiotr: Was thinking about checking for Alloy's `/-/ready` endpoint, do you reckon this could still result in missed logs? 

Or could we write one log line first and wait for it to appear?

Both of these ideas may turn out to be overcomplicated, so let me know if that ends up too ugly and we may need to live with the sleep... but I worry that tests would become too flaky.

kalleep: Ready is not enough, the first line idea would work and is better than the sleep for sure. Will try that out!

kalleep: Waiting for first line worked :+1:

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
```diff
@@ -0,0 +1,339 @@
...
+	// even if file is rotated.
+	totalLinesWritten int
+
+	rotateFn rotateFn
+
+	results []common.ExpectedLogResult
+}
+
+func (w *writer) run() error {
+	// We need an initial sleep so that alloy has time to discover the first set of files.
```
thampiotr: Stale comment. Maybe say that `// We poll for the first line to start the writer and file rotations only after Alloy's pipeline is confirmed to be running`

kalleep: I just forgot to remove this one. I have a comment mentioning why we need to wait for first line

integration-tests/docker/tests/loki-file-rotation/loki_file_rotation_test.go
kalleep: I created a issue for it https://github.com/grafana/alloy/issues/5310

PR #5285, opened 2026-01-19
chore: Update build image to use go v1.25.6

kalleep:


blewis12 (approved):
checking out the [release docs for alpine 3.23.0](https://www.alpinelinux.org/posts/Alpine-3.23.0-released.html) the only warning is regarding `libapk` which we don't use as far as I can see 👍

kalleep:
> checking out the [release docs for alpine 3.23.0](https://www.alpinelinux.org/posts/Alpine-3.23.0-released.html) the only warning is regarding `libapk` which we don't use as far as I can see 👍

Yeah I just figured we jump to the latest anyway

PR #5288, opened 2026-01-19
otelcol/receiver/prometheus: add script to sync with upstream version

juliusmh:

Adds script to automatically update to latest prometheus otelcol receiver and apply local patches to make it work in alloy.

### Pull Request Details

n/a

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


go.mod
ptodev, line 1: Which version of the upstream repo was used for this sync?

There are more updates here than I'd expect, so I suppose v0.139.0 isn't the one that was used.

Issue #5142, opened 2025-12-18
[Tracking Issue] Flaky tests


kalleep:
# Flaky Tests Tracking

This issue tracks tests that fail intermittently in CI.  
Add new items to the list below when a flaky test is discovered, and remove them when fixed.

## How to add a flaky test
- Add a new bullet under **Current flaky tests**
- Include:
  - Test name
  - OS
  - Link to a failing CI run

## Current flaky tests
TestMissingBearerAuthFile - [Windows](https://github.com/grafana/alloy/actions/runs/23339330864/job/67888853484)
TestEndpointBlockOnOverflow - [MacOS](https://github.com/grafana/alloy/actions/runs/23479491638/job/68319489308?pr=5850)

dehaansa:
A flaky test in loki.source.file - https://github.com/grafana/alloy/issues/5482

PR #5289, opened 2026-01-19
chore: Move integration tests and update how we run our tests

kalleep:
### Pull Request Details
We have a issue running our tests on windows
```
usr/bin/bash: line 1: /c/hostedtoolcache/windows/go/1.25.6/x64/bin/go: Argument list too long
```

Essentially the output of `$(go list ./... | grep -v -E "/integration-tests/|/integration-tests-k8s/")` is to large to pass as arguments for windows.

We run this command to exclude our integration tests. To remove this workaround I moved all integration tests from `internal/cmd` to `tests` folder. So `make tests` will now run all tests in `internal`, `extentions/alloyengine` and `collector`. The last two are have their own go mod and we never triggered these tests in ci before.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,8 +177,8 @@ else
 	go test -tags=packaging -race ./internal/tools/packaging_test
 endif
 
-.PHONY: integration-test
-integration-test:
+.PHONY: integration-test-docker
+integration-test-docker:
 	cd integration-tests/docker && $(GO_ENV) go run .
 
 .PHONY: integration-test-k8s


kalleep:
We can now run our tests on windows.

The tests that are failing are already noted down here: https://github.com/grafana/alloy/issues/5142

Can address them in a followup pr, don't want it to block this one

ptodev (commented):
I'm worried that if someone adds a test that's not in `/internal`, then there's a chance that they won't realise their tests are not being ran by the CI. Isn't it customary to be able to run tests from the same folder where the go.mod is? 

Ideally, there should be a good way to exclude integration tests from being ran on windows via the cmd line. If there isn't a good way to do so, we could use comments like `//go:build !windows`? Those tests use k8s and docker which are more of a Linux thing anyway. To make this more explicit, you could rename the `integration-tests` folder to `integration-tests-linux`? In the future we could add another folder called `integration-tests-windows` to test Windows-specific features.

integration-tests/k8s/tests/mimir-alerts-kubernetes/testdata/expected_1.yml
thampiotr, line 1: Could we maybe do:
`integration-tests/docker`
`integration-tests/k8s`

I think `tests` is too generic and we keep unit tests next to code.

.github/workflows/test_windows.yml
```diff
@@ -31,5 +31,4 @@ jobs:
         # We'll need to make sure the same cache is reused by the workflow to build Windows binaries.
         cache: false
     - name: Test
-      run: '& "C:/Program Files/git/bin/bash.exe" -c ''go test -tags="nodocker,nonetwork"
-        $(go list ./... |  grep -v -E "/integration-tests/|/integration-tests-k8s/")'''
+      run: '& "C:/Program Files/git/bin/bash.exe" -c ''go test -tags="nodocker,nonetwork" ./internal/...'''
```
thampiotr: `./internal/...` is not enough I think because there are more top-level directories that may now or in the future contain more tests.

We could list all the top-level directories. Or we could use some script to do that automatically and exclude integration-tests/. Or we can use some other go test setting / env variable to skip the integration tests.

integration-tests/k8s/tests/mimir-alerts-kubernetes/testdata/expected_1.yml
kalleep: Fixed

kalleep:
@ptodev @thampiotr in [a4d50ad](https://github.com/grafana/alloy/pull/5289/commits/a4d50ad246c2a025d694869e323137064ca3e45b) I updated to set `integration` as build tags for all of our integration tests and update make test to find all directories that contains go.mod and trigger tests there.

I just need to figure out how we can do this for windows too. I also need to add back to manually trigger node_exporter test because it has !race flag. Let me know what you think

thampiotr:
> @ptodev @thampiotr in [a4d50ad](https://github.com/grafana/alloy/pull/5289/commits/a4d50ad246c2a025d694869e323137064ca3e45b) I updated to set `integration` as build tags 

One issue with tag `integration` is that it's likely to be a build tag that one of our dependencies would use, which can affect us. I would call it `alloy-integration-tests` - it doesn't cost much to be a bit more verbose.

kalleep:
> One issue with tag `integration` is that it's likely to be a build tag that one of our dependencies would use, which can affect us. I would call it `alloy-integration-tests` - it doesn't cost much to be a bit more verbose.

Apparently you cannot have dashes in build tags so had to go with alloyintegrationtests

Makefile
```diff
@@ -163,10 +163,12 @@ run-alloylint: alloylint
 # more for packages that exclude tests via //go:build !race due to known race detection issues. The
 # final command runs tests for syntax module.
 test:
-	$(GO_ENV) go test $(GO_FLAGS) -race $(shell go list ./... | grep -v -E '/integration-tests/|/integration-tests-k8s/')
-	$(GO_ENV) go test $(GO_FLAGS) ./internal/static/integrations/node_exporter
-	$(GO_ENV) cd ./syntax && go test -race ./...
-
+	@for dir in $$(find . -name go.mod -type f -exec sh -c 'dirname "$$1"' _ {} \;); do \
+		if echo "$$dir" | grep -qv testdata; then \
```
kalleep, line 167: This is a bit ugly but we have some modules here https://github.com/grafana/alloy/tree/main/tools/generate-module-dependencies/testdata that don't have a main package and if we try to run tests there they will fail so we need to skip them..

thampiotr: Yeah, a bit ugly, but I think that's reasonable.

Makefile
```diff
@@ -177,19 +179,18 @@ endif
 
 .PHONY: integration-test
 integration-test:
-	cd internal/cmd/integration-tests && $(GO_ENV) go run .
+	cd integration-tests/docker && $(GO_ENV) go run .
```
thampiotr, line 182: While at it, can we also name the `integration-test` target an `integration-test-docker` to align with the k8s one?

internal/cmd/integration-tests/tests/read-log-file/config.alloy
thampiotr, line 1: Are these deleted by accident?

kalleep: No these are duplicates of `loki-file`. I reworked this test in https://github.com/grafana/alloy/pull/5237 but forgot to remove the old folder.

Makefile
ptodev: Would it not be better to rename the [go.mod](https://github.com/grafana/alloy/blob/main/tools/generate-module-dependencies/testdata/basic-mod/go.mod) to `go.mod.original`, and to have [the test](https://github.com/grafana/alloy/blob/main/tools/generate-module-dependencies/e2e_test.go) use these files in a tmp folder?

kalleep: Is it fine to leave this as a followup?

ptodev: Sure!

PR #5292, opened 2026-01-20
chore: Fix flaky tests on linux

kalleep:
Fix flaky tests that I have discovered on linux

#### TestCounterExpiration
This test is here to ensure that counter will prune expired counters, this had a timing issue (sleep) that we sometimes observe in CI. To fix this I used the new `testing/synctest` package.

#### TestFileWatch
This test is like a basic smoke test to ensure tiler will react to file changes when blocked waiting on events.
The timing issue was the check for `timeBeforeWriting`, this one was sometimes faster than 1 sec. But I fell this is an unnecessary check to have.

While not strictly necessary for this test I also used `testing/synctest` here, mostly to speed it up. I also removed that last sleep and used a waitgroup to ensure component have existed before the test finish.

Part of: https://github.com/grafana/alloy/issues/5142

### Notes to the Reviewer

`testing/synctest` is a really useful package that we should start leveraging more, it will speed up a lot of our timing based tests and make them more consistent. It will hijack the clock and only advance time  when there is no runnable goroutines. It can be used to make timings deterministic.

Because of that test using sleeps will also be a lot faster

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/file_test.go
+++ b/internal/component/loki/source/file/file_test.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/file/file_test.go
```diff
@@ -267,61 +268,56 @@ func TestUpdateRemoveFileWhileReading(t *testing.T) {
...
-			})},
-			ForwardTo: []loki.LogsReceiver{ch1},
-			FileWatch: FileWatch{
-				MinPollFrequency: time.Millisecond * 500,
-				MaxPollFrequency: time.Millisecond * 500,
-			},
-			FileMatch: match,
-		}
+		synctest.Test(t, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
```
x1unix, line 272:
```suggestion
			ctx, cancel := context.WithCancel(t.Context())
```

PR #5294, opened 2026-01-20
fix: Database_observability: grant check only require SELECT *.* on perf_schema

matthewnolf:
This change fixes the MySQL health_check for permissions being overly restrictive. For `SELECT` permissions, we only need `performance_schema.*` - not `*.*`. This updates the logic to more thoroughly check for `performance_schema` access.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -160,30 +162,44 @@ func checkRequiredGrants(ctx context.Context, db *sql.DB) healthCheckResult {
...
+				strings.Contains(up, " ON `PERFORMANCE_SCHEMA`.") ||
+				strings.Contains(up, " ON *.*") {
+
+				req["SELECT"] = true
 			}
 		}
+
+		if strings.Contains(up, "SHOW VIEW") {
+			req["SHOW VIEW"] = true
+		}
```
cristiangreco, line 187: Maybe for a followup: would be cool to show on which objects `SHOW VIEW` is granted (e.g. `*.*` vs `payments.*` or something else), as it helps debug why e.g. a specific schema is not reported.

internal/component/database_observability/mysql/collector/health_check.go
```diff
@@ -160,30 +162,44 @@ func checkRequiredGrants(ctx context.Context, db *sql.DB) healthCheckResult {
...
-			r.result = false
-			if r.value == "" {
-				r.value = "missing: " + k
-			} else {
-				r.value += "," + k
-			}
-		}
-	}
-
+	r.result = req["PROCESS"] && req["REPLICATION CLIENT"] && req["SELECT"] && req["SHOW VIEW"]
```
cristiangreco, line 202: Could we add to `r.result` the list of expected vs missing grants?

matthewnolf: `r.result` is a boolean whether the check passed or failed.
`r.value` can be used to send details of the check.

This comment seems similar to https://github.com/grafana/alloy/pull/5294#discussion_r2712285154 in which we attach more information about which grant is missing?

cristiangreco: > `r.result` is a boolean whether the check passed or failed.
> `r.value` can be used to send details of the check.

Sorry I confused the two fields. But yeah, basically I was thinking of attaching more info like I mentioned in the other comment.

cristiangreco (approved):

PR #5295, opened 2026-01-20
chore: Fix flaky tests on windows

kalleep:
### Pull Request Details
Only tests in for `loki.source.windowsevent` was actually flaky. The prometheus related tests would never pass.

The issue in loki.source.windowsevent is not limited to test but we started two goroutines and incremented the wg counter inside them, that is an actual bug. The counter should always be started before the goroutine and using wg.Go ensures that.

Part of: https://github.com/grafana/alloy/issues/5142

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/tools/generate-module-dependencies/e2e_test.go
+++ b/tools/generate-module-dependencies/e2e_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/grafana/replace-generator/cmd"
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/windowsevent/target.go
```diff
@@ -32,10 +32,9 @@ type Target struct {
 	bm      *bookMark // bookmark to save positions.
 	fetcher *win_eventlog.EventFetcher
 
-	ready bool
```
kalleep, line 35: This was never used

internal/component/loki/source/windowsevent/target.go
```diff
@@ -86,24 +85,18 @@ func NewTarget(
...
 func (t *Target) loop() {
-	t.ready = true
-	t.wg.Add(1)
 	interval := time.NewTicker(t.cfg.PollInterval)
-	defer func() {
-		t.ready = false
-		t.wg.Done()
-		interval.Stop()
-	}()
+	defer interval.Stop()
```
dehaansa, line 97: There should still be a wg.Done somewhere, right?

kalleep: Nope, change to use `wg.Go`. This function will call add and done for you https://pkg.go.dev/sync#WaitGroup.Go

dehaansa: Oh, of course, my brain interpreted the change as removing the wg.Wait() but that's not the change at all. 👍

PR #5299, opened 2026-01-20
refactor: Call UnregisterCollector on remotecfg shutdown

spartan0x117:

Adds a call to `UnregisterCollector` when the `remotecfg` service stops. The idea behind this is to have collectors send a signal when they shut down to the remote so that it can track state accordingly.

### PR Checklist

- [ ] Documentation added
- [X] Tests updated


tpaschalis (approved):
LGTM!

PR #5304, opened 2026-01-21
chore: Replace interface{} with any

kalleep:
### Pull Request Details
Add a rewrite rule to always use `any` instead of `interface{}`. Just so we are consistent within our codebase.

This rule will fail linter if `interface{}` is used. Running e.g. `golangci-lint fmt` replace `interface{}` with `any`.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


go.mod
```diff
@@ -1012,13 +1012,17 @@ require (
 	gotest.tools/gotestsum v1.13.0 // indirect
 )
 
+require github.com/google/go-tpm-tools v0.4.4
```
kalleep: Not sure why this is required all of a sudden 🤔

kalleep: Ah figured it out. Will make another pr to fix this issue.

grcevski (approved):
LGTM! for the Beyla changes.

kgeckhart (approved):

matthewnolf (approved):
Looks good for database_observability components

PR #5308, opened 2026-01-21
chore: Sync the translator for otelcol.receiver.prometheus with upstream

ptodev:

feat(otelcol.receiver.prometheus): It is no longer required to run Alloy with `--stability.level=experimental`  in order to translate Prometheus native histograms into OTLP exponential histograms.

BREAKING CHANGE: `otelcol.receiver.prometheus` no longer sets start times of OTLP metrics. Grafana Cloud and Mimir do not currently use OTLP metric start times. If you do want your metrics to have them, you can use `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point` to get the same behaviour.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

BEGIN_COMMIT_OVERRIDE
feat(otelcol.receiver.prometheus): Remove requirement to run Alloy with `--stability.level=experimental` in order to translate Prometheus native histograms into OTLP exponential histograms. (#5308)

BREAKING CHANGE: `otelcol.receiver.prometheus` no longer sets start times of OTLP metrics. Grafana Cloud and Mimir do not currently use OTLP metric start times. If you do want your metrics to have them, you can use `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point` to get the same behaviour.
END_COMMIT_OVERRIDE

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
+++ b/docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
@@ -13,8 +13,8 @@ title: otelcol.processor.metric_start_time
 `otelcol.processor.metric_start_time` accepts metrics from other `otelcol` components and sets the start time for cumulative metric datapoints which do not already have a start time.
 This processor is commonly used with `otelcol.receiver.prometheus`, which produces metric points without a [start time][otlp-start-time].
 
-Grafana Mimir ingests OTLP metric start times only when it is configured with the -distributor.otel-created-timestamp-zero-ingestion-enabled flag.
-Without this configuration, setting start times in Alloy has no effect on ingestion behavior.
+Grafana Mimir ingests OTLP metric start times only when it is configured with the `-distributor.otel-created-timestamp-zero-ingestion-enabled` flag.
+Without this configuration, setting start times in {{< param "PRODUCT_NAME" >}} has no effect on ingestion behavior.
 
 {{< admonition type="note" >}}
 `otelcol.processor.metric_start_time` is a wrapper over the upstream OpenTelemetry Collector [`metricstarttime`][] processor.
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
@@ -24,19 +24,19 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 
 
 {{< admonition type="note" >}}
-In Alloy v1.14.0, `otelcol.receiver.prometheus` no longer sets OTLP metric [start times][otlp-start-time] for Prometheus metrics. The receiver forwards metrics without adding start times when the Prometheus input does not have a [created time][om-suffixes].
+In {{< param "PRODUCT_NAME" >}} v1.14.0, `otelcol.receiver.prometheus` no longer sets OTLP metric [start times][otlp-start-time] for Prometheus metrics. The receiver forwards metrics without adding start times when the Prometheus input does not have a [created time][om-suffixes].
 
 Start time is a way to tell when a cumulative metric such as an OTLP "sum" or a Prometheus "counter" was last reset.
 If your database uses start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it.
 Grafana Mimir's OTLP endpoint and Grafana Cloud's OTLP Gateway both support OTLP metric start time.
 
-To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.14.0, 
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to {{< param "PRODUCT_NAME" >}} v1.14.0, 
 configure `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point`.
 
 In practice it is only necessary to use `true_reset_point` if your Mimir/Grafana Cloud instance is explicitly configured with a 
 `-distributor.otel-created-timestamp-zero-ingestion-enabled=true` flag, 
 since `true_reset_point` will send a sample with the same start and end time and Mimir by default won't create a zero sample for it.
-So for most Mimir/Grafana Cloud use cases there would be no change in behaviour even if no `otelcol.processor.metric_start_time` is added.
+For most Mimir and Grafana Cloud use cases there would be no change in behaviour even if no `otelcol.processor.metric_start_time` is added.
 
 [om-suffixes]: https://github.com/prometheus/OpenMetrics/blob/v1.0.0/specification/OpenMetrics.md#suffixes
 [otlp-start-time]: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L181-L187


docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,13 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 `otelcol.receiver.prometheus` is a custom component built on a fork of the upstream OpenTelemetry receiver.
 {{< /admonition >}}
 
+As of Alloy v1.13.0, `otelcol.receiver.prometheus` no longer sets a [start time][otlp-start-time] for the translated OTLP metric datapoint.
+Start time is a way to tell when a cumulative metric such as an OTLP "sum" or a Prometheus "counter" was last reset.
+If your database requires start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it. 
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.13.0, set `strategy` to `true_reset_point`.
```
ptodev: TODOs:
- [ ] Make sure the version of Alloy which introduces this breaking change is correct.
- [ ] Clarify if Mimir uses start times of OTLP metrics

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,13 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 `otelcol.receiver.prometheus` is a custom component built on a fork of the upstream OpenTelemetry receiver.
 {{< /admonition >}}
 
+As of Alloy v1.13.0, `otelcol.receiver.prometheus` no longer sets a [start time][otlp-start-time] for the translated OTLP metric datapoint.
```
krajorama: As discussed, I think you mean 
```suggestion
As of Alloy v1.13.0, `otelcol.receiver.prometheus` if a Prometheus metric lacks [start time][otlp-start-time] (formerly known as created time, showing as `..._created` series on the `/metrics` endpoint), then `otelcol.receiver.prometheus` no longer sets a [start time][otlp-start-time] for the translated OTLP metric datapoint.
```

ptodev: Nice clarification, thank you!

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,22 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
...
+Start time is a way to tell when a cumulative metric such as an OTLP "sum" or a Prometheus "counter" was last reset.
+If your database uses start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it.
+Mimir's OTLP endpoint and Grafana Cloud's OTLP Gateway both support OTLP metric start time.
+
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.14.0, 
+configure `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point`.
+
+[om-suffixes]: https://github.com/prometheus/OpenMetrics/blob/v1.0.0/specification/OpenMetrics.md#suffixes
+[otlp-start-time]: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L181-L187
+{{< /admonition >}}
```
ptodev: @krajorama does this sound good? I wish we could add more detail on why Mimir is more correct with those start timestamps, but if you're not sure we could just leave it like this.

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,22 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 `otelcol.receiver.prometheus` is a custom component built on a fork of the upstream OpenTelemetry receiver.
 {{< /admonition >}}
 
+
+{{< admonition type="warning" >}}
+In Alloy v1.14.0 `otelcol.receiver.prometheus` has been changed to no longer set a [start time][otlp-start-time]
```
ptodev: @clayton-cornell Is it ok to mention a particular version like this?

clayton-cornell: We do mention specific version for things like deprecation notes - you can see it in the docs for `prometheus.exporter.windows` (Warning and Note in the Blocks section) for example.  

The deprecation/remove guidelines here: https://grafana.com/docs/writers-toolkit/write/deprecate-remove/ mention specific versions in the recommended text.

If this is a notice of a future removal, we should follow the standard wording for this that's given in the Writer's Toolkit.

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
krajorama: It's almost correct, but it doesn't cover the case when `use_start_time_metric: true` was set, see my conclusion from the slack thread:

Ok, so conclusion: use the metric start time processor now.if the config used to have  something like
```yaml
prometheus:
    use_start_time_metric: true
    start_time_metric_regex: foo_bar_.*
```
Then use the processor with something like:
```
metricstarttime:
    strategy: start_time_metric
    gc_interval: 1h
    start_time_metric_regex: "foo_bar_.*"
```
if nothing was set, then
```
metricstarttime:
    strategy: true_reset_point
```

Regarding Mimir: since with the above settings we're making sure that the otel collector output is unchanged, it doesn't matter what Mimir does.

But for completeness: Mimir and Grafana Cloud can optionally (not default at the moment) can inject zero samples at the start time indicated by OTLP. In later versions it is likely that this will be handled differently.

What happens if people *don't do* the above? 
In case of the true_reset_point strategy, nothing happens at the moment, since the start time and the timestamp of the first sample in a series will be the same - in which case Mimir currently ignores the start time as it cannot ingest a 0 sample and real sample at the same timestamp.

In case of the start_time_metric, some rate calculations where the time range includes the start of the process will be different. E.g. the `rate(some_metric)[5m]` will start one scrape time later than before as the first sample will not have a preceding zero sample.

ptodev: Thank you so much for the detailed reply!

Alloy [always sets](https://github.com/grafana/alloy/pull/5308/files#diff-da1b2a93f76f17c5f044b243bcef91a3835d7d4a4e85748fd5a7f483a6f71714R123) `use_start_time_metric` to `false`. We don't need to worry about that use case :)

> In case of the true_reset_point strategy, nothing happens at the moment

I did my best to clarify this in deb5bb1acf775143a28b3732d6f3f321395af28f. I hope it looks good to you?

docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
```diff
@@ -13,6 +13,8 @@ title: otelcol.processor.metric_start_time
 `otelcol.processor.metric_start_time` accepts metrics from other `otelcol` components and sets the start time for cumulative metric datapoints which do not already have a start time.
 This processor is commonly used with `otelcol.receiver.prometheus`, which produces metric points without a [start time][otlp-start-time].
 
+Grafana Mimir supports OTLP metric start times only if it is configured with `-distributor.otel-created-timestamp-zero-ingestion-enabled`.
```
clayton-cornell:
```suggestion
Grafana Mimir ingests OTLP metric start times only when it is configured with the -distributor.otel-created-timestamp-zero-ingestion-enabled flag.
Without this configuration, setting start times in Alloy has no effect on ingestion behavior.
```

Some clarification and something about "why it matters"

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,27 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
...
+for the translated OTLP metric datapoint if the incoming Prometheus metric lacks a ["created time"][om-suffixes].
```
clayton-cornell:
```suggestion
In Alloy v1.14.0, `otelcol.receiver.prometheus` no longer sets OTLP metric [start times][otlp-start-time] for Prometheus metrics. The receiver forwards metrics without adding start times when the Prometheus input does not have a [created time][om-suffixes].
```

Are we implementing this in Alloy v1.14.0?

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,27 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
...
+If your database uses start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it.
+Grafana Mimir's OTLP endpoint and Grafana Cloud's OTLP Gateway both support OTLP metric start time.
+
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.14.0, 
+configure `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point`.
+
+In practice it is only necessary to use `true_reset_point` if your Mimir/Grafana Cloud instance is explicitly configured with a 
+`-distributor.otel-created-timestamp-zero-ingestion-enabled=true` flag, 
+since `true_reset_point` will send a sample with the same start and end time and Mimir by default won't create a zero sample for it.
+So for most Mimir/Grafana Cloud use cases there would be no change in behaviour even if no `otelcol.processor.metric_start_time` is added.
```
clayton-cornell:
```suggestion
This change reflects how Prometheus metrics represent start times and avoids creating values that are not present in the input. It can affect downstream systems that rely on start times to interpret metric resets or cumulative values. For example, Grafana Mimir ingests OTLP metric start times only when it runs with the `-distributor.otel-created-timestamp-zero-ingestion-enabled` flag enabled.

If your setup depends on start times for correct ingestion or query behavior, configure `otelcol.processor.metric_start_time` with `strategy = "true_reset_point"` to restore the previous behavior.
```

Could we rework and simplify this information? The original felt like it wandered around a lot without really getting to the point.  This change explains why the change was made (this was missing)... and contains an actionable solution if the change negatively affects the users.

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,27 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 `otelcol.receiver.prometheus` is a custom component built on a fork of the upstream OpenTelemetry receiver.
 {{< /admonition >}}
 
+
+{{< admonition type="warning" >}}
```
clayton-cornell:
```suggestion
{{< admonition type="note" >}}
```

I don't think this is suitable as a warning. Warnings are for data loss, incorrect data, or negative behavior by default.  This is a configuration-dependant behaviour change .... correct?

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
ptodev: Yes

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
ptodev: > The original felt like it wandered around a lot without really getting to the point.

It's because start times are a complicated topic and most people won't be familiar with them. We need to provide background information in order for the explanation to make sense to most readers. It also doesn't help that the code supports this only in very sporadic and incomplete ways.

Do you mind if we keep the original explanation? It's more concrete, whereas the new one is more abstract.

internal/component/otelcol/receiver/prometheus/internal/transaction_bench_test.go
kgeckhart, line 1: Is this worth leaving around commented out or should we remove it?

internal/component/otelcol/receiver/prometheus/internal/transaction.go
```diff
@@ -285,17 +292,20 @@ func (t *transaction) AppendExemplar(_ storage.SeriesRef, l labels.Labels, e exe
...
 	}
 
+	var schema int32
+	if h != nil {
+		schema = h.Schema
+	} else if fh != nil {
+		schema = fh.Schema
+	}
 	t.addingNativeHistogram = true
+	t.addingNHCB = schema == -53
```
kgeckhart: Is this -53 documented and/or exposed as a const somewhere we could reference? If not could we make one with some comments about it to reduce how much of a magic number it is?

internal/component/otelcol/receiver/prometheus/internal/transaction.go
```diff
@@ -332,21 +342,38 @@ func (t *transaction) AppendHistogram(_ storage.SeriesRef, ls labels.Labels, atM
...
+	}
 	if err != nil {
 		t.logger.Warn("failed to add histogram datapoint", zap.Error(err), zap.String("metric_name", metricName), zap.Any("labels", ls))
+		// never return errors, as that fails the while scrape
+		// return ref==0 indicating that the series was not added
+		return 0, nil
 	}
 
-	return 0, nil // never return errors, as that fails the whole scrape
+	// never return errors, as that fails the whole scrape
```
kgeckhart:
```suggestion
```

Since there's no possible error to return here

internal/component/otelcol/receiver/prometheus/internal/transaction.go
```diff
@@ -332,21 +342,38 @@ func (t *transaction) AppendHistogram(_ storage.SeriesRef, ls labels.Labels, atM
...
+	// return ref==1 indicating that the series was added and needs staleness tracking
```
kgeckhart, line 357: Is this part of a spec somewhere we should reference? Just knowing what I do about seriesrefs it feels :suspect: to always return the same non-zero ref.

internal/component/otelcol/receiver/prometheus/internal/transaction.go
```diff
@@ -332,21 +342,38 @@ func (t *transaction) AppendHistogram(_ storage.SeriesRef, ls labels.Labels, atM
...
 
-	err = curMF.addExponentialHistogramSeries(t.getSeriesRef(ls, curMF.mtype), metricName, ls, atMs, h, fh)
+	if schema == -53 {
+		err = curMF.addNHCBSeries(t.getSeriesRef(ls, curMF.mtype), metricName, ls, atMs, h, fh)
+	} else {
+		err = curMF.addExponentialHistogramSeries(t.getSeriesRef(ls, curMF.mtype), metricName, ls, atMs, h, fh)
+	}
 	if err != nil {
 		t.logger.Warn("failed to add histogram datapoint", zap.Error(err), zap.String("metric_name", metricName), zap.Any("labels", ls))
+		// never return errors, as that fails the while scrape
```
kgeckhart:
```suggestion
		// Individual datapoint failures should not return an error, as that fails the whole scrape
```

internal/component/otelcol/receiver/prometheus/internal/staleness_end_to_end_test.go
```diff
@@ -3,17 +3,45 @@
 
 package internal_test
 
-// TODO: Uncomment this in the future when there's no more dependency hell for the
```
kgeckhart, line 6: I see this set of commented out tests had a TODO, should it remain? Should these tests still be here?

internal/component/otelcol/receiver/prometheus/internal/transaction.go
dehaansa: https://github.com/prometheus/prometheus/blob/main/model/histogram/generic.go#L28 is the only thing I could find.

internal/component/otelcol/receiver/prometheus/internal/transaction_bench_test.go
ptodev: I think there's value in keeping it with a comment that says why it's commented out. That way someone won't wonder about it the next time this code is synced.

I added a comment at the top of the file.

internal/component/otelcol/receiver/prometheus/internal/staleness_end_to_end_test.go
ptodev: Similarly to the other file, I added a comment to say why the tests are commented out.

internal/component/otelcol/receiver/prometheus/internal/transaction.go
ptodev: Yes, the location Sam linked to is where this is defined. I don't know if it's in any documents.

internal/component/otelcol/receiver/prometheus/internal/transaction.go
ptodev: I don't think "1" is an official Prometheus-approved magic number. Judging by [the upstream PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43925) it is just a fake ID which is unlikely to be used in practice.

internal/component/otelcol/receiver/prometheus/internal/transaction.go
kgeckhart: 👍 could we reference that constant or perhaps IsCustomBucketsSchema? Both look to be exported from that package

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
clayton-cornell: Sounds good. Leave it as-is.

internal/component/otelcol/receiver/prometheus/internal/transaction.go
ptodev: Yes, good idea. It's done now.

kgeckhart: There are other references to -53, should we adjust those too? Edit: I think this one for sure, https://github.com/grafana/alloy/pull/5308/changes#diff-a02b23d6f359465c78be6038b7e415ebfd19774b5acbec4d4c0c5bab68b5cbb6R345, the tests hard code it too but that might be fine.

ptodev: Sorry, you're right, I should have searched the whole folder. I changed those occurrences, including the tests, because in the tests this was used in the input and not in the output.

docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
```diff
@@ -13,6 +13,9 @@ title: otelcol.processor.metric_start_time
 `otelcol.processor.metric_start_time` accepts metrics from other `otelcol` components and sets the start time for cumulative metric datapoints which do not already have a start time.
 This processor is commonly used with `otelcol.receiver.prometheus`, which produces metric points without a [start time][otlp-start-time].
 
+Grafana Mimir ingests OTLP metric start times only when it is configured with the -distributor.otel-created-timestamp-zero-ingestion-enabled flag.
```
clayton-cornell:
```suggestion
Grafana Mimir ingests OTLP metric start times only when it is configured with the `-distributor.otel-created-timestamp-zero-ingestion-enabled` flag.
```

docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
```diff
@@ -13,6 +13,9 @@ title: otelcol.processor.metric_start_time
...
+Without this configuration, setting start times in Alloy has no effect on ingestion behavior.
```
clayton-cornell:
```suggestion
Without this configuration, setting start times in {{< param "PRODUCT_NAME" >}} has no effect on ingestion behavior.
```

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,26 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
 `otelcol.receiver.prometheus` is a custom component built on a fork of the upstream OpenTelemetry receiver.
 {{< /admonition >}}
 
+
+{{< admonition type="note" >}}
+In Alloy v1.14.0, `otelcol.receiver.prometheus` no longer sets OTLP metric [start times][otlp-start-time] for Prometheus metrics. The receiver forwards metrics without adding start times when the Prometheus input does not have a [created time][om-suffixes].
```
clayton-cornell:
```suggestion
In {{< param "PRODUCT_NAME" >}} v1.14.0, `otelcol.receiver.prometheus` no longer sets OTLP metric [start times][otlp-start-time] for Prometheus metrics. The receiver forwards metrics without adding start times when the Prometheus input does not have a [created time][om-suffixes].
```

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,26 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
...
+
+Start time is a way to tell when a cumulative metric such as an OTLP "sum" or a Prometheus "counter" was last reset.
+If your database uses start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it.
+Grafana Mimir's OTLP endpoint and Grafana Cloud's OTLP Gateway both support OTLP metric start time.
+
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.14.0,
```
clayton-cornell:
```suggestion
To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to {{< param "PRODUCT_NAME" >}} v1.14.0, 
```

docs/sources/reference/components/otelcol/otelcol.receiver.prometheus.md
```diff
@@ -22,6 +22,26 @@ You can specify multiple `otelcol.receiver.prometheus` components by giving them
...
+If your database uses start times for OTLP metrics, you can use `otelcol.processor.metric_start_time` to set it.
+Grafana Mimir's OTLP endpoint and Grafana Cloud's OTLP Gateway both support OTLP metric start time.
+
+To add the start time in the same way that `otelcol.receiver.prometheus` did it prior to Alloy v1.14.0, 
+configure `otelcol.processor.metric_start_time` with `strategy` set to `true_reset_point`.
+
+In practice it is only necessary to use `true_reset_point` if your Mimir/Grafana Cloud instance is explicitly configured with a 
+`-distributor.otel-created-timestamp-zero-ingestion-enabled=true` flag, 
+since `true_reset_point` will send a sample with the same start and end time and Mimir by default won't create a zero sample for it.
+So for most Mimir/Grafana Cloud use cases there would be no change in behaviour even if no `otelcol.processor.metric_start_time` is added.
```
clayton-cornell:
```suggestion
For most Mimir and Grafana Cloud use cases there would be no change in behaviour even if no `otelcol.processor.metric_start_time` is added.
```

Issue #5254, opened 2026-01-13
Metadata support prometheus.receive_http


kgeckhart:
### Add a description

Support sending metric metadata in the `prometheus.receive_http` component.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

Issue #5036, opened 2025-12-09
Expose the new appendMetadata setting in prometheus


thampiotr:
### Add a description

e.g. in `promremote.NewWriteHandler` https://github.com/prometheus/prometheus/blob/main/storage/remote/write_handler.go#L68

In Alloy, for example in receive_http.go, we don't expose the appendMetadata option to users.

Also the scrape manager reference in crdmanager 

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

Issue #4659, opened 2025-10-23
Expose the experimental enabling of type and unit labels to metrics


thampiotr:
### Add a description

There is a new experimental option to add type and unit labels to metrics. Expose this in alloy components.

`prometheus.scrape`, `prometheus.receive_http`, and the operator

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5317, opened 2026-01-21
feat(prometheus.receive_http): Add experimental support for AppendMetadata and TypeAndUnitLabels

kgeckhart:
Adds support for two upstream arguments, `AppendMetadata`, and `EnableTypeAndUnitLabels`. These options only apply to receiving Remote Write V2 payloads support for receiving Remote Write V2 was also added to the component as well.

Resolves: https://github.com/grafana/alloy/issues/5036
Related to: https://github.com/grafana/alloy/issues/4659
Resolves: https://github.com/grafana/alloy/issues/5254

### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/prometheus/prometheus.scrape.md
```diff
@@ -97,8 +97,6 @@ You can use the following arguments with `prometheus.scrape`:
 > * `prometheus.remote_write` only when configured for Remote Write v2.
 > * `prometheus.write_queue`
 > 
-> Metadata support for Remote Write v1 in `prometheus.remote_write` will be added soon.
```
kgeckhart: I removed this because it's unclear when support will come ATM so suggesting it will be soon feels a bit misleading.

internal/component/prometheus/receive_http/receive_http.go
```diff
@@ -86,9 +91,9 @@ func New(opts component.Options, args Arguments) (*Component, error) {
 			opts.Registerer,
 			fanout,
 			supportedRemoteWriteProtoMsgs,
-			ingestCTZeroSample,
-			enableTypeAndUnitLabels,
-			appendMetadata,
+			false,
```
blewis12, line 115: majorly nit but I feel like keeping `ingestCTZeroSample := false` makes it clear what this `false` is doing

internal/component/prometheus/receive_http/receive_http_test.go
```diff
@@ -615,3 +645,252 @@ func getFreePort(t *testing.T) int {
...
+				{ts: timestamp, val: 100, l: labels.FromStrings("__name__", "http_requests_total", "__type__", "counter", "__unit__", "requests", "method", "GET", "status", "200")},
+				{ts: timestamp + 1, val: 150, l: labels.FromStrings("__name__", "http_requests_total", "__type__", "counter", "__unit__", "requests", "method", "GET", "status", "200")},
+				{ts: timestamp, val: 45.2, l: labels.FromStrings("__name__", "cpu_usage_seconds", "__type__", "gauge", "__unit__", "seconds", "cpu", "0")},
+			},
+			expectedMetadata: nil, // No metadata should be forwarded
+		},
+		{
+			name:                    "with both metadata and type/unit labels",
+			appendMetadata:          true,
+			enableTypeAndUnitLabels: true,
```
blewis12, line 758: would it be worth adding a test case where both are `false`? Perhaps this is the same behavior as before, but this test is specifically about writev2 which is not covered in the other test cases I don't think

internal/component/prometheus/receive_http/receive_http.go
kgeckhart: I added a comment on this, I removed the original variable because the feature is deprecated and I didn't want it to seem like we should add support for it.

blewis12: makes sense! just good to have the visibility

internal/component/prometheus/receive_http/receive_http.go
```diff
@@ -67,17 +72,17 @@ func New(opts component.Options, args Arguments) (*Component, error) {
...
 
-	// TODO: Make these configurable in the future:
-	// TODO: Expose ingestCTZeroSample: https://github.com/grafana/alloy/issues/4045
-	// TODO: Expose enableTypeAndUnitLabels: https://github.com/grafana/alloy/issues/4659
-	// TODO: Expose appendMetadata: https://github.com/grafana/alloy/issues/5036
-	supportedRemoteWriteProtoMsgs := remote.MessageTypes{remote.WriteV1MessageType}
-	ingestCTZeroSample := false
-	enableTypeAndUnitLabels := false
-	appendMetadata := false
+	supportedRemoteWriteProtoMsgs := remote.MessageTypes{remote.WriteV1MessageType, remote.WriteV2MessageType}
```
dehaansa: Should v2 here be experimental?

kgeckhart: I wasn't sure about this one, if I did do it then I would probably need to expose the option in our config to set which message types are allowed. I didn't want to make the config even more confusing so decided against it. WDYT?

dehaansa: I'm hesitant to support it without experimental since it's experimental in other Alloy components that support it and I don't want to confuse users.

I agree that if we make it experimental we would want it to be clear to users which message formats are accepted and likely make them configure it explicitly.

docs/sources/reference/components/prometheus/prometheus.receive_http.md
clayton-cornell:
```suggestion
prometheus.receive_http "<LABEL>" {
```

docs/sources/reference/components/prometheus/prometheus.receive_http.md
```diff
@@ -35,16 +35,31 @@ prometheus.receive_http "<LABEL?" {
 The component starts an HTTP server supporting the following endpoint:
 
 * `POST /api/v1/metrics/write`: Sends metrics to the component, which in turn is forwarded to the receivers as configured in `forward_to` argument.
-  The request format must match that of [Prometheus `remote_write` API][prometheus-remote-write-docs].
+  The request format must match that of the [Prometheus `remote_write` API][prometheus-remote-write-docs], supporting both remote write v1 and v2 formats.
```
clayton-cornell:
```suggestion
  The request format must be compatible with the [Prometheus remote_write API][prometheus-remote-write-docs] and can use either the v1 or v2 format.
```

docs/sources/reference/components/prometheus/prometheus.receive_http.md
```diff
@@ -35,16 +35,31 @@ prometheus.receive_http "<LABEL?" {
...
+
+| Name                                      | Type                    | Description                                                                  | Default                      | Required |
+| ----------------------------------------- | ----------------------- | ---------------------------------------------------------------------------- | ---------------------------- | -------- |
+| `forward_to`                              | `list(MetricsReceiver)` | List of receivers to send metrics to.                                        |                              | yes      |
+| `append_metadata`                         | `bool`                  | Whether metric metadata should be passed to downstream components.           | `false`                      | no       |
+| `enable_type_and_unit_labels`             | `bool`                  | Whether metric type and unit should be added as labels to the metric.        | `false`                      | no       |
+| `accepted_remote_write_protobuf_messages` | `list(string)`          | Accepted remote write protobuf message types.                                | `["prometheus.WriteRequest"]` | no       |
```
clayton-cornell:
```suggestion
| Name                                      | Type                    | Description                                                           | Default                       | Required |
| ----------------------------------------- | ----------------------- | --------------------------------------------------------------------- | ----------------------------- | -------- |
| `forward_to`                              | `list(MetricsReceiver)` | List of receivers to send metrics to.                                 |                               | yes      |
| `accepted_remote_write_protobuf_messages` | `list(string)`          | Accepted remote write protobuf message types.                         | `["prometheus.WriteRequest"]` | no       |
| `append_metadata`                         | `bool`                  | Pass metric metadata to downstream components.                        | `false`                       | no       |
| `enable_type_and_unit_labels`             | `bool`                  | Add the metric type and unit as labels to the metric.                 | `false`                       | no       |
```

Slight tweak to table column alignment

Alpha sort the rows, required first.

I also tweaked the descriptions for the booleans to use a direct imperative. It's a bool, so we know that if set to true, then it enables the described behaviour.

dehaansa (approved):

PR #5321, opened 2026-01-22
feat: Add a `honor_metadata` config argument to `prometheus.operator` components

### LINKED ISSUE #4615: Add honor_metadata to prometheus.operator.servicemonitors
### Component(s)

prometheus.operator.servicemonitors

### Request

To get the [recently added](https://github.com/grafana/alloy/pull/4098) `honor_metadata` added to `prometheus.operator.servicemonitors`

### Use case

To have metadata forwarded to `otelcol.receiver.prometheus` and onward to Mimir with `otelcol.exporter.otlphttp` after processing.

### Tip

ptodev:
### Pull Request Details

Currently this only works if Remote Write v2 is enabled. This is done in the `prometheus.remote_write` component:

```hcl
  endpoint {
    url = "http://mimir-nginx.mimir-test.svc:80/api/v1/push"
    protobuf_message = "io.prometheus.write.v2.Request"
    send_native_histograms = true
  }
}
```

Fixes #4615

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/k8s/tests/prometheus-operator/testdata/kubernetes.yml
+++ b/integration-tests/k8s/tests/prometheus-operator/testdata/kubernetes.yml
@@ -298,7 +298,7 @@ data:
 
     // Test prometheus.operator.servicemonitors
     prometheus.operator.servicemonitors "default" {
-      forward_to = [prometheus.remote_write.mimir.receiver]
+      forward_to = [prometheus.relabel.servicemonitors.receiver]
       namespaces = ["testing"]
       scrape {
         honor_metadata = true
@@ -308,6 +308,10 @@ data:
           alloy = "yes",
         }
       }
+    }
+
+    prometheus.relabel "servicemonitors" {
+      forward_to = [prometheus.remote_write.mimir.receiver]
       rule {
         action       = "replace"
         target_label = "alloy_int_test"
@@ -325,7 +329,7 @@ data:
 
     // Test prometheus.operator.podmonitors
     prometheus.operator.podmonitors "default" {
-      forward_to = [prometheus.remote_write.mimir.receiver]
+      forward_to = [prometheus.relabel.podmonitors.receiver]
       namespaces = ["testing"]
       scrape {
         honor_metadata = true
@@ -335,6 +339,10 @@ data:
           alloy = "yes",
         }
       }
+    }
+
+    prometheus.relabel "podmonitors" {
+      forward_to = [prometheus.remote_write.mimir.receiver]
       rule {
         action       = "replace"
         target_label = "alloy_int_test"
@@ -352,7 +360,7 @@ data:
 
     // Test prometheus.operator.probes
     prometheus.operator.probes "default" {
-      forward_to = [prometheus.remote_write.mimir.receiver]
+      forward_to = [prometheus.relabel.probes.receiver]
       namespaces = ["testing"]
       scrape {
         honor_metadata = true
@@ -362,6 +370,10 @@ data:
           alloy = "yes",
         }
       }
+    }
+
+    prometheus.relabel "probes" {
+      forward_to = [prometheus.remote_write.mimir.receiver]
       rule {
         action       = "replace"
         target_label = "alloy_int_test"
@@ -379,7 +391,7 @@ data:
 
     // Test prometheus.operator.scrapeconfigs
     prometheus.operator.scrapeconfigs "default" {
-      forward_to = [prometheus.remote_write.mimir.receiver]
+      forward_to = [prometheus.relabel.scrapeconfigs.receiver]
       namespaces = ["testing"]
       scrape {
         honor_metadata = true
@@ -389,6 +401,10 @@ data:
           alloy = "yes",
         }
       }
+    }
+
+    prometheus.relabel "scrapeconfigs" {
+      forward_to = [prometheus.remote_write.mimir.receiver]
       rule {
         action       = "replace"
         target_label = "alloy_int_test"


integration-tests/k8s/util/mimir.go
ptodev, line 1: The test util package could use further refactoring, but I don't want to do it on this PR to keep the diffs more simple. I used similar functions to the ones in the Docker integration tests, but later on we should use one shared function.

docs/sources/shared/reference/components/prom-operator-scrape.md
```diff
@@ -9,3 +9,4 @@ headless: true
 | `default_scrape_interval`  | `duration` | The default interval between scraping targets. Used as the default if the target resource doesn't provide a scrape interval. | `1m`    | no       |
 | `default_scrape_timeout`   | `duration` | The default timeout for scrape requests. Used as the default if the target resource doesn't provide a scrape timeout.        | `10s`   | no       |
 | `scrape_native_histograms` | `bool`     | Whether to scrape native histograms from targets.                                                                            | `false` | no       |
+| `honor_metadata`           | `bool`     | (Experimental) Indicator whether metric metadata should be sent to downstream components.                                    | `false` | no       |
```
kgeckhart: I think this might need some of the extra callouts we have with `prometheus.scrape`

internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -149,26 +155,39 @@ func (c *crdManager) Run(ctx context.Context) error {
...
 	go func() {
-		err := c.scrapeManager.Run(targetSetsChan)
+		err := scrapeManager.Run(targetSetsChan)
 		level.Info(c.logger).Log("msg", "scrape manager stopped")
 		if err != nil {
 			level.Error(c.logger).Log("msg", "scrape manager failed", "err", err)
 		}
 	}()
 
+	// Assign managers under lock to avoid races with apply() and other methods
```
kgeckhart: Was this an existing problem that the tests caught or just an extra safety measure?

internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -118,13 +118,19 @@ func newCrdManager(opts component.Options, cluster cluster.Cluster, logger log.L
...
 func (c *crdManager) Run(ctx context.Context) error {
-	restConfig, err := c.args.Client.BuildRESTConfig(c.logger)
-	if err != nil {
-		return fmt.Errorf("creating rest config: %w", err)
-	}
-	c.client, err = kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		return fmt.Errorf("creating kubernetes client: %w", err)
+	// Build REST config (used for both k8s client and informers)
+	// Only build if client is not already set (i.e., not injected by tests)
```
kgeckhart: WDYT about `newCrdManager` accepting a bool to enable/disable informers + accept a client? The real manager injects true + nil allowing the client to be created with informers and your test manager can provide the test client + false. 

This way when you create a `new` manager you are being very explicit about the behavior vs allowing the behavior to be implicitly dictated by the client field being mutated after you have a newly constructed manager.

internal/component/prometheus/operator/servicemonitors/servicemonitors_test.go
```diff
@@ -0,0 +1,266 @@
...
+			}
+		},
+	}
+
+	// Create component arguments
+	var args operator.Arguments
+	args.SetToDefault()
+	args.ForwardTo = []storage.Appendable{mockAppendable}
+	args.Namespaces = []string{"monitoring"}
+	args.Scrape.HonorMetadata = true
```
kgeckhart: Would it be possible to parameterize this to allow testing with HonorMetadata false and true?

integration-tests/k8s/util/mimir.go
kgeckhart: I was certainly getting a fair amount of dejavu looking at some of the util code here as well 😅

docs/sources/shared/reference/components/prom-operator-scrape.md
ptodev: True! Thanks! I had forgotten about those.

internal/component/prometheus/operator/common/crdmanager.go
ptodev: I don't think it's a production problem. I think it's only a problem for the tests, because the tests start trying to add k8s resources as soon as Run() starts.

In the production code path the informers are started differently and from what I can tell this won't be a problem.

I should probably clarify this in the comment?

internal/component/prometheus/operator/common/crdmanager.go
ptodev: I try to avoid adding test-specific struct fields. That said, I agree that leveraging already existing fields as a switch is also not good.

On Monday I'll have another go at standardising the prod and test paths a bit more. I was really hoping to only fake the k8s dependencies and reuse all other production code, including the informers. I'll timebox it and if I don't find a good way then I'll implement your suggestion.

docs/sources/shared/reference/components/prom-operator-scrape.md
```diff
@@ -9,3 +9,21 @@ headless: true
...
+> Enabling it may increase resource consumption, particularly if a lot of metrics with different names are ingested.
+> Not all downstream components may be compatible with Prometheus metadata yet.
+> Currently, the compatible components are:
+>
+> * `otelcol.receiver.prometheus`
+> * `prometheus.remote_write` only when configured for Remote Write v2.
+> * `prometheus.write_queue`
+> 
+> Metadata support for Remote Write v1 in `prometheus.remote_write` will be added soon.
+>
```
clayton-cornell:
```suggestion
> 
```

Remove this. Documentation should never promise future features or enhancements.

docs/sources/shared/reference/components/prom-operator-scrape.md
```diff
@@ -9,3 +9,21 @@ headless: true
 | `default_scrape_interval`  | `duration` | The default interval between scraping targets. Used as the default if the target resource doesn't provide a scrape interval. | `1m`    | no       |
 | `default_scrape_timeout`   | `duration` | The default timeout for scrape requests. Used as the default if the target resource doesn't provide a scrape timeout.        | `10s`   | no       |
 | `scrape_native_histograms` | `bool`     | Whether to scrape native histograms from targets.                                                                            | `false` | no       |
+| `honor_metadata`           | `bool`     | (Experimental) Indicator whether metric metadata should be sent to downstream components.                                    | `false` | no       |
+
+> **EXPERIMENTAL**: The `honor_metadata` argument is an [experimental][] feature.
+> Enabling it may increase resource consumption, particularly if a lot of metrics with different names are ingested.
+> Not all downstream components may be compatible with Prometheus metadata yet.
+> Currently, the compatible components are:
```
clayton-cornell:
```suggestion
> **EXPERIMENTAL**: The `honor_metadata` argument is an [experimental][] feature.
> If you enable this argument, resource consumption may increase, particularly if you ingest many metrics with different names.
> Some downstream components aren't compatible with Prometheus metadata.
> The following components are compatible:
```

docs/sources/shared/reference/components/prom-operator-scrape.md
```diff
@@ -9,3 +9,21 @@ headless: true
 | `default_scrape_interval`  | `duration` | The default interval between scraping targets. Used as the default if the target resource doesn't provide a scrape interval. | `1m`    | no       |
 | `default_scrape_timeout`   | `duration` | The default timeout for scrape requests. Used as the default if the target resource doesn't provide a scrape timeout.        | `10s`   | no       |
 | `scrape_native_histograms` | `bool`     | Whether to scrape native histograms from targets.                                                                            | `false` | no       |
+| `honor_metadata`           | `bool`     | (Experimental) Indicator whether metric metadata should be sent to downstream components.                                    | `false` | no       |
```
clayton-cornell:
```suggestion
| `honor_metadata`           | `bool`     | (Experimental) Indicates whether to send metric metadata to downstream components.                                          | `false` | no       |
```

There's probably more that we could do to clarify this, but... this is enough for now.

internal/component/prometheus/operator/servicemonitors/servicemonitors_test.go
ptodev: Done

internal/component/prometheus/operator/common/crdmanager.go
ptodev: Since we had this conversation I changed the code in a couple of ways.

To avoid the race condition, the test is now looking for the log line that informers started. I rolled back the change in the Run() function which populates the crdManager with a mutex, since that's no longer needed.

The informers are now faked. A factory is inserted so that the pdn and the test code paths align as much as possible.

I'm much happier with the tests now!

internal/component/prometheus/operator/common/crdmanager.go
ptodev: See [my comment above](https://github.com/grafana/alloy/pull/5321#discussion_r2733169325) for a summary the latest improvements :)

clayton-cornell:
Docs look OK :-)

PR #5322, opened 2026-01-22
fix: Invalid handling of `id` in `foreach` when using discovery components

### LINKED ISSUE #5297: foreach id doesn't work with discovery.file
### Component(s)

_No response_

### What's wrong?

The `foreach` component has an `id` parameter that's supposed to set the component name.  This works when the collection is passed inline, but it does not work when the collection comes from `discovery.file`. In this case, the `id` parameter seems to be ignored, and the component is named with a long random hash.

This example config includes 2 separate foreach blocks, one working and one not, as shown in these screenshots.

Image

Image

### Steps to reproduce

Using the attached config and run `alloy run xerus2.alloy --stability.level experimental`.

### System information

_No response_

### Software version

Grafana Alloy v1.12.2

### Configuration

```text
discovery.file "xerus_debug" {
  files = ["debug/alloy_targets/servertech_xerus.yml"]
}

foreach "xerus_debug_ok"  {
  collection = [
    {
      "__address__" = "10.59.31.124",
      "netbox_id" = "43356",
      "instance" = "10.59.31.124",
    },
  ]
  var = "each"
  id = "netbox_id"
  template {
    prometheus.scrape "xerus_debug" {
      forward_to = [prometheus.remote_write.debug.receiver]
      targets         = [each]
    }
  }
}

foreach "xerus_debug_nok"  {
  collection = discovery.file.xerus_debug.targets
  var = "each"
  id = "netbox_id"
  template {
    prometheus.scrape "xerus_debug" {
      forward_to = [prometheus.remote_write.debug.receiver]
      targets         = [each]
    }
  }
}

prometheus.remote_write "debug" {
  endpoint {
      url = "https://prometheus-prod-13-prod-us-east-0.grafana.net/api/prom/push"
      basic_auth {
        username = sys.env("PROMETHEUS_USER")
        password = sys.env("PROMETHEUS_TOKEN")
      }
  }
}
```

### Logs

```text

```

### Tip

thampiotr:

In the `foreach` close there is an option to use `id` for each element off the `collection`. This requires looking at the element conceptually as a map and extracting the required `id`. But in Alloy unfortunately there are many ways of expressing an conceptual map. The key ways are:

- **`map[string]any`**: Inline object literals with simple values.
  - Example: `collection = [{name = "one", port = "8080"}, {name = "two", port = "8081"}]`

- **`map[string]string`**: Plain Go maps - used to be common, but are now replaced by Target capsules for performance. We keep it for maximum compatibility in case it's needed in the future.

- **`map[string]syntax.Value`**: Inline object literals with expressions or computed values.
  - Example: `collection = [{name = "one", url = "http://" + hostname}]`

- **`syntax.ConvertibleIntoCapsule`**: Capsules from component exports, such as `discovery.Target`.
  - Example: `collection = discovery.kubernetes.pods.targets`

Previously we didn't support or test them all. Now we do.

Fixes #5297

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/runtime/internal/controller/node_config_foreach.go
+++ b/internal/runtime/internal/controller/node_config_foreach.go
@@ -475,6 +475,7 @@ func collectionItemID(item any, key string, logger log.Logger) (any, bool) {
 		// Example: collection = discovery.kubernetes.pods.targets
 		return collectionItemIDFromCapsule(value, key, logger)
 	default:
+		level.Debug(logger).Log("msg", "unsupported collection item type encountered in foreach", "item", fmt.Sprintf("%#v", item))
 		return nil, false
 	}
 }


internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -444,6 +441,62 @@ func objectFingerprint(id any, hashId bool) string {
...
+			logMissingCollectionID(logger, key)
+			return nil, false
+		}
+		return val.Interface(), true
+	case syntax.ConvertibleIntoCapsule:
+		// Capsules from component exports, such as discovery.Target.
+		// Example: collection = discovery.kubernetes.pods.targets
+		return collectionItemIDFromCapsule(value, key, logger)
+	default:
+		return nil, false
```
ptodev, line 479: Is it worth having a debug log line or a metric for unsupported types?

internal/runtime/internal/controller/node_config_foreach.go
```diff
@@ -444,6 +441,62 @@ func objectFingerprint(id any, hashId bool) string {
...
+	case map[string]any:
+		// Inline object literals with simple values.
+		// Example: collection = [{name = "one", port = "8080"}, {name = "two", port = "8081"}]
+		val, ok := value[key]
+		if !ok {
+			logMissingCollectionID(logger, key)
+			return nil, false
+		}
+		return val, true
+	case map[string]string:
```
ptodev, line 455: Isn't there a generic way to return any value as `any`, so that we don't have to list all the combinations? Then we only need to check for `syntax.ConvertibleIntoCapsule`.

internal/runtime/internal/controller/node_config_foreach.go
thampiotr: good idea, I've added it

internal/runtime/internal/controller/node_config_foreach.go
thampiotr: Not without using reflection, which I would prefer not to use in this context.

Issue #5272, opened 2026-01-16
Integration test improvments


kalleep:
### Add a description

In our [integration tests](https://github.com/grafana/alloy/tree/main/internal/cmd/integration-tests) we currently hardcode special setups that different tests needs:
1. [beyla](https://github.com/grafana/alloy/blob/main/internal/cmd/integration-tests/utils.go#L113)
2. [loki-enrich](https://github.com/grafana/alloy/blob/main/internal/cmd/integration-tests/utils.go#L122)
3. [ loki-file-rotation](https://github.com/grafana/alloy/pull/5271)

Things like adding mounts, exposing ports etc could be useful for new tests and is annoying having to hardcode these things.

I think we should add something like a test.yaml file that we can describe certain setup needed for a test, e.g. having a section that list ports we want to expose. Thay way we could make these parts reusable.

ptodev:
SGTM 👍 A yaml file inside each test directory would be ok.

kalleep:
Great I already have something working. Just want to get https://github.com/grafana/alloy/pull/5271 merged first

PR #5328, opened 2026-01-22
chore: Add test.yaml to integration tests

kalleep:
### Pull Request Details
Every integration test can now optionally add test.yaml.

Right now I just added the config we need to support to make all test work. I explored using something like https://github.com/compose-spec/compose-go and the types defined in there, essentially supporting compose server config.

But I think this is simpler and fine for now.

Fixes: #5272

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/docker/utils.go
+++ b/integration-tests/docker/utils.go
@@ -152,12 +152,12 @@ func createContainerRequest(dirName, testDir string, port int, networkName strin
 }
 
 // Configure the test command with appropriate environment variables if needed
-func setupTestCommand(testDir string, testTimeout time.Duration) (*exec.Cmd, error) {
+func setupTestCommand(testDir string, testTimeout time.Duration) *exec.Cmd {
 	testCmd := exec.Command("go", "test", "-tags", "alloyintegrationtests")
 	testCmd.Dir = testDir
 
 	testCmd.Env = append(testCmd.Environ(), fmt.Sprintf("%s=%s", common.TestTimeout, testTimeout.String()))
-	return testCmd, nil
+	return testCmd
 }
 
 var logMux sync.Mutex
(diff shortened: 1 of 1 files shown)


integration-tests/docker/config.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+// TestConfig is used by tests to describe special setup requirements.
+type TestConfig struct {
+	Container ContainerConfig `yaml:"container"`
```
kgeckhart: Nit: It takes some internal knowledge of our integration tests to know the container config referenced here is the alloy container. 

```suggestion
	Container ContainerConfig `yaml:"alloy_container"`
```

integration-tests/docker/README.md
```diff
@@ -30,7 +30,8 @@ Follow these steps to add a new integration test to the project:
 2. Create a new directory under the tests directory to house the files for the new test.
 3. Within the new test directory, create a file named `config.alloy` to hold the pipeline configuration you want to test.
 4. Create a `_test.go` file within the new test directory. This file should contain the Go code necessary to run the test and verify the data processing through the pipeline. All test file should have `alloyintegrationtests` as a build tag.
-5. Ensure any data is written with a unique `test_name` label that matches your assertions. 
+5. If a test folder contains test.yaml this file will be read and parsed. This file can specify different setup requirements that the test have like mount or port mapping.
```
kgeckhart: nit: WDYT about having a link to the config struct? This way it's easy to get to the config options.

integration-tests/docker/utils.go
```diff
@@ -86,19 +87,64 @@ func prepareContainerFiles(absTestDir string) ([]testcontainers.ContainerFile, [
...
+			hc.CapAdd = cfg.Container.CapAdd
+			hc.SecurityOpt = cfg.Container.SecurityOpt
+			hc.PidMode = container.PidMode(cfg.Container.PIDMode)
+
+			if dirName == "beyla" {
+				hc.Privileged = true
+				hc.CapAdd = []string{"SYS_ADMIN", "SYS_PTRACE", "SYS_RESOURCE"}
+				hc.SecurityOpt = []string{"apparmor:unconfined"}
+				hc.PidMode = container.PidMode("host")
+			}
```
kgeckhart: I think we can drop this one now, right?

kalleep: Yes, missed that part :see_no_evil:

PR #5329, opened 2026-01-22
chore: Use shared source library for loki.source.journal

kalleep:
### Pull Request Details
Similiar to other prs. Use shared components we have, Fanout and Drain. We no longer need to drain while updating tailer because we are always reading.

### Notes to the Reviewer

I moved "target" to journal package and renamed to tailer.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/journal/metrics.go
+++ b/internal/component/loki/source/journal/metrics.go
@@ -1,3 +1,5 @@
+//go:build linux && cgo && promtail_journal_enabled
+
 package journal
 
 // This code is copied from Promtail (https://github.com/grafana/loki/commit/954df433e98f659d006ced52b23151cb5eb2fdfa) with minor edits. The target package is used to


internal/component/loki/source/journal/journal.go
```diff
@@ -86,59 +90,47 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
-			for _, r := range c.args.Receivers {
-				select {
-				case <-ctx.Done():
-					return nil
-				case r.Chan() <- lokiEntry:
-				}
+
+	var wg sync.WaitGroup
+	wg.Go(func() { source.Consume(ctx, c.recv, c.fanout) })
+	wg.Go(func() {
```
kalleep, line 110: I don't think we need to have this reload anymore. Before this we had lock contention on mutex but the fanout have it's own locking now so update should generally not be blocked by that.

kalleep:
>LGTM! would be good to add an integration test for it, in the next PR is okay

Great will do a follow up pr for that!

PR #5337, opened 2026-01-23
perf: Reduce allocations in stage.cri

kalleep:
### Pull Request Details
Before this `stage.CRI` used a "sub" pipeline to parse CRI formatted logs. It did so by running `stage.regex`, `stage.labels`, `stage.timestamp` and `stage.output`.

We used `^(?s)(?P<time>\S+?) (?P<stream>stdout|stderr) (?P<flags>\S+?) (?P<content>.*)$` as our regex. Running "sub" pipelines is not that great because we communicate with channels between the stages, every stage run in it's own goroutine and regex is not always that efficient.

I cannot find an official spec for CRI but I found the [design proposal](https://github.com/kubernetes/design-proposals-archive/blob/main/node/kubelet-cri-logging.md).

From the proposal:
1. Time is always RFC 3339Nano 
2. Stream is either `stdout` or `stderr`
3. Flags is either `F` (Full) or `P` (Partial)
4. Rest is considered log content

The previous implementation was a bit lenient in that we did not required the timestamp to be RFC 3339Nano and the regex did not force flags to be either `P` or `F`. So I kept that behavior with this handwritten parser:
1. If timestamp is not a valid RFC 3339Nano we keep the old timestamp
2. We default to `F` flag and optionally validate last white-space

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/cri/parse.go
+++ b/internal/component/loki/process/stages/cri/parse.go
@@ -36,12 +36,10 @@ const (
 // String returns the CRI stream representation ("stdout" or "stderr").
 func (s Stream) String() string {
 	switch s {
-	case StreamStdOut:
-		return "stdout"
 	case StreamStdErr:
 		return "stderr"
 	default:
-		return "unknown"
+		return "stdout"
 	}
 }
 


kalleep:
Benchmark diff:
```
goos: linux
goarch: amd64
       │ bench_old.txt │          bench_string.txt           │
       │    sec/op     │   sec/op     vs base                │
CRI-16    4393.0n ± 1%   631.2n ± 1%  -85.63% (p=0.000 n=10)

       │ bench_old.txt │          bench_string.txt          │
       │     B/op      │    B/op     vs base                │
CRI-16     1795.0 ± 0%   200.0 ± 0%  -88.86% (p=0.000 n=10)

       │ bench_old.txt │          bench_string.txt          │
       │   allocs/op   │ allocs/op   vs base                │
CRI-16     33.000 ± 0%   6.000 ± 0%  -81.82% (p=0.000 n=10)
```

internal/component/loki/process/stages/cri.go
```diff
@@ -0,0 +1,159 @@
...
+
+		// NOTE: Previous implementation used a "sub-pipeline"
+		// to parse CRI logs where the regex stage added these fields
+		// as "extracted" values so the other stages could operate on them.
+		// We don't need this anymore but it would be a breaking change to
+		// no longer set these.
+		e.Extracted[criTime] = parsed.Timestamp
+		e.Extracted[criFlags] = parsed.Flag.String()
+		e.Extracted[criStream] = parsed.Stream.String()
+		e.Extracted[criContent] = parsed.Content
```
kalleep, line 92: These are not really needed anymore because we don't have a "sub" pipeline that operates on them.

But we document these as extracted fields for the stage so we need to keep them https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagecri

internal/component/loki/process/stages/docker.go
```diff
@@ -0,0 +1,48 @@
...
+	RFC3339Nano = "RFC3339Nano"
+)
+
+// DockerConfig is an empty struct that is used to enable a pre-defined
+// pipeline for decoding entries that are using the Docker logs format.
+type DockerConfig struct{}
+
+// NewDocker creates a predefined pipeline for parsing entries in the Docker
+// json log format.
+func NewDocker(logger log.Logger, registerer prometheus.Registerer, minStability featuregate.Stability) (Stage, error) {
```
kalleep, line 19: This is untouched. Just moved it to it's own file

internal/component/loki/process/stages/extensions.go
```diff
@@ -1,209 +0,0 @@
-package stages
```
blewis12, line 1: Something I don't quite understand: why was the cri stage defined as an "extension" before? From what I gather, the new cri still inherits the same interfaces but how it runs internally is changing

kalleep: Well I don't think the filename matter that much but both docker and CRI was there because the use / used "sub" pipeline.

If you look at e.g. docker stage it is just constructing it's own pipeline using json, timestamp, labels and output stages. I guess that is what made them "extentions".

blewis12: I see, just want to make sure i'm not missing something important conceptually

internal/component/loki/process/stages/cri.go
```diff
@@ -0,0 +1,159 @@
...
+
+		e.Line = parsed.Content
+
+		ts, err := time.Parse(time.RFC3339Nano, parsed.Timestamp)
```
blewis12, line 96: Trying to wrap my head around how it was done before and what's being done now: in the timestamp stage we'd attempt to [parse the timestamp](https://github.com/grafana/alloy/blob/810331b005aea58ee97a898b11ef7883a3d4a01b/internal/component/loki/process/stages/timestamp.go#L145) and then [update the entry](https://github.com/grafana/alloy/blob/810331b005aea58ee97a898b11ef7883a3d4a01b/internal/component/loki/process/stages/timestamp.go#L152) with whatever the result was. But if for some reason we got an error we go through this ["fudge" handler.](https://github.com/grafana/alloy/blob/810331b005aea58ee97a898b11ef7883a3d4a01b/internal/component/loki/process/stages/timestamp.go#L196). Is that behavior preserved in this implementation?

kalleep: I should have called this out but IMO the fudge functionality is not really necessary for CRI logs. These logs always comes from files and when we ingest logs from files the default timestamp will be at the time we read the line.

If we for some reason cannot parse correctly formatted timestamps the timestamp this should be more correct than what we got with "fudge". Fudge will look at the last timestamp we saw and [increase it by 1 nano second](https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/timestamp.go#L207). So the time that we read the log line should be much closer than the fudged time.

internal/component/loki/process/stages/cri/parse.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return FlagFull, line
+}
+
+func skipWhitespace(line string) (string, bool) {
+	if len(line) > 0 && line[0] == ' ' {
+		return line[1:], true
+	}
+	return line, false
+}
```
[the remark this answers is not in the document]
blewis12: I was also wondering this, is there no chance that we might have more than a single whitepace?

internal/component/loki/process/stages/cri.go
blewis12: ah ok - good to know that the default ts is from when the line is read, makes sense on a high level. Thanks for explaining!

internal/component/loki/process/stages/cri/parse.go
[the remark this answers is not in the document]
blewis12: oh wait, the previous implementation also checked against single spaces I realize - and it's defined that way in our documentation. nevermind!

blewis12 (approved):
Nice :) makes sense to me, and seems like a great improvement

internal/component/loki/process/stages/cri.go
```diff
@@ -0,0 +1,159 @@
...
+		// We got full-line 'F'.
+		// If any old partial lines matches with this full-line stream, merge it,
+		// else just return the full line.
+		prev, ok := c.partialLines[fingerprint]
+		if ok {
+			var builder strings.Builder
+			builder.WriteString(prev.Line)
+			builder.WriteString(e.Line)
+			e.Line = builder.String()
+			c.ensureTruncateIfRequired(&e)
```
ptodev, line 145: Might be worth adding a todo for the future to trim the line before it is constructed?

internal/component/loki/process/stages/extensions.go
```diff
@@ -1,209 +0,0 @@
...
-			},
-		},
-		{
-			OutputConfig: &OutputConfig{
-				Source: "content",
-			},
-		},
-	}
-
-	p, err := NewPipeline(logger, base, nil, registerer, minStability)
```
ptodev, line 198: I wonder if the debug metrics of the CRI stage are different now? I suppose with the current implementation it'd inherit some of the debug metrics of the sub pipelines.

internal/component/loki/process/stages/cri/parse.go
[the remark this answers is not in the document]
kalleep: Yes exactly. Initially I did advance over all white-spaces until I realized that we did not do that before and it's fine not to

ptodev (commented):
I'm happy that the PR is merged, but I just saw it and had a few minor thoughts

internal/component/loki/process/stages/cri.go
kalleep: Maybe but we did not do that before either

internal/component/loki/process/stages/extensions.go
kalleep: No there was no debug metrics for any of the stages used in the sub pipeline and I don't see any other metrics that would be affected

PR #5342, opened 2026-01-23
feat: Add an `otelcol.processor.metric_start_time` component

ptodev:
### Pull Request Details

This component creates "start time" timestamps for OTLP metrics. Those timestamps are similar to the `_created` metrics in the [Prometheus spec](https://prometheus.io/docs/specs/om/open_metrics_spec/). AFAIK they are not widely used.

This component is required in order for the translator package in `otelcol.receiver.prometheus` to be updated (#5308). The new translator outsources the responsibility of creating start timestamps to this new processor.

The upstream component is called [metricstarttime](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.142.0/processor/metricstarttimeprocessor). I called this one `metric_start_time`, because there is an [open issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339) upstream to rename it.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
+++ b/docs/sources/reference/components/otelcol/otelcol.processor.metric_start_time.md
@@ -1,7 +1,5 @@
 ---
 canonical: https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.metric_start_time/
-aliases:
-  - ../otelcol.processor.metric_start_time/ # /docs/alloy/latest/reference/otelcol.processor.metric_start_time/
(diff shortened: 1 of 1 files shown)


kalleep:
>I called this one metric_start_time, because there is an https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339 upstream to rename it.

We have some of the components on that list and maybe we should rename them as well in a backwards compatible way, either we add some kind of alias support for components or we register them twice.

Would be good to have a issue to track this

ptodev:
@kalleep good idea! I opened #5349 to track this.

PR #5350, opened 2026-01-26
feat: Database_observability.mysql: support excluding schemas in all collectors

cristiangreco:
Propagate `exclude_schemas` configuration to all MySQL collectors and append the user-defined schemas to the default excluded schemas.
Additionally, change the behaviour of explain_plans collector to skip queries from excluded schemas instead of processing them.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/explain_plans_test.go
+++ b/internal/component/database_observability/mysql/collector/explain_plans_test.go
@@ -1529,7 +1529,7 @@ func TestExplainPlans(t *testing.T) {
 
 		t.Run("uses argument value on first request", func(t *testing.T) {
 			nextSeen := lastSeen.Add(time.Second * 45)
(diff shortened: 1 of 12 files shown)


internal/component/database_observability/mysql/collector/explain_plans_test.go
```diff
@@ -1899,17 +1899,12 @@ func TestSchemaDenylist(t *testing.T) {
...
 		"schema_name",
 		"digest",
 		"query_sample_text",
 		"last_seen",
 	}).AddRow(
-		"some_schema",
-		"some_digest1",
-		"select * from some_table where id = 1",
-		lastSeen,
-	).AddRow(
```
cristiangreco, line 1912: Removed because now with `exclude_schemas` we don't expect to fetch queries that belong to `some_schema`.

rgeyer (approved):
LGTM 🚀

PR #5353, opened 2026-01-26
chore: Point release please to generated collector main.go file

blewis12:
We're currently encountering an issue in release please [like so](https://github.com/grafana/alloy/actions/runs/21362169018/job/61484082322?pr=5351)

This is because release please is updating the version, which reflects in the [build manifest](https://github.com/grafana/alloy/blob/a94357fe33a90860807445bac8d12def9b36e37d/collector/builder-config.yaml#L5). However, the version is also part of the generated code [here](https://github.com/grafana/alloy/blob/a94357fe33a90860807445bac8d12def9b36e37d/collector/main.go#L23). If we update one and not the other, the `check generate-otel-collector-distro` action will fail as the output of the code generation does not match what is checked in to the repo

One way to get around this is to include the `// x-release-please-version` in the generated files so that release please can also be pointed there and update the version, which would remove the need to re-generate files as the expected state should be correct

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -30,6 +30,10 @@
         {
           "type": "generic",
           "path": "collector/builder-config.yaml"
+        },
+        {
+          "type": "generic",
+          "path": "collector/main.go"
         }
       ]
     }


blewis12:
discussed a bit outside this PR - but this is likely not the best way forward. closing for now, if we decided to revisit we can reopen :)

PR #5358, opened 2026-01-27
chore: Modify collector main.go to resolve version from release please

blewis12:
As a follow up to the [previous PR](https://github.com/grafana/alloy/pull/5353) - this PR introduces another solution to solve the same problem, but avoids a solution where release-please is itself modifying generated code. Instead, our generated code will be re-jigged slightly to resolve it's version from the `CollectorVersion` function, which contains a version that is maintained by release-please

There's a bit of refactoring to the generator code as well just to make things easier to read


collector/main.go
```diff
@@ -20,7 +20,7 @@ func main() {
 	info := component.BuildInfo{
 		Command:     "alloy",
 		Description: "Alloy OTel Collector distribution.",
-		Version:     "v1.12.1",
```
blewis12, line 23: this is the main output - our generated `main.go` file now contains a reference to the `CollectorVersion()` function. 

running `./build/alloy otel --version` gives us `v1.12.1`

collector/VERSION
```diff
@@ -0,0 +1 @@
+1.12.1 # x-release-please-version
```
jharvey10:
```suggestion
1.12.1 # x-release-please-version

```

(basically to add a new line at the end of the file)

blewis12: oh true, updating the test as well just to make sure it handles newlines

PR #5363, opened 2026-01-27
fix: Add support for compressed files for tail package

kalleep:
When `decompression` is configured for `loki.soure.file` we now use the same code internally as we do when it's not configured. This aligns the feature between them like BOM detection.

### Pull Request Details
Before this pr we had two different implementation that used when tailing file, tailer and decompressor. The latter was used any time decompression was configured.

Ever since my major refactors to tail package I was certain that we could add support for compressed files and reuse the same implementation.

That is what I have done here so most code are now shared. When compression is configured for `tail.File` it will not wait if EOF is returned. It will then check if we have any remaining data to flush and the return EOF. This will stop the tailer.

One issue we have is that the previous implementation tracked position by line numbers but this implementation will track offset on uncompressed data, not sure how we can handle that.

In addition to decompression support I fixed an issue that I noticed by adding `flush` to reader, if we hit EOF we drain all remaining data. But if we had data that did not include a newline it would never be consumed.

### Notes to the Reviewer

If we consume a compressed file we will exit and remove the stored position, while alloy is running this is fine. But if alloy is restarted we will ingest the files again. This is true with both this implementation and the previous one and we should fix that in another pr. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/tailer.go
+++ b/internal/component/loki/source/file/tailer.go
@@ -280,7 +280,7 @@ func (t *tailer) readLines(done chan struct{}) {
 			// We get context.Canceled if tail.File was stopped so we don't have to log it.
 			// If we get context.Canceled it means that tail.File was stopped. If we get EOF
 			// that means that we consumed the file fully and don't wait for more events, this
-			// happens when compression is configured..
+			// happens when compression is configured.
 			if !errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) {
 				level.Error(t.logger).Log("msg", "failed to tail file", "err", err)
 			}


internal/component/loki/source/file/tailer.go
```diff
@@ -276,9 +277,13 @@ func (t *tailer) readLines(done chan struct{}) {
 	for {
 		line, err := t.file.Next()
 		if err != nil {
+			fmt.Println("Exit due to error here")
```
ptodev: Should this be removed?

internal/component/loki/source/file/tailer.go
```diff
@@ -276,9 +277,13 @@ func (t *tailer) readLines(done chan struct{}) {
...
 			// We get context.Canceled if tail.File was stopped so we don't have to log it.
-			if !errors.Is(err, context.Canceled) {
-				level.Error(t.logger).Log("msg", "tail routine: stopping tailer", "err", err)
+			// If we get context.Canceld it means that tail.File was stopped. If we get EOF
```
ptodev:
```suggestion
			// If we get context.Canceled it means that tail.File was stopped. If we get EOF
```

internal/component/loki/source/file/tailer.go
```diff
@@ -276,9 +277,13 @@ func (t *tailer) readLines(done chan struct{}) {
...
+			// that means that we consumed the file fully and don't wait for more events, this
+			// happens when compression is configured..
```
ptodev:
```suggestion
			// happens when compression is configured.
```

kalleep:
>Is this going to be a problem for users who have a partially consumed archive? Is there a chance of logs being lost during Alloy upgrades?

It would be the other way around, we would most likely read lines that have already been consumed again

kalleep:
I want to add integration tests for compression usage. If it's okay I will leave that to a followup

kalleep:
>It'd be good to note this with a fix: in the changelog.

Yeah this is probably good because now we get the same support for BOM detection when reading compressed data.

PR #5366, opened 2026-01-27
feat: Database_observability.postgres: support excluding DBs in all collectors

cristiangreco:
Propagate `exclude_databases` configuration to all PostgreSQL collectors and append the user-defined databases to the default excluded databases.

Additionally, change the behaviour of `explain_plans` collector to skip queries from excluded databases instead of processing them.

Followup of https://github.com/grafana/alloy/pull/5350

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/mysql/collector/exclude_schemas.go
```diff
@@ -1,33 +1,18 @@
 package collector
 
-import "strings"
+import "github.com/grafana/alloy/internal/component/database_observability"
```
cristiangreco, line 3: Most of the code here has been moved to shared `sql_exclusion.go`

internal/component/database_observability/sql_exclusion_test.go
```diff
@@ -0,0 +1,84 @@
+package database_observability
```
cristiangreco, line 1: Tests moved here from deleted file `exclude_schemas_test.go`

internal/component/database_observability/mysql/collector/exclude_schemas.go
```diff
@@ -1,33 +1,18 @@
...
 
 var defaultExcludedSchemas = []string{"mysql", "performance_schema", "sys", "information_schema"}
 
-var defaultExclusionClause = buildExclusionClause(defaultExcludedSchemas)
+var defaultExclusionClause = database_observability.BuildExclusionClause(defaultExcludedSchemas)
 
 func buildExcludedSchemasClause(excludedSchemas []string) string {
```
matthewnolf: Can we move this to `sql_exclusion` given it's pretty generic appending of slices

internal/component/database_observability/mysql/collector/exclude_schemas.go
```diff
@@ -1,33 +1,18 @@
 package collector
 
-import "strings"
+import "github.com/grafana/alloy/internal/component/database_observability"
 
 var defaultExcludedSchemas = []string{"mysql", "performance_schema", "sys", "information_schema"}
```
matthewnolf:
```suggestion
var excludedSchemas = []string{"mysql", "performance_schema", "sys", "information_schema"}
```

internal/component/database_observability/mysql/collector/exclude_schemas.go
```diff
@@ -1,33 +1,18 @@
...
 
-var defaultExclusionClause = buildExclusionClause(defaultExcludedSchemas)
+var defaultExclusionClause = database_observability.BuildExclusionClause(defaultExcludedSchemas)
```
matthewnolf:
```suggestion
var exclusionClause = database_observability.BuildExclusionClause(defaultExcludedSchemas)
```

internal/component/database_observability/postgres/collector/exclude_databases.go
matthewnolf, line 1: Same comments as in internal/component/database_observability/mysql/collector/exclude_schemas.go

internal/component/database_observability/mysql/collector/exclude_schemas.go
cristiangreco: As optimisation, the var `defaultExclusionClause` is precomputed as string and returned by this method, with the default being different for each engine.

We could move everything into `sql_exclusion` but would need to carry over the engine-specific lists as well in the top package. I don't have a strong preference, wdyt?

matthewnolf: I prefer not to move engine-specifics to top package so let's keep as is. Thanks

matthewnolf (approved):

internal/component/database_observability/mysql/collector/exclude_schemas.go
cristiangreco: Have pushed a second commit with the rename.

internal/component/database_observability/mysql/collector/exclude_schemas.go
cristiangreco: Ok I'll leave as-is. (but I've applied the rename you suggested in the other comment)

Issue #2194, opened 2024-11-29
loki.process forwarding to multiple components blocks all destinations if one is blocked


claytonpeters:
### What's wrong?

We have a `loki.process` component in our setup which for testing we had set to `forward_to` multiple endpoints as we needed K8s logs to go to multiple places. When one of the `loki.write` components in this block was unable to send on metrics (because the server it was sending to was temporarily unavailable), we noted that the other components that `loki.process` was forwarding on to also stopped receiving logs.

### Steps to reproduce

* Setup alloy to take a Loki source (for example from Kubernetes pod logs, i.e. `loki.source.kubernetes`)
* Have a `loki.process` component take those logs and forward them on to multiple write end points (a `loki.write`, and at least one other)
* Take down the destination server for one of those write `loki.write` end points so that the logs fail to deliver
* All destinations will now stop receiving logs

### System information

Linux 5.15.0-122

### Software version

Grafana Alloy v1.4.2

### Configuration

```text
discovery.kubernetes "pods" {
  role = "pod"
}
discovery.relabel "pods_logs" {
  targets = discovery.kubernetes.pods.targets
  rule {
    source_labels = ["__meta_kubernetes_namespace"]
    action        = "replace"
    target_label  = "namespace"
  }
  // ..etc..
}
loki.source.kubernetes "pods_logs" {
  targets    = discovery.relabel.pods_logs.output
  forward_to = [loki.process.pods_logs.receiver]
  clustering {
    enabled = true
  }
}
loki.process "pods_logs" {
  stage.static_labels {
    values = {
      cluster = "cluster-name",
    }
  }
  forward_to = [
    loki.write.default.receiver,
    loki.write.second_test.receiver,
  ]
}
loki.write "default" {
  endpoint {
    url       = "http://default-loki-endpoint:3100/loki/api/v1/push"
    tenant_id = "public"
  }
}
loki.write "second_test" {
  endpoint {
    url       = "http://second-loki-endpoint:3100/loki/api/v1/push"
    tenant_id = "public"
  }
}
```

### Logs

```text

```

digitalsoba:
Any folks have updates on this issue?

michaelfedell:
[...]

PR #5371, opened 2026-01-28
feat: Add block_on_overflow to loki.write queue_config

kalleep:
### Pull Request Details
`loki.write` would always block the pipeline when we wait for a item to be queued. There are cases where we don't want this behavior, one of them listed in https://github.com/grafana/alloy/issues/2194.

```
We just run into this issue as well, we write to prod / stage loki envs. When stage went down over weekend our whole ingestion stopped. When we restore stage cluster the writes continued.
```
Currently there is no way to handle this kind of issues.

In this pr I added `block_on_overflow`. By default this is set to true to keep existing behavior but if set to false all entries that could not be queued are dropped. This is configurable per `endpoint`
 
Related to https://github.com/grafana/alloy/issues/2194

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/endpoint_test.go
+++ b/internal/component/common/loki/client/endpoint_test.go
@@ -478,7 +478,7 @@ func TestEndpointBlockOnOverflow(t *testing.T) {
 		entry := loki.Entry{Entry: push.Entry{Line: "my entry"}}
 
 		// NOTE: We have configured batch size to 1 so only one entry will fit in each batch.
-		// To fill up the queue we need to pass 4 entries. We have one batch that we are actively trying
+		// To exceed the queue's capacity we need to pass 4 entries. We have one batch that we are actively trying
 		// to send, one batch that is queued and one batch that we are currently working with filling up.
 		require.NoError(t, e.enqueue(entry, 0))
 		require.NoError(t, e.enqueue(entry, 0))


internal/component/common/loki/client/endpoint_test.go
```diff
@@ -448,3 +451,80 @@ func TestEndpoint(t *testing.T) {
...
+				BlockOnOverflow: false,
+			},
+		}, log.NewNopLogger(), internal.NewNopMarkerHandler())
+		require.NoError(t, err)
+		defer e.stop()
+
+		entry := loki.Entry{Entry: push.Entry{Line: "my entry"}}
+
+		// NOTE: We have configured batch size to 1 so only one entry will fit in each batch.
+		// To fill up the queue we need to pass 4 entries. We have one batch that we are actively trying
```
dehaansa:
```suggestion
		// To exceed the queue's capacity we need to pass 4 entries. We have one batch that we are actively trying
```

internal/component/common/loki/client/endpoint_test.go
```diff
@@ -448,3 +451,80 @@ func TestEndpoint(t *testing.T) {
...
+		// to send, one batch that is queued and one batch that we are currently working with filling up.
+		require.NoError(t, e.enqueue(entry, 0))
+		require.NoError(t, e.enqueue(entry, 0))
+		// The third enqueue is a bit flaky because it will depend if a shard has grabbed a queued batch or not.
```
dehaansa, line 485: Is there a way to make this more consistent? I don't love that this would obscure a real behavior change with capacity management.

dehaansa (approved):
Overall LGTM, one nit and one test question.

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -134,11 +134,12 @@ The optional `queue_config` block configures how the endpoint queues batches of
...
-| `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it enqueues pending batches and drains the send queue sending each.  | `"1m"`  | no       |
-| `min_shards`    | `number`   | Minimum number of concurrent shards sending samples to the endpoint.                                                                                                          | `1`      | no       |
+| Name                 | Type       | Description                                                                                                                                                                   | Default | Required |
+| -------------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `block_on_overflow`  | `bool`     | If `true`, block until there is space in the queue; if `false`, drop entries when queue is full.                                                                              | `true`  | no       |
```
clayton-cornell, line 139:
```suggestion
| `block_on_overflow`  | `bool`     | If `true`, block until there is space in the queue. If `false`, drop entries when queue is full.                                                                              | `true`  | no       |
```

PR #5374, opened 2026-01-28
fix: Remove 'nodes/proxy' RBAC rule and replace with 'nodes/pods'.

### LINKED ISSUE #5364: docs/rbac: Is nodes/proxy permission required for discovery.kubernetes? (potentially excessive + blocks GKE Autopilot)
### Component(s)

RBAC

### Summary

The Grafana Alloy Helm chart / documentation currently suggests granting the `discovery.kubernetes` component get/list/watch permissions for the `nodes/proxy` subresource:
	•	Alloy values.yaml reference:
https://github.com/grafana/alloy/blob/d3542bc2bc2a677a66b98478289af37bd022ff22/operations/helm/charts/alloy/values.yaml#L203-L206

However, Prometheus Kubernetes discovery RBAC example (which appears to use the same discovery mechanisms) does not request `nodes/proxy` at all:
 * Prometheus RBAC example:
https://github.com/prometheus/prometheus/blob/8721871cf76c713192609c144ebd7b42f5bc846b/documentation/examples/rbac-setup.yml#L13-L20

### Question / request

Is `nodes/proxy` actually required for Alloy’s `discovery.kubernetes` in any supported configuration?
* If yes, could you document the exact scenario(s) that require it (e.g., which discovery role or endpoint needs it)?
* If no, could the chart/docs be updated to remove the nodes/proxy permissions from the default ClusterRole and/or mark them as optional?

### Why this matters (security)

Granting get on `nodes/proxy` is high impact in many clusters. It can enable node-local execution against pods on that node (and, depending on cluster configuration, can lead to privilege escalation, network-policy bypass, and reduced audit visibility because kubelet interactions may not be captured in the Kubernetes audit log).
* Background: https://grahamhelton.com/blog/nodes-proxy-rce
* I have reproduced this risk using the default Grafana Alloy ClusterRole (i.e., nodes/proxy access was sufficient for the demonstrated behavior in my environment).

### Why this matters (compatibility)

Some managed Kubernetes offerings restrict or disable access to nodes/proxy entirely. For example, GKE Autopilot environments prevent giving out `nodes/proxy` privilege entirely:
* GKE Autopilot related documentation:
https://docs.cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port

Removing `nodes/proxy` from the default RBAC (or making it explicitly opt-in) would improve Alloy deployability in such environments.

### Proposed change
* **Preferred**: Remove nodes/proxy from the default Alloy ClusterRole if not required for discovery.kubernetes.
* **Alternative**: Make it opt-in (values flag), and clearly document when/why it’s needed. 

### Steps to reproduce

Deploy the default chart of Grafana Alloy, open a shell to the Grafana Alloy pod and send any exec command to any pod on the same node using the Websocket interface on port 10250 (kubelet API) and it will run.

### System information

Latest GKE Kubernetes

### Software version

Grafana Alloy v1.3.0

### Configuration

```text
rules:
  # Rules which allow discovery.kubernetes to function.
  - apiGroups:
      - ""
      - "discovery.k8s.io"
      - "networking.k8s.io"
    resources:
      - endpoints
      - endpointslices
      - ingresses
      - nodes
      - nodes/proxy
      - nodes/metrics
      - pods
      - services
    verbs:
      - get
      - list
      - watch
```

### Logs

```text

```

### Tip

petewall:

This change removes the `nodes/proxy` ClusterRole from the default RBAC rule set. This permission has been linked to a security issue in Kubernetes where this permission can cause [remote code execution](https://grahamhelton.com/blog/nodes-proxy-rce).

It also adds the `nodes/pods` permission, which should allow for the `discovery.kubelet` component to still function.

Fixes #5364

### Notes to the Reviewer

Ran a test in the k8s-monitoring helm chart, manually removing that permission.
All tests passed: https://github.com/grafana/k8s-monitoring-helm/pull/2279

### PR Checklist

- [X] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -10,7 +10,7 @@ internal API changes are not present.
 Unreleased
 ----------
 
-### Bug fixes
+### Enhancements
 
 - Remove `nodes/proxy` RBAC rule and replace with `nodes/pods`. (@petewall)
 


operations/helm/charts/alloy/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ internal API changes are not present.
 Unreleased
 ----------
 
+### Bug fixes
```
dehaansa: I think this should probably be an enhancement?

operations/helm/charts/alloy/values.yaml
```diff
@@ -200,9 +200,17 @@ rbac:
 
   # -- The rules to create for the ClusterRole objects.
   clusterRules:
-    # -- Rules required for the `discovery.kubernetes` component.
+    # -- Rules required for the Nodes role in the `discovery.kubernetes` component.
+    - apiGroups: [""]
+      resources: ["nodes"]
```
dehaansa, line 205: Is there a reason this needed to be changed from one rule w/ 3 resources to 3 rules (one for each resource)? I think it's cleaner, but wanted to understand if it was required.

petewall: Not required, but it means that the documentation is more specific to their purpose.

devops-taehoon-kim:
Hi. Because of this, we face 403 forbidden issue.

```

job | "prometheus.scrape.cadvisor"
-- | --
url | "https://kubernetes.default.svc.cluster.local:443/api/v1/nodes/ip-172-31-197-4.ap-northeast-2.compute.internal/proxy/metrics/cadvisor"
health | "down"
labels | { 	instance = "ip-172-31-197-4.ap-northeast-2.compute.internal", 	job      = "prometheus.scrape.cadvisor", 	nodepool = "infra", }
last_error | "server returned HTTP status 403 Forbidden"
last_scrape | "2026-02-10T02:54:22.9939514Z"
last_scrape_duration | "1.731773ms"
```

I fixed this issue like this

```
  - patch: |-
      - op: add
        path: /rules/-
        value:
          apiGroups: [""]
          resources: ["nodes/proxy"]
          verbs: ["get", "list", "watch"]
    target:
      kind: ClusterRole
      name: grafana-alloy.*
```

petewall:
@jessebot, @devops-taehoon-kim , see if the strategy outlined in this issue solves your problem: https://github.com/grafana/alloy/issues/5417

devops-taehoon-kim:
As security is not a primary concern in my case, I will proceed by retaining the nodes/proxy permission.

Thank you very much for your support.

Issue #5367, opened 2026-01-27
proposal: evolving position files


kalleep:
### Background

Currently we for position tracking we [Entry](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/internal/positions/positions.go#L67-L70) as key with a string key.

This works in most cases but we have a need to be able to store additional metadata along side the key value. E.g. for loki.source.file we want to be able to store file signature so we can detect rotated files while alloy was stopped.

### Proposal

To be able to store additional data along side values I propose that we add another field to [position.File](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/internal/positions/positions.go#L73-L75)

```go
type File struct {
  Positions map[Entry]string `yaml:"positions"`
  PositionsV2 map[Entry]Value `yaml:"positions_v2"`
}

type Value struct {
  Content string `yaml:"content"`
  Metadata map[string]string `yaml:"metadata"`
}
```

And we would update both the old format and new for a while. So we support downgrading alloy to a version that don't support the new format in case that is needed.

But currently we use [strict unmarshalling](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/internal/positions/positions.go#L344) of yaml files. This do not allow to unmarshal yaml with unknown feilds.

So we could change this to not be strict, wait until next release of alloy and then add a new format. That way we could still support downgrading one minor without issues.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

ptodev:
Ideally we can support the old format for a long time, so that users who migrate to new versions a few months from now can do so smoothly.

Would it be possible to add a schema header on top of the file? And to have a position file interface which has different implementations, depending on the schema of the file being read? For unmarshaling we could use something generic like mapstructure if we need to handle all sorts of fields?

kalleep:
>Would it be possible to add a schema header on top of the file? And to have a position file interface which has different implementations, depending on the schema of the file being read? For unmarshaling we could use something generic like mapstructure if we need to handle all sorts of fields?

Yes we have talked about that. Currently the problem is that we cant even parse the yaml if we add additional fields. 

Need to merge https://github.com/grafana/alloy/pull/5392 first and let that sit for a while.

ptodev:
> let that sit for a while

[...]

PR #5392, opened 2026-01-29
chore: Change to use non strict unmarshal for position file

kalleep:
### Pull Request Details
Main change is to use non strict Unmarshal for position file. If we want to be able to change this format like adding new properties in the future while still being able to downgrade alloy we need to change this.

For more details read https://github.com/grafana/alloy/issues/5367

Part of: https://github.com/grafana/alloy/issues/5367

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/internal/positions/positions_test.go
+++ b/internal/component/loki/source/internal/positions/positions_test.go
@@ -179,7 +179,7 @@ positions:
 		}()
 
 		yaml := []byte(`
-Version: v2
+some_new_property: new_unexpected_value
 positions:
   ? path: /tmp/random.log
     labels: '{job="tmp"}'


internal/component/loki/source/internal/positions/positions.go
```diff
@@ -15,13 +16,13 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
+	yamlv3 "gopkg.in/yaml.v3"
```
kalleep: Not to sure about this change yet. Need to dig into this a bit more before commiting

internal/component/loki/source/internal/positions/positions_test.go
```diff
@@ -143,32 +143,63 @@ func TestConvertLegacyPositionsFileJournal(t *testing.T) {
...
+			_ = os.Remove(temp)
+		}()
 
-	require.NoError(t, err)
-	require.Equal(t, "17623", pos[Entry{
-		Path:   "/tmp/random.log",
-		Labels: `{job="tmp"}`,
-	}])
+		yaml := []byte(`
+Version: v2
```
thampiotr:
```suggestion
some_new_property: new_unexpected_value
```
we may end up adding version property
also it is easier to read this way

PR #5396, opened 2026-01-29
perf: Reduce allocations in stage.docker

kalleep:
### Pull Request Details
Similiar to https://github.com/grafana/alloy/pull/5337 but much simpler.
The format for json logs emitted by docker can be found here: https://docs.docker.com/engine/logging/drivers/json-file

### Notes to the Reviewer

I think these where defined as "extentions" because it was / is possible to do all this by combining other existing stages and where added because it's convenient to have.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/cri.go
+++ b/internal/component/loki/process/stages/cri.go
@@ -68,10 +68,10 @@ type cri struct {
 func (c *cri) Name() string { return StageTypeCRI }
 
 const (
-	criFlags     = "flags"
-	criStream    = "stream"
-	criContent   = "content"
-	criTimestamp = "timestamp"
+	criFlags   = "flags"
+	criStream  = "stream"
+	criContent = "content"
+	criTime    = "time"
 )
 
 func (c *cri) Run(in chan Entry) chan Entry {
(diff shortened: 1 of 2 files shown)


kalleep:
```
goos: linux
goarch: amd64
          │ bench_old.txt │            bench_new.txt            │
          │    sec/op     │   sec/op     vs base                │
Docker-16     4.099µ ± 1%   1.712µ ± 1%  -58.25% (p=0.000 n=10)

          │ bench_old.txt │           bench_new.txt            │
          │     B/op      │    B/op     vs base                │
Docker-16     2033.0 ± 0%   576.0 ± 0%  -71.67% (p=0.000 n=10)

          │ bench_old.txt │           bench_new.txt            │
          │   allocs/op   │ allocs/op   vs base                │
Docker-16      44.00 ± 0%   13.00 ± 0%  -70.45% (p=0.000 n=10)
```

internal/component/loki/process/stages/cri.go
```diff
@@ -68,10 +68,10 @@ type cri struct {
...
 
 const (
-	criTime    = "time"
-	criFlags   = "flags"
-	criStream  = "stream"
-	criContent = "content"
+	criFlags     = "flags"
+	criStream    = "stream"
+	criContent   = "content"
+	criTimestamp = "timestamp"
```
kalleep: In previous implementation cri used `time` but documentation says it extracts [timestamp](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagecri).

We can change to timestamp here to:
1. Follow the documentation
2. Align with docker

ptodev: Isn't this a breaking change?

kalleep: Yeah I guess it is even though we documented that it extracted `timestamp`. I can change it back and update docs

clayton-cornell (approved):

PR #5398, opened 2026-01-29
ci: Allow release note enrichment when "closes" is present

### LINKED ISSUE #5297: foreach id doesn't work with discovery.file

jharvey10:
An interesting thing I discovered with release please is that if a commit has a reference to a closed issue, this will be called out in the changelog :tada:

For example:

```
* Invalid handling of id in foreach when using discovery components (#5322) (61fe184), closes #5297
```

The only problem is this breaks the way I wrote the regex for the release not enrichment script. This PR fixes that.

### DIFF
--- a/tools/release/enrich-release-notes/main.go
+++ b/tools/release/enrich-release-notes/main.go
@@ -13,6 +13,10 @@ import (
 	"github.com/grafana/alloy/tools/release/internal/version"
 )
 
+// commitPattern matches a commit SHA in markdown link format: "([abc1234](https://github.com/.../commit/...))"
+// It captures the short SHA from the link text, regardless of surrounding context.
+var commitPattern = regexp.MustCompile(`\(\[([a-f0-9]{7,40})\]\(https://github\.com/[^)]+\)\)`)
+
 // commitAuthorsQuery is the GraphQL query to fetch all authors for a commit.
 const commitAuthorsQuery = `query($owner: String!, $repo: String!, $oid: GitObjectID!) {
 	repository(owner: $owner, name: $repo) {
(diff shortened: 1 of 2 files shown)


tools/release/enrich-release-notes/main.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/grafana/alloy/tools/release/internal/version"
 )
 
+// commitPattern matches a commit SHA in markdown link format: "([abc1234](https://github.com/.../commit/...))"
+// It captures the short SHA from the link text, regardless of surrounding context.
+var commitPattern = regexp.MustCompile(`\(\[([a-f0-9]{7,40})\]\(https://github\.com/[^)]+\)\)`)
```
jharvey10, line 18: the main difference is the absence of `\s*$`

PR #5399, opened 2026-01-29
feat(loki.process): Add debug metrics for CRI stage to track truncation of lines and partial line flushing

ptodev:
### Pull Request Details

For partial lines there is currently a log line, but having a metric could be an easier way to tell if something wrong is going on. And it means we could alert on it.

For truncation there are no logs and metrics, so this will be the first time we can track it.

Those are both low cardinality metrics and I don't expect them to have much impact. 

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/cri.go
+++ b/internal/component/loki/process/stages/cri.go
@@ -12,6 +12,7 @@ import (
 	crip "github.com/grafana/alloy/internal/component/loki/process/stages/cri"
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
+	"github.com/grafana/alloy/internal/util"
 	"github.com/grafana/alloy/syntax"
 )
 
@@ -50,14 +51,8 @@ func (args *CRIConfig) Validate() error {
 }
 
 func NewCRI(logger log.Logger, cfg CRIConfig, registerer prometheus.Registerer, _ featuregate.Stability) (Stage, error) {
-	partialLinesFlushedMetric, err := getPartialLinesFlushedMetric(registerer)
-	if err != nil {
-		level.Error(logger).Log("msg", "failed to register cri partial lines flushed metric", "err", err)
-	}
-	linesTruncatedMetric, err := getLinesTruncatedMetric(registerer)
-	if err != nil {
-		level.Error(logger).Log("msg", "failed to register cri lines truncated metric", "err", err)
-	}
+	partialLinesFlushedMetric := getPartialLinesFlushedMetric(registerer)
+	linesTruncatedMetric := getLinesTruncatedMetric(registerer)
 	return &cri{
 		logger:                    logger,
 		cfg:                       cfg,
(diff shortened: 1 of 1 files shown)


internal/component/loki/process/stages/cri.go
```diff
@@ -153,6 +198,9 @@ func (c *cri) Run(in chan Entry) chan Entry {
 func (c *cri) ensureTruncateIfRequired(e *Entry) {
 	if c.cfg.MaxPartialLineSizeTruncate && len(e.Line) > int(c.cfg.MaxPartialLineSize) {
 		e.Line = e.Line[:c.cfg.MaxPartialLineSize]
+		if c.linesTruncatedMetric != nil {
+			c.linesTruncatedMetric.Inc()
```
ptodev, line 181: I considered adding the log labels as metric labels, to make it easier to identify log streams with long lines. I suspect most of the time it'll be a particular stream. But for now I don't want to make changes that could lead to too many metrics.

internal/component/loki/process/stages/cri.go
```diff
@@ -105,6 +147,9 @@ func (c *cri) Run(in chan Entry) chan Entry {
 		if parsed.Flag == crip.FlagPartial {
 			if len(c.partialLines) >= c.cfg.MaxPartialLines {
 				level.Warn(c.logger).Log("msg", "cri stage: partial lines upperbound exceeded. merging it to single line", "threshold", c.cfg.MaxPartialLines)
+				if c.partialLinesFlushedMetric != nil {
+					c.partialLinesFlushedMetric.Add(float64(len(c.partialLines)))
```
ptodev, line 130: Tracking the total number of lines is preferable to tracking the number of times the counter reset. That's because if the user updates the limit in the configuration, then the meaning of the metric would change.

internal/component/loki/process/stages/cri.go
```diff
@@ -49,20 +49,62 @@ func (args *CRIConfig) Validate() error {
...
+		Help: "A count of partial lines that were flushed prematurely due to the max_partial_lines limit being exceeded",
+	})
+	err := registerer.Register(metric)
+	if err != nil {
+		if existing, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			return existing.ExistingCollector.(prometheus.Counter), nil
+		}
+		return nil, err
+	}
+	return metric, nil
```
kalleep: We have a helper function for this https://github.com/grafana/alloy/blob/main/internal/util/metrics.go#L8

clayton-cornell:
No suggestions for docs. Looks OK as-is.

ptodev:
> Could you explain how these could be used to actually find issues?

It's hard to tell what values to set for `max_partial_line_size` and `max_partial_lines` without some feedback from Alloy. For  truncated lines there are neither logs nor metrics. For partial lines there is a low-cardinality log line which should really be a metric instead, so that users can track better at what point in time the problem happened. Maybe there could also be an info-level alert for such things.

ptodev:
Hi @kalleep, have you had a chance to think about this please? I'm open to other ways of debug this but a metric seems like the easy and low cost way of doing it.

kalleep:
It looks good to me but still we should use the helper function I mentioned here https://github.com/grafana/alloy/pull/5399#discussion_r2797833940

internal/component/loki/process/stages/cri.go
ptodev: Thanks, I updated the code to use it.

It's not great how the helper function uses `AlreadyRegisteredError`, but since it's already the case for most of the code in Alloy it's ok to have it here. I hope over time we can get rid of it.

PR #5401, opened 2026-01-29
fix(scheduling): Shutdown runnables with a timeout before starting new ones

kgeckhart:

Adjust the ordering for component scheduling so that we give components time to shutdown before we start new components. If components take too long we will still move on but wait for all to shutdown during scheduling. 

### Pull Request Details

We have run in to issues with components which bind shared resources not being able to release those resources when the component is rescheduled due to a module rename / component rename. An example of this is the `loki.source.syslog` component which will bind 1 or more ports depending on the listener config. This causes issues with our current implementation where shutdown/startup order is not defined,

1. Old components are told to shutdown
2. As soon as all old components have been told to shutdown, new components are started
3. If the syslog component did not shutdown fast enough the new component will fail to bind ports
4. Failing to bind ports does not cause the syslog component to exit we just won't open the listener forcing a user to restart alloy to resolve the problem

I need to look to see if there are issues which describe this sort of behavior

### Notes to the Reviewer

This PR is intended to work with components which bind resources when Run. We have multiple components which bind resources when the is newly created before running. This causes a panic instead which will be resolved in a separate PR.

I chose to leave an escape hatch on shutdown because our default component shutdown timeout is 10 minutes. This is incredibly long to leave a pipeline offline. Using the warning log window of 1 minute felt much safer. 

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/runtime/internal/controller/scheduler_test.go
+++ b/internal/runtime/internal/controller/scheduler_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component"
 	"github.com/grafana/alloy/internal/runtime/internal/controller"
@@ -115,10 +116,9 @@ func TestScheduler_Synchronize(t *testing.T) {
 		finished.Wait()
 	})
 
-	t.Run("Timeout allows new tasks to start but waits for old tasks to finish", func(t *testing.T) {
+	t.Run("Shutdown will stop waiting after TaskShutdownWarningTimeout to startup components and wait for shutdown after", func(t *testing.T) {
 		synctest.Test(t, func(t *testing.T) {
-			oldTaskExited := false
-			newTaskStarted := false
+			var oldTaskExited, newTaskStarted atomic.Bool
 
 			// Old task that takes a long time to stop
 			slowStop := func(ctx context.Context) error {
@@ -126,13 +126,13 @@ func TestScheduler_Synchronize(t *testing.T) {
 
 				// Simulate slow shutdown
 				time.Sleep(2 * controller.TaskShutdownWarningTimeout)
-				oldTaskExited = true
+				oldTaskExited.Store(true)
 				return nil
 			}
 
 			// New task
 			basicRun := func(ctx context.Context) error {
-				newTaskStarted = true
+				newTaskStarted.Store(true)
 				<-ctx.Done()
 				return nil
 			}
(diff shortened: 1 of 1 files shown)


internal/runtime/internal/controller/scheduler.go
```diff
@@ -51,17 +51,23 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
+// 4. Wait for any remaining nodes to shutdown
+//
+// Nodes are shutdown first to ensure any shared resources, such as ports,
+// are allowed to be freed before new nodes are scheduled. As a means to avoid,
+// long stretches of downtime we give this a 1 minute timeout.
 //
 // Existing components will be restarted if they stopped since the previous
 // call to Synchronize.
+//
+// Synchronize is not goroutine safe and should not be called concurrently.
```
kalleep, line 69: You say this but we still have a mutex internally.

How come we cannot lock and defer the unlock, what issues would we get?

internal/runtime/internal/controller/scheduler.go
```diff
@@ -51,17 +51,23 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
 func (s *Scheduler) Synchronize(rr []RunnableNode) error {
-	s.tasksMut.Lock()
-
 	if s.ctx.Err() != nil {
 		return fmt.Errorf("Scheduler is closed")
 	}
```
kalleep: Now that this is no longer protected by a mutex we need to check the channel instead of error state directly.
```suggestion
select {
	case <-ctx.Done():
		return errors.New("Scheduler is closed")
	default:
	}
```

internal/runtime/internal/controller/scheduler.go
kgeckhart: Holding the lock is the source of the problem. A task can only be considered shutdown if it can remove itself from the task map, https://github.com/grafana/alloy/blob/29284d25aabf5af390fe91e8749aa27bd1263733/internal/runtime/internal/controller/scheduler.go#L127-L139

Since we were holding the lock the bulk of the duration of synchronize we were only telling the tasks to shutdown before starting new tasks which is where the conflict can happen. We need to release the lock if we want to wait for tasks to shutdown completely or redesign how shutdown works. A task can shutdown at anytime outside of calls to synchronize which leads us to needing the mutex still. 

The new approach has a _slight_ race condition where a task shuts down while synchronize is waiting for the other tasks to shut down. I think this is okay and falls in to the "Existing components will be restarted if they stopped since the previous call to Synchronize." case.

internal/runtime/internal/controller/scheduler.go
kgeckhart: Great catch, thank you! I'll include it with a fix for whatever is going on with the logger in tests now.

PR #5403, opened 2026-01-30
feat(loki.source.file): Track duplicate targets with a metric and logs

kalleep:
### Pull Request Details

Alternative to https://github.com/grafana/alloy/pull/5397.

Cannot detect duplicated file through e.g. symlinks.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/file/file.go
+++ b/internal/component/loki/source/file/file.go
@@ -361,6 +361,24 @@ func (c *Component) scheduleSources() {
 			})
 		},
 	)
+
+	var (
+		duplicationCount = float64(0)
+		scheduled        = make(map[string]struct{}, c.scheduler.Len())
+	)
+
+	for s := range c.scheduler.Sources() {
+		path := s.Key().Path
+		if _, ok := scheduled[path]; ok {
+			duplicationCount += 1
+			level.Warn(c.opts.Logger).Log(
+				"msg", "multiple targets with different labels which will cause duplicate log lines",
+				"path", path,
+			)
+		}
+		scheduled[path] = struct{}{}
+	}
+	c.metrics.duplicatedFiles.Set(float64(duplicationCount))
 }
 
 type debugInfo struct {
--- a/internal/component/loki/source/file/metrics.go
+++ b/internal/component/loki/source/file/metrics.go
@@ -15,10 +15,11 @@ type metrics struct {
 	reg prometheus.Registerer
 
 	// File-specific metrics
-	readBytes   *prometheus.GaugeVec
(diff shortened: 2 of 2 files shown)


internal/component/loki/source/file/file.go
```diff
@@ -361,6 +361,24 @@ func (c *Component) scheduleSources() {
...
+		duplicationCount = float64(0)
+		scheduled        = make(map[string]struct{}, c.scheduler.Len())
+	)
+
+	for s := range c.scheduler.Sources() {
+		path := s.Key().Path
+		if _, ok := scheduled[path]; ok {
+			duplicationCount += 1
+			level.Warn(c.opts.Logger).Log(
+				"msg", "multiple targets with different labels which will cause duplicate log lines",
```
ptodev, line 375: It'd be really useful to log the labels causing duplication. Otherwise users have to open the Alloy UI.

internal/component/loki/source/file/file.go
```diff
@@ -361,6 +361,24 @@ func (c *Component) scheduleSources() {
 			})
 		},
 	)
+
+	var (
+		duplicationCount = float64(0)
+		scheduled        = make(map[string]struct{}, c.scheduler.Len())
+	)
+
+	for s := range c.scheduler.Sources() {
```
ptodev, line 370: I like how this is done after the reconcile, it doesn't slow it down as much this way.

ptodev (commented):
I'll have a quick look if I can optimise something with the symlink handling, but honestly I don't think symlink support is a must. Your solution would be enough in a lot of cases

internal/component/loki/source/file/file.go
ptodev, line 370: The fact that I'm checking `targets` instead of `c.scheduler.Sources()` also made me realise that I could be erroneously reporting duplicates in situations where the tailer in reality deduplicated the targets. It's better to check the files we know for sure are tailing, rather than the raw targets list.

internal/component/loki/source/file/file.go
kalleep: Yeah I wanted to try to keep this pretty fast and not required to much memory just to produce output for a log.

But I will see what I can do :+1:

ptodev: It's fine if we do more thorough checking only when there's duplicates. This means the happy path with no duplicates is still performant. That's why I'm not too bothered by the extra allocations in https://github.com/grafana/alloy/pull/5397.

PR #5405, opened 2026-01-30
perf(loki.source.file): Reduce allocations during file reading

kalleep:
### Pull Request Details
There is a lot of allocations being performed while reading files. I can clearly see this being a hot spot in our internal clusters.

Two of the biggest issues are:
1. We allocate a new pending buffer as soon as we have completed a line
2. We use ReadBytes api causing new allocation for every call that we just move to pending

Instead we can use `ReadSlice`, this api will not allocate but return a slice pointing to the internal buffer while still advancing read position. We copy it into pending so this is fine and we remove the intermediate allocation done by `ReadBytes`.

We can also reset length of `pending` and this buffer. This is fine because we never read past newline boundary. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/file_test.go
+++ b/internal/component/loki/source/file/internal/tail/file_test.go
@@ -553,7 +553,7 @@ func verifyResult(t *testing.T, f *File, expectedLine *Line, expectedErr error) 
 var benchLine *Line
 
 func BenchmarkFile(b *testing.B) {
-	// we create a file with 1000 lines and each line is 500 bytes
+	// we create a file with 1000 lines and each line is 500 bytes.
 	line := bytes.Repeat([]byte{'a'}, 500)
 	lines := strings.Repeat(string(line)+"\n", 1000)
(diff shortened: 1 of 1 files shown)


kalleep:
Benchmark diff against main
```
goos: linux
goarch: amd64
pkg: github.com/grafana/alloy/internal/component/loki/source/file/internal/tail
        │ bench_old.txt │            bench_new.txt             │
        │    sec/op     │    sec/op     vs base                │
File-16   1287.0µ ± 19%   410.5µ ± 57%  -68.10% (p=0.000 n=10)

        │ bench_old.txt │            bench_new.txt             │
        │     B/op      │     B/op      vs base                │
File-16   5057.0Ki ± 0%   557.0Ki ± 0%  -88.99% (p=0.000 n=10)

        │ bench_old.txt │            bench_new.txt            │
        │   allocs/op   │  allocs/op   vs base                │
File-16     4.022k ± 0%   2.022k ± 0%  -49.73% (p=0.000 n=10)
```

Issue #5410, opened 2026-02-01
loki.process stage.truncate panic when explicitly defining source_type="line"


a-abella:
### Component(s)

loki.process

### What's wrong?

Hello,

Stumbled on a panic against a recently added `loki.process` stage, `stage.truncate`;

```
panic: reflect.Set: value of type string is not assignable to type stages.TruncateSourceType
```

Only occured when explicitly setting the argument `source_type = "line"`; "line" is the default value so its probably not commonly explicitly defined. After commenting out `source_type = "line"` it loaded up fine.

### Steps to reproduce

Configure a `stage.truncate` rule with explicitly defined `source_type = "line"`

### System information

Linux 5.4.0-216-generic aarch64

### Software version

alloy, version v1.12.2 (branch: HEAD, revision: 477e314)

### Configuration

```text
stage.truncate {
            rule {
                limit       = "31KiB"
                source_type = "line"
                suffix      = "...[truncated]"
            }
        }
```

### Logs

```text
Feb 01 02:05:23 ubuntu20 alloy[187324]: panic: reflect.Set: value of type string is not assignable to type stages.TruncateSourceType
Feb 01 02:05:23 ubuntu20 alloy[187324]: goroutine 1 [running]:
Feb 01 02:05:23 ubuntu20 alloy[187324]: reflect.Value.assignTo({0x9817740?, 0x4001df1b60?, 0x4001be4218?}, {0xba87bad, 0xb}, 0x9822800, 0x0)
Feb 01 02:05:23 ubuntu20 alloy[187324]:         /__t/go/1.25.5/x64/src/reflect/value.go:3146 +0x204
Feb 01 02:05:23 ubuntu20 alloy[187324]: reflect.Value.Set({0x9822800?, 0x4001dd7d40?, 0xa23d160?}, {0x9817740?, 0x4001df1b60?, 0x9822800?})
Feb 01 02:05:23 ubuntu20 alloy[187324]:         /__t/go/1.25.5/x64/src/reflect/value.go:2131 +0xcc
Feb 01 02:05:23 ubuntu20 alloy[187324]: github.com/grafana/alloy/syntax/internal/value.(*decoder).decode(0x4001e2e6a8, {{0x9817740?, 0x4001df1b50?, 0xbac928?}, 0x38?}, {0x9822800, 0x4001dd7d40, 0x198})
[...]

PR #5419, opened 2026-02-02
fix(loki.process): Implement encoding.TextMarshaler and encoding.TextUnmarshaler for TruncateSourceType

kalleep:
### Pull Request Details

It was impossible to configure `source_type` for `stage.truncate` and would always result in a panic. For these custom enum types we need to implement `encoding.TextUnmarshaler` for it to be configurable. 

Fixes: #5410

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


kalleep:
@blewis12 I added a test to parse alloy syntax for this block.

I also noticed that the documentation was wrong. We document `source` but it's actually `sources` and is an array of strings.

Issue #5423, opened 2026-02-03
bug: tail_from_end can produces invalid first line


kalleep:
### Component(s)

loki.source.file

### What's wrong?

When using `tail_from_end` we run [getLastLinePosition](https://github.com/grafana/alloy/blob/0044e0365710c7394c994c57690a585caab22440/internal/component/loki/source/file/tailer.go#L87). This one does not take encoding into account. This would be a problem for e.g. UTF-16LE because it encodes newlines as `0A 00` so we would point offset at `00` messing up the first line.

### Steps to reproduce

Use `loki.source.file` on a UTF-16LE encoded file and set `tail_from_end = true`.

### Configuration

```text

```

### Logs

```text

```

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5429, opened 2026-02-04
fix(loki.source.file): Update `tail_from_end` to properly handle file encoding

kalleep:
When `tail_from_end = true` we no longer produce invalid line on first read when reading `UTF-16LE` encoded files. 

### Pull Request Details
The issue was that `getLastLinePosition` did not take file encoding into account. I moved the logic into `tail.reader` so that we can perform this kind of operation with either configured or detected encoding.

Fixes: #5423

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/file_test.go
+++ b/internal/component/loki/source/file/internal/tail/file_test.go
@@ -439,23 +439,35 @@ func TestFile(t *testing.T) {
 	})
 
 	t.Run("start from end", func(t *testing.T) {
-		startFromEndTest(t, "utf-8", nopEncoder, nopEncoder, 0, []Line{{Text: "line3", Offset: 18}})
-		startFromEndTest(t, "utf-16be", utf16beBOMEncoder, utf16beEncoder, 0, []Line{{Text: "line3", Offset: 38}})
-		startFromEndTest(t, "utf-16le", utf16leBOMEncoder, utf16leEncoder, 0, []Line{{Text: "line3", Offset: 38}})
+		startFromEndTest(t, "utf-8", nopEncoder, nopEncoder, false, 0, []Line{{Text: "line3", Offset: 18}})
+		startFromEndTest(t, "utf-16be", utf16beBOMEncoder, utf16beEncoder, false, 0, []Line{{Text: "line3", Offset: 38}})
+		startFromEndTest(t, "utf-16le", utf16leBOMEncoder, utf16leEncoder, false, 0, []Line{{Text: "line3", Offset: 38}})
 	})
 
 	t.Run("start from end with start offset", func(t *testing.T) {
-		startFromEndTest(t, "utf-8", nopEncoder, nopEncoder, 6, []Line{
+		startFromEndTest(t, "utf-8", nopEncoder, nopEncoder, false, 6, []Line{
 			{Text: "line2", Offset: 12},
 			{Text: "line3", Offset: 18},
 		})
-		startFromEndTest(t, "utf-16be", utf16beBOMEncoder, utf16beEncoder, 14, []Line{
+		startFromEndTest(t, "utf-8-cr", nopEncoder, nopEncoder, true, 7, []Line{
+			{Text: "line2", Offset: 14},
+			{Text: "line3", Offset: 21},
+		})
+		startFromEndTest(t, "utf-16be", utf16beBOMEncoder, utf16beEncoder, false, 14, []Line{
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/file/internal/tail/bom.go
```diff
@@ -30,15 +30,18 @@ const (
...
 // and underlying file are positioned at the beginning of the file
-// when calling this function, as it peeks at the first bytes to detect the BOM.
-func detectBOM(br *bufio.Reader, offset int64) (int64, BOM) {
-	// Peek up to 4 bytes (longest BOM)
-	buf, err := br.Peek(4)
+// when calling this function, as it reads the first 4 bytes to detect the BOM.
+func detectBOM(r io.Reader, offset int64) (int64, BOM) {
+	buf := make([]byte, 4)
+
+	n, err := r.Read(buf)
```
ptodev: Why is the switch from Peek to Read?

internal/component/loki/source/file/internal/tail/file_test.go
```diff
@@ -418,16 +418,45 @@ func TestFile(t *testing.T) {
 		utf16offsets = [3]int64{14, 26, 38}
 	)
 
+	var (
+		nopEncoder        = encoding.Nop.NewEncoder()
+		utf16beEncoder    = unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM).NewEncoder()
+		utf16beBOMEncoder = unicode.UTF16(unicode.BigEndian, unicode.UseBOM).NewEncoder()
+		utf16leEncoder    = unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder()
+		utf16leBOMEncoder = unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewEncoder()
+	)
```
ptodev, line 427: I think encoders are stateful? Instead of reusing them, we could use factory functions to return an encoder just to make sure we use a clean one?

internal/component/loki/source/file/internal/tail/file_test.go
```diff
@@ -473,6 +502,35 @@ func compressionTest(t *testing.T, name, compression string, enc *encoding.Encod
 	})
 }
 
+func startFromEndTest(t *testing.T, name string, encoder, appendEncoder *encoding.Encoder, offset int64, expected []Line) {
+	t.Run(name, func(t *testing.T) {
+		content, err := encoder.String("line1\nline2\n")
```
ptodev: Do we want to also test `\r\n`?

internal/component/loki/source/file/internal/tail/util_test.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func TestLastNewline(t *testing.T) {
+	encoder := encoding.Nop.NewEncoder()
+	t.Run("empty file", func(t *testing.T) {
+		lastNewlineTest(t, "empty", encoder, "", 0)
+	})
+
+	t.Run("UTF-8 no newline", func(t *testing.T) {
+		lastNewlineTest(t, "no-nl", encoder, "line1", 0)
```
ptodev, line 20: Is it worth testing with a corrupted position? Like a negative number or a very large one?

internal/component/loki/source/file/internal/tail/bom.go
ptodev, line 37: So Read instead of Peek is not a problem in this case since we reset the buffer after detectBOM?

internal/component/loki/source/file/internal/tail/file_test.go
kalleep: No encoders _and_ decoders are not stateful, if you use writer or reader that will have state e.g. `encoder.Writer(f)`.

But calling e.g. `encoder.String` if fine and will not store any state.

internal/component/loki/source/file/internal/tail/file_test.go
kalleep: Sure I can add a test for that

internal/component/loki/source/file/internal/tail/bom.go
kalleep: yes!

internal/component/loki/source/file/internal/tail/util_test.go
kalleep: That would be for the other test, this one will just extract position after last nl found

PR #5433, opened 2026-02-04
feat(pyroscope.java): expand async-profiler options

ivanape:

Expand async-profiler options for `pyroscope.java`

### Pull Request Details

This PR expands the configuration options available in the `pyroscope.java` component to expose more async-profiler capabilities, allowing users to leverage additional async-profiler features.

Almost all available options from async-profiler 4.3.0 have been added. If any of these options don't make sense from Alloy's perspective, I'm happy to make the necessary code changes.

## Configuration example

```alloy
pyroscope.java "example" {
    forward_to = [pyroscope.write.endpoint.receiver]
    
    targets = discovery.kubernetes.pods.targets
    
    profiling_config {
        interval = "60s"
        cpu      = true
        event    = "itimer"
        
        alloc      = "512k"
        live       = true
        native_mem = "1m"
        
        lock        = "10ms"
        native_lock = "5ms"
        
        include = ["com.example.*", "org.myapp.*"]
        exclude = ["*Unsafe.park*"]
        
        jstackdepth = 1024
        cstack      = "dwarf"
        features    = ["stats", "vtable"]
    }
}
```

### Notes to the Reviewer

- Removed the `Validate()` method as it appears it wasn't being used
- Changed default values for `alloc` (from `10ms` to `512k`) and `lock` (from `512k` to `10ms`)
- 
### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+| `jstackdepth`   | `int`          | Maximum Java stack depth to capture.                                                              | `2048`     | no       |
+| `cstack`        | `string`       | C stack walking mode: `fp`, `dwarf`, `lbr`, `vm`, `vmx`, or `no`.                                 | `""`       | no       |
+| `features`      | `list(string)` | Stack walking features to enable (e.g., `["stats", "vtable"]`).                                   | `[]`       | no       |
+| `trace`         | `list(string)` | Java methods to trace with optional latency threshold (e.g., `["my.pkg.Method:50ms"]`).           | `[]`       | no       |
+| `jfrsync`       | `string`       | Start Java Flight Recording with specified configuration (e.g., `"default"`, `"profile"`).        | `""`       | no       |
+| `signal`        | `string`       | Alternative signal for profiling (e.g., `"SIGUSR1"` or `"SIGUSR1/SIGUSR2"` for CPU/WALL).         | `""`       | no       |
+| `clock`         | `string`       | Clock source for JFR timestamps: `tsc` or `monotonic`.                                            | `"tsc"`    | no       |
```
clayton-cornell:
```suggestion
| Name          | Type           | Description                                                                                             | Default    | Required |
| ------------- | -------------- | ------------------------------------------------------------------------------------------------------- | ---------- | -------- |
| `all`         | `bool`         | Enable all profiling types simultaneously: `cpu`, `wall`, `alloc`, `live`, `lock`, and native memory.   | `false`    | no       |
| `all_user`    | `bool`         | Include only user-mode events. This is helpful when kernel profiling is restricted.                     | `false`    | no       |
| `alloc`       | `string`       | Allocation profiling interval, for example, `512k` or `1m`. Empty string disables allocation profiling. | `"512k"`   | no       |
| `begin`       | `string`       | Automatically start profiling when specified native function is executed.                               | `""`       | no       |
| `clock`       | `string`       | Clock source for Java Flight Recording timestamps: `tsc` or `monotonic`.                                | `""`       | no       |
| `cpu`         | `bool`         | Enable CPU profiling using the event specified in `event`.                                              | `true`     | no       |
| `cstack`      | `string`       | C stack walking mode: `fp`, `dwarf`, `lbr`, `vm`, `vmx`, or `no`.                                       | `""`       | no       |
| `end`         | `string`       | Automatically stop profiling when specified native function is executed.                                | `""`       | no       |
| `event`       | `string`       | Profiling event: `cpu`, `itimer`, `wall`, `alloc`, `lock`, `nativemem`, `cache-misses`                  | `"itimer"` | no       |
| `exclude`     | `list(string)` | Exclude stack traces matching these patterns, for example, `["*Unsafe.park*"]`.                         | `[]`       | no       |
| `features`    | `list(string)` | Stack walking features to enable, for example, `["stats", "vtable"]`.                                   | `[]`       | no       |
| `filter`      | `string`       | Profile only specific thread IDs, for example, `"120-127,132"`. Only applicable in wall-clock mode.     | `""`       | no       |
| `include`     | `list(string)` | Include only stack traces matching these patterns, for example, `["Primes.*", "java/*"]`.               | `[]`       | no       |
| `interval`    | `duration`     | How frequently to collect profiles from the targets.                                                    | `"60s"`    | no       |
| `jfrsync`     | `string`       | Start Java Flight Recording with specified configuration, for example, `"default"` or `"profile"`.      | `""`       | no       |
| `jstackdepth` | `int`          | Maximum Java stack depth to capture.                                                                    | `2048`     | no       |
| `live`        | `bool`         | Retain only live objects in allocation profiling. This is useful for finding memory leaks.              | `false`    | no       |
| `lock`        | `string`       | Lock profiling threshold, for example, `10ms` or `100us`. Empty string disables lock profiling.         | `"10ms"`   | no       |
| `log_level`   | `string`       | Async-profiler log level: `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, or `NONE`.                         | `"INFO"`   | no       |
| `native_lock` | `string`       | Native lock `pthread` profiling threshold, for example, `10ms`.                                         | `""`       | no       |
| `native_mem`  | `string`       | Native memory allocation profiling interval, for example, `1m`.                                         | `""`       | no       |
| `no_free`     | `bool`         | Don't record free calls in native memory profiling.                                                     | `false`    | no       |
| `nostop`      | `bool`         | Record profiling window between `begin` and `end` without stopping outside that window.                 | `false`    | no       |
| `per_thread`  | `bool`         | Profile threads separately. Each stack trace ends with a thread identifier frame.                       | `false`    | no       |
| `proc`        | `string`       | Collect system process statistics at specified interval, for example, `30s`.                            | `""`       | no       |
| `quiet`       | `bool`         | Suppress "Profiling started/stopped" log messages.                                                      | `false`    | no       |
| `record_cpu`  | `bool`         | Capture which CPU each sample was taken on.                                                             | `false`    | no       |
| `sample_rate` | `int`          | CPU profiling sample rate in Hz (samples per second).                                                   | `100`      | no       |
| `sched`       | `bool`         | Group threads by Linux scheduling policy (BATCH/IDLE/OTHER).                                            | `false`    | no       |
| `signal`      | `string`       | Alternative signal for profiling, for example, `"SIGUSR1"` or `"SIGUSR1/SIGUSR2"` for CPU and wall.     | `""`       | no       |
| `target_cpu`  | `int`          | Sample only threads running on specified CPU. Set to `-1` for all CPUs.                                 | `-1`       | no       |
| `trace`       | `list(string)` | Java methods to trace with optional latency threshold, for example, `["my.pkg.Method:50ms"]`.           | `[]`       | no       |
| `ttsp`        | `bool`         | Enable time-to-safepoint profiling.                                                                     | `false`    | no       |
| `wall`        | `string`       | Wall-clock profiling interval, for example, `100ms`. Profiles all threads regardless of status.         | `""`       | no       |
```

Remove brackets (not all were removed - some legacy ones were left as-is). Sort rows alphabetically. Align table columns for markdown readbility/consistency.

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+Refer to [profiler-options](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilerOptions.md) for more information about async-profiler configuration.
 
 #### `event`
 
-The `event` argument sets the CPU profiling event:
+The `event` argument specifies what to profile. It supports various event types:
+
+**Predefined events:**
 
 * `itimer` - Default. Uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) syscall, which generates a signal every time a process consumes CPU.
```
clayton-cornell:
```suggestion
* `itimer` - Default. Uses the [`setitimer(ITIMER_PROF)`](http://man7.org/linux/man-pages/man2/setitimer.2.html) system call, which generates a signal every time a process consumes CPU.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
 * `cpu` - Uses PMU-case sampling (like Intel PEBS or AMD IBS), can be more accurate than `itimer`, but it's not available on every platform.
-* `wall` - This samples all threads equally every given period of time regardless of thread status: Running, Sleeping, or Blocked.
+* `wall` - Samples all threads equally every given period of time regardless of thread status: Running, Sleeping, or Blocked.
    For example, this can be helpful when profiling application start-up time or IO-intensive processes.
```
clayton-cornell:
```suggestion
   For example, this can be helpful when profiling application start-up time or I/O-intensive processes.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+* `lock` - Java lock profiling.
+* `nativemem` - Native memory allocation profiling.
+
+**Hardware events:**
+
+* `cache-misses`, `cycles`, `instructions`, `branch-misses`, and other PMU events.
+
+**Java method profiling:**
+
+* `ClassName.methodName` - Profile all invocations of a specific Java method (e.g., `java.util.Properties.getProperty`).
```
clayton-cornell:
```suggestion
* `ClassName.methodName` - Profile all invocations of a specific Java method, for example, `java.util.Properties.getProperty`.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+
+**Native function profiling:**
+
+* `<symbol>` - Profile calls to a native function (e.g., `strcmp`, `malloc`).
```
clayton-cornell:
```suggestion
* `<symbol>` - Profile calls to a native function, for example, `strcmp`, `malloc`.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+* `mem:<func>` - Hardware breakpoint on a function or memory address.
```
clayton-cornell:
```suggestion
* `mem:<func>` - Hardware break point on a function or memory address.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+
+**Kernel events:**
+
+* `<tracepoint>` - Kernel tracepoint (e.g., `syscalls:sys_enter_open`).
```
clayton-cornell:
```suggestion
* `<tracepoint>` - Kernel tracepoint, for example, `syscalls:sys_enter_open`.
```

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -127,28 +127,79 @@ The `profiling_config` block describes how async-profiler is invoked.
...
+* `kprobe:<func>` - Kernel probe on a function.
+* `uprobe:<func>` - Userspace probe on a function.
```
clayton-cornell:
```suggestion
* `uprobe:<func>` - User space probe on a function.
```

korniltsev-grafanista:
@ivanape I am not super excited to maintain a duplicate(need to be in sync with async-profiler) list of flags and documentation. Do you think instead  we can create an "ExtraAsyncProfilerArguments []string" and just append everything there for "power users"

We have this in our sdk https://github.com/grafana/pyroscope-java/blob/17bec4822cbb8e3daa18dc853b41b0f4bf54d528/agent/src/main/java/io/pyroscope/javaagent/config/Config.java#L306C20-L306C55

This way we don't have to maintain/introduce/deprecate alloy configuration options when upgrade async-profiler.

ivanape:
@korniltsev-grafanista,


I think this is a very good approach since both the code and the Alloy configuration are much simpler.

I've opened [a new PR](https://github.com/grafana/alloy/pull/5472) with these changes.

If you prefer, we can close this one and continue in the new one.

Thank you very much!

PR #5445, opened 2026-02-04
feat(database_observability.postgres): Add logs collector

gaantunes:
Adds a Postgres logs collector, at this first iteration focused on generating error count metrics.

This component supports reading stderr postgres logs configured with the log_line_prefix format `%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a`, supported by AWS RDS.

### DIFF
--- a/docs/sources/reference/compatibility/_index.md
+++ b/docs/sources/reference/compatibility/_index.md
@@ -234,6 +234,10 @@ The following components, grouped by namespace, _export_ Loki `LogsReceiver`.
 
 <!-- START GENERATED SECTION: EXPORTERS OF Loki `LogsReceiver` -->
 
+{{< collapse title="database_observability" >}}
+- [database_observability.postgres](../components/database_observability/database_observability.postgres)
+{{< /collapse >}}
+
 {{< collapse title="loki" >}}
 - [loki.echo](../components/loki/loki.echo)
 - [loki.enrich](../components/loki/loki.enrich)
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -35,6 +35,15 @@ You can use the following arguments with `database_observability.postgres`:
 | `enable_collectors`  | `list(string)`       | A list of collectors to enable on top of the default set.   |         | no       |
 | `exclude_databases`  | `list(string)`       | A list of databases to exclude from monitoring.             |         | no       |
 
+## Exports
+
+The following fields are exported and can be referenced by other components:
+
+| Name                  | Type            | Description                                                                      |
+|-----------------------|-----------------|----------------------------------------------------------------------------------|
+| `targets`             | `list(map(string))` | Targets that can be used to collect metrics from the component.               |
+| `error_logs_receiver` | `LogsReceiver`  | Receiver for PostgreSQL error logs that processes and exports metrics.           |
+
 The following collectors are configurable:
 
 | Name             | Description                                                           | Enabled by default |
@@ -43,6 +52,7 @@ The following collectors are configurable:
 | `query_samples`  | Collect query samples and wait events information.                    | yes                |
 | `schema_details` | Collect schemas, tables, and columns from PostgreSQL system catalogs. | yes                |
 | `explain_plans`  | Collect query explain plans.                                          | yes                |
+| `error_logs`     | Process PostgreSQL error logs and export metrics (always enabled).    | yes                |
 
 ## Blocks
 
@@ -134,6 +144,67 @@ The `azure` block supplies the identifying information for the database being mo
 |--------------------|------------|------------------------------------------------------|---------|----------|
 | `collect_interval` | `duration` | How frequently to collect information from database. | `"1h"`  | no       |
 
+## Error Logs Collector
+
+The `error_logs` collector processes PostgreSQL error logs in real-time and exports both structured logs and Prometheus metrics.
+Unlike other collectors, it **runs independently of the database connection** and starts immediately when the component is created.
+
+### Key Features
+
+- **Always-on collection**: Collects logs even when the database is unavailable
+- **RDS log format support**: Parses structured PostgreSQL logs using AWS RDS format
+- **SQLSTATE extraction**: Automatically extracts and classifies errors by SQLSTATE codes
+- **Prometheus metrics**: Exports detailed error metrics with labels for severity, SQLSTATE, database, user, and instance
+- **Format validation**: Validates log format and provides warnings for misconfigured log output
+
+### Exported Receiver
+
+The component exports an `error_logs_receiver` that can be consumed by log source components:
+
+- `loki.source.file` - for reading PostgreSQL log files
+- `loki.source.cloudwatch` - for reading CloudWatch Logs (RDS)
+- `otelcol.receiver.otlp` - for receiving OTLP logs
+
+### Metrics
+
+The error_logs collector exports the following Prometheus metrics:
+
+| Metric Name                                  | Type    | Description                                          | Labels                                                             |
+|----------------------------------------------|---------|------------------------------------------------------|--------------------------------------------------------------------|
+| `postgres_errors_total`                      | counter | Total PostgreSQL errors by severity and SQLSTATE     | `severity`, `sqlstate`, `sqlstate_class`, `database`, `user`, `instance`, `server_id` |
+| `postgres_error_log_parse_failures_total`    | counter | Number of log lines that failed to parse             | -                                                                  |
+
+### Required PostgreSQL Configuration
+
+For the error_logs collector to work correctly, PostgreSQL must be configured with the RDS log format:
+
+```sql
+-- Set log format (requires superuser or rds_superuser)
+ALTER SYSTEM SET log_line_prefix = '%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a';
+
+-- Reload configuration
+SELECT pg_reload_conf();
+```
+
+**For AWS RDS/Aurora**, this is typically the default format. Verify with:
+
+```sql
+SHOW log_line_prefix;
+```
+
+### Supported Log Format
+
+The collector expects PostgreSQL logs in the RDS format with these fields:
+
+```
+<timestamp>:<remote_host:port>:<user>@<database>:[<pid>]:<line>:<SQLSTATE>:<session_start>:<vtxid>:<txid>:<session_id>:<query><app><severity>: <message>
+```
+
+Example log line:
+```
+2026-02-02 21:35:40.130 UTC:10.24.155.141(34110):app_user@books_store:[32032]:2:40001:2026-02-02 21:33:19 UTC:25/112:0:693c34cb.2398::psqlERROR:  canceling statement due to user request
+```
+
 ## Example
 
 ```alloy
@@ -154,6 +225,16 @@ database_observability.postgres "orders_db" {
 prometheus.exporter.postgres "orders_db" {
   data_source_name   = "postgres://user:pass@localhost:5432/dbname"
   enabled_collectors = ["stat_statements"]
+}
+
+// Read PostgreSQL log files and forward to error_logs collector
+loki.source.file "postgres_logs" {
+  targets = [{
+    __path__ = "/var/log/postgresql/postgresql-*.log",
+    job      = "postgres-errors",
+  }]
+  
+  forward_to = [database_observability.postgres.orders_db.error_logs_receiver]
 }
 
 loki.relabel "orders_db" {
@@ -230,6 +311,7 @@ Replace the following:
 `database_observability.postgres` has exports that can be consumed by the following components:
 
 - Components that consume [Targets](../../../compatibility/#targets-consumers)
+- Components that consume [Loki `LogsReceiver`](../../../compatibility/#loki-logsreceiver-consumers)
 
 {{< admonition type="note" >}}
 Connecting some components may not be sensible or components may require further configuration to make the connection work correctly.
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -35,6 +35,8 @@ You can use the following arguments with `database_observability.postgres`:
 | `enable_collectors`  | `list(string)`       | A list of collectors to enable on top of the default set.   |         | no       |
 | `exclude_databases`  | `list(string)`       | A list of databases to exclude from monitoring.             |         | no       |
 
+[Data Source Name format]: https://pkg.go.dev/github.com/lib/pq#hdr-URL_connection_strings-NewConfig
+
 ## Exports
 
 The following fields are exported and can be referenced by other components:
@@ -201,18 +203,9 @@ Example log line:
 The `logs` collector uses a watermark file to track the last processed log timestamp. This prevents re-counting historical logs on component restart and maintains proper Prometheus counter semantics.
 
 **Default behavior:**
-- Watermark file: `<data_path>/dbo11y_pg_logs_watermark.txt`
+- Watermark file localtion: `<data_path>/dbo11y_pg_logs_watermark.txt`
 - On first run: Starts processing logs from the current time (skips historical logs)
 - On restart: Resumes from the last processed timestamp
-- Sync frequency: Every 10 seconds (atomic writes for crash safety)
-
-**Benefits:**
-- `postgres_errors_total` maintains monotonically increasing values
-- No duplicate counting on Alloy restarts
-- Proper `rate()` and `increase()` calculations in Prometheus
-- Resumes from last position after downtime (no data loss)
-
-**Important:** When using log sources with `start_from` set to historical timestamps (e.g., `otelcol.receiver.awscloudwatch` with `start_from = "2026-01-01T00:00:00Z"`), the watermark ensures that historical logs are only counted once, even if Alloy restarts multiple times.
 
 ## Examples
 
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -35,26 +35,26 @@ You can use the following arguments with `database_observability.postgres`:
 | `enable_collectors`  | `list(string)`       | A list of collectors to enable on top of the default set.   |         | no       |
 | `exclude_databases`  | `list(string)`       | A list of databases to exclude from monitoring.             |         | no       |
 
-[Data Source Name]: format must adhere to the [pq library standards](https://pkg.go.dev/github.com/lib/pq#hdr-URL_connection_strings-NewConfig).
+[data_source_name]: format must adhere to the [pq library standards](https://pkg.go.dev/github.com/lib/pq#hdr-URL_connection_strings-NewConfig).
 
 ## Exports
 
 The following fields are exported and can be referenced by other components:
 
-| Name             | Type                | Description                                                                      |
-|------------------|---------------------|----------------------------------------------------------------------------------|
-| `targets`        | `list(map(string))` | Targets that can be used to collect metrics from the component.                  |
-| `logs_receiver`  | `LogsReceiver`      | Receiver for PostgreSQL logs that processes and exports error metrics.           |
+| Name            | Type                | Description                                                            |
+| --------------- | ------------------- | ---------------------------------------------------------------------- |
+| `logs_receiver` | `LogsReceiver`      | Receiver for PostgreSQL logs that processes and exports error metrics. |
+| `targets`       | `list(map(string))` | Targets that can be used to collect metrics from the component.        |
 
 The following collectors are configurable:
 
 | Name             | Description                                                           | Enabled by default |
 |------------------|-----------------------------------------------------------------------|--------------------|
+| `explain_plans`  | Collect query explain plans.                                          | yes                |
+| `logs`           | Process PostgreSQL logs and export error metrics.                     | yes                |
 | `query_details`  | Collect queries information.                                          | yes                |
 | `query_samples`  | Collect query samples and wait events information.                    | yes                |
 | `schema_details` | Collect schemas, tables, and columns from PostgreSQL system catalogs. | yes                |
-| `explain_plans`  | Collect query explain plans.                                          | yes                |
-| `logs`           | Process PostgreSQL logs and export error metrics.                     | yes                |
 
 ## Blocks
 
@@ -173,18 +173,18 @@ The logs collector exports the following Prometheus metrics:
 
 The collector expects PostgreSQL logs with these prefixed fields:
 
-```
+```text
 <timestamp>:<remote_host:port>:<user>@<database>:[<pid>]:<line>:<SQLSTATE>:<session_start>:<vtxid>:<txid>:<session_id>:<query><app><severity>: <message>
 ```
 
 Example log line:
-```
+```text
 2026-02-02 21:35:40.130 UTC:10.24.155.141(34110):app_user@books_store:[32032]:2:40001:2026-02-02 21:33:19 UTC:25/112:0:693c34cb.2398::psqlERROR:  canceling statement due to user request
 ```
 
 This is done by setting the log_line_prefix param to `%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a`.
 
-#### How to configure the log_line_prefix
+#### Configure the log_line_prefix
 
 **Self hosted Postgres server**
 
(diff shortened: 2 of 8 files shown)


docs/sources/reference/compatibility/_index.md
```diff
@@ -234,6 +234,10 @@ The following components, grouped by namespace, _export_ Loki `LogsReceiver`.
 
 <!-- START GENERATED SECTION: EXPORTERS OF Loki `LogsReceiver` -->
 
+{{< collapse title="database_observability" >}}
+- [database_observability.postgres](../components/database_observability/database_observability.postgres)
+{{< /collapse >}}
+
 {{< collapse title="loki" >}}
```
cristiangreco, line 241: hmm what is this change?

internal/component/database_observability/postgres/collector/error_logs.go
```diff
@@ -0,0 +1,359 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const (
+	ErrorLogsCollector = "error_logs"
+	OP_ERROR_LOGS      = "error_logs"
+)
```
cristiangreco, line 22: Generally speaking, why calling it `error_logs` and not just `logs_collector`?

internal/component/database_observability/postgres/component.go
```diff
@@ -183,6 +184,10 @@ type Component struct {
 	dbConnection *sql.DB
 	healthErr    *atomic.String
 	openSQL      func(driverName, dataSourceName string) (*sql.DB, error)
+
+	errorLogsReceiver  loki.LogsReceiver
+	errorLogsIn        chan loki.Entry
+	errorLogsCollector *collector.ErrorLogs // Always-running collector (no DB required)
```
cristiangreco: I understand that it's convenient to have an always-running collector so you don't have to set it up explicitely (like `connection_info`), but it seems to me that we're stretching the requirement a bit. Does the collector startup process really need to be so different than the rest? This introduces a bunch of unwanted complexity, e.g. like the `SystemID` that is set after startup.

If possible, this collector should be started like other collectors and just sit there to wait to receive logs on the channel. If `error_logs_receiver` is not wired, then the collector will just wait forever. wdyt?

docs/sources/reference/compatibility/_index.md
gaantunes: That is generated by make generate-docs, basically adds our component to the listing of those that exposes a `LogsReceiver`

internal/component/database_observability/postgres/component.go
gaantunes: The idea was to enable collecting logs even when the db is down, so any server level logs that help troubleshooting the db failure are parsed. But on a second thought, since SystemID is our unique id and we can't rely solely on instance name, it would be hard to match the actual server owning those logs.

It is a tradeoff, since customers using the instance level uniquely could gain from it. happy to take the other route.

internal/component/database_observability/postgres/collector/error_logs.go
gaantunes: Good point, will make this change.

internal/component/database_observability/postgres/component.go
cristiangreco: > The idea was to enable collecting logs even when the db is down, so any server level logs that help troubleshooting the db failure are parsed. But on a second thought, since SystemID is our unique id and we can't rely solely on instance name, it would be hard to match the actual server owning those logs.

Hmm yeah I like the idea in principle, but I'd postpone it and start with a simpler implementation.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,6 +144,57 @@ The `azure` block supplies the identifying information for the database being mo
 |--------------------|------------|------------------------------------------------------|---------|----------|
 | `collect_interval` | `duration` | How frequently to collect information from database. | `"1h"`  | no       |
 
+## Logs Collector
+
+The `logs` collector processes PostgreSQL logs received through the `logs_receiver` entry point and exports Prometheus metrics for query and server errors.
+
+
```
```suggestion
The `logs` collector processes PostgreSQL logs received through the `logs_receiver` entry point and exports Prometheus metrics for query and server errors.

```

Remove extra blank line

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,6 +144,57 @@ The `azure` block supplies the identifying information for the database being mo
...
+- `otelcol.receiver.otlp` + `otelcol.exporter.loki` - receives OTLP logs and forwards to the receiver
+
+### Metrics
+
+The logs collector exports the following Prometheus metrics:
+
+| Metric Name                                  | Type    | Description                                          | Labels                                                             |
+|----------------------------------------------|---------|------------------------------------------------------|--------------------------------------------------------------------|
+| `postgres_errors_total`                      | counter | Total PostgreSQL errors by severity and SQLSTATE     | `severity`, `sqlstate`, `sqlstate_class`, `database`, `user`, `instance`, `server_id` |
+| `postgres_error_log_parse_failures_total`    | counter | Number of log lines that failed to parse             | -                                                                  |
```
```suggestion
| Metric Name                               | Type    | Description                                       | Labels                                                                                |
| ----------------------------------------- | ------- | ------------------------------------------------- | ------------------------------------------------------------------------------------- |
| `postgres_errors_total`                   | counter | Total PostgreSQL errors by severity and SQLSTATE. | `severity`, `sqlstate`, `sqlstate_class`, `database`, `user`, `instance`, `server_id` |
| `postgres_error_log_parse_failures_total` | counter | Number of log lines that failed to parse.         | -                                                                                     |
```
Minor tidy to add fullstops and align columns (just to be tidy in source)

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,6 +144,57 @@ The `azure` block supplies the identifying information for the database being mo
...
+
+```sql
+-- Set log format (requires superuser or rds_superuser)
+ALTER SYSTEM SET log_line_prefix = '%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a';
+
+-- Reload configuration
+SELECT pg_reload_conf();
+```
+
+```sql
```
clayton-cornell, line 218:
````suggestion
```

Use the following SELECT statement to show and verify the current string format applied to the beginning of each log line.

```sql
````

We should explain why the SELECT is given here.  I took a guess at the explanation...

File: internal/component/database_observability/postgres/component.go [hunk 7]
```diff
@@ -66,6 +67,7 @@ type Arguments struct {
 	EnableCollectors  []string            `alloy:"enable_collectors,attr,optional"`
 	DisableCollectors []string            `alloy:"disable_collectors,attr,optional"`
 	ExcludeDatabases  []string            `alloy:"exclude_databases,attr,optional"`
+	WatermarkPath     string              `alloy:"watermark_path,attr,optional"`
```
cristiangreco: I think we can move this under a dedicated config for the `logs` collector, so it'll be

```
logs {
  watermark_path = ...
}
```

internal/component/database_observability/postgres/component.go
```diff
@@ -589,6 +620,17 @@ func (c *Component) CurrentHealth() component.Health {
...
+func (c *Component) getWatermarkPath() string {
+	// If user specified a path, use it
+	if c.args.WatermarkPath != "" {
+		return c.args.WatermarkPath
+	}
+
+	// Default to <data_path>/dbo11y_pg_logs_watermark.txt
+	return filepath.Join(c.opts.DataPath, "dbo11y_pg_logs_watermark.txt")
+}
+
```
cristiangreco: Maybe use just `logs_watermark.txt` as file name, given that the file will live under a folder named after the component anyway, so pretty recognizable.

cristiangreco: Also, I think we can drop these comments :)

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+			}
+		}
+	}
+}
+
+func (l *Logs) processLogLine(entry loki.Entry) error {
+	return l.parseTextLog(entry)
+}
+
+func (l *Logs) parseTextLog(entry loki.Entry) error {
```
cristiangreco: Can we merge `processLogLine` and `parseTextLog`?

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/runtime/logging/level"
+)
+
+const LogsCollector = "logs"
+
+// RDS log format: %m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a
+var rdsLogFormatRegex = regexp.MustCompile(
```
cristiangreco: We can call this just `logFormatRegex` for now.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	)
+
+	l.parseErrors = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "postgres_error_log_parse_failures_total",
+			Help: "Failed to parse log lines",
+		},
+	)
+
+	if l.registry != nil {
```
cristiangreco: In other collectors we assume the registry passed in is not nil, I think we can do the same here.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+
+	return nil
+}
+
+func (l *Logs) Stop() {
+	l.cancel()
+	l.stopped.Store(true)
+	close(l.watermarkQuit)
+	l.wg.Wait()
+}
```
cristiangreco, line 146: Let's also unregister metrics like we do in ConnectionInfo

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+		l.lastProcessedTimestamp.Store(l.startTime)
+	}
+
+	return l, nil
+}
+
+func (l *Logs) initMetrics() {
+	l.errorsBySQLState = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "postgres_errors_total",
```
cristiangreco: I think we should use either `database_observability_` or `pg_` as prefix, as it aligns with current aggregation exclusion rules.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+func NewLogs(args LogsArguments) (*Logs, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	l := &Logs{
+		logger:                 log.With(args.Logger, "collector", LogsCollector),
+		entryHandler:           args.EntryHandler,
+		instanceKey:            args.InstanceKey,
+		systemID:               args.SystemID,
```
cristiangreco: Sorry I didn't spot this on the first pass... we don't really need `systemID` here because the `server_id` label is automatically injected by the component to all metrics emitted by collectors.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	l.invalidLogsThisMinute++
+
+	// Emit warning once per minute if ALL logs were invalid
+	now := time.Now()
+	if now.Sub(l.lastFormatWarning) >= time.Minute {
+		if l.validLogsThisMinute == 0 && l.invalidLogsThisMinute > 0 {
+			level.Warn(l.logger).Log(
+				"msg", "all PostgreSQL error logs in the last minute had invalid format",
+				"invalid_count", l.invalidLogsThisMinute,
+				"expected_format", "%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a",
```
cristiangreco: Let's put the expected format in a const for readability

internal/component/database_observability/postgres/collector/logs.go
cristiangreco: Same for the `instance` label.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	return l, nil
+}
+
+func (l *Logs) initMetrics() {
+	l.errorsBySQLState = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "postgres_errors_total",
+			Help: "PostgreSQL errors by severity with database, user, SQLSTATE, and instance tracking",
+		},
+		[]string{"severity", "sqlstate", "sqlstate_class", "database", "user", "instance", "server_id"},
```
cristiangreco: Maybe let's use `datname` instead of `database` for consistency?

cristiangreco: Also, let's remove `instance` and `server_id` as mentioned in the other comment.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	stopped *atomic.Bool
+	wg      sync.WaitGroup
+
+	formatCheckMutex      sync.Mutex
+	lastFormatWarning     time.Time
+	validLogsThisMinute   int
+	invalidLogsThisMinute int
+
+	// Watermark tracking
+	watermarkPath          string
```
cristiangreco: I might be missing something here, but I've just found out that:
- `loki.source.file` already handles its own positions file
- `otelcol.receiver.awscloudwatch` has a `storage` config that allows adding a storage file for checkpointing.

Do we need this implementation? (even if it might be needed for other CSPs, I'd still leave it out of this PR)

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -0,0 +1,453 @@
...
+	line := entry.Entry.Line
+
+	if strings.HasPrefix(line, "{") {
+		var jsonLog struct {
+			Body string `json:"body"`
+		}
+		if err := json.Unmarshal([]byte(line), &jsonLog); err == nil && jsonLog.Body != "" {
+			line = jsonLog.Body
+		}
+	}
```
cristiangreco: In which case the log line can be a json?

cristiangreco (commented):
Left some comments.

internal/component/database_observability/postgres/component.go
gaantunes: On a second thought, I am removing this param. I don't think we should allow them to change the default path, since `data_path` itself is an alloy native configuration.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Problem is `otelcol.receiver.awscloudwatch` only allows you to set a specific `starts_from` value, which is fixed in time, so most probably we will be always counting past logs at first run.

gaantunes: This also frees us up from needing to configure different source components in a specific way to make the counts work right.

internal/component/database_observability/postgres/component.go
gaantunes: Will remove it and set it at a const.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Nice catch

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Ah yeah true, nice catch

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Nice catch as well, thanks.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: This is dead code, sorry.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -35,6 +35,17 @@ You can use the following arguments with `database_observability.postgres`:
 | `enable_collectors`  | `list(string)`       | A list of collectors to enable on top of the default set.   |         | no       |
 | `exclude_databases`  | `list(string)`       | A list of databases to exclude from monitoring.             |         | no       |
 
+[Data Source Name format]: https://pkg.go.dev/github.com/lib/pq#hdr-URL_connection_strings-NewConfig
```
clayton-cornell: Where is this used? This is a link definition, but I don't see it actually referenced anywhere.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -35,6 +35,17 @@ You can use the following arguments with `database_observability.postgres`:
...
+
+## Exports
+
+The following fields are exported and can be referenced by other components:
+
+| Name             | Type                | Description                                                                      |
+|------------------|---------------------|----------------------------------------------------------------------------------|
+| `targets`        | `list(map(string))` | Targets that can be used to collect metrics from the component.                  |
+| `logs_receiver`  | `LogsReceiver`      | Receiver for PostgreSQL logs that processes and exports error metrics.           |
```
```suggestion
| Name            | Type                | Description                                                            |
| --------------- | ------------------- | ---------------------------------------------------------------------- |
| `logs_receiver` | `LogsReceiver`      | Receiver for PostgreSQL logs that processes and exports error metrics. |
| `targets`       | `list(map(string))` | Targets that can be used to collect metrics from the component.        |
```

Alpha sort and tighten up spacing in the markdown table

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -43,6 +54,7 @@ The following collectors are configurable:
 | `query_samples`  | Collect query samples and wait events information.                    | yes                |
 | `schema_details` | Collect schemas, tables, and columns from PostgreSQL system catalogs. | yes                |
 | `explain_plans`  | Collect query explain plans.                                          | yes                |
+| `logs`           | Process PostgreSQL logs and export error metrics.                     | yes                |
```
```suggestion
| `explain_plans`  | Collect query explain plans.                                          | yes                |
| `logs`           | Process PostgreSQL logs and export error metrics.                     | yes                |
| `query_details`  | Collect queries information.                                          | yes                |
| `query_samples`  | Collect query samples and wait events information.                    | yes                |
| `schema_details` | Collect schemas, tables, and columns from PostgreSQL system catalogs. | yes                |
```

We can alpha sort to be consistent with other tables.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+The logs collector exports the following Prometheus metrics:
+
+| Metric Name                               | Type    | Description                                       | Labels                                                                                |
+| ----------------------------------------- | ------- | ------------------------------------------------- | ------------------------------------------------------------------------------------- |
+| `postgres_errors_total`                   | counter | Total PostgreSQL errors by severity and SQLSTATE. | `severity`, `sqlstate`, `sqlstate_class`, `database`, `user`, `instance`, `server_id` |
+| `postgres_error_log_parse_failures_total` | counter | Number of log lines that failed to parse.         | -                                                                                     |
+
+### Required PostgreSQL Configuration
+
+#### Supported Log Format
```
clayton-cornell, line 172: For future reference. We should avoid having 2 headings without any text between. We can leave it as-is here for this iteration.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+
+The collector expects PostgreSQL logs with these prefixed fields:
+
+```
```
````suggestion
```text
````

File: docs/sources/reference/components/database_observability/database_observability.postgres.md [hunk 24]
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+<timestamp>:<remote_host:port>:<user>@<database>:[<pid>]:<line>:<SQLSTATE>:<session_start>:<vtxid>:<txid>:<session_id>:<query><app><severity>: <message>
+```
+
+Example log line:
+```
```
````suggestion
```text
````

File: docs/sources/reference/components/database_observability/database_observability.postgres.md [hunk 25]
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+2026-02-02 21:35:40.130 UTC:10.24.155.141(34110):app_user@books_store:[32032]:2:40001:2026-02-02 21:33:19 UTC:25/112:0:693c34cb.2398::psqlERROR:  canceling statement due to user request
+```
+
+This is done by setting the log_line_prefix param to '%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a'.
+
+#### How to configure the log_line_prefix
```
```suggestion
#### Configure the log_line_prefix
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+```
+
+**AWS RDS**
+
+On AWS RDS, you cannot use `ALTER SYSTEM` commands. Instead, configure `log_line_prefix` via RDS Parameter Groups:
+
+1. Open the AWS RDS Console → Parameter Groups
+2. Create or modify your parameter group
+3. Set `log_line_prefix` to: `%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a`
+4. Apply the parameter group to your RDS instance
```
```suggestion
1. Create or modify your parameter group
1. Set `log_line_prefix` to: `%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a`
1. Apply the parameter group to your RDS instance
```

We should try to use "lazy numbering"

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+
+**Note:** Ensure CloudWatch Logs export is enabled for the `postgresql` log in your RDS instance settings.
```
```suggestion
{{< admonition type="note" >}}
Ensure CloudWatch Logs export is enabled for the `postgresql` log in your RDS instance settings.
{{< /admonition >}}
```

Always wrap notes, tips, warnings, and cautions with the admonition shortcode.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -134,28 +146,114 @@ The `azure` block supplies the identifying information for the database being mo
...
+```sql
+SHOW log_line_prefix;
+```
+
+### Watermark and Historical Log Processing
+
+The `logs` collector uses a watermark file to track the last processed log timestamp. This prevents re-counting historical logs on component restart and maintains proper Prometheus counter semantics.
+
+**Default behavior:**
+- Watermark file localtion: `<data_path>/dbo11y_pg_logs_watermark.txt`
```
```suggestion
- Watermark file location: `<data_path>/dbo11y_pg_logs_watermark.txt`
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -208,6 +306,138 @@ loki.write "logs_service" {
 }
 ```
 
+**Persistent storage:** Alloy's data path (`--storage.path`) must be persisted across restarts to maintain loki.source.file positions file.
```
```suggestion
**Persistent storage:** The {{< param "PRODUCT_NAME" >}} data path (`--storage.path`) must be persisted across restarts to maintain `loki.source.file` positions file.
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -208,6 +306,138 @@ loki.write "logs_service" {
...
+
+### With otelcol.receiver.awscloudwatch
```
```suggestion
### With `otelcol.receiver.awscloudwatch`
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -208,6 +306,138 @@ loki.write "logs_service" {
...
+
+This needs to be used with `--stability.level=experimental`
```
```suggestion
This requires `--stability.level=experimental`
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -208,6 +306,138 @@ loki.write "logs_service" {
...
+    url = sys.env("<GRAFANA_CLOUD_HOSTED_LOGS_URL>")
+    basic_auth {
+      username = sys.env("<GRAFANA_CLOUD_HOSTED_LOGS_ID>")
+      password = sys.env("<GRAFANA_CLOUD_RW_API_KEY>")
+    }
+  }
+}
+```
+
+**AWS Credentials:** The `otelcol.receiver.awscloudwatch` component requires AWS credentials. Configure via:
```
```suggestion
**AWS Credentials:** The `otelcol.receiver.awscloudwatch` component requires AWS credentials. Configure with:
```

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -208,6 +306,138 @@ loki.write "logs_service" {
...
+      "logs:GetLogEvents",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams"
+    ],
+    "Resource": "arn:aws:logs:*:*:log-group:/aws/rds/instance/*" // Place your log-group arn(s) here
+  }]
+}
+```
+
+**Persistent storage:** Both `otelcol.storage.file` directory and Alloy's data path (`--storage.path`) must be persisted across restarts to maintain CloudWatch state and watermarks.
```
```suggestion
**Persistent storage:** Both `otelcol.storage.file` directory and the {{< param "PRODUCT_NAME" >}} data path (`--storage.path`) must be persisted across restarts to maintain CloudWatch state and watermarks.
```

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Will remove the watermark as discussed.

docs/sources/reference/components/database_observability/database_observability.postgres.md
gaantunes: This is one of the Arguments on the table above.

cristiangreco (approved):

PR #5456, opened 2026-02-05
fix(loki.process): Prevent panic when stage config is invalid

kalleep:
### Pull Request Details
It is possible to trigger a panic when reloading config for `loki.process`. If you update syntactically correct alloy config but invalid config internal to loki.process you will get a panic that we are trying to send on a closed channel.

E.g. updating alloy with 
```
stage.match {
    selector = "{}"
}
```

This is valid alloy syntax invalid logql expression. So when this happened we first stopped running pipeline but then failed to create the new one. Because we stopped current pipeline we also close [processIn channel](https://github.com/grafana/alloy/blob/main/internal/component/loki/process/process.go#L62). So when `loki.process` start receiving logs again we try to send on a closed channel resulting in a panic.

Added a test that fails on main. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/process_test.go
+++ b/internal/component/loki/process/process_test.go
@@ -61,7 +61,7 @@ func TestComponent(t *testing.T) {
 		err = ctrl.Update(Arguments{
 			ForwardTo: []loki.LogsReceiver{collector.Receiver()},
 			Stages: []stages.StageConfig{
-				stages.StageConfig{
+				{
 					MatchConfig: &stages.MatchConfig{
 						// {} is not a valid selector.
 						Selector: "{}",


internal/component/loki/process/stages/pipeline.go
```diff
@@ -73,36 +73,39 @@ func NewPipeline(logger log.Logger, stages []StageConfig, jobName *string, regis
...
-// RunWith will read from the input channel entries, mutate them with the process function and returns them via the output channel.
-func RunWith(input chan Entry, process func(e Entry) Entry) chan Entry {
-	out := make(chan Entry)
-	go func() {
-		defer close(out)
-		for e := range input {
-			out <- process(e)
+// Start will start the pipeline and forward entries to next.
+// The returned EntryHandler should be used to pass entries through the pipeline.
+func (p *Pipeline) Start(next chan<- loki.Entry) loki.EntryHandler {
```
kalleep, line 78: Changes done in this file is not necessary to the pr just tried to clear up some things.

internal/component/loki/process/process.go
```diff
@@ -138,16 +131,17 @@ func (c *Component) Update(args component.Arguments) error {
...
 	if stagesChanged(c.stages, newArgs.Stages) || c.stages == nil {
-		if c.entryHandler != nil {
-			c.entryHandler.Stop()
-		}
-
 		pipeline, err := stages.NewPipeline(c.opts.Logger, newArgs.Stages, &c.opts.ID, c.opts.Registerer, c.opts.MinStability)
 		if err != nil {
 			return err
 		}
-		entryHandler := loki.NewEntryHandler(c.processOut, func() { pipeline.Cleanup() })
```
kalleep, line 149: No need to wrap in a entry handler here. pipeline will trigger cleanup when it's stopped. In fact this entry handler was just passed to pipeline.Wrap and stop was never called on that one.

internal/component/loki/process/stages/pipeline.go
kgeckhart: Did anything change functionally or was it mostly a rename + code move?

kalleep: Nothing changed, Rename and cleaned up the code a bit e.g. using wg.Go

PR #5465, opened 2026-02-05
ci: Attach release artifacts before release publish

### LINKED ISSUE #5462: Release Please to create the release artefacts prior to publishing the releases
### Component(s)

_No response_

### Request

Currently, most release artifacts (other than source code) are created after the release is already published. This means that right after the release is published it doesn't have usable artifacts that users can install. This request is to create the artifacts before the release is ready for publishing, so that when it's published all artifacts are attached.

### Use case

It could have a couple of advantages:

- If the creation of artifacts fails, then the new release won't be published
- People who monitor the repo for a new release can download it as soon as it's available

### Tip

jharvey10:

Merging a release-please PR will now create a *draft* release instead of a published one. Artifacts will now build based on that event and attach themselves to the PR once finished. Once this is done, the release can be published and have everything attached from the start.

Also updated some release shepherd docs.

Fixes #5462

### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/developer/shepherding-releases.md
+++ b/docs/developer/shepherding-releases.md
@@ -22,20 +22,25 @@ point, only critical fixes should be merged into the branch until the release is
 
 1.  For `Use workflow from`, select the release branch in question.
 2.  Make sure to uncheck the `Dry run` box.
-3.  (This will create a draft release on GitHub.)
-4.  Add any relevant changelog details to the RC draft release and publish it.
-5.  (This will trigger CI to build/publish release artifacts and attach them to the release.)
+3.  (This will trigger workflows to create a draft release on GitHub, build the release artifacts,
+    and attach them to the release.)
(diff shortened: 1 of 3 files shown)


docs/developer/shepherding-releases.md
```diff
@@ -164,7 +171,16 @@ are the steps you can take to do it manually:
...
+   ```
+6. **DO NOT MERGE THIS PR**. This is only needed to get a passing `zizmor` check for the new commit.
+7. Once `zizmor` reports green, run the following commands:
+   ```bash
+   git checkout main
    git push
    ```
-6. **Delete the Bypass** from `Important branches require pull requests (except trusted apps)` for
+8. The PR will automatically close itself.
+9. **Delete the Bypass** from `Important branches require pull requests (except trusted apps)` for
```
ptodev, line 185: It'd be nice if we didn't have to push directly to main 😅 Is it really required, given that there is a PR?

PR #5466, opened 2026-02-05
feat(otel): Add receivers used in the otel k8s helm chart presets

kgeckhart:

Adds missing components which are used in the OTel Collector k8s helm chart via presets (https://opentelemetry.io/docs/platforms/kubernetes/helm/collector/#presets). This should enable someone to use the alloy distro for monitoring k8s with OTel successfully.

### Notes to the Reviewer

I also included the `prometheusreceiver` which isn't necessary for the k8s use case but seemed useful in general.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
         # TODO: Try enabling caching later. It might use up too much disk space on runners so needs extra testing.
         cache: false
     - run: make generate-ui
-    - run: GO_TAGS="builtinassets promtail_journal_enabled" GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} GOARM= make alloy
+    - run: GO_TAGS="embedalloyui promtail_journal_enabled" GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} GOARM= make alloy
 
   build_linux_boringcrypto:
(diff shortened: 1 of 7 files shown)


blewis12:
Thanks @kgeckhart! The `prometheusreceiver` is part of the core distro and should have been included, but I seemed to have missed it - gonna do another check just to make sure I didn't miss anything else. The other additions also make sense.

Seems like we're getting an error on build, we had some issues with the prometheus exporter I remember - I can take a quick look today

blewis12:
while I still don't fully understand the issue with building, I can reproduce it locally when I include the `builtinassets` tag like we do when building on CI, with the error coming from [here](https://github.com/prometheus/prometheus/blob/3155c95c1f58542ed98d265b0bae6675df433f8a/web/ui/assets_embed.go#L24)

**edit** the `builtinassets` tag conflicts with a tag of the same name in the prometheus library, which was running the problematic code. Renaming our tag to get rid of the conflict fixes the issue

blewis12 (approved):
not sure if this review counts because I have commits in this PR lol, but lgtm

thampiotr (approved):

blewis12:
(not sure if we actually want to backport this if we plan to release the engine with v1.14 - gonna chat with Piotr and Joe today to align there)

PR #5470, opened 2026-02-06
feat(local.file_match, loki.source.file): Match multiple files using doublestar `{...}` expressions

### LINKED ISSUE #4423: Alloy local.file_match supporting include/exclude multiple path

ptodev:
### Pull Request Details

Upgrading to doublestar/v4. The new version of the library is more performant and has the new `{...}` matching syntax. 

However, it does have a few [breaking changes](https://github.com/bmatcuk/doublestar/blob/master/UPGRADING.md). The breaking change which affects us the most is not documented - apparently the library is now case-sensitive by default even on Windows.

Fixes #4423

### Notes to the Reviewer

I checked the tests on Windows by removing the case sensitivity glob option and they indeed failed:

Screenshot 2026-02-06 at 17 53 39
Screenshot 2026-02-06 at 17 53 01

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/local/file_match/testutil/path_windows.go
+++ b/internal/component/local/file_match/testutil/path_windows.go
@@ -11,9 +11,10 @@ import (
 // PathEndsWith checks if any target's __path__ ends with the given suffix.
 // On Windows, this is case-insensitive.
 func PathEndsWith(sources []discovery.Target, suffix string) bool {
+	suffix = strings.ToLower(suffix)
 	for _, s := range sources {
 		p, _ := s.Get("__path__")
-		if strings.HasSuffix(strings.ToLower(p), strings.ToLower(suffix)) {
+		if strings.HasSuffix(strings.ToLower(p), suffix) {
 			return true
 		}
 	}


internal/component/local/file_match/testutil/path_windows.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/grafana/alloy/internal/component/discovery"
+)
+
+// PathEndsWith checks if any target's __path__ ends with the given suffix.
+// On Windows, this is case-insensitive.
+func PathEndsWith(sources []discovery.Target, suffix string) bool {
+	for _, s := range sources {
+		p, _ := s.Get("__path__")
+		if strings.HasSuffix(strings.ToLower(p), strings.ToLower(suffix)) {
```
x1unix: wdyt to lowercase the suffix once?

ptodev: Good idea! I updated it.

internal/component/local/file_match/testutil/testutil.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/grafana/alloy/internal/component/discovery"
+)
+
+// WriteFile creates a test file with the given name in the specified directory.
+func WriteFile(t *testing.T, dir string, name string) {
+	t.Helper()
+	err := os.WriteFile(path.Join(dir, name), []byte("test content"), 0664)
+	require.NoError(t, err)
+	time.Sleep(20 * time.Millisecond)
```
kalleep, line 21: I know this was here before but did you identify why/if we need it?

internal/component/loki/source/file/resolver.go
```diff
@@ -82,7 +87,7 @@ func (s *globResolver) Resolve(targets []discovery.Target) iter.Seq[resolvedTarg
 
 			for _, m := range matches {
 				if exclude != "" {
-					if match, _ := doublestar.PathMatch(exclude, m); match {
+					if match, _ := s.globber.PathMatch(filepath.FromSlash(exclude), m); match {
```
kalleep, line 90: Should we move `filepath.FromSlash` usage into globber(s)?

internal/util/glob/glob_default.go
```diff
@@ -0,0 +1,21 @@
+//go:build !windows
```
kalleep, line 1: should we name this file `glob_unix`?

kalleep (commented):
Nice, great work. Just a few comments

internal/util/glob/glob_windows.go
```diff
@@ -0,0 +1,23 @@
...
+type windowsGlobber struct{}
+
+// NewGlobber creates a new Globber appropriate for the current platform.
+// On Windows, this returns a case-insensitive globber.
+func NewGlobber() Globber {
+	return &windowsGlobber{}
+}
+
+func (g *windowsGlobber) FilepathGlob(pattern string) ([]string, error) {
+	return doublestar.FilepathGlob(pattern, doublestar.WithCaseInsensitive())
```
x1unix, line 18: Why an interface is required here? Is it necessary to do tests/mocks?

If the iface isn't strictly necessary, would suggest having a wrapper func that checks `runtime.GOOS` for case sensitivity option.

Downside of having separate build-constrained files is that syntax errors/linter/formatters may not run <ins>locally</ins> against build-constrained files for windows.

clayton-cornell:
Docs are fine as-is

internal/util/glob/glob_default.go
ptodev: Would there be an advantage in doing so? If we do it then we'd have to also make a `_default` implementation for non-unix and non-windows platforms, which is in anycase not a platform we'd support. And I think it makes sense that the default implementation is case sensitive anyway.

internal/util/glob/glob_windows.go
ptodev: It feels more defensive than using the doublestar functions directly.

If we use the functions directly and only switch to swapper funcs wherever is needed, then it would be very easy to mistakenly call the wrong function in case the code is refactored or extended.

internal/component/local/file_match/testutil/testutil.go
ptodev: I think it's because the component polls the filesystem occasionally:
```
c.args.SyncPeriod = 10 * time.Millisecond
```

After the file is created and the test has slept for 10ms, the test will run and wait for 20ms for the file to be in the list of targets.

I'd prefer if we remove the sleep statements and instead wrap `getWatchedFiles` with a `requireEventually`. I could change this in a subsequent PR?

internal/component/loki/source/file/resolver.go
ptodev: I'm not sure. The reason I left it like this is so that it's more flexible if someone doesn't want the `FromSlash` in the future.

internal/util/glob/glob_default.go
kalleep: I think it's mostly to follow the naming scheme for these files that we have in all other places in alloy. It's not important but mostly I suggested it for consistency

internal/component/local/file_match/testutil/testutil.go
ptodev: Follow up PR: #5523

internal/util/glob/glob_default.go
ptodev: Sure, I don't mind :) I opened #5524 for it. I see that indeed there are a few cases in the codebase where we have this convention. Since we only support Unix and Windows platforms, having two files is ok.

PR #5472, opened 2026-02-06
feat(pyroscope): Add support for extra async-profiler CLI arguments

ivanape:

As suggested [in the previous PR](https://github.com/grafana/alloy/pull/5433#issuecomment-3860674758), this PR adds support for any `asprof` arguments.

### Pull Request Details

Configuration example

```
  profiling_config {
    interval = "15s"
    cpu = false
    event = "itimer"
    sample_rate = 200
    alloc = "512k"
    lock = "1ms"
    extra_arguments = [
      "--live",
      "--wall",
      "10ms",
      "--nativemem", "512k",
      "--jfrsync", "profile",
      "-j", "100",      
      "-F", "stats,vtable,comptask",
      "--clock", "monotonic",
      "--cstack", "dwarf",
    ]
  }
```

### Notes to the Reviewer

- I've had doubts about how to handle the cpu property and its relationship with event. Right now it simply overwrites the event that the user has set to have the same meaning, but I'm not sure if it's the best approach.
- I've also added a note (that I had forgotten in the first PR) explaining why the `perthread` flag doesn't apply when exporting to JFR.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated


internal/component/pyroscope/java/args.go
```diff
@@ -38,27 +38,23 @@ func (rc *Arguments) UnmarshalAlloy(f func(any) error) error {
 }
 
 func (arg *Arguments) Validate() error {
-	switch arg.ProfilingConfig.Event {
-	case "itimer", "cpu", "wall":
-		return nil
-	default:
-		return fmt.Errorf("invalid event: '%s'. Event must be one of 'itimer', 'cpu' or 'wall'", arg.ProfilingConfig.Event)
-	}
+	return nil
```
korniltsev-grafanista, line 41: why remove validation? is it not used?

korniltsev-grafanista (approved):
lgtm with one nit regarding validation

internal/component/pyroscope/java/args.go
ivanape: The validate method is not currently being used. Additionally, with the current version of async-profiler there are more defined events that can be used or use more than one event, i.e: `asprof -e cpu,alloc,lock -f profile.jfr ...`

ivanape:
Hi @korniltsev-grafanista, 

I see the tests are failing. It seems this is the cause:

image

The failing test uses this configuration:

```
	cfg := ProfilingConfig{
		Interval:   10 * time.Millisecond,
		SampleRate: 1000,
		CPU:        true,
		Event:      "cpu",
	}
```

~~The code of this PR overwrites the event to `itimer` if the user sets `cpu: true`.~~

~~This is related to the doubts about how to handle the `cpu` and `event` together, as the current async-profiler allows different kinds of profiles.~~

~~I can upload the changes shown in the screenshot, but I'd like to know your thoughts first because with this change, the `cpu` property will have no sense.~~

For backward compatibility, the `itimer` event is set as default when the user has not specified another event

Thanks!

clayton-cornell:
Docs are OK. It is a bid odd around the references to async-profiler, but as that's the upstream product name, it's OK as-is.

clayton-cornell (approved):

internal/component/pyroscope/java/loop.go
```diff
@@ -226,14 +226,19 @@ func (p *profilingLoop) start() error {
...
 	}
+	argv = append(argv, "-e", cfg.Event)
+	if cfg.PerThread {
+		argv = append(argv, "-t")
+	}
+	profilingInterval := time.Second.Nanoseconds() / int64(cfg.SampleRate)
+	argv = append(argv, "-i", strconv.FormatInt(profilingInterval, 10))
 	if cfg.Alloc != "" {
 		argv = append(argv, "--alloc", cfg.Alloc)
 	}
```
korniltsev-grafanista: before I try to understand this, why do we need this change?

ivanape: The primary reason for this code change is to decouple the async-profiler event (`-e`) from Alloy's `cpu` property. With async-profiler, you can specify events in various ways, including [Java methods](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilingModes.md#java-method-profiling), such as:

`-e java.util.Properties.getProperty`

Or [JVM methods](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilingModes.md#native-function-profiling):

`-e JVM_StartThread`

The same applies to the interval (`-i`) parameter, which [has different meanings](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilerOptions.md#general-options) depending on the profiling type, and is not tied exclusively to CPU profiling.


In the current PR implementation, the `cpu` attribute is just a fallback to the default `itimer` event when the user hasn't explicitly specified a configuration. For example:

```
  profiling_config {
    interval = "15s"
    cpu = true
  }
```

This will generate the command with the default event option:

`asprof -e itimer ....`

I felt this was the most conservative approach to maintain coherence with the current `pyroscope.java` behavior. However, I'm open to adjusting this if you'd prefer a different approach.

Best regards,

ivanape: Hi @korniltsev-grafanista,

Any chance you could revisit this? I'm available to make adjustments if needed. Thanks! 🙏

nachopv:
Hi @korniltsev-grafanista 

Just to double check , is there anything else preventing the workflow to run?

internal/component/pyroscope/java/loop.go
simonswine: I wonder if it would be helpful to make `extra_arguments` mutually exclusive with all the other flags that are there and passed into arguments to send to `asprof`, maybe `custom_arugments` (edit) might be the right name,

This would give the advanced users what they need and not confuse starters or break existing configs

Wdyt?

internal/component/pyroscope/java/loop.go
```diff
@@ -226,14 +226,19 @@ func (p *profilingLoop) start() error {
...
+		if cfg.Event == "" {
+			cfg.Event = "itimer"
 		}
-		profilingInterval := time.Second.Nanoseconds() / int64(cfg.SampleRate)
-		argv = append(argv, "-i", strconv.FormatInt(profilingInterval, 10))
 	}
+	argv = append(argv, "-e", cfg.Event)
+	if cfg.PerThread {
+		argv = append(argv, "-t")
+	}
```
simonswine: As you say "per thread" is not used for the jfr output format,  we probably should remove this and mark the argument deprecated.

simonswine (approved):
Thanks for your contribution this is useful!

LGTM. Just a couple of thoughts, we can revisit early next week with @korniltsev-grafanista

internal/component/pyroscope/java/args.go
```diff
@@ -38,27 +38,23 @@ func (rc *Arguments) UnmarshalAlloy(f func(any) error) error {
...
-			Lock:       "512k",
-			CPU:        true,
-			Event:      "itimer",
-			PerThread:  false,
-			LogLevel:   "INFO",
-			Quiet:      false,
+			Interval:       60 * time.Second,
+			SampleRate:     100,
+			Alloc:          "512k",
+			Lock:           "10ms",
```
aleks-p: Just to call this out, this seems to fix an existing bug with these two being swapped by default. We can keep it (but document it in the changelog or similar) or move it to a separate PR.

internal/component/pyroscope/java/loop.go
aleks-p: I could be wrong but It looks to me that this breaks the existing behavior when `cpu=false`. Since we always add `-i` and `-e` (defaulting to `itimer`), the user has no way to disable cpu profiling?

Maybe we keep the `if cfg.CPU {` as is on main? You should still be able to do everything you need then by setting `cpu=false` and passing everything you need via extra arguments?

aleks-p (commented):
Overall seems like a simple change that allows the component to do much more than today. I left a couple of comments, one for what seems to me like a breaking change.

Another thing we want to have before approving this is a unit test covering the addition of extra arguments.

internal/component/pyroscope/java/loop.go
aleks-p: > maybe `customer_arugments` might be the right name

I'd suggest `extra_async_profiler_arguments` for clarity or `ap_extra_arguments` if we want to be consistent with [pyroscope-java](https://github.com/grafana/pyroscope-java/blob/17bec4822cbb8e3daa18dc853b41b0f4bf54d528/agent/src/main/java/io/pyroscope/javaagent/config/Config.java#L306C20-L306C55).

ivanape:
Thanks, @aleks-p and @simonswine, for the review!

To avoid blocking the PR, should I wait for your internal discussion on the open questions, or start addressing the comments now?

korniltsev-grafanista (changes requested):
> I wonder if it would be helpful to make extra_arguments mutually exclusive with all the other flags that are there and passed into arguments to send to asprof, maybe custom_arugments (edit) might be the right name,
> 
> This would give the advanced users what they need and not confuse starters or break existing configs
> 

I like this idea, lets implement it.
@ivanape do you think it would suit your needs?

ivanape:
Hi @korniltsev-grafanista,

That works for me.

I have applied the changes, and now the configuration with `custom_arguments` looks like this:

```
  profiling_config {
    interval = "15s"
    custom_arguments = [
      "-e", "itimer,alloc,lock,wall,mutex,native",
      "-i", "5ms",
      "--alloc", "512k",
      "--lock", "1ms",
    ]
  }
```

I also added a `warning` in the docs about the `per_thread` argument.

Would you like me to also add a runtime warning log when `-t` is used in the Alloy configuration? I have seen the same approach used in other components.

Thanks!

korniltsev-grafanista (approved):

korniltsev-grafanista:
```

error[unpinned-uses]: unpinned action reference
  --> ./.github/workflows/deploy-pr-preview.yml:18:11
   |
18 |     uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High
   = help: audit documentation → https://docs.zizmor.sh/audits/#unpinned-uses

203 findings (1 ignored, 201 suppressed): 0 informational, 0 low, 0 medium, 1 high
zizmor-exit-code=14

```

not sure whatsup with the zizmor check, the action is not pinned on main https://github.com/grafana/alloy/blob/5a4d03b0b3afe82bcab40ba9e50b212800c32ea1/.github/workflows/deploy-pr-preview.yml#L18
@grafana/grafana-alloy-maintainers how do we merge this?

ivanape:
Hi @korniltsev-grafanista,

I just wanted to kindly check on the status of this PR. Please let me know if I can help with anything. Thanks!

korniltsev-grafanista:
Thank you for the contribution! Apologies for long reviews and CI failures. I will try to do better next time.

PR #5474, opened 2026-02-06
test: Add OTel engine via helm chart example in kind

thampiotr:

Adds the ability to deploy OpenTelemetry Collector helm chart using Alloy OTel Engine in `example/kind` local k8s cluster.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/example/kind/config/otelcol-helm/values.yaml
+++ b/example/kind/config/otelcol-helm/values.yaml
@@ -6,12 +6,3 @@ command:
   name: "bin/otelcol"
 
 mode: deployment
-
-config:
-  receivers:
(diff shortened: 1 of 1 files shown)


blewis12 (approved):
Nice, it's probably also worth linking this in the bug bash (i'll also update the devel there to be the same one mentioned here, just for alignment)

thampiotr:
> Nice, it's probably also worth linking this in the bug bash (i'll also update the devel there to be the same one mentioned here, just for alignment)

@blewis12 thanks, that's a good point, I've added it to our internal bug bash doc.

PR #5477, opened 2026-02-09
perf(loki.write): Pre-allocate buffers used for batch encoding

kalleep:
By pre-allocating buffers for requests we can reduce total amount of memory allocated by `loki.write`

### Pull Request Details
Before this change we allocated two buffers every time we encoded a batch:
* One temporary buffer to for the proto serialized data of `push.PushRequest`
* One buffer that snappy wrote to and we actually sent with the request.

Because a shard will only ever send one batch at a time we can allocate two buffers that we re-suse for each request. My first attempt to fix this was to fix the size check in a batch. We only check size of line and size of structured metadata when appending to a batch.

Ideally we should count the serialized bytes it would take to add an entry to a batch and labels. I made two attempts to do this. They where both correct but it consumed a lot of CPU to do this. If this had worked out we could always pre-allocate these two buffers and always re-used them _and_ `batch_size` would actually be respected. Currently we always go over the configured value a lot.

Because it's really expensive to keep a correct count I reverted back to the old behavior _and_ changed the API a bit. We now call `request` on batch to get the request we have built and number of entries it contains. We then adjust the buffers after the largest batch we have seen and reuse them. 

### Notes to the Reviewer
I did some other changes to batch:
* Move the check to see if we can add to a batch into `add` itself and handle errors at call sites instead.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/batch_test.go
+++ b/internal/component/common/loki/client/batch_test.go
@@ -90,7 +90,7 @@ func TestBatch_add(t *testing.T) {
 				{Labels: model.LabelSet{"type": "a"}, Entry: logEntries[1].Entry},
 				{Labels: model.LabelSet{"type": "b"}, Entry: logEntries[2].Entry},
 			},
-			expectedSizeBytes: batchSize(logEntries[0], logEntries[1], logEntries[0]),
+			expectedSizeBytes: batchSize(logEntries[0], logEntries[1], logEntries[2]),
 		},
 	}
 


internal/component/common/loki/client/batch.go
```diff
@@ -31,135 +31,115 @@ type SentDataMarkerHandler interface {
...
 type batch struct {
-	streams map[string]*push.Stream
-	// totalBytes holds the total amounts of bytes, across the log lines in this batch.
-	totalBytes int
-	createdAt  time.Time
-
+	// req is the push request holding streams and entries to send to Loki.
+	req *push.PushRequest
+	// index maps label strings to stream indices in req.Streams for fast lookup.
+	index map[string]int
```
kalleep: Not sure we even need this and could be a premature optimization. If it's faster or not would highly depend on how many streams usually are in a batch

internal/component/common/loki/client/batch.go
```diff
@@ -31,135 +31,115 @@ type SentDataMarkerHandler interface {
...
+// sizeBytes returns the current batch size in bytes.
 func (b *batch) sizeBytes() int {
-	return b.totalBytes
-}
-
-// sizeBytesAfter returns the size of the batch after the input entry
-// will be added to the batch itself
-func (b *batch) sizeBytesAfter(entry push.Entry) int {
-	return b.totalBytes + entrySize(entry)
+	return b.req.Size()
```
kalleep: Need to investigate how expensive this call we be. We call it every time we want to add to a batch. 

If it's is a bit expensive there are workaround for caching it but will leave that out for now

kalleep: Okay calling b.req.Size() on every add was really expensive so need to look into this

kalleep: Current implementation is still to expensive. We trade less memory for more cpu, not great.

internal/component/common/loki/client/batch_test.go
```diff
@@ -36,102 +37,78 @@ func TestBatch_MaxStreams(t *testing.T) {
...
-			inputEntries: []loki.Entry{
+		{
+			name: "multiple streams with multiple log entries",
+			entries: []loki.Entry{
 				{Labels: model.LabelSet{"type": "a"}, Entry: logEntries[0].Entry},
 				{Labels: model.LabelSet{"type": "a"}, Entry: logEntries[1].Entry},
 				{Labels: model.LabelSet{"type": "b"}, Entry: logEntries[2].Entry},
 			},
-			expectedSizeBytes: len(logEntries[0].Entry.Line) + len(logEntries[1].Entry.Line) + len(logEntries[2].Entry.Line),
+			expectedSizeBytes: batchSize(logEntries[0], logEntries[1], logEntries[0]),
```
ptodev: Why are we adding `logEntries[0]` twice?

internal/component/common/loki/entry.go
```diff
@@ -19,8 +19,14 @@ func (e *Entry) Clone() Entry {
 	}
 }
 
+// Returns the size of the entry in bytes.
 func (e *Entry) Size() int {
-	// FIXME(kalleep): this is not correct but it is how size has been calculated so far.
-	// Size of an entry is more than just the line.
-	return len(e.Line)
+	// FIXME(kalleep): This is not correct but computing
```
ptodev, line 24: Do you think there is a way to do a cheap rough estimation?

kalleep: This is the "rough estimation", but even for a single entry this is wrong because the size of timestamp is not considered and there is a bunch of stuff like tag and variable length encoding not considered. Also we don't considers labels at all, the are de-duplicated at stream level.

But I can create a issue for more accurately count entries in the batch.

internal/component/common/loki/client/batch_test.go
kalleep: That was a mistake, they are the same size so it worked but I fixed it.

kalleep:
>The PR is quite big and although I tried to understand most if it, honestly I can't verify every detail. Since you said you're bringing back old behaviour, I suppose this is already battle tested and not brand new code :)

They for the size calculation. I changed so that `add` is now doing the size check instead of having the caller do it _and_ we moved `encode` to it's own function and just return the built request from batch.

internal/component/common/loki/entry.go
ptodev: Honestly I think it's not necessary. We could just leave it like this and just reuse the buffer.

ptodev:
Thanks! LGTM

PR #5480, opened 2026-02-09
docs: Add endpoints deprecation note to docs

ptodev:
This is similar to a note in the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints).

If users use `endpoints`, Alloy will log a warning.

### DIFF
--- a/docs/sources/reference/components/discovery/discovery.kubernetes.md
+++ b/docs/sources/reference/components/discovery/discovery.kubernetes.md
@@ -158,7 +158,7 @@ If a Pod backs the endpoint, the component discovers all container ports of the 
 
 {{< admonition type="warning" >}}
 The Endpoints API is deprecated in Kubernetes v1.33+.
-It is recommended to use EndpointSlices instead and switch to the `endpointslice` role below.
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/discovery/discovery.kubernetes.md
```diff
@@ -156,6 +156,11 @@ The `endpoints` role discovers targets from listed endpoints of a service.
 The component discovers one target per port for each endpoint address.
 If a Pod backs the endpoint, the component discovers all container ports of the Pod as targets even if they aren't bound to an endpoint port.
 
+{{< admonition type="warning" >}}
+The Endpoints API is deprecated in Kubernetes v1.33+.
+It is recommended to use EndpointSlices instead and switch to the `endpointslice` role below.
```
clayton-cornell:
```suggestion
Use the EndpointSlice API instead, and switch to the `endpointslice` role below.
```

We can be more explicit and just say "Use this instead".

I also tweaked the text a little to align with https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/

PR #5404, opened 2026-01-30
docs: Iterate on the OTel Engine Documentation

blewis12:
Part of https://github.com/grafana/alloy/issues/4722

This addresses a couple points of feedback in the first iteration:

* This PR introduces a separate "Introduction" page, which has a high level explanation of the two engines, and outlines the motivation/benefits of the OTel Engine
* This main page has a list of components, so far I can't find a way to automate this - but using a dropdown component I feel like the information is more approachable. I've also linked to the OCB manifest just in case folks want to look in more detail
* I've included a "Getting Started" page which has concrete examples to get off the ground with the new engine. I've moved the examples away from the CLI docs page to deduplicate

PR #5487, opened 2026-02-10
chore: Cherry Pick OTel Engine New Components and Documentation

blewis12:
I'd like to cherry-pick the following:

* https://github.com/grafana/alloy/pull/5404 -> Adds OTel Documentation changes (required to avoid merge conflicts, as later commits add components which are then included in the docs)
* https://github.com/grafana/alloy/pull/5316 -> Adds all components from the default engine
* https://github.com/grafana/alloy/pull/5466 -> Adds a couple more components used in otel k8s helm chart presets. This PR also renames the `builtinassets` build tag to `embedalloyui` as the former conflicts with an underlying build tag in the otel prometheus receiver. This should not change any behavior
* https://github.com/grafana/alloy/pull/5473 -> Adds alias to make the OTel Engine play nicely with the upstream helm chart

The reason i'd like to cherry-pick this is so that v1.13.x can be properly used in the upcoming bug bash to test out the OTel Engine internally. This way folks can pull an official tagged version of Alloy, which will be included in our different installation flows and can be easily referenced in helm charts. This might not be necessary though since our next release is scheduled for around the time of the bug bash - gonna keep this in draft and either open it or close it based on a final decision

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,11 @@ RUN apt-get update \
 COPY --from=build --chown=${UID}:${UID} /src/alloy/build/alloy /bin/alloy
 COPY --chown=${UID}:${UID} example-config.alloy /etc/alloy/config.alloy
 
+# Provide /bin/otelcol compatibility entrypoint. Useful when using Alloy's OTel Engine with
+# OpenTelemetry Collector helm chart and other ecosystem tools that expect otelcol binary.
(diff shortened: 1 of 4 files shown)


blewis12:
It was decided this isn't necessary for an internal bug bash, and best that we stick to including only bug fixes/cve's in patches

PR #5490, opened 2026-02-10
docs: Add tree layout to Alloy component docs

clayton-cornell:
This PR adds the tree layout to the Alloy docs. It's a trial run/test to see if the new shortcode can be used in Alloy component docs.


docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -57,6 +57,24 @@ You can use the following blocks with `otelcol.connector.count`:
 The `>` symbol indicates deeper levels of nesting.
 For example, `spans` > `attributes` refers to an `attributes` block defined inside a `spans` block.
 
+<!-- prettier-ignore-start -->
+{{</* docs/api-tree */>}}
```
clayton-cornell:
```suggestion
{{< docs/api-tree >}}
```

docs/sources/reference/components/otelcol/otelcol.connector.count.md
```diff
@@ -57,6 +57,24 @@ You can use the following blocks with `otelcol.connector.count`:
...
+├── [`debug_metrics`][debug_metrics] \[optional\]
+├── [`logs`][logs] \[optional\]
+│   └── [`attributes`][attributes] \[optional\]
+├── [`metrics`][metrics] \[optional\]
+│   └── [`attributes`][attributes] \[optional\]
+├── [`spanevents`][spanevents] \[optional\]
+│   └── [`attributes`][attributes] \[optional\]
+└── [`spans`][spans] \[optional\]
+    └── [`attributes`][attributes] \[optional\]
+{{</* /docs/api-tree */>}}
```
clayton-cornell:
```suggestion
{{< /docs/api-tree >}}
```

ptodev (commented):
This is awesome! I love how we can show the blocks hierarchically. However, I wonder if we can also include arguments on the tree layout. I might be able to generate those layouts from schema files after #5539 is merged. Would you like to do it that way? It would make your job much easier, with a lot less manual work. Then we could show the treemap in a "config" section which shows not just the blocks but the whole config, with arguments and blocks. This feels much more natural to me.

clayton-cornell:
>  I wonder if we can also include arguments on the tree layout. 

I was wondering about that too. Since this is kind of experimental for how we might use it in the Alloy docs, there's room to improve.

> I might be able to generate those layouts from schema files after #5539 is merged. Would you like to do it that way?

Sure :-)  Anything that makes it easier in the long run. Any effort now to automate some of these things is worth it.

> we could show the treemap in a "config" section which shows not just the blocks but the whole config, with arguments and blocks. 

Yes, that would present the info nicely.

We could also provide feedback to the platform team. Maybe ask for a change or two to extend the shortcode capability? Or for a parallel shortcode that takes the API one as a baseline and adds the extra bits we'd like to see for Alloy doc needs?

ptodev:
> Maybe ask for a change or two to extend the shortcode capability?

The main reason I'm on the fence for this particular PR is that we lose the "description" field for the blocks. Is there a way to add this in?

clayton-cornell:
> we lose the "description" field for the blocks. Is there a way to add this in?

I think that's a must have for an implementation of this shortcode. It's something I plan to provide as feedback to the platform team.

The styling is a big step in the right direction... but it's not 100% there for what the Alloy docs need.

clayton-cornell:
Closing this PR. There is now a new shortcode that handles what was attempted in this PR, so this is no longer needed.

Issue #5506, opened 2026-02-11
Mark prometheus.write.queue as deprecated


kgeckhart:
### Add a description

Update in docs and log a warning if the component is in use

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5509, opened 2026-02-11
feat: Deprecate prometheus.write.queue

kgeckhart:

`prometheus.write.queue` is deprecated and will be removed in the future. We have decided to invest further in `prometheus.remote_write` over further investments in `prometheus.write.queue` as an outcome of #5039.

Resolves: #5506 

### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.write.queue.md
+++ b/docs/sources/reference/components/prometheus/prometheus.write.queue.md
@@ -16,7 +16,9 @@ title: prometheus.write.queue
 Metrics are sent over the network using the [Prometheus Remote Write protocol][remote_write-spec].
 
 {{< admonition type="note" >}}
-This component is deprecated and will be removed in a future version of {{< param "PRODUCT_NAME" >}}. Please use `prometheus.remote_write` instead.
+Starting with {{< param "PRODUCT_NAME" >}} v1.14, `prometheus.write.queue` is deprecated.
+It will be removed in a future release.
+Use `prometheus.remote_write` instead.
 {{< /admonition >}}
 
 You can specify multiple `prometheus.write.queue` components by giving them different labels.


docs/sources/reference/components/prometheus/prometheus.write.queue.md
```diff
@@ -15,6 +15,10 @@ title: prometheus.write.queue
 `prometheus.write.queue` collects metrics sent from other components into a Write-Ahead Log (WAL) and forwards them over the network to a series of user-supplied endpoints.
 Metrics are sent over the network using the [Prometheus Remote Write protocol][remote_write-spec].
 
+{{< admonition type="note" >}}
+This component is deprecated and will be removed in a future version of {{< param "PRODUCT_NAME" >}}. Please use `prometheus.remote_write` instead.
```
clayton-cornell: We shoudl update the wording here to follow the guidelines here https://grafana.com/docs/writers-toolkit/write/deprecate-remove/

So something like this... assuming that it'll be deprecated in v1.14

```suggestion
Starting with {{< param "PRODUCT_NAME" >}} v1.14, `prometheus.write.queue` is deprecated.
It will be removed in a future release.
Use `prometheus.remote_write` instead.
```

internal/component/prometheus/write/queue/component.go
```diff
@@ -26,6 +28,8 @@ func init() {
 }
 
 func NewComponent(opts component.Options, args Arguments) (*Queue, error) {
+	level.Warn(opts.Logger).Log("msg", "prometheus.write.queue is deprecated and will be removed in a future version. Migrate to prometheus.remote_write to prevent future errors.")
```
kalleep, line 31: I know we have talked about deprecation of components in general. Maybe we could add support for that during component registration.

Not required for that in this pr but something we should consider at lest

clayton-cornell (approved):
Approving for docs

internal/component/prometheus/write/queue/component.go
kgeckhart: I ended up trying this out with a deprecated stability level. I might push it for feedback but I didn't feel great about it. I had to make it an "internal" stability level equivalent to Experimental which I think makes sense just felt a bit awkward.

kgeckhart:
I did end up changing this from chore -> feat so it shows up in the changelog.

PR #5511, opened 2026-02-12
docs: Restructure Modules topic and add more info for new users

clayton-cornell:
Restructure Modules topic and add more info for new users

Add info about when to use Modules and set some expectations about what modules can/can't do.

Shuffle info into a more logical flow/order, and simplify some statements.

Fixes: https://github.com/grafana/support-escalations/issues/15661


docs/sources/get-started/modules.md
```diff
@@ -141,6 +153,32 @@ This example shows how:
 1. **Exports enable connections**: The `filter_input` export allows other components to send data to the filter.
 1. **Component references work across modules**: The main configuration references exports from the imported module using `helpers.log_filter.default.filter_input`.
 
+## Import modules
+
+To use modules in your configuration, you import them using one of the `import` configuration blocks.
+The component controller processes import statements during configuration loading to make custom components available in the importing module's namespace.
```
ptodev, line 162: This line uses the term "namespace", which is defined in the next paragraph. Should this line be moved to the end of the section? That way it'll appear after the term definition.

clayton-cornell: I simplified the first sentence. That way the info flow remains in place... and namespaces are introduced later.

ptodev: Honestly I think the previous sentence was more meaningful. Let's just revert the latest change and just leave it as it was? It's not too bad and I'm not sure how to make it sound much better.

docs/sources/get-started/modules.md
```diff
@@ -19,47 +19,59 @@ Modules enable powerful configuration management capabilities:
 1. **Sharing**: Distribute configurations across teams and projects.
 1. **Composition**: Combine multiple modules to build sophisticated data collection systems.
 
+{{< admonition type="note" >}}
+Most {{< param "PRODUCT_NAME" >}} users don't need modules when they first get started.
```
dehaansa, line 23: One interesting thing is that basic understanding of modules is pretty useful for users just getting started with Fleet Management, might be worth a mention.

docs/sources/get-started/modules.md
```diff
@@ -141,6 +153,32 @@ This example shows how:
...
+
+You import modules into a _namespace_.
+This exposes the top-level custom components of the imported module to the importing module.
+The label of the import block specifies the namespace of an import.
+
+For example, if a configuration contains a block called `import.file "my_module"`, then custom components defined by that module appear as `my_module.CUSTOM_COMPONENT_NAME`.
+Namespaces for imports must be unique within a given importing module.
+
+{{< admonition type="warning" >}}
+If you use a label for an `import` or `declare` block that matches a built-in component namespace, such as `prometheus` or `mimir`, the built-in namespace becomes shadowed and unavailable in your configuration.
```
dehaansa, line 181: I wasn't aware of this, IMO this is pretty bad, we should consider blacklisting these namespaces... Not a comment on this PR but on the behavior.

docs/sources/get-started/modules.md
clayton-cornell: Definitely.... added a brief mention about FM and linked to the docs.

docs/sources/get-started/modules.md
clayton-cornell: Should we spawn off an Issue on this so it isn't forgotten?

docs/sources/get-started/modules.md
clayton-cornell: Done.  I'm not sure how to improve it either... I'll leave is as-is for now and think about possible content shuffle. It's hard to shuffle too much without losing the context and flow.

ptodev (approved):

PR #5512, opened 2026-02-12
chore: Remove name from loki.process stages

kalleep:
This is not used at all so lets just clean it up.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


ptodev (approved):
I suppose it's a remnant from Promtail which used to be used there 🤔

kalleep:
>I suppose it's a remnant from Promtail which used to be used there

Yes that seems to be the case

PR #5514, opened 2026-02-12
chore: Cleanup channel usage in loki.process

kalleep:
### Pull Request Details
Channel usage within `loki.process` is really messy. I decided clean up the outer most layer of this component and reuse shared functions we have.

Notable changes:
1. processIn is now static for the life-cycle of the component so we no longer need to take a lock when reading from exported channel and passing it down into the pipeline.
2. Use source.Drain when component is stopping. Before we only ever finished the forward routine after we had stopped the pipeline. While this worked in most cases it would not work if alloy is shutting down and this would lock. So using source.Drain is safer.
3. Pre allocate the extracted map to size of entry labels. We always add these all labels to extracted when passing through pipeline.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/process.go
+++ b/internal/component/loki/process/process.go
@@ -190,6 +190,10 @@ func (c *Component) handleIn(ctx context.Context) {
 			case <-ctx.Done():
 				return
 			case c.processIn.Chan() <- entry.Clone():
+				// TODO(@tpaschalis) Instead of calling Clone() at the
+				// component's entrypoint here, we can try a copy-on-write
+				// approach instead, so that the copy only gets made on the
+				// first stage that needs to modify the entry's labels.
 			}
 		}
 	}


internal/component/loki/process/stages/pipeline.go
```diff
@@ -73,8 +74,8 @@ func NewPipeline(logger log.Logger, stages []StageConfig, registerer prometheus.
 
 // Start will start the pipeline and forward entries to next.
 // The returned EntryHandler should be used to pass entries through the pipeline.
-func (p *Pipeline) Start(next chan<- loki.Entry) loki.EntryHandler {
-	handlerIn := make(chan loki.Entry)
+func (p *Pipeline) Start(in chan loki.Entry, out chan<- loki.Entry) loki.EntryHandler {
```
ptodev, line 77:
```suggestion
func (p *Pipeline) Start(in <-chan loki.Entry, out chan<- loki.Entry) loki.EntryHandler {
```
Should `in` have a direction?

internal/component/loki/process/stages/pipeline.go
```diff
@@ -83,24 +84,32 @@ func (p *Pipeline) Start(next chan<- loki.Entry) loki.EntryHandler {
...
-			pipelineIn <- Entry{
-				Extracted: map[string]any{},
-				Entry:     e,
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case e := <-in:
+				pipelineIn <- Entry{
+					// NOTE: When entires pass through the the pipeline
```
ptodev:
```suggestion
					// NOTE: When entires pass through the pipeline
```

internal/component/loki/source/consume.go
```diff
@@ -13,13 +13,43 @@ import (
 // of log entries from a component's internal handler to downstream receivers.
 // The fanout allows entries to be sent to multiple receivers concurrently.
 func Consume(ctx context.Context, recv loki.LogsReceiver, f *loki.Fanout) {
+	consume(ctx, recv, f, func(e loki.Entry) loki.Entry { return e })
+}
+
+// ConsumeAndProcess continuously reads log entries from recv, processes them using processFn,
+
```
ptodev:
```suggestion
```

internal/component/loki/process/process.go
```diff
@@ -48,18 +49,15 @@ var (
 // Component implements the loki.process component.
 type Component struct {
 	opts       component.Options
+	processOut loki.LogsReceiver
+	processIn  loki.LogsReceiver
 	receiver   loki.LogsReceiver
```
ptodev, line 54: It's weird how all these things are `LogsReceiver`. Should the code be refactored so that we use more specific types? Also, could a `LogsReceiver` change while we use something from it, e.g. a channel? Then we might end up using an old channel?

internal/component/loki/process/stages/pipeline.go
kalleep: Ideally it should but it cannot because we return loki.EntryHandler

internal/component/loki/process/process.go
kalleep: I mean it could also be plain channels, don't matter much.

>Also, could a LogsReceiver change while we use something from it, e.g. a channel?

Not sure I understand but before processIn was changed on every update but I changed it to be static for the lifetime of the component so we don't have use mutex when sending to it.

internal/component/loki/process/process.go
```diff
@@ -171,54 +185,11 @@ func (c *Component) handleIn(ctx context.Context, wg *sync.WaitGroup) {
...
 			))
+
 			select {
 			case <-ctx.Done():
 				return
-			case c.processIn <- entry.Clone():
-				// TODO(@tpaschalis) Instead of calling Clone() at the
-				// component's entrypoint here, we can try a copy-on-write
-				// approach instead, so that the copy only gets made on the
-				// first stage that needs to modify the entry's labels.
```
ptodev, line 181: I wonder why you deleted the TODO?

internal/component/loki/process/process.go
ptodev: Nevermind, I checked out the code locally to look at it and I understand it better now.

I just thought the naming of `loki.LogsReceiver` mis weird given that sometimes we use it to send from something and sometimes it's to send to something. But I guess having both `loki.LogsReceiver` and `loki.LogsSender` would be unnecessary.

Regarding changing the channels while they're used - I understand that you refactored it so that it doesn't happen, but I was wondering if can be done in a more defensive way so that someone doesn't accidentally change the code in a way that the channels get mutated incorrectly. But I don't have any concrete ideas to suggest, it's just a thought.

internal/component/loki/process/process.go
kalleep: Oh that was by mistake

Issue #439, opened 2023-05-30
Write documentation about using the Agent as a proxy for other Agents


ptodev:
### Request

In the last few months the topic of using the Agent as a proxy came up several times. It would be great to have a page in the documentation to which we could refer people on such occasions.

We could cover topics such as:

- Pushing vs pulling from the proxy Agents
- Load balancing (e.g. sharding)
- Replication (sometimes also called "high availability"). Running more than one Agent with the same series and avoiding out of order [errors](https://grafana.com/docs/mimir/latest/operators-guide/mimir-runbooks/#err-mimir-sample-out-of-order). Note that Agents running with a replication mode usually have extra settings such as [cluster and replica labels](https://grafana.com/docs/mimir/latest/configure/configure-high-availability-deduplication/#how-to-configure-prometheus).
- How much load is acceptable per agent. When should the customer spread the load to more than 1 proxy Agent?
- Which components to use for receiving the data in the proxy (e.g. [prometheus.receive_http](https://grafana.com/docs/agent/next/flow/reference/components/prometheus.receive_http/))

A few thoughts, to get us started:

- If the proxy agents expect data to be pushed to them, it could work via [loki.source.api](https://grafana.com/docs/agent/next/flow/reference/components/loki.source.api/) and [prometheus.receive_http](https://grafana.com/docs/agent/next/flow/reference/components/prometheus.receive_http/). However, I'm not sure what would then be used for load balancing.
- If the proxy agents pull the data from the other agents, then they could use normal Prometheus hashmod sharding for load balancing, but I'm not sure what they could use for logs. I don't think Agents can pull logs from other Agents in the way they can pull metrics.
- Is [clustering](https://github.com/grafana/agent/pull/3151) going to work for signals other than metrics?
- The most likely recommended route seems to be a sticky sessions load balancer which redirects remote write requests from non-Proxy Agents to Proxy Agents.
  - Can we test this and show an example load balancer configuration?
  - Which signal types would not be supported? E.g. for profile signals (Phlare) I don't think the Agent supports being pushed to.
- Can we dogfood the recommended approach internally? Is there a good use case?

### Use case

There are multiple reasons why a proxy Agent would be needed. Examples include:

- Air gapped networks where the proxy Agent is in a DMZ
- Storing database credentials in only one place - with the proxy Agents

clayton-cornell:
This has been discussed a lot internally. Specific links to Slack discussions:

[...]

PR #5515, opened 2026-02-12
docs: Describe Alloy as proxy

clayton-cornell:
Add new topic discussing Alloy as a proxy. The information here was pulled primarily from internal Slack discussions. 

Fixes: https://github.com/grafana/alloy/issues/439

### DIFF
--- a/docs/sources/configure/proxy.md
+++ b/docs/sources/configure/proxy.md
@@ -17,7 +17,7 @@ In OpenTelemetry terminology, this deployment model is often referred to as _gat
 
 {{< admonition type="note" >}}
 The proxy configuration described here refers to using {{< param "PRODUCT_NAME" >}} as a telemetry proxy that aggregates and forwards telemetry between instances.
-It doesn't cover configuring {{< param "PRODUCT_NAME" >}} to use a corporate HTTP proxy for outbound traffic, such as `proxy_url` or `proxy_from_environment` in `remote_write` or `loki.write`.
+It doesn't cover configuring {{< param "PRODUCT_NAME" >}} to use a corporate HTTP proxy for outbound traffic, such as `proxy_url` or `proxy_from_environment` in [`prometheus.remote_write`](../../reference/components/prometheus/prometheus.remote_write/) or [`loki.write`](../../reference/components/loki/loki.write/).
 {{< /admonition >}}
 
 ## Before you begin
@@ -57,14 +57,14 @@ flowchart LR
   class EdgeAlloy,LoadBalancer,ProxyAlloy,Backend grafana
 {{< /mermaid >}}
 
-For metrics, edge instances push data using `prometheus.remote_write` to proxy instances running `prometheus.receive_http`.
-For logs, edge instances push data using `loki.write` to proxy instances running `loki.source.api`.
+For metrics, edge instances push data using [`prometheus.remote_write`](../../reference/components/prometheus/prometheus.remote_write/) to proxy instances running [`prometheus.receive_http`](../../reference/components/prometheus/prometheus.receive_http/).
+For logs, edge instances push data using [`loki.write`](../../reference/components/loki/loki.write/) to proxy instances running [`loki.source.api`](../../reference/components/loki/loki.source.api/).
 
 #### Sticky load balancing for metrics
 
 Sticky load balancing ensures that requests with the same identifier, such as a time series or trace ID, are consistently routed to the same backend instance.
 
-For Prometheus `remote_write` traffic, you must ensure consistent routing per time series.
+For Prometheus [`prometheus.remote_write`](../../reference/components/prometheus/prometheus.remote_write/) traffic, you must ensure consistent routing per time series.
 When different proxy instances receive samples for the same series, you encounter out-of-order sample errors, increased ingestion load, and write-ahead log (WAL) churn.
 
 {{< admonition type="warning" >}}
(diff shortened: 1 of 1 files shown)


docs/sources/configure/proxy.md
```diff
@@ -0,0 +1,293 @@
...
+The pull model doesn't apply to logs.
+Logs must use a push model.
+{{< /admonition >}}
+
+While technically possible for metrics, using proxy instances to scrape other {{< param "PRODUCT_NAME" >}} instances isn't recommended as a primary aggregation strategy.
+Push-based aggregation using `prometheus.remote_write` provides clearer scaling characteristics, simpler configuration management, and better compatibility with dynamic environments.
+
+## Configure metrics proxying
+
+Configure metrics proxying using the push pattern with `prometheus.remote_write` on edge instances and `prometheus.receive_http` on proxy instances.
```
dehaansa: This is just a repeat of line 46?

docs/sources/configure/proxy.md
```diff
@@ -0,0 +1,293 @@
...
+| Signal   | Push through proxy | Pull with sharding | Notes                                  |
+| -------- | ------------------ | ------------------ | -------------------------------------- |
+| Metrics  | Supported          | Supported          | Sticky routing required for push       |
+| Logs     | Supported          | Not supported      | Push only                              |
+| Traces   | Depends            | Generally no       | Use OpenTelemetry-compatible receivers |
+| Profiles | Limited            | No                 | Depends on backend ingestion model     |
+
+For traces, you typically configure edge instances to send data to an OpenTelemetry-compatible receiver, such as `otelcol.receiver.otlp`, on proxy instances.
+The proxy instances then export to the backend using an appropriate exporter.
+Trace proxying doesn't require sticky routing in the same way as Prometheus `remote_write`, but capacity planning and retry behavior still apply.
```
dehaansa: Stickiness is still relevant when doing aggregations and producing spanmetrics or service graph. In those cases the edge should use the loadbalancing exporter, or they an add unique labels and aggregate in promql (or adaptive).

docs/sources/configure/proxy.md
clayton-cornell: Oh.. yeah.. it is. Word blind is a very real thing some days.

I reworked that sentence to read:   `You can use the push pattern to proxy metrics between edge and proxy instances.` as a lead-in to the following H3 headings.

docs/sources/configure/proxy.md
clayton-cornell: I added a definition of sticky routing further up in the doc where the concept is introduced. Here I reworked the text to add more information and context - I used the rewording suggested by teh Claude agent to try and capture all the edge case info.

```
Basic trace forwarding doesn't require sticky routing, but if proxy instances run trace-derived components such as `otelcol.connector.spanmetrics` or `otelcol.connector.servicegraph`, you need consistent routing so all spans for a trace or service reach the same instance.
You can use `otelcol.exporter.loadbalancing` on the edge instances to route by trace ID or service name.
Alternatively, you can add a unique label per proxy instance and aggregate the resulting metrics in PromQL or Adaptive Metrics.
```

docs/sources/configure/proxy.md
```diff
@@ -0,0 +1,311 @@
...
+prometheus.remote_write "to_proxy" {
+  endpoint {
+    url = "https://<PROXY_LOAD_BALANCER>/api/v1/metrics/write"
+  }
+}
+```
+
+Replace the following:
+
+- _`<PROXY_LOAD_BALANCER>`_: The URL of your load balancer in front of the proxy {{< param "PRODUCT_NAME" >}} instances.
```
jharvey10, line 117:
```suggestion
- _`<PROXY_LOAD_BALANCER>`_: The URL of your load balancer in front of the proxy {{< param "PRODUCT_NAME" >}} instances.

{{< admonition type="note" >}}
`<PROXY_LOAD_BALANCER>` is the address to which edge instances send data.
Your load balancer must forward each request path to the port exposed by your proxy {{< param "PRODUCT_NAME" >}} instances.

For example:

- Metrics: forward `https://<PROXY_LOAD_BALANCER>/api/v1/metrics/write` to proxy instances listening on port `12345`
- Logs: forward `https://<PROXY_LOAD_BALANCER>/loki/api/v1/push` to proxy instances listening on port `3100`
{{< /admonition >}}
```

docs/sources/configure/proxy.md
```diff
@@ -0,0 +1,311 @@
...
+## Signal support
+
+The following table shows what patterns each signal type supports:
+
+| Signal   | Push through proxy | Pull with sharding | Notes                                  |
+| -------- | ------------------ | ------------------ | -------------------------------------- |
+| Metrics  | Supported          | Supported          | Sticky routing required for push       |
+| Logs     | Supported          | Not supported      | Push only                              |
+| Traces   | Depends            | Generally no       | Use OpenTelemetry-compatible receivers |
+| Profiles | Limited            | No                 | Depends on backend ingestion model     |
```
jharvey10: I did some AI-assisted digging and found that we have a GA component `pyroscope.receive_http` whose docs say:

> # `pyroscope.receive_http`
> 
> `pyroscope.receive_http` receives profiles over HTTP and forwards them to `pyroscope.*` components capable of receiving profiles.
> 
> The HTTP API exposed is compatible with both the Pyroscope HTTP ingest API and the pushv1.PusherService Connect API.
> This allows `pyroscope.receive_http` to act as a proxy for Pyroscope profiles, enabling flexible routing and distribution of profile data.

Does this imply more than a "Limited" here? Is there a 'paved path' that puts this at more of a "Supported"? I'm not really sure of the nuances here, but it *feels* like an undersell 😄

clayton-cornell: You're right.  I can slightly expand this to account for `pyroscope.write` → `pyroscope.receive_http` → `pyroscope.write`

PR #5517, opened 2026-02-12
fix: Add /FORCEREGISTRY flag to windows installer

kalleep:
### Pull Request Details
When using the windows installer to upgrade alloy it's not possible to update arguments, environment etc if it was previously written. I added a flag that will force set these in windows registry.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/set-up/install/windows.md
+++ b/docs/sources/set-up/install/windows.md
@@ -73,8 +73,8 @@ To do a silent install of {{< param "PRODUCT_NAME" >}} on Windows, perform the f
 - `/DISABLEREPORTING=<yes|no>` Disable [data collection][]. Default: `no`
 - `/DISABLEPROFILING=<yes|no>` Disable profiling endpoint. Default: `no`
 - `/ENVIRONMENT="KEY=VALUE\0KEY2=VALUE2"` Define environment variables for Windows Service. Default: ``
-- `/FORCEREGISTRY=yes` Force the installer to write install options such as `/STABILITY` and `/DISABLEREPORTING` to the Windows registry even when upgrading.
-  By default, the installer preserves registry values on upgrade, so options passed during an upgrade aren't applied unless you use this flag.
+- `/FORCEREGISTRY=yes` At the start of installation, delete all Alloy registry keys and then run the rest of the install as normal so that install options such as `/STABILITY` and `/DISABLEREPORTING` are written fresh.
+  By default, the installer preserves existing registry values on upgrade; use this flag to reset them and apply the options passed on the command line, and to remove any old or unused registry keys from previous versions.
 - `/RUNTIMEPRIORITY="normal|below_normal|above_normal|high|idle|realtime"` Set the runtime priority of the {{< param "PRODUCT_NAME" >}} process. Default: `normal`
 - `/STABILITY="generally-available|public-preview|experimental"` Set the stability level of {{< param "PRODUCT_NAME" >}}. Default: `generally-available`
 - `/USERNAME="<username>"` Set the fully qualified user that Windows uses to run the service. Default: `NT AUTHORITY\LocalSystem`
--- a/packaging/windows/install_script.nsis
+++ b/packaging/windows/install_script.nsis
@@ -72,6 +72,14 @@ Section "install"
   # Calls to functions like nsExec::ExecToLog below push the exit code to the
   # stack, and must be popped after calling.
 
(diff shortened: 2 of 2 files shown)


kalleep:
I don't have much experience with nsis scripts and maybe there is a better way to do this

docs/sources/set-up/install/windows.md
```diff
@@ -73,6 +73,7 @@ To do a silent install of {{< param "PRODUCT_NAME" >}} on Windows, perform the f
 - `/DISABLEREPORTING=<yes|no>` Disable [data collection][]. Default: `no`
 - `/DISABLEPROFILING=<yes|no>` Disable profiling endpoint. Default: `no`
 - `/ENVIRONMENT="KEY=VALUE\0KEY2=VALUE2"` Define environment variables for Windows Service. Default: ``
+- `/FORCEREGISTRY=yes` Force the installer to write install options (for example, `/STABILITY`, `/DISABLEREPORTING`) to the Windows registry even when upgrading over an existing installation. By default, existing registry values are left unchanged on upgrade, so new options passed during an upgrade are not applied unless you use this flag.
```
clayton-cornell:
```suggestion
- `/FORCEREGISTRY=yes` Force the installer to write install options such as `/STABILITY` and `/DISABLEREPORTING` to the Windows registry even when upgrading.
  By default, the installer preserves registry values on upgrade, so options passed during an upgrade aren't applied unless you use this flag.
```

clayton-cornell:
Doc changes look OK to me.

kalleep:
>This feels like a workaround. Can we just recommend to users to run the uninstaller if they want to reset the installation?

Yes that's the point. No if we uninstall the data directory will be removed so we lose WAL, position files etc. 

>then could we just add an installer option to delete all Alloy registry keys prior to the installation?

Sure that could work but then the updating alloy is two steps no? One to prune registry and then one to update it again or did you have something else in mind?

clayton-cornell:
Would it help to add a little clarification around upgrade behavior?  Maybe something like this (just below the install options and above the note)?

```
### Upgrade behavior

On a fresh install, the installer writes all options you pass to the Windows registry.

On an upgrade, the installer preserves registry values by default.
The installer ignores options you pass during an upgrade unless you use `/FORCEREGISTRY=yes`.

When you use `/FORCEREGISTRY=yes`, the installer writes all options to the registry.
To change one setting while preserving others, include all the options you want to keep along with the changed setting.
```

packaging/windows/install_script.nsis
```diff
@@ -176,16 +178,26 @@ FunctionEnd
...
 
 # InitializeRegistry initializes the keys in the registry that the service
-# runner uses. If the registry values already exist, they are not overwritten.
+# runner uses. If the registry values already exist, they are not overwritten,
+# unless /FORCEREGISTRY=yes was passed.
 Function InitializeRegistry
   !define REGKEY "HKLM\Software\GrafanaLabs\Alloy"
 
-  # Define the default key, which points to the service.
+  # Define the default key, which points to the service. Write when missing or when /FORCEREGISTRY=yes.
```
ptodev: Why not just have a `/FORCEREGISTRY` flag instead of `/FORCEREGISTRY=yes`? If the flag is not set, default it to false. 

I don't know what the standard way to approach this in Windows installers is.

kalleep: No me neither but I saw that we had `=yes` for another flag so [I followed that](https://github.com/grafana/alloy/blob/main/packaging/windows/install_script.nsis#L197)

kalleep:
>By default the "data" files are in a different directory from the one the Alloy executable resides in. I don't know if the uninstaller will remove it. If it doesn't then this could be a problem in itself, since people would expect the uninstaller to remove that too.

[It will](https://github.com/grafana/alloy/blob/18737d8396160aea7a4db0f2f69bf0e14c9172b6/packaging/windows/install_script.nsis#L272-L291)

kalleep:
>No, what I mean is that if FORCEREGISTRY is set, then install_script.nsis can delete all the Alloy registry keys at the start of the installation. Then the rest of the installation can proceed as normal. It'd make the script cleaner and easier to understand. I'd also clean up the registry from keys that are old or unused.

Sure I guess that could work

packaging/windows/install_script.nsis
ptodev: Oh okay... we could leave it as it is then 😆

kalleep:
@ptodev I change to the suggestion you made, much simpler

PR #5518, opened 2026-02-12
feat(database_observability): Introduce wait_event_v2 with pre-classified wait event type

gaantunes:
### Brief description

Introduces `wait_event_v2` — a cleaner alternative to the baseline `wait_event` op that pre-classifies `wait_event_type` in Alloy at emit time rather than computing it with a 6-level `label_replace` chain at query time.

Also adds a `database_observability_wait_event_seconds_total` Prometheus counter (MySQL only) as a pre-aggregated alternative for wide-window TotalWaitEventsTime queries that hit Loki's series limit.

Context and benchmark results: https://github.com/grafana/grafana-dbo11y-app/issues/2350

### What changed

**New op: `wait_event_v2`** (MySQL + PostgreSQL)
- Identical fields to `wait_event` (v1) plus `wait_event_type` pre-classified into `IO Wait`, `Lock Wait`, `Network Wait`, or `Other Wait` in the log body
- Controlled by the new `enable_pre_classified_wait_events` flag (replaces the experimental `enable_structured_metadata` flag)
- When the flag is ON, `wait_event_v2` is emitted **instead of** `wait_event` (v1) — never both

**Prometheus counter: `database_observability_wait_event_seconds_total`** (MySQL only)
- Labels: `server_id`, `digest`, `schema`
- Increments unconditionally on every wait event row, regardless of which Loki op is active
- Exposed at the component's `/metrics` endpoint via the existing registry

**Not changed**
- `query_sample`, `query_sample_v2`, `query_association`, `query_association_v2` — untouched

### Why pre-classification

Benchmarks (n=143–144 iterations, 30m–2h windows) showed:
- `wait_event_v2` delivers **−83 to −90% p95** on WaitEventsPanel duration queries by eliminating the Loki 500-series fallback retry loop
- The retry loop adds 2–7s of latency at wide unfiltered windows for v1
- SM-based versions (old v2/v3/v5) hit the same series limit at 2h+ windows for TotalWaitEventTime — net zero benefit

### PR Checklist

- [x] Documentation updated (`enable_pre_classified_wait_events` in both component docs)
- [x] Tests updated (mutual exclusion tests, `TestClassifyMySQLWaitEventType`, `TestQuerySamples_WaitEventCounter`)

To be used with https://github.com/grafana/grafana-dbo11y-app/pull/2693

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -36,7 +36,6 @@ You can use the following arguments with `database_observability.mysql`:
 | `enable_collectors`                        | `list(string)`       | A list of collectors to enable on top of the default set.                   |         | no       |
 | `exclude_schemas`                          | `list(string)`       | A list of schemas to exclude from monitoring.                               | `["alloydbadmin", "alloydbmetadata", "azure_maintenance", "azure_sys", "cloudsqladmin", "rdsadmin"]` | no       |
 | `allow_update_performance_schema_settings` | `boolean`            | Whether to allow updates to `performance_schema` settings in any collector. Enable this in conjunction with other collector-specific settings where required. | `false` | no       |
-| `enable_pre_classified_wait_events` | `boolean` | When `true`, emits `wait_event_v2` entries with `wait_event_type` pre-classified as `IO Wait`, `Lock Wait`, `Network Wait`, or `Other Wait` in the log body, instead of the baseline `wait_event` entries. | `false` | no |
 
 The following collectors are configurable:
 
@@ -169,6 +168,7 @@ The `gcp` block supplies the identifying information for the GCP Cloud SQL datab
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
 | `sample_min_duration`            | `duration` | Minimum duration for query samples to be collected. Set to "0s" to disable filtering and collect all samples regardless of their duration.| `"0s"`  | no       |
 | `wait_event_min_duration`        | `duration` | Minimum duration for a wait event to be collected. Set to "0s" to disable filtering and collect all wait events regardless of their duration.  | `"1us"` | no       |
+| `enable_pre_classified_wait_events` | `boolean` | When `true`, emits `wait_event_v2` entries with `wait_event_type` pre-classified as `IO Wait`, `Lock Wait`, `Network Wait`, or `Other Wait` in the log body, instead of the baseline `wait_event` entries. | `false` | no       |
 
 ### `setup_actors`
 
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -36,6 +36,7 @@ You can use the following arguments with `database_observability.mysql`:
 | `enable_collectors`                        | `list(string)`       | A list of collectors to enable on top of the default set.                   |         | no       |
 | `exclude_schemas`                          | `list(string)`       | A list of schemas to exclude from monitoring.                               | `["alloydbadmin", "alloydbmetadata", "azure_maintenance", "azure_sys", "cloudsqladmin", "rdsadmin"]` | no       |
 | `allow_update_performance_schema_settings` | `boolean`            | Whether to allow updates to `performance_schema` settings in any collector. Enable this in conjunction with other collector-specific settings where required. | `false` | no       |
+| `enable_pre_classified_wait_events` | `boolean` | When `true`, emits `wait_event_v2` entries with `wait_event_type` pre-classified as `IO Wait`, `Lock Wait`, `Network Wait`, or `Other Wait` in the log body, instead of the baseline `wait_event` entries. | `false` | no |
 
 The following collectors are configurable:
 
--- a/internal/component/database_observability/mysql/collector/query_samples.go
+++ b/internal/component/database_observability/mysql/collector/query_samples.go
@@ -492,19 +492,26 @@ func (c *QuerySamples) determineTimerClauseAndLimit(uptime float64) (string, flo
 	return timerClause, limit
 }
 
+// classifyMySQLWaitEventType maps a raw MySQL performance_schema wait event name
+// to a standardized category, aligned with the wait taxonomy used elsewhere:
+//
+//	IO Wait      = wait/io/(file|table).+
+//	Network Wait = wait/io/socket.+
+//	Lock Wait    = wait/(io/lock|synch|lock).+
 func classifyMySQLWaitEventType(waitEventName string) string {
 	rest, ok := strings.CutPrefix(waitEventName, "wait/")
 	if !ok {
-		return waitEventName
+		return "Other Wait"
 	}
-	category, _, _ := strings.Cut(rest, "/")
-	switch category {
-	case "io":
+	switch {
+	case strings.HasPrefix(rest, "io/file/"), strings.HasPrefix(rest, "io/table/"):
 		return "IO Wait"
-	case "lock":
+	case strings.HasPrefix(rest, "io/socket/"):
+		return "Network Wait"
+	case strings.HasPrefix(rest, "io/lock/"),
+		strings.HasPrefix(rest, "synch/"),
+		strings.HasPrefix(rest, "lock/"):
 		return "Lock Wait"
-	case "synch":
-		return "Synch Wait"
 	}
-	return waitEventName
+	return "Other Wait"
 }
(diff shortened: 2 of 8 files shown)


internal/component/database_observability/loki_entry.go
```diff
@@ -24,3 +24,59 @@ func BuildLokiEntryWithTimestamp(level logging.Level, op, line string, timestamp
...
+			continue
+		}
+		if enableStructuredMetadata {
+			smLabels = append(smLabels, push.LabelAdapter{Name: f.Name, Value: f.Value})
+		} else {
+			logLine += fmt.Sprintf(` %s="%s"`, f.Name, f.Value)
+		}
+	}
+
+	return loki.Entry{
```
kalleep: Can you use `loki.NewEntry`. We set created timestamp when using this so we can record how long it takes for an entry to be propagate through the pipeline.

gaantunes: I can look into that before I merge this branch.

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -108,6 +115,10 @@ type QuerySamples struct {
 	setupConsumersCheckInterval time.Duration
 	sampleMinDuration           time.Duration
 	waitEventMinDuration        time.Duration
+	enablePreClassifiedWaitEvents bool
+
+	waitEventCounter *prometheus.CounterVec
+	serverID         string
```
cristiangreco: `serverID` shouldn't be needed in the collectors, it's added automatically by the relabeling in the component.

internal/component/database_observability/postgres/collector/query_details.go
```diff
@@ -83,9 +83,9 @@ func NewQueryDetails(args QueryDetailsArguments) (*QueryDetails, error) {
 		excludeUsers:     args.ExcludeUsers,
 		entryHandler:     args.EntryHandler,
 		tableRegistry:    args.TableRegistry,
-		normalizer:       sqllexer.NewNormalizer(sqllexer.WithCollectTables(true), sqllexer.WithCollectComments(true), sqllexer.WithKeepIdentifierQuotation(true)),
-		logger:           log.With(args.Logger, "collector", QueryDetailsCollector),
-		running:          &atomic.Bool{},
+		normalizer:               sqllexer.NewNormalizer(sqllexer.WithCollectTables(true), sqllexer.WithCollectComments(true), sqllexer.WithKeepIdentifierQuotation(true)),
+		logger:                   log.With(args.Logger, "collector", QueryDetailsCollector),
+		running:                  &atomic.Bool{},
```
cristiangreco: These look like just whitespace changes?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -567,6 +578,38 @@ func (c *QuerySamples) buildWaitEventLabels(state *SampleState, we WaitEventOccu
...
+	)
+}
+
+// classifyPostgresWaitEventType returns a standardized classification for a raw PostgreSQL
+// wait event type. Currently returns the raw type unchanged.
+// TODO: implement classification mapping (e.g. "IO" -> "IO Wait", "Lock" -> "Lock Wait")
+func classifyPostgresWaitEventType(waitEventType string) string {
+	return waitEventType
+}
+
```
cristiangreco: This means that the postgres log line is not really changing with v2?

cristiangreco (commented):
left some questions

internal/component/database_observability/mysql/collector/query_samples.go
gaantunes: Addressed — removed `serverID` from the collector. The component now uses `CurryWith(prometheus.Labels{"server_id": serverID})` to pre-set the label before passing the curried counter to the collector.

internal/component/database_observability/postgres/collector/query_details.go
gaantunes: Addressed — reverted the unintentional extra whitespace on those lines.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Good catch — the passthrough was a TODO placeholder. Implemented the full mapping now: `IO` → `IO Wait`, `Lock`/`LWLock`/`Activity`/... → `Lock Wait`, `Client` → `Network Wait`, everything else → `Other Wait`. Also added `TestClassifyPostgresWaitEventType` to cover all branches.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -36,6 +36,7 @@ You can use the following arguments with `database_observability.mysql`:
 | `enable_collectors`                        | `list(string)`       | A list of collectors to enable on top of the default set.                   |         | no       |
 | `exclude_schemas`                          | `list(string)`       | A list of schemas to exclude from monitoring.                               | `["alloydbadmin", "alloydbmetadata", "azure_maintenance", "azure_sys", "cloudsqladmin", "rdsadmin"]` | no       |
 | `allow_update_performance_schema_settings` | `boolean`            | Whether to allow updates to `performance_schema` settings in any collector. Enable this in conjunction with other collector-specific settings where required. | `false` | no       |
+| `enable_pre_classified_wait_events` | `boolean` | When `true`, emits `wait_event_v2` entries with `wait_event_type` pre-classified as `IO Wait`, `Lock Wait`, `Network Wait`, or `Other Wait` in the log body, instead of the baseline `wait_event` entries. | `false` | no |
```
cristiangreco, line 39: We document `enable_pre_classified_wait_events` as a flag for the component in mysql (ie for `database_observability.mysql`), while we document it as a flag for the `query_samples` collector for postgres (ie for `database_observability.postgres.query_samples`).
However, in code we define it as a flag at the component level, for both mysql and postgres.

All in all, I think we should:
- make it a flag for `query_samples` collector, not for the component (both for mysql and postgres)
- fix this mysql doc

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -472,3 +491,20 @@ func (c *QuerySamples) determineTimerClauseAndLimit(uptime float64) (string, flo
...
+	category, _, _ := strings.Cut(rest, "/")
+	switch category {
+	case "io":
+		return "IO Wait"
+	case "lock":
+		return "Lock Wait"
+	case "synch":
+		return "Synch Wait"
+	}
+	return waitEventName
```
cristiangreco, line 509: This might be a bit misaligned with what we've used in other places, that is:

```
io_wait = 'wait/io/(file|table).+';
lock_wait = 'wait/((io/lock|synch)|lock).+';
network_wait = 'wait/io/socket.+';
```

internal/component/database_observability/mysql/component.go
```diff
@@ -274,6 +276,7 @@ type Component struct {
 	healthErr         *atomic.String
 	openSQL           func(driverName, dataSourceName string) (*sql.DB, error)
 	exporterCollector prometheus.Collector
+	waitEventCounter  *prometheus.CounterVec
```
cristiangreco, line 279: Going to ask if we can slim down this initial PR a bit more: can we defer this metric to a followup PR? It helps a bit to focus on the exact change being tested.

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -606,3 +618,42 @@ func isIdleState(state string) bool {
...
+		state.LastRow.DatabaseName.String,
+		state.LastRow.PID,
+		leaderPID,
+		state.LastRow.Username.String,
+		state.LastRow.BackendType.String,
+		we.LastState,
+		state.LastRow.BackendXID.Int64,
+		state.LastRow.BackendXmin.Int64,
+		we.LastWaitTime,
+		classifyPostgresWaitEventType(we.WaitEventType),
```
cristiangreco, line 639: I see that here we're reusing `wait_event_type` in v2, which already existed in v1 for postgres. Was the override done on purpose?

We're kinda losing such information by overriding it... although not a big deal maybe. An alternative would be to use an entirely new name, e.g. `wait_event_category` in place of `wait_event_type`, but want to hear your thoughts.

internal/component/database_observability/mysql/collector/query_samples.go
gaantunes: Yeah that is wrong. I wanted to propose we place Sync and IO Lock into a Others bucket though, so we do not inflate the Lock category which is normally associated with query locking contention.

internal/component/database_observability/mysql/component.go
gaantunes: We could... but the metric is an essential part of the performance improvement being developed and implementing the frontend counterpart without it would be a lot of rework.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Yeah that was on purpose. WaitEventType in v1 exists solely to make the regex a bit simpler, the information itself will still live in the `wait_event_name` field.

docs/sources/reference/components/database_observability/database_observability.mysql.md
gaantunes: Makes sense, I will address that.

internal/component/database_observability/mysql/collector/query_samples.go
gaantunes: For now, I will just fix it using the current categorization.

PR #5522, opened 2026-02-12
feat: Introduce SeriesRefMappingStore

### LINKED ISSUE #5062: Make labelstore.LabelStore in to a SeriesRefMappingStore

x1unix:

This PR implements SeriesRefMappingStore which was proposed in https://github.com/grafana/alloy/issues/5062.

The PR introduces `--feature.series-ref-mapping.enabled` experimental flag to disable label store in favor of the new series ref mapping store.

Fixes #5062

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/appenders/seriesrefmapping.go
+++ b/internal/component/prometheus/appenders/seriesrefmapping.go
@@ -188,7 +188,7 @@ func (s *seriesRefMapping) appendToChildren(ref storage.SeriesRef, lbls labels.L
 
 	// No existing mapping, proceed with normal append to all children.
 	var nonZeroCount int
-	var firstNonZeroRef storage.SeriesRef
+	var nonZeroRef storage.SeriesRef
 	for _, child := range s.children {
 		childRef, err := af(child, ref)
 		if err != nil {
@@ -198,7 +198,7 @@ func (s *seriesRefMapping) appendToChildren(ref storage.SeriesRef, lbls labels.L
 		s.childRefs = append(s.childRefs, childRef)
 		if childRef != 0 {
 			nonZeroCount++
-			firstNonZeroRef = childRef
+			nonZeroRef = childRef
 		}
 	}
 
@@ -213,7 +213,7 @@ func (s *seriesRefMapping) appendToChildren(ref storage.SeriesRef, lbls labels.L
 
 	if nonZeroCount == 1 {
 		// Only one child allocated a ref; return it directly — no mapping needed.
-		return firstNonZeroRef, nil
+		return nonZeroRef, nil
 	}
 
 	uniqueRef := s.store.CreateMapping(s.childRefs, lbls)
@@ -394,8 +394,6 @@ func (s *SeriesRefMappingStore) UpdateMapping(uniqueRef storage.SeriesRef, refRe
 	s.refMappingMu.Lock()
 	defer s.refMappingMu.Unlock()
 
-	// Ensure that label hash index is up to date to handle possible hash collisions.
-	// TODO: is this necessary?
 	newHash := lbls.Hash()
 	prev, ok := s.uniqueRefToChildRefs[uniqueRef]
 	if ok && prev.labelHash != newHash {


internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+
+	// We got different refs back and need to create a new mapping
+	uniqueRef := s.store.CreateMapping(s.childRefs, lbls)
+	s.uniqueRefCell.Refs = append(s.uniqueRefCell.Refs, uniqueRef)
+	return uniqueRef, nil
+}
+
+type uniqRefChildren struct {
+	childRefs *[]storage.SeriesRef
```
kgeckhart: I don't think we need this as a pointer anymore. IIRC we added it when we didn't have a struct due to weirdness with map growth on a slice.

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,508 @@
...
+		refUpdateRequired := false
+		for childIndex, childRef := range existingChildRefs {
+			newChildRef, err := af(s.children[childIndex], childRef)
+			if err != nil {
+				appendErr = multierror.Append(appendErr, err)
+			}
+
+			if newChildRef != childRef {
+				// Child ref changed, need to update mapping
+				existingChildRefs[childIndex] = newChildRef
```
kgeckhart: I think I did this originally, but pretty sure there's a race here because we are updating the slice returned from the mapping. I think we'll need to use s.childRefs to track diffs.

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,508 @@
...
+	//  are the same value. This isn't safe as we will mix downstream refs with unique refs which could
+	//  collide. We could start at max unit64 for unique refs and go backwards lessening the chance of
+	// 	collisions but it's rather dangerous for an unlikely edge case. If two components are returning
+	// 	the same ref it's two remote_write components which should probably be merged in to one.
+	for _, child := range s.children {
+		childRef, err := af(child, ref)
+		if err != nil {
+			appendErr = multierror.Append(appendErr, err)
+
+			// TODO should I make the childRef zero here?
```
kgeckhart:
```suggestion
			appendErr = multierror.Append(appendErr, err)
```

We don't care because any appender error will cause us to abort without doing anything with mappings.

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,508 @@
...
+
+	var multiErr error
+	for _, c := range s.children {
+		err := c.Rollback()
+		if err != nil {
+			multiErr = multierror.Append(multiErr, err)
+		}
+	}
+	return multiErr
+}
```
kgeckhart, line 89: I forgot to add latency tracking to this one like, https://github.com/grafana/alloy/pull/5522/changes#diff-013c9adca557619d7ae47d087315b534f31ba672237cd7612741df349b6e2f7aR29-R45

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,508 @@
...
+	// cleanupStaleRefs. We stopped and waited for the background worker that calls it
+	// to finish but some extra safety won't hurt.
+	s.timestampTrackingMu.Lock()
+	defer s.timestampTrackingMu.Unlock()
+
+	s.refMappingMu.Lock()
+	defer s.refMappingMu.Unlock()
+
+	clear(s.uniqueRefToChildRefs)
+	clear(s.uniqueRefTimestamps)
```
kgeckhart, line 502:
```suggestion
	clear(s.uniqueRefToChildRefs)
	clear(s.uniqueRefTimestamps)
	clear(s.labelHashToUniqueRef)
```

internal/component/prometheus/pipeline_test.go
```diff
@@ -33,166 +39,413 @@ import (
...
+	return noopAppender(n)
+}
+
+func (n noopStore) StartTime() (int64, error) {
+	return 0, nil
+}
+
+func (n noopStore) Close() error {
+	return nil
+}
```
kgeckhart: I think you can move this over to fanout_test as I believe it's the only thing using it ATM

internal/component/prometheus/appenders/seriesrefmapping_test.go
kgeckhart, line 1: I realized I added a bunch of tests for the store but not the appender 🙈

internal/component/prometheus/appenders/seriesrefmapping.go
x1unix: @kgeckhart done in https://github.com/grafana/alloy/pull/5522/changes/cffe650ddd990871fb9b33b74a1de53394c6cdd8

internal/component/prometheus/appenders/seriesrefmapping.go
x1unix: Done in https://github.com/grafana/alloy/pull/5522/changes/cffe650ddd990871fb9b33b74a1de53394c6cdd8

internal/component/prometheus/appenders/seriesrefmapping.go
x1unix: @kgeckhart done in https://github.com/grafana/alloy/pull/5522/changes/be4408bcde2309482cacd6f74799d0cc8f08c966

internal/component/prometheus/appenders/seriesrefmapping.go
x1unix: @kgeckhart addressed via https://github.com/grafana/alloy/pull/5522/changes/0fbc47874e0a6de8ba55752db599a883e4df71e9

internal/component/prometheus/appenders/seriesrefmapping_test.go
x1unix: @kgeckhart addressed via https://github.com/grafana/alloy/pull/5522/changes/5b527f336af82f9afba4dce0f2e3cd46737ad652

internal/alloycli/cmd_run.go
```diff
@@ -169,6 +170,7 @@ depending on the nature of the reload error.
 		cmd.Flags().StringVar(&r.windowsPriority, "windows.priority", r.windowsPriority, fmt.Sprintf("Process priority to use when running on windows. This flag is currently in public preview. Supported values: %s", strings.Join(slices.Collect(windowspriority.PriorityValues()), ", ")))
 	}
 	cmd.Flags().DurationVar(&r.taskShutdownDeadline, "feature.component-shutdown-deadline", r.taskShutdownDeadline, "Maximum duration to wait for a component to shut down before giving up and logging an error")
+	cmd.Flags().BoolVar(&r.enableSeriesRefMapping, "feature.series-ref-mapping.enabled", r.enableSeriesRefMapping, "Enable experimental SeriesRefMapping implementation instead of LabelStore")
```
dehaansa: Should this have a name that is clearer to a user who is not technically familiar with the implementation details? `replace-labelstore`, `simpler-labelstore` etc? More user friendly, and doesn't require renaming if we change the details of the replacement based on experiments.

internal/component/prometheus/appenders/passthrough.go
```diff
@@ -0,0 +1,98 @@
...
+	duration := time.Since(p.start)
+	p.writeLatency.Observe(duration.Seconds())
+}
+
+func (p *passthrough) SetOptions(opts *storage.AppendOptions) {
+	p.wrapping.SetOptions(opts)
+}
+
+func (p *passthrough) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	if p.start.IsZero() {
```
dehaansa, line 65: Why is start not initialized on initialization of the appender?

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,533 @@
...
+		s.start = time.Now()
+	}
+
+	// Check if the incoming ref has ref mappings
+	existingChildRefs := s.store.GetMapping(ref, lbls)
+
+	var appendErr error
+
+	// Sanity check: if we have existing child refs, they must match the number of children
+	if existingChildRefs != nil && len(existingChildRefs) == len(s.children) {
```
dehaansa, line 160: If this is a sanity check, should we log a warning if it fails?

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,533 @@
...
+
+		uniqueRef := s.store.CreateMapping(s.childRefs, lbls)
+		s.uniqueRefCell.Refs = append(s.uniqueRefCell.Refs, uniqueRef)
+		return uniqueRef, nil
+	}
+
+	// We got different refs back and need to create a new mapping
+	uniqueRef := s.store.CreateMapping(s.childRefs, lbls)
+	s.uniqueRefCell.Refs = append(s.uniqueRefCell.Refs, uniqueRef)
+	return uniqueRef, nil
```
dehaansa: This is the same logic, right? We can simplify the guard to return ref if `nonZeroCount == 1 && ref != 0`, otherwise do this logic.

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,533 @@
...
+		return
+	}
+
+	childRefSlice := make([]storage.SeriesRef, len(refResults))
+	copy(childRefSlice, refResults)
+
+	s.refMappingMu.Lock()
+	defer s.refMappingMu.Unlock()
+
+	// Ensure that label hash index is up to date to handle possible hash collisions.
```
dehaansa: How does this handle hash collisions? Seems more like it ignores/overrides hash collisions?

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+func (s *SeriesRefMappingStore) GetCellForAppendedSeries() *Cell {
+	return s.cellPool.Get().(*Cell)
+}
+
+func (s *SeriesRefMappingStore) cleanupStaleRefs() {
+	defer close(s.cleanupStopped)
+
+	ticker := time.NewTicker(15 * time.Minute)
```
dehaansa, line 431: Should this be configurable?

internal/component/prometheus/appenders/seriesrefmapping.go
```diff
@@ -0,0 +1,533 @@
...
+			s.timestampTrackingMu.Unlock()
+
+		case <-s.stopCleanup:
+			return
+		}
+	}
+}
+
+// Clear will clear all internal mappings and stop the cleaner goroutine if it is running.
+// It is safe to re-use the same instance after calling Clear.
```
dehaansa, line 477: How is it safe to reuse when you're closing and not replacing a chan?

internal/component/prometheus/remotewrite/interceptor.go
```diff
@@ -70,17 +78,25 @@ func NewInterceptor(componentID string, exited *atomic.Bool, debugDataPublisher
...
 
-			localRef := ls.GetLocalRefID(componentID, uint64(globalRef))
-			newLocalRef, nextErr := next.AppendHistogram(storage.SeriesRef(localRef), l, t, h, fh)
-			if nextErr == nil {
-				handleLocalLink(uint64(globalRef), l, localRef, uint64(newLocalRef))
+			var finalRef storage.SeriesRef
+			var err error
+			if ls.Enabled() {
+				localRef := ls.GetLocalRefID(componentID, uint64(ref))
+				newLocalRef, nextErr := next.AppendHistogram(storage.SeriesRef(localRef), l, t, h, fh)
```
dehaansa, line 92: Should newLocalRef be directly used somewhere, or is passing it to handleLocalLink on non-err enough?

internal/component/prometheus/remotewrite/interceptor.go
```diff
@@ -99,17 +115,25 @@ func NewInterceptor(componentID string, exited *atomic.Bool, debugDataPublisher
...
+			if ls.Enabled() {
+				localRef := ls.GetLocalRefID(componentID, uint64(ref))
+				newLocalRef, nextErr := next.UpdateMetadata(storage.SeriesRef(localRef), l, m)
+				if nextErr == nil {
+					handleLocalLink(uint64(ref), l, localRef, uint64(newLocalRef))
+				}
+				finalRef = ref
+				err = nextErr
+			} else {
+				finalRef, err = next.UpdateMetadata(ref, l, m)
```
dehaansa, line 136: It seems like this block could be refactored into a function that accepts a wrapper around the appender func that returns the ref & err for a lot less copy paste, but if it's not clean to do it's fine.

internal/service/labelstore/service.go
```diff
@@ -36,12 +36,35 @@ type staleMarker struct {
...
+	_ alloy_service.Service = (*disabledStore)(nil)
+)
+
+type LabelStoreService interface {
+	alloy_service.Service
+	LabelStore
+}
 
-func New(l log.Logger, r prometheus.Registerer) *Service {
+func New(l log.Logger, r prometheus.Registerer, enabled ...bool) LabelStoreService {
```
dehaansa, line 49: Why does this take a variable number of booleans?

internal/component/prometheus/appenders/passthrough.go
kgeckhart: This is mirrored functionality from the labelstore based appender https://github.com/grafana/alloy/blob/21416263bd3fe39b8a018f739782143b06a2fb7f/internal/component/prometheus/fanout.go#L165-L167 It's used for write latency and we don't start writing until the first append happens.

internal/component/prometheus/appenders/seriesrefmapping.go
kgeckhart: I'm not sure when we would want to change it? The same behavior for labelstore is also not configurable today, https://github.com/grafana/alloy/blob/93ad170a295d1064cf45cd0048e3bb6f7d96a270/internal/service/labelstore/service.go#L118

internal/component/prometheus/appenders/seriesrefmapping.go
kgeckhart: Which channel are we not replacing? AFAICT we replace all channels
https://github.com/grafana/alloy/blob/92cf809c46e989f6a51ae45be5060c4bd3fab748/internal/component/prometheus/appenders/seriesrefmapping.go#L528-L532

internal/service/labelstore/service.go
kgeckhart: It's a pattern to enable optional parameters, https://leapcell.io/blog/handling-optional-parameters-in-golang. Trying to prevent further superficial changes in the PR.

internal/alloycli/cmd_run.go
kgeckhart: This is one of those areas where I don't see anyone using it unless we explicitly tell to use it or they are familiar with the implementation details. Even with a friendlier name explaining what it does is incredibly implementation detail specific. The CLI flag is here to enable easy migration as ideally labelstore dies after enough testing and this flag never moves beyond being experimental so we can remove it.

internal/component/prometheus/appenders/seriesrefmapping.go
kgeckhart: This was changed very recently and I was talking with Denys about if it should be revered. If we had only 1 non-zero ref we were returning it vs creating a mapping which is a nice fast path. AI flagged it but I'm not convinced it's really a problem.

internal/component/prometheus/appenders/seriesrefmapping.go
kgeckhart: It doesn't, the comment is misleading 😅

internal/component/prometheus/remotewrite/interceptor.go
kgeckhart: Handling the local link is the only thing that is important. FTR this part didn't functionally change

internal/alloycli/cmd_run.go
kgeckhart: I changed it to `feature.prometheus.direct-fanout.enabled` to be more generic

internal/component/prometheus/appenders/seriesrefmapping.go
kgeckhart: I revered it and added safety around various state transitions from passthrough -> seriesrefmapping -> passthrough

dehaansa (approved):

PR #5528, opened 2026-02-13
fix(loki.source.file): Close file if we cannot find encoding

kalleep:
### Pull Request Details
Noticed that we do not close file if we cannot get encoding. Review all other paths and we close file properly.

I really wish go had something like [errdefer like zig does](https://ziglang.org/documentation/master/#errdefer).

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


blewis12 (approved):
oh that's interesting, wondering if this could be applicable to the escalation you're looking at currently?

**edit** ah nevermind, I see the code changed was introduced more recently

kalleep:
>edit ah nevermind, I see the code changed was introduced more recently

No but it got me to review the paths we are opening files

PR #5525, opened 2026-02-12
ci: Update build images to go 1.25.7

jharvey10:
## Post-merge TODO:

- [x] tag the image `build-image/v0.1.27`

PR #5529, opened 2026-02-13
ci: Update Go Version to v1.25.7

blewis12:
A follow up to https://github.com/grafana/alloy/pull/5525

Should also fix [CVE-2025-68121](https://ops.grafana-ops.net/a/grafana-vulnerabilityobs-app/cves/details/CVE-2025-68121)


kalleep:
You also need to update GHA to use new build image. And there is a bunch of other places we we can just update the version. See https://github.com/grafana/alloy/pull/5286/changes as a template

Issue #5533, opened 2026-02-13
Concurrent map read and write panic in getCachedTags


bennettatoms:
### What's wrong?

We have encountered this concurrent map read/write panic in /syntax/internal/value.getCachedTags a couple times in recent weeks.

The panic occurred in goroutine 226 with the error internal/runtime/maps.fatal at /usr/local/go/src/runtime/panic.go:1046, indicating a concurrent map access violation in Go's runtime, likely triggered by Alloy's internal state management during pod log collection.

### Steps to reproduce

1. Deploy Alloy v1.13.0 in Kubernetes with profiling enabled (alloy-profiles deployment)  
2. Configure remote configuration with self-monitoring logs pipeline  
3. Run under concurrent load with Kubernetes pod log collection  
4. Panic occurs during map access in runtime: "fatal error: concurrent map read and map write"  

### System information

Operating System: Linux (Kubernetes container) Architecture: linux/amd64 Platform: Kubernetes (prd cluster) Go Version: go1.23.7 (based on typical Alloy v1.13.0 builds) GOMAXPROCS: 8 (detected from logs)

### Software version

Alloy v1.13.0

### Configuration

```text
loki.write "grafana_cloud_loki" {  
	endpoint {  
		url = "https://logs-prod-021.grafana.net/loki/api/v1/push"  
		basic_auth {  
			username = "REDACTED"  
			password = sys.env("GCLOUD_RW_API_KEY")  
		}  
	}  
}  

discovery.kubernetes "pods" {  
	role = "pod"  
	selectors {  
		role  = "pod"  
		field = "spec.nodeName=" + sys.env("HOSTNAME")  
	}  
}  

discovery.relabel "filtered_pods" {  
	targets = discovery.kubernetes.pods.targets  
	rule {  
		action        = "keep"  
		source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_name", "__meta_kubernetes_pod_container_name"]  
		separator     = "/"  
		regex         = sys.env("NAMESPACE") + "/" + sys.env("POD_NAME") + "/alloy"  
	}  
	# ... additional relabel rules for namespace, pod, container, job labels  
}  

loki.source.kubernetes "pod_logs" {  
	targets    = discovery.relabel.filtered_pods.output  
	forward_to = [loki.process.pod_logs.receiver]  
}  

loki.process "pod_logs" {  
	stage.match {  
		selector = "{tmp_container_runtime=~\"containerd|cri-o\"}"  
[...]

PR #5534, opened 2026-02-13
fix: Use read-write mutex locks to prevent concurrent tagsCache map reads and writes

bennettatoms:
Fixes cause of panic described in https://github.com/grafana/alloy/issues/5533

### DIFF
--- a/syntax/internal/value/tag_cache.go
+++ b/syntax/internal/value/tag_cache.go
@@ -70,10 +70,6 @@ func getCachedTags(t reflect.Type) *objectFields {
 	}
 
 	tagsCacheMutex.Lock()
-	if entry, ok := tagsCache[t]; ok {
-		tagsCacheMutex.Unlock()
-		return entry
-	}
 	tagsCache[t] = tree
 	tagsCacheMutex.Unlock()
 


syntax/internal/value/tag_cache.go
```diff
@@ -16,9 +17,14 @@ func getCachedTags(t reflect.Type) *objectFields {
 		panic("getCachedTags called with non-struct type")
 	}
 
+	// Mutex to protect concurrent reads and writes to tagsCache
+	var tagCacheMutex sync.RWMutex
```
kalleep: This is wrong. This will create a new lock every time `getCachedTags` is called so it still won't protect the map.

The lock need's to be a package level variable.

bennettatoms: PR has been updated. Thank you for the feedback.

syntax/internal/value/tag_cache.go
```diff
@@ -62,7 +69,14 @@ func getCachedTags(t reflect.Type) *objectFields {
 		}
 	}
 
+	tagsCacheMutex.Lock()
+	if entry, ok := tagsCache[t]; ok {
+		tagsCacheMutex.Unlock()
+		return entry
+	}
```
kalleep: This feel a bit unnecessary, we have already taken a write lock and already built the tree so I don't see the need for this extra lookup

bennettatoms: no worries, removed it

kalleep (approved):

PR #5535, opened 2026-02-13
chore: Add tool and make targets to update go version

kalleep:
I built a small tool with two different commands `pr-1` and `pr-2`, these should be triggered from our make targets `make update-go-version-pr-1 VERSION=1.25.7` and `make update-go-version-pr-2 VERSION=1.25.7`.

There is also a diff for some modules that was missed in previous pr:s to update version

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/go-version/main.go
+++ b/tools/go-version/main.go
@@ -113,7 +113,7 @@ func updateGoModFiles(root, version string) error {
 }
 
 func updateDockerFiles(root, version string) error {
-	paths, err := getPaths(root, "Dockerfile", "Dockerfile", "Dockerfile.windows", "tools/build-image")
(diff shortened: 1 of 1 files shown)


tools/go-version/main.go
```diff
@@ -0,0 +1,280 @@
...
+		if err := os.WriteFile(path, newContent, 0644); err != nil {
+			return fmt.Errorf("write %s: %w", path, err)
+		}
+	}
+
+	return nil
+}
+
+func updateDockerFiles(root, version string) error {
+	paths, err := getPaths(root, "Dockerfile", "Dockerfile", "Dockerfile.windows", "tools/build-image")
```
jharvey10: Should this just have `Dockerfile` once instead of twice?

PR #5539, opened 2026-02-13
ci: Generate config argument tables in docs from schema

ptodev:
### Pull Request Details

This could be useful for a few reasons:
* One day the config struct code can also be generated from schema files.
* It makes it easier to keep the docs in sync. Currently the docs sometimes drift from the implementation. Default values are not up to date, and sometimes arguments or blocks are missing completely.
* [Upstream OTel](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.145.0/processor/tailsamplingprocessor/config.schema.yaml) also use schema files. We could use them to make OTel upgrades easier.

### Notes to the Reviewer

Before I open the PR for review I'll need to add a GitHub Action to make sure the generated docs are up to date.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -50,37 +50,26 @@ The component starts an HTTP server on the configured port and address with the
...
-| `forward_to`                | `list(LogsReceiver)` | List of receivers to send log entries to.                                          |            | yes      |
-| `labels`                    | `map(string)`        | The labels to associate with each received logs record.                            | `{}`       | no       |
-| `relabel_rules`             | `RelabelRules`       | Relabeling rules to apply on log entries.                                          | `{}`       | no       |
-| `use_incoming_timestamp`    | `bool`               | Whether to use the timestamp received from request.                                | `false`    | no       |
-| `max_send_message_size`     | `size`               | Maximum size of a request to the push API.                                         | `"100MiB"` | no       |
-| `graceful_shutdown_timeout` | `duration`           | Timeout for server's graceful shutdown. If configured, should be greater than zero. | `"30s"`    | no       |
+{{< docs/shared lookup="generated/components/loki/source/api/__arguments.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
thampiotr, line 53: It seems that you are not checking in the generated code? 
Now if someone wants to quickly take a look at markdown documentation, it won't be possible. 
I like the pattern where we do the following:
- check in the generated code
- have a PR validation that makes sure that code has been re-generated (there is no diff after running generation)

what do you think?

docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -50,37 +50,26 @@ The component starts an HTTP server on the configured port and address with the
...
 ## Blocks
 
+<!-- TODO: Add this note to the docgen -->
 You can use the following blocks with `loki.source.api`:
 
-| Name                  | Description                                        | Required |
-| --------------------- | -------------------------------------------------- | -------- |
-| [`http`][http]        | Configures the HTTP server that receives requests. | no       |
-| `http` > [`tls`][tls] | Configures TLS for the HTTP server.                | no       |
+{{< docs/shared lookup="generated/components/loki/source/api/__blocks.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
thampiotr, line 66: We already have some generation in this code that uses comments to identify the sections which need updating. See below for the `<!-- START GENERATED COMPATIBLE COMPONENTS -->`. 

Have you considered this approach to make it easier to read the text markdown file? That way we could avoid having multiple files to look up to create the full page to read. 

These references will IMO make it harder to just open a full file and read it, which will be a bit of a pain for us, but also will be a struggle for AI which will likely perform worse when told to read the docs.

thampiotr: Also, if we decide to change the approach, maybe we should also migrate the 'compatible components' section to the new method for consistency.

internal/component/common/net/schema.yml
thampiotr, line 1: Is this file written entirely by hand? It does look a bit intimidating.

docs/sources/reference/components/loki/loki.source.api.md
ptodev: I agree that it makes it a little harder to read, but most of the time we don't look at raw markdown files. It's not hard to generate a preview locally and open it in a browser. I have to open browser previews a lot of the time anyway, because there's lots of stuff that can only render properly on a browser:
* Admonition blocks
* Markdown files that are shared between components. E.g. HTTP server/client arguments are often in such shared files
* Links with variables in them, e.g. a link that depends on the OTel version that Alloy uses

The main reason I don't like using shared files in this PR are:
* It makes it harder to review since the tables are moved elsewhere.
* A user might forget to add a newly added block to the component's markdown. There's nothing to check whether all markdowns that should be referenced are indeed referenced.

We plan to pilot this on a limited number of components, and there are much bigger problems right now (e.g. code validation). I think it would be ok to proceed like this, and then fix this later if necessary? I could open an issue to track this suggestion?

docs/sources/reference/components/loki/loki.source.api.md
ptodev: Apologies, I forgot to push it! I pushed the generated code now. If the PR looks ok, I will also add a GitHub action to check that the `docs/sources/shared/generated/` folder is up to date.

thampiotr: Yes, thanks! we definitely do want the GH action to verify it's up-to-date 👍

docs/sources/reference/components/loki/loki.source.api.md
thampiotr: sure, we can trial this 👍 
I do reserve a right to complain about this in the future 😂

internal/component/common/net/schema.yml
ptodev: Yes, for now. Eventually there will be automation for this too.

.github/workflows/check-generate-docs-args.yml
ptodev, line 1: I tested the GitHub action by modifying the schema file and it correctly identified that the generated docs are out of date:
Screenshot 2026-02-25 at 10 12 25
Screenshot 2026-02-25 at 10 12 32

docs/sources/shared/generated/components/loki/source/api/__blocks.md
```diff
@@ -0,0 +1,13 @@
...
+| [`http`][http] | Configures the HTTP server that receives requests. | no |
+| [`tls`][tls] | Configures TLS for the server. | no |
+| grpc > [`tls`][tls] | Configures TLS for the gRPC server. | no |
+| http > [`tls`][tls] | Configures TLS for the HTTP server. | no |
+
+[grpc]: #grpc
+[http]: #http
+[tls]: #tls
+[tls]: #tls
+[tls]: #tls
```
ptodev: I'll need to make sure there is only one such link

docs/sources/shared/generated/components/loki/source/api/__blocks.md
```diff
@@ -0,0 +1,13 @@
+| Block | Description | Required |
+| ----- | ----------- | -------- |
+| [`grpc`][grpc] | Configures the gRPC server that receives requests. | no |
+| [`http`][http] | Configures the HTTP server that receives requests. | no |
+| [`tls`][tls] | Configures TLS for the server. | no |
+| grpc > [`tls`][tls] | Configures TLS for the gRPC server. | no |
```
ptodev: The parent (`grpc` in this case) should be surrounded by backticks.

docs/sources/shared/generated/components/loki/source/api/__arguments.md
ptodev, line 1: It'd be nice if the rows in the tables are sorted in the same way as before.

docs/sources/shared/generated/components/loki/source/api/__blocks.md
```diff
@@ -0,0 +1,13 @@
+| Block | Description | Required |
+| ----- | ----------- | -------- |
+| [`grpc`][grpc] | Configures the gRPC server that receives requests. | no |
```
ptodev: gRPC was not documented before. I wonder if it actually works? I need to test it and add a unit test for it.

docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -50,37 +50,26 @@ The component starts an HTTP server on the configured port and address with the
...
 The > symbol indicates deeper levels of nesting.
 For example, `http` > `tls` refers to a `tls` block defined inside an `http` block.
 
-[http]: #http
-[tls]: #tls
-
 ### `http`
 
-{{< docs/shared lookup="reference/components/server-http.md" source="alloy" version="<ALLOY_VERSION>" >}}
+{{< docs/shared lookup="generated/components/loki/source/api/http.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
ptodev: This config is shared between components and so ideally it shouldn't be in the `loki/source/api` directory. It should be in a location where other component docs can also import it.

clayton-cornell:
Do we want to hold on this for a few days while we wait to see what changes may be in the pipeline for the block hierarchy  documentation syntax?

docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -50,37 +50,26 @@ The component starts an HTTP server on the configured port and address with the
...
 
-| Name                  | Description                                        | Required |
-| --------------------- | -------------------------------------------------- | -------- |
-| [`http`][http]        | Configures the HTTP server that receives requests. | no       |
-| `http` > [`tls`][tls] | Configures TLS for the HTTP server.                | no       |
+{{< docs/shared lookup="generated/components/loki/source/api/__blocks.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
+<!-- TODO: Add this note to the docgen -->
 The > symbol indicates deeper levels of nesting.
 For example, `http` > `tls` refers to a `tls` block defined inside an `http` block.
```
clayton-cornell: We will be able to drop this part with the new Block element shortcode.

docs/sources/shared/generated/components/loki/source/api/__blocks.md
ptodev: Done

docs/sources/shared/generated/components/loki/source/api/__arguments.md
ptodev: I made them so the required attributes are shown first. Then attributes are sorted by name.

They won't always appear as before, since before they weren't always sorted correctly.

docs/sources/reference/components/loki/loki.source.api.md
ptodev: Good point! I'll update the PR to use the `{{< docs/alloy-config >}}` shortcode.

docs/sources/reference/components/loki/loki.source.api.md
ptodev: Done - it's now in `docs/sources/shared/generated/common/net/http.md`

docs/sources/shared/generated/components/loki/source/api/__blocks.md
ptodev: Done

docs/sources/shared/generated/components/loki/source/api/__blocks.md
ptodev: I decided to add schema for loki.write and to delete the one for loki.source.api.

We can re-add the one for loki.source.api when we have an answer to whether grpc actually works.

docs/sources/reference/components/loki/loki.source.api.md
ptodev: I agree that we will drop this part, but it's not necessary for this PR. This PR is just to import the table from a folder. In a subsequent PR this paragraph can be removed, `{{< docs/alloy-config >}}` can be added, and everything will continue to work normally.

internal/component/common/net/schema.yml
ptodev: I added a Cursor skill to write these files and to double check them.

ptodev:
@thampiotr @clayton-cornell I updated the PR to polish a few rough edges, and added a Cursor skill to help with the writing of the schema files. Please feel free to review again!

If the PR looks good to you I will double check the arguments in the tables one last time before merging. I haven't double checked them very thoroughly yet because the arguments on the `main` branch may change by the time we need to merge it.

internal/component/common/net/schema.yml
ptodev: It also checks for discrepancies between code and docs before writing the schemas.

Screenshot 2026-03-27 at 17 35 10

Makefile
```diff
@@ -325,6 +325,23 @@ else
...
+else
+	@echo "Generating documentation config arguments for all components..."
+	@go list -f '{{.Dir}}' $$(go list -f '{{join .Imports " "}}' ./internal/component/all) | while read dir; do \
+		if [ -f "$$dir/metadata.yml" ]; then \
+			component_name=$$(echo $$dir | sed 's|.*/internal/component/||'); \
+			output_dir="docs/sources/shared/generated/components/$$component_name"; \
+			shared_base="docs/sources/shared/generated"; \
+			echo "Generating docs for component: $$component_name"; \
+			mkdir -p "$$output_dir"; \
+			go run ./internal/tools/docs_args_generator/ "$$dir/metadata.yml" "$$output_dir" "$$shared_base"; \
```
thampiotr, line 340: Since we already have a Go code, wouldn't it make sense to replace this makefile scripting with Go code inside that tool? To keep makefile simpler - like a single Go tool call?

internal/tools/docs_args_generator/main.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func main() {
+	cmd := NewCommand()
+	cobra.CheckErr(cmd.Execute())
+}
+
+func NewCommand() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:          "docs_args_generator",
```
thampiotr, line 14: I'd like to see some documentation / help message about what the args mean.

internal/tools/docs_args_generator/blocks.go
```diff
@@ -0,0 +1,167 @@
...
+type BlocksTableRow struct {
+	Name           string
+	ParentPath     []string
+	Description    string
+	Required       string
+	SourceSchemaID string
+	// DefName is set when the block was declared via an internal $ref (e.g.
+	// "$ref: \"#/$defs/tls\""). It overrides the parent-path component of the
+	// link ID so that every occurrence of the same $defs entry shares one link.
+	DefName string
```
thampiotr, line 20: I don't quite understand that comment...

docs/sources/reference/components/loki/loki.source.api.md
```diff
@@ -91,7 +79,7 @@ The `tls` block configures TLS for the HTTP server.
 
 ## Exported fields
 
-`loki.source.api` doesn't export any fields.
+{{< docs/shared lookup="generated/components/loki/source/api/__exports.md" source="alloy" version="<ALLOY_VERSION>" >}}
```
thampiotr, line 82: I don't see this file right now?
Also the paths are confusing. Looking at this magic tag, I would expect the file to be at $repo/docs/shared/generated/components/loki/source/api/__exports.md

But I think it would magically be looked up at docs/sources/shared/generated/...


Since we are now heavily using imports, could we clean up the paths to be easier to follow?

Makefile
```diff
@@ -384,6 +401,9 @@ else
 	bash ./.github/ISSUE_TEMPLATE/scripts/update-gh-issue-templates.sh
 endif
 
+validate-metadata:
+	bash ./tools/validate-metadata
```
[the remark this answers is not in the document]
thampiotr: Yeah, I don't see this one

internal/component/loki/write/metadata.yml
```diff
@@ -0,0 +1,248 @@
...
+      type: object
+      alloy:
+        type: attribute
+        type_override: "map(string)"
+        default_override: "{}"
+
+    max_streams:
+      description: >-
+        Maximum number of active streams.
+      type: integer
```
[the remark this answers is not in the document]
thampiotr: looks legit IMO

internal/tools/docs_args_generator/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,215 @@
...
+		if prop.Ref != "" {
+			referencePath := filepath.Join(ymlPathDir, prop.Ref)
+			log.Printf("Processing YAML subschema: %s", referencePath)
+
+			parsedProp, err := unmarshalFromPath[Schema](referencePath)
+			if err != nil {
+				return fmt.Errorf("failed to parse schema file: %w", err)
+			}
+
+			if err := mergeSubschemas(ymlPathDir, parsedProp); err != nil {
```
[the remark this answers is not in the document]
thampiotr: It does look like we keep the same `ymlPathDir` is that correct?

internal/tools/docs_args_generator/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,215 @@
...
+				return err
+			}
+
+			resolveInternalRefs(parsedProp, parsedProp.Definitions)
+
+			for name, def := range parsedProp.Definitions {
+				def.SourceID = parsedProp.ID
+				schema.Properties[name] = def
```
thampiotr, line 114: Isn't it possible that schema.Properties is nil and this would panic?

ptodev:
@thampiotr thank you so much for reviewing this PR earlier this year! Unfortunately, I lack the time to polish it enough to make it good enough for merging. It's been stale for months and I'll close it for now. We could reopen it later if we want to have something like this in the future. 

I still think there's value in this kind of project. I see that #6477 also uses a schema, and also @kalleep had mentioned it may be useful for something he has in mind.

PR #5540, opened 2026-02-13
fix(pyroscope.ebpf): Move Pyroscope ebpf metrics registration after component error handling

### LINKED ISSUE #5156: pyrocope.ebpf confusing panic
### Component(s)

pyroscope.ebpf

### What's wrong?

`panic: duplicate metrics collector registration attempted`

The ebpf component seem to be created second time with the same prom registry.
```
Error: data-alloy/ebpf.txt:12:1: Failed to build component: building component: open /tmp/symb-cache/lidia1: permission denied

11 |
12 |   pyroscope.ebpf "default" {
   |  _^^^^^^^^^^^^^^^^^^^^^^^^^^
13 | |     targets_only = false
14 | |     forward_to   = [pyroscope.write.staging.receiver]
15 | | }
   | |_^
16 |

interrupt received
panic: duplicate metrics collector registration attempted

goroutine 253 [running]:
github.com/prometheus/client_golang/prometheus.(*wrappingRegisterer).MustRegister(0xc001a81aa0, {0xc0019dd5c0?, 0x0?, 0x0?})
        /home/korniltsev/go/pkg/mod/github.com/prometheus/client_golang@v1.23.2/prometheus/wrap.go:138 +0x126
github.com/grafana/pyroscope/ebpf/metrics.NewSymtabMetrics({0xeba6ff0, 0xc001a81aa0})
        /home/korniltsev/go/pkg/mod/github.com/grafana/pyroscope/ebpf@v0.4.11/metrics/symtab.go:43 +0x88d
github.com/grafana/pyroscope/ebpf/metrics.New({0xeba6ff0, 0xc001a81aa0})
        /home/korniltsev/go/pkg/mod/github.com/grafana/pyroscope/ebpf@v0.4.11/metrics/metrics.go:12 +0x25
github.com/grafana/alloy/internal/component/pyroscope/ebpf.newMetrics({0xeba6ff0, 0xc001a81aa0})
        /home/korniltsev/alloy/internal/component/pyroscope/ebpf/metrics.go:54 +0x49c
github.com/grafana/alloy/internal/component/pyroscope/ebpf.New({_, _}, {_, _}, {_, _}, {{0xc0011360b0, 0x1, 0x1}, {0x0, ...}, ...})
        /home/korniltsev/alloy/internal/component/pyroscope/ebpf/ebpf_linux.go:56 +0x11d
github.com/grafana/alloy/internal/component/pyroscope/ebpf.init.0.func1({{0x7f9bad2431f8, 0xc001a81980}, {0xc001596cf0, 0x16}, {0xeaef520, 0xc00159b770}, {0xc000dbb7a0, 0x21}, 0xc00117c650, {0xeba6ff0, ...}, ...}, ...)
        /home/korniltsev/alloy/internal/component/pyroscope/ebpf/ebpf_linux.go:41 +0xfe
github.com/grafana/alloy/internal/runtime/internal/controller.(*BuiltinComponentNode).evaluate(0xc000b53b88, 0xc001971260)
        /home/korniltsev/alloy/internal/runtime/internal/controller/node_builtin_component.go:290 +0x496
github.com/grafana/alloy/internal/runtime/internal/controller.(*BuiltinComponentNode).Evaluate(0xc000b53b88, 0xd2f1ea0?)
        /home/korniltsev/alloy/internal/runtime/internal/controller/node_builtin_component.go:263 +0x1c
github.com/grafana/alloy/internal/runtime/internal/controller.(*Loader).concurrentEvalFn(0xc001570b60, {0x7f9bad243218, 0xc000b53b88}, {0xebddac0, 0xc0019e22a0}, {0xeb28e98, 0xc001b0c030}, 0xc001a9f380)
        /home/korniltsev/alloy/internal/runtime/internal/controller/loader.go:854 +0x553
github.com/grafana/alloy/internal/runtime/internal/controller.(*Loader).EvaluateDependants.func2()
        /home/korniltsev/alloy/internal/runtime/internal/controller/loader.go:791 +0x33
github.com/grafana/alloy/internal/runtime/internal/worker.(*workQueue).emitNextTask.func1()
        /home/korniltsev/alloy/internal/runtime/internal/worker/worker_pool.go:195 +0x5c
github.com/grafana/alloy/internal/runtime/internal/worker.(*fixedWorkerPool).start.func1()
        /home/korniltsev/alloy/internal/runtime/internal/worker/worker_pool.go:101 +0x67
created by github.com/grafana/alloy/internal/runtime/internal/worker.(*fixedWorkerPool).start in goroutine 1
        /home/korniltsev/alloy/internal/runtime/internal/worker/worker_pool.go:94 +0x25
```

### Steps to reproduce

run ebpf component with insufficient permissions - no root

### System information

Linux fedora 6.17.9-300.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Nov 24 23:31:27 UTC 2025 x86_64 GNU/Linux

### Software version

e2e96e95ff0dab600befbe63165e10eea096b968

### Configuration

```text

```

### Logs

```text

```

### Tip

crbednarz:

This change moves the metrics registration to the end of the pyroscope.ebpf component construction. This allows for errors to occur during construction without leaving unused metrics registered.

### Pull Request Details

This is a fix for #5156. I've written up the specifics of the issue here: https://github.com/grafana/alloy/issues/5156#issuecomment-3814096205

In short, during the construction of the pyroscope.ebpf component, metrics get registered for Prometheus. If the construction process fails, those metrics remain registered. Since components may be re-evaluated asynchronously, this would lead to a secondary "duplicate registration" error occasionally printed, distracting from the actual error.

Fixes #5156 

### Notes to the Reviewer

I went with the lowest impact change I could think of, as the issue itself is fairly minor. Since `ms` wasn't referenced before the final return, moving it (and `newMetrics(reg)`) lower meant no failures could occur after metrics were registered.

I've also verified the new test fails without the change.

### PR Checklist

- [x] Tests updated


crbednarz:
I realize the commit text is missing the word "metrics" in "ebpf _**metrics**_ registration". I can change it, but it seems like the PR title is what gets used anyway?

korniltsev-grafanista:
I dont think this should be fixed in the pyroscope components. I believe the alloy itself either should not create a component second time, or should not reuse registrer if they for some strange reason have to

crbednarz:
> I dont think this should be fixed in the pyroscope components. I believe the alloy itself either should not create a component second time, or should not reuse registrer if they for some strange reason have to

That was my thought too initially- as it seemed strange to me that a component build error that would end execution would also be retried. Although I'm still learning my way around the code base, so perhaps there is some underlying reason.

That said, I would argue that there are two separate issues at play here:

1. The component is rebuilt when it (probably) shouldn't be.
2. `ebpf.New(..., reg, ...)` registers entries on `reg` even when it fails.

I definitely want to get a better understanding of the rationale behind the rebuild behavior, but ignoring that for a moment, (2) still feels like a valid issue to address.

Consider this call:

```go
ebpf, err = ebpf.New(logger, reg, "test-ebpf", args)
```

On success, `ebpf, nil` is returned and the expected registration of our metrics into `reg` occurs.
However, in the event of an error, the call returns `nil, err`. At this point we have no way of knowing that a side effect has occurred, nor do we have a mechanism to undo it. Our only option would be to completely reconstruct `reg`, which feels less than ideal. Of course the caller wouldn't intuitively know to do that either, as I think there's an implicit expectation for failed constructions to clean up after themselves.

I figure fixing either issue would resolve #5156. The rebuild issue is definitely a curious one, but I suspect it'll involve a more complex solution as the calls are spread across multiple threads and have a much wider impact.

I'm curious what your thoughts are here?

korniltsev-grafanista:
Appologise for late response.


> The component is rebuilt when it (probably) shouldn't be.

I am still thinking the component either should not be recreated, or the registry should be a pristine one.

> ebpf.New(..., reg, ...) registers entries on reg even when it fails.

Again, I beleive this is a problem with alloy. If the contract is that we do not or register any metrics unless the component is created with no error - it should be *enforced* through code or lint *and documented*

I will close this for now as I believe this should be fixed in a different way. BTW, it's not clear why you want this to be fixed. If you still believe you want this to land - I suggest you to work with @grafana/grafana-alloy-maintainers  and the rest of @grafana/grafana-alloy-profiling-maintainers

crbednarz:
> I will close this for now as I believe this should be fixed in a different way. BTW, it's not clear why you want this to be fixed. If you still believe you want this to land - I suggest you to work with @grafana/grafana-alloy-maintainers and the rest of @grafana/grafana-alloy-profiling-maintainers

That's honestly a bit disappointing to hear.

For context, I don't have a specific reason to fix this particular issue. I was actually just looking to contribute periodically, and this seemed like a good jumping off point to better understand the code base and surrounding processes. (Which is why the write-ups for this probably exceeds what the issue merits)

Though at this point I'm more confused now than before. This might not resolve the linked issue, but it still felt worthwhile and a net positive.

I'd really like to continue making an effort here, but I'm not sure how to even loop in the groups you've mentioned, nor is it clear to me where the appropriate place for this discussion should have been, given the small size of the change.

Would you be able to help loop in the appropriate party(s) here? Or clarify where I should be discussing this? Any help is appreciated.

korniltsev-grafanista:
> That's honestly a bit disappointing to hear.

I understand and appologise.

> and this seemed like a good jumping off point

Unfortunately this issue seem to be a controversial one. I believe this should be fixed in alloy, alloy seem to believe components have to "fix" and employ these ugly easy to forget and hard to deploy for 3party code `util.MustRegisterOrGet` or some thing else , for example your solution with reordering of metric.


> better understand the code base and surrounding processes

I would start with understanding why the component is created second time if it failed.

> but it still felt worthwhile and a net positive.

This looks like hiding a problem to me rather than fixing.

> but I'm not sure how to even loop in the groups you've mentioned

I would expect tagging them here should be enough. You can also try your luck in grafana community slack in pyroscope and/or alloy channels.

> Would you be able to help loop in the appropriate party(s) here?

I've requested review from @grafana/grafana-alloy-maintainers , other than that I recommend you  to try the slack way.


OK, maybe closing the PR was too much, I just wanted to give at least some feedback instead of permanent silence. lets reopen and you seek for other approver. I will not block or rever this "fix", just do not expect an approve from me.

marcsanmi:
Hey @crbednarz, thanks for the thorough write-up and patience here.

I agree with @korniltsev-grafanista that the root cause lives in the alloy framework — ideally a failed `Build()` should not leave the registerer in a state that breaks subsequent retries.

That said, I do think your reordering is a reasonable defensive fix

I'm on the pyroscope/profiling team as well, and I'd like to wait for the @grafana/grafana-alloy-maintainers to weigh in before we move forward. Specifically, it would be good to hear whether:

- The framework intends to guarantee a fresh registerer on component rebuild, or
- Components are expected to handle this themselves

kalleep:
AFAIK there is nothing in place today to protect against it in alloy. But I agree that we should handle it inside the alloy runtime instead of having to handle this at component level.

Will create a issue for it, in the meantime I don't see the harm of merging this one

marcsanmi (approved):
LGTM, thanks for the contribution!

crbednarz:
Thanks everyone. I really appreciate you all taking the time to look over the change.

Hopefully I'll have some time soon to look into the component rebuild issue as well, as it's definitely interesting.

PR #5547, opened 2026-02-17
build: Run Collector/Dependency Generation on Build Locally

blewis12:
After the introduction of the replacement generation tooling and the otel collector generation, there are now multiple targets in the Alloy makefile that might need to be run to build the project if dependencies are changed. For local development there are a couple different scenario's you might find yourself in

1. You've updated the source code but not any dependencies - in which case you just want to build the code as is (ie the `alloy` make target)
2. You've updated a dependency, through a `replace` directive or otherwise - in which case you'd need to run `sync-module-dependencies` to ensure that the `alloy` and `alloy_engine` go.mod files are synced up and tidied , and then the `generate-otel-collector-distro` which would re-generate and tidy `go.mod` file for the otel engine (that needs to be updated because the engine imports both `alloy` and `alloy_engine` and so would have updated transitive dependencies)
3. You've updated the builder manifest file, in which case you'd need to run `generate-otel-collector-distro`

For CI - we'd never need to do any code generation, we assume that all generated code has been checked in. We ensure this by having the `check generate-module-dependencies` and `check generate-otel-collector-distro` workflows that run in every PR.

An attempt to summarize this with a diagram:
Screenshot 2026-02-17 at 11 17 52

### Pull Request Details
This PR introduces a small changes: the `generate-source-code` make target that will run as a pre-requisite for the `alloy` target. It will not run if `CI=true`,  a pattern also used [elsewhere](https://github.com/search?q=org%3Agrafana%20%22ifeq%20(%24(CI)%2Ctrue)%22&type=code). We can see this in action in the CI for this PR:

Screenshot 2026-02-17 at 11 31 52

 It will also not run if `SKIP_CODE_GENERATION` is set to `1` - so developers can also opt-out if they don't need it, much like what we do for `SKIP_UI_BUILD`. The individual targets remain individual, so they can be run separate from one another if necessary.

My hope is that this change should reduce some confusion for contributors, who ideally shouldn't have to think too hard about code generation

It's true that we could also modify our pipelines to just run the code generation and avoid the conditional, but it would both add unnecessary overhead and mean we'd need to make some changes to have that run smoothly.


Makefile
```diff
@@ -274,6 +275,18 @@ else
 	cd ./tools/generate-module-dependencies && $(GO_ENV) go generate
 endif
 
+generate-source-code:
+ifeq ($(USE_CONTAINER),1)
+	$(RERUN_IN_CONTAINER)
+else ifeq ($(CI),true)
```
thampiotr, line 295: Is `CI` documented to be set in GH workflows?

jharvey10:
> For CI - we'd never need to do any code generation, we assume that all generated code has been checked in. We ensure this by having the `check generate-module-dependencies` and `check generate-otel-collector-distro` workflows that run in every PR.

I'm wondering if it's worth running these somehow in CI mode. That way we have one flow where it's not CI mode where they are expected to run, and then another flow (CI enabled) where we don't generate anything new, but we still validate the current state of things. That way CI mode isn't strictly a bypass of some checks/builds we expect to do, but rather puts it into read-only validation mode.

blewis12:
@jharvey10 I was trying to workshop a way to do this actually - but from what I can see the only way to really validate things _is_ to generate and check the state 🤔 and if we want to regenerate when we build in CI then we'd need to make some changes specifically to our publishing action where it's running our build step for each architecture in parallel, last time we got issues doing that with the generation script stepping on its own toes when reading/writing files in parallel.

What we could do perhaps is have our `check` github actions run also on PR merge, that way we can ensure that the main branch (or release branch, what have you) is always validated to be in a correct state. Wdyt?

Makefile
blewis12: yup! it's in the docs [here](https://docs.github.com/en/actions/reference/workflows-and-actions/variables)

jharvey10:
@blewis12 I feel like what you suggested makes sense. I forgot about the parallel thing and it messing up its own files. I suppose at a minimum where this catches up to us is on the release-please PRs. Since I believe those run a full CI suite and act as a release gate anyway. I could be persuaded to be okay with how it is 😆  as long as it's a gate someplace we actively see. I know the main ci often fails (for a few reasons), so I doubt the team is monitoring for things like these. It might just end up as more noise until we get main consistently green.

blewis12:
@jharvey10 I just pushed a commit that will ensure we run gen-checks on main, I know the failing pipelines are overloaded but if something _does_ go wonky this might at least help us retrace back to understand what merge caused the issue. Could I get a ✅ from you if you feel this is good at least for now?

Makefile
```diff
@@ -274,6 +275,18 @@ else
...
+	@echo "Skipping code generation (CI=1)"
+else ifeq ($(SKIP_CODE_GENERATION),1)
+	@echo "Skipping code generation (SKIP_CODE_GENERATION=1)"
+else
+	@$(MAKE) generate-module-dependencies generate-otel-collector-distro
+endif
+
+
 generate-otel-collector-distro:
```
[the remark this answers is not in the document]
blewis12: we should never be generating things in parallel, this would fail anyway - I prefer to have a separate make target specifically addressing the building of source code and have the targets runnable by themselves so I will keep this here

Issue #1310, opened 2024-07-17
Docs feedback: Review guidance for sending OTel to Grafana


ptodev:
### URL

https://grafana.com/docs/alloy/latest/collect/opentelemetry-to-lgtm-stack/

### Feedback

We currently have a [Collect OpenTelemetry data and forward to Grafana](https://grafana.com/docs/alloy/latest/collect/opentelemetry-to-lgtm-stack/) doc, which advises users to translate their telemetry to Loki-native and Prometheus-native formats prior to sending it to Grafana Cloud. Now that Grafana Cloud has OTLP endpoints which are Generally Available, there is no need to do these translations. In fact, it's not even recommended to do the translations anymore since the way they are done in the cloud gateway might be different from the way it's done in Alloy.

We could do a few things:
* Delete the [Collect OpenTelemetry data and forward to Grafana](https://grafana.com/docs/alloy/latest/collect/opentelemetry-to-lgtm-stack/) page.
* Update the [Collect OpenTelemetry data](https://grafana.com/docs/alloy/latest/collect/opentelemetry-data/) doc to mention some more things:
  * That Grafana Cloud has OTLP endpoints and users could send OTLP to them directly.
  * It's possible to translate data between OTLP and Prometheus, and between OTLP and Loki. Users are free to do so whenever they like, but their experience with Grafana Cloud might be best if they avoid doing translations. That way the original data will be preserved as much as possible, and Grafana Cloud will have a chance to translate the telemetry in the best way. E.g. it might use [structured metadata](https://grafana.com/docs/loki/latest/get-started/labels/structured-metadata/#when-to-use-structured-metadata) when converting OTLP resource attributes in logs.

cc @cyrille-leclerc

cyrille-leclerc:
Looks great.
One minor comment
> That Grafana Cloud has OTLP endpoints and users could send OTLP to them directly.
We have consensus to call it the "Grafana Cloud OTLP Endpoint", singular

derek-cadzow:
@Clayton, FYI for visibility.

cyrille-leclerc:
FYI the following docs update will improve the guidance for sending OTLP to Grafana:
* https://github.com/grafana/website/pull/21511

PR #5565, opened 2026-02-17
docs: Use otelcol.exporter.otlphttp in most docs examples

ptodev:
`otelcol.exporter.otlphttp` is what the Grafana databases and Grafana Cloud support. We should steer users towards that component so that they have a smoother onboarding experience.

The reason why `otelcol.exporter.otlp` has been used so far is that historically Tempo only supported gRPC, and it was also the first Grafana DB to support OTel. Today Tempo supports HTTP, and Mimir and Loki only support HTTP OTLP.

Related to #1310

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.auth.basic.md
+++ b/docs/sources/reference/components/otelcol/otelcol.auth.basic.md
@@ -142,7 +142,7 @@ These examples show how to perform basic authentication using the `client_auth` 
 
 #### Use client authentication
 
-This example configures [`otelcol.exporter.otlp`][otelcol.exporter.otlp] to use basic authentication with a single username and password combination:
+This example configures [`otelcol.exporter.otlphttp`][otelcol.exporter.otlphttp] to use basic authentication with a single username and password combination:
 
 ```alloy
 otelcol.receiver.otlp "example" {
@@ -267,3 +267,4 @@ otelcol.auth.basic "creds" {
 
 [otelcol.receiver.otlp]: ../otelcol.receiver.otlp/
 [otelcol.exporter.otlp]: ../otelcol.exporter.otlp/
+[otelcol.exporter.otlphttp]: ../otelcol.exporter.otlphttp/


docs/sources/reference/components/otelcol/otelcol.auth.basic.md
clayton-cornell, line 123: Are these supposed to stay with otlp?

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -151,23 +151,23 @@ otelcol.receiver.otlp "example" {
...
+    auth     = otelcol.auth.basic.creds.handler
   }
 }
 
 otelcol.auth.basic "creds" {
   client_auth {
-    username = "demo"
-    password = sys.env("API_KEY")
+    username = sys.env("OTLP_USERNAME")
+    password = sys.env("OTLP_API_KEY")
```
clayton-cornell:
```suggestion
    username = sys.env("<OTLP_USERNAME>")
    password = sys.env("<OTLP_API_KEY>")
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
clayton-cornell:
```suggestion

Replace the following:

* _`<OTLP_USERNAME>`_: Your OTLP username.
* _`<OTLP_API_KEY>`_: Your OTLP API key.

{{< admonition type="note" >}}
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
```diff
@@ -151,23 +151,23 @@ otelcol.receiver.otlp "example" {
...
+    traces  = [otelcol.exporter.otlphttp.default.input]
   }
 }
 
-otelcol.exporter.otlp "default" {
+otelcol.exporter.otlphttp "default" {
   client {
-    endpoint = "my-otlp-grpc-server:4317"
-    auth = otelcol.auth.basic.creds.handler
+    endpoint = `https://otlp-gateway-prod-gb-south-0.grafana.net/otlp`
```
clayton-cornell, line 162: Do we want to use a more generic endpoint? Grafana Cloud endpoints work fine of course...

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
clayton-cornell, line 170: These are env vars correct? Do we need to explain what there are?

docs/sources/shared/reference/components/otelcol-tls-tpm-block.md
```diff
@@ -17,7 +17,7 @@ The [trusted platform module][tpm] (TPM) configuration can be used for loading T
 
 The `path` attribute is not supported on Windows.
 
-### Example
+In the example below, the private key `my-tss2-key.key` in TSS2 format will be loaded from the TPM device `/dev/tmprm0`:
```
clayton-cornell:
```suggestion
In the following example, the private key `my-tss2-key.key` in TSS2 format is loaded from the TPM device `/dev/tmprm0`:
```

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
ptodev: I think it's self explanatory and doesn't need further explanation. There are no "correct" env vars - Alloy doesn't ship env vars by default, and this is just an example how to retrieve values from an env var that the user set up on their own.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
ptodev: The reason I'm putting this in is that people are sometimes confused what endpoint to use for grafana cloud. E.g. whether it should start with "http:", or whether it should contain a more specific endpoint like "/otlp/v1/metrics". 

I worked on an escalation recently where I wasn't sure if the customer used a properly formatted endpoint and I wanted to check our docs to make sure the format is correct, but I couldn't find any place in our docs where we have an example endpoint for Grafana Cloud.

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
ptodev: Yes, this is a grpc example which requires `otelcol.exporter.otlp`.

ptodev: Oops, I forgot to update this.

ptodev: Fixed

docs/sources/reference/components/otelcol/otelcol.auth.basic.md
clayton-cornell: That's something that's been on my ToDo list... to clarify or add "opinionated" examples... probably in parallel with 3rd party... something like "Here's how to do it with Grafana Cloud, and here's a not-Grafana-Cloud example too.. you can choose."

ptodev: > here's a not-Grafana-Cloud example

I think there's no need to say this. For Grafana products the Grafana Cloud example can also work for Grafana OSS DBs. Historically the only place that differed was Tempo, but as I mentioned in the PR description Tempo now also has an HTTP OTLP endpoint:

> The reason why otelcol.exporter.otlp has been used so far is that historically Tempo only supported gRPC, and it was also the first Grafana DB to support OTel. Today Tempo supports HTTP, and Mimir and Loki only support HTTP OTLP.

I'd rather not document non-Grafana DBs since it would be very hard for us to keep the docs up to date. We could do it in limited situations like with otelcol.exporter components, but that's mostly because the documentation tends to be duplicated from upstream and we can count on the community expertise there.

PR #5571, opened 2026-02-18
docs: Add a new topic, troubleshoot memory issues

### LINKED ISSUE #5387: docs: Troubleshooting memory issues
### Component(s)

_No response_

### Add a description

Troubleshooting Memory Issues

- Memory profiling guide
- GOMEMLIMIT configuration
- WAL and positions persistence
- Kubernetes resource configuration
- Memory leak detection

### Tip

clayton-cornell:
Add a new topic explaining how to troubleshoot memory issues.

Clean up the Generate a support bundle topic and add an example

Fixes: https://github.com/grafana/alloy/issues/5387 and https://github.com/grafana/support-escalations/issues/20552


docs/sources/troubleshoot/memory_issues.md
```diff
@@ -0,0 +1,218 @@
...
+
+   Verify that memory requests and limits exist.
+   If no limit is defined, set one.
+   If the limit is close to observed usage, increase it.
+
+   Set the limit high enough to absorb WAL replay and temporary queue growth.
+   In most environments, this means at least two to four times steady-state usage.
+   Refer to [Estimate resource usage][estimate-resource-usage] for baseline guidance.
+
+1. Configure `GOMEMLIMIT`.
```
kalleep: It is worth to note that alloy uses https://github.com/KimMachineGun/automemlimit

This will already configure GOMEMLIMIT to 90% if not manually configured

>By default, it sets `GOMEMLIMIT` to 90% of cgroup's memory limit.

docs/sources/troubleshoot/memory_issues.md
```diff
@@ -0,0 +1,218 @@
...
+
+   Set `GOMEMLIMIT` to approximately 90% of your container memory limit.
+   For example, with a 2GiB limit, use `GOMEMLIMIT=1800MiB`.
+
+   Without `GOMEMLIMIT`, the Go runtime may expand memory until Kubernetes terminates the container.
```
kalleep: This is not really correct. GOMEMLIMIT is a soft limit and memory can go over it. What happens is that the go runtime will start to aggressively run GC to try to stay under it.

docs/sources/troubleshoot/memory_issues.md
```diff
@@ -0,0 +1,218 @@
...
+   - Ensure [persistent storage](#configure-kubernetes-correctly) exists for WAL and positions data.
+
+1. Avoid restart loops.
+
+   Frequent restarts increase replay cost and can create repeated spikes.
+
+If spikes persist after adjusting these settings, [capture a profile][profile] during startup to identify the source.
+If profiling suggests a memory leak, refer to [Report a potential memory leak](#report-a-potential-memory-leak).
+
+## Memory grows steadily over time
```
kalleep: This is a common issue in loki pipelines caused by sustained back pressure.

For components that exposes a http server (`loki.source.api`, `loki.source.firehose` etc) this will look like a memory leak but it's not. It's just that we get more http request than alloy can process.

clayton-cornell: Added a note about this.  I'm not sure if that's the best way to handle it though...

docs/sources/troubleshoot/memory_issues.md
clayton-cornell: Changed / updated this section

docs/sources/troubleshoot/memory_issues.md
kalleep: Maybe we should keep these troubleshooting guides really general and then have specific topic for troubleshooting prometheus, loki and otel pipelines. Because the have different behaviors, pitfalls etcs

docs/sources/troubleshoot/memory_issues.md
```diff
@@ -0,0 +1,225 @@
...
+
+   Set the limit high enough to absorb WAL replay and temporary queue growth.
+   In most environments, this means at least two to four times steady-state usage.
+   Refer to [Estimate resource usage][estimate-resource-usage] for baseline guidance.
+
+1. Understand `GOMEMLIMIT` behavior.
+
+   {{< param "PRODUCT_NAME" >}} uses [automemlimit][] to automatically set `GOMEMLIMIT` to 90% of the container memory limit.
+   `GOMEMLIMIT` is a soft limit.
+   When memory approaches this threshold, the Go runtime runs garbage collection to try to stay under it.
```
kalleep:
```suggestion
   When memory approaches this threshold, the Go runtime runs garbage collection more aggressively to try to stay under it.
```

docs/sources/troubleshoot/memory_issues.md
clayton-cornell: Specific follow-up topics is easy enough we could spawn tracking issues and put them in my ToDo.

How can we keep this really general here? It is fairly general as it is... it says that data arriving faster than Alloy process it can sometimes be a leak... has a few checkpoints and then instructions for how to report to support. We could drop the note as it's a specific use-case for backpressure?

kalleep: I mean this is mentioning WAL and promethues components a lot, that I would not consider general but specific issues for promethues pipelines.

clayton-cornell: Understood. I've done a deep dive into this and mined the support playbooks/runbooks as well as the Slack triage discussions and built up a much more comprehensive set of memory troubleshooting docs. I've asked #customer-support to take a look and weigh in on the content.

docs/sources/troubleshoot/memory/loki.md
```diff
@@ -0,0 +1,196 @@
...
+menuTitle: Loki
+weight: 300
+---
+
+# Loki component memory issues
+
+Loki components in {{< param "PRODUCT_NAME" >}} handle log ingestion and forwarding.
+Memory issues with these components usually occur when back pressure builds up or when the positions file isn't persisted correctly across restarts.
+
+## Back pressure from HTTP sources
```
kalleep, line 14: Should we mention here that we can limit these http sources with `conn_limit`?

e.g. https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.awsfirehose/#http

docs/sources/troubleshoot/memory/loki.md
```diff
@@ -0,0 +1,196 @@
...
+If you're unsure whether back pressure causes the memory growth, refer to [Diagnose back pressure and queue buildup][memory-backpressure] in the memory troubleshooting overview.
+
+{{< admonition type="note" >}}
+Back pressure issues aren't memory leaks.
+Memory grows because incoming log data accumulates in memory buffers faster than {{< param "PRODUCT_NAME" >}} can process or forward it.
+{{< /admonition >}}
+
+### Diagnose back pressure
+
+1. Check log ingestion latency.
```
kalleep, line 30: This is a list using only `1.`, I guess that is not intentional

docs/sources/troubleshoot/memory/loki.md
clayton-cornell: I added something about `conn_limit` to the `Resolve back pressure `section.

docs/sources/troubleshoot/memory/loki.md
```diff
@@ -0,0 +1,199 @@
...
+   Inspect {{< param "PRODUCT_NAME" >}} logs for HTTP errors or retries when sending logs to Loki endpoints.
+   Repeated failures or retry loops can cause queues to grow and increase memory usage.
+
+1. Confirm whether traffic volume increased.
+
+   Compare log ingestion rate to historical baselines.
+   Increased load results in higher steady-state memory.
+   Refer to [Estimate resource usage][estimate-resource-usage] for more information.
+
+1. Inspect component queues.
```
kalleep: What queues are you referring to here?

In loki pipeline we only have queues in loki.write and when those are full alloy start to block. So the build up would be the number of requests. 

There are currently some metrics that can be used for components that exposes http servers to identify this but they are undocumented and I want to fix that https://github.com/grafana/alloy/issues/5751

docs/sources/troubleshoot/memory/loki.md
clayton-cornell: This is intentional. We use "lazy numbering" in the markdown. Refer tohttps://grafana.com/docs/writers-toolkit/write/markdown-guide/#numbered-lists

docs/sources/troubleshoot/memory/loki.md
clayton-cornell: Oooops, that looks like a step that was copied in from elsewhere and not removed in my own reviews.  The first 3 steps identify causes of back pressure, and the step after this confirms things... and this step isn't helping.

I will remove it. It doesn't belong here.

docs/sources/troubleshoot/memory/_index.md
```diff
@@ -0,0 +1,140 @@
...
+
+- **`OOMKilled` or startup crashes**: Refer to [Kubernetes memory issues][kubernetes] for resource configuration and persistent storage guidance.
+- **Memory spikes after restart or WAL issues**: Refer to [Prometheus component memory issues][prometheus] for WAL replay and retention configuration.
+- **Back pressure from HTTP ingestion sources**: Refer to [Loki component memory issues][loki] for [`loki.source.api`][loki-source-api] and [`loki.source.awsfirehose`][loki-source-awsfirehose] troubleshooting.
+- **Memory grows steadily and series count increases**: Refer to [High cardinality memory issues][high-cardinality] for guidance on identifying and reducing high-cardinality labels.
+- **Gradual memory growth during normal operation**: This usually indicates queue buildup caused by slow or failing remote systems.
+  Review endpoint latency and internal queue metrics.
+  If your configuration includes Prometheus or other metrics ingestion pipelines, refer to [Prometheus component memory issues][prometheus] for remote write queues, WAL replay behavior, and cardinality-related memory usage.
+
+## Diagnose back pressure and queue buildup
```
ptodev: Where did you see this? Usually there is a setting to drop telemetry in those cases. I've never seen high memory usage due to queue buildup.

docs/sources/troubleshoot/memory/high-cardinality.md
```diff
@@ -0,0 +1,174 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/troubleshoot/memory/high-cardinality/
+description: Learn how high cardinality can cause memory issues in Grafana Alloy
+menuTitle: High cardinality
+title: High cardinality memory issues
+weight: 250
+---
+
+# High cardinality memory issues
```
ptodev, line 9: Does high cardinality actually cause lots of memory? I can see how it would increase it somewhat, but I wonder if there's a real world situation which prompted this paragraph to be written?

docs/sources/troubleshoot/memory/high-cardinality.md
ptodev, line 1: In my experience a far more common culprit of high memory usage with metrics is a high WAL truncation time. Shortening the truncation time usually helps reduce memory. However, it comes as a tradeoff that Alloy is less resilient to network and DB outages.

Another common reason for resource consumption (for CPU, not memory) is the cache in prometheus.relabel. It needs to be configured high enough to acommodate all active series + any series that might be dropped. But it's a complicated topic to say how it should be configured best... E.g. if there's a relabel for each scrape function the cache doesn't need to be as big.

ptodev (commented):
This is so much content, and I'm honestly not sure whether it is helping. The noise-to-signal ratio in these docs tends more towards noise and less towards signal. Why not just advise users to use our existing doc for profiling, and to then refer to component reference docs for any component that is flagged as consuming lots of memory? The component reference docs should have been documented with best practices and they should state whether high memory usage could occur in some configurations.

ptodev:
Another problem with adding such verbose and high-volume documentation is that it makes it harder to check their correctness and to keep them up to date.

docs/sources/troubleshoot/memory/high-cardinality.md
clayton-cornell: Digging into this comment.  Claude AI made some suggestions and elped me update that info a little:

> For metrics pipelines, WAL retention and truncation settings are often a more common root cause than generic queue buildup, and relabel cache sizing often shows up as CPU pressure first when cache misses thrash. We updated wording to reflect that ordering: cardinality and WAL behavior are treated as primary suspects, and queue buildup is now a conditional path that requires queue metrics evidence.

Does this cover it more accurately? Is there something else we should/could say here?

For the advanced/complicated explanation surrounding the `prometheus.relabel` best practices, can we tag this as an Issue for a future update?

docs/sources/troubleshoot/memory/high-cardinality.md
clayton-cornell: I had to ask Claude for help answering this one too:

> High cardinality tends to show up first as rising series count, WAL churn, and CPU cost, then memory growth.

clayton-cornell:
> Another problem with adding such verbose and high-volume documentation is that it makes it harder to check their correctness and to keep them up to date.

That's an overall doc issue. We shouldn't avoid writing docs just because they might need future maintenance.

PR #5575, opened 2026-02-18
feat(database_observability): Add scaffolding for db-o11y integration tests

matthewnolf:
This adds the scaffolding for integration tests on the `database_observability.mysql` and `database_observability.postgres` components. It adds minimal assertions that metrics and logs are produced when the component is ran.

Will followup with tests for each respective collector.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


integration-tests/docker/tests/database-observability-mysql/config.alloy
```diff
@@ -0,0 +1,52 @@
+prometheus.exporter.mysql "mysql" {
```
cristiangreco, line 4: I'm thinking how we can possibly avoid anybody blindly copy-pasting these test configs for their prod setup. WDYT about adding a warning message of some sort insaying that they're used here just for testing and are not intended to be fully representative of a recommended configuration?

integration-tests/docker/docker-compose.yaml
```diff
@@ -122,6 +122,52 @@ services:
...
+      - mysql
+    volumes:
+      - ./tests/database-observability-mysql/init_mysql.sh:/init_mysql.sh
+    command: sh /init_mysql.sh
+    restart: "no"
+    networks:
+      - integration-tests
+
+  postgres:
+    image: postgres:16
```
cristiangreco, line 148: Are we thinking of eventually having a matrix of versions here?

integration-tests/docker/docker-compose.yaml
```diff
@@ -122,6 +122,52 @@ services:
...
+    container_name: mysql
+    ports:
+      - "3306:3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: rootpassword
+      MYSQL_DATABASE: testdb
+    networks:
+      - integration-tests
+
+  mysql-init:
```
cristiangreco: I think we could just mount the script under `/docker-entrypoint-initdb.d` and it should be executed automatically?

integration-tests/docker/docker-compose.yaml
matthewnolf: Eventually, but I'd like to start simpler

matthewnolf:
@grafana/grafana-alloy-maintainers Please can I get a review of these changes

cristiangreco (approved):

PR #5580, opened 2026-02-18
feat: GraphQL server

jharvey10:

Adds a queryable graphql endpoint to Alloy's web server. This is part 2/3 to a foundation that allows Alloy to query itself for things like readiness and component status (Part 1 was agreement on and adding of the initial schema layout). It will eventually be used for the UI (future enhancement) and the REPL (part 3/3).

This includes:
- Code models for the Alloy and Component gql types
- Resolvers for the `alloy`, `components`, and `component` queries
- Supporting utilities for readiness and local/remote component retrieval

Also includes some other random enhancements:
- Proper starting logs for Alloy at two points:
   - As early as possible after logging is ready in the command Run method
   - After the initial reload has completed

## TODO:
- [x] update public docs calling this out as a feature (do we need a feature flag?)
- [x] investigate adding an integration test for a basic pipeline

### Notes to the Reviewer

See the README for instructions on how to run and test it locally.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


internal/service/graphql/graph/resolver.go
```diff
@@ -0,0 +1,15 @@
...
+//
+// It serves as dependency injection for your app, add any dependencies you require here.
+
+import (
+	"github.com/grafana/alloy/internal/service"
+)
+
+type Resolver struct {
+	Host service.Host
+}
```
jharvey10, line 15: this is essentially a global object that becomes available to all resolvers as they process queries.

internal/service/graphql/utils/print_graphql_response.go
```diff
@@ -0,0 +1,26 @@
+package utils
+
```
jharvey10, line 2: This function is not currently used, but will be when the REPL is added later on

internal/service/graphql/graph/components.resolvers.go
```diff
@@ -0,0 +1,46 @@
...
+
+	for _, comp := range allComponents {
+		if comp.ID.String() == id {
+			result := model.NewComponent(comp)
+			return &result, nil
+		}
+	}
+
+	// Component not found
+	return nil, nil
```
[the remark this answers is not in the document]
jharvey10: This won't handle remote component retrieval as written, but I'll think about it.

docs/sources/reference/http/graphql.md
```diff
@@ -0,0 +1,132 @@
...
+None.
+
+#### Fields
+
+- **`branch`:** The git branch from which this build was created.
+- **`buildDate`:** The timestamp of when this build was created.
+- **`buildUser`:** The user account that initiated this build.
+- **`isReady`:** Whether the Alloy instance is up and running.
+- **`revision`:** The git commit hash from which this build was created.
+- **`version`:** The semantic version of this Alloy build.
```
clayton-cornell:
```suggestion
- **`isReady`:** Whether the {{< param "PRODUCT_NAME" >}} instance is up and running.
- **`revision`:** The git commit hash from which this build was created.
- **`version`:** The semantic version of this {{< param "PRODUCT_NAME" >}} build.
```

docs/sources/reference/http/graphql.md
```diff
@@ -0,0 +1,132 @@
...
+
+### `components`
+
+All components running in Alloy.
```
clayton-cornell:
```suggestion
All components running in {{< param "PRODUCT_NAME" >}}.
```

clayton-cornell:
Doc look pretty good :-)

internal/service/graphql/graph/model/alloy.go
```diff
@@ -0,0 +1,51 @@
...
+	// The git commit hash this build was created from
+	Revision string `json:"revision"`
+	// The semantic version of this Alloy build
+	Version string `json:"version"`
+
+	// Used to get at the http service for readiness checks
+	serviceHost service.Host
+}
+
+func NewAlloy(alloy Alloy, serviceHost service.Host) Alloy {
```
x1unix: Question - why there is a `Alloy` instance constructor that accepts another `Alloy` itself?

jharvey10: This is a great point and it made me rethink where the resolver vs model logic should live. After doing more research, I learned that models should be as thin as possible and query resolvers should be where the bulk of the logic lives.

I'm updating this accordingly, which means that the custom Alloy model goes away entirely. It is replaced by a simple, generated struct with all of the fields, and relies on the _resolver_ layer to have an `IsReady` method. Model structs will be built with as many "cheap" fields as they can, and will defer to the field resolvers for everything else.

The `Components` model had a similar issue, but that one reads better now too. That one still has a constructor, but this is only used to stash away a reference to the "real" component info object to be used later on when a field resolver requests Arguments, Exports, or DebugInfo. More like standard dependency injection for that one.

Will push an update momentarily!

internal/service/graphql/server.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+const globalRequestTimeout = 10 * time.Second
+
+type AlloyGraphQLProvider struct {
+	srv        *handler.Server
+	playground http.Handler
+}
+
+func RegisterRoutes(urlPrefix string, r *mux.Router, host service.Host, logger log.Logger, enablePlayground bool) {
```
x1unix: Small nitpick: as this function has 5 arguments - is it possible to consider moving at least some of them into a struct? E.g `urlPrefix`, `enablePlayground` and `logger` (or `host`).

We might get more configuration parameters in a future and I've a concern of potential arguments creep.

internal/service/graphql/utils/print_graphql_response.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/grafana/alloy/internal/service/graphql/client"
+)
+
+// PrintGraphQLResponse pretty prints a GraphQL response.
```
x1unix, line 10: Is this func used anywhere?

docs/sources/reference/http/graphql.md
```diff
@@ -0,0 +1,132 @@
...
+
+Information about the running Alloy instance.
+
+#### Arguments
+
+None.
+
+#### Fields
+
+- **`branch`:** The git branch from which this build was created.
```
x1unix: imo having GQL API described in a doc introduces additional maintenance burden + person needs to know that this doc needs to be updated as well.

Can we consider publishing (by CI) something like Swagger doc and keep a link to it here?

internal/alloycli/cmd_run.go
```diff
@@ -205,19 +208,33 @@ type alloyRun struct {
...
-		return fmt.Errorf("the '--feature.prometheus.direct-fanout.enabled' can be used only at experimental stability level")
+		return fmt.Errorf("'--feature.prometheus.direct-fanout.enabled' %s", errMsg)
+	}
+
+	if fr.enableGraphQL {
+		return fmt.Errorf("'--feature.graphql.enabled' %s", errMsg)
+	}
+
+	if fr.enableGraphQLPlayground {
+		return fmt.Errorf("'--feature.graphql-playground.enabled' %s", errMsg)
```
x1unix, line 237: The doc states that this flag requires requires `--feature.graphql.enabled`. Can you please add validation whether `fr.enableGraphQL` is also true.

internal/service/graphql/utils/marshal_body_to_string.go
```diff
@@ -0,0 +1,15 @@
+package utils
+
+import "github.com/grafana/alloy/syntax/encoding/alloyjson"
+
+// MarshalBodyToString marshals the provided value to a JSON string, safely accounting for nil values.
+func MarshalBodyToString(v any) string {
+	if v == nil {
+		return "{}"
```
x1unix: Shouldn't this be `null`?

internal/service/graphql/server.go
```diff
@@ -0,0 +1,76 @@
...
+
+	r.Handle(path.Join(urlPrefix, "/graphql"), provider.srv)
+
+	if enablePlayground {
+		level.Info(logger).Log("msg", "GraphQL playground is enabled")
+		r.Handle(path.Join(urlPrefix, "/graphql/playground"), provider.playground)
+	}
+}
+
+func NewAlloyGraphQLProvider(urlPrefix string, host service.Host, enablePlayground bool) *AlloyGraphQLProvider {
```
x1unix: Seems like this func is used just by `RegisterRoutes` and can be private.

internal/service/graphql/tools.go
```diff
@@ -0,0 +1,7 @@
+//go:build tools
+
+package graphql
+
+import (
+	_ "github.com/99designs/gqlgen"
```
x1unix: This file and the build tag is used for a specific file generation case?

I couldn't find the `tools` tag reference in Makefile changes.

internal/service/graphql/graph/model/models_gen.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Health status of the component.
+type Health struct {
+	// Message of the health status.
+	Message string `json:"message"`
+	// Last updated time of the health status.
+	LastUpdated time.Time `json:"lastUpdated"`
+}
+
+type Query struct {
```
x1unix, line 34: Is this type used somewhere?

internal/service/graphql/client/graphql_client.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	req, err := http.NewRequest("POST", c.endpoint, bytes.NewBuffer(jsonBody))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	for key, value := range c.headers {
+		req.Header.Set(key, value)
```
x1unix: You can use the [`http.Header`](https://pkg.go.dev/net/http#Header) type in `GraphQLClient.headers` and instead of copy, just share it between requests:

```go
req.Header = c.headers
```

internal/service/graphql/server.go
jharvey10: done! Went with:

```
type RegisterRoutesParams struct {
	Router           *mux.Router
	Logger           log.Logger
	URLPrefix        string
	Host             service.Host
	EnablePlayground bool
}

func RegisterRoutes(params RegisterRoutesParams) {
```

I'm generally a big fan of kwargs / struct-oriented args.

As a follow on to this work, it might be worth revisiting the web server infrastructure in alloy. It's a lot of hand-crafed things and alloy-specific patterns which I think could be simplified by a library like Gin, Fiber, or Echo. The as-is state was a little challenging to find a proper place to extend the logic which had everything I needed.

internal/service/graphql/utils/print_graphql_response.go
jharvey10: No, not currently. This will be used in "part 3" of this work (the REPL). Could just as easily be put into a separate PR, but I left it here since it's already done.

docs/sources/reference/http/graphql.md
jharvey10: Definitely agree that we don't want to hand-tend this. I took a page from GitHub as far as how they manifest their graphql API. They have a full schema and they also have rich docs pages for the various pieces as well. 

As a medium/long term goal, I think we could follow a similar pattern as we do for other generated things in our repo, whereby we have some `make` rules to build this stuff automatically and CI checks to ensure it happens when the source changes. That could extend to a full schema file as well that we publish. Given the limited scope of the API right now, I don't think we need it out of the gate, but I can make a follow on issue for that.

For this pass, I pointed AI at the schema, told it the format I wanted, and let it generate the docs automatically with just a one-shot prompt.

internal/alloycli/cmd_run.go
jharvey10: This check is covered in [ui.go](https://github.com/grafana/alloy/pull/5580/changes#diff-fa22d57fc99731fbe10b64291582775ff1594d6d1861c44822114721f0cb9578R91). I debated whether both checks should live in the same function or not, but ultimately decided to split them.

That way when `RegisterRoutes` is called, it's gurarnteed to do some amount of registration. It'll register one route if gql is on and playground is off, and two routes if they're both on.

internal/service/graphql/server.go
jharvey10: done!

internal/service/graphql/utils/marshal_body_to_string.go
jharvey10: Yeah. I think what happened here was I originally was using `""`, but that fails parsing via `JSON.parse` on the UI side, but alloyjson actually uses `[]` to represent the empty case, so the new logic is mirroring that (for Exports and DebugInfo). It lets `nil` flow through untouched and will use `[]` as a fallback for `Arguments`, which is never supposed to be nil (i.e. an error case that's likely impossible to hit).

example:

```go
// Exports is the resolver for the exports field.
func (r *componentResolver) Exports(ctx context.Context, obj *model.Component) (*string, error) {
	if obj.ComponentInfo.Exports == nil {
		return nil, nil
	}
	body, err := alloyjson.MarshalBody(obj.ComponentInfo.Exports)
	if err != nil {
		return nil, err
	}
	result := string(body)
	return &result, nil
}
```

internal/service/graphql/tools.go
jharvey10: I found a more modern way to specify a tool dependency, so I switched to that!

internal/service/graphql/graph/model/models_gen.go
jharvey10: The short answer is, "no". lol. This is part of gqlgen's generated stuff, so even if I remove it, it comes right back on the next generate.

I think it's probably them adding a `type` in go for every `type` found in the graphql schema (see base.graphqls), but yeah there's no references to it in the code anywhere that I can find.

internal/service/graphql/client/graphql_client.go
jharvey10: I adjusted this to use `http.Header` and clone the configured headers for each request. I’m hesitant to assign `req.Header = c.headers` directly because that would share a mutable header map between the client and every request. Since there’s no locking around `SetHeader` or request execution, cloning avoids introducing an implicit immutability/thread-safety contract for callers.

internal/service/ui/ui.go
```diff
@@ -25,9 +27,11 @@ const ServiceName = "ui"
...
 // lifetime of the UI service.
 type Options struct {
-	UIPrefix        string                        // Path prefix to host the UI at.
-	CallbackManager livedebugging.CallbackManager // CallbackManager is used for live debugging in the UI.
-	Logger          log.Logger
+	UIPrefix                string                        // Path prefix to host the UI at.
+	CallbackManager         livedebugging.CallbackManager // CallbackManager is used for live debugging in the UI.
+	Logger                  log.Logger
+	EnableGraphQL           bool // Whether the GraphQL API is enabled.
+	EnableGraphQLPlayground bool // Whether the GraphQL playground UI is enabled.
```
x1unix:
```suggestion
// UIPrefix is the path prefix to host the UI at.
	UIPrefix string

	// CallbackManager is used for live debugging in the UI.
	CallbackManager livedebugging.CallbackManager

	// Logger is used for structured logging across the service.
	Logger log.Logger

	// EnableGraphQL specifies whether the GraphQL API is enabled.
	EnableGraphQL bool

	// EnableGraphQLPlayground specifies whether the GraphQL playground UI is enabled.
	EnableGraphQLPlayground bool
```

x1unix (approved):

internal/service/graphql/utils/components_helper.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/grafana/alloy/internal/service"
+	"github.com/grafana/alloy/internal/service/remotecfg"
+)
+
+var defaultInfoOpts = component.InfoOptions{
+	GetHealth:    true,
+	GetArguments: true,
+	GetExports:   true,
+	GetDebugInfo: true,
+}
```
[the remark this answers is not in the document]
jharvey10: This is not a bad idea, but given the limited scope and additional arg passing it will require, I'll defer it to later.

internal/service/graphql/graph/components.resolvers.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/grafana/alloy/internal/service/graphql/graph/model"
+	"github.com/grafana/alloy/internal/service/graphql/utils"
+	"github.com/grafana/alloy/syntax/encoding/alloyjson"
+)
+
+// Arguments is the resolver for the arguments field.
+func (r *componentResolver) Arguments(ctx context.Context, obj *model.Component) (string, error) {
+	body, err := alloyjson.MarshalBody(obj.ComponentInfo.Arguments)
+	if err != nil {
+		return "[]", err // [] is the "no arguments" behavior from alloyjson
```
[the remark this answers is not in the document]
jharvey10: Both values (the value and the error) will be returned as per GraphQL semantics for partial results.

internal/service/ui/ui.go
jharvey10: done! I always forget about the `// <symbol name> does this thing` comment format of Go. Thanks for keeping me honest!

PR #5581, opened 2026-02-18
feat(loki.secretfilter): Add configurable processing timeout with timeout metrics

kleimkuhler:
Progress on https://github.com/grafana/alloy/issues/3694. Adds a per-entry processing timeout to `loki.secretfilter`. If filtering a line takes too long, it can be eagerly terminated so that it doesn't back up the rest of the processing pipeline.

### New configuration

- `processing_timeout`: maximum time allowed to process a single log entry. Defaults to `0` (disabled)
- `drop_on_timeout`: when `true`, entries that exceed `processing_timeout` are dropped instead of forwarded unredacted. Defaults to `false`.

### New metrics

- `loki_secretfilter_lines_timed_out_total`: incremented on every timeout, regardless of whether the entry was dropped or forwarded.
- `loki_secretfilter_lines_dropped_total`: incremented only when `drop_on_timeout = true` and an entry is dropped.


internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -38,11 +38,13 @@ func init() {
...
+	RedactWith        string              `alloy:"redact_with,attr,optional"`        // Template for redaction placeholder; $SECRET_NAME and $SECRET_HASH are replaced. When set, percentage-based redaction is not used.
+	RedactPercent     uint                `alloy:"redact_percent,attr,optional"`     // When redact_with is not set: percent of the secret to redact (1-100; gitleaks-style: show leading (100-N)% + "...", 100 = "REDACTED"). 0 or unset defaults to 80.
+	GitleaksConfig    string              `alloy:"gitleaks_config,attr,optional"`    // Path to a gitleaks TOML config file; if empty, the default gitleaks config is used
+	ProcessingTimeout time.Duration       `alloy:"processing_timeout,attr,optional"` // Maximum time allowed to process a single log entry. 0 (default) disables the timeout.
+	DropOnTimeout     bool                `alloy:"drop_on_timeout,attr,optional"`    // When true, entries that exceed processing_timeout are dropped instead of forwarded unredacted. Requires processing_timeout to be set.
```
mikefat, line 47: Do we really want this to exist? I would not give this option - if the line takes too long, we should always forward it. 
1. It's worse to drop legit lines w/o secrets than to have a long line with a secret show up in loki (you can mitigate a leaked secret and even detect a secret later in the pipeline, but a dropped line isn't coming back).
2. A misconfiguration or overlook could cause lines to be silently dropped. Even for an experimental feature this seems overly risky.

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -89,6 +91,8 @@ type Component struct {
 //   - loki_secretfilter_secrets_redacted_by_rule_total: Number of secrets redacted, partitioned by rule name.
 //   - loki_secretfilter_secrets_redacted_by_origin: Number of secrets redacted, partitioned by origin label value (only registered when origin_label is set).
 //   - loki_secretfilter_processing_duration_seconds: Summary of time taken to process and redact log entries.
+//   - loki_secretfilter_lines_timed_out_total: Total number of log lines that exceeded the processing timeout (regardless of whether they were dropped or forwarded).
+//   - loki_secretfilter_lines_dropped_total: Total number of log lines dropped due to processing timeout (subset of lines_timed_out_total, only when drop_on_timeout is true).
```
mikefat, line 95: Since I would not keep the dropping lines functionality, I would also remove all the code with regards to the `lines_dropped_total` metric

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -253,21 +281,47 @@ func (c *Component) Run(ctx context.Context) error {
...
+			ch <- entry
+			return
+		}
+		ch <- c.redactLine(entry, findings)
+	}()
+
+	select {
+	case result := <-ch:
+		return result, false
+	case <-time.After(timeout):
```
mikefat: In go it's more idiomatic to use `context.WithTimeout` rather than just a timer. The `cancel()` method also ensures one doesn't leave timers ticking in the background like it would happen with this code (the timer is not reclaimed until it fires).
It looks like
```go
timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()
```
and later on you add a clause to check if the context exceeded the deadline instead of the `time.After` call
```go
case <-timeoutCtx.Done():
 ...
```
You will have to pass the original context to the `processEntry` method as well.
This will also prevent issues like the parent context being cancelled but the component still running and waiting for its own timeout

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -253,21 +281,47 @@ func (c *Component) Run(ctx context.Context) error {
...
-	return c.redactLine(entry, findings)
+	ch := make(chan loki.Entry, 1)
+	go func() {
+		findings := c.detector.DetectString(entry.Line)
+		if len(findings) == 0 {
+			ch <- entry
+			return
+		}
+		ch <- c.redactLine(entry, findings)
+	}()
```
mikefat: This spawns one goroutine per line, and that will be a big impact on performance under load (not so much for the goroutine per se but the number and their management). I don't think this is a good approach.

I think the only realistic option would be for gitleaks' `DetectString" to take a context (with timeout) and honor ctx.Done(). That way we wouldn't have to spawn a goroutine. I've checked their implementation and they do create a context almost right away, so if we forked it it looks like it could be an easy change.

There's also [NewDetectorContext](https://github.com/gitleaks/gitleaks/blob/v8.30.0/detect/detect.go#L121) which is the same as the Detector but allows a custom context to be passed. The caveat is that we'd be provisioning a detector on every line. I'd prefer the refactor above.

Either way, needs more investigation.

internal/component/loki/secretfilter/secretfilter.go
mikefat: Thinking about [the refactor mentioned in my other comment](https://github.com/grafana/alloy/pull/5581/changes#r2825712287), we would create a new context with timeout on a new line and then use that context down the line w/o spawning additional goroutines

internal/component/loki/secretfilter/secretfilter.go
mikefat: Actually, something like this should work (instead of DetectString and the go routines):

```go
c.detector.DetectContext(ctx, detect.Fragment{Raw: entry.Line})
```
Where `ctx` is the context with timeout

internal/component/loki/secretfilter/secretfilter.go
kelnage: I disagree; we should leave it user configurable. Some of our users will want to be more cautious, others will want to be less so. We have similar behavior in other components (e.g., [stage.eventlogmessage](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stageeventlogmessage), [stage.limit](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagelimit)) so I don't consider this to be outside of the norm for Alloy.

internal/component/loki/secretfilter/secretfilter.go
kleimkuhler: Good call yea I made the change to use `DetectContext` and got rid of the Go routine 👍

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -253,21 +281,39 @@ func (c *Component) Run(ctx context.Context) error {
...
+	}
+
+	//nolint:staticcheck // DetectContext still requires detect.Fragment in v8
+	findings := c.detector.DetectContext(ctx, detect.Fragment{Raw: entry.Line})
 
-	// If no secrets found, return the original entry
-	if len(findings) == 0 {
-		return entry
+	if ctx.Err() != nil {
+		c.metrics.linesTimedOutTotal.Inc()
```
kelnage, line 305: At this point we may have `findings`, even though the detection function timed out. If we are doing to pass this line through, it would seem appropriate to still redact those `findings`.

kleimkuhler: Yep good find! Added

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -253,21 +281,44 @@ func (c *Component) Run(ctx context.Context) error {
...
+	if timeout := c.args.ProcessingTimeout; timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
 
-	// If no secrets found, return the original entry
-	if len(findings) == 0 {
-		return entry
+	//nolint:staticcheck // DetectContext still requires detect.Fragment in v8
```
mikefat, line 301: Nit: do we need this comment?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -46,13 +46,15 @@ loki.secretfilter "<LABEL>" {
...
+| `drop_on_timeout`    | `bool`               | When true, drop entries that exceed `processing_timeout` instead of forwarding them unredacted.                      | `false` | no       |
+| `gitleaks_config`    | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                           | `""`    | no       |
+| `origin_label`       | `string`             | Loki label to use for the `secrets_redacted_by_origin` metric. If empty, that metric is not registered.              | `""`    | no       |
+| `processing_timeout` | `duration`           | Maximum time allowed to process a single log entry. `0` disables the timeout.                                        | `0`     | no       |
+| `redact_percent`     | `uint`               | When `redact_with` is not set: percent of the secret to redact (1–100), where 100 is full redaction.                 | `80`    | no       |
+| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`. E.g.: `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
```
```suggestion
| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`, for example, `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
```

We shouldn't use E.g. in the docs.

internal/component/loki/secretfilter/secretfilter.go
kleimkuhler: Yep!

PR #5589, opened 2026-02-19
docs: Refactor the clustering documentation

### LINKED ISSUE #226: Docs feedback: /docs/sources/flow/tasks/configure-agent-clustering.md
The [task page](https://grafana.com/docs/agent/latest/flow/tasks/configure-agent-clustering/) for configure agent clustering should have notice to inform users that when enabling clustering, components intended to distribute work (such as prometheus.scrape) must be modified to include

```
clustering {
  enabled = true 
}
```

And warn users that if they don't do this, all agents in the cluster will be doing the same work.

clayton-cornell:
Clarify and clean up the Clustering topic. Add info about enabling per component, split topics more logically.  Moved a bunch of info that was buried in word-dense Clustering section of run.md to this Clustering task topic

Fixes: https://github.com/grafana/alloy/issues/226

Review the new/updated content here: https://deploy-preview-alloy-5589-zb444pucvq-vp.a.run.app/docs/alloy/next/configure/clustering/

Followup PR: https://github.com/grafana/alloy/pull/6968


docs/sources/configure/clustering/configure-alloy.md
```diff
@@ -0,0 +1,106 @@
...
+You can customize clustering behavior with additional flags:
+
+| Flag                             | Description                                     | Default    |
+| -------------------------------- | ----------------------------------------------- | ---------- |
+| `--cluster.name`                 | Name to prevent mixing clusters.                | `""`       |
+| `--cluster.advertise-address`    | Address to advertise to other cluster nodes.    | `""`       |
+| `--cluster.advertise-interfaces` | Network interfaces to use for advertisement.    | `eth0,en0` |
+| `--cluster.rejoin-interval`      | Interval to rejoin the cluster.                 | `60s`      |
+| `--cluster.wait-for-size`        | Minimum cluster size before traffic processing. | `0`        |
+| `--cluster.wait-timeout`         | Timeout for cluster size wait.                  | `0`        |
```
ptodev: These two are highly recommended. If they are not used then users might see "err-mimir-sample-out-of-order" errors while the Alloy nodes in the cluster are discovering each other and distributing the work with each other.

clayton-cornell: Maybe I could add something like this after the optional flags table:

````
For production deployments, set `--cluster.wait-for-size` to your expected cluster size and `--cluster.wait-timeout` to a reasonable duration.
This ensures all nodes join before processing begins, which prevents duplicate data collection and out-of-order sample errors during startup.

The following example configures a 3-node cluster where each node waits up to 5 minutes for all cluster members to join before it starts processing traffic:

```bash
alloy run config.alloy \
  --cluster.enabled \
  --cluster.join-addresses=alloy-1:7946,alloy-2:7946 \
  --cluster.wait-for-size=3 \
  --cluster.wait-timeout=5m
```
````

Comment (stefan-fast) [reply docs/sources/configure/clustering/configure-alloy.md]: Hey everyone,

I caught some interest in this conversation because I’m seeing `err-mimir-sample-out-of-order` errors whenever an instance joins or leaves my Alloy cluster. While the [documentation mentions](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#out-of-order-errors) that these errors are "normal" when the number of instances changes, I’m trying to understand if using `--cluster.wait-for-size` and `--cluster.wait-timeout` is the recommended mitigation.

Currently, the [Alloy Helm chart](https://github.com/grafana/alloy/blob/main/operations/helm/charts/alloy/templates/containers/_agent.yaml#L15-L20) doesn't seem to set these values by default.

I have two specific questions to help clarify the relationship between these settings:

1.  How exactly does waiting for a specific cluster size prevent out-of-order errors? Is it because it prevents "early" instances from starting scrapers and sending samples before the hash ring has fully settled/rebalanced? Why does it also happen when scaling down then?
2.  The [clustering docs](https://grafana.com/docs/alloy/latest/get-started/clustering/#use---clusterwait-for-size-but-with-caution) suggest these flags are for when a single instance can't handle the full load, whereas the [out-of-order docs](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/#out-of-order-errors) don't explicitly point to these flags as a solution. Are these addressing two separate problems, or is the "wait-for-size" flag the standard way to smooth out rebalancing during rolling updates?

Since I see these errors consistently during pod churn, I’d love to understand if I should be manually overriding the Helm args to include these flags.

If it makes sense the mentioned parts in the documentation could be updated accordingly as well.

Comment (clayton-cornell) [reply docs/sources/configure/clustering/configure-alloy.md]: Maybe soem slight adjustment to the text will help?


````suggestion
| `--cluster.wait-for-size`        | Minimum cluster size before traffic processing. | `0`        |
| `--cluster.wait-timeout`         | Timeout for cluster size wait.                  | `0`        |

For production deployments, set `--cluster.wait-for-size` to your expected cluster size and `--cluster.wait-timeout` to a reasonable duration.
This ensures all nodes join before processing begins, which reduces duplicate scrapes and out-of-order samples at remote write during startup.

If you leave `--cluster.wait-for-size` at the default `0`, nodes can scrape before enough peers have joined, so two nodes may write the same series and trigger errors.
Refer to [Out of order errors][remote-write-out-of-order] for additional troubleshooting.

The following example configures a 3-node cluster where each node waits up to 5 minutes for all cluster members to join before it starts processing traffic:

```bash
alloy run config.alloy \
  --cluster.enabled \
  --cluster.join-addresses=alloy-1:7946,alloy-2:7946 \
  --cluster.wait-for-size=3 \
  --cluster.wait-timeout=5m
```

If your replica count changes during normal operation, for example with a Horizontal Pod Autoscaler, choose `--cluster.wait-for-size` with a safety margin so traffic is not blocked when the cluster is legitimately smaller.
Refer to [Use `--cluster.wait-for-size`, but with caution][clustering-wait] in the clustering introduction.

[clustering-wait]: ../../get-started/clustering/#use---clusterwait-for-size-but-with-caution
{{< /admonition >}}
````

Comment (stefan-fast) [reply docs/sources/configure/clustering/configure-alloy.md]: This sounds good and makes sense to me! Thank you! 👍

File: docs/sources/configure/clustering/_index.md [hunk 2]
```diff
@@ -2,54 +2,49 @@
...
+To configure clustering, complete these two steps:
 
-## Configure {{% param "PRODUCT_NAME" %}} clustering with Helm Chart
+1. **Enable clustering in {{< param "PRODUCT_NAME" >}}**: Configure {{< param "PRODUCT_NAME" >}} itself to join a cluster using Helm chart settings or command-line flags.
+1. **Enable clustering in components**: Add a `clustering` block to each component that should participate in workload distribution.
 
-This section guides you through enabling clustering when {{< param "PRODUCT_NAME" >}} is installed on Kubernetes using the {{< param "PRODUCT_NAME" >}} [Helm chart][install-helm].
+{{< admonition type="note" >}}
+Cluster mode at the {{< param "PRODUCT_NAME" >}} level doesn't automatically distribute workload.
+You must also enable clustering in each component that should participate in the cluster.
```
dehaansa, line 22:
```suggestion
You must also enable clustering in each component that should participate in the cluster. Only components with a configured `clustering` block will distribute workload.
```

docs/sources/configure/clustering/configure-alloy.md
```diff
@@ -0,0 +1,133 @@
...
+To distribute workload, you must also [enable clustering in individual components][distribute-workload].
+{{< /admonition >}}
+
+## Configure clustering with a Helm chart
+
+You can enable clustering when {{< param "PRODUCT_NAME" >}} is installed on Kubernetes with a [Helm chart][install-helm].
+
+### Before you begin
+
+For multi-replica clustering on Kubernetes, Grafana Alloy recommends setting `controller.type` to `statefulset`.
```
dehaansa: The recommendation feels like it should have some kind of justification or reasoning. I don't love my rewrite but something like this?

```suggestion
For multi-replica clustering on Kubernetes, {{< param "PRODUCT_NAME" >}} typically recommends setting `controller.type` to `statefulset` to reduce the resource impact on a cluster of having a `daemonset` deploy {{< param "PRODUCT_NAME" >}} on each node.
```

docs/sources/configure/clustering/configure-alloy.md
```diff
@@ -0,0 +1,133 @@
...
+
+You can customize clustering behavior with additional flags:
+
+| Flag                             | Description                                      | Default    |
+| -------------------------------- | ------------------------------------------------ | ---------- |
+| `--cluster.advertise-address`    | Address to advertise to other cluster nodes.     | `""`       |
+| `--cluster.advertise-interfaces` | Network interfaces to use for advertisement.     | `eth0,en0` |
+| `--cluster.enable-tls`           | Enables TLS for cluster communication.           | `false`    |
+| `--cluster.max-join-peers`       | Number of peers to join from the discovered set. | `5`        |
+| `--cluster.name`                 | Name to prevent mixing clusters.                 | `""`       |
```
dehaansa: Not sure on this one, but I feel like "mixing clusters" isn't a clear description of the behavior.

```suggestion
| `--cluster.name`                 | Shared name to prevent nodes from joining the wrong cluster(s).                 | `""`       |
```

docs/sources/configure/clustering/distribute-workload.md
```diff
@@ -0,0 +1,147 @@
...
+    enabled = true
+  }
+}
+```
+
+## How workload distribution works
+
+When you enable clustering in a component:
+
+1. All cluster nodes use a consistent hashing algorithm to determine target ownership.
```
dehaansa, line 128: Maybe worth a note to say "If target labels are not consistent across the cluster, the hashing algorithm will not be able to correctly determine target ownership. Ensure that target discovery does not add labels like constants.hostname that will prevent consistent hashing"

docs/sources/configure/clustering/configure-alloy.md
clayton-cornell: Ah.. I missed that one in my cleanup. :-(

how about something like this:

> For multi-replica clustering on Kubernetes, set `controller.type` to `statefulset`.

This removes the `recommends` part... 

Then I asked Claude to help explain and several iterations ended up with this:

> A StatefulSet gives each Pod a stable network identity for peer discovery.
> It also lets you choose how many {{< param "PRODUCT_NAME" >}} Pods to run instead of running one Pod on every node.
> Use a DaemonSet only when your configuration must collect node-local data.

docs/sources/configure/clustering/configure-alloy.md
clayton-cornell: What about this:

> Cluster name that nodes must share to join the same cluster.

docs/sources/configure/clustering/distribute-workload.md
clayton-cornell: Added a note

docs/sources/configure/clustering/_index.md
clayton-cornell: Done

docs/sources/configure/clustering/configure-alloy.md
dehaansa: Yeah, I think that works. 👍

dehaansa (approved):

PR #5613, opened 2026-02-20
fix: Stop components in a deterministic order

kalleep:
### Pull Request Details
The order we use to stop components either on config update or when alloy stops is not deterministic and is done async.

This is not optimal due to the nature of how pipeline works. Take a loki pipeline as an example
```
loki.source.file "f" {}

loki.process "p" {}

loki.write "w" {}
```

In this pipeline the optimal order is to stop at the source, then process and last write. When we create our components it's done in the opposite order and that is correct.

To do this I added `WeaklyConnectedComponents` with filter support and updated `Synchronize` to accept the graph instead.

We use WeaklyConnectedComponents and filter out only leaf nodes for each "sub" graph. For these leaves we can the do a topological walk on each sub graph. The order we get if from leaves to roots, for every node we increase it's rank. For starting we sort all task in ascending order and when stopping we sort them by descending order. Each sub graph also get's a ephemeral groupID that we can use to run start/stop for different graphs in parallel. The rank and groupID is only valid until next Syncronize call so we need to update stored tasks.

Fixes: 330

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/runtime/internal/controller/scheduler.go
+++ b/internal/runtime/internal/controller/scheduler.go
@@ -53,12 +53,10 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
 //
 // 1. Nodes already managed by the scheduler will be unchanged.
 // 2. Nodes which are no longer present will be told to shutdown.
-// 3. Nodes will be given 1 minute to shutdown before new nodes are created.
-// 4. Wait for any remaining nodes to shutdown
+// 3. Nodes that is not already managed by the scheduler will be started.
 //
 // Nodes are shutdown first to ensure any shared resources, such as ports,
-// are allowed to be freed before new nodes are scheduled. As a means to avoid,
-// long stretches of downtime we give this a 1 minute timeout.
+// are allowed to be freed before new nodes are scheduled.
 //
 // Tasks are stopped from roots to leaves and started from leaves to roots.
 //


internal/runtime/alloy.go
```diff
@@ -265,9 +265,11 @@ func newController(o controllerOptions) (*Runtime, error) {
 // Run starts the Alloy controller, blocking until the provided context is
 // canceled. Run must only be called once.
 func (f *Runtime) Run(ctx context.Context) {
-	defer func() { _ = f.sched.Close() }()
-	defer f.loader.Cleanup(!f.opts.IsModule)
-	defer level.Debug(f.log).Log("msg", "Alloy controller exiting")
+	defer func() {
+		level.Debug(f.log).Log("msg", "Alloy controller exiting")
+		f.loader.Cleanup(!f.opts.IsModule)
+		f.sched.Stop()
```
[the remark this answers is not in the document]
kalleep: Bad bot. This is the same order

internal/runtime/alloy.go
```diff
@@ -285,35 +287,8 @@ func (f *Runtime) Run(ctx context.Context) {
...
-				runnables = append(runnables, fe)
-			}
-
-			// Only the root controller should run services, since modules share the
-			// same service instance as the root.
-			if !f.opts.IsModule {
-				for _, svc := range services {
-					runnables = append(runnables, svc)
-				}
-			}
```
kalleep, line 309: When using the graph we don't have to care about this because services are never added to the graph in a module

internal/runtime/internal/controller/scheduler_test.go
```diff
@@ -234,12 +172,40 @@ func TestScheduler_Synchronize(t *testing.T) {
...
+				return n == id
+			})
+		}
+
+		for _, e := range g.Edges() {
+			from := indexOf(stopOrder, e.From.NodeID())
+			to := indexOf(stopOrder, e.To.NodeID())
+			require.Less(t, from, to)
+		}
+	})
```
kalleep, line 208: We cant guarantee the exact order for components on the same "level" because internally graph uses a bunch of maps and iteration of maps are randomized. So the test just asserts that we stop from before to

internal/runtime/internal/controller/scheduler.go
```diff
@@ -63,113 +60,115 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
-		close(doneStopping)
-	}()
-
-	stoppingTimedOut := false
-	select {
-	case <-doneStopping:
-		// All tasks stopped successfully within timeout.
-	case <-time.After(TaskShutdownWarningTimeout):
-		level.Warn(s.logger).Log("msg", "Some tasks are taking longer than expected to shutdown, proceeding with new tasks")
-		stoppingTimedOut = true
```
kalleep, line 112: I removed this for now, I don't see the point of us having this, we need to wait for stopped components anyway and without this at least we can garantuee that we don't start any new tasks before we have stopped tasks that was removed

internal/runtime/internal/controller/scheduler.go
```diff
@@ -63,113 +60,115 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
+	s.tasksMut.Unlock()
+
+	for _, group := range toStop {
+		go func() {
+			// NOTE: we cannot hold lock when calling Stop because onDone will mutate running tasks.
+			for _, t := range stopOrder(group) {
+				t.Stop()
+			}
+		}()
+	}
```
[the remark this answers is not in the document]
kalleep: Not true since [v1.22](https://go.dev/doc/go1.22#language)

internal/runtime/internal/controller/scheduler_test.go
```diff
@@ -282,3 +248,31 @@ func (sb *syncBuffer) String() string {
...
+		set[e.To] = struct{}{}
+	}
+
+	nodes := make(map[string]*fakeRunnable, len(set))
+	for id := range set {
+		nodes[id] = &fakeRunnable{
+			ID: id,
+			Component: mockComponent{RunFunc: func(ctx context.Context) error {
+				<-ctx.Done()
+				*stopOrder = append(*stopOrder, id)
```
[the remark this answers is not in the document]
kalleep: Not true since [v1.22](https://go.dev/doc/go1.22#language)

internal/runtime/internal/controller/scheduler.go
```diff
@@ -63,113 +60,115 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
 			continue
 		}
-
-		level.Debug(s.logger).Log("msg", "Stopping task", "id", id)
-		stopping.Add(1)
-		go func(t *task) {
-			defer stopping.Done()
-			t.Stop()
-		}(t)
+		toStop[t.groupID] = append(toStop[t.groupID], t)
```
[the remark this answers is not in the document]
kalleep: This is fine, we just update for tasks that are kept. So every task we want to stop still belongs to the same groupID.

internal/dag/weak_test.go
```diff
@@ -0,0 +1,96 @@
...
+		g.AddEdge(Edge{From: c, To: d})
+
+		got := WeaklyConnectedComponents(&g, FilterAllFunc)
+		got = sortComponents(got)
+
+		require.Equal(t, [][]Node{{a, b}, {c, d}}, got)
+	})
+
+	t.Run("single graph", func(t *testing.T) {
+		// One component: a->b->c
```
ptodev: What does "component" mean in this case? Aren't there three components?

internal/dag/weak.go
```diff
@@ -0,0 +1,70 @@
...
+
+func FilterAllFunc(_ *Graph, _ Node) bool {
+	return true
+}
+
+func FilterLeavesFunc(g *Graph, n Node) bool {
+	return len(g.outEdges[n]) == 0
+}
+
+func FilterRootsFunc(g *Graph, n Node) bool {
```
ptodev, line 16: The name `FilterRootsFunc` doesn't make it clear whether roots are retained or removed. It'd be good to have comments here and for the above functions, or to have a more clear name.

internal/runtime/internal/controller/scheduler.go
```diff
@@ -63,113 +60,115 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
...
-	var stopping sync.WaitGroup
 	for id, t := range s.tasks {
-		if _, keep := newRunnables[id]; keep {
+		if dt, keep := desiredTasks[id]; keep {
+			t.groupID = dt.groupID
+			t.rank = dt.rank
 			continue
 		}
-
-		level.Debug(s.logger).Log("msg", "Stopping task", "id", id)
```
ptodev, line 90: Do we need to keep log lines like this? If there are any issues with the change do we have enough observability to diagnose them?

internal/runtime/internal/controller/scheduler_test.go
ptodev, line 1: I wonder if subtle bugs can happen with partial graph updates that partially fail. I think Alloy will apply as many changes as it can while rejecting changes that it can't. It won't transactionally reject the whole update if it anything doesn't work.

Do we have tests for such partial graph updates that fail? E.g. if component A is referenced by component B, and A gets deleted, then I suppose A will still be deleted and B will be unhealthy?

internal/dag/weak_test.go
kalleep: Its just the wording used by this kind of algorithm, I removed the comment but essentially one component is an isolated "sub" graph https://www.geeksforgeeks.org/dsa/find-weakly-connected-components-in-a-directed-graph/

internal/runtime/internal/controller/scheduler_test.go
kalleep: Maybe I have not checked what happens in this case but IMHO that would be another problem that is unrelated to these changes, but still worth to take a look at.

This only changes they way we schedule start / stop of components.

internal/runtime/internal/controller/scheduler.go
```diff
@@ -63,113 +60,113 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
 // are allowed to be freed before new nodes are scheduled. As a means to avoid,
 // long stretches of downtime we give this a 1 minute timeout.
 //
+// Tasks are stopped in from roots to leaves and started from leaves to roots.
```
kgeckhart:
```suggestion
// Tasks are stopped from roots to leaves and started from leaves to roots.
```

internal/runtime/internal/controller/scheduler.go
```diff
@@ -52,7 +49,7 @@ func NewScheduler(logger log.Logger, taskShutdownDeadline time.Duration) *Schedu
 }
 
 // Synchronize adjusts the set of running components based on the provided
-// RunnableNodes in the following way,
+// graph in the following way,
```
kgeckhart, line 52: The sequence of events below is no longer valid

internal/dag/weak.go
```diff
@@ -0,0 +1,73 @@
...
+
+// FilterAllFunc includes every node.
+func FilterAllFunc(_ *Graph, _ Node) bool {
+	return true
+}
+
+// FilterLeavesFunc includes only leaves.
+func FilterLeavesFunc(g *Graph, n Node) bool {
+	return len(g.outEdges[n]) == 0
+}
```
kgeckhart, line 13: This is the only defined func we use, is it useful to keep the others or wait to add them when we need them?

internal/runtime/internal/controller/scheduler.go
kgeckhart: Since startup/shutdown is still all done together I'm not a huge fan of letting a single task bottleneck scheduling for 10 mins. Outside of the port collision issue I'm not sure how much it is a problem that in exceptional cases we can startup a new component before the other is shutdown. 

I would feel more comfortable about dropping this if we were doing it at the pipeline level rather than globally.

internal/runtime/internal/controller/scheduler_test.go
```diff
@@ -234,12 +172,40 @@ func TestScheduler_Synchronize(t *testing.T) {
...
 
-			require.NoError(t, sched.Close())
+			sched.Stop()
 
 			// Sleep long enough to let the runFunc exit to preventing a synctest panic
 			time.Sleep(time.Minute)
 		})
 	})
+
+	t.Run("Tasks are shutdown from roots to leaves within each subgraph", func(t *testing.T) {
```
kgeckhart, line 182: Should we have a test to cover the same for startup?

internal/runtime/internal/controller/scheduler.go
```diff
@@ -231,3 +236,50 @@ func (t *task) Stop() {
...
+func desiredTasksFromGraph(g *dag.Graph) map[string]desiredTask {
+	var (
+		desiredTasks = make(map[string]desiredTask, g.Len())
+		components   = dag.WeaklyConnectedComponents(g, dag.FilterLeavesFunc)
+	)
+
+	for groupID, leaves := range components {
+		var rank int
+		_ = dag.WalkTopological(g, leaves, func(n dag.Node) error {
+			if runnable, ok := n.(RunnableNode); ok {
```
kgeckhart, line 256: Is this going to be equivalent to the old code which was targeting loader components/services/imports/foreaches? Could we possibly pickup unexpected runnables now?

internal/runtime/alloy.go
kgeckhart: They still end up being picked up at the root scheduler though, right?

kalleep: Yes they are, because they are [added to the graph ](https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/loader.go#L436)

internal/runtime/internal/controller/scheduler.go
kalleep: Pipeline level would be tricky. Yes one slow task would potentially bottleneck startup of all new ones and sure this is mostly related to shared resources such as ports, maybe others but cannot think of an example right now.

We can add it back if you think it's that important but future scheduling will be blocked anyway until we have stopped them.

internal/runtime/internal/controller/scheduler.go
kalleep: will rewrite it

internal/dag/weak.go
kalleep: Yeah so first I just implemented `WeaklyConnectedComponents` without filtering but for scheduling to work correctly I only need leaves, but felt bad about only adding a specialized function for it. 

I am not sure we need the rest of them I just added it for consistency but I can remove unused ones, though I think FilterAllFunc is useful just to test that the algorithm is correct in general

internal/runtime/internal/controller/scheduler.go
kalleep: Yes it's going to be equivalent. If we did not pass runnable nodes before that would most likely been a bug.

There are some nodes in the graph that is not runnable like `logging` and `tracing`.

internal/runtime/internal/controller/scheduler_test.go
kalleep: The reason why I did not do it is because even though we call start in a specific order they run in their own goroutine so it's hard to verify in test

kalleep:
>The part I'm most concerned about is that we don't have any tests on alloy.go that we could use to prove we are going to schedule the same things at the end of this change.

We do have some, like specific tests for modules, services, foreach and a couple of others. But yes I should probably add some test to ensure that all works correctly. 

I can do that in another pr and then once that is merged I will rebase this branch to make sure they are still passing.

internal/runtime/internal/controller/scheduler.go
kgeckhart: Yeah no matter what we have to wait and it's probably better to wait out the stop vs enter in to more undefined behavior with potentially having overlapping components running at the same time.

Final thing, in the event we did get stalled on shutdown is our logging good enough to make it obvious? I think so based on us logging the warning after 1 minute and the failure at 10 mins but WDYT?

kgeckhart:
> > The part I'm most concerned about is that we don't have any tests on alloy.go that we could use to prove we are going to schedule the same things at the end of this change.
> 
> We do have some, like specific tests for modules, services, foreach and a couple of others. But yes I should probably add some test to ensure that all works correctly.
> 
> I can do that in another pr and then once that is merged I will rebase this branch to make sure they are still passing.

Great idea!

internal/runtime/internal/controller/scheduler.go
kalleep: I would say logging is fine but shutdown could be way longer now because we stop synchronously and the deadline is per task, are we fine with that?

kalleep:
Added a test in https://github.com/grafana/alloy/pull/5679

Let me know if you want me to expand it @kgeckhart

internal/runtime/internal/controller/scheduler.go
kgeckhart: I think at worst we are waiting a little over 10 mins because still we launch a goroutine to stop every background task. Did I miss something that might have changed that?

kalleep: So before we stopped every task async with a default timeout of 10min. Now when we stop it synchronously within each sub pipeline the max time would be e.g. 10min * number of components

kgeckhart: Ahhhh, I still think it's okay maybe we could add more diagnostic info when shutdown is taking a long time? How hard would it be to provide the number of components left to stop for us to log?

kalleep: Hm not that hard, but we do have warning log for when things are slow for each task we stop. Adding one where we log how many are left during stopping would be a bit noisy I think

kgeckhart: I don't think we need another log, I was thinking we could include the number of components left in the warning / error when it is taking too long.

PR #5615, opened 2026-02-20
fix(loki.process): New `action_on_duplicate_timestamp` config attribute to fudge identical log timestamps

### LINKED ISSUE #3837: [loki.process] Out-of-order of log lines with same timestamps parsed using stage.timestamp
### What's wrong?

When using a `loki.process` pipeline with stage `stage.timestamp`, log lines having the same timestamps (which can occur quite frequently, depending on the timestamp precision) are properly parsed, but order is not guaranteed by alloy.

This is assuming that the incoming log stream is strictly ordered, of course, which seems to be the vast majority of cases to me.

As a consequence, downstream Loki-based Grafana dashboards may display these lines out-of-order: refreshing the dashboard is enough to shuffle these log lines, which is very disturbing.

Along the same line as https://github.com/grafana/loki/issues/3250 on `promtail`.

The `fudge` action-on-failure mechanism would actually do the trick if applied to successfully parsed entries as well (performance impacts set aside).
As a result, when enabled, this would ensure that log lines having the same timestamps keep their order respective to each other, even when Loki is configured to accept out-of-order log shipping.

I've confirmed (unit test available) that the `stage.timestamp` does indeed keep the parsed timestamp without further processing, confirming the code analysis of:
https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/timestamp.go#L139-L158

### Steps to reproduce

* Use a log file have different log lines with the same timestamp:
```
2025-06-13T09:04:05.000 Msg one
2025-06-13T09:04:05.000 Msg two
2025-06-13T09:04:05.000 Msg three
```
* Use a vanilla processing stage (see below)
* Ship it to Loki / Grafana
* Explore the logs

### System information

Linux 4.18.0-553.54.1.el8_10.x86_64

### Software version

Grafana Alloy v1.9.1

### Configuration

```text
loki.process "test" {
  forward_to = [ loki.process.local.receiver ]
  stage.timestamp {
    source = "timestamp"
    format = "2006-01-02T15:04:05.000"
    action_on_failure = "fudge"
  }
}
```

### Logs

```text

```

boussaffawalid:

When multiple log entries parse to the exact same timestamp (e.g. millisecond precision), they kept that identical value and their order was not guaranteed in Loki/Grafana. This change extends the existing "fudge" behaviour to the success path: when action_on_failure is "fudge", entries whose parsed timestamp is equal to or before the last emitted timestamp for that stream are advanced by 1 nanosecond so message order is preserved.

### Pull Request Details
**Problem:** With stage.timestamp, lines that parse to the same timestamp (common with limited precision formats) kept the same value. Alloy did not guarantee their relative order, so Loki could return them in any order and Grafana dashboards could show lines shuffled on refresh.

**Solution:** When action_on_failure = "fudge" (default), the stage now also applies fudge on success: if the parsed timestamp is not strictly after the last known timestamp for that stream, the entry’s timestamp is set to lastKnown + 1ns and the cache is updated. This reuses the existing LRU cache and 1ns step; no new config. With action_on_failure = "skip", behaviour is unchanged.

**Code:** In timestamp.go, the success path checks the last known timestamp when fudge is enabled and, when parsedTs <= lastKnown, assigns lastKnown.Add(1*time.Nanosecond) and updates the cache.
Test: New case in TestTimestampStage_ProcessActionOnFailure: multiple entries with the same "time" value; expected output timestamps are strictly increasing (parsed, +1ns, +2ns).

**Docs:** loki.process.md updated to state that with fudge, successfully parsed timestamps that are equal to or before the last emitted for the stream are advanced by 1ns to preserve message order.

Fixes: #3837
 
### Notes to the Reviewer

Fudge-on-success only runs when action_on_failure = "fudge"; the cache is already created in that case, so no extra allocation.
Existing test "should keep the parsed timestamp on success" still passes (distinct timestamps 03.400 vs 03.500, so no fudging).
Run: go test -v -race ./internal/component/loki/process/stages/... -run TestTimestampStage to verify.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1880,7 +1880,7 @@ The `action_on_failure` field defines what should happen when the source field d
 
 The supported actions are:
 
-* fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering.
+* fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering. When parsing succeeds, the stage also ensures that entries whose parsed timestamp is equal to (or before) the last emitted timestamp for that stream are advanced by 1 nanosecond, so that multiple messages with the exact same timestamp keep a stable order in Loki and Grafana.
 * skip: Don't change the timestamp and keep the time when the log entry was scraped.
 
 The following stage fetches the `time` value from the shared values map, parses it as a RFC3339 format, and sets it as the log entry's timestamp.
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1807,7 +1807,7 @@ The following arguments are supported:
 | `format`                        | `string`       | Determines how to parse the source string.                  |           | yes      |
 | `source`                        | `string`       | Name from extracted values map to use for the timestamp.    |           | yes      |
 | `action_on_failure`             | `string`       | What to do when the timestamp can't be extracted or parsed. | `"fudge"` | no       |
-| `action_on_duplicate_timestamp` | `string`       | What to do when parsing duplicate timestamps                | `"fudge"` | no       |
+| `action_on_duplicate_timestamp` | `string`       | What to do when parsing duplicate timestamps.               | `"fudge"` | no       |
 | `fallback_formats`              | `list(string)` | Fallback formats to try if the `format` field fails.        | `[]`      | no       |
 | `location`                      | `string`       | IANA Timezone Database location to use when parsing.        | `""`      | no       |
 
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1802,13 +1802,14 @@ When no timestamp stage is set, the log entry timestamp defaults to the time whe
 
 The following arguments are supported:
 
-| Name                | Type           | Description                                                 | Default   | Required |
-| ------------------- | -------------- | ----------------------------------------------------------- | --------- | -------- |
-| `format`            | `string`       | Determines how to parse the source string.                  |           | yes      |
-| `source`            | `string`       | Name from extracted values map to use for the timestamp.    |           | yes      |
-| `action_on_failure` | `string`       | What to do when the timestamp can't be extracted or parsed. | `"fudge"` | no       |
-| `fallback_formats`  | `list(string)` | Fallback formats to try if the `format` field fails.        | `[]`      | no       |
-| `location`          | `string`       | IANA Timezone Database location to use when parsing.        | `""`      | no       |
+| Name                            | Type           | Description                                                 | Default   | Required |
+| ------------------- ------------| -------------- | ----------------------------------------------------------- | --------- | -------- |
+| `format`                        | `string`       | Determines how to parse the source string.                  |           | yes      |
+| `source`                        | `string`       | Name from extracted values map to use for the timestamp.    |           | yes      |
+| `action_on_failure`             | `string`       | What to do when the timestamp can't be extracted or parsed. | `"fudge"` | no       |
+| `action_on_duplicate_timestamp` | `string`       | What to do when parsing duplicate timestamps                | `"fudge"` | no       |
+| `fallback_formats`              | `list(string)` | Fallback formats to try if the `format` field fails.        | `[]`      | no       |
+| `location`                      | `string`       | IANA Timezone Database location to use when parsing.        | `""`      | no       |
 
 {{< admonition type="note" >}}
 Be careful with further stages which may also override the timestamp.
@@ -1880,8 +1881,15 @@ The `action_on_failure` field defines what should happen when the source field d
 
 The supported actions are:
 
-* fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering. When parsing succeeds, the stage also ensures that entries whose parsed timestamp is equal to (or before) the last emitted timestamp for that stream are advanced by 1 nanosecond, so that multiple messages with the exact same timestamp keep a stable order in Loki and Grafana.
+* fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering.
 * skip: Don't change the timestamp and keep the time when the log entry was scraped.
+
+The `action_on_duplicate_timestamp` field defines what to do when parsing succeeds but the parsed timestamp is equal to (or before) the last emitted timestamp for that stream (for example, when multiple messages share the same second or millisecond).
+
+The supported actions are:
+
+* fudge (default): Set the entry timestamp to the last emitted timestamp plus 1 nanosecond so that message order is preserved in Loki and Grafana.
+* keep: Leave the parsed timestamp as-is; duplicate timestamps may appear out of order downstream.
 
 The following stage fetches the `time` value from the shared values map, parses it as a RFC3339 format, and sets it as the log entry's timestamp.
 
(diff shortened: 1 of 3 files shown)


docs/sources/reference/components/loki/loki.process.md
ptodev, line 1880: It's weird to lump this feature onto the existing `action_on_failure` attribute. Do you mind adding a new `action_on_duplicate_timestamp` attribute  that can be set to `keep` or `fudge`? It can default to `fudge`. This would make the config more consistent with `action_on_failure`.

Under the hood both config attributes can reuse the same timestamp cache.

boussaffawalid: @ptodev could you review it again

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1779,13 +1779,14 @@ When no timestamp stage is set, the log entry timestamp defaults to the time whe
...
-| `fallback_formats`  | `list(string)` | Fallback formats to try if the `format` field fails.        | `[]`      | no       |
-| `location`          | `string`       | IANA Timezone Database location to use when parsing.        | `""`      | no       |
+| Name                            | Type           | Description                                                 | Default   | Required |
+| ------------------- ------------| -------------- | ----------------------------------------------------------- | --------- | -------- |
+| `format`                        | `string`       | Determines how to parse the source string.                  |           | yes      |
+| `source`                        | `string`       | Name from extracted values map to use for the timestamp.    |           | yes      |
+| `action_on_failure`             | `string`       | What to do when the timestamp can't be extracted or parsed. | `"fudge"` | no       |
+| `action_on_duplicate_timestamp` | `string`       | What to do when parsing duplicate timestamps                | `"fudge"` | no       |
```
```suggestion
| `action_on_duplicate_timestamp` | `string`       | What to do when parsing duplicate timestamps.               | `"fudge"` | no       |
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1860,6 +1861,13 @@ The supported actions are:
 * fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering.
 * skip: Don't change the timestamp and keep the time when the log entry was scraped.
 
+The `action_on_duplicate_timestamp` field defines what to do when parsing succeeds but the parsed timestamp is equal to (or before) the last emitted timestamp for that stream (for example, when multiple messages share the same second or millisecond).
```
```suggestion
The `action_on_duplicate_timestamp` field defines what to do when parsing succeeds but the parsed timestamp is equal to, or before, the last emitted timestamp for that stream, for example, when multiple messages share the same second or millisecond.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1860,6 +1861,13 @@ The supported actions are:
...
+
+The supported actions are:
+
+* fudge (default): Set the entry timestamp to the last emitted timestamp plus 1 nanosecond so that message order is preserved in Loki and Grafana.
```
clayton-cornell, line 1891:
```suggestion
* fudge: Default. Set the entry timestamp to the last emitted timestamp plus 1 nanosecond so that message order is preserved in Loki and Grafana.
```

clayton-cornell:
Docs input is OK as-is.

boussaffawalid:
@ptodev The failing test appears to be flaky. It passes consistently when executed locally on macOS.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1860,6 +1861,13 @@ The supported actions are:
 * fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering.
 * skip: Don't change the timestamp and keep the time when the log entry was scraped.
 
+The `action_on_duplicate_timestamp` field defines what to do when parsing succeeds but the parsed timestamp is equal to, or before, the last emitted timestamp for that stream, for example, when multiple messages share the same second or millisecond.
+
+The supported actions are:
+
+* fudge: Default. Set the entry timestamp to the last emitted timestamp plus 1 nanosecond so that message order is preserved in Loki and Grafana.
```
ptodev:
```suggestion
* fudge (default): Set the entry timestamp to the last emitted timestamp plus 1 nanosecond so that message order is preserved in Loki and Grafana.
```
Changing this to be consistent with the paragraph above.

internal/component/loki/process/stages/timestamp_test.go
ptodev:
```suggestion
		"sets default action on failure and on duplicate timestamp": {
```

internal/component/loki/process/stages/timestamp_test.go
```diff
@@ -116,9 +116,23 @@ func TestTimestampValidation(t *testing.T) {
...
+				Format: time.RFC3339,
+			},
+			err: nil,
+			expectedConfig: &TimestampConfig{
+				Source:                     "source1",
+				Format:                     time.RFC3339,
+				ActionOnFailure:            "fudge",
+				ActionOnDuplicateTimestamp: "fudge",
 			},
 		},
```
ptodev:
```suggestion
```
This test case is the same as the one above.

boussaffawalid:
@ptodev Is there anything else preventing this PR from being merged? Could you please re-review it?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1860,6 +1861,13 @@ The supported actions are:
 * fudge (default): Change the timestamp to the last known timestamp, summing up 1 nanosecond to guarantee log entries ordering.
 * skip: Don't change the timestamp and keep the time when the log entry was scraped.
 
+The `action_on_duplicate_timestamp` field defines what to do when parsing succeeds but the parsed timestamp is equal to, or before, the last emitted timestamp for that stream, for example, when multiple messages share the same second or millisecond.
```
ptodev: > or before

Sorry, I just realised that old timestamps will also be fudged. Is this necessary? It could cause a few problems:

* It's confusing, since the name of the config attribute implies that it's only for duplicate timesamps.
* It could mask unintended ingestion of old logs.

I think that for the purpose of fixing the duplicate timestamp issue we shouldn't also be fudging previous timestamps?

boussaffawalid: @ptodev do you suggest changing 
`if !parsedTs.After(lastKnown)` by `parsedTs.Equal(lastKnown)`

so that it only check for duplicate ts

ptodev: Yes, it should only be checking for duplicate timestamps.

boussaffawalid: considering this test
```
{timestamp: time.Unix(1, 0), extracted: map[string]any{"time": "2019-10-01T01:02:03.400000000Z"}},
{timestamp: time.Unix(1, 0), extracted: map[string]any{"time": "2019-10-01T01:02:03.400000000Z"}},
{timestamp: time.Unix(1, 0), extracted: map[string]any{"time": "2019-10-01T01:02:03.400000000Z"}},
```

The issue with `parsedTs.Equal(lastKnown)` is that it breaks the accumulation case (3+ identical timestamps) because after the first fudge, lastKnown becomes +1ns and the next entry's parsedTs is no longer equal to it.

Either I keep it `parsedTs.After(lastKnown)` or I will have to track two separate values per label: the original parsed timestamp (for comparison) and the last adjusted/output timestamp (as the base for +1ns). 
Here is the implementation https://github.com/boussaffawalid/alloy/pull/2 

@ptodev what do you think ?

ptodev: Hi, apologies for the late reply! I was away for a few weeks.

I'm happy with your proposed way of dealing with this. It's intuitive and it works in the way a user would expect it to work. That way we also don't need to add extra info in the documentation - we just need to remove the ", or before," part. Feel free to merge it to this branch. Thanks!

boussaffawalid: Hi, no worries at all, thanks for getting back to me.
Glad we're aligned on the approach

ptodev (approved):

PR #5620, opened 2026-02-20
feat(loki.secretfilter): Add configurable `disabled_rules` config option for default gitleaks config

mikefat:

This PR adds a new optional argument for secretfilter called `disabled_rules`. This is a set of rule names for gitleaks that will be disabled when `gitleaks_config` is not set, that is, when we're using the default gitleaks config.
It defaults to `['generic-api-key]`. We have added this because we've identified this rule to create too many false positives.

### Additional Details

The gitleaks library provides an easy way to start a detector without having to ship the default config file alongside. In order to keep it that way the loading of the default config has been modified, so as to take into account the `disabled_rules` argument.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/secretfilter/secretfilter.go
+++ b/internal/component/loki/secretfilter/secretfilter.go
@@ -43,6 +43,7 @@ type Arguments struct {
 	RedactWith     string              `alloy:"redact_with,attr,optional"`     // Template for redaction placeholder; $SECRET_NAME and $SECRET_HASH are replaced. When set, percentage-based redaction is not used.
 	RedactPercent  uint                `alloy:"redact_percent,attr,optional"`  // When redact_with is not set: percent of the secret to redact (1-100; gitleaks-style: show leading (100-N)% + "...", 100 = "REDACTED"). 0 or unset defaults to 80.
 	GitleaksConfig string              `alloy:"gitleaks_config,attr,optional"` // Path to a gitleaks TOML config file; if empty, the default gitleaks config is used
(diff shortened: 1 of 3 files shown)


internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -173,11 +174,47 @@ func loadGitleaksConfig(path string) (config.Config, error) {
...
+// If GitleaksConfig is empty, the built-in default is used with args.DisabledRules removed
+// (defaulting to ["generic-api-key"] when DisabledRules is not specified).
 func newDetectorFromArgs(args Arguments) (*detect.Detector, error) {
 	if args.GitleaksConfig == "" {
-		return detect.NewDetectorDefaultConfig()
+		disabledRules := args.DisabledRules
+		if disabledRules == nil {
+			disabledRules = []string{"generic-api-key"}
+		}
+		return defaultDetectorConfig(disabledRules)
```
kleimkuhler, line 217: What about setting the default in `func SetToDefault`? It calls `*args = DefaultArguments`, so we'd change `DefaultArguments` to

```
var DefaultArguments = Arguments{
    DisabledRules: []string{"generic-api-key"},
}
```

But that way, we're using `SetToDefault` to set a default, and not setting a default here where it's not as obvious. I think this would just change to

```
return defaultDetectorConfig(args.DisabledRules)
```

mikefat:
Instead of disabling the rule we are going to add an extension of the gitleaks config file, and leave this as is, since anyone can configure the secret filter with the existing functionality to disable the generic-api-key rule if they want.

PR #5621, opened 2026-02-22
feat: Add healthcheck CLI subcommand for container health checks

### LINKED ISSUE #1891: [Feature Request] Alloy container health check
### Request

Hi,

The standard OTel collector has a health check extension that can be used in deployments to restart the container if it fails:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckv2extension/README.md

I understand the v2 extension is still experimental, and the v1 is being deprecated due to limitations. However, would still be useful to see if the container had failed. I haven't yet set up the Alloy integration with Grafana Cloud, but I will be doing so soon. 

Is there a plan to have a native health check and ready check we can use for deploying Alloy containers?

Thanks!

### Use case

As a new Grafana Cloud user, I've been setting up an observability solution with metrics, traces, and logs.

I requested to enable native histogram support in Grafana, but this seemed to break my Alloy configuration without any error why:

> ts=2024-10-15T13:11:36.774541357Z level=info msg="node exited without error" node=prometheus.remote_write.metrics

Reverting the `send_native_histograms` to `false` fixed the issue:

```
prometheus.remote_write "metrics" {
	// Exports metrics to Prometheus backend
	// https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.remote_write/
	endpoint {
		url                    = env("PROMETHEUS_SERVER_URL")
		send_native_histograms = false

		tls_config {
			insecure_skip_verify = env("TLS_ENABLED") == "false"
		}

		basic_auth {
			username = env("PROMETHEUS_USERNAME")
			password = env("PROMETHEUS_PASSWORD")
		}
	}
}
```

However, the deployment succeeded as the container wasn't detected as unhealthy.

goto1134:
## Summary

- Add a built-in `alloy healthcheck` CLI command that queries a running Alloy instance's health endpoint via HTTP
- Enable Docker `HEALTHCHECK` instructions in both Linux and Windows Dockerfiles without requiring external utilities like `curl` or `wget`
- Add CLI reference documentation for the new command

Closes #1891

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,8 @@ RUN groupadd --gid $UID $USERNAME \
     && chown -R $USERNAME:$USERNAME /var/lib/alloy \
     && chmod -R 770 /var/lib/alloy
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+  CMD ["/bin/alloy", "healthcheck"]
 ENTRYPOINT ["/bin/alloy"]
 ENV ALLOY_DEPLOY_MODE=docker
 CMD ["run", "/etc/alloy/config.alloy", "--storage.path=/var/lib/alloy/data"]
--- a/Dockerfile.windows
+++ b/Dockerfile.windows
@@ -92,6 +92,8 @@ COPY --from=builder ["/src/alloy/example-config.alloy", "C:/Program Files/Grafan
 # OpenTelemetry Collector helm chart and other ecosystem tools that expect otelcol binary.
 COPY ["packaging/docker/otelcol.cmd", "C:/Program Files/GrafanaLabs/Alloy/otelcol.cmd"]
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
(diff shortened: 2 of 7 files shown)


jharvey10:
Hey! Thanks for submitting this PR :) I'd like to actually hold off on this because I'm solving the same problem in a broader way over in this PR: https://github.com/grafana/alloy/pull/5580 (and a follow on that will come soon after to expose this through a gql and repl command).

This is basically a formalized version of a hackathon project that started over here: https://github.com/grafana/alloy/pull/4012

100% Agree that this is a problem that needs solving, and in fact I had to work around the exact issue in our integration tests [here](https://github.com/grafana/alloy/blob/df0a3ced3706f6c81425b8a7e3bcede0b4856005/integration-tests/docker/tests/otlp-metrics-default-engine/docker-compose.yaml#L17-L21).

Once the work I'm doing goes in, you'll be able to do `alloy gql "alloy{ isReady }"` as a healthcheck (and much more).

PR #5626, opened 2026-02-23
feat: Add option to "logging" block to redirect logs to Windows Event Log

### LINKED ISSUE #2525: Add config option to disable logging to stderr/winevent
### Request

Add configuration option to disable sending own logs to `stderr`.

### Use case

Currently Alloy sends all own logs to `stderr` (or, on windows, to windows events).
These logs cannot be staged, the only option to control `stderr` output is `logging.level`.
In windows environment, these logs can flood windows event journal.
If one uses `logging.write_to` to completely process Alloy's own logs, then duplicating those logs to `stderr` becomes unnessesary and excessive.
In this case, an extra option to disable output to `stderr` (while keeping it to `write_to`) would be handy.

### LINKED ISSUE #4070: The Alloy agent writes multiple events into same Windows Event Log entry
### What's wrong?

When the Alloy Agent reports its own events to the Windows Event Log, a couple of issues is raised:
1. Multiple Alloy log lines are written into same event
2. The individual log level is thereby not revealed
3. The Alloy Application Event log becomes rather special to most others
4. The Alloy Event log need special treatment in Grafana dashboards/alerts to capture errors, as most most event logs will be written as info even though there might be several error.

A parser possibility to split these into individual events before they are sent to Loki would be fine, which don't seem available for the loki.source.windowsevent, or is there that any workaround available for this?

Image

level statements in the event log

### Steps to reproduce

Write a single error in an <config>.alloy file, and there will be one or more error statements among multiple info statements in one windows event log.

### System information

Microsoft Windows [Version 10.0.20348.2031]

### Software version

Grafana Alloy v1.9.1

### Configuration

```text
declare "fetch" {
  argument "forward_to" {
    comment = "Must be a list(LogsReceiver) where collected logs should be forwarded to"
  }

  // Collect the Windows Application Event Logs
  loki.source.windowsevent "application" {
    eventlog_name          = "Application"
    forward_to             = [loki.process.app.receiver]
    use_incoming_timestamp = true
  }

  // Collect the Windows Security Event Logs
  loki.source.windowsevent "security" {
    eventlog_name          = "Security"
    forward_to             = [loki.process.sys.receiver]
    use_incoming_timestamp = true
  }

  // Collect the Windows Setup Event Logs
  loki.source.windowsevent "setup" {
    eventlog_name          = "Setup"
    forward_to             = [loki.process.app.receiver]
    use_incoming_timestamp = true
  }

  // Collect the Windows System Event Logs
  loki.source.windowsevent "system" {
    eventlog_name          = "System"
    forward_to             = [loki.process.app.receiver]
    use_incoming_timestamp = true
  }


  // Send logs to system tenant
  loki.process "sys" {
    forward_to = argument.forward_to.value

    stage.json {
      expressions = {
        level  = "levelText",
        source = "source",
        id     = "event_id",
      }
    }

    stage.labels {
      values = {
        level = "",
      }
    }

    stage.static_labels {
      values = {
        service_name = "system/eventlog",
      }
    }

    stage.structured_metadata {
      values = {
        event_log    = "channel",
        event_source = "source",
        event_id     = "id",
      }
    }

    stage.label_drop {
      values = [
        "channel",
        "computer",
      ]
    }

    stage.tenant {
      value = string.format("%s_sys", coalesce(sys.env("TENANT_PREFIX"), "cust"))
    }
  }


  // Send logs to application related tenant
  loki.process "app" {
    forward_to = argument.forward_to.value

    stage.json {
      expressions = {
        level  = "levelText",
        source = "source",
        id     = "event_id",
      }
    }

    stage.labels {
      values = {
        level = "",
      }
    }

    stage.static_labels {
      values = {
        service_name = "system/eventlog",
      }
    }

    stage.structured_metadata {
      values = {
        event_log    = "channel",
        event_source = "source",
        event_id     = "id",
      }
    }

    stage.label_drop {
      values = [
        "channel",
        "computer",
      ]
    }


    stage.tenant {
      value = string.format("%s_app", coalesce(sys.env("TENANT_PREFIX"), "cust"))
    }
  }
}
```

### Logs

```text
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=eManager_logs.fetch.default trace_id=492b353f018857368d59d8c3d7e57997 node_id=argument.tail_from_end duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=eManager_logs.fetch.default trace_id=492b353f018857368d59d8c3d7e57997 node_id=argument.forward_to duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=eManager_logs.fetch.default trace_id=492b353f018857368d59d8c3d7e57997 node_id=loki.process.file duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=eManager_logs.fetch.default trace_id=492b353f018857368d59d8c3d7e57997 node_id=loki.source.file.default duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished complete graph evaluation" controller_path=/ controller_id=eManager_logs.fetch.default trace_id=492b353f018857368d59d8c3d7e57997 duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=55b06823895f93db923e9a824896aa57 node_id=eManager_logs.fetch.default duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id="" trace_id=55b06823895f93db923e9a824896aa57 node_id=import.file.windows_event_logs duration=1.0005ms
ts=2025-07-29T07:40:23.7202804Z level=info msg="starting complete graph evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=argument.forward_to duration=0s
ts=2025-07-29T07:40:23.7202804Z level=error msg="failed to evaluate config" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node=loki.process.alloy_app err="decoding configuration: windows_event_logs:136:3: unrecognized block name \"stage.xml\""
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.process.alloy_app duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.process.app duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.source.windowsevent.setup duration=999.8µs
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.source.windowsevent.system duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.source.windowsevent.application duration=1.0013ms
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.process.sys duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished node evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 node_id=loki.source.windowsevent.security duration=0s
ts=2025-07-29T07:40:23.7202804Z level=info msg="finished complete graph evaluation" controller_path=/ controller_id=windows_event_logs.fetch.default trace_id=7163a22fa5c76c3a79384c17b7b84269 duration=2.0011ms
ts=2025-07-29T07:40:23.7202804Z level=error msg="failed to evaluate config" controller_path=/ controller_id="" trace_id=55b06823895f93db923e9a824896aa57 node=windows_event_logs.fetch.default err="updating custom component: windows_event_logs:136:3: unrecognized block name \"stage.xml\""
```

ptodev:
### Pull Request Details

Currently, Alloy can send its own logs to stderr and to loki components. This PR adds the ability to send to the Windows Event log instead of stderr.

Fixes #4070

In subsequent PRs I will add a new "none" destination to disable logging.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

BEGIN_COMMIT_OVERRIDE
BREAKING CHANGE: If Alloy is ran in a Windows service, it will now send logs to Windows Event Log by default.
END_COMMIT_OVERRIDE

### DIFF
--- a/docs/sources/reference/config-blocks/logging.md
+++ b/docs/sources/reference/config-blocks/logging.md
@@ -25,13 +25,14 @@ logging {
 
 You can use the following arguments with `logging`:
 
-| Name       | Type                 | Description                                | Default    | Required |
-| ---------- | -------------------- | ------------------------------------------ | ---------- | -------- |
-| `format`   | `string`             | Format to use for writing log lines        | `"logfmt"` | no       |
-| `level`    | `string`             | Level at which log lines should be written | `"info"`   | no       |
-| `write_to` | `list(LogsReceiver)` | List of receivers to send log entries to   | `[]`       | no       |
+| Name          | Type                 | Description                                  | Default       | Required |
+| ------------- | -------------------- | -------------------------------------------- | ------------- | -------- |
+| `destination` | `string`             | Primary log destination.                     | __See below__ | no       |
+| `format`      | `string`             | Format to use for writing log lines.         | `"logfmt"`    | no       |
+| `level`       | `string`             | Level at which log lines should be written.  | `"info"`      | no       |
+| `write_to`    | `list(LogsReceiver)` | List of receivers to send log entries to.    | `[]`          | no       |
 
-### Log level
+### `level`
 
 The following strings are recognized as valid log levels:
 
@@ -40,30 +41,29 @@ The following strings are recognized as valid log levels:
 * `"info"`: Only write logs at _info_ level or above.
 * `"debug"`: Write all logs, including _debug_ level logs.
 
-### Log format
+### `format`
 
 The following strings are recognized as valid log line formats:
 
 * `"json"`: Write logs as JSON objects.
 * `"logfmt"`: Write logs as [`logfmt`][logfmt] lines.
 
-### Log receivers
+### `write_to`
 
 The `write_to` argument allows {{< param "PRODUCT_NAME" >}} to tee its log entries to one or more `loki.*` component log receivers in addition to the default [location][].
 This, for example can be the export of a `loki.write` component to send log entries directly to Loki, or a `loki.relabel` component to add a certain label first.
 
-## Log location
+### `destination`
 
-{{< param "PRODUCT_NAME" >}} writes all logs to `stderr`.
+The following strings are recognized as valid log destinations:
 
-When you run {{< param "PRODUCT_NAME" >}} as a systemd service, you can view logs written to `stderr` through `journald`.
+* `"stderr"`: Write logs to `stderr`.
+* `"windows_event_log"`:  Windows only. Write logs to the Windows Event Log under the "Alloy" source.
 
-When you run {{< param "PRODUCT_NAME" >}} as a container, you can view logs written to `stderr` through `docker logs` or `kubectl logs`, depending on whether Docker or Kubernetes was used for deploying {{< param "PRODUCT_NAME" >}}.
+The default value of `destination` is set to `"windows_event_log"` when {{< param "PRODUCT_NAME" >}} runs as a Windows service.
+Otherwise, `destination` defaults to `"stderr"`.
 
-When you run {{< param "PRODUCT_NAME" >}} as a Windows service, logs are written as event logs.
-You can view the logs through Event Viewer.
-
-In other cases, redirect `stderr` of the {{< param "PRODUCT_NAME" >}} process to a file for logs to persist on disk.
+{{< param "PRODUCT_NAME" >}} fails to start if `destination` is set to `"windows_event_log"` and {{< param "PRODUCT_NAME" >}} is not running on Windows.
 
 ## Retrieve logs
 
@@ -72,6 +72,9 @@ You can retrieve the logs in different ways depending on your platform and insta
 **Linux:**
 
 * If you're running {{< param "PRODUCT_NAME" >}} with systemd, use `journalctl -u alloy`.
+
+**Docker:**
+
 * If you're running {{< param "PRODUCT_NAME" >}} in a Docker container, use `docker logs CONTAINER_ID`.
 
 **macOS:**
@@ -89,6 +92,7 @@ You can retrieve the logs in different ways depending on your platform and insta
 **All platforms:**
 
 * {{< param "PRODUCT_NAME" >}} writes logs to `stderr` if started directly without a service manager.
+  Redirect `stderr` of the {{< param "PRODUCT_NAME" >}} process to a file for logs to persist on disk.
 
 ## Example
 
--- a/internal/runtime/logging/handler.go
+++ b/internal/runtime/logging/handler.go
@@ -37,6 +37,17 @@ type handler struct {
 	currentFormat Format
 	inner         slog.Handler
 	replacer      func(groups []string, a slog.Attr) slog.Attr
+
+	eventLogPool sync.Pool
+}
+
+// eventLogScratch bundles a reusable buffer with a slog handler that
+// writes into it. format records which Format the cached handler was
+// built for, so a runtime format change triggers a rebuild.
+type eventLogScratch struct {
+	buf    *bytes.Buffer
+	hdlr   slog.Handler
+	format Format
 }
 
 // nesting is used since attrs and groups need to be nested in the order they were entered.
@@ -78,26 +89,37 @@ func (h *handler) Handle(ctx context.Context, r slog.Record) error {
 // which knows about the event log and the record level. Only used when
 // the event log destination is active.
 func (h *handler) handleWithEventLog(ctx context.Context, r slog.Record) error {
-	buf := bytesPool.Get().(*bytes.Buffer)
-	buf.Reset()
+	// A pool (rather than buildHandler's single shared handler) because this
+	// path reads buf.Bytes() back after Handle to pass them plus the level to Dispatch.
+	s := h.getEventLogScratch()
 	defer func() {
-		buf.Reset()
-		bytesPool.Put(buf)
+		s.buf.Reset()
+		h.eventLogPool.Put(s)
 	}()
 
-	if err := h.newSlogHandler(buf).Handle(ctx, r); err != nil {
+	if err := s.hdlr.Handle(ctx, r); err != nil {
 		return err
 	}
-	if buf.Len() == 0 {
+	if s.buf.Len() == 0 {
 		return nil // No need to dispatch empty log entries.
 	}
-	return h.w.Dispatch(buf.Bytes(), &r.Level)
+	return h.w.Dispatch(s.buf.Bytes(), &r.Level)
 }
 
-// bytesPool reuses bytes.Buffer instances across event-log slow-path
-// calls to avoid an allocation per record.
-var bytesPool = sync.Pool{
-	New: func() any { return new(bytes.Buffer) },
+// getEventLogScratch returns scratch space whose cached handler matches
+// the current format, rebuilding the handler only when the format
+// changed (or on first use). The returned buffer is already empty.
+func (h *handler) getEventLogScratch() *eventLogScratch {
+	expectFormat := h.formatter.Format()
+	s, _ := h.eventLogPool.Get().(*eventLogScratch)
+	if s == nil {
+		s = &eventLogScratch{buf: new(bytes.Buffer)}
+	}
+	if s.hdlr == nil || s.format != expectFormat {
+		s.hdlr = h.newSlogHandler(s.buf)
+		s.format = expectFormat
+	}
+	return s
 }
 
 // newSlogHandler constructs a fresh slog handler writing into w, with the
--- a/internal/runtime/logging/handler.go
+++ b/internal/runtime/logging/handler.go
@@ -22,10 +22,8 @@ import (
 // JSON or logfmt, and create a new inner handler if needed.
 
 type handler struct {
-	// w owns every sink: innerWriter (stderr), lokiWriter, tmpWriter, and the
-	// optional Windows event log. Records are formatted through a pooled slog
-	// handler that writes into a leveledWriter, which forwards the bytes plus
-	// the record's level to w.Dispatch.
+	// w owns every sink: innerWriter (stderr), lokiWriter, tmpWriter,
+	// and the optional Windows event log.
 	w         *writerVar
 	leveler   slog.Leveler
 	formatter formatter
@@ -41,14 +39,15 @@ type handler struct {
 	scratchPool sync.Pool
 }
 
-// scratch bundles a leveledWriter with a slog handler that writes through it.
-// The slog handler formats the record and writes the bytes straight to the
-// configured sinks via the leveledWriter, which carries the record's level.
-// format records which Format the cached handler was built for, so a runtime
-// format change triggers a rebuild.
 type scratch struct {
-	lw     *leveledWriter
-	hdlr   slog.Handler
+	// Carries the record's level.
+	lw *leveledWriter
+
+	// Formats the record and writes the bytes straight to the configured sinks.
+	hdlr slog.Handler
+
+	// Records which format the cached handler was built for,
+	// so a runtime format change triggers a rebuild.
 	format Format
 }
 
@@ -70,12 +69,6 @@ func (h *handler) Enabled(ctx context.Context, l slog.Level) bool {
 }
 
 func (h *handler) Handle(ctx context.Context, r slog.Record) error {
-	// Record the level on the leveledWriter, then let the cached slog handler
-	// format the record and write straight through to writerVar.Dispatch. The
-	// level lets Dispatch pick the event log's Info/Warning/Error API when one
-	// is attached; the stderr/loki/tmp sinks ignore it. Dispatch fans out to
-	// whichever sinks are active (writing nowhere if none are), so there's no
-	// separate "is anything listening?" pre-check here.
 	s := h.getScratch()
 	defer h.scratchPool.Put(s)
 
@@ -92,15 +85,15 @@ func (h *handler) getScratch() *scratch {
 		s = &scratch{lw: &leveledWriter{w: h.w}}
 	}
 	if s.hdlr == nil || s.format != expectFormat {
-		s.hdlr = h.newSlogHandler(s.lw)
+		s.hdlr = h.buildHandler(s.lw)
 		s.format = expectFormat
 	}
 	return s
 }
 
-// newSlogHandler constructs a fresh slog handler writing into w, with the
+// buildHandler constructs a fresh slog handler writing into w, with the
 // current format and the WithAttrs/WithGroup chain applied.
-func (h *handler) newSlogHandler(w io.Writer) slog.Handler {
+func (h *handler) buildHandler(w io.Writer) slog.Handler {
 	handlerOpts := slog.HandlerOptions{
 		AddSource: false,
 		Level:     h.leveler,
--- a/internal/runtime/logging/handler.go
+++ b/internal/runtime/logging/handler.go
@@ -1,7 +1,6 @@
 package logging
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -23,29 +22,32 @@ import (
 // JSON or logfmt, and create a new inner handler if needed.
 
 type handler struct {
-	// w owns every sink (innerWriter, lokiWriter, tmpWriter, and the optional event log).
-	// It also implements io.Writer so it can be passed straight into slog.NewTextHandler
-	// and slog.JSONHandler on the fast path. The slow path (event log attached) uses
-	// *writerVar's Dispatch method directly.
+	// w owns every sink: innerWriter (stderr), lokiWriter, tmpWriter, and the
+	// optional Windows event log. Records are formatted through a pooled slog
+	// handler that writes into a leveledWriter, which forwards the bytes plus
+	// the record's level to w.Dispatch.
 	w         *writerVar
 	leveler   slog.Leveler
 	formatter formatter
+	replacer  func(groups []string, a slog.Attr) slog.Attr
 
 	nested []nesting
 
-	mut           sync.RWMutex
-	currentFormat Format
-	inner         slog.Handler
-	replacer      func(groups []string, a slog.Attr) slog.Attr
-
-	eventLogPool sync.Pool
+	// scratchPool holds per-call scratch (a cached slog handler + leveledWriter).
+	// It's per handler instance so each scratch's cached handler matches this
+	// handler's nested attrs/groups, and pooled (rather than a single shared
+	// handler) because the per-call level on the leveledWriter is mutable state
+	// that can't be shared across goroutines.
+	scratchPool sync.Pool
 }
 
-// eventLogScratch bundles a reusable buffer with a slog handler that
-// writes into it. format records which Format the cached handler was
-// built for, so a runtime format change triggers a rebuild.
-type eventLogScratch struct {
-	buf    *bytes.Buffer
+// scratch bundles a leveledWriter with a slog handler that writes through it.
+// The slog handler formats the record and writes the bytes straight to the
+// configured sinks via the leveledWriter, which carries the record's level.
+// format records which Format the cached handler was built for, so a runtime
+// format change triggers a rebuild.
+type scratch struct {
+	lw     *leveledWriter
 	hdlr   slog.Handler
 	format Format
 }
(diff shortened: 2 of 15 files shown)


ptodev:
For now this requires administrator privileges. In a future PR I will get the Alloy installer to register Alloy in the Event Log so that you don't have to run Alloy with administrator rights just so that it can log to the event log.

Screenshot 2026-02-23 at 21 59 17

I tested this change in a Windows VM and I do see properly formatted Windows Event log. The warning/error/info labels work better than with the service Alloy's installer currently uses.

Screenshot 2026-02-23 at 22 01 36

I changed the config to use stderr and reloaded it. The terminal started populating with logs:

Screenshot 2026-02-23 at 22 02 31

docs/sources/reference/config-blocks/logging.md
```diff
@@ -40,30 +41,26 @@ The following strings are recognized as valid log levels:
...
 
-## Log location
+### `destination`
 
-{{< param "PRODUCT_NAME" >}} writes all logs to `stderr`.
+The following strings are recognized as valid log destinations:
 
-When you run {{< param "PRODUCT_NAME" >}} as a systemd service, you can view logs written to `stderr` through `journald`.
+* `"stderr"`: Write logs to `stderr`.
+* `"windows_event_log"` (Windows only): Write logs to the Windows Event Log under the "Alloy" source.
```
clayton-cornell:
```suggestion
* `"windows_event_log"`:  Windows only. Write logs to the Windows Event Log under the "Alloy" source.
```

docs/sources/reference/config-blocks/logging.md
```diff
@@ -25,13 +25,14 @@ logging {
...
-| `format`   | `string`             | Format to use for writing log lines        | `"logfmt"` | no       |
-| `level`    | `string`             | Level at which log lines should be written | `"info"`   | no       |
-| `write_to` | `list(LogsReceiver)` | List of receivers to send log entries to   | `[]`       | no       |
+| Name          | Type                 | Description                                                                 | Default    | Required |
+| ------------- | -------------------- | --------------------------------------------------------------------------- | ---------- | -------- |
+| `format`      | `string`             | Format to use for writing log lines                                         | `"logfmt"` | no       |
+| `level`       | `string`             | Level at which log lines should be written                                  | `"info"`   | no       |
+| `write_to`    | `list(LogsReceiver)` | List of receivers to send log entries to                                    | `[]`       | no       |
+| `destination` | `string`             | Primary log destination.                                                    | `"stderr"` | no       |
```
clayton-cornell:
```suggestion
| Name          | Type                 | Description                                                                 | Default    | Required |
| ------------- | -------------------- | ---------------------------------------------------------------------------- | ---------- | -------- |
| `destination` | `string`             | Primary log destination.                                                    | `"stderr"` | no       |
| `format`      | `string`             | Format to use for writing log lines.                                         | `"logfmt"` | no       |
| `level`       | `string`             | Level at which log lines should be written.                                 | `"info"`   | no       |
| `write_to`    | `list(LogsReceiver)` | List of receivers to send log entries to.                                    | `[]`       | no       |
```

docs/sources/reference/config-blocks/logging.md
```diff
@@ -40,30 +41,26 @@ The following strings are recognized as valid log levels:
...
+* `"stderr"`: Write logs to `stderr`.
+* `"windows_event_log"`:  Windows only. Write logs to the Windows Event Log under the "Alloy" source.
 
-When you run {{< param "PRODUCT_NAME" >}} as a container, you can view logs written to `stderr` through `docker logs` or `kubectl logs`, depending on whether Docker or Kubernetes was used for deploying {{< param "PRODUCT_NAME" >}}.
-
-When you run {{< param "PRODUCT_NAME" >}} as a Windows service, logs are written as event logs.
-You can view the logs through Event Viewer.
-
-In other cases, redirect `stderr` of the {{< param "PRODUCT_NAME" >}} process to a file for logs to persist on disk.
+When you run {{< param "PRODUCT_NAME" >}} as a Windows service, logs are written as event logs even if `destination` is set to `"stderr"`.
```
x1unix: I've a question:

What happens when Alloy is running on a Windows as a service and destination is set to `stderr`?

When service is running inside svchost, the stdio isn't available and writing into `os.Stderr` will block a goroutine infinitely.

docs/sources/reference/config-blocks/logging.md
```diff
@@ -25,13 +25,14 @@ logging {
...
 You can use the following arguments with `logging`:
 
-| Name       | Type                 | Description                                | Default    | Required |
-| ---------- | -------------------- | ------------------------------------------ | ---------- | -------- |
-| `format`   | `string`             | Format to use for writing log lines        | `"logfmt"` | no       |
-| `level`    | `string`             | Level at which log lines should be written | `"info"`   | no       |
-| `write_to` | `list(LogsReceiver)` | List of receivers to send log entries to   | `[]`       | no       |
+| Name          | Type                 | Description                                  | Default    | Required |
+| ------------- | -------------------- | -------------------------------------------- | ---------- | -------- |
+| `destination` | `string`             | Primary log destination.                     | `"stderr"` | no       |
```
kgeckhart: It feels weird to introduce this as a flag when the bulk of the combos are either invalid or the default behavior,

1. Windows service - stderr - ignored / invalid
2. Windows service - windows_event_log - default behavior
3. Windows executable - stderr - default behavior
4. Windows executable - windows_event_log - new intended behavior
5. linux/mac - stderr - default behavior
6. linux/mac - windows_event_log - invalid

Can we add a write_to for windows event log to get the new behavior without the additional invalid states?

docs/sources/reference/config-blocks/logging.md
ptodev: Currently, Alloy's Windows service doesn't run Alloy directly. It runs a [thin client executable](https://github.com/grafana/alloy/tree/main/internal/cmd/alloy-service) which then starts Alloy. That executable reads Alloy's logs from stderr and forwards them to the Event Log.

That thin client executable doesn't know the severity level of each log line. That's because all it sees is blobs of text. To infer the severity it'd have to parse that text. For performance reasons the parsing is currently very rudimentary and is leading to log lines being incorrectly labeled (e.g. an info log could show up as an error). Also, multiple lines get lumped together because it's hard to separate them out.

In a subsequent PR I'm hoping to get rid of that thin client service and to instead start Alloy directly.

docs/sources/reference/config-blocks/logging.md
ptodev: There are a few reasons why I didn't go with a write_to:

1. In a future PR I want to add a `none` option to `destination`.  It'll be useful for users who use loki.write to send Alloy's logs (#2525). They may not want to spam their event logs or journals.
2. In a future PR I'll need to remove the windows service executable so that Alloy is started directly by the Windows service (there's more info in my comment to Denys). When that happens, the Windows service will need a way to enable Windows Event Log logging without changing the config file. It'd be very impractical for the service to have to use the write_to param. My plan is to introduce an `ALLOY_LOG_DESTINATION` env var which (if defined) will change the default `destination` param in the `logging` block. This is the simplest way out that I can think of. Then, if a user who installed Alloy as a Windows service doesn't want Event Log they can set `destination` to `none` (which will override `ALLOY_LOG_DESTINATION`) and use `write_to` instead.

I agree that all those options look messy right now, but I think the 1 and 2 options are only there in the short term. After the Windows Service simplification we will simply be able to say:

> On Windows, `destination` defaults to `"windows_event_log"` when Alloy is ran as a Windows service. Otherwise,`destination` defaults to `"stderr"`.

ptodev:
I did some manual testing and everything worked ok.

On a Mac Alloy refuses to start when destination is set to windows event log:

Screenshot 2026-03-31 at 18 09 07

On a Mac Alloy starts normally without the destination parameter:

Screenshot 2026-03-31 at 18 21 06

On a Mac Alloy works the same way with an explicit stderr param:

Screenshot 2026-03-31 at 18 21 44

On a Windows service without an explicitly set "destination" parameter the logs are being shipped directly from Alloy and the formatting is correct:

Screenshot 2026-03-31 at 18 10 32

A service with an explicit destination of Windows Event log works:

Screenshot 2026-03-31 at 18 19 02

A Windows service with an explicitly set `destination = "stderr"` parameter works the same way as Alloy used to work:

Screenshot 2026-03-31 at 18 12 09

A non-service with explicit stderr destination works:

Screenshot 2026-03-31 at 18 15 24

By default a non-service uses stderr as expected:
 
Screenshot 2026-03-31 at 18 16 16

A non-service can now also log to the Event Log:

Screenshot 2026-03-31 at 18 17 40

internal/runtime/logging/logger.go
```diff
@@ -125,7 +156,33 @@ func (l *Logger) Update(o Options) error {
...
+	// destination.
+	if o.EffectiveDestination() == LogDestinationWindowsEventLog {
+		el, err := l.eventLogOpener("Alloy")
+		if err != nil {
+			return fmt.Errorf("failed to open Windows Event Log: %w", err)
+		}
+		l.windowsEventLogHandler = newWindowsEventLogHandler(el, l.level, replace)
+		if l.windowsEventLogHandler == nil {
+			return fmt.Errorf("failed to create Windows Event Log handler: %w", err)
+		}
```
kgeckhart: The semantics here are really weird, `newWindowsEventLogHandler`returning nil means we wrap a nil error? Looks like `newWindowsEventLogHandler` can only return nil when el is nil. If el is nil with a nil error that really breaks a lot of go idioms?

internal/runtime/logging/logger.go
```diff
@@ -31,15 +33,42 @@ type Logger struct {
...
+	// record may dispatch through the previous configuration — e.g. an event
+	// log → stderr reload could duplicate a record (old fanoutHandler still
+	// writes to event log while the new stderr path also writes), and an
+	// event log reload could send a record to a just-closed event log handle
+	// (which returns an error from el.Info but does not crash). Accepted
+	// because (1) config reloads are rare, (2) the worst case is a couple of
+	// duplicated or dropped records per reload, and (3) the alternative —
+	// holding bufferMut for the entire dispatch — would let a slow sink
+	// (e.g. a blocked Loki receiver) stall Update indefinitely, which is a
+	// worse failure mode.
```
kgeckhart: I think this comment might be better off in the Update function where the tradeoff is being made. It's a bit out of context here.

internal/runtime/logging/windows_handler.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Handle handles the Record.
+func (h *windowsEventLogHandler) Handle(ctx context.Context, r slog.Record) error {
+	if !h.Enabled(ctx, r.Level) {
+		return nil
+	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
kgeckhart: I'm wondering what this mu is guarding here. I see a lot of local calls and other things that appear to be thread safe?

internal/runtime/logging/options.go
```diff
@@ -12,16 +12,65 @@ import (
...
+	Destination *LogDestination `alloy:"destination,attr,optional"`
 
 	WriteTo []loki.LogsReceiver `alloy:"write_to,attr,optional"`
 }
 
+// EffectiveDestination returns the log destination to use. When Destination is
+// nil (not explicitly set by the user), it returns LogDestinationWindowsEventLog
+// if the process is running as a Windows service, and LogDestinationStderr
+// otherwise.
+func (o *Options) EffectiveDestination() LogDestination {
```
kgeckhart: How does this end up working? There's a SetToDefault() from syntax that's going to make it Stderr and UnmarshalText will also make it Stderr if it's empty.

internal/runtime/logging/fanout_handler.go
kgeckhart, line 1: Might be worth it to have some tests on this even if it's small

kgeckhart (commented):
Some questions about the implementation

internal/runtime/logging/windows_handler.go
ptodev: Yes, I think you're right that it's unnecessary. I'll delete it.

internal/runtime/logging/logger.go
ptodev: I think we'd be idiomatically ok if newWindowsEventLogHandler also returns an `error`? I pushed this change now.

I like the idea of checking both `eventLogOpener` and `newWindowsEventLogHandler` for errors since I don't want to assume that `newWindowsEventLogHandler`  can only fail if `eventLogOpener` is nil. And I don't want to pass `eventLogOpener` to `newWindowsEventLogHandler`.

internal/runtime/logging/options.go
ptodev: I'd added this function because I wanted to make sure Event Log is the destination of choice when Alloy is ran as a service. I also wanted to make sure that if the user explicitly sets a destination, the user destination is used instead of the default one. I wasn't sure if `SetToDefault` alone would be able to do this, so I added it to a function. This is also why in `Options`, `Destination` is a pointer to `LogDestination`.

I updated the code just now to remove this function and the simplified code passes the newly added unit tests. So I think it's ok just to have a smarter `SetToDefault`.

internal/runtime/logging/fanout_handler.go
ptodev: Sure, I added a `fanout_handler_test.go` file.

internal/runtime/logging/logger.go
ptodev: True. I moved it to Update.

internal/runtime/logging/logger.go
kgeckhart: Yeah both returning an error here is more helpful. Even if the only error it can return is when el is nil that's totally fine.

internal/runtime/logging/logger.go
```diff
@@ -125,7 +143,46 @@ func (l *Logger) Update(o Options) error {
 	l.level.Set(slogLevel(o.Level).Level())
 	l.format.Set(o.Format)
 
-	l.writer.SetInnerWriter(l.inner)
+	// Close any existing Windows Event Log handler; we'll open a fresh one
+	// below if the destination still calls for it.
+	if l.windowsEventLogHandler != nil {
+		_ = l.windowsEventLogHandler.Close()
```
kgeckhart: What's going to happen if we error while starting the new one after just closing this one? Is logging going to be completely broken? 

Could we capture the current value and close it only after we have successfully swapped?

ptodev: > What's going to happen if we error while starting the new one after just closing this one? Is logging going to be completely broken?

No, only the event logger will not work. The write_to will still get the log messages. But the event logger will return an error.

> Could we capture the current value and close it only after we have successfully swapped?

It's more complex and honestly I'm not sure it's worth it. These loggers won't be changed frequently.

What bothers me more is that `write_to` misses a lot of log messages. I've been testing this locally and on startup it misses more than half of the first few log lines due to lack of buffers. We should probably spend more effort on fixing that instead, in a subsequent PR. It's not related to this change. It's missing log lines on the main branch too.

kgeckhart (dismissed):
Nice work!

internal/runtime/logging/logger.go
```diff
@@ -256,9 +324,10 @@ func (f *formatVar) Set(format Format) {
 type writerVar struct {
 	mut sync.RWMutex
 
-	lokiWriter  *lokiWriter
-	innerWriter io.Writer
-	tmpWriter   io.Writer
+	lokiWriter    *lokiWriter
+	innerWriter   io.Writer
+	tmpWriter     io.Writer
+	suppressInner bool // when true, Write skips innerWriter
```
kalleep: Cant we just set innerWriter to nil and do a check like we do for lokiWriter and tmpWriter?

internal/runtime/logging/windows_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	defer h.state.mu.RUnlock()
+	return h.state.el != nil
+}
+
+// Enabled reports whether the handler dispatches records at the given level.
+// Returns false when the handle has been closed.
+func (h *windowsEventLogHandler) Enabled(_ context.Context, level slog.Level) bool {
+	h.state.mu.RLock()
+	defer h.state.mu.RUnlock()
+	return h.state.el != nil && level >= h.level.Level()
```
kalleep: Not sure we need to do a lock and check state here. leveler is safe to call concurrently and we also do this exact check in Handle and abort, so if enabled we do 2 read locks on every log

internal/runtime/logging/windows_handler.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Handle dispatches the record to the Windows Event Log. Holds the state
+// RLock for the entire call, so a concurrent Close blocks until this call
+// returns — no risk of calling Info/Warning/Error on a closed handle.
+func (h *windowsEventLogHandler) Handle(_ context.Context, r slog.Record) error {
```
kalleep: on none windows system we always fan-out too this handler performing unnecessary check that will never be true. Could we handle that?

kalleep (commented):
Looks OK to me but there are a couple of things I am concerned about mostly on non windows we will do extra locking + checks. We also always call With* functions on both bytes and windows handler always regardless of what is used.

internal/runtime/logging/logger.go
ptodev: You can, but then you'd have to remember the original write buffer in case you want to re-enable it in the future. Since I deleted `inner` from `Logger`, it is now saved in `innerWriter` and I can't set it to nil.

internal/runtime/logging/handler.go
```diff
@@ -21,8 +22,13 @@ import (
 // expected to behave. Handlers will look up whether they should be logging as
 // JSON or logfmt, and create a new inner handler if needed.
 
-type handler struct {
-	w         io.Writer
+type bytesHandler struct {
```
kalleep: nit: not sure this rename should be done now, wdyt?

internal/runtime/logging/handler.go
```diff
@@ -44,77 +50,129 @@ type formatter interface {
...
+		return nil
 	}
-	h.mut.RUnlock()
 
-	// Slow path: we need to build a new handler.
-	h.mut.Lock()
-	defer h.mut.Unlock()
+	buf := bytesPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	defer bytesPool.Put(buf)
```
kalleep: nit: I prefer to make sure the state is correct when fetching from pools. I.e. Reset should be called before we put it back

internal/runtime/logging/handler.go
```diff
@@ -44,77 +50,129 @@ type formatter interface {
...
+	}
+	if !hasEventLog {
+		// Stderr happy path: the cached slog handler writes directly into
+		// writerVar via the io.Writer interface, no buffer, no per-call
+		// rebuild.
+		return h.buildHandler().Handle(ctx, r)
+	}
+	// Event log attached: format once into a buffer so we can hand both
+	// the bytes and the record level to writerVar.WriteRecord.
+	return h.handleWithEventLog(ctx, r)
```
kalleep: I am a bit confused why we need two different handle, Sure we use io.Write for the other ones but could we not just reuse the same function for all of them and handle it in the writer? It's a bit confusing that both writer and the handle need to know the state

internal/runtime/logging/handler.go
ptodev: Yes, I've been meaning to rename this back to `handler` during a final cleanup.

internal/runtime/logging/handler.go
ptodev: The problem is that the handler we return is just a wrapper around either `slog.NewTextHandler` or `slog.NewJSONHandler`, both of which simply write to a `io.Writer`. 

If we want to keep wrapping the slog wrappers then we have to write to a temporary buffer, read it, then write the same string to actual windows event log.

internal/runtime/logging/handler.go
```diff
@@ -44,77 +50,129 @@ type formatter interface {
...
-	defer h.mut.Unlock()
+	buf := bytesPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	defer bytesPool.Put(buf)
+
+	if err := h.newSlogHandler(buf).Handle(ctx, r); err != nil {
+		return err
+	}
+	if buf.Len() == 0 {
+		return nil // belt-and-suspenders for handlers that drop records silently
```
kalleep: what does this comment even mean?

internal/runtime/logging/handler.go
```diff
@@ -44,77 +50,129 @@ type formatter interface {
...
+// handleWithEventLog formats the record into a per-call buffer using a
+// freshly-built slog handler, then dispatches via writerVar.WriteRecord
+// which knows about the event log and the record level. Only used when
+// the event log destination is active.
+func (h *bytesHandler) handleWithEventLog(ctx context.Context, r slog.Record) error {
+	// Cheap level filter before we allocate the buffer and build the
+	// per-call slog handler. The inner slog handler filters too, but doing
+	// it here also skips the event-log dispatch (which would otherwise
+	// receive an empty message).
+	if !h.Enabled(ctx, r.Level) {
```
kalleep: Do we really need this check here? I am pretty sure slog.Logger will not call into this if it's not enabled

internal/runtime/logging/handler.go
ptodev: I reworded this comment so that it's more clear.

internal/runtime/logging/handler.go
ptodev: True, we don't need the check. I removed it.

ptodev:
I profiled the code using a mock component which logs 200 times per second. Then I looked at how much CPU time `log/slog.(*Logger).Info` takes per hour:
* On the `main` branch it took 23 seconds.
* On this branch, with `destination = "stderr"` it took 30 seconds.
* On this branch, with `destination = "windows_event_log"` it took 43.5 seconds.

I don't like how this branch reports slower time than `main` for `stderr`, but it's probably not a real issue. Some of the increase in time is not due to my code - e.g. `runtime.Callers` going from 7.4s to 9.4s. This happens even with `AddSource` set to `false`. It is a [known problem in Go](https://github.com/rs/zerolog/issues/571#issuecomment-1697479194) and I don't think my branch affects anything about how this function is called.

I could do Go benchmarks to calculate this more accurately, but I don't think it's really necessary.

internal/runtime/logging/handler.go
```diff
@@ -32,6 +37,17 @@ type handler struct {
 	currentFormat Format
 	inner         slog.Handler
 	replacer      func(groups []string, a slog.Attr) slog.Attr
+
+	eventLogPool sync.Pool
```
kalleep: Where do we set this one? I cant seem to find it

internal/runtime/logging/handler.go
```diff
@@ -52,59 +68,115 @@ func (h *handler) Enabled(ctx context.Context, l slog.Level) bool {
...
+	return h.w.Dispatch(s.buf.Bytes(), &r.Level)
+}
 
-	var newHandler slog.Handler
+// getEventLogScratch returns scratch space whose cached handler matches
+// the current format, rebuilding the handler only when the format
+// changed (or on first use). The returned buffer is already empty.
+func (h *handler) getEventLogScratch() *eventLogScratch {
+	expectFormat := h.formatter.Format()
+	s, _ := h.eventLogPool.Get().(*eventLogScratch)
```
kalleep: We should never be able to get a nil eventLogScratch

internal/runtime/logging/options.go
```diff
@@ -12,23 +12,58 @@ import (
...
 
 var _ syntax.Defaulter = (*Options)(nil)
 
-// SetToDefault implements syntax.Defaulter.
+// SetToDefault implements syntax.Defaulter. Destination is re-evaluated so
+// tests that stub isWindowsService after package init still observe the
+// expected default.
 func (o *Options) SetToDefault() {
 	*o = DefaultOptions
+	o.Destination = defaultDestination()
```
kalleep: We don't need this call right? We already store this is DefaultOptions

internal/runtime/logging/logger.go
```diff
@@ -287,32 +312,131 @@ func (w *writerVar) SetLokiWriter(receivers []loki.LogsReceiver) {
...
+		w.tmpWriter != nil
+	return
+}
 
-	if w.innerWriter == nil {
-		return 0, fmt.Errorf("no writer available")
-	}
+// HasSink reports whether any active sink will accept bytes. Kept for
+// callers that don't need to distinguish the event-log case.
+func (w *writerVar) HasSink() bool {
```
kalleep: I don't think we call this function

internal/runtime/logging/handler.go
ptodev: I updated the code to use a New() function

ptodev:
I ran a few quick profiles after the latest refactoring and now I for a 30 min run time I get this much time spent on `slog.Info()`:
* 13 seconds on the `main` branch
* 18 seconds for this branch. It's the same amount of time regardless whether we use event log or stderr.

`main`:
Screenshot 2026-06-05 at 17 33 57

This branch with stderr:
Screenshot 2026-06-05 at 17 34 10

This branch with event log:
Screenshot 2026-06-05 at 17 34 04

kalleep (dismissed):

kalleep (approved):

Issue #5602, opened 2026-02-19
bufio.Scanner: token too long


angelcarballeda:
### Component(s)

loki.source.docker

### Add a description

We have deployed a Docker container with the Alloy agent to collect logs from the rest of the Docker containers using the docker.discovery and loki.source.docker.nebula modules.
When we change our application logs to TRACE level, we get the following error:

{"ts":"2026-02-11T13:00:06.824884944Z","level":"error","msg":"error reading docker log line","component_path":"/","component_id":"loki.source.docker.nebula","component":"tailer","container":"docker/a6e0613d3cfec8606e2cb8025e7186d6e9dc64494b3396a66bf9e4caa7ada97f","err":"bufio.Scanner: token too long"}

What is the maximum number of characters for this module? Could it be 100,000?

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

jerbob92:
I have the same issue. When this happens it completely stops reading the container logs of that specific container.
According to the code, it seems to be 1048576 bytes.

kalleep:
Hey, yes this is it limit we have for each line, do you actually have logs that are bigger than this size? 

There was some changes recently to loki.source.docker so I am wondering if some regression was introduced there or if you actually have logs that are bigger than this size.

jerbob92:
@kalleep Yeah we do have lines longer that 1MB. We are working on removing them, they should not be there in the first place.

I get that the limit should be there, the main issue is that the whole log ingestion stops working for that single container after it has seen a line that's over 1MB in size. The only fix is recreating the container after that.

kalleep:
Great, I will do a fix where we log an error but don't exit when this happens so ingestion don't stop.

PR #5648, opened 2026-02-25
fix(loki.source.docker): Skip lines that are too big to consume

kalleep:
### Pull Request Details
Before scanner stream process jobs would exit whenever we tried to consume lines that are too big.

There is not really a good way to restore a scanner state when this happens so I had to do a workaround.

1. Wrap our reader with bufio.Reader.
2. When we hit ErrToLong from scanner we skip until next newline (without allocating temporary buffer).
3. Create new scanner and continue.

Fixes: #5602

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/tailer_test.go
+++ b/internal/component/loki/source/docker/tailer_test.go
@@ -271,7 +271,7 @@ func TestTailerConsumeLines(t *testing.T) {
 		require.Equal(t, expectedTimestamp, entry.Timestamp)
 	})
 
-	t.Run("bigger than mix size", func(t *testing.T) {
+	t.Run("bigger than max size", func(t *testing.T) {
 		collector := loki.NewCollectingHandler()
 		tailer := &tailer{
 			logger:            log.NewJSONLogger(os.Stdout),


internal/component/loki/source/docker/tailer.go
```diff
@@ -258,17 +272,42 @@ func extractTsFromBytes(line []byte) (time.Time, []byte, error) {
 }
 
 func (t *tailer) process(r io.Reader, logStreamLset model.LabelSet) {
-	defer t.wg.Done()
```
blewis12, line 261: I think removing this interferes with the [wg setup](https://github.com/grafana/alloy/blob/26cbc4d392e5298c813bbc89166fe195b7fa2657/internal/component/loki/source/docker/tailer.go#L154) done before the goroutines are launched no?

internal/component/loki/source/docker/tailer.go
```diff
@@ -258,17 +272,42 @@ func extractTsFromBytes(line []byte) (time.Time, []byte, error) {
...
+	for {
+		if !scanner.Scan() {
+			err := scanner.Err()
+			// We got EOF and should stop scanning.
+			if err == nil {
+				return
+			}
+
+			if errors.Is(err, bufio.ErrTooLong) {
+				level.Error(t.logger).Log("msg", "line to big skipping")
```
blewis12:
```suggestion
				level.Error(t.logger).Log("msg", "line too big, skipping")
```

internal/component/loki/source/docker/tailer.go
kalleep: Yes I added it back.

kalleep:
flog used for test actually had a line that was too big, preciously we hit this exact error and only consumed 5 lines even though the file had 1k lines so the test passed. When I fixed it the test broke... 

So I only keept the five lines we validate in the test file

internal/component/loki/source/docker/tailer.go
```diff
@@ -325,6 +365,16 @@ func (t *tailer) getStreamLabels(logStream string) model.LabelSet {
 	return filtered
 }
 
+func skipUntilNewline(r *bufio.Reader) error {
+	for {
+		_, err := r.ReadSlice('\n')
```
blewis12, line 370: Trying to wrap my head around the layers of buffers data is being read and passed to  - does the `bufio.Reader` track the last position consumed by the scanner? I'm wondering if there's an edge case where the `bufio.Reader` has advanced its position to beyond the newline, but the scanner has only read up until before that newline before it throws an `errTooLong`. Then we attempt to find the next newline in the `bufio.Reader` buffer, but we'd essentially skip a line in that case

kalleep: Yes bufio.Reader knows where next read is at, as soon as it have been consumed it's discarded. We re-create scanner when this happen and we have already moved passed the long line at this point. So next call to bufio.Reader will be after newline

blewis12: ahh ok

PR #5652, opened 2026-02-25
feat(loki.write): Add metric to track request size

kalleep:
BREAKING-CHANGE: removed `loki_write_encoded_bytes_total`, `loki_write_mutated_entries_total` and `loki_write_mutated_bytes_total`
* The first one is not used in any official dashboard and the same data can be observed using `loki_write_request_size_bytes`.
* The last two was undocumented and used in code.

### Pull Request Details
Add histogram metric for request size, this is useful when tweaking `batch_size` and `batch_wait`.

Removed the "mutated" metrics, these are undocumented and unused. 
Removed `loki_write_stream_lag_seconds`, this one was documented but don't exists.
Removed `loki_write_encoded_bytes_total`, this is now covered the new histogram added.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/client/endpoint_test.go
+++ b/internal/component/common/loki/client/endpoint_test.go
@@ -60,6 +60,7 @@ func TestEndpoint(t *testing.T) {
                                # TYPE loki_write_dropped_entries_total counter
                                loki_write_dropped_entries_total{host="__HOST__",reason="ingester_error",tenant=""} 0
                                loki_write_dropped_entries_total{host="__HOST__",reason="line_too_long",tenant=""} 0
+							   loki_write_dropped_entries_total{host="__HOST__",reason="queue_is_full",tenant=""} 0
                                loki_write_dropped_entries_total{host="__HOST__",reason="rate_limited",tenant=""} 0
                                loki_write_dropped_entries_total{host="__HOST__",reason="stream_limited",tenant=""} 0
                        `,
@@ -91,6 +92,7 @@ func TestEndpoint(t *testing.T) {
                               # TYPE loki_write_dropped_entries_total counter
                               loki_write_dropped_entries_total{host="__HOST__",reason="ingester_error",tenant=""} 0
                               loki_write_dropped_entries_total{host="__HOST__",reason="line_too_long",tenant=""} 0
+							  loki_write_dropped_entries_total{host="__HOST__",reason="queue_is_full",tenant=""} 0
                               loki_write_dropped_entries_total{host="__HOST__",reason="rate_limited",tenant=""} 0
                               loki_write_dropped_entries_total{host="__HOST__",reason="stream_limited",tenant=""} 0
                        `,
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/loki/loki.write.md
```diff
@@ -196,14 +196,14 @@ The following fields are exported and can be referenced by other components:
...
+* `loki_write_encoded_bytes_total` (counter): Number of bytes encoded and ready to send.
 * `loki_write_batch_retries_total` (counter): Number of times batches have had to be retried.
 * `loki_write_dropped_bytes_total` (counter): Number of bytes dropped because failed to be sent to the ingester after all retries.
 * `loki_write_dropped_entries_total` (counter): Number of log entries dropped because they failed to be sent to the ingester after all retries.
-* `loki_write_encoded_bytes_total` (counter): Number of bytes encoded and ready to send.
-* `loki_write_request_duration_seconds` (histogram): Duration of sent requests.
 * `loki_write_sent_bytes_total` (counter): Number of bytes sent.
 * `loki_write_sent_entries_total` (counter): Number of log entries sent to the ingester.
-* `loki_write_stream_lag_seconds` (gauge): Difference between current time and last batch timestamp for successful sends.
+* `loki_write_request_size_bytes` (histogram): Number of bytes for requests.
```
```suggestion
* `loki_write_request_size_bytes` (histogram): Number of bytes for encoded requests.
```

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -196,14 +196,14 @@ The following fields are exported and can be referenced by other components:
 
 ## Debug metrics
 
+* `loki_write_encoded_bytes_total` (counter): Number of bytes encoded and ready to send.
```
thampiotr: Isn't this now redundant metric? It can be replaced by the histogram sum.

kalleep: Yes that is true. Would it be a breaking change to remove it?

clayton-cornell:
Just a thought... do we want to "disappear" the `loki.write` metrics from the docs? Or note somewhere (in addition to the readme) on the topic page that they were removed?

kalleep:
> Just a thought... do we want to "disappear" the `loki.write` metrics from the docs? Or note somewhere (in addition to the readme) on the topic page that they were removed?

Do you mean something that mentions that we removed these in a specific version?

kalleep:
Also just to be clear:
* loki_write_stream_lag_seconds this one never existed so it would not be possible to use that metric
* loki_write_mutated_bytes_total and loki_write_mutated_entries_total was not documented and was not used in code at all

So the only one that we actually removed was loki_write_encoded_bytes_total

clayton-cornell:
I'm just thinking of whichever ones we exposed in the documentation. If they were never used/existed/documented, then there's nothing to do.

For something that was exposed in the docs, it'd be useful... maybe?... to have a trail so to speak and add:

```
{{< admonition type="warning" >}}
`loki_write_encoded_bytes_total` is removed.
{{< /admonition >}}
```

Or we just leave it alone and take it out of the docs. Not trying to complicate things here. If it was more "internal" and not something a user needs to know about beyond a mention in the changelog... then ignore me :-)

PR #5656, opened 2026-02-25
feat: Add fanout latency metric for loki pipelines

kalleep:
### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -2056,6 +2056,8 @@ The following fields are exported and can be referenced by other components:
 
 ## Debug metrics
 
+
+* `loki_fanout_latency` (histogram): Write latency for sending to components.
 * `loki_process_dropped_lines_total` (counter): Number of lines dropped as part of a processing stage.
 * `loki_process_dropped_lines_by_label_total` (counter):  Number of lines dropped when `by_label_name` is non-empty in [stage.limit][].
 * `loki_process_truncated_fields_total` (counter): Number of lines, label values, extracted field values, and structured metadata values truncated as part of a `truncate` stage.
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -2056,7 +2056,6 @@ The following fields are exported and can be referenced by other components:
 
 ## Debug metrics
 
-
 * `loki_fanout_latency` (histogram): Write latency for sending to components.
 * `loki_process_dropped_lines_total` (counter): Number of lines dropped as part of a processing stage.
 * `loki_process_dropped_lines_by_label_total` (counter):  Number of lines dropped when `by_label_name` is non-empty in [stage.limit][].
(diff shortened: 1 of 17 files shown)


internal/component/common/loki/fanout.go
```diff
@@ -4,12 +4,23 @@ import (
...
 )
 
 // NewFanout creates a new Fanout that will send log entries to the provided
 // list of LogsReceivers.
-func NewFanout(children []LogsReceiver) *Fanout {
+func NewFanout(children []LogsReceiver, register prometheus.Registerer) *Fanout {
+	wl := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    "loki_fanout_latency",
+		Help:    "Write latency for sending to components",
+		Buckets: []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 30, 60},
```
thampiotr: Do we need all these buckets? I think we can do with fewer and still have a good idea of 'something is wrong' or 'everything looks fine'.

internal/component/loki/process/process.go
```diff
@@ -73,7 +73,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		processOut:         loki.NewLogsReceiver(),
 		processIn:          loki.NewLogsReceiver(),
 		receiver:           loki.NewLogsReceiver(loki.WithComponentID(o.ID)),
-		fanout:             loki.NewFanout(args.ForwardTo),
+		fanout:             loki.NewFanout(args.ForwardTo, o.Registerer),
```
thampiotr, line 76: Is this registerer scoped to the component?

thampiotr (commented):
Tests are missing.

internal/component/loki/process/process.go
kalleep: [Yes it is](https://github.com/grafana/alloy/blob/main/internal/runtime/internal/controller/node_builtin_component.go#L188)!

internal/component/common/loki/fanout.go
kalleep: Absolutely, I just used the same bucktes we have for prometheus and pyroscope pipelines

kalleep:
>Tests are missing.

Well it's a draft, also need to update docs for each component using fanout to include this metric

internal/component/common/loki/fanout.go
kalleep: I changed buckets in https://github.com/grafana/alloy/pull/5656/changes/f62e04f6d02f0afc94762eac21e688efc693b42d.

These would be able to measure potential throughput at 5k/s to 1/s but I am happy to adjust these more if you think we should do it

kalleep:
Btw right now I just fixed existing tests that check metrics to ignore these, do you think we should have test for the metric itself?

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -2056,6 +2056,8 @@ The following fields are exported and can be referenced by other components:
 
 ## Debug metrics
 
+
+* `loki_fanout_latency` (histogram): Write latency for sending to components.
```
```suggestion
* `loki_fanout_latency` (histogram): Write latency for sending to components.
```

Remove extra blank line

clayton-cornell:
Docs changes are OK for now. Are there more changes incoming for docs?

kalleep:
Closing in favor of https://github.com/grafana/alloy/pull/5702

PR #5663, opened 2026-02-26
feat(loki.secretfilter): Add sampling for secretfilter entries

mikefat:
This PR introduces the argument `rate` for loki.secretfilter to sample which log entries are processed by the secret filter; non-sampled entries are forwarded unchanged.

### Pull Request Details

- **`rate`** (optional, `float`, default `1.0`): Sampling rate in `[0.0, 1.0]`. Fraction of entries that are run through detection/redaction; the rest are passed through unmodified. `1.0` = process all (current behavior).
- Validation: `rate` must be in `[0.0, 1.0]`
- Sampling uses the same boundary + integer RNG approach as the `loki.process` sampling stage for consistency.
- New metric: **`loki_secretfilter_entries_bypassed_total`** — count of entries forwarded without processing due to sampling.
- UT refactor: Due to inconsistencies in how the gitleaks handles config reloads (it uses global variables to track its state with itself and within `viper`), I've refactored the unit tests so that we can test consistently the cases with custom config file versus default config. This was a long tangent, but the gist is that multiple reloads of the config or extensions to the default config were causing inconsistencies in the tests due to gitleaks using global variables to track state.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.secretfilter.md
+++ b/docs/sources/reference/components/loki/loki.secretfilter.md
@@ -51,9 +51,9 @@ You can use the following arguments with `loki.secretfilter`:
 | `forward_to`         | `list(LogsReceiver)` | List of receivers to send log entries to.                                                                            |         | yes      |
 | `gitleaks_config`    | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                           | `""`    | no       |
 | `origin_label`       | `string`             | Loki label to use for the `secrets_redacted_by_origin` metric. If empty, that metric is not registered.              | `""`    | no       |
-| `rate`               | `float`              | Entry sampling rate in `[0.0, 1.0]` where `1` processes all entries. Unsampled entries are forwarded unchanged       | `1.0`   | no       |
-| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`. E.g.: `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
-| `redact_percent`     | `uint`               | When `redact_with` is not set: percent of the secret to redact (1–100), where 100 is full redaction                  | `80`    | no       |
+| `rate`               | `float`              | Entry sampling rate in `[0.0, 1.0]` where `1` processes all entries. Unsampled entries are forwarded unchanged.      | `1.0`   | no       |
+| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`, for example, `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
+| `redact_percent`     | `uint`               | When `redact_with` is not set: percent of the secret to redact (1–100), where 100 is full redaction.                 | `80`    | no       |
 | `drop_on_timeout`    | `bool`               | When true, drop entries that exceed `processing_timeout` instead of forwarding them unredacted.                      | `false` | no       |
 | `processing_timeout` | `duration`           | Maximum time allowed to process a single log entry. `0` disables the timeout.                                        | `0`     | no       |
 
@@ -77,8 +77,9 @@ For consistent behavior, use an external configuration file via `gitleaks_config
   When `redact_percent` is 0 or unset, 80% redaction is used.
 
 **Sampling:** The `rate` argument controls what fraction of log entries are processed by the secret filter.
-Entries that are not selected by the sampling rate are forwarded unchanged (no detection or redaction).
-Use a value below `1.0` (for example, `0.1` for 10%) to reduce CPU usage when processing high-volume logs; monitor `loki_secretfilter_entries_bypassed_total` to observe how many entries were skipped.
+Entries that {{< param "PRODUCT_NAME" >}} does not select based on the sampling rate pass through unchanged, with no detection or redaction applied.
+Use a value below `1.0`, for example, `0.1` for 10%, to reduce CPU usage when processing high-volume logs.
+Monitor `loki_secretfilter_entries_bypassed_total` to observe how many entries were skipped.
 
 **Origin metric:** The `origin_label` argument specifies which Loki label to use for the `secrets_redacted_by_origin` metric, so you can track how many secrets were redacted per source or environment.
 


kelnage:
Is this two PRs? One to refactor the unit tests, and a second to add the sampling feature?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -51,6 +51,7 @@ You can use the following arguments with `loki.secretfilter`:
 | `forward_to`      | `list(LogsReceiver)` | List of receivers to send log entries to.                                                                            |         | yes      |
 | `gitleaks_config` | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                           | `""`    | no       |
 | `origin_label`    | `string`             | Loki label to use for the `secrets_redacted_by_origin` metric. If empty, that metric is not registered.              | `""`    | no       |
+| `rate`            | `float`              | Entry sampling rate in `[0.0, 1.0]` where `1` processes all entries. Sampled entries are forwarded unchanged         | `1.0`   | no       |
```
kelnage: > Sampled entries are forwarded unchanged

I'd argue that the sampled entries are _processed_ by the secretfilter, the unsampled ones are forwarded unchanged (else the sampling rate should be defined the other way around as 1 would process **no** entries)

mikefat: Thanks, fixed

mikefat:
> Is this two PRs? One to refactor the unit tests, and a second to add the sampling feature?

Good point, thanks for flagging. I refactored the UT first since they could be inconsistent as I explained in the description, and I didn't want to add the sampling first since that would have made the refactor more annoying.
I'm happy to split it in two PRs if you prefer. I can open the refactor PR first and then open the sampling PR targeting the refactor branch

internal/component/loki/secretfilter/run_test_cases.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/grafana/alloy/internal/util"
+	"github.com/grafana/alloy/syntax"
+	"github.com/grafana/loki/pkg/push"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCase is a single test case for RunTestCases (name, log line, expected redaction).
+type TestCase struct {
```
kalleep, line 24: Nit: Nothing in this file need to be exported

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -52,16 +56,38 @@ type Exports struct {
...
 
 // SetToDefault implements syntax.Defaulter.
 func (args *Arguments) SetToDefault() {
 	*args = DefaultArguments
 }
 
-// defaultRedactPercent is the percentage used for gitleaks-style redaction when redact_with is not set and redact_percent is 0 or out of range.
-const defaultRedactPercent uint = 80
+// ErrInvalidRate is the error returned when rate is not in [0, 1].
+const ErrInvalidRate = "secretfilter rate must be between 0.0 and 1.0, received %f"
```
kalleep: Nit: No need to export this string, also I cannot see any usage of it so maybe just inline it in `fmt.Errorf`

internal/component/loki/secretfilter/run_test_cases.go
mikefat: Actually everything needs to be exported for the [extend_test.go](https://github.com/grafana/alloy/pull/5663/changes#diff-7dc38dc567f316534b46219872d6df827fef9e15f475e3bd70889627e490a12c) so that we can reuse the test data & test cases when running with an extended gitleaks config file

mikefat:
> Nice, since this shared a lot of similarities with `stage.sampling` it would be useful if the sampling logic was moved to some shared library, WDYT?

In order to keep the PR small and localized I will create a follow up issue for this and add it after we merge this PR 👍

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -49,12 +49,13 @@ You can use the following arguments with `loki.secretfilter`:
...
-| `drop_on_timeout`    | `bool`               | When true, drop entries that exceed `processing_timeout` instead of forwarding them unredacted.                      | `false` | no       |
 | `gitleaks_config`    | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                           | `""`    | no       |
 | `origin_label`       | `string`             | Loki label to use for the `secrets_redacted_by_origin` metric. If empty, that metric is not registered.              | `""`    | no       |
+| `rate`               | `float`              | Entry sampling rate in `[0.0, 1.0]` where `1` processes all entries. Unsampled entries are forwarded unchanged       | `1.0`   | no       |
+| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`. E.g.: `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
+| `redact_percent`     | `uint`               | When `redact_with` is not set: percent of the secret to redact (1–100), where 100 is full redaction                  | `80`    | no       |
```
```suggestion
| `rate`               | `float`              | Entry sampling rate in `[0.0, 1.0]` where `1` processes all entries. Unsampled entries are forwarded unchanged.      | `1.0`   | no       |
| `redact_with`        | `string`             | Template for the redaction placeholder. Use `$SECRET_NAME` and `$SECRET_HASH`, for example, `"<$SECRET_NAME:$SECRET_HASH>"` | `""`    | no       |
| `redact_percent`     | `uint`               | When `redact_with` is not set: percent of the secret to redact (1–100), where 100 is full redaction.                 | `80`    | no       |
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -75,6 +76,10 @@ For consistent behavior, use an external configuration file via `gitleaks_config
   `100` replaces the entire secret with `"REDACTED"`.
   When `redact_percent` is 0 or unset, 80% redaction is used.
 
+**Sampling:** The `rate` argument controls what fraction of log entries are processed by the secret filter.
+Entries that are not selected by the sampling rate are forwarded unchanged (no detection or redaction).
```
```suggestion
Entries that {{< param "PRODUCT_NAME" >}} does not select based on the sampling rate pass through unchanged, with no detection or redaction applied.
```

Alternatively we could also say: ``Entries that `loki.secretfilter` does not select based on the sampling rate pass through unchanged, with no detection or redaction applied.``

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -75,6 +76,10 @@ For consistent behavior, use an external configuration file via `gitleaks_config
...
+Use a value below `1.0` (for example, `0.1` for 10%) to reduce CPU usage when processing high-volume logs; monitor `loki_secretfilter_entries_bypassed_total` to observe how many entries were skipped.
```
```suggestion
Use a value below `1.0`, for example, `0.1` for 10%, to reduce CPU usage when processing high-volume logs.
Monitor `loki_secretfilter_entries_bypassed_total` to observe how many entries were skipped.
```

PR #5670, opened 2026-02-26
perf(discovery): Return full slice to pool after use

bboreham:

Without a closure, `defer`` captures the value as seen when that line was executed, whereas the value we want to put into the pool is the full-sized slice after appending to it. The closure achieves that effect.

Found using https://github.com/egonelbre/exp/tree/main/check-deferslice.

### DIFF
--- a/internal/component/discovery/target.go
+++ b/internal/component/discovery/target.go
@@ -292,7 +292,7 @@ func (t Target) SpecificLabelsHash(labelNames []string) uint64 {
 func (t Target) HashLabelsWithPredicate(pred func(key string) bool) uint64 {
 	// For hash to be deterministic, we need labels order to be deterministic too. Figure this out first.
 	labelsInOrder := borrowLabelsSlice()
-	defer releaseLabelsSlice(labelsInOrder)
+	defer func() { releaseLabelsSlice(labelsInOrder) }()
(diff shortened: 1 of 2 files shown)


kalleep (commented):
TestHashLabelsWithPredicateClearsStringSlicePool fails with this pr but I am not sure that test is entirely correct, or at least a bit confusing.

In that test we replace both `releaseLabelsSlice` and `borrowLabelsSlice` to make sure caller reset length before returning the slice. But the actual implementation do that [it self ](https://github.com/grafana/alloy/blob/main/internal/component/discovery/target.go#L36).

cc @thampiotr

bboreham:
I agree that test is odd.  I don't see any callers of `releaseLabelSlice` beside the two I changed in this PR, and those don't attempt to do what the test is expecting.  It only passes the test because of the bug I'm fixing.

kalleep:
>It only passes the test because of the bug I'm fixing

Exactly. So if we want to test for this we should reset length at call-site but I don't really see the point if the actual function already do it

bboreham:
OK I can add a commit removing that test, and the indirection mechanism.

thampiotr:
@kalleep I've removed the test and did minor refactor, ready for review

kalleep (approved):

PR #5672, opened 2026-02-26
chore: Trigger release artifact when tag is created

kgeckhart:

Switch publish artifacts to run when the tag for the release is created, which triggers at the same time as the draft release is created. `created` is not triggered on draft release,

image

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release

I also added a way to manually trigger it so I could get the artifacts for the current release.

### DIFF
--- a/.github/workflows/release-publish-alloy-artifacts.yml
+++ b/.github/workflows/release-publish-alloy-artifacts.yml
@@ -1,16 +1,16 @@
 name: Publish Alloy release artifacts
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*'
   workflow_dispatch:
     inputs:
       tag:
         description: 'Release tag to publish artifacts for (e.g., v1.14.0-rc.0)'
         required: true
         type: string
-
 env:
-  RELEASE_TAG: ${{ inputs.tag || github.event.release.tag_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-publish-alloy-artifacts.yml
```diff
@@ -1,7 +1,16 @@
...
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to publish artifacts for (e.g., v1.14.0-rc.0)'
+        required: true
+        type: string
+
+env:
+  RELEASE_TAG: ${{ inputs.tag || github.event.release.tag_name }}
```
jharvey10: Triggering off of tags should have the intended behavior. Release-please will create the draft release and the tag at the same time, so the release can stay in draft while the publish workflow runs. Once all artifacts are attached, the release can be published as usual.

```suggestion
name: Publish Alloy release artifacts
on:
  push:
    tags:
      - 'v*'
  workflow_dispatch:
    inputs:
      tag:
        description: 'Release tag to publish artifacts for (e.g., v1.14.0-rc.0)'
        required: true
        type: string

env:
  RELEASE_TAG: ${{ inputs.tag || github.ref_name }}
```

jharvey10 (approved):

PR #5673, opened 2026-02-27
ci: Force publish artifacts to be from a tag when triggered by dispatch

kgeckhart:

Ensure the dispatch of the artifacts job will be from a tag. Previously this was not required which could accidentally cause artifacts to be generated from main.

This also adds a dry run mode which prints the ref and exits.


jharvey10:
I think we can simplify this so that for the manual trigger, we just require the user to change the ref from which it's being run to the release tag.

PR #5678, opened 2026-02-27
feat(database_observability.mysql): Add filtering of query samples and wait events by minimum duration

cristiangreco:

In the `query_samples` collector, allow filtering of query samples and wait events based on a minimum duration.

Relates to https://github.com/grafana/grafana-dbo11y-app/issues/2355

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -152,6 +152,8 @@ The `azure` block supplies the identifying information for the database being mo
 | `disable_query_redaction`        | `bool`     | Collect unredacted SQL query text including parameters.                        | `false` | no       |
 | `auto_enable_setup_consumers`    | `boolean`  | Enables specific `performance_schema.setup_consumers` options. You must also enable `allow_update_performance_schema_settings`. | `false` | no       |
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
+| `sample_min_duration`            | `duration` | Minimum elapsed time for a query sample to be collected. Samples below this threshold are excluded. A value of `"0s"` disables filtering. | `"0s"`  | no       |
+| `wait_event_min_duration`              | `duration` | Minimum duration for a wait event to be collected. Wait events below this threshold are excluded. A value of `"0s"` disables filtering.   | `"1us"` | no       |
 
 ### `setup_actors`
 
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -152,7 +152,7 @@ The `azure` block supplies the identifying information for the database being mo
 | `disable_query_redaction`        | `bool`     | Collect unredacted SQL query text including parameters.                        | `false` | no       |
 | `auto_enable_setup_consumers`    | `boolean`  | Enables specific `performance_schema.setup_consumers` options. You must also enable `allow_update_performance_schema_settings`. | `false` | no       |
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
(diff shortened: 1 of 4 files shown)


docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -152,6 +152,8 @@ The `azure` block supplies the identifying information for the database being mo
 | `disable_query_redaction`        | `bool`     | Collect unredacted SQL query text including parameters.                        | `false` | no       |
 | `auto_enable_setup_consumers`    | `boolean`  | Enables specific `performance_schema.setup_consumers` options. You must also enable `allow_update_performance_schema_settings`. | `false` | no       |
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
+| `sample_min_duration`            | `duration` | Minimum elapsed time for a query sample to be collected. Samples below this threshold are excluded. A value of `"0s"` disables filtering. | `"0s"`  | no       |
+| `wait_event_min_duration`              | `duration` | Minimum duration for a wait event to be collected. Wait events below this threshold are excluded. A value of `"0s"` disables filtering.   | `"1us"` | no       |
```
cristiangreco: By default, only filter wait events with duration less than 1 microsecond. Do not filter samples.

cristiangreco: I feel we should also log a warn in case `wait_event_min_duration` is greater than `sample_min_duration`, which doesn't make much sense.

gaantunes: Should we use 1ms as the default for `wait_event_min_duration`? I feel like that is quite low yet and will filter much more.

cristiangreco: Hmm I didn't want to start with an aggressive value until we have tested this a bit more.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -152,6 +152,8 @@ The `azure` block supplies the identifying information for the database being mo
 | `disable_query_redaction`        | `bool`     | Collect unredacted SQL query text including parameters.                        | `false` | no       |
 | `auto_enable_setup_consumers`    | `boolean`  | Enables specific `performance_schema.setup_consumers` options. You must also enable `allow_update_performance_schema_settings`. | `false` | no       |
 | `setup_consumers_check_interval` | `duration` | How frequently to check if `setup_consumers` are correctly enabled.            | `"1h"`  | no       |
+| `sample_min_duration`            | `duration` | Minimum elapsed time for a query sample to be collected. Samples below this threshold are excluded. A value of `"0s"` disables filtering. | `"0s"`  | no       |
```
matthewnolf: I don't think we should allow 0s. Having 0s disable sampling gives 2 different ways to disable sampling which I don't think we need.

matthewnolf (commented):
Looks good

docs/sources/reference/components/database_observability/database_observability.mysql.md
cristiangreco: Hmm `0s` here disables filtering, not sampling. i.e. it means "don't filter, collect all samples regardless of their duration"

I guess the wording can be improved... but wording aside, wdyt?

clayton-cornell: You could make it more active...


```suggestion
| `sample_min_duration`            | `duration` | Minimum duration for query samples to be collected. Set to "0s" to disable filtering and collect all samples regardless of their duration.| `"0s"`  | no       |
```

You can drop the second sentence as it's redundant. The "minimum" part already tells the reader that anything below it won't be collected.

matthewnolf: Okay yeah that makes sense, sorry I got confused by the wording

matthewnolf (approved):

PR #5679, opened 2026-02-27
chore: Add test for alloy runtime to check that we start and stop components

kalleep:
Add test that verifies different component statues for alloy runtime, this is to increase confidence with upcoming scheduling changes https://github.com/grafana/alloy/pull/5613#issuecomment-3964960504

### DIFF
--- a/internal/runtime/alloy.go
+++ b/internal/runtime/alloy.go
@@ -289,7 +289,6 @@ func (f *Runtime) Run(ctx context.Context) {
 				components = f.loader.Components()
 				services   = f.loader.Services()
 				imports    = f.loader.Imports()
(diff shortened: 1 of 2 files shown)


internal/runtime/alloy.go
```diff
@@ -301,10 +300,6 @@ func (f *Runtime) Run(ctx context.Context) {
 				runnables = append(runnables, i)
 			}
 
-			for _, fe := range forEachs {
-				runnables = append(runnables, fe)
-			}
```
kalleep, line 306: This is not needed we already add all foreach nodes to components

kgeckhart: By this you mean, the foreach nodes are included in f.loader.Components()?

PR #5687, opened 2026-02-27
feat: Expose Functionality to Handle syslogs with Empty MSG Field

blewis12:
This relates to this escalation: https://github.com/grafana/support-escalations/issues/20753

Currently, in our `loki.source.syslog` component, we will drop logs with empty messages and increase our `loki_source_syslog_empty_messages_total` counter to indicate this. This has different ramifications for the different formats we support

### **RAW**
Our raw format basically parses the entire log line as the message, and so an empty message in this case should be treated as an anomaly and be dropped

### RFC3164
In this case, the log format is `<PRI> <HEADER> <MSG>`

For example `<13>Mar 13 12:01:33 firewall01 sshd: I am a log`. This would be parsed like:
- `PRI`: <13> 
- `HEADER.timestamp`: Mar 13 12:01:33 
- `HEADER.hostname`: firewall01
- `MSG.appname`: sshd 
- `MSG.message`: I am a log -> previously if this was empty, we would drop the log

In our parser (`leodido/go-syslog/v4`), omitting the "I am a log" part would result in a `message` set to `sshd`. That means that an empty message in this case would mean a log like `<13>Mar 13 12:01:33 firewall01`. This log line doesn't provide a lot of use, therefore I think it would make sense to continue dropping this, though ensure that it is clearly documented

It's also worth noting the the [RFC](https://datatracker.ietf.org/doc/html/rfc3164#section-5.4) doesn't provide any message-less examples 

### RFC5424
In this case, the log format is more granular `<PRI>VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID STRUCTURED-DATA [MSG]` 

For example `<13>1 2025-03-13T12:01:33.000Z firewall01 sshd 12345 msg-001 [example@12345 key="value"] I am a log`. This would be parsed like

- `PRI`: <13> 
- `VERSION`: 1
- `TIMESTAMP`: Mar 13 12:01:33 
- `HOSTNAME`: firewall01
- `APP-NAME`: sshd 
- `PROCID`: 12345
- `MSGID`: msg-001
- `STRUCTURED-DATA`: map[example@12345:map[key:value]]
- `MSG`: I am a log -> previously if this was empty, we would drop the log

In this case, an empty message does not mean an entirely useless log - the system can still be sending over structured data with contextual information, which can also be applied in relabelling/processing rules in later pipeline steps to output a useful log line. This is the same scenario the customer found themselves in, and was confused to see log lines "silently" being dropped

My proposal is that we introduce opt-in functionality to allow for empty message fields in `RFC5424` formatted logs.  It's worth noting that the RFC explicitly states that an empty MSG field is a valid log, see example 4 in https://www.rfc-editor.org/rfc/rfc5424#section-6.5

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.syslog.md
+++ b/docs/sources/reference/components/loki/loki.source.syslog.md
@@ -112,13 +112,14 @@ The `listener` block defines the listen address and protocol where the listener 
 The following arguments can be used to configure a `listener`.
 Only the `address` field is required and any omitted fields take their default values.
 
-| Name                              | Type          | Description                                                                            | Default     | Required |
-|-----------------------------------|---------------|----------------------------------------------------------------------------------------|-------------|----------|
-| `address`                         | `string`      | The `<host:port>` address to listen to for syslog messages.                            |             | yes      |
-| `rfc5424_allow_empty_msg`         | `bool`        | Whether to forward RFC5424 messages with empty MSG content. When `false`, such messages are dropped. Only applies when `syslog_format` is `rfc5424`. | `false`     | no       |
-| `idle_timeout`                    | `duration`    | The idle timeout for TCP connections.                                                  | `"120s"`    | no       |
-| `label_structured_data`           | `bool`        | Whether to translate syslog structured data to Loki labels.                            | `false`     | no       |
-| `labels`                          | `map(string)` | The labels to associate with each received syslog record.                              | `{}`        | no       |
+| Name                      | Type          | Description                                                                                                                                          | Default  | Required |
+|---------------------------|---------------|------------------------------------------------------------------------------------------------------------------------------------------------------|----------|----------|
+| `address`                 | `string`      | The `<host:port>` address to listen to for syslog messages.                                                                                          |          | yes      |
+| `allow_empty_rfc5424_msg` | `bool`        | Whether to forward RFC5424 messages with empty MSG content. When `false`, such messages are dropped. Only applies when `syslog_format` is `rfc5424`. | `false`  | no       |
+| `idle_timeout`            | `duration`    | The idle timeout for TCP connections.                                                                                                                | `"120s"` | no       |
+| `label_structured_data`   | `bool`        | Whether to translate syslog structured data to Loki labels.                                                                                          | `false`  | no       |
+| `labels`                  | `map(string)` | The labels to associate with each received syslog record.                                                                                            | `{}`     | no       |
+
 | `max_message_length`              | `int`         | The maximum limit to the length of syslog messages.                                    | `8192`      | no       |
 | `protocol`                        | `string`      | The protocol to listen to for syslog messages. Must be either `tcp` or `udp`.          | `"tcp"`     | no       |
 | `rfc3164_default_to_current_year` | `bool`        | Whether to default the incoming timestamp of an `rfc3164` message to the current year. | `false`     | no       |


internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget.go
```diff
@@ -283,6 +290,7 @@ func (t *SyslogTarget) handleMessageRFC3164(connLabels labels.Labels, msg *rfc31
 }
 
 func (t *SyslogTarget) handleMessageRaw(connLabels labels.Labels, msg *syslog.Base) {
+	// Drop raw logs with nil or empty message - this would mean the line itself is empty (e.g. "\n")
 	if msg.Message == nil || *msg.Message == "" {
 		t.metrics.syslogEmptyMessages.Inc()
 		return
```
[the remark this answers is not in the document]
blewis12: it's not changing behavior but I guess we should have a test for it

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -115,6 +115,7 @@ Only the `address` field is required and any omitted fields take their default v
 | Name                              | Type          | Description                                                                            | Default     | Required |
 |-----------------------------------|---------------|----------------------------------------------------------------------------------------|-------------|----------|
 | `address`                         | `string`      | The `<host:port>` address to listen to for syslog messages.                            |             | yes      |
+| `allow_empty_rfc5424_msg`         | `bool`        | Whether to forward RFC5424 messages with empty MSG content. When `false`, such messages are dropped. Only applies when `syslog_format` is `rfc5424`. | `false`     | no       |
```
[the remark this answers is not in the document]
blewis12: I think this just adds confusion, it can't be used either for `raw` or `rfc3164` - imo this should be clear enough from the name of the config itself and is also included in the validation logic

internal/component/loki/source/syslog/internal/syslogtarget/syslogtarget_test.go
```diff
@@ -567,6 +567,102 @@ func TestSyslogTarget_RFC5424Messages(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = writeMessagesToStream(c, []string{msg}, fmtNewline)
+	require.NoError(t, err)
+	require.NoError(t, c.Close())
+
+	// Give a moment for any async processing
+	time.Sleep(100 * time.Millisecond)
+
+	require.Empty(t, handler.Received(), "empty MSG should be rejected when AllowEmptyRFC5424Msg is false")
```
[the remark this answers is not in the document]
blewis12: I don't think this really makes sense, even if a message arrived after we wait - stopping the target wouldn't change that

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -147,12 +148,15 @@ The `rfc3164_default_to_current_year`, `use_incoming_timestamp` and `use_rfc5424
 * **`rfc3164`**
   A legacy syslog format, also known as BSD syslog.
   Example: `<34>Oct 11 22:14:15 my-server-01 sshd[1234]: Failed password for root from 192.168.1.10 port 22 ssh2`
+  Messages with empty MSG content are dropped and the `loki_source_syslog_empty_messages_total` counter is incremented.
```
clayton-cornell:
```suggestion
  `loki.source.syslog` drops messages with empty MSG content and increments the `loki_source_syslog_empty_messages_total` counter.
```

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -147,12 +148,15 @@ The `rfc3164_default_to_current_year`, `use_incoming_timestamp` and `use_rfc5424
...
 * **`rfc5424`**
   A modern, structured syslog format. Uses ISO 8601 for timestamps.
   Example: `<165>1 2025-12-18T00:33:00Z web01 nginx - - [audit@123 id="456"] Login failed`.
+  Messages with empty MSG content are dropped by default; set `allow_empty_rfc5424_msg` to `true` to forward them. The `loki_source_syslog_empty_messages_total` counter is incremented in both cases for debugging.
```
clayton-cornell:
```suggestion
  `loki.source.syslog` drops messages with empty MSG content by default.
  Set `allow_empty_rfc5424_msg` to `true` to forward them.
  `loki.source.syslog` increments the `loki_source_syslog_empty_messages_total` counter in both cases for debugging.
```

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -147,12 +148,15 @@ The `rfc3164_default_to_current_year`, `use_incoming_timestamp` and `use_rfc5424
...
 * **`raw`**
   Disables log line parsing. This format allows receiving non-RFC5424 compliant logs, such as [CEF][cef].
   Raw logs can be forwarded to [`loki.process`](./loki.process.md) component for parsing.
+  Messages with nil or empty body are dropped and the `loki_source_syslog_empty_messages_total` counter is incremented.
```
clayton-cornell:
```suggestion
  `loki.source.syslog` drops messages with nil or empty body and increments the `loki_source_syslog_empty_messages_total` counter.
```

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -115,6 +115,7 @@ Only the `address` field is required and any omitted fields take their default v
...
 |-----------------------------------|---------------|----------------------------------------------------------------------------------------|-------------|----------|
 | `address`                         | `string`      | The `<host:port>` address to listen to for syslog messages.                            |             | yes      |
+| `allow_empty_rfc5424_msg`         | `bool`        | Whether to forward RFC5424 messages with empty MSG content. When `false`, such messages are dropped. Only applies when `syslog_format` is `rfc5424`. | `false`     | no       |
 | `idle_timeout`                    | `duration`    | The idle timeout for TCP connections.                                                  | `"120s"`    | no       |
 | `label_structured_data`           | `bool`        | Whether to translate syslog structured data to Loki labels.                            | `false`     | no       |
 | `labels`                          | `map(string)` | The labels to associate with each received syslog record.                              | `{}`        | no       |
```
x1unix: Nitpick: can please reformat the markdown table width

```suggestion
| Name                      | Type          | Description                                                                                                                                          | Default  | Required |
|---------------------------|---------------|------------------------------------------------------------------------------------------------------------------------------------------------------|----------|----------|
| `address`                 | `string`      | The `<host:port>` address to listen to for syslog messages.                                                                                          |          | yes      |
| `allow_empty_rfc5424_msg` | `bool`        | Whether to forward RFC5424 messages with empty MSG content. When `false`, such messages are dropped. Only applies when `syslog_format` is `rfc5424`. | `false`  | no       |
| `idle_timeout`            | `duration`    | The idle timeout for TCP connections.                                                                                                                | `"120s"` | no       |
| `label_structured_data`   | `bool`        | Whether to translate syslog structured data to Loki labels.                                                                                          | `false`  | no       |
| `labels`                  | `map(string)` | The labels to associate with each received syslog record.                                                                                            | `{}`     | no       |

```

docs/sources/reference/components/loki/loki.source.syslog.md
```diff
@@ -147,12 +148,17 @@ The `rfc3164_default_to_current_year`, `use_incoming_timestamp` and `use_rfc5424
 * **`rfc3164`**
   A legacy syslog format, also known as BSD syslog.
   Example: `<34>Oct 11 22:14:15 my-server-01 sshd[1234]: Failed password for root from 192.168.1.10 port 22 ssh2`
```
x1unix: Added a dot to separate the next new sentence.

```suggestion
  Example: `<34>Oct 11 22:14:15 my-server-01 sshd[1234]: Failed password for root from 192.168.1.10 port 22 ssh2`.
```

internal/component/loki/source/syslog/config/config.go
```diff
@@ -126,6 +126,9 @@ type SyslogTargetConfig struct {
 	// When false, the year will default to 0.
 	RFC3164DefaultToCurrentYear bool `yaml:"rfc3164_default_to_current_year"`
 
+	// AllowEmptyRFC5424Msg when true, forwards RFC5424 messages with empty MSG content. Default false.
+	AllowEmptyRFC5424Msg bool `yaml:"allow_empty_rfc5424_msg"`
```
x1unix: I propose to use the format prefix for a config field, to be consistent with `rfc3164_default_to_current_year` and `rfc3164_cisco_components` fields.

```suggestion
	RFC5424AllowEmptyMsg bool `yaml:"rfc5424_allow_empty_msg"`
```

blewis12: ah yeah, good shout 👍

blewis12: https://github.com/grafana/alloy/pull/5687/changes/12894ee3a626ed8bdb9e587c4c4a810f9d8a47da

PR #5676, opened 2026-02-27
chore: Update Helm chart to Alloy v1.13.2

icelynjennings:

PR #5694, opened 2026-03-02
chore: Update helm for 1.13.2

prateekpandey14:


blewis12:
Hiya @prateekpandey14 - there's already a [PR up](https://github.com/grafana/alloy/pull/5676), do you mind if we close this one to avoid duplicates?

prateekpandey14:
Hi @blewis12 , I understand the concern about duplicates. This change is time-sensitive for us, so merging this PR would help in case the other one takes longer to push the required update. Thanks !

blewis12:
understood @prateekpandey14

by the way, you can point to any image/tag of alloy in the helm chart and override the default. For example, in the values.yaml:

```
image:
  tag: v1.13.2
```

If that helps at all with the time sensitivity factor

prateekpandey14:
@blewis12 can you please approve and  merge . Thanks

PR #5701, opened 2026-03-02
perf(prometheus.relabel): Reduce lock contention

kgeckhart:

Remove `cacheMut`: The LRU cache was protected by both `c.mut` (outer read/write lock) and a separate `cacheMut`. The outer lock already serializes pointer replacement in `clearCache`, and `hashicorp/golang-lru/v2` is internally thread-safe for concurrent reads/writes. `cacheMut` was redundant.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/relabel/relabel.go
+++ b/internal/component/prometheus/relabel/relabel.go
@@ -90,8 +90,7 @@ type Component struct {
 
 	debugDataPublisher livedebugging.DebugDataPublisher
 
-	cacheMut sync.RWMutex
-	cache    *lru.Cache[uint64, labels.Labels]
+	cache *lru.Cache[uint64, labels.Labels]
 }
 
 var (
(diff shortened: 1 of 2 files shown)


internal/component/prometheus/relabel/relabel.go
```diff
@@ -207,8 +229,15 @@ func New(o component.Options, args Arguments) (*Component, error) {
 				return 0, nil
 			}
 
-			// Since SeriesRefs are tied to the labels, we send zero to indicate the seriesRef should be recalculated downstream.
-			return next.AppendHistogram(0, newLbl, t, h, fh)
+			// With labelstore disabled the incoming ref is the downstream ref returned
```
kalleep: I guess there are plans to remove labelstore in favor of seriersrefstore in the future?

Could we add a FIXME/TODO that we should remove this when that happens?

kalleep (approved):

internal/component/prometheus/relabel/relabel.go
kgeckhart: I ended up backing this change out. It's more complex than expected to do properly.

PR #5702, opened 2026-03-03
feat(loki.write): Add loki pipeline latency metric

kalleep:
### Pull Request Details
Alternative to https://github.com/grafana/alloy/pull/5656.

In this pr I added `loki_write_entry_propagation_latency`. This is a histogram with the time it takes from from a source until it's written over the wire.

Every source now uses `loki.NewEntry` or `loki.NewEntryWithCreated`, and multiline will propagate created from first entry.

When the entry was either sent or dropped by loki write we record the time it took for all of them. I also added a new version for entries in WAL so that the created time is encoded.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/wal/encoding.go
+++ b/internal/component/common/loki/wal/encoding.go
@@ -207,7 +207,7 @@ func DecodeEntries(b []byte, version RecordType, rec *Record) error {
 		}
 
 		if dec.Err() != nil {
-			return fmt.Errorf("entry decode error after %d RefEntries: %w", n-rem, dec.Err())
+			return fmt.Errorf("entry decode error after decoding %d entries in current RefEntries: %w", n-rem, dec.Err())
 		}
 
 		rec.RefEntries = append(rec.RefEntries, refEntries)


internal/component/common/loki/wal/encoding.go
```diff
@@ -55,28 +80,6 @@ func (r *Record) Reset() {
 	}
 
 	r.RefEntries = r.RefEntries[:0]
-	r.entryIndexMap = make(map[uint64]int)
-}
-
-func (r *Record) AddEntries(fp uint64, counter int64, entries ...push.Entry) {
```
kgeckhart, line 61: Was this unused?

internal/component/common/loki/wal/encoding.go
```diff
@@ -26,22 +29,44 @@ const (
...
 
-// The current type of Entries that this distribution writes.
-// Loki can read in a backwards compatible manner, but will write the newest variant.
-const CurrentEntriesRec = WALRecordEntriesV3
+// The current type of Entries that WAL writes.
+const CurrentEntriesRec = WALRecordEntriesV4
+
+type RefEntries struct {
+	// Counter is unused.
+	Counter int64
```
kgeckhart, line 41: Is this true? It looks like it's used with `EncodeSeries` when the entry version >= 2

internal/component/common/loki/wal/encoding.go
```diff
@@ -26,22 +29,44 @@ const (
...
 type Record struct {
-	UserID string
-	Series []record.RefSeries
-
-	// entryIndexMap coordinates the RefEntries index associated with a particular fingerprint.
-	// This is helpful for constant time lookups during ingestion and is ignored when restoring
-	// from the WAL.
-	entryIndexMap map[uint64]int
-	RefEntries    []RefEntries
+	// UserID is unused.
```
kgeckhart, line 59: Is it kept for backwards compatibility reasons?

internal/component/common/loki/client/metrics.go
```diff
@@ -56,6 +57,11 @@ func newMetrics(reg prometheus.Registerer) *metrics {
 		MiB = 1024 * KiB
 	)
 
+	m.entryLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "loki_write_entry_propagation_latency",
+		Help:    "Write latency for entries",
+		Buckets: []float64{0.0002, 0.001, 0.005, 0.02, 0.1, 1.0},
```
kgeckhart: Is it possible for us to use a native histogram here instead? I feel like 0.002 might be a bit excessive on the low end for measurements.

internal/component/common/loki/wal/encoding.go
```diff
@@ -26,22 +29,44 @@ const (
...
+const CurrentEntriesRec = WALRecordEntriesV4
+
+type RefEntries struct {
+	// Counter is unused.
+	Counter int64
+	// Ref identifies the series these entries belong to.
+	Ref chunks.HeadSeriesRef
+	// Entries are log entries belonging to the same series.
+	Entries []push.Entry
+	// Created stores entry creation timestamps in Unix nanoseconds, aligned by index with Entries.
```
kgeckhart: Do we need to have nanosecond precision here? It seems like milliseconds would be acceptable.

internal/component/common/loki/entry.go
```diff
@@ -1,21 +1,44 @@
...
+}
+
+// Entry is a push.Entry with labels.
+// It should be created using either NewEntry or NewEntryWithCreated.
 type Entry struct {
 	Labels model.LabelSet
 	push.Entry
+
+	// FIXME check if we should use unix nano instead..
+	created time.Time
```
kgeckhart: I think we might want to do this? Probably depends on the overhead of a time.Time struct vs the int64 though. See prior comment regarding the need for nanosecond precision.

kgeckhart (commented):
Nice work, I did have some questions.

1. Do we have a sense of the overhead this adds for WAL/in general? 
2. If we are trending towards a batched based Loki pipeline solution (I haven't seen much in the way of updates on this to know what direction we are heading) is this entry level metric still going to make sense?

internal/component/common/loki/wal/encoding.go
kalleep: Yeah we don't use this in Alloy, in loki this would be another story. Keep it here because we need to write it to be backwards compatible. Same for UserID.

internal/component/common/loki/wal/encoding.go
kalleep: Yes

internal/component/common/loki/client/metrics.go
kalleep: Is there a way to support both or how does native histogram work?

For the buckets I am happy to adjust them, 0.002 would give us a resolution of 5k entries/sec. It does not seem unreasonable.

internal/component/common/loki/wal/encoding.go
kalleep: With current buckets we cannot use milliseconds but we can use micro seconds

internal/component/common/loki/entry.go
kalleep: Yeah so time.Time is 24 bytes so using a unix timestamp (mili,micro or nano) would save bytes per entry. 

But if we do this we loose the monotonic clock reading. So if any adjustments is done to the system clock while in flight we can get inaccurate recordings. But we need to serialize this time when WAL is used so we will always have this problem there.

internal/component/common/loki/wal/encoding.go
kalleep: Yes in alloy this is unused. Also I realized that we always encode one series and one entry [for every entry we put in WAL](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/wal/writer.go#L257-L278).

This is not that efficient but with batching we would make this better

kalleep:
>Do we have a sense of the overhead this adds for WAL/in general?

Yes it would add 8 additional bytes in WAL for every entry.

>If we are trending towards a batched based Loki pipeline solution (I haven't seen much in the way of updates on this to know what direction we are heading) is this entry level metric still going to make sense

Good point, will have to think a bit about this one but I think it would be acceptable to have one created timestamp per batch. If that is the case we could store one Created timestamp for a `RefEntries` instead. It would also work currently because in alloy `RefEntries` only contain one entry at the moment. WDYT?

The overhead in WAL would be less with batches then because we would store 8 bytes for the entire batch instead of 8 bytes per entry in batch.

And yes we are likely going to batches, we are working on the design doc just going a bit slow

internal/component/common/loki/client/metrics.go
kgeckhart: >  0.002 would give us a resolution of 5k entries/sec. It does not seem unreasonable.

I thought rate would be based off the total count without buckets playing a factor, what am I missing? 

On the whole you're well under the normal amount of buckets I'm used to seeing which is like 10. I don't think it's bad it just seems unnecessary to track below 1ms. I don't know what the typical distribution would look like here though.

>  Is there a way to support both or how does native histogram work?

Based on the docs it looks like you can define buckets as well as the native histogram options in the HistogramOpts and it will do both. Since they are newer it might be better because prometheus scraping can handle both, https://prometheus.io/docs/specs/native_histograms/#scraping-both-classic-and-native-histograms. Using native histograms should help with a scenario like this where we are unsure of the ideal bucket boundaries and it should adapt accordingly. Might take some reading / thought for the config values though https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L405-L480.

kgeckhart:
> Yes it would add 8 additional bytes in WAL for every entry.

I guess I was also thinking about general pipeline overhead but that's more nebulous / hard to gauge. It's an extra field on a struct so probably not important.

> Good point, will have to think a bit about this one but I think it would be acceptable to have one created timestamp per batch. If that is the case we could store one Created timestamp for a `RefEntries` instead. It would also work currently because in alloy `RefEntries` only contain one entry at the moment. WDYT?
> 

I think that makes a lot of sense!

> And yes we are likely going to batches, we are working on the design doc just going a bit slow

FTR my comment wasn't suggesting it's moving too slow just qualifying my question in case it was out of date 😅

internal/component/common/loki/client/metrics.go
kalleep: Yeah it's true that if we want to graph throughput using rate buckets would not affect it, I see you point :+1:

kalleep: For native histogram I guess supporting both would be best for now will read a bit about it

kalleep: Ah somehow forgot that this is the time it takes for entries from ingestion to sent over the wire, including retries.

Yeah I agree that `0.0002` is way to optimistic

I changed the buckets for classic histogram and added settings for native histograms

internal/component/common/loki/entry.go
kalleep: Okay I changed to storing created as UnixMicro, we could use UnixMili but for storage cost that won't matter.

kalleep:
We are now writing one created to WAL for each `RefEntries`, it works currently because we are only writing one entry in each RefEntry and when we change to batching RefEntries would contain all entries sharing labels from a batch so the created should be the same for them

clayton-cornell:
No input on docs content - one-line change

docs/sources/reference/components/loki/loki.write.md
```diff
@@ -203,6 +203,7 @@ The following fields are exported and can be referenced by other components:
 * `loki_write_sent_entries_total` (counter): Number of log entries sent to the ingester.
 * `loki_write_request_size_bytes` (histogram): Number of bytes for encoded requests.
 * `loki_write_request_duration_seconds` (histogram): Duration of sent requests.
+* `loki_write_entry_propagation_latency_seconds` (histogram): Time in seconds from entry creation until it's either successfully sent or dropped
```
```suggestion
* `loki_write_entry_propagation_latency_seconds` (histogram): Time in seconds from entry creation until it's either successfully sent or dropped.
```

internal/component/common/loki/client/batch.go
```diff
@@ -31,6 +32,8 @@ type SentDataMarkerHandler interface {
 // streams for each tenant are stored in a dedicated batch.
 type batch struct {
 	streams map[string]*push.Stream
+	// created stores per-entry creation timestamps in unix micro seconds for latency observation.
+	created []int64
 	// createdAt is when the batch was created.
 	createdAt time.Time
```
thampiotr, line 38: Thought we would be moving towards using only `createdAt` and then `Observe` N times for all N entries in the batch?

kalleep: Yes we can do this for every entry in a batch that is passed along the pipeline when we have it. This is the batch structure we append to before sending to loki. Because this one can and will contain entries from multiple different streams we for now need to store one created per entry.

When we move to batched pipelines we would store one created for each batch added here if that makes sense

thampiotr: ok, gotcha, this code is after WAL.

internal/component/loki/process/stages/multiline.go
```diff
@@ -188,26 +192,12 @@ func (m *multilineStage) flush(out chan Entry, s *multilineState) {
...
-			},
-		},
-	}
-	s.buffer.Reset()
-	s.currentLines = 0
 
-	out <- collapsed
+	entry := s.startLineEntry
+	entry.Line = s.buffer.String()
+
```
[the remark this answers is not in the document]
kalleep: This is safe, the state is local to a specific stream. When we detect the start of a multi-line we set startLineEntry to that entry and continue. 

When the line is finished we update this entry's line with everything we have written into the buffer and reset the state. In reset we clear buffer, set currentLines to 0 and set startLineEntry to it's zero value so there is no more references to extractex, structured metadata and labels.

The copy of these we have before was unnecessary and is just doing additional allocations

PR #5703, opened 2026-03-03
ci: Add alloy-cd trigger-argo-workflow Action Invoked on Alloy Release Event 

blewis12:
### Pull Request Details
This PR introduces a workflow that will trigger the `alloy-cd` workflow on Alloy release/publish - it will exit early if `helm-chart/...` or `build-image/...` tags are detected

### Notes to the Reviewer
I tested this out in my own test repo [here](https://github.com/blewis12/test-project/actions), by commenting out the action action invocation and just logging the resolved tag, that [seems to work](https://github.com/blewis12/test-project/actions/runs/22630548081/job/65579125876)

Links to specific tests:
* I added some debug steps [here](https://github.com/blewis12/test-project/actions/runs/22631870213/job/65583936262) that print what would be passed to the argo workflow in the `Print ArgoCD Parameters` step, to me this looks correct 👍 
* I also tested out the validation step [here](https://github.com/blewis12/test-project/actions/runs/22631925054) - I don't think we'd encounter that sort of issue here but it's good to have just in case the first step mucks up and outputs a bad tag

I think it's ok to test this for real when for our v1.14.0 RC/R process, worst case this workflow fails in which case there aren't any side effects - and we can keep an eye on the first PR (auto-merging) that arrives in the `dev1` wave from this triggered workflow to make sure it's working as expected - if not, the workflow can be cancelled in the UI and the PR closed whilst I investigate

### DIFF
--- a/.github/workflows/trigger-argo-workflow-on-tag.yml
+++ b/.github/workflows/trigger-argo-workflow-on-tag.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Release tag (e.g., v1.14.0)'
+        description: 'Release tag (e.g. v1.14.0)'
         required: true
         default: ''
         type: string


.github/workflows/trigger-argo-workflow-on-tag.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          TAG_NAME="${{ steps.tag.outputs.tag }}"
+          if ! echo "${TAG_NAME}" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Invalid tag format: ${TAG_NAME}. Expected semantic version (e.g. v1.2.3) with no prefix."
+            exit 1
+          fi
+          echo "Tag format valid: ${TAG_NAME}"
+
+      - name: Trigger ArgoCD Workflow
+        uses: grafana/shared-workflows/actions/trigger-argo-workflow@trigger-argo-workflow-v1.2.0
```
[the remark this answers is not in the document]
blewis12: this has been fixed (not sure how to resolve this comment itself though)

.github/workflows/trigger-argo-workflow-on-tag.yml
```diff
@@ -0,0 +1,50 @@
...
+          if ! echo "${TAG_NAME}" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Invalid tag format: ${TAG_NAME}. Expected tag in format vX.Y.Z (no additional prefix/suffix)."
+            exit 1
+          fi
+          echo "Tag format valid: ${TAG_NAME}"
+
+      - name: Trigger ArgoCD Workflow
+        uses: grafana/shared-workflows/actions/trigger-argo-workflow@a37de51f3d713a30a9e4b21bcdfbd38170020593 # v1.3.0
+        with:
+          instance: "ops"
```
blewis12, line 68: if we don't want to test this "live" we can replace this with "dev", but as I said in the description - I think it is ok to give it a go in ops

blewis12:
actually, putting this back in draft until stuff in `deployment_tools` is ready to go

.github/workflows/trigger-argo-workflow-on-tag.yml
```diff
@@ -0,0 +1,50 @@
+name: Trigger Rollout
+
+on:
+  push:
+    tags:
+      - 'v*'
```
kgeckhart:
```suggestion
on:
  release:
    types: [published]
```

ATM the tag is created when the release is drafted causing the artifacts to build. This workflow would race with that one for the artifacts. After the artifacts finish the release is published which is a good time to trigger this workflow. Published should fire for pre-release and releases according to https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release

You should be able to get the tag from `github.event.release.tag_name`

.github/workflows/trigger-argo-workflow-on-tag.yml
```diff
@@ -0,0 +1,50 @@
+name: Trigger Rollout
+
+on:
```
kgeckhart, line 3: I would consider adding a workflow_dispatch here so we can easily trigger manually if something transient happens.

```
  workflow_dispatch:
    inputs:
      tag:
        description: 'Release tag (e.g., v1.14.0)'
        required: true
        default: ''
        type: string
```

.github/workflows/trigger-argo-workflow-on-tag.yml
blewis12: ah nice - that's a good point 👍

.github/workflows/trigger-argo-workflow-on-tag.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          echo "tag=${TAG_NAME}" >> $GITHUB_OUTPUT
+          echo "Detected tag: ${TAG_NAME}"
+
+      - name: Verify tag format
+        run: |
+          if ! echo "${TAG_NAME}" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Invalid tag format: ${TAG_NAME}. Expected tag in format vX.Y.Z (no additional prefix/suffix)."
+            exit 1
+          fi
```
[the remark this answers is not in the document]
blewis12: valid - pushed a change that should gracefully handle RC tags and exit in cases where we're tagging the helm chart or any other sub module: https://github.com/grafana/alloy/pull/5703/changes/932c1476d99e47404b6b24d3ae6cbdfb6f8118d2

PR #5707, opened 2026-03-04
fix(database_observability): Ensure that `connection_info` metric is only emitted for a given DB instance when it is available

rgeyer:
We noticed a discrepancy between the number of DB instances reported in the dbo11y app, and in the billing data.

This is due to the fact that we are using different metrics/queries, namely.

App: `last_over_time((count(count by (server_id) (mysql_perf_schema_events_statements_total{job=~"integrations/db-o11y", digest=~".+", } or pg_stat_statements_calls_total{job=~"integrations/db-o11y", queryid=~".+", })))[$__range:])`

Billing: `{__name__="database_observability_connection_info", server_id!="", job="integrations/db-o11y"}`

Further research revealed that a database which successfully connected, then at some point was unable to be reached would continue emitting the `database_observability_connection_info` metric. This is in opposition to our desired billing model, where a customer could "turn off" a database and not be billed for the time it is down.

### Notes to the Reviewer

~This starts yet-another-goroutine to check the DB connectivity on a regular cadence (every 1 minute) and unregister the metric if there are 3 consecutive failures. It will also re-register the metric if the database ping is successful 3 consecutive times after a failure.~

~This now performs the checks in the main `component.go` file for each dbo11y component, reusing the goroutine that's already present, with different timers. Should be a bit cleaner.~

Nevermind, we went back to putting the checks in the connection_info collector. Final decision :)

This *will* increase pings/connects to the database, but the alternative is having some other collector (which may not be running) report to this collector, which seems the worse of two evils.

### PR Checklist
- [x] Tests updated

Screenshot from the local test. This is starting alloy monitoring a database which is unavailable, starting the database, then stopping, and resuming it.

image


internal/component/database_observability/connection_info_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+	if config != nil {
+		if config.CheckInterval > 0 {
+			interval = config.CheckInterval
+		}
+		if config.ChecksThreshold > 0 {
+			threshold = config.ChecksThreshold
+		}
+	}
+	ctx, cancel = context.WithCancel(ctx)
+	go func() {
```
gaantunes, line 47: Could we use the component.go existing goroutine instead of starting a new one? I think it would make it cleaner.

internal/component/database_observability/connection_info_monitor.go
```diff
@@ -0,0 +1,87 @@
...
+				}
+			} else {
+				consecutiveFailures = 0
+				if metricRegistered {
+					consecutiveSuccesses = 0
+				} else {
+					consecutiveSuccesses++
+					if consecutiveSuccesses >= threshold {
+						registry.MustRegister(infoMetric)
+						infoMetric.WithLabelValues(labelValues[0], labelValues[1], labelValues[2], labelValues[3], labelValues[4], labelValues[5]).Set(1)
```
gaantunes: Can we make labelValues a typed struct instead? this is hard to read and understand.

internal/component/database_observability/connection_info_monitor.go
cristiangreco: Hmm yeah I'm wondering if it's easier to build this on top of the reconnection logic. Today we keep the collectors in "running" state even if the database connection is lost. Maybe we should just "stop()" collectors until database connection retries succeed?

gaantunes: It makes sense, but I wonder if it may cause unnecessary stop/starts in a flapping db connection. 
We would also need to separate the logs collector from the pool, since it is independent from the db connection, once it is started. Right now it only count errors, but once we implement detailed error scraping it will be useful to troubleshoot a flapping db.

rgeyer: Refactored this to (mostly) reuse the existing reconnect logic.

The existing logic seems to only be used when the db is unavailable on first startup, so there is an additional check now to ensure that the DB *remains* connected, and stops the connection_info collector if the DB becomes unavailable.

internal/component/database_observability/mysql/component.go
```diff
@@ -450,6 +483,74 @@ func (c *Component) connectAndStartCollectors(ctx context.Context) error {
 	return nil
 }
 
+// hasConnectionInfoCollector reports whether the connection_info collector is currently in c.collectors.
+// Must be called with c.mut held.
```
cristiangreco: Nit: given this and the next functions are just from one place, maybe they could take care of mutex handling rather than having this reminder?

internal/component/database_observability/connection_info_monitor.go
cristiangreco: I'm a bit confused though by the various tickets though. So, to recap, it seems we have:
- the existing ticker loop for "initial" connection
- a new ticker loop dedicated to `connection_info` with 
I understand we want to keep those activities separate, but I'm wondering if we should just put the new logic inside the `connection_info` collector? i.e. that collector would have its own loop like other collectors

rgeyer: The only *real* reason for having two tickers is that they're on different cadences (30s and 60s).

We could probably combine them, and checking for DB connectivity every 30s is probably not a big deal. 🤔

internal/component/database_observability/mysql/component.go
rgeyer: Yeah.. I don't have a strong opinion here. They're called in a few places, and on lines 321-325 there is only a single mutex lock rather than two separate ones.

We could just make each of those helper functions hold the mutex individually tho, since they are always accessing a property that needs to avoid thread contention. 🤔

rgeyer: Made this change

matthewnolf:
I'm coming from being a bit out of the loop on this topic, but the changes involved seem fairly involved with multiple tickers and being in `component.go` are not tested at all. Is there a reason we can't go with something simpler like passing the db connection to the `connection_info` collector that can simply unregister the metric if a ping fails?

rgeyer:
> I'm coming from being a bit out of the loop on this topic, but the changes involved seem fairly involved with multiple tickers and being in `component.go` are not tested at all. Is there a reason we can't go with something simpler like passing the db connection to the `connection_info` collector that can simply unregister the metric if a ping fails?

@matthewnolf That was the initial implementation. There was a separate goroutine started in each of the `connection_info.go` collector for each DB. It checked with a DB ping every 60 seconds, and would deregister the metric from the registry.

[This thread](https://github.com/grafana/alloy/pull/5707#discussion_r2886519901) is what prompted the changes. I may have overly influenced that, given that my initial PR description pointed out the additional goroutines and DB contacts as a negative thing. Perhaps it's fine?

matthewnolf:
> > I'm coming from being a bit out of the loop on this topic, but the changes involved seem fairly involved with multiple tickers and being in `component.go` are not tested at all. Is there a reason we can't go with something simpler like passing the db connection to the `connection_info` collector that can simply unregister the metric if a ping fails?
> 
> @matthewnolf That was the initial implementation. There was a separate goroutine started in each of the `connection_info.go` collector for each DB. It checked with a DB ping every 60 seconds, and would deregister the metric from the registry.
> 
> [This thread](https://github.com/grafana/alloy/pull/5707#discussion_r2886519901) is what prompted the changes. I may have overly influenced that, given that my initial PR description pointed out the additional goroutines and DB contacts as a negative thing. Perhaps it's fine?

But what is stopping the `connection_info` collector behaving like all the others: taking a DB connection and running in a loop. The loop in this case would be to ping the DB and register/unregister the metric.
Overall I'd prefer to keep logic out of `component.go` that is mainly for wiring, and is less tested.

rgeyer:
> > > I'm coming from being a bit out of the loop on this topic, but the changes involved seem fairly involved with multiple tickers and being in `component.go` are not tested at all. Is there a reason we can't go with something simpler like passing the db connection to the `connection_info` collector that can simply unregister the metric if a ping fails?
> > 
> > 
> > @matthewnolf That was the initial implementation. There was a separate goroutine started in each of the `connection_info.go` collector for each DB. It checked with a DB ping every 60 seconds, and would deregister the metric from the registry.
> > [This thread](https://github.com/grafana/alloy/pull/5707#discussion_r2886519901) is what prompted the changes. I may have overly influenced that, given that my initial PR description pointed out the additional goroutines and DB contacts as a negative thing. Perhaps it's fine?
> 
> But what is stopping the `connection_info` collector behaving like all the others: taking a DB connection and running in a loop. The loop in this case would be to ping the DB and register/unregister the metric. Overall I'd prefer to keep logic out of `component.go` that is mainly for wiring, and is less tested.

Nothing at all, and in fact, that was my initial implementation (see [here](https://github.com/grafana/alloy/pull/5707/changes/524bb097638c585f9de3dc86515ca23da31d05a5)), but there was discussion about changing it in the thread I linked above.

I'm not particular to either approach, but we should come to some consensus and get this merged.

Thoughts @matthewnolf @cristiangreco @gaantunes ?

gaantunes:
> > > > I'm coming from being a bit out of the loop on this topic, but the changes involved seem fairly involved with multiple tickers and being in `component.go` are not tested at all. Is there a reason we can't go with something simpler like passing the db connection to the `connection_info` collector that can simply unregister the metric if a ping fails?
> > > 
> > > 
> > > @matthewnolf That was the initial implementation. There was a separate goroutine started in each of the `connection_info.go` collector for each DB. It checked with a DB ping every 60 seconds, and would deregister the metric from the registry.
> > > [This thread](https://github.com/grafana/alloy/pull/5707#discussion_r2886519901) is what prompted the changes. I may have overly influenced that, given that my initial PR description pointed out the additional goroutines and DB contacts as a negative thing. Perhaps it's fine?
> > 
> > 
> > But what is stopping the `connection_info` collector behaving like all the others: taking a DB connection and running in a loop. The loop in this case would be to ping the DB and register/unregister the metric. Overall I'd prefer to keep logic out of `component.go` that is mainly for wiring, and is less tested.
> 
> Nothing at all, and in fact, that was my initial implementation (see [here](https://github.com/grafana/alloy/pull/5707/changes/524bb097638c585f9de3dc86515ca23da31d05a5)), but there was discussion about changing it in the thread I linked above.
> 
> I'm not particular to either approach, but we should come to some consensus and get this merged.
> 
> Thoughts @matthewnolf @cristiangreco @gaantunes ?

My initial PR comment was more targeted on the additional ticker being added for this new functionality. I agree that keeping it inside the collectors look cleaner on Matt's [PR](https://github.com/grafana/alloy/pull/5759/changes) suggestion, but Ryan's logic for a consecutive failure threshold is important to avoid unneeded billing flapping.

cristiangreco (approved):

cristiangreco:
@rgeyer approved, and renamed the PR to get a nicer changelog entry

PR #5709, opened 2026-03-05
feat(loki.source.syslog): Fix buffer read in raw parser

x1unix:

This PR fixes buffer read in the raw parser for logs with octet-counting framing type.

PR addresses the issue when multiple, long octet-counted log-lines could be read as a single Loki log entry.

### PR Checklist

- [x] Tests updated


internal/component/loki/source/syslog/internal/syslogtarget/syslogparser/rawparser.go
```diff
@@ -41,23 +41,26 @@ func IterStreamRaw(r io.Reader, delimiter byte) iter.Seq2[*syslog.Base, error] {
...
-	b, err := buf.ReadByte()
+func parseLineRaw(src *bufio.Reader, delimiter byte) (*syslog.Base, error) {
+	b, err := src.ReadByte()
 	if err != nil {
 		return nil, err
 	}
 
 	// TODO: use bytebufferpool?
-	_ = buf.UnreadByte()
+	_ = src.UnreadByte()
```
kalleep, line 51: nit: you could use `src.Peek(1)` this will not advance positions and does not allocate

internal/component/loki/source/syslog/internal/syslogtarget/syslogparser/rawparser.go
```diff
@@ -41,23 +41,26 @@ func IterStreamRaw(r io.Reader, delimiter byte) iter.Seq2[*syslog.Base, error] {
...
 	ftype := framingTypeFromFirstByte(b)
 	if ftype == framingTypeOctetCounting {
-		contentLength, err := readFrameLength(buf)
+		contentLength, err := readFrameLength(src)
 		if err != nil {
 			return nil, fmt.Errorf("failed to read octet length header: %w", err)
 		}
 
+		// The .Read() method in bufio.Reader will return only buffered data.
```
kalleep, line 59: This it not really true. It will return buffered data if it have it otherwise it will do one `Read` on the underlying io.Reader

PR #5711, opened 2026-03-05
feat(database_observability.mysql): Embed prometheus exporter within db-o11y component

matthewnolf:
### Pull Request Details
This change enables the `prometheus.exporter.mysql` component to be embedded inside the `database_observability.mysql` component so that it encapsulates the functionality. This enables end users to create a single component, instead of multiple and wiring them up.

This also will aid our support for clustering mode in a future change.

If the `mysql_exporter` config block is provided, we create the exporter and attach it to the prometheus registry already used for `database_observbaility_*` metrics.

Example config:
```
database_observability.mysql "database" {
   data_source_name = "..."
   forward_to       = [loki.relabel.database.receiver]
   targets = []
   mysql_exporter {
    enable_collectors = ["perf_schema.eventsstatements", "perf_schema.eventswaits"]
   }
}
```

 

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/component.go
+++ b/internal/component/database_observability/mysql/component.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"database/sql"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"path"
 	"regexp"
@@ -17,6 +18,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/model"
+	mysqld_collector "github.com/prometheus/mysqld_exporter/collector"
 	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component"
@@ -25,9 +27,12 @@ import (
 	"github.com/grafana/alloy/internal/component/database_observability"
 	"github.com/grafana/alloy/internal/component/database_observability/mysql/collector"
 	"github.com/grafana/alloy/internal/component/discovery"
+	exporter_mysql "github.com/grafana/alloy/internal/component/prometheus/exporter/mysql"
 	"github.com/grafana/alloy/internal/featuregate"
+	"github.com/grafana/alloy/internal/runtime/logging"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
 	http_service "github.com/grafana/alloy/internal/service/http"
+	"github.com/grafana/alloy/internal/static/integrations/mysqld_exporter"
 	"github.com/grafana/alloy/syntax"
 	"github.com/grafana/alloy/syntax/alloytypes"
 )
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/component.go
```diff
@@ -425,6 +446,26 @@ func (c *Component) connectAndStartCollectors(ctx context.Context) error {
 		cp = cloudProvider
 	}
 
+	if c.exporterCollector != nil {
+		c.registry.Unregister(c.exporterCollector)
+		c.exporterCollector = nil
+	}
+
+	if c.args.MySQLExporter != nil {
```
cristiangreco: Should we automatically wire `targets`? And maybe having a check to prevent using both the "embedded" exporter and the targets coming from another block?

internal/component/database_observability/mysql/component.go
```diff
@@ -72,6 +77,7 @@ type Arguments struct {
 	LocksArguments          LocksArguments          `alloy:"locks,block,optional"`
 	QuerySamplesArguments   QuerySamplesArguments   `alloy:"query_samples,block,optional"`
 	HealthCheckArguments    HealthCheckArguments    `alloy:"health_check,block,optional"`
+	MySQLExporter           *MySQLExporterArguments `alloy:"mysql_exporter,block,optional"`
```
cristiangreco: nit: I feel this should be named somewhat similar to `prometheus.exporter.mysql` but not exactly the same.

internal/component/database_observability/mysql/component.go
```diff
@@ -132,6 +138,20 @@ type HealthCheckArguments struct {
...
+//
+// It is a distinct type (not an embedded struct) because the Alloy syntax
+// system does not support anonymous/embedded fields.
+type PrometheusExporterArguments exporter_mysql.Arguments
+
+func (a *PrometheusExporterArguments) SetToDefault() {
+	*a = PrometheusExporterArguments(exporter_mysql.DefaultArguments)
+}
+
+func (a *PrometheusExporterArguments) Validate() error { return nil }
```
cristiangreco: Should we call `Validate()` on the args here? Or maybe we could call it in the `Validate()` if this component itself? (line 206)

internal/component/database_observability/mysql/component.go
```diff
@@ -425,6 +449,26 @@ func (c *Component) connectAndStartCollectors(ctx context.Context) error {
...
+		exporterArgs := exporter_mysql.Arguments(*c.args.PrometheusExporter)
+		exporterCfg := exporterArgs.Convert()
+		scrapers := mysqld_exporter.GetScrapers(exporterCfg)
+		slogLogger := slog.New(logging.NewSlogGoKitHandler(c.opts.Logger))
+		exporter := mysqld_collector.New(context.Background(), string(c.args.DataSourceName), scrapers, slogLogger, mysqld_collector.Config{
+			LockTimeout:   exporterCfg.LockWaitTimeout,
+			SlowLogFilter: exporterCfg.LogSlowFilter,
+		})
+		if err := c.registry.Register(exporter); err != nil {
+			return fmt.Errorf("failed to register mysqld_exporter collector: %w", err)
```
cristiangreco:
```suggestion
			return fmt.Errorf("failed to register prometheus_exporter collector: %w", err)
```

cristiangreco (approved):

PR #5713, opened 2026-03-05
fix(loki.source.gcplog): Update to pubsub v2 and fix shutdown semantics

kalleep:
### Pull Request Details
Going through the code I found multiple issues we should fix.

#### Component
Use `loki.Fanout` and `source.Consume`.

#### PullTarget
1. Update to use pubsub v2
*  [SubscriptionInProject was removed](https://github.com/googleapis/google-cloud-go/blob/main/pubsub/MIGRATING.md#topicinproject-and-subscriptioninproject-removed). It's enough to call `Subscriber` with our [subscription id because the client is created with project id](https://github.com/googleapis/google-cloud-go/blob/main/pubsub/v2/subscriber.go#L65)
2. Fixed `Ack/Nack MUST be called within the Subscriber.Receive handler function, and not from a goroutine`. This is part of the contract for both v1 and v2. Previously we had one go rountine consuming messages and send to another where we Ack/Nack.
3. If target is stopped we now cancel any sends and call `msg.Nack()`. 

#### PushTarget
1. Implement same force shutdown semantics as [I did for loki.source.api](https://github.com/grafana/alloy/pull/4818). I.e. when graceful period have passed we cancel all blocked in-flight request and return 503. This will cause clients to retry.
2. Update metric namespace to `loki_source_gcplog_push` for metrics registered by the server, before this had the component id but because the registry is scoped to component this is not necessary and matches what we use for other components. These metrics are not documented.
 
### Notes to the Reviewer

I also included some general refactor to clean up the code, can split it out into another pr if that is prefered

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
+++ b/internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
@@ -6,12 +6,12 @@ package gcplogtarget
 // from GCP.
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/grafana/loki/pkg/push"
+	jsoniter "github.com/json-iterator/go"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/relabel"
@@ -58,7 +58,7 @@ type parseOptions struct {
 func parseLogEntry(data []byte, builder *labels.Builder, relabelConfig []*relabel.Config, opts parseOptions) (loki.Entry, error) {
 	var entry LogEntry
 
-	if err := json.Unmarshal(data, &entry); err != nil {
+	if err := jsoniter.Unmarshal(data, &entry); err != nil {
 		return loki.Entry{}, err
 	}
 


internal/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
```diff
@@ -60,89 +58,84 @@ type pubsubSubscription interface {
...
-			t.handler.Chan() <- entry
-			m.Ack() // Ack only after log is sent.
-			t.metrics.gcplogEntries.WithLabelValues(t.config.ProjectID).Inc()
 		}
-	}
+	})
+	return nil
 }
 
-func (t *PullTarget) consumeSubscription() {
```
kalleep, line 125: Not even sure why there was two loops and a extra channel for message since we can just parse and pass them directly.

internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
kgeckhart, line 1: Was this just a rename or did code change too? If code changed could you put the name back to make diff'ing easier. Edit: Looks like this merged the push translate + pull formatter?

internal/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
```diff
@@ -60,89 +58,84 @@ type pubsubSubscription interface {
...
+
+		for t.backoff.Ongoing() {
+			err := t.sub.Receive(t.ctx, func(ctx context.Context, m *pubsub.Message) {
+				entry, err := parseLogEntry(m.Data, labels.NewBuilder(labels.EmptyLabels()), t.relabelConfig, parseOptions{
+					fixedLabels:          lbls,
+					useFullLine:          t.config.UseFullLine,
+					useIncomingTimestamp: t.config.UseIncomingTimestamp,
+				})
+				if err != nil {
+					level.Error(t.logger).Log("event", "cloud not parse log entry", "error", err)
```
kgeckhart:
```suggestion
					level.Error(t.logger).Log("msg", "could not parse log entry", "error", err)
```

internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
kalleep: >Looks like this merged the push translate + pull formatter?

The push translate always called the parse function in formatter.go.

Changes I made here:
1. Renamed Structures, no longer have GCP prefix.
2. Renamed functions
3. Update functions signatures to accept parseOptions, relabel rules, use incomming timestamt and fixed label set
4. Pass labels.Builder instead of creating a builder, converting to labels and then create a builder again.

Functionality wise nothing changed but I can revert it you want.

internal/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
blewis12: I'm guessing because both `sub.Receive()` and writing to `handler.Chan()` block, so they both run in their own goroutine to avoid blocking one another? but your revamp of things looks like a good approach to me

internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
```diff
@@ -6,23 +6,26 @@ package gcplogtarget
 // from GCP.
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
-	json "github.com/json-iterator/go"
```
kgeckhart, line 13: I remember this being a rather big performance win it's unfortunate the project was archived and json v2 is still experimental

internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
```diff
@@ -44,47 +47,48 @@ type GCPLogEntry struct {
...
 	} else {
-		processed = lbs.Labels()
+		processed = builder.Labels()
 	}
 
-	// final labelset that will be sent to loki
 	lbls := make(model.LabelSet)
 	processed.Range(func(lbl labels.Label) {
-		if strings.HasPrefix(lbl.Name, "__") {
+		if strings.HasPrefix(lbl.Name, "__") && lbl.Name != reservedLabelTenantID {
```
kgeckhart, line 91: Wait how did this work before? Is there a test we are missing we should add?

blewis12, line 91: what does it mean now to keep `__tenant_id__` labels that were previously dropped?

blewis12: ohh nevermind, it was handled [here](https://github.com/grafana/alloy/pull/5713/changes#diff-230b13359bd2e8cb81b047485c817be07cf0ea4740cafb39d7b581ea3ef72839L81) before

blewis12: oh sorry didn't see your comment, I made a duplicate one [here](https://github.com/grafana/alloy/pull/5713#discussion_r2913891469) and linked to where (I think) this was handled before

kalleep: Yes it was handled before, we added it too both the label builder and label set. Figured that was wasteful so now we just set it here.

internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go
kalleep: Okay, we can keep it. I check the repo and yes it was archived that's why I changed to stdlib.

When / if json v2 is out we should use that one instead!

kalleep: Changed it back

internal/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
kalleep: Yes sure but I don't really see why that was even necessary, just another channel in the mix and extra go routine for seemingly no reason :man_shrugging:

PR #5717, opened 2026-03-06
feat: Add `key_mode` config option for position tracking

### LINKED ISSUE #5135: Don't always reset offset in positions.yaml if targets change
### Component(s)

loki.source.file

### Request

Hello :)

we recently ran into a problem with Alloy + the k8s-monitoring-helm chart (see here https://github.com/grafana/k8s-monitoring-helm/issues/2218).

I proposed a change to the chart but also want to talk about how Alloy (loki.source.file I guess?) handles the positions.yaml if it discovers new targets within the same Pod. 

Right now it seems that a change to the labels that are available to the pipeline lead to the current target being removed and a new one being created. Including the offset of the tailed logfile being reset in the positions.yaml

This together with a pipeline that injects Kubernetes annotations as labels leads to a "loop of death" if a workload uses the annotations of a pod to store state (as it does with our TimescaleDB cluster with Patroni). 

While I understand that the target is recalculated if a new label for a workload appears, I think there is some benefit of keeping the position offset for the affected logfile. Otherwise the same logs are exported again and again. Even if it wouldn't be an automated process that updates an annotation value in a given interval, I personally wouldn't want to have loglines more than once in my log system.

I'm not sure if the current behavior is working as intended, therefore I added this issue as a feature request and not a bug. 

Happy to have a discussion on that / hear your opinions. If you need more information I'd be more than happy to provide them. 

P.S. here is a thread in the #alloy channel where I had a chat with @dehaansa about that issue.  

### Use case

To not have logs ingested more than once. 

### Tip

kalleep:
### Pull Request Details
This pr will expose config options for all loki source components relevant to position tracking, `key_mode` and `sync_period` under `position` block.

E.g. for `loki.source.file`

```
loki.source.file "f" {
  position {
    key_mode = "include_labels"
    sync_period = "10s"
  }
}
```

The important changes is `key_mode` where `include_labels` will include labels in position tracking and `exclude_labels` will only track by path, contatinerid etc. Default value is to include labels because that is what alloy _always_ did before this change.

When switching between the two different modes I have added some fallback logic. When using `exclude_labels` we first check if we have a stored position without any labels, if not we check with labels. When using `include_labels` we fist check if we have a stored position with labels, if not we check without.

Remove always removes both so that we are not left with any orphaned positions. If we track the same file twice without any labels they will both update the same position entry.  

Fixes: #5135

### Notes to the Reviewer

This pr also includes a some of code cleanup for position package:
1.  Move file implementation from `positions.go` to `file.go`.
2.  Moved legacy structure and functions from `position.go` to `legacy.go`. 
3. Removed some options that we don't use in config like `ReadOnly` and `IngoreInvalidYaml`.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/loki/loki.source.cloudflare.md
```diff
@@ -213,7 +213,17 @@ A sample log looks like this:
 
 ## Blocks
 
-The `loki.source.cloudflare` component doesn't support any blocks. You can configure this component with arguments.
+You can use the following blocks with `loki.source.cloudflare`:
```
clayton-cornell:
```suggestion
You can use the following block with `loki.source.cloudflare`:
```

docs/sources/reference/components/loki/loki.source.cloudflare.md
```diff
@@ -213,7 +213,17 @@ A sample log looks like this:
...
+
+| Name                     | Description                                         | Required |
+|--------------------------|-----------------------------------------------------|----------|
+| [`position`][position]   | Configure how positions are keyed and synced to disk. | no    |
```
clayton-cornell:
```suggestion
| Name                   | Description                                           | Required |
| ---------------------- | ----------------------------------------------------- | -------- |
| [`position`][position] | Configure how positions are keyed and synced to disk. | no       |
```

docs/sources/shared/reference/components/loki-source-position-block.md
```diff
@@ -0,0 +1,13 @@
...
+
+| Name          | Type       | Description                                   | Default            | Required |
+|---------------|------------|-----------------------------------------------|--------------------|----------|
+| `key_mode`    | `string`   | How position entries are keyed.               | `"include_labels"` | no       |
+| `sync_period` | `duration` | How often to sync the positions file to disk. | `"10s"`            | no       |
+
+The `key_mode` argument must be one of:
+
+* `"include_labels"`: track positions by key and labels.
+* `"exclude_labels"`: track positions by key only.
```
clayton-cornell:
```suggestion
* `"include_labels"`: Track positions by key and labels
* `"exclude_labels"`: Track positions by key only
```

kalleep:
Need to fix this pr. The changes are not this big but because I moved stuff around the diff is huge.

docs/sources/shared/reference/components/loki-source-position-block.md
```diff
@@ -0,0 +1,13 @@
...
+
+| Name          | Type       | Description                                   | Default            | Required |
+|---------------|------------|-----------------------------------------------|--------------------|----------|
+| `key_mode`    | `string`   | How position entries are keyed.               | `"include_labels"` | no       |
+| `sync_period` | `duration` | How often to sync the positions file to disk. | `"10s"`            | no       |
+
+The `key_mode` argument must be one of:
+
+* `"include_labels"`: Track positions by key and labels
+* `"exclude_labels"`: Track positions by key only
```
thampiotr, line 11: As a user how do I know what `key` means here?

kalleep: That is a fair point. Either we clarify how it works here or we come up with a better name

internal/component/loki/source/docker/docker.go
ptodev, line 1: Do we really have to expose extra config options for every component? Can we not change the code and the position file format so that they are better and backwards compatible? E.g. the positions for loki.source.docker can contain the same thing they do now + container ID. Then if the container ID is present, it will be used instead of the labels?

docs/sources/reference/components/loki/loki.source.cloudflare.md
```diff
@@ -213,7 +213,17 @@ A sample log looks like this:
 
 ## Blocks
 
-The `loki.source.cloudflare` component doesn't support any blocks. You can configure this component with arguments.
+You can use the following block with `loki.source.cloudflare`:
+
```
clayton-cornell, line 217:
```suggestion

{{< docs/alloy-config >}}

```

We need to add the new blocks shortcode

docs/sources/reference/components/loki/loki.source.cloudflare.md
```diff
@@ -213,7 +213,17 @@ A sample log looks like this:
...
+| Name                   | Description                                           | Required |
+| ---------------------- | ----------------------------------------------------- | -------- |
+| [`position`][position] | Configure how positions are keyed and synced to disk. | no       |
+
+[position]: #position
+
```
clayton-cornell, line 223:
```suggestion

{{< /docs/alloy-config >}}

```

docs/sources/shared/reference/components/loki-source-position-block.md
```diff
@@ -0,0 +1,13 @@
+The `position` block configures how a component tracks and persists read offsets.
```
clayton-cornell, line 1:
```suggestion
---
canonical: https://grafana.com/docs/alloy/latest/shared/reference/components/loki-source-position-block/
description: Shared content, loki source position block
headless: true
---

The `position` block configures how a component tracks and persists read offsets.
```

Issue #5365, opened 2026-01-27
loki: loki.source.file will read already consumed compressed files on restarts.


kalleep:
### Component(s)

loki.source.file

### What's wrong?

In alloy we store offsets where we should start reading next into a position file. When we stop reading a file (both compressed and non compressed files) we check if the component is stopping. If it's not stopping we handle that as target was removed so either the file removed or we no longer want to track it.

When reading compressed files we stop at EOF and remove position. So if alloy is restarted it will re-ingest that file  

### Steps to reproduce

Configure alloy to read a compressed file. Let it consume it fully and restart alloy.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

kalleep:
This is a regression and have been in alloy since https://github.com/grafana/alloy/pull/2428

PR #5723, opened 2026-03-09
fix(loki.source.file): Keep positions for compressed files when reading is finished

kalleep:
### Pull Request Details

This is a [regression that have been in alloy since 1.8](https://github.com/grafana/alloy/pull/2428). When tailing compressed files we stop once we hit EOF. The issue is that tailer, and previously decompressor, will remove the stored position because component is not stopping. Then when alloy is re-started we read the files again.

Instead we should never remove the position here when decompression is enabled, position structure will run [cleanup](https://github.com/grafana/alloy/blob/c47338fec6eb26fa9f2d040cc002f78721b93427/internal/component/loki/source/internal/positions/positions.go#L303) on files that no longer exists so it will be able to clean it up when files disapears. 

Fixes: #5365

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/tailer.go
+++ b/internal/component/loki/source/file/tailer.go
@@ -268,7 +268,6 @@ func (t *tailer) stop(done chan struct{}) {
 	if !t.shouldKeepPosition() {
 		t.positions.Remove(t.key.Path, t.key.Labels)
 	}
-
 }
 
 func (t *tailer) Key() positions.Entry {


kalleep:
>A test is missing

@thampiotr TestTailer_Compressions was already trying to test this behavior but because it always returned  true for componentIsStopping it passed. 

I changed that to false, this makes this test fail on main but pass in this pr

PR #5735, opened 2026-03-10
fix(beyla): Inject Beyla version into binary via ldflags

### LINKED ISSUE #4385: beyla_internal_build_info
### Component(s)

_No response_

### Add a description

beyla_internal_build_info metric show version as unset I'm using alloy version 1.10.2 

### Tip

pratik50:

When Beyla is used as a library inside Alloy, the `beyla_build_info` metric was reporting `version="unset"` because Beyla's build-time version flags were never passed during Alloy's build process.

This PR injects the Beyla version into the Alloy binary via `-ldflags` at compile time, so the correct version is exposed via the `beyla_build_info` metric.

### Pull Request Details

Alloy builds Beyla as a Go library dependency. Unlike standalone Beyla builds, Alloy's build system was not passing the `ldflags` needed to set `pkg/buildinfo.Version`, causing it to fall back to its default value of `"unset"`.

The fix retrieves the Beyla version directly from `go.mod` using `go list -m` and injects it at build time.

### Issue fixed by this Pull Request

Fixes https://github.com/grafana/beyla/issues/2584
Fixes #4385

Changes are minimal and focused:
- `Makefile` — fetch Beyla version from `go.mod` and inject via ldflags
- `common/metric.go` — add `Version` field to parse the version label
- `beyla_test.go` — add test to verify version is not `"unset"`

### PR Checklist

- [x] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,9 @@ GO_ENV := GOEXPERIMENT=$(GOEXPERIMENT) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOA
 VERSION      ?= $(shell bash ./tools/image-tag)
 GIT_REVISION := $(shell git rev-parse --short HEAD)
 GIT_BRANCH   := $(shell git rev-parse --abbrev-ref HEAD)
-BEYLA_VERSION := $(shell go list -m github.com/grafana/beyla/v2 | cut -d' ' -f2)
-BEYLA_PKG     := github.com/grafana/beyla/v2/pkg/buildinfo
+BEYLA_MODULE  := $(shell go list -m all | grep "^github.com/grafana/beyla" | head -1)
+BEYLA_VERSION := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f2)
+BEYLA_PKG     := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f1)/pkg/buildinfo
 VPREFIX      := github.com/grafana/alloy/internal/build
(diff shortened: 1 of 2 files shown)


pratik50:
Hi @grafana/beyla!

This PR fixes grafana/beyla#2584 where `beyla_build_info` was reporting `version="unset"` when Beyla runs inside Alloy.

The fix injects Beyla's version via `-ldflags` at Alloy's build time. Would love your guidance and review on this!

Makefile
```diff
@@ -130,6 +130,8 @@ GO_ENV := GOEXPERIMENT=$(GOEXPERIMENT) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOA
 VERSION      ?= $(shell bash ./tools/image-tag)
 GIT_REVISION := $(shell git rev-parse --short HEAD)
 GIT_BRANCH   := $(shell git rev-parse --abbrev-ref HEAD)
+BEYLA_VERSION := $(shell go list -m github.com/grafana/beyla/v2 | cut -d' ' -f2)
```
marctc: Can we make this that doesn't depend on the major version? now is pointing to `v2` but we are changing to v3, so this wont work.

pratik50: Thanks for the feedback @marctc 
I'll update it to not depend on the major version.

pratik50:
Updated to dynamically detect the Beyla module so will work with latest version without hardcode!

Makefile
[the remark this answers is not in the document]
pratik50: @marctc what are your point of views on this?

PR #5737, opened 2026-03-10
test: Fix several flaky tests

thampiotr:

Fixes several flaky tests as reported in https://github.com/grafana/alloy/issues/5142

### Pull Request Details

Following test issues addressed:

TestUpdate_detectsWhenTargetRequiresARestart
(internal/component/loki/source/heroku/heroku_test.go)

TestClient
(internal/component/otelcol/auth/bearer/bearer_test.go)

processortest.go/Test_Update (reported as startup-timeout flake)
Addressed via shared timeout hardening in processortest helper.

Test_Hash
Addressed via the same processortest timeout hardening used by attributes processor tests.

Test_ComponentIO (spanlogs / otelcol connector path)
Addressed via shared processortest timeout hardening.

TestEndpoint/batch_log_entries_together_until_the_batch_wait_time_is_reached
(current test name in file is the “batch wait time is reached” subtest in internal/component/common/loki/client/endpoint_test.go)

TestFile/deleted_while_reading
(internal/component/loki/source/file/internal/tail/file_test.go)

Test_SchemaDetails_populates_TableRegistry
(internal/component/database_observability/postgres/collector/schema_details_test.go, with lifecycle fix in schema_details.go)
### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/processor/processortest/processortest.go
+++ b/internal/component/otelcol/processor/processortest/processortest.go
@@ -39,11 +39,7 @@ type ProcessorRunConfig struct {
 	L                     log.Logger
 }
 
-const (
-	componentStartTimeout  = 5 * time.Second
-	componentExportTimeout = 5 * time.Second
-	signalOutputTimeout    = 3 * time.Second
-)
+const signalOutputTimeout = 3 * time.Second
 
 func TestRunProcessor(c ProcessorRunConfig) {
 	go func() {
@@ -51,8 +47,8 @@ func TestRunProcessor(c ProcessorRunConfig) {
 		require.NoError(c.T, err)
 	}()
 
-	require.NoError(c.T, c.Ctrl.WaitRunning(componentStartTimeout), "component never started")
-	require.NoError(c.T, c.Ctrl.WaitExports(componentExportTimeout), "component never exported anything")
+	require.NoError(c.T, c.Ctrl.WaitRunning(0), "component never started")
+	require.NoError(c.T, c.Ctrl.WaitExports(0), "component never exported anything")
 
 	// Send signals in the background to our processor.
 	go func() {
--- a/internal/component/otelcol/processor/processortest/processortest.go
+++ b/internal/component/otelcol/processor/processortest/processortest.go
@@ -39,14 +39,20 @@ type ProcessorRunConfig struct {
 	L                     log.Logger
 }
 
+const (
+	componentStartTimeout  = 5 * time.Second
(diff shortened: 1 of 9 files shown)


internal/component/otelcol/processor/processortest/processortest.go
```diff
@@ -39,14 +39,20 @@ type ProcessorRunConfig struct {
...
+
 func TestRunProcessor(c ProcessorRunConfig) {
 	go func() {
 		err := c.Ctrl.Run(c.Ctx, c.Args)
 		require.NoError(c.T, err)
 	}()
 
-	require.NoError(c.T, c.Ctrl.WaitRunning(time.Second), "component never started")
-	require.NoError(c.T, c.Ctrl.WaitExports(time.Second), "component never exported anything")
+	require.NoError(c.T, c.Ctrl.WaitRunning(componentStartTimeout), "component never started")
```
kalleep: I have seen these tests fail all over the place. Maybe we should hardcode the timeout inside `WaitRunning`. That way it would be consistent in all places we use it at least, WDYT?

internal/component/common/loki/client/endpoint_test.go
```diff
@@ -353,7 +355,11 @@ func TestEndpoint(t *testing.T) {
 				receivedReqs = append(receivedReqs, req)
 			}
 
-			assert.ElementsMatch(t, tt.expectedReqs, receivedReqs)
+			if tt.allowBatchMerge {
```
kalleep: I investigated this test a bit yesterday and noticed that windows required longer delay in CI because of scheduling.

The problem is that the loop for shards are not started before we start queuing. I increase the wait time a bit in https://github.com/grafana/alloy/pull/5738 for windows and it worked in the pr but I fear it still could be really flaky. 

I will check if there is a better way to wait for shards to be started before we start queuing

internal/component/database_observability/postgres/collector/schema_details.go
```diff
@@ -416,7 +419,10 @@ func (c *SchemaDetails) Stopped() bool {
 
 // Stop should be kept idempotent
 func (c *SchemaDetails) Stop() {
-	c.cancel()
+	if c.cancel != nil {
+		c.cancel()
+	}
```
cristiangreco, line 419: I think this and `defer ticker.Stop()` are the important change. We shouldn't need the wg.

kalleep (approved):

PR #5739, opened 2026-03-10
chore: Use shared source structures for aws firehose 

kalleep:
### Pull Request Details
1. Use loki.Fanout and source.Consume
2. Remove grpc block, dskit will always start a grpc server along side http but there is no way for aws to call this server.
3. Don't use component id in metric names, dskit server exposes some metrics that we currently don't document but these should not be prefixed with component id but with component name like we do in [loki.source.api](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/api/internal/lokipush/push_api_server.go#L68). This is fine because all metircs have component_id as label and these are not documented.

There is other fixes I want to do too this component but will do that in a follow up.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated

BREAKING-CHANGE: Renamed undocumented metrics that was previously prefixed with <component_id>_<metric_name> to loki_source_awsfirehose_<metric_name>

### DIFF
--- a/internal/component/loki/source/aws_firehose/component.go
+++ b/internal/component/loki/source/aws_firehose/component.go
@@ -16,6 +16,7 @@ import (
 	"github.com/grafana/alloy/internal/component/common/loki"
 	fnet "github.com/grafana/alloy/internal/component/common/net"
 	alloy_relabel "github.com/grafana/alloy/internal/component/common/relabel"
+	"github.com/grafana/alloy/internal/component/loki/source"
 	"github.com/grafana/alloy/internal/component/loki/source/aws_firehose/internal"
 	"github.com/grafana/alloy/internal/util"
 )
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/aws_firehose/component.go
```diff
@@ -148,20 +137,18 @@ func (c *Component) Update(args component.Arguments) error {
...
 	// update relabel rules in component if needed
 	if handlerNeedsUpdate {
 		c.rbs = newRelabels
 	}
 
-	r := strings.NewReplacer(".", "_", "/", "_")
-	jobName := r.Replace(c.opts.ID)
-
 	registry := prometheus.NewRegistry()
 	c.serverMetrics.SetCollector(registry)
```
ptodev, line 150: Are you sure that this component has a `component_id` label? It doesn't use the normal registry that would be passed in component `opts`.

It would be nice to have the debug metrics in the reference docs.

It could be worth pointing out this change in the changelog, maybe as a breaking change? Just so people who have dashboards or alerts on those metrics can change them.

kalleep: >Are you sure that this component has a component_id label

Yes i verified it

>It would be nice to have the debug metrics in the reference docs.

Also Yes but I have the same change in another pr for loki.source.gcplog. After both this one and that one is merged I will update the docs for all of them. Will create issue for that

>It could be worth pointing out this change in the changelog

Good call, will add that

kalleep:
@ptodev Created a issue about documenting these metrics https://github.com/grafana/alloy/issues/5751

kalleep:
It's a bit weird that we cannot prevent the grpc server from starting. Currently these components do not pass a port for it so a port will be selected at random. But this also means a random port will be occupied and while it's unlikely it could collide with another configured port.

PR #5746, opened 2026-03-10
fix(loki.process): Multiline no longer pass empty entry if start was flushed

kalleep:
In https://github.com/grafana/alloy/pull/5702 some changes to `stage.multiline` was made to support created timestamp along with some optimizations (No longer do a deep clone). But we also made sure to always reset to a zero entry.

This is a issue when a multiline entry is flushed before completion, either by max line or max wait. Then all lines that is not considered start line will be added to an empty entry and later flushed.

To fix this we revert to the old behavior where we clone the stored entry and keep it around for other lines. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/multiline_test.go
+++ b/internal/component/loki/process/stages/multiline_test.go
@@ -191,8 +191,8 @@ func TestMultilineStageStartLineFlushedBeforeNew(t *testing.T) {
 	require.Equal(t, startTs, out[0].Timestamp)
 	require.Equal(t, "START line 1\ncontinuation line 1", out[0].Line)
 
-	require.Equal(t, lset, out[0].Labels)
-	require.Equal(t, startTs, out[0].Timestamp)
+	require.Equal(t, lset, out[1].Labels)
+	require.Equal(t, startTs, out[1].Timestamp)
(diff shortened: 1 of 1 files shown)


internal/component/loki/process/stages/multiline_test.go
```diff
@@ -156,6 +152,50 @@ func TestMultilineStageMaxWaitTime(t *testing.T) {
...
+		},
+	)
+
+	require.Len(t, out, 2)
+	require.Equal(t, lset, out[0].Labels)
+	require.Equal(t, startTs, out[0].Timestamp)
+	require.Equal(t, "START line 1\ncontinuation line 1", out[0].Line)
+
+	require.Equal(t, lset, out[0].Labels)
+	require.Equal(t, startTs, out[0].Timestamp)
```
thampiotr:
```suggestion
	require.Equal(t, lset, out[1].Labels)
	require.Equal(t, startTs, out[1].Timestamp)
```
I gather we wanted to check entry 1 here?

kalleep: yes :facepalm:

Issue #5669, opened 2026-02-26
[otel-engine] unable to start multiple `alloyengine` instances


theSuess:
### What's wrong?

I tried running two instances of the `alloyengine` extension just to see if it works and alloy fails to start the second instance with the following error:
```
2026-02-26T14:15:32.452Z        warn    alloyengine/extension.go:152    run command failed, will retry  {"resource": {"service.instance.id": "8029dacf-9cfa-41ea-bdc5-7fe737309a56", "service.name": "alloy", "service.version": "v1.13.2"}, "otelcol.component.id": "alloyengine/one", "otelcol.component.kind": "extension", "error": "failed to build loader: a loader exists already exists for \"\"", "retry_delay": 4, "attempt": 2}
```

### Steps to reproduce

1. Create a dummy `alloy.config`
2. Create a otelcol configuration which spawns two instances of the `alloyengine` extension like this:
  ```
  receivers:
    otlp:
      protocols:
        grpc:
          endpoint: 0.0.0.0:4317
        http:
          endpoint: 0.0.0.0:4318
  exporters:
    debug:
  extensions:
    alloyengine/one:
      config:
        file: /opt/pwd/config.alloy
      flags:
        server.http.listen-addr: 0.0.0.0:12345
    alloyengine/two:
      config:
        file: /opt/pwd/config.alloy
      flags:
        server.http.listen-addr: 0.0.0.0:12346

  service:
    extensions: [alloyengine/one,alloyengine/two]
    pipelines:
      metrics:
        receivers: [otlp]
        processors: []
        exporters: [debug]
  ```
3. Run `alloy otel --config=config.yaml`

### Software version

docker.io/grafana/alloy-dev:v1.13.2-devel-5bf4dcf

### Configuration

```text

```

### Logs

```text
2026-02-26T14:15:32.452Z        warn    alloyengine/extension.go:152    run command failed, will retry  {"resource": {"service.instance.id": "8029dacf-9cfa-41ea-bdc5-7fe737309a56", "service.name": "alloy", "service.version": "v1.13.2"}, "otelcol.component.id": "alloyengine/one", "otelcol.component.kind": "extension", "error": "failed to build loader: a loader exists already exists for \"\"", "retry_delay": 4, "attempt": 2}
```

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

blewis12:
hiya, I can have a look into this one

blewis12:
[...]

PR #5763, opened 2026-03-11
fix: Enforce Singleton For Running Alloy Extension Instances

blewis12:
This addresses an issue brought up in https://github.com/grafana/alloy/issues/5669, where if you attempt to run _multiple_ instances of the `alloyengine` extension you will encounter errors. These errors are because the default engine has global state that gets mucked up when you try to run multiple pipelines in the same process 

### Pull Request Details
My proposal is: I don't think we should support this. We don't currently have a way of doing that outside of the extension, and if we enable the ability to run multiple default pipelines in one collector instance via the extension, it opens up for quite some complexity. Instead, I think we should do what extensions like `pprof` do and enforce a rule that only one extension instance can be started/running at once. I have stolen [their implementation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/acb90fdf6aa319b6aaf717d2895b95f843e59eba/extension/pprofextension/pprofextension.go#L22) and applied it here to the alloy engine extension, and adjusted the documentation accordingly

This change means that users can do something like 

```
extensions:
  alloyengine/one:
    config:
      file: /opt/pwd/config.alloy
    flags:
      server.http.listen-addr: 0.0.0.0:12345
  alloyengine/two:
    config:
      file: /opt/pwd/config.alloy
    flags:
      server.http.listen-addr: 0.0.0.0:12346

service:
  extensions: [alloyengine/one]
```

but as soon as they add an extension to their running list like

```
service:
  extensions: [alloyengine/one, alloyengine/two]
```

they will encounter an error with a clear indicator that only one extension can be running.

The other approach could be to _merge_ the config for multiple extensions into one running extension - but this isn't very conceptually clean, and we already allow users to pass in a directory (since the extension just exposes the same functionality as the `run` CLI command). I've made it a bit more clear in the docs here that you can indeed pass a directory through the extension, I feel that wasn't so clear before.

### DIFF
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -133,7 +133,7 @@ service:
 
 Replace the following:
 
-- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file.
+- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file or directory. If you provide a directory, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files in it and loads them as a single configuration source. See the [run command reference](../../reference/cli/run/) for details.
 - _`<USERNAME>`_: Your username. If you're using Grafana Cloud, this is your Grafana Cloud instance ID.
 - _`<PASSWORD>`_: Your password. If you're using Grafana Cloud, this is your Grafana Cloud API token.
 - _`<URL>`_: The URL to export data to. If you're using Grafana Cloud, this is your Grafana Cloud OTLP endpoint URL.
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -150,7 +150,7 @@ The output of both engines is visible in the logs.
 You can access the {{< param "DEFAULT_ENGINE" >}} UI and metrics on port `12345`.
 
 {{< admonition type="warning" >}}
-Only one `alloyengine` extension can be active per process, please ensure you only specify a single extension in your `service.extensions` definition. If you specify multiple, you will see a clear error indicating that only one alloy engine extension can be live at once per collector instance. 
+Only one `alloyengine` extension can be active per process.
 {{< /admonition >}}
 
 ## Run with the OpenTelemetry Collector Helm chart
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -150,7 +150,7 @@ The output of both engines is visible in the logs.
 You can access the {{< param "DEFAULT_ENGINE" >}} UI and metrics on port `12345`.
 
 {{< admonition type="warning" >}}
-Only one `alloyengine` extension can be active per process, please ensure you only specify a single extension in your `service.pipelines` definition. If you specify multiple, you will see a clear error indicating that only one alloy engine extension can be live at once per collector instance. 
+Only one `alloyengine` extension can be active per process, please ensure you only specify a single extension in your `service.extensions` definition. If you specify multiple, you will see a clear error indicating that only one alloy engine extension can be live at once per collector instance. 
 {{< /admonition >}}
 
 ## Run with the OpenTelemetry Collector Helm chart
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -149,6 +149,10 @@ This command starts both the {{< param "DEFAULT_ENGINE" >}} and {{< param "OTEL_
 The output of both engines is visible in the logs.
 You can access the {{< param "DEFAULT_ENGINE" >}} UI and metrics on port `12345`.
 
+{{< admonition type="warning" >}}
+Only one `alloyengine` extension can be active per process, please ensure you only specify a single extension in your `service.pipelines` definition. If you specify multiple, you will see a clear error indicating that only one alloy engine extension can be live at once per collector instance. 
+{{< /admonition >}}
+
 ## Run with the OpenTelemetry Collector Helm chart
 
 Use the upstream [OpenTelemetry Collector Helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector) to run the {{< param "OTEL_ENGINE" >}}.
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -133,7 +133,9 @@ service:
 
 Replace the following:
 
-- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file or directory. If you provide a directory, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files in it and loads them as a single configuration source. See the [run command reference](../../reference/cli/run/) for details.
+- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file or directory.
+  If you provide a directory, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files in it and loads them as a single configuration source.
+  Refer to the [run command reference](../../reference/cli/run/) for more information.
 - _`<USERNAME>`_: Your username. If you're using Grafana Cloud, this is your Grafana Cloud instance ID.
 - _`<PASSWORD>`_: Your password. If you're using Grafana Cloud, this is your Grafana Cloud API token.
 - _`<URL>`_: The URL to export data to. If you're using Grafana Cloud, this is your Grafana Cloud OTLP endpoint URL.
--- a/extension/alloyengine/extension.go
+++ b/extension/alloyengine/extension.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.opentelemetry.io/collector/component"
@@ -17,6 +18,10 @@ import (
 )
 
 var _ extension.Extension = (*alloyEngineExtension)(nil)
+
+// running tracks whether any alloyengine instance is currently active.
+// Only one instance can be active per process
+var running = &atomic.Bool{}
 
 type state int
 
@@ -86,19 +91,33 @@ func newAlloyEngineExtension(config *Config, settings component.TelemetrySetting
 }
 
 func (e *alloyEngineExtension) Start(ctx context.Context, host component.Host) error {
+	var startErr error
+	defer func() {
+		if startErr != nil {
+			running.Store(false)
+		}
+	}()
+
 	currentState := e.getState()
 	switch currentState {
 	case stateNotStarted:
 		break
 	default:
-		return fmt.Errorf("cannot start alloyengine extension in current state: %s", currentState)
+		startErr = fmt.Errorf("cannot start alloyengine extension in current state: %s", currentState)
+		return startErr
+	}
+
+	// The Default Engine uses process-global state, so we only want one instance per process.
+	if !running.CompareAndSwap(false, true) {
+		startErr = fmt.Errorf("only one alloyengine extension can be active per process; an instance is already running")
+		return startErr
 	}
 
 	runCommand := e.runCommandFactory()
 	runCommand.SetArgs([]string{e.config.AlloyConfig.File})
-	err := runCommand.ParseFlags(e.config.flagsAsSlice())
-	if err != nil {
-		return fmt.Errorf("failed to parse flags: %w", err)
+	if err := runCommand.ParseFlags(e.config.flagsAsSlice()); err != nil {
+		startErr = fmt.Errorf("failed to parse flags: %w", err)
+		return startErr
 	}
 
 	runCtx, runCancel := context.WithCancel(context.Background())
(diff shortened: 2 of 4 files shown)


extension/alloyengine/extension_test.go
```diff
@@ -143,6 +143,23 @@ func TestLifecycle_StartTwiceFails(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestLifecycle_SecondInstanceFailsWhileFirstRunning(t *testing.T) {
+	ext1 := newTestExtension(t, blockingCommand, defaultTestConfig())
+	ext2 := newTestExtension(t, blockingCommand, defaultTestConfig())
+
+	require.NoError(t, ext1.Start(context.Background(), componenttest.NewNopHost()))
+	require.Eventually(t, func() bool { return ext1.getState() == stateRunning }, 1*time.Second, 25*time.Millisecond, "first extension did not reach stateRunning")
```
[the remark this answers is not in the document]
blewis12: huh, that wasn't introduced in this PR but is valid

extension/alloyengine/extension.go
```diff
@@ -86,19 +90,33 @@ func newAlloyEngineExtension(config *Config, settings component.TelemetrySetting
...
-		return fmt.Errorf("cannot start alloyengine extension in current state: %s", currentState)
+		startErr = fmt.Errorf("cannot start alloyengine extension in current state: %s", currentState)
+		return startErr
+	}
+
+	// Here we check if another extension instance is already running, if so we return an error
+	if !running.CompareAndSwap(false, true) {
+		startErr = fmt.Errorf("only one alloyengine extension can be active per process; an instance is already running")
+		return startErr
 	}
```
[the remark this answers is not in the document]
blewis12: if `CompareAndSwap` fails then we're going to shut down the entire collector, so setting `running.Store(false)` holds

Invocation of Start() is inside the collector service and managed upstream, if we're calling Start twice that's a collector bug - I don't think we should be handling that

extension/alloyengine/extension_test.go
```diff
@@ -141,6 +141,28 @@ func TestLifecycle_StartTwiceFails(t *testing.T) {
 	require.NoError(t, e.Start(context.Background(), componenttest.NewNopHost()))
 	err := e.Start(context.Background(), componenttest.NewNopHost())
 	require.Error(t, err)
+
+	// Cleanup: release the singleton slot so subsequent tests can start extensions
```
blewis12: (this was a pre-existing issue with tests fixed here)

docs/sources/set-up/otel_engine.md
```diff
@@ -149,6 +149,10 @@ This command starts both the {{< param "DEFAULT_ENGINE" >}} and {{< param "OTEL_
 The output of both engines is visible in the logs.
 You can access the {{< param "DEFAULT_ENGINE" >}} UI and metrics on port `12345`.
 
+{{< admonition type="warning" >}}
+Only one `alloyengine` extension can be active per process, please ensure you only specify a single extension in your `service.extensions` definition. If you specify multiple, you will see a clear error indicating that only one alloy engine extension can be live at once per collector instance.
```
ptodev:
```suggestion
Only one `alloyengine` extension can be active per process.
```

It's not necessary to go in so much detail 😄 The 2nd sentence states the same as the 1st. The 3rd one is self explanatory.

ptodev: Also, @clayton-cornell doesn't allow us to say "please" 😛

extension/alloyengine/extension.go
```diff
@@ -86,19 +90,32 @@ func newAlloyEngineExtension(config *Config, settings component.TelemetrySetting
...
 
 	runCommand := e.runCommandFactory()
 	runCommand.SetArgs([]string{e.config.AlloyConfig.File})
-	err := runCommand.ParseFlags(e.config.flagsAsSlice())
-	if err != nil {
-		return fmt.Errorf("failed to parse flags: %w", err)
+	if err := runCommand.ParseFlags(e.config.flagsAsSlice()); err != nil {
+		startErr = fmt.Errorf("failed to parse flags: %w", err)
+		return startErr
```
ptodev:
```suggestion
		return fmt.Errorf("failed to parse flags: %w", err)
```

Why not just return it?

extension/alloyengine/extension_test.go
ptodev: Why not use defer for cancel and for e.shutdown? The defer can be called at the beginning of the test.

docs/sources/set-up/otel_engine.md
blewis12: Sure if you reckon that still reads well 👍 didn't know how thorough to be

extension/alloyengine/extension.go
blewis12: it's there because `startErr` is used later with:

```
defer func() {
		if startErr != nil {
			running.Store(false)
		}
	}()
```

If there was an error on startup, then the extension is not actually running so we set `running.Store(false)`. This is what the pprof extension also does, I wanted to follow a [similar pattern](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/acb90fdf6aa319b6aaf717d2895b95f843e59eba/extension/pprofextension/pprofextension.go#L45-L47)

extension/alloyengine/extension_test.go
blewis12: could do that too - is there an upside to doing that here?

ptodev: In general it's good to instruct the cleanup as soon as the resources are allocated. In this case this happens after the first `e.Start`. We can't do all the cleanup in a `t.Cleanup` since we need to use `require.NoError` in it, so it'd have to be a `defer`. I think that's ok - `t.Cleanup` seems to be more useful for more complex situations like parallel tests or for freeing up resources that may otherwise persist on the machine (like a file that really needs to be deleted even if the test ends abruptly).

That said, I'm confused about this code since it the production code doesn't behave the way I'd expect. `e.Shutdown` should not need a context to be canceled in order to shut down.  Not sure if the reason why the test code is convoluted is that the production code may be convoluted?

I'd expect either of those mechanisms for shutting down:
1. Call `Shutdown` without needing a context to also be canceled for it to do the shut down.
2. No `Shutdown` function. Instead just cancel the context passed to `Start`.

This is 1):
```
	err := e.Start(t.Context(), componenttest.NewNopHost())
	require.NoError(t, err)

	defer require.NoError(t, e.Shutdown(2*time.Second)

	err = e.Start(t.Context(), componenttest.NewNopHost())
	require.Error(t, err)
```

This is 2):
```
	err := e.Start(t.Context(), componenttest.NewNopHost())
	require.NoError(t, err)

	err = e.Start(t.Context(), componenttest.NewNopHost())
	require.Error(t, err)
```

In this second case you don't need special cleanup operations because the test suite will cancel the context when the test end.

blewis12: You're right, it is quite convoluted. We have three contexts:

1. `ctx`: passed through to Start() - unused. It's [documented](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L34-L38) that if you want to have a long running operation you should _not_ use this context and instead create your own (`runCtx` in this case) that is cancelled in the Shutdown method
2. `runCtx`: created in Start() and used to assign the runCancel func. it's passed to `runCommand.ExecuteContext` so it manages the Default engine lifecycle
3. `ctx`: passed through to Shutdown(), used as a safety mechanism if for some reason the default engine hangs despite `runCtx` being cancelled, which we mean we cannot shut it down and thus `runExited` is blocked


Cancelling the 3rd context isn't necessary in tests, so we can definitely remove that to make it simpler. I'd prefer though to keep Shutdown out of a `defer` statement here though because it shows a clear order of operations, and allows us to check for the `terminated` state after we call Shutdown

I've just pushed some changes now  - let me know what you reckon

extension/alloyengine/extension.go
```diff
@@ -85,20 +89,33 @@ func newAlloyEngineExtension(config *Config, settings component.TelemetrySetting
 	}
 }
 
-func (e *alloyEngineExtension) Start(ctx context.Context, host component.Host) error {
+func (e *alloyEngineExtension) Start(_ context.Context, host component.Host) error {
```
ptodev, line 92: Should `running` be set to `true` in the beginning to avoid race conditions? Or we could use a mutex instead?

extension/alloyengine/extension_test.go
```diff
@@ -119,10 +119,7 @@ func TestLifecycle_SuccessfulStartAndShutdown(t *testing.T) {
 	require.NoError(t, e.Ready())
 	require.NoError(t, e.NotReady())
 
-	// Perform graceful shutdown with timeout to avoid hanging tests.
-	shutdownCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
-	t.Cleanup(cancel)
-	require.NoError(t, e.Shutdown(shutdownCtx))
+	require.NoError(t, e.Shutdown(context.Background()))
```
ptodev: I believe that in general `t.Context()` is preferred to `t.Background()` in tests.

extension/alloyengine/extension_test.go
ptodev: Oh ok, I didn't realise this is part of an OTel interface and that it requires 2 different contexts.

In the new code I'm just not sure if `require.NoError(t, e.Shutdown(shutdownCtx))` will be called on timeout, because I think `t.Cleanup(cancel)` runs after that `require`.

docs/sources/set-up/otel_engine.md
```diff
@@ -133,7 +133,7 @@ service:
 
 Replace the following:
 
-- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file.
+- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file or directory. If you provide a directory, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files in it and loads them as a single configuration source. See the [run command reference](../../reference/cli/run/) for details.
```
clayton-cornell:
```suggestion
- _`<ALLOY_CONFIG_PATH>`_: The path to your {{< param "DEFAULT_ENGINE" >}} configuration file or directory.
  If you provide a directory, {{< param "PRODUCT_NAME" >}} finds `*.alloy` files in it and loads them as a single configuration source.
  Refer to the [run command reference](../../reference/cli/run/) for more information.
```

extension/alloyengine/extension.go
blewis12: `running` already uses an [atomic.Bool](https://github.com/grafana/alloy/pull/5763/changes#diff-6402cb86c7951e02ab8812eca363e29799581a9a1ca5a28ec385d8a79101c696R23) type so it should be able to handle race conditions. But i'm i've implemented this in a way that will set this after flags have been parsed/validated so we don't set `running` to `true` unless we're _actually_ running

extension/alloyengine/extension_test.go
blewis12: ah ok - good to know, i'll change that 👍

extension/alloyengine/extension_test.go
blewis12: I think I see what you mean, `t.Cleanup(cancel)` isn't really necessary here because `Shutdown` will already hit [this line](https://github.com/grafana/alloy/blob/d777ed157c4d8df15c4298dbc7f7b6ab05a94975/extension/alloyengine/extension.go#L176) and initiate shutdown, it'll never hit [this line](https://github.com/grafana/alloy/blob/d777ed157c4d8df15c4298dbc7f7b6ab05a94975/extension/alloyengine/extension.go#L178) because that context is cancelled once the test finishes and everything is already exited. I'll update the test to remove that cleanup and avoid confusion 

Edit: yeah we don't even need a context with a timeout, we can also just use `t.Context()`

blewis12: https://github.com/grafana/alloy/pull/5763

extension/alloyengine/extension.go
ptodev: Ok, having an atomic bool is fine. It might be better to delete the defer statement from Start altogether? It'd make it easier to reason about edge cases.

For example:
* Is there a bug if there is already an alloy extension running that set "running" to true, and then the config is refreshed and startErr set to something non-nil causes "running" to become "false"?
* The "only one alloyengine extension can be active" error on line 118 is also technically a start error but startErr isn't set there.

docs/sources/set-up/otel_engine.md
clayton-cornell: Has this been fixed in another commit? The text I see still isn't aligning well with Grafana styles.

blewis12: hum, I thought I did - but will commit this suggestion nwo

blewis12: (I might have forced pushed over it, in hindsight)

extension/alloyengine/extension.go
blewis12: that's a good point, I think we can just do: https://github.com/grafana/alloy/pull/5763/commits/66191300016bff20597b308f64bcc856b3d04a58

wdyt?

extension/alloyengine/extension_test.go
kgeckhart, line 1: 🤔 IIUC we are mutating a global on startup or shutdown. If you want test isolation you might want to add a `t.Cleanup` which will shutdown any started component. Otherwise, when a require fails it might not actually shutdown the component and cause all the other tests to fail unnecessarily.

kgeckhart (approved):
I think the implementation makes sense, just the comment related to test isolation.

PR #5769, opened 2026-03-12
feat(pyroscope.ebpf): Add comm, pid labels and kernel frame options

korniltsev-grafanista:
## Summary
- Add configurable `comm` option (none/label/stackframe/both) to control how process command name appears in generated pprof profiles
- Add `pid_label` option to include PID as a pprof sample label
- Add `kernel_frames` option to include kernel frames in profiles (default: true)
- All three options are undocumented, to address a Grafana Labs customer's escalation
- Add test binary and symbolization tests for PLT/zero-size function resolution
- Move `CommMode` type to `reporter/args` package for cross-platform build support (fixes Windows CI)
- Validate `comm` mode at config time — invalid values (e.g. typos like `"labels"`) now fail fast instead of silently behaving as `"none"`

## Test plan
- [ ] Run `pyroscope.ebpf` with `comm = "label"` and verify comm appears as sample label
- [ ] Run with `comm = "stackframe"` and verify comm appears as root frame
- [ ] Run with `comm = "both"` and verify both behaviors
- [ ] Run with `kernel_frames = false` and verify kernel frames are excluded
- [ ] Verify invalid `comm` values (e.g. `"labels"`) are rejected at config validation
- [ ] `GOOS=windows go build ./internal/component/pyroscope/ebpf/reporter/...` succeeds
- [ ] Run existing unit tests

### DIFF
--- a/internal/component/pyroscope/ebpf/args.go
+++ b/internal/component/pyroscope/ebpf/args.go
@@ -30,13 +30,13 @@ type Arguments struct {
 	DeprecatedArguments DeprecatedArguments    `alloy:",squash"`
 
 	// undocumented
-	PyroscopeDynamicProfilingPolicy bool     `alloy:"targets_only,attr,optional"`
-	SymbCachePath                   string   `alloy:"symb_cache_path,attr,optional"`
-	SymbCacheSizeEntries            int      `alloy:"symb_cache_size,attr,optional"`
-	ReporterUnsymbolizedStubs       bool     `alloy:"reporter_unsymbolized_stubs,attr,optional"`
-	PIDLabel                        bool     `alloy:"pid_label,attr,optional"`
+	PyroscopeDynamicProfilingPolicy bool   `alloy:"targets_only,attr,optional"`
+	SymbCachePath                   string `alloy:"symb_cache_path,attr,optional"`
+	SymbCacheSizeEntries            int    `alloy:"symb_cache_size,attr,optional"`
+	ReporterUnsymbolizedStubs       bool   `alloy:"reporter_unsymbolized_stubs,attr,optional"`
+	PIDLabel                        bool   `alloy:"pid_label,attr,optional"`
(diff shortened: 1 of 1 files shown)


internal/component/pyroscope/ebpf/reporter/pprof.go
```diff
@@ -282,6 +320,18 @@ func (p *PPROFReporter) createProfile(containerID samples.ContainerID, origin li
 			}
 			s.Location = append(s.Location, location)
 		}
+		if comm != "" && p.cfg.CommMode.stackframe() {
+			commLocation := &profile.Location{
```
korniltsev-grafanista: do we need to do this on each trace? can't we lazy create one?

korniltsev-grafanista: Good point, moved comm location creation to a lazy-cached `CommLocation(comm)` method on `ProfileBuilder` — now it's created once per unique comm string per builder.

korniltsev-grafanista:
@cursor-bugbot review

internal/component/pyroscope/ebpf/args.go
```diff
@@ -33,6 +34,9 @@ type Arguments struct {
 	SymbCachePath                   string `alloy:"symb_cache_path,attr,optional"`
 	SymbCacheSizeEntries            int    `alloy:"symb_cache_size,attr,optional"`
 	ReporterUnsymbolizedStubs       bool   `alloy:"reporter_unsymbolized_stubs,attr,optional"`
+	PIDLabel                        bool   `alloy:"pid_label,attr,optional"`
+	Comm                            string `alloy:"comm,attr,optional"` // to address a Grafana Labs customer's escalation
```
marcsanmi, line 38: maybe?
```suggestion
	Comm                            string `alloy:"comm,attr,optional"`
```

korniltsev-grafanista: no, i put this comment so that we do not remove this in the future

PR #5779, opened 2026-03-13
feat(pyroscope): Replace Parca gRPC debuginfo upload with Pyroscope Connect API

korniltsev-grafanista-yolo-vibecoder239:
## Summary

Replace the Parca gRPC-based debuginfo/symbol upload with the new Pyroscope Connect bidirectional streaming API ([pyroscope#4648](https://github.com/grafana/pyroscope/pull/4648)). This simplifies the upload protocol from a 4-step gRPC flow (`ShouldInitiateUpload` → `InitiateUpload` → `Upload` → `MarkUploadFinished`) to a single bidi stream (`Upload` with init/chunk messages).

- Replace `ParcaSymbolUploader` with `PyroscopeSymbolUploader` using Connect bidi streaming
- Replace `Client() debuginfogrpc.DebuginfoServiceClient` with `ConnectClient()`/`ConnectClients()` across the `Appender` interface hierarchy
- Rewrite `receive_http` debuginfo proxy to fan out uploads to **all** downstream Connect clients (not just the first one)
- Add undocumented `symbol_upload_enabled` config knob to `pyroscope.ebpf` (disabled by default, for internal use)
- Remove all Parca gRPC dependencies (`buf.build/gen/go/parca-dev/parca`)
- Bump `github.com/grafana/pyroscope/api` to the `debuginfo-upload` branch

### New upload protocol (Connect bidi stream)

The new protocol uses a single `Upload` RPC with bidirectional streaming:
1. Client sends `ShouldInitiateUploadRequest` with `FileMetadata` (build ID, file name, type)
2. Server responds with `ShouldInitiateUploadResponse` (boolean + reason)
3. If approved, client streams `UploadChunk` messages (3MB chunks)
4. Client closes send side to signal EOF

### receive_http proxy fan-out

The `receive_http` component now collects **all** Connect clients from all downstream appendables via `ConnectClients()` and replicates uploads to every endpoint that responds with `should_initiate_upload = true`. Previously the proxy only forwarded to the first available gRPC client.

### Configuration

New undocumented attribute in `pyroscope.ebpf`:
```
symbol_upload_enabled = true  // default: false
```

When enabled, `ReportExecutable` calls `reportExecutableForDebugInfoUpload` which dispatches upload jobs through the appendable chain to `pyroscope.write` endpoints.

### Files changed

**Deleted:**
- `reporter/grpc_upload_client.go` — old Parca gRPC streaming upload
- `reporter/parca_uploader.go` — old Parca symbol uploader  
- `write/debuginfo_client.go` — old gRPC connection factory
- `receive_http/grpc.go` — old gRPC server helper

**Created:**
- `reporter/pyroscope_uploader.go` — new Connect-based uploader (linux/amd64+arm64)
- `reporter/pyroscope_uploader_stub.go` — non-linux stub

**Renamed:**
- `write/debuginfo/parca.go` → `write/debuginfo/upload.go`

## Test plan

- [x] `go build ./internal/component/pyroscope/...` compiles
- [x] `go test ./internal/component/pyroscope/...` passes (all unit tests green)
- [x] `go mod tidy` clean — no parca dependencies remain
- [ ] Manual test with `symbol_upload_enabled = true` against a Pyroscope instance with debuginfo support
- [ ] Verify upload is **not** triggered when `symbol_upload_enabled` is omitted (default false)


internal/component/pyroscope/ebpf/reporter/parca/reporter/pyroscope_uploader_test.go
```diff
@@ -3,12 +3,99 @@
 package reporter
 
 import (
```
korniltsev-grafanista, line 5: rename this file

internal/component/pyroscope/ebpf/reporter/parca/reporter/pyroscope_uploader_stub.go
```diff
@@ -0,0 +1,6 @@
+//go:build !(linux && (arm64 || amd64))
+
+package reporter
+
+// This file provides a no-op stub for platforms that don't support eBPF profiling.
```
korniltsev-grafanista: why do we need this? it used to work before

internal/component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -304,6 +309,11 @@ func (c *Component) reportExecutableForDebugInfoUpload(args *reporter2.Executabl
 	c.appendable.Upload(debuginfo.UploadJob{
 		FrameMappingFileData: mf,
 		Open:                 open,
+		InitArguments: debuginfo.Arguments{
```
korniltsev-grafanista: put this into ebpf component args, pass the debuginfo.Arguments as is from ebpf args down to job.

internal/component/pyroscope/write/write.go
```diff
@@ -755,6 +754,22 @@ func validateLabels(lbls labels.Labels) error {
 	return err
 }
 
+// newHTTP2Client wraps an existing HTTP client's transport with HTTP/2 support.
+// For HTTPS, HTTP/2 is negotiated via ALPN automatically. For plain HTTP (h2c),
+// we use the http2 transport with AllowHTTP and a custom dialer.
+func newHTTP2Client(base *http.Client) *http.Client {
```
korniltsev-grafanista: this is garbage. is there no way to configure h2 with 		commonconfig.NewClientFromConfig?

internal/component/pyroscope/write/debuginfo/common.go
```diff
@@ -4,21 +4,19 @@ import (
...
-	//down to Client and therefore parca's uploader
+	// down to Client and therefore the uploader.
 	Upload(j UploadJob)
-	// Client returns the direct grpc client of the first nested child (down to write component)
-	// this is a best-effort support for the proxy case - we do not support fan-out for proxy only one-to-one
-	// forwarding to the first write endpoint.
-	Client() debuginfogrpc.DebuginfoServiceClient
+	// ConnectClients returns ALL Connect debuginfo clients from all nested children.
+	// This is used by the receive_http proxy to fan-out uploads to all downstream endpoints.
+	ConnectClients() []debuginfov1alpha1connect.DebuginfoServiceClient
```
korniltsev-grafanista: rename to DebugInfoClients

internal/component/pyroscope/write/write.go
korniltsev-grafanista: this still looks like garbage and idk how to make it nice

korniltsev-grafanista-yolo-vibecoder239 (commented):
## Code Review

Overall this is a well-structured migration from Parca gRPC to Pyroscope Connect bidi streaming. Clean dependency removal, good fan-out upgrade, solid test coverage. A few issues worth addressing:

---

### HIGH

#### 1. Bidi stream never fully closed — `CloseResponse()` / final response never read

**`pyroscope_uploader.go:attemptUpload`** and **`receive_http/debuginfo.go:Upload`**

After sending all chunks and calling `stream.CloseRequest()`, neither the uploader nor the proxy ever calls `stream.CloseResponse()` or reads the server's final response. This means:
- The server may send a final error/confirmation that goes unread → **silent upload failures**
- HTTP/2 streams may leak (client holds the receive half-open)

The Connect bidi protocol expects the client to close both sides.

**Fix:** After `CloseRequest()`, drain the server response (or at minimum call `CloseResponse()`) in both files.

---

#### 2. Proxy: all downstreams failing init silently returns "upload not needed"

**`receive_http/debuginfo.go`**, init loop (~lines 53-73 in the new code)

When ALL downstream clients fail during the init handshake (Send or Receive errors), the code `continue`s past each. Result: `downstreams` is empty, `anyShouldUpload` is `false`, proxy sends back `ShouldInitiateUpload: false` — telling the caller "no upload needed" when all downstreams are unreachable. Silent data loss.

**Fix:** After the init loop, check `len(downstreams) == 0 && len(clients) > 0` and return `connect.NewError(connect.CodeUnavailable, ...)`.

---

#### 3. h2c client drops auth headers and tracing from base client

**`write/write.go:newH2CClient()`**

The h2c client creates a fresh `http2.Transport` that does NOT inherit the base client's transport chain:
- **No auth headers** (basic auth, bearer token from `HTTPClientConfig`)
- **No tracing** (base transport is wrapped with `otelhttp.NewTransport`)
- **No custom dialer** (proxy settings, `DialContext` from base)

Only `Timeout`, `CheckRedirect`, and `Jar` are copied. Debuginfo uploads over plain HTTP will lack authentication.

**Fix:** Wrap the h2c transport in the same tracing/auth middleware, or document this as a known limitation with a TODO.

---

### MEDIUM

#### 4. No stream cleanup on early error returns in `attemptUpload()`

**`pyroscope_uploader.go`** — when `attemptUpload()` returns early (e.g., `stream.Send()` fails in Step 1), the bidi stream is never closed. No `defer` for cleanup.

**Fix:** Add `defer func() { _ = stream.CloseRequest(); _ = stream.CloseResponse() }()` after `stream := client.Upload(ctx)`.

---

#### 5. Chunk read loop may lose last bytes (fragile `io.EOF` handling)

**`pyroscope_uploader.go`**, Step 4 chunk loop:

```go
n, err := reader.Read(buffer)
if err == io.EOF {
    break  // If n > 0 here, those bytes are lost
}
```

Per `io.Reader` contract, a reader MAY return `n > 0, io.EOF` on the final read. Safe with `bufio.Reader` today, but fragile if the reader changes. Standard pattern:

```go
n, err := reader.Read(buffer)
if n > 0 { /* send chunk */ }
if errors.Is(err, io.EOF) { break }
if err != nil { return ... }
```

---

#### 6. Proxy chunk fan-out silently ignores downstream send errors

**`receive_http/debuginfo.go`**, chunk loop:

```go
_ = ds.stream.Send(req)  // error discarded
```

A downstream that fails mid-upload continues receiving (and failing on) all subsequent chunks. Never removed from `activeStreams`, never logged.

**Fix:** Track per-stream errors atomically and either remove failed streams or log the failure.

---

#### 7. Proxy: non-EOF receive errors treated as success

**`receive_http/debuginfo.go`**:

```go
if err == io.EOF || err != nil {
    break
}
```

If upstream disconnects mid-stream (transport error, not EOF), the proxy breaks and returns `nil` — success. Partial uploads to downstreams are never flagged.

**Fix:** Differentiate `io.EOF` (normal) from other errors (propagate).

---

### LOW / nits

- **`parca` directory name** is now misleading — contains exclusively Pyroscope code. Follow-up rename recommended.
- **`readAtCloserSize` returning `(0, nil)`** for non-Stater readers → permanent skip via retry cache. Pre-existing.
- **Fragile string matching**: `err.Error() == "no backing file for anonymous memory"` and `initResp.Reason == ReasonUploadInProgress`. Pre-existing pattern.
- **`context.Background()` in upload path** (`upload.go`) — no cancellation possible for individual uploads.
- **No test for `stripTextSection=true` path** (temp file + `elfwriter.OnlyKeepDebug()`).

### Positive observations

- Clean dependency removal — all `buf.build/gen/go/parca-dev/parca` gone from go.mod, go.sum, and source
- `DebugInfoClients() []DebuginfoServiceClient` is the right design for fan-out
- Good integration tests with real HTTP/2 TLS test servers
- Thread safety properly handled throughout
- Safe defaults (`UploadEnabled: false`)
- Smart vdso skip

PR #5786, opened 2026-03-13
fix(loki.secretfilter): Fix bug where entries were being shadow dropped

mikefat:

This fixes a bug in secretfilter so that processed entries are correctly forwarded.
Previously, line 289 declared `newEntry` but inside the `if` conditional for sampling `newEntry` was getting shadowed and therefore processed entries were being forwarded empty.
A test has been added for regression and debug logs have been added for the component at various places.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/secretfilter/secretfilter.go
+++ b/internal/component/loki/secretfilter/secretfilter.go
@@ -274,7 +274,6 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 	level.Debug(c.log).Log(
 		"msg", "loki.secretfilter initialized",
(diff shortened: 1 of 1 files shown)


internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -287,9 +303,11 @@ func (c *Component) Run(ctx context.Context) error {
 			c.mut.RLock()
 
 			var newEntry loki.Entry
+			var dropped bool
```
kleimkuhler: Nit: we don't use `dropped` outside of the `if`, so it could be moved into that block

PR #5799, opened 2026-03-17
perf(discovery.process): Use BootTimeCache to avoid re-reading btime value

luweglarz:
### Pull Request Details

Currently we are running Alloy using the discovery.process component, and I have noticed a high usage of cpu on the discovery.process part
argos-alloy_process_cpu_cpu_nanoseconds_cpu_nanoseconds_2026-03-17_1018-to-2026-03-17_1048(1)
mostly on the `readBootTimeStat`  method
argos-alloy_process_cpu_cpu_nanoseconds_cpu_nanoseconds_2026-03-17_1018-to-2026-03-17_1048(2) 
The gopsutil library used for process discovery allows to use a cache for the `btime` value which can lead to CPU usage reduction.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


internal/component/discovery/process/process.go
```diff
@@ -12,9 +12,12 @@ import (
 	"github.com/grafana/alloy/internal/component/discovery"
 	"github.com/grafana/alloy/internal/featuregate"
 	"github.com/grafana/alloy/internal/service/livedebugging"
+	gopsutil "github.com/shirou/gopsutil/v3/process"
 )
 
 func init() {
+	gopsutil.EnableBootTimeCache(true)
```
kalleep, line 19: I really don't like that this is a global config but it is what it is.

Issue #5803, opened 2026-03-17
loki: align implementations for sources that uses servers


kalleep:
### Add a description

We have a couple of loki sources that exposes http server:
* loki.source.api
* loki.source.firehose
* loki.source.gcplog
* loki.source.heroku

We should align server implementations for these so that we can:
1. Prepare a bit for https://github.com/grafana/alloy/pull/4940
2. Create a shared server abstraction for these.

They should all forward batches to the components and implement the same  shutdown semantics we have for `loki.source.api`.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5804, opened 2026-03-17
fix(loki.source.heroku): Fix shutdown semantics and consume logs in batches

kalleep:
### Pull Request Details

This pr change to use the shared `loki.Fanout` and `source.ConsumeBatch`. Like `loki.source.api` the server we create now forward batches of entries _and_ implement the same shutdown semantics.

When we reload we stop the server and try to wait for in-flight request to finish. When stopping component we abort all in-flight request directly. 

Part of https://github.com/grafana/alloy/issues/5803

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/api/internal/lokipush/push_api_server_test.go
+++ b/internal/component/loki/source/api/internal/lokipush/push_api_server_test.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"net/http"
 	"strconv"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/component/common/loki/receiver.go
```diff
@@ -65,3 +67,51 @@ type logsBatchReceiver struct {
...
+			c.mtx.Unlock()
+		}
+	})
+	return c
+}
+
+// CollectingBatchReceiver is a LogsBatchReceiver that will
+// collect all received entries so it can later be inspected.
+// Used in tests.
+type CollectingBatchReceiver struct {
```
kgeckhart, line 88: Does it make sense to move this in to one of the test packages?

kalleep: Yeah maybe, currently I am just preparing to rebuild the pipeline so these should eventually be removed

PR #5813, opened 2026-03-18
fix(loki.source.file): Add file path to error output for windows

mbaykara:
### Pull Request Details
Add path for file read error for windows to increase verbosity and added corresponding test

### PR Checklist

- [ x] Tests updated

### DIFF
--- a/internal/component/loki/source/file/internal/tail/fileext/file_posix_test.go
+++ b/internal/component/loki/source/file/internal/tail/fileext/file_posix_test.go
@@ -1,0 +1,24 @@
+//go:build linux || darwin || freebsd || netbsd || openbsd
+
+package fileext
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


mbaykara:
hi @kalleep this is the error we get on windows.
`component_id=loki.source.file.inhouse_classic_logs component=tailer err="failed to tail the file: The file cannot be accessed by the system."`
it is not showing failed path. Or am I missing something? 👀

kalleep:
> hi @kalleep this is the error we get on windows. `component_id=loki.source.file.inhouse_classic_logs component=tailer err="failed to tail the file: The file cannot be accessed by the system."` it is not showing failed path. Or am I missing something? 👀

On what version of alloy?

kalleep:
It was added in https://github.com/grafana/alloy/pull/5139 and is included from v.1.13.0

mbaykara:
in our case it is -> `v1.10.2`

kalleep:
Okay so this is already fixed in a later version and this pr won't change that. So lets close it?

Issue #5806, opened 2026-03-17
prometheus.enrich: Support multi-label matching


dtrejod:
### Component(s)

prometheus.enrich

## Problem

`prometheus.enrich` currently supports matching on a single label pair (`target_match_label` / `metrics_match_label`). In practice, enriching metrics with Kubernetes metadata requires matching on multiple labels simultaneously — for example, matching cadvisor metrics to discovery targets by namespace + pod + container.

### Current workaround

The workaround requires chaining 3 components to synthesize composite keys, implement cascading match logic, then dropping the composite key:

```alloy
// 1. Create composite keys on discovery targets
discovery.relabel "k8s_meta_enrich_keys" {
  targets = discovery.kubernetes.pods.targets
  rule {
    source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_name", "__meta_kubernetes_pod_container_name"]
    separator     = "/"
    target_label  = "__enrich_container_key"
    action        = "replace"
  }
}

// 2. Second pass: container-level enrichment (overwrites with precise match)
prometheus.enrich "container_level" {
    targets             = discovery.relabel.k8s_meta_enrich_keys.output
    target_match_label  = "__enrich_container_key"
    metrics_match_label = "__enrich_container_key"
    forward_to          = [prometheus.relabel.drop_composite_key.receiver]
}

// 3. Drop the composite key 
discovery.relabel "drop_composite_key" {
  rule {
    action = "labeldrop"
    regex = "__enrich_.*"
  }
  ...
}
```

This workaround is verbose — it requires synthesizing composite keys on both sides (discovery targets and metrics) before the actual enrichment can happen. Native multi-label matching would eliminate the `discovery.relabel` and `prometheus.relabel` steps entirely.

## Proposed solution

Add a new `target_to_metric_match` attribute that accepts a map of target label name → metric label name. All entries in the map must match for enrichment to occur.

```alloy
prometheus.enrich "container_level" {
    targets = discovery.kubernetes.pods.targets

    target_to_metric_match = {
        "__meta_kubernetes_namespace"          = "namespace",
        "__meta_kubernetes_pod_name"           = "pod",
        "__meta_kubernetes_pod_container_name" = "container",
    }

    forward_to = [prometheus.remote_write.default.receiver]
}
```

[...]

PR #5822, opened 2026-03-18
feat(prometheus.enrich): Support multi-label matching

dtrejod:

- Add `target_to_metric_match` attribute to `prometheus.enrich` for multi-label matching 
- The strategy is a `map[string]string` (target label → metric label), making the relationship explicit
- Internally uses FNV-64a hashed composite keys (values only, sorted by target label name) for O(1) cache lookup
- Backward compatible: existing `target_match_label` / `metrics_match_label` continue to work, mutually exclusive with new attribute

https://github.com/grafana/alloy/issues/5806

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.enrich.md
+++ b/docs/sources/reference/components/prometheus/prometheus.enrich.md
@@ -13,17 +13,39 @@ title: prometheus.enrich
 {{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
 The `prometheus.enrich` component enriches metrics with additional labels from service discovery targets.
-It matches a label from incoming metrics against a label from discovered targets, and copies specified labels from the
-matched target to the metric sample. If match doesn't happen, the metrics will be passed through.
+It matches labels from incoming metrics against labels from discovered targets, and copies specified labels from the
+matched target to the metric sample. If no match occurs, the metrics are passed through unchanged.
+
+The component supports two matching modes:
+
+- **Single-label matching (legacy):** Match on a single label pair using `target_match_label` and optionally `metrics_match_label`.
+- **Multi-label matching:** Match on multiple labels simultaneously using `target_to_metric_match`. The map keys are target label names and the values are the corresponding metric label names. All labels in the map must match for enrichment to occur.
 
 ## Usage
+
+### Single-label matching
 
 ```alloy
 prometheus.enrich "<LABEL>" {
   targets = <DISCOVERY_COMPONENT>.targets
-  
+
   target_match_label = "<LABEL>"
-  
+
+  forward_to = [<RECEIVER_LIST>]
+}
+```
+
+### Multi-label matching
+
+```alloy
+prometheus.enrich "<LABEL>" {
+  targets = <DISCOVERY_COMPONENT>.targets
+
+  target_to_metric_match = {
+    "<TARGET_LABEL_1>" = "<METRIC_LABEL_1>",
+    "<TARGET_LABEL_2>" = "<METRIC_LABEL_2>",
+  }
+
   forward_to = [<RECEIVER_LIST>]
 }
 ```
@@ -32,15 +54,21 @@ prometheus.enrich "<LABEL>" {
 
 You can use the following arguments with `prometheus.enrich`:
 
-| Name                  | Type                    | Description                                                                                        | Default | Required |
-|-----------------------|-------------------------|----------------------------------------------------------------------------------------------------|---------|----------|
-| `forward_to`          | `list(MetricsReceiver)` | Where the metrics should be forwarded to, after enrichment.                                        |         | yes      |
-| `target_match_label`  | `string`                | The label from discovered targets to match against, for example, `"__inventory_consul_service"`.   |         | yes      |
-| `targets`             | `list(map(string))`     | List of targets from a discovery component.                                                        |         | yes      |
-| `labels_to_copy`      | `list(string)`          | List of labels to copy from discovered targets to metrics. If empty, all labels are copied.        |         | no       |
-| `metrics_match_label` | `string`                | The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
+| Name                                | Type                           | Description                                                                                        | Default | Required |
+|-------------------------------------|--------------------------------|----------------------------------------------------------------------------------------------------|---------|----------|
+| `forward_to`                        | `list(MetricsReceiver)`        | Where the metrics should be forwarded to, after enrichment.                                        |         | yes      |
+| `targets`                           | `list(map(string))`            | List of targets from a discovery component.                                                        |         | yes      |
+| `target_to_metric_match`            | `map(string)`                  | Map of target label name to metric label name. All entries must match for enrichment.                   |         | no       |
+| `target_match_label`                | `string`                       | The label from discovered targets to match against, for example, `"__inventory_consul_service"`.   |         | no       |
+| `metrics_match_label`               | `string`                       | The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
+| `labels_to_copy`                    | `list(string)`                 | List of labels to copy from discovered targets to metrics. If empty, all labels are copied.        |         | no       |
 
-If not provided, the `metrics_match_label` attribute defaults to the value of `target_match_label`.
+You must specify exactly one of:
+
+- `target_match_label` (with optional `metrics_match_label`) for single-label matching, or
+- `target_to_metric_match` for multi-label matching.
+
+These two modes are mutually exclusive. If not provided, the `metrics_match_label` attribute defaults to the value of `target_match_label`.
 
 ## Blocks
 
@@ -62,9 +90,9 @@ The following values are exported:
 
 * `prometheus_fanout_latency` (histogram): Write latency for sending to direct and indirect components.
 * `prometheus_forwarded_samples_total` (counter): Total number of samples sent to downstream components.
-* `prometheus_target_cache_size` (gauge): The number of targets in the `targets` list.
+* `prometheus_target_cache_size` (gauge): Total number of cached target entries.
 
-## Example
+## Examples
 
 ### Enrich metrics from `prometheus.scrape`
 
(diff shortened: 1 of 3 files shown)


docs/sources/reference/components/prometheus/prometheus.enrich.md
```diff
@@ -13,17 +13,39 @@ title: prometheus.enrich
...
 
 The `prometheus.enrich` component enriches metrics with additional labels from service discovery targets.
-It matches a label from incoming metrics against a label from discovered targets, and copies specified labels from the
-matched target to the metric sample. If match doesn't happen, the metrics will be passed through.
+It matches labels from incoming metrics against labels from discovered targets, and copies specified labels from the
+matched target to the metric sample. If no match occurs, the metrics are passed through unchanged.
+
+The component supports two matching modes:
+
+- **Single-label matching (legacy):** Match on a single label pair using `target_match_label` and optionally `metrics_match_label`.
```
dtrejod: 💭 Given `prometheus.enrich` is marked as experimental, do we want to consider a breaking change where we only support the new syntax going forward?

docs/sources/reference/components/prometheus/prometheus.enrich.md
```diff
@@ -32,15 +54,21 @@ prometheus.enrich "<LABEL>" {
...
-| `metrics_match_label` | `string`                | The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
+| Name                                | Type                           | Description                                                                                        | Default | Required |
+|-------------------------------------|--------------------------------|----------------------------------------------------------------------------------------------------|---------|----------|
+| `forward_to`                        | `list(MetricsReceiver)`        | Where the metrics should be forwarded to, after enrichment.                                        |         | yes      |
+| `targets`                           | `list(map(string))`            | List of targets from a discovery component.                                                        |         | yes      |
+| `target_to_metric_match`            | `map(string)`                  | Map of target label name to metric label name. All entries must match for enrichment.                   |         | no       |
```
clayton-cornell:
```suggestion
| `target_to_metric_match`            | `map(string)`                  | Map of target label name to metric label name. All entries must match for enrichment.              |         | no       |
```

Just aligning the columns (cleaner markdown).

docs/sources/reference/components/prometheus/prometheus.enrich.md
```diff
@@ -32,15 +54,21 @@ prometheus.enrich "<LABEL>" {
...
+| `metrics_match_label`               | `string`                       | The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
+| `labels_to_copy`                    | `list(string)`                 | List of labels to copy from discovered targets to metrics. If empty, all labels are copied.        |         | no       |
 
-If not provided, the `metrics_match_label` attribute defaults to the value of `target_match_label`.
+You must specify exactly one of:
+
+- `target_match_label` (with optional `metrics_match_label`) for single-label matching, or
+- `target_to_metric_match` for multi-label matching.
+
+These two modes are mutually exclusive. If not provided, the `metrics_match_label` attribute defaults to the value of `target_match_label`.
```
clayton-cornell:
```suggestion
You must specify exactly one of the following:

- `target_match_label` (with optional `metrics_match_label`) for single-label matching
- `target_to_metric_match` for multi-label matching.

These two modes are mutually exclusive.

If you don't set the `metrics_match_label`, it defaults to the value of `target_match_label`.
```

docs/sources/reference/components/prometheus/prometheus.enrich.md
```diff
@@ -107,31 +135,71 @@ The following example shows how the `prometheus.enrich` enriches incoming metric
...
+```
+
+### Multi-label matching with Kubernetes metadata
 
-prometheus.remote_write "default" {  
-  endpoint {  
-    url = "http://mimir:9009/api/v1/push"    
-  }    
+The following example enriches cadvisor metrics with Kubernetes pod metadata,
+matching on namespace, pod, and container labels simultaneously.
```
clayton-cornell:
```suggestion
The following example enriches cadvisor metrics with Kubernetes Pod metadata, matching on namespace, Pod, and container labels simultaneously.
```

dtrejod:
@clayton-cornell address your feedback. Thank you!

docs/sources/reference/components/prometheus/prometheus.enrich.md
ptodev: We can make the new mode the preferred one. The old mode can be documented as "deprecated" and can be removed in a future release. The old mode can still be documented but in a less prominent way. We'd need to also instruct users how to migrate by listing examples.

@clayton-cornell I'd say in this paragraph we can just list the multi-label matching? And under the paragraph we can add a `{{< admonition type="note" >}}` to state the old arguments and that they will be removed in the future?

internal/component/prometheus/enrich/enrich.go
```diff
@@ -38,24 +39,103 @@ type Arguments struct {
...
+	}
+	return nil
+}
+
 // Exports holds values which are exported by the prometheus.enrich component.
 type Exports struct {
 	Receiver storage.Appendable `alloy:"receiver,attr"`
 }
 
+// FNV-64a helpers (same algorithm as
```
ptodev: Is there any way to not have to maintain such hashing algorithms in the Alloy codebase? Can we import something from Prometheus?

docs/sources/reference/components/prometheus/prometheus.enrich.md
```diff
@@ -32,15 +54,23 @@ prometheus.enrich "<LABEL>" {
...
+|-------------------------------------|--------------------------------|----------------------------------------------------------------------------------------------------|---------|----------|
+| `forward_to`                        | `list(MetricsReceiver)`        | Where the metrics should be forwarded to, after enrichment.                                        |         | yes      |
+| `targets`                           | `list(map(string))`            | List of targets from a discovery component.                                                        |         | yes      |
+| `target_to_metric_match`            | `map(string)`                  | Map of target label name to metric label name. All entries must match for enrichment.              |         | no       |
+| `target_match_label`                | `string`                       | The label from discovered targets to match against, for example, `"__inventory_consul_service"`.   |         | no       |
+| `metrics_match_label`               | `string`                       | The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
```
ptodev:
```suggestion
| `target_match_label`                | `string`                       | (Deprecated) The label from discovered targets to match against, for example, `"__inventory_consul_service"`.   |         | no       |
| `metrics_match_label`               | `string`                       | (Deprecated) The label from incoming metrics to match against discovered targets, for example `"service_name"`. |         | no       |
```

internal/component/prometheus/enrich/enrich.go
dtrejod: The fnv helpers in [prometheus/common/model/fnv.go](https://github.com/prometheus/common/blob/0dfcdfb00df68e0b14a98f20d90f4b3ff12432e6/model/fnv.go) are unexported so can't be used directly. 

I took a look and noticed [target.go](https://github.com/grafana/alloy/blob/main/internal/component/discovery/target.go) was using a xxhash pkg already so reused it here to avoid the custom hashing implementation.

internal/component/prometheus/enrich/enrich.go
```diff
@@ -38,24 +40,83 @@ type Arguments struct {
...
+// from a model.LabelSet. Returns (0, false) if any label is missing or empty.
+func hashValuesFromLabelSet(ls model.LabelSet, names []string) (uint64, bool) {
+	h := xxhash.New()
+	for _, name := range names {
+		v := string(ls[model.LabelName(name)])
+		if v == "" {
+			return 0, false
+		}
+		_, _ = h.WriteString(v)
+		_, _ = h.Write(sep)
```
dtrejod, line 96: Ignoring the errors here since the error is just to satisfy the io.Writer interface. In reality these funcs don't return errors:
https://github.com/cespare/xxhash/blob/ab37246c889f9db16b606fda1c232d659df9271d/xxhash.go#L75

clayton-cornell (dismissed):
Approving for docs.

dtrejod:
@ptodev are you still the code reviewer for this PR? Can you take another pass?

dtrejod:
@ptodev anything bump on reviewing this PR please 🙏

kgeckhart:
> @ptodev anything bump on reviewing this PR please 🙏

Hey sorry for the delay. @ptodev is OOO ATM, we're looking for someone else to review it while he's out. Might be a little bit but we didn't forget about you.

internal/component/prometheus/enrich/enrich.go
```diff
@@ -173,36 +234,33 @@ func (c *Component) Update(args component.Arguments) error {
 	defer c.mut.Unlock()
 
 	newArgs := args.(Arguments)
+	c.args = newArgs
```
kgeckhart: Looks like there's a subtle data race here now that we are updating args. Line 266 `if len(c.args.LabelsToCopy) == 0 {` doesn't hold any locks when it does the read. 

I think you can avoid this if you snapshot `c.args.LabelsToCopy` on to the matchCache when you rebuild the new instance in `refreshCacheFromTargets`

Either way, it's probably worth it to add a test with a goroutine tightly sending data through enrich and calling update.

internal/component/prometheus/enrich/enrich.go
```diff
@@ -38,24 +40,83 @@ type Arguments struct {
 	// The targets to use for enrichment
 	Targets []discovery.Target `alloy:"targets,attr"`
 
-	// Which label from targets to use for matching (e.g. "hostname", "ip")
-	TargetMatchLabel string `alloy:"target_match_label,attr"`
+	// Multi-label matching: a map of target_label -> metric_label.
+	// Mutually exclusive with target_match_label / metrics_match_label.
```
kgeckhart:
```suggestion
	// Takes precedence over target_match_label / metrics_match_label.
```

I think this is right based on the docs?

internal/component/prometheus/enrich/enrich.go
```diff
@@ -38,24 +40,83 @@ type Arguments struct {
...
 // Exports holds values which are exported by the prometheus.enrich component.
 type Exports struct {
 	Receiver storage.Appendable `alloy:"receiver,attr"`
 }
 
+var sep = []byte{0xff} // separator to prevent hash collisions across value boundaries
+
+// hashValuesFromLabelSet hashes the values of the given label names (in order)
+// from a model.LabelSet. Returns (0, false) if any label is missing or empty.
+func hashValuesFromLabelSet(ls model.LabelSet, names []string) (uint64, bool) {
```
kgeckhart, line 85: Could this return (0, false) if names are empty? It looks like validate protects from a scenario where it could happen but if that is refactored away every metric gets enriched with whatever target happened to be last IIUC.

kgeckhart (commented):
Two nitty things and a data race. Thanks for this!

internal/component/prometheus/enrich/enrich.go
dtrejod: thank you. Yes we changed the behavior and i forgot to update the docstring.

internal/component/prometheus/enrich/enrich.go
dtrejod: nice catch. I saved off a copy to matchCache as you suggested to avoid race

internal/component/prometheus/enrich/enrich.go
```diff
@@ -38,24 +40,90 @@ type Arguments struct {
...
+// matchCache holds the hash-based lookup for a match strategy.
+type matchCache struct {
+	sortedMetricLabels []string                  // metric label names to hash, sorted by corresponding target label name
+	cache              map[uint64]model.LabelSet // hash of values -> target label set
+	labelsToCopy       []string                  // snapshot of which target labels to copy (empty means copy all)
+}
+
 type Component struct {
 	opts component.Options
 	args Arguments
```
kgeckhart: I think we can drop args all together now that the last field was moved under mc. It doesn't look like anything reads it anymore. If that's wrong don't worry about it just looks like it might be dead now.

internal/component/prometheus/enrich/enrich_test.go
```diff
@@ -145,6 +204,112 @@ func TestEnricher(t *testing.T) {
...
+			app := c.receiver.Appender(t.Context())
+			_, _ = app.Append(0, lbls, 0, 0)
+			_ = app.Commit()
+		}
+	})
+
+	// continuously rotate Targets and LabelsToCopy.
+	wg.Go(func() {
+		for range iterations {
+			require.NoError(t, c.Update(args))
```
kgeckhart:
```suggestion
			assert.NoError(t, c.Update(args))
```

I think this needs the import too. IIRC require is going to do a FailNow which will kill the goroutine and since this one isn't the test goroutine it can cause the test to end in a weird state. Assert won't stop the test but it will make sure it ends it failure.

kgeckhart (approved):
Will merge after CI is green, thanks for doing this and iterating with our feedback!

PR #5830, opened 2026-03-19
fix(loki): Make drain forward entries with fallback timeout

kalleep:
### Pull Request Details
Since https://github.com/grafana/alloy/pull/5804 is merged we now have a defined shutdown order of components, where we start from sources. This means that we no longer have to drain into nothing but could forward them to fanout.

But I added in some protection in that if we cannot forward within `DefaultDrainTimeout` we will start to discard entries.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/drain.go
+++ b/internal/component/common/loki/drain.go
@@ -29,8 +29,8 @@ func Drain(recv LogsReceiver, fanout *Fanout, timeout time.Duration, fn func()) 
 		defer consumeCancel()
 		Consume(consumeCtx, recv, fanout)
 
-		// NOTE: If we could not forward entries within fallbackDuration we drain to nothing.
-		// This is just to guard against deadlock. If/when fn finish sucessfully this will stop.
+		// NOTE: If we could not forward entries during the configured timeout we discard entries.
+		// This is just to guard against deadlock. When fn finishes successfully this will stop.
 		discard(ctx, recv)
 	})
 
--- a/internal/component/common/loki/drain_test.go
+++ b/internal/component/common/loki/drain_test.go
@@ -1,8 +1,10 @@
 package loki
 
 import (
+	"strconv"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/grafana/loki/pkg/push"
@@ -30,16 +32,13 @@ func TestDrain(t *testing.T) {
 			}
 		})
 
-		completed := false
 		Drain(recv, NewFanout([]LogsReceiver{collector.Receiver()}), time.Second, func() {
 			require.Eventually(t, func() bool {
 				return len(collector.Received()) == 1
 			}, time.Second, 10*time.Millisecond)
-			completed = true
 		})
 
 		producer.Wait()
-		require.True(t, completed)
 		require.Len(t, collector.Received(), 1)
 		require.Equal(t, "forwarded", collector.Received()[0].Line)
 	})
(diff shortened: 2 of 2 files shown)


internal/component/common/loki/drain.go
```diff
@@ -2,32 +2,51 @@ package loki
...
+
+	defer func() {
+		cancel()
+		wg.Wait()
+	}()
+
+	wg.Go(func() {
+		consumeCtx, consumeCancel := context.WithTimeout(ctx, timeout)
+		Consume(consumeCtx, recv, fanout)
+		consumeCancel()
```
kgeckhart: Is this doing anything for us if we made it here this would be doing nothing or do we need to call cancel to free resources? If it's to free resources, could we do it as a defer or add a comment here.

internal/component/common/loki/drain.go
```diff
@@ -2,32 +2,51 @@ package loki
...
+
+		// NOTE: If we could not forward entries within fallbackDuration we drain to nothing.
+		// This is just to guard against deadlock. If/when fn finish sucessfully this will stop.
+		discard(ctx, recv)
```
kgeckhart, line 34: WDYT about a warning log if we fallback to discarding?

internal/component/common/loki/drain.go
```diff
@@ -2,32 +2,51 @@ package loki
...
-// This prevents deadlocks that can occur when stopping components that may still be
-// sending entries to the receiver channel. The draining goroutine will continue
-// consuming entries until f returns, at which point the context is cancelled and
-// the goroutine exits.
+const DefaultDrainTimeout = 2 * time.Minute
+
+// Drain forwards log entries from recv to fanout in a background goroutine while
+// fn executes. If forwarding blocks for longer than timeout, Drain falls back
+// to discarding entries from recv until fn returns. This prevents deadlocks in
+// shutdown paths where component may still send to recv while fn is stopping them.
```
kgeckhart, line 14:
```suggestion
// Drain forwards log entries from recv to fanout in a background goroutine while
// fn executes. It will continue to forward up to the timeout and then falls back
// to discarding entries from recv until fn returns. This prevents deadlocks in
// shutdown paths where component may still send to recv while fn is stopping them.
```

Is this a bit more accurate? Forwarding could be slow and we still fall back right?

internal/component/common/loki/drain_test.go
kgeckhart, line 1: Should we add `defer goleak.VerifyNone(t)` on these or does something else do leak detection for us? I don't think you're leaking anything but having the verification can't hurt.

internal/component/common/loki/drain_test.go
```diff
@@ -11,28 +11,60 @@ import (
 )
 
 func TestDrain(t *testing.T) {
-	recv := NewLogsReceiver()
+	t.Run("forwards while fn runs", func(t *testing.T) {
```
kgeckhart, line 19: WDYT about a combined test where we expect some entries to be forwarded and other to be dropped? This way we are testing the transition happens successfully vs each independently.

kalleep: Yes I agree and I started with writing one but it's really hard to not make it racy. But I think maybe synctest can help here, will add it.

internal/component/common/loki/drain.go
kalleep: Yes we need to call cancel to free resources but we can do that in a defer you are right :+1:

PR #5833, opened 2026-03-19
feat(beyla.ebpf): Bump Beyla to v3.6

marctc:

- Bump `beyla.ebpf` to Beyla v3.6
- Add missing fields from last updates
- Add integration test for java injector sdk

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/beyla/ebpf/beyla_linux.go
+++ b/internal/component/beyla/ebpf/beyla_linux.go
@@ -468,15 +468,6 @@ func (args Metrics) Convert() prom.PrometheusConfig {
 	return p
 }
 
-func (args Metrics) hasNetworkFeature() bool {
-	for _, feature := range args.Features {
-		if feature == "network" {
-			return true
-		}
-	}
-	return false
-}
-
 func (args Metrics) hasAppFeature() bool {
 	for _, feature := range args.Features {
 		switch feature {


collector/go.mod
```diff
@@ -1111,7 +1112,7 @@ replace github.com/prometheus/node_exporter => github.com/grafana/node_exporter
 
 replace github.com/github/smimesign => github.com/grafana/smimesign v0.2.1-0.20220408144937-2a5adf3481d3
 
-replace go.opentelemetry.io/obi => github.com/grafana/opentelemetry-ebpf-instrumentation v1.4.11
+replace go.opentelemetry.io/obi => github.com/grafana/opentelemetry-ebpf-instrumentation v1.12.2-0.20260318145328-e31c5acda289
```
rafaelroquetto, line 1115: I believe this should point to v1.15

kalleep:
Not sure you want to use chore in the title, it will not be added to changelog with that prefix. Probably feat instead?

marctc:
We have automatic changelog now? :heart_eyes:

kalleep:
>We have automatic changelog now?

Yes we do! It will also pick up changes from description if you want to add more details https://github.com/grafana/alloy/blob/main/docs/developer/contributing.md#pull-request-titles-and-commit-messages

collector/go.mod
marctc: It's actually pointing to that, but somewhat Go needs to have valid semver tag: https://go.dev/doc/modules/version-numbers

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -503,15 +510,14 @@ func (args Metrics) Validate() error {
 
 func (args Network) Convert(enable bool) obi.NetworkConfig {
```
skl: Is `enable` still used? Looks like it could be removed.

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -581,6 +590,118 @@ func (args Filters) Convert() filter.AttributesConfig {
...
+		w.AddK8sUIDAttributes = *args.AddK8sUIDAttributes
+	}
+	if args.UseLabelsForResourceAttributes != nil {
+		w.UseLabelsForResourceAttributes = *args.UseLabelsForResourceAttributes
+	}
+
+	return w
+}
+
+func (args Injector) Convert() (beyla.SDKInject, error) {
```
skl, line 634: Suggest adding a Validate method (similar to the other Validate methods) to check, e.g.

- If `webhook.enable` is `true`, `cert_path` and `key_path` should also be set
- `enabled_sdks` checked as known SDKs
- `otel_endpoint` checked as valid URL

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -790,6 +914,18 @@ func (a *Arguments) Convert() (*beyla.Config, error) {
...
 
+	cfg.Injector, err = a.Injector.Convert()
+	if err != nil {
+		return nil, err
+	}
+
+	if a.Injector.OTELEndpoint != "" {
+		endpoint := a.Injector.OTELEndpoint
+		cfg.Traces.OTLPEndpointProvider = func() (string, bool) {
+			return endpoint, true
```
skl, line 947: This means that the injector OTEL endpoint config takes priority for the global traces export endpoint, is that ok/intentional?

marctc: yeah, Alloy for default case it's using `cfg.TracesReceiver`

PR #5847, opened 2026-03-20
feat(helm): Allow setting `revisionHistoryLimit` in the helm chart

hegerdes:

Allows setting the `revisionHistoryLimit` for Deployments, StatefulSet and DaemonSet

### Pull Request Details

Some managed Kubernetes offerings limit the size of their etcd. Having lots of unused revisions can fill up a users quota.  
Additionally, GitOps setups like ArgoCD track every resource created by controllers. The PR allows users to lower the number of resources tracked by their GitOps tool. 

The limit is set to 10 - Kubernetes default value and is not included in the rendered chart if unchanged. So it should not have user facing impact.

```yaml
{{- if ne (int .Values.controller.revisionHistoryLimit) 10 }}
revisionHistoryLimit: {{ .Values.controller.revisionHistoryLimit }}
{{- end }}
```
The initial proposal was just setting `revisionHistoryLimit: {{ .Values.controller.revisionHistoryLimit }}` without a conndition

None

### Notes to the Reviewer

Either GitHub has problems again or the CLA site; It loads forever for me.

### PR Checklist

I ran `make` in the chart dir.

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/README.md
+++ b/operations/helm/charts/alloy/README.md
@@ -123,6 +123,7 @@ useful if just using the default DaemonSet isn't sufficient.
 | controller.podLabels | object | `{}` | Extra pod labels to add. |
 | controller.priorityClassName | string | `""` | priorityClassName to apply to Grafana Alloy pods. |
 | controller.replicas | int | `1` | Number of pods to deploy. Ignored when controller.type is 'daemonset'. |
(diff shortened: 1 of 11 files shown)


hegerdes:
Switched to the conditional include and updated the tests.  
Only included one for DS, thought every deployment type would be overkill since its the same for every type.

PR #5848, opened 2026-03-20
feat(mixin): Update loki dashboard

kalleep:
Update loki dashboard to include panel for request size and entry lentency propagation. Also enable native histogram for loki_write_request_size_bytes and updated bytes dropped to sum by reason.

### DIFF
--- a/operations/alloy-mixin/rendered/dashboards/alloy-loki.json
+++ b/operations/alloy-mixin/rendered/dashboards/alloy-loki.json
@@ -192,7 +192,7 @@
       },
       {
          "datasource": "${datasource}",
-         "description": "Percentile write latency\n",
+         "description": "Percentile write latency.\n",
          "fieldConfig": {
             "defaults": {
                "unit": "s"


operations/alloy-mixin/dashboards/loki.libsonnet
```diff
@@ -159,14 +159,83 @@ local filename = 'alloy-loki.json';
...
         Bytes dropped per second.
       |||) +
-      panel.withStacked() +
       panel.withUnit('Bps') +
+      panel.withStacked(stackingMode='off') +
+      panel.withPosition({ x: 12, y: 1 + y_offset, w: 12, h: 10 }) +
+      panel.withQueries([
+        panel.newQuery(
+          expr=|||
+            sum by(${groupby}, reason) (rate(loki_write_dropped_bytes_total{%(instanceSelector)s, host=~"$url"}[$__rate_interval]))
```
thampiotr, line 168: I don't think you mean to leave the `host=` selector here?

internal/component/common/loki/client/metrics.go
```diff
@@ -68,9 +68,12 @@ func newMetrics(reg prometheus.Registerer) *metrics {
...
 	m.requestSize = prometheus.NewHistogramVec(prometheus.HistogramOpts{
-		Name:    "loki_write_request_size_bytes",
-		Help:    "Number of bytes for requests.",
-		Buckets: []float64{1 * KiB, 4 * KiB, 16 * KiB, 64 * KiB, 256 * KiB, 512 * KiB, 1 * MiB, 2 * MiB, 4 * MiB, 8 * MiB, 16 * MiB, 20 * MiB},
+		Name:                            "loki_write_request_size_bytes",
+		Help:                            "Number of bytes for requests.",
+		Buckets:                         []float64{1 * KiB, 4 * KiB, 16 * KiB, 64 * KiB, 256 * KiB, 512 * KiB, 1 * MiB, 2 * MiB, 4 * MiB, 8 * MiB, 16 * MiB, 20 * MiB},
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: 1 * time.Hour,
```
thampiotr, line 76: Are these the right defaults? We give explicit buckets, but also have bucket factor and max number of buckets. What does that mean?

operations/alloy-mixin/dashboards/loki.libsonnet
kalleep: I did, we had that one before. But with the adhoc filter I am not sure this is even needed anymore 🤔

internal/component/common/loki/client/metrics.go
kalleep: This will expose this metric as both native and classic histogram.

As for the default I am not sure what are good values here, I just took whatever we use in other places

operations/alloy-mixin/dashboards/loki.libsonnet
thampiotr: ah, okay - let's leave as-is

PR #5849, opened 2026-03-20
chore: Ignore false positive trufflehog report

kalleep:
These are reported on pr. First one was a commit hash from loki repo, don't really see the point keeping this. The second one is a fake request id used in test.

https://github.com/grafana/alloy/pull/5848#issuecomment-4098176764


internal/component/loki/source/cloudflare/cloudflare.go
```diff
@@ -1,10 +1,5 @@
 package cloudflare
 
-// This code is copied from Promtail (a1c1152b79547a133cc7be520a0b2e6db8b84868).
```
jharvey10, line 3: If worth keeping, could probably shorten this to less characters to avoid trufflehog.

Issue #5826, opened 2026-03-19
chore: Prepare for loki architecture changes


kalleep:
### Add a description

I experimented a bit with how we can make incremental changes for https://github.com/grafana/alloy/pull/4940.

During that experiment I noticed it was much easier to partially migrate components that used the shared loki structures like `Fanout` and `Consume*` functions. While that design doc is still being discussed I want to move current components to use these functions and structures.

Components left to migrate:
- [ ] `loki.relabel`
- [ ] `loki.secretfilter`
- [ ] `loki.source.azure_event_hubs`
- [ ] `loki.source.gelf`
- [ ] `loki.source.kafka`
- [ ] `loki.source.kubernetes`
- [ ] `loki.source.podlogs`
- [ ] `loki.source.syslog`
- [ ] `loki.source.windowsevent`
- [ ] `faro.receiver`
- [ ] `database_observability.mysql`
- [ ] `database_observability.postgres` 

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5850, opened 2026-03-20
chore(loki.source.syslog): Use Fanout and Consume and Drain

kalleep:
### Pull Request Details
Now that we run target reload loop and consume loop in seperate go routine we don't have to start drain updating targets. We also use loki.Drain during shutdown.

Part of https://github.com/grafana/alloy/issues/5826

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/syslog/syslog.go
+++ b/internal/component/loki/source/syslog/syslog.go
@@ -94,14 +94,18 @@ func (c *Component) Run(ctx context.Context) error {
 		})
 	}()
 
-	var wg sync.WaitGroup
+	var (
+		wg                 sync.WaitGroup
+		consumeCtx, cancel = context.WithCancel(context.Background())
+	)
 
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/syslog/syslog.go
```diff
@@ -81,36 +81,39 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
-		case entry := <-c.handler.Chan():
-			c.mut.RLock()
-			for _, receiver := range c.fanout {
-				receiver.Chan() <- entry
+	var (
+		wg                 sync.WaitGroup
+		consumeCtx, cancel = context.WithCancel(context.Background())
+	)
+
+	wg.Go(func() { loki.Consume(consumeCtx, c.handler, c.fanout) })
```
kalleep, line 102: We pass a context here that will only cancel consume after reload loop has exited. That we way we cannot deadlock if reload happens just before components exit

PR #5859, opened 2026-03-24
chore(loki.sourcer.kubernetes, loki.source.podlogs): Use fanout, consume and drain

kalleep:
### Pull Request Details
Use shared functions and structure to make migration easier in the future

Part of: https://github.com/grafana/alloy/issues/5826

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/source/podlogs/podlogs.go
+++ b/internal/component/loki/source/podlogs/podlogs.go
@@ -160,7 +160,6 @@ func (c *Component) Run(ctx context.Context) error {
 			if c.tailer != nil {
 				c.tailer.Stop()
 			}
-
 		})
 	}()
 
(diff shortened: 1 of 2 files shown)


internal/component/loki/source/podlogs/podlogs.go
```diff
@@ -152,73 +150,47 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 // Run implements component.Component.
 func (c *Component) Run(ctx context.Context) error {
```
ptodev, line 152: I wonder if all those loki components can share the same Run() function in a similar way to how the otelcol ones do. It'd avoid repetition and in the future if anyone wants to update this patter, it'd be in one place.

kalleep: Yes I think so. I want to migrate all components first and then we can create abstractions for it

PR #5864, opened 2026-03-24
chore(faro): Use fanout

kalleep:
### Pull Request Details
Use fanout for faro log exporter.

Fixes: https://github.com/grafana/alloy/issues/5826

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/faro/receiver/exporters.go
+++ b/internal/component/faro/receiver/exporters.go
@@ -86,8 +86,7 @@ type logsExporter struct {
 	sourceMaps sourceMapsStore
 	format     LogFormat
 
-	receiversMut sync.RWMutex
-	receivers    []loki.LogsReceiver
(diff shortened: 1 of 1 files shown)


kalleep:
Before we had this
```go
	// Grab the current value of exp.receivers so sendKeyValsToLogsPipeline
	// doesn't block updating receivers.
	exp.receiversMut.RLock()
	var (
		receivers = exp.receivers
	)
	exp.receiversMut.RUnlock()
```

This is not safe in current loki pipelines and could cause deadlocks. If a downstream component was removed we could get stuck sending in a dead channel so we need to hold the look during send, [explained here](https://github.com/grafana/alloy/blob/87ceef3d4e68db074e89224d7637406f0c1b24dd/internal/component/common/loki/fanout.go#L27-L36).

PR #5866, opened 2026-03-24
chore(database_observability): Use Fanout, Consume and Drain

kalleep:
### Pull Request Details
We have added some shared functionallity for working with loki pipelines and should use them in our `database_observability`. This will make it easier in the future to start to migrate to [new architecture for loki pipelines](https://github.com/grafana/alloy/pull/4940) and avoid common pitfalls. 

I did not review all `collectors` but just did a sample and e.g. [explain_plan](https://github.com/grafana/alloy/blob/main/internal/component/database_observability/mysql/collector/explain_plans.go#L469) can cause a deadlock when component is stopping and in turn preventing alloy from stopping because nothing is consuming from handler.

Changes:
1. Use [loki.Fanout](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/fanout.go#L19) - This is doing internal locking of logs receivers so components do not have to care about it. It will also stop forwarding when context is canceled.
2. Use [loki.Consume](https://github.com/grafana/alloy/blob/0ac6d7ca7c744268668189c0f11460a84c9ef458/internal/component/common/loki/consume.go#L13) - Runs the consume loop from a `LogsReceiver` and will exit when context is canceled.
3. Use [loki.Drain](https://github.com/grafana/alloy/blob/0ac6d7ca7c744268668189c0f11460a84c9ef458/internal/component/common/loki/drain.go#L18) - When components is stopping we need to still forward / drain to make sure collectors can stop. Since  https://github.com/grafana/alloy/pull/5613 we have a deterministic order for stopping components so it is safe to forward to downstream components during stops but we still have a timeout after we would drain into nothing.
4. Use [loki.NewEntry](https://github.com/grafana/alloy/blob/0ac6d7ca7c744268668189c0f11460a84c9ef458/internal/component/common/loki/entry.go#L10) - This makes sure we set `created` timestramp of entries so they can be properly tracked by `loki_write_entry_propagation_latency_seconds`.

Part of: https://github.com/grafana/alloy/issues/5826

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/component.go
+++ b/internal/component/database_observability/mysql/component.go
@@ -241,9 +241,9 @@ type Collector interface {
 type Component struct {
 	opts              component.Options
 	args              Arguments
-	mut               sync.RWMutex
-	receivers         []loki.LogsReceiver
 	handler           loki.LogsReceiver
+	fanout            *loki.Fanout
+	mut               sync.RWMutex
 	registry          *prometheus.Registry
 	baseTarget        discovery.Target
 	collectors        []Collector
@@ -262,7 +262,7 @@ func new(opts component.Options, args Arguments, openFn func(driverName, dataSou
 	c := &Component{
 		opts:      opts,
 		args:      args,
-		receivers: args.ForwardTo,
+		fanout:    loki.NewFanout(args.ForwardTo),
 		handler:   loki.NewLogsReceiver(),
 		registry:  prometheus.NewRegistry(),
 		healthErr: atomic.NewString(""),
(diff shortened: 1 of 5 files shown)


internal/component/database_observability/mysql/component.go
```diff
@@ -373,6 +371,7 @@ func (c *Component) Update(args component.Arguments) error {
 	defer c.mut.Unlock()
 
 	c.args = args.(Arguments)
+	c.fanout.UpdateChildren(c.args.ForwardTo)
```
kalleep, line 374: Before both of these components stored both args and `receivers` but never actually updated the second one. So if a config change was made with new receivers that would not work and we would keep sending logs to the old receiver list causing all kinds of problems.

matthewnolf (approved):
LGTM 👍

PR #5885, opened 2026-03-24
test(loki.secretfilter): Add tests for skipping `tokenId` variants

kleimkuhler:

Skip `token_id` and its variants redactions in `loki.secretfilter`.

### Pull Request Details

Add test for skipping `token_id` and its variants in the custom gitleaks extended configuration file.

We noticed that we had `generic-api-key` redactions happening but no actual value being redacted ([slack](https://raintank-corp.slack.com/archives/C044LC5GMGF/p1774378872467969)). The reason for this is the substring `… token_id= total=123.456 … foo bar`. There is no value for `token_id` so it’s just a space, but then `total=123.456` has enough entropy to cause a redaction. When this happens we see `token_id= <ALLOY-REDACTED-SECRET:… foo bar` — notice that `total` is now missing from the log line.

We’ve considered allowing `token_id` because it triggers a lot of `generic-api-key` redactions that are false positives, but now that we see it’s leading to incorrect reductions it seems to be the best path forward. This adds a test that demonstrates the behavior, and asserts that when it’s added to `gitleaks.toml`, we properly skip redaction.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/secretfilter/extend/extend_test.go
+++ b/internal/component/loki/secretfilter/extend/extend_test.go
@@ -46,3 +46,25 @@ func TestSecretFiltering_WithGitleaksConfigFile(t *testing.T) {
 
 	secretfilter.RunTestCases(t, config, cases)
 }
+
+func TestSecretFiltering_RequestTimingsLog(t *testing.T) {
+	configPath := gitleaksConfigPath(t)
+	if _, err := os.Stat(configPath); err != nil {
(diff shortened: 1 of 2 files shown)


kleimkuhler:
Closing this as we are not testing new secretfilter functionality here... just asserting a gitleaks configuration change would work as expected before https://github.com/grafana/deployment_tools/pull/526855

PR #5894, opened 2026-03-25
fix(database_observability.mysql): Refactor explain plan loop batch size

cristiangreco:
The fetchExplainPlans() function in the explain_plans collector used defer inside a for loop to handle per-iteration bookkeeping. Given defer is not executed until the enclosing function returns, the count of processed queries was never incremented during the loop, making batch size ineffective (i.e. every cache entry is processed every scrape).

This change moves the logic into a separate function and simplifies the bookkeeping in the caller.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/explain_plans.go
+++ b/internal/component/database_observability/mysql/collector/explain_plans.go
@@ -582,149 +582,147 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 
 	processedCount := 0
 	for _, qi := range c.queryCache {
-		generatedAt := time.Now().Format(time.RFC3339)
-		nonRecoverableFailureOccurred := false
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/mysql/collector/explain_plans.go
```diff
@@ -582,149 +582,147 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 
 	processedCount := 0
 	for _, qi := range c.queryCache {
-		generatedAt := time.Now().Format(time.RFC3339)
-		nonRecoverableFailureOccurred := false
```
cristiangreco, line 586: The diff is easier to follow by [hiding whitespaces](https://github.com/grafana/alloy/pull/5894/changes?w=1).

rgeyer (approved):

PR #5897, opened 2026-03-25
chore(beyla.ebpf): Add agentic GH action to bump Beyla releases automatically

marctc:

This PR adds an agentic workflow to help upgrading beyla.ebpf component to an specific version.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


.github/workflows/agent_bump_beyla.yml
```diff
@@ -0,0 +1,206 @@
...
+
+      - name: Configure Git
+        env:
+          APP_ID: ${{ steps.get-app.outputs.app-id }}
+          APP_LOGIN: ${{ steps.get-app.outputs.app-login }}
+        run: |
+          git config user.name "${APP_LOGIN}"
+          git config user.email "${APP_ID}+${APP_LOGIN}@users.noreply.github.com"
+
+      - name: Run Claude Code
```
skl: I think it might be worth splitting this into two jobs, plan and implement.

The reason is that having a single one-shot 90-minute job is going to be difficult to debug if it fails, and it doesn't provide any incremental benefit. What I mean is if it fails, no draft PR is created, and we're left with nothing. If a human had to pickup the job, they'd be starting from the beginning again.

If you at least split it into a plan job, and output a `plan.md` artifact for the implement job to use, then if that fails - at least a human could look at the `plan.md` for reference.

Ideally it might be useful for the implement job to keep the Claude code comment updated every time it goes around the change/test failure cycle. It could for example summarise what it's tried so far, and what failed. Again just helping a human if they have to pick up from a failure it would provide some clues on what went wrong.

PR #5900, opened 2026-03-25
docs: Update release cadence mentions across the docs

jharvey10:
Release cadence is now 3 weeks instead of 6.


clayton-cornell:
Do we want to backport to v1.15?

jharvey10:
@clayton-cornell Good point, I'll add the label!

jharvey10:
Oh wait I just realized that the label shouldn't exist yet. 1.15 hasn't gone final. I need to think about this. It looks like the backport label was created manually, but that might break the automation that runs down the road.

clayton-cornell:
> Oh wait I just realized that the label shouldn't exist yet. 1.15 hasn't gone final. I need to think about this. It looks like the backport label was created manually, but that might break the automation that runs down the road.

Sorry that was me that manually created the label. I didn't know there was an automation in place... and historically we had to create it at the first instance where we decided to add it.... so I did... :-P

jharvey10:
No worries! It actually sparked me to improve the automation for when they get made. Now (soon), if it exists, you can safely use it, if not then backporting isn't [yet] necessary based on what stage of the release process we're in.

#5908

PR #5901, opened 2026-03-26
chore(promhttp2): Use less reflection&unsafe, nosemgrep: use-of-unsafe-block

korniltsev-grafanista:
…e-block

Change the way we use unsafe in promhttp2 package
- Do not use unsafe to set each field one by one
- just transpose one struct pointer to another once
- remove nil interface check handling
- nosemgrep: use-of-unsafe-block
Context:
https://raintank-corp.slack.com/archives/C03NCLB4GG7/p1774413479849249

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/pyroscope/write/promhttp2/reflect_oauth2.go
+++ b/internal/component/pyroscope/write/promhttp2/reflect_oauth2.go
@@ -1,6 +1,7 @@
 package promhttp2
 
 import (
+	"fmt"
 	"net/http"
 	"reflect"
 	"unsafe"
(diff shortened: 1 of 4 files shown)


internal/component/pyroscope/write/promhttp2/reflect_oauth2.go
```diff
@@ -15,26 +16,27 @@ import (
...
-		field := optsType.Field(i)
-		src := reflect.NewAt(field.Type, unsafe.Add(unsafe.Pointer(opts), field.Offset)).Elem()
-		dst := reflect.NewAt(field.Type, unsafe.Add(promPtr, field.Offset)).Elem()
-		dst.Set(src)
+func newOAuth2RoundTripper(oauthCredential commonconfig.SecretReader, config *commonconfig.OAuth2, next http.RoundTripper, opts *httpClientOptions) (http.RoundTripper, error) {
+	if oauthCredential == nil {
+		oauthCredential = commonconfig.NewInlineSecret("")
+	}
+	if config == nil {
+		return nil, fmt.Errorf("newOAuth2RoundTripper: config == nil")
```
marcsanmi, line 24: would this be a bit redundant? ` unable to create OAuth2RoundTripper: newOAuth2RoundTripper: ....`

marcsanmi (approved):
LGTM

internal/component/pyroscope/write/promhttp2/reflect_oauth2.go
korniltsev-grafanista: sorry I dont understand. What do you propose?

marcsanmi: I was thinking if using something like this `fmt.Errorf("config must not be nil")` would be less redundant  for the inner errors, since the caller already provides the context. It's a nit, feel free to skip.

Issue #5862, opened 2026-03-24
pyroscope.ebpf: potential memory leak in pprofReporter


luweglarz:
### Component(s)

pyroscope.ebpf

### What's wrong?

Some of our alloy pods using the pyroscope.ebpf component stopped to send profiles, it was then followed by a continuous memory usage increase
This graph shows when one of the alloy pod having the issue stopped pushing profiles
<img width="959" height="237" alt="Image" src="https://github.com/user-attachments/assets/53f38b18-03ca-4287-a917-ff233c685d5e" />
The drops correlated with the continuous memory increase
<img width="1992" height="319" alt="Image" src="https://github.com/user-attachments/assets/c40381c7-8d98-4fb1-af57-9ccbee602db7" />
The flamegraph show that the memory inuse is on pprof reporter
<img width="2210" height="858" alt="Image" src="https://github.com/user-attachments/assets/9e306cdb-86f8-49d9-b749-bbb596dd690e" />
I am not fully sure what is the root cause yet and need to investigate more

### Steps to reproduce

Not sure yet, just alloy running with pyroscope.ebpf component

### System information

Linux 5.14 x86_64

### Software version

Alloy 1.14.0

### Configuration

```text
pyroscope.ebpf "profile_ebpf" {
      forward_to     = [pyroscope.write.default.receiver]
      targets        = discovery.relabel.k8s_ebpf_pods_relabel.output
      collect_interval = "15s"
      container_id_cache_size = 1024
      demangle = "full"
      dotnet_enabled = true
      hotspot_enabled = true
      perl_enabled = true
      php_enabled = true
      python_enabled = true
      ruby_enabled = true
      sample_rate = 20
      v8_enabled = true
      off_cpu_threshold = 0
      targets_only =  true
      lazy_mode = true
}
```

### Logs

```text

```

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

luweglarz:
to add more info I have a found a potential deadlock between 2 goroutine after doing a goroutien dump on an alloy having this issue

```
goroutine 740 [sync.Mutex.Lock, 96 minutes]:
internal/sync.runtime_SemacquireMutex(0xc003043028?, 0xc8?, 0x200000003?)
	/usr/local/go/src/runtime/sema.go:95 +0x25
internal/sync.(*Mutex).lockSlow(0xc000bcdc50)
	/usr/local/go/src/internal/sync/mutex.go:149 +0x15d
[...]

PR #5629, opened 2026-02-24
fix: Close memory-mapped tables on LRU eviction in  pyroscope.ebpf

keyolk:
from https://github.com/grafana/opentelemetry-ebpf-profiler/pull/45

## Summary

Fix memory leak where lidia symbol tables (memory-mapped files) are never closed when evicted from the LRU cache, causing unbounded file memory growth.

## Problem

The `irsymcache.Resolver` maintains two data structures:

1. `cache` - LRU cache mapping FileID to cached marker
2. `tables` - Map of FileID to memory-mapped lidia Table

When entries are evicted from the LRU cache, the `SetOnEvict` callback only removes the cache file from disk but **never closes the memory-mapped table**:

```go
// Before: Only removes file, table stays in memory
cache.SetOnEvict(func(id libpf.FileID, marker cachedMarker) {
    if marker == erroredMarker {
        return
    }
    filePath := res.tableFilePath(id)
    if err = os.Remove(filePath); err != nil {
        log.Errorf("symbcache eviction error: %v", err)
    }
})
```

This causes:
- **~115MB of file memory growth** that accumulates over time
- Memory-mapped files remain open indefinitely
- No upper bound on memory usage for symbol tables

## Solution

Close the memory-mapped table and remove from `tables` map before deleting the file:

```go
// After: Close table before removing file
cache.SetOnEvict(func(id libpf.FileID, marker cachedMarker) {
    if marker == erroredMarker {
        return
    }
    // Close and remove the memory-mapped table to prevent memory leak
    res.mutex.Lock()
    if table, ok := res.tables[id]; ok {
        table.Close()
        delete(res.tables, id)
    }
    res.mutex.Unlock()

    filePath := res.tableFilePath(id)
    if err = os.Remove(filePath); err != nil {
        log.Errorf("symbcache eviction error: %v", err)
    }
})
```

## Test results

Deployed to Kubernetes cluster:

| Metric | Before | After |
|--------|--------|-------|
| File memory | ~135 MiB (growing) | ~20 MiB (stable) |

## Note

This PR is based on commit 6d4ae03 (before the pyroscope-specific code was moved). If the irsymcache code has been moved to Alloy, this fix should be applied there instead.

## Related

- Issue: https://github.com/grafana/alloy/issues/5442
- Related PR: https://github.com/grafana/opentelemetry-ebpf-profiler/pull/44

BEGIN_COMMIT_OVERRIDE
[...]

PR #5911, opened 2026-03-27
fix(pyroscope.ebpf): Fix deadlock on LRU eviction in irsymcache

luweglarz:

Fix deadlock between ResolveAddress and the LRU eviction callback in `irsymcache`. Move `mutex.Lock` after  `cache.Get` so the two locks are never held at the same time.

### Pull Request Details

When the cache is full, `ResolveAddress` and the LRU evict callback deadlock each other:
1. `ResolveAddress`: locks `Resolver.mutex`, then calls `cache.Get` (needs LRU lock)
2. Eviction callback: holds LRU lock, then calls `res.mutex.Lock`

The fix moves cache.Get before `mutex.Lock` in `ResolveAddress`. using explicit `Unlock` instead of `defer` because the error path calls `cache.Remove` which also takes the LRU lock.

The deadlock was introduced in #5629, the reporter goroutine stucks, and memory keeps growing and no profiles are being reported.

I also added a test case that simulate the deadlock.

#5862 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


luweglarz:
Hi @korniltsev-grafanista, wondering if you could have a look, thanks !

Issue #5549, opened 2026-02-17
proposal: Add alloy syntax types for sub expressions


kalleep:
### Background

In alloy we sometime have "sub" expression in config e.g. Regexp and OTTL.

In config alloy config we use the string type and perform validation when creating the component.

For example in stage.regex the [config attribute is a string ](https://github.com/grafana/alloy/blob/6087b4eab8876cbba8874eb7b9404f2a9bccc325/internal/component/loki/process/stages/regex.go#L26) and we [validate when creating the stage](https://github.com/grafana/alloy/blob/6087b4eab8876cbba8874eb7b9404f2a9bccc325/internal/component/loki/process/stages/regex.go#L41).

### Proposal

We should implement our own wrapper types that implements `encoding.TextUnmarshaller` and `encoding.TextMarshaller` for these sub expressions. This would allow us to validate these expression at an earlier stage _and_ make it possible for validate command to validate the syntax as well.

POC for regexp: https://github.com/grafana/alloy/pull/5548

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

PR #5916, opened 2026-03-27
fix(validation): Improve type checking of ast.LiteralExpr

kalleep:
Followup to https://github.com/grafana/alloy/pull/5076. We can use `value.decoder` on ast.LiteralExpr. This will handle capsule conversions and checking things like `Validate` and `TextUnmarshal`. 

When updating to use `value.Decoder` I realized that the precious checks was to restrictive for e.g. using numbers instead of string, our syntax will convert numbers to strings so that is totally valid.

Part of: https://github.com/grafana/alloy/issues/3712
Part of: https://github.com/grafana/alloy/issues/5549

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


blewis12 (approved):
nice!

Issue #5852, opened 2026-03-20
pullPolicy for configReloader


mbecker-tw20:
### Request

There is no option to set pullPolicy in configReloader section

### Use case

We using Kyverno in our Clusters and need to set pullPolicy to always.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

M-A-X-I-M:
this was fast, thanks! 🚀

PR #5923, opened 2026-03-30
feat(helm): Allow configuring image pull policy for config reloader

kalleep:
### Pull Request Details

Allow to configure `imagePullPolicy` for config reloader. If it's empty it will not set it and use the default from k8s.

Fixes: https://github.com/grafana/alloy/issues/5852

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


operations/helm/charts/alloy/values.yaml
```diff
@@ -244,6 +244,8 @@ configReloader:
     tag: v0.81.0
     # -- SHA256 digest of image to use for config reloading (either in format "sha256:XYZ" or "XYZ"). When set, will override `configReloader.image.tag`
     digest: ""
+    # -- Config reloader image pull policy.
+    pullPolicy: ""
```
blewis12: Do you think it makes sense to have the default value mirror that of the [image.pullPolicy](https://github.com/grafana/alloy/blob/ea0edff8aea23d945f4d6e95bd806f3d66a32e9b/operations/helm/charts/alloy/values.yaml#L151) which is explicitly `IfNotPresent`? This is anyways the [default](https://kubernetes.io/docs/concepts/containers/images/?utm_source=chatgpt.com#updating-images) from k8s

kalleep: Sure that makes sense, it is what we have for alloy image :+1:

M-A-X-I-M:
thanks!🚀

PR #5954, opened 2026-03-31
chore: Update Alloy Helm Chart to Point to Alloy v1.15.0

blewis12:
Updates our Alloy helm chart to reference the newly published Alloy v1.15.0

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -10,7 +10,7 @@ internal API changes are not present.
 Unreleased
 ----------
 
-1.7.0 (2026-03-31)
+1.7.0 (2026-04-01)
 ----------
 
 ### Bug fixes


operations/helm/charts/alloy/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ internal API changes are not present.
 Unreleased
 ----------
 
+1.6.3 (2026-03-31)
```
ptodev: Let's make it 1.7.0 instead? If it's a minor release of Alloy it's better to make a minor release of the helm chart since it's a more risky upgrade. For major breaking changes maybe we can even bump the major version number. Thankfully 1.15.0 doesn't have such change.

blewis12: ah yes you're right 👍 I can do that

blewis12: alright - done now

PR #5956, opened 2026-03-31
ci(extension/alloyengine): Publish as independent Go module

### LINKED ISSUE #5823: [otel-engine] Ensure alloy extension is OCB-able
### Component(s)

_No response_

### Add a description

The extension is not OCB-able today because it's a separate module we are not currently publishing with our releases. We will need to update the release process to create new tagged versions and any other blockers.

### Tip

kgeckhart:

Publishes `extension/alloyengine` as an independently versioned Go module so it can be referenced from external OpenTelemetry Collector Builder distros.

### Pull Request Details

The `extension/alloyengine` module has its own `go.mod` but has never been tagged or tracked by release-please, making it impossible to reference from external OpenTelemetry Collector Builder distros.

This adds `extension/alloyengine` to release-please so it gets tagged alongside each Alloy release (e.g. `extension/alloyengine/v1.15.0`). It also updates the `require` for `github.com/grafana/alloy` from the sentinel version (`v0.0.0-00010101000000-000000000000`) to `v1.15.0` so external consumers can resolve the transitive dependency from the module proxy — the existing `replace` directive continues to override this locally.

Finally, `extension/alloyengine/go.mod` is added to the root package's `extra-files` so the require version stays in sync with the Alloy version on each future release.

Fixes #5823

### Notes to the Reviewer

I'll create the `extension/alloyengine/v1.15.0` tag manually after merging to backfill the current release. Future releases will be tagged automatically by release-please.

### Workflow Testing

Tested the updated tag creation script against a private throwaway repo with an annotated tag (`v1.14.0`) to verify annotated tag dereference and annotated tag creation both work correctly.

**Input:** `v1.14.0` is an annotated tag (type: `tag`), SHA points to tag object `df35f835...`

**Dereference step:** resolved to underlying commit `e9f2f140...`

**Tag creation:** created annotated tag object `97d9820d...` pointing to commit, then created ref:

```json
{
  "ref": "refs/tags/extension/alloyengine/v1.14.0",
  "object": {
    "sha": "97d9820d34a9ae46978fdd8b737588de8740fb48",
    "type": "tag"
  }
}
```

Tag object resolves to the correct commit:

```json
{
  "tag": "extension/alloyengine/v1.14.0",
  "message": "extension/alloyengine v1.14.0",
  "object": {
    "sha": "e9f2f14074703c82aa2e82b62b3701f666b57869",
    "type": "commit"
  }
}
```


blewis12:
yay!

jharvey10 (commented):
Even though we *can* version the alloyengine extension separately from base Alloy, I'd recommend we *don't* do that. 

To me, it seems fair to have them always synchronized since the extension is pretty much just a wrapper around Alloy. The only potential "gotcha" with doing that is that if the extension's wrapper code itself needs an update (even if base Alloy *doesn't* need an update), we'd need to bump Alloy in order to do that as a BAU PR like for anything else Alloy-related.

Said differently: A merged change affecting just the alloyengine extension would create a normal changelog entry for base Alloy, which would get picked up as a change for the next "normal" Alloy release by the existing release-please process.

Given that approach, all we'd need to do is have a basic workflow which runs when new alloy release tags get published and reacts by pushing a corresponding tag for the extension, e.g. `extension/alloyengine@v1.15.0`.

The go.mod "sync" would be handled by release-please (via an "extra file") and the tagging would be a small workflow file. No separate release-please package needed. Since we're locally replacing the module, there's no `go.sum` update needed to go along with that either.

There's an extremely tiny gap between when the go.mod file gets updated with the new alloy version (when the PR gets merged) and when the tag actually gets created, but we're talking maybe 30 seconds, max, and none of the build artifacts would have started building until the tag exists anyway, so the risk is incredibly minimal.

kgeckhart:
I think that's all fair, our goal wasn't really to have a process to release it independently so keeping it a post process makes sense. I updated the PR accordingly.

.github/workflows/release-tag-alloyengine.yml
```diff
@@ -0,0 +1,48 @@
...
+          private-key: ${{ fromJSON(steps.get-secrets.outputs.secrets).ALLOYBOT_PRIVATE_KEY }}
+          owner: grafana
+          repositories: alloy
+
+      - name: Create extension/alloyengine tag
+        env:
+          ALLOY_TAG: ${{ inputs.tag || github.ref_name }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          ALLOY_SHA=$(gh api repos/grafana/alloy/git/ref/tags/"${ALLOY_TAG}" --jq '.object.sha')
```
jharvey10: since we use signed/annotated tags, this will point to the sha of the tag object and not the underlying commit. for release branch creation we use [GetRefSHA](https://github.com/grafana/alloy/blob/0495682b64e58883301166c19543ec1f25a508ff/tools/release/internal/github/client.go#L153) to cover all the cases. Could be used here as well. Similarly for creating the tag, could use `CreateTag` ([example](https://github.com/grafana/alloy/blob/0495682b64e58883301166c19543ec1f25a508ff/tools/release/create-rc/main.go#L151)).

extension/alloyengine/go.mod
```diff
@@ -7,7 +7,7 @@ replace github.com/grafana/alloy => ../..
 replace github.com/grafana/alloy/syntax => ../../syntax
 
 require (
-	github.com/grafana/alloy v0.0.0-00010101000000-000000000000
+	github.com/grafana/alloy v1.15.0
```
jharvey10:
```suggestion
	github.com/grafana/alloy v1.15.0 // x-release-please-version
```

This should allow release-please to find and update this as a part of the release PRs. One thing to check perhaps would be whether or not `make generate-module-dependencies` clobbers a comment here.

jharvey10 (commented):
Summarizing a couple messages from slack here on the PR.

.github/workflows/release-tag-alloyengine.yml
```diff
@@ -0,0 +1,67 @@
...
+            -f tag="${TAG_NAME}" \
+            -f message="extension/alloyengine ${ALLOY_TAG}" \
+            -f object="${ALLOY_SHA}" \
+            -f type="commit" \
+            --jq '.sha')
+
+          # Create the ref pointing to the tag object
+          gh api repos/grafana/alloy/git/refs \
+            -f ref="refs/tags/${TAG_NAME}" \
+            -f sha="${TAG_OBJECT_SHA}"
```
jharvey10:
```suggestion
            -f sha="${TAG_OBJECT_SHA}"

```

jharvey10 (approved):
Approving despite tiny newline comment

Issue #4953, opened 2025-11-28
[Epic] change architecture for loki pipelines


kalleep:
# Loki Pipeline Migration Plan
Plan to make the changes from[ this design doc](https://github.com/grafana/alloy/blob/4f49b57829bcbb8f5ecd18daa8b58127743562ba/docs/design/4940-reliable-loki-pipelines.md)

## Phase 1: Increase integration test coverage for Loki pipelines

Before changing pipeline internals, we should make sure we have strong integration/unit test coverage for the current Loki pipeline behavior. 

## Phase 2: Introduce new pipeline primitives in `common/loki`

This phase is one PR focused only on introducing the shared synchronous pipeline primitives in `common/loki`. These components should have extensive testing including tests for congestion and backpressure.

- define the `Consumer` interface
- define the batch structure that will represent a unit of synchronous processing
- define `Fanout` for synchronous forwarding to multiple downstream consumers
- define `Sharding` for per-stream ordered execution
- define any supporting error types or result semantics the sources will later need
- add a temporary single-entry method on `Consumer` so components can switch from channels to function calls before they switch to batch processing.

Why the temporary single-entry method exists:

- moving from channels to function calls is one migration step
- moving from single-entry flow to batch flow is a separate migration step
- it allows source components to be converted over multiple PRs, with batching support added incrementally instead of being required in the first migration step

## Phase 3: Switch consuming components to the `Consumer` interface

All consuming Loki components should implement `Consumer` and export that interface instead of `LogsReceiver`. Sources should then call the consumer functions directly instead of forwarding entries on channels.

This phase should only change the component boundary, not the internal implementation of each component. This means sources will call consuming components synchronously at the boundary, but the final internal semantics of those components will still be completed in later.

We are also going to deprecate `LogsReceiver` and `EntryHandler` interface and implementations here and annotate code that still use it.

For example, `loki.write` should accept entries through `Consumer` while still forwarding internally through its existing channel, and `loki.process` should accept entries through `Consumer` while still running its internal stage pipeline through channels as it does today.

It is important to call out that success will still be reported earlier in the pipeline than in the final design. In this phase, success may still be returned before logs reach the final acceptance boundary. For example, `loki.process` may return success after forwarding into its existing internal channel-based pipeline, rather than after `loki.write` has accepted the logs.

So this phase changes how components are connected, but it does not yet give us the final success and error propagation semantics.

## Phase 4: Migrate component internals to the new architecture

[...]

PR #5957, opened 2026-03-31
chore: Add loki processing integration test

kalleep:
### Pull Request Details
Add a new  integration test for `loki-process` that tests several `loki.process` stages against different inputs.

I tried to mix different kind of parser and stages and draw inspiration from what we have in our docs. Also all the test fixtures are written when test run so that timestamps are not stale and get to old to be ingested into loki.

### Notes to the Reviewer

I added the ability to verify structured metedata. For this to work we have to pass a special header to loki:
`X-Loki-Response-Encoding-Flags: categorize-labels`. This will change the response so I had to update the json parsing logic for it.

### PR Checklist
- [ ] Documentation added
- [x] Tests updated


kalleep:
@codex review

integration-tests/docker/tests/loki-file/loki_file_test.go
```diff
@@ -13,27 +13,9 @@ func TestReadLogFile(t *testing.T) {
 		t,
 		common.ExpectedLogResult{
 			Labels: map[string]string{
-				"detected_level": "info",
+				"filename": "/etc/alloy/logs.txt",
```
kalleep, line 16: We can no longer check `detected_level` as a labels because we now tell loki to categorize our labels, meaning `detected_level` is not treated as label but a parsed value.

Parsed seems to be similar to structured_metadata but is "metadata" that loki can derive from requests.

PR #5965, opened 2026-04-01
fix(loki.process): Prevent stage.structured_metadata from adding the same metadata several times

kalleep:
### Pull Request Details
I was working with increasing test coverage for `loki.process`. I wanted to setup more full pipeline tests for this component.

When writing the first test case "simple cri pipeline" I notcied that `stage.structured_metadata` can add metadata several times. StructuredMetadata is stored as a slice and we would just append to it.

So when using something like `stage.cri` before this one we will have `stream` both as a label _and_ in the extracted map. To fix this we check if it already exists and if it does we replace it otherwise we append.

I also cleanup both the stage and tests  for it and added regression test to make sure we don't add metadata several times.

Relates to https://github.com/grafana/alloy/issues/4953

### Notes to the Reviewer

Part of cleaning up this stage is to have similar function for every place / way we try to extract metadata.

Before we had `processLabelsConfigs`, inline regex check for extracted map and later we called `extractFromLabels`.

I created a function for each combination of source and config instead so we have:
* `processExtractedLabelsByConfig` - source is extracted using labelConfig
* `processExtractedLabelsByRegex` - source is extracted using regex
* `processEntryLabelsByConfig` - source is labels using labelConfig
* `processEntryLabelsByRegex` - source is labels using regex

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/process_test.go
+++ b/internal/component/loki/process/process_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/go-kit/log"
 	"github.com/grafana/loki/pkg/push"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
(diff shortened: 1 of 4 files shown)


kalleep:
@codex review

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -75,40 +76,32 @@ func (*structuredMetadataStage) Cleanup() {
...
-							level.Debug(s.logger).Log("msg", "invalid label value parsed", "value", labelValue)
-						}
-						continue
-					}
-					e.StructuredMetadata = append(e.StructuredMetadata, push.LabelAdapter{Name: lName, Value: string(labelValue)})
-				}
+		appendStructureMetadata := func(labelName model.LabelName, labelValue model.LabelValue) {
+			if !containsStructuredMetadataLabel(e.StructuredMetadata, string(labelName)) {
+				e.StructuredMetadata = append(e.StructuredMetadata, push.LabelAdapter{Name: string(labelName), Value: string(labelValue)})
 			}
```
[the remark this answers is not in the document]
kalleep: This should be fine

thampiotr (commented):
If we have a pipeline:

stage A sets foo=old
stage B sets foo=new

it would result previously in foo=old,foo=new
but I think the correct result is foo=new
we should be able to override labels with new values - it should work like the normal indexed labels IMO. 

and I think that there are missing tests that would be verifying this override behaviour - can we make sure they are there?

internal/component/loki/process/process_test.go
```diff
@@ -31,51 +30,146 @@ import (
 )
 
 func TestComponent(t *testing.T) {
-	t.Run("update with invalid stage config", func(t *testing.T) {
```
thampiotr, line 34: I think this is now missing?

kalleep:
As discussed in slack I changed it so that we only append if it does not exist otherwise we replace.

As for testing I two tests originally and after changes 2 more:
- `expected structured metadata to be extracted once when values and regex both match extracted values` - verifies we do not append the same structured metadata key twice when both values and regex select the same extracted field
- `expected structured metadata to be extracted once when present in both extracted values and label` - verifies we do not append a duplicate when the same key is available from both extracted values and labels in the same stage
- `expected later structured metadata stage to replace earlier stage output` - verifies that we override the value if we have several stage.structured_metadata.
- `expected later source within a stage to replace existing structured metadata` - If value exists in both labels and extracted we get it from labels.

internal/component/loki/process/process_test.go
kalleep: Added it back in [a4fdcd1](https://github.com/grafana/alloy/pull/5965/commits/a4fdcd11daa628aef2ab2be716a024440c8069a6)

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -75,40 +76,40 @@ func (*structuredMetadataStage) Cleanup() {
...
+			if i != -1 {
+				e.StructuredMetadata[i] = metadata
+				return
 			}
+
+			e.StructuredMetadata = append(e.StructuredMetadata, metadata)
 		}
 
-		return s.extractFromLabels(e)
+		// Try to add structured metdata from extracted map using labelsConfig.
```
thampiotr:
```suggestion
		// Try to add structured metadata from extracted map using labelsConfig.
```

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -75,40 +76,40 @@ func (*structuredMetadataStage) Cleanup() {
...
+		processExtractedLabelsByConfig(s.logger, e.Extracted, s.labelsConfig, appendStructureMetadata)
+
+		// Try to add structured metadata from extraced map using regex.
```
thampiotr:
```suggestion
		// Try to add structured metadata from extracted map using regex.
```

internal/component/loki/process/stages/structured_metadata.go
```diff
@@ -75,40 +76,40 @@ func (*structuredMetadataStage) Cleanup() {
...
+		processExtractedLabelsByRegex(s.logger, e.Extracted, s.regex, appendStructureMetadata)
+
+		// Try to add structured metadata from labels using labelsConfig.
+		processEntryLabelsByConfig(e.Labels, s.labelsConfig, appendStructureMetadata)
+
+		// Tru to add structured metadata from labels using regex.
```
thampiotr:
```suggestion
		// Try to add structured metadata from labels using regex.
```

thampiotr (approved):

thampiotr (approved):
LGTM % some typos

PR #5968, opened 2026-04-01
Refactor task based docs

ptodev:
This is an attempt to simplify and consolidate the examples and task based docs:

* The current docs are sometimes too verbose.
* The current docs are not consistent. Some use docker containers for examples, some use killercoda, some include a picture of the component graph. They all vary.
* The examples often include outdated practices and deprecated functionality.
* There is lots of duplication. E.g. there could be a task based doc with a certain Alloy config, and literally the same config may also be listed as an example in a component reference page
* Some examples in component reference pages list very elaborate examples which span several components. Those should be task based docs
* When users unfamiliar with Alloy browse the docs, they don't know what components to use. It'd be good to use the component reference only for component reference and list examples elsewhere.


clayton-cornell:
Did some digging on this.  Came up with a few points that we can start with.

1. The component reference.  These remain as reference topics with arguments, blocks, exports, health, debug info, and **one** minimal single-component syntax snippet (just enough to show usage). Real examples are elsewhere.
2. he alloy-scenarios could become the single source of runnable configurations. This external repo has a growing number of docker container examples that can be run local and used as a learning tool.
3. Expand the `choose-component.md` topic. This is really high level right now, but this could be expanded into a better grounded guidance for people, answering the "where do I go?" question.
4. Rework the current (and very stale)  `collect/`, `monitor/`, and `tutorials/` content into a single (?) format/template style (it's really mixed up right now). Strip out the alloy-scenarios (there are 8?) that are documented here (it just duplicates the scenario readmes... was put in place before the scenarios repo matured).

ptodev:
> 1. The component reference.  These remain as reference topics with arguments, blocks, exports, health, debug info, and **one** minimal single-component syntax snippet (just enough to show usage). Real examples are elsewhere.

SGTM 👍 

> 2. The alloy-scenarios could become the single source of runnable configurations. This external repo has a growing number of docker container examples that can be run local and used as a learning tool.

SGTM 👍 There are too many possible configs. It's good to outsource them to a repo where people can also easily contribute something new, and where CI can check them. The scenarios repo is quite good for this.

> 3. Expand the `choose-component.md` topic. This is really high level right now, but this could be expanded into a better grounded guidance for people, answering the "where do I go?" question.

Could do, as long as it doesn't turn it into a really large doc :) 

> 4. Rework the current (and very stale)  `collect/`, `monitor/`, and `tutorials/` content into a single (?) format/template style (it's really mixed up right now). Strip out the alloy-scenarios (there are 8?) that are documented here (it just duplicates the scenario readmes... was put in place before the scenarios repo matured).

+1 for having just one section. How do we decide what will be in alloy-secnarios repo and what will be in the official docs "tasks" section? I know "tasks" may not be the official name, but I suppose that's where the task docs we've been talking about will go - those would be the docs that mostly replace the Example sections in reference docs?

PR #5855, opened 2026-03-21
feat(loki.secretfilter): Add `secrets_redacted_by_category_total` metric combining rule and origin

kleimkuhler:
### Brief description of Pull Request

Adds a new `loki_secretfilter_secrets_redacted_by_category_total` counter metric with `rule` and `origin` labels, combining the information from the existing `secrets_redacted_by_rule_total` and `secrets_redacted_by_origin` metrics into a single metric. This will allow us to correlate which rules are firing for which jobs

### Pull Request Details

Unlike `secrets_redacted_by_origin` (which is only registered when `origin_label` is set), this metric is always registered. When `origin_label` is not configured or the label is absent on a log entry, the `origin` label value defaults to an empty string.

The existing metrics are unchanged. We can consider removing them in a future change.

### Issue(s) fixed by this Pull Request

Closes https://github.com/grafana/secops-research/issues/284

### PR Checklist

- [x] Documentation added
- [x] Tests updated

PR #5970, opened 2026-04-01
feat(loki.secretfilter): Remove redundant `secrets_redacted_by_rule_total` and `secrets_redacted_by_origin metrics`

kleimkuhler:

The `loki_secretfilter_secrets_redacted_by_category_total` metric (introduced in #5855) partitions by both `rule` and `origin`, making the two single-dimension metrics redundant. Any query with them can be expressed via `sum by (rule|origin) (...)` on the category metric.

This change removes those metrics now that they are no longer needed.

Redundant metrics lead to high cardinality and complicate secretfilter.

### PR Checklist

- [x] Documentation updated
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.secretfilter.md
+++ b/docs/sources/reference/components/loki/loki.secretfilter.md
@@ -52,7 +52,7 @@ You can use the following arguments with `loki.secretfilter`:
 | `drop_on_timeout`    | `bool`               | When true, drop entries that exceed `processing_timeout` instead of forwarding them unredacted.                             | `false` | no       |
 | `gitleaks_config`    | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                                  | `""`    | no       |
 | `label_timed_out`    | `bool`               | When true, adds `secretfilter="timed-out"` to entries forwarded after a processing timeout.                                 | `false` | no       |
-| `origin_label`       | `string`             | Loki label to use as the `origin` dimension in `secrets_redacted_by_origin` and `secrets_redacted_by_category_total`. If empty, `secrets_redacted_by_origin` is not registered and the `origin` label on `secrets_redacted_by_category_total` is set to `""`. | `""`    | no       |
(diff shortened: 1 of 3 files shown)


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -52,7 +52,7 @@ You can use the following arguments with `loki.secretfilter`:
 | `drop_on_timeout`    | `bool`               | When true, drop entries that exceed `processing_timeout` instead of forwarding them unredacted.                             | `false` | no       |
 | `gitleaks_config`    | `string`             | Path to a custom Gitleaks TOML config file. If empty, the default Gitleaks config is used.                                  | `""`    | no       |
 | `label_timed_out`    | `bool`               | When true, adds `secretfilter="timed-out"` to entries forwarded after a processing timeout.                                 | `false` | no       |
-| `origin_label`       | `string`             | Loki label to use as the `origin` dimension in `secrets_redacted_by_origin` and `secrets_redacted_by_category_total`. If empty, `secrets_redacted_by_origin` is not registered and the `origin` label on `secrets_redacted_by_category_total` is set to `""`. | `""`    | no       |
+| `origin_label`       | `string`             | Loki label to use as the `origin` dimension in `secrets_redacted_by_category_total`. If empty, the `origin` label on `secrets_redacted_by_category_total` is set to `""`. | `""`    | no       |
```
mikefat: The `|` should be aligned for all the rows so they need some refactoring here

mikefat (commented):
LGTM just minor comment on the docs formatting

docs/sources/reference/components/loki/loki.secretfilter.md
clayton-cornell: Before any table refactoring... I think we could drop the second sentence in the description.  The default label is `""` and the second sentence is... if I read it right... saying that if you don't set the label it defaults to the default.

kleimkuhler: Done!

kalleep (approved):

mikefat (approved):

PR #5974, opened 2026-04-02
chore: Add integration tests for loki.source.syslog

kalleep:
### Pull Request Details

Add a new Docker integration test for `loki.source.syslog` that sends RFC5424 and RFC3164 syslog messages over both TCP and UDP.

### Notes to the Reviewer

This also updates `common.AssertLogsPresent` to take an explicit total log count so tests can assert multiple label combinations.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated
- [ ] Config converters updated****

### DIFF
--- a/integration-tests/docker/common/logs_assert.go
+++ b/integration-tests/docker/common/logs_assert.go
@@ -32,22 +32,15 @@ type ExpectedLogResult struct {
 	EntryCount int
 }
 
-// AssertLogsPresent checks that logs are present in Loki and match expected labels
-func AssertLogsPresent(t *testing.T, expected ...ExpectedLogResult) {
+// AssertLogsPresent checks that logs are present in Loki and match expected labels.
+func AssertLogsPresent(t *testing.T, totalCount int, expected ...ExpectedLogResult) {
 	t.Helper()
 	AssertStatefulTestEnv(t)
 
-	var (
-		totalExpected int
-		logResponse   LogResponse
-	)
-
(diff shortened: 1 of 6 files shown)


integration-tests/docker/common/logs_assert.go
```diff
@@ -32,30 +32,23 @@ type ExpectedLogResult struct {
 	EntryCount int
 }
 
-// AssertLogsPresent checks that logs are present in Loki and match expected labels
-func AssertLogsPresent(t *testing.T, expected ...ExpectedLogResult) {
+// AssertLogsPresent checks that logs are present in Loki and match expected labels.
+func AssertLogsPresent(t *testing.T, totalCount int, expected ...ExpectedLogResult) {
```
thampiotr, line 36: Nit: I find the calls to this a bit harder to read now as they have 
```
common.AssertLogsPresent(t, 3, common.ExpectedLogResult{ ...
```
in above that `3` is not clear what it means

kalleep: Yeah I agree that it's not great, before we calculated total count by walking all expected result and adding up EntryCount for them. But this did not allow us to perform different checks for the same entries.

Now that we have separate them we can assert that We have 5 entries with `format = rfc5424 protocol = tcp` and two of them have labels `format = rfc5424 severity = debug` without affecting the total count. 

I will merge as is but if you have any suggestion to make this better I can fix that in future pr:s

PR #5985, opened 2026-04-03
ci: Wait for docker on windows runners

jharvey10:
To avoid errors such as this: https://github.com/grafana/alloy/actions/runs/23961165082

### DIFF
--- a/.github/workflows/publish-alloy-windows.yml
+++ b/.github/workflows/publish-alloy-windows.yml
@@ -35,6 +35,23 @@ jobs:
         os: [windows-2022]
     runs-on: ${{ matrix.os }}
     steps:
+    - name: Wait for Docker to be ready
+      run: |
+        $retries = 0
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-alloy-windows.yml
ptodev, line 1: There's [official advice](https://github.com/actions/runner-images/issues/13729#issuecomment-4183956802) to try starting the docker service:

```yaml
      - name: Docker issue workaround
        id: docker-workaround
        run: |
          $dockerState = Get-Service -Name "docker"
          if ($dockerState.Status -ne "Running") {
            Start-Service -Name "docker"
          }
```

Let's switch to that?

jharvey10: Great find! Switched over to that logic

x1unix (approved):

PR #5994, opened 2026-04-06
perf(loki.secretfilter): Small improvements in `redactLine`

kleimkuhler:

Two small performance improvements to the `loki.secretfilter` hot path

### Pull Request Details

- Move origin label lookup out of the per-finding loop in `redactLine`
- Remove per-redaction overhead of new hasher and `fmt.Sprintf`


kleimkuhler:
This will likely conflict with https://github.com/grafana/alloy/pull/5970

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -428,9 +433,8 @@ func (c *Component) redactLine(entry loki.Entry, findings []report.Finding) loki
 
 // hashSecret returns a short hex-encoded SHA1 hash of the secret for use in redaction placeholders.
 func hashSecret(secret string) string {
-	hasher := sha1.New()
-	hasher.Write([]byte(secret))
-	return fmt.Sprintf("%x", hasher.Sum(nil))
+	sum := sha1.Sum([]byte(secret))
```
ghost, line 406: <strike>Detected SHA1 hash algorithm which is considered insecure. SHA1 is not collision resistant and is therefore not suitable as a cryptographic signature. Use SHA256 or SHA3 instead.</strike>

✨ _Fixed in commit 66b1a1ec17fd07b08ade342763cac72883c6fe3c_ ✨

kalleep (approved):

kleimkuhler:
After some team discussion, we changed back to sha1 as mentioned in the commit message

> We originally opted for SHA1 in large part for speed - the goal with hashing
the secret is not to keep it entirely secret, but to efficiently conceal it
from all but the most determined attackers.
>
> After running benchmarks, sha256 is 2.09% slower, and we're looking for
every gain we can take right now.

mikefat (approved):
LGTM

PR #5998, opened 2026-04-07
chore(loki): Add integration test for loki.source.docker

kalleep:
### Pull Request Details

Add Docker integration test for `loki.source.docker`. This test also includes a simple `loki.relabel` and `loki.process` stages.

### Notes to the Reviewer

This also extends the Docker integration-test so each test can start additional containers, build their images, and mount the Docker socket into the Alloy container.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/docker/tests/loki-docker/test.yaml
+++ b/integration-tests/docker/tests/loki-docker/test.yaml
@@ -1,0 +1,17 @@
+alloy_container:
+  use_docker_sock: true
+
+additional_containers:
+  - name: producer-1
+    image: loki-docker-producer:latest
+    build:
+      context: .
+      dockerfile: Dockerfile
+  - name: producer-2
+    image: loki-docker-producer:latest
+  - name: producer-3
+    image: loki-docker-producer:latest
+  - name: producer-4
+    image: loki-docker-producer:latest
+ 
+ 
--- a/integration-tests/docker/README.md
+++ b/integration-tests/docker/README.md
@@ -32,7 +32,7 @@ Follow these steps to add a new integration test to the project:
 2. Create a new directory under the tests directory to house the files for the new test.
 3. Within the new test directory, create a file named `config.alloy` to hold the pipeline configuration you want to test.
 4. Create a `_test.go` file within the new test directory. This file should contain the Go code necessary to run the test and verify the data processing through the pipeline. All test file should have `alloyintegrationtests` as a build tag.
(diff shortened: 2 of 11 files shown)


integration-tests/docker/tests/loki-docker/test.yaml
```diff
@@ -0,0 +1,17 @@
+alloy_container:
+  use_docker_sock: true
+
+additional_containers:
+  - name: producer-1
```
blewis12, line 5: just so I understand correctly, is the reason we only have `build/*` args only in `producer-1` because we only want to build the image once and then in the subsequent references (`product-2` ... `producer-4`) it's already built so we can grab from the local docker repo? 

If so, I think maybe we could represent it in a way in yaml that doesn't rely on folks having to understand that, maybe something like

```
images:
  producer:
    build:
      context: .
      dockerfile: Dockerfile
    tag: loki-docker-producer:latest

additional_containers:
  - name: producer-1
    image: producer
  - name: producer-2
    image: producer
  - name: producer-3
    image: producer
  - name: producer-4
    image: producer
```

wdyt?

kalleep: Yes you understand it correctly but I am not sure about that. This is how docker compose does it and I just implemented it with the same properties and the same behavior

blewis12: I see - if this is how docker compose does it, could we make sure it's implemented in a way that's not relying on order? The way i understand it, if someone accidentally did

```
additional_containers:
  - name: producer-1
    image: loki-docker-producer:latest
  - name: producer-2
    image: loki-docker-producer:latest
  - name: producer-3
    image: loki-docker-producer:latest
  - name: producer-4
    image: loki-docker-producer:latest
    build:
      context: .
      dockerfile: Dockerfile
 ```
 
 then we wouldn't be able to grab `producter-1`...`producer-3` because we haven't yet built the image eh? If this is the way docker-compose behaves maybe that's fine, but if I recall that's not relying on ordering

kalleep: Yes that is a good idea!

kalleep: fixed in [d7ab8d5](https://github.com/grafana/alloy/pull/5998/commits/d7ab8d51d8f76341deb5f9859e0524dfbb495b6b)

blewis12 (approved):

PR #6010, opened 2026-04-08
feat: Add zone-aware scraping to prometheus operator CRDs

timonegk:

This pull request adds a new option `zone_aware` to the `clustering` section in the `prometheus.operator.*` components that enables zone-aware clustering. When the option is set to `true`, each Alloy instance has a zone-local hash rings that only contain zone-local targets (determined by a Pod's Node). This option heavily reduces costly cross-zone traffic.

### Pull Request Details

In our Alloy setup, we deploying across three availability zones. Since Alloy is not zone-aware, its scraping behavior results in a lot of expensive cross-zone traffic. While it would be possible to deploy three zone-local alloy deployments that each filter for targets in the own zone, I thought that it would be nicer and more useful for other people to integrate the functionality directly into Alloy.
I tested the change in a 25-node, three-zone cluster with six Alloy instances. The cross-zone traffic by Alloy was reduced from up to 80% to less than 0.5%.
The change is not breaking because the default is `zone_aware=false` which does not change the current behavior.

### Notes to the Reviewer

This PR was created with the help of AI. I reviewed the changes thoroughly, made adaptions where necessary and deployed the code to a live cluster, as mentioned above.
I made some decisions that we can discuss:
- I added the option with the name `zone_aware` to the clustering section of the prometheus.operator components. Maybe there is a better name or place for the option.
- I am using K8S_NODE_NAME to get the current node. Some components use HOSTNAME, but i figured K8S_NODE_NAME is semantically more correct and since #5484 it is always present when rolled out with Helm.
- I am rebuilding the zone-local rings on every update of the discovery manager. Maybe that is too often. We could also do it periodically (unknown nodes fall through to the global ring).
- I was also playing with attach_metadata to attach the node's zone metadata directly to the pod. In the end, using the node name of the pod was the simplest and most reliable approach.
- I added (=generated) a lot of tests. We can also reduce them if you think it's too much.
- I also added debug info and info logs. We can reduce/remove them if you want.
- I originally planned to implement it only for PodMonitors and ServiceMonitors. ScrapeConfigs and Probes came for free. For Probes, the change never applies because they are only discovered through static config or ingress. For ScrapeConfig, it can theoretically apply if `kubernetesSDConfigs` is used.
- @clayton-cornell I added the same documentation to all four components. If you have a better idea, feel free to suggest it.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
+++ b/docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
@@ -125,9 +125,10 @@ The following arguments are supported:
 
 ### `clustering`
 
-| Name      | Type   | Description                                       | Default | Required |
-| --------- | ------ | ------------------------------------------------- | ------- | -------- |
-| `enabled` | `bool` | Enables sharing targets with other cluster nodes. | `false` | yes      |
+| Name         | Type   | Description                                                              | Default | Required |
+| ------------ | ------ | ------------------------------------------------------------------------ | ------- | -------- |
+| `enabled`    | `bool` | Enables sharing targets with other cluster nodes.                        | `false` | yes      |
+| `zone_aware` | `bool` | Enables availability-zone-aware target filtering when clustering is on.  | `false` | no       |
 
 When {{< param "PRODUCT_NAME" >}} is [using clustering][], and `enabled` is set to true, then this component instance opts-in to participating in the cluster to distribute scrape load between all cluster nodes.
 
@@ -139,6 +140,19 @@ When a node joins or leaves the cluster, every peer recalculates ownership and c
 This performs better than hashmod sharding where _all_ nodes have to be re-distributed, as only 1/N of the target's ownership is transferred, but is eventually consistent (rather than fully consistent like hashmod sharding is).
 
 If {{< param "PRODUCT_NAME" >}} is _not_ running in clustered mode, then the block is a no-op, and `prometheus.operator.podmonitors` scrapes every target it receives in its arguments.
+
+#### Zone-aware clustering
+
+When `zone_aware` is set to `true`, each {{< param "PRODUCT_NAME" >}} instance only scrapes targets that are in the same availability zone as the instance itself.
+Targets in a different zone are skipped before the hash ring is consulted, which reduces cross-zone network traffic.
+Targets whose zone can't be determined fall through to normal hash ring distribution so they aren't dropped.
+You should make sure that {{< param "PRODUCT_NAME" >}} is deployed in all availability zones, e.g. by using `topologySpreadConstraints`.
+
+The component determines the local availability zone by reading the `topology.kubernetes.io/zone` label from the Kubernetes node object.
+This requires the `K8S_NODE_NAME` environment variable to be set, typically through the [Kubernetes downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).
+The default {{< param "PRODUCT_NAME" >}} Helm chart already sets this variable.
+
+Setting `zone_aware` to `true` requires `enabled` to also be `true`.
 
 [using clustering]: ../../../../get-started/clustering/
 
--- a/docs/sources/reference/components/prometheus/prometheus.operator.probes.md
+++ b/docs/sources/reference/components/prometheus/prometheus.operator.probes.md
@@ -125,9 +125,10 @@ The following arguments are supported:
 
 ### `clustering`
 
-| Name      | Type   | Description                                       | Default | Required |
-| --------- | ------ | ------------------------------------------------- | ------- | -------- |
-| `enabled` | `bool` | Enables sharing targets with other cluster nodes. | `false` | yes      |
+| Name         | Type   | Description                                                              | Default | Required |
+| ------------ | ------ | ------------------------------------------------------------------------ | ------- | -------- |
+| `enabled`    | `bool` | Enables sharing targets with other cluster nodes.                        | `false` | yes      |
+| `zone_aware` | `bool` | Enables availability-zone-aware target filtering when clustering is on.  | `false` | no       |
 
 When {{< param "PRODUCT_NAME" >}} is running in [clustered mode][], and `enabled` is set to true, then this component instance opts-in to participating in the cluster to distribute scrape load between all cluster nodes.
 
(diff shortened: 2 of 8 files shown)


docs/sources/reference/components/prometheus/prometheus.operator.podmonitors.md
```diff
@@ -140,6 +141,19 @@ This performs better than hashmod sharding where _all_ nodes have to be re-distr
...
+When `zone_aware` is set to `true`, each {{< param "PRODUCT_NAME" >}} instance only scrapes targets that are in the same availability zone as the instance itself.
+Targets in a different zone are skipped before the hash ring is consulted, which reduces cross-zone network traffic.
+Targets whose zone can't be determined fall through to normal hash ring distribution so they aren't dropped.
+You should make sure that {{< param "PRODUCT_NAME" >}} is deployed in all availability zones, e.g. by using `topologySpreadConstraints`.
+
+The component determines the local availability zone by reading the `topology.kubernetes.io/zone` label from the Kubernetes node object.
+This requires the `K8S_NODE_NAME` environment variable to be set, typically through the [Kubernetes downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).
+The default {{< param "PRODUCT_NAME" >}} Helm chart already sets this variable.
+
+Setting `zone_aware` to `true` requires `enabled` to also be `true`.
```
clayton-cornell, line 155:
```suggestion
When you set `zone_aware` to `true`, each {{< param "PRODUCT_NAME" >}} instance scrapes targets in its availability zone.
The component skips targets in different zones before it consults the hash ring, which reduces cross-zone network traffic.
If the component can't determine a target's zone, it uses the normal hash ring distribution so it doesn't drop the target.
Deploy {{< param "PRODUCT_NAME" >}} in every availability zone, for example with `topologySpreadConstraints`.

The component reads the `topology.kubernetes.io/zone` label from the Kubernetes node object to learn the local availability zone.
Set the `K8S_NODE_NAME` environment variable through the [Kubernetes downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).
The default {{< param "PRODUCT_NAME" >}} Helm chart sets this variable.

If `zone_aware` is `true`, then `enabled` must be `true` too.
```

Some rework to make it active voice, remove gerunds (ing words), and better align with Grafana doc styles.

clayton-cornell: These changes should be applied in the other topics as well

docs/sources/reference/components/prometheus/prometheus.operator.probes.md
```diff
@@ -140,6 +141,24 @@ This performs better than hashmod sharding where _all_ nodes have to be re-distr
...
+#### Zone-aware clustering
+
+When `zone_aware` is set to `true`, each {{< param "PRODUCT_NAME" >}} instance only scrapes targets that are in the same availability zone as the instance itself.
+Targets in a different zone are skipped before the hash ring is consulted, which reduces cross-zone network traffic.
+Targets whose zone can't be determined fall through to normal hash ring distribution so they aren't dropped.
+You should make sure that {{< param "PRODUCT_NAME" >}} is deployed in all availability zones, e.g. by using `topologySpreadConstraints`.
+
+{{< admonition type="note" >}}
+Zone-aware clustering only applies to targets with Kubernetes pod metadata (when Probe targets are discovered through Kubernetes service discovery).
+For Probe resources that target arbitrary IPs or hostnames without pod metadata, the `zone_aware` setting has no effect and targets are distributed using the global cluster ring.
```
clayton-cornell, line 153:
```suggestion
Zone-aware clustering applies only when Probe targets include Kubernetes Pod metadata from Kubernetes service discovery.
For Probe resources that target arbitrary IP addresses or hostnames without Pod metadata, `zone_aware` has no effect.
The component still assigns those targets on the global cluster ring.
```

docs/sources/reference/components/prometheus/prometheus.operator.scrapeconfigs.md
```diff
@@ -182,6 +183,24 @@ This performs better than hashmod sharding where _all_ nodes have to be re-distr
...
+#### Zone-aware clustering
+
+When `zone_aware` is set to `true`, each {{< param "PRODUCT_NAME" >}} instance only scrapes targets that are in the same availability zone as the instance itself.
+Targets in a different zone are skipped before the hash ring is consulted, which reduces cross-zone network traffic.
+Targets whose zone can't be determined fall through to normal hash ring distribution so they aren't dropped.
+You should make sure that {{< param "PRODUCT_NAME" >}} is deployed in all availability zones, e.g. by using `topologySpreadConstraints`.
+
+{{< admonition type="note" >}}
+Zone-aware clustering only applies to targets discovered via Kubernetes pod service discovery (when ScrapeConfig uses Pod or Endpoint discovery).
+For other service discovery methods (static configs, HTTP SD, etc.), the `zone_aware` setting has no effect and targets are distributed using the global cluster ring.
```
clayton-cornell, line 195:
```suggestion
Zone-aware clustering applies only when Probe targets include Kubernetes Pod metadata from Kubernetes service discovery.
For Probe resources that target arbitrary IP addresses or hostnames without Pod metadata, `zone_aware` has no effect.
The component still assigns those targets on the global cluster ring.
```

clayton-cornell: I've only added a suggestion on the note here. the other text looks the same as the previous topic and the suggestions there should be applied here as well.

timonegk:
@clayton-cornell Thanks for the review. I applied your suggestions but I removed all docs changes from `prometheus.operator.probes` because I think my change actually doesn't apply there because it is not possible to configure a probe target using Kubernetes Service Discovery.

clayton-cornell:
>  I removed all docs changes from `prometheus.operator.probes` because I think my change actually doesn't apply there because it is not possible to configure a probe target using Kubernetes Service Discovery.

That may need some additional verification. I poked it with Claude, and it stated that 

1. You can configure Probe targets using Kubernetes service discovery, by using the Ingress target type on the Probe.
2. The note around “Kubernetes service discovery” + “pod metadata” does not match the implementation: neither Probe mode in this codebase attaches `__meta_kubernetes_pod_node_name`, so availability-zone-aware filtering does not apply to Probe-discovered targets the way it can for SD roles that expose that meta label (pods/endpoints/endpoint slices).

It then suggested the following updated admonition:

```
Zone-aware clustering maps each target to an availability zone using the Prometheus meta label `__meta_kubernetes_pod_node_name`.
A Probe scrape configuration uses only static targets or Kubernetes Ingress discovery, and neither path adds that label to targets.
For `prometheus.operator.probes`, `zone_aware` doesn't filter targets by zone.
Those targets stay on the global cluster ring.
```

I'll leave that to you and the @grafana/grafana-alloy-maintainers to comment on whether this AI finding is valid...

timonegk:
My understanding is that it is possible to configure Kubernetes Probe targets using the Ingress type, and they do use KubernetesSD, but no `__meta_kubernetes_pod_node_name` labels are attached. See [this documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#ingress) where pod (used by PodMonitor) and endpoint/endpointslice (used by ServiceMonitor) state that the pod-specific labels are added, but ingress does not. That's why I removed the documentation for Probe (I also adapted the "notes to the reviewer" accordingly).

kgeckhart:
👋  thanks for the PR I appreciate that you validated it in production before submitting. It's certainly a really useful feature. 

I think it is complex enough it warrants a https://github.com/grafana/alloy/blob/main/docs/design/README.md before we commit to an implementation — would you be open to starting there? We are currently evaluating options for zone aware scraping ourselves and were leaning towards isolated deployments to allow for independent scaling. Maintaining this implementation could be a real challenge for us if we don't have an easy test setup. 

Something else I was wondering was this would handle a situation where multi-AZ is setup but alloy is not deployed to all AZs? It looks like they wouldn't be scraped and we wouldn't know. Multiple deployments doesn't solve this either but I was just wondering if my read on it was correct.

timonegk:
Hi @kgeckhart and thanks for your kind comment. I can certainly set up a proposal document but will have to look into the process. I also thought about isolated deployments but I wanted to have the best user experience, and setting up several deployments is certainly more cumbersome. Independent scaling is not really an issue for us because the workloads are evenly spread across zones. Since zones are usually used to ensure high availability, I think it would be valid to assume that even scaling is the case in most setups.
Regarding zones without Alloy instances, you are correct in your assumption that they wouldn't be scraped. As you said, you would have the same problem with multiple deployments. That's also why I added to the documentation that users have to make sure that the pods are spread across all zones. I could probably add a safeguard to ensure the spread, but anything that goes beyond a warning message would probably increase complexity. But we can discuss that as well once the proposal document is set up.

timonegk:
Hi @kgeckhart, I proposed a design document in #6128. Please take a look, especially if I correctly captured your proposal in option 2.

PR #6026, opened 2026-04-09
feat(loki.source.file): Add option to set `max_line_size`

kalleep:
Add config `max_line_size` to protect against reading to large lines. Currently it is possible to OOM alloy if you read a really big file without any newlines.

Part of: #6014

### Notes to the Reviewer

This pr includes some small cleanups.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.file.md
+++ b/docs/sources/reference/components/loki/loki.source.file.md
@@ -45,12 +45,15 @@ You can use the following arguments with `loki.source.file`:
 | `targets`                 | `list(map(string))`  | List of files to read from.                                    |                            | yes      |
 | `encoding`                | `string`             | The encoding to convert from when reading files.               | `""`                       | no       |
 | `legacy_positions_file`   | `string`             | Allows conversion from legacy positions file.                  | `""`                       | no       |
+| `max_line_size`           | `string`             | Maximum size of a single emitted log line.                     | `"1MiB"`                   | no       |
 | `on_positions_file_error` | `string`             | How to handle a corrupt positions file entry for a given file. | `"restart_from_beginning"` | no       |
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
 The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} can automatically change
 the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
 The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
+
+The `max_line_size` argument limits the size of a single emitted log line. If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads. Set `max_line_size` to `0` to disable this limit.
 
 You can use the `tail_from_end` argument when you want to tail a large file without reading its entire content.
 When set to true, only new logs are read, ignoring the existing ones.
--- a/docs/sources/reference/components/loki/loki.source.file.md
+++ b/docs/sources/reference/components/loki/loki.source.file.md
@@ -53,7 +53,9 @@ The `encoding` argument must be a valid [IANA encoding][] name and if not set, i
 the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
 The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
 
-The `max_line_size` argument limits the size of a single emitted log line. If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads. Set `max_line_size` to `0` to disable this limit.
+The `max_line_size` argument limits the size of a single emitted log line.
+If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads.
+Set `max_line_size` to `0` to disable this limit.
 
 You can use the `tail_from_end` argument when you want to tail a large file without reading its entire content.
 When set to true, only new logs are read, ignoring the existing ones.
--- a/docs/sources/reference/components/loki/loki.source.file.md
+++ b/docs/sources/reference/components/loki/loki.source.file.md
@@ -45,16 +45,12 @@ You can use the following arguments with `loki.source.file`:
 | `targets`                 | `list(map(string))`  | List of files to read from.                                    |                            | yes      |
 | `encoding`                | `string`             | The encoding to convert from when reading files.               | `""`                       | no       |
 | `legacy_positions_file`   | `string`             | Allows conversion from legacy positions file.                  | `""`                       | no       |
-| `max_line_size`           | `string`             | Maximum size of a single emitted log line.                     | `"1MiB"`                   | no       |
 | `on_positions_file_error` | `string`             | How to handle a corrupt positions file entry for a given file. | `"restart_from_beginning"` | no       |
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
 The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} can automatically change
 the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
 The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
-
-The `max_line_size` argument limits the size of a single emitted log line, set to `0` disable.
-If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads.
 
 You can use the `tail_from_end` argument when you want to tail a large file without reading its entire content.
 When set to true, only new logs are read, ignoring the existing ones.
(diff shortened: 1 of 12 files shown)


internal/component/loki/source/file/file.go
```diff
@@ -344,16 +346,33 @@ func (c *Component) scheduleSources() {
...
-				path:                 target.Path,
-				labels:               target.Labels,
-				encoding:             c.args.Encoding,
-				decompressionConfig:  c.args.DecompressionConfig,
-				fileWatch:            c.args.FileWatch,
-				tailFromEnd:          c.args.TailFromEnd,
-				onPositionsFileError: c.args.OnPositionsFileError,
-				legacyPositionUsed:   c.args.LegacyPositionsFile != "",
-			})
+			tailer := newTailer(
```
kalleep, line 360: Just removed the helper `newSource` since it's not needed any more

internal/component/loki/source/file/tailer.go
```diff
@@ -30,23 +30,27 @@ type tailer struct {
...
-	key                positions.Entry
-	labels             model.LabelSet
-	legacyPositionUsed bool
-
-	tailFromEnd          bool
-	onPositionsFileError OnPositionsFileError
-	watcherConfig        tail.WatcherConfig
-
-	running *atomic.Bool
+	opts   tailerOptions
```
kalleep, line 32: Simplified this alot. Instead of storing every property from options we just store options

kalleep:
@codex review

kalleep:
@codex review

kalleep:
@codex review

internal/component/loki/source/file/file.go
```diff
@@ -84,6 +85,7 @@ func (o *OnPositionsFileError) UnmarshalText(text []byte) error {
 func (a *Arguments) SetToDefault() {
 	a.FileWatch.SetToDefault()
 	a.FileMatch.SetToDefault()
+	a.MaxLineSize = 1 * units.MiB
```
kalleep: Before this was unlimited and by default loki will reject entries over 256KB.

But we should find a reasonable default here.

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -45,13 +45,16 @@ You can use the following arguments with `loki.source.file`:
...
 | `on_positions_file_error` | `string`             | How to handle a corrupt positions file entry for a given file. | `"restart_from_beginning"` | no       |
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
 The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} can automatically change
 the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
 The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
 
+The `max_line_size` argument limits the size of a single emitted log line. If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads. Set `max_line_size` to `0` to disable this limit.
```
```suggestion
The `max_line_size` argument limits the size of a single emitted log line.
If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads.
Set `max_line_size` to `0` to disable this limit.
```

Just putting each sentence on its own line

clayton-cornell (commented):
Minor suggestion. Everything else is OK for doc input.

docs/sources/reference/components/loki/loki.source.file.md
```diff
@@ -45,13 +45,18 @@ You can use the following arguments with `loki.source.file`:
...
 | `tail_from_end`           | `bool`               | Whether to tail from end if a stored position isn't found.     | `false`                    | no       |
 
 The `encoding` argument must be a valid [IANA encoding][] name and if not set, it defaults to UTF-8. {{< param "PRODUCT_NAME" >}} can automatically change
 the encoding to `UTF-16` if the file includes a Byte Order Mark (BOM) for either `UTF-16BE` or `UTF-16LE`.
 The BOM will be taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file. This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
 
+The `max_line_size` argument limits the size of a single emitted log line.
+If a line grows beyond this limit before a newline is encountered, {{< param "PRODUCT_NAME" >}} emits the first `max_line_size` bytes and continues reading the remainder on subsequent reads.
+Set `max_line_size` to `0` to disable this limit.
+
```
jharvey10, line 59: Wondering if it would make sense to be able to configure this behavior? maybe so that you could either chunk it or drop it? I'm thinking if someone had huge json log lines, chunking them could break downstream stages and isn't all that helpful as a chunk.

I think you could kind of do that with like `max_line_size 1.1MiB` followed by `stage.drop { longer_than "1MiB"}`, but it feels like it might be nice to drop it right at the source.

kalleep: >Wondering if it would make sense to be able to configure this behavior?

Not sure I follow, this is already configurable in the pr. Or are you referring to something else?

internal/component/loki/source/file/internal/tail/reader.go
```diff
@@ -137,21 +143,58 @@ func (r *reader) decode(line []byte) (string, error) {
 // consumeLine checks pending for the delimiter; if found, it splits
 // pending into line and remainder.
 func (r *reader) consumeLine() ([]byte, bool) {
-	// Check if pending contains a full line.
-	i := bytes.Index(r.pending, r.nl)
-	if i < 0 {
+	// If we have no pending data we need to read more.
+	if len(r.pending) == 0 {
```
blewis12: Is `len(r.pending) == 0` (no pending data left) the same condition as `bytes.Index(r.pending, r.nl) < 0` (pending does not contain new line) - or is there another reason for this particular check? I would have thought that the final `return nil, false` includes this case

internal/component/loki/source/file/internal/tail/reader.go
```diff
@@ -62,20 +62,22 @@ func newReader(logger log.Logger, f *os.File, offset int64, enc encoding.Encodin
...
 }
 
 type reader struct {
-	pos     int64
-	br      *bufio.Reader
-	pending []byte
+	br          *bufio.Reader
+	pos         int64
+	pending     []byte
+	maxLineSize int
```
blewis12, line 78: could we not have this be a `uint` and validate against the incoming `int` to ensure it's >0? `fitsMaxLineSize` accounts for a value <0 but I don't think we have to

docs/sources/reference/components/loki/loki.source.file.md
kalleep: Oh you mean discarding the full line if it's too big? Maybe.. I think we could have behavioral config for when line reached the limit:
1. Default split/chunk (current)
2. Emit first chunk, discard until newline.
3. Drop until newline

internal/component/loki/source/file/file.go
blewis12: Yeah I think 1 MiB is a totally fine limit to start with, a log line with ~1mil characters is pretty unreadable anyway - so having that split into more entries feels OK. I know we're throwing ideas in the team to have OOTB grafana defaults for our config too, maybe that would be the time to override this with 256KB

internal/component/loki/source/file/internal/tail/reader.go
kalleep: They annoying thing about using uint here is that we have cast is a a couple of places, but it's technically more correct.

In [dca8709](https://github.com/grafana/alloy/pull/6026/commits/dca8709feef79316ff8059f7355cffed15c44748) I changed to that we always set `maxLineSize` to a positive value

blewis12: ah right, do you think it could make more sense to explicitly fail validation if we detect a negative value instead of falling back to clamping like we do for `file_match .sync_period`? That seems to the pattern I see across alloy

kalleep: So as a argument we use[ units.Base2Bytes](https://github.com/grafana/alloy/blob/dca8709feef79316ff8059f7355cffed15c44748/internal/component/loki/source/file/file.go#L54). I am pretty sure you cannot get a negative value

blewis12: I see 👍 if the error message there looks OK to a user then sounds good to me

docs/sources/reference/components/loki/loki.source.file.md
blewis12: I like the idea of having some behavioral options - though perhaps as a follow up / in response to user feedback?

blewis12 (dismissed):
LGTM - lmk if you need a re-review though with the merge conflicts and/or any follow ups

kalleep:
Yeah so the only thing left to think about i what default value should be for `max_line_size`, 1MiB is technically a breaking change

blewis12:
> max_line_size

True, though isn't _any_ value a breaking change? It should at least be very lenient to start, which I think 1mil characters is quite. Unless we made the entire thing opt-in and have the default be `0`?

kalleep:
Yea, IMO `1MiB` should be enough, loki by default rejects entries that are larger than `256KB`. But we can always leave it as an opt-in.

blewis12:
> Yea, IMO `1MiB` should be enough, loki by default rejects entries that are larger than `256KB`. But we can always leave it as an opt-in.

Is a >1MiB log line a common enough scenario? My 2c is I think it makes sense to have sensible defaults in Alloy that protect against things like OOM and clearly document the change - but if it's actually not uncommon to have very long log lines in the wild then I get it could cause confusion to update Alloy and see an "increase" in log lines coming through from splits

kalleep:
>Is a >1MiB log line a common enough scenario

I would not think so but it's hard to say.

blewis12:
If we want to err on the side of caution we could always have it be opt-in, and if we get enough escalations or issues related to it we can re-think our defaults?

docs/sources/reference/components/loki/loki.source.file.md
jharvey10: Sorry, I totally lost track of this PR somehow :( But yeah you nailed it. Just some options for those behaviors you outlined was what I was suggesting.

kalleep: It's a great suggestion but lets leave that as a follow up, will create an issue for it. But this makes me wonder if `max_line_size` should be configured inside a block instead. Then we could group related config like these two under this block.

e.g.
```
loki.source.file "f" {
  line {
    max_size = "1MiB"
    behavior = "splitt"
  }
}
```

jharvey10: I could definitely see that being useful. Especially if we wanted to add more properties in the future.

Maybe something like `line_config`, `line_options`, `line_rules`, or `line_behavior`?

Not sure if we have precedence for one block name style versus another.

kalleep: I went with just `line` in [6200cca](https://github.com/grafana/alloy/pull/6026/commits/6200cca10b1c9fe635841765f3a5efe463e09fd8). I don't like the `_config` or `_option` suffix variants but the other two I guess could work.

For existing blocks we have `decompression`, `file_match` and `file_watch`.

jharvey10: Yeah, probably `line` works. I hadn't thought that if it were `line_behavior` then it would end up as:
```
line_behavior {
  behavior = "..."
}
```
behavior behavior lol. `line.behavior` reads nicer.

docs/sources/reference/components/loki/loki.source.file.md
```suggestion
The BOM is taken into account even if {{< param "PRODUCT_NAME" >}} resumes tailing a file from the middle of the file.
This can happen after {{< param "PRODUCT_NAME" >}} is restarted.
```

We can fix the Vale linting with a small change.

Issue #3717, opened 2025-05-28
Clustered loki.source.kubernetes produces duplicates when cluster nodes go down


skolyszewski:
### What's wrong?

We run Alloy in STS mode for collecting pod logs. We use `loki.source.kubernetes` with enabled clustering. 
We configured Alloy to use PV for preserving positions files during restarts - each instance has its own EBS-backed volume. When all cluster nodes are operational, everything works perfectly fine.
When one cluster node goes down, the other cluster nodes seem to be completely unaware of where the node that just disappeared left off wrt. to pod logs positions - that causes the remaining nodes to seemingly re-read the entire log stream for each pod, that was previously being read by the now-offline cluster node.

While Loki configured downstream seem to handle that by deduping these events, we also use `otelcol.exporter.awss3` which results in write spikes - see the image below

![Image](https://github.com/user-attachments/assets/889f9819-96b5-4b32-ab02-320015463897)

### Steps to reproduce

1. Set up Alloy in StatefulSet mode, with clustering enabled.
2. Configure pod discovery and use `loki.source.kubernetes` with clustering enabled.
3. Configure Alloy to send events to Loki and s3.
4. Bring down one of the alloy cluster nodes.
5. Observe spikes in events being written to all downstream components.

### Software version

v1.8.2 (we use grafana-alloy chart 1.0.2)

### Configuration

```text
// 1st step for log data collection - discover all pods in the cluster
discovery.kubernetes "pod" {
  role = "pod"
}

// Relabel Pod metadata for further processing and sending downstream
discovery.relabel "pod_logs" {
  targets = discovery.kubernetes.pod.targets

  rule {
    source_labels = ["__meta_kubernetes_namespace"]
    action = "replace"
    target_label = "namespace"
  }

  rule {
    source_labels = ["__meta_kubernetes_pod_name"]
    action = "replace"
    target_label = "pod"
  }

  rule {
    source_labels = ["__meta_kubernetes_pod_container_name"]
    action = "replace"
    target_label = "container"
  }

  rule {
    source_labels = ["__meta_kubernetes_pod_label_app_kubernetes_io_name"]
    action = "replace"
    target_label = "app"
  }

  rule {
    source_labels = ["__meta_kubernetes_pod_uid", "__meta_kubernetes_pod_container_name"]
    action = "replace"
    target_label = "__path__"
    separator = "/"
    replacement = "/var/log/pods/*$1/*.log"
  }
}

// Read logs from discovered and relabeled pods
[...]

PR #6027, opened 2026-04-09
feat: Add clustering for `loki.source.kubernetes_events`

### LINKED ISSUE #401: Make `loki.source.kubernetes_events` clusterable in flow-mode
### Request

I would like the loki source for `kubernetes_events` to be able to distribute the logs or only run on a single node when in clustered mode on the Grafana-Agent when deployed with Flow mode.

Unfortunately there doesn't seem to be a way to use `loki.source.kubernetes_events` if you are only running a cluster of Grafana-Agents in flow mode. If you add this to your configuration you will get duplicate logs.

### Use case

For ease of use, we are wanting to run a cluster of grafana-agents in flow-mode to handle our logging and metric ingestion needs. Right now we'd need to deploy some static grafana agents and some flow-mode agents to accomplish this.

petewall:

This change adds a `clustering` option for `loki.source.kubernetes_events` that will distribute the work according to the  list. If no namespaces are specificied, then only a single instance will run. This is great, because it means that it will be safe to run on Alloy instances with multiple replicas without resulting in duplication.

Fixes #401

### PR Checklist

- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.kubernetes_events.md
+++ b/docs/sources/reference/components/loki/loki.source.kubernetes_events.md
@@ -76,10 +76,12 @@ You can use the following blocks with `loki.source.kubernetes_events`:
 | `client` > [`oauth2`][oauth2]                    | Configure OAuth 2.0 for authenticating to the endpoint.    | no       |
 | `client` > `oauth2` > [`tls_config`][tls_config] | Configure TLS settings for connecting to the endpoint.     | no       |
 | `client` > [`tls_config`][tls_config]            | Configure TLS settings for connecting to the endpoint.     | no       |
+| [`clustering`][clustering]                       | Configure the component for when {{< param "PRODUCT_NAME" >}} is running in clustered mode. | no       |
 
 [authorization]: #authorization
 [basic_auth]: #basic_auth
 [client]: #client
+[clustering]: #clustering
 [oauth2]: #oauth2
 [tls_config]: #tls_config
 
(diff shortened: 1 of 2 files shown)


kalleep:
Cool but I do think we will have at least one problem, that `loki.source.kuberenetes` also is affected by. It's that we use position file for tracking where we last read from.

The problem here and with that other component is that if a target moves to another instance we loose track of where we should start reading from and would start from beginning.

Not really a blocker itself but we should look into how we can solve this for both this one and `loki.source.kubernetes`.

petewall:
A bit of testing:
1. Created a kind cluster
2. Deployed two alloys, both as 2-replica deployments, both using the same config:
```
          loki.source.kubernetes_events "cluster_events" {
            job_name   = "integrations/kubernetes/eventhandler"
            log_format = "logfmt"
            // These lines are only on the test alloy
            clustering {
              enabled = true
            }
          }

          loki.write "loki" {
            endpoint {
              url = "http://loki.loki.svc:3100/loki/api/v1/push"
              tenant_id = "1"
            }
            external_labels = {
              "source" = "control",
              "collector" = constants.hostname,
            }
          }
```
3. First alloy using latest published version, second alloy using my source-built image

Both control-group alloy replicas reported the same events (duplication)
only one of the test group alloy replicas send events (deduplication)

petewall:
> Cool but I do think we will have at least one problem, that `loki.source.kuberenetes` also is affected by. It's that we use position file for tracking where we last read from.
> 
> The problem here and with that other component is that if a target moves to another instance we loose track of where we should start reading from and would start from beginning.
> 
> Not really a blocker itself but we should look into how we can solve this for both this one and `loki.source.kubernetes`.

There's some good points here, but I'm not sure if it's big enough deal to hold this up. By default the TTL for kubernetes events is 1 hour, so any duplication would be bounded by that. And I feel the benefits (no more requirement for a singleton deployment) outweigh the concerns (potential short-term log duplication).

Is there an existing issue or a TODO for this, and we can try and fix that in another PR?

kalleep:
>Is there an existing issue or a TODO for this, and we can try and fix that in another PR?

Yeah there is a issue for [loki.source.kubernetes](https://github.com/grafana/alloy/issues/3717) as mentioned in slack.  But I saw that you added a note about this and that is enough I think.

kalleep (approved):

Issue #4813, opened 2025-11-10
internal: replace go-kit/log


kalleep:
### Add a description

Logging in alloy is a bit of a mess. Throughout the code base we use `go-kit/log` interface.
The actual logger that is used is [implemented here](https://github.com/grafana/alloy/blob/main/internal/runtime/logging/logger.go#L24). This logger implements bot `go-kit/log` interface and `log/slog` handler interface.

### Proposal
Use [zap](https://github.com/uber-go/zap). This is the logging library used by otel-collector.

We would need to implement [Core](https://github.com/uber-go/zap/blob/master/zapcore/core.go#L25) to handle dynamic updates and possibilities to forward logs to loki pipelines.

Then we would need to use `*zap.Logger` everywhere and rewrite various adapters.

kalleep:
After https://github.com/grafana/alloy/pull/6039 is merged I want to go through loki.process and adress this https://github.com/grafana/alloy/pull/6039#discussion_r3085284976.

PR #6031, opened 2026-04-10
chore: Use slog in runtime package

kalleep:
### Pull Request Details

Start moving to use `slog.Logger` instead of `go-kit/log`.

This will only change usage in runtime and I will follow up with pr(s) to migrate component usage. Eventually we can get rid of some wrappers and clean-up it a bit.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/runtime/alloy.go
+++ b/internal/runtime/alloy.go
@@ -48,10 +48,10 @@ package runtime
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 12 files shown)


kgeckhart (approved):
It's so nice to no longer have to do "msg", .............

PR #6035, opened 2026-04-10
docs: Replace examples in reference docs with links to task docs

ptodev:


docs/sources/monitor/monitor-windows.md
```diff
@@ -228,7 +228,7 @@ loki.process "endpoint" {
       }
   }
 
-  stage.eventlogmessage {
+  stage.windowsevent {
```
ptodev, line 231: We'd need to update the `alloy-scenarios` repo.

docs/sources/monitor/monitor-windows.md
ptodev, line 1: @clayton-cornell  It'd be nice to show the full config in one place? E.g. both the "metrics" and the "logs" sections can show the config at the start of the section? Then this will look more like a drop-in replacement for the current examples in reference docs.

PR #6036, opened 2026-04-10
fix(loki.process): Eliminate per-stream goroutines in multiline stage

kgeckhart:

Replace per-stream goroutines in the `loki.process` multiline stage with a single goroutine and a shared `time.Timer`, reducing allocations by up to 76% and latency by up to 39% under benchmark.

### Pull Request Details

The multiline stage previously spawned one goroutine per active log stream to handle `max_wait_time` timeouts via `time.After`. In environments with many label combinations this creates O(N) goroutines and allocations that grow with cardinality.

The new implementation uses a single goroutine with a shared timer set to the earliest per-stream deadline. When it fires, all expired streams are flushed and the timer is rescheduled to the next deadline. Streams are removed from the map after a timeout flush, preventing unbounded map growth for idle streams.

All debug log calls are now wrapped in `if Debug { ... }` guards, consistent with the `drop`, `regex`, and `labels` stages, to avoid `fmt.Sprintf` and `buffer.String()` more overhead when debug logging is disabled.

**Benchmarks**

```
                          │   main   │            this PR            │
                          │  sec/op  │   sec/op     vs base          │
MultilineStage/debug=false  2.131µ ±1%   1.310µ ±1%  -38.55% (p=0.000 n=10)
MultilineStage/debug=true   2.131µ ±1%   1.610µ ±1%  -24.45% (p=0.000 n=10)

                          │   main   │            this PR            │
                          │  B/op    │   B/op       vs base          │
MultilineStage/debug=false  2112.0 ±0%    808.0 ±0%  -61.74% (p=0.000 n=10)
MultilineStage/debug=true  2.062Ki ±0%  1.310Ki ±0%  -36.51% (p=0.000 n=10)

                          │   main   │            this PR            │
                          │ allocs/op │  allocs/op  vs base          │
MultilineStage/debug=false  21.000 ±0%    5.000 ±0%  -76.19% (p=0.000 n=10)
MultilineStage/debug=true    21.00 ±0%   12.00 ±0%  -42.86% (p=0.000 n=10)
```

### Notes to the Reviewer

Two intentional behavioural changes:

1. **Stale-block flush on entry receipt**: Previously a timed-out block was only flushed when the `time.After` goroutine fired. The new code also flushes inline in `processEntry` when a new entry arrives for a stream that has been idle past `max_wait_time`. The outcome is the same — the block is flushed before the new entry is processed — but the trigger differs.

2. **Cross-stream output ordering is now deterministic**: The old goroutine-per-stream model had multiple goroutines writing concurrently to `out`, so ordering across streams was non-deterministic. The new model serializes all output through a single goroutine, so entries are emitted in arrival order.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/multiline.go
+++ b/internal/component/loki/process/stages/multiline.go
@@ -7,15 +7,15 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/prometheus/common/model"
 
+	"github.com/grafana/loki/pkg/push"
+
 	"github.com/grafana/alloy/internal/component/common/loki"
 	"github.com/grafana/alloy/internal/runtime/logging/level"
-	"github.com/grafana/loki/pkg/push"
 )
 
 // Configuration errors.
@@ -68,22 +68,18 @@ func validateMultilineConfig(cfg MultilineConfig) (*regexp.Regexp, error) {
 
 // multilineStage matches lines to determine whether the following lines belong to a block and should be collapsed
 type multilineStage struct {
-	logger log.Logger
-	cfg    MultilineConfig
-	regex  *regexp.Regexp
+	logger  log.Logger
+	cfg     MultilineConfig
+	regex   *regexp.Regexp
+	streams map[model.Fingerprint]*multilineState
 }
 
 // multilineState captures the internal state of a running multiline stage.
 type multilineState struct {
-	buffer         *bytes.Buffer // The lines of the current multiline block.
-	startLineEntry Entry         // The entry of the start line of a multiline block.
-	currentLines   uint64        // The number of lines of the current multiline block.
-}
-
-func (s *multilineState) Reset() {
-	// We don't reset startLineEntry here to keep old behaviour.
-	s.buffer.Reset()
-	s.currentLines = 0
+	buffer         *bytes.Buffer
+	startLineEntry Entry  // The entry of the start line of a multiline block.
+	currentLines   uint64 // The number of lines of the current multiline block.
+	lastSeen       time.Time
 }
 
 // newMultilineStage creates a MulitlineStage from config
@@ -94,9 +90,10 @@ func newMultilineStage(logger log.Logger, config MultilineConfig) (Stage, error)
 	}
 
 	return &multilineStage{
-		logger: log.With(logger, "component", "stage", "type", "multiline"),
-		cfg:    config,
-		regex:  regex,
+		logger:  log.With(logger, "component", "stage", "type", "multiline"),
+		cfg:     config,
+		regex:   regex,
+		streams: make(map[model.Fingerprint]*multilineState),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/component/loki/process/stages/multiline.go
```diff
@@ -105,97 +102,171 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+					out <- r
+				}
+				// Arm the timer for any stream that now has the earliest deadline,
+				// including streams where currentLines==0 (just hit max_lines) so
+				// the timer fires to remove them if they subsequently go idle.
+				if key := e.Labels.FastFingerprint(); m.streams[key] != nil {
+					if dl := m.streams[key].lastSeen.Add(m.cfg.MaxWaitTime); nearestDeadline.IsZero() || dl.Before(nearestDeadline) {
+						armTimer(dl)
+					}
+				}
```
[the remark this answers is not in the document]
kgeckhart: I don't think this is worth it to fix

internal/component/loki/process/stages/multiline.go
```diff
@@ -105,97 +102,171 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+			select {
+			case e, ok := <-in:
+				if !ok {
+					// Flush all per-stream buffers when the input closes.
+					for _, state := range m.streams {
+						if state.currentLines > 0 {
+							out <- m.flushState(state)
+						}
+					}
+					m.streams = nil
```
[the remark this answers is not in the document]
kgeckhart: I don't think this is worth it to fix

internal/component/loki/process/stages/multiline.go
```diff
@@ -105,97 +102,177 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+					timer.Stop()
+					return
+				}
+				// Capture the stream key before emitting entries downstream.
+				// A downstream stage goroutine may mutate e.Labels concurrently
+				// once the entry is sent on out, which would race with a
+				// post-emit FastFingerprint() call.
+				key := e.Labels.FastFingerprint()
```
kalleep, line 168: We should pass this hash to process entries, the first thing we do in that function is to compute it again

internal/component/loki/process/stages/multiline.go
```diff
@@ -105,97 +102,177 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+				for _, r := range m.processEntry(e) {
+					out <- r
+				}
+				// Arm the timer for any stream that now has the earliest deadline,
+				// including streams where currentLines==0 (just hit max_lines) so
+				// the timer fires to remove them if they subsequently go idle.
+				if m.streams[key] != nil {
+					if dl := m.streams[key].lastSeen.Add(m.cfg.MaxWaitTime); nearestDeadline.IsZero() || dl.Before(nearestDeadline) {
```
kalleep: We have this exact same check in two places, should probably be a helper function

kalleep (commented):
Nice! Yeah the go routine per stream is not great here. 

The over all changes makes sense but I left two comments

internal/component/loki/process/stages/multiline.go
```diff
@@ -105,97 +102,177 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+				select {
+				case <-timer.C:
+				default:
+				}
+			}
+			if d := time.Until(deadline); d > 0 {
+				timer.Reset(d)
+			} else {
+				timer.Reset(0)
+			}
```
x1unix:
```suggestion
      d := max(0, time.Until(deadline))
      timer.Reset(d)
```

kalleep (approved):

PR #6039, opened 2026-04-13
chore(loki): Migrate loki processing components to slog

kalleep:
### Pull Request Details

Migrate `loki.echo`, `loki.enrich`, `loki.process`, `loki.relabel`, and `loki.secretfilter` to `slog`. Since we can inspect log level with slog I removed `Debug` variable.

### Notes to the Reviewer

All `loki.Process` stages now scope their logger consistently with `stage=<stage>`.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


kalleep:
@codex review

internal/component/registry.go
```diff
@@ -70,8 +71,18 @@ type Options struct {
 
 	// Logger the component may use for logging. Logs emitted with the logger
 	// always include the component ID as a field.
+	//
+	// Logger is kept for compatibility with components that have not yet been
+	// migrated to slog. Prefer SLogger for new code.
```
kalleep, line 76: I did not mark this a deprecated because then we would have to put no lint comments all over the code

internal/runtime/logging/handler.go
```diff
@@ -76,7 +76,7 @@ func (h *handler) buildHandler() slog.Handler {
 	var newHandler slog.Handler
 
 	handlerOpts := slog.HandlerOptions{
-		AddSource: true,
+		AddSource: false,
```
kalleep, line 79: This is adding a `source: <path to line where we log>` on every log emitted through slog. We did not do this before migrating to slog so we can remove it for now

internal/component/loki/process/stages/windowsevent.go
```diff
@@ -73,14 +72,14 @@ func (w *WindowsEventStage) Run(in chan Entry) chan Entry {
 func (w *WindowsEventStage) processEntry(extracted map[string]any, key string) error {
 	value, ok := extracted[key]
 	if !ok {
-		if Debug {
-			level.Debug(w.logger).Log("msg", "source not in the extracted values", "source", key)
+		if debugEnabled(w.logger) {
```
bboreham, line 75: Is it worth doing this?  Slog does pretty much what `debugEnabled` does right at the top of the `log` function.

I could see the point if the arguments were costly to compute.

kalleep: I just kept it in every place we had it before, yeah for this one specifically and many like it I don't think it makes sense. 

I will go through and only keep this one where arguments can be costly, but I will do that in a [follow up](https://github.com/grafana/alloy/issues/4813#issuecomment-4250989418).

internal/component/loki/process/stages/docker.go
```diff
@@ -2,25 +2,24 @@ package stages
...
-	"github.com/grafana/alloy/internal/runtime/logging/level"
 )
 
 type DockerConfig struct{}
 
 // NewDocker creates a predefined pipeline for parsing entries in the Docker json log format.
-func NewDocker(logger log.Logger, _ prometheus.Registerer, _ featuregate.Stability) (Stage, error) {
-	return toStage(&DockerStage{logger}), nil
+func NewDocker(logger *slog.Logger, _ prometheus.Registerer, _ featuregate.Stability) (Stage, error) {
+	return toStage(&DockerStage{logger: logger.With("stage", "docker")}), nil
```
ptodev, line 18: Where was this done before? Or was the logger not wrapped?

internal/component/loki/process/stages/cri.go
```diff
@@ -50,11 +49,11 @@ func (args *CRIConfig) Validate() error {
 	return nil
 }
 
-func NewCRI(logger log.Logger, cfg CRIConfig, registerer prometheus.Registerer, _ featuregate.Stability) (Stage, error) {
+func NewCRI(logger *slog.Logger, cfg CRIConfig, registerer prometheus.Registerer, _ featuregate.Stability) (Stage, error) {
 	partialLinesFlushedMetric := getPartialLinesFlushedMetric(registerer)
 	linesTruncatedMetric := getLinesTruncatedMetric(registerer)
 	return &cri{
-		logger:                    logger,
+		logger:                    logger.With("stage", "cri"),
```
ptodev, line 56: Would it be better if whoever calls NewCRI wraps the logger? That way we can make sure it's done and that it's done in a consistent way for all components. For example, the caller can call a Name() function to get the name of each stage.

internal/component/loki/process/stages/docker.go
kalleep: It was not wrapped. Some stage did wrap it and others not. If a stage wrapped it before it was using `log.With("component", "stage", "type", "multiline")` and logger we pass to components always have `log.With("component_id", "<id>", "component_path", "<path>")`. 

I figured it cleaner to just have `logger.With("stage", "<stagename>")`.

internal/component/loki/process/stages/cri.go
kalleep: Maybe, but can I do that in a followup?

ptodev: Yes, sure

internal/component/loki/process/stages/docker.go
ptodev: Aha ok. I still think it's better if the caller of New does the wrapping, but it's not necessary to do this in this PR.

ptodev (approved):

PR #6049, opened 2026-04-14
chore: Add pipeline tests

kalleep:
Add an initial declarative pipeline test framework under `internal/pipelinetest` for running real Alloy log pipelines from YAML schemas.

### Pull Request Details
This adds a small framework for running pipeline tests against an in-process Alloy runtime and asserting on the resulting Loki output. The schema currently lets a test define the Alloy config under test, Loki inputs, and Loki assertions.

The test automatically injects a `pipelinetest.source` component for configured Loki inputs and a `pipelinetest.sink` component for collecting output. Configs can also use symbolic output references such as `pipelinetest.loki.url` and `pipelinetest.loki.receiver`, these are translated by to the concrete sink exports before the test is run.

The current tests are mostly intended to demonstrate what this framework can express and how it can be used. It's currently limited to loki signals but can be extended to others.

It's pretty trivial to expose this as a external command `alloy test` that lets users run these kind of tests but we should only do that after we have added more support and are happy with the schema. Code removed in https://github.com/grafana/alloy/pull/6049/changes/e222319d1d9850674ace31fddfa36c191e3c5c62 was all it took to expose this.

### Notes to the Reviewer

Currently there are three tests for loki pipeline, two of them are sharing the same processing module:
1. loki-pipeline - use the shared processing module and inject entries and perform strict assertion on entries.
2. loki-pipeline-file - uses the same module as the other test but is ingesting logs from different files and perform `partial` assertion on labels.
3. loki-pipeline-fanout - added to make sure we can fan-out an entry and modify parts of it like labels and structured_metadata.

### PR Checklist
- [ ] Documentation added
- [x] Tests updated


internal/pipelinetest/loki_test.go
```diff
@@ -0,0 +1,74 @@
...
+		config       string
+		entryPoints  []string
+		inputEntries []loki.Entry
+		assertions   []harness.Assertion
+	}
+
+	tests := []testCase{
+		{
+			name: "rename foo label",
+			config: `
```
ptodev, line 26: It'd be good to make it clear what the entry and exit points are in the config itself. Can we include the `source` and `sink` components in the config? This would be especially helpful if we need more than one entry/exit point per pipeline for some reason.

internal/pipelinetest/harness/alloy.go
```diff
@@ -0,0 +1,129 @@
...
+	Config          string
+	LogsEntryPoints []string
+}
+
+func NewAlloy(t *testing.T, opts Options) *Alloy {
+	t.Helper()
+
+	injectedComponents := func(opts Options) string {
+		return `
+			pipelinetest.source "in" {
```
ptodev: Can we use real components instead of dummies? Is the problem that the real components do too much? I wonder if we can make simple real components and use those instead? E.g. in the past we talked about a generic component that can ingest log files that can come from a string provided by local.file or remote.http.

internal/pipelinetest/harness/alloy.go
```diff
@@ -0,0 +1,129 @@
...
+				forward_to {
+					logs = [` + strings.Join(opts.LogsEntryPoints, ", ") + `]
+				}
+			}
+
+			pipelinetest.sink "out" {}
```
ptodev: Can we use live debugging for output instead?

ptodev (commented):
I'm happy to have something like this, but I want to challenge it a little bit to make sure we can't just reuse standard Alloy features :)

internal/pipelinetest/loki_test.go
kalleep: I had it like that first but moved it to the harness itself, it's just easier not having to repeat it for every test. The downside is what you pointed out, that you cannot have multiple inputs / output components I guess.

I don't really have a strong preference here.

internal/pipelinetest/harness/alloy.go
kalleep: We should keep the sink component at least, that will let us use e.g. loki.write and assert on the requests later.

Not sure about source/reciever compoents, on one hand it would be great to include them but we would be limited to components that don't requires external dependencies, we have integration tests for that. It's also way harder to control the shape of the input data exactly.

internal/pipelinetest/harness/alloy.go
kalleep: That would be really annoying and limited to do, then we would be dependent on the format that we emit there. I do not consider this a good option

internal/pipelinetest/harness/alloy.go
kalleep: I did some changes and added an example using pipelinetest.sink and another one using loki.source.file.

internal/pipelinetest/loki_test.go
kalleep: I changed it so that source is optional but we always inject the sink

internal/pipelinetest/schema.go
```diff
@@ -0,0 +1,196 @@
...
+	return out
+}
+
+func toLabelsAdapter(labels map[string]string) push.LabelsAdapter {
+	if len(labels) == 0 {
+		return nil
+	}
+
+	out := make(push.LabelsAdapter, 0, len(labels))
+	for name, value := range labels {
```
thampiotr, line 228: Because we build it from labels map, the order won't be deterministic. This can lead to flaky tests in some cases. We probably want to sort them as they become a slice?

internal/pipelinetest/schema.go
```diff
@@ -0,0 +1,196 @@
...
+		parsed, err := time.Parse(time.RFC3339Nano, match.Timestamp)
+		if err != nil {
+			return nil, fmt.Errorf("parse match timestamp %q: %w", match.Timestamp, err)
+		}
+		matchers = append(matchers, harness.LokiEntryTimestamp(parsed))
+	}
+	if len(match.StructuredMetadata) > 0 {
+		matchers = append(matchers, harness.LokiEntryStructuredMetadata(toLabelsAdapter(match.StructuredMetadata)))
+	}
+
```
thampiotr, line 202: I think we should check that there are some kind of matchers present, otherwise it's easy to produce a matcher that matches everything? AFAIU the intended use is to have len(matchers) > 0

internal/pipelinetest/harness/alloy.go
```diff
@@ -0,0 +1,146 @@
...
+		a.Stop()
+		return nil, err
+	}
+
+	if err := eventually(func() error {
+		if ctrl.LoadComplete() {
+			return nil
+		}
+		return fmt.Errorf("runtime has not finished loading")
+	}, 2*time.Second, 50*time.Millisecond); err != nil {
```
thampiotr: In CI a pause of 2+ seconds can happen. I think we should give it 10s here.

internal/pipelinetest/harness/alloy.go
```diff
@@ -0,0 +1,146 @@
...
+
+		if len(errs) == 0 {
+			return nil
+		}
+
+		return AssertionErrors{
+			Errors:   errs,
+			Snapshot: s,
+		}
+	}, time.Second, 50*time.Millisecond)
```
thampiotr: That's also very tight time frame. A GC pause or any kind of VM freeze in CI can be 5s occasionally. 10s timeout probably is a better idea here. And both should ideally be configured by a constant.

internal/pipelinetest/tests/file-source/test.yaml
```diff
@@ -0,0 +1,47 @@
+config: |
+  loki.source.file "file" {
```
thampiotr, line 2: I think that ergonomics could be a bit better (easier to get identiation right and IDE support) if we made one-test-per-folder approach. 
You would have folder file_source/ and in it:
- config.alloy
- input.yaml
- assert.yaml

or maybe the yaml files combined into single `test.yaml`, that may be simpler. 
But then you can even see us taking the config.alloy and using it in docs examples. 

As we have 50+ tests in the future, it would make them a bit easier to manage too, if there can be more than one file per test, like in the case of file_source test.

internal/pipelinetest/tests/file_source.yaml
```diff
@@ -0,0 +1,47 @@
+config: |
+  loki.source.file "file" {
+    targets = [{__path__ = "./tests/file_source.log"}]
```
thampiotr: I'd expect the test framework to run from inside of `./tests/` directory already. Can we make sure we CD correctly?

internal/pipelinetest/pipeline_test.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		if filepath.Ext(entry.Name()) != ".yaml" {
+			continue
+		}
+
+		name := strings.TrimSuffix(entry.Name(), filepath.Ext(entry.Name()))
+		path := filepath.Join("tests", entry.Name())
+
+		t.Run(name, func(t *testing.T) {
+			bb, err := os.ReadFile(path)
```
thampiotr, line 26: I think if we were to scale this up to 50+ tests, we will want to run some of them in parallel. 
The issue is with shared resources like ports (might need to not run those in parallel in the end) or disk paths (would need to run them in temporary test folders). So I think that tests should be able to opt in whether they can be run in parallel or not. What do you think?

internal/pipelinetest/tests/file-source/test.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - type: has_entry
+      match:
+        labels:
+          test: test
+        line: line 1
+    - type: has_entry
+      match:
+        labels:
+          test: test
+        line: line 2
```
thampiotr, line 47: I see the list of these likely grow large and become a bit harder to discover / correctly write these assertions. If they are in Go code, it's slightly easier to get autocomplete, type checking and compilation errors. Yet something is quite nice about just adding a text file though I cannot verbalise what that exactly is. 

Thoughts on how to mitigate the risk of this becoming its own language that nobody really knows how to use? Especially over time as we add more complex assertions.

thampiotr (commented):
Overall I like this approach and it's quite simple and should be manageable to add new tests. Left some comments on specific things we probably want to address.

internal/pipelinetest/harness/alloy.go
kalleep: Ah yes for sure! I even think this should be configurable at least if we expose this as a command.

internal/pipelinetest/tests/file-source/test.yaml
kalleep: Yes I agree that having one folder per test is better. I am thinking a bit how this would look like if we exposed it as a test command to users in the future, we should probably not force any conventions.

What do you say about supporting both inline config or specifying the path to the config?

internal/pipelinetest/tests/file-source/test.yaml
kalleep: >If they are in Go code, it's slightly easier to get autocomplete, type checking and compilation errors. Yet something is quite nice about just adding a text file though I cannot verbalise what that exactly is.

Yes I started with this and [using table tests](https://github.com/grafana/alloy/blob/667702189b23d136330b476ea37fca6a621b8e82/internal/pipelinetest/loki_test.go). Then I started to experiment how to use this as an external command and that how the currect "schema" was born. So I figured if we want to expose this in the future  we should also write our tests using the schema so we can discover what is good / bad about it.


>Thoughts on how to mitigate the risk of this becoming its own language that nobody really knows how to use?

Yes this is something I worry about and it's also a big part why I would not feel comfortable exposing this to users yes. I really don't want to invent our own dsl but we need some way to configure different kind of assertions without creating to much friction. I am open to ideas!

internal/pipelinetest/schema.go
kalleep: Yes that is true! Either we accept a slice of name and values for the assertion input or we sort and don't perform strict comparison in the assertion

internal/pipelinetest/pipeline_test.go
kalleep: I changed it so that our internal test have directories for every test instead and will look for test.yaml in all sub dirs instead.

kalleep: > disk paths (would need to run them in temporary test folders)

This is not a problem any more since I pass t.TempDir as data-path for each test.

For other shared resources, yes they are problematic with t.Parallel, will think a bit about this case

internal/pipelinetest/tests/file_source.yaml
kalleep: So I don't think chaining directory will work correctly, at least not with t.Parallel. Will think a bit how this could be resolved.

And I would like to only solve it for internal tests, if we expose this as a test command it should work like `alloy run` i.e. requires absolute path or relative from where it was called from.

internal/pipelinetest/tests/file-source/test.yaml
kalleep: Btw I refactored the loki assertions. 

I changed the types to `contains` and `count`

1. Count supports matches like contains does but asserts that we get exactly the number of entires specified, without matchers every entry will be counted.
1. Contains make sure we have at least one entry using matchers.

labels and structured metadata also have a mode, partial or exact, to make it more flexible

internal/pipelinetest/prelude.go
```diff
@@ -0,0 +1,96 @@
...
+// withPrelude builds the final Alloy source for a pipeline test by prepending
+// the test sink, injecting any generated input sources, and rewriting
+// pipelinetest symbolic output references.
+func withPrelude(schema TestSchema) string {
+	sink := buildSink()
+	sources := buildSources(schema.Inputs)
+
+	return sink + "\n\n" + sources + "\n\n" + rewritePipelineTestRefs(string(schema.Config))
+}
+
```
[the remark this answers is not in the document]
kalleep: Skipping this one for now

internal/pipelinetest/prelude.go
```diff
@@ -0,0 +1,96 @@
...
+func rewritePipelineTestRefs(source string) string {
+	replacer := strings.NewReplacer(
+		"pipelinetest.loki.url", "pipelinetest.sink.out.loki_push_url",
+		"pipelinetest.loki.receiver", "pipelinetest.sink.out.lokireceiver",
+	)
+	return replacer.Replace(source)
+}
```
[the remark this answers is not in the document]
kalleep: Yes this is intentional because the implementation is much simpler than rewriting the AST. If this ever becomes a problem we can change that

thampiotr (approved):

PR #6056, opened 2026-04-15
ci: Start additional containers in parallel

ptodev:
It reduces the runtime of integration tests if some containers take a long time to start.

This doesn't actually save any build time at the moment, because most containers are quick to start. The only slow one is the single Oracle DB one. Concurrent starts will be important later since I intend to start more than one Oracle DB.

### DIFF
--- a/integration-tests/docker/additional_containers.go
+++ b/integration-tests/docker/additional_containers.go
@@ -74,22 +74,50 @@ func startAdditionalContainers(ctx context.Context, absTestDir, networkName stri
 		requests = append(requests, req)
 	}
 
-	containers := make([]testcontainers.Container, 0, len(cfg.AdditionalContainers))
-	for _, r := range requests {
-		container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
-			ContainerRequest: r,
-			Started:          true,
-			Logger:           log.Default(),
+	// Start all additional containers in parallel.
+	// That way containers which are slow to start won't stop others from starting up.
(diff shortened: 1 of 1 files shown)


integration-tests/docker/additional_containers.go
```diff
@@ -74,22 +74,50 @@ func startAdditionalContainers(ctx context.Context, absTestDir, networkName stri
...
+	// Start all additional containers in parallel.
+	// That way containers which are slow to start won't stop others from starting up.
+	containers := make([]testcontainers.Container, len(requests))
+	gctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+	var startErr error
+	for i := range requests {
+		i := i
```
kalleep, line 86: No need to copy i here, LLMs have not caught up to changes in go yet

integration-tests/docker/additional_containers.go
```diff
@@ -74,22 +74,50 @@ func startAdditionalContainers(ctx context.Context, absTestDir, networkName stri
...
 		}
-
-		containers = append(containers, container)
+		_ = terminateAdditionalContainers(ctx, partial)
+		return nil, startErr
 	}
 
-	return containers, nil
+	out := make([]testcontainers.Container, 0, len(containers))
+	out = append(out, containers...)
```
kalleep, line 119: Why do you need this clone?

ptodev: I don't :) Thanks, I'll delete it

kalleep (approved):

PR #5919, opened 2026-03-28
docs: Add gRPC block documentation for loki.source.api

boinger:
### Brief description of Pull Request

The `loki.source.api` component starts both HTTP and gRPC servers via the
shared `ServerConfig`, but only the HTTP block was documented. This adds
the missing `grpc` block documentation by including the existing shared
`loki-server-grpc.md` reference, matching the pattern used by
`loki.source.heroku` and `loki.source.gcplog`.

Also fixes a copy-paste error in the shared gRPC docs where `conn_limit`
was described as limiting "HTTP connections" instead of "gRPC connections".

### Pull Request Details

The undocumented gRPC listener has caused confusion for users whose
security teams flagged unexpected network bindings (see issue comments).

Changes:
- Updated intro to mention gRPC alongside HTTP
- Added `grpc` and `grpc > tls` to the Blocks table
- Added `### grpc` section using the existing shared include
- Updated `### tls` description to cover both HTTP and gRPC
- Fixed `conn_limit` description in `loki-server-grpc.md` (HTTP → gRPC)

### Issue(s) fixed by this Pull Request

Fixes #2889

### Notes to the Reviewer

The shared gRPC docs file (`loki-server-grpc.md`) already existed and is
used by `loki.source.heroku` and `loki.source.gcplog`. This PR wires it
into `loki.source.api` using the same pattern.

### PR Checklist

- [x] Documentation added

PR #6057, opened 2026-04-15
docs: Add gRPC block documentation for loki.source.awsfirehose

### LINKED ISSUE #2889: Docs feedback: loki.source.api missing documentation of grpc block
### URL

https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.api/

### Feedback

The loki.source.api configuration uses the shared internal config block for defining a server, which includes an HTTP and/or GRPC server. The code supports the GRPC pathways, however it's missing from the documentation.

boinger:

The `loki.source.awsfirehose` component uses the shared `ServerConfig` via `fnet.NewTargetServer`, which always binds both HTTP and gRPC listeners regardless of whether the blocks are configured. The docs only referenced the `http` block, leaving the gRPC port undocumented.

This is the same class of issue addressed for `loki.source.api` in #5919 (closed #2889), replicated for a sibling component. The `tls` block description already acknowledged "HTTP and gRPC servers", so the gap was plainly visible in the docs themselves.

### Pull Request Details

Changes:
- Added `grpc` and `grpc > tls` rows to the Blocks table
- Added `[grpc]: #grpc` reference
- Added `### grpc` section using the existing shared include (`loki-server-grpc.md`), matching the pattern in `loki.source.api`, `loki.source.heroku`, and `loki.source.gcplog`

The component intro is intentionally left unchanged. AWS Data Firehose only ever POSTs over HTTP, so claiming the component "receives over HTTP and gRPC" would be misleading. The gRPC listener is a side effect of the shared server config, not a functional ingestion path.

Related to #2889 (same root cause, sibling component).

### Notes to the Reviewer

Follow-up to #5919. Two other components (`pyroscope.receive_http` and `prometheus.receive_http`) have the same gap and will be addressed in separate PRs for focused review.

### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.awsfirehose.md
+++ b/docs/sources/reference/components/loki/loki.source.awsfirehose.md
@@ -107,13 +107,20 @@ You can use the following blocks with `loki.source.awsfirehose`:
 
 | Name                  | Description                                        | Required |
 | --------------------- | -------------------------------------------------- | -------- |
+| [`grpc`][grpc]        | Configures the gRPC server that receives requests. | no       |
+| `grpc` > [`tls`][tls] | Configures TLS for the gRPC server.                | no       |
 | [`http`][http]        | Configures the HTTP server that receives requests. | no       |
 | `http` > [`tls`][tls] | Configures TLS for the HTTP server.                | no       |
 
 [http]: #http
+[grpc]: #grpc
 [tls]: #tls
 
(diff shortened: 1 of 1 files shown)


boinger:
Superseded by #6071 — prefer the architectural fix (stop starting the phantom gRPC listener) over documenting a surface that shouldn't exist. Closing in favor of that issue.

PR #6058, opened 2026-04-15
docs: Add gRPC block documentation for pyroscope.receive_http

boinger:
### Brief description of Pull Request

The `pyroscope.receive_http` component uses the shared `ServerConfig` via `fnet.NewTargetServer`, which always binds both HTTP and gRPC listeners regardless of whether the blocks are configured. The docs only referenced the `http` block, leaving the gRPC port undocumented.

The component name implies HTTP-only, which makes the surprise gRPC port especially confusing for operators whose security teams flag unexpected network bindings. This is the same class of issue addressed for `loki.source.api` in #5919 (closed #2889).

Note: the Connect handlers in this component (`pushv1connect.PusherService`, `debuginfov1alpha1connect.DebuginfoService`) are mounted on the HTTP router, not the native gRPC server. The gRPC listener opened by the shared server config therefore has no application-level handlers bound to it — it is purely a side effect of the shared server configuration.

### Pull Request Details

Changes:
- Added `grpc` and `grpc > tls` rows to the Blocks table
- Added `[grpc]: #grpc` reference
- Added `### grpc` section using the existing shared include (`loki-server-grpc.md`)
- Updated `### tls` description from "HTTP server" to "HTTP and gRPC servers"

Pattern matches the merged `loki.source.api` (#5919), as well as pre-existing `loki.source.heroku` and `loki.source.gcplog` docs.

### Issue(s) fixed by this Pull Request

Related to #2889 (same root cause, sibling component).

### Notes to the Reviewer

Follow-up to #5919 and a companion PR for `loki.source.awsfirehose` (#6057). A third PR for `prometheus.receive_http` will follow.

### PR Checklist

- [x] Documentation added

PR #6059, opened 2026-04-15
docs: Add gRPC block documentation for prometheus.receive_http

### LINKED ISSUE #2889: Docs feedback: loki.source.api missing documentation of grpc block

boinger:

The `prometheus.receive_http` component uses the shared `ServerConfig` via `fnet.NewTargetServer`, which always binds both HTTP and gRPC listeners regardless of whether the blocks are configured. The docs only referenced the `http` block, leaving the gRPC port undocumented.

The component name implies HTTP-only, which makes the surprise gRPC port especially confusing for operators whose security teams flag unexpected network bindings. This is the same class of issue addressed for `loki.source.api` in #5919 (closed #2889).

Note: `prometheus.receive_http` only serves the Prometheus remote write protocol over HTTP — the gRPC listener opened by the shared server config has no application-level handlers bound to it. It is purely a side effect of the shared server configuration.

### Pull Request Details

Changes:
- Added `grpc` and `grpc > tls` rows to the Blocks table
- Added `[grpc]: #grpc` reference
- Added `### grpc` section using the existing shared include (`loki-server-grpc.md`)
- Updated `### tls` description from "HTTP server" to "HTTP and gRPC servers"

Pattern matches the merged `loki.source.api` (#5919) and pre-existing `loki.source.heroku` / `loki.source.gcplog` docs.

Related to #2889 (same root cause, sibling component).

### Notes to the Reviewer

Follow-up to #5919 and companion PRs for `loki.source.awsfirehose` (#6057) and `pyroscope.receive_http` (#6058). This completes the set of components that inherit the always-on gRPC listener from the shared `ServerConfig` but don't yet document it.

### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.receive_http.md
+++ b/docs/sources/reference/components/prometheus/prometheus.receive_http.md
@@ -69,13 +69,20 @@ You can use the following blocks with `prometheus.receive_http`:
 
 | Name                  | Description                                        | Required |
 | --------------------- | -------------------------------------------------- | -------- |
+| [`grpc`][grpc]        | Configures the gRPC server that receives requests. | no       |
+| `grpc` > [`tls`][tls] | Configures TLS for the gRPC server.                | no       |
 | [`http`][http]        | Configures the HTTP server that receives requests. | no       |
 | `http` > [`tls`][tls] | Configures TLS for the HTTP server.                | no       |
 
+[grpc]: #grpc
 [http]: #http
 [tls]: #tls
 
 {{< /docs/alloy-config >}}
+
+### `grpc`
+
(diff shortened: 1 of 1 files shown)


kalleep:
Hey thanks for the pr(s). Instead of updating docs I created https://github.com/grafana/alloy/issues/6071. We should fix this issue instead. It does not make sense for us to even start this http over grpc bridge for all of these components.

boinger:
Superseded by #6071 — prefer the architectural fix (stop starting the phantom gRPC listener) over documenting a surface that shouldn't exist. Closing in favor of that issue.

PR #6062, opened 2026-04-15
ci: Let require.Eventually increase its timeout via the --test-timeout cmd arg

ptodev:
Currently, there is a `TEST_TIMEOUT` env var which governs how long `require.Eventually` should wait for. There's also a `--test-timeout` cmd arg which sets `TEST_TIMEOUT`, but only if the integration test framework is ran for a specific test. 

Neither `--test-timeout` nor `TEST_TIMEOUT` increase the timeout of the child `go test` sub process. So if you set `--test-timeout` or `TEST_TIMEOUT` to a high value, then the test will just panic in the end due to a timeout. This is not helpful because you won't see the output from `require.Eventually` - it won't say why it failed. This can happen, if for example `TEST_TIMEOUT` is 15 min. Then Go by default will run the test for a maximum of 10 min, and the test will time out and panic in 10 min before reaching the full 15 min.

This is all confusing and inconsistent. IMO it'd be more intuitive if the command line argument sets the global timeout limit for the `go test` process, and then `require.Eventually`'s timeout expands and contracts according to that.

### DIFF
--- a/integration-tests/docker/common/common.go
+++ b/integration-tests/docker/common/common.go
@@ -18,12 +18,12 @@ type Unmarshaler interface {
 const (
 	DefaultRetryInterval = 100 * time.Millisecond
 
-	// Timeout to use if no -timeout is provided to go test.
-	defaultTimeout = 10 * time.Minute
+	// Default timeout for polling/assertions if no -timeout is provided to go test.
+	defaultTimeout = 120 * time.Second
 
 	// Timeout padding to subtract from the time left until the go test -timeout deadline.
 	// This makes sure that polling/assertions fail before the process watchdog kills the test.
-	timeoutPadding = 30 * time.Second
+	timeoutPadding = 15 * time.Second
 )
 
 func FetchDataFromURL(url string, target Unmarshaler) (string, error) {
@@ -131,6 +131,10 @@ func isStatefulFromEnv() (bool, error) {
 	}
 
 	return isStateful, nil
+}
+
+func DefaultProcessTimeout() time.Duration {
+	return defaultTimeout + timeoutPadding
 }
 
 // TestTimeout returns how long polling helpers should keep retrying. When the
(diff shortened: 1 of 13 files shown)


integration-tests/docker/common/common.go
```diff
@@ -17,7 +17,13 @@ type Unmarshaler interface {
 
 const (
 	DefaultRetryInterval = 100 * time.Millisecond
-	DefaultTimeout       = 90 * time.Second
+
+	// Timeout to use if no -timeout is provided to go test.
+	defaultTimeout = 10 * time.Minute
```
kgeckhart: Do we really want such a high test timeout like this? A test that can take up to 10 minutes to run is not a test I really want to have to run ever.

integration-tests/docker/main.go
```diff
@@ -40,7 +38,7 @@ func main() {
 		"This is useful for a fast iteration loop locally but should not be used in CI." +
 		"You must run 'docker compose down' manually if you want to switch from stateful to stateless mode."
 	rootCmd.PersistentFlags().BoolVar(&stateful, "stateful", false, statefulUsageString)
-	rootCmd.PersistentFlags().DurationVar(&testTimeout, "test-timeout", common.DefaultTimeout, "Timeout for each individual test")
+	rootCmd.PersistentFlags().DurationVar(&testTimeout, "test-timeout", 10*time.Minute, "Timeout for each individual test")
```
kgeckhart: Why move this away from the common default?

integration-tests/docker/common/common.go
ptodev: I agree that we can lower it. I was using high timeouts in recent testing due to waiting several minutes for databases to initialise. But that's rare and I agree that the default can be lower.

integration-tests/docker/main.go
ptodev: I didn't want to export the constant timeout setting :) I refactored it a bit, now this uses the common default and I think it's more clear.

integration-tests/docker/common/common.go
ptodev: I lowered it to a similar value as the previous default

Issue #6032, opened 2026-04-10
Alloy - AWS SDK v1 still present via transitive dependencies after #2936 closure


lpospiech-lumen:
### Add a description

**Overview**

Issue #2936 was closed on March 30, 2026, tracking the migration from AWS SDK for Go v1 to v2. However, checking the current dependency graph shows that v1 is still present through transitive dependencies from upstream packages.

Since AWS SDK for Go v1 reached end-of-life on **July 31, 2025**, these remaining dependencies mean Alloy is still exposed to unsupported code that will no longer receive security patches.

AWS's official announcement: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

The v1 SDK has known vulnerabilities that will remain unpatched:

- **CVE-2020-8911**: Padding oracle vulnerability in AWS S3 Crypto SDK - allows attackers with S3 write access to exploit AES-CBC encryption when MAC validation is absent
- **CVE-2020-8912**: In-band key negotiation vulnerability in S3 Crypto SDK - enables encryption algorithm downgrade attacks (AES-GCM → AES-CTR) for attackers with bucket write access

**Changes description**

While Alloy itself may have completed its direct migration to v2, several upstream dependencies still pull in v1. Here's the current state:

```
$ go mod graph | grep github.com/aws/aws-sdk-go | grep -v github.com/aws/aws-sdk-go-v2
github.com/go-kit/kit@v0.13.0 github.com/aws/aws-sdk-go@v1.40.45
github.com/grafana/loki/v3@v3.6.5 github.com/aws/aws-sdk-go@v1.55.7
github.com/heroku/x@v0.5.2 github.com/aws/aws-sdk-go@v1.34.0
github.com/mongodb/mongo-tools@v0.0.0-20250715155915-887641268977 github.com/aws/aws-sdk-go@v1.53.11
```

These are all indirect/transitive dependencies from:
- **go-kit/kit** (v0.13.0)
- **grafana/loki/v3** (v3.6.5) - has its own tracking issue at https://github.com/grafana/loki/issues/16309
- **heroku/x** (v0.5.2)
- **mongodb/mongo-tools**

**Summary**

This issue is meant to track the remaining v1 dependencies until upstream packages complete their migrations. Options include:

1. Monitor upstream migration efforts (especially Loki #16309 which is a Grafana project)
[...]

PR #6064, opened 2026-04-15
feat(loki.source.heroku): Drop github.com/heroku/x package

x1unix:

Copy `github.com/heroku/x/logplex/encoding` package and drop `github.com/heroku/x` dependency.

### Pull Request Details

The `loki.source.heroku` uses `github.com/heroku/x/logplex/encoding` package to parse Heroku Syslog-based log messages.

The heroku/x Go module brings aws-sdk-go as an indirect dependency. The SDK is EOL and having 2 known security issues.

Although the component pulls only a single `encoding` package - this still triggers security scanners (see: https://github.com/grafana/alloy/issues/6032).

Considering that the [upstream PR](https://github.com/heroku/x/pull/256) is open for quite a long time and Alloy uses only one small package, to address the problem I've copied it and removed unused functionality.

**Note:** although the module is removed from `go.mod`, it *still present* in dependency tree as it is used by Loki v3.6.5.

Related to https://github.com/grafana/alloy/issues/6032

### Notes to the Reviewer

I've considered using `github.com/leodido/go-syslog/v4` package instead, but go-syslog's RFC 5424 parser is fundamentally incompatible with the Heroku drain format.

Best-effort mode doesn't help - it loses the message body, which is the most important field.


jharvey10 (commented):
> Copy github.com/heroku/x/logplex/encoding package

It looks like this means copy as in literal copy/paste. If so, I see their repo is governed by BSD 3 clause, which means we need to carry forward their copyright notice either in each of the files or in a file alongside the source.

Since this is more of an external dependency, isolating it from our internal code might be a good idea. We don't appear to have a great directory for that sort of stuff right now, but there seems to be some conventions online indicating to use something like `internal/third_party/...`.

So in this case, perhaps: `internal/third_party/heroku/x/logplex/encoding`, etc.

x1unix:
@jharvey10 thanks for taking a look.

> It looks like this means copy as in literal copy/paste. 

Almost, I've removed parts that we don't use to reduce maintenance surface area.

> I see their repo is governed by BSD 3 clause, which means we need to carry forward their copyright notice either in each of the files or in a file alongside the source.

As the upstream don't have copyright header in their source files, I can put the copyright in the readme or pull the license file.

> Since this is more of an external dependency, isolating it from our internal code might be a good idea.

Regarding a potential package path - this package is also used in integration tests (`loki_heroku_test.go`) so I can't put that into `/internal`.

Some repos put common reused packages into `/pkg`. What do you think about `pkg/logplex-encoding` to avoid having depth nesting?

jharvey10:
> Regarding a potential package path - this package is also used in integration tests (loki_heroku_test.go) so I can't put that into `/internal`.

Is this a technical limitation? Or do you feel weird about putting it in that folder because of its usage elsewhere?

x1unix:
> Is this a technical limitation? Or do you feel weird about putting it in that folder because of its usage elsewhere?

@jharvey10 this is a technical limitation, `internal` packages can't be imported from outside:

```
loki_heroku_test.go:14:2: use of internal package github.com/grafana/alloy/internal/component/loki/source/heroku/internal/encoding not allowed
```

I had to move out `encoding` package from `internal` to make tests pass, see logs for a failed pipeline:

https://github.com/grafana/alloy/actions/runs/24519700698/job/71673695646#step:4:3911

jharvey10:
It looks like the CI failure you linked was perhaps because of the _nested_ `internal` that was there previously?

I'm just suggesting to use:
```
herokuencoding "github.com/grafana/alloy/internal/third_party/heroku/x/logplex/encoding"
```
instead of
```
herokuencoding "github.com/grafana/alloy/internal/component/loki/source/heroku/encoding"
```

A small amount of nesting, but it would help to mark this heroku module as an internal so outside users aren't tempted to use it.

I did test it locally and it seemed to work, but want to make sure that pathing works for you as well.

x1unix:
@jharvey10 I've moved the package to proposed location and added a license.

jharvey10 (approved):
🎉  LGTM!

PR #6068, opened 2026-04-15
refactor(pyroscope): Remove reflection workaround for NewOAuth2RoundTripper

korniltsev-grafanista-yolo-vibecoder239:
## Summary

- Remove the `reflect`+`unsafe` workaround (`reflect_oauth2.go`) that was needed to call `commonconfig.NewOAuth2RoundTripper` with its previously unexported `*httpClientOptions` parameter
- Delete the associated layout-verification tests (`reflect_oauth2_test.go`) that ensured struct compatibility
- Call `commonconfig.NewOAuth2RoundTripper` directly using its new variadic `...HTTPClientOption` signature
- Bump `prometheus/common` to include [prometheus/common#898](https://github.com/prometheus/common/pull/898)

## Details

The upstream `NewOAuth2RoundTripper` accepted an unexported `*httpClientOptions` as its fourth parameter, making it uncallable from external packages without reflection. The `promhttp2` package maintained a local mirror of the struct and used `reflect.NewAt` + `unsafe.Pointer` to pass it through — with extensive tests to assert struct layout, field order, and byte-for-byte default matching against the upstream type.

prometheus/common#898 changed the signature to accept variadic `...HTTPClientOption` (an exported interface), matching the pattern already used by `NewRoundTripperFromConfigWithContext`. This makes the workaround unnecessary.

A new `toUpstreamOpts()` method on the local `httpClientOptions` converts fields to upstream option values so that any options passed by callers are correctly forwarded to the OAuth2 round tripper.

**Net change:** 194 lines deleted, 36 added.

## Test plan

- [x] `go test ./internal/component/pyroscope/write/...` passes
- [x] All OAuth2-related tests (`TestOAuth2`, `TestOAuth2WithFile`, `TestOAuth2WithJWTAuth`, etc.) pass


korniltsev-grafanista:
Superseded by #6099 — recreated from a maintainer account with dependency-change permissions.

PR #6078, opened 2026-04-16
fix(loki.process): Return errors from metric registration instead of panicking

boinger:

Two functions in the `loki.process` pipeline stages package (`getDropCountMetric` and `registerCounterVec`) called `panic(err)` when Prometheus metric registration failed for non-duplicate reasons. This could crash the entire Alloy process on an otherwise recoverable error.

A TODO at `match.go:107` (added in #5399) already acknowledged this:

> This code should neither panic nor use AlreadyRegisteredError. Register it without these, and return error if it fails.

This PR implements that fix.

### Pull Request Details

Changes:
- Both helpers now return `error` instead of panicking; six callers and the stage factory propagate it
- `newSamplingStage` and `newPackStage` signatures changed from `Stage` to `(Stage, error)` for consistency with the other stage constructors

The `AlreadyRegisteredError` handling is preserved (returning the existing collector), as this is expected behavior when multiple stages share a metric. Only unexpected registration failures are now surfaced as errors instead of panics.

None. Resolves TODO at match.go:107 (#5399).

### Notes to the Reviewer

`pack_test.go` has the explicit `require.NoError` on the new return; other callers route through existing `NewPipeline` tests in `pipeline_test.go`, `drop_test.go`, etc.

Verified with `go test -race -count=1 ./internal/component/loki/process/stages/...` and `make lint`.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/drop.go
+++ b/internal/component/loki/process/stages/drop.go
@@ -84,11 +84,16 @@ func newDropStage(logger *slog.Logger, config DropConfig, registerer prometheus.
 		return nil, err
 	}
 
+	dropCount, err := getDropCountMetric(registerer)
+	if err != nil {
+		return nil, err
+	}
+
 	return &dropStage{
 		logger:    logger.With("stage", "drop"),
 		cfg:       &config,
 		regex:     regex,
-		dropCount: getDropCountMetric(registerer),
+		dropCount: dropCount,
 	}, nil
 }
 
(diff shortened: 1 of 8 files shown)


dehaansa:
This needs a rebase from main to fix CI.

dehaansa:
Now that 1.18 has released, another rebase from main should resolve CI and we can get this merged! Sorry for the back and forth.

Edit: nevermind, I was able to re-run it. 👍

dehaansa (approved):

PR #6081, opened 2026-04-16
feat(database_observability): Always extract traceparent in MySQL query_samples collector

fridgepoet:

The MySQL query_samples collector is changed to extract and log the traceparent from SQL_TEXT.

### Pull Request Details

In an instrumented app, middleware for a DB driver/ORM like SQLCommenter can add embed trace information to each SQL statement in a comment. 

If the trace information is available in the SQL statement (SQL_TEXT), this PR makes a change so that the traceparent is extracted. It is then logged under the `traceparent` label. 

Example log:
Screenshot 2026-04-16 at 16 00 03

Contributes to https://github.com/grafana/grafana-dbo11y-app/issues/2659

### Notes to the Reviewer

Original work: https://github.com/grafana/alloy/pull/4196

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -472,3 +462,34 @@ func (c *QuerySamples) determineTimerClauseAndLimit(uptime float64) (string, flo
...
+		return ""
+	}
+
+	idx := strings.LastIndex(strings.ToLower(sqlText), "traceparent=")
+	if idx < 0 {
+		return ""
+	}
+	tp := sqlText[idx+len("traceparent="):]
+
+	quote := tp[0]
```
matthewnolf, line 484: We should check this slice isn't empty first

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -74,8 +73,9 @@ LEFT JOIN
 	ON statements.THREAD_ID = threads.THREAD_ID
 WHERE
 	statements.DIGEST IS NOT NULL
+    AND statements.DIGEST_TEXT IS NOT NULL
```
cristiangreco: `digest_text` is not used anymore, this check could be dropped.

cristiangreco: We're also removing the clause `AND statements.SQL_TEXT IS NOT NULL`... not a big deal, but might be worth considering keeping it?

internal/component/database_observability/mysql/collector/query_samples.go
cristiangreco, line 484: If the text is truncated exactly after `traceparent=`, I think this will panic.

cristiangreco (changes requested):
Left some comments for discussion

internal/component/database_observability/mysql/collector/query_samples.go
fridgepoet: Thanks, OK, let's include `SQL_TEXT IS NOT NULL` as you mention since we are trying to extract traceparent from the SQL_TEXT.

So how about we go with `AND statements.SQL_TEXT IS NOT NULL` and drop the `AND statements.DIGEST_TEXT IS NOT NULL`?

internal/component/database_observability/mysql/collector/query_samples.go
fridgepoet: Thanks, added a guard and test cases.

fridgepoet: Thanks, added a guard and test cases.

internal/component/database_observability/mysql/collector/query_samples.go
matthewnolf: Yes, let's drop `digest_text` and it's `not null` condition

matthewnolf: Will adding `SQL_TEXT IS NOT NULL` remove some queries from being collected though? Is it possible sql_text isn't enabled on some DBs?

fridgepoet: Looked at the docs and `SQL_TEXT` should be present as long as it's a SQL statement and `performance_schema_max_sql_text_length` is not 0.

This is as opposed to `DIGEST_TEXT` where the `statements_digest` consumer also needs to be set to `yes`.

matthewnolf (approved):

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -74,8 +73,9 @@ LEFT JOIN
 	ON statements.THREAD_ID = threads.THREAD_ID
 WHERE
 	statements.DIGEST IS NOT NULL
+    AND statements.SQL_TEXT IS NOT NULL
```
cristiangreco:
```suggestion
	AND statements.SQL_TEXT IS NOT NULL
```

cristiangreco: nit: keep using tabs instead of spaces

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -472,3 +462,38 @@ func (c *QuerySamples) determineTimerClauseAndLimit(uptime float64) (string, flo
...
+	if strings.HasSuffix(sqlText, "...") {
+		return ""
+	}
+
+	idx := strings.LastIndex(strings.ToLower(sqlText), "traceparent=")
+	if idx < 0 {
+		return ""
+	}
+
+	tp := sqlText[idx+len("traceparent="):]
```
cristiangreco: I think we'd better go with a key-val based parsing as described in the [specs](https://google.github.io/sqlcommenter/spec/) ("Parsing" section).

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -374,9 +361,10 @@ func (c *QuerySamples) fetchQuerySamples(ctx context.Context) error {
 		row.TimestampMilliseconds = calculateWallTime(serverStartTime, row.TimerEndPicoseconds.Float64, uptime)
 		cpuTime := picosecondsToMilliseconds(row.CPUTime)
 		elapsedTime := picosecondsToMilliseconds(row.ElapsedTimePicoseconds.Float64)
+		traceParent := tryExtractTraceParent(row.SQLText.String)
 
 		logMessage := fmt.Sprintf(
-			`schema="%s" user="%s" client_host="%s" thread_id="%s" event_id="%s" end_event_id="%s" digest="%s" rows_examined="%d" rows_sent="%d" rows_affected="%d" errors="%d" max_controlled_memory="%db" max_total_memory="%db" cpu_time="%fms" elapsed_time="%fms" elapsed_time_ms="%fms"`,
+			`schema="%s" user="%s" client_host="%s" thread_id="%s" event_id="%s" end_event_id="%s" digest="%s" rows_examined="%d" rows_sent="%d" rows_affected="%d" errors="%d" max_controlled_memory="%db" max_total_memory="%db" cpu_time="%fms" elapsed_time="%fms" elapsed_time_ms="%fms" traceparent="%s"`,
```
cristiangreco: Do we still want to log the traceparent key even if it's empty?

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -409,7 +398,7 @@ func (c *QuerySamples) fetchQuerySamples(ctx context.Context) error {
 		if row.WaitEventID.Valid && row.WaitTime.Valid {
 			waitTime := picosecondsToMilliseconds(row.WaitTime.Float64)
 			waitLogMessage := fmt.Sprintf(
-				`schema="%s" user="%s" client_host="%s" thread_id="%s" digest="%s" event_id="%s" wait_event_id="%s" wait_end_event_id="%s" wait_event_name="%s" wait_object_name="%s" wait_object_type="%s" wait_time="%fms"`,
+				`schema="%s" user="%s" client_host="%s" thread_id="%s" digest="%s" event_id="%s" wait_event_id="%s" wait_end_event_id="%s" wait_event_name="%s" wait_object_name="%s" wait_object_type="%s" wait_time="%fms" traceparent="%s"`,
```
cristiangreco: Traceparent might not be needed here for `op=wait_event`.

internal/component/database_observability/mysql/collector/query_samples.go
fridgepoet: I'll go ahead and remove it when it's empty. That kind of matches the conditional presence of the sql_text key.

cristiangreco (approved):

PR #5035, opened 2025-12-09
Update Prometheus and Loki

thampiotr:
#### PR Description

This finally gets rid of our Prometheus fork. 
Not all new experimental features in Prometheus are exposed. Instead there are some TODOs left in the code to GH issues which should address this. This is to keep the PR smaller and prioritise the work accordingly. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Config converters updated

PR #6087, opened 2026-04-17
fix(prometheus.scrape): Update scrape_native_histograms to be updated at runtime

kalleep:
### Pull Request Details

Allow scrape_native_histograms updates to take effect in prometheus.scrape and keep the runtime guard only for the settings that still live on the scrape manager. The original runtime restriction was introduced in #4044 when native histogram ingestion was configured through scrape.Options, but it stopped being true after #5035 when this option moved into scrape config.

I also added gaurds / logging for the things that are not currently updatable.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/scrape/scrape.go
+++ b/internal/component/prometheus/scrape/scrape.go
@@ -478,9 +478,14 @@ func (c *Component) Update(args component.Arguments) error {
 	if !c.firstUpdateDone {
 		c.firstUpdateDone = true
 	} else {
-		if c.args.ScrapeNativeHistograms != newArgs.ScrapeNativeHistograms {
-			level.Warn(c.opts.Logger).Log("msg", "scrape_native_histograms cannot be changed at runtime; the component will continue using the original setting until Alloy is restarted", "current", c.args.ScrapeNativeHistograms, "requested", newArgs.ScrapeNativeHistograms)
-			newArgs.ScrapeNativeHistograms = c.args.ScrapeNativeHistograms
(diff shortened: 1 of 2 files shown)


internal/component/prometheus/scrape/scrape.go
```diff
@@ -478,9 +478,14 @@ func (c *Component) Update(args component.Arguments) error {
 	if !c.firstUpdateDone {
 		c.firstUpdateDone = true
 	} else {
-		if c.args.ScrapeNativeHistograms != newArgs.ScrapeNativeHistograms {
-			level.Warn(c.opts.Logger).Log("msg", "scrape_native_histograms cannot be changed at runtime; the component will continue using the original setting until Alloy is restarted", "current", c.args.ScrapeNativeHistograms, "requested", newArgs.ScrapeNativeHistograms)
-			newArgs.ScrapeNativeHistograms = c.args.ScrapeNativeHistograms
+		// TODO: When these change we could stop and re-create scrape manager.
```
kgeckhart, line 481: I wonder if this was behavior carried from upstream which has the ability to do live config updates but not recycle the whole manager?

It would certainly require some thought about what could go wrong.

kgeckhart (approved):

internal/component/prometheus/scrape/scrape.go
kalleep: Yes that was what I was thinking too and did not want to dig into it here.

PR #6088, opened 2026-04-17
fix(documentation): Fix documentation for loki.rules.kubernetes

timonegk:

Fix some broken links and case errors in the `loki.rules.kubernetes` component documentation.

### Notes to the Reviewer

I find the difference between blocks and arguments in the documentation a bit confusing, especially regarding the `rule_selector`. `match_labels` is an argument, `match_expression` is a block. So the two options are explained in completely different places, but tightly belong together. When I read the documentation, I thought that maybe one or the other was not supported because sometimes only one, sometimes only the other was listed. After I understood that one is an argument and one is a block, it was clear, but initially it was quite confusing.
That's why I added the sentence to explain that `match_expression` can also be used.

### PR Checklist

- none applicable


clayton-cornell (approved):
Thanks for catching and fixing the links. They sometimes get missed/broken when other refactoring happens. I try to catch it.. but sometimes they slip through.

The clarification sentence is a good idea.

timonegk:
Of course @clayton-cornell, it's only normal that something slips through from time to time.
By the way, thank you for your thorough reviews in general and your keen eye on detail and consistency. I really appreciate it!

PR #6089, opened 2026-04-17
feat: Integration tests for aws firehose and cloudflare logpull

x1unix:

* Add integration tests for Loki components:
  * `loki.source.cloudflare`
  * `loki.source.awsfirehose`
* Add CloudFlare LogPull mock API server for integration tests.
* Add option to override Cloudflare API URL via env var - this is necessary for tests.

### DIFF
--- a/integration-tests/docker/tests/loki-cloudflare/logpullmock/main.go
+++ b/integration-tests/docker/tests/loki-cloudflare/logpullmock/main.go
@@ -1,0 +1,67 @@
+// logpullmock is a minimal mock of the Cloudflare Logpull API for integration testing.
+// It serves NDJSON log entries on the /client/v4/zones/{zoneID}/logs/received endpoint.
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+	"time"
+)
+
+func main() {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/health", handleHealth)
+	mux.HandleFunc("/client/v4/zones/", handleLogpull)
+
+	log.Println("logpullmock listening on :8080")
+	if err := http.ListenAndServe(":8080", mux); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func handleHealth(w http.ResponseWriter, _ *http.Request) {
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintln(w, "ok")
+}
+
+func handleLogpull(w http.ResponseWriter, r *http.Request) {
+	// Expected path: /client/v4/zones/{zoneID}/logs/received
+	if !strings.HasSuffix(r.URL.Path, "/logs/received") {
+		http.NotFound(w, r)
+		return
+	}
+
+	log.Printf("logpull request: %s %s", r.Method, r.URL.String())
+
+	now := time.Now()
+	entries := []struct {
+		ClientIP            string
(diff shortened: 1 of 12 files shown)


integration-tests/docker/tests/loki-cloudflare/logpullmock/main.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func handleLogpull(w http.ResponseWriter, r *http.Request) {
+	// Expected path: /client/v4/zones/{zoneID}/logs/received
+	if !strings.HasSuffix(r.URL.Path, "/logs/received") {
+		http.NotFound(w, r)
+		return
+	}
+
+	log.Printf("logpull request: %s %s", r.Method, r.URL.String())
```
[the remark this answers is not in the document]
x1unix: False positive alert on a CF logpull mocking tool used for integration tests.

integration-tests/docker/tests/loki-firehose/loki_firehose_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type firehoseRecord struct {
+	Data string `json:"data"`
+}
+
+// pushFirehoseRecords sends Direct PUT Firehose requests to the running Alloy
+// component. It sends `batches` requests per source ARN, each containing
+// `recordsPerBatch` base64-encoded plain text records.
+func pushFirehoseRecords(batches, recordsPerBatch int, sourceARNs ...string) error {
```
kalleep: So IIRC we can get two kinds of payload, this is doing what we call `direct-put` right?

Could we also send in the other format we accept?

internal/component/loki/source/cloudflare/client.go
```diff
@@ -30,7 +31,11 @@ func (w *wrappedClient) LogpullReceived(ctx context.Context, start, end time.Tim
 }
 
 var getClient = func(apiKey, zoneID string, fields []string) (Client, error) {
-	c, err := cloudflare.NewWithAPIToken(apiKey)
+	var opts []cloudflare.Option
+	if apiURL := os.Getenv("ALLOY_CLOUDFLARE_API_URL"); apiURL != "" {
```
kalleep, line 37: Ah yeah this is a bit annoying. I am not too big of a fan using env here to override.

I guess the other option would be to add another argument but I don't _think_ that is useful for anything other than testing. But if we keep env lets add a comment why we do this

x1unix: > I guess the other option would be to add another argument but I don't think that is useful for anything other than testing.

@kalleep this was the exact reason why I decided to pick an environment variable. Adding an extra argument would also increase PR scope as that would affect docs and component unit tests.

Addressed in https://github.com/grafana/alloy/pull/6089/changes/339feffa63a89907460dcd7b5c767f8838d6555e

integration-tests/docker/tests/loki-firehose/loki_firehose_test.go
x1unix: @kalleep added logs payload type in addition to `direct-put`: https://github.com/grafana/alloy/pull/6089/changes/8bbd703f755adc91a2ad318da91dd730f556e819

kalleep (approved):
LGTM thanks!

PR #6096, opened 2026-04-20
feat(prometheus.exporter.azure): Add Azure Monitor Batch API support

robert-c-ewing:

Add support for the Azure Monitor Batch API (`metrics:getBatch`) to `prometheus.exporter.azure`, enabling up to 50 resources per API call using the Azure Monitor data plane. This data plane has separate rate limits from the ARM management plane, dramatically reducing 429 throttling for organizations monitoring many resources across multiple subscriptions.

### Pull Request Details

**Problem:** The existing per-resource collection path makes one ARM API call per discovered resource. At scale (e.g., 40 subscriptions × dozens of resource types), this quickly exhausts the [ARM rate limit](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling) (250-token bucket, 25 tokens/second refill per subscription), producing 429 errors and gaps in metrics.

**Solution:** A new `use_batch_api = true` config option routes metric collection through the [Azure Monitor Batch API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch) (`POST https://{region}.metrics.monitor.azure.com/subscriptions/{sub}/metrics:getBatch`). This:

- Batches up to **50 resources** per request (vs. 1)
- Uses the Azure Monitor **data plane**, which has separate rate limits from ARM
- Auto-discovers resources via Resource Graph and groups them by (subscription, region)
- Chunks metric names in groups of 20 (Azure API limit per request)
- Produces the **same Prometheus labels** as the existing per-resource path
- Automatically widens the query time window when `interval > timespan`
- Supports all Azure cloud environments (public, China, US Government)

**What this does NOT change:** The existing per-resource and subscription-scope (`regions`) paths are completely untouched. `use_batch_api = false` (the default) preserves all existing behavior.

### Notes to the Reviewer

**Architecture:** The batch API implementation lives entirely in a single new file (`batch.go`), with minimal wiring into the existing code:
- `config.go`: 1 new field + 1 validation rule
- `azure_exporter.go`: 1 new branch in `MetricsHandler()`, before the existing `regions` and per-resource branches
- `azure.go` (component layer): 1 new field in `Arguments` + `Convert()`

**Auth:** The batch API uses a data-plane OAuth2 scope (`https://metrics.monitor.azure.com/.default`) rather than the ARM management scope. The token is acquired per-request using the same credential the ARM client uses. In most environments this just works; restrictive environments may need to grant the identity access to the data plane.

**Time grains:** Some Azure resource types (e.g., PostgreSQL Flexible Servers) only support coarse time grains (≥ PT30M). This is an Azure-side constraint, not specific to the batch API. Users must set `interval` and `timespan` appropriately. The docs now include a note about this.

**`computeTimeWindow` auto-widening:** If the configured `interval` (time grain) is larger than the `timespan` (query window), the batch API would return no data. `computeTimeWindow()` automatically widens the window to at least match the interval. This is a safety net — `Config.Validate()` already rejects `timespan < interval`, but the batch API computes absolute start/end times and this guards against edge cases.

**`parseISO8601Duration`:** We implement a minimal parser rather than using the `iso8601duration` library already in the dependency tree. This is intentional — the library's `ToDuration()` doesn't handle all cases cleanly, and we only need D/H/M/S for Azure Monitor's supported intervals. Invalid inputs panic, since `Config.Validate()` runs the full library parser first.

**Query-param exclusion:** `UseBatchAPI` is intentionally excluded from `MergeConfigWithQueryParams` since enabling batch mode changes the data plane routing and auth scope.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
@@ -29,8 +29,8 @@ The exporter offers the following three options for gathering metrics.
    1. This approach doesn't work with all resource types, and Azure doesn't document which resource types do or don't work.
    1. A resource type that's not supported produces errors that look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`.
    1. If you encounter one of these errors you must use the default Azure Resource Graph based option to gather metrics.
-1. Use the [Azure Monitor Batch API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch) by setting `use_batch_api` to `true`.
-   1. This batches up to 50 resources per API call using the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`).
+1. Set use_batch_api` to `true` to use the [Azure Monitor Batch API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch).
+   1. This uses the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`) to batch up to 50 resources per API call.
    1. The data plane has **separate rate limits** from the ARM management plane, which significantly reduces 429 (Too Many Requests) throttling when monitoring many resources.
    1. Resources are discovered using Azure Resource Graph (like option 1), then grouped by subscription and region for batch retrieval.
    1. Cannot be used together with `regions`. Use `resource_graph_query_filter` to target specific regions instead.
@@ -45,7 +45,7 @@ The account used by {{< param "PRODUCT_NAME" >}} needs:
 * When using an Azure Resource Graph query, [read access to the resources that will be queried by Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#permissions-in-azure-resource-graph).
 <!-- vale Grafana.GoogleSpacing = NO -->
 * Permissions to call the [Microsoft.Insights Metrics API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics/list) which should be the `Microsoft.Insights/Metrics/Read` permission.
-* When using `use_batch_api`, the identity also needs access to the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`). This uses a different OAuth2 scope than the ARM management plane. In most environments this works with the same identity, but restrictive configurations may require additional permissions.
+* When you use `use_batch_api`, the identity also needs access to the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`). This uses a different OAuth2 scope than the ARM management plane. In most environments this works with the same identity, but restrictive configurations may require additional permissions.
 <!-- vale Grafana.GoogleSpacing = YES -->
 
 ## Usage
@@ -201,8 +201,8 @@ Replace the following:
 
 ### Batch API example
 
-The following example uses the Batch API to collect metrics from Azure Key Vault resources across multiple subscriptions
-with reduced API call overhead. This is recommended when monitoring many resources to avoid ARM rate limiting.
+The following example uses the Batch API to collect metrics from Azure Key Vault resources across multiple subscriptions with reduced API call overhead.
+This is recommended when monitoring many resources to avoid ARM rate limiting.
 
 ```alloy
 prometheus.exporter.azure "keyvault" {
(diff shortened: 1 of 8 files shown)


docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
```diff
@@ -29,6 +29,11 @@ The exporter offers the following two options for gathering metrics.
    1. This approach doesn't work with all resource types, and Azure doesn't document which resource types do or don't work.
    1. A resource type that's not supported produces errors that look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`.
    1. If you encounter one of these errors you must use the default Azure Resource Graph based option to gather metrics.
+1. Use the [Azure Monitor Batch API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch) by setting `use_batch_api` to `true`.
```
clayton-cornell:
```suggestion
1. Set use_batch_api` to `true` to use the [Azure Monitor Batch API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics-batch/batch).
```

docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
```diff
@@ -29,6 +29,11 @@ The exporter offers the following two options for gathering metrics.
...
+   1. This batches up to 50 resources per API call using the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`).
```
clayton-cornell:
```suggestion
   1. This uses the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`) to batch up to 50 resources per API call.
```

docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
```diff
@@ -40,6 +45,7 @@ The account used by {{< param "PRODUCT_NAME" >}} needs:
 * When using an Azure Resource Graph query, [read access to the resources that will be queried by Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#permissions-in-azure-resource-graph).
 <!-- vale Grafana.GoogleSpacing = NO -->
 * Permissions to call the [Microsoft.Insights Metrics API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics/list) which should be the `Microsoft.Insights/Metrics/Read` permission.
+* When using `use_batch_api`, the identity also needs access to the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`). This uses a different OAuth2 scope than the ARM management plane. In most environments this works with the same identity, but restrictive configurations may require additional permissions.
```
clayton-cornell:
```suggestion
* When you use `use_batch_api`, the identity also needs access to the Azure Monitor data plane (`https://<region>.metrics.monitor.azure.com`). This uses a different OAuth2 scope than the ARM management plane. In most environments this works with the same identity, but restrictive configurations may require additional permissions.
```

docs/sources/reference/components/prometheus/prometheus.exporter.azure.md
```diff
@@ -192,6 +199,43 @@ Replace the following:
 * _`<USERNAME>`_: The username to use for authentication to the `remote_write` API.
 * _`<PASSWORD>`_: The password to use for authentication to the `remote_write` API.
 
+### Batch API example
+
+The following example uses the Batch API to collect metrics from Azure Key Vault resources across multiple subscriptions
+with reduced API call overhead. This is recommended when monitoring many resources to avoid ARM rate limiting.
```
clayton-cornell:
```suggestion
The following example uses the Batch API to collect metrics from Azure Key Vault resources across multiple subscriptions with reduced API call overhead.
This is recommended when monitoring many resources to avoid ARM rate limiting.
```

robert-c-ewing:
Got some time to perform some stress testing against our cloud environment. So far so good; no regressions I've spotted in CPU or memory usage that jump out, and the series count matches for legacy and batch API scrapers.

**Stress test configuration**
- Subscriptions:     40
- Resource types:  4 (KeyVault, PostgreSQL, ContainerApps, ContainerAppEnvs)
- Scrapers:             8 (4 batch + 4 legacy)
- PostgreSQL:        31 metrics (2 batch chunks of 20+11)

**Results (30m stress):**
- 7,128 total active series in Prometheus
- 3,336 series from Azure exporters across 8 scrape targets (4 batch + 4 legacy)

I'll update if the soak testing reveals anything major.

kgeckhart:
Hi @robert-c-ewing, thanks for the PR and testing it out. I am not sure that this is work that we can accept though. 

Today the component is a wrapper around an upstream exporter and this PR adds what is effectively a new exporter implementation sitting next to it. As soon as we merge this PR we become response for the maintenance burden of it. I would encourage you to redirect this PR to the upstream exporter which has a maintainer that is more familiar with the Azure domain.

PR #6103, opened 2026-04-21
fix(faro.receiver) Use Background context for log sending

### LINKED ISSUE #6102: Faro receiver logs multiple context canceled after upgradring from 1.13.0 to 1.15.1
### Component(s)

faro.receiver

### What's wrong?

After upgrading  Alloy from 1.13.0 to 1.15.1 quite a lot of context canceled errors are logged.

We are receiving about 1k Faro requests / second and about once every five minutes this errors is logged. We have limited control of the clients.

### Steps to reproduce

Set up Alloy to receive Faro with quite a bit of requests (1000/s) from global end users.

### System information

AKS 1.34.4

### Software version

Grafana Alloy 1.15.1

### Configuration

```text
declare "input_faro" {
	argument "logs_targets" {
		optional = false
	}

	argument "traces_targets" {
		optional = false
	}

	faro.receiver "faro" {
		server {
			listen_address = "0.0.0.0"
			listen_port    = 8081

			rate_limiting {
				rate       = 2000
				burst_size = 4000
			}
		}

		output {
			logs   = [loki.process.set_faro_labels.receiver]
			traces = argument.traces_targets.value
		}
		extra_log_labels = {
			source_type = "faro",
		}

		sourcemaps {
			download              = true
			download_from_origins = [
                                <redacted>
			]
			download_timeout = "10s"

			cache {
				ttl = "10m"
			}
		}
	}

	loki.process "set_faro_labels" {
		forward_to = argument.logs_targets.value

		stage.logfmt {
			mapping = {
				"app_name"  = "",
				"kind"      = "",
				"level"     = "",
				"view_name" = "",
				"page_url"  = "",
				"timestamp" = "",
			}
		}

		stage.timestamp {
			source           = "timestamp"
			format           = "RFC3339"
			fallback_formats = [
				"2006-01-02 15:04:05.999 -0700 UTC",
			]
		}

		stage.regex {
			source     = "page_url"
			expression = "^(?:https?:\\/\\/)?(?P<host>[^\\/\n?]+)"
		}

		stage.labels {
			values = {
				app          = "app_name",
				kind         = "",
				level        = "",
				view_name    = "",
				host         = "",
				service_name = "app_name",
				gp_category  = "kind",
				gp_view_name = "view_name",
				host_name    = "host",
			}
		}

		stage.structured_metadata {
			values = {
				severity_text = "level",
			}
		}
	}
}

// Receive traces and logs from Grafana Faro (Client)
components.input_faro "input_faro" {
	logs_targets   = [loki.write.loki3.receiver]
	traces_targets = [otelcol.processor.batch.batch_traces.input]
}

loki.write "loki3" {
	endpoint {
		url = "http://<redacted>{{ .Release.Namespace }}.svc.cluster.local./enrich/loki/api/v1/push"
	}
}
```

### Logs

```text
{"ts":"2026-04-21T07:54:58.325261114Z","level":"error","msg":"exporter failed with error","component_path":"/components.input_faro.input_faro","component_id":"faro.receiver.faro","subcomponent":"handler","exporter":"logs exporter","err":"context canceled\ncontext canceled\ncontext canceled\ncontext canceled\ncontext canceled\ncontext canceled\ncontext canceled\ncontext canceled\ncontext canceled"}
```

### Tip

JohanLindvall:
Change context to Background for log sending to avoid cancellation from incoming requests.

Use Background context for log sending

### Pull Request Details

Use Background context for log sending

Fixes #6102

### DIFF
--- a/internal/component/faro/receiver/exporters.go
+++ b/internal/component/faro/receiver/exporters.go
@@ -167,7 +167,8 @@ func (exp *logsExporter) sendKeyValsToLogsPipeline(ctx context.Context, kv *payl
 		return err
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, 2*time.Second) // TODO(rfratto): potentially make this configurable
+	// Use Background so that incoming request cancellation doesn't cancel sending already received logs.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second) // TODO(rfratto): potentially make this configurable
 	defer cancel()
 	return exp.fanout.Send(ctx, loki.NewEntry(exp.labelSet(kv), push.Entry{
 		Timestamp: time.Now(),


kalleep:
Hey I don't really see the issue here. The context that is propagated is from the request. If that is canceled it means that client have canceled the request. We cloud ignore context.Canceled from error logs but usually when the request context is canceled it means that the client sending the request timed out and we should abort.

JohanLindvall:
> Hey I don't really see the issue here. The context that is propagated is from the request. If that is canceled it means that client have canceled the request. We cloud ignore context.Canceled from error logs but usually when the request context is canceled it means that the client sending the request timed out and we should abort.

As far as I understand things, the request has been completely received and the client may for some reason chose to end the request. This should not cancel sending the already received Faro data to the backend(s).

Additional context: we are receiving ~1k Faro requests per second and this error is logged maybe once every five minutes. We have limited control of the clients.

kalleep:
>As far as I understand things, the request has been completely received and the client may for some reason chose to end the request. This should not cancel sending the already received Faro data to the backend(s).

Yeah we have received the request but alloy has not finished processing it yet. Consider a client that is making a request with a timeout of e.g. 10s. After this 10s have passed the client will consider the request failed and abort, retry or do what ever it needs to do. 

If alloy is not aborting at this point we risk e.g. ingesting duplicated request (in case of retries). We also just keep piling up requests even though client has already aborted and is maybe retrying making the back pressure worse

JohanLindvall:
> > As far as I understand things, the request has been completely received and the client may for some reason chose to end the request. This should not cancel sending the already received Faro data to the backend(s).
> 
> Yeah we have received the request but alloy has not finished processing it yet. Consider a client that is making a request with a timeout of e.g. 10s. After this 10s have passed the client will consider the request failed and abort, retry or do what ever it needs to do.
> 
> If alloy is not aborting at this point we risk e.g. ingesting duplicated request (in case of retries). We also just keep piling up requests even though client has already aborted and is maybe retrying making the back pressure worse

Understood - I am not too knowledgeable about the client and what it does. It is just a shame that we receive a complete client request but fail in the middle of sending the parsed data to Loki.

A "proper" fix would perhaps be to separate the read and write paths so that partial writes due to client cancellation simply isn't possible?

kalleep:
>A "proper" fix would perhaps be to separate the read and write paths so that partial writes due to client cancellation simply isn't possible?

That would require buffering request and also not a good fix IMHO, same issue with back presure. But there is still some fixes that should be done here. After every call to `sendKeyValsToLogsPipeline` we should check for context cancellation and abort on first one.

A fix for handling partial requests is way bigger and is already being worked as a bigger effort to loki pipelines in general

PR #6106, opened 2026-04-22
feat(database_observability.mysql): Add database_observability_wait_event_seconds_total counter

gaantunes:
## Summary

Adds a `database_observability_wait_event_seconds_total` Prometheus counter to the MySQL component that pre-aggregates wait time per query digest and database schema.

This counter provides an alternative backend for wide-window TotalWaitEventsTime queries, bypassing the Loki 500-series fallback loop entirely. Benchmarks show −52 to −65% p50 at 1h30+ unfiltered windows where Loki fallback adds 2–7s of retry overhead.

## Details

**Metric:** `database_observability_wait_event_seconds_total` (counter)
**Labels:** `server_id` (curried at component level), `digest`, `schema`
**Emitted:** only when `enable_pre_classified_wait_events` is true (i.e. when the collector emits `wait_event_v2` log lines). When the flag is off, the counter is not registered and does not appear in `/metrics`.
**Exposed:** at the component's `/metrics` endpoint via the existing registry

`server_id` is curried at the component level using `CounterVec.CurryWith`, so the collector only needs `digest` and `schema` label values.

## Changes

- `waitEventCounter *prometheus.CounterVec` field on `Component` (lifecycle mirrors `exporterCollector`)
- Counter creation, registration, and `CurryWith` in `startCollectors`, gated on `EnablePreClassifiedWaitEvents`; nil is passed to the collector when the flag is off
- `WaitEventCounter *prometheus.CounterVec` passed to `QuerySamples` collector (nil-safe)
- Counter increment after each wait event emit
- `TestQuerySamples_WaitEventCounter` unit test

## Test plan

- [x] `TestQuerySamples_WaitEventCounter` passes
- [ ] Counter visible at `/metrics` on a running instance with `enable_pre_classified_wait_events = true`
- [ ] Counter absent from `/metrics` when `enable_pre_classified_wait_events = false`
- [x] Labels `server_id`, `digest`, `schema` present on each series


internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -427,6 +431,11 @@ func (c *QuerySamples) fetchQuerySamples(ctx context.Context) error {
 					waitV2LogMessage,
 					int64(millisecondsToNanoseconds(row.TimestampMilliseconds)),
 				)
+
+				if c.waitEventCounter != nil {
+					waitTimeSeconds := waitTime / 1000.0 // waitTime is in ms
```
cristiangreco: There's should be an utility conversion function in `time.go`.

internal/component/database_observability/mysql/component.go
```diff
@@ -615,6 +616,26 @@ func (c *Component) startCollectors(serverID string, engineVersion string, parse
...
+		var curriedCounter *prometheus.CounterVec
+		if c.args.QuerySamplesArguments.EnablePreClassifiedWaitEvents {
+			c.waitEventCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+				Name: "database_observability_wait_event_seconds_total",
+				Help: "Total wait time in seconds per query, aggregated by server, query digest, and schema.",
+			}, []string{"server_id", "digest", "schema"})
+			if err := c.registry.Register(c.waitEventCounter); err != nil {
+				level.Warn(c.opts.Logger).Log("msg", "failed to register wait event counter", "err", err)
+			}
+			cc, err := c.waitEventCounter.CurryWith(prometheus.Labels{"server_id": serverID})
```
cristiangreco: I think this should live in the collector, i.e. we have other examples of collectors that use the registry to add their own metrics.

cristiangreco: Currying shouldn't be needed. No need to add the `server_id` label at all because it's added in `relabeling.go`.

internal/component/database_observability/mysql/collector/query_samples.go
gaantunes: Done in e3c6d7913 — added `millisecondsToSeconds` to `collector/time.go` and the emit site now calls it.

internal/component/database_observability/mysql/component.go
gaantunes: Done in e3c6d7913 — counter creation, `MustRegister`, and `Unregister` now live in `NewQuerySamples` / `Stop`, mirroring `SetupConsumers` and `ConnectionInfo`. Dropped the `server_id` label and the `CurryWith` call; relabeling adds `server_id` at scrape time.

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -139,6 +143,15 @@ func NewQuerySamples(args QuerySamplesArguments) (*QuerySamples, error) {
 		running:                       &atomic.Bool{},
 	}
 
+	if args.EnablePreClassifiedWaitEvents && args.Registry != nil {
+		c.registry = args.Registry
+		c.waitEventCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "database_observability_wait_event_seconds_total",
+			Help: "Total wait time in seconds per query, aggregated by query digest and schema.",
```
cristiangreco: very nit: I'd keep this generic to something like `Total duration of wait events in secods` or similar.

Also, it could be defined as:

```go
		Namespace: "database_observability",
		Name:      "wait_event_seconds_total",
		Help:      "....",
```

cristiangreco (approved):

internal/component/database_observability/mysql/collector/query_samples.go
gaantunes: Done in 99c1ec778 — split into `Namespace: "database_observability"` / `Name: "wait_event_seconds_total"` and trimmed the help to "Total duration of wait events in seconds." Exposed metric name is unchanged.

PR #6110, opened 2026-04-23
chore: Release alloy helm chart v1.8.0

jharvey10:


blewis12 (approved):

kgeckhart (approved):

dcelasun:
It seems like the chart for v1.8.0 didn't get generated correctly and the tgz archive is missing: https://github.com/grafana/helm-charts/issues/4159

jharvey10:
Should be good now! We had two workflow runs collide. I started a discussion internally to see if we need to tweak the release process at all to prevent this from happening again.

PR #6112, opened 2026-04-23
fix(logging): Fix startup deadlock when components log before logging config is evaluated

kgeckhart:

Fix a deadlock that causes Alloy to hang at startup with no log output and the HTTP server never binding.

### Pull Request Details

When components start logging during graph evaluation before the `logging` config block is processed, a lock-order cycle between `Logger.Update` and concurrent `Handle` calls on child deferred handlers permanently deadlocks the process. The bug has been present since #707.

### Notes to the Reviewer

Proof-of-deadlock test hangs for 5s on the unfixed code.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/runtime/logging/logger.go
+++ b/internal/runtime/logging/logger.go
@@ -116,12 +116,12 @@ func (nopSlogHandler) WithGroup(string) slog.Handler { return nopSlogHandler{} }
 // Update re-configures the options used for the logger.
 func (l *Logger) Update(o Options) error {
 	l.bufferMut.Lock()
-	defer l.bufferMut.Unlock()
 
 	switch o.Format {
 	case FormatLogfmt, FormatJSON:
 		l.hasLogFormat = true
 	default:
+		l.bufferMut.Unlock()
 		return fmt.Errorf("unrecognized log format %q", o.Format)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/runtime/logging/logger.go
```diff
@@ -133,25 +133,35 @@ func (l *Logger) Update(o Options) error {
...
 		l.deferredSlog.buildHandlers(nil)
 	}
-	// Print out the buffered logs since we determined the log format already
-	for _, bufferedLogChunk := range l.buffer {
+
+	// Re-acquire to drain and clear the buffer now that handlers are built.
+	l.bufferMut.Lock()
+	buffer := l.buffer
+	l.buffer = nil
+	l.bufferMut.Unlock()
```
jharvey10: Noticed that `Update` now has three separate lock/unlock sites for `bufferMut` which makes the flow a bit hard to follow at a glance. The function seems to naturally break into three phases:

- validate & apply options (locked)
- build handlers (unlocked)
- drain buffer (locked)

What are your thoughts on extracting the two locked phases into private helpers to help clean up the flow?

Each helper could use `defer` normally, and `Update` itself becomes a clear three-ish-line orchestrator. The early-return unlock at line 124 (the invalid-format case) would also disappear as a side effect.

kgeckhart: Before the refactors from copilots review this made sense. I think it's a lot more concise now where splitting it in to helper functions might make it harder to follow.

kalleep (approved):

internal/runtime/logging/logger.go
jharvey10: 👍  Agree. It reads pretty well now!

PR #6113, opened 2026-04-24
feat: Integration tests for loki.source.azure_event_hubs

x1unix:

This PR adds integration tests for `loki.source.azure_event_hubs` component.

The test exercises the full chain: SASL_PLAIN-over-TLS Kafka connect -> Azure Resource Log parse -> relabel -> Loki write.

### Pull Request Details

Under `integration-tests/docker/tests/loki-azure-event-hubs/`:

* `docker-compose.yaml` — three services: 
    - `cert-init`: Generates SSL certificate
    - `kafka-eh`: Kafka 4.2.0 KRaft w/ SASL_SSL + PEM keystore
    - `alloy`
* `config.alloy` — `loki.source.azure_event_hubs` with `connection_string` auth and a `__azure_event_hubs_category → service_name` relabel rule on the source.
* `loki_azure_event_hubs_test.go` — Sarama producer over SASL_SSL, sends Azure Resource Log JSON envelopes (`records[].category` = topic), asserts via `common.AssertLogsPresent`.
* `gen-certs.sh`, `kafka_server_jaas.conf`, `keystore.creds` — static infrastructure copied into the generated certs dir at boot.
* `certs/.gitignore` + `.gitkeep` — keep the bind-mount directory tracked while ignoring runtime PEMs.

Related to https://github.com/grafana/alloy/issues/4953

### Notes to the Reviewer

- The `loki.source.azure_event_hubs` hard-codes `UseTLS: true` for both auth mechanisms with no `tls_config` escape hatch, so the shared plaintext Kafka broker in the root compose is unusable. Hence the dedicated SASL_SSL broker.
- `apache/kafka:latest` unconditionally exports `KAFKA_SSL_KEYSTORE_PASSWORD` (even when empty) whenever SSL is enabled, which Kafka rejects for PEM keystores. Compose overrides the entrypoint to `unset KAFKA_SSL_KEYSTORE_PASSWORD KAFKA_SSL_KEY_PASSWORD` between `configure` and `launch`.
- Single self-signed cert covers `DNS:kafka-eh, DNS:localhost, IP:127.0.0.1`, so it verifies cleanly both from Alloy inside compose and from the host-run Go test — no `InsecureSkipVerify` anywhere.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/integration-tests/docker/tests/loki-azure-event-hubs/docker-compose.yaml
+++ b/integration-tests/docker/tests/loki-azure-event-hubs/docker-compose.yaml
@@ -1,10 +1,17 @@
 services:
   cert-init:
-    image: alpine/openssl:latest
+    # Reuse the kafka-eh image — it already bundles openssl, so we avoid pulling
+    # a second base image. Runs as root so it can write to a freshly-bind-mounted
+    # /certs dir; the script chowns outputs back to uid 1000 (kafka appuser,
+    # typical host user) before exiting.
+    image: apache/kafka:latest@sha256:9516fb7634bad307d17c33b589fde9023003b0cb761374f500002b980a3149b9
+    user: "0:0"
     restart: "no"
     entrypoint: ["sh", "/gen-certs.sh"]
     volumes:
       - ./gen-certs.sh:/gen-certs.sh:ro
+      - ./kafka_server_jaas.conf:/input/kafka_server_jaas.conf:ro
+      - ./keystore.creds:/input/keystore.creds:ro
       - ./certs:/certs
 
   kafka-eh:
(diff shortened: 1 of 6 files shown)


x1unix:
CC @kalleep

integration-tests/docker/tests/loki-azure-event-hubs/docker-compose.yaml
```diff
@@ -0,0 +1,116 @@
+services:
```
kalleep, line 1: How is this docker-compose started? Do we have support for this in our integration tests, if so I had no idea about that

x1unix: @kalleep yes, apparently we are. The `beyla-java` and `otlp-metrics-*` use that.

image

kalleep: Hm interesting, we also have have additional_containers we can configure in test.yaml file..

kalleep (approved):

PR #6125, opened 2026-04-24
test(integration-tests): Refactor k8s integration tests framework

thampiotr:
## Summary

These are integration tests that run Alloy against real workloads on a local
Kubernetes cluster. We use [kind](https://kind.sigs.k8s.io/) for the cluster
and Helm chart to deploy Alloy.

CI parallelises these tests by sharding test **packages** (one package == one heavy
test that brings up its own dependencies, e.g. Mimir, Loki, etc.). Each CI job runs with
`--shard i/n`, and a package runs only when it hashes to the current shard. This
way every package runs on exactly one shard, and as we add more tests we
just bump `n` to keep wall-clock time flat — the test suite shouldn't
become a bottleneck.

The other priority is local developer experience. `make
integration-test-k8s-local-dev` opens an interactive menu for the common
options (reuse cluster, skip image build, pick a single package, pick a
shard) so you can iterate fast without remembering flags. The kubeconfig
is dropped at a known path (see below) so `kubectl` and `k9s` work
straight away.

Adding a new test should stay easy as the suite grows. Each test is a
plain Go test that calls `harness.Setup` with a list of dependencies
(`deps.Namespace`, `deps.Mimir`, `deps.Alloy`, ...). The harness installs
them in order and tears them down in reverse. To add a test: drop a new directory
under `tests/`, and use the existing tests as an inspiration.

Currently the cluster itself is reused across tests in a single run, but every
dependency is installed and torn down per-test. That keeps tests isolated
and easy to reason about, at the cost of some setup time (~20s per test). If specific
heavy dependencies (e.g. Mimir, the prometheus-operator CRDs) become a
real bottleneck we can promote them to cluster-scoped install-once
fixtures — the harness is structured to allow adding this later if needed.

### One-shot run all tests

```sh
make integration-test-k8s
```

### Local dev interactive menu (recommended)

```sh
make integration-test-k8s-local-dev
```

Opens a small TUI to pick the common run options before tests start.

- Reusing kind cluster or skipping Alloy image build to speed up local development.
- Filtering tests by shard or by package.

This allows to save some time when iterating locally on config:
image

This allows to select what tests to run:
image

Neat shard validation. Shows how we can hopefully keep developer experience smooth.
image

### Inspecting the running cluster

The runner writes its kubeconfig to `integration-tests/k8s/.kube/kubeconfig`
(gitignored). Pass it explicitly to your tools to connect to the cluster.

```sh
kubectl --kubeconfig integration-tests/k8s/.kube/kubeconfig get pods -A
k9s    --kubeconfig integration-tests/k8s/.kube/kubeconfig
```

### Run specific tests from CLI (if you don't want to use the interactive menu)

```sh
# Split test packages across 2 shards and run shard index 0. Use this to reproduce a CI run.
make integration-test-k8s RUN_ARGS='--shard 0/2'

# Run a single test package.
make integration-test-k8s RUN_ARGS='--package ./integration-tests/k8s/tests/prometheus-operator'

# Skip building the Alloy image.
make integration-test-k8s RUN_ARGS='--skip-alloy-build'
```


kalleep:
>missing required command: rg

Should probably use grep for better portability?

integration-tests/k8s/tests/prometheus-operator/main_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+package prometheusoperator
+
+import (
+	"testing"
+
+	"github.com/grafana/alloy/integration-tests/k8s/harness"
+)
+
+func TestMain(m *testing.M) {
```
kalleep: Ah nice, I really like this! Hardly every use TestMain so kinda forgot it existed

integration-tests/k8s/harness/shard.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"hash/fnv"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+var shardFlag = flag.String("shard", "", "run only tests for shard i/n")
+
+type shardConfig struct {
```
kalleep, line 17: I have a hard time figuring out what this is, mind explaning?

integration-tests/k8s/harness/testmain.go
```diff
@@ -0,0 +1,196 @@
...
+	Namespace            string
+	TestID               string
+	MimirLocalPort       string
+	Shard                shardConfig
+	AlloyImageRepository string
+	AlloyImageTag        string
+	client               *kubernetes.Clientset
+}
+
+var current *TestContext
```
kalleep: I don't like this but as far as I know there is no good way to inject this into every test.

I tried to see if it would be possible to inject this into the context used for every test from TestMain but that does not seem possible

kalleep (commented):
This is nice, great work. 

Left a couple of comments but mainly I don't really understand shards

integration-tests/k8s/harness/shard.go
thampiotr: The idea is that when we have a lot of integration tests, these may become slow to run, so we can shard, for example, 50 tests to run in parallel on 10 shards, each shard running 5 tests. Helps to scale integration tests.

integration-tests/k8s/harness/install.go
```diff
@@ -0,0 +1,133 @@
...
+
+func applyWorkloads(path, namespace string) error {
+	absPath, err := filepath.Abs(path)
+	if err != nil {
+		return fmt.Errorf("resolve workloads path: %w", err)
+	}
+	_ = namespace
+	return step("apply workloads", func() error {
+		return runCommand("kubectl", "apply", "-f", absPath)
+	})
```
kgeckhart: Looks like this is going to discard the namespace param, should it be part of the command?

integration-tests/k8s/harness/mimir.go
```diff
@@ -0,0 +1,194 @@
...
+		case <-time.After(5 * time.Second):
+		}
+	}, nil
+}
+
+func collectFailureDiagnostics(ctx *TestContext) {
+	fmt.Printf("[k8s-itest] collecting failure diagnostics namespace=%s\n", ctx.Namespace)
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "get", "pods", "-o", "wide")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "describe", "pods")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "logs", "deployment/alloy", "--all-containers=true", "--tail", "200")
```
kgeckhart: Is alloy always going to be a deployment here? We support statefulset and daemonset too right?

integration-tests/k8s/run.sh
kgeckhart, line 1: Just wondering about this being a shell script vs doing a main entrypoint which can reuse some of the harness code to ensure config/help stays in sync easier.

integration-tests/k8s/README.md
```diff
@@ -1,23 +1,27 @@
...
-To debug the test you can also set two environment variables:
-* `ALLOY_STATEFUL_K8S_TEST=true` will retain the k8s clusters after the test terminates.
-* `ALLOY_K8S_TEST_LOGGING=debug` will get the test to print log messages.
-
+`integration-tests/k8s/run.sh` is the canonical entrypoint. It always uses a
+runner-managed kind cluster and kubeconfig (never your default kube context),
+then executes `go test` for `integration-tests/k8s/tests/...`.
 
-For example:
+Useful options (forwarded with `RUN_ARGS`):
```
kgeckhart: I think some of these could use an explanation as to why they are useful. Like `--shard 0/2` is that doing something special? Why would I want to change the controller type in my test run?

integration-tests/k8s/harness/shard.go
kgeckhart: I think some comments about _what_ is being sharded would be helpful. IIUC it's the packages.

integration-tests/k8s/harness/shard.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (s shardConfig) shouldRun(key string) bool {
+	if s.total == 0 {
+		return true
+	}
+	hasher := fnv.New32a()
+	_, _ = hasher.Write([]byte(key))
+	return int(hasher.Sum32()%uint32(s.total)) == s.index
+}
```
kgeckhart, line 65: Could we replace this with `index % total` for a more deterministic round-robin? I feel like hashing has the potential to create an uneven distribution.

integration-tests/k8s/harness/testmain.go
```diff
@@ -0,0 +1,209 @@
...
+const (
+	BackendMimir Backend = "mimir"
+)
+
+type Options struct {
+	Name         string
+	ConfigPath   string
+	Workloads    string
+	Backends     []Backend
+	PodWaits     []string
```
kgeckhart: Looks like this is unused

integration-tests/k8s/harness/testmain.go
```diff
@@ -0,0 +1,209 @@
...
+	Namespace    string
+	AlloyRelease string
```
kgeckhart: Also looks unused

integration-tests/k8s/harness/mimir.go
```diff
@@ -0,0 +1,194 @@
...
+		pods, err := ctx.client.CoreV1().Pods(namespace).List(t.Context(), metav1.ListOptions{
+			LabelSelector: labelSelector,
+		})
+		require.NoError(c, err)
+		require.NotEmpty(c, pods.Items, "no pods for namespace=%s selector=%s", namespace, labelSelector)
+		for _, pod := range pods.Items {
+			require.Nil(c, pod.DeletionTimestamp, "pod %s is deleting", pod.Name)
+			require.Equal(c, corev1.PodRunning, pod.Status.Phase, "pod %s is not running", pod.Name)
+		}
+	}, 5*time.Minute, 2*time.Second)
```
kgeckhart:
```suggestion
	}, timeout, 2*time.Second)
```

integration-tests/k8s/harness/mimir.go
```diff
@@ -0,0 +1,194 @@
...
+	}, nil
+}
+
+func collectFailureDiagnostics(ctx *TestContext) {
+	fmt.Printf("[k8s-itest] collecting failure diagnostics namespace=%s\n", ctx.Namespace)
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "get", "pods", "-o", "wide")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "describe", "pods")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "logs", "deployment/alloy", "--all-containers=true", "--tail", "200")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "logs", "deployment/prom-gen", "--all-containers=true", "--tail", "200")
+	_ = runCommand("kubectl", "--namespace", ctx.Namespace, "logs", "deployment/blackbox-exporter", "--all-containers=true", "--tail", "200")
```
kgeckhart: Do these deployments exists in all the tests? Is it something that might need to configurable per test?

integration-tests/k8s/tests/mimir-alerts-kubernetes/config/config.alloy
```diff
@@ -0,0 +1,24 @@
+remote.kubernetes.configmap "default" {
+  namespace = "test-mimir-alerts-kubernetes"
```
kgeckhart: Should we potentially allow namespace templating here like `${NAMESPACE}`? This would prevent someone from needing to know the conventional namespace name (https://github.com/grafana/alloy/pull/6125/changes#diff-f1dc67f6a8018e711590a9537c8e75385b02c0b4502b39073284cb882dd32e11R80) and/or making sure it isn't overridden by setting `harness.Options.Namespace`

kgeckhart (commented):
The ideal looks solid! Some suggestions

integration-tests/k8s/README.md
thampiotr: Sorry, this should have been still a draft, but your feedback has been useful regardless :) 
- yes I agree, will document a bit better what these shards are and why would you want to manually set them (this is for repro of a CI failure, for example)
- I am refining the draft so that tests themselves define the controller type and in the future they can also define a completely custom values.yaml for Alloy helm chart for full control.

integration-tests/k8s/harness/install.go
thampiotr: that was old draft idea not cleaned up - we allow tests to put workloads in other namespaces too, so we don't pass namespace here any more

integration-tests/k8s/harness/mimir.go
thampiotr: yes, all deployment modes must be supported by framework - this was WIP

integration-tests/k8s/harness/mimir.go
thampiotr: I'm going to try an approach where we can register diagnostic hooks so we can configure these per test

integration-tests/k8s/harness/shard.go
thampiotr: good points, I will make it more clear before finalising the draft

integration-tests/k8s/harness/shard.go
thampiotr: sure! we can always tweak this stuff in the future

thampiotr: Hmm, getting a list of all packages from inside a go test run of a single package is a bit messy. So I'm leaving this for now as a hash(name) % total.

integration-tests/k8s/tests/mimir-alerts-kubernetes/config/config.alloy
thampiotr: I would maybe defer this for later if this turns out to be an annoyance, because adding templating on top of the configs adds a bit more complexity and things to also document... and some tests may use a few namespaces.

integration-tests/k8s/harness/testmain.go
thampiotr: I thought more about this and managed to get rid of it by making the approach to dependency slightly different, but should be still good.

integration-tests/k8s/tests/mimir-alerts-kubernetes/config/config.alloy
thampiotr: ended up doing this as it helped in the end make things easier!

integration-tests/k8s/deps/alloy.go
```diff
@@ -0,0 +1,171 @@
...
+		if err := createAlloyConfigMap(a.opts.Namespace, a.opts.ConfigPath); err != nil {
+			return err
+		}
+	}
+
+	image := os.Getenv(alloyImageEnv)
+	imageRepo, imageTag, ok := strings.Cut(image, ":")
+	if !ok {
+		return fmt.Errorf("%s must be in repo:tag format (the test runner sets this), got %q", alloyImageEnv, image)
+	}
```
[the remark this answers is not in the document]
thampiotr: I don't think we need to be doing it since we have full control over setting ALLOY_TESTS_IMAGE.

integration-tests/k8s/run.sh
thampiotr: turned it into Go and even added some interactive menus for local development, because I started to dislike all the flags etc...

integration-tests/k8s/harness/env.go
```diff
@@ -0,0 +1,54 @@
...
+// CommandEnv returns the process environment with KUBECONFIG forced to the
+// managed test kubeconfig (when set). Pass it as cmd.Env when running a long
+// lived command directly with exec.Cmd; for one-shot invocations prefer the
+// RunCommand* helpers which apply this automatically.
+func CommandEnv() []string {
+	env := os.Environ()
+	if kubeconfig := os.Getenv(kubeconfigEnv); kubeconfig != "" {
+		env = append(env, "KUBECONFIG="+kubeconfig)
+	}
+	return env
```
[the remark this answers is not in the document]
thampiotr: fixed

integration-tests/k8s/tests/mimir-alerts-kubernetes/config/workloads.yaml
```diff
@@ -0,0 +1,108 @@
...
+  receivers:
+  - name: "null"
+  - name: "database-pager"
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: othernamespace
+  labels:
+    alloy-integration-test: "true"
```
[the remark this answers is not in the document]
thampiotr: that was intentional for test, added comment

integration-tests/k8s/deps/mimir.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func curl(c *assert.CollectT, targetURL string) string {
+	resp, err := http.Get(targetURL)
+	require.NoError(c, err)
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(c, err)
+	return string(body)
```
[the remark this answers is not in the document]
thampiotr: fixed

integration-tests/k8s/runner/main.go
```diff
@@ -0,0 +1,254 @@
...
+
+func main() {
+	cfg, err := parseFlags()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if err := requireCommands("docker", "kind", "kubectl", "helm", "go", "make"); err != nil {
+		fmt.Println(err)
```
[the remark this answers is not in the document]
thampiotr: we consider make to be broadly available

integration-tests/k8s/runner/main.go
```diff
@@ -0,0 +1,258 @@
...
+	fs.StringVar(&cfg.alloyImage, "alloy-image", "grafana/alloy:latest", "Alloy image (repo:tag) used by tests; must exist locally or in the kind cluster")
+	fs.BoolVar(&cfg.interactive, "interactive", false, "Pick run options (reuse-cluster, skip-image-builds, shard/packages) via an interactive menu before running")
+	fs.Usage = func() {
+		fmt.Println("Usage: go run ./integration-tests/k8s/runner [flags]")
+		fmt.Println()
+		fs.PrintDefaults()
+	}
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return config{}, err
+	}
```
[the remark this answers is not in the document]
thampiotr: the current code path handles this correctly, this is more of a cosmetic and not obviously better

integration-tests/k8s/runner/main.go
```diff
@@ -0,0 +1,256 @@
...
+	// -count=1 disables Go's test cache so re-runs always exercise the live cluster.
+	args := []string{"test", "-v", "-count=1", "-timeout", "30m"}
+	args = append(args, patterns...)
+	if cfg.shard != "" {
+		args = append(args, "-args", "-shard="+cfg.shard)
+	}
+	return util.Step("go test "+strings.Join(patterns, " "), func() error {
+		return harness.RunCommand("go", args...)
+	})
```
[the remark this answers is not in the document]
thampiotr: We want parallelism here. Isolation is by namespaces. If we do share deps in the future, we will handle their multi-tenancy in some way and idempotent installs.

kalleep (approved):

integration-tests/k8s/deps/mimir.go
```diff
@@ -0,0 +1,287 @@
...
+			_ = cmd.Process.Kill()
+		}
+		select {
+		case <-waitCh:
+		case <-time.After(5 * time.Second):
+		}
+	}, nil
+}
+
+func pickFreeLocalPort() (string, error) {
```
kgeckhart, line 261: There's a bit of a race here it's probably fine at low scale. I don't think there's really any amazing solutions other than retrying on port bind failures.

kgeckhart (approved):
The PR is too big to review at much depth. AI commented on the piece I mentioned and a bit about gore2regex which is only relevant for loki secret filter,

```
runGoTests drops gore2regex tag — integration-tests/k8s/runner/main.go line 244 builds go test args as []string{"test", "-v", "-count=1", "-timeout", "30m"} with no -tags flag. The Makefile (lines 130–134) ensures gore2regex is injected for Makefile targets, but the new runner calls go test as a subprocess directly, bypassing that. Any loki component relying on loki.secretfilter will silently fall back to stdlib regex.
```

I don't know how long that tags going to live around though?

integration-tests/k8s/deps/mimir.go
thampiotr: Yes, that's why we're retrying in `startPortForwardWithRetries` 👍

thampiotr:
> runGoTests drops gore2regex tag — integration-tests/k8s/runner/main.go line 244 builds go test args as > 
> I don't know how long that tags going to live around though?

Ah, I didn't think we need `gore2regex`, but now it makes sense. I will bring it back in a follow-up 👍

PR #6128, opened 2026-04-24
docs(design): Add design document for zone-aware scraping

timonegk:

Add a design document for zone-aware scraping. Created from discussion in #6010.

### DIFF
--- a/docs/design/6128-zone-aware-scraping.md
+++ b/docs/design/6128-zone-aware-scraping.md
@@ -1,0 +1,120 @@
+# Proposal: Zone-Aware Metrics Scraping
+
+* Author(s): Timon Engelke
+* Last updated: 2026-04-24
+* Proposal issue: https://github.com/grafana/alloy/pull/6010
+
+## Abstract
+
+This proposal introduces a way for zone-aware metrics scraping by considering
+the zone of a metrics target when choosing the Alloy instance that scrapes it.
+
+## Problem
+
+Currently, in a distributed Alloy setup with clustering enabled, scrape targets
+are distributed to Alloy instances using a hash ring. The target distribution
+is deterministic based on the name of the Alloy instance and the scrape target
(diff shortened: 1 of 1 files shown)


kgeckhart (commented):
Thanks again for taking the time to write up the proposal and engage in the discussion. I really appreciate that flexibility as this is a really valuable feature for alloy.

I think I framed the conversation wrong when I started as it came off as needing to discuss the merits of the two options. Realistically both options have their own merits and they don't really conflict with each other. 

The main concern mostly comes back to our ability to support the local zone ring solution long term if we merge the PR. For a feature with this level of complexity, we should be confident that maintainers can validate future changes and troubleshoot issues when they arise. It's wonderful you're validating it by running it in the PR but when the first bug report lands or the operator evolves it's us who has to be able to do the same level of validation.

In practice, that confidence usually comes from us using it internally at scale. The way our environments work makes independent alloy clusters more attractive than one large multi-az alloy cluster. Since it's not something we can easily verify ourselves at scale long term we likely need a strong integration test suite which can validate standard operation and various failure modes. I'll discuss further with the team and see if there's other areas that would be good to refine further.

timonegk:
@kgeckhart have you discussed with the team?
I have another proposal that could possible reduce the extent of the change but still fulfill our need. Since Kubernetes 1.35, topology labels are propagated from a node to the pods. So the scrapetarget -> node -> AZ mapping is no longer required. Filtering of scrape targets could be done via alloy config. But we would need to have separate hash rings for the availability zones. Would it be an option to add `clustering { ring = "<...>" }` or similar to the Prometheus Operator components? All instances with the same `ring` value would be in a dedicated ring. This way, we could do the AZ-local scaping using three separate rings.
Otherwise, if this is still too much added complexity, would you consider extending the Helm Chart to deploy multiple zone-specific deployments/stateful sets? Similar to how we have zone-aware stateful sets for Mimir ingesters or store gateways.

PR #6134, opened 2026-04-27
chore(loki): Add new pipeline primitives

kalleep:
### Pull Request Details

This PR introduces Loki pipeline primitives that should be used by the new pipeline architecture. For now, these primitives are unused by the existing pipeline.

#### `Consumer`

A new `Consumer` interface with both `Consume` and `ConsumeEntry` methods. `Consume` is the long-term batch-oriented interface, while `ConsumeEntry` exists as a temporary migration helper so components can move from channel-based handoff to synchronous function calls before they are converted to full batch processing. For more context on that migration step, see #4953.

#### `Batch`

It groups entries into streams by label set and stores created time once per batch instead of once per entry.

Two important operations on `Batch` are `Apply` and `ConsumeStreams`:
1. `Apply` makes it possible to walk entries in place, rewrite them, drop them, or move them between streams when labels change. One use case for this is `loki.process`, where a batch should be passed through the stage pipeline while still exposing the internal `loki.Entry` type that those stages already operate on.
2. `ConsumeStreams` hands batch contents off stream-by-stream. This is what `ShardingConsumer` uses to preserve per-stream ordering while still allowing different streams to make progress independently. It should also be useful when writing batches to the WAL, where entries for the same stream should be stored together in a record, and when adapting the `loki.write` client sharding and batch queueing mechanisms to operate on batch streams instead of individual entries.

#### `FanoutConsumer`

`FanoutConsumer` forwards batches or entries to multiple downstream consumers synchronously. It clones data for all but the final consumer so downstream consumers can safely mutate batches or entries without affecting each other.

#### `ShardingConsumer`

`ShardingConsumer` shards batches / entries by stream labels so that entries for the same stream stay ordered while different streams can make progress independently. It also gives sources and downstream components a synchronous backpressure boundary instead of relying on unbounded internal fanout.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -8,28 +8,21 @@ import (
 	"github.com/prometheus/common/model"
 )
 
-type EntryAction uint8
-
-const (
-	ActionKeep EntryAction = iota
-	ActionDrop
-)
-
 type Batch struct {
 	// created is a unix timestamp in micro seconds.
 	created  int64
-	streams  []Stream
 	entryLen int
+	streams  []Stream
 }
 
 // NewBatch creates an empty Batch.
 func NewBatch() Batch {
-	b := Batch{
-		created:  0,
-		streams:  []Stream{},
-		entryLen: 0,
-	}
-	return b
+	return Batch{}
+}
+
+// NewBatchWithCreatedUnixMicro creates an empty Batch with created.
+func NewBatchWithCreatedUnixMicro(created int64) Batch {
+	return Batch{created: created}
 }
 
 // Add adds a stream to the batch.
@@ -56,6 +49,13 @@ func (b *Batch) add(labels model.LabelSet, entries ...push.Entry) {
 	b.streams = append(b.streams, NewStream(labels, entries...))
 }
 
+type EntryAction uint8
+
+const (
+	ActionKeep EntryAction = iota
+	ActionDrop
+)
+
 // IterMut calls fn for each entry in the batch.
 // Kept entries are written back, dropped entries are removed,
 // and entries whose labels change are moved to a different stream.
@@ -71,8 +71,8 @@ func (b *Batch) IterMut(fn func(entry *Entry) EntryAction) {
 	)
 
 	// Process each entry and compact each stream in place.
-	// The callback mutates a temporary Entry view; kept entries are written back,
-	// dropped entries are skipped, and entries whose labels change are deferred
+	// The callback mutates a temporary Entry view. Kept entries are written back,
+	// dropped entries are skipped, and moved entries are deferred
 	// so we do not mutate the stream set while iterating it.
 	for i := range b.streams {
 		var (
@@ -133,6 +133,10 @@ func (b *Batch) EntryLen() int {
 	return b.entryLen
 }
 
+func (b *Batch) Created() int64 {
+	return b.created
+}
+
 // Clone returns a clone of the batch.
 func (b *Batch) Clone() Batch {
 	clone := Batch{
@@ -151,9 +155,9 @@ func (b *Batch) Clone() Batch {
 
 // ConsumeStreams calls fn for each stream in the batch and then resets the batch.
 // The callback receives ownership of the stream.
-func (b *Batch) ConsumeStreams(fn func(Stream)) {
+func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64)) {
 	for _, s := range b.streams {
-		fn(s)
+		fn(s, b.created)
 	}
 	b.Reset()
 }
@@ -187,13 +191,15 @@ func (s Stream) Clone() Stream {
 	}
 	for _, entry := range s.Entries {
 		e := push.Entry{
-			Timestamp:          entry.Timestamp,
-			Line:               entry.Line,
-			StructuredMetadata: make(push.LabelsAdapter, 0, len(entry.StructuredMetadata)),
+			Timestamp: entry.Timestamp,
+			Line:      entry.Line,
 		}
 
-		for _, s := range entry.StructuredMetadata {
-			e.StructuredMetadata = append(e.StructuredMetadata, push.LabelAdapter{Name: s.Name, Value: s.Value})
+		if entry.StructuredMetadata != nil {
+			e.StructuredMetadata = make(push.LabelsAdapter, 0, len(entry.StructuredMetadata))
+			for _, s := range entry.StructuredMetadata {
+				e.StructuredMetadata = append(e.StructuredMetadata, push.LabelAdapter{Name: s.Name, Value: s.Value})
+			}
 		}
 
 		cloned.Entries = append(cloned.Entries, e)
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -82,7 +82,7 @@ func (b *Batch) FilterMap(fn func(entry *Entry) FilterMapAction) {
 
 		for _, e := range stream.Entries {
 			// FIXME(kalleep): This clone protects stream.Labels from in-place mutation
-			// through Entry.Labels, which is a map. Most IterMut callbacks do not change
+			// through Entry.Labels, which is a map. Most FilterMap callbacks do not change
 			// labels, so once the new batched pipeline owns this path, consider replacing
 			// direct label access with copy-on-write label mutation methods.
 			entry := NewEntryWithCreatedUnixMicro(stream.Labels.Clone(), b.created, e)
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -49,17 +49,10 @@ func (b *Batch) add(labels model.LabelSet, entries ...push.Entry) {
 	b.streams = append(b.streams, NewStream(labels, entries...))
 }
 
-type FilterMapAction uint8
-
-const (
-	FilterMapActionKeep FilterMapAction = iota
-	FilterMapActionDrop
-)
-
-// FilterMap calls fn for each entry in the batch.
-// Kept entries are written back, dropped entries are removed,
-// and entries whose labels change are moved to a different stream.
-func (b *Batch) FilterMap(fn func(entry *Entry) FilterMapAction) {
+// FilterMap calls fn for each entry in the batch. If fn returns true the
+// entry is kept, if fn returns false the entry is dropped. Kept entries are
+// written back, and entries whose labels change are moved to a different stream.
+func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
 	type movedEntry struct {
 		labels model.LabelSet
 		entry  push.Entry
@@ -86,9 +79,7 @@ func (b *Batch) FilterMap(fn func(entry *Entry) FilterMapAction) {
 			// labels, so once the new batched pipeline owns this path, consider replacing
 			// direct label access with copy-on-write label mutation methods.
 			entry := NewEntryWithCreatedUnixMicro(stream.Labels.Clone(), b.created, e)
-			action := fn(&entry)
-
-			if action == FilterMapActionDrop {
+			if !fn(&entry) {
 				continue
 			}
 
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -111,7 +111,7 @@ func (b *Batch) IterMut(fn func(entry *Entry) EntryAction) {
 	}
 	b.entryLen = newLen
 
-	// Remove any empty stream.
+	// Remove any empty streams.
 	streamDst := 0
 	for i := range b.streams {
 		if len(b.streams[i].Entries) == 0 {
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -1,0 +1,199 @@
+package loki
+
+import (
+	"slices"
+	"time"
+
+	"github.com/grafana/loki/pkg/push"
+	"github.com/prometheus/common/model"
+)
+
+type EntryAction uint8
+
+const (
+	ActionKeep EntryAction = iota
+	ActionDrop
+)
+
+type Batch struct {
+	// created is a unix timestamp in micro seconds.
+	created  int64
+	streams  []Stream
+	entryLen int
+}
+
+// NewBatch creates an empty Batch.
+func NewBatch() Batch {
+	b := Batch{
+		created:  time.Now().UnixMicro(),
+		streams:  []Stream{},
+		entryLen: 0,
+	}
+	return b
+}
+
+// Add adds a stream to the batch.
+// Ownership of the stream data is transferred to the batch and it must not be
+// mutated or retained after calling Add.
+func (b *Batch) Add(stream Stream) {
+	b.add(stream.Labels, stream.Entries...)
+	b.entryLen += len(stream.Entries)
+}
+
+func (b *Batch) add(labels model.LabelSet, entries ...push.Entry) {
+	i := slices.IndexFunc(b.streams, func(s Stream) bool {
+		return s.Labels.Equal(labels)
+	})
+
+	if i >= 0 {
+		b.streams[i].Entries = append(b.streams[i].Entries, entries...)
+		return
+	}
+
+	b.streams = append(b.streams, NewStream(labels, entries...))
+}
+
+// IterMut calls fn for each entry in the batch.
+// Kept entries are written back, dropped entries are removed,
+// and entries whose labels change are moved to a different stream.
+func (b *Batch) IterMut(fn func(entry *Entry) EntryAction) {
+	type movedEntry struct {
+		labels model.LabelSet
+		entry  push.Entry
+	}
+
+	var (
+		newLen int
+		moves  []movedEntry
+	)
+
+	// Process each entry and compact each stream in place.
+	// The callback mutates a temporary Entry view; kept entries are written back,
+	// dropped entries are skipped, and entries whose labels change are deferred
+	// so we do not mutate the stream set while iterating it.
+	for i := range b.streams {
+		var (
+			dst    = 0
+			stream = &b.streams[i]
+		)
+
+		for _, e := range stream.Entries {
+			entry := NewEntryWithCreatedUnixMicro(stream.Labels.Clone(), b.created, e)
+			action := fn(&entry)
+
+			if action == ActionDrop {
+				continue
+			}
+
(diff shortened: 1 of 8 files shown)


internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	ActionKeep EntryAction = iota
+	ActionDrop
+)
+
+// IterMut calls fn for each entry in the batch.
+// Kept entries are written back, dropped entries are removed,
+// and entries whose labels change are moved to a different stream.
+func (b *Batch) IterMut(fn func(entry *Entry) EntryAction) {
```
kalleep: Any suggestions for a better name would be great. Essentially this functions is supposed to be used when we need to mutate entries, for now this returns our internal `loki.Entry` structure since all components operate on that thus we have to clone the shared labels when doing so since they can be mutated, ideally we could optimize this part, maybe some kind of Copy on Write wrapper since we don't always will mutate labels.

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,208 @@
...
+
+import (
+	"slices"
+	"time"
+
+	"github.com/grafana/loki/pkg/push"
+	"github.com/prometheus/common/model"
+)
+
+type Batch struct {
```
kalleep, line 11: I realize that this `Batch` implementation could be a bit controversial. Ideally we should do something like this since a lot of entries will share labels, e.g. lets say that we read 100 lines from a single file and forward that as a batch, all those 100 entries will share the same labels, filename + discovery labels. 

But when using e.g. `loki.process` we can add / remove / change labels for a single entry based on parsed log lines etc so we must clone shared labels during iteration.

A much simpler implementation of batch would be to just store []loki.Entry directly. We can still support the same operations such as `ConsumeStreams` and `IterMut`, but we would probably still have to clone the labels like we do [inline in loki.process currently](https://github.com/grafana/alloy/blob/fcae6f4beff1d582c7e3ac0708256ef12ab10302/internal/component/loki/process/process.go#L191)

kalleep:
@codex review

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,208 @@
...
+func (s Stream) Clone() Stream {
+	cloned := Stream{
+		Labels:  s.Labels.Clone(),
+		Entries: make([]push.Entry, 0, len(s.Entries)),
+	}
+	for _, entry := range s.Entries {
+		e := push.Entry{
+			Timestamp: entry.Timestamp,
+			Line:      entry.Line,
+		}
```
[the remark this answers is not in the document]
kalleep: In alloy we never use `Parsed`, it's internal field used by loki

internal/component/common/loki/consumer_fanout.go
```diff
@@ -0,0 +1,99 @@
...
+
+	for i, consumer := range f.consumers {
+		if i == len(f.consumers)-1 {
+			if err := consumer.ConsumeEntry(ctx, entry); err != nil {
+				errs = append(errs, err)
+			}
+			continue
+		}
+
+		if err := consumer.ConsumeEntry(ctx, entry.Clone()); err != nil {
```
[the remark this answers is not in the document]
kalleep: Yeah I am aware of this, this is how it currently functions but need to look into if this is correct in fan-out scenarios.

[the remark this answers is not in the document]
kalleep: Fixed in https://github.com/grafana/alloy/pull/6138

kalleep:
@codex review

internal/component/common/loki/consumer_sharding.go
```diff
@@ -0,0 +1,155 @@
...
+	errChans := make([]<-chan error, 0, streamLen)
+
+	// NOTE: when we only have one stream in a batch there is no need to split it.
+	if streamLen == 1 {
+		errChans = append(errChans, s.consume(ctx, s.shardFor(batch.streams[0].Labels), batch))
+	} else {
+		batch.ConsumeStreams(func(stream Stream, created int64) {
+			streamBatch := NewBatchWithCreatedUnixMicro(created)
+			streamBatch.Add(stream)
+			errChans = append(errChans, s.consume(ctx, s.shardFor(stream.Labels), streamBatch))
```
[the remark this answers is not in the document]
kalleep: This is a good point but I will not do anything about this for now. We could think about chaining this in the future

internal/component/common/loki/consumer.go
```diff
@@ -0,0 +1,69 @@
...
+
+var _ Consumer = (*CollectingConsumer)(nil)
+
+func NewCollectingConsumer() *CollectingConsumer {
+	return &CollectingConsumer{}
+}
+
+// CollectingConsumer is a Consumer that will collect all received entries
+// and batches so it can be inspected later.
+// Used in tests.
```
kgeckhart, line 22: Nit: should we move this to a test package?

internal/component/common/loki/consumer_fanout.go
```diff
@@ -0,0 +1,99 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (f *FanoutConsumer) Update(consumers []Consumer) {
+	f.mut.RLock()
+	if requireUpdate(f.consumers, consumers) {
```
kgeckhart, line 90: This is referencing the collecting consumer implementation right? Since that one is used for tests perhaps it should live here?

internal/component/common/loki/consumer_fanout.go
```diff
@@ -0,0 +1,99 @@
...
+		// Upgrade lock to write.
+		f.mut.RUnlock()
+		f.mut.Lock()
```
kgeckhart, line 93: Can update be called multiple times in parallel? There's a gap here where something else could have updated the config rendering this update ineffective. I don't think it matters since taking the write lock is the heaviest bit and we need to do that before we can double check.

internal/component/common/loki/consumer_fanout_test.go
kgeckhart, line 1: Could we add some tests for `Update`?

internal/component/common/loki/consumer_sharding_test.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+)
+
+func TestShardingConsumer_Consume(t *testing.T) {
+	t.Run("splits batch into streams", func(t *testing.T) {
+		const created = int64(1234)
+
+		first := model.LabelSet{"job": "fist"}
```
kgeckhart:
```suggestion
		first := model.LabelSet{"job": "first"}
```

kgeckhart (commented):
Small things, will respond to the batching thing inline

internal/component/common/loki/batch.go
kgeckhart: Is your plan to keep this functionality the same through the implementation or to aide in the migration? IIUC it's a deviation from the original plan which is probably fine. I'm not sure I fully grasp the tradeoff yet. 

Is it something where we could potentially quantify tradeoffs in a benchmark simulating some realistic use cases?

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,206 @@
...
+		return s.Labels.Equal(labels)
+	})
+
+	if i >= 0 {
+		b.streams[i].Entries = append(b.streams[i].Entries, entries...)
+		return
+	}
+
+	b.streams = append(b.streams, NewStream(labels, entries...))
+}
```
jharvey10, line 50: Would it be cleaner to have `add` own the `EntryLen` tracking? I see `Add` and `IterMut` both modify it. If another call site to `add` ever gets added, this could be easily missed as state that needs updating as well.

internal/component/common/loki/consumer_fanout.go
```diff
@@ -0,0 +1,99 @@
...
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (f *FanoutConsumer) ConsumeEntry(ctx context.Context, entry Entry) error {
+	// NOTE: It's important that we hold a read lock for the duration of ConsumeEntry
+	// rather than making a copy of children and releasing the lock early.
```
jharvey10: Should this be `consumers`?

```suggestion
	// rather than making a copy of consumers and releasing the lock early.
```

internal/component/common/loki/batch.go
jharvey10: I’m wondering if a way to split this could be: callers classify each entry’s outcome, while `Batch` owns applying that outcome to its stream storage. The current `*Entry` mutation API blurs that a bit because the callback mutates data directly and `Batch` then infers whether the entry was dropped/moved/kept.

Would a “filter/map result” style API make this easier to name and reason about? Something like `FilterMap(func(Entry) FilterMapResult)`, where the callback returns `Keep`, `Drop`, or `Move(labels, entry)`. That reads more like filter/map/re-key, and keeps compaction/re-bucketing as an internal `Batch` responsibility.

e.g.:

```go
for _, e := range stream.Entries {
    // No need to clone now
    res := fn(e) 

    switch res.Action {
    case ActionDrop:
        continue
    case ActionMove:
        moves = append(moves, movedEntry{
            labels: res.NewLabels,
            entry:  e,
        })
        newLen++
    case ActionKeep:
        stream.Entries[dst] = e
        dst++
        newLen++
    }
}
```

internal/component/common/loki/consumer.go
kalleep: Maybe, we have a couple of these, like [CollectingHandler](https://github.com/grafana/alloy/blob/9e6bc5e4ec2774cb829f6604b5e2b16a809d701d/internal/component/common/loki/entry_handler.go#L148) and [CollectingBatchReceiver](https://github.com/grafana/alloy/blob/9e6bc5e4ec2774cb829f6604b5e2b16a809d701d/internal/component/common/loki/receiver.go#L88).

But it's not a bad idea to have a "test" package for loki pipeline specific implementations

internal/component/common/loki/consumer_fanout.go
kalleep: No update cannot be called multiple times in parallel. 

But you are correct that it could technically happen, i.e. a TOCTOU. But still this is a good point. We could perform the check again but as you said I don't think that matters

internal/component/common/loki/consumer_fanout_test.go
kalleep: :+1:

internal/component/common/loki/consumer_fanout.go
kalleep: The thing about the collecting variant is that we will most likely use that one a lot of other tests as well. It's convenient to use when testing e.g. components since we have an implementation that will just store the entries for us so we can assert on the result. So I don't think they are tied together.

internal/component/common/loki/batch.go
kalleep: >Is your plan to keep this functionality the same through the implementation or to aide in the migration? IIUC it's a deviation from the original plan which is probably fine. I'm not sure I fully grasp the tradeoff yet.

Yeah so first I want to have a data structure to hide the internals of a batch, in the design doc I just represented it as a slice of entries.

I started with just storing a slice of loki.Entry but there are problems with that too that still requires us to clone the labels. In some sources we set this labels without performing clone e.g. [loki.source.file](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/file/tailer.go#L227-L230) and in other we do a [clone before creating an entry](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/api/routes.go#L124).

That is why we currently always clone in [loki.process](https://github.com/grafana/alloy/blob/9e6bc5e4ec2774cb829f6604b5e2b16a809d701d/internal/component/loki/process/process.go#L191).

In a lot of sources we produce entries that at least originally shares labels, e.g. lines from a single file or from `loki.source.api` where entries already belongs to the same stream so it makes sense to me to internally store them as streams where they share labels, but that also means that we have to clone labels each time we want to mutate entries to make it safe, hence the callout to maybe use some copy on write semantics.

I want to eventually change / remove `loki.Entry`, i.e. add copy on write semantics so we can avoid the clone if we don't touch labels. But for the initial migration it is easier to do it like this because then we don't have to change stuff to much when we are in the middle of it.

internal/component/common/loki/batch.go
kalleep: I started with an api just like this. But there are some issues with it:
1. We cannot just pass a push.Entry but we also need to pass the current labels associated with it. I think can be a bit confusing but in alloy we currently have our own [Entry](https://github.com/grafana/alloy/blob/main/internal/component/common/loki/entry.go#L35-L45) wrapper around `push.Entry` that have labels attached to it.
2. If we don't clone the labels the caller can mutate them and it would change labels for all entries associated with the stream, and it would be up to the caller to make it safe, e.g. clone them.

internal/component/common/loki/batch.go
kalleep: Maybe but `IterMut` becomes a bit tricky then. We would have to decrement entryLen when entry gets removed and when entry move, since adding it to the correct stream would increment it.

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,206 @@
...
+
+const (
+	ActionKeep EntryAction = iota
+	ActionDrop
+)
+
+// IterMut calls fn for each entry in the batch.
+// Kept entries are written back, dropped entries are removed,
+// and entries whose labels change are moved to a different stream.
+func (b *Batch) IterMut(fn func(entry *Entry) EntryAction) {
```
kalleep: I really don't like the name `IterMut`, it's a convention from creating an iterator of mutable entities in rust.

Other suggestions:
* Rewrite
* Apply
* FilterMap

Any preference?

kalleep: For now I changed to Apply

x1unix: `FilterMap` imo seems more self-explanatory

jharvey10: +1. That's the direction I was headed with [my other comment](https://github.com/grafana/alloy/pull/6134#discussion_r3157785994) as well.

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,207 @@
...
+
+const (
+	ApplyActionKeep EntryAction = iota
+	ApplyActionDrop
+)
+
+// Apply calls fn for each entry in the batch.
+// Kept entries are written back, dropped entries are removed,
+// and entries whose labels change are moved to a different stream.
+func (b *Batch) Apply(fn func(entry *Entry) EntryAction) {
```
x1unix: Do we plan to support more actions?

If not - not sure whether enum is necessary here, boolean would work here.

internal/component/common/loki/batch.go
```diff
@@ -0,0 +1,207 @@
...
+	// so we do not mutate the stream set while iterating it.
+	for i := range b.streams {
+		var (
+			dst    = 0
+			stream = &b.streams[i]
+		)
+
+		for _, e := range stream.Entries {
+			// FIXME(kalleep): This clone protects stream.Labels from in-place mutation
+			// through Entry.Labels, which is a map. Most IterMut callbacks do not change
```
x1unix:
```suggestion
			// through Entry.Labels, which is a map. Most Apply callbacks do not change
```

internal/component/common/loki/consumer_sharding.go
```diff
@@ -0,0 +1,155 @@
...
+var _ Consumer = (*ShardingConsumer)(nil)
+
+var ErrConsumerStopped = errors.New("consumer stopped")
+
+// NewShardingConsumer creates a Consumer which shards streams across a fixed
+// number of shards before forwarding them to the downstream consumer.
+func NewShardingConsumer(shards int, consumer Consumer) *ShardingConsumer {
+	if shards <= 0 {
+		shards = 1
+	}
```
x1unix:
```suggestion
	shards = max(shards, 1)
```

internal/component/common/loki/batch.go
kalleep: Sure lets go with that

internal/component/common/loki/batch.go
jharvey10: Ahh, I think I see what you mean. 

> but we also need to pass the current labels associated with it

Rather than passing the labels directly, could we pass a small "view" with a `Get` method to approximate read-only access? And/or an [iterator](https://pkg.go.dev/iter) over the map instead of the raw map itself. Also maybe a `Clone` method to assist the copy-on-write in the callback?

Mostly just thinking out loud. I don't have a good handle on how the callbacks would work and if this would make their life easier or harder.

kalleep: >Rather than passing the labels directly, could we pass a small "view" with a Get method to approximate read-only access? And/or an [iterator](https://pkg.go.dev/iter) over the map instead of the raw map itself. Also maybe a Clone method to assist the copy-on-write in the callback?

Yes we should have some kind of copy-on-write structure here instead but I don't want to do that when doing the migration, it will require us to change to much at once. So I want to start with something like this and change it after like I mentioned in the comment

internal/component/common/loki/batch.go
kalleep: At some point I had a "Move" action but it was not necessary to have. I changed to use a bool instead

internal/component/common/loki/consumer_sharding_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+		batches := c.Batches()
+		require.Len(t, batches, 1)
+		require.Equal(t, 1, batch.StreamLen())
+		require.Equal(t, 2, batch.EntryLen())
+		batch.ConsumeStreams(func(stream Stream, _ int64) {
+			require.Equal(t, labels, stream.Labels)
+			require.Equal(t, "1", stream.Entries[0].Line)
+			require.Equal(t, "2", stream.Entries[1].Line)
+		})
```
kgeckhart: Do you need to get the first batch off of batches here to assert on?

kgeckhart (dismissed):
Very minor comment, nice work 🚀

internal/component/common/loki/consumer_sharding_test.go
kalleep: Ah yeah right..

kgeckhart (approved):

PR #6138, opened 2026-04-28
fix(loki): Clone structured metadata so we can perform fan-out correctly

kalleep:
### Pull Request Details
When performing fan-out of loki entries to multiple `loki.process` components that will mutate structured_metadata it will share the underlying slice even though we clone it, because it was only performing a shallow clone on embedded `push.Entry`.

We also have a slice for `Parsed` but this is not something we use in alloy. Ideally this would be solved at pipeline level and not in individual components like `loki.process` and `loki.relabel` as we currently do. I will change that in the bigger loki pipeline refactoring I am working on. But for now we can make a quick fix for it.

Its not trivial to make a test for it but I have added a test in https://github.com/grafana/alloy/pull/6049 that currently fails because we get a data race.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/entry.go
+++ b/internal/component/common/loki/entry.go
@@ -1,6 +1,7 @@
 package loki
 
 import (
+	"slices"
 	"time"
 
 	"github.com/grafana/loki/pkg/push"
@@ -47,8 +48,12 @@ type Entry struct {
 func (e *Entry) Clone() Entry {
 	return Entry{
 		Labels:  e.Labels.Clone(),
-		Entry:   e.Entry,
 		created: e.created,
+		Entry: push.Entry{
+			Timestamp:          e.Entry.Timestamp,
+			Line:               e.Entry.Line,
+			StructuredMetadata: slices.Clone(e.Entry.StructuredMetadata),
+		},
 	}
 }
 


internal/component/common/loki/entry.go
```diff
@@ -47,8 +48,12 @@ type Entry struct {
 func (e *Entry) Clone() Entry {
 	return Entry{
 		Labels:  e.Labels.Clone(),
-		Entry:   e.Entry,
 		created: e.created,
+		Entry: push.Entry{
+			Timestamp:          e.Entry.Timestamp,
+			Line:               e.Entry.Line,
+			StructuredMetadata: slices.Clone(e.Entry.StructuredMetadata),
```
blewis12, line 55: Do we not also want to copy over `e.Parsed`?

kalleep: We could but it's not something we use in alloy, it's internal to Loki. It where they out eg detected_level etc

blewis12: oh sorry, I just saw you already mentioned that in the PR description 🤦

blewis12 (approved):

PR #6143, opened 2026-04-28
feat(database_observability): Update wait_event_v2 to a 6-bucket taxonomy and surface Mysql nested events

gaantunes:
## Summary

Two changes to MySQL/Postgres `wait_event_v2` emission:

**1. 6-bucket taxonomy.** Replaces the 4-bucket scheme (`IO Wait` / `Network Wait` / `Lock Wait` / `Other Wait`) with a 6-bucket scheme that adds `Engine Wait` and `Replication Wait`. Engine-internal synchronization (Postgres `LWLock`/`BufferPin`/`IPC`, MySQL `wait/synch/*`) is split out of `Lock Wait` into `Engine Wait`. A name-prefix rule routes replication events into `Replication Wait` regardless of how the engine types them.

**2. Surface nested MySQL handler waits** (incorporated from #5935). When MySQL reports an outer `wait/io/table/sql/handler` event, that's a molecule wrapper, not the actual I/O wait. The query now `LEFT JOIN`s `events_waits_history` a second time to fetch the nested child event, and substitutes its fields when emitting `wait_event` / `wait_event_v2`. Both v1 and v2 paths benefit, so `wait_event_v2`'s classifier sees the real underlying event (e.g. `wait/io/file/innodb/innodb_data_file`) and routes it to `IO Wait` rather than misclassifying the wrapper.

## Why the taxonomy split

The previous scheme collapsed engine-internal synchronization (Postgres `LWLock`/`BufferPin`/`IPC`, MySQL `wait/synch/*`) and cascade locks (Postgres `Lock`, MySQL `wait/lock/*`) into a single `Lock Wait` bucket. The two have nothing in common operationally:

| | Lock Wait (cascade) | Engine Wait (internal synch) |
|---|---|---|
| Visible in `pg_locks` / `data_locks`? | yes | no |
| Hold duration | ms → seconds | ns → μs |
| Remediation | kill session / change app pattern | tune config / change query plan |
| Time to fix | minutes | hours-days |

A name-prefix rule (`SyncRep`, `WalSender`, `Recovery`, `LogicalApply`, etc. on Postgres; `relaylog*`, `Slave_*`, `Relay_log_info*`, `Replica_*`, `Source_info`, `Master_info`, `MYSQL_RELAY_LOG`, `Mts_*` on MySQL) runs *before* the type/prefix switch so replication events route to `Replication Wait` regardless of how the engine types them — stable across Postgres versions where the type assignment for replication events shifts (`Activity` in older PG → `Client`/`IPC`/`Timeout` in newer) and across MySQL inclusive-language renames (`Slave_*` → `Replica_*`, `Master_info` → `Source_info` in 8.0.22+).

## Why surfacing nested handler events

`wait/io/table/sql/handler` is reported by `performance_schema` as a molecule (wrapper) event — its `timer_wait` covers the whole table-access op, but the actual I/O is recorded as a nested child event in `events_waits_history`. Without surfacing the child, every InnoDB row-read shows up as the same `wait/io/table/sql/handler` line and downstream consumers cannot distinguish data-file reads from index lookups, log writes, etc.

Fixes https://github.com/grafana/grafana-dbo11y-app/issues/1301.

## Behaviour

- The taxonomy change is gated behind the existing `enable_pre_classified_wait_events` collector arg (off by default). Users who haven't opted in see no change to bucket labels.
- The v1 `wait_event` log line emits raw `wait_event_type` / `wait_event_name` (unchanged for opted-out users), but its `wait_event_name` and timer fields **do** now reflect the substituted nested event when the outer event is `wait/io/table/sql/handler`. This is the user-visible behavior change from #5935 and is the intended fix — wrapper events were previously hiding the actual I/O.
- For users on `enable_pre_classified_wait_events=true`: the bucket label changes for events previously assigned to `Lock Wait` or `Other Wait` that move to `Engine Wait` or `Replication Wait` per the canonical mapping below.

### Postgres — match on `(wait_event_type, wait_event)`

| `wait_event_type` | `wait_event` match | Bucket |
|---|---|---|
| `IO` | * | `IO Wait` |
| `Client` | * | `Network Wait` |
| `Lock` | * | `Lock Wait` |
| any type | name starts with `SyncRep`, `WalSender`, `WalReceiver`, `Recovery`, `LogicalApply`, `LogicalLauncher`, `LogicalSync`, `LogicalParallelApply`, `LogicalRep`, `ReplicationSlot`, or `ReplicationOrigin` | `Replication Wait` |
| `LWLock`, `BufferPin`, `IPC` | (not replication by name rule above) | `Engine Wait` |
| `Activity`, `Timeout`, `Extension`, `InjectionPoint` | (not replication by name rule above) | `Other Wait` |
| anything unknown | — | `Other Wait` |

The replication name-rule runs first, so `Activity:WalSenderMain`, `Timeout:RecoveryApplyDelay`, and `IPC:SyncRep` route to `Replication Wait` instead of their type's default bucket.

### MySQL — match on `wait_event_name`

| name prefix / pattern | Bucket |
|---|---|
| `wait/io/file/sql/relaylog*` | `Replication Wait` |
| `wait/synch/*/sql/Slave_*` | `Replication Wait` |
| `wait/synch/*/sql/Replica_*` | `Replication Wait` |
| `wait/synch/*/sql/Master_info*` | `Replication Wait` |
| `wait/synch/*/sql/Source_info*` | `Replication Wait` |
| `wait/synch/*/sql/Relay_log_info*` | `Replication Wait` |
| `wait/synch/*/sql/MYSQL_RELAY_LOG*` | `Replication Wait` |
| `wait/synch/*/sql/Mts_*` | `Replication Wait` |
| `wait/io/file/*` | `IO Wait` |
| `wait/io/table/*` | `IO Wait` |
| `wait/io/socket/*` | `Network Wait` |
| `wait/io/lock/*` | `Lock Wait` |
| `wait/lock/*` | `Lock Wait` |
| `wait/synch/*` (everything else) | `Engine Wait` |
| anything else | `Other Wait` |

Replication carve-outs are checked before the generic prefix switch, so e.g. `wait/synch/mutex/sql/Relay_log_info::pending_jobs_lock` routes to `Replication Wait`, not `Engine Wait`. The classifier is fed the **substituted** event name when the raw row is `wait/io/table/sql/handler` with a nested child, so an InnoDB data-file read shows up as `IO Wait` rather than depending on how the wrapper would be classified.

## Files changed

- `internal/component/database_observability/postgres/collector/query_samples.go` + `_test.go` — new `isPostgresReplicationWaitEvent` helper, `classifyPostgresWaitEventType` extended to `(rawType, waitEvent string)`.
- `internal/component/database_observability/mysql/collector/query_samples.go` + `_test.go` —
  - new `isMySQLReplicationWaitEvent` helper, `classifyMySQLWaitEventType` adds `Engine Wait` for `wait/synch/*`, narrows `Lock Wait` to `wait/lock/*` + `wait/io/lock/*`;
  - second `LEFT JOIN events_waits_history nested_waits` and 6 new struct/scan fields for the nested event;
  - emission block rewritten so both v1 and v2 wait-event lines use the substituted fields (event id, name, object, timer) when an outer `wait/io/table/sql/handler` has a nested child.

## Test plan

- [x] `go test -race -tags="nodocker" ./internal/component/database_observability/...` passes
- [x] `go vet ./internal/component/database_observability/...` clean
- [x] `gofmt -l` clean on modified files
- [x] `go generate ./internal/tools/docs_generator/` produces no drift
- [x] Test tables cover: all 6 buckets per engine; replication name-rule firing across multiple `wait_event_type`s on PG (Client/Activity/IPC/Timeout) and across MySQL replication renames (Slave_/Replica_/Master_info/Source_info/MYSQL_RELAY_LOG/Mts_); MDL_lock::rwlock vs mdl distinction on MySQL; defensive empty/unknown inputs
- [x] Nested-handler test (with-nested → uses nested fields) ported from #5935. The fall-through case (no nested) is exercised implicitly by every existing wait-event fixture, all of which have nil nested values.
- [x] Cross-engine consistency: bucket strings, helper naming, replication-first invariant aligned between PG and MySQL

## Out of scope

- A 7th `Commit Wait` bucket for `MYSQL_BIN_LOG::COND_done` (write-path serialization) — argued for during analysis but not justified by enough evidence to commit to here.
- Sub-categorization of `Replication Wait` into idle vs lag — DBAs can filter by `wait_event` for the distinction.
- Routing `IO:SnapbuildSync` (logical-replication snapshot builder) to Replication — the wait *type* is genuinely `IO` and the diagnostic playbook is the IO one; classification stays with the type.

## Supersedes

- https://github.com/grafana/alloy/pull/5935 — those three commits are cherry-picked into this branch (preserving authorship and messages); #5935 can be closed once this merges.


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -636,23 +637,70 @@ func (c *QuerySamples) buildWaitEventV2Labels(state *SampleState, we WaitEventOc
...
-	case "Lock", "LWLock", "Activity", "Extension", "InjectionPoint", "IPC", "Timeout", "BufferPin":
-		return "Lock Wait"
 	case "Client":
 		return "Network Wait"
+	case "Lock":
+		return "Lock Wait"
+	case "LWLock", "BufferPin", "IPC":
+		return "Engine Wait"
+	case "Activity", "Timeout", "Extension", "InjectionPoint":
+		return "Other Wait"
```
cristiangreco: nit: I guess this can be merged with the `default` clause

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -170,6 +170,26 @@ The `gcp` block supplies the identifying information for the GCP Cloud SQL datab
 | `wait_event_min_duration`        | `duration` | Minimum duration for a wait event to be collected. Set to "0s" to disable filtering and collect all wait events regardless of their duration.  | `"1us"` | no       |
 | `enable_pre_classified_wait_events` | `boolean` | When `true`, emits telemetry data with pre-classified wait event information. | `false` | no       |
 
+#### `wait_event_type` values emitted by `wait_event_v2`
```
cristiangreco: TBH I think all this new documentation is unnecessary here. These files should document mainly component settings.

internal/component/database_observability/mysql/collector/query_samples.go
```diff
@@ -402,7 +430,27 @@ func (c *QuerySamples) fetchQuerySamples(ctx context.Context) error {
...
+			endEventID := row.WaitEndEventID.String
+			eventName := row.WaitEventName.String
+			objectName := row.WaitObjectName.String
+			objectType := row.WaitObjectType.String
+			waitTime := row.WaitTime.Float64
+
+			// wait/io/table/sql/handler is a molecule event wrapping the actual I/O;
+			// the SQL JOIN populates nested_waits.* only for that outer name, so a
+			// valid nested row means we should surface its fields instead.
+			// See https://dev.mysql.com/doc/refman/5.7/en/performance-schema-atom-molecule-events.html
```
cristiangreco:
```suggestion
			// See https://dev.mysql.com/doc/refman/8.0/en/performance-schema-atom-molecule-events.html
```

cristiangreco: Mainly because we don't support 5.7

cristiangreco (approved):

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -636,23 +637,50 @@ func (c *QuerySamples) buildWaitEventV2Labels(state *SampleState, we WaitEventOc
...
 		we.WaitEvent,
 		waitEventFullName,
 		we.BlockedByPIDs,
 		state.LastRow.QueryID.Int64,
 	)
 }
 
-// classifyPostgresWaitEventType maps a raw PostgreSQL wait event type to a standardized category.
-func classifyPostgresWaitEventType(rawType string) string {
+var postgresReplicationWaitEventPrefixes = []string{
```
matthewnolf, line 648: nit: we don't need to prefix things with postgres as we're in a postgres package already

matthewnolf: Same applies to below funcs

matthewnolf (approved):
Looks good!

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: I had it squashed so this wasn't addressed, but will address in a subsequent pr.

PR #6150, opened 2026-04-29
docs: Add information about the availability of windows exporter collectors being tied to the upstream collector.

clayton-cornell:
Minor linting cleanup

Add information about the availability of windows exporter collectors being tied to the upstream collector.

NOTE for the reviewer: This PR adds new text on lines 18-25, and then some minor cleanup throughout (table spacing, acronym vs word use, etc.).

Fixes: https://github.com/grafana/support-escalations/issues/17923


docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
```diff
@@ -381,9 +392,9 @@ For example, you can set `enabled_list` to `["ServerShares"]`.
...
 ### `smb_client`
 
-| Name           | Type           | Description                                      | Default | Required |
-| -------------- | -------------- | ------------------------------------------------ | ------- | -------- |
-| `enabled_list` | `list(string)` | Deprecated (no-op), a list of collectors to use. | `[]`    | no       |
+| Name           | Type           | Description                           | Default | Required |
+| -------------- | -------------- | ------------------------------------- | ------- | -------- |
+| `enabled_list` | `list(string)` | Deprecated. This field has no effect. | `[]`    | no       |
 
 The collectors specified by `enabled_list` can include the following:
```
dehaansa: If the field has no effect why are we telling users how to use it?

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
```diff
@@ -369,9 +380,9 @@ You can still include these attributes in your configuration files, but they hav
...
 ### `smb`
 
-| Name           | Type           | Description                                      | Default | Required |
-| -------------- | -------------- | ------------------------------------------------ | ------- | -------- |
-| `enabled_list` | `list(string)` | Deprecated (no-op), a list of collectors to use. | `[]`    | no       |
+| Name           | Type           | Description                           | Default | Required |
+| -------------- | -------------- | ------------------------------------- | ------- | -------- |
+| `enabled_list` | `list(string)` | Deprecated. This field has no effect. | `[]`    | no       |
 
 The collectors specified by `enabled_list` can include the following:
```
dehaansa: Same question

docs/sources/reference/components/prometheus/prometheus.exporter.windows.md
clayton-cornell: We have a few cases of no-op (30 cases across 17 components). It's all related to deprecated args that stay in the component, but if a user actually uses it, it has.. no-op... no effect. Last time I discussed this with someone on the team - I forget who though - the no-op args were left in place on purpose so things didn't break if someone left it in an old config.  I'm happy to do a sweep and remove them all if that's the right option... or we leave them.. either works.

clayton-cornell: Ah I misunderstood your question. I checked the source and it shows  SMB block config types are explicitly marked deprecated with Go deprecation comments in config.go line 189 and config.go line 195.

dehaansa: Right, I was looking to understand why on line 397 we say it has no effect, but then on 399 we tell users how to use it?

dehaansa (approved):

PR #6161, opened 2026-04-30
feat: Integration tests for loki.source.gcplog

x1unix:

Add integration tests for `loki.source.gcplog`

### Pull Request Details

Covered both HTTP push and Pub/Sub pull scenarios through relabeling into Loki.

The [gcloud pubsub emulator](https://github.com/marcelcorso/gcloud-pubsub-emulator) is used to cover the Pub/Sub pull scenario.

Related to https://github.com/grafana/alloy/issues/4953

### Notes to the Reviewer

Pub/Sub server address override is done using `PUBSUB_EMULATOR_HOST` env var, which is used by `cloud.google.com/go/pubsub/v2` package under the hood.

Also, I've chose `messagebird/gcloud-pubsub-emulator` over `gcr.io/google.com/cloudsdktool/cloud-sdk:emulators` because it's ~150 MB vs ~1 GB and provisions topics/subscriptions from a single env var on boot — no separate init container or `gcloud beta emulators pubsub` orchestration.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/integration-tests/docker/tests/loki-gcplog/docker-compose.yaml
+++ b/integration-tests/docker/tests/loki-gcplog/docker-compose.yaml
@@ -1,0 +1,47 @@
+services:
+  pubsub-emulator:
+    image: messagebird/gcloud-pubsub-emulator:latest
+    container_name: pubsub-emulator
+    environment:
+      # Pre-create the topic+subscription pair on boot so Alloy can start subscribing
+      # immediately and the test publisher has a topic to publish to.
+      PUBSUB_PROJECT1: "test-project,gcplog-topic:gcplog-subscription"
+    ports:
+      - "8681:8681"
+    healthcheck:
+      test: ["CMD-SHELL", "nc -z localhost 8681 || exit 1"]
+      interval: 1s
+      timeout: 3s
+      retries: 30
+      start_period: 10s
+
+  alloy:
+    image: alloy-integration-tests
+    depends_on:
+      pubsub-emulator:
+        condition: service_healthy
+    command:
+      - run
+      - /etc/alloy/config.alloy
+      - --server.http.listen-addr
+      - 0.0.0.0:12345
+      - --stability.level
+      - generally-available
+    environment:
+      PUBSUB_EMULATOR_HOST: pubsub-emulator:8681
+    volumes:
+      - ./config.alloy:/etc/alloy/config.alloy:ro
+    ports:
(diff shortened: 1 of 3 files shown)


integration-tests/docker/tests/loki-gcplog/loki_gcplog_test.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// publishViaEmulator publishes per-resource-type batches to the emulator topic.
+// Alloy's pull subscriber consumes them through the pre-created subscription.
+func publishViaEmulator(ctx context.Context, t *testing.T, shapes []resourceShape, perShape int) error {
+	t.Helper()
+
+	if err := os.Setenv(emulatorEnvVar, emulatorHost); err != nil {
+		return fmt.Errorf("set %s: %w", emulatorEnvVar, err)
+	}
```
jharvey10: Since we have `t` in scope, could this be replaced with `t.SetEnv()` so that it is automatically torn down when the test finishes?

integration-tests/docker/tests/loki-gcplog/docker-compose.yaml
```diff
@@ -0,0 +1,47 @@
+services:
+  pubsub-emulator:
+    image: messagebird/gcloud-pubsub-emulator:latest
```
jharvey10: We've gotten burned a few times where `latest` images change in a breaking way. Since Renovate will find and update these on its own, what do you think about pinning this to a specific hash instead? I realize it's diminishing returns here because there's only one tag on dockerhub for this image, but if it does ever change, our tests wouldn't break and Renovate will help us with tracking the update.

jharvey10 (commented):
Just a couple minor things for consideration!

integration-tests/docker/tests/loki-gcplog/docker-compose.yaml
x1unix: Thanks, addressed.

integration-tests/docker/tests/loki-gcplog/loki_gcplog_test.go
x1unix: Thanks, addressed.

jharvey10 (approved):
LGTM. Up to you if you'd like Kalle to take a look from a more detailed loki perspective.

PR #6168, opened 2026-05-01
fix(oracledb_exporter): Support CGO_ENABLED=0 cross-compilation

korniltsev-grafanista:
The oracledb integration imported github.com/oracle/oracle-db-appdev-monitoring/collector, which transitively pulls in github.com/godror/godror. godror requires CGO, so `cd collector && CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build .` failed with "undefined: VersionInfo" and similar errors from godror's orahlp.go.

The fix follows the existing loki.source.journal dual-file pattern: pure-Go config and helpers stay unconditional, the godror import and New() body move behind `//go:build cgo`, and a `//go:build !cgo` stub New() returns a clear error so configs surface a meaningful message in no-cgo builds.

To prevent regressions, two new CI jobs cross-compile the collector (in test_mac.yml) and the pyroscope playground (in test_pyroscope_pr.yml) for linux/amd64 and linux/arm64 with CGO_ENABLED=0 on macos-15-xlarge.

### DIFF
--- a/.github/workflows/test_mac.yml
+++ b/.github/workflows/test_mac.yml
@@ -30,3 +30,23 @@ jobs:
         cache: true
     - name: Test
       run: CGO_LDFLAGS="-ld_classic $CGO_LDFLAGS" make GO_TAGS="nodocker" test
+
+  cross-compile-collector-linux-nocgo:
+    name: Cross-compile collector for linux/${{ matrix.goarch }} with CGO disabled
+    strategy:
+      matrix:
+        platform: [macos-15-xlarge]
+        goarch: [amd64, arm64]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+    - name: Set up Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        go-version-file: go.mod
+        cache: true
+    - name: Build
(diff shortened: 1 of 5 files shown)


.github/workflows/test_mac.yml
```diff
@@ -30,3 +30,24 @@ jobs:
...
+        persist-credentials: false
+    - name: Set up Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        go-version-file: go.mod
+        cache: false
+    - run: cd collector && GOARCH=amd64 go build .
+    - run: cd collector && GOARCH=arm64 go build .
+    - run: GOARCH=amd64 go build ./internal/component/pyroscope/util/internal/cmd/playground
+    - run: GOARCH=arm64 go build ./internal/component/pyroscope/util/internal/cmd/playground
```
thampiotr: Would only one arch here be sufficient?

.github/workflows/test_mac.yml
```diff
@@ -30,3 +30,24 @@ jobs:
         cache: true
     - name: Test
       run: CGO_LDFLAGS="-ld_classic $CGO_LDFLAGS" make GO_TAGS="nodocker" test
+
+  cross-compile-mac-to-linux-nocgo:
+    name: Maintain developer-friendly mac-to-linux cross-compile
```
thampiotr: The name should be more descriptive like the other checks. Maybe "Verify mac-to-linux cross-compile" and leave a comment `# This is to maintain good developer experience on macOS`.

.github/workflows/test_mac.yml
korniltsev-grafanista: no

thampiotr (approved):

PR #6169, opened 2026-05-01
feat(prometheus.relabel): Add opt-in TTL cache mode

### LINKED ISSUE #5225: LRU cache thrashing on prometheus.relabel
### Component(s)

prometheus.relabel

### Background

Hello 🙂 

I noticed that the prometheus.relabel component is using an LRU cache with seriesRef as a key.

There is absolutely no issue as long as the cache is greater or equal to the cardinality of metrics scraped. Cache hit rate would then be at 100%.

What I'm worried about is that, since metrics scraping is cyclic (since they usually always appear in the same order in /metrics endpoints), the LRU cache may drop to 0% hit rate as soon as the cardinality of metrics is greater than the cache capacity. 

For example, if you have a cache size of 3, and 4 metrics that are scraped and relabeled, the LRU cache will evict the oldest metric first, which happen to be the very next to process.

It would look like this:
* |A|B|C| // cache is filled with A,B,C and we process D
* |D|B|C| // D was not in cache, we evict A as it is oldest and put D in its place. Next metrics to process is A
* |D|A|C| // A was not in cache, we evict B as it is oldest and put A in its place. Next metrics to process is B
* |D|A|B| // B was not in cache, we evict C as it is oldest and put B in its place. Next metrics to process is C
* etc.

To convince myself further, I tried to run the following tests in https://github.com/grafana/alloy/blob/main/internal/component/prometheus/relabel/relabel_test.go#L88

```go
func TestLRUNoThrashing(t *testing.T) {
	// cache size is set at 100_000
	relabeller := generateRelabel(t)

	for range 2 { // 2 pass to measure cache hits
		for i := 0; i < 100_000; i++ { // exactly cache size, no thrashing expected
			lbls := labels.FromStrings("__address__", "localhost", "inc", strconv.Itoa(i))
			relabeller.relabel(storage.SeriesRef(i), 0, lbls)
		}
	}
	require.Equal(t, 100_000, relabeller.cache.Len())                   // cache is full
	assert.Equal(t, 100_000., getCounterValue(t, relabeller.cacheHits)) // test passes, hit rate is 100% for the second pass
}

func TestLRUThrashing(t *testing.T) {
	relabeller := generateRelabel(t)

	for range 2 {
		for i := 0; i < 100_001; i++ { // 1 more than cache size to
			lbls := labels.FromStrings("__address__", "localhost", "inc", strconv.Itoa(i))
			relabeller.relabel(storage.SeriesRef(i), 0, lbls)
		}
	}
	require.Equal(t, 100_000, relabeller.cache.Len())                // cache is full
	assert.Greater(t, 0.0, getCounterValue(t, relabeller.cacheHits)) // test fails, hit rate is 0%
}

// just a helper to read cache counters values
func getCounterValue(t *testing.T, metric prom.Counter) float64 {
	var m = &dto.Metric{}
	err := metric.Write(m)
	require.NoError(t, err)
	return m.Counter.GetValue()
}
```

I do realize though that this might be a bit naive, since there are a lot of parameters in a real life scenario that can render metrics scraping acyclic. And I have no idea how other processings, such as batching, might affect this behavior. 

### Proposal

Switching to an MRU (most recently used) cache seems to be the obvious solution, but again, since I have no idea how other processings work, this might not be the best suited for every use case.

There doesn't seem to be an obvious pick for an MRU lib  in the golang ecosystem though. This is first result on google : https://github.com/Code-Hex/go-generics-cache

Thank you for your work, I hope I was clear enough 🙂 

### Tip

kgeckhart:

Adds a TTL cache mode to `prometheus.relabel` so the cache can size itself to the working set of series, instead of an operator-guessed `max_cache_size`.

### Pull Request Details

The component currently uses a fixed-size LRU cache. When scraped cardinality exceeds `max_cache_size`, the LRU degenerates to ~0% hit rate — series arrive in a stable cyclic order, so the entry the LRU just evicted is often the next one requested.

| `max_cache_size` | `cache_ttl` | Mode |
|---|---|---|
| `> 0` | `0` | LRU (today's default, unchanged) |
| `0` | `> 0` | TTL — entries expire after `cache_ttl` of inactivity, no size cap |

`max_cache_size` and `cache_ttl` are mutually exclusive (validation error). `cache_ttl` minimum is `1m`. TTL semantics are sliding: each cache hit pushes the entry's expiry forward, so active series stay cached. Default behavior is unchanged; flipping the default to TTL mode is deferred to a follow-up once the opt-in has production exposure.

Steady-state hot-path benchmark (`BenchmarkCacheModes`, M3 Pro, single label set):

| Mode | ns/op | B/op | allocs/op |
|---|---|---|---|
| lru | 192 | 284 | 1 |
| ttl | 203 | 269 | 1 |

Fixes #5225

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/relabel/cache.go
+++ b/internal/component/prometheus/relabel/cache.go
@@ -2,6 +2,7 @@ package relabel
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2"
@@ -9,16 +10,9 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 )
 
-// relabelCache abstracts the per-component cache of relabeled labels keyed
-// by the input labels' hash. Implementations cover two modes selectable
-// via Arguments: bounded LRU (default) and time-bounded ("size dictated
-// by working set"). All methods are safe for concurrent use.
-//
-// Implementations own any background work (e.g. periodic eviction).
-// `newRelabelCache` returns a ready-to-use cache; callers must invoke
-// Close when the cache is no longer needed. Tests that want to drive
-// eviction manually can use the per-impl bare constructors (e.g.
-// `newTTLRelabelCache`) which don't spawn the background goroutine.
+// relabelCache is the per-component cache of relabeled labels keyed by
+// the input labels' hash. Implementations are safe for concurrent use
+// and may own background goroutines; callers must Close when done.
 type relabelCache interface {
 	// Get returns the cached relabel result for the given hash. The
 	// boolean is false if no entry is cached (or, in TTL mode, if the
@@ -35,11 +29,8 @@ type relabelCache interface {
 	Close()
 }
 
-// newRelabelCache constructs a cache impl based on Arguments and
-// readies it for use (TTL caches have their background scan goroutine
-// running on return). Validation has already ensured exactly one of
-// CacheSize and CacheTTL is non-zero, so the lru.New error here is
-// effectively unreachable in production but propagated for safety.
+// newRelabelCache constructs a cache from args. TTL caches return with
+// their scan goroutine already running; the caller owns Close.
 func newRelabelCache(args Arguments, evictions prometheus_client.Counter) (relabelCache, error) {
 	if args.CacheTTL > 0 {
 		c := newTTLRelabelCache(args.CacheTTL, evictions)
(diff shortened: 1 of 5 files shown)


internal/component/prometheus/relabel/cache.go
```diff
@@ -0,0 +1,206 @@
...
+	lru "github.com/hashicorp/golang-lru/v2"
+	prometheus_client "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/model/labels"
+	"go.uber.org/atomic"
+)
+
+// relabelCache is the per-component cache of relabeled labels keyed by
+// the input labels' hash. Implementations are safe for concurrent use
+// and may own background goroutines; callers must Close when done.
+type relabelCache interface {
```
x1unix, line 16: Private interface with public method names?

x1unix (dismissed):

x1unix (dismissed):

internal/component/prometheus/relabel/cache.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Phase 2: prune. Re-check expiry per key in case a concurrent Get
+	// slid the entry forward (or Add refreshed it) between the phases.
+	for _, h := range expired {
+		if e, ok := c.entries[h]; ok && nowSec > e.expires.Load() {
+			delete(c.entries, h)
```
kalleep, line 226: I don't now if this is going to be a problem but deleting from a map will not reclaim any memory. So if you for some reason have a period with a lot of relabels being cached and then later it goes down the memory for the cache would be the same. 

We could perform compaction at some point, e.g. if we shrink the cache by some threshold (just create a new map and copy over everything)

kgeckhart: I should have left a comment about this. I originally did it and walked it back to see if we need it mostly because when we reach a size where it's worth it to compact we end up with two rather large maps sitting around for a short time. 

Perhaps a shrink when it's an order of magnitude smaller makes sense? Like 1 million down to 100k?

kgeckhart: WDYT about https://github.com/grafana/alloy/pull/6169/commits/23e880ab997b0b4dec7ad92ca6b43c6fd1d6f6b8?

kalleep (approved):
LGTM

internal/component/prometheus/relabel/cache.go
kalleep: I like it, great!

Issue #6173, opened 2026-05-01
Proposal: Revisit alloyengine module structure


kgeckhart:
## Background

[Design doc 4719](https://github.com/grafana/alloy/blob/main/docs/design/4719-otel-engine-dependency-management.md) established the three-module structure (alloy, `extension/alloyengine`, `collector`) and a `dependency-replacements.yaml` DSL that fans replace directives across them. Maintaining three `go.mod` files in sync plus the DSL has measurable ongoing overhead, which #6131 currently proposes to automate around.

## Question

Does `extension/alloyengine` need its own `go.mod` at all, given that it is intrinsically coupled to the Alloy runtime via `flowcmd`?

## Proposal

Design PR will be linked when opened.

Related: #6131

thampiotr:
I believe this is closed by? https://github.com/grafana/alloy/pull/6258 
If not, feel free to reopen

PR #6131, opened 2026-04-24
ci(release-please): Track syntax as its own package and auto-regenerate on release PRs

kgeckhart:
### Brief description of Pull Request

Two changes that work together:

1. **Track `syntax` as its own release-please package** so it gets independently semver-versioned and tagged. External consumers (cockroach, mattermost, alloy2otel, several grafana-internal projects) need a versioning line that reflects the syntax library's own API stability — not alloy's release cadence.

2. **Auto-regenerate `module dependencies` and `collector distribution` on every release-please PR**, eliminating the manual fixup step where a human had to check out the release-please branch and run `make generate-*` to satisfy the `check-*` workflows. Same step also closes the pre-existing drift on `extension/alloyengine`'s alloy require and `collector/builder-config.yaml`'s alloyengine gomod entry by marking them so release-please bumps them on each release.

### Issue(s) fixed by this Pull Request

Part of #5823

### Notes to the Reviewer

After merge, release-please's next run on `main` should detect commits touching `syntax/` since `syntax/v0.1.0` (the typecheck additions, etc.) and cut a `syntax/v0.2.0` release PR automatically — no manual tagging required. If for some reason it doesn't (e.g. no conventional-commit-formatted commits in that path), we can either manually tag or land a `feat(syntax):` commit to nudge it.

Going forward, contributors making syntax-only changes should scope commits as `feat(syntax):` / `fix(syntax):`. Path-based detection covers the case where they don't, but the scope makes the changelog cleaner.

The auto-regen step runs with `alloybot` write credentials and only acts when a `release-please--branches--<target>` branch exists.

PR #6174, opened 2026-05-01
docs: Design doc for revisiting alloyengine's module structure

kgeckhart:

Adds a mini design doc for proposal #6173, revisiting the module structure decision from [design doc 4719](https://github.com/grafana/alloy/blob/main/docs/design/4719-otel-engine-dependency-management.md). Three options analyzed: 

(0) automate version-bump fan-out via release-please (the current path in #6131)
(1) tooling simplification keeping the extension go.mod
(2) eliminate the extension go.mod and embed alloyengine via OCB's `gomod + import + name` form.

This came about originally from Joe as we were trying to plug all the holes to keep versions consistent to support OCB'ing and he suggested "do we even need the module at all". At that point we both proved out that it would work without the module and that there is a middle ground to keep it but simplify our tooling. 

### Notes to the Reviewer

Proposal 2 verified end-to-end. POC branch: [`kgeckhart/verify-extension-no-gomod`](https://github.com/grafana/alloy/tree/kgeckhart/verify-extension-no-gomod) — six-file commit, +5/−4535 lines.

### DIFF
--- a/docs/design/6173-alloyengine-module-structure.md
+++ b/docs/design/6173-alloyengine-module-structure.md
@@ -1,0 +1,130 @@
+# Mini design: Revisiting Module Structure for Alloy Engine Extension
+
+#### Authors: [Kyle Eckhart](https://github.com/kgeckhart), [Joe Harvey](https://github.com/jharvey10), Claude (AI assistant)
+
+* Proposal issue: https://github.com/grafana/alloy/issues/6173
+
+# Context
+
+[Design doc 4719](./4719-otel-engine-dependency-management.md) introduced the three-module structure (Alloy, `extension/alloyengine`, `collector`) and chose Proposal 1: a `dependency-replacements.yaml` DSL that fans `replace` directives into all three targets. That tooling has been operating since.
+
+The pain points are now visible:
+
+- Three `go.mod` files plus `builder-config.yaml` must stay in sync. Every release bumps versions in three places.
+- Component versions (e.g. otel `v0.147.0`) are still hand-coordinated between root `go.mod` and `builder-config.yaml`.
+- The YAML DSL is a less-natural spelling of go.mod's native `replace` syntax. It exists *because* the same data must live in multiple modules.
+- [PR #6131](https://github.com/grafana/alloy/pull/6131) is open to automate the multi-target version bumps via release-please. This automates around the structural cost rather than removing it.
+
+A reviewer on #6131 raised the question: does `extension/alloyengine` need its own `go.mod` at all? The extension imports `flowcmd`, which transitively pulls in the entire Alloy component registry. Any consumer of the extension already depends on all of Alloy. The version coupling is intrinsic — a separate `go.mod` makes the form look independent while the substance is fully coupled.
+
+# Goals
+
+- Reduce ongoing maintenance overhead of the multi-module structure
+- Remove the cause of the dependency-sync tooling, not just automate around it
+- Keep external OCB consumers able to embed the extension with a documented and reasonable workflow
+
+# Non-Goals
+
+- Changing `syntax/` (has external consumers; its `go.mod` is non-negotiable)
+- Changing `collector/` (it is the OCB-generated build target)
+- Solving component-version sync between root `go.mod` and `builder-config.yaml` — separable problem
+- Moving the extension to a separate repo
+
+# Proposal 0 — Do Nothing
+
+Land [PR #6131](https://github.com/grafana/alloy/pull/6131) to automate version bumps via release-please. Keep the three-module structure and the YAML DSL.
+
+```mermaid
+graph LR
+  DR[dependency-replacements.yaml] -->|generator| RG[go.mod]
+  DR -->|generator| EG[extension/alloyengine/go.mod]
+  DR -->|generator| BC[builder-config.yaml]
+  BC -->|OCB| CG[collector/go.mod]
+```
+
+**Pros**
+- Smallest scope; PR already drafted
+- Zero user-facing change; preserves OCB convention
+
+**Cons**
+- Automates around complexity rather than removing it
+- Three modules + DSL + multi-target fan-out remain forever
+- Component-version hand-coordination unchanged
+
+# Proposal 1 — Tooling Simplification (Keep extension go.mod)
+
+Make `builder-config.yaml` canonical for `replace` directives. Replace `generate-module-dependencies` with a thin propagator that injects replaces into root `go.mod` and `extension/alloyengine/go.mod` between BEGIN/END markers. Delete `dependency-replacements.yaml`. Fold into `generate-otel-collector-distro`.
+
+`builder-config.yaml` already owns OCB-shape decisions (component list, versions, dist info). Adding `replace` directives to that ownership extends an existing boundary.
+
+```mermaid
+graph LR
+  BC[builder-config.yaml] -->|propagator| RG[go.mod]
+  BC -->|propagator| EG[extension/alloyengine/go.mod]
+  BC -->|OCB| CG[collector/go.mod]
+```
+
+**Pros**
+- Drops the YAML DSL; tool shrinks from 341 LOC to ~150
+- Coherent canonical boundary in `builder-config.yaml`
+- Replaces sit next to the component versions they constrain
+- Zero user-facing change; OCB convention preserved
+- Sets up future canonicalization of component versions
+- Reversible — Proposal 2 reachable as a follow-up with no rework
+
+**Cons**
+- Three modules and ~5000 lines of `extension/alloyengine/go.mod`+`go.sum` still remain
+- Two propagation targets still need marker injection
+- Doesn't fix the structural cost driver
+
+# Proposal 2 — Eliminate the Extension go.mod (Recommended)
+
+Make `extension/alloyengine/` a sub-package of Alloy. Delete its `go.mod`/`go.sum`. Reference it in `builder-config.yaml` via OCB's `gomod + import + name` form:
+
+```yaml
+extensions:
+  - gomod: github.com/grafana/alloy v1.16.0
+    import: github.com/grafana/alloy/extension/alloyengine
+    name: alloyengine
+```
+
+Combined with Proposal 1's tooling: `builder-config.yaml` canonical for replaces, propagator only fans out to root `go.mod`.
+
+```mermaid
+graph LR
+  BC[builder-config.yaml] -->|propagator| RG[go.mod]
+  BC -->|OCB| CG[collector/go.mod]
+```
+
+**Pros**
+- ~5200 lines deleted (DSL + extension go.mod + go.sum)
+- Two modules (alloy + collector), not three
+- One propagation target instead of two
+- Tool shrinks to ~80 LOC
+- Removes the structural cause, not just the symptom
+- Aligns module structure with actual coupling — there is no realistic independent-versioning use case for the extension
+
+**Cons**
+- External OCB consumers must use `gomod + import + name` for this one component instead of the standard one-line form
+- First-time copy-paste of the standard form fails with an opaque "module not found" error before the user discovers `import:`
+- `go get github.com/grafana/alloy/extension/alloyengine@vX` no longer works
+- No standalone pkg.go.dev page for the extension
+- Reversing requires recreating the go.mod and propagation target
+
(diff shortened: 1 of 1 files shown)


docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Compatibility:** No external migration concern — `extension/alloyengine` has not been published as a usable release and has no external consumers yet. Documentation in `extension/alloyengine/README.md` should describe the new form from the outset, alongside the `cgo_enabled: true` requirement and the syntax-replace requirement (both true regardless of this proposal).
+
+# Verification
+
+Proposal 2 was verified end-to-end before authoring. The POC lives on branch [`kgeckhart/verify-extension-no-gomod`](https://github.com/grafana/alloy/tree/kgeckhart/verify-extension-no-gomod). With those changes applied: `make generate-module-dependencies` and OCB regen succeed; `go build ./...` produces a working Alloy binary with the `otel` subcommand intact. A synthetic external distro using the `gomod + import + name` form, with path-replaces to the worktree and full fork-replace block, also builds cleanly under OCB compilation and registers `alloyengine` at `module: github.com/grafana/alloy v1.16.0`. Pseudo-version resolution from the public Go module proxy is not yet exercised; will be tested when the implementation PR pushes a branch.
```
thampiotr, line 118: Isn't it using a fixed, published v1.16.0 of Alloy, instead of the Alloy in the local repo checkout because the `replace` with a local path is removed? 

Image

Before proceeding I'd like to verify that it can work with the actual current repo's Alloy version.

docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Authors: [Kyle Eckhart](https://github.com/kgeckhart), [Joe Harvey](https://github.com/jharvey10), Claude (AI assistant)
+
+* Proposal issue: https://github.com/grafana/alloy/issues/6173
+
+# Context
+
+[Design doc 4719](./4719-otel-engine-dependency-management.md) introduced the three-module structure (Alloy, `extension/alloyengine`, `collector`) and chose Proposal 1: a `dependency-replacements.yaml` DSL that fans `replace` directives into all three targets. That tooling has been operating since.
+
+The pain points are now visible:
```
thampiotr, line 11: The reason this was set up this way is to follow the typical conventions of how OTel components and extensions are structured: each one has its own go.mod. 

The path of automating the version updates with tooling is what upstream uses, [for example this](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48002/changes#diff-9f94c82086362a3f578d6495d951a32c4316a086ca9298a4a66ef56781b1bfb6).

I'm open to find ways that we can do what we need to in a more convenient way, but we have to keep in mind that diverging too far away from the upstream OTel can result in integration pains in other situation.

docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Proposal 2 — Eliminate the Extension go.mod (Recommended)
+
+Make `extension/alloyengine/` a sub-package of Alloy. Delete its `go.mod`/`go.sum`. Reference it in `builder-config.yaml` via OCB's `gomod + import + name` form:
+
+```yaml
+extensions:
+  - gomod: github.com/grafana/alloy v1.16.0
+    import: github.com/grafana/alloy/extension/alloyengine
+    name: alloyengine
```
thampiotr, line 88: Didn't know this was possible! Is this somehow new? nice :)

docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+**Pros**
+- Drops the YAML DSL; tool shrinks from 341 LOC to ~150
+- Coherent canonical boundary in `builder-config.yaml`
+- Replaces sit next to the component versions they constrain
+- Zero user-facing change; OCB convention preserved
+- Sets up future canonicalization of component versions
+- Reversible — Proposal 2 reachable as a follow-up with no rework
+
+**Cons**
+- Three modules and ~5000 lines of `extension/alloyengine/go.mod`+`go.sum` still remain
```
thampiotr, line 76: Counting `go.sum` is a bit of a hack 😛

docs/design/6173-alloyengine-module-structure.md
jharvey10: IIRC this was [modified from the default module version in this commit](https://github.com/grafana/alloy/commit/af6327535f4bc117a5ca05ca6387d22727756d51) and doesn't actually work properly, which is what drove a lot of this discussion. 

There's a couple parts to clarify:

1. From inside Alloy, this is has no impact because of the replace directive.
2. For external consumers it is broken as is, because the alloyengine extension depends on the Alloy module, which depends on the Syntax module, which is not properly tagged/published.
3. This will not remove the need to run `make generate-otel-collector-distro` (or equivalent), which will keep that Alloy version up to date across releases. This is more-or-less what is implied [here](https://github.com/grafana/alloy/pull/6174/changes#diff-bfcf45237bc6963d16db2c3748ae42a17cd85fa607e1b37fc1924ee268574143R56)

docs/design/6173-alloyengine-module-structure.md
jharvey10: We plan to maintain this paradigm of `one extension == one go.mod`, and this proposal is no exception.

Right now, the `alloyengine` extension is effectively managed by *two* `go.mod` files with almost identical content. The `alloyengine` `go.mod` file being removed does not result in bundling multiple extensions together; it just simplifies the graph for us by acknowledging that the `alloyengine` extension's `go.mod` is essentially an exact copy of the top-level `go.mod` file (with direct dependencies moved to indirect ones).

Under this proposal, the `alloyengine` extension will be managed by a single `go.mod` file. It just happens to be the same one that core Alloy uses as well.

Additional extensions will *not* use the same `go.mod` file; they will be self-isolated.

docs/design/6173-alloyengine-module-structure.md
thampiotr: I think I misread the repalces. Thanks for double-checking!

docs/design/6173-alloyengine-module-structure.md
thampiotr: Sounds good, thanks for the extra context!

thampiotr (approved):
+1 to Proposal 2

docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+
+* Proposal issue: https://github.com/grafana/alloy/issues/6173
+
+# Context
+
+[Design doc 4719](./4719-otel-engine-dependency-management.md) introduced the three-module structure (Alloy, `extension/alloyengine`, `collector`) and chose Proposal 1: a `dependency-replacements.yaml` DSL that fans `replace` directives into all three targets. That tooling has been operating since.
+
+The pain points are now visible:
+
+- Three `go.mod` files plus `builder-config.yaml` must stay in sync. Every release bumps versions in three places.
```
thampiotr, line 13: I think even with Proposal 2 we will still have several go.mod files, especially if we have more OTel components in Alloy repo. So we will need updating them when we update dependencies and this pain point is not addressed? Are we okay with that and can perhaps release please help us automate these updates?

docs/design/6173-alloyengine-module-structure.md
jharvey10: haha true. I think a more worthwhile stat here might be how many conflicts we've had against these files ;)

docs/design/6173-alloyengine-module-structure.md
jharvey10: We'll have a bit of a proof of concept with this with the syntax module. The rough plan is to manage each with a release-please sub module and use github PRs sort of like we do with Alloy now. Tracking changes will be driven by affected files in specific directories. I could also see us ending up with a single "release" PR each minor that encompasses base alloy and all the extensions we ship.

jharvey10 (approved):

docs/design/6173-alloyengine-module-structure.md
kgeckhart: I certainly didn't either until AI pointed it out. The fact that it's not widely known was a bit of a concern since we'll need docs + guides to be crystal clear about it.

blewis12: same here, It's mentioned in the [github](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder#configuration) docs which is linked from the [web docs](https://opentelemetry.io/docs/collector/extend/ocb/) though - which is something we can point to in our own documentation in case that's confusing for users

docs/design/6173-alloyengine-module-structure.md
```diff
@@ -0,0 +1,130 @@
...
+- Zero user-facing change; preserves OCB convention
+
+**Cons**
+- Automates around complexity rather than removing it
+- Three modules + DSL + multi-target fan-out remain forever
+- Component-version hand-coordination unchanged
+
+# Proposal 1 — Tooling Simplification (Keep extension go.mod)
+
+Make `builder-config.yaml` canonical for `replace` directives. Replace `generate-module-dependencies` with a thin propagator that injects replaces into root `go.mod` and `extension/alloyengine/go.mod` between BEGIN/END markers. Delete `dependency-replacements.yaml`. Fold into `generate-otel-collector-distro`.
```
blewis12, line 56: > Replace `generate-module-dependencies` with a thin propagator that injects replaces into root `go.mod` and `extension/alloyengine/go.mod` between BEGIN/END markers

More of an implementation Q: will the final destination `go.mod` files that are to be modified by the wrapper around `builder-config.yaml` be hard coded or stored in some config file? I remember having clarity on which files exactly were going to be modified was part of the appeal of a centralized YAML file that stores both the replacements to be propagated as well as the go.mod destinations.

Another part of the idea behind having a generic `dependency-replacements.yaml` means that we can extend it, and add for example an `exclude` statement that would be propagated. Implementing purely from the `builder-config.yaml` means we are bound only to the structure of that file which currently only contains `replace` directives.  I can actually see we have an [example](https://github.com/grafana/alloy/blob/a3db6110931e62e58135f3fbc654e0ce3c0170b3/go.mod#L1107) of this already not being propagated. We can probably remove that since it's not causing issues (😄) but more something that could pop up in the future.

blewis12 (approved):
Thanks for putting this together! I think proposal 2 makes sense and simplifies our dependency management process quite some

docs/design/6173-alloyengine-module-structure.md
jharvey10: Good question for sure. The intent here is to manage it all from the `builder-config.yml` file as opposed to our own domain-specific language as much as we can.

Regarding the extension aspect and the exclude stuff, is this something where we could extend the OCB tool to accomplish what we want and then push for these sorts of enhancements upstream at the same time?

IIRC we bundle the builder in the alloy repo, so I'm wondering if the more upstream-friendly path is to simply extend that tooling if we need additional functionality.

But I'm curious of your thoughts on that having a lot more experience with how these things are wired together.

blewis12: Ah yeah so we don't actually embed the OCB tool, we just do a [go run](https://github.com/grafana/alloy/blob/57de4f4aab2c2cc4602f1c63a8f9989f0ed1143e/collector/generate.go#L3) which is sort like running `npx run` - it just grabs the upstream library and runs it. We _could_ bundle in the OCB tool, it's [not huge](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder) but it would be additional maintenance for us and more code in Alloy. Given that exclude statements are relatively uncommon, maybe it's worth just manually ensuring those are propagated for now. But if they become more common or an issue we could always look into extending the OCB tool, or just proposing the changes upstream if it's not a time rush.

I think the way proposal 2 is laid out is totally fine for now! But just something worth noting for the future in case it comes up

blewis12: hah, disregard about the `exclude` statement - it's not documented anywhere I can see but it's in the [code](https://github.com/open-telemetry/opentelemetry-collector/blob/6aeee1722dd6bc8aab2e32dc614bdf01b045dae6/cmd/builder/internal/builder/config.go#L54) 🤦

PR #6180, opened 2026-05-04
feat(database_observability.postgres): Emit `op=error_message` Loki entries with query fingerprint

gaantunes:
## Summary

Adds an `op="error"` Loki entry on `database_observability.postgres` that pairs each PostgreSQL `ERROR`/`FATAL`/`PANIC` log line with its `STATEMENT:` continuation and emits a stable `query_fingerprint`. This lets consumers compute per-query error rate without `pg_stat_statements.queryid` in `log_line_prefix` — which RDS/Aurora don't expose.

Also introduces the freestanding `internal/component/database_observability/postgres/fingerprint` package — a 16-hex hash of the parsed SQL AST via `libpg_query`, invariant across comments/whitespace/literals — which #6297 builds on.

## Configuration

```alloy
database_observability.postgres "x" {
  ...
  logs_processing {
    enable_error_logs = true
  }
}
```

Defaults to `false`. With the flag off, behavior is unchanged: the existing `database_observability_pg_errors_total` counter still increments and no `op="error"` entries are emitted.

## Entry shape

```
op="error"
body: level="info" severity=ERROR datname=books_store query_fingerprint=<16-hex>
```

The SQL text is not emitted — consumers recover it by joining on `query_fingerprint` (see #6297's `op="query_association_v2"`).

## Scope

Rate-only, intentionally minimal. A single in-flight pending error pairs each ERROR with its following STATEMENT (PostgreSQL's logging collector writes each message atomically and contiguously). Error-detail fields (sqlstate, pid, client/session, application_name, error_message) are **deferred to a follow-up PR**.

## Test plan

- [x] `go test -race -tags=nodocker ./internal/component/database_observability/postgres/...` — green
- [x] `golangci-lint` — clean
- [x] `CGO_ENABLED=0` cross-compile (the `!cgo` fingerprint stub)
- [x] Covered: happy path, multi-line STATEMENT, displaced pending, no-STATEMENT timeout, flag-off / default-off


docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -27,16 +27,16 @@ database_observability.postgres "<LABEL>" {
...
-| `targets`            | `list(map(string))`  | List of external targets to scrape for Prometheus metrics.  |         | no       |
-| `disable_collectors` | `list(string)`       | A list of collectors to disable from the default set.       |         | no       |
-| `enable_collectors`  | `list(string)`       | A list of collectors to enable on top of the default set.   |         | no       |
-| `exclude_databases`  | `list(string)`       | A list of databases to exclude from monitoring.             | `["alloydbadmin", "alloydbmetadata", "azure_maintenance", "azure_sys", "cloudsqladmin", "rdsadmin"]` | no       |
-| `exclude_users`      | `list(string)`       | A list of users to exclude from monitoring.                 | `["azuresu", "cloudsqladmin", "db-o11y", "rdsadmin"]` | no       |
-| `exclude_current_user` | `bool`             | Exclude from activity monitoring the user that Alloy uses to connect to the database. The resolved username is automatically added to `exclude_users`, if not already present. | `true` | no       |
+| Name                       | Type                 | Description                                                 | Default | Required |
```
clayton-cornell: Are there any changes here other than adding spaces to the table?

gaantunes: No not really. i will drop these changes.

docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -167,6 +169,12 @@ The `cache_enabled`, `cache_size`, and `cache_ttl` settings are deprecated: they
 |--------------------|------------|------------------------------------------------------|---------|----------|
 | `collect_interval` | `duration` | How frequently to collect information from database. | `"1h"`  | no       |
 
+### `logs_processing`
+
+| Name                | Type   | Description                                                                                                                  | Default | Required |
+|---------------------|--------|------------------------------------------------------------------------------------------------------------------------------|---------|----------|
+| `enable_error_logs` | `bool` | Emit `op="error"` Loki entries that pair PostgreSQL `ERROR`/`FATAL`/`PANIC` log lines with their `STATEMENT:` continuations. | `false` | no       |
+
```
gaantunes: @clayton-cornell I need to keep this config option hidden (not documented), since this new feature will need more testing within our dev and ops clusters. If I just delete this doc from the PR, the Ci will fail right? How should I proceeed?

clayton-cornell: I don't see anything that would cause the PR CI/CD to fail if you remove the doc from the PR. It's stand-alone and nothing directly references it.  The `check_docs.yml` is triggered if you touch a file in `/docs/sources`... but it's not a check that looks for doc needs in a code change (as far as I know).

If this PR isn't backported, it'll land in `next` when it's merged... so any doc changes will only show up once the next release is cut.

internal/component/database_observability/postgres/component.go
```diff
@@ -72,13 +72,18 @@ type Arguments struct {
...
-	ExplainPlansArguments  ExplainPlansArguments        `alloy:"explain_plans,block,optional"`
-	HealthCheckArguments   HealthCheckArguments         `alloy:"health_check,block,optional"`
-	PrometheusExporter     *PrometheusExporterArguments `alloy:"prometheus_exporter,block,optional"`
+	CloudProvider           *CloudProvider               `alloy:"cloud_provider,block,optional"`
+	QuerySampleArguments    QuerySampleArguments         `alloy:"query_samples,block,optional"`
+	QueryDetailsArguments   QueryDetailsArguments        `alloy:"query_details,block,optional"`
+	SchemaDetailsArguments  SchemaDetailsArguments       `alloy:"schema_details,block,optional"`
+	ExplainPlansArguments   ExplainPlansArguments        `alloy:"explain_plans,block,optional"`
+	HealthCheckArguments    HealthCheckArguments         `alloy:"health_check,block,optional"`
+	LogsProcessingArguments LogsProcessingArguments      `alloy:"logs_processing,block,optional"`
```
cristiangreco: Could we call this block just `logs` perhaps? (like the name of the collector)

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -78,24 +99,34 @@ type Logs struct {
 	validLogsThisMinute   int
 	invalidLogsThisMinute int
 
+	// op="error" pairing state. PostgreSQL's logging collector writes each
+	// message (the ERROR line, its STATEMENT, and continuations) atomically and
+	// contiguously per backend, so a single in-flight pending is sufficient —
+	// no PID-keyed map is needed.
+	pending             *pendingError
+	pendingMu           sync.Mutex
```
cristiangreco: Is this mutex useful now or is it a relic from refactoring?

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -15,8 +15,13 @@ import (
 	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/component/database_observability"
+	"github.com/grafana/alloy/internal/component/database_observability/postgres/fingerprint"
+	"github.com/grafana/alloy/internal/runtime/logging"
 )
 
+const OP_ERROR = "error"
```
cristiangreco: Might consider calling this `error_line` or `error_message`, i.e. something that is not confused with a regular app error statement.

internal/component/database_observability/postgres/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,98 @@
...
+
+var (
+	sentinelTruncatedFp  = FingerprintOf(SentinelTruncated)
+	sentinelUnparsableFp = FingerprintOf(SentinelUnparsable)
+)
+
+// Fingerprint parses query and returns its fingerprint, falling back to a
+// quote/paren repair pass and then to a sentinel hash. trackActivityQuerySize
+// is consulted only when source == SourcePgStatActivity.
+func Fingerprint(query string, source Source, trackActivityQuerySize int) (fp string, repaired bool, err error) {
```
cristiangreco: Seems that in this PR and in https://github.com/grafana/alloy/pull/6297 we use sources `SourceLog` and `SourcePgStatStatements`. The `SourcePgStatActivity` is never used.

If this is confirmed, it might be unnecessary to pass `trackActivityQuerySize` around, correct? And if this is the case, we only need the sentinel `sentinelUnparsableFp` (i.e. `sentinelTruncatedFp` would not be needed as `pg_stat_statements` doesn't truncate the text).

gaantunes: You are right, this is dead code from when I was planning on shifting the pg query key to be the fingerprint. Will remove this and we can add it back if needed afterwards.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: I agree, lets go with error_message, since an error is composed of multiple log lines.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: A relic, good catch.

internal/component/database_observability/postgres/component.go
gaantunes: Yes, makes sense.

cristiangreco (approved):

PR #6195, opened 2026-05-06
feat(faro.receiver): Support gzip-compressed request bodies

d0ugal:
## Summary

- Add gzip decompression support to `faro.receiver`'s HTTP handler so it can accept `Content-Encoding: gzip` request bodies
- Add `content-encoding` to the CORS `defaultAllowedHeaders` so browser preflight requests are not rejected

## Context

The [Faro web SDK is adding optional gzip request compression](https://github.com/grafana/faro-web-sdk/pull/2028) to `FetchTransport` using the browser's native `CompressionStream` API. This compresses the entire JSON batch before sending, reducing payload sizes significantly with zero bundle cost.

The `otelcol.receiver.faro` component already handles this transparently via the OTel collector's `httpContentDecompressor` middleware. However, the native `faro.receiver` component uses a plain `http.Server` with no decompression — `json.NewDecoder(req.Body)` reads raw gzip bytes and fails with a parse error.

### Changes

**CORS** (`handler.go:22`): Added `content-encoding` to `defaultAllowedHeaders`. Without this, cross-origin requests with `Content-Encoding: gzip` trigger a preflight that gets rejected because the header isn't in the allow list.

**Decompression** (`handler.go:120-131`): Check `Content-Encoding: gzip` and wrap `req.Body` with `gzip.NewReader` before JSON decoding. This follows the same pattern used by `loki.source.aws_firehose` and `loki.source.api` in this codebase.

## Test plan

- [x] Gzip-compressed payload is properly decoded (`TestGzipCompressedPayload`)
- [x] Invalid gzip body returns 400 (`TestGzipInvalidBody`)
- [x] Uncompressed payloads continue to work (`TestUncompressedPayloadStillWorks`)
- [x] CORS preflight with `content-encoding` header succeeds (`TestCORSPreflightWithContentEncoding`)
- [x] All existing tests pass


kalleep (commented):
LGTM but we should mention in documentation that gzip is supported

d0ugal:
> LGTM but we should mention in documentation that gzip is supported

Thanks @kalleep. I updated with a note about gzip.

kalleep (approved):

PR #6196, opened 2026-05-06
ci: Cross-compile darwin/amd64 from ARM runner

kgeckhart:

Removes the dedicated `build_mac_intel` job (~17 min, long pole on PR critical path). Cross-compiles `darwin/amd64` from the existing macOS ARM runner instead. Native Intel hardware testing remains in the release pipeline.

### Pull Request Details

Measured on this PR:

| | Before | After |
|---|---|---|
| `build_mac_intel` | 17 min (parallel) | removed |
| `build_mac_arm` / `build_mac` | 7 min | 12m 40s |
| **PR critical path contribution** | **17 min** | **12m 40s** |

### Notes to the Reviewer

Apple shipped its last Intel Macs in 2023 (Mac Pro) and the lineup has been ARM-only since. The cost of a dedicated 17-minute runner per PR to verify native Intel hardware compatibility is increasingly hard to justify when nearly all users are on Apple Silicon. A successful cross-compile on every PR plus a native build on every release strikes a better balance.

The `Build on MacOS (Intel)` status check is removed and `Build on MacOS (ARM)` is renamed to `Build on MacOS`. I'll update the branch protection rule to match before merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,23 +62,8 @@ jobs:
         cache: false
     - run: GO_TAGS="embedalloyui promtail_journal_enabled" GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} GOARM= GOEXPERIMENT=boringcrypto make alloy
 
-  build_mac_intel:
-    name: Build on MacOS (Intel)
-    runs-on: macos-15-large
-    steps:
-    - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      with:
-        persist-credentials: false
-    - name: Set up Go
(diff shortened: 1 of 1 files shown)


jharvey10 (commented):
I'm wondering if there's any way we could/should run this on the release-please PR. Either that or I feel like we should have some better alerting to CI failures on `main` so this is more in-your-face. Could be done as a separate PR though because I'm not sure what the best ergonomics are.

kgeckhart:
Good point. The release-please PR is a natural place to catch darwin/amd64 regressions before release — I can ship a follow-up that adds the native Intel build there. We will probably want similar behavior for the windows tests job which is also main only and also fails in a manner that's not quite loud enough sometimes.

jharvey10 (approved):

PR #6199, opened 2026-05-06
feat(database_observability/postgres): add query_fingerprint to op=query_association entries

gaantunes:
## Summary

Adds `query_fingerprint="..."` to the existing `op="query_association"` Loki entry that the `query_details` collector emits per `pg_stat_statements` row. The fingerprint becomes a queryid → logical-query lookup table that downstream consumers can query via LogQL to translate `pg_stat_statements_*` PromQL counters into per-fingerprint rollups.

Stacked on top of #6180 (`gaantunes/pg-errors-by-fingerprint-counter`) — this PR depends on the `fingerprint` package introduced there. **Base branch is set accordingly.** Will rebase to `main` once #6180 lands.

### What ships

- New body field `query_fingerprint` on every `op="query_association"` entry, computed via `pg_query.Fingerprint` from the raw `pg_stat_statements.query` text *before* comment stripping (so the value is stable across comment differences).
- Identical fingerprint identity across all three surfaces — `pg_query.Fingerprint` canonicalizes literals at the AST level, so the fingerprint computed from the partially-normalized `pg_stat_statements.query` text matches the one computed from raw `pg_stat_activity.query` (future `op="query_sample"`) and from server-log `STATEMENT:` continuations (`op="error"` from #6180).
- Reference docs entry under "Emitted Loki entries" describing the new field, including the LogQL pattern for queryid lookup.

### LogQL → PromQL workflow this unlocks

Two-step "errors per logical query" / "calls per logical query" aggregation:

1. `{op="query_association"} | logfmt | query_fingerprint="<fp>"` → list of queryids belonging to a fingerprint
2. `pg_stat_statements_calls_total{queryid=~"<id1>|<id2>|..."}` (or any other counter) → aggregate metrics

This is the lighter-weight alternative to a join metric (`database_observability_query_hash_info`) — keeps `pg_stat_statements_*` cardinality untouched and avoids new high-cardinality Prometheus labels.

### Test discipline

- `query_details.go` change is small (~10 lines): one `fingerprint.Fingerprint(...)` call before the existing `removeComments`, one new field in the `Sprintf` format string.
- The 22 sub-tests in `TestQueryDetails` and 3 sub-tests in `TestQueryDetails_SQLDriverErrors` were updated using a `<fp>` placeholder + `substituteFingerprints` helper (mirrors the pattern used in #6180's `query_samples` work). No fingerprint hex hard-coded — every expected value is computed via `fingerprint.Fingerprint(...)` at test time, so `pg_query_go` version bumps don't silently invalidate assertions.

### What this is *not*

- Not a Prometheus join metric (`database_observability_query_hash_info`) — that's deferred unless concrete PromQL aggregation use cases show up.
- Not a registry / cache — fingerprint is computed inline per scrape.
- Not the `query_samples` / `wait_event` integration — separate follow-up PR (PR 3 in the breakdown).

## Test plan

- [x] `go test -count=1 ./internal/component/database_observability/postgres/...` — all green
- [x] `go vet ./internal/component/database_observability/postgres/...` — clean
- [x] All 22+3 pre-existing `TestQueryDetails*` sub-tests pass with the new `query_fingerprint=` field
- [ ] Manual verification: bring up local `postgres:18` with `pg_stat_statements`, run `database_observability.postgres`, confirm `op="query_association"` Loki entries on `forward_to` carry the new field

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.postgres.md
+++ b/docs/sources/reference/components/database_observability/database_observability.postgres.md
@@ -229,28 +229,40 @@ window around the entry's timestamp. If the failed query was inside a write
 transaction, `xid` is also a deterministic key against
 `pg_stat_activity.backend_xid`.
 
-### `op="query_association"` entries from `query_details`
+### `op="query_association"` / `op="query_association_v2"` entries from `query_details`
 
 The `query_details` collector emits one Loki entry per `pg_stat_statements`
-row on each scrape. The line carries the row's `queryid`, the computed
-`query_fingerprint`, and the partially-normalized SQL text from the view.
-Use it as a queryid → fingerprint lookup table to translate aggregate
-metrics in `pg_stat_statements_*` into per-fingerprint rollups.
+row on each scrape. The exact `op` label depends on the parent block's
+`enable_query_fingerprint` argument:
 
-| Field                          | Source                          | Notes                                                                |
-|--------------------------------|---------------------------------|----------------------------------------------------------------------|
-| Label `op`                     | constant                        | `query_association`                                                  |
-| Body field `level`             | constant                        | `info`                                                               |
-| Body field `queryid`           | `pg_stat_statements.queryid`    | PostgreSQL's native, version-dependent identifier                    |
-| Body field `query_fingerprint` | computed                        | `libpg_query` fingerprint of the SQL text (16-char hex)              |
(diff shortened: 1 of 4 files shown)


gaantunes:
Rolled into #6200 — that PR has been retargeted to base directly on #6180 and now contains the full diff for both `query_association_v2` and `query_sample_v2`/`wait_event_v3`/`wait_event_v4` plus the `track_activity_query_size` plumbing and pipeline counters. Closing this in favor of the consolidated PR.

PR #6202, opened 2026-05-07
feat(prometheus): Add prometheus.expose component to expose metrics on /metrics

### LINKED ISSUE #576: ability to expose metrics from multiple integrations on a /metrics endpoint
### Request

I want to use grafana alloy for exposing node exporter and process exporter metrics, in the documentation I can see that there's a way to forward those metrics to a prometheus remote write url, I wanted to know if there's a way to expose those metrics on a /metrics type endpoint in grafana alloy.

### Use case

We want to use it in our already existing monitoring system which scrapes metrics from a huge fleet of EC2 instances, we don't want to remote write from every EC2 instance.

yogeshmahajan-1903:

Add a new `prometheus.expose` component that receives Prometheus metrics via
`forward_to` and re-exposes them on Alloy's own `/metrics` endpoint. This
allows scraped or forwarded metrics to be surfaced as part of Alloy's
self-monitoring without requiring an external remote_write target.

### Pull Request Details

The component:
- Accepts metrics through a `receiver` export (compatible with any
  `forward_to` target)
- Stores the latest value per time series in memory
- Registers a dynamic Prometheus collector against `prometheus.DefaultRegisterer`
  so metrics appear at `http://<alloy-host>:12345/metrics`
- Supports optional `namespace`, `subsystem`, and global `labels` arguments
  for metric name prefixing and label injection

**Known limitation:** multiple `prometheus.expose` instances in the same
config will conflict if they receive metrics with identical names, since all
instances share `DefaultRegisterer`.

Fixes #576 

### Notes to the Reviewer

1. Stability is set to Experimental as this is a new component.
2.  The component uses an unchecked prometheus. Collector (empty Describe) because metric names are determined dynamically from incoming time series.
3. All metrics are treated as gauges — counter/histogram types are not distinguished at this stage.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.expose.md
+++ b/docs/sources/reference/components/prometheus/prometheus.expose.md
@@ -1,0 +1,106 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.expose/
+description: Learn about prometheus.expose
+labels:
+  stage: experimental
+  products:
+    - oss
+title: prometheus.expose
+---
+
+# `prometheus.expose`
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+The `prometheus.expose` component receives Prometheus metrics and re-exposes them on Alloy's own `/metrics` endpoint.
+This is useful when you want scraped or forwarded metrics to be visible as part of Alloy's self-monitoring metrics.
+
+## Usage
+
+```alloy
+prometheus.expose "<LABEL>" {
+}
+```
+
+## Arguments
+
+You can use the following arguments with `prometheus.expose`:
+
+| Name        | Type                | Description                                      | Default | Required |
+| ----------- | ------------------- | ------------------------------------------------ | ------- | -------- |
+| `namespace` | `string`            | Prefix added to all metric names as `namespace_`.| `""`    | no       |
+| `subsystem` | `string`            | Prefix added to all metric names as `subsystem_`.| `""`    | no       |
+| `labels`    | `map(string)`       | Extra labels added to every metric.              | `{}`    | no       |
+
+When both `namespace` and `subsystem` are set, the resulting metric name is `namespace_subsystem_<original_name>`.
+
+Global labels defined in `labels` are only added when the incoming metric does not already carry a label with the same name.
+
+## Blocks
+
+The `prometheus.expose` component doesn't support any blocks. You can configure this component with arguments.
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+| Name       | Type              | Description                                               |
+| ---------- | ----------------- | --------------------------------------------------------- |
+| `receiver` | `MetricsReceiver` | A value that other components can use to send metrics to. |
+
+## Component health
+
+`prometheus.expose` is only reported as unhealthy if given an invalid configuration.
+In those cases, exported fields retain their last healthy values.
+
+## Debug information
+
+`prometheus.expose` doesn't expose any component-specific debug information.
+
+## Debug metrics
+
+`prometheus.expose` doesn't expose any component-specific debug metrics.
+
+## Example
+
+Expose scraped application metrics on Alloy's own `/metrics` endpoint:
+
+```alloy
+prometheus.scrape "app" {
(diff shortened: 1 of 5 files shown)


yogeshmahajan-1903:
I have read the CLA and agree to the terms

jharvey10:
Hello! Could you give some specific examples of what you'd use this component for? You mention wanting some external prom metrics to be exposed as a part of Alloy's self-monitoring, but I'm having trouble understanding what types of external metrics you might want to add that would make sense to be a part of Alloy's meta monitoring metrics.

yogeshmahajan-1903:
> Hello! Could you give some specific examples of what you'd use this component for? You mention wanting some external prom metrics to be exposed as a part of Alloy's self-monitoring, but I'm having trouble understanding what types of external metrics you might want to add that would make sense to be a part of Alloy's meta monitoring metrics.

Hi,

I want to expose metrics which are scraped using `prometheus.exporter.unix `.
Here is full example - 
```
// Prometheus Exporters and Metrics
prometheus.exporter.unix "system_metrics" {
  set_collectors = ["cpu", "filesystem", "meminfo", "diskstats", "netdev","loadavg", "boottime"]
  include_exporter_metrics = true
}

prometheus.scrape "system_metrics" {
  targets    = prometheus.exporter.unix.system_metrics.targets
  forward_to = [prometheus.remote_write.endpoint.receiver]
}

prometheus.scrape "self_metrics" {
  targets = [
    {
      "__address__" = "127.0.0.1:12345",
      "job"         = "alloy-self-monitoring",
      "instance" = sys.env("HOSTNAME"),
    },
  ]
  forward_to = [prometheus.remote_write.endpoint.receiver]
}


prometheus.remote_write "endpoint" {
  endpoint {
    url = "http://172.17.0.2:9090/api/v1/write"
  }
}
```
So, instead of remote write, this PR enables to expose endpoint that can be scraped by Prometheus. Endpoint  `<alloy_host>:12345/metrics` will be used to expose the same. You can see multiple requests for the same in  #576

kalleep:
Hey, thanks for the pr. I am not sure we want to add something like this into alloy. Also note that we would prefer that you write a proper proposal for something like this before making a pr, that gives us (alloy maintainers) a change to review the proposal before any work has been done. 

>Known limitation: multiple prometheus.expose instances in the same
config will conflict if they receive metrics with identical names, since all
instances share DefaultRegisterer.

This is not an acceptable limitation.

yogeshmahajan-1903:
> lso note that we would prefer that you write a proper proposal for something like this before making a pr, that gives u

I would rename know limitation to warning. I think, if Prometheus is receiving metrics with an identical name, it will be mixed.
I thought this was just an addition of one more component, hence I created PR. Could you please let me know if you (maintainers) want to go ahead with this or not?

kalleep:
>I thought this was just an addition of one more component, hence I created PR. Could you please let me know if you (maintainers) want to go ahead with this or not?

The issue you linked is just describing that they want to have some way to expose metrics on a `/metrics` endpoint. But no where in there does it describe how this would work. AFAICT this would expose metrics on alloys internal `/metrics` endpoint and that is one way to solve it but it's not the only way. It will come with limitations like the ones you mention that IMO is not acceptable. Another solution would be to collect metrics and expose them on it's own `/metrics` endpoints. That is why I said we would want a proposal for something like this before a pr so we can look it over and reject / make suggestions before any work is being put in to a pr.

yogeshmahajan-1903:
> > I thought this was just an addition of one more component, hence I created PR. Could you please let me know if you (maintainers) want to go ahead with this or not?
> 
> The issue you linked is just describing that they want to have some way to expose metrics on a `/metrics` endpoint. But no where in there does it describe how this would work. AFAICT this would expose metrics on alloys internal `/metrics` endpoint and that is one way to solve it but it's not the only way. It will come with limitations like the ones you mention that IMO is not acceptable. Another solution would be to collect metrics and expose them on it's own `/metrics` endpoints. That is why I said we would want a proposal for something like this before a pr so we can look it over and reject / make suggestions before any work is being put in to a pr.

Thanks for your inputs. Could you please provide me with a link/format for raising the proposal?

kalleep:
https://github.com/grafana/alloy/blob/main/docs/design/README.md

PR #6204, opened 2026-05-07
feat: Migrate loki pipeline to single entry consumer

kalleep:
### Pull Request Details
This it the first step to convert loki pipeline to use function calls, phase 3 from https://github.com/grafana/alloy/issues/4953. 

This pr only changes how components communicate with each other and I did not go deeper into component implementations, this will also not leverage the batch functionality yet.

All sources are still collecting entries from a channel, e.g. every file sends on a channel and the component will forward it through the pipeline, so it _should_ behave as it did before.

I also added `InterceptorConsumer` that is useful to use for components that will both consume and forward entries, we have a similar thing for promtheus pipelines.

Most processor components where simple to migrate and also make it possible for several components to call at the same time, so I had to make sure they are safe to process entries in parallel:
* loki.secretfilter
* loki.relabel
* loki.enrich

The one that stands out here is `loki.process`. This is a much bigger refactor because we need to redesign the whole pipeline logic. To be able to do this in another pr we implement the Consumer interface and try to pass entries to the processing pipeline channel. Then we have to read from the out channel and forward it futher in the pipeline. This will disconnect entries from where it originated, e.g. `loki.source.file` will see it as sucessful even though it was just consumed by `loki.process` and not yet at a sink component, suck as `loki.write`, `loki.echo` or `otelcol.receiver.loki`.

After this pr is merged I will create individual pr:s to fix internals of components and add support from batching, starting from sink components, the processor components and last sources.

### Notes to the Reviewer

* I do not use `ShardingConsumer` in sources yet, since all sources are still only forwarding one log at a time.
* I made some changes to when we call OnStateChanged in components, specifically we should do that after we are sure that we could build component.
* I added nil guard to `fanout.Consume` and `fanout.ConsumeEntry`. If we fail to create a component on initial load alloy will stop and report a problem. But if it happens during a config reload, we create  a new component that fails, then we don't export any receiver and alloy will pass a nil receiver to components.
* Fanout will now snapshot consumers instead of holding read lock for the duration of send. When using channels this was required so that downstream components was not stopped and thus stop reading from channels causing a deadlock. Now this is safe because we just call a function that will return an error if it has already stopped, same as we do for Prometheus pipelines.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/loki/process/metric/counters.go
```diff
@@ -92,24 +92,24 @@ func (c *Counters) With(labels model.LabelSet) prometheus.Counter {
 
 type expiringCounter struct {
 	prometheus.Counter
-	lastModSec int64
+	lastModSec atomic.Int64
```
kalleep, line 95: This is not directly related to this pr but I detected this in tests. Before there was a race here where the metric stage can mutate this at the same time we collect metrics

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -40,17 +40,15 @@ var supportedSeverities = map[string]struct{}{
 
 type LogsArguments struct {
 	Receiver         loki.LogsReceiver
-	EntryHandler     loki.EntryHandler
```
kalleep, line 47: This was not used, The same receiver was passed twice as a `loki.LogsReceiver` and `loki.EntryHandler`

kalleep:
In all of our docs we have `loki.LogsReceiver` as the type, should we bother to change this or is it fine?

internal/runtime/logging/logger.go
```diff
@@ -207,31 +207,22 @@ func (l *Logger) addRecord(r slog.Record, df *deferredSlogHandler) {
...
 			},
-		}:
-		default:
-			return 0, fmt.Errorf("lokiWriter failed to forward entry, channel was blocked")
-		}
+		),
+	)
+	if err != nil {
+		return 0, err
 	}
```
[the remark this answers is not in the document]
kalleep: Yeah I was thinking about this one but just moved on, this for sure needs a fix

internal/component/common/loki/consumer_interceptor.go
```diff
@@ -0,0 +1,76 @@
...
+func (i *InterceptorConsumer) ConsumeEntry(ctx context.Context, entry Entry) error {
+	if i.onConsumeEntry != nil {
+		entry, keep, err := i.onConsumeEntry(ctx, entry)
+		if err != nil || !keep {
+			return err
+		}
+		return i.next.ConsumeEntry(ctx, entry)
+	}
+
+	return errors.New("loki interceptor: unimplemented consume entry")
```
x1unix, line 71: Shouldn't `i.next.ConsumeEntry` still be called but with an original `entry` if `WithConsumeEntryHook` wasn't provided?

kalleep: I don't think so because that would mean that we would skip a component. In reality all components are currently only implementing `ConsumeEntry` and this is just temporary until we have moved to batching. After that  we should remove `ConsumeEntry`

internal/runtime/logging/logger.go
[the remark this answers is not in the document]
kalleep: fixed in [dedec69](https://github.com/grafana/alloy/pull/6204/commits/dedec69ab9aa3315bf743da43cdbc75e58e08ad0)

internal/component/common/loki/consumer_interceptor.go
kalleep: And when that is done it should not be optional to add a function for Consume.

kalleep:
@codex review

gaantunes (dismissed):
Reviewed db-o11y related components, looks clean, thanks for the bugfix.

docs/sources/reference/components/loki/loki.enrich.md
```diff
@@ -44,10 +44,10 @@ You can use the following arguments with `loki.enrich`:
...
+| `forward_to`         | `list(LogsReceiver)`  | List of receivers to send enriched logs to.                                                      |                        | yes      |
 | `target_match_label` | `string`              | The label from discovered targets to match against, for example, `"__inventory_consul_service"`. |                        | yes      |
-| `targets`            | `[]discovery.Target`  | List of targets from a discovery component.                                                      |                        | yes      |
-| `labels_to_copy`     | `[]string`            | List of labels to copy from discovered targets to logs. If empty, all labels will be copied.     |                        | no       |
+| `targets`            | `list(map(string))`   | List of targets from a discovery component.                                                      |                        | yes      |
+| `labels_to_copy`     | `list(string)`        | List of labels to copy from discovered targets to logs. If empty, all labels will be copied.     |                        | no       |
```
clayton-cornell:
```suggestion
| `labels_to_copy`     | `list(string)`        | List of labels to copy from discovered targets to logs. If empty, all labels are copied.     |                        | no       |
```

docs/sources/reference/components/loki/loki.enrich.md
```diff
@@ -44,10 +44,10 @@ You can use the following arguments with `loki.enrich`:
...
 | `logs_match_label`   | `string`              | The label from incoming logs to match against discovered targets, for example `"service_name"`.  |                        | no       |
 
 If not provided, the `logs_match_label` attribute will default to the value of `target_match_label`.
```
clayton-cornell:
```suggestion
If not provided, the `logs_match_label` attribute defaults to the value of `target_match_label`.
```

clayton-cornell:
Doc changes look OK

kalleep:
In [bb9ae57](https://github.com/grafana/alloy/pull/6204/commits/bb9ae57844337e4ec25bedb8c799f16b79ca86f1) I fixed some issues we have in some of our loki components. We exported receivers before we had successfully created components

clayton-cornell:
Doc changes look OK to me.

internal/component/common/loki/consume.go
```diff
@@ -4,19 +4,46 @@ import (
...
-func Consume(ctx context.Context, recv LogsReceiver, f *Fanout) {
+func ConsumeBatch(ctx context.Context, recv LogsBatchReceiver, f *FanoutConsumer) {
+	for {
+		// NOTE: Select is not deterministic so we should check if context was canceled
+		// before we start waiting on channel again.
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
```
ptodev, line 21: If select is non-deterministic, then even with this extra select statement we can still have a non-deterministic select case a few lines down? Should we instead modify the 2nd select statement to take into account queued batches and closed channels? E.g. maybe `case batch := <-recv.Chan()` can also check for `<-ctx.Done()`? Or the other way around, if ctx.Done() then we could drain anythhing left before returning?

internal/component/loki/secretfilter/secretfilter.go
```diff
@@ -248,12 +250,48 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
+		c.fanout,
+		loki.WithConsumeEntryHook(func(ctx context.Context, entry loki.Entry) (loki.Entry, bool, error) {
+			c.mut.RLock()
+			defer c.mut.RUnlock()
+
+			if c.stopped {
+				return entry, false, loki.ErrConsumerStopped
+			}
+
+			if c.shouldProcessEntry() {
```
ptodev, line 269: Is this ok to run in parallel? The [Int63](https://pkg.go.dev/math/rand#Source) func is not thread safe.

ptodev (commented):
Would it be possible to split the PR into smaller ones? It'd be easier to review 😅  Otherwise it'd take a long time to look into everything properly.

internal/component/common/loki/consume.go
kalleep: We cannot close the channel since we are reading from it. Then it would case a panic from sender.

But yes we could move check it again after we have received a batch / entry, that is better 👍 

These functions will eventually go away when we have rebuilt users of it

internal/component/loki/secretfilter/secretfilter.go
kalleep: Good catch, no it's not!

kalleep:
>Would it be possible to split the PR into smaller ones? It'd be easier to review 

Yeah and it's probably better to get this one moving, what do you think about doing a pr for each component instead and just implement the interface but not use it yet, in that work I will make sure components are safe to call in parallel. It would reduce the scope a lot.

Then when it's done we switch over the pipeline to use functions instead of channels. WDYT?

cristiangreco:
@kalleep what's the latest on this work? `@grafana/db-o11y-squad` is still tagged for review but it seems this PR is either paused or has been split into multiple smaller parts?

Asking because the squad gets actively notified about this open PR pending review, wondering if it could be either closed or converted to draft.

kalleep:
@cristiangreco I was OOO for a while but yes I am splitting this up in smaller pr:s. So no need to monitor this one for now :)

PR #6205, opened 2026-05-07
chore: Refactor logger handlers

ptodev:
### Pull Request Details

`Logger.handler` now dispatches through a generic `multiSlogHandler` that fans every record out to both the `write_to` components and to `stderr`.

This makes the `write_to` and the `stderr` paths more distinct. It will be useful in a future PR (#5626) when we add an option to make it possible to change the default `stderr` output to Windows Event Log. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/runtime/logging/logger.go
+++ b/internal/runtime/logging/logger.go
@@ -31,9 +31,15 @@ type Logger struct {
 	level  *slog.LevelVar // Current configured level.
 	format *formatVar     // Current configured format.
 
-	// handler is the unified dispatch point: a multiSlogHandler that fans out
-	// to the primary destination handler and the aux fan-out handler.
-	// Constructed in NewDeferred and rebuilt by Update.
+	// primary writes the primary log output to l.inner (stderr).
+	// aux writes the auxiliary fan-out (write_to + tmp) to auxWriter.
+	// Both are constructed in NewDeferred and not replaced afterward, so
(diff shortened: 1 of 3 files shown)


internal/runtime/logging/logger.go
```diff
@@ -22,17 +22,35 @@ type EnabledAware interface {
...
+	// handler is the unified dispatch point: a multiSlogHandler that fans out
+	// to the primary destination handler and the aux fan-out handler.
+	// Constructed in NewDeferred and rebuilt by Update.
+	handler slog.Handler
+
+	// auxWriter holds the dynamic write_to (loki) and temporary writer (used
+	// by the support bundle endpoint) for the aux handler. These are
+	// independent of the primary destination, so they receive logs whichever
+	// primary is in use.
+	auxWriter *auxWriterVar
```
kalleep, line 49: What does aux stand for? 🤔

ptodev: auxiliary

kalleep (commented):
I like this change but IMO then we should also make loki and tmp writer independent of each other, WDYT?

ptodev:
> I like this change but IMO then we should also make loki and tmp writer independent of each other, WDYT?

I agree that this would make sense. I didn't do it here so that we can keep the PR small. We can do it in a future PR?

kalleep (approved):

PR #6206, opened 2026-05-07
chore(database_observability): Refactor health check tests

cristiangreco:
Refactor health check tests to use constants for queries.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


matthewnolf (commented):
Can we prefix the const with QueryXYZ rather than xyzQuery?
Can we also move the functions that generate queries next to the const?

matthewnolf (approved):

PR #6216, opened 2026-05-09
fix(loki.process): Wrap NewPipeline error correctly in match stage

boinger:
## Pull Request Details

In `newMatcherStage`, the error wrap for a failed inner-pipeline
construction had its `fmt.Errorf` arguments reversed. The buggy line
at `match.go:76`:

```go
return nil, fmt.Errorf("%v: %w", err, fmt.Errorf("match stage failed to create pipeline from config: %v", config))
```

The `%w` verb wraps an error so `errors.Is` and `errors.As` can
unwrap it. Here the `%w` slot wraps the *inner* `fmt.Errorf` ("match
stage failed..."), not the actual `err` from `NewPipeline`. So
`errors.Unwrap`, `errors.Is`, and `errors.As` all silently fail to
reach the original error, which is reduced to a stringified message
in the `%v` slot.

The adjacent line 82 has the correct shape: `fmt.Errorf("%v: %w",
"error parsing pipeline", err)`, with a static string in `%v` and
the real `err` in `%w`.

The fix:

```go
return nil, fmt.Errorf("match stage failed to create pipeline from config %+v: %w", config, err)
```

No observable user-visible bug today (no callers in alloy use
`errors.Is`/`errors.As` on the returned error from match stage
construction), but the wrap is incorrect and would silently fail
sentinel-error matching if any caller starts using it. This is
correctness preservation for future callers.

### Notes to the Reviewer

`TestMatchStage_NewPipelineErrorIsWrapped` builds a match config with
a deliberately invalid inner regex, then asserts `errors.Unwrap`
reaches `NewPipeline`'s wrap (which contains `"invalid stage config"`).
Without the fix, `errors.Unwrap` returns the dummy fresh `fmt.Errorf`
and the assertion fails with a clear error. With the fix it passes.

Verified deterministically with `go test -race -count=3` and `make lint`.

### PR Checklist

- [ ] Documentation added (n/a)
- [x] Tests updated
- [ ] Config converters updated (n/a)

### DIFF
--- a/internal/component/loki/process/stages/match_test.go
+++ b/internal/component/loki/process/stages/match_test.go
@@ -1,6 +1,7 @@
 package stages
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/component/loki/process/stages/match_test.go
```diff
@@ -227,3 +228,42 @@ func TestValidateMatcherConfig(t *testing.T) {
...
+// TestMatchStage_NewPipelineErrorIsWrapped verifies that when newMatcherStage's
+// NewPipeline call fails, the returned error wraps the original error from
+// NewPipeline so callers can use errors.Unwrap / errors.Is / errors.As to
+// inspect the underlying cause.
+//
+// Regression test for the reversed %w wrap that was previously at match.go:76,
+// where the %w slot held a fresh fmt.Errorf instead of the actual err from
+// NewPipeline. With the bug, errors.Unwrap returned the dummy fmt.Errorf
+// ("match stage failed to create pipeline...") and lost the original error
+// context. With the fix, errors.Unwrap returns NewPipeline's error.
```
kalleep: We would we need this amount of comments, including those below, for this simple test? 

IMHO this is just create a lot of noise

boinger: Ah, you're right, I'm sorry.  I defaulted to writing it like I write my own tests (and I'm admittedly very verbose with comments) and I should have paid more attention to the project's pattern/preference.

I'll trim it and force-push.

internal/component/loki/process/stages/match_test.go
```diff
@@ -227,3 +228,26 @@ func TestValidateMatcherConfig(t *testing.T) {
 		})
 	}
 }
+
+// TestMatchStage_NewPipelineErrorIsWrapped verifies the match stage wraps NewPipeline's error for errors.Unwrap.
```
kalleep: Not useful at all, this could all be inferred from the test name

kalleep (approved):

PR #6223, opened 2026-05-11
chore: Migrate loki source components to slog

kalleep:
Migrate loki source components to use slog

### DIFF
--- a/internal/component/loki/source/aws_firehose/component_test.go
+++ b/internal/component/loki/source/aws_firehose/component_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/grafana/alloy/internal/component/common/loki"
 	fnet "github.com/grafana/alloy/internal/component/common/net"
 	alloy_config "github.com/grafana/alloy/internal/component/common/relabel"
-	"github.com/grafana/alloy/internal/util"
+	"github.com/grafana/alloy/internal/runtime/logging"
 )
 
 const singleRecordRequest = `{"requestId":"a1af4300-6c09-4916-ba8f-12f336176246","timestamp":1684422829730,"records":[{"data":"eyJDSEFOR0UiOi0wLjIzLCJQUklDRSI6NC44LCJUSUNLRVJfU1lNQk9MIjoiTkdDIiwiU0VDVE9SIjoiSEVBTFRIQ0FSRSJ9"}]}`
@@ -62,7 +62,7 @@ func TestComponentFromNestedController(t *testing.T) {
 
 	opts := component.Options{
 		ID:            "foo/loki.source.awsfirehose.default",
-		Logger:        util.TestAlloyLogger(t),
+		SLogger:       logging.NewSlogNop(),
 		Registerer:    prometheus.NewRegistry(),
 		OnStateChange: func(e component.Exports) {},
 	}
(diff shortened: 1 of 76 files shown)


internal/component/loki/source/podlogs/podlogs.go
```diff
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-kit/log"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
```
ptodev, line 14: `newReconciler` and `newController` are defined in `reconciler.go` and `controller.go` with `log.Logger` signatures and were not migrated in this PR. Since `c.log` was removed from the Component struct, these are the only remaining go-kit logger references in the package. Consider either migrating those two functions to accept `*slog.Logger` or adding a FIXME comment here consistent with the pattern used for `slogadapter.GoKit` calls elsewhere.

ptodev (dismissed):
Clean and thorough migration. The "Address comments" sweep fixed the remaining small issues (err vs err.Error(), the log message typos, and the fmt.Sprintf("%T", m) in syslogtarget). FIXME comments clearly document the slogadapter.GoKit bridge calls.

One remaining inconsistency: in podlogs/podlogs.go, kubetail.NewManager was switched to o.SLogger but the adjacent newReconciler and newController calls still pass o.Logger because reconciler.go and controller.go have not been migrated. The c.log field was removed from the struct, so these are the only go-kit logger references left in the package. Worth either migrating those two functions or adding a FIXME comment matching the pattern used elsewhere. Not a blocker.

internal/component/loki/source/aws_firehose/component_test.go
```diff
@@ -177,7 +177,7 @@ func TestComponent(t *testing.T) {
 func TestComponent_UpdateWithNewArguments(t *testing.T) {
 	opts := component.Options{
 		ID:            "loki.source.awsfirehose",
-		Logger:        util.TestAlloyLogger(t),
+		SLogger:       logging.NewSlogNop(),
```
ptodev, line 180: The previous logger was not a no-op. Does that mean that if the tests fail now, we won't see log messages in the logs?

ptodev (changes requested):

internal/component/loki/source/podlogs/podlogs.go
kalleep: Ah yeah missed those, thanks :+1:

ptodev: ^ This was from the AI, but seems like a legit comment.

ptodev: No problem 😆

internal/component/loki/source/aws_firehose/component_test.go
kalleep: That is correct, I don't generally see any usefulness in emitting logs in tests because is creates a lot of noise. But if we want that I can do something similar, I am fine with whatever :)

ptodev (approved):

internal/component/loki/source/aws_firehose/component_test.go
ptodev: Sometimes it's useful, but it's not too important. I'm happy to keep it a no-op.

PR #6231, opened 2026-05-11
fix(loki.source.syslog): Fix goroutine leak in UDP listener

### LINKED ISSUE #6207: loki.source.syslog leaks memory when processing UDP packets
### Component(s)

loki.source.syslog

### What's wrong?

The UDP code for loki.source.syslog is written without considering that UDP is a connectionless protocol. In 
`internal/component/loki/source/syslog/internal/syslogtarget/transport.go`, the acceptPackets function has the following code:
```go
	for {
		...
		n, addr, err = t.udpConn.ReadFrom(buf)
		if n <= 0 && err != nil {
			level.Warn(t.logger).Log("msg", "failed to read packets", "addr", addr, "err", err)
			continue
		}

		stream, ok := streams[addr.String()]
		if !ok {
			stream = NewConnPipe(addr)
			streams[addr.String()] = stream
			t.openConnections.Add(1)
			go t.handleRcv(stream)
		}
	}
```
handleRcv has the following loop:
```go
	for {
		datagram := make([]byte, t.maxMessageLength())
		n, err := c.Read(datagram)
		if err != nil {
			if err == io.EOF {
				break
			}

			level.Warn(t.logger).Log("msg", "error reading from pipe", "err", err)
			continue
		}
		...
	}
```

Because UDP is connectionless, c.Read will never return EOF under normal circumstances. This means handleRcv is an infinite loop, so alloy will hold onto this goroutine, with `datagram` allocated with t.maxMessageLength(), forever.

Since UDP is connectionless, clients are likely to use different source ports to send syslog messages over time, meaning alloy will have hundreds to thousands of these handleRecv goroutines for each syslog client.

See attached [pprof](https://github.com/user-attachments/files/27493618/profile.pb.gz) for evidence of memory leak

### Steps to reproduce

1. Start alloy with a loki.source.syslog component with a UDP listener.
2. Send UDP syslog messages to the listener from a large number of IP:source port pairs
3. Observe memory leak

### System information

_No response_

### Software version

_No response_

### Configuration

```text

```

### Logs

```text

```

### Tip

x1unix:

This PR refactors the UDP listener to address unbound goroutine leak.

### Pull Request Details

Instead of having separate per-host goroutine, the new approach is using a single separate processor goroutine + buffered channel as a queue.

This is a breaking change as it may require users to adjust queue size with the new `udp_queue_size` attribute. 

Fixes #6207 

### Notes to the Reviewer

I was also considered more complex implementation with adjustable processor goroutine pool.

It can be a single pool, just for processors, or 2 separate pools - for datagram receivers and processors.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


blewis12:
> This is a breaking change as it may require users to adjust queue size with the new udp_queue_size attribute.

just so I understand, would it be a breaking change because users would need to invoke this argument in order to access the fix? Or does it mean that users could get unexpected behavior without changing their config at all but upgrading alloy?

kalleep:
>just so I understand, would it be a breaking change because users would need to invoke this argument in order to access the fix? Or does it mean that users could get unexpected behavior without changing their config at all but upgrading alloy?

The breaking change will be that before we would start a consume + parse go routine for every unique host so if many different host where pushing at the same time it would parse them all in separate goroutines. With this change we don't do that specific part but only have goroutine for consuming messages and one for parsing / processing them and we would start to block when queue is full.

x1unix:
> just so I understand, would it be a breaking change because users would need to invoke this argument in order to access the fix? Or does it mean that users could get unexpected behavior without changing their config at all but upgrading alloy?

@blewis12 this changeset introduces backpressure because it brings a single goroutine instead of having per-sender goroutines.

If incoming datagrams frequency is higher than consumption rate (the default buffered channel size) - datagram receive loop will pause until at least one item is consumed by processor.

Users who observe this behavior will need to update component config to adjust `udp_queue_size` value.

internal/component/loki/source/syslog/config/config.go
```diff
@@ -131,6 +133,9 @@ type SyslogTargetConfig struct {
 
 	// RFC3164CiscoComponents enables and configures Cisco IOS syslog parsing.
 	RFC3164CiscoComponents *RFC3164CiscoComponents `yaml:"rfc3164_cisco_components"`
+
+	// UDPQueueSize determines max number of messages which may be waiting for a processor.
+	UDPQueueSize int `yaml:"udp_queue_length"`
```
kgeckhart: Should we be consistent with these?

kgeckhart:
I can see how it might technically count as a breaking change but it seems unlikely someone hits it without potentially running up against the limitations of the single channel pipelines.

internal/component/loki/source/syslog/internal/syslogtarget/transport.go
```diff
@@ -436,98 +436,119 @@ func (t *UDPTransport) Close() error {
...
-			if err == io.EOF {
-				break
-			}
-
-			level.Warn(t.logger).Log("msg", "error reading from pipe", "err", err)
-			continue
-		}
+func (t *UDPTransport) handleDatagram(src net.Addr, buf []byte) {
+	srcIP := hostFromAddr(src)
+	lbs := t.connectionLabels(srcIP)
```
kgeckhart: Claude pointed out that the calls behind this include `net.LookupAddr(` which could add some real overhead to each call if DNS resolution is slow. I feel like that's a pretty valid concern. You could go LRU cache or switch back to goroutine per host with a TTL.

internal/component/loki/source/syslog/config/config.go
x1unix: @kgeckhart fixed, just didn't push the commit yet.

x1unix: @kgeckhart fixed in https://github.com/grafana/alloy/pull/6231/changes/d721532d01a77c1f86db1b9466a1c2f3e22f3824

internal/component/loki/source/syslog/internal/syslogtarget/transport.go
x1unix: @kgeckhart thnx. I've used LRU cache with configurable size: https://github.com/grafana/alloy/pull/6231/changes/bd4d64b23ad03e025c3117415dddfd150141c537

kgeckhart (approved):
Might be worth it to ping the original reporter to see if this makes sense for their use case. Otherwise 👍

PR #6234, opened 2026-05-12
chore: Add integration test for loki.source.kubernetes

kalleep:
### Pull Request Details
Add k8s integration test for `loki.source.kubernetes`. To make this work I added:
1. Loki dep that will start a loki instance that we later can query for entries.
2. A `LogGen` output the supplied file to stdout.

In this test I created 3 namespaces and deploying 3 statefulset to each namesapce, I used statefulset to get stable pod names so I later can assert on structured_metadata.

Alloy will then discover pods in these three namespaces and consume logs from pods discovered and assert on number of entries for each namespace and pod.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


integration-tests/internal/lokihttp/loki.go
```diff
@@ -1,4 +1,4 @@
-package common
+package lokihttp
```
kalleep, line 1: I just moved these shared structured into an internal dicetory that can be shared by docker and k8s tests.

x1unix (approved):

PR #6239, opened 2026-05-12
feat(database_observability.mysql): Refactor schema_details logging strategy

cristiangreco:
Refactor the collector to log at most one OP_CREATE_STATEMENT per table every 30m, while still checking for new tables at every collect interval tick.

Remove the internal cache (settings are ignored/deprecated) and replace it with an in-memory map to throttle logging of OP_CREATE_STATEMENT to at most once per 30 minutes per table.

The metadata queries (columns, indexes, foreign keys) are also narrowed with TABLE_NAME IN (...) to fetch only "due" tables.

Note that we partially give up on schema DDL changes detection, which was anyway not working exactly as expected: the old internal cache was keyed by UPDATE_TIME timestamp, which however is not guaranteed to be updated on every schema changes (mysql caches metadata [24h](https://dev.mysql.com/doc/refman/8.0/en/information-schema-tables-table.html) by default).

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -136,12 +136,14 @@ The `gcp` block supplies the identifying information for the GCP Cloud SQL datab
 
 ### `schema_details`
 
-| Name               | Type       | Description                                                           | Default | Required |
-|--------------------|------------|-----------------------------------------------------------------------|---------|----------|
-| `collect_interval` | `duration` | How frequently to collect information from database.                  | `"1m"`  | no       |
-| `cache_enabled`    | `boolean`  | Whether to enable caching of table definitions.                       | `true`  | no       |
-| `cache_size`       | `integer`  | Cache size.                                                           | `256`   | no       |
-| `cache_ttl`        | `duration` | Cache TTL.                                                            | `"10m"` | no       |
+| Name               | Type       | Description                                                 | Default | Required |
+|--------------------|------------|-------------------------------------------------------------|---------|----------|
+| `collect_interval` | `duration` | How frequently to collect information from database.        | `"1m"`  | no       |
+| `cache_enabled`    | `boolean`  | Deprecated. Whether to enable caching of table definitions. | `true`  | no       |
+| `cache_size`       | `integer`  | Deprecated. Cache size.                                     | `256`   | no       |
+| `cache_ttl`        | `duration` | Deprecated. Cache TTL.                                      | `"10m"` | no       |
+
+The `cache_enabled`, `cache_size`, and `cache_ttl` settings are deprecated: they are accepted for backward compatibility, but ignored.
 
 ### `explain_plans`
 
--- a/internal/component/database_observability/mysql/collector/schema_details.go
+++ b/internal/component/database_observability/mysql/collector/schema_details.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/hashicorp/golang-lru/v2/expirable"
 	"go.uber.org/atomic"
 
 	"github.com/grafana/alloy/internal/component/common/loki"
@@ -25,6 +24,11 @@ const (
 	OP_TABLE_DETECTION     = "table_detection"
 	OP_CREATE_STATEMENT    = "create_statement"
 )
+
+// EmitInterval is the minimum amount of time that must elapse between
+// successive OP_CREATE_STATEMENT emissions for the same table, regardless of
+// the configured collect_interval.
+const EmitInterval = 30 * time.Minute
 
 const (
 	selectTablesTemplate = `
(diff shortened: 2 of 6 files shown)


docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -136,12 +136,16 @@ The `gcp` block supplies the identifying information for the GCP Cloud SQL datab
...
+| Name               | Type       | Description                                          | Default | Required |
+|--------------------|------------|------------------------------------------------------|---------|----------|
+| `collect_interval` | `duration` | How frequently to collect information from database. | `"1m"`  | no       |
+
+`schema_details` emits two operations per collected table:
+
+* `table_detection` is emitted once per detected table on every collection cycle, so it acts as a heartbeat for schema discovery.
+* `create_statement` carries the table's DDL and column/index/foreign-key metadata. To limit log volume and load on `information_schema`, it's emitted at most once every 30 minutes per `(schema, table)`, regardless of `collect_interval`. If `collect_interval` is set to a value larger than 30 minutes, {{< param "PRODUCT_NAME" >}} logs a warning and `create_statement` entries are emitted at most once per `collect_interval`.
+
+The `cache_enabled`, `cache_size`, and `cache_ttl` arguments are accepted for backward compatibility, but they're ignored. The 30-minute throttle described above replaces the previous in-memory cache.
```
cristiangreco: ✂️

internal/component/database_observability/mysql/collector/schema_details.go
```diff
@@ -391,7 +407,7 @@ func (c *SchemaDetails) fetchSchemaSpecs(ctx context.Context, schema string) (ma
 		return s
 	}
 
-	colRS, err := c.dbConnection.QueryContext(ctx, selectColumnNames, schema)
+	colRS, err := c.dbConnection.QueryContext(ctx, fmt.Sprintf(selectColumnNames, database_observability.BuildExclusionClause(tableNames)), schema)
```
cristiangreco, line 411: I'm reusing `BuildExclusionClause` to actually build an inclusion clause. 

Can address this in a followup to avoid this PR growing too large.

internal/component/database_observability/mysql/component.go
```diff
@@ -587,13 +587,18 @@ func (c *Component) startCollectors(serverID string, engineVersion string, parse
 	}
 
 	if collectors[collector.SchemaDetailsCollector] {
+		if c.args.SchemaDetailsArguments.CollectInterval > collector.EmitInterval {
+			level.Warn(c.opts.Logger).Log(
+				"msg", "schema_details.collect_interval is greater than the create_statement emission interval; create_statement entries will be emitted at most once per collect_interval",
+				"collect_interval", c.args.SchemaDetailsArguments.CollectInterval,
+				"emit_interval", collector.EmitInterval,
```
cristiangreco: This can be shortened

internal/component/database_observability/mysql/collector/schema_details.go
```diff
@@ -26,6 +25,11 @@ const (
 	OP_CREATE_STATEMENT    = "create_statement"
 )
 
+// EmitInterval is the minimum amount of time that must elapse between
+// successive OP_CREATE_STATEMENT emissions for the same table, regardless of
+// the configured collect_interval.
+const EmitInterval = 30 * time.Minute
```
cristiangreco, line 31: This is hardcoded for now. We could consider making it configurable in the future.

internal/component/database_observability/mysql/component.go
```diff
@@ -587,13 +589,18 @@ func (c *Component) startCollectors(serverID string, engineVersion string, parse
 	}
 
 	if collectors[collector.SchemaDetailsCollector] {
```
matthewnolf, line 591: I think we should also log a warning that deprecated config values are used

internal/component/database_observability/mysql/collector/schema_details.go
matthewnolf: Does this need to be exposed?

internal/component/database_observability/mysql/collector/schema_details.go
```diff
@@ -111,11 +111,13 @@ type SchemaDetails struct {
...
-	// the last update time of the table (this allows capturing schema changes
-	// at each scan, regardless of caching).
-	cache *expirable.LRU[string, *tableInfo]
+	// lastEmittedAt records the wall-clock time at which OP_CREATE_STATEMENT
+	// was last emitted for a "schema.table" key. Used to throttle logging
+	// to at most one per EmitInterval per table.
+	lastEmittedAt map[string]time.Time
+
+	// nowFn allows overriding time.Now() in tests
+	nowFn func() time.Time
```
matthewnolf, line 120:
```suggestion
	now func() time.Time
```

matthewnolf (approved):

Issue #6155, opened 2026-04-30
[Tracking Issue] Release 1.17


kgeckhart:
### Version

1.17

### Release Candidate Date

2026-06-09

### Stable Release Date

2026-06-12

### Link

#6111

kgeckhart:
We have been delayed waiting for prometheus to do a release as we hit the same blocking changes as last time. It's not looking great to have these line up so we might move forward again without updates.

kgeckhart:
Still delayed will post a new timeline when we have a better idea

kgeckhart:
We have been through a large round of refactor our of build processes. Prometheus did do a release and that release included bumping OTel from v1.151.0 -> v1.153.0 which adds more to the already really large diff (https://github.com/grafana/alloy/pull/6240). Our plan is to do this minor release without upgrading again and get to latest for the next minor release. Still planning a date as we want to try to include #5823 as it's a feature we have wanted since the OTel Engine released.

kgeckhart:
I was planning to start the release today but need to do another patch on 1.16 for #6436

jharvey10:
We're finalizing some improved workflow permissions changes and plan to cut an RC early this week.

kgeckhart:
Currently on track to finalize the release tomorrow

kgeckhart:
Release has shipped https://github.com/grafana/alloy/releases/tag/v1.17.0

PR #6240, opened 2026-05-12
feat!: Upgrade Prometheus to v3.12.0-rc.0, otel-contrib to v0.151.0, Loki to v3.7.1

### LINKED ISSUE #6503: prometheus to otel metric convertion broken for certain metrics
### Component(s)

prometheus.exporter.self

### What's wrong?

After updating to alloy v1.16.3 from v1.14.2 I see logs being thrown:

```
"rpc.client.call.duration_seconds" is not a valid metric name
"rpc.server.call.duration_seconds" is not a valid metric name
```

Checking our config/metrics the `rpc_(client|server)_*` metric seems to be coming from the `prometheus.exporter.self`. At least that's the only component that is emitting `rpc_(client|server)_*` metrics (identified via `job` label). In live debugging I don't see any `rpc_*` metric at all not there nor any other prometheus relabling component or the `otelcol.receiver.prometheus`.

We convert to otel before sending it to our backend. This seems to not working properly, but only for those two metrics, no other metrics is reporter as invalid. Also I don't see any historic data `rpc_(client|server)_call_*`. I'm 100 % sure that the logs are only written after the v1.16.3 update.

I'm not saying prom to otel conversion is completely broken. E.g. `probe_duration_seconds` from the blackbox exporter component works as expected. 

I've also tested v1.17.0, but I'm getting the same logs.

### Steps to reproduce

- use `prometheus.exporter.self "this" {}`
- scrape with `promtheus.scrape "self"`, we forward to a relabling component, but I don't think thats relevant
- receiver via `otelcol.receiver.prometheus"
- forward it to `otelcol.processor.batch` and from there to `otelco.exporter.otlp`
- see log line in alloy logs

### System information

aws ecs task with dockerhub image

### Software version

Grafana Alloy v1.16.3

### Configuration

```text
alloy run --server.http.listen-addr=... /etc/alloy
```

### Logs

```text
<date-time> "rpc.server.call.duration_seconds" is not a valid metric name
<date-time> "rpc.server.call.duration_seconds" is not a valid metric name
<date-time> "rpc.client.call.duration_seconds" is not a valid metric name
<date-time> "rpc.client.call.duration_seconds" is not a valid metric name

this is logged once a minute, which matches the scrape interval of `promtheus.scrape "self"`
```

### Tip

kgeckhart:

Bumps Prometheus, OpenTelemetry, and Loki for the next release:

- `prometheus/prometheus` → **v3.12.0-rc.0** (`v0.312.0-rc.0`); drops the `grafana/prometheus` fork.
- `opentelemetry-collector-contrib` → **uniform v0.151.0**; `go.opentelemetry.io/collector` core → **v0.151.0 / v1.57.0**.
- `grafana/loki/v3` → **v3.7.1** (forced — v3.6.x imports the removed `tsdb/errors`).
- `prom-label-proxy` → upstream **v0.13.0** (no fork); `prometheus-operator` → upstream **v0.91.0** + a one-line `kgeckhart/*` fork (details below).

Closes #6503 — the `otel/exporters/prometheus` bump (v0.60 → v0.65) stops Alloy's self-telemetry emitting the dotted `rpc.client.call.duration_seconds` name that failed metric-name validation.

### Pull Request Details

This is scoped to the upgrade and the config changes v0.151 forces — new fields default to their upstream values, and the remaining optional v0.151 config fields are deferred to a follow-up.

**Why `v3.12.0-rc.0` and not GA.** Prometheus GA (`v0.312.0`) couples the OTel stack to v0.153 — its native OTLP write path imports `deltatocumulativeprocessor` v0.153, which drags collector core to v1.59 and the otel SDK to v1.44 through MVS. rc.0 is the last tag before GA's CVE-driven `go get -u`, so it keeps those at v0.151 / v1.57 — matching this PR's uniform otel-contrib pin. GA + otel v0.153 is a focused follow-up. (Released v0.311.x can't be used either: `scrape.NewManager` gained an `AppendableV2` arg and `discovery.CreateAndRegisterSDMetrics` flipped to `*SDMetrics` on different release tracks, so no released v0.311.x is otel-contrib-compatible.) The pin lives in `dependency-replacements.yaml`; drop it when otel-contrib pins a released Prometheus tag.

**Forks.** `prometheus-operator` carries a one-line `kgeckhart/*` patch over v0.91.0 (the only release; its `rulefmt.Parse` call predates the `*slog.Logger` arg prom v3.12 requires). Upstream `main` has the fix but pulls `k8s.io/client-go` v0.36, incompatible with the pinned `controller-runtime` and otel-contrib v0.151 (both on v0.35) — and otel-contrib hasn't shipped the k8s 1.36 uplift in any release yet (latest is v0.154), so the fork stays until it does. `prom-label-proxy` needs no fork: prom restored `ExperimentalDurationExpr` before v3.12, so upstream v0.13.0 works as-is.

**Breaking changes for users.**

BREAKING-CHANGE: `otelcol.exporter.splunkhec` — `batcher` block removed (upstream deletion in otel-contrib v0.151). Configs using it will fail to load. Migration: move the equivalent fields under `sending_queue.batch` (`enabled` → presence of the block; `flush_timeout`/`sizer`/`min_size`/`max_size` keep their names).

BREAKING-CHANGE: `loki.process` — a label produced by a parser stage (e.g. `json`, `logfmt`) no longer overrides structured metadata of the same name (behavior change in Loki v3.7, pulled in by the v3.7.1 bump; no opt-out).

### PR Checklist

- [x] Documentation added
- [x] Tests updated


raymond-hardy-envoy:
Hi @kgeckhart — would you consider bumping to contrib v0.152.0 (or v0.153.0) instead of v0.151.0?

  open-telemetry/opentelemetry-collector-contrib#45564 landed in v0.152.0 and adds an idle_series_timeout option to otelcol.receiver.datadog. It addresses the unbounded MetricsTranslator series-state accumulation in contrib#45523. We ran into this with otelcol.receiver.datadog in an alloy-cluster deployment — ~1 GB/pod/hr heap growth, OOM-killed within a few hours of starting up. Restart cycles are the only mitigation today.

  v0.152.0 was tagged 2026-05-11 and v0.153.0 on 2026-05-26. The v0.151→v0.152 diff is small: spanmetrics adds a collector.instance.id attribute (gated), a handful of feature-gate promotions, snake_case renames with deprecated aliases.
  v0.152→v0.153 is similar.

  Can help test on our side if useful.

  Thanks!

kgeckhart:
Hey @raymond-hardy-envoy, thanks for the ping and offering to help! 

The plan we came up with internally was to finish the #6155 release. As soon as that's done I was planning to cleanup this PR and keep it only focused on the upgrades necessary to bump prometheus. Then do a follow up to bump OTel again to get us as close to latest as possible for 1.18 (ticket TBD). This way we get this likely large PR and hopefully something smaller to focus on for OTel. It will be documented on the 1.18 ticket when I make it.

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -91,6 +91,8 @@ You can use the following blocks with `otelcol.exporter.kafka`:
 | [`metrics`][metrics]                                    | Configures how to send metrics to Kafka brokers.                               | no       |
 | [`producer`][producer]                                  | Kafka producer configuration,                                                  | no       |
 | `producer` > [`compression_params`][compression_params] | Configures the compression parameters for the kafka producer.                  | no       |
```
clayton-cornell:
```suggestion
| `producer` > [`compression_params`][compression_params] | Configures the compression parameters for the Kafka producer.                  | no       |
```

internal/component/otelcol/processor/tail_sampling/tail_sampling.go
```diff
@@ -84,18 +84,20 @@ func (args Arguments) Convert() (otelcomponent.Config, error) {
...
-		BlockOnOverflow:               args.BlockOnOverflow,
-		ExpectedNewTracesPerSec:       args.ExpectedNewTracesPerSec,
-		SampleOnFirstMatch:            args.SampleOnFirstMatch,
-		DropPendingTracesOnShutdown:   args.DropPendingTracesOnShutdown,
-		MaximumTraceSizeBytes:         args.MaximumTraceSizeBytes,
-		PolicyCfgs:                    otelPolicyCfgs,
-		DecisionCache:                 args.DecisionCache.Convert(),
-	}, nil
+	// Start from the upstream default config so fields Alloy doesn't expose keep
+	// their upstream defaults, including the unexported sampling strategy.
```
dehaansa: Comment is confusing

dehaansa (dismissed):
One nit

internal/static/metrics/wal/wal.go
```diff
@@ -644,7 +649,7 @@ func (w *Storage) Truncate(mint int64) error {
 	w.metrics.checkpointCreationTotal.Inc()
 
 	// TODO(x1unix): pass EnableSTStorage when Prometheus will be upgraded
```
blewis12: is this comment outdated now that we actually do pass `EnableSTStorage` as false? or are we intending to set the argument to `true` at some point

clayton-cornell (dismissed):
Approving for docs. The one comment I had isn't that important.. just a nit.

internal/component/otelcol/processor/tail_sampling/tail_sampling.go
kgeckhart: Yeah I can see it, this particular area is really interesting. Upstream OTel has SamplingStrategy which is not exported (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/config.go#L360). I couldn't find any real reason why it was done this way but because it's unexported starting from the default makes it easier to not do really hacky things here.

internal/static/metrics/wal/wal.go
kgeckhart: I think we will want to write it up like upstream does, https://prometheus.io/docs/prometheus/latest/feature_flags/#start-timestamp-st-native-storage

blewis12: aha ok

alloy-service.exe
x1unix, line 1: Do we need to update our `.gitignore` to exclude `.exe`?

kgeckhart: Good catch, probably lol

dehaansa (approved):

PR #6242, opened 2026-05-12
feat: Add gql subcommand

### LINKED ISSUE #477: Feature Request: Add cURL to the Container Images
Hello,

Would it be possible to add cURL to the Container Images by default?

This helps with troubleshooting connectivity from the Agent Container to an Endpoint and I often find myself having to install it manually when testing/troubleshooting issues and it aids me in resolving them.

Additionally, it also allows you to define a custom Healthcheck in a Docker Compose file to use the `health` endpoint.

Thanks

### LINKED ISSUE #1891: [Feature Request] Alloy container health check

### LINKED ISSUE #5001: ecs healthcheck needed inside docker image
### Component(s)

_No response_

### Request

Can we get something like this built and available inside the alloy docker image?  https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/healthcheck/main.go

### Use case

AWS ECS Fargate health check of the docker image is very difficult to accomplish.  Having a binary like the URL linked about would allow this to be a simple task. 

### Tip

jharvey10:

Adds the `gql` subcommand to Alloy, allowing GraphQL queries to be invoked against a running Alloy instance.

Fixes #477  
Fixes #1891
Fixes #5001 

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/cli/gql.md
+++ b/docs/sources/reference/cli/gql.md
@@ -1,0 +1,101 @@
+---
+canonical: https://grafana.com/docs/alloy/latest/reference/cli/gql/
+description: Learn about the gql command
+labels:
+  stage: experimental
+  products:
+    - oss
+title: gql
+weight: 250
+---
+
+# `gql`
+
+{{< docs/shared lookup="stability/experimental_feature.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+The `gql` command runs a GraphQL query against the [{{< param "PRODUCT_NAME" >}} GraphQL API][GraphQL API].
+Use this command to query information from a running {{< param "PRODUCT_NAME" >}} instance, such as build metadata, readiness, and component health.
+
+The `gql` command doesn't enable the GraphQL API.
+The target {{< param "PRODUCT_NAME" >}} instance must run with the GraphQL API enabled.
+
+## Usage
+
+```shell
+alloy gql [<FLAG> ...] <QUERY>
+```
+
+Replace the following:
+
+* _`<FLAG>`_: One or more flags that define the GraphQL endpoint.
+* _`<QUERY>`_: Required. The GraphQL query to run.
+
+The `graphql` command is an alias for `gql`.
+
+## Enable the GraphQL API
+
+The GraphQL API is disabled by default.
+To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=experimental` and `--feature.graphql.enabled` flags:
+
+```shell
+alloy run --stability.level=experimental --feature.graphql.enabled <PATH_NAME>
+```
+
+Replace _`<PATH_NAME>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
+
+## Query formats
+
+The `gql` command accepts complete GraphQL queries.
+For example, you can run an anonymous query:
+
+```shell
+alloy gql '{ alloy { version isReady } }'
+```
+
+The command also accepts query body shorthand.
+When the query doesn't start with `{`, `query`, `mutation`, or `subscription`, the command wraps it in `query { ... }`.
+
+For example, the following command:
+
+```shell
+alloy gql 'alloy { version isReady }'
+```
+
+Runs this GraphQL query:
+
+```graphql
+query { alloy { version isReady } }
+```
+
+## Flags
+
+The following flags are supported:
+
+* `--server.graphql.endpoint`: Address of the GraphQL endpoint (default `"http://127.0.0.1:12345/graphql"`).
+
+## Examples
+
+To query build and readiness information from the default GraphQL endpoint, use the following command:
+
+```shell
+alloy gql 'alloy { isReady version }'
+```
+
+The output resembles the following:
+
+```json
+{
+  "alloy": {
+    "isReady": true,
+    "version": "v1.14.0"
+  }
+}
+```
+
+To query a GraphQL endpoint on a different address, use the `--server.graphql.endpoint` flag:
+
+```shell
+alloy gql --server.graphql.endpoint=http://localhost:12345/graphql 'components { id health { message } }'
+```
+
+[GraphQL API]: ../../http/graphql/
--- a/docs/sources/reference/cli/gql.md
+++ b/docs/sources/reference/cli/gql.md
@@ -38,7 +38,7 @@ The GraphQL API is disabled by default.
 To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=experimental` and `--feature.graphql.enabled` flags:
 
 ```shell
-alloy run --stability.level=experimental --feature.graphql.enabled <PATH_NAME>
+alloy run --stability.level=experimental --feature.graphql.enabled <CONFIG_PATH>
 ```
 
 Replace _`<PATH_NAME>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
@@ -71,7 +71,7 @@ query { alloy { version isReady } }
 
 The following flags are supported:
 
-* `--server.graphql.endpoint`: Address of the GraphQL endpoint (default `"http://127.0.0.1:12345/graphql"`).
+* `--endpoint`: Address of the GraphQL endpoint (default `"http://127.0.0.1:12345/graphql"`).
 
 ## Examples
 
@@ -85,17 +85,19 @@ The output resembles the following:
 
 ```json
 {
-  "alloy": {
-    "isReady": true,
-    "version": "v1.14.0"
+  "data": {
+    "alloy": {
+      "isReady": true,
+      "version": "v1.14.0"
+    }
   }
 }
 ```
 
-To query a GraphQL endpoint on a different address, use the `--server.graphql.endpoint` flag:
+To query a GraphQL endpoint on a different address, use the `--endpoint` flag:
 
 ```shell
-alloy gql --server.graphql.endpoint=http://localhost:12345/graphql 'components { id health { message } }'
+alloy gql --endpoint=http://localhost:12345/graphql 'components { id health { message } }'
 ```
 
 [GraphQL API]: ../../http/graphql/
--- a/docs/sources/reference/cli/gql.md
+++ b/docs/sources/reference/cli/gql.md
@@ -41,7 +41,7 @@ To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=exp
 alloy run --stability.level=experimental --feature.graphql.enabled <CONFIG_PATH>
 ```
 
-Replace _`<PATH_NAME>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
+Replace _`<CONFIG_PATH>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
 
 ## Query formats
 
--- a/internal/alloycli/cmd_gql.go
+++ b/internal/alloycli/cmd_gql.go
@@ -1,0 +1,74 @@
+package alloycli
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/grafana/alloy/internal/service/graphql/client"
+	"github.com/grafana/alloy/internal/service/graphql/utils"
+)
+
+type alloyGql struct {
+	httpAddr string
+}
+
+func gqlCommand() *cobra.Command {
+	g := &alloyGql{
+		httpAddr: "http://127.0.0.1:12345/graphql",
+	}
+
+	cmd := &cobra.Command{
+		Use:   "gql <query>",
+		Short: "[EXPERIMENTAL] Run a GraphQL query against a running Alloy instance",
+		Long: `The gql subcommand runs a GraphQL query against a running Alloy instance.
+The query is provided as a single argument to the command.
+
+It requires the --feature.graphql.enabled flag on the running Alloy instance to
+be set, as well as the stability.level flag set to "experimental".
+
+This command is experimental and may be modified or removed in the future. Use
+with caution in production.
+`,
+		Args:         cobra.ExactArgs(1),
+		SilenceUsage: true,
+		Aliases:      []string{"graphql"},
+		RunE: func(_ *cobra.Command, args []string) error {
+			return g.Run(args[0])
+		},
+	}
+
+	cmd.Flags().StringVar(
+		&g.httpAddr,
+		"server.graphql.endpoint",
+		g.httpAddr,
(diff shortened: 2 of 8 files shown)


docs/sources/reference/cli/gql.md
```diff
@@ -0,0 +1,101 @@
...
+
+The `graphql` command is an alias for `gql`.
+
+## Enable the GraphQL API
+
+The GraphQL API is disabled by default.
+To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=experimental` and `--feature.graphql.enabled` flags:
+
+```shell
+alloy run --stability.level=experimental --feature.graphql.enabled <PATH_NAME>
```
thampiotr: Instead of <PATH_NAME> I'd call it <CONFIG_PATH> so it's not ambiguous with something like URL path for gql like `/api/v1/gql`

docs/sources/reference/cli/gql.md
```diff
@@ -0,0 +1,101 @@
...
+
+```graphql
+query { alloy { version isReady } }
+```
+
+## Flags
+
+The following flags are supported:
+
+* `--server.graphql.endpoint`: Address of the GraphQL endpoint (default `"http://127.0.0.1:12345/graphql"`).
```
thampiotr: This flag is on a client side, so I'm not sure it should be in `server` namespace. Maybe `--endpoint` is enough as it's scoped to `gql` command. We needed namespaces on `run` command as there are so many flags.

thampiotr (commented):
Technically this looks good. I do wonder what use cases we are aiming to solve with this. Let me know if there is some existing design doc. Feeling a bit hesitant to add a solution without looking into alternatives and problem space too much.

docs/sources/reference/cli/gql.md
jharvey10: Good call! I renamed it to just `endpoint`.

jharvey10:
> Technically this looks good. I do wonder what use cases we are aiming to solve with this. Let me know if there is some existing design doc. Feeling a bit hesitant to add a solution without looking into alternatives and problem space too much.

@thampiotr To start, definitely the CLI-executable health/readiness checks, but for that alone this would be overkill. To mention some of the points from our discussion eariler:

- **Write once, run anywhere:** The graphql api can/will be used to drive this CLI utility as well as the UI and other agentic tooling without needing special handling for each of those mechanisms.
- **Reduced initial setup friction:** Prior to getting Alloy pipelines hooked up, this can be used to introspect Alloy in detail to understand how pipelines are wired, what arguments/exports are going in/out, what targets have been discovered and their statuses, etc. We have flexibility to expose high-cardinality data points beyond what we'd want to include in metrics/logs/traces being sent to the cloud.
- **Just-in-time debugging:** automated (and other) tooling can introspect the API and utilize any available data points to assist with failure diagnostics. This can also supplement (or in some cases replace) support bundle content, and can supplement metrics/logs/traces.

To fully realize these, there will need to be several more data points exposed on the api, but for now we just have a small starting set.

jharvey10:
@x1unix You mentioned maybe there's a way here that we could bake a readiness check into the docker image. Did you happen to have a link I could browse to take a look into that?

docs/sources/reference/cli/gql.md
```diff
@@ -0,0 +1,103 @@
...
+
+* _`<FLAG>`_: One or more flags that define the GraphQL endpoint.
+* _`<QUERY>`_: Required. The GraphQL query to run.
+
+The `graphql` command is an alias for `gql`.
+
+## Enable the GraphQL API
+
+The GraphQL API is disabled by default.
+To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=experimental` and `--feature.graphql.enabled` flags:
```
dehaansa, line 38: We have server.http.enable-pprof, should this be more similar to that? server.graphql.enable?

docs/sources/reference/cli/gql.md
```diff
@@ -0,0 +1,103 @@
...
+
+```graphql
+query { alloy { version isReady } }
+```
+
+## Flags
+
+The following flags are supported:
+
+* `--endpoint`: Address of the GraphQL endpoint (default `"http://127.0.0.1:12345/graphql"`).
```
dehaansa, line 74: Should we note that this can then be used to query remote Alloy instances if a non-local endpoint is used, not just Alloys on the local machine?

docs/sources/reference/cli/gql.md
```diff
@@ -0,0 +1,103 @@
...
+## Enable the GraphQL API
+
+The GraphQL API is disabled by default.
+To enable it, start {{< param "PRODUCT_NAME" >}} with the `--stability.level=experimental` and `--feature.graphql.enabled` flags:
+
+```shell
+alloy run --stability.level=experimental --feature.graphql.enabled <CONFIG_PATH>
+```
+
+Replace _`<PATH_NAME>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
```
thampiotr:
```suggestion
Replace _`<CONFIG_PATH>`_ with the {{< param "PRODUCT_NAME" >}} configuration file or directory path.
```

internal/alloycli/cmd_gql.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(response.Errors) > 0 {
+		return fmt.Errorf("GraphQL response contains errors")
+	}
+	return nil
+}
+
+func formatGraphQLQuery(query string) string {
+	trimmedQuery := strings.TrimSpace(query)
+
+	for _, prefix := range []string{
```
thampiotr: I'm reading that query($id: ID){...} or mutation($x:Int){...} are also possible and wouldn't be handled here. Maybe let's add them and have a table test?

internal/alloycli/cmd_gql.go
```diff
@@ -0,0 +1,88 @@
...
+		Aliases:      []string{"graphql"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return g.Run(args[0], cmd.OutOrStdout())
+		},
+	}
+
+	cmd.Flags().StringVar(
+		&g.httpAddr,
+		"endpoint",
+		g.httpAddr,
```
thampiotr: Let's rename `httpAddr` to `endpoint` too so it matches the flag and is overall better name I think (can contain URL path)

internal/service/graphql/client/graphql_client.go
thampiotr: Is this used anywhere?

thampiotr (commented):
Only minor stuff

internal/service/graphql/client/graphql_client.go
```diff
@@ -72,17 +75,22 @@ func (c *GraphQLClient) Execute(query string) (*GraphQLResponse, error) {
...
-		return nil, fmt.Errorf("failed to execute request: %s", resp.Status)
-	}
-
-	buf := new(bytes.Buffer)
-	_, err = buf.ReadFrom(resp.Body)
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodySize+1))
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %w", err)
 	}
+	if len(data) > maxResponseBodySize {
```
dehaansa, line 77: Why is the client defining a maximum response size?

internal/service/graphql/client/graphql_client_test.go
```diff
@@ -0,0 +1,56 @@
...
+		_, err := w.Write([]byte(`{"errors":[{"message":"bad query"}]}`))
+		require.NoError(t, err)
+	}))
+	defer server.Close()
+
+	c := NewGraphQLClient(server.URL)
+	_, err := c.Execute("{ alloy {")
+
+	require.Error(t, err)
+	require.ErrorContains(t, err, "400 Bad Request")
```
dehaansa, line 23: Why are we looking for two partial contains, rather than just the full expected error?

x1unix:
> @x1unix You mentioned maybe there's a way here that we could bake a readiness check into the docker image. Did you happen to have a link I could browse to take a look into that?

@jharvey10 Docker has `HEALTHCHECK` instruction: https://docs.docker.com/reference/dockerfile/#healthcheck

It's not as mature as K8s readiness vs liveness probes, but docker-compose can use it as a readiness probe.

So, if you have a container that depends on it, you can set `depends_on` and `condition: service_healthy` so compose will wait for a dependency to become healthy before starting a dependent container.

internal/alloycli/cmd_gql.go
jharvey10: Good call out! I definitely need to adjust this logic. I want to explicitly *not* allow parameterized queries like what you pasted. For one-shot, CLI-driven queries, they're really not that useful and they complicate the query parsing because you have to define an additional block with their values. I'll also update tests to be more comprehensive.

internal/service/graphql/client/graphql_client.go
jharvey10: No, not currently. More wanted to establish a pattern for this, but I also don't really anticipate any usages of this in the near future, so I'll remove it.

docs/sources/reference/cli/gql.md
jharvey10: I'm open to any options for these flags. We did merge them as part of the other PR and are just referencing them here, but neither has made it into a minor yet, so we could tweak them in a follow on PR. 

It's technically just another http endpoint, so it could even be like
```
--server.http.enable-graphql
```
and
```
--server.http.enable-graphql-playground
```

I'm not sure when it's more appropriate to use something like that versus the `feature.*` pattern we have for a few things.

docs/sources/reference/cli/gql.md
jharvey10: We *sort of* have this down below where we say:

> To query a GraphQL endpoint on a different address, use the `--endpoint` flag:
> 
> ```shell
> alloy gql --endpoint=http://localhost:12345/graphql 'components { id health { message } }'

but i could clarify it even further to be:

> To query a GraphQL endpoint on a different address (including other Alloy deployments), use the `--endpoint` flag:
> 
> ```shell
> alloy gql --endpoint=http://localhost:12345/graphql 'components { id health { message } }'

WDYT?

internal/service/graphql/client/graphql_client.go
jharvey10: This is to prevent resource exhaustion. If someone were to run a gql command on an alloy instance that was already unhealthy/borderline and cause it to receive a huge amount of data, it could make things even worse. Especially because the gql endpoint is configurable.

internal/service/graphql/client/graphql_client_test.go
jharvey10: Mostly a style thing. `400 Bad Request` asserts the http error, and `bad query` asserts the graphql error from the `errors` property in the response object, without caring about the exact format we get back surrounding each of those.

internal/alloycli/cmd_gql.go
jharvey10: Done! New logic should be more reliable and cover a lot more cases (governed by tests)

internal/service/graphql/client/graphql_client.go
jharvey10: Removed!

docs/sources/reference/cli/gql.md
thampiotr: I like `--server.http.enable-graphql`

thampiotr (approved):
From my pov - LGTM!

docs/sources/reference/cli/gql.md
jharvey10: 👍 Works for me too. I'll make that change across the existing infra and docs in a follow up PR

docs/sources/reference/cli/gql.md
jharvey10: follow on PR is up with this change

bvahdat:
Hi @jharvey10 

Thanks for this PR. I am now using [the image](https://hub.docker.com/layers/grafana/alloy/v1.17.0-rc.1) your PR is [part of it](https://github.com/grafana/alloy/pull/6111), inside my docker compose file with the command section as:

```
  alloy:
    image: docker.io/grafana/alloy:v1.17.0-rc.1
...
...
  command: run --server.http.listen-addr=0.0.0.0:12345 --storage.path=/var/lib/alloy --disable-reporting=true --stability.level=experimental --feature.graphql.enabled /etc/alloy/config.alloy
```

And while starting the container I get:

```
Error: unknown flag: --feature.graphql.enabled
2026/06/11 15:48:15 collector server run finished with error: unknown flag: --feature.graphql.enabled
```

Do you maybe have an idea?

Thanks in advance

jharvey10:
@bvahdat Sorry about that! We had a last-minute change of the feature flag from
`--feature.graphql.enabled`
to
`--server.http.enable-graphql` (and `--server.http.enable-graphql-playground`)

If you use that new flag it should get past that check.

Could you let me know if you're still seeing references to `feature.graphql.enabled` anywhere? I want to make sure those get updated.

bvahdat:
Thanks @jharvey10!

Looking at the [latest documentation](https://github.com/grafana/alloy/blob/main/docs/sources/reference/cli/gql.md) of the main branch revealed everything, and no stale documentation found.

PR #5695, opened 2026-03-02
HACKATHON: component level metrics

theSuess:
HACKATHON PROJECT, no review required, keeping this open as a draft :)

### Brief description of Pull Request

This PR adds the `destination` label to the metrics produced by the fanout processor. This can then be used to build live traffic graphs visualizing the data flow across components

### Notes to the Reviewer

HACKATHON PROJECT, no review required, keeping this open as a draft :)

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #6243, opened 2026-05-13
feat(telemetry): Add graph connection metrics

theSuess:

This PR adds graph connection info metrics that signify a connection in the component graph.

### Pull Request Details

Connection metrics tell the story of how components are connected. An example might look like this:
```
alloy_component_graph_connection{controller_id="",controller_path="/",from="prometheus.scrape.integrations_alloy_health",to="prometheus.relabel.integrations_alloy_health"} 1
```
This shows a component `prometheus.scrape.integrations_alloy_health` which provides data to `prometheus.relabel.integrations_alloy_health`. This could be telemetry data, discovery data or an auth reference.

The metric can be used to gather data flow between components by combining this with metrics for forwarded samples like this:

```
(alloy_component_graph_connection{} * on(from) group_left  label_replace(
        sum(rate(prometheus_forwarded_samples_total{}[$__rate_interval])) by (component_id),
        "from","$1","component_id","(.*)"
        ))
```

This PR takes the idea from #5695 and makes it maintainable & scaleable by reducing the number of required series (just one instead of one per export target) and not requiring passing down component IDs to all fanouts.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/troubleshoot/controller_metrics.md
+++ b/docs/sources/troubleshoot/controller_metrics.md
@@ -25,6 +25,7 @@ The controller exposes the following metrics:
 * `alloy_component_evaluation_seconds` (Histogram): The time it takes to evaluate components after one of their dependencies is updated.
 * `alloy_component_dependencies_wait_seconds` (Histogram): Time spent by components waiting to be evaluated after one of their dependencies is updated.
 * `alloy_component_evaluation_queue_size` (Gauge): The current number of component evaluations waiting to be performed.
+* `alloy_component_graph_connection` (Gauge): Indicates that two components are connected in the graph.
 
 [component controller]: ../../get-started/component_controller/
 [alloy run]: ../../reference/cli/run/
--- a/internal/runtime/internal/controller/loader.go
+++ b/internal/runtime/internal/controller/loader.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/grafana/dskit/backoff"
 	"github.com/hashicorp/go-multierror"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/storage"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
@@ -206,6 +207,7 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 	}()
 
 	l.cache.ClearModuleExports()
+	l.cm.graphEdgeConnection.Reset()
 
 	// Evaluate all the components.
 	_ = dag.WalkTopological(&newGraph, newGraph.Leaves(), func(n dag.Node) error {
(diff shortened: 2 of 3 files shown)


theSuess:
Tested this with fleet management as well and the metrics correctly appear & disappear when the config is modified

internal/runtime/internal/controller/loader.go
```diff
@@ -224,6 +226,14 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 		case ComponentNode:
 			components = append(components, n)
 			componentIDs[n.ID().String()] = n.ID()
+			outgoing := n.GetDataFlowEdgesTo()
+
+			for _, e := range outgoing {
+				l.cm.graphEdgeConnection.With(prometheus.Labels{
+					"from": n.ID().String(),
```
kalleep: nit: This will cause a allocation for each call and we use it in componentIDs too. Lets do this once for each ComponentNode and reuse it

kalleep (commented):
Looks good to me, just one nit

internal/runtime/internal/controller/loader.go
```diff
@@ -222,8 +224,17 @@ func (l *Loader) Apply(options ApplyOptions) diag.Diagnostics {
 
 		switch n := n.(type) {
 		case ComponentNode:
+			nodeID := n.ID()
```
kalleep:
```suggestion
			nodeID := n.ID().String()
```

kalleep: The string call is the one doing allocations

blewis12:
Since IIUC this will be automatically shipped now for any folks who are using meta-monitoring, do you reckon it makes sense to add a sentence or two about this somewhere in our docs, maybe [here](https://grafana.com/docs/alloy/latest/troubleshoot/component_metrics/)? I also don't _think_ these metrics would be kept for folks using the alloy health integration to ship metamonitoring info to the backend 🤔 but maybe that's not the intended outcome anyway, if this is only something for the FM side of things

theSuess:
Preallocated the string now and added docs in the controller metrics. Right now, these metrics are not included in the meta-monitoring set of metrics but this will change once we use them in product

docs/sources/troubleshoot/controller_metrics.md
```diff
@@ -25,6 +25,7 @@ The controller exposes the following metrics:
 * `alloy_component_evaluation_seconds` (Histogram): The time it takes to evaluate components after one of their dependencies is updated.
 * `alloy_component_dependencies_wait_seconds` (Histogram): Time spent by components waiting to be evaluated after one of their dependencies is updated.
 * `alloy_component_evaluation_queue_size` (Gauge): The current number of component evaluations waiting to be performed.
+* `alloy_component_graph_connection` (Info): Indicates that two components are connected in the graph.
```
kalleep:
```suggestion
* `alloy_component_graph_connection` (Gauge): Indicates that two components are connected in the graph.
```

theSuess: Technically it's a gauge but since the value isn't the important thing but rather the presence of the metric, IMHO info is clearer here

kalleep: It's mostly for consistency because all other metrics have the type included

kalleep (approved):

PR #6245, opened 2026-05-13
feat(telemetry): Add loki_forwarded_entries_total metric

theSuess:

This PR adds the `loki_forwarded_entries_total` metric with the same semantics as `prometheus_forwarded_samples_total`. It records the number of entries forwarded to the next components.

Partial continuation of #5695 but includes all current components

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/common/loki/consume_test.go
+++ b/internal/component/common/loki/consume_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/grafana/loki/pkg/push"
 	"github.com/stretchr/testify/require"
 )
@@ -13,7 +15,7 @@ import (
 func TestConsume(t *testing.T) {
 	consumer := NewLogsReceiver()
 	producer := NewLogsReceiver()
-	fanout := NewFanout([]LogsReceiver{consumer})
+	fanout := NewFanout([]LogsReceiver{consumer}, prometheus.DefaultRegisterer)
 
 	t.Run("should fanout any consumed entries", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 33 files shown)


theSuess:
Still need to update tests in some components, the fanout struct & general approach is ready for review though

kalleep:
Hey, we are in the middle of doing a big refactor of loki pipelines, and have a draft pr up that will remove this fanout https://github.com/grafana/alloy/pull/6204

I am expecting to merge that branch in a couple of days, have it running in internal dev cluster to verify everything. So depending on how fast you want to get this in we can either:
1. Wait for that pr to be merged and make this change after
2. Make this change now and I will resolve conflicts and make sure we have this metric in the new structure

Up to you :)

theSuess:
Thanks for the heads up @kalleep! I'll take a look at your PR tomorrow and can rebase this on top, maybe it'll simplify things :)

theSuess:
@kalleep I took a look and it looks like these changes are mostly parallel (but sadly touch the same call sites :sweat_smile:)

We _could_ use the new interceptors, but I think keeping this functionality in the fanout(consumer) is the cleanest solution and keeps consistency with prometheus & pyroscope

I guess the way forward depends on how fast you want to proceed with your PR but it's really up to you :shrug: I'm just a drive-by contributor :grimacing:

kalleep:
> @kalleep I took a look and it looks like these changes are mostly parallel (but sadly touch the same call sites 😅)
> 
> We _could_ use the new interceptors, but I think keeping this functionality in the fanout(consumer) is the cleanest solution and keeps consistency with prometheus & pyroscope
> 
> I guess the way forward depends on how fast you want to proceed with your PR but it's really up to you 🤷 I'm just a drive-by contributor 😬

Yes we should do this in fan-out for sure, interceptor is not always used and is more of a way to implement this interface for components. I want to get my pr in pretty soon but it will happen after v1.17 release, we are supposed to do rc today.

kalleep:
Hey @theSuess I am taking a different approach to migrate our pipelines so we could merge this pr if you want that so you don't have to wait.

PR #6246, opened 2026-05-13
test(loki.source.cloudflare): Stop mutating production defaultBackoff from tests

kgeckhart:

Three tests mutate the package-level `defaultBackoff` to retry fast, causing cross-test contamination and an ordering bug: `TestTailer_CloudflareTargetError`/`...168h` captures `defaultBackoff` into `cfg.Backoff` *before* the in-test mutation runs, so they only pass when `TestTailer_RetryErrorIterating` has already mutated the package var. Switch all three to a local `backoff.Config{MinBackoff: 0, MaxBackoff: 0, MaxRetries: 5}` literal (matching what `TestTailer_RetryErrorLogpullReceived` already does), so global state is not mutated from tests.

### DIFF
--- a/internal/component/loki/source/cloudflare/tailer_test.go
+++ b/internal/component/loki/source/cloudflare/tailer_test.go
@@ -171,17 +171,18 @@ func TestTailer_RetryErrorIterating(t *testing.T) {
 	getClient = func(apiKey, zoneID string, fields []string) (Client, error) {
 		return cfClient, nil
 	}
-	// retries as fast as possible.
-	defaultBackoff.MinBackoff = 0
-	defaultBackoff.MaxBackoff = 0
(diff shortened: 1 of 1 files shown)


internal/component/loki/source/cloudflare/tailer_test.go
```diff
@@ -284,7 +278,6 @@ func TestTailer_CloudflareTargetError168h(t *testing.T) {
 
 	ta, err := newTailer(newMetrics(prometheus.NewRegistry()), logger, handler.Receiver(), ps, cfg)
 	require.NoError(t, err)
-	require.True(t, ta.ready())
```
kgeckhart, line 287: This was also dropped because it's racy. With no backoff on retry it's possible we run all the retries before we end up verifying ready.

kalleep (approved):

PR #6251, opened 2026-05-13
refactor: Migrate from deprecated relabel.Process to relabel.ProcessBuilder

kgeckhart:

`relabel.Process` is the older form that allocates and returns a new `labels.Labels`; `relabel.ProcessBuilder` mutates a `*labels.Builder` in-place. Upstream Prometheus keeps `ProcessBuilder` going forward and will drop the free function. Both APIs exist on current main's `grafana/prometheus` fork, so this is a behavior-preserving migration that also skips the Builder allocation entirely in the no-rules case in `loki.relabel`.


kalleep:
Nice, I did this when I tried to update Prometheus in the latest otel update but reverted that prat because I had to roll back the promethues version anyway. Way better to do this as a separate pr!

internal/component/loki/relabel/relabel.go
```diff
@@ -250,7 +250,13 @@ func (c *Component) process(e loki.Entry) model.LabelSet {
 	}
 	c.builder.Sort()
 	lbls := c.builder.Labels()
-	lbls, _ = relabel.Process(lbls, c.rcs...)
+	if len(c.rcs) > 0 {
```
kalleep, line 253: In this component we re-use a `ScratchBuilder`. In my pr that refactors loki pipelines I had to switch to using a sync pool instead and I also changed to use `Builder` instead. Using `Builder` directly simplifies things and removes a bunch of conversions, see 
https://github.com/grafana/alloy/blob/6ea01728cd81f986959f65c660c3b13088fdf483/internal/component/loki/relabel/relabel.go#L257-L286

kalleep (commented):
You are also missing a bunch of places where we use the old api e.g. https://github.com/grafana/alloy/blob/main/internal/component/loki/source/gcplog/internal/gcplogtarget/formatter.go#L84

internal/component/loki/relabel/relabel.go
kgeckhart: I wasn't sure about this one, from what I understand scratchbuilder might be better for this use case because it's doing many O(1) appends and a final sort. The labelbuilder set is first going to iterate to see if the label is already pending in add before it appends and then calling Labels() incurs a similar sorting overhead.

kgeckhart: Ahh I also forgot, I chose not to go too crazy here because no worse and better if someone ends up using relabel with no configs (which would be really silly but I'm sure happens). 

Internally relabel.Process allocates a labels.NewBuilder even if there are no relabel configs.

kalleep: Good point, we should probably benchmark this

kalleep: Ran the benchmark we have for relabel on this branch,

With `labels.ScratchBuilder`:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/alloy/internal/component/loki/relabel
BenchmarkRelabelComponent-16    	  296055	      3944 ns/op	    1967 B/op	      31 allocs/op
```

With `labels.Builder`:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/alloy/internal/component/loki/relabel
BenchmarkRelabelComponent-16    	  325496	      3856 ns/op	    1558 B/op	      28 allocs/op
```

kgeckhart: Nice! Okay, so I think what we landed on is that it's not ideal but it's okay for this PR because it's functionally the same which seems ideal for this PR.

kalleep: We have also talked about storing label.Labels instead of model.LabelSet in the future. That would remove a bunch of these conversions in various places. But that will most likely only happen after the pipeline migration with batching is finished

kalleep (approved):

PR #6261, opened 2026-05-15
chore(loki.write): Migrate to slog

kalleep:
### Pull Request Details

Migrate loki.write and used structures to use slog.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


dehaansa (approved):
Looks straightforward. Only question would be if there's value in leaving in the debug logging, just commented out, for future use, but it can always be dug out of git history.

PR #6262, opened 2026-05-15
ci: Run OTel Engine OCB Generation As Pre-req For Alloy Build

blewis12:

A follow up to [this PR](https://github.com/grafana/alloy/pull/5547) with a bit of a simpler solution. This PR means that running `make alloy` will _also_ run the code generation steps for replacement generation + OCB. You won't need to manually run `generate-module-dependencies` and `generate-otel-collector-distro`. You can explicitly opt-out of the generation pre-req with `SKIP_CODE_GENERATION=1`

Previously, I tried to achieve this by ensuring that generation as pre-req to build only occurred locally. This is because we already check in generated code and so running the generation steps before build should be a no-op. However, doing that meant juggling lots of env variables to toggle the generation on/off. I think instead we can take the hit of generating code in CI, and having our `check` workflows ensuring that the generation _is_ in fact a no-op so that we can always know when generated code has been properly checked into version control.

The one thing we have to be careful of is ensuring that we _don't_ run our generation step in our packaging workflows. That's because our [build step](https://github.com/grafana/alloy/blob/e4aada8e67eed1af33c3a86a37158e0429bb7a77/.github/workflows/release-publish-alloy-artifacts.yml#L105) in that workflow runs parallel builds for each architecture - and our code generation _cannot_ be run concurrently. This PR separates out the generation step to be run before the parallel build, which explicitly disables the code generation

I've also extended the `check` workflows to also run on merge to main - so we can sanity check that the generated code checked in to the main branch is always in the correct state. Since we're now also generating code on CI, i'd rather err on the side of caution that the checked in code is in a valid state


Makefile
```diff
@@ -333,7 +336,16 @@ generate-module-dependencies:
 ifeq ($(USE_CONTAINER),1)
 	$(RERUN_IN_CONTAINER)
 else
-	$(GO_ENV) go run -C tools ./cmd generate module-dependencies --dependency-yaml=$(CURDIR)/dependency-replacements.yaml
```
blewis12, line 336: We don't need `GO_ENV` when running `module-dependencies` - in fact we need to empty `GOOS` and `GOARCH` as we do [here](https://github.com/grafana/alloy/blob/6430123c2d9bc6b6a53c86ba164344e09f1814ee/Makefile#L347) to avoid cross compiling in the CI, without doing this see CI failures

.github/workflows/release-publish-alloy-artifacts.yml
```diff
@@ -101,9 +101,12 @@ jobs:
         go-version-file: go.mod
         cache: false
 
-    - name: Build
+    - name: Generate source code
       run: |
-        RELEASE_BUILD=1 make -j4 dist
+        RELEASE_BUILD=1 make generate-source-code
```
dehaansa: What elements of the code generation are not committed and need to be built in the release process?

blewis12: ah, actually it doesn't. I got a bit carried away here and started to operate under the assumption that we _needed_ to run this and forgot the generated code has already been checked in, so this is a no-op. The only important thing to do here is ensure we skip generation when we run the parallel builds in the next step

blewis12: removed: https://github.com/grafana/alloy/pull/6262/changes/af28af5e0e14815b4a70bd752056bbea956b7b40

jharvey10 (approved):

PR #6268, opened 2026-05-15
revert(extension/alloyengine): Undo release-please tracking

jharvey10:

Reverts the changes from #5956 so that we can then eventually implement https://github.com/grafana/alloy/blob/main/docs/design/6173-alloyengine-module-structure.md


kgeckhart (commented):
Any reason to avoid fully reverting, https://github.com/grafana/alloy/pull/5956, and set the version back to v0.0.0-00010101000000-000000000000?

jharvey10:
> Any reason to avoid fully reverting, #5956, and set the version back to v0.0.0-00010101000000-000000000000?

I can do that. Shouldn't matter either way since the tags aren't present yet, so might as well do a full revert.

jharvey10:
After re-running the make scripts, there were some go mod/sum changes it picked up.

kgeckhart (approved):
> I can do that. Shouldn't matter either way since the tags aren't present yet, so might as well do a full revert.

Seeing the version in there just felt like noise that could be misleading. It _should_ be temporary but you never know.

PR #6271, opened 2026-05-18
fix(loki.process): No longer mutate rules in stage.truncate causing every config update to reload pipeline when this stage is used

kalleep:
### Pull Request Details

If `loki.process` contains `stage.truncate` it will reload the pipeline on every update even though the pipeline have not changed.

This happened because we mutated `RuleConfig` directly in validateTruncateStage function, setting default value for `SourceType` and setting `effectiveLimit`. When we later compared new config against old stored these could never match so pipeline is always reloaded. 

This pr will use `syntax.Defaulter` and `syntaxt.Validator` instead. We will catch invalid configs before we even try to build the pipeline.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/truncate.go
+++ b/internal/component/loki/process/stages/truncate.go
@@ -21,10 +21,11 @@ var (
 	truncateExtractedField          = "extracted"
 )
 
-const (
-	errLimitMustBeGreaterThanZero = "limit must be greater than zero"
-	errSourcesForLine             = "sources cannot be set when source_type is 'line'"
-	errAtLeastOneRule             = "at least one truncate rule must be defined"
+var (
+	errTruncateLimit          = errors.New("limit must be greater than zero")
+	errTruncateSourcesForLine = errors.New("sources cannot be set when source_type is 'line'")
+	errTruncateAtLeastOneRule = errors.New("at least one truncate rule must be defined")
+	errTruncateSuffixLength   = errors.New("suffix length cannot be greater than or equal to limit")
 )
 
 // TruncateConfig contains the configuration for a truncateStage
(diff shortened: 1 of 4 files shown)


internal/component/loki/process/stages/truncate.go
```diff
@@ -37,61 +37,46 @@ type RuleConfig struct {
...
 
-	effectiveLimit units.Base2Bytes
+// SetToDefault implements syntax.Defaulter.
+func (r *RuleConfig) SetToDefault() {
+	*r = RuleConfig{SourceType: SourceTypeLine}
 }
 
-// validateTruncateConfig validates the TruncateConfig for the truncateStage
-func validateTruncateConfig(cfg *TruncateConfig) error {
-	if len(cfg.Rules) == 0 {
```
blewis12, line 46: Previously we were validating on a `TruncateConfig` level and so could ensure that `len(cfg.Rules)` was above zero. Now that we're validating on the `RuleConfig` level, are we doing this length validation anymore?

internal/component/loki/process/stages/truncate.go
```diff
@@ -37,61 +37,46 @@ type RuleConfig struct {
...
-				return errors.New("suffix length cannot be greater than or equal to limit")
-			}
-
-			r.effectiveLimit -= units.Base2Bytes(len(r.Suffix))
-		}
+	if r.SourceType == SourceTypeLine && len(r.Sources) > 0 {
+		return errors.New(errSourcesForLine)
+	}
+	if len(r.Suffix) > 0 && int(r.Limit) <= len(r.Suffix) {
+		return errors.New("suffix length cannot be greater than or equal to limit")
```
blewis12: nit: should we also add this to the error consts for consistency?

internal/component/loki/process/stages/truncate.go
kalleep: This is already handled on alloy config level, rule block is required and you get an error if none is defined. We also [have a test for it ](https://github.com/grafana/alloy/blob/8edf2e447b34f9abba5a7cb88658b6cb3c1a7a80/internal/component/loki/process/stages/truncate_test.go#L259-L263)

internal/component/loki/process/stages/truncate.go
kalleep: Fixed in [f221d22](https://github.com/grafana/alloy/pull/6271/commits/f221d22cad8e32844456def510e3c8fa5fbf1006)

I also made them package level errors instead of strings that we create errors from.

internal/component/loki/process/stages/truncate.go
blewis12: ah ok nice, didn't catch that

blewis12 (dismissed):

blewis12 (approved):

PR #6273, opened 2026-05-18
fix(database_observability.postgres): Bound wait_event wait_time across identity shifts

gaantunes:
## Summary

This PR improves the accuracy and consistency of wait event timing and labeling in the Postgres query samples collector. The main focus is ensuring that wait event durations are correctly bounded by both the state change and the time of the prior sample, which addresses over-counting in cases where the same wait event occurs multiple times.

### `wait_time` inflation across wait_event identity shifts

Each `WaitEventOccurrence` stored a single `LastWaitTime` that grew monotonically (`now − state_change`) every scrape and was emitted once at finalize. Because PG keeps `pg_stat_activity.state_change` anchored on the `state` column transition (`idle`/`active`), **not** on the `wait_event` column, every secondary wait identity in a multi-wait query re-credits the full prior state duration. The visible symptom: `sum_over_time({op="wait_event"} | unwrap duration_seconds(wait_time))` shows roughly 2× the real wait time for any query that passes through multiple wait events.

This PR keeps the existing emit cadence — **one `wait_event` log per occurrence at finalize**, so the existing wait-event count viz is unaffected — and fixes the math.

#### The fix

Each `WaitEventOccurrence` gets a `waitStartedAt` field, set once when the occurrence opens to:

```
waitStartedAt = max(state_change, priorLastSeen-when-this-occurrence-opened)
```

`priorLastSeen` is the `LastSeenAt` value on the `SampleState` before it gets overwritten by the current scrape — i.e., the timestamp of the prior scrape that observed this same sample. `upsertActiveSample` captures it and passes it into `upsertWaitEvent`.

At finalize, the wait-event loop in `emitAndDeleteSample` emits:

```
wait_time = LastTimestamp − waitStartedAt
```

For the first occurrence on a brand-new sample, `priorLastSeen` is zero, so `waitStartedAt = state_change` — same as before. For any secondary occurrence (whether the prior wait ended via an identity change or via a no-wait scrape), `waitStartedAt = priorLastSeen`, which cancels out the prior occurrence's window.

#### Effect on a multi-wait scenario

90s query, Lock 5–45s, IO 65–85s (real PG: `state_change=0` throughout), scrape every 10s starting T=7:

| | Pre-fix `wait_time` | This PR |
|---|---|---|
| Lock A | 37s | 37s |
| IO B | **77s** — re-credits A's full window | **20s** |
| Sum (real = 60s) | **114s** (~2× real) | **57s** (within 3s tail-loss) |

#### Tradeoffs worth flagging for reviewers

- **Multi-wait collapse on the timestamp axis.** All `wait_event` entries for one query share the sample's `query_start`. Distinguishability moves from the timestamp into the `wait_event_type`, `wait_event`, and `blocked_by_pids` labels. Sufficient for the trace-correlation use case driving the change; consumers that previously relied on the timestamp to order waits within a query need to switch to the labels.

### One-shot / pre-existing-wait observations

Brand-new occurrence on a brand-new sample uses `state_change` as the anchor, unchanged from before. A query observed on a single scrape still emits the full state-anchored duration; a backend that was already waiting when alloy started credits its historical wait time on first emit. This matches the prior behavior and is intentional — bounding by `priorLastSeen` would otherwise drop short-lived one-shot observations under typical 30–60s scrape intervals.

## Test plan

- [x] `go test -race -count=1 -tags=nodocker ./internal/component/database_observability/postgres/...` — green
- [x] `go vet ./internal/component/database_observability/postgres/...` — clean
- [x] `TestQuerySamples_FinalizationScenarios` subtests updated with the new wait_time values; multi-wait `wait-event_starts_new_occurrence_on_set_change` asserts A=5s, B=3s instead of A=5s, B=8s (telescoped), and pins all three entries to the sample's `query_start` timestamp.
- [x] `TestQuerySamples_WaitEventBoundedByPriorScrape` pins the priorLastSeen bound: existing sample, state_change 1h in the past, scrape gap of 5s → `wait_time="5s"`; both entries stamp at `query_start` (1h in the past), not the scrape time.
- [x] `TestQuerySamples_FetchQuerySamples` updated with `expectedTimestamps` per case; "query with wait event" exercises the collapse (sample + wait_event entries both at `queryStartTime`).
- [x] `TestQuerySamples_IdleScenarios` (3 subtests) updated: expected timestamp `stateChangeTime` → `queryStartTime`.
- [x] Verified on local playground (image `dbo11y-11203710b` running in `db-o11y-playground`); Loki shows expected per-occurrence wait_time values, no inflation.


internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -489,21 +509,25 @@ func (c *QuerySamples) emitAndDeleteSample(key SampleKey) {
...
 			continue
 		}
+		waitTime := we.LastTimestamp.Sub(we.boundedStart)
+		if waitTime < 0 {
+			waitTime = 0
+		}
+		waitTimeStr := waitTime.String()
+
+		op := OP_WAIT_EVENT
+		labels := c.buildWaitEventLabels(state, we, waitTimeStr)
```
cristiangreco: nit: can we have an if/else here again? For `OP_WAIT_EVENT_V2` we end up building the `labels` twice otherwise.

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -196,14 +196,20 @@ func (t *WaitEventTracker) CloseOpen()                        { t.openIdx = -1 }
...
-	BlockedByPIDs []int64 // normalized set (sorted, unique)
-	LastWaitTime  string  // last stateDuration seen for this wait event
-	LastState     string
-	LastTimestamp time.Time
+	WaitEventType   string
+	WaitEvent       string
+	BlockedByPIDs   []int64 // normalized set (sorted, unique)
+	LastState       string
+	LastTimestamp   time.Time // scrape time of the most recent observation
+	FirstObservedAt time.Time // first scrape that saw this occurrence
```
cristiangreco: Seems that `FirstObservedAt` is unused now?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -196,14 +196,20 @@ func (t *WaitEventTracker) CloseOpen()                        { t.openIdx = -1 }
...
+	boundedStart    time.Time // wait_time anchor: max(state_change, priorLastSeen at first observation)
```
cristiangreco: nit: wondering if `waitStartedAt` could convey better the concept?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -418,19 +424,22 @@ func (c *QuerySamples) validateQuerySample(sample QuerySamplesInfo) error {
...
-	state, ok := c.samples[key]
-	if !ok {
+	state, existed := c.samples[key]
+	var priorLastSeen time.Time
+	if existed {
+		priorLastSeen = state.LastSeenAt
+	} else {
 		state = &SampleState{tracker: newWaitEventTracker()}
 		c.samples[key] = state
 	}
```
cristiangreco, line 436: Seems that this if/else branch is mixing two different thing? Maybe could be made a bit more explicit with:

```go
state, existed := c.samples[key]
if !existed {
    state = &SampleState{tracker: newWaitEventTracker()}
    c.samples[key] = state
}

var priorLastSeen time.Time
if existed {
    priorLastSeen = state.LastSeenAt
}
```

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Done in 0da681e7b808e20f11a2646e8e8c80ef4c4394fd — switched to if/else so labels build only once.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Removed in 0da681e7b808e20f11a2646e8e8c80ef4c4394fd — it was a leftover from when I was leaving a hook for #6214's counter, but it's dead in this PR.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Done in 0da681e7b808e20f11a2646e8e8c80ef4c4394fd — renamed to waitStartedAt.

internal/component/database_observability/postgres/collector/query_samples.go
gaantunes: Done in 0da681e7b808e20f11a2646e8e8c80ef4c4394fd — split into two blocks as suggested.

fridgepoet:
Regarding https://github.com/grafana/alloy/pull/6273/commits/3187dbff74e3b2d26efdfd9e5d7d0ea0bf7b5d1d: 
I’d recommend keeping the timestamp changes in a separate PR. 

I noticed the timestamp assertion was folded into part of a larger test. I’d recommend a dedicated test for that behavior so it is easier to maintain/debug are easier to isolate and reason about.

cristiangreco (dismissed):

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -474,6 +496,9 @@ func (c *QuerySamples) emitAndDeleteSample(key SampleKey) {
 		return
 	}
 	sampleLabels := c.buildQuerySampleLabelsWithEnd(state, state.EndAt)
+	// Anchor on query_start so the entry timestamp lines up with the
+	// trace span's start on the timeline; application traces stamp their
+	// span at query start, and the DB O11y app navigates to that point.
```
fridgepoet:
```suggestion
```

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -196,14 +197,19 @@ func (t *WaitEventTracker) CloseOpen()                        { t.openIdx = -1 }
...
+// state duration).
 type WaitEventOccurrence struct {
 	WaitEventType string
 	WaitEvent     string
 	BlockedByPIDs []int64 // normalized set (sorted, unique)
-	LastWaitTime  string  // last stateDuration seen for this wait event
 	LastState     string
-	LastTimestamp time.Time
+	LastTimestamp time.Time // scrape time of the most recent observation
+	waitStartedAt time.Time // wait_time anchor: max(state_change, priorLastSeen at first observation)
```
fridgepoet: Should we remove these explanatory comments? What do you think?

internal/component/database_observability/postgres/collector/query_samples.go
```diff
@@ -196,14 +197,19 @@ func (t *WaitEventTracker) CloseOpen()                        { t.openIdx = -1 }
 func (t *WaitEventTracker) WaitEvents() []WaitEventOccurrence { return t.waitEvents }
 
 // WaitEventOccurrence tracks a continuous occurrence of the same wait event
-// with the same blocked_by_pids set.
+// with the same blocked_by_pids set. wait_time at emit time is
+// LastTimestamp - waitStartedAt; waitStartedAt is anchored on state_change
+// but offset to priorLastSeen when a prior wait identity ended before this
+// one began (PG's state_change is stable across wait_event changes, so
+// without the offset every secondary occurrence would re-credit the full
+// state duration).
```
fridgepoet, line 205: Should we remove any verbose explanatory comments? What do you think?

internal/component/database_observability/postgres/collector/query_samples_test.go
```diff
@@ -117,7 +117,8 @@ func TestQuerySamples_FetchQuerySamples(t *testing.T) {
 				`level="info" datname="testdb" pid="102" leader_pid="" user="testuser" app="testapp" client="127.0.0.1:5432" backend_type="client backend" state="waiting" xid="0" xmin="0" xact_time="2m0s" query_time="30s" queryid="124"`,
 				`level="info" datname="testdb" pid="102" leader_pid="" user="testuser" backend_type="client backend" state="waiting" xid="0" xmin="0" wait_time="10s" wait_event_type="Lock" wait_event="relation" wait_event_name="Lock:relation" blocked_by_pids="[103 104]" queryid="124"`,
 			},
-			expectedTimestamps: []time.Time{queryStartTime, now},
+			// Both sample and wait_event collapse to query_start.
+			expectedTimestamps: []time.Time{queryStartTime, queryStartTime},
```
fridgepoet, line 121: This change might be a leftover from the log timestamp change?

internal/component/database_observability/postgres/collector/query_samples.go
cristiangreco:
```suggestion
	LastTimestamp time.Time
	waitStartedAt time.Time
```

internal/component/database_observability/postgres/collector/query_samples.go
cristiangreco: I think this one is worth keeping though? It describes how `state_change` works upstream and how we use it in computations, which I personally found useful in this PR.

internal/component/database_observability/postgres/collector/query_samples_test.go
cristiangreco: I think this is still needed here

fridgepoet (approved):

PR #6274, opened 2026-05-18
fix(database_observability.postgres): Count error logs when log_timezone is non-UTC

gaantunes:
## Summary

- Fixes `database_observability_pg_errors_total` returning no data when PostgreSQL is configured with a non-UTC `log_timezone` (e.g. `America/Los_Angeles`). Go's `time.Parse` fabricates a zero-offset `Location` for any timezone abbreviation it doesn't recognize (PST, PDT, EDT, AEDT, ...), so the collector's historical-log filter saw every live entry as hours in the past and silently dropped it.
- The collector now queries `log_timezone` from `pg_settings` (with a 1-hour refresh, since the GUC is `sighup`-reloadable) and uses the resolved `*time.Location` to reconstruct each log line's wall-clock into a real UTC instant. The historical filter is preserved for all IANA configurations.
- Layered safety net: if the cached Location's abbreviation disagrees with the log line's (stale cache between a `pg_reload_conf` and the next refresh), or `log_timezone` is a POSIX spec Go's tzdata can't load (`EST5EDT,M3.2.0,M11.1.0`, `<+05>5`, ...), the collector falls back to skipping the historical filter — live errors are still counted instead of being mis-classified as historical. A one-shot warn (deduped per distinct value) tells operators to switch to an IANA name.

Refs: grafana/grafana-dbo11y-app#2877

## Test plan

- [x] New unit tests in `logs_test.go` covering: bug repro (no resolver, PST log counted), reconstruction counts recent non-UTC log with filter preserved, reconstruction filters historical non-UTC log, abbrev mismatch falls back to counting.
- [x] `go test -race -tags=nodocker ./internal/component/database_observability/...` — full suite green.
- [x] `gofmt -d` clean, `go vet` clean.
- [x] Manual verification against the reproducer in grafana/grafana-dbo11y-app#2877 (PostgreSQL 15 + `log_timezone='America/Los_Angeles'`, confirm `database_observability_pg_errors_total` increments).

### DIFF
--- a/internal/component/database_observability/postgres/collector/logs.go
+++ b/internal/component/database_observability/postgres/collector/logs.go
@@ -2,6 +2,7 @@ package collector
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"log/slog"
 	"regexp"
@@ -19,7 +20,12 @@ import (
 const (
 	LogsCollector         = "logs"
 	expectedLogLinePrefix = "%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a"
+	selectLogTimezone     = `SELECT setting FROM pg_settings WHERE name = 'log_timezone';`
 )
+
+// log_timezone has context=sighup, so operators can change it without a
+// restart; re-read on a tick so we stay in sync.
+const logTimezoneRefreshInterval = time.Minute
 
 // Postgres log format regex
 var logFormatRegex = regexp.MustCompile(
@@ -44,9 +50,10 @@ type LogsArguments struct {
 	Registry         *prometheus.Registry
 	ExcludeDatabases []string
 	ExcludeUsers     []string
-	// LogTimezone returns PostgreSQL's configured log_timezone as a Location,
-	// or nil if it is unknown. Called on every log line, so must be cheap.
-	LogTimezone func() *time.Location
+	// DB, if non-nil, is polled for PostgreSQL's log_timezone setting so the
+	// historical-log filter can resolve non-UTC abbreviations. When nil, the
+	// collector skips the filter for ambiguous timezones (still counts the log).
+	DB *sql.DB
 }
 
 type Logs struct {
@@ -57,7 +64,10 @@ type Logs struct {
 	receiver         loki.LogsReceiver
 	excludeDatabases []string
 	excludeUsers     []string
-	getLogTimezone   func() *time.Location
+
+	db                     *sql.DB
+	logTimezone            atomic.Pointer[time.Location]
+	lastLogTimezoneWarning atomic.Pointer[string] // dedupes unparseable-log_timezone warns across refresh ticks
 
 	errorsBySQLState *prometheus.CounterVec
 	parseErrors      prometheus.Counter
@@ -78,11 +88,6 @@ type Logs struct {
 func NewLogs(args LogsArguments) (*Logs, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 
-	getLogTimezone := args.LogTimezone
-	if getLogTimezone == nil {
-		getLogTimezone = func() *time.Location { return nil }
-	}
-
 	l := &Logs{
 		logger:           args.Logger.With("collector", LogsCollector),
 		entryHandler:     args.EntryHandler,
@@ -90,7 +95,7 @@ func NewLogs(args LogsArguments) (*Logs, error) {
 		receiver:         args.Receiver,
 		excludeDatabases: args.ExcludeDatabases,
 		excludeUsers:     args.ExcludeUsers,
-		getLogTimezone:   getLogTimezone,
+		db:               args.DB,
 		ctx:              ctx,
 		cancel:           cancel,
 		stopped:          atomic.NewBool(false),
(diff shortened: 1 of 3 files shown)


internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -20,8 +21,12 @@ import (
 const (
 	LogsCollector         = "logs"
 	expectedLogLinePrefix = "%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a"
+	selectLogTimezone     = `SELECT setting FROM pg_settings WHERE name = 'log_timezone';`
 )
 
+// log_timezone is sighup-reloadable; poll for changes.
+const logTimezoneRefreshInterval = time.Minute
```
cristiangreco: 1m might be a bit too much? In other places where we run these "background checks" we usually set 1h.

(I understand that since it's a live-reloadable setting we want to react asap, but I guess it's not the kind of settings that is modified often)

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -56,6 +63,10 @@ type Logs struct {
 	excludeDatabases []string
 	excludeUsers     []string
 
+	db                     *sql.DB
+	logTimezone            atomic.Pointer[time.Location]
+	lastLogTimezoneWarning atomic.Pointer[string]
```
cristiangreco: nit: this could be just `lastLogTimezone`, as we're storing the value, not the warning message

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -129,11 +141,53 @@ func (l *Logs) Receiver() loki.LogsReceiver {
...
+			l.refreshLogTimezone(l.ctx)
+		}
+	}
+}
+
+func (l *Logs) refreshLogTimezone(ctx context.Context) {
+	queryCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	var name string
```
cristiangreco: nit: can we call this `tz` or `tzName` or similar?

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -129,11 +141,53 @@ func (l *Logs) Receiver() loki.LogsReceiver {
...
+	if err := l.db.QueryRowContext(queryCtx, selectLogTimezone).Scan(&name); err != nil {
+		level.Debug(l.logger).Log("msg", "failed to query log_timezone", "err", err)
+		l.logTimezone.Store(nil)
+		return
```
cristiangreco: In case of transient query failure we wipe `l.logTimezone`, if any. Maybe on query error we should keep the previous value around?

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Done — bumped refresh interval from 1m to 1h. Agreed log_timezone changes are rare, and the abbrev-mismatch fallback in resolveAbsolute keeps live errors counted during the gap.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Done — renamed to lastLogTimezone.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Done — renamed local to tzName.

internal/component/database_observability/postgres/collector/logs.go
gaantunes: Good catch — fixed. Query failure now leaves any previously cached Location in place (transient blip → no behavior change). Only LoadLocation failure (PG returned a value Go can't load → cache is now stale by definition) clears it.

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -45,6 +50,8 @@ type LogsArguments struct {
 	Registry         *prometheus.Registry
 	ExcludeDatabases []string
 	ExcludeUsers     []string
+	// DB, if non-nil, is polled for log_timezone to resolve non-UTC log abbreviations.
+	DB *sql.DB
```
cristiangreco:
```suggestion
	DB               *sql.DB
```

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -20,8 +21,12 @@ import (
 const (
 	LogsCollector         = "logs"
 	expectedLogLinePrefix = "%m:%r:%u@%d:[%p]:%l:%e:%s:%v:%x:%c:%q%a"
+	selectLogTimezone     = `SELECT setting FROM pg_settings WHERE name = 'log_timezone';`
 )
 
+// log_timezone is sighup-reloadable; poll for changes.
```
cristiangreco:
```suggestion
// log_timezone is a sighup-reloadable Postgres setting so we periodically poll for changes
```

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -129,11 +141,54 @@ func (l *Logs) Receiver() loki.LogsReceiver {
...
+func (l *Logs) refreshLogTimezone(ctx context.Context) {
+	queryCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	var tzName string
+	if err := l.db.QueryRowContext(queryCtx, selectLogTimezone).Scan(&tzName); err != nil {
+		// Keep any previously cached Location: a transient query failure
+		// doesn't mean PG's setting changed.
+		level.Debug(l.logger).Log("msg", "failed to query log_timezone", "err", err)
+		return
```
cristiangreco, line 171:
```suggestion
	if err := l.db.QueryRowContext(queryCtx, selectLogTimezone).Scan(&tzName); err != nil {
		level.Debug(l.logger).Log("msg", "failed to query log_timezone", "err", err)
		return
```

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -128,11 +139,52 @@ func (l *Logs) Receiver() loki.LogsReceiver {
...
+	}
+}
+
+func (l *Logs) refreshLogTimezone(ctx context.Context) {
+	queryCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	var tzName string
+	if err := l.db.QueryRowContext(queryCtx, selectLogTimezone).Scan(&tzName); err != nil {
+		level.Debug(l.logger).Log("msg", "failed to query log_timezone", "err", err)
```
cristiangreco:
```suggestion
		l.logger.Debug("failed to query log_timezone", "err", err)
```

internal/component/database_observability/postgres/collector/logs.go
```diff
@@ -128,11 +139,52 @@ func (l *Logs) Receiver() loki.LogsReceiver {
...
+		return
+	}
+
+	loc, err := time.LoadLocation(tzName)
+	if err != nil {
+		// PG accepts POSIX specs (e.g. 'EST5EDT,M3.2.0,M11.1.0') that Go's tzdata can't load.
+		if prev := l.lastLogTimezone.Load(); prev == nil || *prev != tzName {
+			level.Warn(l.logger).Log("msg", "PostgreSQL log_timezone is not a Go-loadable IANA name; logs collector will skip its historical-log filter. Consider setting log_timezone to an IANA name (e.g. 'America/New_York') in postgresql.conf.", "log_timezone", tzName, "err", err)
```
cristiangreco:
```suggestion
			l.logger.Warn("PostgreSQL log_timezone is not a Go-loadable IANA name; logs collector will skip its historical-log filter. Consider setting log_timezone to an IANA name (e.g. 'America/New_York') in postgresql.conf.", "log_timezone", tzName, "err", err)
```

fridgepoet (dismissed):
I'd suggest replacing the time.Sleep in the tests. Also it might be a good idea to exhaustively add unit tests to resolveAbsolute, both for assertions and because it might be nice to illustrate/document through tests how it works.

fridgepoet (approved):

fridgepoet:
Nice, hope we can add unit tests to resolveAbsolute sometime in a follow-up if we think it could be useful.

cristiangreco:
> Nice, hope we can add unit tests to resolveAbsolute sometime in a follow-up if we think it could be useful.

It's implicitly covered, but agree we could have more specific tests

PR #6276, opened 2026-05-18
ci: Switch workflows to grafana/create-github-app-token

### LINKED ISSUE #4727: Update workflows from actions/create-github-app-token to grafana's
https://github.com/grafana/shared-workflows/tree/main/actions/create-github-app-token

relevant slack thread:
https://raintank-corp.slack.com/archives/CKF4X4H1D/p1761667491336099

jharvey10:
Closes #4727

### DIFF
--- a/.github/workflows/bump-formula-pr.yml
+++ b/.github/workflows/bump-formula-pr.yml
@@ -3,8 +3,10 @@ on:
   release:
     types: [released]
 
-permissions:
-  contents: read
+permissions: {}
+
+env:
+  ALLOYBOT_APP_NAME: grafana-alloybot
 
 jobs:
   homebrew-grafana:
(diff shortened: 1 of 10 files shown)


.github/workflows/bump-formula-pr.yml
```diff
@@ -3,43 +3,34 @@ on:
...
     # These permissions are needed to assume roles from Github's OIDC.
     permissions:
-      contents: read
       id-token: write
     steps:
     # TODO: Remove this when we no longer need a forked action in the "Update Homebrew formula" step.
+    - uses: grafana/shared-workflows/actions/create-github-app-token@259ba21cb3ff07724f331e26d926d655d24b317b # create-github-app-token/v0.2.3
+      id: app-token
+      with:
+        github_app: ${{ env.ALLOYBOT_APP_NAME }}
```
blewis12, line 23: I know this action doesn't work anyway so not really a blocker, but I think we'd need to specify a [permission_set](https://github.com/grafana/shared-workflows/blob/7ae85085c0bed854b41785ba266604b08e50abe0/actions/create-github-app-token/action.yaml#L4) here - we could probably just do the minimal permission set of [alloy](https://github.com/grafana/deployment_tools/blob/7b6293cc91ebe8985554f72d94d075ac2d62e5c5/terraform/repositories/alloy/github-app-configs/config.yaml#L95) for now - happy to add the actual fixing and refactoring of this workflow to the follow up items

blewis12:
Just one comment - then @jharvey10 if you're happy to move this out of draft I can approve

.github/workflows/bump-formula-pr.yml
jharvey10: Done!

kgeckhart (dismissed):

.github/workflows/bump-formula-pr.yml
```diff
@@ -3,43 +3,35 @@ on:
...
+        permission_set: alloy
```
kgeckhart, line 24: I think we were previously using the homebrew-grafana permission + alloy permission right? Do we need both or just the one?

jharvey10: So at the moment, this whole workflow is broken 😆 so this is essentially a minimally permissive change to get us moved over to the token broker, but this whole workflow needs to be revamped. I believe this would mean we'd need two separate token requests which are used in corresponding parts in the rest of the workflow, but if a rewrite happens, that might not even be necessary if we simplify things.

blewis12 (approved):

kalleep (approved):

PR #6280, opened 2026-05-19
ci: Harden PR workflows

thampiotr:

Two small hardening changes to PR-triggered workflows so a single missed manual approval cannot leak secrets or escalate token scope.

### Pull Request Details

**`ai-dependency-review.yml`** — checkout now pins to `${{ github.event.pull_request.base.sha }}` instead of relying on the default `pull_request` checkout ref (`refs/pull/<N>/merge`, which contains the PR author's code merged onto the base).

**`release-lint-pr-title.yml`** — removes the unused `id-token: write` permission. The only step in the workflow is `amannn/action-semantic-pull-request`, which does not use OIDC. 

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.github/workflows/ai-dependency-review.yml
+++ b/.github/workflows/ai-dependency-review.yml
@@ -20,15 +20,10 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     steps:
-      # SECURITY: Check out the PR's base SHA (trusted code already merged to the
-      # target branch), not the PR head/merge ref. The tool below runs with
-      # OPENAI_API_KEY and a write-scoped GITHUB_TOKEN in its environment, so it
-      # must not execute any code or read any prompt from the PR. The PR's diff
-      # itself is fetched via the GitHub API inside the tool (see
-      # tools/ai-review/github.go: getPRDiff), so we never need PR files on disk.
-      - name: Checkout base ref (trusted code only)
+      - name: Checkout base ref
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          # Use the trusted base SHA so the tooling we run is not from the PR.
           ref: ${{ github.event.pull_request.base.sha }}
           persist-credentials: false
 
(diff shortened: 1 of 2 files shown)


kgeckhart (dismissed):

.github/workflows/ai-dependency-review.yml
```diff
@@ -20,9 +20,11 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
+      - name: Checkout code (base ref)
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          # Use the trusted base SHA so the tooling we run is not from the PR.
+          ref: ${{ github.event.pull_request.base.sha }}
```
jharvey10: I understand the intent here, but I'm not sure it's actually effective. There's a couple potential issues:

1. If we do this, we lose a way to test the script and prompt file used by this workflow in the context of a 'real' PR. They'd have to either already be present on the base branch or be tested in on a dev's machine.
2. The vault secret fed to the `ai-review` script is not present on forks. Even if the workflow runs, the request will not succeed.
3. Since this is using `pull_request` as the trigger, an attack would have to come from a branch within the Alloy repo and not a fork. In that case, this workflow file could simply be changed in the same PR to remove this restriction. For a fork, it runs in the source repo's security context anyway, so there's no exposure.

The danger would be if we left the checkout alone, but then used the dangerous `pull_request_target` trigger.

thampiotr: Indeed, I'm aware you can remove the `ref:` or change it if needed. But it would make it pretty obvious that workflows were changed if we get a PR like that. On workflow changes we expect to implement more thorough review, or limit them to only PRs authored by maintainers. 

What this prevents is someone sneaking a change into the ai-review tool and being able to get it running when we approve their PR checks without noticing that the tool was changed.

jharvey10: > when we approve their PR checks without noticing that the tool was changed.

From a fork-based PR, this workflow will indeed run, but it will always fail by design. The Vault retrieval action will fail first, which prevents secret material from being passed to any later step. This workflow is (by design) not applicable to fork-based PRs due to this exact type of thing: prompt injection or other risks.

kalleep (approved):

jharvey10 (approved):

PR #6281, opened 2026-05-19
feat(otelcol.exporter.prometheus): Convert classic histograms to NHCB

### LINKED ISSUE #6183: Implement convert_classic_histograms_to_nhcb in the otel collector.
### Component(s)

otelcol.exporter.prometheus

### Request

Prometheus has the ability to convert incoming histograms from the bloated classic format to native histograms with pre-defined buckets (nhcb) via the prometheus.scrape functionality. The same conversion isn't possible with histogram data coming from otel collectors. While in both cases there is the possibility of modifying the data source to send native/exponential histograms, that doesn't help if the data source cannot be quickly modified, if it can be at all.

Given the conversion logic already exists, it would be helpful to be able to funnel otel histograms through the convert_classic_histograms_to_nhcb logic, thus saving on data transfer and storage costs on the back end.

Another option would be to put the logic into a new processor (ex: otelcol.processor.convert_nhcb)

### Use case

We need to scrape many legacy otel data sources that we do not have direct control over. Classic histograms are large and expensive to transmit, process, and store. Allowing for a conversion/compression at the collector level gives control back to the team/user using alloy rather than relying on upstream legacy sources to update.

### Tip

madaraszg-tulip:

Add an experimental `convert_classic_histograms_to_nhcb` argument that turns each OTel explicit-bucket histogram data point into a single Prometheus native histogram with custom buckets, instead of expanding into _bucket, _sum, and _count series. Gated behind the experimental stability level.

Fixes #6183

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
+++ b/docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
@@ -62,17 +62,19 @@ The following components are compatible:
 * `prometheus.remote_write` only when configured for Remote Write v2.
 * `prometheus.write_queue`
 
-[experimental]: ../../../get-started/configuration-syntax/expressions/function_calls/#experimental-functions
+[experimental]: https://grafana.com/docs/release-life-cycle/
 {{< /admonition >}}
 
 {{< admonition type="warning" >}}
 **EXPERIMENTAL**: The `convert_classic_histograms_to_nhcb` argument is an [experimental][] feature.
 
 When set to `true`, each OTel explicit-bucket histogram data point is converted to a single Prometheus native histogram with custom buckets (NHCB) instead of being expanded into `_bucket`, `_sum`, and `_count` series.
-The OTel `explicit_bounds` array maps directly to the NHCB custom bucket boundaries.
-The downstream appender must support native histograms.
-If the metrics are forwarded over remote write, both the sending component and the receiver must speak Prometheus Remote Write 2.0, since NHCB isn't representable in Remote Write 1.0.
-With `prometheus.remote_write`, set `protobuf_message = "io.prometheus.write.v2.Request"` in the `endpoint` block (the default is `"prometheus.WriteRequest"`, which is Remote Write 1.0).
+The downstream components must support native histograms.
+
+To enable and use an experimental feature, you must set the `stability.level` [flag][] to `experimental`.
+
+[experimental]: https://grafana.com/docs/release-life-cycle/
+[flag]: https://grafana.com/docs/alloy/<ALLOY_VERSION>/reference/cli/run/
 {{< /admonition >}}
 
 When `include_scope_labels` is `true`  the `otel_scope_name` and `otel_scope_version` labels are added to every converted metric sample.
@@ -204,6 +206,33 @@ otelcol.exporter.prometheus "default" {
 prometheus.remote_write "mimir" {
   endpoint {
     url = "http://mimir:9009/api/v1/push"
+  }
+}
+```
+
+## Convert classic histograms to NHCB
+
+This example enables `convert_classic_histograms_to_nhcb` and forwards the converted metrics to a `prometheus.remote_write` endpoint configured for Prometheus Remote Write 2.0.
+NHCB samples can't be sent over Remote Write 1.0, so the `protobuf_message` argument must be set to `io.prometheus.write.v2.Request` on the `endpoint` block.
+
+```alloy
+otelcol.receiver.otlp "default" {
+  grpc {}
+
+  output {
+    metrics = [otelcol.exporter.prometheus.default.input]
+  }
+}
+
+otelcol.exporter.prometheus "default" {
+  convert_classic_histograms_to_nhcb = true
+  forward_to                         = [prometheus.remote_write.mimir.receiver]
+}
+
+prometheus.remote_write "mimir" {
+  endpoint {
+    url              = "http://mimir:9009/api/v1/push"
+    protobuf_message = "io.prometheus.write.v2.Request"
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
```diff
@@ -64,6 +65,16 @@ The following components are compatible:
 [experimental]: ../../../get-started/configuration-syntax/expressions/function_calls/#experimental-functions
 {{< /admonition >}}
 
+{{< admonition type="warning" >}}
+**EXPERIMENTAL**: The `convert_classic_histograms_to_nhcb` argument is an [experimental][] feature.
```
ptodev, line 69: I think the `[experimental][]` link won't work unless this is added inside the admonition block, like in the block above:

```
[experimental]: ../../../get-started/configuration-syntax/expressions/function_calls/#experimental-functions
```

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
```diff
@@ -64,6 +65,16 @@ The following components are compatible:
...
+
+When set to `true`, each OTel explicit-bucket histogram data point is converted to a single Prometheus native histogram with custom buckets (NHCB) instead of being expanded into `_bucket`, `_sum`, and `_count` series.
+The OTel `explicit_bounds` array maps directly to the NHCB custom bucket boundaries.
+The downstream appender must support native histograms.
+If the metrics are forwarded over remote write, both the sending component and the receiver must speak Prometheus Remote Write 2.0, since NHCB isn't representable in Remote Write 1.0.
```
ptodev:
```suggestion
NHCB isn't representable in Remote Write 1.0.
```
Is this true? I thought that both native histograms and native histograms with classic buckets are available in v1, although they need to be enabled explicitly. It might be best just to omit this statement and instead to replace the entire admonition block with an example config in the Example section of the doc? The example can use RWv2 and avoid making statements about RWv1 if we're not sure whether they are accurate and whether they will stay true over time.

internal/component/otelcol/exporter/prometheus/internal/convert/histograms.go
```diff
@@ -150,3 +151,125 @@ func convertBucketsLayout(buckets pmetric.ExponentialHistogramDataPointBuckets,
...
+		h.Count = p.Count()
+	}
+	return h, nil
+}
+
+// convertCustomBucketsLayout converts a dense slice of cumulative-zero-trimmed
+// bucket counts into the sparse spans/deltas representation used by native
+// histograms with custom buckets. Unlike the exponential variant, indices are
+// not scaled and the initial offset is not adjusted by 1.
+func convertCustomBucketsLayout(bucketCounts []uint64, offset int32) ([]histogram.Span, []int64) {
```
ptodev, line 198: This does seem to align with what's in the [Prometheus code](https://github.com/prometheus/prometheus/blob/ea787c89cbe7fbc4e4cf1cbe4785d2de4c83a042/storage/remote/otlptranslator/prometheusremotewrite/histograms.go#L170-L256), so I'll assume it's ok.

madaraszg-tulip:
Something seems to be off with the CLA check. I have signed it back here: https://github.com/grafana/alloy/pull/2185#issuecomment-2506215876

Now when I click on it, I get:

```
Error
There is no CLA to sign for grafana/alloy

(Bad credentials)
```

ptodev (commented):
Thanks! LGTM, as long as move some of the docs content to an example. The non-example section can just state:

> When set to `true`, each OTel explicit-bucket histogram data point is converted to a single Prometheus native histogram with custom buckets (NHCB) instead of being expanded into `_bucket`, `_sum`, and `_count` series.
The downstream components must support native histograms.

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
madaraszg-tulip: I had a long discussion with prometheus devs about this last year when I experimented with prometheus and NHCB and it silently broke those metrics. They said RW1.0 will not change and explicitly added this code:

https://github.com/prometheus/prometheus/blob/ea787c89cbe7fbc4e4cf1cbe4785d2de4c83a042/storage/remote/queue_manager.go#L855-L860

```
		if t.protoMsg == remoteapi.WriteV1MessageType && h.H != nil && h.H.Schema == histogram.CustomBucketsSchema {
			// We cannot send native histograms with custom buckets (NHCB) via remote write v1.
			t.metrics.droppedHistogramsTotal.WithLabelValues(reasonNHCBNotSupported).Inc()
			t.logger.Warn("Dropped native histogram with custom buckets (NHCB) as remote write v1 does not support itB", "ref", h.Ref)
			continue
		}
```

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
clayton-cornell: That's correct. During the site generation, the admonition contents are built before the general links are parsed. We have to define all links used in an admonition within the admonition somewhere.. either inline or as a separate link definition line.

madaraszg-tulip: Actually this link is not the right one, it's broken (404), I'll fix the one for honor_metadata too

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
madaraszg-tulip: Admonition block reduced, example added.

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
ptodev: Thank you!

ptodev (approved):

PR #6284, opened 2026-05-19
docs: Add project file for Claude documentation skill integration

clayton-cornell:
This PR adds the project-context.md file that is used by the documentation skill integrations. This file can be used by anyone within Grafana contributing to the Alloy docs in combination with the Claude skills defined in the docs-ai repo.

This file should have no impact for anyone (community or Grafana) that does not use the skills for doc maintenance.

### DIFF
--- a/docs/project-context.md
+++ b/docs/project-context.md
@@ -1,0 +1,75 @@
+# Project context (local)
+
+<!-- Copied from docs-ai/skills/context-templates/project-context.template.md -->
+<!-- Location: docs/project-context.md -->
+
+## Identity
+
+- **Product name (short):** Alloy
+- **Product name (first mention in prose):** Grafana Alloy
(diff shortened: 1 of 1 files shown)


thampiotr (commented):
In case we use something other than Claude, would this be detected? 

I'm thinking we should maybe link to this doc from docs-writing.md or from AGENTS.md

clayton-cornell:
> In case we use something other than Claude, would this be detected?

project-context.md is for Claude. It functions in the same general way that copilot-instructions.md does. Content-wise it's just a markdown file and the content can be referenced by most AI engines if you explicitly call it out.

> I'm thinking we should maybe link to this doc from docs-writing.md or from AGENTS.md

We could do that in the future. The docs-ai repo isn't public yet.

An alternative if we don't want to push to the Alloy repo is I keep it local and add the filename to git ignore.

PR #6291, opened 2026-05-21
feat(database_observability): Set query_sample, wait_event timestamps at query start for MySQL

fridgepoet:

This PR changes the query_sample collectors to log timestamps at query _start_ time.

In the context of application traces, traces start their span at query start time. The DB O11y app's trace linkage navigates to the trace's start time on the timeline, so the previous end timestamp query_sample point in the UI can appear _too late_.

This fix changes the timestamp of the log to the start time for the MySQL query_sample collector.

Contributes to https://github.com/grafana/grafana-dbo11y-app/issues/2818

### Notes to the Reviewer

To test this manually, application trace timestamps for queries were compared and found similar to op=query_sample log timestamps. 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


gaantunes:
Mysql looks clean, but the Postgres part will cause a few issues. My [PR](https://github.com/grafana/alloy/pull/6273) is touching some of the same code lines, and fixing your problem statement there seems easier than syncing our work, and it could be merged independently this way as well. I propose that you just touch MySQL here, and I will fix the issue for postgres in my PR. WDYT?

gaantunes:
Done in https://github.com/grafana/alloy/pull/6273/changes/3187dbff74e3b2d26efdfd9e5d7d0ea0bf7b5d1d

fridgepoet:
Ok, I removed my changes for Postgres here.
@gaantunes I suggest we can just follow up with a Postgres PR for this separately rather than add it to your changes. That'll keep your PR focused. Happy to follow up on this.

gaantunes:
> Ok, I removed my changes for Postgres here. @gaantunes I suggest we can just follow up with a Postgres PR for this separately rather than add it to your changes. That'll keep your PR focused. Happy to follow up on this.

I think it is fine, still a smallish PR.

fridgepoet:
> I think it is fine, still a smallish PR.

Yeah, I’d still lean toward separating unrelated changes when possible. Keeping the scope of a PR focused makes it easier to review and reason about, and it also keeps the history cleaner for future debugging or reverts if something goes wrong.

matthewnolf (approved):

PR #6297, opened 2026-05-22
feat(database_observability.postgres): Emit op=query_association_v2 carrying query_fingerprint

gaantunes:
## Summary

Emits `op="query_association_v2"` — the existing `query_association` line plus a stable `query_fingerprint` between `queryid` and `querytext` — giving consumers a `queryid` ↔ `query_fingerprint` mapping. Gated behind the same `logs_processing { enable_error_logs }` flag from #6180 (no separate argument); with it off, the legacy `op="query_association"` is emitted unchanged — never both.

Stacked on #6180: reuses its `fingerprint` package and shares its `enable_error_logs` flag.

## Entry shape

```
op="query_association_v2"
body: level="info" queryid="<int>" query_fingerprint="<16-hex>" querytext=<sql> datname="<db>"
```

The fingerprint is computed client-side via `libpg_query` on the raw `pg_stat_statements.query` (before comment-stripping), so it matches the value on #6180's `op="error"`. Recover the queryids for a fingerprint with `{op="query_association_v2"} | logfmt | query_fingerprint="<fp>"`.

## Scope

The same `enable_error_logs` toggle enables `op="error"` (#6180) and this op together. Default off preserves `op="query_association"` byte-for-byte.

## Test plan

- [x] `go test -race -tags=nodocker ./internal/component/database_observability/postgres/...` — green
- [x] `golangci-lint` — clean
- [x] v2-on emits the `query_fingerprint`; v2-off preserves the legacy op with no `query_fingerprint`


docs/sources/reference/components/database_observability/database_observability.postgres.md
```diff
@@ -37,6 +37,7 @@ You can use the following arguments with `database_observability.postgres`:
 | `exclude_databases`        | `list(string)`       | A list of databases to exclude from monitoring.             | `["alloydbadmin", "alloydbmetadata", "azure_maintenance", "azure_sys", "cloudsqladmin", "rdsadmin"]` | no       |
 | `exclude_users`            | `list(string)`       | A list of users to exclude from monitoring.                 | `["azuresu", "cloudsqladmin", "db-o11y", "rdsadmin"]` | no       |
 | `exclude_current_user`     | `bool`               | Exclude from activity monitoring the user that Alloy uses to connect to the database. The resolved username is automatically added to `exclude_users`, if not already present. | `true` | no       |
+| `enable_query_fingerprint` | `bool`               | Compute a stable semantic `query_fingerprint` per observed query and emit it on versioned Loki ops alongside the queryid. | `false` | no       |
```
```suggestion
| `enable_query_fingerprint` | `bool`               | Compute a stable semantic `query_fingerprint` per observed query and emit it on versioned Loki ops alongside the `queryid`. | `false` | no       |
```

internal/component/database_observability/postgres/component.go
```diff
@@ -622,6 +622,7 @@ func (c *Component) startCollectors(systemID string, engineVersion string, cloud
 			ExcludeUsers:     effectiveExcludeUsers,
 			EntryHandler:     entryHandler,
 			TableRegistry:    tableRegistry,
+			EnableErrorLogs:  c.args.Logs.EnableErrorLogsProcessing,
```
cristiangreco: Nit: can we use the same name `EnableErrorLogsProcessing`?

gaantunes: Yes, renaming all over.

internal/component/database_observability/postgres/collector/query_details.go
```diff
@@ -153,16 +158,34 @@ func (c *QueryDetails) fetchAndAssociate(ctx context.Context) error {
...
+
 		queryText, err = removeComments(c.normalizer, queryText)
 		if err != nil {
 			c.logger.Error("failed to remove comments", "err", err)
 			continue
 		}
 
+		op := OP_QUERY_ASSOCIATION
+		body := fmt.Sprintf(`queryid="%s" querytext=%q datname="%s"`, queryID, queryText, databaseName)
+		if c.enableErrorLogsProcessing {
```
cristiangreco: @gaantunes should we also check that `fp != ""`? (e.g. when fingerprinting fails)

internal/component/database_observability/postgres/collector/query_details.go
```diff
@@ -153,16 +158,34 @@ func (c *QueryDetails) fetchAndAssociate(ctx context.Context) error {
...
+			op = OP_QUERY_ASSOCIATION_V2
+			body = fmt.Sprintf(`queryid="%s" query_fingerprint="%s" querytext=%q datname="%s"`, queryID, fp, queryText, databaseName)
+		}
```
cristiangreco, line 194: nit: micro-optimization, I know, could we avoid filling `body` twice? i.e.

```
if c.enableErrorLogsProcessing {
  op = v2
  body = ...
} else {
  op = v1
  body = ...
}
```

internal/component/database_observability/postgres/collector/query_details.go
gaantunes: Makes sense, I am not emitting a warning if `fp` is empty and holding the emition.

cristiangreco (approved):

PR #6300, opened 2026-05-22
ci: Harden npm supply-chain config (allow-git, min-release-age)

kgeckhart:

Adds `allow-git=none` and `min-release-age=3d` to the UI's `.npmrc` and pins `packageManager` to `npm@11.15.0` in `package.json` per the org's recommended npm supply-chain hardening. Follow-up to #6241.

### Notes to the Reviewer

On the npm currently bundled with our pinned Node (11.13.0 via Node 24.13.0), `allow-git` and `min-release-age` are unknown directives and silently ignored — they only take effect once the runtime npm reaches 11.15.0. Open question I don't have a strong view on: do we want to bridge that gap in CI now (e.g., `npm install -g npm@11.15.0` or `corepack enable` before `npm install` in `lint.yml`), or leave the settings as declared intent and let them activate when a future Node patch ships npm 11.15.0+ naturally? Happy to follow up either way.


thampiotr (dismissed):

jharvey10:
I like the npmrc changes for sure. We may want more than just the `packageManager` one in the package.json file though. Also we should extend this to the other npm files we have for the release-please-runner.

Were you able to confirm that these settings do in fact result in npm 11.15 being used in our CI runners?

For local development, packageManager won't have any effect, and to enforce that, we'd likely want to include an `engines` block.

```diff
diff --git a/internal/web/ui/.npmrc b/internal/web/ui/.npmrc
index 97b895e2f..6e29f2f17 100644
--- a/internal/web/ui/.npmrc
+++ b/internal/web/ui/.npmrc
@@ -1 +1,2 @@
 ignore-scripts=true
+engine-strict=true
diff --git a/internal/web/ui/package.json b/internal/web/ui/package.json
index 15ae10260..9efbbd2e8 100644
--- a/internal/web/ui/package.json
+++ b/internal/web/ui/package.json
@@ -3,6 +3,11 @@
   "private": true,
   "version": "0.0.0",
   "type": "module",
+  "packageManager": "npm@11.15.0",
+  "engines": {
+    "node": ">=24.13.0",
+    "npm": ">=11.15.0"
+  },
   "scripts": {
     "dev": "vite",
     "generate:fixtures": "tsx src/test/fixtures/writeFixture.ts",
```

the only problem with that right now is that since there's no node version with 11.15 in it yet, we'd need to update our instructions for how a dev picks that up, because `nvm use` by itself will no longer work.

what works for me locally to pick up a usable npm is `nvm install --latest-npm`

kgeckhart:
> Were you able to confirm that these settings do in fact result in npm 11.15 being used in our CI runners?

I think the Notes to the Reviewer section addresses this bit. I wasn't quite sure how far to take this. 

Started a convo offline about the other bits for now.

jharvey10:
Did some more research into this. Right now, we're using [24.13.0](https://nodejs.org/en/download/archive/v24.13.0) in our nvmrc, which comes with npm version `11.6.2`.

`allow-git` was introduced in 11.9.0, so as long as we move to at least node 24.14.0, we can take advantage of this, but there's a catch... According to [this PR](https://github.com/npm/cli/pull/9348), the implementation had some gaps until this fixed it and was released in 11.15.0.

The latest node in v24 [v24.16.0](https://nodejs.org/en/download/archive/v24.16.0) ships with npm v11.13.0. Still doesn't have the fix, but is getting close. In roughly a month I'd expect there to be a new node v24 version that we can pick up with that fix, but in the meantime, I'd recommend we upgrade node to v24.16.0, use the CLI arg in its current state, and then upgrade as soon as 24.17.0 becomes available.

Regarding `min-release-age` anything v24.14.1 or higher should have this, so we're good there as long as we upgrade.

jharvey10:
In either case, adding `packageManager` doesn't appear to buy us much unless we're also using corepack. And given that corepack is deprecated, I'd recommend we do the following:

- Upgrade node
- Add config settings for `allow-git=none` in both .npmrc files
- Add config settings for `min-release-age=7d` in both .npmrc files
- Add `engines` blocks to both package.json files
- Point `with.node-version-file` to package.json files in our workflows
- [maybe] Add `engine-strict=true` in both .npmrc files (though this needs to be checked to ensure it doesn't break things like dependabot/renovate)

jharvey10 (approved):

PR #6301, opened 2026-05-23
fix(loki.source.syslog): Treat RFC5424 structured-data as meaningful

DominoTree:

`loki.source.syslog` now forwards RFC5424 messages that have an empty MSG field
when their STRUCTURED-DATA section is populated, instead of dropping them.
Previously such messages were discarded (and counted in
`loki_source_syslog_empty_messages_total`) unless `rfc5424_allow_empty_msg` was
set, even though their payload lived entirely in the structured-data section.

### Pull Request Details

RFC5424 explicitly allows MSG to be absent (`MSG = MSG-ANY / MSG-UTF8 / [empty]`),
and many real-world senders put their whole payload in STRUCTURED-DATA with no
MSG at all: Juniper Junos `sd-syslog` flow records
(`RT_FLOW_SESSION_CREATE/CLOSE/DENY`, …), OTel-style collectors, and assorted
network gear. For these, the structured-data section *is* the message, so
dropping them loses data even though parsing succeeded. The only prior workaround
was running senders in RFC3164 (`format syslog`) mode, which gets the bytes
through but forfeits first-class structured-data fields.

This change treats "empty MSG **with** non-empty STRUCTURED-DATA" as meaningful:
it is always forwarded and the empty-message counter is not incremented,
regardless of `rfc5424_allow_empty_msg`. The flag still governs the truly-empty
case (NILVALUE structured data `-` and no MSG), which is unchanged. The change is
scoped to the RFC5424 path (`handleMessageRFC5424`); RFC3164 has no structured
data and is intentionally left alone.

`StructuredData` on `leodido/go-syslog/v4` `rfc5424.SyslogMessage` is a
`*map[string]map[string]string`, so the guard is
`msg.StructuredData != nil && len(*msg.StructuredData) > 0`.

### Notes to the Reviewer

The existing negative test `TestSyslogTarget_RFC5424MessageEmptyMSGWhenNotAllowed`
previously used a message that *had* structured data, which would now (correctly)
be forwarded; it has been updated to a true NILVALUE-SD message
(`… RT_FLOW - … -`) so it still exercises the real drop path. A new test,
`TestSyslogTarget_RFC5424EmptyMSGWithStructuredDataAlwaysAllowed`, covers the new
behavior including structured-data label extraction.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/loki/loki.source.syslog.md
+++ b/docs/sources/reference/components/loki/loki.source.syslog.md
@@ -170,6 +170,7 @@ You can't use the `rfc3164_default_to_current_year`, `use_incoming_timestamp`, a
   `loki.source.syslog` drops messages with empty MSG content by default.
   Set `rfc5424_allow_empty_msg` to `true` to forward them.
   `loki.source.syslog` increments the `loki_source_syslog_empty_messages_total` counter in both cases for debugging.
+  Messages with empty MSG content but non-empty STRUCTURED-DATA are always forwarded regardless of `rfc5424_allow_empty_msg`, since their payload lives in the structured-data section. This is common in Junos `sd-syslog` flow records and similar network gear.
 - **`raw`**
   Disables log line parsing. This format allows receiving non-RFC5424 compliant logs, such as [CEF][cef].
   Raw logs can be forwarded to [`loki.process`](./loki.process.md) component for parsing.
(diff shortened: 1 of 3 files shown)


DominoTree:
It is entirely possible that this is the entire point of the `rfc5424_allow_empty_msg`

blewis12:
Hiya @DominoTree - indeed that's exactly what the `alloy_empty_msg` field was introduced for [here](https://github.com/grafana/alloy/pull/5687), we decided to make it opt-in to avoid any breaking behavior. The test `RFC5424MessageEmptyMSGWhenNotAllowed` was supposed to replicate a scenario in which the SD contained the "real" message but the MSG was empty, and users had _not_ opted in to the keep behavior and so the log would be dropped. If you still feel that the testing could be improved or made clearer, feel free to submit another PR!

PR #6314, opened 2026-05-26
chore: Use mise for repo dependencies 

kalleep:
### Pull Request Details
After internal discussions we a leaning towards using mise for tool installation and Taskfile for as task runner, replacement for make.

This pr adds mise.toml and mise.lock so it can be used to install common tooling. I left out a lot of tools that I think we can assume already exists such as docker, kubeclt, make, git etc. Not really sure where we should draw the line but I figured this was a good starting point.

For now I migrated lint workflow to use mise, installed tools is cached in workflows so we will benifit more when we migrate others to use it. 

### Notes to the Reviewer

Before we exectuted govoulcheck by running `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0`. Now it's installed through mise and ran directly. 

Renovate have support for mise https://github.com/renovatebot/renovate/pull/33113#issuecomment-2907645456

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


x1unix (approved):

x1unix:
Merging this with Kalle's agreement

PR #6315, opened 2026-05-26
chore: Tools restructure

kalleep:
### Pull Request Details
Our tools directory have been a dumping ground for a lot of different things:
* Small go programs
* Various bash scripts
* alloy build image
* release-please-runner 

Since we want to reduce the number of bash scripts I figured it would be easiest to have one go cli with various commands we need. The place for that cli is the tools directory which means we should consolidate every go program we have under one cli. But before I do that I wanted to clean up the structure a bit and follow up with a pr to convert all go code we currently have in tools to one cli. After that we can start to migrate our various bash scripts into that cli.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


jharvey10:
The release please runner I wouldn't really classify as a build tool. Its job is to manage the release PRs we have open and and is triggered by CI. It's kind of a misfit because it's more of a github action, but we had to add a custom wrapper around it to handle our versioning.

jharvey10 (approved):

PR #6322, opened 2026-05-27
ci: Publish Alloy Images to GAR 

blewis12:
We will now be publishing our Alloy images to GAR, which will then automatically be mirrored back to our Dockerhub repositories. This means that our CI workflows are no longer directly authenticating with Dockerhub. Part of following [these steps](https://github.com/grafana/deployment_tools/blob/master/docs/platform/gar-image-mirror.md)

Publishing `devel` images on PR merge to `main` hook into the `publish-alloy-*` workflows, so those will also be using GAR/mirroring after this PR is merged

Part of https://github.com/grafana/alloy-internal/issues/362

### DIFF
--- a/.github/workflows/create_build_image.yml
+++ b/.github/workflows/create_build_image.yml
@@ -42,7 +42,6 @@ jobs:
       uses: grafana/shared-workflows/actions/login-to-gar@c7ad5f57693ac858c698d95783685a9cbfd91223 # login-to-gar/v1.0.1
       with:
         registry: us-docker.pkg.dev
-        environment: prod
 
     - name: Setup QEMU
       uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
--- a/.github/workflows/create_build_image.yml
+++ b/.github/workflows/create_build_image.yml
@@ -38,8 +38,11 @@ jobs:
       id: get_image_version
       run: echo "image_tag=${FULL_TAG##*/}${{ matrix.build.suffix }}" >> $GITHUB_OUTPUT
 
-    - name: Login to DockerHub (from vault)
-      uses: grafana/shared-workflows/actions/dockerhub-login@081a366728379fd0426b9cfef190e9a21c2d5418 # dockerhub-login/v1.0.3
+    - name: Login to Google Artifact Registry
+      uses: grafana/shared-workflows/actions/login-to-gar@c7ad5f57693ac858c698d95783685a9cbfd91223 # login-to-gar/v1.0.1
+      with:
+        registry: us-docker.pkg.dev
+        environment: prod
 
     - name: Setup QEMU
       uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4.0.0
@@ -53,6 +56,6 @@ jobs:
         platforms: linux/amd64,linux/arm64
         context: ./build-tools/build-image
         push: true
-        tags: grafana/alloy-build-image:${{ steps.get_image_version.outputs.image_tag }}
+        tags: us-docker.pkg.dev/grafanalabs-global/dockerhub-alloy-prod-mirror/alloy-build-image:${{ steps.get_image_version.outputs.image_tag }}
         build-args: |
           GO_RUNTIME=${{ matrix.build.runtime }}
--- a/scripts/docker-containers-windows
+++ b/scripts/docker-containers-windows
@@ -41,7 +41,7 @@ export DEVEL_ALLOY_IMAGE=$GAR_MIRROR/alloy-dev
 
 # TODO: Unit test this script? Test cases:
 # * Input:  ALLOY_GO_VERSION=1.24 WINDOWS_VERSION=windows-2022 GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v1.8.0 ./scripts/docker-containers-windows alloy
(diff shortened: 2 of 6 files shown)


.github/workflows/create_build_image.yml
```diff
@@ -38,8 +38,11 @@ jobs:
       id: get_image_version
       run: echo "image_tag=${FULL_TAG##*/}${{ matrix.build.suffix }}" >> $GITHUB_OUTPUT
 
-    - name: Login to DockerHub (from vault)
-      uses: grafana/shared-workflows/actions/dockerhub-login@081a366728379fd0426b9cfef190e9a21c2d5418 # dockerhub-login/v1.0.3
+    - name: Login to Google Artifact Registry
+      uses: grafana/shared-workflows/actions/login-to-gar@c7ad5f57693ac858c698d95783685a9cbfd91223 # login-to-gar/v1.0.1
+      with:
+        registry: us-docker.pkg.dev
+        environment: prod
```
kgeckhart: Is this doing anything? Looks like it's not currently listed as an input https://github.com/grafana/shared-workflows/tree/main/actions/login-to-gar

scripts/docker-containers-windows
kgeckhart: Is this still valid?

kgeckhart (commented):
Do we need to do something with our post steps that might assume the image exists? Tasks which trigger other workflows?

.github/workflows/create_build_image.yml
blewis12: ah you're right! i'll remove that

scripts/docker-containers-windows
blewis12: hm yeah the image/tag needs updating now to GAR, incoming

blewis12: https://github.com/grafana/alloy/pull/6322/commits/4c9b525b4c0d66c92aacaf910e0c62eb81eb5279

blewis12:
> Do we need to do something with our post steps that might assume the image exists? Tasks which trigger other workflows?

That's a good question, having a look over the different workflows and I can see two places this might cause issues, sort of for the same reason:

1. `publish-alloy-devel.yml` => we push to GAR and then immediately create a PR in deployment_tools that relies on the dockerhub image
2. `trigger-argo-workflow-on-tag.yml` => same as above, we publish to GAR but will start rollout relying on dockerhub image

Additionally there's a separate issue, but one I argue can be addressed in a follow up: We cannot push to GAR from a pull request CI and thus cannot run `publish-alloy-devel-pr.yml` => I think we can address this as a follow up

I've pinged platform about both of these to get their opinion, in particular I would like to understand how long the mirroring process takes and if maybe we need to be referencing GAR instead of dockerhub in `deployment_tools` now

ptodev (commented):
Have the changes been tested? Or are we smoke testing it on the main branch? I'd be good if they can be tested on the main branch of a fork. Then you can also test other workflows that depend on it by creating a release on the fork, etc.

blewis12:
> Have the changes been tested? Or are we smoke testing it on the main branch? I'd be good if they can be tested on the main branch of a fork. Then you can also test other workflows that depend on it by creating a release on the fork, etc.

@ptodev The idea would be to smoke test it on the main branch yes, we can't really test it on a fork because pushing to GAR will only work in this repository due to OIDC. The smoke testing here will be the `devel` image workflow, where we merge this into `main` and should see us pushing to GAR for our `alloy-dev` image, which should be mirrored to dockerhub - and then see a PR open against `deployment_tools`. 

Let me know if I misunderstood your comment though

blewis12:
Also, as a follow up from the questions I asked to the platform team: mirroring typically takes 30 seconds, but can be longer with larger images. There's an [action in progress](https://github.com/grafana/shared-workflows/pull/1956) that should allow us to wait until mirroring is complete before progressing with a workflow which will be useful

The ETA on that mirror-wait workflow is end of this week / beginning of next week, so might make sense to wait for that since it's not far off

blewis12:
Another idea as well in terms of testing: we could go ahead with the GAR migration _just_ for the build image? This feels lower in risk, and that workflow doesn't rely on waiting for mirroring. We have [this PR](https://github.com/grafana/alloy/pull/6277) open that we could test against once it's merged/tagged

ptodev:
> pushing to GAR will only work in this repository due to OIDC

Could we use a personal, non-Grafana repo for this? This is how I used to test publishing to dockerhub. I'd fork and change the dockerhub location to my own one. It'd be worth asking the platform team for recommended ways of testing this. They might also have test GAR repos.

blewis12:
@ptodev unless there are strong objections, let's test out the E2E flow just for the build-image (using a [smaller PR](https://github.com/grafana/alloy/pull/6335)). Beyla is [already using this workflow](https://github.com/grafana/beyla/pull/2821) so it's not completely unchartered territory. I think if that goes well we could move on to enabling the E2E GAR flow for `devel` images. Is there something in particular you'd be worried about and would want to be tested in a fork? A fork environment just feels like a lot of work when smoke testing from `main` doesn't carry a big risk, we'd need to set up a separate GAR repo that can be pushed to, our own dockerhub and so forth

blewis12:
It works for build images 🥳 so now to decide if we want to have a temporary solution in the meantime to block/poll dockerhub or wait for the action from platform that will do that for us. I'll look into a simple interim action since i'd like to get this whole thing tested sooner rather than later

kgeckhart (approved):
I think it's fine to move as is. The worst case scenario is that we have some image pull backoffs for a period of time in dev until the sync'ing happens. But given the short window it's more likely the end to end time is long enough to cover the gap and we'll be able to inject the wait step later to be sure about it.

PR #6333, opened 2026-05-28
chore: Fix lints

kalleep:
### Pull Request Details
This pr fixes the lints we are running.

Previously `make lint` only ran lint on root module (several times) and in CI we only linted syntax and main modules.
We also did not run shellcheck on a lot of scripts, I adress this by creating a script and make target we use both locally and in ci.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


scripts/lint-go
ptodev, line 1: Could this be a simple Go application instead of a bash script? Or some other kind of script that's easier to read and safer?

kalleep: Yes I plan to convert all of these scripts but I don't want to do it until https://github.com/grafana/alloy/pull/6318 is meged. Then we can just convert them to a command there

.github/workflows/lint.yml
```diff
@@ -22,24 +22,21 @@ jobs:
...
+    - name: Lint Go
+      run: make lint-go
 
-    - name: Run custom linters
+    - name: Run Alloy Lint
       run: make run-alloylint
 
   shellcheck:
-    name: Shellcheck
+    name: Lint shell
```
ptodev: It'd be best to revert this line since the CI keeps waiting for a "Shellcheck" to complete. We could reconfigure the CI but then it could have an effect on other PRs and it's not worth it.

.github/workflows/scripts/generate-component-labels
```diff
@@ -16,22 +18,18 @@ COLORS["remote"]="#4B5320"
 
 FALLBACK_COLOR="#999999"
 
-for README in $(find ./docs/sources/reference/components -name '*.md' ! -name '*index.md'); do
-    # The find ends up with an empty string in some OSes
-    if [[ -z "${README}" ]]; then
-        continue
-    fi
+while IFS= read -r -d '' README; do
```
ptodev, line 21: What is the point of this change? I can read the previous code but I don't even understand the new one 😆

scripts/image-tag
```diff
@@ -19,10 +19,10 @@ fi
 
 DETECTED_TAG=$(git describe --match 'v*' --exact-match 2>/dev/null || echo -n "")
 
-if [ ! -z "${RELEASE_TAG}" ]; then
+if [ -n "${RELEASE_TAG}" ]; then
```
ptodev, line 22: How does this help? Is it that it catches more empty types?

scripts/lint-go
ptodev, line 1: Would be nice to have a comment on top of both of these bash files to summarise what they do. I can sort of read them but there is stuff I really don't understand like `IFS= read -r -d ''`

ptodev (commented):
I think it's ok but I added a few minor comments mostly for my own understanding.

scripts/image-tag
kalleep: It is basically the same check, change because the other version failed lint

.github/workflows/scripts/generate-component-labels
kalleep: Was flagged by linter, the other version would break on spaces and newlines in file names.

scripts/lint-go
kalleep: I don't think it's necessary since I plan to replace these scripts with go code in a not to distant future, that will make it much easier to reason about

ptodev (approved):

PR #6338, opened 2026-05-28
chore: Tools cli part 1

kalleep:
Bootstrap the common tools cli and add goversion as command to it, extracted from https://github.com/grafana/alloy/pull/6318


tools/goversion/command.go
```diff
@@ -1,69 +1,86 @@
...
+	"github.com/grafana/alloy/tools/internal/git"
 )
 
-func main() {
-	if len(os.Args) < 3 {
-		printUsage()
-		os.Exit(1)
+func Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-go-version <command>",
```
ptodev: Do you think it'd be better if the folder name matches the command name?

ptodev (dismissed):

ptodev (approved):

PR #6339, opened 2026-05-28
fix(mimir.rules.kubernetes) Fix non-recovering unhealthy state after a failed restart when Mimir is unreachable

### LINKED ISSUE #1076: [mimir.rules.kubernetes] When mimir is unavailable when the component start, it never tries again
### Request

Component startup should retry when connection to mimir is unavailable during the compoent starts but it does not and we have to manually restart the pod:

```
ts=2024-06-18T13:27:21.649522346Z level=error msg="failed to list rules from mimir" component=mimir.rules.kubernetes.local err="Get \"http://mimir-ruler.mimir.svc:8080/prometheus/config/v1/rules\": dial tcp 172.31.62.76:8080: connect: connection refused"
ts=2024-06-18T13:27:21.649584571Z level=error msg="starting up component failed" component=mimir.rules.kubernetes.local err="Get \"http://mimir-ruler.mimir.svc:8080/prometheus/config/v1/rules\": dial tcp 172.31.62.76:8080: connect: connection refused"
```

### Use case

We need this because we deploy the agent at the same time as Mimir and we have to manually intervene

TheoBrigitte:

Fixes an issue where `mimir.rules.kubernetes` component stops working and stays as unhealthy in case a restart occurred while Mimir was unreachable 

### Pull Request Details

Under some specific conditions the `mimir.rules.kubernetes` would stop working and no longer sync rules with Mimir.

This happens when Alloy becomes the new leader and restarts the `mimir.rules.kubernetes`, if Mimir is unreachable during the component restart, the `startup` method terminates and does not launch the new `eventProcessor.run`. This leaves the component in a unhealthy state and also prevent any rule from being synced with Mimir.

The fix proposed here is to make sure the eventProcessor is always run regardless of the Mimir being reachable or not, and enqueue a new sync event right away, rather than having the sync happening synchronously. 

I am unsure if this is the right approach as this would still leave the eventProcessor runing without any state in case Mimir is unreachable (as the eventProcessor would fail after 5 retry). It might be worth making sure the state is being synced first when an event is received.

Fixes #1076 

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


QuentinBisson:
Hey @dehaansa, do you think you could check out this PR to unblock an issue we have?

PR #6348, opened 2026-05-28
chore: Migrate direct pkg/errors usage to stdlib errors

thampiotr:
## Summary
- replace the remaining direct `github.com/pkg/errors` imports in Loki journal/config code with stdlib `errors`
- convert `errors.Wrap(...)` calls to `fmt.Errorf(...: %w)` while preserving context
- run `go mod tidy`, which moves `github.com/pkg/errors` from direct to indirect because it is still transitively required by `github.com/grafana/dskit/flagext`

## Test plan
- [x] `go test ./internal/loki/util/cfg ./internal/component/loki/source/journal`
- [x] `go mod tidy`
- [x] `git grep -n "github.com/pkg/errors" -- '*.go'`


kalleep (approved):
:tada:

PR #6349, opened 2026-05-28
fix(ui): Reduce UI dependencies

jharvey10:

Removes the following direct dependencies:

- `@grafana/data`
- `@grafana/ui`
- `d3`
- `d3-dag`
- `d3-zoom`
- `rc-slider`
- `@types/d3`
- `@types/d3-zoom`

This results in a net reduction of **330+** unique packages from the lockfile.

### Pull Request Details

The grafana UI package was being used for the Slider, Field, and Input controls. These are replaced by a basic HTML-native slider element and corresponding input/label elements.

It was also being used for theme boostrapping, but since multiple themes are not being used, that functionality was removed with no change in behavior.

Some failing unit tests were fixed, which involved adding jsdom as a dependency and correcting some assertions.

### Before

image

### After

image

### Before

image

### After

image 2

### PR Checklist

- [x] Tests updated


kgeckhart (commented):
Nice work! Could you possibly do a recording showing the various places still working? It doesn't look like we have any end-to-end tests that can help verify but I might have missed something.

jharvey10:
@kgeckhart Not me forgetting to click "save" after adding my screenshots 🤦 

I'll record a small vid and attach it as well!

jharvey10:
Demo vid:

demo

kgeckhart (approved):
🔥

PR #6352, opened 2026-05-28
feat(sigil): Add sigil.receive and sigil.write components

alexander-akhmetov:

Add two experimental components for sending [Grafana AI Observability](https://grafana.com/docs/grafana-cloud/machine-learning/ai-observability/) (Sigil) conversation data through Alloy.

`sigil.receive` starts an HTTP server that accepts generation export requests (`POST /api/v1/generations:export`).

`sigil.write` forwards records to one or more remote Sigil endpoints, serializing each request as JSON.

<details>
  <summary>Config examples</summary>

Receive over HTTP and forward to a single Sigil endpoint:

```alloy
sigil.receive "default" {
  forward_to = [sigil.write.default.receiver]
}

sigil.write "default" {
  endpoint {
    url = "https://<sigil>"

    basic_auth {
      username = env("SIGIL_USER")
      password = env("SIGIL_API_KEY")
    }

    tenant_id = env("SIGIL_TENANT_ID")
  }
}
```

Fan out to multiple endpoints with different tenant IDs:

```alloy
sigil.write "default" {
  endpoint {
    url = "https://<sigil>"

    basic_auth {
      username = env("SIGIL_USER")
      password = env("SIGIL_API_KEY")
    }

    tenant_id = env("SIGIL_TENANT_ID")
  }

  endpoint {
    url       = "https://<sigil-2>"
    tenant_id = "staging"
  }
}

sigil.receive "default" {
  forward_to = [sigil.write.default.receiver]
}
```

</details>

### Notes to the Reviewer

- [Documentation](https://deploy-preview-alloy-6352-zb444pucvq-vp.a.run.app/docs/alloy/latest/reference/components/sigil/)

This is a follow-up for the closed https://github.com/grafana/alloy/pull/6029.

The Sigil API can return partial success: some generations in a batch are accepted and some are rejected. The Sigil components in Alloy do not merge the responses. The first successful receiver response wins. So if a client sends `[g1, g2]` and two receivers are configured, both receivers get the full `[g1, g2]` batch. Even if the second receiver rejects some generations, the client only sees receiver 1's body and the responses aren't merged.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


internal/component/sigil/receiver/receiver.go
```diff
@@ -0,0 +1,117 @@
...
+	defer c.mut.Unlock()
+
+	c.handler.update(args.ForwardTo, int64(args.MaxRequestBodySize))
+
+	serverNeedsRestarting := !reflect.DeepEqual(c.serverConfig, args.Server)
+	if !serverNeedsRestarting {
+		return nil
+	}
+
+	c.shutdownServer()
```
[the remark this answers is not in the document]
alexander-akhmetov: I see the same pattern in other places ([example](https://github.com/grafana/alloy/blob/35ea0a86151705361dc41d46ceaf2fd29f26021b/internal/component/prometheus/receive_http/receive_http.go#L156)), keeping this.

kalleep:
We need to update [this function](https://github.com/grafana/alloy/blob/fc98bb5980037e2989884e4ab5f3ca7463f865e8/internal/runtime/internal/controller/loader.go#L1012) with support for this new "signal". This is used by alloy front-end to render the graph and data flowing through it

internal/component/sigil/receiver/arguments.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(a.ForwardTo) == 0 {
+		return fmt.Errorf("forward_to must contain at least one receiver")
+	}
+	for i, recv := range a.ForwardTo {
+		if recv == nil {
+			return fmt.Errorf("forward_to[%d] must not be null", i)
+		}
+	}
+	if a.MaxRequestBodySize <= 0 {
+		return fmt.Errorf("max_request_body_size must be greater than zero")
```
kalleep: nit: No need to call fmt.Errorf here since we don't format anything. errors.New is fine here

internal/component/sigil/fanout.go
```diff
@@ -0,0 +1,98 @@
...
+
+// FanOut forwards req to every receiver in parallel. Each branch receives an
+// independently mutable clone of the request so processors can mutate freely.
+//
+// Result selection: FanOut returns the first non-nil response from a
+// successful branch, with branches identified by their index. If every branch
+// errored, it returns the joined errors. A partial failure (some branches
+// succeed, some error) is logged at warn level and recorded on metrics if
+// PartialFailures is set.
+func FanOut(
```
kalleep, line 25: Not sure what I think of this because we don't really do this pattern for any other pipeline (spawning go routine for each downstream component). We call components synchronously for each request.

The most common case will be only one downstream components and with this kind of implementation we get some unnecessary overhead:
1. One extra go routine 
2. Unnecessary clone

internal/component/sigil/fanout.go
```diff
@@ -0,0 +1,98 @@
...
+	ctx context.Context,
+	req *GenerationsRequest,
+	receivers []GenerationsReceiver,
+	logger *slog.Logger,
+	metrics FanOutMetrics,
+) (*GenerationsResponse, error) {
+
+	if len(receivers) == 0 {
+		return nil, errors.New("no downstream receivers configured")
```
kalleep: Should this really be an error?

internal/component/sigil/fanout.go
```diff
@@ -0,0 +1,98 @@
...
+	for i, recv := range receivers {
+		if recv == nil {
+			return nil, fmt.Errorf("downstream receiver %d is nil", i)
+		}
+	}
+
+	var wg sync.WaitGroup
+	for i, recv := range receivers {
+		wg.Add(1)
+		branch := req.Clone()
```
kalleep: When doing fanout we don't always have to clone. A good pattarn is to clone for every receiver except for the last one.

Something like this https://github.com/grafana/alloy/blob/main/internal/component/common/loki/consumer_fanout.go#L39-L50

Then if only one receiver is configured no clone will happen

internal/component/sigil/fanout.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// FanOut forwards req to every receiver in parallel. Each branch receives an
+// independently mutable clone of the request so processors can mutate freely.
+//
+// Result selection: FanOut returns the first non-nil response from a
+// successful branch, with branches identified by their index. If every branch
+// errored, it returns the joined errors. A partial failure (some branches
+// succeed, some error) is logged at warn level and recorded on metrics if
+// PartialFailures is set.
```
kgeckhart: Just wondering how this works with the API being able to support partial success, https://grafana.com/docs/grafana-cloud/machine-learning/ai-observability/reference/generation-ingest-api/#response? When there's a partial success is the status code success?

internal/component/sigil/types.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+
+	sigilv1 "github.com/grafana/sigil-sdk/go/proto/sigil/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// GenerationsReceiver is the interface connecting sigil.receive to
+// sigil.write and any intermediate processors.
+type GenerationsReceiver interface {
```
kgeckhart: Could we rename this to GenerationsForwarder to help differentiate it from the receiver component? We aren't super consistent with the names but for Loki/Prometheus we don't have components with the Receiver as "sources".

internal/component/sigil/write/write_test.go
```diff
@@ -0,0 +1,594 @@
...
+		})
+	}
+}
+
+func TestEndpointClient_RetriesRemoteTimeout(t *testing.T) {
+	var attempts atomic.Int32
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		if attempts.Add(1) == 1 {
+			time.Sleep(50 * time.Millisecond)
```
kgeckhart:
```suggestion
			<-r.Context().Done()
```

I think this might be safer vs depending on time. We might not have it flake but our CI can be really resource limited at times.

internal/component/sigil/codec.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return wire.MarshalExportGenerationsJSON(req.Request)
+}
+
+// ParseGenerationsResponse decodes a JSON response body. An empty body
+// produces a nil ExportGenerationsResponse pointer.
+func ParseGenerationsResponse(body []byte) (*sigilv1.ExportGenerationsResponse, error) {
+	if len(body) == 0 {
+		return nil, nil
+	}
```
kgeckhart: Is it valid for the API to return an empty response? The contract seems to signify that a body is expected.

internal/component/sigil/receiver/arguments.go
alexander-akhmetov: Updated: https://github.com/grafana/alloy/pull/6352/changes/c6ee878b6e0e476e3041dbc711c6c2225e839d70

internal/component/sigil/fanout.go
alexander-akhmetov: Makes sense, updated to be synchronous and clone only when there are more than 1 receiver: https://github.com/grafana/alloy/pull/6352/changes/45f3bcf30f19ebfdb2f3ed9e865d42d5f6fd12b3 and https://github.com/grafana/alloy/pull/6352/changes/aa031a684bc71e940d1100369d5e28807a5d6115. Also removed the metric that wasn't needed anymore and now it returns an error any one of the receivers errors, same as in Loki components.

internal/component/sigil/fanout.go
alexander-akhmetov: Updated: https://github.com/grafana/alloy/commit/45f3bcf30f19ebfdb2f3ed9e865d42d5f6fd12b3

internal/component/sigil/types.go
alexander-akhmetov: Renamed: https://github.com/grafana/alloy/pull/6352/changes/09cbd09f119b5bda0dc8e0c56caa602586df7f39

internal/component/sigil/write/write_test.go
alexander-akhmetov: Updated: https://github.com/grafana/alloy/pull/6352/changes/6df426f2c1cef5848766c3583861d284a95f4751

internal/component/sigil/fanout.go
alexander-akhmetov: Yes, the code from Sigil itself is `202 Accepted`, but I also removed this part and made it similar to Loki components when if there is at least one error when sending to receivers, it returns an error here: https://github.com/grafana/alloy/commit/45f3bcf30f19ebfdb2f3ed9e865d42d5f6fd12b3 and https://github.com/grafana/alloy/pull/6352/changes/aa031a684bc71e940d1100369d5e28807a5d6115

internal/component/sigil/receiver/handler.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		h.logger.Debug("failed to parse generation export request", "err", err)
+		http.Error(w, err.Error(), status)
+		return
+	}
+
+	resp, fanErr := sigil.FanOut(r.Context(), req, forwardTo, h.logger, h.fanOutMetrics())
+
+	// If every branch failed, return 502.
+	if resp == nil && fanErr != nil {
```
dehaansa: The remainder of the logic in this function all feels strange/not clean.

- We only care about fanErr if resp is nil?
- If the status code appears invalid, we respond with a 200 to client unless marshal fails?
- if the resp is nil (and err is also nil), we still try to marshal it and respond?

It seems like we should handle a nil resp consistently and handle a fanErr even if resp != nil. If I'm misunderstanding, maybe we need some comments for added clarity.

internal/component/sigil/fanout.go
```diff
@@ -0,0 +1,98 @@
...
+		if firstResp == nil {
+			firstResp = r.resp
+		}
+	}
+
+	if failed > 0 && succeeded > 0 {
+		if logger != nil {
+			logger.Warn("partial fan-out failure", "succeeded", succeeded, "failed", failed, "err", errs)
+		}
+		if metrics.PartialFailures != nil {
```
dehaansa: Rather than track partial failures, should we track total failures and total successes? Otherwise the metric will be the same with 1 success 10 failures (given 11 endpoints, 10 of them failing) as 10 successes 1 failure.

alexander-akhmetov: I changed that to track total failures and fail on any error, so there is no `metrics.PartialFailures`: https://github.com/grafana/alloy/commit/45f3bcf30f19ebfdb2f3ed9e865d42d5f6fd12b3#diff-18e5a7d2a86bdb85697f7a0d68476991c2275dfbb0966d621a489e4600dada04.

internal/component/sigil/receiver/handler.go
alexander-akhmetov: It's related to the partial success scenaio that I removed and simplified to match pyroscope: https://github.com/grafana/alloy/pull/6352/changes/45f3bcf30f19ebfdb2f3ed9e865d42d5f6fd12b3 and https://github.com/grafana/alloy/pull/6352/changes/aa031a684bc71e940d1100369d5e28807a5d6115. Now first upstream error is propagated and on success 202 is returned.

internal/component/sigil/codec.go
alexander-akhmetov: Not really. I wasn't sure if it should error but it's probably cleaner since it's not expected to have empty body. Updated: https://github.com/grafana/alloy/pull/6352/changes/5a68b78920daaa071f99b6c86a614890c5782e37

internal/component/sigil/fanout.go
kgeckhart: Nice thank you! How does the sdk handle the partial error? Like if the response indicates some generations failed is it going to retry them or just log and continue? I don't know that you need to do anything differently here but if there's some behavior triggered by partial failure in the response I think we will end up silently dropping those if they happened in the second response.

alexander-akhmetov: The SDK doesn't retry per-generation failures on partial errors, currently it just logs them and continues. The assumption is that these are permanent validation errors. That's not always true though, sometimes backend errors also come back as failures, so we may want to change this, but I'm not sure yet on which side.

I didn't try to merge errors from multiple receivers since the SDK doesn't do much in this case anyway, and merging gets tricky because a duplicated generation is also an error. I thought about returning a generation as failed if it failed in at least one receiver, but not sure if it's the right way 🤷 For example if we retried it wouldn't work well since duplicates aren't allowed and merging + retrying in this case would create duplicates. We need to return something from the backend to show if it's a retryable error or not per generation.

kgeckhart: Yeah I don't think there are any amazing solutions here. My only lean towards trying to merge is to ensure that all generations that failed are logged somewhere. If a user expects the sdk to log them it might be weird that it's not comprehensive? 

No matter what it's probably worth a comment so it's clear that whatever direction we go was chosen intentionally.

alexander-akhmetov: My initial thought was that marking a generation as failed when it succeeded in some cases wasn't correct. But after thinking about it more, I think you're right.

Especially since we need to document this behavior anyway, it feels clearer to mark a generation as "failed" if it fails for any configured receiver, and as "succeeded" only if it succeeds for all configured receivers. I'll make this change.

alexander-akhmetov: Updated: https://github.com/grafana/alloy/pull/6352/changes/4a672580b7ea94d5d979214d751666f4854cdbe4

kgeckhart:
I think I'm good with where things stand today @kalleep / @dehaansa did either of you have anything else that's unresolved at this point?

dehaansa (approved):

kalleep (approved):

alexander-akhmetov:
Hey, sorry for the inconvenience and the back and forth on this. I'm looking into the OTel path again on our side and we made a few changes for that. There are still some open questions with a few of the signals we send, so I'd like to experiment more before committing to these experimental components in Alloy, to be 100% sure before merging. Let's hold this PR for now until I've worked out whether the OTLP path is possible now. I'll follow up either way.

PR #6362, opened 2026-05-29
chore: Migrate generate-module-dependency to tools cli

kalleep:
This is no longer it's own module but a sub command in our tools cli, also made --root flag optional and it will resolve git root if not passed.


blewis12 (approved):
nice, LGTM 👍

PR #6363, opened 2026-05-29
chore(deps): Remove outdated pyroscope/ebpf module dependency

korniltsev-grafanista:


thampiotr (approved):
Confirmed offline that this removes only undocumented metrics so does not require documentation updates.

PR #6364, opened 2026-05-29
chore(renovate): Lower abandonment threshold

thampiotr:
## Summary
- set Renovate `abandonmentThreshold` to `365 days`
- surface abandoned dependency signals in the dependency dashboard sooner

## Test plan
- [x] confirmed `.github/renovate.json5` includes `\"abandonmentThreshold\": \"365 days\"`

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -17,6 +17,9 @@
   // Updates must be published for at least 7 to be considered
   "minimumReleaseAge": "7 days",
 
+  // Surface abandonment signal in dependency dashboard after one year
+  "abandonmentThreshold": "365 days",
+
   "prConcurrentLimit": 5,
 
   // Move dependency files forward along with lock files


.github/renovate.json5
```diff
@@ -17,6 +17,9 @@
   // Updates must be published for at least 7 to be considered
   "minimumReleaseAge": "7 days",
 
+  // Surface abandonment signal in dependency dashboard after one year
```
blewis12, line 20: what was the default value before this?

thampiotr: I thought it was 2 years, but couldn't find where is this set 😅 
Finally found that we have a [base config here](https://github.com/grafana/deployment_tools/blob/master/ksonnet/lib/renovate-self-hosted/renovate-base.libsonnet#L84).
So yeah, it's 2 years by default.

thampiotr: Hmm, but perhaps we can't override it because it's in a `force` section? 🤔 
I also don't see the footers that are added in deployment_tools config on our GH issue. 
@jharvey10 would you know how this is wired? do we use the defaults from deployment_tools?

jharvey10: I do see the footer on PRs, for example this one: 
image

but I think the reason it doesn't show on the dependency dashboard is that it's so huge that it literally just breaks things 😆 . We may even be exceeding the max length of a PR body or something. I'm like 90% sure even the output that we do have is truncated.

Ah yes... confirmation.
image

If ever there was a dependency hell, we are in it and have front row seats 😈.

I believe you're correct though that those forced settings are things we can't override even if we tried; they're baked in at the runner host-level. We could potentially petition for lowering that globally to 1 year or seeing if there's options for them to enforce a max value while still allowing teams to lower it from there.

PR #6368, opened 2026-05-29
ci: Use "npm ci" instead of "npm install"

ptodev:
It's considered a [best practice](https://github.com/lirantal/npm-security-best-practices#6-use-npm-ci) to use `npm ci` since it checks the lock file and returns an error if there are inconsistencies in the checksum and the actual files on disc.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -84,7 +84,7 @@ jobs:
         install_args: node
 
     - name: Install dependencies
-      run: npm install
+      run: npm ci
 
     - name: Lint
       run: npm run lint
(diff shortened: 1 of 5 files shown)


jharvey10 (dismissed):
This looks great! One bit of food for thought: I have some stuff I run in Google Cloud using their buildpacks, and the command they use is:

```
npm ci --quiet --no-fund --no-audit
```

I don't think we need quiet, but no fund and no audit probably make a lot of sense to add.

Funding is not something we need to think about in CI builds, and auditing is already being done by renovate/dependabot/other scanning.

ptodev:
@jharvey10 sounds good, thank you for the tip! I added `--no-audit --no-fund`. I don't see a downside in having them.

jharvey10 (approved):

PR #6369, opened 2026-05-29
ci: Sign release container images with cosign

kgeckhart:

Keyless-signs Alloy's release images (Linux `alloy` + `alloy-boringcrypto`, Windows `alloy`) with cosign via GitHub Actions OIDC. 

### Pull Request Details

Signing is done via a new `sign-and-attest.yml`. Using the OIDC method this matter long term because it becomes part of the verification process. We cannot change after it is introduced without changing the verification steps a user must use making it a breaking change. There are other options that are a bit more complicated but could decouple us. I went with this one because it's what OTel does today.

| Option | What it is | Identity you verify against | Used by |
|---|---|---|---|
| **GitHub Actions OIDC** (this PR) | Keyless. The signing workflow is the identity — no key or stored secret. | `https://github.com/<org>/<repo>/.github/workflows/<file>@<ref>`<br>issuer `https://token.actions.githubusercontent.com` | OpenTelemetry Collector, Cilium, Trivy, Chainguard |
| **Service account / federated identity** | Keyless, but the workflow federates to a stable cloud or custom OIDC identity instead of the workflow. Works with any Fulcio-supported issuer (GCP, AWS, Azure, custom); prominent examples are GCP. | `signer@<project>.iam.gserviceaccount.com`<br>issuer `https://accounts.google.com` (GCP example) | Kubernetes, distroless, Knative |
| **Self-managed key (KMS)** | Not keyless. Long-lived key; consumers pin the public key. | a public key (`cosign verify --key …`) | teams wanting full key ownership |

### Notes to the Reviewer

Some more references on image signing,

Keyless signing model (Fulcio + Rekor)
- How Sigstore works — https://docs.sigstore.dev/about/overview/
- Cosign signing overview (the keyless flow) — https://docs.sigstore.dev/cosign/signing/overview/
- Fulcio — the CA that issues the short-lived cert from an OIDC identity (not a signing service) — https://docs.sigstore.dev/certificate_authority/overview/ + how OIDC is used https://docs.sigstore.dev/certificate_authority/oidc-in-fulcio/#source-repository
- Rekor — the transparency log that timestamps the signature so a short-lived cert stays verifiable — https://docs.sigstore.dev/logging/overview/

Signing & verifying container images
- Signing with containers — https://docs.sigstore.dev/cosign/signing/signing_with_containers/
- Verifying signatures (--certificate-identity / --certificate-oidc-issuer) — https://docs.sigstore.dev/cosign/verifying/verify/
- Cosign signature spec — documents the tag-based sha256-<digest>.sig scheme (the thing the GAR→Docker Hub mirror relies on) — https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md

cosign in GitHub Actions
- cosign-installer action (what we pin) — https://github.com/sigstore/cosign-installer


.github/workflows/release-publish-alloy-artifacts.yml
```diff
@@ -75,6 +75,36 @@ jobs:
...
+    needs: publish_windows_container
+    if: ${{ needs.publish_windows_container.outputs.image != '' }}
+    uses: ./.github/workflows/sign-and-attest.yml
+    permissions:
+      contents: read
+      id-token: write
+    with:
+      image: ${{ needs.publish_windows_container.outputs.image }}
+
   build_alloy:
```
blewis12, line 108: Would it be possible to add an action towards the end here that verifies that the docker image has indeed been signed? Maybe we'd need to wait for [this action](https://github.com/grafana/shared-workflows/pull/1956) to be merged so that we can wait for mirroring to be complete - but then it would be a matter of calling `cosign verify ...` right?

blewis12: _or_ we could just verify against the image in GAR and not dockerhub, but it would certainly be nice to test against the end artifact that users download

.github/workflows/publish-alloy-linux.yml
```diff
@@ -18,6 +18,10 @@ on:
         required: false
         type: string
         default: ""
+    outputs:
+      image:
+        description: "Digest-pinned ref of the pushed release image (e.g. <repo>@sha256:...). Empty when not pushed or for devel/PR builds."
```
blewis12, line 23: Do we not also want to set up artifact signing even for `devel` images, since those run on our internal infrastructure? Or is that something we intend to do down the line?

.github/workflows/release-publish-alloy-artifacts.yml
kgeckhart: I added it targeting GAR, it's especially helpful with immutable releases so we know it's going to be broken before we publish

.github/workflows/publish-alloy-linux.yml
kgeckhart: Good question, I can't really see a good reason to do it today unless we also want to only run signed images in our environments. We can always add it later if it's something we want to target. I imagine most security conscious customers who want signed images won't be running dev builds but who knows.

.github/workflows/release-publish-alloy-artifacts.yml
blewis12: nice, that sounds good to me!

.github/workflows/publish-alloy-linux.yml
blewis12: that's true, i'm thinking of if there's some scenario where an attacker gains access to our dockerhub credentials and pushes some devel looking image to our dockerhub. That said though, the only way they could deploy that to our internal infra would be if our CI was compromised too and they ran our workflow which submits a PR against dev tools.

I think if anything it could just be an extra hardening measure, and we can verify the images we run on our internal infrastructure are signed across our environments using something like [kyverno](https://kyverno.io/)

kgeckhart: We could certainly do something like that in a follow up. I think signing our released images is a good place to start.

.github/workflows/sign-and-attest.yml
thampiotr, line 1: the file name indicates we are attesting, but we are not yet doing this here - is there a plan to add it or should we rename the file?

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,91 @@
...
+      # Fail the run if the just-written signature doesn't verify, so a broken
+      # signing job is caught before the draft release is published — with
+      # immutable releases the image can't be re-signed after publication. This
+      # checks the signature in GAR; propagation to Docker Hub via the mirror is
+      # verified separately (smoke test), since the mirror copies asynchronously.
+      env:
+        IMAGE: ${{ inputs.image }}
+        REPO: ${{ github.repository }}
+      run: |
+        cosign verify \
```
thampiotr, line 67: Should we add some docs that would explain how to verify the images for users?

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,91 @@
...
+#     --recursive later is a one-line change.
+#
+#   * Legacy tag-based signatures, not OCI 1.1 referrers. By default cosign
+#     writes the signature as a `sha256-<digest>.sig` tag on the same repo. That
+#     matters because release images push to GAR and the gar-image-mirror
+#     replicates to Docker Hub by copying tags for allowlisted images — so the
+#     .sig tag rides the mirror downstream for free. OCI 1.1 referrers are
+#     untagged and the mirror filters them out, so they would never reach Docker
+#     Hub. Switching to referrers requires BOTH --registry-referrers-mode=oci-1-1
+#     and COSIGN_EXPERIMENTAL=1; we set neither, so we get tag-based signatures.
```
thampiotr: I'm not sure I follow this comment, maybe I'm missing some context, but perhaps it can be improved.

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,91 @@
...
+#
+# Two deliberate choices, load-bearing enough to call out:
+#
+#   * Index-only signing. Alloy builds one multi-platform manifest via buildx,
+#     so `cosign sign <index-digest>` signs the manifest-list as a whole. We do
+#     NOT pass --recursive (which would also sign each per-arch child). This
+#     covers every consumer pulling by tag or by index digest; it does not cover
+#     pinning an individual arch's leaf digest, which we have no evidence Alloy
+#     consumers do. Matches Cilium and grafana/docker-otel-lgtm; flipping to
+#     --recursive later is a one-line change.
```
thampiotr:
```suggestion
#     consumers do. Matches Cilium and grafana/docker-otel-lgtm;
```

seems more like a note to self 😅

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,91 @@
...
+        IMAGE: ${{ inputs.image }}
+      run: |
+        cosign sign --yes "${IMAGE}"
+
+    - name: Verify signature
+      # Fail the run if the just-written signature doesn't verify, so a broken
+      # signing job is caught before the draft release is published — with
+      # immutable releases the image can't be re-signed after publication. This
+      # checks the signature in GAR; propagation to Docker Hub via the mirror is
+      # verified separately (smoke test), since the mirror copies asynchronously.
```
thampiotr: "propagation to Docker Hub via the mirror is verified separately (smoke test), since the mirror copies asynchronously."

Is this part of this PR? I couldn't find the smoke test for Docker Hub.

thampiotr (commented):
That's going to be a great addition for those that want to verify integrity!

.github/workflows/sign-and-attest.yml
kgeckhart: That will be handled in a follow up, didn't want to add documentation for something that doesn't exist yet.

.github/workflows/sign-and-attest.yml
kgeckhart: We can't verify it without having the action to wait for the sync which can happen after we close that gap.

.github/workflows/sign-and-attest.yml
kgeckhart: So with the flow we are using the identity that signs our releases ends up being our workflow and even though we are not attesting today the plan was to follow up with VEX/SBOM which will be attested. Changing the workflow name will end up changing the cert and the verification. You can see how OTel handles it here, https://github.com/open-telemetry/opentelemetry-collector/blob/main/README.md#verifying-the-images-signatures

.github/workflows/sign-and-attest.yml
thampiotr: So the comment should be removed?

.github/workflows/publish-alloy-linux.yml
blewis12: agree 👍

.github/workflows/sign-and-attest.yml
kgeckhart: I adjust it to acknowledge we can't verify docker hub because it's async

kgeckhart:
We decided to start with going through dev first #6454 so we can try to build out a full verification workflow. 

This PR will stay as an easy reference and I'll keep it in sync if it diverges.

PR #6370, opened 2026-05-29
ci: Rework backport workflow to not use workflow_run

ptodev:
It's best to avoid `workflow_run` due to [security issues](https://securitylab.github.com/resources/github-actions-new-patterns-and-mitigations/). I think the current workflow we have is not exploitable since it checks out the main branch of 'grafana/alloy' and not a forked branch, but it's good to reduce the attach surface anyway by removing `workflow_run` altogether. That way a slight misconfiguration won't suddenly make the github action more vulnerable.

The downside is that for already merged PRs @clayton-cornell (or another member of the team) would have to trigger the workflow manually instead of just adding a label.


clayton-cornell:
Manual workflow triggering for backporting docs is already a thing on community PRs.  I don't see any real issues with manual triggers as long as it's a known thing and we/I don't forget to click through when it's needed :-)

Do I understand this right that the "new" manual triggering is only if we decide to backport a PR after it is already merged into `next`?

jharvey10:
I understand the intent of this PR, but I want to call out a few things/concerns.

These were my acceptance criteria for the backport workflow rewrite I did a few months back:

1. It must work for both internal and community PRs.
2. It must be low friction to encourage its use.
3. It must have clear indications of failure.
4. It must be safe.

The as-is implementation solves for all four of these, but can raise some red flags when `workflow_run` is taken in isolation.

The suggested implementation solves (4), but that comes at the expense of (2) and (3).

Because of that, there's an inherent tradeoff. We can remove the potentially unsafe trigger, but worsen our own release ergonomics in the process. Or, we can accept the risk of the trigger and maintain the existing ease of use.

It's important to understand *why* this might be unsafe, and how the existing implementation might be exploited by an attacker. As you point out, the trigger is only unsafe because it has the *potential* to be used incorrectly, not that it's actively being used incorrectly. To that end, this backport workflow has undergone significant [auditing and scrutiny](https://grafanalabs.enterprise.slack.com/archives/C0A7T2SSN6S). This is why the workflow is divided as it is. The trigger's responsibility is to gather only the minimum amount of information from the event which triggered the workflow (the pull request number) and to "pass" that to the privileged workflow which does all of its own data gathering directly from the alloy repo for the remainder of its logic.

Yes, someone could propose a PR that breaks this trust, but we have both zizmor and human reviewers as layers to catch that. Not to mention that this would need to make it all the way through undetected to release publication to actually have an impact.

The patterns I've observed from devs on our team are that backport labels often get added after PRs are already merged; typically as we start thinking about an upcoming patch and deciding if there's more things to add there. This is especially noticeable for dependency updates and CVE mitigations.

The proposed change, while it does solve for (4), turns this common team pattern into a multi-step process: You must label the PR and you must also either run a workflow from the UI or run a command from the CLI to trigger the backport. In either of those cases, you've done a context switch off of the PR page, which is where the success or failure result of the backport is most obvious.

ptodev:
@jharvey10 thank you for reviewing this PR! I agree and I'll close it.

This was just an exploration that I was doing when I first learned about security issues with `workflow_run`. I didn't know this workflow went through so much rigorous testing :) My main worry is that someone who doesn't know all this may change the workflow in a bad way without knowing, but as you say we have plenty of checks to make sure this doesn't happen.

PR #6371, opened 2026-05-30
fix(discovery.dns): use Windows system resolver

### LINKED ISSUE #6238: `discovery.dns` fails on Windows attempting to read /etc/resolv.conf
### Component(s)

discovery.dns

### What's wrong?

On Windows, `discovery.dns` fails to perform any DNS lookup because the underlying resolver tries to read `/etc/resolv.conf`, which does not exist on Windows. The error happens for every record type tested (SRV, A), so it appears to be a generic resolver-initialisation problem rather than something specific to SRV.

System DNS resolution is healthy on the host — `Resolve-DnsName` / `nslookup` for the same record from the same machine returns the expected answer.

### Steps to reproduce

1. On a Windows host with a working DNS configuration (e.g. an AD-joined workstation), save the following as `dns-srv-test.alloy`:

```alloy
logging {
  level  = "debug"
  format = "logfmt"
}

discovery.dns "domain_controllers" {
  names            = ["_ldap._tcp.dc._msdcs.example.com"]
  type             = "SRV"
  refresh_interval = "30s"
}

prometheus.scrape "dns_probe_srv" {
  targets         = discovery.dns.domain_controllers.targets
  job_name        = "dns_srv_test"
  scrape_interval = "1m"
  scrape_timeout  = "10s"
  forward_to      = []
}
```

2. Run Alloy:

```powershell
& "C:\Program Files\GrafanaLabs\Alloy\alloy-windows-amd64.exe" run dns-srv-test.alloy
```

3. Observe the error in the logs / live debugging tab for `discovery.dns.domain_controllers`.

## What did you expect to see?

The component should resolve the SRV record using the system DNS configuration (the resolvers configured via Windows network settings, the same ones used by `Resolve-DnsName` / `nslookup`) and emit one target per SRV result.

## What did you see instead?

```
ts=2026-05-12T14:45:57.2977654Z level=error msg="Error refreshing DNS targets" component_path=/ component_id=discovery.dns.domain_controllers err="could not load resolv.conf: open /etc/resolv.conf: The system cannot find the path specified."
```

The same error appears when `type` is changed to `A` (with an explicit `port`) and `names` is a public hostname like `example.com`, suggesting the failure is at resolver init, not SRV-specific.

### System information

Windows 11 Enterprise 10.0.26200

### Software version

v1.16.1

### Configuration

```text
logging {
  level  = "debug"
  format = "logfmt"
}

discovery.dns "domain_controllers" {
  names            = ["_ldap._tcp.dc._msdcs.example.com"]
  type             = "SRV"
  refresh_interval = "30s"
}

prometheus.scrape "dns_probe_srv" {
  targets         = discovery.dns.domain_controllers.targets
  job_name        = "dns_srv_test"
  scrape_interval = "1m"
  scrape_timeout  = "10s"
  forward_to      = []
}
```

### Logs

```text
ts=2026-05-12T14:45:57.2977654Z level=error msg="Error refreshing DNS targets" component_path=/ component_id=discovery.dns.domain_controllers err="could not load resolv.conf: open /etc/resolv.conf: The system cannot find the path specified."
```

### Tip

ADITYA-CODE-SOURCE:
## Summary
- add a Windows-specific `discovery.dns` implementation that uses Go's system resolver instead of Prometheus DNS SD's `/etc/resolv.conf` path
- keep the existing Prometheus DNS SD behavior on non-Windows platforms
- add tests for Windows target formatting and resolver error handling

## Testing
- `go test common.go common_test.go` in `internal/component/discovery/dns`
- package-wide Go tests could not complete in this environment because the Windows host ran out of memory/paging file during compilation

Closes #6238.

### DIFF
--- a/internal/component/discovery/dns/common.go
+++ b/internal/component/discovery/dns/common.go
@@ -1,0 +1,62 @@
+package dns
+
+import (
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/common/model"
+)
+
+const (
+	dnsNameLabel            = model.MetaLabelPrefix + "dns_name"
+	dnsSrvRecordPrefix      = model.MetaLabelPrefix + "dns_srv_record_"
+	dnsSrvRecordTargetLabel = dnsSrvRecordPrefix + "target"
+	dnsSrvRecordPortLabel   = dnsSrvRecordPrefix + "port"
+	dnsMxRecordPrefix       = model.MetaLabelPrefix + "dns_mx_record_"
+	dnsMxRecordTargetLabel  = dnsMxRecordPrefix + "target"
+	dnsNsRecordPrefix       = model.MetaLabelPrefix + "dns_ns_record_"
+	dnsNsRecordTargetLabel  = dnsNsRecordPrefix + "target"
+
+	dnsMetricsNamespace = "prometheus"
+)
+
+func hostPort(host string, port int) model.LabelValue {
+	return model.LabelValue(net.JoinHostPort(host, strconv.Itoa(port)))
+}
+
+func trimRootedName(name string) string {
+	return strings.TrimRight(name, ".")
+}
+
+func buildIPTarget(name string, ip net.IP, port int) model.LabelSet {
+	return model.LabelSet{
+		model.AddressLabel: hostPort(ip.String(), port),
+		dnsNameLabel:       model.LabelValue(name),
+	}
+}
+
+func buildSRVTarget(name string, record *net.SRV) model.LabelSet {
+	return model.LabelSet{
+		model.AddressLabel:      hostPort(trimRootedName(record.Target), int(record.Port)),
+		dnsNameLabel:            model.LabelValue(name),
+		dnsSrvRecordTargetLabel: model.LabelValue(record.Target),
+		dnsSrvRecordPortLabel:   model.LabelValue(strconv.Itoa(int(record.Port))),
+	}
+}
+
+func buildMXTarget(name string, record *net.MX, port int) model.LabelSet {
+	return model.LabelSet{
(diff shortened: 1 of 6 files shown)


kgeckhart:
Hi, while we appreciate attempting to solve this issue we would prefer this behavior to be adopted upstream and brought in to alloy that way vs alloy owning a secondary windows specific implementation.

Edit: This was mentioned in the original issue as well https://github.com/grafana/alloy/issues/6238#issuecomment-4439192550

ADITYA-CODE-SOURCE:
> Hi, while we appreciate attempting to solve this issue we would prefer this behavior to be adopted upstream and brought in to alloy that way vs alloy owning a secondary windows specific implementation.
> 
> Edit: This was mentioned in the original issue as well [#6238 (comment)](https://github.com/grafana/alloy/issues/6238#issuecomment-4439192550)

Thanks, that makes sense.
I traced the failing path to the upstream Prometheus DNS discovery code used by Alloy, where `lookupWithSearchPath` calls `dns.ClientConfigFromFile("/etc/resolv.conf")` on Windows.
I’ll redirect this toward the upstream fix instead of keeping a Windows-specific Alloy implementation. Do you prefer that to target `grafana/prometheus` first, or `prometheus/prometheus` directly?

kalleep:
>Do you prefer that to target grafana/prometheus first, or prometheus/prometheus directly?

prometheus/prometheus, we only use grafana/prometheus when we need to apply fixes that either have not been accepted upstream or if we need to cherry-pick a fix that have not been released yet. I imagine that is a fix is made in the upstream library that promethues used it's just a dependency bump in promethues

kalleep:
Since the fix needs to be implemented upstream I will close this pr

PR #6373, opened 2026-05-30
docs: Update support bundle flag prefix

jakobzudrell:
Add the `server.http` prefix to the `disable-support-bundle` as this is what (at least) version 1.16.1 requires.

This PR adds the `server.http` prefix to the `disable-support-bundle` in the `run` reference documentation.

### PR Checklist

- [x ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/cli/run.md
+++ b/docs/sources/reference/cli/run.md
@@ -41,9 +41,9 @@ The following flags are supported:
 * `--server.http.memory-addr`: Address to listen for [in-memory HTTP traffic][] on (default `"alloy.internal:12345"`).
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `"127.0.0.1:12345"`).
 * `--server.http.ui-path-prefix`: Base path where the UI is exposed (default `"/"`).
+* `--server.http.disable-support-bundle`: Disable [support bundle][] endpoint (default `false`).
 * `--storage.path`: Base directory where components can store data (default `"data-alloy/"`).
 * `--disable-reporting`: Disable [data collection][] (default `false`).
-* `--disable-support-bundle`: Disable [support bundle][] endpoint (default `false`).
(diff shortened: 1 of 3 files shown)


dehaansa (approved):
Thanks for catching this! It looks like this is also incorrect in two other places, `docs/sources/troubleshoot/support_bundle.md` and an error message in `internal/service/http/http.go`. Would you like to update the PR with those two fixes as well?

jakobzudrell:
Done, sorry could've done a full text search in the first place 😅

I am not quite sure, but is it possible that `internal/static/config.go` still expects the flag in the global scope and not within the `server.http`? 

https://github.com/grafana/alloy/blob/3c48891b67e4b0bdaf079d19a8f11b164f59c0d4/internal/static/config/config.go#L46-L59

https://github.com/grafana/alloy/blob/3c48891b67e4b0bdaf079d19a8f11b164f59c0d4/internal/static/config/config.go#L322

clayton-cornell (approved):

dehaansa:
> Done, sorry could've done a full text search in the first place 😅
> 
> I am not quite sure, but is it possible that `internal/static/config.go` still expects the flag in the global scope and not within the `server.http`?
> 
> https://github.com/grafana/alloy/blob/3c48891b67e4b0bdaf079d19a8f11b164f59c0d4/internal/static/config/config.go#L46-L59
> 
> https://github.com/grafana/alloy/blob/3c48891b67e4b0bdaf079d19a8f11b164f59c0d4/internal/static/config/config.go#L322

The `internal/static` subdirectory is code copied in from the previous Grafana Agent project, and is only relevant when called from another place. Some if it is definitely cruft that should be cleaned up, just hasn't been a priority! I saw that as well when searching and feel pretty confident that it's not relevant.

dehaansa:
Thanks for the contribution!

PR #6377, opened 2026-06-01
chore: Migrate lint-go to tools cli

kalleep:
### Pull Request Details
Migrate `lint-go` bash script to a command in our tools cli `./tools lint go`. The script would exit on first module that failed but the command will run for all modules before exiting.

I did some other refactors/cleanup in this pr:
1. Common RootFlag under `tools/internal/cli`. A lot of our commands takes a root that we should trigger command for and had similar way to resolve default root (git root).
2. Function for "discovering" files in `tools/internal/discover` - We have several commands that tries to e.g. find all directories that contains a go.mod file or looking for all docker files.

### PR Checklist

- [x] Documentation added
- [ ] Tests updated


kalleep:
@codex review

tools/lint/golint/command.go
```diff
@@ -0,0 +1,65 @@
...
+	var errs []error
+	for _, dir := range result.Dirs() {
+		fmt.Println("Lint: ", dir)
+		cmd := exec.Command(bin, "run", "-v", "--timeout=10m")
+		cmd.Dir = dir
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			errs = append(errs, fmt.Errorf("lint %s", dir))
+		}
```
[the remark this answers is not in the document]
kalleep: This will just add `exit status 1` to every error, not that useful

tools/lint/golint/command.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	var errs []error
+	for _, dir := range result.Dirs() {
+		fmt.Println("Lint: ", dir)
+		cmd := exec.Command(bin, "run", "-v", "--timeout=10m")
+		cmd.Dir = dir
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
```
thampiotr, line 59: This `err` should be wrapped and preserved.

thampiotr (commented):
One thing I've noticed, otherwise good

tools/lint/golint/command.go
kalleep: The only error we will get here is `exit status 1`. We already stream stderr and stdout from the command.

Still think it's valuable?

kalleep: Okay discussed in slack added wrapping in [d6e1d29](https://github.com/grafana/alloy/pull/6377/commits/d6e1d29a2cca5f2141eaf8856a6fa4f5b135c32c)

thampiotr (approved):

PR #6378, opened 2026-06-01
chore: Migrate lint-shell to tools cli

kalleep:
### Pull Request Details
Replace `lint-shell` script with a command in tools cli `tools lint shell`.

Previously we only linted scripts certain directories (docs, scripts, packaging, .github) but now we scan the full repo, skipping node_modules and vendor.

To do this I changed to that `discover.File` takes a match function instead and pass a function that will match on file extensions. This discovered more scripts that we did not lint before and I have updated them to pass linter.

I also updated the regex to match shebangs with trailing arguments, before it was a strict match.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


operations/helm/scripts/rebuild-tests.sh
```diff
@@ -3,24 +3,31 @@
...
       # Render chart
       if [[ "${TESTNAME}" == *"vertical"* ]]; then
-        helm template --namespace default --kube-version 1.26 --debug ${CHART_NAME} ${CHART_DIR} -f ${FILE_PATH} --output-dir ${TEST_DIR}/${TESTNAME} --set '$chart_tests=true' --api-versions autoscaling.k8s.io/v1
+        # shellcheck disable=SC2016 
+		#'$chart_tests' is a literal Helm value key, not a shell variable.
+        helm template --namespace default --kube-version 1.26 --debug "${CHART_NAME}" "${CHART_DIR}" -f "${FILE_PATH}" --output-dir "${TEST_DIR}/${TESTNAME}" --set '$chart_tests=true' --api-versions autoscaling.k8s.io/v1
       else
-        helm template --namespace default --kube-version 1.26 --debug ${CHART_NAME} ${CHART_DIR} -f ${FILE_PATH} --output-dir ${TEST_DIR}/${TESTNAME} --set '$chart_tests=true'
+        # shellcheck disable=SC2016
+		#'$chart_tests' is a literal Helm value key, not a shell variabl.
```
ptodev, line 29: Isn't there a way to define '$chart_tests' differently so that it doesn't get flagged?

tools/lint/shelllint/command.go
```diff
@@ -0,0 +1,87 @@
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("shellcheck failed: %w", err)
+	}
+
+	return nil
+}
+
+var shellShebang = regexp.MustCompile(`^#!.*[/ ](sh|bash)(\s|$)`)
+
+func hasShebang(path string) (bool, error) {
```
ptodev, line 75: This would be cleaner and more unit testable if it doesn't try to open the file.

ptodev (approved):
LGTM, just a couple of minor comments.

Issue #5878, opened 2026-03-24
docs: Look at the need for hardening topics in the Alloy docs.


clayton-cornell:
### Add a description

Is there a need for hardening topics in the Alloy docs? What is the effort vs return for our customers?

70% of the information is already there in the existing docs, but it's scattered all over.

ChatGPT analysis of the potential gap:

--- 

## A general hardening topic (new conceptual doc)

This is the biggest missing piece: a unifying mental model.

Section: “Core security principles for Alloy”

Pull directly from existing behavior and make it explicit:

* Least privilege
  * Run as non-root wherever possible (already documented in K8s + containers)
* Minimize network exposure
  * HTTP server binds to localhost by default (already true, just not framed as security)
* Protect secrets
  * Don’t store credentials in config; use env vars / secret stores
* Isolate telemetry access
  * Alloy often reads host logs, metrics, or kernel data → high sensitivity
* Component-specific risk
  * eBPF components require elevated privileges (already noted, but buried)

👉 This section is mostly reframing, not new content.

## Add a “Hardening checklist” (high-value, low effort)

Users love this, and it doesn’t exist today.

Example:

Minimal production hardening checklist

- Run Alloy as non-root (UID 473 where supported)
- Set readOnlyRootFilesystem: true (containers)
- Disable privilege escalation
- Restrict HTTP server to localhost or private network
- Use TLS for all remote writes
- Store secrets outside config files
- Scope Kubernetes RBAC to minimum required
- Avoid running eBPF components unless needed
- Use rootless containers where possible

👉 Every single one of these already exists somewhere in docs.

## Consolidate Kubernetes hardening (existing → structured)

Right now this is the strongest area, but fragmented.

Create a single section:

“Harden Alloy on Kubernetes”

Combine:

- Non-root doc
- OpenShift doc
- SecurityContext examples

Structure it like:

- Workload security
  - runAsUser, runAsGroup, fsGroup
  - readOnlyRootFilesystem
  - allowPrivilegeEscalation: false
- Cluster security
  - RBAC scoping guidance (currently implicit)
  - Namespace isolation
- Network security
  - Limit service exposure
  - Avoid exposing UI externally
- Component caveats
  - eBPF requires elevated privileges → call this out clearly

👉 This becomes your reference implementation of “secure Alloy”

## Add Standalone hardening (this is the biggest gap)

Right now this is barely covered.

New section: “Harden standalone Alloy (Linux/macOS/Windows)”

- Cross-platform
[...]

PR #6386, opened 2026-06-01
docs: Add topics about hardening Alloy deployments.

clayton-cornell:
Harden Alloy topic additions

Part of: https://github.com/grafana/alloy/issues/5878


docs/sources/secure/_index.md
```diff
@@ -0,0 +1,80 @@
...
+weight: 110
+---
+
+# Secure {{% param "FULL_PRODUCT_NAME" %}}
+
+{{< param "FULL_PRODUCT_NAME" >}} collects telemetry from hosts, containers, and APIs, then forwards it to observability backends.
+That work requires read access to logs, process data, and cluster resources, plus credentials for remote write and similar endpoints.
+Your configuration determines which permissions {{< param "PRODUCT_NAME" >}} needs, and your deployment platform determines how you enforce them.
+
+## Security best practices
```
thampiotr: Discussed offline, I think there are a lot of exceptions and these won't always apply. If something applies always, we should try to make it the default.

docs/sources/access_permissions/kubernetes.md
```diff
@@ -0,0 +1,235 @@
...
+{{< param "PRODUCT_NAME" >}} requires read access to Kubernetes API resources, node and container telemetry, and credentials for observability backends.
+The {{< param "PRODUCT_NAME" >}} container image runs as `root` by default and defines a non-root `alloy` user with UID `473` and GID `473`.
+You can configure `securityContext`, RBAC, and network settings for the components in your configuration.
+
+## Run as a non-root user
+
+The [{{< param "PRODUCT_NAME" >}} container image][image] defines two users: `root` and a non-root user named `alloy` with UID `473` and GID `473`.
+The container runs the `alloy` binary as `root` by default because some components, such as [beyla.ebpf][], need elevated privileges.
+
+The Grafana Helm chart doesn't set `runAsUser` or `runAsGroup` by default, so the container also runs as `root` until you add a `securityContext`.
```
thampiotr: Which helm chart specifically are we referring to here?

docs/sources/secure/kubernetes.md
```diff
@@ -0,0 +1,235 @@
...
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: scc-alloy
+runAsUser:
+  type: MustRunAs
+  uid: 473
+fsGroup:
+  type: MustRunAs
+  uid: 1000
```
thampiotr: I'm getting LLM flag this as invalid and it should go with:
```
fsGroup:
  type: MustRunAs
  ranges:
    - min: 1000
      max: 1000
```
and confirmed this here: https://github.com/openshift/api/blob/master/security/v1/types.go#L257

thampiotr: On top of my previous comment, we should set this to 473, not 1000. This is because Alloy's GID is 473.

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,80 @@
...
+## Network exposure
+
+{{< param "PRODUCT_NAME" >}} runs an HTTP server for its UI, API, and `/metrics` endpoint.
+The binary binds to `127.0.0.1:12345` by default, which limits exposure to the local machine.
+The Grafana Helm chart sets `alloy.listenAddr` to `0.0.0.0` by default so other Pods can reach the container on port `12345`.
+Review what you expose before you change either default.
+
+## Secrets and credentials
+
+{{< param "PRODUCT_NAME" >}} loads secrets at runtime through several patterns:
```
ptodev: This sentence makes it sound like secrets are always loaded. But they're only loaded if those functionalities are used. Maybe it's better to rephrase it to something like "you can load secrets using..."

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,80 @@
...
+For `secret`-typed values at runtime and protection from exposure in the UI and component exports, refer to [Types and values][types-values].
+
+## Components that require elevated access
+
+Some components need root, extra capabilities, or group membership.
+
+| Component                  | Requirement                        | Notes                                                        |
+| -------------------------- | ---------------------------------- | ------------------------------------------------------------ |
+| `beyla.ebpf`               | Root or Linux capabilities         | Kernel-level eBPF. Incompatible with strict non-root setups. |
+| `prometheus.exporter.unix` | Read access to `/proc`, `/sys`     | Usually satisfied by the `alloy` user on Linux.              |
```
ptodev: What is "the alloy user on linux"?

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,80 @@
...
+| `loki.source.journal`      | `adm` and `systemd-journal` groups | The package installer adds the `alloy` user to both groups.  |
+| `loki.source.file`         | Read access to target log paths    | Grant per-path ACLs instead of broad permissions.            |
```
ptodev: What is ACL?

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,80 @@
...
+| `pyroscope.ebpf`           | Root or Linux capabilities         | Same constraint as `beyla.ebpf`.                             |
```
ptodev: It's not exactly the same. It's the same in the sense that it requires "root". It's not the same in the sense that it doesn't require eBPF capabilities. 

"root" is whether the user is 0. Usually that user has all capabilities, but that's not guaranteed. In theory you can even restrict the 0 user from doing things.

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,80 @@
...
+{{< param "PRODUCT_NAME" >}} loads secrets at runtime through several patterns:
+
+- `sys.env()` in configuration to reference environment variables
+- [`remote.vault`][remote-vault] to load secrets from HashiCorp Vault
+- Secrets from the cluster: [remote.kubernetes.secret][remote-k8s-secret]
+- [`remote.s3`][remote-s3] to load configuration or secrets from AWS S3
+
+For `secret`-typed values at runtime and protection from exposure in the UI and component exports, refer to [Types and values][types-values].
+
+## Components that require elevated access
```
ptodev: Should this table be here? Is it not better to have a Security section in the reference docs of the components? It's very easy for people to forget to update this page. Also, the component reference pages already have security-related info which could use a better organisation.

docs/sources/access_permissions/kubernetes.md
```diff
@@ -0,0 +1,235 @@
...
+          drop:
+            - ALL
+```
+
+- `runAsNonRoot: true` causes Kubernetes to reject the Pod if the image tries to run as root.
+- `readOnlyRootFilesystem: true` blocks writes to the container filesystem except on mounted volumes.
+- `allowPrivilegeEscalation: false` blocks privilege escalation beyond the parent process, regardless of file capabilities or `setuid` bits.
+- `capabilities.drop: [ALL]` removes all Linux capabilities from the container.
+
+The Helm chart sets `alloy.storagePath` to `/tmp/alloy` by default.
```
ptodev: Can statements like this go out of date? I'm not sure I wan to be this specific.

docs/sources/access_permissions/kubernetes.md
ptodev, line 1: Shouldn't the helm charts configure things securely out of the box? Why do users need to configure things explicitly?

docs/sources/access_permissions/kubernetes.md
clayton-cornell: Dropped this part

docs/sources/access_permissions/_index.md
clayton-cornell: Dropped the table and added a link to the component docs instead. Any access/permissions can be explicitly documented per component that requires something beyond the default

docs/sources/access_permissions/_index.md
clayton-cornell: The table has been dropped.

docs/sources/access_permissions/_index.md
clayton-cornell: ACLs is like RBAC?  It's dropped when the table was removed.,

docs/sources/access_permissions/kubernetes.md
clayton-cornell: It defaults to root... and users have asked for non-root.

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,76 @@
...
+Use only what matches your components and environment.
+
+1. When your components allow it, run {{< param "PRODUCT_NAME" >}} as a non-root user on [Linux][linux], [Kubernetes][kubernetes], or a dedicated service account on [Windows][windows].
+1. If you don't need remote access to the UI or `/metrics`, restrict the HTTP server to `127.0.0.1` or a private network address with the [`http` block][http-block].
+1. When you expose the HTTP server beyond localhost, enable TLS with the [`http` block][http-block].
+1. Use TLS for outbound remote write and OTLP connections with the [`tls` block][tls-block].
+1. Avoid `insecure_skip_verify = true` in production.
+   Refer to the [`tls` block][tls-block].
+1. Store credentials outside configuration files when you can.
+   Refer to [Types and values][types-values].
```
ptodev, line 27: It's ok to have a link to "Types and values" for now, but in the future it'd be so nice to have a page in the new Access & Permissions category solely dedicated to best practices with accessing secrets. It could show how to get them from vault, from an env var, how to use the coalesce std lib funtion, etc.

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,76 @@
...
+1. On Kubernetes, set RBAC to the permissions your configuration uses.
+   Refer to [Access and permissions on Kubernetes][kubernetes].
+1. For container deployments, set `readOnlyRootFilesystem: true` when your volume mounts and components allow it.
+   Refer to [Access and permissions on Kubernetes][kubernetes].
+1. For container deployments, set `allowPrivilegeEscalation: false` when your components don't need privilege escalation.
+   Refer to [Access and permissions on Kubernetes][kubernetes].
+1. Don't use eBPF components unless your use case requires them.
```
ptodev: This could be rephrased as something like "components which require elevated access should be ran in an Alloy instance dedicated to them". BPF is widely regarded as safe, and we don't want to discourage people from using it. It's just that they need a bit more guardrails.

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,76 @@
...
+   Refer to [Components that require elevated access](#components-that-require-elevated-access).
+
+## Process identity and privilege
+
+Create a dedicated service account or user on your deployment platform:
+
+- [Access and permissions on Linux][linux]: systemd service, file permissions, and the `alloy` user
+- [Access and permissions on Kubernetes][kubernetes]: `securityContext`, non-root UID, and RBAC
+- [Access and permissions on Windows][windows]: service accounts, Windows security groups, and filesystem ACLs
```
ptodev:
```suggestion
- [Linux][linux]: systemd service, file permissions, and the `alloy` user
- [Kubernetes][kubernetes]: `securityContext`, non-root UID, and RBAC
- [Windows][windows]: service accounts, Windows security groups, and filesystem ACLs
```

No need to repeat "Access and permissions"?

docs/sources/access_permissions/_index.md
```diff
@@ -0,0 +1,76 @@
...
+{{< param "PRODUCT_NAME" >}} runs an HTTP server for its UI, API, and `/metrics` endpoint.
+The binary binds to `127.0.0.1:12345` by default, which limits exposure to the local machine.
+The Grafana Helm chart sets `alloy.listenAddr` to `0.0.0.0` by default so other Pods can reach the container on port `12345`.
+Review what you expose before you change either default.
+
+## Secrets and credentials
+
+You can load secrets at runtime through several patterns:
+
+- `sys.env()` in configuration to reference environment variables
```
ptodev: There could be a link to the [reference doc](https://deploy-preview-alloy-6386-zb444pucvq-vp.a.run.app/docs/alloy/latest/reference/stdlib/sys/#sysenv). But it might be even better if there's a dedicated page for secret management, with examples?

docs/sources/access_permissions/linux.md
```diff
@@ -0,0 +1,211 @@
...
+Don't expand the `alloy` user's group membership to reach those files.
+
+```shell
+sudo setfacl -R -m u:alloy:rx /var/log/myapp
+sudo setfacl -R -d -m u:alloy:rx /var/log/myapp
+```
+
+The `-d` flag sets a default ACL so new files in the directory inherit the permission.
+
+## Restrict the HTTP server
```
ptodev, line 177: This was already mentioned in the index page?

docs/sources/access_permissions/linux.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+Reload and restart the service:
+
+```shell
+sudo systemctl daemon-reload
+sudo systemctl restart alloy
+```
+
+## Restrict file and directory permissions
```
ptodev, line 61: Maybe the installer should restrict this by default 🤔 Anyway it's good to document this in case someone doesn't use the installer.

docs/sources/access_permissions/linux.md
```diff
@@ -0,0 +1,211 @@
...
+**`prometheus.exporter.unix`** reads from `/proc` and `/sys`.
+The `alloy` user can read most of these paths on a typical Linux system without elevated privileges.
+If you see permission errors, check the metric collector that causes the issue.
+Don't switch the service to root.
+
+## Next steps
+
+- [Access and permissions for {{< param "PRODUCT_NAME" >}}][access]
+- [Access and permissions on Kubernetes][kubernetes]
+- [Access and permissions on Windows][windows]
```
ptodev: Not sure why those are next steps? If you're on Linux you don't need to read about Windows. And the index page is theoretically the previous step.

docs/sources/access_permissions/kubernetes.md
```diff
@@ -0,0 +1,139 @@
...
+
+{{< admonition type="note" >}}
+If you use components that need elevated host access, such as `beyla.ebpf`, add back the capabilities those components need.
+Don't drop all capabilities when these components are in your configuration.
+Refer to the [beyla.ebpf component reference][beyla-ebpf-cap].
+
+[beyla-ebpf-cap]: ../../reference/components/beyla/beyla.ebpf/
+{{< /admonition >}}
+
+## Restrict the HTTP server
```
ptodev, line 95: Not sure if we need to mention this again.

docs/sources/access_permissions/kubernetes.md
```diff
@@ -0,0 +1,139 @@
...
+Set `alloy.listenAddr` to `127.0.0.1` in `values.yaml` or restrict access with a NetworkPolicy when you don't need cross-Pod access to the UI or `/metrics` endpoint.
+The container image uses the binary default of `127.0.0.1:12345` when you don't pass `--server.http.listen-addr`.
+Refer to the [`http` block][http-block] for TLS and authentication options.
+
+## Kubernetes RBAC
+
+{{< param "PRODUCT_NAME" >}} needs RBAC permissions to interact with Kubernetes APIs.
+The Helm chart creates a `ClusterRole` and `ClusterRoleBinding` with a fixed rule set in `values.yaml` when `rbac.create` is `true`.
+
+Remove permissions for resources your configuration doesn't use.
```
ptodev: Is it easy to remove permissions which the helm chart installs? Not sure how doable it would be in practice.

docs/sources/access_permissions/_index.md
clayton-cornell: Created https://github.com/grafana/alloy/issues/6435 to track this

docs/sources/access_permissions/_index.md
clayton-cornell: Created #6435 to track this

docs/sources/access_permissions/kubernetes.md
clayton-cornell: Easiest to simply remove it. It's not critical info

docs/sources/access_permissions/kubernetes.md
clayton-cornell: It's pretty generic in the index file.  Here it dives a bit deeper about what you can do in K8s.

docs/sources/access_permissions/kubernetes.md
clayton-cornell: Updated with new info here. I dig a bit deeper into where the permissions are set and it seems to be set in the rbac.rule and rbac.clusterrules.

docs/sources/access_permissions/linux.md
clayton-cornell: Yes, but we can't assume people will always read the parent. For info like this we apply the principle of "Every page is page one"

docs/sources/access_permissions/linux.md
clayton-cornell: Good point.  Cleaning it up on all new topics.

clayton-cornell:
OK, I think this is ready for final review now. I've updated, and simplified things based on your feedback/comments.

docs/sources/introduction/requirements.md
```diff
@@ -191,7 +191,7 @@ If something doesn't behave as expected after deployment:
 [http]: ../../reference/http/
 [reference]: ../../reference/
 [run]: ../../reference/cli/run/
-[nonroot]: ../../configure/nonroot/
+[access-kubernetes]: ../../access_permissions/kubernetes/
```
clayton-cornell:
```suggestion
[access-kubernetes]: ../../../access_permissions/kubernetes/
```

docs/sources/reference/components/prometheus/prometheus.exporter.cadvisor.md
```diff
@@ -78,8 +78,9 @@ For Docker, grant permissions using one of these approaches:
   Only use this approach if necessary for your environment.
   {{< /admonition >}}
 
-For more information about running {{< param "PRODUCT_NAME" >}} without root privileges, refer to [Configure {{< param "PRODUCT_NAME" >}} to run as a nonroot user][nonroot].
+For more information about running {{< param "PRODUCT_NAME" >}} without root privileges, refer to [Access and permissions for {{< param "PRODUCT_NAME" >}} on Kubernetes][access-kubernetes].
 
+[access-kubernetes]: ../../../access_permissions/kubernetes/
```
clayton-cornell:
```suggestion
[access-kubernetes]: ../../../../access_permissions/kubernetes/
```

docs/sources/access_permissions/kubernetes.md
```diff
@@ -0,0 +1,135 @@
...
+{{< param "PRODUCT_NAME" >}} requires read access to Kubernetes API resources, node and container telemetry, and credentials for observability backends.
+The {{< param "PRODUCT_NAME" >}} container image runs as `root` by default and defines a non-root `alloy` user with UID `473` and GID `473`.
+Set `securityContext`, RBAC, and network settings to match the components in your configuration.
+
+The [{{< param "PRODUCT_NAME" >}} Docker container image][image] defines two users:
+
+- A `root` user.
+- A non-root user named `alloy` with UID `473` and GID `473`.
+
+You can configure a non-root user when you deploy Alloy in Kubernetes.
```
dehaansa: This says just above that a non-root user is defined, but then that you can configure one, I'm a bit confused by how those two statements work together.

dehaansa (dismissed):
One nit

docs/sources/access_permissions/kubernetes.md
clayton-cornell: Hmm yeah I could have defined that a bit better. The section below that at line 35 explains it better... I think just dropping that sentence is the cleanest solution to the confusion.

dehaansa (approved):

PR #6388, opened 2026-06-01
chore: Use Taskfile for lint

x1unix:

Add taskfile with lint tasks.

### Pull Request Details

This PR introduces Taskfile as a future Makefile replacement and in this, first PR, adds a lint task that is also used in CI.

Based on #6294

### Notes to the Reviewer

Docs will be updated in a separate PR when more tasks will be added.


Taskfile.yml
```diff
@@ -0,0 +1,18 @@
+# yaml-language-server: $schema=https://taskfile.dev/schema.json
+version: "3"
+
+includes:
+  lint: tools/task/lint.yml
```
kalleep: Shoult we have all tasks under `./task` directly? Currently tools is just a cli and nothing else, all scripts and build stuff have been moved out

.github/workflows/lint.yml
```diff
@@ -73,21 +70,21 @@ jobs:
         working-directory: internal/web/ui
 
     steps:
-    - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      with:
-        persist-credentials: false
+      - name: Checkout code
```
ptodev: Why is the whitespace changing?

build-tools/task/lint.yml
ptodev, line 1: In the makefile, could you remove the lint-related targets and replace them with a call to the task file? That way in the PR review we can make sure the new task does the same as the current make command.

.github/workflows/lint.yml
kalleep: I guess we should have a formatter for our yaml files

x1unix: @ptodev Its my yaml formatter, I apologize

We can adopt prettier to have a consistent style & formatting for yaml

Taskfile.yml
x1unix: @kalleep as we're going to have more tasks, I planned to keep them in separate files like I did in the original PR #6294.

If this is a concern, I can put lint tasks in the main file and decouple them later.

kalleep: Yeah and I think it's good to a separate files, what I meant is that in `./tools` directory we only have code for our tools-cli. And I think all of these files can live under a ./tasks directory (not nested under tools)

x1unix: @kalleep what do you think about `./build-tools/task`? Similar to where makefiles currently are.

But, `/tasks` is also fine by me

kalleep: either one is fine with me

build-tools/task/lint.yml
x1unix: @ptodev I've updated our lint workflow to use the new lint task instead but can update make as well

Taskfile.yml
x1unix: @kalleep done in https://github.com/grafana/alloy/pull/6388/changes/f76cd668b9806edfee05c4926ec9d1f3aace6f16

.github/workflows/lint.yml
x1unix: @ptodev fixed formatting: https://github.com/grafana/alloy/pull/6388/changes/9c1e90885941a7602fb1e8c5e34fe3ba9e754c96

imo we should consider using a standardized formatter like Prettier.

build-tools/task/lint.yml
x1unix: @ptodev see https://github.com/grafana/alloy/pull/6388/changes/0960d3df8187417d84b948602a3ff2f43ae4f35d

kalleep (approved):

PR #6390, opened 2026-06-02
chore: Migrate test workflows mise setup

kalleep:
Migrate test_* workflows to install go thorugh mise.

For the mac workflow we did have cache enabled. Mise will cache tools installed but not go modules or go build output so I created a resuable action for it.

### DIFF
--- a/.github/workflows/test_full.yml
+++ b/.github/workflows/test_full.yml
@@ -23,15 +23,16 @@ jobs:
       uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       with:
         persist-credentials: false
+
     - name: Set ownership
       # https://github.com/actions/runner/issues/2033#issuecomment-1204205989
       run: |
           # this is to fix GIT not liking owner of the checkout dir
           chown -R $(id -u):$(id -g) $PWD
-    - name: Set up Go
-      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+
+    - name: Setup tools
+      uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
       with:
-        go-version-file: go.mod
-        # TODO: Enable caching later.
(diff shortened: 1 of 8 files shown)


.github/workflows/test_full.yml
```diff
@@ -23,15 +23,16 @@ jobs:
...
     - name: Set ownership
       # https://github.com/actions/runner/issues/2033#issuecomment-1204205989
       run: |
           # this is to fix GIT not liking owner of the checkout dir
           chown -R $(id -u):$(id -g) $PWD
-    - name: Set up Go
-      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+
+    - name: Setup tools
+      uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
```
zeitlinger, line 34: you can also pin the mise version if you want to https://github.com/grafana/docker-otel-lgtm/blob/main/.github/workflows/acceptance-tests.yml#L20

needs an extra renovate rule (which is included in https://github.com/grafana/flint/ if you want to use that too)

kalleep: Thanks for the suggestion! Added https://github.com/grafana/alloy/issues/6405. We already use mise for some workflows so will followup after this pr and pin it for all of them

kalleep: Nvm I pinned the mise version but update the issue to add a renovate rule. In alloy we use mac, linux and windows runners so we need to handle that in renovate somehow

zeitlinger: see https://github.com/grafana/flint/blob/main/.github/workflows/test.yml#L19-L27 on how to do that using https://github.com/grafana/flint/blob/main/.github/renovate.json5#L35-L44

zeitlinger: The rule is here: https://github.com/grafana/flint/blob/main/.github/renovate.json5#L20-L27

kgeckhart (approved):

PR #6393, opened 2026-06-02
feat(database_observability): Set timestamp query_sample at query start for Postgres

fridgepoet:

Similar to https://github.com/grafana/alloy/pull/6291, this PR changes the query_sample collectors to log timestamps at query start time.

In the context of application traces, traces start their span at query start time. The DB O11y app's trace linkage navigates to the trace's start time on the timeline, so the previous end timestamp query_sample point in the UI can appear too late.

This fix changes the timestamp of the log to the start time for the Postgres query_sample collector.

Closes https://github.com/grafana/grafana-dbo11y-app/issues/2818

### Notes to the Reviewer

To test this manually, application trace timestamps for queries were compared and found similar to op=query_sample log timestamps.

* Built and ran these changes locally, pointing to a DB in the cloud, collecting telemetry to a personal Grafana Cloud stack
* An unchanged Alloy was also collecting from the same DB in the cloud, sending to another Grafana Cloud stack
* I found log lines for the same query on both stacks (same trace-id-span-id comment) 
* Comparing the log timestamps, I can see this PR changes it to be at the start time. (query execution was 10s, so the difference between both log timestamps was indeed confirmed as 10s)

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/database_observability/postgres/collector/query_samples_test.go
```diff
@@ -100,7 +103,7 @@ func TestQuerySamples_FetchQuerySamples(t *testing.T) {
 						"testuser", "testapp", "127.0.0.1", 5432,
 						"client backend", backendStartTime, sql.NullInt32{}, sql.NullInt32{},
 						xactStartTime, "waiting", stateChangeTime, sql.NullString{String: "Lock", Valid: true},
-						sql.NullString{String: "relation", Valid: true}, pq.Int64Array{103, 104}, now, sql.NullInt64{Int64: 124, Valid: true},
+						sql.NullString{String: "relation", Valid: true}, pq.Int64Array{103, 104}, queryStartTime, sql.NullInt64{Int64: 124, Valid: true},
```
fridgepoet, line 106: mock is now modified to return a different time for the query_start time so that the assertions will have distinct timestamps for op=query_sample vs op=wait_event

internal/component/database_observability/postgres/collector/query_samples_test.go
```diff
@@ -1340,7 +1345,7 @@ func TestQuerySamples_WaitEvents_PreClassifiedFlag(t *testing.T) {
 				"testuser", "testapp", "127.0.0.1", 5432,
 				"client backend", backendStartTime, sql.NullInt32{}, sql.NullInt32{},
 				xactStartTime, "waiting", now.Add(-5*time.Second), sql.NullString{String: "IO", Valid: true},
-				sql.NullString{String: "DataFileRead", Valid: true}, pq.Int64Array{}, now, sql.NullInt64{Int64: 5002, Valid: true},
+				sql.NullString{String: "DataFileRead", Valid: true}, pq.Int64Array{}, now.Add(-30*time.Second), sql.NullInt64{Int64: 5002, Valid: true},
```
fridgepoet, line 1348: similar to above, mock is now modified to return a different time for the query_start time

cristiangreco (approved):

matthewnolf (approved):

PR #6395, opened 2026-06-02
feat: Make alloyengine extension OCB-able

jharvey10:

- Removes the separate Go module for the `alloyengine` extension.
- Removes dependency-replacements.yaml in favor of using the OCB manifest (`collector/builder-config.yaml`) as the source of truth for shared external `replace` directives.
- Runs a post-process on release-please PRs to rebuild the collector and commit any changes to the release PR.

### Pull Request Details

This PR moves the `extension/alloyengine` to be part of the root Alloy module since their contents and dependencies are almost identical. It also updates the generated collector distro to reference `github.com/grafana/alloy` directly with an explicit `import` and `name` in the OCB builder config. It is the implementation of the chosen proposal of https://github.com/grafana/alloy/pull/6174.

It removes the old `dependency-replacements.yaml` flow and the `generate module-dependencies` tooling, replacing it with `tools sync-replaces`. The new command reads shared replace directives from the marked block in `collector/builder-config.yaml`, syncs them into the root `go.mod`, preserves root-local path replaces, and runs `go mod tidy`.

The collector generation target now runs `sync-replaces` as the primary portion of `make generate-otel-collector-distro`. Release automation is also updated to regenerate the collector distro on release-please branches when needed.

Docs and agent instructions have also been updated.

### Notes to the Reviewer

Shared remote replaces now belong in `collector/builder-config.yaml` between the `<BEGIN_SHARED_REPLACE_DIRECTIVES>` and `<END_SHARED_REPLACE_DIRECTIVES>` markers. Local path replaces should stay outside that shared block.

Focused verification run:

`go test -C tools ./sync-replaces/...`

### PR Checklist

- [x] Documentation added
- [x] Tests updated


extension/alloyengine/README.md
```diff
@@ -47,7 +47,53 @@ service:
...
+
+## Build with OCB
+
+To include the extension in an external OpenTelemetry Collector Builder (OCB) distribution, add it as an extension using the Alloy module plus the extension import path. The extension doesn't have its own Go module, so the one-line `gomod: github.com/grafana/alloy/extension/alloyengine ...` form doesn't work.
+
+```yaml
+dist:
+  name: custom-alloy-otel
+  output_path: ./dist
+  cgo_enabled: true
```
jharvey10, line 67: @kgeckhart From the testing that you were doing before, is this something you actually needed to do?

tools/cmd/main.go
```diff
@@ -6,14 +6,21 @@ import (
...
 )
 
 func main() {
+	cmd := newRootCommand()
+	if err := cmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func newRootCommand() *cobra.Command {
```
jharvey10, line 23: This made things easier to test locally

.github/workflows/release-please.yml
```diff
@@ -28,6 +28,13 @@ jobs:
           token: ${{ steps.app-token.outputs.token }}
           persist-credentials: false
 
+      - name: Configure Git user
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          APP_SLUG: ${{ env.ALLOYBOT_APP_NAME }}
+        run: |
+          go run -C tools ./cmd release configure-git-user
```
thampiotr, line 43: Before running go, shouldn't we run the 'Set up Go' step?

.github/workflows/release-please.yml
```diff
@@ -47,3 +54,48 @@ jobs:
...
+        if: steps.root-release-branch.outputs.exists == 'true'
+        run: |
+          make generate-otel-collector-distro
+
+          if [ -z "$(git status --porcelain)" ]; then
+            echo "Collector generation produced no diff; nothing to amend."
+            exit 0
+          fi
+
+          git add .
```
thampiotr: Could we limit this to more specifically what we want to add and commit? it's not uncommon mistake for some side-effects to appear in repo and in this case they would get committed. And it could be safety feature too.

extension/alloyengine/README.md
```diff
@@ -47,7 +47,53 @@ service:
 In this example, the extension will:
 1. Start the default engine with the configuration file at the relative path `./config.alloy`
 2. Pass the `--server.http.listen-addr=0.0.0.0:12345` and `--stability.level=experimental` flags to the `alloy run` command
-4. Run the Alloy configuration concurrently with the OpenTelemetry Collector pipeline
+3. Run the Alloy configuration concurrently with the OpenTelemetry Collector pipeline
+
+## Build with OCB
```
thampiotr, line 59: Could this go to our docs website? I think that would be cool. The PR title is `feat:` so I think we did want to make it user-facing?

.github/workflows/release-please.yml
```diff
@@ -47,3 +54,48 @@ jobs:
...
+          persist-credentials: true # Needed for subsequent git push.
+
+      - name: Set up Go 🏗️
+        if: steps.root-release-branch.outputs.exists == 'true'
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: Regenerate collector distro and stage changes
```
thampiotr, line 83: I wonder if there is a safer, more targetted way of doing this. Maybe configuring the files that need to be updated in release-please-config.json? 

It's probably okay as-is, but we do run code in the repo via `make` below and then push whatever it produced without human review.

thampiotr (commented):
Nice! that's cleaning things up a lot! Just a few questions.

.github/workflows/release-please.yml
jharvey10: good call! I have a separate TODO to switch our setup-go usages to mise in CI, but for this PR, I'll add what's needed for this particular workflow.

.github/workflows/release-please.yml
jharvey10: I'm definitely not opposed to that, and that's similar to what it was before. During iteration on this PR it was a lot simpler to have this be a catch-all and rely on the porcelain checks elsewhere as the gatekeepers.

extension/alloyengine/README.md
jharvey10: Yeah, it definitely can. If we get this PR out there and well-validated, I bet we'll have a good recipe for this to put on our public docs.

.github/workflows/release-please.yml
jharvey10: I did a ton of research on this exact point that's not really captured in the PR description or the final implementation, and there's a few layers of safety here.

I would love to use the release-please-config as the sole mechanism for this, but based on earlier research that @kgeckhart and I did, it's really not feasible to do that given that we want to commit generated files to the repo. If we *didn't* commit generated files, then release-please would be able to handle this 100% on its own, but that's not the Go convention.

The branch from which the code is run here is always a release-please branch, and those are only ever created by release please, and as a baseline require write/push access to those branches (i.e. maintainers or admins). It's always `main` + 1 commit, which is the result of the release-please run.

If we adjust the follow on commit to just specific files that we want, it'll lock this down even more.

Any *next* time that release please runs, it will get rid of any existing content and use the latest workflow/script versions from main, so it's very easy to "redo" content if we see that it's incorrect/inappropriate.

Lastly, this still requires a human to both take the resulting PR out of draft and do merge it for it to affect our mainline code.

.github/workflows/release-please.yml
jharvey10: done!

clayton-cornell:
Public doc change is minimal. No comments from my side.

extension/alloyengine/README.md
kgeckhart: gco_enabled? Yep it's required for alloy to build. I added https://github.com/grafana/alloy/issues/6467 to GA to see if we could drop it via stubs.

tools/sync-replaces/main_test.go
kgeckhart, line 1: Any reason these tests skipped testify `assert`/`require`?

tools/sync-replaces/main_test.go
```diff
@@ -0,0 +1,175 @@
...
+				t.Fatal("expected error")
+			}
+			if !strings.Contains(err.Error(), tt.wantErr) {
+				t.Fatalf("expected error containing %q, got %v", tt.wantErr, err)
+			}
+		})
+	}
+}
+
+func TestSyncBuilderConfigReplacesToGoMod(t *testing.T) {
```
kgeckhart: This looks like a flow test since we run it twice, could you rename the test around the flow or add comments about it?

tools/sync-replaces/builder_config_scanner.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil
+		}
+	}
+	if err := s.scanner.Err(); err != nil {
+		return fmt.Errorf("scan builder config: %w", err)
+	}
+	return fmt.Errorf("missing shared replace start marker")
+}
+
+func (s *builderConfigScanner) readSharedReplaceEntry() (*replaceEntry, error) {
```
kgeckhart, line 34: I can't tell for sure but it looks like this validates the replace a little but will this or something else catch a totally invalid replace like fat finger of `=>`

tools/sync-replaces/main.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	return appendSharedReplaces(formatted, replaces), nil
+}
+
+func appendSharedReplaces(data []byte, replaces []replaceEntry) []byte {
+	if len(replaces) == 0 {
+		return data
+	}
+
+	var builder strings.Builder
```
kgeckhart, line 173: Should we put a banner here to start this that points people away form editing here? I think they might see all the individual comments and think it's still fine to add a direct "REPLACES NEED TO GO TO builder-config.yml") might make it more obvious?

kgeckhart (commented):
Nice work! Some thoughts

tools/sync-replaces/main_test.go
jharvey10: We don't currently have testify in the tools module, but I can add it and adjust the tests accordingly.

tools/sync-replaces/main_test.go
jharvey10: decided on `TestSyncBuilderConfigReplacesToGoModIdempotence`

tools/sync-replaces/main_test.go
jharvey10: added!

tools/sync-replaces/builder_config_scanner.go
jharvey10: Since OCB itself will eventually fail with malformed entries, I was thinking of just leaving that as the failure mode, but I tried out a few things and noticed that that's probably too late. A go.mod file has already been written in such a way where if you correct the builder config and re-run, it'll never self-correct. Catching it earlier is the way to go, so I adjusted the logic to add validation.

tools/sync-replaces/main.go
jharvey10: done

kgeckhart (approved):
🔥

PR #796, opened 2024-05-08
Enable linux boringcrypto builds

mattdurham:
#### PR Description

Enable linux boringcrypto builds for executables.

PR #6407, opened 2026-06-03
chore: Remove unused Microsoft Go boringcrypto build image

kgeckhart:

Removes the `-boringcrypto` build-image variant from both build-image workflows, repoints the boringcrypto compile-check at the plain image, and drops the `goversion` automation that maintained its pin. 

The remove variant builds FROM the Microsoft Go fork, which doesn't support `GOEXPERIMENT=boringcrypto`, and nothing uses it for that anyway: release images build boringcrypto from the plain build-image (standard Go), and the PR check gets standard Go from `actions/setup-go` regardless. It's a leftover from the Windows cngcrypto path removed in #796.


.github/workflows/build.yml
```diff
@@ -55,11 +55,6 @@ jobs:
       run: |
           # this is to fix GIT not liking owner of the checkout dir
           chown -R $(id -u):$(id -g) $PWD
-    - name: Set up Go
-      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
-      with:
-        go-version-file: go.mod
-        cache: false
```
kgeckhart, line 57: Same as, https://github.com/grafana/alloy/pull/6408, there's no reason to re-install go when using a build image that's supposed to provide it for us

x1unix (dismissed):

x1unix (approved):

PR #6408, opened 2026-06-03
chore: Drop redundant setup-go from build-image container jobs

kgeckhart:

Drops the redundant `actions/setup-go` step from the CI jobs that run inside the `alloy-build-image` container. The image already ships Go matching `go.mod` (1.26.3), so those jobs sourced Go twice — from the image and from `setup-go` — which can silently drift; the released container binary already builds from the image's Go (the root `Dockerfile` has no `setup-go`), so this makes every container job use that single source. The Mac/Windows jobs keep `setup-go` since they run on bare runners with no image. 

https://github.com/grafana/alloy/pull/6407 handles the boring crypto variant since there's more to be done there

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,11 +30,6 @@ jobs:
       run: |
           # this is to fix GIT not liking owner of the checkout dir
           chown -R $(id -u):$(id -g) $PWD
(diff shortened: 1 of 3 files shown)


kalleep (approved):
Ah right, good catch

PR #6410, opened 2026-06-04
feat: Add otelcol.connector.signaltometrics component

### LINKED ISSUE #6409: Add otelcol.connector.signaltometrics component
### Component(s)

_No response_

### Request

Add the upstream signaltometricsconnector https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/signaltometricsconnector

### Use case

This has come up repeatedly across internal discussions over roughly the past year, from several different teams, and there's currently no clean way to achieve it in Alloy. The demand clusters into four concrete use cases:

1. **Logs → metrics (most requested).** Extract a numeric value from log-line content and record it as a histogram or sum — e.g. a duration field into a native histogram — so the source log can be dropped after extraction and only the metric is retained and paid for.
2. **Span/trace attributes → metrics.** Produce a metric whose value comes from a span attribute, not just a count of spans — e.g. turning a latency/duration attribute into a histogram for SLOs and alerting.
3. **Cardinality / cost reduction.** Collapse high-cardinality per-instance series (e.g. many per-pod gauges) into a handful of histogram buckets, significantly cutting active series / DPM.
4. **Sampling-adjusted span metrics.** Use the connector's `adjustedCount` OTTL function (OTEP-235) to weight span-derived metrics by the probabilistic sampling rate — accurate metrics under head sampling, avoiding the cost of tail sampling.

**Why now / no alternative in Alloy**

The existing `otelcol.connector.count` only counts incoming spans/logs/datapoints by filter — it can't read an attribute's *value* or produce histograms/sums from values. Today the only fallbacks are `loki.process` `stage.metrics` (logs-only, scrape-based, classic fixed-bucket histograms, not OTel-native) or running a separate upstream `otelcol-contrib` distribution alongside Alloy. Neither is a clean, native Alloy path.

`signaltometrics` is effectively a superset of the `count` connector and covers all four cases in a single component. The demand spans users on both Prometheus/Loki-style (default-engine) pipelines and OTel-native pipelines, so a single first-class component would serve a broad set of existing requests.

### Tip

bradleypettit:

Add the `otelcol.connector.signaltometrics` component, wrapping the upstream OpenTelemetry Collector Contrib [`signaltometrics`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/signaltometricsconnector) connector. It generates metrics — sums, gauges, explicit bucket histograms, and exponential (native) histograms — from spans, metric data points, and log records using OTTL expressions.

### Pull Request Details

- New component at `internal/component/otelcol/connector/signaltometrics/`, registered at **experimental** stability with `spans`, `datapoints`, and `logs` blocks, each producing one of `histogram` / `exponential_histogram` / `sum` / `gauge`, plus `error_mode`.
- Added to the embedded OTel Collector distribution (`collector/builder-config.yaml` + regenerated `collector/`).
- Reference documentation, included-components list, and regenerated compatibility index.
- Config converter (`converter_signaltometricsconnector.go`) with testdata.
- Pinned to upstream contrib `v0.147.0` (matching Alloy's other otelcol deps).

### Notes to the Reviewer

The upstream connector also supports a **profiles** signal, but it's intentionally omitted here: Alloy's shared `otelcol.Consumer`/connector framework only wires traces/metrics/logs (no `xconnector`), so no Alloy otelcol component exposes profiles. Adding it would be a separate, framework-wide change.

Closes #6409

### PR Checklist

- [x] Documentation added
- [x] Tests updated


blewis12:
Adding here to reflect the offline discussion, the related feature request will have some more info/motivation added to it and any concerns can be discussed there - assigning myself for review here once that's done!

internal/component/otelcol/connector/signaltometrics/types.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (h *Histogram) Convert() configoptional.Optional[config.Histogram] {
+	if h == nil {
+		return configoptional.None[config.Histogram]()
+	}
+
+	buckets := h.Buckets
+	if len(buckets) == 0 {
+		buckets = defaultHistogramBuckets
```
blewis12: Instead of applying the defaults in Convert(), I think we can implement the syntax.Defaulter interface on the block types:

```
var _ syntax.Defaulter = (*Histogram)(nil)

func (h *Histogram) SetToDefault() {
	h.Buckets = append([]float64{}, defaultHistogramBuckets...)
}

...
var _ syntax.Defaulter = (*ExponentialHistogram)(nil)

func (eh *ExponentialHistogram) SetToDefault() {
	eh.MaxSize = defaultExponentialHistogramMaxSize
}
```
This way, if buckets or max_size isn't set in the config, it's populated with the default at parse time, so Convert() can just pass the value through rather than defaulting it. We follow this pattern elsewhere, e.g. [servicegraph](https://github.com/grafana/alloy/blob/a987e33af93f6756c32d776c37a6af4484b66183/internal/component/otelcol/connector/servicegraph/servicegraph.go#L96).

blewis12: That implementation also means that the the generated alloy config from the convert will have default values removed (since its implied to be the default under the hood). So instead of 

```
	spans {
		histogram {
            buckets = [2, 4, 6, 8, 10, 50, 100, 200, 400, 800, 1000, 1400, 2000, 5000, 10000, 15000]
			value = "Milliseconds(end_time - start_time)"
		}
	}
```

it would just be 

```
	spans {
		histogram {
			value = "Milliseconds(end_time - start_time)"
		}
	}
```

internal/component/otelcol/connector/signaltometrics/signaltometrics_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/grafana/alloy/internal/component/otelcol/connector/signaltometrics"
+	"github.com/grafana/alloy/syntax"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/connector/signaltometricsconnector/config"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/config/configoptional"
+)
+
+func TestArguments_UnmarshalAlloy(t *testing.T) {
+	tests := []struct {
```
blewis12, line 15: nit: can we add a test here which covers the user setting a non-default `buckets` and `max_size`? Just a sanity check to make sure they are overridden when needed

blewis12:
To fix that `govulncheck` error I think you'll need to pull changes from the main repo into your fork which will update the go version 🔒

bradleypettit:
Thanks for the suggestions @blewis12 - updated those and the examples too (was concerned about DPM so added some info around that too).

internal/component/otelcol/connector/signaltometrics/types.go
```diff
@@ -105,18 +106,22 @@ type Histogram struct {
 	Value string `alloy:"value,attr"`
 }
 
+var _ syntax.Defaulter = (*Histogram)(nil)
+
+// SetToDefault implements syntax.Defaulter, populating the default buckets when
+// a histogram block is defined without any. This means Convert can pass the
+// value straight through.
+func (h *Histogram) SetToDefault() {
```
blewis12, line 112: nit: for the `SetToDefault` description we tend to just add [this comment](https://github.com/search?q=repo%3Agrafana%2Falloy%20%22%2F%2F%20SetToDefault%20implements%20syntax.Defaulter%22&type=code) - devs can look into the `Defaulter` to get a more in depth understanding and we can avoid a longer redundant comment here

blewis12:
thanks @bradleypettit - added one more nit in there but other than that LGTM. We'll also want a review from @clayton-cornell before merging since this adds to the docs

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+    - oss
+title: otelcol.connector.signaltometrics
+---
+
+# `otelcol.connector.signaltometrics`
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`otelcol.connector.signaltometrics` accepts spans, metrics, and logs from other `otelcol` components and generates metrics from them using [OTTL][] expressions.
+You can use it to produce sums, gauges, explicit bucket histograms, and exponential (native) histograms from any signal.
```
clayton-cornell:
```suggestion
`otelcol.connector.signaltometrics` accepts spans, metric data points, and logs from other `otelcol` components and generates metrics from them using [OpenTelemetry Transformation Language][OTTL] expressions.
You can use it to produce sums, gauges, explicit bucket histograms, and exponential native histograms from any signal.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+}
+```
+
+## Arguments
+
+You can use the following arguments with `otelcol.connector.signaltometrics`:
+
+| Name         | Type     | Description                                                   | Default       | Required |
+| ------------ | -------- | ------------------------------------------------------------- | ------------- | -------- |
+| `error_mode` | `string` | How to react to errors when processing OTTL expressions.      | `"propagate"` | no       |
```
clayton-cornell:
```suggestion
| Name         | Type     | Description                                                            | Default       | Required |
| ------------ | -------- | ---------------------------------------------------------------------- | ------------- | -------- |
| `error_mode` | `string` | How the connector reacts to errors when it evaluates OTTL expressions. | `"propagate"` | no       |
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+
+The `error_mode` argument determines how the connector reacts to errors that occur while evaluating an OTTL condition or expression during runtime data consumption.
+This setting doesn't affect errors during OTTL parsing at configuration time, which always cause a startup failure.
```
clayton-cornell:
```suggestion
The `error_mode` argument determines how the connector reacts to errors that occur while it evaluates an OTTL condition or expression during runtime data consumption.
This setting doesn't affect errors during OTTL parsing at configuration time, which always causes a startup failure.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+You can set `error_mode` to one of the following values:
+
+* `propagate`: Return the error up the pipeline. This results in the payload being dropped from {{< param "PRODUCT_NAME" >}}.
+* `ignore`: Ignore the error and continue processing. If an error occurs, the record is skipped and the error is logged.
+* `silent`: Ignore the error and continue processing. If an error occurs, the record is skipped and the error isn't logged.
```
clayton-cornell:
```suggestion
* `propagate`: Return the error up the pipeline. {{< param "PRODUCT_NAME" >}} drops the payload from the pipeline.
* `ignore`: Ignore the error and continue to process data. If an error occurs, {{< param "PRODUCT_NAME" >} skips the record and logs the error.
* `silent`: Ignore the error and continue to process data. If an error occurs, {{< param "PRODUCT_NAME" >} skips the record and doesn't log the error.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+The `spans` block configures a metric generated from spans.
+
+You can specify multiple `spans` blocks to define different metrics.
+
+| Name          | Type           | Description                                            | Default | Required |
+| ------------- | -------------- | ------------------------------------------------------ | ------- | -------- |
+| `name`        | `string`       | Name of the generated metric.                          |         | yes      |
+| `conditions`  | `list(string)` | OTTL conditions, ORed together, that select the data.  | `[]`    | no       |
+| `description` | `string`       | Description of the generated metric.                   | `""`    | no       |
+| `unit`        | `string`       | Unit associated with the generated metric.             | `""`    | no       |
```
clayton-cornell:
```suggestion
| Name          | Type           | Description                                 | Default | Required |
| ------------- | -------------- | ------------------------------------------- | ------- | -------- |
| `name`        | `string`       | Name of the generated metric.               |         | yes      |
| `conditions`  | `list(string)` | OTTL conditions that filter incoming spans. | `[]`    | no       |
| `description` | `string`       | Description of the generated metric.        | `""`    | no       |
| `unit`        | `string`       | Unit associated with the generated metric.  | `""`    | no       |
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+
+Each `spans` block must contain exactly one of the [`histogram`][histogram], [`exponential_histogram`][exponential_histogram], [`sum`][sum], or [`gauge`][gauge] blocks, which defines the type of metric to generate.
+
+The `conditions` argument accepts a list of [OTTL][] conditions that are ORed together.
+The connector only processes data into the metric if at least one condition evaluates to `true`.
+If you don't specify any conditions, the connector processes all data.
```
clayton-cornell:
```suggestion
The `conditions` argument accepts a list of [OTTL][] conditions that the connector combines with a logical OR.
The connector processes data into the metric only when at least one condition evaluates to `true`.
If you don't specify `conditions`, the connector processes all data.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+Refer to the [`spans`][spans] block documentation for the complete list of supported arguments and blocks.
+
+### `attributes`
+
+The `attributes` block specifies a data point attribute to group the generated metric by.
+
+Each unique combination of attribute values generates a separate data point on the metric.
+You can specify multiple `attributes` blocks within a `spans`, `datapoints`, or `logs` block to group by multiple attributes.
+
+The following arguments are supported:
```
clayton-cornell:
```suggestion
You can use the following arguments:
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+
+| Name            | Type      | Description                                                  | Default | Required |
+| --------------- | --------- | ------------------------------------------------------------ | ------- | -------- |
+| `key`           | `string`  | Attribute key name.                                          |         | yes      |
+| `default_value` | `any`     | Value used when the attribute is missing from the data.      |         | no       |
+| `optional`      | `boolean` | Generate the metric even when the attribute is missing.      | `false` | no       |
+
+You can set at most one of `default_value` or `optional` for a given attribute.
+If neither is set, data points that don't have the attribute are dropped from the generated metric.
```
clayton-cornell:
```suggestion
If you don't set either option, the connector drops data points that don't have the attribute from the generated metric.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+| `count`   | `string`        | OTTL expression that produces the recorded count.  | `""`                     | no       |
+
+If you don't specify `buckets`, the connector uses the following default boundaries:
+`[2, 4, 6, 8, 10, 50, 100, 200, 400, 800, 1000, 1400, 2000, 5000, 10000, 15000]`.
+
+If you don't specify `count`, each matching record increments the count by one.
+
+### `exponential_histogram`
+
+The `exponential_histogram` block generates a base-2 exponential (native) histogram metric.
```
clayton-cornell:
```suggestion
The `exponential_histogram` block generates a base-2 exponential native histogram metric.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Debug information
+
+`otelcol.connector.signaltometrics` doesn't expose any component-specific debug information.
+
+## Examples
+
+### Generate a native histogram from log lines
+
+The following example generates an exponential (native) histogram of log body lengths and exports it over OTLP.
```
clayton-cornell:
```suggestion
The following example generates an exponential native histogram of log body lengths and exports it over OTLP.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+  client {
+    endpoint = "http://localhost:4318"
+  }
+}
+```
+
+### Generate metrics from spans
+
+The following example generates a span duration histogram and a request counter grouped by HTTP method.
+As with the previous example, the connector's delta metrics are routed through [`otelcol.processor.deltatocumulative`][deltatocumulative] and [`otelcol.processor.interval`][interval] to bound the number of data points produced per minute.
```
clayton-cornell:
```suggestion
As with the previous example, the connector's delta metrics are routed through [`otelcol.processor.deltatocumulative`][deltatocumulative] and [`otelcol.processor.interval`][interval] to limit the number of data points produced per minute.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+    endpoint = "http://localhost:4318"
+  }
+}
+```
+
+### Control the number of data points per minute
+
+`otelcol.connector.signaltometrics` produces metrics whenever it receives a batch of telemetry, using delta temporality.
+It has no flush interval of its own, so the rate at which it produces data points is governed by how often upstream components send it data, not by any scrape interval.
+Exporting these metrics directly can produce many data points per minute for each series, which increases cost on backends that price per data point, such as those contracted to one data point per minute, equivalent to a 60-second scrape interval.
```
clayton-cornell:
```suggestion
If you export these metrics directly, it can produce many data points per minute for each series, which increases cost on backends that price per data point, such as those contracted to one data point per minute, equivalent to a 60-second scrape interval.
```

docs/sources/reference/components/otelcol/otelcol.connector.signaltometrics.md
```diff
@@ -0,0 +1,379 @@
...
+
+To bound the rate, place an [`otelcol.processor.deltatocumulative`][deltatocumulative] and an [`otelcol.processor.interval`][interval] processor between the connector and the exporter, as shown in the previous examples:
```
clayton-cornell:
```suggestion
To limit the rate, place an [`otelcol.processor.deltatocumulative`][deltatocumulative] and an [`otelcol.processor.interval`][interval] processor between the connector and the exporter, as shown in the previous examples:
```

clayton-cornell (approved):

bradleypettit:
Thanks @clayton-cornell & @blewis12 - I think we're in a good spot now 😄

blewis12 (approved):

PR #6412, opened 2026-06-04
chore: Fix lint task, workflow and fix lint issues

kalleep:
https://github.com/grafana/alloy/pull/6388 broke go linting. I fixed it and the issues we accumulated since and split up go lint from alloy lint.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,6 +24,12 @@ jobs:
     - name: Lint Go
       run: task lint:go
 
+    - name: Lint Go
+      run: task lint:go
+
+    - name: Run Alloy Lint
+      run: task lint:alloylint 
+
   govulncheck:
     name: Govulncheck
     # xlarge (64 GB): govulncheck's SSA call graph on the
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,9 +24,6 @@ jobs:
     - name: Lint Go
       run: task lint:go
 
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -24,6 +24,12 @@ jobs:
     - name: Lint Go
       run: task lint:go
 
+    - name: Lint Go
+      run: task lint:go
```
ptodev: Why is this the same as above?

kalleep: fixed in [9ec60d2](https://github.com/grafana/alloy/pull/6412/commits/9ec60d22ac1df8425c15eadaa203a5dfeae91aac)

build-tools/task/lint.yml
```diff
@@ -11,17 +11,10 @@ tasks:
       - task: govulncheck
 
   go:
-    desc: Run Go linters.
-    cmds:
-      - task: golangci-lint
-      - task: alloylint
-
-  golangci-lint:
     desc: Run golangci-lint.
```
thampiotr, line 14: This is stale now?

thampiotr: is this still running golangci-lint? it's not clear to me from `go run ./cmd lint go`

kalleep: It's not stale. It is runinng golangci-lint through that command. It does it for every go.mod file

ptodev (approved):

PR #6413, opened 2026-06-04
chore: Migrate promethues components to slog

kalleep:
Migrate promethues components to use `slog/log` instead of `go-kit/log`.

### DIFF
--- a/internal/component/prometheus/echo/echo.go
+++ b/internal/component/prometheus/echo/echo.go
@@ -94,21 +94,19 @@ func (c *Component) Appender(ctx context.Context) storage.Appender {
 	c.mut.RUnlock()
 
 	return &echoAppender{
-		logger:      c.opts.SLogger,
-		componentID: c.opts.ID,
-		format:      format,
-		samples:     make(map[string]sample),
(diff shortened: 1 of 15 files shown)


internal/component/prometheus/echo/echo.go
```diff
@@ -240,21 +237,20 @@ func (a *echoAppender) Commit() error {
 	case "text", "":
 		expFormat = expfmt.NewFormat(expfmt.TypeTextPlain)
 	default:
-		level.Warn(a.logger).Log("component", a.componentID, "msg", "unknown format, using text", "format", a.format)
+		a.logger.Warn("unknown format, using text", "format", a.format)
```
x1unix, line 240: Component ID isn't needed anymore?

x1unix (approved):

internal/component/prometheus/echo/echo.go
kalleep: Loggers passed to components is already prefixed with component is https://github.com/grafana/alloy/blob/main/internal%2Fruntime%2Finternal%2Fcontroller%2Fnode_builtin_component.go#L188

PR #6432, opened 2026-06-05
feat(prometheus.remote_write): Support Azure workload identity and custom OAuth scope in azuread

rknightion:

Expose Azure Workload Identity and a custom OAuth 2.0 `scope` in the `prometheus.remote_write` `azuread` block, allowing Alloy to authenticate with secretless AKS federated tokens and to request Microsoft Entra tokens for a custom (non–Azure Monitor) audience.

### Pull Request Details

Alloy's pinned Prometheus dependency already implements both features in `azuread.AzureADConfig` (the `WorkloadIdentity` and `Scope` fields), but Alloy never surfaced them in its own `prometheus.remote_write` config schema or in the `alloy convert` path. This change adds that plumbing; no dependency bump is required.

The `azuread` block gains an optional `workload_identity` block with `client_id`, `tenant_id`, and an optional `token_file_path` that defaults to the standard Azure Workload Identity projected-token path (`/var/run/secrets/azure/tokens/azure-identity-token`). It also gains an optional `scope` attribute. When `scope` is empty, Alloy keeps using the per-cloud Azure Monitor ingestion audience; when it's set, Alloy requests a token for that audience instead.

Validation is reworked to mirror upstream's `azuread.AzureADConfig.Validate()`: authenticator exclusivity is now count-based (a single "cannot provide multiple authentication methods" message that includes Workload Identity), `client_id` and `tenant_id` are validated as UUIDs, and `scope` is checked against the same character set upstream uses. This also corrects a pre-existing copy-paste error where a failed SDK `tenant_id` validation reported an "Azure OAuth" message.

The `alloy convert` path (`toAzureAD`) maps both fields so existing Prometheus `azuread` configurations that use `workload_identity` or `scope` convert correctly, with `token_file_path` defaulted by the Prometheus loader.

Existing `azuread` configurations are unaffected: both the `workload_identity` block and the `scope` attribute are optional, and the default behavior is unchanged.

N/A

### Notes to the Reviewer

Scope is metrics-only by design. `loki.write` has no `azuread` block (it uses Prometheus's generic HTTP client config) and `otelcol` exporters have no Azure auth extension, so neither is touched here.

Validation run:

- `go test ./internal/component/prometheus/remotewrite/... ./internal/converter/internal/prometheusconvert/...`
- `gofmt -l` (clean) and `go vet` (clean) on both packages
- `build/alloylint ./internal/component/prometheus/remotewrite/... ./internal/converter/internal/prometheusconvert/...` (clean)
- `alloy validate` on a config that uses `workload_identity` and `scope` (succeeds)

`golangci-lint run --new-from-rev=HEAD` reports no new issues for this change. A full `golangci-lint run` over these packages surfaces only pre-existing repo-wide `goconst`/`govet` findings in unrelated files.

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.remote_write.md
+++ b/docs/sources/reference/components/prometheus/prometheus.remote_write.md
@@ -140,6 +140,8 @@ If the endpoint doesn't support receiving native histogram samples, pushing metr
 
 {{< docs/shared lookup="reference/components/azuread-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
+You must configure exactly one of the [`managed_identity`][managed_identity], [`oauth`][oauth], [`sdk`][sdk], or [`workload_identity`][workload_identity] blocks.
+
 ### `managed_identity`
 
 {{< badge text="Required" >}}
--- a/docs/sources/reference/components/prometheus/prometheus.remote_write.md
+++ b/docs/sources/reference/components/prometheus/prometheus.remote_write.md
@@ -55,6 +55,7 @@ You can use the following blocks with `prometheus.remote_write`:
 | `endpoint` > `azuread` > [`managed_identity`][managed_identity] | Configure Azure user-assigned managed identity.                            | yes      |
 | `endpoint` > `azuread` > [`oauth`][oauth]                       | Configure Azure OAuth.                                                     | yes      |
 | `endpoint` > `azuread` > [`sdk`][sdk]                           | Configure Azure SDK authentication.                                        | yes      |
+| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Azure Workload Identity.                                       | yes      |
 | `endpoint` > [`basic_auth`][basic_auth]                         | Configure `basic_auth` for authenticating to the endpoint.                 | no       |
 | `endpoint` > [`metadata_config`][metadata_config]               | Configuration for how metric metadata is sent.                             | no       |
 | `endpoint` > [`oauth2`][oauth2]                                 | Configure OAuth 2.0 for authenticating to the endpoint.                    | no       |
@@ -75,6 +76,7 @@ You can use the following blocks with `prometheus.remote_write`:
 [oauth2]: #oauth2
 [queue_config]: #queue_config
 [sdk]: #sdk
+[workload_identity]: #workload_identity
 [sigv4]: #sigv4
 [tls_config]: #tls_config
 [wal]: #wal
@@ -155,6 +157,12 @@ If the endpoint doesn't support receiving native histogram samples, pushing metr
 {{< badge text="Required" >}}
 
 {{< docs/shared lookup="reference/components/azuread-sdk.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+### `workload_identity`
+
+{{< badge text="Required" >}}
+
+{{< docs/shared lookup="reference/components/azure-workload_identity-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
 
 ### `basic_auth`
 
--- a/docs/sources/reference/components/prometheus/prometheus.remote_write.md
+++ b/docs/sources/reference/components/prometheus/prometheus.remote_write.md
@@ -55,7 +55,7 @@ You can use the following blocks with `prometheus.remote_write`:
 | `endpoint` > `azuread` > [`managed_identity`][managed_identity] | Configure Azure user-assigned managed identity.                            | yes      |
 | `endpoint` > `azuread` > [`oauth`][oauth]                       | Configure Azure OAuth.                                                     | yes      |
 | `endpoint` > `azuread` > [`sdk`][sdk]                           | Configure Azure SDK authentication.                                        | yes      |
-| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Azure AD Workload Identity.                                    | yes      |
+| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Microsoft Entra Workload ID.                             | yes      |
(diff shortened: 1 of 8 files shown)


docs/sources/shared/reference/components/azure-workload_identity-block.md
```diff
@@ -0,0 +1,24 @@
...
+
+`client_id` and `tenant_id` must be valid [UUID][]s in one of the supported formats:
+
+* `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+* `urn:uuid:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+* Microsoft encoding: `{xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx}`
+* Raw hex encoding: `xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`
+
+Use `workload_identity` to authenticate from an AKS Pod configured with [Microsoft Entra Workload ID][workload-id] without requiring a client secret.
+Alloy reads the projected service account token from `token_file_path` and federates it for a Microsoft Entra token.
```
clayton-cornell:
```suggestion
{{< param "PRODUCT_NAME" >}} reads the projected service account token from `token_file_path` and federates it for a Microsoft Entra token.
```

docs/sources/shared/reference/components/azuread-block.md
```diff
@@ -4,12 +4,15 @@ description: Shared content, azuread block
...
+| `cloud` | `string` | The Azure Cloud.                                                    | `"AzurePublic"` | no       |
+| `scope` | `string` | Custom OAuth 2.0 scope (audience) to request when acquiring tokens. | `""`            | no       |
 
 The supported values for `cloud` are:
 
 * `"AzurePublic"`
 * `"AzureChina"`
 * `"AzureGovernment"`
+
+When `scope` is left empty, Alloy uses the per-cloud Azure Monitor ingestion audience, for example, `https://monitor.azure.com//.default` for `"AzurePublic"`. Set `scope` to request a token for a different audience, such as a custom Microsoft Entra app registration.
```
clayton-cornell:
```suggestion
When `scope` is left empty, {{< param "PRODUCT_NAME" >}} uses the per-cloud Azure Monitor ingestion audience, for example, `https://monitor.azure.com//.default` for `"AzurePublic"`.
Set `scope` to request a token for a different audience, such as a custom Microsoft Entra app registration.
```

docs/sources/reference/components/prometheus/prometheus.remote_write.md
```diff
@@ -55,6 +55,7 @@ You can use the following blocks with `prometheus.remote_write`:
 | `endpoint` > `azuread` > [`managed_identity`][managed_identity] | Configure Azure user-assigned managed identity.                            | yes      |
 | `endpoint` > `azuread` > [`oauth`][oauth]                       | Configure Azure OAuth.                                                     | yes      |
 | `endpoint` > `azuread` > [`sdk`][sdk]                           | Configure Azure SDK authentication.                                        | yes      |
+| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Azure Workload Identity.                                       | yes      |
```
clayton-cornell: What's the vendor name here? it looks like the proper product name has been rebranded to **Microsoft Entra Workload ID**.

rknightion: it has been yes. Although the [upstream prometheus lib](https://pkg.go.dev/github.com/prometheus/prometheus/storage/remote/azuread) stil uses azuread and workload_identity in the package (but happy to change the descriptions?)

clayton-cornell: I did some digging here and it seems to be a bit of a hot issues when the name was changed.... with no real "right" solution. RedHat uses `Microsoft Entra Workload ID (formerly Azure AD Workload Identity)`.

We should at least use the full former name (with the AD part)... and since the keys, vars etc all use formatting like `azure.workload.identity`... keeping teh former name isn't terrible :-)

```suggestion
| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Azure AD Workload Identity.                                    | yes      |
```

clayton-cornell: Oh wait I see it's `Microsoft Entra Workload ID` in the shared content... we need to make this consistent.

docs/sources/reference/components/prometheus/prometheus.remote_write.md
```diff
@@ -55,6 +55,7 @@ You can use the following blocks with `prometheus.remote_write`:
 | `endpoint` > `azuread` > [`managed_identity`][managed_identity] | Configure Azure user-assigned managed identity.                            | yes      |
 | `endpoint` > `azuread` > [`oauth`][oauth]                       | Configure Azure OAuth.                                                     | yes      |
 | `endpoint` > `azuread` > [`sdk`][sdk]                           | Configure Azure SDK authentication.                                        | yes      |
+| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Azure AD Workload Identity.                                    | yes      |
```
clayton-cornell:
```suggestion
| `endpoint` > `azuread` > [`workload_identity`][workload_identity] | Configure Microsoft Entra Workload ID.                             | yes      |
```

Revisiting this (sorry for the churn).. the shared element uses `Microsoft Entra Workload ID` so we should use the same here for consistency.

clayton-cornell:
Docs look OK to me for now. Over to @grafana/grafana-alloy-maintainers for a code review :-)

kgeckhart:
I just noticed this but I believe https://github.com/grafana/alloy/pull/6637 covered this and more which will land in the next minor that will happen soon.

rknightion:
ah yes it does! I'll close this :)

PR #6442, opened 2026-06-05
feat: Inline config support for Alloy Otel Extension

### LINKED ISSUE #6420: Explore Implementation for Inline Config in Alloy Engine Extension
### Component(s)

_No response_

### Add a description

With the current setup, if a user is using the alloy engine extension and is managing the OTel engine with the supervisor, they could technically do something like the following:

```
extensions:
  alloyengine:
    config:
      file: ./my_config.alloy
```

and within `my_config.alloy` they could define

```
remotecfg {
...
}
```

Then, if they ran a supervised OTel Engine, we'd have **both** engines connecting to FM. Although this works, it's a clunky experience, and we should aim for users to be able to manage both engines through a single FM connection. A good way to do this is via inline configuration, something like

```
extensions:
  alloyengine:
    config:
      inline:  """
        prometheus.exporter.self "self" {}
        prometheus.scrape "demo" {
          targets    = prometheus.exporter.self.example.targets
          forward_to = [....]
        } 
        """
```

This way, the backend OpAmp server include the default engine config embedded in its OTel config that it sends over. 

We should also disable the use of `remotecfg` in whatever config is passed through the extension - we don't want to support two different connections managing two different engines in the same collector instance. 

[Relevant design doc](https://docs.google.com/document/d/1JkSRsaP-69Jac1uaP-Mup607ygaywTkFJNx5qzRxxOA/edit?tab=t.xwaw2ojna2tc#heading=h.rbvj3dxc0o2j) which talks about this and why running both engines through two FM connections is a bad idea.

### Tip

x1unix:

Support passing Alloy config to the extension as inline string.

### Pull Request Details

This PR adds `alloyengine.config.content` which allows specifying Alloy inline config.

Motivation is to be able to provision Alloy config via OpAMP.

Besides that:
- `alloyengine.config.file` was renamed to `alloyengine.config.path` as Alloy config path parameter can be a path to a directory as well.
- Remote configuration feature is disabled when Alloy is running inside the collector. \
  Declaring `remotecfg {}` will throw an error during startup.

#### Caveats

The Alloy configuration language has [`module_path` keyword](https://github.com/grafana/alloy/blob/x1unix/feat/alloy-ext-inline-cfg/internal/runtime/alloy.go#L314) which provides a directory where configuration file is stored.

The extension doesn't have information about a location of loaded configuration file(s). Also, bootstrap configuration can be provided by other config providers.

Therefore, this PR adds an optional `alloyengine.config.module_path` parameter to set a value for the keyword.

By default, `module_path` points to a current working directory.

```yaml
extensions:
  alloyengine:
    config:
      inline:
        module_path:  "/foo/bar" # defaults to current working directory
        content: |
          ...
```

Fixes #6420

### PR Checklist

- [x] Tests updated


extension/alloyengine/config.go
```diff
@@ -25,13 +36,8 @@ func (cfg *Config) flagsAsSlice() []string {
...
 
 func (cfg *Config) Validate() error {
-	if cfg.AlloyConfig.File == "" {
-		return fmt.Errorf("config.file is required")
-	}
-
-	_, err := os.Stat(cfg.AlloyConfig.File)
-	if err != nil {
-		return fmt.Errorf("provided config path %s does not exist or is not readable: %w", cfg.AlloyConfig.File, err)
+	if cfg.AlloyConfig.Content == "" {
```
blewis12: we'll need to add back the validation logic that checks on cfg.AlloyConfig.File - since a user can specify either

extension/alloyengine/config.go
```diff
@@ -2,18 +2,29 @@ package alloyengine
...
 type AlloyConfig struct {
+	// ModulePath is value to be resolved for "module_path" alloy config keyword.
+	//
+	// Has no effect if [File] is set.
+	ModulePath string `mapstructure:"module_path"`
+
+	// File is a path to Alloy config file or a directory containing config files.
+	//
+	// Note: either [File] or [Content] can be set.
 	File string `mapstructure:"file"`
```
blewis12, line 22: I'm wondering about the best way we can define the schema here. I'm siding on something like 

```
config
  file: string
  inline:
    module_path: optional string
    content: string
```

since `module_path` currently is only relevant to whenever the config is `inline`. WDYT?

extension/alloyengine/extension.go
```diff
@@ -220,3 +236,107 @@ func (e *alloyEngineExtension) NotReady() error {
...
+		if err := validateAlloyConfig("config.alloy", data); err != nil {
+			return "", nil, fmt.Errorf("invalid inline Alloy config: %w", err)
+		}
+
+		files = map[string][]byte{"config.alloy": data}
+		return modulePath, files, nil
+	}
+
+	// Alloy supports accepting a directory as config source
+	stat, err := os.Lstat(cfg.File)
```
blewis12: Good point in that a directory is also a possibility, I'd rather we defined this as a separate config param called `directory` - having a param called `file` which could actually be a directory and is handled as either feels confusing. 

Maybe we could handle passing directories through as a follow up and in this PR only add validation that `file` is in fact a file and not a directory? We don't document anywhere that the extension accepts a directory anyway

extension/alloyengine/config.go
blewis12: There's also overlapping validation in `buildAlloyConfig`. We could keep validation here as simply checking that the user has specified _any_ config (aka both `content` and `file` are not empty) and `buildAlloyConfig` can take care of checking that the provided context itself is syntactically valid. WDYT?

extension/alloyengine/extension.go
x1unix: @blewis12 That's fair. What do you think about using a more general `config.path` parameter instead?

The extension already supported directories, similar to regular Alloy, and the new implementation keeps that behavior. That would let us preserve the existing behavior while avoiding a misleading file name.

extension/alloyengine/config.go
x1unix: @blewis12 my initial motivation was to avoid nesting, but I see your point :+1:

extension/alloyengine/config.go
x1unix: @blewis12 I missed that method, my bad.

extension/alloyengine/config.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6442/changes/3e75fb5289e7a8d8f9df82d4d4f7965927938b3e

extension/alloyengine/config.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6442/changes/0e6b432e7d85e71141cb27d25bb0ba88d7282e66

extension/alloyengine/extension.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6442/changes/f72b9fb93c374e20737bde3a23a5578299724fe4

extension/alloyengine/extension.go
```diff
@@ -220,3 +236,107 @@ func (e *alloyEngineExtension) NotReady() error {
...
+}
+
+// validateAlloyConfig checks whether Alloy config contains statements unsupported in extension mode.
+func validateAlloyConfig(fname string, data []byte) error {
+	tree, err := parser.ParseFile(fname, data)
+	if err != nil {
+		return fmt.Errorf("failed to parse config file: %w", err)
+	}
+
+	// TODO: throw warning if 'import.file' uses 'module_path' in inline config.
```
blewis12, line 334: A thought regarding `import.file`/`module_path`, there are two scenario's to consider:

1. The user has defined inline configuration in the extension, in which case `module_path` doesn't make a whole lot of sense because we have no relative path to the default engine configuration (because it's not a file on disk anymore). It's this case that we should probably just explicitly just disallow the usage of `module_path`
2. The user has defined a `path` to the configuration in the extension, in which case we _do_ now have a path to the config on disk - which means we can determine `module_path` automatically

Basically, I feel like we might not even need `module_path` in the extension config? Because in case 1 we just disallow it, and in case 2 it's available anyway 🤔

extension/alloyengine/extension.go
blewis12: discussed offline - I think that makes sense!

extension/alloyengine/extension.go
x1unix: @blewis12 I would prefer to keep `module_path` for now to provide a smooth migration path for users with existing Alloy configs which may use `module_path` keyword in inline config.

blewis12: discussed offline - let's do this for now yeah 👍

extension/alloyengine/extension.go
```diff
@@ -220,3 +236,107 @@ func (e *alloyEngineExtension) NotReady() error {
 		return fmt.Errorf("alloyengine extension not ready in current state: %s", currentState.String())
 	}
 }
+
+func buildAlloyConfig(cfg AlloyConfig) (modulePath string, files map[string][]byte, err error) {
+	if cfg.Inline.Content == "" && cfg.Path == "" {
+		return "", nil, errors.New("missing alloy config. Either config.file or config.inline.content must be set")
+	}
```
blewis12, line 249: This is already checked in `Validate()` right?  The collector already validates config on parse based on `Validate()` which runs before `Start()` - so we can assume that by the time we get to `buildAlloyConfig` these fields are already checked against

extension/alloyengine/README.md
```diff
@@ -17,20 +17,27 @@ The extension accepts the following configuration fields:
...
 ### Config Object
 
-The `config` object specifies the Alloy configuration source.
-Currently we only support `file` as an input type, but this is planned to be extended to other formats
+The `config` object specifies the Alloy configuration source. Either `path` or `inline` must be set, but not both.
 
 | Field | Type | Required | Default | Description |
 |-------|------|----------|---------|-------------|
-| `file` | string | Yes | - | The path to the Alloy configuration file to run. |
+| `path` | string | No | - | Path to an Alloy config file or a directory containing `.alloy` files. |
```
blewis12, line 24: nit: let's update this in our [setup docs](https://github.com/grafana/alloy/blob/e11d0f03e2b2e727161fb2169cade54698e17bd8/docs/sources/set-up/otel_engine.md?plain=1#L104) as well (rename `file` to `path`)

extension/alloyengine/config.go
```diff
@@ -25,13 +39,17 @@ func (cfg *Config) flagsAsSlice() []string {
...
-		return fmt.Errorf("config.file is required")
-	}
+	hasPath := cfg.AlloyConfig.Path != ""
+	hasContent := cfg.AlloyConfig.Inline.Content != ""
 
-	_, err := os.Stat(cfg.AlloyConfig.File)
-	if err != nil {
-		return fmt.Errorf("provided config path %s does not exist or is not readable: %w", cfg.AlloyConfig.File, err)
+	if !hasPath && !hasContent {
+		return fmt.Errorf("either config.file or config.inline.content must be set")
```
blewis12: these error messages still say `config.file` but should say `config.path`

internal/alloycli/cmd_run.go
```diff
@@ -106,57 +110,140 @@ depending on the nature of the reload error.
...
+
+// NewRunAsExtensionCommand returns a standalone cobra command to run Alloy inside OTel collector as an extension.
+//
+// In extension mode:
+//   - Certain features like remote config are disabled.
+//   - Config reload on NOHUP signal is disabled.
+//   - Alloy config contents passed directly, instead of file path cli argument.
+func NewRunAsExtensionCommand(params ExtensionModeParams) *cobra.Command {
+	r := newAlloyRun()
+	r.isExtension = true
```
blewis12: is this used anywhere?

extension/alloyengine/extension_test.go
```diff
@@ -43,14 +44,16 @@ func requireShutdownAndTerminated(t *testing.T, e *alloyEngineExtension) {
 func defaultTestConfig() *Config {
```
blewis12, line 45: Could we add a test that covers the blocking of `remotecfg` in both inline and path based configuration? Since that's a key addition in this PR

blewis12 (commented):
I think we're almost there :) left some last comments but nothing major. In terms of documentation, the rename of `file` to `path` is enough for this PR I think since it's a breaking change - as we discussed before we should have a follow up PR which introduces more in-depth documentation to reflect the other changes made here (`remotecfg` disallow, `module_path` usage etc)

extension/alloyengine/config.go
```diff
@@ -2,18 +2,32 @@ package alloyengine
...
 type Config struct {
 	AlloyConfig AlloyConfig       `mapstructure:"config"`
 	Flags       map[string]string `mapstructure:"flags"`
 }
 
-// This type represents the incoming format of the Alloy configuration
-// This is a one-of type, and it is expected that only one of the fields will be set (ie, we cannot define multiple config sources of different types)
+// AlloyConfig represents the incoming format of the Alloy configuration.
 type AlloyConfig struct {
-	File string `mapstructure:"file"`
```
thampiotr, line 16: That's a breaking change. I know we're still not in GA, but I wonder if it's worth the hassle to rename. Personally if I see `file` and `inline` it makes sense to me, so moving to `path` and `inline` may not be worth it. What do you think?

x1unix: @thampiotr Alloy config path can be either a path to a file or a path to a dir that contains configs, therefore the field was renamed.

This came in during offline discussion of https://github.com/grafana/alloy/pull/6442#discussion_r3387994171 and https://github.com/grafana/alloy/pull/6442#discussion_r3394205893.

x1unix: @thampiotr per offline discussion - I'll add a fallback to `file` with a deprecation notice.

extension/alloyengine/README.md
x1unix: Done: https://github.com/grafana/alloy/pull/6442/changes/d6526f36afffafc1400f3cccf91765a6303c8bc8

extension/alloyengine/extension.go
x1unix: Done: https://github.com/grafana/alloy/pull/6442/changes/41d40420a3d68c74071ad62b78da566195d2fd42

internal/alloycli/cmd_run.go
x1unix: https://github.com/grafana/alloy/pull/6442/changes/2e2b79331d8f2b15af153ae5d6f00572aee3e77c

extension/alloyengine/extension_test.go
x1unix: Done: https://github.com/grafana/alloy/pull/6442/changes/9999e27d82b57b83254872b27840b691811cc62c

extension/alloyengine/config.go
x1unix: @thampiotr added a fallback: https://github.com/grafana/alloy/pull/6442/changes/d2f7faf5fa5dcc801620155e86f2d67743dbd731

extension/alloyengine/config.go
x1unix: Done: https://github.com/grafana/alloy/pull/6442/changes/df7f0ea0075197bd9bbdb75213ce857a7d018921

extension/alloyengine/extension.go
```diff
@@ -220,3 +240,109 @@ func (e *alloyEngineExtension) NotReady() error {
...
+		if err := validateAlloyConfig("config.alloy", data); err != nil {
+			return "", nil, fmt.Errorf("invalid inline Alloy config: %w", err)
+		}
+
+		files = map[string][]byte{"config.alloy": data}
+		return modulePath, files, nil
+	}
+
+	// Alloy supports accepting a directory as config source
+	stat, err := os.Lstat(effectivePath)
```
thampiotr: os.Stat would follow symlinks and it was used before - is it intentional? I think we do want to follow symlinks

extension/alloyengine/extension.go
```diff
@@ -220,3 +240,109 @@ func (e *alloyEngineExtension) NotReady() error {
...
+	// TODO: throw warning if 'import.file' uses 'module_path' in inline config.
+	for _, stmt := range tree.Body {
+		block, ok := stmt.(*ast.BlockStmt)
+		if !ok {
+			continue
+		}
+
+		blockName := block.GetBlockName()
+		if blockName == remotecfg.ServiceName {
+			return fmt.Errorf("block %q is not supported in Alloy extension", blockName)
```
thampiotr, line 343: if the main file has an import statement that imports another file e.g. via http that has remotecfg service, this check will not catch this, right?

Should we add a test for this edge case (as it's a bit hard to follow with all the modules) and make sure that remotecfg is disallowed in extension even if the extension gets it via import.* statements?

extension/alloyengine/extension.go
x1unix: @thampiotr thanks for notice, I thought the opposite by mistake - that `Lstat` follows symlinks but `Stat` doesn't.

extension/alloyengine/extension.go
x1unix: @thampiotr I have a todo there for `import.*` statements and planned to do that in a separate PR.

The `remotecfg` block check in imports should probably be done in Alloy runtime which handles imports.

extension/alloyengine/extension.go
x1unix: Fixed in https://github.com/grafana/alloy/pull/6442/changes/46c4db2a5991e7cadc7111908912d5f41e0fcbbd

blewis12 (approved):

PR #6451, opened 2026-06-08
chore: Finish up slog migration

### LINKED ISSUE #4813: internal: replace go-kit/log

kalleep:
### Pull Request Details
With this pr alloy:s slog migration is complete:
1. Rename `SLogger` to `Logger`
2. Update util.TestLogger to return *slog.Logger
3. Update windows_service to use slog 
4. Remove now unused level package
5. Clean up tests for the `go-kit/log` interface in logging package
6. Update so that we only buffer slog.Records since we no longer expose go-kit interface

Fixes: https://github.com/grafana/alloy/issues/4813

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/runtime/logging/logger_test.go
```diff
@@ -37,155 +35,8 @@ func (b *safeBuffer) String() string {
...
-BenchmarkLogging_GoKitLevel_Prints-8          	  678214	      1669 ns/op	     849 B/op	      16 allocs/op
-BenchmarkLogging_Slog_Drops-8                 	79687671	        15.09 ns/op	       8 B/op	       0 allocs/op
-BenchmarkLogging_Slog_Prints-8                	 1000000	      1119 ns/op	      32 B/op	       2 allocs/op
-BenchmarkLogging_AlloyLevel_Drops-8            	21693330	        58.45 ns/op	     168 B/op	       2 allocs/op
-BenchmarkLogging_AlloyLevel_Prints-8           	  720554	      1672 ns/op	     833 B/op	      15 allocs/op
-*/
-
 const testStr = "this is a test string"
 
-func TestLevels(t *testing.T) {
```
dehaansa, line 58: Why were these tests removed?

kalleep: These just tested levels for both `go-kit/log/level` and internal package we have.

I can add tests back for slog if you think it's important

kalleep: I added some tests back in [42f694d](https://github.com/grafana/alloy/pull/6451/commits/42f694dfde965a2fe933130f21def69794ddf0ee).

Just to test different levels

dehaansa: Mainly just was curious because I thought they would validate the Copilot concern of the log representation changing from warn to WARN and it causing issues with the windows service logger.

kalleep: >The Windows Event Log writer is doing a case-sensitive substring check for "warn"/"error" to decide which eventlog method to call. With the new slog.TextHandler output, levels are typically rendered as "level=WARN"/"level=ERROR" (uppercase), so warnings/errors will be misclassified as Info events.

Well they are lower case, and after https://github.com/grafana/alloy/pull/5626 we write directly to windows event log when using windows service

dehaansa (approved):

PR #6454, opened 2026-06-08
ci: Sign dev container images with cosign

kgeckhart:

Sign the `alloy-dev` container images (Linux, Linux boringcrypto, Windows) on every push to `main` using keyless cosign via GitHub Actions OIDC. No Grafana product signs its images today, so this lights it up on dev first — proving the full path (digest capture on real runners, keyless OIDC, signature push to GAR, and the GAR→Docker Hub mirror carrying the `.sig`) on low-stakes images before release signing goes live.

### Pull Request Details

A reusable `sign-and-attest.yml` does the work: it signs the image index by digest (`cosign sign`, index-only, public Rekor), then verifies its own signature against the GAR digest before the job exits. The Linux/Windows build scripts surface the pushed digest as a job output, and the dev publish workflow wires a sign job per image. `alloy-dev` is on the `gar-image-mirror` allowlist, so the legacy-tag signature replicates to `docker.io/grafana/alloy-dev` for free.

### DIFF
--- a/.github/workflows/publish-alloy-devel.yml
+++ b/.github/workflows/publish-alloy-devel.yml
@@ -39,7 +39,37 @@ jobs:
     with:
       img-name: alloy-devel
       dev: true
-        
+
+  sign_linux_container:
+    needs: publish_linux_container
+    if: ${{ needs.publish_linux_container.outputs.image != '' }}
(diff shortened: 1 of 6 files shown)


blewis12 (approved):
will be cool to see how this runs e2e 👀

PR #6459, opened 2026-06-09
perf(loki.source.journal): Reduce allocations for labels

kalleep:
### Pull Request Details
When building labels for journal entries we do a lot of unnecessary allocations and this shows up in profiles.

What we did for each:
1. We created a map and add entry fields and static labels
2. We constructed `labels.Labels` from this map and passed it to a labels.Builder
3. After relabeling we converted the result back to labels.Labels so we could build a map from it
4. We loop over this map and build the final model.LabelSet

This is a lot of unnecessary allocations. In this pr I changed it so that we reuse the same labels.Builder pre-populated with the static labels. We then add all journal entry fields to this builder and run relabel rules. After we loop over entries in builder and construct the final model.LabelSet.

The only annoying thing is that there is no way to get the length for resulting labels from the builder without doing allocations, e.g. by converting it to labels.Labels. For now I left it empty but we could set the length to `len(entry.Fields)+t.labels.Len()+1`  but this would most of the time over allocate.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/journal/tailer.go
+++ b/internal/component/loki/source/journal/tailer.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/coreos/go-systemd/v22/sdjournal"
 	"github.com/grafana/alloy/internal/loki/promtail/scrapeconfig"
+	"github.com/grafana/alloy/internal/loki/util"
 	"github.com/grafana/loki/pkg/push"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/prometheus/common/model"
@@ -106,7 +107,9 @@ type tailer struct {
 	positionPath  string
 	relabelConfig []*relabel.Config
 	config        *scrapeconfig.JournalTargetConfig
-	labels        model.LabelSet
+
+	builder *labels.Builder
+	labels  labels.Labels
 
 	r     journalReader
 	until chan time.Time
@@ -158,6 +161,8 @@ func newTailerWithReader(
 		entryFunc = defaultJournalEntryFunc
 	}
 
+	lbls := labels.FromMap(util.ModelLabelSetToMap(targetConfig.Labels))
+
 	until := make(chan time.Time)
 	t := &tailer{
 		metrics:       metrics,
@@ -166,7 +171,8 @@ func newTailerWithReader(
 		positions:     pos,
 		positionPath:  positionPath,
 		relabelConfig: relabelConfig,
-		labels:        targetConfig.Labels,
+		labels:        lbls,
+		builder:       labels.NewBuilder(lbls),
 		config:        targetConfig,
 
 		until: until,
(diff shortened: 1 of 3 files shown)


internal/component/loki/source/journal/tailer.go
```diff
@@ -311,31 +316,22 @@ func (t *tailer) formatter(entry *sdjournal.JournalEntry) (string, error) {
...
-	lb := labels.NewBuilder(labels.FromMap(entryLabels))
-	var processedLabels labels.Labels
-	if relabel.ProcessBuilder(lb, t.relabelConfig...) {
-		processedLabels = lb.Labels()
-	} else {
-		processedLabels = labels.EmptyLabels()
+	addJournalFields(t.builder, entry.Fields)
+	if !relabel.ProcessBuilder(t.builder, t.relabelConfig...) {
+		t.logger.Debug("journal entry dropped by relabel rules", "unit", entry.Fields["_SYSTEMD_UNIT"])
+		t.metrics.journalErrors.WithLabelValues(emptyLabelsError).Inc()
```
kgeckhart, line 322: It's consistent with the current behavior but it feels weird to track this under the journal errors counter. Not sure how consistent we are with this in other components.

kgeckhart (dismissed):
Looks clean, would it be useful to have a benchmark on to see what we are looking at +-?

internal/component/loki/source/journal/tailer.go
kalleep: Yes you are right, it's pretty weird it's recorded as an error :thinking:

Ideally sources should have similar metrics but I don't think we are consistent around them. I can create a issue for it and see what we have and what we ideally should have for metrics.

In my mind there are at least two kind of metrics that should always be useful:
1. Lines successfully forwarded
2. Lines dropped, with a potential reason. I guess this metric is under this category now but that name is not really fitting.

kalleep:
>Looks clean, would it be useful to have a benchmark on to see what we are looking at +-?

Yes we should. I will add a benchmark  for the whole component because I have an [pr with bigger changes ](https://github.com/grafana/alloy/pull/6508) that I would also like to compare against.

kalleep:
Added a benchmark that start this components and consumes up to 10k entries, it always stops after 1 second. 

Diff against main:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/alloy/internal/component/loki/source/journal
           │ bench_old.txt │         bench_new.txt          │
           │    sec/op     │   sec/op     vs base           │
Journal-16     1.003 ± ∞ ¹   1.002 ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

           │ bench_old.txt  │          bench_new.txt           │
           │      B/op      │     B/op       vs base           │
Journal-16   122.84Mi ± ∞ ¹   53.25Mi ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

           │ bench_old.txt │          bench_new.txt          │
           │   allocs/op   │  allocs/op    vs base           │
Journal-16    1.631M ± ∞ ¹   1.181M ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

kgeckhart (approved):
Nice work!

PR #6475, opened 2026-06-10
feat(pyroscope.ebpf): Add no_kernel_version_check argument

marcsanmi:

Expose the eBPF profiler's `NoKernelVersionCheck` option as a `no_kernel_version_check` component argument.

### Pull Request Details

The underlying profiler refuses to load the eBPF tracer on kernel versions affected by a known probe-read bug, and suggests overriding the check with `-no-kernel-version-check`. The option exists in the profiler's controller config but was never mapped in the component's `Convert()`, so users hitting this on kernels with backported fixes had no way to override it.

The argument defaults to `false`, which keeps the kernel version check enabled, the same behavior as before, so existing configurations are unaffected.

### Notes to the Reviewer

The flag is plumbed through `controller.Config.NoKernelVersionCheck`, which the profiler already honors (`KernelVersionCheck: !NoKernelVersionCheck` when building the tracer config).


korniltsev-grafanista (dismissed):

dehaansa (dismissed):

dehaansa (dismissed):

docs/sources/reference/components/pyroscope/pyroscope.ebpf.md
```diff
@@ -84,6 +84,7 @@ You can use the following arguments with `pyroscope.ebpf`:
 | `go_table_fallback`       | `bool`                   | Deprecated (no-op), previously enabled symbol lookup in `.sym` / `.dynsym` sections when `.gopclntab` lookup failed. | `false`  | no       |
 | `hotspot_enabled`         | `bool`                   | A flag to enable or disable HotSpot profiling.                                                                       | `true`   | no       |
 | `kernel_frames`           | `bool`                   | Include kernel-space frames in collected profiles. Set to `false` to drop kernel frames from each stack trace.       | `true`   | no       |
+| `no_kernel_version_check` | `bool`                   | Disable checking the kernel version for eBPF support. Use at your own risk, to run on older kernels with backported eBPF features. | `false`  | no       |
```
clayton-cornell:
```suggestion
| `no_kernel_version_check` | `bool`                   | Skip the kernel version check for eBPF support.                                                                      | `false`  | no       |
```

We can move the risk statement out of the table and into a caution admonition.

docs/sources/reference/components/pyroscope/pyroscope.ebpf.md
clayton-cornell, line 104:
```suggestion

{{< admonition type="caution" >}}
Use `no_kernel_version_check` only when you run on an older kernel that includes backported eBPF features from a later release.
When you skip this check, the profiler can't verify that your kernel has the eBPF features it needs, and profiling may fail or produce incomplete results.
Proceed at your own risk.
{{< /admonition >}}
```

Maybe something like this?

dehaansa (approved):

PR #6477, opened 2026-06-10
feat: Implement Beyla as a subprocess

rafaelroquetto:

This PR reimplements the `beyla.ebpf` component, changing it to simply embed Beyla as a binary blob and execute it as a subprocess, doing without the source level integration.

### Pull Request Details

This is a major refactoring of the `beyla.ebpf` component. The main but not single motivation behind it is to do without the dependency hell that tends to plague every Beyla release inside alloy. By embedding Beyla as an opaque binary blob, we define a clean interface between the Alloy and Beyla domains.

`beyla.ebpf` will now simply extract the embedded Beyla binary to an anonymous memory location, then execute the binary. Once the binary is executed, that memory location is no longer required and is therefore freed.

The same approach is used to pass the Beyla configuration file. This mechanism is implemented with the help of `memfd()`. A typical instance of Alloy running one `beyla.ebpf` instance now appears as such:

```
3695624 ?        Ssl    4:44  \_ /bin/alloy run /etc/alloy/config.alloy --storage.path=/tmp/alloy --server.http.listen-addr=0.0.0.0:12345 --server.http.ui-path-prefix=/ --disable-reporting -
3695702 ?        Sl    85:42      \_ /proc/self/fd/6 -config /proc/self/fd/11
```

The file descriptor `6` points to the Beyla binary, whereas the file descriptor `11` points to the config file.

Communication between the parent and child process is done via abstract Unix socket endpoints: `@alloy-beyla-health-<id>-<pid>-<n>`) for health checks, `@alloy-beyla-otlp-<id>-<pid>-<n>` for OTLP.

Additionally, TCP endpoints are proxied for prometheus metrics and profiling. The usual Alloy endpoint works as usual.

Metrics example:

```
curl http://localhost:12345/api/v0/component/beyla.ebpf.default/metrics
```

Profiling example:

```
curl http://localhost:12345/api/v0/component/beyla.ebpf.default/debug/pprof/
``` 

The recommended config is `output {...}` as the OTLP pipeline is leaner; the exported targets + prometheus.scrape path remains for users who already have a Prometheus-scrape pipeline.

Subprocess metrics feature the `subprocess` label:

```
alloy_resources_process_cpu_seconds_total{component_path="/",component_id="beyla.ebpf.a",subprocess="beyla"} 12.47
alloy_resources_process_cpu_seconds_total{component_path="/",component_id="beyla.ebpf.b",subprocess="beyla"} 8.10

alloy_resources_process_resident_memory_bytes{component_path="/",component_id="beyla.ebpf.a",subprocess="beyla"} 9.8643968e+07
alloy_resources_process_resident_memory_bytes{component_path="/",component_id="beyla.ebpf.b",subprocess="beyla"} 7.5497472e+07
```

The component enforces that the subprocess is terminated when the parent Alloy process terminates. It also implements a watchdog that is able to restart the subprocess if it hangs or becomes unhealthy, using a backoff mechanism to avoid churning.

Another noteworthy improvement brought by this PR is the handling of the configuration code: during the build process, Alloy downloads the Beyla configuration scheme, and auto-generates a lot of the configuration boilerplate that traditionally had to be hand-crafted. [Check out the generator README file for further information](https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/README.md).

Testing has been primarily, but not limited to, done with dogfooding. This approach has been succesfully running inside Grafana's clusters as well as in the notel demo app for several weeks. No remarkable changes to resource usage or bugs have been observed.

With regards to security, the updated documentation documents the capabilities process, but long story short: nothing changes in terms of permission (it has actually become possible to grant permissions only to the subprocess, but that has trade-offs and is therefore not considered "official". There are no regressions in terms of functionality).

### PR Checklist

- [x] Documentation added
- [x] Tests updated


mariomac:
This is too amazing!! Very excited. I will take some time to review this big PR

internal/component/beyla/ebpf/internal/config/gen/download.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	for _, p := range paths {
+		if _, err := os.Stat(p); err != nil {
+			return false
+		}
+	}
+	return true
+}
+
+func downloadBinary(url, destPath string) error {
```
skl: This should probably require checksum validation at least.

internal/component/beyla/ebpf/watchdog_linux.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// "beyla" is a placeholder, DialContext below ignores it and
+	// dials the abstract unix socket
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://beyla/healthz", nil)
+
+	if err != nil {
+		return err
+	}
+
+	client := &http.Client{
```
skl: This creates a new client every 10s, maybe:

1. reuse the client, or:
2. `defer client.CloseIdleConnections()`

internal/component/beyla/ebpf/internal/config/gen/download.go
rafaelroquetto: Done

internal/component/beyla/ebpf/watchdog_linux.go
rafaelroquetto: fixed

internal/component/beyla/ebpf/internal/config/gen/download.go
```diff
@@ -0,0 +1,129 @@
...
+	resp, err := http.Get(url) //nolint:noctx
+	if err != nil {
+		return fmt.Errorf("GET %s: %w", url, err)
+	}
+	defer resp.Body.Close()
+
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

internal/component/beyla/ebpf/otlp_linux.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		c.opts.Logger.Error("failed to read OTLP request body", "err", err)
+		http.Error(w, "failed to read request", http.StatusBadRequest)
+		return
+	}
+
+	defer r.Body.Close()
+
+	item := otlpItem{isTraces: isTraces, body: body, contentType: r.Header.Get("Content-Type")}
+
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

Makefile
```diff
@@ -85,6 +86,7 @@
 ##   GOEXPERIMENT         Used to enable Go features behind feature flags.
 ##   SKIP_UI_BUILD        Set to 1 to skip the UI build (assumes UI assets already exist).
 ##   SKIP_CODE_GENERATION Set to 1 to skip code generation before building the alloy binary
+##   BEYLA_VERSION        Version of Beyla to download and embed (default v3.9.3).
```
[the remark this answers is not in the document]
rafaelroquetto: removed the explicit version, it's just a maintenance burden

internal/component/beyla/ebpf/watchdog_linux.go
```diff
@@ -0,0 +1,112 @@
...
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				return (&net.Dialer{}).DialContext(ctx, "unix", addr)
+			},
+		},
+	}
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

internal/component/beyla/ebpf/internal/config/gen/download.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+)
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

internal/alloycli/resources_collector.go
```diff
@@ -33,7 +33,7 @@ func newResourcesCollector(l *slog.Logger) *resourcesCollector {
 
 		processStartTime: prometheus.NewDesc(
 			"alloy_resources_process_start_time_seconds",
-			"Start time of the process since Unix epoch in seconds.",
+			"Start time of the process since unix epoch in seconds.",
 			nil, nil,
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

internal/component/beyla/ebpf/watchdog_linux.go
```diff
@@ -0,0 +1,112 @@
...
+		return fmt.Errorf("subprocess not started")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// "beyla" is a placeholder, DialContext below ignores it and
+	// dials the abstract unix socket
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://beyla/healthz", nil)
+
```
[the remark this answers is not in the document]
rafaelroquetto: fixed

skl (dismissed):
lgtm! 🚀

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
 The component exposes metrics that can be collected by a Prometheus scrape component, and traces that can be forwarded to an OTel exporter component.
 
-{{< admonition type="note" >}}
-To run this component, {{< param "PRODUCT_NAME" >}} requires administrative privileges, or at least it needs to be granted the following capabilities: `BPF`, `SYS_PTRACE`, `NET_RAW`, `CAP_CHECKPOINT_RESTORE`, `DAC_READ_SEARCH`, and `PERFMON`.
-The number of required capabilities depends on the specific use case.
-Refer to the [Beyla capabilities](https://grafana.com/docs/beyla/latest/security/#list-of-capabilities-required-by-beyla) for more information.
+## Permissions
+
+`beyla.ebpf` uses eBPF, which requires elevated privileges.
+{{< param "PRODUCT_NAME" >}} spawns Beyla as a child process and transfers the required capabilities to it via the kernel's inheritable and ambient capability sets, so no `SETPCAP` is required.
```
clayton-cornell:
```suggestion
`beyla.ebpf` uses eBPF and needs elevated privileges.
{{< param "PRODUCT_NAME" >}} spawns Beyla as a child process and transfers the required capabilities through the kernel's inheritable and ambient capability sets.
You don't need `SETPCAP`.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+
+The required capabilities are: `BPF`, `NET_ADMIN`, `NET_RAW`, `PERFMON`, `DAC_READ_SEARCH`, `SYS_PTRACE`, `CHECKPOINT_RESTORE`, `SYS_RESOURCE` (kernels earlier than 5.11), and `SYS_ADMIN` (only for library-level instrumentation).
```
clayton-cornell:
```suggestion
The required capabilities are: `BPF`, `NET_ADMIN`, `NET_RAW`, `PERFMON`, `DAC_READ_SEARCH`, `SYS_PTRACE`, `CHECKPOINT_RESTORE`, `SYS_RESOURCE` on kernels earlier than 5.11, and `SYS_ADMIN` for library-level instrumentation only.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+The exact set needed depends on your use case; refer to [Beyla capabilities][] for more information.
```
clayton-cornell:
```suggestion
The exact set depends on your use case.
Refer to [Beyla capabilities][] for more information.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+
+In Kubernetes, you must also set `hostPID: true` in the Pod spec and configure an [Unconfined AppArmor profile][].
```
clayton-cornell:
```suggestion
In Kubernetes, set `hostPID: true` in the Pod spec and configure an [Unconfined AppArmor profile][].
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+
+### Standalone: root
+
+Run {{< param "PRODUCT_NAME" >}} as root. On a standard Linux system, root processes inherit all capabilities from the bounding set, so no additional configuration is required.
+If the bounding set is restricted (for example, by a systemd unit), grant the required capabilities explicitly:
```
clayton-cornell:
```suggestion
Run {{< param "PRODUCT_NAME" >}} as root.
On a standard Linux system, root processes inherit all capabilities from the bounding set, so you don't need extra configuration.
If systemd or another tool restricts the bounding set, grant the required capabilities explicitly:
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
-In Kubernetes environments, the [AppArmor profile must be `Unconfined`](https://kubernetes.io/docs/tutorials/security/apparmor/#securing-a-pod) for the Deployment or DaemonSet running {{< param "PRODUCT_NAME" >}}.
-You must also set the `hostPID` flag to `true` in the Pod spec so that in can access all the processes running on the host.
+```bash
+setcap 'cap_bpf,cap_net_admin,cap_net_raw,cap_perfmon,cap_dac_read_search,cap_sys_ptrace,cap_checkpoint_restore,cap_sys_resource,cap_sys_admin+ep' /path/to/alloy
+```
+
+### Standalone: non-root
+
+Set file capabilities on the {{< param "PRODUCT_NAME" >}} binary using the `+ip` flag.
+This seeds the permitted set without granting the capabilities to {{< param "PRODUCT_NAME" >}}'s own effective set, so {{< param "PRODUCT_NAME" >}} holds them only to transfer to Beyla:
```
clayton-cornell:
```suggestion
Set file capabilities on the {{< param "PRODUCT_NAME" >}} binary with the `+ip` flag.
This seeds the permitted set but doesn't grant effective capabilities to {{< param "PRODUCT_NAME" >}}, so {{< param "PRODUCT_NAME" >}} holds them only to pass to Beyla:
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+
+```bash
+setcap 'cap_bpf,cap_net_admin,cap_net_raw,cap_perfmon,cap_dac_read_search,cap_sys_ptrace,cap_checkpoint_restore,cap_sys_resource,cap_sys_admin+ip' /path/to/alloy
+```
+
+{{< admonition type="note" >}}
+File capabilities are not scoped to a container boundary and travel with the binary. Treat this as a deliberate security decision.
```
clayton-cornell:
```suggestion
File capabilities aren't scoped to a container boundary and travel with the binary.
Treat this as a deliberate security decision.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
 {{< /admonition >}}
 
+### Kubernetes: privileged
+
+Set `privileged: true` in the container's `securityContext`. This grants all capabilities and disables `seccomp` and AppArmor profiles. This approach is **not recommended** for production environments.
```
clayton-cornell:
```suggestion
Set `privileged: true` in the container's `securityContext`.
This grants all capabilities and disables `seccomp` and AppArmor profiles.
This approach is **not recommended** for production environments.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,102 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+            - CHECKPOINT_RESTORE
+            - SYS_RESOURCE  # kernels < 5.11
+            - SYS_ADMIN  # only for library-level instrumentation
+```
+
+Unlike `privileged: true`, this keeps `seccomp` and AppArmor profiles active.
+
+### Kubernetes: unprivileged, non-root user
+
+For the most restrictive posture, run as a non-root UID. Add `setcap +ip` to your container image (the official {{< param "PRODUCT_NAME" >}} image already includes this):
```
clayton-cornell:
```suggestion
For the most restrictive posture, run as a non-root UID.
Add `setcap +ip` to the binary when you build a custom image:
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -820,59 +837,54 @@ The matcher tags can be in the `:name` or `{name}` format.
 
 ### `injector`
 
-The `injector` block configures Beyla SDK injection feature used with the Beyla `k8s-injection-controller`.
+The `injector` block configures Beyla's SDK injection feature, which automatically instruments services by injecting OpenTelemetry SDKs without requiring eBPF.
```
clayton-cornell:
```suggestion
The `injector` block configures the Beyla SDK injection feature, which automatically instruments services by injecting OpenTelemetry SDKs without requiring eBPF.
```

Don't use possessive for product names.

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
 
 `beyla.ebpf` doesn't expose any component-specific debug information.
 
+## Observability considerations
+
+`beyla.ebpf` runs Beyla as a separate child process. This isolates failures in Beyla from {{< param "PRODUCT_NAME" >}} but changes how its resource usage and profile data are exposed.
```
clayton-cornell:
```suggestion
`beyla.ebpf` runs Beyla as a separate child process.
This isolates failures in Beyla from {{< param "PRODUCT_NAME" >}} but changes how {{< param "PRODUCT_NAME" >}} exposes resource usage and profile data.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+
+### Resource metrics
+
+{{< param "PRODUCT_NAME" >}} exposes process-level metrics for the Beyla process under the same name as its own, distinguished by a `subprocess="beyla"` label:
```
clayton-cornell:
```suggestion
{{< param "PRODUCT_NAME" >}} exposes process-level metrics for the Beyla process under the same name as its own, with a `subprocess="beyla"` label:
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+
+```promql
+alloy_resources_process_resident_memory_bytes                    # {{< param "PRODUCT_NAME" >}} process only
+alloy_resources_process_resident_memory_bytes{subprocess="beyla"} # Beyla subprocess only
+```
+
+Dashboards or alerts that previously used `alloy_resources_process_*` to approximate total container resource usage now see only {{< param "PRODUCT_NAME" >}}'s share.
```
clayton-cornell:
```suggestion
If you use `alloy_resources_process_*` to approximate total container resource usage, the metrics include only the {{< param "PRODUCT_NAME" >}} process.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+```promql
+sum without(subprocess) (alloy_resources_process_resident_memory_bytes)
+```
+
+For container-level limits and OOM monitoring, prefer `kubelet`/cAdvisor metrics such as `container_memory_working_set_bytes`, which already account for every process in the container.
+
+### Profiling
+
+{{< param "PRODUCT_NAME" >}}'s `/debug/pprof/*` endpoints reflect only {{< param "PRODUCT_NAME" >}}'s own goroutines, allocations, and CPU.
+Profile data for the Beyla process is exposed by Beyla on its own HTTP port and is reachable through the component's reverse-proxy URL, for example:
```
clayton-cornell:
```suggestion
The {{< param "PRODUCT_NAME" >}} `/debug/pprof/*` endpoints reflect goroutines, allocations, and CPU from the {{< param "PRODUCT_NAME" >}} process only.
Beyla exposes profile data on its own HTTP port.
You can reach it through the component's reverse-proxy URL, for example:
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+
+```
+<alloy>/api/v0/component/beyla.ebpf.<LABEL>/debug/pprof/heap
+```
+
+The Beyla process only exposes pprof endpoints when {{< param "PRODUCT_NAME" >}} itself has profiling enabled.
+The toggle is the existing `--server.http.enable-pprof` flag (default `true`); when set to `false`, the proxied URL above also returns 404, matching {{< param "PRODUCT_NAME" >}}'s own behavior.
+There is no separate Beyla-specific switch.
```
clayton-cornell:
```suggestion
When you enable profiling for {{< param "PRODUCT_NAME" >}}, Beyla also exposes pprof endpoints.
The toggle is the `--server.http.enable-pprof` flag, which defaults to `true`.
If you set it to `false`, the proxied URL returns 404, which matches how the {{< param "PRODUCT_NAME" >}} `/debug/pprof/*` endpoints behave.
Beyla doesn't have a separate switch.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+
+Each `beyla.ebpf` component instance has its own pprof URL scoped by component ID.
+With multiple instances configured (for example, `beyla.ebpf.foo` and `beyla.ebpf.bar`), each is reachable at its own component path with no extra wiring.
```
clayton-cornell:
```suggestion
If you define multiple instances, such as `beyla.ebpf.foo` and `beyla.ebpf.bar`, each one is reachable at its own component path without extra configuration.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -910,6 +922,47 @@ The exported targets use the configured [in-memory traffic][] address specified
...
+
+Continuous-profiling pipelines that previously scraped {{< param "PRODUCT_NAME" >}}'s pprof endpoints to capture Beyla profile data must now also scrape the component's proxied pprof URL.
+The existing scrape job continues to work but only captures {{< param "PRODUCT_NAME" >}}'s own profile.
```
clayton-cornell:
```suggestion
To capture Beyla profile data in continuous-profiling pipelines, scrape the component's proxied pprof URL in addition to the {{< param "PRODUCT_NAME" >}} pprof endpoints.
A scrape job that targets only the {{< param "PRODUCT_NAME" >}} pprof endpoints captures only the {{< param "PRODUCT_NAME" >}} profile.
```

clayton-cornell:
Docs are good as-is now :-)

clayton-cornell (dismissed):
Approving for docs.

rafaelroquetto:
ping

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -86,15 +180,8 @@ You can use the following blocks with `beyla.ebpf`:
...
-| `ebpf` > `payload_extraction` > `http` > `genai` > [`mcp`][mcp payload extraction] | Configures **Model Context Protocol** (MCP) payload extraction and parsing.                                             | no       |
-| `ebpf` > `payload_extraction` > `http` > `genai` > [`embedding`][embedding payload extraction] | Configures generic **embedding** provider (**Voyage AI**, **Cohere**, **Jina AI**) payload extraction and parsing.                                                      | no       |
-| `ebpf` > `payload_extraction` > `http` > `genai` > [`rerank`][rerank payload extraction] | Configures **Rerank** (**Cohere**, **Jina AI**, **Voyage AI**, etc.) payload extraction and parsing.                                             | no       |
-| `ebpf` > `payload_extraction` > `http` > `genai` > [`retrieval`][retrieval payload extraction] | Configures vector **retrieval** (**Pinecone**, **Qdrant**, **Milvus**, **Chroma**, **Weaviate**, etc.) payload extraction and parsing.                                             | no       |
```
ptodev, line 97: Why are those options deleted? Is it because of a Beyla update?

ptodev (commented):
This is very large PR... is there any way to split it into smaller parts? E.g. the schema part could be a smaller PR which helps validate even the current (non-embedded) version of Beyla?

internal/component/beyla/ebpf/internal/config/args.go
```diff
@@ -0,0 +1,201 @@
...
+	MaxBucketNumber  int           `alloy:"max_bucket_number,attr,optional"`
+	MinResetDuration time.Duration `alloy:"min_reset_duration,attr,optional"`
+}
+
+type Traces struct {
+	Instrumentations []string      `alloy:"instrumentations,attr,optional"`
+	Sampler          SamplerConfig `alloy:"sampler,block,optional"`
+}
+
+type Network struct {
```
kgeckhart, line 147: Looks like some fields might be missing here like `geo_ip` and `reverse_dns` edit: unless we can guarantee no one is using them we'll need them to stay so someone who upgrades is not broken.

internal/component/beyla/ebpf/exec_linux.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func findFreePort() (int, error) {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return 0, err
+	}
+	defer listener.Close()
+	return listener.Addr().(*net.TCPAddr).Port, nil
+}
```
kgeckhart, line 135: This ends up being a race because the port is released before your process can bind it and in that time something else could have bound the port. 

Ideally, the listeners you use can accept a zero and treat that as "find an open port" instead.

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -1147,379 +303,174 @@ func getLatestArgsFromChannel[A any](ch chan A, current A) A {
...
+		c.opts.Logger.Error("Beyla subprocess exceeded maximum restart attempts, giving up", "max_restarts", subprocess.MaxRestarts)
+		return
 	}
 
-	if _, err := glob.Compile(s); err != nil {
-		return services.GlobAttr{}, fmt.Errorf("invalid glob pattern %q: %w", s, err)
-	}
-	return services.NewGlob(s), nil
+	c.opts.Logger.Info("scheduling subprocess restart", "backoff", backoff, "restart_count", count)
+	c.restartTimer.Reset(backoff)
```
kgeckhart, line 450: I think this might need some synchronization. The various triggers for happen in go routines and nothing prevents a config update from firing at the same time which turns in to a data race.

kgeckhart (commented):
Still need more time to work through it

internal/component/beyla/ebpf/internal/config/gen/schema.json
ptodev, line 1: Should this file live in the Beyla repo? Alloy can just import it?

internal/component/beyla/ebpf/internal/config/gen/sample_config.alloy
ptodev, line 1: Is this file used anywhere? Do we need to generate it and store it in the repo? Can it be gitignored or not generated?

ptodev (commented):
I added a few comments regarding reducing the size of the PR.

What will be the long term strategy with making Beyla an OTel-native component? Ideally one day we can let users run it in the OTel Engine as an OCB-able component. Would that also work as a subprocess? Do we expect it to be a subprocess in the upstream OTel repo, and then we can import it from there in the future?

Makefile
```diff
@@ -108,6 +110,18 @@ CGO_ENABLED          		?= 1
 RELEASE_BUILD        		?= 0
 GOEXPERIMENT         		?= $(shell go env GOEXPERIMENT)
 
+# Beyla embedding configuration
+BEYLA_VERSION        ?= v3.25.0
```
marctc, line 114: The docs page uses `{{< param "BEYLA_VERSION" >}}` pulled from `docs/sources/_index.md.t:11`. We probably should unify these values

internal/component/beyla/ebpf/internal/config/args.go
rafaelroquetto: ah good catch - these got accidentally dropped, I've readded them, thanks!

internal/component/beyla/ebpf/exec_linux.go
rafaelroquetto: yeah that's indeed unfortunate - the problem here is that we aren't the ones binding the port - OBI/Beyla is - all we are doing is guessing a port that we can pass to down to Beyla and say "hey, bind your exporter to this port" - and then we reverse-proxy it.

In the unlikely event of a race, Beyla will fail, the process will restart and try again at a different port, so it should be pretty harmless - albeit not ideal.

I suppose a more proper would be to change OBI/Beyla to accept file descriptors instead - this way, we could do what you are proposing and bind/hold the port, then pass down the fd to Beyla - no race. I can work on that, but that'd be best handled as a separate/follow up PR IMHO. For now, I believe this should be acceptable (but feel free to disagree).

internal/component/beyla/ebpf/beyla_linux.go
rafaelroquetto: fixed

docs/sources/reference/components/beyla/beyla.ebpf.md
rafaelroquetto: Ah, that section got reworked (the river config layout changed to be more in line with the Beyla config) and the docs were never updated. The features are still there, I've updated the docs now.

internal/component/beyla/ebpf/internal/config/gen/schema.json
rafaelroquetto: So, this actually does live in the Beyla repo at grafana/beyla/docs/config-schema.json . `make download-beyla-schema` pulls it pinned to `BEYLA_VERSION` (same version as the embedded binary, so they can't drift). We can't Go-import it since the subprocess model drops the Beyla module dependency, so it's fetched as an artifact. We commit the pinned snapshot so codegen is reproducible offline and CI can verify the generated code against a known schema version, just like it's usually done when checking in a generated-code input like a proto/OpenAPI spec.

internal/component/beyla/ebpf/internal/config/gen/sample_config.alloy
rafaelroquetto: This is just used as a guide - sample for people looking at what the current alloy config looks like. It is not used anywhere and it could be git ignored - I'd recommend keeping it committed for 2 reasons:
 - it allows people to go to https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/sample_config.alloy and use it as a reference
 - it is useful when doing a `git diff` and `git add -p` for when people want to know what really changed in the config.

rafaelroquetto:
@ptodev unfortunately I think it is difficult to split this PR - think of it as a brand new component rather than an evolution of the old beyla.ebpf one. The fact that we now have a subprocess requires that we write a YAML file, which needs to be generated. I could have written all the code manually that translates from Alloy's arguments into a YAML structure, but that would have been error-prone and wouldn't help with speeding the release cadence - hence the config generation based on the JSON schema. So things are somewhat tightly coupled here, sorry.

rafaelroquetto:
> What will be the long term strategy with making Beyla an OTel-native component? Ideally one day we can let users run it in the OTel Engine as an OCB-able component. Would that also work as a subprocess? Do we expect it to be a subprocess in the upstream OTel repo, and then we can import it from there in the future?

@ptodev I am honestly not sure, this is a bigger discussion that is slowly taking place within OBI itself. subprocess may be a consideration, but I can't tell at this time. The coralogix guys are the ones driving it afaik, and there is currently a discussion between OBI and the upstream OCB folks about how to deal with committed binaries... so there's nothing concrete yet in that direction.

Makefile
rafaelroquetto: fixed

rafaelroquetto:
@ptodev @kgeckhart @grafana/beyla - I have raised an alternative PR that is simpler - see https://github.com/grafana/alloy/pull/6695 and please read the PR description as it explains the trade-offs involved. If we prefer to go down that route, let me know so I can run some further tests on that PR and dump this one.

rafaelroquetto:
Superseded by https://github.com/grafana/alloy/pull/6695

Issue #2524, opened 2025-01-24
Add config option equivalent to --disable-reporting


dmitrydonskih:
### Request

Add configuration option to disable sending stat reports to `stats.grafana.org`.

### Use case

When Alloy runs in closed network without access to internet, sending stats reports fails, it floods own logs with errors and therefore should be disabled.
Currently the only way to disable it is to issue command-line option `--disable-reporting`.
On linux, this requires modification of systemd service.
On windows, installed service runs different executable (`alloy-service-windows-amd64.exe`), so modifying command-line options is not possible (at least not documented).
We definitely need a more convenient way to control this feature.

mzurborg-scopevisio:
Please prioritize this.

We want to roll-out alloy to our customers who use parts of our software in their local networks on Windows-based computers in order to be able to collect metrics and logs and we cannot seriously (due to strict data protection agreements with our customers) do this unless this analytics can be easily disabled via config file.

As @dmitrydonskih stated above, easily disabling reporting for the windows service installer is currently neither supported nor documented.

dmitrydonskih:
Please prioritize and assign!
This should only take a couple lines of code.
At least several users cannot use Alloy in their production environments due to this nasty issue.

edufur:
please do this! clearly there was an intent to allow us to turn off collection. but we are unable to do that for the service execution

TomKeur:
We're facing the same issue, but we're not running alloy with `systemd`, but in a container with the default command at the moment.
Since we are not using a custom command, we need to change our deployments, and I'm kind of afraid this will break something in the future, because of "custom".

Maybe it would also be an idea to add an environment variable to be able to disable it.
For example example at NextJS you have: `NEXT_TELEMETRY_DISABLED=1`.

sedmaste:
Hi, we are facing very similar issue with production enviroment - as we have installed alloy using helm chart - is it possible to disable reporting via _values.yaml_ file or so ?

michaelkebe:
For those installed on Debian/Ubuntu with the `.deb` file (mayby `.rpm`, too).

A systemd service `alloy.service` is installed with it. In the service description there are these lines:

```
...
EnvironmentFile=/etc/default/alloy
...
ExecStart=/usr/bin/alloy run $CUSTOM_ARGS --storage.path=/var/lib/alloy/data $CONFIG_FILE
...
```

This means we can add `CUSTOM_ARGS` via the `/etc/default/alloy`.

[...]

PR #6478, opened 2026-06-10
fix: Back off usage reporting on persistent failure instead of retrying every minute

### LINKED ISSUE #6474: Usage reporting retries every ~3s with no backoff when stats.grafana.org fails to resolve, causing a fleet-wide DNS query storm
### What's wrong?

Alloy's anonymous usage reporting is enabled by default and periodically contacts `stats.grafana.org`. When that hostname **fails to resolve** — for example it is `NXDOMAIN`'d by a DNS blocklist (it is independently present on common lists such as Hagezi/OISD), or sinkholed in a closed network — Alloy appears to retry on a very short, fixed interval (**~1 attempt every ~3 seconds, per agent**) with no exponential backoff or give-up.

On a single host this is just log noise, but across a fleet it becomes a **DNS query storm** against the local recursive resolver.

This is **distinct from #2524**, which asks for a config-file toggle to disable reporting. Even when an operator leaves reporting enabled, the *failure handling itself* is too aggressive: a single unreachable telemetry endpoint should never generate hundreds of thousands of DNS queries per day. (Also related to the hardening-docs discussion in #5878.)

### Steps to reproduce

1. Run Alloy with default settings (usage reporting on).
2. Make `stats.grafana.org` fail to resolve — e.g. add it to a DNS blocklist so the resolver returns `NXDOMAIN` (Hagezi/OISD already list it), or block it on the resolver in an air-gapped network.
3. Watch the resolver's query log and the agent's outbound DNS.

### Observed behavior

- ~1 resolution attempt for `stats.grafana.org` **every ~3 seconds, per agent, indefinitely** — no backoff.
- On a **~57-agent fleet**: roughly **900k DNS queries/day** for this single name; the recursive resolver's query log grew from ~20 MB/day to ~147 MB/day.
- Workaround applied fleet-wide: `--disable-reporting` (Debian: `CUSTOM_ARGS` in `/etc/default/alloy`; Alpine: `command_args` in `/etc/conf.d/alloy`).

### Expected behavior

When the usage-reporting endpoint repeatedly fails (DNS or transport), the reporter should **back off exponentially with a cap**, or treat a persistent resolution failure as "reporting unavailable" and stop hammering. The failure mode should degrade gracefully rather than amplify into a per-host tight loop.

### System information

Linux (Debian 12/13 and Alpine), fleet of ~57 agents installed from the Grafana `apt`/`apk` repositories. Local recursive resolver (Technitium) with a public DNS blocklist that independently lists `stats.grafana.org`.

### Software version

Observed with current stable Alloy as of June 2026 (v1.16.x). The behavior is in the usage-stats reporter and looks independent of a specific release.

### Configuration

Default configuration (usage reporting left at its default of enabled); a `loki.source.journal` → `loki.relabel` → `loki.write` logs pipeline. The reporting traffic is unrelated to the configured components.

### Logs

The agent side surfaces repeated reporting failures; the resolver side shows the `stats.grafana.org` lookups recurring at the ~3s cadence described above. (Happy to attach a redacted resolver query-log sample if useful.)

kgeckhart:

When the usage-reporting endpoint persistently fails — e.g. `stats.grafana.org` is `NXDOMAIN`'d by a DNS blocklist or sinkholed in a closed network — the reporter re-attempted on every check tick (~1/minute) indefinitely, because `lastReport`/`next` only advanced on success. Across a fleet that turns one unreachable endpoint into a DNS query storm (the issue reports ~900k queries/day from 57 agents). `reportUsage` already retries with dskit backoff (5 attempts, up to 30s); this advances the schedule whether or not the report succeeded, so a spent window waits a full interval (4 hours) instead of looping every 1 minute.

Fixes #6474

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/usagestats/reporter.go
+++ b/internal/usagestats/reporter.go
@@ -15,6 +15,12 @@ import (
 var (
 	reportCheckInterval = time.Minute
 	reportInterval      = 4 * time.Hour
+
+	reportBackoffConfig = backoff.Config{
+		MinBackoff: time.Second,
+		MaxBackoff: 30 * time.Second,
+		MaxRetries: 5,
+	}
 )
 
 // Reporter holds the Alloy seed information and sends report of usage
@@ -56,10 +62,15 @@ func (rep *Reporter) Start(ctx context.Context, metricsFunc func() map[string]an
 				continue
 			}
 			rep.logger.Info("reporting Alloy stats", "date", time.Now())
-			if err := rep.reportUsage(ctx, next, metricsFunc()); err != nil {
+			err := rep.reportUsage(ctx, next, metricsFunc())
+			if err != nil {
 				rep.logger.Warn("failed to report usage", "err", err)
-				continue
 			}
+			// Advance the schedule whether or not the report succeeded.
+			// reportUsage already retries with backoff; on persistent failure
+			// (e.g. the endpoint won't resolve) we treat the window as spent and
+			// wait a full interval rather than re-attempting on every tick, which
+			// would otherwise hammer the endpoint indefinitely.
 			rep.lastReport = next
 			next = next.Add(reportInterval)
 		case <-ctx.Done():
@@ -70,11 +81,7 @@ func (rep *Reporter) Start(ctx context.Context, metricsFunc func() map[string]an
 
 // reportUsage reports the usage to grafana.com.
 func (rep *Reporter) reportUsage(ctx context.Context, interval time.Time, metrics map[string]any) error {
-	backoff := backoff.New(ctx, backoff.Config{
-		MinBackoff: time.Second,
-		MaxBackoff: 30 * time.Second,
-		MaxRetries: 5,
-	})
+	backoff := backoff.New(ctx, reportBackoffConfig)
 	var errs multierror.MultiError
 	for backoff.Ongoing() {
 		if err := sendReport(ctx, rep.seed, interval, metrics); err != nil {
--- a/internal/usagestats/reporter_test.go
+++ b/internal/usagestats/reporter_test.go
@@ -8,8 +8,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/dskit/backoff"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


internal/usagestats/reporter.go
```diff
@@ -56,10 +62,15 @@ func (rep *Reporter) Start(ctx context.Context, metricsFunc func() map[string]an
 				continue
 			}
 			rep.logger.Info("reporting Alloy stats", "date", time.Now())
-			if err := rep.reportUsage(ctx, next, metricsFunc()); err != nil {
+			err := rep.reportUsage(ctx, next, metricsFunc())
+			if err != nil {
 				rep.logger.Warn("failed to report usage", "err", err)
-				continue
```
kalleep, line 61: sneaky

kalleep (approved):

ferr079:
Thanks for turning this around so fast @kgeckhart — the fix reads correct to me and matches what I see on the fleet.

**On the "~3s" in my report vs the "~1/minute" tick — they compose, and your fix targets the right layer.** The outer `Start` loop re-enters every `reportCheckInterval` (1 min) and, because `lastReport`/`next` only advanced on success, it re-ran `reportUsage` on every tick. Each `reportUsage` then spends its full dskit budget — 5 attempts at ~1s/2s/4s/8s/16s — so each minute contains a ~30s burst of ~5 lookups. Eyeballed at the resolver that reads as "a query every few seconds, indefinitely." Advancing the schedule unconditionally turns a spent window into **5 attempts per 4h** instead of **5 attempts per minute** — a ~240× cut per agent, which is exactly the fleet-wide drop the issue is about.

The test captures the invariant cleanly (exactly-once contact over a window longer than the run, independent of host speed).

This is also nicely orthogonal to #2524: even when an operator leaves reporting **on**, the failure handling is what was too aggressive — so fixing the loop is the right primitive regardless of whether a disable toggle ever lands.

Happy to validate the built binary against the original repro — 57-agent fleet with `stats.grafana.org` NXDOMAIN'd by a local resolver (Hagezi blocklist) — and confirm the resolver query volume collapses. Point me at a build or I'll just compile the branch. Thanks again! 🙏

PR #6486, opened 2026-06-12
ci: Auto freeze and unfreeze PRs

jharvey10:
This addresses two gaps:
- PRs that were all green, but unmerged prior to RC mode starting could still be merged in during code freeze. This is because they have a stale green check that needs to be dismissed.
- Upon release publication, PRs with a stale red code freeze check have no lightweight way to unfreeze except merging `main` into their PR, which is significantly more expensive than toggling a label.


jharvey10:
This is not currently enough of an issue to warrant introducing this new workflow.

Issue #6495, opened 2026-06-15
Support Formatter Specification from treefmt


Scrumplex:
### Background

I want to use [treefmt](https://treefmt.com/latest/) which helps manage many different formatters for monorepos. Alloy currently does not support its specification for how a formatter should behave, as it must only write to the target file, if it has actually been changed by the formatter.

### Proposal

Only write formatted output to the file(s) if they actually changed.

See https://treefmt.com/latest/reference/formatter-spec/

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

vishalmore90:
Approach Taken:

1. Change logic in format: In internal/alloycli/cmd_fmt.go, we read the original bytes (bb) and compared them to the final formatted bytes (buf.Bytes()) using bytes.Equal().
2. Conditional writing: If bytes.Equal(bb, buf.Bytes()) is true and the -w/--write flag is active, we immediately return nil without opening or truncating the file.
3. Optimized comparison: We replaced reflect.DeepEqual with bytes.Equal for faster and more idiomatic slice comparison, allowing us to drop the unused "reflect" import.
4. Unit testing: We added a new test suite in internal/alloycli/cmd_fmt_test.go including a specialized test (TestAlloyFmt_NoWriteIfUnchanged) that flags the target file as read-only (0400) before running Run(). If alloyFmt tries to open the file for writing, it fails due to the read-only permissions, confirming that it correctly avoids modifying already-formatted files.

thampiotr:
Closing as per https://github.com/grafana/alloy/blob/main/docs/developer/genai.md, if this is mistake, please reopen and let us know.

Scrumplex:
I don't use GenAI/LLM for any of my open source contributions, including issues. Please reopen.

Scrumplex:
I am not able to reopen the issue, as it was closed by a project collaborator. @thampiotr please reopen this

Scrumplex:
Reminder to please reopen this issue.

thampiotr:
Hi @Scrumplex thanks for clarification. The comment that raised a concern wasn't from you. Apologies for misunderstanding, let me take a look at your proposal.

thampiotr:
Since this is not a change to user-facing interface and relatively minor ask, I don't think this needs to be a proposal. Removing the proposal label and I have a PR open ready for review.

thampiotr:
[...]

Issue #2936, opened 2025-03-09
Drop AWS SDK for Go v1 and upgrade to v2 before EOL/end-of-support at July 31, 2025 / fix CVE-2020-8911 + CVE-2020-8912 - Alloy


rgoltz:
### Request

Currently Alloy using GO dependencies and libraries, which are using/reference the **version 1 of the AWS SDK for Go** (github.com/aws/aws-sdk-go). AWS as maintainer of this AWS SDK for Go package announced the **EOL of the version 1 for July 31, 2025**. After this date, there are **no** critical bug fixes and **no** security issues provided. For details: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

It’s recommended to migrate to [AWS SDK for Go v2](https://aws.github.io/aws-sdk-go-v2/docs/) - https://github.com/aws/aws-sdk-go-v2/ 
Since Alloy is still indirectly using the EOL-version of AWS SDK for Go v1, it’s also affected be at least two security issues (which will never be fixed in version 1 codestream):

- CVE-2020-8911: A padding oracle vulnerability exists in the AWS S3 Crypto SDK for GoLang versions prior to V2. The SDK allows users to encrypt files with AES-CBC without computing a Message Authentication Code (MAC), which then allows an attacker who has write access to the target's S3 bucket.

- CVE-2020-8912: A vulnerability in the in-band key negotiation exists in the AWS S3 Crypto SDK for GoLang versions prior to V2. An attacker with write access to the targeted bucket can change the encryption algorithm of an object in the bucket, which can then allow them to change AES-GCM to AES-CTR.

### Use case

I’d like to start the joint discussion for a step-by-step investigation and upgrade to AWS SDK for Go v2 for Alloy. I’m aware that is not an easy step. Furthermore, some dependency already using v2. It’s more like a "progress-tracking and discussion-ticket". I assume in some cases, it’s a update of a dependency, in other cases, it might be necessary to address this v1 to v2 upgrade to other package maintainers - See lists below!

In order to get an overview of the "areas" of affected Go dependencies for , I’ve used [deptree](https://github.com/vc60er/deptree) to identify `github.com/aws/aws-sdk-go` references (in total 28x times based on main from 09.03.2025):

```
github.com/grafana/alloy (go1.23.5)
…
 ├── <529> github.com/aws/aws-sdk-go@v1.55.6 (go1.19)
[...]

PR #6499, opened 2026-06-16
feat(cli): Only write formatted files if content changed

vishalmore90:
## Description

This PR addresses issue #6495. Currently, `alloy fmt` truncates and writes the formatted bytes back to disk even if no formatting changes were made. This causes files to appear modified in orchestration systems like `treefmt`.
We now compare the parsed and formatted byte slice to the original file content using `bytes.Equal`. If they are identical, the file is not opened/written to.

## Changes
- Modified `internal/alloycli/cmd_fmt.go` to check if `bytes.Equal(bb, buf.Bytes())` is true before initiating file writes.
- Cleaned up unused `"reflect"` package import.
- Created `internal/alloycli/cmd_fmt_test.go` to test that already-formatted files are not touched, unformatted files are correctly formatted, and the `--test` flag behaves as expected. #2936 
 
## Verification
Executed specific unit tests successfully on Windows:
```powershell
go test -tags="nodocker" internal/alloycli/cmd_fmt.go internal/alloycli/cmd_fmt_test.go

### DIFF
--- a/internal/alloycli/cmd_fmt.go
+++ b/internal/alloycli/cmd_fmt.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"reflect"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


vishalmore90:
Hello @grafanabot 

Could you please take a final look when you have time? If everything looks good, I would appreciate an approval and merge.

Regards,
Vishal

PR #6500, opened 2026-06-16
ci: Cache the AccessControl NSIS plugin

ptodev:
AccessControl is used for changing file and folder permissions in NSIS. Alloy's NSIS script creates the Windows installer. AccessControl doesn't come pre-bundled with NSIS - it's a plugin which is downloaded separately. 

There are rare times when the [download fails](https://github.com/grafana/alloy/actions/runs/27411398458/job/81013283189). Caching the plugin binary will make the tests faster and reduce the chance of flaky failures.

In the future we can also consider #6502 to remove AccessControl entirely.

This PR relies on functionality which is not very clearly [documented](https://github.com/actions/cache/tree/cdf6c1fa76f9f475f3d7449005a359c84ca0f306/), but it seems to work. The first run of this branch showed a cache miss:

Screenshot 2026-06-16 at 14 21 44

The second run had a cache hit:
Screenshot 2026-06-16 at 15 46 15

### DIFF
--- a/.github/workflows/integration-test-windows-service.yml
+++ b/.github/workflows/integration-test-windows-service.yml
@@ -36,12 +36,25 @@ jobs:
         run: choco install nsis -y
         shell: pwsh
 
+      - name: Cache NSIS AccessControl plugin
+        id: cache-accesscontrol
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
+        with:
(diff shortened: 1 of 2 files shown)


ptodev:
I tested this with an incorrect hash and got an error, as expected:
Screenshot 2026-06-18 at 13 46 22

jharvey10 (approved):
Did some additional security vetting and couldn't find any issues. 🎉

PR #6501, opened 2026-06-16
 feat(collector): Add slim build that trims Alloy to a Prometheus+Loki distro (516MB → 165MB)

shalk:

  Produces a trimmed-down ("slim") build of Alloy that keeps only the components
  needed for a Prometheus-scrape + remote_write + node_exporter + Loki file-logs
  pipeline, shrinking the binary from ~516MB to ~165MB (-68%). Achieved by
  trimming the two component registries and gating the heavy `alloy convert`
  converters and legacy static-mode integrations behind a `slim` build tag.

  ### Pull Request Details

  The size reduction comes in two layers:

  **1. Component registry trim (always on in this distro)**
  - `internal/component/all/all.go`: reduced from ~180 blank imports to the 10
    components actually used — `prometheus.scrape`, `prometheus.remote_write`,
    `prometheus.relabel`, `prometheus.exporter.unix`, `prometheus.exporter.self`,
    `discovery.relabel`, `loki.source.file`, `loki.process`, `loki.write`,
    `local.file_match`.
  - `collector/builder-config.yaml`: removed all OTel collector business
    components, keeping only the `alloyengine` extension, the confmap providers,
    and `nop` receiver/exporter placeholders. Regenerated `main.go`,
    `components.go`, `go.mod`, `go.sum` via OCB.

  **2. `slim` build tag (opt-in, default build unchanged)**
  Following the existing `embedalloyui`/`boringcrypto` split-file pattern, three
  unconditional imports that statically pull in huge dependency trees are now
  gated behind `!slim`:
  - `flowcmd`: the `internal/static/integrations/install` registration
    (legacy static-mode integrations → vmware/azure/gcp exporters).
  - `internal/converter`: the `otelcolconvert` and `staticconvert` converters
    (pull in every OTel collector component → AWS/GCP/Azure/DataDog SDKs).

  Under `slim`, `alloy convert` for `otelcol`/`static` formats returns a clear
  "not supported in this slim build" diagnostic instead of being compiled in.

  Build command:
  GO_TAGS=slim SKIP_UI_BUILD=1 SKIP_CODE_GENERATION=1 RELEASE_BUILD=1 make alloy

  Dependency impact (full → slim package counts): DataDog 193→0, arrow-go 31→0,
  otel-contrib 317→47, total compiled packages 3992→2450. Residual AWS/Azure/GCP
  SDK packages remain only for the kept components' auth features
  (`prometheus.remote_write` sigv4/Azure-AD/GCP, `loki.write`).

  | Stage | Size |
  |---|---|
  | Baseline (debug, full) | 516 MB |
  | Registry trim + `-s -w` strip | 321 MB |
  | + `slim` tag (converter/static gated) | **165 MB** |

  ### Notes to the Reviewer

  - The default (no-tag) build is unchanged — both `go build` and
    `go build -tags slim` of `internal/converter` and `flowcmd` compile cleanly.
  - Smoke-tested: the slim binary runs a config exercising all 10 components with
    zero component-load errors; `alloy convert --source-format=otelcol` degrades
    gracefully with a critical diagnostic rather than crashing.
  - The registry trim in layer 1 is distro-specific (it removes components other
    users rely on). If this targets upstream rather than a custom distro, only the
    layer-2 `slim` build tag is intended to be general-purpose; the registry trim
    would need to be dropped or made tag-conditional.

  ### PR Checklist

  - [ ] Documentation added
  - [ ] Tests updated


thampiotr:
Hi! Thanks for the suggestion. I can see this part:

> internal/component/all/all.go: reduced from ~180 blank imports to the 10
components actually used — prometheus.scrape, prometheus.remote_write,
prometheus.relabel, prometheus.exporter.unix, prometheus.exporter.self,
discovery.relabel, loki.source.file, loki.process, loki.write,
local.file_match.

This is going to make this feature very narrowly useful - only for those who use these 10 components. What if someone uses one extra component that is not on the list? We don't want currently to add `slim-2` or something else that will orchestrate the list of components to include. If we do that in the future, it would be more similar to what `OCB` does for OpenTelemetry - with a YAML manifest and some code generation. But that will require a discussion through our [proposal process](https://github.com/grafana/alloy/blob/main/docs/design/908-proposal-process.md). So if you'd like to continue, I'd encourage opening a proposal that would make this feature more generic so it can apply to more use cases then users that use these sepcifc 10 components.

shalk:
> proposal process

Thank you for your reply.  I will try to opening a proposal after some simple design. 

I hope this will be a generic feature.

 Before that, I'll take some time to investigate OCB.

PR #6508, opened 2026-06-16
perf(loki.source.journal): Reduce allocations

kalleep:
### Pull Request Details
While working on https://github.com/grafana/alloy/pull/6459 I noticed that journal reader from from `github.com/coreos/go-systemd/v22/sdjournal` is not really a great fit and a bit akward to use.

An example of the awkwardness is that we emit entries from the [formatter function](https://github.com/grafana/alloy/blob/9e7da573b914acccf6cdda57ec32f4b593775708/internal/component/loki/source/journal/tailer.go#L257), and we do a lot of unnecessary work like calling journal functions for things we don't ever need, not reusing allocations etc.

This library do expose a wrapper for journal api but it will still allocate a new map for every [GetEntry](https://github.com/coreos/go-systemd/blob/main/sdjournal/journal.go#L751) call.

So I figured we could define the handful of functions we need and create our own wrapper around it and simplify the tailing part a lot. This wrapper just defines the handful of apis we need and only call the functions for data we actually use.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


kalleep:
Benchmark against main after https://github.com/grafana/alloy/pull/6459 was merged

```
goos: linux
goarch: amd64
pkg: github.com/grafana/alloy/internal/component/loki/source/journal
           │ bench_old.txt │         bench_new.txt          │
           │    sec/op     │   sec/op     vs base           │
Journal-16     1.003 ± ∞ ¹   1.003 ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

           │ bench_old.txt │          bench_new.txt           │
           │     B/op      │     B/op       vs base           │
Journal-16   53.25Mi ± ∞ ¹   23.93Mi ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

           │ bench_old.txt │          bench_new.txt          │
           │   allocs/op   │  allocs/op    vs base           │
Journal-16   1180.7k ± ∞ ¹   820.3k ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

internal/component/loki/source/journal/internal/sdjournal/sdjournal.go
```diff
@@ -0,0 +1,472 @@
...
+	Name  string
+	Value string
+}
+
+var ErrNoData = errors.New("no data")
+
+// Next returns a slice with fields from the next entry in the journal and the cursor associated with that entry.
+// The returned slice is borrowed and invalidated after the next call to Next.
+// If there are no more entries ErrNoData is returned.
+func (j *Journal) Next() ([]Field, string, error) {
```
kalleep, line 380: This is almost identical to what the library does when fetching entires with some differences:
1. The library always allocates a new map and store fields in that map, instead we can reuse the same slice and put our fields in there.
2. We avoid 2 string allocations by slicing the buffer we have already allocated

internal/component/loki/source/journal/internal/sdjournal/sdjournal.go
ptodev, line 1: I suppose there won't be much maintenance work with keeping this file up to date since the API is stable.

However, how would we detect the need to update it? Would we get an error at build time? Potentially we could add snippets like this:
```
// Fails to compile if our assumed signature drifts from the real one:
static int (*_chk_open)(sd_journal **, int) = sd_journal_open;
static int (*_chk_open_dir)(sd_journal **, const char *, int) = sd_journal_open_directory;
```
This is a bit like the `var _ component.Component     = (*Component)(nil)` in Go.

We could also use `go test -asan`? I didn't know about it before, but apparently it's an address sanitiser which catches common CGO bugs.

internal/component/loki/source/journal/journal.go
```diff
@@ -157,52 +140,40 @@ func (c *Component) CurrentHealth() component.Health {
...
 func (c *Component) reloadTailer() {
-	// Grab current state
 	c.mut.RLock()
 	var tailerToStop *tailer
 	if c.tailer != nil {
 		tailerToStop = c.tailer
 	}
-	rcs := alloy_relabel.ComponentToPromRelabelConfigs(c.args.RelabelRules)
-	c.mut.RUnlock()
+	c.mut.Unlock()
```
ptodev:
```suggestion
	c.mut.RUnlock()
```

internal/component/loki/source/journal/tailer.go
```diff
@@ -26,349 +19,213 @@ import (
...
+				t.pos.PutString(t.key, "", cursor)
+				continue
+			}
 
-	var msg string
+			if !relabel.ProcessBuilder(t.br, t.rcs...) {
+				t.logger.Debug("journal entry dropped by relabel rules")
+				t.pos.PutString(t.key, "", cursor)
+				continue
+			}
```
ptodev, line 177: Should this increment `emptyLabelsError`? It used to do it. I suppose it should keep doing it. If not, then we'd need a different debug metric label.

kalleep: Ah Yes. Me and kyle discussed that in another pr too. But Yes it should

internal/component/loki/source/journal/internal/sdjournal/sdjournal.go
kalleep: That check would only work if we have headers at compile time, so if we want the have that requirement then yes we should do something like that. If the signature of these functions actually changed we would also need to have headers with those changes at compile time otherwise it would not catch it.

ptodev: What does including the headers consist of? Is it just a matter of having a `.h` file in the repo? Or do we need to compile on Linux and install something via the Linux package manager?

kalleep: Currently we do not vendor these headers in alloy but install them in our build image. https://github.com/grafana/alloy/blob/main/build-tools%2Fbuild-image%2FDockerfile#L68

My initial idea was to not even include them so its easier to cross compile alloy to Linux but we can still require them.

If we do we should do what you suggested. The library we use do not do that so it would be better than we have today

ptodev: If we can cross-compile using C code in a CGO comment, then why wouldn't we cross-compile a more complete `.h` file?

Maybe I'm missing something 🤔 I don't understand why the stripped down C code compiles on non-Linux platforms. Basically on Windows we need to get "lfcn.h", which is probably only accessible if we build under something like Cygwin which emulates a unix system. And if we have to do that, then I'm not sure it's true cross compilation.

kalleep: We can do it but then we have a requirement that this header is present at compile time. That is fine and more safe with the thing you suggested. I just started out with not having that as a requirement

ptodev: Ok, thank you, I see now. Yeah I'm happy to include a `.h` instead, it'd be safer.

kalleep: So I made some changes in https://github.com/grafana/alloy/pull/6508/changes/95d3102478f19d5f20a043a2b5a488690fddfe94

internal/component/loki/source/journal/internal/sdjournal/sdjournal.go
```diff
@@ -1,10 +1,5 @@
-//go:build linux && cgo
+//go:build linux && cgo && promtail_journal_enabled
```
ptodev, line 2: A comment like this would be useful:

```
// Package sdjournal is a minimal wrapper over libsystemd's sd-journal API.
//
// We don't link libsystemd at build time. Instead openLib resolves each
// function at runtime with dlopen/dlsym, so the default Alloy binary has no
// hard dependency on libsystemd; it is only needed when loki.source.journal
// runs. dlsym returns an untyped void*, which Go can't call directly, so for
// each function we define a tiny C wrapper (named j_<func>) that casts the
// void* back to the real signature and calls it.
//
// To wrap a new sd_journal_<x> function, copy its real signature from
// systemd/sd-journal.h and, following the existing examples in the cgo preamble
// below, add:
//  1. a typedef for that signature,
//  2. a _Static_assert(... "drift") that checks our typedef still matches the
//     header (this is the safety net: a mismatch fails the build),
//  3. a j_<x> wrapper that casts the void* and calls the function.
//
// Then, on the Go side:
//  4. add a field for it to the lib struct and dlsym it in openLib,
//  5. add the Go method that calls C.j_<x>.
//
// The _Static_assert checks only run when this package is compiled against the
// systemd headers, which CI does via the promtail_journal_enabled build (see
// build.yml and packaging.mk). Never drop the #include <systemd/sd-journal.h>:
// that would silently disable every check.
```

ptodev (commented):
LGTM! It'd just be nice to clarify how sdjournal.go should be maintained. A comment at the top is enough.

ptodev (approved):

PR #6514, opened 2026-06-16
ci: Fix backport workflow when multiple commits have same title

jharvey10:
The backport workflow had a bug whereby if two commits being backported share the same git commit title, the second one will silently fail to create its PR, thinking that the commit was already backported.

Instead, now the exact line placed by `git cherry-pick -x` is searched for in the commit messages.

### DIFF
--- a/tools/release/backport/command.go
+++ b/tools/release/backport/command.go
@@ -25,7 +25,6 @@ type backportInfo struct {
 	PRNumber       int
 	OriginalPR     *github.PullRequest
 	MergeCommitSHA string
-	CommitSHA      string
 	AppIdentity    gh.AppIdentity
 	TargetBranch   string
 	BackportBranch string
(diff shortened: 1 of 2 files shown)


tools/release/backport/command.go
```diff
@@ -168,27 +168,23 @@ func resolveBackportInfo(ctx context.Context, client *gh.Client, prNumber int, t
 		return nil, fmt.Errorf("getting app identity: %w", err)
 	}
 
-	// Check if backport was already merged by looking for the original PR
-	// title in the release branch history.
-	alreadyMerged, err := client.CommitExistsWithPattern(ctx, gh.FindCommitParams{
-		Branch:  targetBranch,
-		Pattern: originalPR.GetTitle(),
+	commitSHA := originalPR.GetMergeCommitSHA()
```
ptodev: `commitSHA` is the same as `MergeCommitSHA`?

ptodev (dismissed):
I just added one minor comment. It makes sense, but I wish we had tests for this. It's hard to test manually otherwise, and a bit hard to review by a person.

tools/release/backport/command.go
jharvey10: Good call. I think this was some slop on slop. I'm removing the `CommitSHA` and changing all of the references to just use `MergeCommitSHA`. I also added a small tweak to the docs to mention that this is specifically for squash-merged PRs where the squash "merge commit" (that's not technically a merge commit) is the same as the entire PR's contents.

ptodev (approved):

PR #6520, opened 2026-06-17
chore: Implement consumer for loki.echo

kalleep:
### Pull Request Details

Extracted from https://github.com/grafana/alloy/pull/6204. Updating `loki.echo` to implement consumer interface. We still expose `LogsReceiver` through the pipeline.

### Notes to the Reviewer

Bring over changes that have to be made to `FanoutConsumer` and implementing `InterceptorConsumer`

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -151,9 +151,12 @@ func (b *Batch) Clone() Batch {
 
 // ConsumeStreams calls fn for each stream in the batch and then resets the batch.
 // The callback receives ownership of the stream.
-func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64)) {
+// If callback returns false iteration will stop.
+func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64) bool) {
 	for _, s := range b.streams {
-		fn(s, b.created)
+		if !fn(s, b.created) {
+			break
+		}
 	}
 	b.Reset()
 }
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -151,14 +151,15 @@ func (b *Batch) Clone() Batch {
 
 // ConsumeStreams calls fn for each stream in the batch and then resets the batch.
 // The callback receives ownership of the stream.
-// If callback returns false iteration will stop.
-func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64) bool) {
+// If callback returns errors interation will stop
+func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64) error) error {
+	defer b.Reset()
 	for _, s := range b.streams {
-		if !fn(s, b.created) {
-			break
+		if err := fn(s, b.created); err != nil {
+			return err
 		}
 	}
-	b.Reset()
+	return nil
 }
 
 // Reset clears the batch so it can be reused.
--- a/internal/component/common/loki/consumer.go
+++ b/internal/component/common/loki/consumer.go
@@ -13,6 +13,7 @@ var ErrConsumerStopped = errors.New("consumer stopped")
 
 type Consumer interface {
 	Consume(ctx context.Context, batch Batch) error
+	// TODO: Remove this when we have moved over to batching.
 	ConsumeEntry(ctx context.Context, entry Entry) error
 }
 
--- a/internal/component/common/loki/consumer.go
+++ b/internal/component/common/loki/consumer.go
@@ -2,9 +2,14 @@ package loki
 
 import (
 	"context"
-	"reflect"
+	"errors"
+	"slices"
 	"sync"
 )
+
+// ErrConsumerStopped is returned by Consumer when an entry is
+// submitted after the consumer has been stopped.
+var ErrConsumerStopped = errors.New("consumer stopped")
 
 type Consumer interface {
 	Consume(ctx context.Context, batch Batch) error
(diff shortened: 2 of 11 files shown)


internal/component/common/loki/consumer_fanout.go
```diff
@@ -53,31 +59,31 @@ func (f *FanoutConsumer) Consume(ctx context.Context, batch Batch) error {
...
-				errs = append(errs, err)
-			}
+	for _, consumer := range consumers {
+		if consumer == nil {
 			continue
 		}
 
-		if err := consumer.ConsumeEntry(ctx, entry.Clone()); err != nil {
+		// ConsumeEntry does not clone entries. Components that mutate an entry must
+		// clone it first, as they already do. Batching will own cloning at the batch/pipeline boundary.
```
ptodev, line 86: It'd be better to have such comments inside `type Consumer interface {`, where the function declarations are.

internal/component/common/loki/consumer_fanout.go
```diff
@@ -97,3 +103,12 @@ func (f *FanoutConsumer) Update(consumers []Consumer) {
 		f.mut.RUnlock()
 	}
 }
+
+func isReportableError(err error) bool {
+	// We can always ignore ErrConsumerStopped.
+	// If a downstream component have been stopped due to config update
+	// we should no longer forward entries there and when alloy it self
```
ptodev:
```suggestion
	// we should no longer forward entries there and when alloy itself
```

internal/component/common/loki/consumer_fanout_test.go
```diff
@@ -96,6 +96,32 @@ func TestFanoutConsumer_Consume(t *testing.T) {
 	require.Equal(t, "original", lastStreams[0].Entries[0].Line)
 }
 
+func TestFanoutConsumer_NilConsumer(t *testing.T) {
+	fanout := NewFanoutConsumer([]Consumer{nil})
+	require.NotPanics(t, func() {
+		_ = fanout.Consume(context.Background(), NewBatch())
+		_ = fanout.ConsumeEntry(context.Background(), NewEntry(model.LabelSet{}, push.Entry{}))
```
ptodev: Why not check for `require.NoError`? It's more specific.

internal/component/common/loki/consumer_fanout.go
```diff
@@ -53,31 +59,31 @@ func (f *FanoutConsumer) Consume(ctx context.Context, batch Batch) error {
 }
 
 func (f *FanoutConsumer) ConsumeEntry(ctx context.Context, entry Entry) error {
```
ptodev, line 62: IIUC, this function is only temporary until all components start using batching? Is it worth adding a TODO to remove it then? It'd be awkward to permanently have two consumer functions.

internal/component/common/loki/consumer.go
```diff
@@ -47,23 +52,11 @@ func (c *CollectingConsumer) Batches() []Batch {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
-	return c.batches
+	return slices.Clone(c.batches)
```
ptodev, line 56: Why do we need this clone?

internal/component/common/loki/consumer_interceptor.go
```diff
@@ -0,0 +1,76 @@
...
+func (i *InterceptorConsumer) Consume(ctx context.Context, batch Batch) error {
+	if i.onConsume != nil {
+		batch, err := i.onConsume(ctx, batch)
+		if err != nil || batch.EntryLen() == 0 {
+			return err
+		}
+		return i.next.Consume(ctx, batch)
+	}
+
+	return errors.New("loki interceptor: unimplemented consume")
```
ptodev: Is this too harsh? After all we could just call ConsumeEntry instead.

internal/component/common/loki/consumer.go
kalleep: Yeah so this is strictly for tests. I remember getting a race where I checked the output while still producing logs.

internal/component/common/loki/consumer_fanout.go
kalleep: I don't think so since this is specific to one implementation of Consumer (fanout). When we move completely over to batching I will be safe to mutate e.g. labels directly since fanout handles it but we don't add it to other implementations since that would be pretty bad.

internal/component/common/loki/consumer_fanout.go
kalleep: Yes it's temporary. Will add a TODO in all places we implement it

internal/component/common/loki/consumer_interceptor.go
kalleep: You mean going from Consume to Consume entry in that case? I guess that could work. Then we could support batching in stages.

kalleep: Okay I updated the interceptor to fallback to ConsumeEntry

internal/component/common/loki/batch.go
```diff
@@ -151,11 +151,15 @@ func (b *Batch) Clone() Batch {
 
 // ConsumeStreams calls fn for each stream in the batch and then resets the batch.
 // The callback receives ownership of the stream.
-func (b *Batch) ConsumeStreams(fn func(stream Stream, created int64)) {
+// If callback returns errors interation will stop
```
ptodev, line 154: I'm not sure how the early return works together with the batch reset. Is it not a problem if we send to 5 downstream components and only the 3rd component fails? Then the last 2 components won't get a batch?

ptodev (approved):
Thank you! LGTM, just one minor comment for ConsumeStreams. Maybe I'm misunderstanding something

internal/component/common/loki/batch.go
kalleep: In the case we are propagating batches through the pipeline we will deep clone the batch to all downstream except one. But we are not passing a pointer to a batch so reset will not affect the batch state of the caller component but only in the place that the batch is "consumed", it more of a semantic in that stream ownership is transferred and we can mutate stream state without corrupting the batch.

For the case of the (temporary) fallback behavior in the interceptor I would imagine that first error would be enough to fail the whole batch and we don't handle the fanout part here. 

But these are good questions to think about and I will follow up with a pr with some tests that composes different Consumers and how they behave with errors etc so it's clearer

PR #6521, opened 2026-06-17
fix: Bump moduledeps fixtures to go 1.26.4

thampiotr:
## Summary
- Update moduledeps test fixture `go.mod` files from `go 1.26.2` to `go 1.26.4` to address CVE-2026-42504.
- Update corresponding `go.mod.expected` fixture snapshots so moduledeps E2E assertions stay aligned.
- Confirm no remaining `1.26.2` references in repository `go.mod` files.

## Test plan
- [x] Run `go test ./...` in `tools/`

### DIFF
--- a/tools/generate/moduledeps/testdata/basic-mod/go.mod
+++ b/tools/generate/moduledeps/testdata/basic-mod/go.mod
@@ -1,3 +1,3 @@
 module test.example.com
 
-go 1.26.2
+go 1.26.4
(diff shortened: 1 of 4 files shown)


kalleep (approved):
We do exclude these when running the update go command https://github.com/grafana/alloy/blob/main/tools/internal/discover/gomod.go#L8 but I cant remember why I did it 🤔

PR #6524, opened 2026-06-17
fix: Stabilize Windows CI in Loki tests and static converter

thampiotr:
## Summary
- Fix Windows test leaks by waiting for `Run(...)` goroutines to exit in `TestEntrySentToTwoProcessComponents` and `TestDeleteRecreateFile`.
- Keep the existing TLS emptiness check (`reflect.DeepEqual(*args.TLS, http.TLSArguments{})`) while normalizing empty Windows TLS filter values to `nil`.
- Normalize empty Windows exporter `Time`/`Filetime` slice representation to defaults to avoid spurious `time { }` output in converter snapshots.

## Validation
- `go test ./internal/component/loki/process -run TestEntrySentToTwoProcessComponents -count=1`
- `go test ./internal/component/loki/source/file -run TestDeleteRecreateFile -count=1`
- `go test ./internal/converter/internal/staticconvert/internal/build -count=1`
- `go test ./internal/converter/internal/staticconvert -count=1`

### DIFF
--- a/internal/component/loki/process/process_test.go
+++ b/internal/component/loki/process/process_test.go
@@ -1194,20 +1194,26 @@ stage.static_labels {
 
 	// The lines were received after processing by each component, with no
 	// race condition between them.
+	seenCh1 := false
+	seenCh2 := false
 	for i := 0; i < 2; i++ {
 		select {
 		case logEntry := <-ch1.Chan():
 			require.WithinDuration(t, time.Now(), logEntry.Timestamp, 1*time.Second)
 			require.Equal(t, "writing some text", logEntry.Line)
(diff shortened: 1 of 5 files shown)


internal/component/loki/process/process_test.go
```diff
@@ -1146,15 +1146,33 @@ stage.static_labels {
 	args2.ForwardTo = []loki.LogsReceiver{ch2}
 
 	ctx, ctxCancel := context.WithCancel(t.Context())
-	defer ctxCancel()
+	var (
+		runWG  sync.WaitGroup
+		runErr = make(chan error, 3)
+	)
+	runComponent := func(runFn func() error) {
+		runWG.Add(1)
```
kalleep, line 1154: nit: use runWg.Go then you don't have to manually call Add and Done

kalleep (approved):

PR #6532, opened 2026-06-17
refactor(http): Make the /-/healthy endpoint suitable for liveness probes

### LINKED ISSUE #6340: The concept of component health is ambiguous
### Component(s)

_No response_

### Add a description

The docs don't make it clear what "component health" means and how to use it. For now it seems to be just for informational purposes and it should not be used as part of automated operational decisions on whether to restart Alloy.

There is a [disclaimer](https://grafana.com/docs/alloy/latest/reference/http/#-healthy) that the `/healthy` endpoint is not suitable for K8s liveness probes:

> The /-/healthy endpoint isn’t suitable for a [Kubernetes liveness probe](https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/).
> 
> You don’t necessarily need to restart an Alloy instance that reports as unhealthy. For example, a component may be unhealthy due to an invalid configuration or an unavailable external resource. In this case, restarting Alloy would not fix the problem. A restart may make it worse, because it would stop the flow of telemetry in healthy pipelines.

This can be counterintuitive for some users.

We need to clarify what component health is. E.g. is it something that gets reported when Alloy needs restarting? If yes, then how do we report components which aren't working but for which a restart isn't going to work? 

What are the conventions used in other similar applications? Prometheus [almost always](https://github.com/prometheus/prometheus/issues/3650) reports HTTP 200 on its `/healthy` endpoint. Grafana Agent used to do the same as Prometheus.

### Tip

vishalmore90:
### Purpose
Resolves #6340 by changing the `/-/healthy` HTTP endpoint behavior to always return `HTTP 200 OK` once Alloy is running.

### Background
Currently, the `/-/healthy` endpoint checks all internal components and returns `HTTP 500` if any component reports as unhealthy. This makes the endpoint unsuitable for Kubernetes liveness probes, as restarting Alloy does not fix common component-level failures (such as configuration errors or temporary unavailability of external backends). In fact, restarts can worsen issues by interrupting other healthy pipelines.

### Changes
- Updated `/-/healthy` in `internal/service/http/http.go` to return `HTTP 200 OK` with "Alloy is healthy." without performing component health checks.
- Updated `TestHTTP` and `TestUnhealthy` in `internal/service/http/http_test.go` to expect the new behavior and status code.
- Updated documentation in `docs/sources/reference/http/_index.md` to state that the endpoint is suitable for liveness probes to verify process status, and added notes on monitoring individual components via the UI or metrics.

### DIFF
--- a/docs/sources/reference/http/_index.md
+++ b/docs/sources/reference/http/_index.md
@@ -26,28 +26,17 @@ Otherwise, if the instance isn't ready, the `/-/ready` endpoint returns `HTTP 50
 
 ## `/-/healthy`
 
-When all {{< param "PRODUCT_NAME" >}} components are working correctly, all components are considered healthy.
-If all components are healthy, the `/-/healthy` endpoint returns `HTTP 200 OK` and the message `All Alloy components are healthy.`.
-Otherwise, if any of the components aren't working correctly, the `/-/healthy` endpoint returns `HTTP 500 Internal Server Error` and an error message.
-You can also monitor component health through the {{< param "PRODUCT_NAME" >}} [UI](../../troubleshoot/debug#alloy-ui).
+The `/-/healthy` endpoint returns `HTTP 200 OK` and the message `Alloy is healthy.` to indicate that the {{< param "PRODUCT_NAME" >}} instance is running.
 
 ```shell
 curl localhost:12345/-/healthy
-All Alloy components are healthy.
-```
-
-```shell
-curl localhost:12345/-/healthy
-unhealthy components: math.add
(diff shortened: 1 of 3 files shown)


ptodev:
I'll close this because resolving this is not as simple as just returning HTTP 200. We need component health to be a genuinely useful concept. This could require setting some conventions and documenting what sort of situations cause a component to be unhealthy. For kubernetes liveness probes I'm not sure whether just returning HTTP 200 at all times is useful. It should probably take the actual component health into account at least a little bit.

PR #6539, opened 2026-06-18
fix(promethues.exporter.mongodb): Add currentop_slow_time argument

### LINKED ISSUE #6516: MongoDB exporter currentops collector cannot be enabled
### Component(s)

prometheus.exporter.mongodb

### What's wrong?

Even with `collect_all` or `enable_currentop_metrics` enabled, the `mongodb_currentop_*` metrics are never scraped.

The problem is, that the `CurrentOpSlowTime` is not set, so the percona exporter disabled the currentop collector completely. 

Alloy v1.17.0 includes percona v0.47.2.
The missing option is defined here: https://github.com/percona/mongodb_exporter/blob/v0.47.2/exporter/exporter.go#L80
and checked here: https://github.com/percona/mongodb_exporter/blob/v0.47.2/exporter/exporter.go#L221-L225

So when creating the exporter here: https://github.com/grafana/alloy/blob/v1.17.0/internal/static/integrations/mongodb_exporter/mongodb_exporter.go#L107
we should pass some time value for `CurrentOpSlowTime`.
Compare that to what the exporter cli uses: https://github.com/percona/mongodb_exporter/blob/v0.47.2/main.go#L80

### Steps to reproduce

Configure the mongodb exporter with default arguments, with include `collect_all`.

### Software version

Grafana Alloy v1.17.0

### Tip

kalleep:
### Pull Request Details
As pointed out by the issue it's currently impossible to run currentop collector since we never set slow time. 

Fixes: #6516

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.exporter.mongodb.md
+++ b/docs/sources/reference/components/prometheus/prometheus.exporter.mongodb.md
@@ -34,26 +34,27 @@ prometheus.exporter.mongodb "<LABEL>" {
 
 You can use the following arguments with `prometheus.exporter.mongodb`:
 
-| Name                           | Type     | Description                                                                                                                            | Default | Required |
-| ------------------------------ | -------- | -------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
-| `mongodb_uri`                  | `secret` | MongoDB node connection URI.                                                                                                           |         | yes      |
-| `collect_all`                  | `bool`   | Enables all collectors.                                                                                                                | `true`  | no       |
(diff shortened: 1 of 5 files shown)


ptodev (dismissed):

ptodev (approved):
It's not necessary to add this to the converter - Agent static mode wouldnt have this feature. But it's ok to have it.

kalleep:
>It's not necessary to add this to the converter - Agent static mode wouldnt have this feature. But it's ok to have it.

Well tell that to our tests :see_no_evil:

PR #6541, opened 2026-06-18
fix(ci): Only stamp BuildDate for release builds

### LINKED ISSUE #4266: Only pass -X .../build.BuildDate for official builds to improve developer experience and compilation time
### Component(s)

_No response_

### Add a description

```
Piotr Gwizdala
One thing I've noticed is that each time you build there's a -X .../build.BuildDate set which changes, so it recompiles as the value is different each time. We probably don't want to eliminate it for official builds, but should think of a way to remove it when not needed for local builds. (edited) 
```

### Tip

Churi12:
## What this does

The BuildDate linker flag is derived from the current time, so it changes on
every build. Because the value is different each time, the Go build cache is
invalidated and the binary is relinked on every local build even when nothing
else changed. This was raised in #4266.

This change computes BuildDate only when it is actually needed:

- Release builds (RELEASE_BUILD=1) stamp the current time, as before.
- Reproducible builds via SOURCE_DATE_EPOCH stamp the pinned time, as before.
- Local development builds leave it empty, so repeated builds stay cached and
  skip the relink.

The SOURCE_DATE_EPOCH path keeps the exact same date invocation as before, so
reproducible builds are unaffected.

## Why empty is fine for dev builds

BuildDate is a plain string passed through to the Prometheus version package.
An empty value simply renders as an empty build date in `alloy --version`; it
is not parsed and does not affect runtime behavior.

## Verification

Built the collector binary locally with the resulting dev-mode flags and
confirmed the build cache behavior:

- Rebuild with the empty (constant) BuildDate: 0 link invocations (cache hit).
- Rebuild changing only BuildDate: 1 link invocation (forced relink), which is
  the current behavior the issue describes.

Also confirmed the binary builds and runs with an empty BuildDate, and that
RELEASE_BUILD=1 still stamps the build time.

Fixes #4266

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,15 +151,20 @@ BEYLA_VERSION := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f2)
 BEYLA_PKG     := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f1)/pkg/buildinfo
 VPREFIX      := github.com/grafana/alloy/internal/build
 VPREFIXSYNTAX := github.com/grafana/alloy/syntax/internal/stdlib
+# Allow the Go build cache to be used except when doing a release build or when
+# an epoch is explicitly set. BuildDate otherwise changes on every invocation,
+# which busts the cache and forces a relink on every local build.
 ifdef SOURCE_DATE_EPOCH
-    DATE_STAMP = -d@$(SOURCE_DATE_EPOCH)
+    BUILD_DATE = $(shell date -u -d@$(SOURCE_DATE_EPOCH) +"%Y-%m-%dT%H:%M:%SZ")
+else ifeq ($(RELEASE_BUILD),1)
+    BUILD_DATE = $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 endif
 GO_LDFLAGS   := -X $(VPREFIX).Branch=$(GIT_BRANCH)                        \
                 -X $(VPREFIX).Version=$(VERSION)                          \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -151,15 +151,21 @@ BEYLA_VERSION := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f2)
 BEYLA_PKG     := $(shell echo $(BEYLA_MODULE) | cut -d' ' -f1)/pkg/buildinfo
 VPREFIX      := github.com/grafana/alloy/internal/build
 VPREFIXSYNTAX := github.com/grafana/alloy/syntax/internal/stdlib
+# BuildDate changes on every invocation, which busts the Go build cache and
+# forces a relink on every local build. Only stamp it for release builds, or
+# when SOURCE_DATE_EPOCH pins it to a fixed value for reproducible builds.
+# Local development builds leave it empty so repeated builds stay cached.
```
jharvey10: The comment sort of describes where we came from, but not where we are now. How about something like this?

> Allow the Go build cache to be used except when doing a release build or when an epoch is explicitly set.

jharvey10 (commented):
🙌

Makefile
Churi12: Good point, that reads better. Adopted your wording and pushed. Thanks for the review.

jharvey10 (approved):

Makefile
jharvey10: Thanks! Looks great. I changed the PR title slightly so it shows up in our next changelog :)

PR #6550, opened 2026-06-18
feat: Add helm option for the Beyla Kubernetes SDK injection controller

grcevski:

When Beyla and the Kubernetes SDK injection controller are deployed in a Kubernetes cluster, for the purpose of automatic injection of OpenTelemetry SDKs, they need to exchange information. This information is passed through specially designed configmap which the Beyla Alloy eBPF component needs to write. For this purpose, when this mode is enabled, Alloy needs RBAC permissions to write this configmap, only to its own namespace.

### Pull Request Details

I'm adding a Helm chart option to enable additional RBAC for alloy to get, create and update configmaps in its own namespace, when this deployment mode is enabled.

### PR Checklist

- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/operations/helm/charts/alloy/ci/sdk-injection-controller-values.yaml
+++ b/operations/helm/charts/alloy/ci/sdk-injection-controller-values.yaml
@@ -1,0 +1,3 @@
+# Enable RBAC permissions required by the Beyla SDK injection controller.
+sdkInjectionController:
+  enabled: true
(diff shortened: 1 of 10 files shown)


operations/helm/charts/alloy/ci/sdk-injection-controller-values.yaml
```diff
@@ -0,0 +1,3 @@
+# Enable RBAC permissions required by the Beyla SDK injection controller.
+sdkInjectionController:
+  enabled: true
```
petewall, line 3: Is there something we could add to:
```yaml
alloy:
  configMap:
    content: "..."
```
so demonstrate why we want this flag to be set to true?

Issue #598, opened 2024-04-17
Reuse official binaries in homebrew


ptodev:
### Request

The [homebrew recipe](https://github.com/grafana/homebrew-grafana/blob/master/alloy.rb) for Alloy is currently building Alloy form source instead of reusing the official binaries from our releases. It would be good to reuse the official binaries - the installation would be faster, and wouldn't [require](https://github.com/grafana/homebrew-grafana/blob/d3830c691877531c9e2955df302e3205c149cef5/alloy.rb#L8-L10) installing things like Go. Maybe it also wouldn't require "node" and "yarn"?

### Use case

Faster installations and simpler code.

rfratto:
I feel like it's against the Homebrew expected patterns to directly download pre-built binaries. IMO we should try to make our Formula feel as native to Homebrew as possible, although that means going through the extra work of automating building bottles for people who don't want to build from source.

moxious:
In the context of early getting started processes with Grafana Cloud, many paths go through installing Alloy.  At present, it takes a substantial amount of time to download the extra packages and compile from source, which slows the getting started process.  Additionally, in lab environments on multiple occasions I have seen breakages (alloy failed to install) that owe to this particular issue.  Typically the breakage is caused by something unrelated to alloy that the user doesn't understand and isn't in a position to fix.   An example would be some other homebrew author changing a needed go dependency package, which isn't available for this-or-that silicon, or a conflict with a different version of a devtool needed for a different install.  This tends to leave users stuck and frustrated, because the step of "install Alloy" is something they're taking within the context of getting started with Grafana Cloud - and they come no further without it.

blewis12:
Hey folks, i'm actually doing some digging into this now and had created a [duplicate issue](https://github.com/grafana/alloy/issues/4627) (apologies, I didn't see this one!)

If it makes sense I can just assign myself to this issue if no one else is actively looking into it? and remove the duplicate one

moxious:
I would LOVE to see this fixed.  It's very much still a thing

blewis12:
Tacking on [this related issue](https://github.com/grafana/homebrew-grafana/issues/138) that I can also take a look at when I get to this. I'm currently working on a high priority project so haven't had time to get round to this one yet, but when things calm down a tad I can dip back in :)

x1unix:
Grafana Alloy is available as a prebuilt binary in homebrew-core.

[...]

PR #6551, opened 2026-06-18
chore: Add Alloy service wrapper generator for Homebrew formula

x1unix:

Add a `alloy-wrapper` generator for Homebrew packaging needed for Homebrew formulas.

### Pull Request Details

Our Alloy formula in [homebrew-grafana](https://github.com/grafana/homebrew-grafana) tap provides a small `alloy-wrapper` shell script for Launchd service.

The wrapper sources the environment file, reads the extra-args file, and `exec`s `alloy run` with the configured storage and config paths.

As we're planning to deprecate our formula in favor of `grafana-alloy` formula from [homebrew-core](https://github.com/Homebrew/homebrew-core), I'm working on upstreaming this wrapper.

As Homebrew maintainers [prefer not to maintain shell scripts](https://github.com/Homebrew/homebrew-core/pull/288356#discussion_r3426441072) that upstream doesn't ship, the tooling to generate the script should be provided by Alloy itself, so formula can use it.

The script should be generated dynamically in the formula so it can use `#{pkgetc}`, `#{opt_bin}` and other Homebrew-specific values.

Also, keeping the generator in-tree means the wrapper is versioned with the Alloy release it ships in.

The generator is meant to be used by Homebrew formula, do not require any additional dependencies and therefore is put into `packaging/homebrew` directory.

#### Otel mode support

Homebrew formula can have only one `service do` block to declare a service.

Therefore, the generated `alloy-wrapper` script supports toggling between alloy and otel modes if `ALLOY_OTEL_MODE` is set.

The environment variable can be set using an env file, which the script reads at startup.

In addition to that, Otel mode has a separate extra args file.

#### Follow-up work

After a new version of Alloy is released with a bundled generator, both formulas (`homebrew-grafana` and `homebrew-core`) should be updated to use the generator:

```ruby
cd "packaging/homebrew/service-wrapper-gen" do
  system "go", "run", ".",
    "--alloy-bin=#{opt_bin}/alloy",
    "--config-path=#{pkgetc}",
    "--storage-path=#{var}/lib/grafana-alloy/data",
    "--env-file=#{pkgetc}/config.env",
    "--extra-args-file=#{pkgetc}/extra-args.txt",
    "--otel-extra-args-file=#{pkgetc}/otel-extra-args.txt",
    "--out=#{buildpath}/alloy-wrapper"
end
bin.install "alloy-wrapper"
```

Related to:

- https://github.com/grafana/homebrew-grafana/issues/134
- https://github.com/grafana/alloy/issues/598

### PR Checklist

- [x] Tests updated


packaging/homebrew/service-wrapper-gen/wrapper.tpl
```diff
@@ -0,0 +1,14 @@
...
+if [ -f "{{.EnvFile}}" ]; then
+  set -a
+  . "{{.EnvFile}}"
+  set +a
+fi
+
+extra_args=""
+[ -f "{{.ExtraArgsFile}}" ] && extra_args=$(cat "{{.ExtraArgsFile}}")
+
+exec "{{.AlloyBin}}" run \
```
thampiotr: I think we want to support OTel Engine here from day one. So we need some way to select the engine and then either run `run` or `otel` command and support both YAML and Alloy configuration.

x1unix: @thampiotr done in https://github.com/grafana/alloy/pull/6551/changes/a45baccb10ff3b184bb0820da4d55001bb8e2a9e and https://github.com/grafana/alloy/pull/6551/changes/c9a076d1f6b6954b6c20e0cb315d5550f3e55206

packaging/homebrew/service-wrapper-gen/README.md
```diff
@@ -0,0 +1,87 @@
...
+The Homebrew formula installs a small `alloy-wrapper` shell script that its
+`service` block runs as the launchd/systemd entrypoint. The wrapper sources the
+environment file, reads the extra-args file, and `exec`s alloy.
+
+The subcommand is chosen at runtime from the environment: if `ALLOY_OTEL_MODE`
+is `1`, `true`, `yes`, or `on`, the wrapper runs `alloy otel`; otherwise it
+runs `alloy run`. The two modes take different flags — `run` uses
+`--storage.path` and the config path argument, while `otel` uses
+`--config="file:<config-path>/config.yaml"` and does not support
+`--storage.path`. Set the variable in the formula's `config.env` to switch a
```
thampiotr: Should we explain that there is also option to pass extra args? I think users will frequently need this:
- Enable public-preview features in Alloy
- Add extra paths to OTel config file via additional `--config` flags (multiple are allowed)
- ...

thampiotr: Or we don't enumerate them here, as they are listed below.

packaging/homebrew/service-wrapper-gen/README.md
```diff
@@ -0,0 +1,87 @@
...
+single service between modes — Homebrew supports only one `service` block per
+formula, so the mode is a runtime toggle rather than a second service.
+
+This program emits that script. All Homebrew paths are passed as flags, and the
+script body lives in `wrapper.tpl` (embedded via `//go:embed`) so it reads and
+edits as a plain shell script. Keeping the generator in-tree means the wrapper
+is versioned with the Alloy release it ships in, rather than maintained
+separately inside the formula.
+
+## Why a separate module
```
thampiotr, line 18: Just a quick check: anything we need to do for a new module? Like enlisting it to Renovate or in Go version update Makefile/Taskfile

packaging/homebrew/service-wrapper-gen/testdata/core.golden.sh
thampiotr, line 1: Would `shellcheck` look at this file if it doesn't have `.sh` extension? I think it would be good idea to run linter on this to avoid any misuse.

thampiotr (commented):
Looking good, few smaller comments.

packaging/homebrew/service-wrapper-gen/README.md
x1unix: @thampiotr renovate and our toolchain [are globbing](https://github.com/grafana/alloy/blob/main/tools/internal/discover/gomod.go#L8) `go.mod` files, so I assume nothing needs to be updated.

packaging/homebrew/service-wrapper-gen/testdata/core.golden.sh
x1unix: @thampiotr done: https://github.com/grafana/alloy/pull/6551/changes/aa3792fa7dead4f13268406d42b66c353a4ece31

packaging/homebrew/service-wrapper-gen/README.md
x1unix: > Should we explain that there is also option to pass extra args?

@thampiotr the `caveats` block of our homebrew formula [does that](https://github.com/grafana/homebrew-grafana/blob/master/alloy.rb#L82).

I will be also updating the `grafana-alloy` formula from the core when this PR gets merged.

I've also cleaned up a bit the readme: https://github.com/grafana/alloy/pull/6551/changes/0a1a8b61b9feb082858fdaa5f088e32a5b275416

thampiotr (approved):

PR #6563, opened 2026-06-20
feat(otelcol.receiver.loki): Add selective label mapping via labels block

vishalmore90:
## What

Adds an optional `labels` configuration block to `otelcol.receiver.loki` that
allows users to selectively control which Loki labels become OTel log record
attributes during ingestion.

Closes #6556

## Why

Currently all Loki labels are unconditionally forwarded as log record
attributes. This is problematic when labels like `instance`, `stream`, or
`level` should remain as log-level attributes, be dropped entirely, or be
renamed to match OTel semantic conventions (e.g. `job` → `service.name`).

The only workaround today is adding a downstream `otelcol.processor.attributes`
step, which is wasteful for high-volume Loki ingestion since the data is
already structured at that point.

## How

The new `labels` block supports three optional attributes:

| Attribute | Type | Description |
|-----------|------|-------------|
| `include` | `list(string)` | Allowlist — only these labels are forwarded |
| `exclude` | `list(string)` | Blocklist — these labels are dropped |
| `rename`  | `map(string)` | Rename labels during conversion |

- `include` and `exclude` are mutually exclusive (validation error if both set)
- `rename` is applied after filtering
- When `labels` block is omitted, all labels are forwarded (backward compatible)
- The `filename` → `log.file.path`/`log.file.name` special handling is
  independent of label filtering

### Example

```alloy
otelcol.receiver.loki "default" {
  labels {
    include = ["job"]
    rename = {
      job = "service.name",
    }
  }

  output {
    logs = [otelcol.exporter.otlphttp.default.input]
  }
}

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.loki.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.loki.md
@@ -42,11 +42,38 @@ You can use the following blocks with `otelcol.receiver.loki`:
 
 | Block              | Description                                        | Required |
 |--------------------|----------------------------------------------------|----------|
+| [`labels`][labels] | Configures selective label forwarding.             | no       |
 | [`output`][output] | Configures where to send converted telemetry data. | yes      |
 
+[labels]: #labels
 [output]: #output
 
 {{< /docs/alloy-config >}}
+
+### `labels`
+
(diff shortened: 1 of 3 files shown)


kalleep:
Since this is a bigger change to this component the issue have been marked as a proposal so I will close this pr until it has a approved design

vishalmore90:
Hello @kalleep, 
Thanks for letting me know, Could you share any template format for reference that you did like me to follow for this?

PR #6566, opened 2026-06-21
feat(ui): Render pipeline graph via @grafana/alloy-pipeline-graph

fabienne-m:
## Summary

Replaces the in-tree React Flow + dagre pipeline visualization (`features/graph`) with the shared `@grafana/alloy-pipeline-graph` package. Alloy now only owns the thin adapter layer that shapes its runtime API into the package's input interface.

https://github.com/user-attachments/assets/ca88da4f-e647-47b9-8e20-f6d0ec1e5f96

- **`buildPipelineGraph`**: maps `ComponentInfo[]` into `PipelineGraphData`, emitting custom-component module loaders as expandable containers with nested child nodes.
- **`useModuleInternals`**: lazily fetches `/modules/{id}/components` for each container to populate inner nodes.
- **`overlayLiveMetrics`**: overlays the live `/graph` debug stream onto edges as per-signal flow rates (`signalMetrics`), grouped per edge and per signal kind (metrics, logs, traces, targets, other), so a single edge can show independent rates for each signal. Stale metrics are cleared when no live data matches an edge in the current window. Covered by unit tests.
- **Graph page**: wires the renderer, Window slider, legend and node-click navigation to component detail.
- Drops `@xyflow/react` and `@dagrejs/dagre`; dedupes `react`/`react-dom` in Vite so the symlinked package shares a single React copy.

## Notes / TODO before merge

- :warning: The package is currently referenced via a local `file:` path. It **must be pinned to a published npm version** before this can pass CI / be merged.

## Test plan

- [ ] `cd internal/web/ui && yarn test` (overlayLiveMetrics unit tests pass)
- [ ] `make lint`
- [ ] Manual: open the pipeline graph UI, verify nodes/edges render, custom components expand, live per-signal rates appear on edges, and node-click navigates to component detail.

Made with [Cursor](https://cursor.com)


internal/web/ui/src/pages/Graph.tsx
```diff
@@ -38,33 +49,49 @@ function Graph() {
...
+
+  const graph: PipelineGraphData = useMemo(() => overlayLiveMetrics(baseGraph, debugData), [baseGraph, debugData]);
+
+  const onNodeClick = (node: PipelineNode) => {
+    const nodeModuleID = String(node.meta?.moduleID ?? '');
+    const localID = String(node.meta?.localID ?? node.id);
+    const baseUrl = globalThis.window.location.origin + pathPrefix;
+    const remoteCfgPrefix = nodeModuleID.startsWith('remotecfg/') ? 'remotecfg/' : '';
+    const path = nodeModuleID !== '' ? `component/${nodeModuleID}/${localID}` : `component/${localID}`;
+    globalThis.window.open(baseUrl + remoteCfgPrefix + path, '_blank');
```
jharvey10: I know this preserves the original behavior, but do you know if there's a way to make it so when you click on the nodes, they behave like normal anchors and respect normal click vs. middle click vs cmd+click to let the user decide whether or not they want to open the link in a new tab? I have a long-standing todo on my list to investigate this for accessibility reasons. Wondering if your new graph implementation would make this more possible than the old one.

internal/web/ui/.prettierignore
```diff
@@ -1,3 +1,4 @@
 build/
 dist/
 **/*tsconfig.json
+src/test/generated_fixtures/discovery.kubernetes.heavy.json
```
jharvey10: minor: missing new line at the end of this file

internal/web/ui/vite.config.ts
```diff
@@ -109,6 +109,14 @@ export default defineConfig(({ mode }) => ({
...
   // The <base> tag in index.html will be set at runtime by the Go server
   base: './',
+  resolve: {
+    // Force a single copy of React across the app and any linked packages
+    // (e.g. the symlinked @grafana/alloy-pipeline-graph library, which ships
+    // its own node_modules/react). Without this, Vite bundles two copies of
+    // React and hooks fail at runtime with "Cannot read properties of null
+    // (reading 'useContext')".
+    dedupe: ['react', 'react-dom'],
+  },
```
jharvey10: I've definitely hit this with other projects before :( I'm not able to see the alloy-pipeline-graph repo at the moment, but I think this usually happens when the pulled in dependency (alloy-pipeline-grpah) has react listed as a `dependencies` dep instead of a `devDependencies` dep (or one of its dependencies has react as a top-level dep). Based on the lockfile alone, I'm not seeing anything indicating that multiple versions of react are being included. Was this perhaps a problem earlier in development that's no longer an issue?

internal/web/ui/.prettierignore
fabienne-m: Fixed — added the trailing newline.

internal/web/ui/vite.config.ts
fabienne-m: I was able to remove this!

internal/web/ui/src/pages/Graph.tsx
fabienne-m: this is now fixed :)

internal/web/ui/src/features/pipeline/buildPipelineGraph.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+function buildComponentNode(component: ComponentInfo, overrides: Partial<PipelineNode> = {}): PipelineNode {
+  return {
+    id: component.localID,
+    componentName: component.name,
+    label: component.label ?? null,
+    stage: inferPipelineStage(component.name),
+    signals: deduplicateSignals(inferSignalsFromComponentName(component.name)),
+    docsUrl: `https://grafana.com/docs/alloy/latest/reference/components/${component.name}`,
```
CatPerry: Possibly broken docs URLs:
The repo's own docs are namespaced: `docs/sources/reference/components/prometheus/prometheus.remote_write.md` → the live URL is `.../reference/components/prometheus/prometheus.remote_write/` (namespace subdir + trailing slash). This template produces `.../reference/components/prometheus.remote_write` — no namespace, no trailing slash — which may404 for the namespaced majority of components

internal/web/ui/src/features/pipeline/overlayLiveMetrics.ts
```diff
@@ -0,0 +1,92 @@
...
+import { parseID } from '../../utils/id';
+import { type DebugData, DebugDataType } from './debugDataType';
+
+/**
+ * Maps a live debug-stream data type onto the signal a graph edge carries.
+ * `target` (service-discovery counts) gets its own signal so it stays visually
+ * distinct from telemetry; `undefined` is the frontend's placeholder and never
+ * arrives from the backend.
+ */
+function debugTypeToSignal(type: DebugDataType): SignalKind {
```
CatPerry, line 12: Signal granularity is silently reduced (intent question):

The old legend distinguished six colors: prometheus_metric, `otel_metric`, `loki_log`, `otel_log`, `otel_trace`, `target`. `debugTypeToSignal` collapses these into Metrics / Logs / Traces / Targets, and `overlayLiveMetrics` sums prometheus+otel into one Metrics number and loki+otel into one Logs number. That's a deliberate simplification, but it's a real information-loss regression that the PR description frames as a neutral "overlay," not a behavior change. Is losing the Prometheus-vs-OTel distinction intended and signed off?

internal/web/ui/src/hooks/useModuleInternals.ts
```diff
@@ -0,0 +1,54 @@
...
+          : `./api/v0/web/modules/${moduleId}/components`;
+        const response = await fetch(url, {
+          cache: 'no-cache',
+          credentials: 'same-origin',
+          signal: abortController.signal,
+        });
+        if (response.ok) {
+          result.set(moduleId, (await response.json()) as ComponentInfo[]);
+        }
+      }
```
CatPerry: **Serial fetches + silent failures.** These module fetches run sequentially (`await` inside the `for…of`), so N custom components = N back-to-back round-trips. And a non-OK response is silently dropped — the container just renders empty with no signal. Suggest parallelizing and logging failures:

```ts
const entries = await Promise.all(
  Array.from(moduleIds, async (moduleId) => {
    const url = isRemotecfg
      ? `./api/v0/web/remotecfg/modules/${moduleId}/components`
      : `./api/v0/web/modules/${moduleId}/components`;
    const response = await fetch(url, { cache: 'no-cache', credentials: 'same-origin', signal: abortController.signal });
    if (!response.ok) {
      console.error(`Failed to load module ${moduleId}: ${response.status}`);
      return null;
    }
    return [moduleId, (await response.json()) as ComponentInfo[]] as const;
  })
);
setInternals(new Map(entries.filter((e) => e !== null)));
```

internal/web/ui/src/features/pipeline/buildPipelineGraph.ts
fabienne-m: Good catch — fixed. The URL now includes the namespace subdirectory and trailing slash (e.g. `.../components/prometheus/prometheus.remote_write/`), via a shared `componentDocsUrl` helper. Also applied the same fix to the pre-existing broken link in `ComponentView.tsx`.

internal/web/ui/src/hooks/useModuleInternals.ts
fabienne-m: Addressed — module fetches now run in parallel via `Promise.all`, and non-OK responses are logged with `console.error` instead of being silently dropped.

internal/web/ui/src/features/pipeline/overlayLiveMetrics.ts
fabienne-m: Yes, this is intentional and was a deliberate contract decision for the shared package: the graph shows what *kind* of telemetry flows (Metrics/Logs/Traces), not which protocol carries it — the protocol distinction is visible from the component names themselves (`prometheus.*` vs `otelcol.*`). Collapsing to signal kinds keeps the legend/color contract consistent with Fleet Management, which consumes the same package. Note it's not purely a reduction either: `Targets` was added as a first-class signal so discovery-target counts stay visually distinct instead of being lumped in as before. Happy to discuss further if you feel the protocol split is worth keeping.

CatPerry (dismissed):

.cursor/plans/shared_pipeline_graph_package_8727b4b6.plan.md
jharvey10, line 1: I think this file should be removed and/or a gitignore added for .cursor/plans

internal/web/ui/src/pages/Graph.tsx
```diff
@@ -89,7 +96,7 @@ function Graph() {
       ) : (
         components.length > 0 && (
           <div className={styles.graphWrapper}>
-            <PipelineGraph graph={graph} onNodeClick={onNodeClick} theme="light" />
+            <PipelineGraph graph={graph} theme="light" />
```
jharvey10, line 96: 🎉

internal/web/ui/src/pages/Graph.tsx
```diff
@@ -51,16 +51,23 @@ function Graph() {
...
-  const graph: PipelineGraphData = useMemo(() => overlayLiveMetrics(baseGraph, debugData), [baseGraph, debugData]);
+  const liveGraph: PipelineGraphData = useMemo(() => overlayLiveMetrics(baseGraph, debugData), [baseGraph, debugData]);
 
-  const onNodeClick = (node: PipelineNode) => {
-    const nodeModuleID = String(node.meta?.moduleID ?? '');
-    const localID = String(node.meta?.localID ?? node.id);
+  // Give each node an `href` to its component page. The graph renders these as
+  // real anchors, so clicks follow native browser semantics: a plain click
+  // opens in the same tab, while ⌘/ctrl/middle-click open a new tab and the
+  // browser's "open in new tab" / "copy link address" context menu works.
```
jharvey10: Minor: I think this comment could be removed or shortened to just mention that the href is later used to generate anchors, but not needing to describe the resulting UI behavior.

internal/web/ui/src/utils/docs.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Builds the reference docs URL for a component. Component reference docs live
+ * under a namespace subdirectory matching the component's prefix, e.g.
+ * `prometheus.remote_write` is documented at
+ * `reference/components/prometheus/prometheus.remote_write/`.
+ */
+export function componentDocsUrl(componentName: string): string {
+  const namespace = componentName.split('.')[0];
+  return `https://grafana.com/docs/alloy/latest/reference/components/${namespace}/${componentName}/`;
```
jharvey10, line 13: Maybe a future TODO: It's fairly likely that a user's alloy version will be outdated. The alloy graphql API does support a `version` field which could be used to get more specific here, but there's no UI-side plumbing yet for it. There might be other existing ways to get the alloy version too, but it's not super important since the docs site has a version selector anyway.

internal/web/ui/vite.config.ts
```diff
@@ -109,13 +109,8 @@ export default defineConfig(({ mode }) => ({
...
   base: './',
-  resolve: {
-    // Force a single copy of React across the app and any linked packages
-    // (e.g. the symlinked @grafana/alloy-pipeline-graph library, which ships
-    // its own node_modules/react). Without this, Vite bundles two copies of
-    // React and hooks fail at runtime with "Cannot read properties of null
-    // (reading 'useContext')".
-    dedupe: ['react', 'react-dom'],
+  build: {
+    chunkSizeWarningLimit: 750,
```
jharvey10: it's honestly probably fine to remove this. we do get a warning during the UI build, but we just ignore it because it's practical impact is so minimal for our use cases.

jharvey10 (commented):
Looking great! Just a couple things that could use removing and a future consideration.

.cursor/plans/shared_pipeline_graph_package_8727b4b6.plan.md
fabienne-m: Yeah good point. I removed it and also the empty folder.

internal/web/ui/src/pages/Graph.tsx
fabienne-m: Addressed, shortened the comment to just mention that the href is rendered as an anchor.

internal/web/ui/src/utils/docs.ts
fabienne-m: Good idea, added a TODO to the docstring mentioning the GraphQL `version` field so we can pick this up in a follow-up.

internal/web/ui/vite.config.ts
fabienne-m: Addressed, removed the `chunkSizeWarningLimit` override.

jharvey10 (approved):

PR #6568, opened 2026-06-22
fix: Report io errors on globbing

### LINKED ISSUE #6542: Alloy silently swallows error and does not send logs on missing log file permission
### Component(s)

loki.write

### What's wrong?

I am using the Grafana K8s monitoring chart to mainly collect logs of my pods. My K8s cluster has some nodes that use OpenSuse (read-only file system). I noticed that pod logs from that node were not getting transmitted to Loki. Inspecting the container to check for the log files yielded a permission denied error, as reading them was not possible without a privileged user. 
This was especially annoying as Alloy was not logging anything about the error, my expected behaviour would be that for every target log file that have been discovered, an error is logged in case it could not been read.

### Steps to reproduce

1. Use read-only OS (OpenSuse)
2. install and run some k8s instance (minikube/k3sd)
3. deploy a basic pod
4. deploy Loki instance
5. deploy Alloy using the K8s monitoring chart

### System information

openSUSE MicroOS, kernal:  7.0.12-1-default, k3s  v1.35.5+k3s1

### Software version

Alloy v1.17.0

### Configuration

```text

```

### Logs

```text

```

### Tip

kalleep:
### Pull Request Details
If alloy tries to glob a directory the user don't have access to the error was ignored because we did not pass the options that surfaced it. 

Fixes: #6542

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/util/glob/glob_unix.go
+++ b/internal/util/glob/glob_unix.go
@@ -13,7 +13,7 @@ func NewGlobber() Globber {
 }
 
 func (g *defaultGlobber) FilepathGlob(pattern string) ([]string, error) {
-	return doublestar.FilepathGlob(pattern)
+	return doublestar.FilepathGlob(pattern, doublestar.WithFailOnIOErrors())
 }
 
 func (g *defaultGlobber) PathMatch(pattern, path string) (bool, error) {
--- a/internal/util/glob/glob_windows.go
+++ b/internal/util/glob/glob_windows.go
@@ -15,7 +15,11 @@ func NewGlobber() Globber {
 }
 
 func (g *windowsGlobber) FilepathGlob(pattern string) ([]string, error) {
-	return doublestar.FilepathGlob(pattern, doublestar.WithCaseInsensitive())
+	return doublestar.FilepathGlob(
+		pattern,
+		doublestar.WithFailOnIOErrors(),
+		doublestar.WithCaseInsensitive(),
+	)
 }
 
 func (g *windowsGlobber) PathMatch(pattern, path string) (bool, error) {


kalleep:
I am a bit worried this can become pretty verbose since it will log this error every time we re-run it, by default every `10s`. 

We could memorize the error and only emit new once for a specific path.

internal/util/glob/glob_unix.go
```diff
@@ -13,7 +13,7 @@ func NewGlobber() Globber {
 }
 
 func (g *defaultGlobber) FilepathGlob(pattern string) ([]string, error) {
-	return doublestar.FilepathGlob(pattern)
+	return doublestar.FilepathGlob(pattern, doublestar.WithFailOnIOErrors())
```
blewis12, line 16: A question around previous vs new behavior with errors: If for example `FilepathGlob` was called on directory `/var/log/*.log` which contained two files: `my_logs.log` which was readable and `my_secrets.log` which was restricted and returned IO errors - would the previous behavior be that only `my_logs.log` be returned as a match and `my_secrets` be skipped? 

I ask because if we return an error now, I think we'd be skipping _all_ targets that return under a glob. Both in [loki.source.file](https://github.com/grafana/alloy/blob/24fc84634920e01fbc289ad421aa0ca5605ed238/internal/component/loki/source/file/resolver.go#L81) and [local.file_match](https://github.com/grafana/alloy/blob/24fc84634920e01fbc289ad421aa0ca5605ed238/internal/component/local/file_match/file.go#L146)

kalleep: Good catch. So lets say we have two target directories A and B. The user running alloy have permissions on A but not B.

If we have two targets it will work since we only report errors on B.

But if we have one single globing pattern, e.g. `/var/logs/mylogs/**/*.log` it will fail.

Will have to see if there is a way we can solve that :+1:

ferr079:
Hit this outside Kubernetes and ended up at the same root cause, so here's an independent reproduction plus an idea for the design question raised in the review.

### Reproduction (plain systemd host, no K8s)

Collecting Proxmox VE's proxy log on a Debian host:

- log file `/var/log/pveproxy/access.log`, mode `0640`, owner `www-data:www-data`
- the `alloy` user *is* a member of `www-data` — `id alloy` confirms it
- but the parent directory `/var/log/pveproxy/` is `0700 root:root`, so `alloy` cannot traverse it

Result: `loki_source_file_files_active_total` stays at `0`, `journalctl -u alloy` is empty at every level, the component stays healthy, and the deployment reports success while collecting nothing. The ops-side fix was `chmod 0750` on the directory.

One debugging note for anyone landing here: `runuser -u alloy -- test -r <file>` was not a reliable oracle (it answered "no" while group membership was in fact correct). `files_active_total` was.

### Why nothing surfaces

This matches @kalleep's reading, and the silence is structural on both sides:

- `internal/util/glob/glob_unix.go` calls `doublestar.FilepathGlob(pattern)` with no options, so `forwardErrIfFailOnIOErrors` drops the `EACCES` and returns `([], nil)`. Both branches of `FilepathGlob` do this — the `Glob`/`os.DirFS` path and the literal-path `os.Lstat` special case.
- `local.file_match`: `watch.go` `getPaths()` only logs from `filepath.Abs` / `os.Stat`, which run over *already matched* paths, so an empty match set reaches neither. `file.go` `getWatchedFiles()` does have an `"error getting paths"` branch, but it is unreachable for I/O errors today.
- `loki.source.file`: `resolver.go` `globResolver.Resolve()` has the same shape — `logger.Error("failed to resolve target")` only fires when the globber itself returns an error.

Both go quiet through the shared `glob.Globber`, which is also why #6568 addresses both at once.

### On the two objections in this PR

@blewis12's (one unreadable directory drops *every* match under the glob) and @kalleep's (a log line every `sync_period`, 10s by default) are both consequences of putting `WithFailOnIOErrors()` on the production glob. There may be a way to get the observability without the semantics change: keep globbing *without* the option, and wrap the filesystem so the errors are observed as the walk goes past them.

`FilepathGlob` is essentially `SplitPattern` + `os.DirFS(base)` + `Glob(fsys, f, opts...)` + re-joining the results. If that were inlined with Alloy's own `fs.FS` in place of `os.DirFS`:

```go
type observingFS struct {
	fs.FS
	onErr func(path string, err error)
}
// ReadDir / Stat: on error, call onErr(path, err), then return the error unchanged
```

then — with `WithFailOnIOErrors` still off — doublestar swallows that returned error exactly as it does today and keeps walking, so partial matches are preserved, while `onErr` sees every individual `EACCES`. That includes the case a top-level error cannot distinguish: a single unreadable subtree in the middle of a `**` pattern, where the glob still returns plenty of valid matches and nothing looks wrong.

`onErr` is also the natural home for the memoization @kalleep suggested: keep the last error per path and emit only on transition (first failure, then recovery), so a permanently closed directory costs one line instead of one every 10s. A counter (`glob_io_errors_total` or similar) would make it visible without depending on log level at all.

Two details for that route: the literal-path `os.Lstat` special case needs the same treatment, and `glob_windows.go` must keep `WithCaseInsensitive()`. Both `NewGlobber()` call sites (`local.file_match`, `loki.source.file`) already have a logger in scope, so threading one in stays contained.

Happy to put this up as a PR with tests — a `0700` directory, and a glob where one subtree is unreadable while the other matches still come through — if that's a direction you'd want to go. Didn't want to push code at a draft you may already have plans for.

kalleep:
@ferr079 Please read our [gen AI policy](https://github.com/grafana/alloy/blob/66ebea856dffb242f921775e9552a133f60dc5c1/docs/developer/genai.md#not-acceptable).

PR #6569, opened 2026-06-22
feat(pyroscope): Add scope labels to Alloy profiles

korniltsev-grafanista:
Add `otel.scope.name` `otel.scope.version` seriels labels to profiles from `pyroscope.scrape`, `pyroscope.ebpf`, `pyroscope.java`

### DIFF
--- a/internal/component/pyroscope/ebpf/reporter/pprof.go
+++ b/internal/component/pyroscope/ebpf/reporter/pprof.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/google/pprof/profile"
+	"github.com/grafana/alloy/internal/component/pyroscope"
 	"github.com/grafana/alloy/internal/component/pyroscope/ebpf/discovery"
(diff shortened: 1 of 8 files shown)


simonswine (approved):
LGTM

What I am wondering is if we somone use this alloy version with an old Pyroscope v1 (before https://github.com/grafana/pyroscope/pull/3335) that doesn't allow dots, the user has no way to opt out. 

Maybe this is acceptable, but let me know what you think. This is obivously a problem wider than just in the Alloy PR.

I also think it is fair to say upgrade your 2 year old Pyroscope.

simonswine:
Also for Alloy it might make sense to document this labels a bit similar to what we have here: https://github.com/grafana/alloy/blob/d55dee06218549804da54d295d9a5f7108a097ca/docs/sources/reference/components/pyroscope/pyroscope.scrape.md?plain=1#L123-L127

korniltsev-grafanista:
Follow-up docs PR addressing the review comment about documenting the scope labels: #6653

cc @simonswine

PR #6592, opened 2026-06-24
ci: Ensure bot-authored commits are signed

jharvey10:

Switches bot-authored workflow commits to paths that produce signed commits.

This is now required by Grafana policy.

The backport workflow now has the release tool prepare the cherry-pick locally, then lets `peter-evans/create-pull-request` create or update the signed backport PR.

The release-please workflow still lets release-please create or update its PR first. A small release-please plugin writes the root PR branch, title, and body to workflow outputs, so the later collector distro regeneration step can update the same PR without re-querying GitHub for metadata.

This still means release-please can trigger CI twice for the same PR. That is left for the follow-up CI work.

### PR Checklist

- [x] Tests updated


tools/release/internal/git/git.go
```diff
@@ -221,31 +217,128 @@ func DeleteLocalBranch(branch string) error {
...
-			continue
+// GetStagedChanges returns the staged changes in the repository, using paths
+// relative to the repository root.
+func GetStagedChanges() (StagedDiff, error) {
+	root, err := repoRoot()
+	if err != nil {
+		return StagedDiff{}, err
+	}
+
+	additionPaths, err := getStagedPaths(root, "ACMRT")
```
kgeckhart: The addition only carries path + contents, no mode. So a staged `chmod +x` (shows up as `M`) commits the unchanged content and drops the executable bit, and a symlink (`T`) gets written as a regular file with the link target as its body. Probably never hits us with go.mod/go.sum/collector regen — but it's a real difference from `git push`, so fine as long as it's a conscious call.

tools/release/backport/command.go
```diff
@@ -112,24 +121,51 @@ func run(ctx context.Context, flags flags) (retErr error) {
...
+	changes, baseSHA, err := prepareBackportCommit(ctx, client, info)
+	if err != nil {
 		return err
 	}
+	if err := client.CreateBranch(ctx, gh.CreateBranchParams{
+		Branch: info.BackportBranch,
+		SHA:    baseSHA,
+	}); err != nil {
+		return fmt.Errorf("creating backport branch: %w", err)
+	}
```
kgeckhart: One thing I noticed here is that it's not atomic anymore. The old `git push` did the branch and the commit together, but now they're two separate API calls. So if `CreateBranch` succeeds and then the commit call fails, we're left with an empty branch on the remote. The annoying part is the next run sees that branch already exists, `resolveBackportInfo` returns nil, and the backport quietly never happens. Retrying won't get us out of it, and since we dropped `DeleteBranch` in this PR there's nothing handy to clean it up with.

Wonder if we could delete the branch when the commit step fails, or check for an empty branch and treat it as unfinished? Curious what you think.

tools/release/internal/git/git.go
kgeckhart: I don't think it's a blocker but something to call out. I think the addition only sends the path and contents, no file mode, so if something staged a `chmod +x` we'd commit the content but lose the executable bit. Symlinks would have the same issue, they'd land as a regular file with the link target as the contents. Honestly I doubt it ever comes up with go.mod/go.sum/collector regen. Just didn't want it to bite us later.

tools/release/internal/git/git_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if headline != "chore: Regenerate collector distro" {
+		t.Fatalf("unexpected headline: %q", headline)
+	}
+	if body != "Generated after release-please updated versions." {
+		t.Fatalf("unexpected body: %q", body)
+	}
+}
+
+func TestGetCherryPickCommitMessagePrintsGitLogOutput(t *testing.T) {
```
kgeckhart: Right now it only checks that git-log gets printed to stdout, but not what the function actually returns, so the headline/body split and the `(cherry picked from commit ...)` line aren't really covered. That line matters since it's what `FindCherryPickedCommit` uses to spot dupes, so might want to assert on the returned message too.

kgeckhart (commented):
Nice work, a few notes

tools/release/internal/git/git.go
jharvey10: I can absolutely see this biting us at some point. I'll see how easy/hard it is to identify these types of changes and likely just require a manual backport of them.

tools/release/backport/command.go
jharvey10: This logic got completely reworked, so this is all managed by the action now.

tools/release/internal/git/git.go
jharvey10: Logic got a rework and this is handled by the action now.

tools/release/internal/git/git_test.go
jharvey10: This (along with a bunch of other logic) got removed in favor of the action approach.

.github/workflows/check-generate-otel-collector-distro.yml
```diff
@@ -6,22 +6,32 @@ on:
...
+
 jobs:
   check:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Debounce release-please PR updates
+        if: github.event_name == 'pull_request' && startsWith(github.head_ref, 'release-please--branches--')
+        # Debounce this check on release PRs while the generated-files update is pushed.
+        run: sleep 30
```
kgeckhart, line 21: Could you expand on the why here?

kgeckhart (approved):
Changes look good I just had a question on one part.

.github/workflows/check-generate-otel-collector-distro.yml
jharvey10: Yeah I can add more comments on there. Basically since we have two commits coming in in quick succession, the first one will always cause the otel check to fail CI. On these PRs that’s probably fine, because nobody is getting spam email from it, but a “canceled” run feels like a nicer thing to have in the logs, and this amount of time should give the second commit enough of a head start to interrupt the first run instead of having them both “complete” in sequence.

jharvey10: I say “quick succession” but internally the second commit is running the regen just like how the CI check does, so it takes at least as long as this first check to run and actually push that second commit.

kgeckhart: Ahhh that's fair sometimes we get some quick merges back-to-back

PR #6593, opened 2026-06-25
fix(windows_service): Gracefully terminate alloy when service is stopped

### LINKED ISSUE #6572: Wrong log positions with Windows Alloy
### Component(s)

loki.source.file

### What's wrong?

`loki.source.file` on Windows under-reports bytes read from log files, storing incorrect offsets in `positions.yml`. On every Alloy restart, files are re-read from the wrong position and duplicate entries are sent to Loki.

Could you please have a look? Thank you in advance!

### Steps to reproduce

1. Deploy Alloy v1.17.0 on Windows
2. Configure `loki.source.file` to tail log files with mixed `LF`/`CRLF` line endings e.g. [1781510805205.log](https://github.com/user-attachments/files/29212176/1781510805205.log)
3. Let Alloy ingest some files
4. Check `positions.yml` - stored offsets will be much smaller than actual file sizes
5. Restart Alloy - files are re-read from the wrong offset, duplicates sent to Loki

### 1. `read_bytes_total` is completely wrong

For **new files** discovered after a fresh Alloy start (no prior position entry):

| File | Actual Size | `read_bytes_total` | `read_lines_total` |
|------|------------|-------------------|-------------------|
| 1781868450754.log | **745,409** | **24** | 82 |
| 1781868732641.log | 14,789 | **24** | 176 |
| 1781868928358.log | 12,923 | **24** | 37 |

All lines are successfully read and sent to Loki (`sent_entries = read_lines = 217,311`), but the byte counter reports **24 bytes** regardless of actual file size.

For **pre-existing files** with a prior position entry:

| File | Actual Size | `read_bytes_total` | `positions.yml` offset |
|------|------------|-------------------|----------------------|
| 1781510805205.log | 4,411 | 1,553 | **"1553"** |

### 2. Massive ingestion spikes on every restart

Querying Loki (`sum(count_over_time({job="desktop-app-logs"}[5m]))`) shows:

| Time (UTC) | Lines ingested in 5min | Event |
|------------|----------------------|-------|
| 07:35 | **315,761** | Alloy restart 1 |
| 11:25 | **212,355** | Alloy restart 2 |
| Normal | ~300-500 | Steady state |

Each restart produces a **600-1000x spike** as Alloy re-reads all files from the wrong offset.

### 3. Same file content sent multiple times

File `1781510805205.log` (a write-once, 4,411-byte file) was ingested:
- **07:32 UTC** - 45 entries (restart 1)
- **11:20 UTC** - 44 entries (restart 2)

Same content, sent twice to Loki.

### 4. Positions file confirms the wrong offset

For the same unchanged 4,411-byte file, different Alloy runs store different wrong offsets:

| Alloy Run | Stored Position | Actual File Size | % of file |
|-----------|----------------|-----------------|-----------|
| Restart ~07:35 | (unknown) | 4,411 | ? |
| Restart 11:20 | **24** | 4,411 | 0.5% |
| Restart 13:20 | **1,553** | 4,411 | 35.2% |

```yaml
? path: C:\Desk\Errors\APP - 7.81.006\PROD\prod\ACTION\1781510805205.log
    labels: '{}'
  : "1553"     # stored offset (WRONG, actual file size: 4,411)
```

On next restart, Alloy seeks to the wrong offset, re-reads part of the file after the offset and sends duplicates.

### System information

Windows Server 2025 Datacenter Azure Edition 24H2 26100.32995 x64

### Software version

Grafana Alloy v1.17.0

### Configuration

```text
local.file_match "app_logs_main" {
  path_targets = [
    { __path__ = "C:\\Desk\\Errors\\APP*\\PROD\\prod\\*\\*.log" },
  ]
  sync_period = "10s"
}


loki.source.file "scrape_app_logs_main" {
  targets       = local.file_match.app_logs_main.targets
  forward_to    = [loki.process.app_logs_main.receiver]
  tail_from_end = false
}


loki.process "app_logs_main" {

  stage.replace {
    expression = "(?P<value>[A-Za-z0-9+/]{20,}==)"
    replace    = "****"
  }

  stage.regex {
    expression = "C:\\\\Desk\\\\Errors\\\\APP(?: - (?P<version>[^\\\\]+))?\\\\PROD\\\\prod\\\\(?P<directory>[^\\\\]+)\\\\(?P<filename>[^\\\\]+)\\.log"
    source     = "filename"
  }

  stage.static_labels {
    values = {
      job  = "desktop-app-logs",
      service_name = "app",
    }
  }

  stage.template {
    source   = "message"
    template = "dir={{ .directory }} version={{ .version }} file={{ .filename }} {{ .Entry }}"
  }

  stage.output {
    source = "message"
  }

  stage.label_drop {
    values = ["filename"]
  }
}

loki.write "loki" {
  endpoint {
    url = "https://loki.platform-lab.internal.mycompany.org/loki/api/v1/push"
  }
}
```

### Logs

```text

```

### Tip

kalleep:
### Pull Request Details
When alloy is running as a windows service we are really running a small wrapper program that spawns alloy as a [sub process](https://github.com/grafana/alloy/blob/31df04cd579920117c8655f2673b46fec9354ebf/internal/cmd/alloy-service/service.go#L65).

On windows when we cancel the provided context Go will end up calling [TerminateProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess) and this can lead to all kinds of problems, e.g. not flushing state to disk.

My first attempt was to remove this wrapper in https://github.com/grafana/alloy/pull/6591 but I cannot even get alloy to start, main is never triggered. My best bet is that some init function in on dependency exits early but I have not confirmed this yet. I also tracked down https://github.com/grafana/agent/issues/3243, this is the original issue for why this wrapper was created in the first place. So for now we keep it.

But signals don't really exist on windows in the same way they do on unix system so it's not possible for us to send e.g. SIGTERM to tell alloy to stop. 

The workaround is to allocate a console for the running process and spawn alloy process in a new process group. That way we can generate a console event that alloy will pick up.

Fixes: https://github.com/grafana/alloy/issues/6572

### Notes to the Reviewer

I made the service and service_test files windows only since we need to use windows specific api:s

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


collector/generator/main_alloy.tpl
```diff
@@ -1,3 +1,5 @@
+// GENERATED CODE: DO NOT EDIT
```
ptodev: This is misleading - the tpl file isn't generated. Let's add it the warning during the generation as before?

collector/main_windows.go
```diff
@@ -1,29 +1,13 @@
...
-			// Per https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicectrldispatchera#return-value
-			// this means that the process is not running as a service, so run interactively.
-			return runInteractive(params)
-		}
-
-		return fmt.Errorf("failed to start collector server: %w", err)
-	}
-
-	return nil
+	return runInteractive(params)
```
ptodev, line 12: Does this mean that OTel Engine doesn't behave the same way as upstream OTel? Should we note this in the OTel Engine docs? There could be some sort of "Differences" section.

internal/cmd/alloy-service/service_windows.go
```diff
@@ -81,11 +84,20 @@ func (svc *serviceManager) Run(ctx context.Context) {
...
+	cmd.Stdout = svc.cfg.stdout
+	cmd.Stderr = svc.cfg.stderr
 	cmd.Env = os.Environ()
-	cmd.Env = append(cmd.Env, svc.cfg.Environment...)
+	cmd.Env = append(cmd.Env, svc.cfg.environment...)
+
+	// Put the child in its own process group so we can target it specifically
+	// with a console control event.
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: windows.CREATE_NEW_PROCESS_GROUP}
+	cmd.Cancel = func() error {
```
ptodev, line 97: Do you think it's worth adding a somewhat long WaitDelay here just in case something goes wrong with the signal?

ptodev (commented):
Claude gave me an interesting alternative:

> Alternative: named event object. Instead of allocating a console and broadcasting CTRL_BREAK across a process group, the wrapper could create a named Windows event (CreateEvent with a name unique per child, e.g. derived from the child PID) and pass the name to Alloy via an env var or flag. Alloy opens it (OpenEvent) and a goroutine blocks on WaitForSingleObject; on stop the wrapper calls SetEvent, the goroutine wakes and triggers the same graceful-shutdown path we already wire to os.Interrupt. This targets exactly the one child process — no AllocConsole, no extra conhost.exe, no process-group semantics that could also hit any subprocess Alloy spawns — and it's deterministic to unit-test (just set the event) rather than depending on the test runner having a console. The tradeoff is it needs a small Alloy-side addition (open + wait on the handle) whereas the console approach reuses Alloy's existing signal handling untouched, and it doesn't sidestep the underlying Go limitation discussion in golang/go#72884 so much as avoid it. Either way I'd still pair it with a WaitDelay/force-kill fallback so a stalled shutdown can't hang the service in STOP_PENDING.

It seems cleaner than the console? I'd never heard of consoles before. It looks like they are normally used by UIs and it's a but of a hacky solution in this situation?

collector/main_windows.go
kalleep: When using otel engine you need to use the otel sub-command in alloy. If we today create our own service for alloy it can _only_ run otel collector, ni sub command required. 

So for now you need run alloy through the wrapper and that way you can run either engine.

internal/cmd/alloy-service/service_windows.go
kalleep: I was thinking about that before making the or. When alloy stops we need to stop all running components so they have a chance to e.g. update position files, flush outgoing data etc. 

We have a flag in alloy to controll max stop time for a component so I think we should use that instead of using WaitTime

kalleep:
> Claude gave me an interesting alternative:
> 
> > Alternative: named event object. Instead of allocating a console and broadcasting CTRL_BREAK across a process group, the wrapper could create a named Windows event (CreateEvent with a name unique per child, e.g. derived from the child PID) and pass the name to Alloy via an env var or flag. Alloy opens it (OpenEvent) and a goroutine blocks on WaitForSingleObject; on stop the wrapper calls SetEvent, the goroutine wakes and triggers the same graceful-shutdown path we already wire to os.Interrupt. This targets exactly the one child process — no AllocConsole, no extra conhost.exe, no process-group semantics that could also hit any subprocess Alloy spawns — and it's deterministic to unit-test (just set the event) rather than depending on the test runner having a console. The tradeoff is it needs a small Alloy-side addition (open + wait on the handle) whereas the console approach reuses Alloy's existing signal handling untouched, and it doesn't sidestep the underlying Go limitation discussion in golang/go#72884 so much as avoid it. Either way I'd still pair it with a WaitDelay/force-kill fallback so a stalled shutdown can't hang the service in STOP_PENDING.
> 
> It seems cleaner than the console? I'd never heard of consoles before. It looks like they are normally used by UIs and it's a but of a hacky solution in this situation?

Yes it is possible to do that and it would probably be a bit more idiomatic. I avoided this for now since it would be a bit more complex to setup and this does not alter the shutdown mechanism we already have.

collector/main_windows.go
ptodev: Yes, but what I mean is that the way OTel engine runs as a service won't be the same as the way upstream OTel runs as a service. We'd need to mention this in the docs at some point. It doesn't have to be mentioned here though.

Do we need an additional changelog note that this change fixes Alloy installations which run Default engine without a service wrapper?

internal/cmd/alloy-service/service_windows.go
ptodev: Which flag is it? I didn't know about it.

ptodev (approved):
Thank you, LGTM. I just wonder if we should improve the docs in the future. The OTel Engine docs could mention how the windows service differs from upstream's. And even Default Engine's docs might need to mention that if people don't run the service wrapper then Alloy may not shut down cleanly. But there's no need to do it in this PR (there's no strong need to do in general, really),

internal/cmd/alloy-service/service_windows.go
kalleep: https://github.com/grafana/alloy/blob/main/internal/alloycli/cmd_run.go#L72

collector/main_windows.go
kalleep: > Do we need an additional changelog note that this change fixes Alloy installations which run Default engine without a service wrapper?

I don't think so, for now we just removed this ability. It's currenyly not possible to run Alloy as a service without the wrapper since we never implement the service event loop in alloy directly.

PR #6595, opened 2026-06-25
chore: Update github.com/wasilibs/go-re2

### LINKED ISSUE #6289: Official Alloy builds require MemoryDenyWriteExecute=no due to go-re2/wazero path
### What's wrong?

The official Alloy v1.16.x Linux package cannot run under a strict systemd unit with `MemoryDenyWriteExecute=yes` in at least one deployment. Alloy panics during startup/config validation with a wazero/go-re2 executable mmap path similar to:

```text
wazero/internal/engine/wazevo.mmapExecutable
```

The practical workaround is to set `MemoryDenyWriteExecute=no`, which weakens the systemd sandbox by allowing runtime executable mappings. This is not ideal for hardened bare-metal/VM deployments where Alloy otherwise does not need extra Linux capabilities.

### Why this appears tied to the release build

From Alloy v1.16.1 source:

- The top-level `Makefile` forcibly prepends the `gore2regex` build tag to `GO_TAGS`:

```make
# If GO_TAGS does not already contain gore2regex, prepend it.
# This makes loki.secretfilter use the go-re2 library, which provides
# a substantial performance improvement over stdlib's regex.
ifeq ($(filter gore2regex,$(GO_TAGS)),)
override GO_TAGS := $(strip gore2regex $(GO_TAGS))
endif
```

- `internal/component/all/all.go` imports `loki.secretfilter` unconditionally.
- `loki.secretfilter` uses Gitleaks.
- Gitleaks has two regex implementations:
  - `regexp/stdlib_regex.go` for `!gore2regex`, using Go stdlib `regexp`
  - `regexp/wasilibs_regex.go` for `gore2regex`, using `github.com/wasilibs/go-re2`

This means stock Alloy appears to load the go-re2/wazero path even if a deployment does not configure `loki.secretfilter`.

### Steps to reproduce

1. Install the official Alloy v1.16.1 Linux package.
2. Use a systemd unit with `MemoryDenyWriteExecute=yes`.
3. Run a config that does not use `loki.secretfilter` but does use normal Loki file/log pipelines.
4. Start Alloy.

### Expected behavior

Alloy should be able to run without executable memory mappings when the configured pipeline does not require JIT/WASM/native-code regex support.

### Actual behavior

Alloy requires relaxing the unit to `MemoryDenyWriteExecute=no` because startup hits the wazero/go-re2 executable mmap path.

### Requested fix / options

Could Alloy provide a W^X-compatible official build path or runtime behavior? Possible options:

- Do not force `gore2regex` for all release builds, or publish a secondary package built without it.
- Avoid importing/initializing `loki.secretfilter` unless it is configured.
- Make `loki.secretfilter` / Gitleaks use stdlib regex unless explicitly configured for go-re2.
- Document that official Alloy packages require executable memory mappings due to go-re2/wazero, so hardened systemd units should set `MemoryDenyWriteExecute=no`.

### Software version

Alloy v1.16.1

### System information

Linux systemd deployment with a hardened unit using `MemoryDenyWriteExecute=yes`.

### Configuration

The affected deployment does not use `loki.secretfilter`; it uses Alloy as a host log collector forwarding to Loki.

kalleep:
### Pull Request Details
We compile alloy with `gore2regex`. This causes https://github.com/gitleaks/gitleaks to use github.com/wasilibs/go-re2 instead of stdlib regex, which have been proven faster when using `loki.secretfilter`. 

But with the version we have the [wasm runtime is always created](https://github.com/wasilibs/go-re2/blob/v1.9.0/internal/re2_wazero.go#L162-L215). Regardless if this component is used or not. In [v1.10.0](https://github.com/wasilibs/go-re2/releases/tag/v1.10.0) they removed the init function and instead lazy load the runtime on first regex creation. 

With this alloy will only create this runtime if `loki.secretfilter` is used.

Fixes: https://github.com/grafana/alloy/issues/6289

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/collector/go.mod
+++ b/collector/go.mod
@@ -918,7 +918,7 @@ require (
 	github.com/vishvananda/netns v0.0.5 // indirect
 	github.com/vmware/govmomi v0.53.1 // indirect
 	github.com/vultr/govultr/v3 v3.31.1 // indirect
-	github.com/wasilibs/go-re2 v1.9.0 // indirect
+	github.com/wasilibs/go-re2 v1.10.0 // indirect
 	github.com/wasilibs/wazero-helpers v0.0.0-20240620070341-3dff1577cd52 // indirect
 	github.com/webdevops/azure-metrics-exporter v0.0.0-20250720221040-5092ac086990 // indirect
 	github.com/webdevops/go-common v0.0.0-20250720220738-a468ce107f36 // indirect
--- a/collector/go.sum
+++ b/collector/go.sum
@@ -2479,8 +2479,8 @@ github.com/vmware/govmomi v0.53.1 h1:6cNudrmbkYTp2onqSLTIsPwP2zj6Fkfb4ymNn7GO5dk
 github.com/vmware/govmomi v0.53.1/go.mod h1:EWfuzPfxT5NV+aS2we02SLFdhvJkgeY7t7+TszgBSMY=
 github.com/vultr/govultr/v3 v3.31.1 h1:AoJRZ0WDS1J1otp2wk3OD4aYh4EQxFU+kvyOPdCe4+Y=
 github.com/vultr/govultr/v3 v3.31.1/go.mod h1:2zyUw9yADQaGwKnwDesmIOlBNLrm7edsCfWHFJpWKf8=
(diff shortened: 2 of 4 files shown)


ptodev (approved):
It'd have been nice to have a fallback for this so that it uses a less efficient regex package but the component still works 🤔 But it's of course not necessary for this PR

Issue #6488, opened 2026-06-12
[Tracking Issue] Release 1.18


kgeckhart:
### Version

1.18

Due to OTel and Prometheus being deeply intertwinded recently we have not been able to do a dependency bump of either. 

The goal for this release will be to get everything upgraded and new features from the upgrades supported.

1. Start by upgrading prometheus to non-GA and otel to 1.151 #6240 (this is right before prometheus updated their own OTel version to 1.153.0)
2. Upgrade OTel to at least 1.153.0 (likely 154 if it's a small uplift) and prometheus to the GA changset.
3. Smaller PRs to introduce new prometheus and OTel features

### Release Candidate Date

2026-07-14 

### Stable Release Date

2026-07-20

kgeckhart:
Okay after doing some digging we will land at Prometheus v3.12.0 and OTel v0.153. OTel v0.154's is coupled Prometheus v3.13.0 which only has an RC. Next release we can try to target Prom v3.13.0 and OTel v0.154.0 (or higher).

blewis12:
Bumping the release dates to 2026-07-6 (for RC) and 2026-07-08 (for actual release). We're a bit behind due to dependency upgrades taking time, but almost there

blewis12:
Bumping once again to 2026-07-14 (for RC) and 2026-07-16 (for actual release) - will update here if anything changes

blewis12:
One final bump :) RC is released internally and marinading in our clusters. So far so good, if that continues we can cut the final release next monday

blewis12:
Released! https://github.com/grafana/alloy/releases/tag/v1.18.0

PR #6597, opened 2026-06-25
feat!: Upgrade Prometheus to v3.12.0 (GA) and OpenTelemetry Collector to v0.153.0

kgeckhart:

Bumps Prometheus to GA `v3.12.0` and the OpenTelemetry Collector stack to v0.153.0 (contrib + core `v1.59.0`), the follow-up to #6240.

### Pull Request Details

The `prometheus-operator` fork stays, since otel-contrib is still on `client-go` v0.35.4 through v0.154. OTel 153 is as far as we can go due to further version coupling between Prom and OTel (https://github.com/grafana/alloy/issues/6488#issuecomment-4802619432)

BREAKING-CHANGE: `otelcol.receiver.kafka` and `otelcol.exporter.kafka`: `resolve_canonical_bootstrap_servers_only` is now a no-op and deprecated. The upstream migration from migration from `sarama` to `franz-go dropped the behavior. The argument still parses but no longer resolves bootstrap server addresses.

BREAKING-CHANGE: `otelcol.processor.filter` and `otelcol.processor.transform`: `error_mode` now defaults to `ignore` instead of `propagate`, matching upstream OpenTelemetry Collector v0.153. Statement and condition errors are now logged and skipped rather than dropping the payload. Set `error_mode = "propagate"` to keep the previous behavior.

BREAKING-CHANGE: `otelcol.processor.k8sattributes`: `deployment_name_from_replicaset` now defaults to `true` instead of `false`, matching upstream v0.153. The processor derives the deployment name by trimming the Pod template hash and no longer watches ReplicaSets. Set `deployment_name_from_replicaset = false` in the `extract` block to keep the previous behavior.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.connector.spanmetrics.md
```diff
@@ -177,7 +177,9 @@ The default dimensions are:
 * `span.kind`
 * `status.code`
 
-The default dimensions are always added if not listed in `exclude_dimensions`. If no additional dimensions are specified, only the default ones will be added.
+The default dimensions are always added if not listed in `exclude_dimensions`. If no additional dimensions are specified, only the default ones are added.
+
+Alloy always excludes the `collector.instance.id` dimension. The connector adds it by default, but Alloy doesn't set a stable `collector.instance.id`, so it would otherwise be a random value that changes on every restart.
```
clayton-cornell:
```suggestion
{{< param "PRODUCT_NAME" >}} adds the default dimensions unless you list them in `exclude_dimensions`.
If you don't specify additional dimensions, {{< param "PRODUCT_NAME" >}} adds only the default ones.

{{< param "PRODUCT_NAME" >}} always excludes the `collector.instance.id` dimension.
The connector adds it by default, but {{< param "PRODUCT_NAME" >}} doesn't set a stable `collector.instance.id`.
Without this exclusion, the value changes on every restart.
```

docs/sources/reference/components/otelcol/otelcol.processor.k8sattributes.md
```diff
@@ -196,7 +196,9 @@ By default, if `metadata` isn't specified, the following fields are extracted an
 
 When `otel_annotations` is set to `true`, annotations such as `resource.opentelemetry.io/exampleResource` will be translated to the `exampleResource` resource attribute, etc.
 
-When `deployment_name_from_replicaset` is set to `true`, the processor will extract deployment name from replicaset name by trimming pod template hash. This will disable watching for replicaset resources, which can be useful in environments with limited RBAC permissions as the processor will not need `get`, `watch`, and `list` permissions for replicasets.
+When `deployment_name_from_replicaset` is set to `true`, the processor extracts the deployment name from the ReplicaSet name by trimming the Pod template hash. This disables watching for ReplicaSet resources, which can be useful in environments with limited RBAC permissions as the processor doesn't need `get`, `watch`, and `list` permissions for ReplicaSets.
+
+The Alloy default remains `false`, which differs from the upstream default of `true` since v0.153.
```
clayton-cornell:
```suggestion
The {{< param "PRODUCT_NAME" >}} default remains `false`, which differs from the upstream default of `true` since v0.153.
```

docs/sources/reference/components/otelcol/otelcol.connector.spanmetrics.md
dehaansa: "The connector adds it by default" is not super clear to me, perhaps "in other OpenTelemetry distributions  `collector.instance.id` is a default dimension"

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -51,7 +51,7 @@ You can use the following arguments with `otelcol.exporter.kafka`:
 | `partition_logs_by_resource_attributes`    | `bool`         | Whether to include the hash of sorted resource attributes as the message partitioning key in log messages sent to Kafka.    | `false`              | no       |
 | `partition_logs_by_trace_id`               | `bool`         | Whether to use the 16-bit hex string of the trace ID as the message partitioning key in log messages sent to Kafka.         | `false`              | no       |
 | `partition_traces_by_id`                   | `bool`         | Whether to include the trace ID as the message key in trace messages sent to Kafka.                                         | `false`              | no       |
-| `resolve_canonical_bootstrap_servers_only` | `bool`         | Whether to resolve then reverse-lookup broker IPs during startup.                                                           | `false`              | no       |
+| `resolve_canonical_bootstrap_servers_only` | `bool`         | Whether to resolve then reverse-lookup broker IP addresses during startup. Deprecated: now a no-op after the upstream Kafka client migration. | `false`              | no       |
```
dehaansa: "upstream Kafka client migration" could use a link, or a note, or more detail "...migration from Sarama to franz-go"

dehaansa (dismissed):
Couple of nits

docs/sources/reference/components/otelcol/otelcol.connector.spanmetrics.md
kgeckhart: The interesting part here is it doesn't look like it's actually set by any distribution. I'm working on a full repro for an upstream issue because it looks like it's going to have the same churning effect.

kgeckhart:
Comments are resolved thank you! I also backed tracked on taking two changes from upstream that are now documented as breaking changes. Unlike the otelcol.connector.spanmetrics `collector.instance.id` change, these weren't harmful upstream changed their default and it's not great for us to diverge.

blewis12 (approved):

PR #6598, opened 2026-06-25
docs: Add common traps section to the key-deps update guide

kgeckhart:

Adds a `Common traps and hard-won lessons` section to the key-deps update guide, folding the judgment from recent and historical Prometheus/OTel/Loki upgrades into the guide so the next upgrade agent doesn't relearn the traps the mechanical steps don't capture.

### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/developer/key-deps-update/key-dep-updates.md
+++ b/docs/developer/key-deps-update/key-dep-updates.md
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
    - Depends on: OpenTelemetry Collector Contrib (internal/exp/metrics, pkg/pdatautil, processor/deltatocumulativeprocessor)
    - Does NOT depend on: Beyla
 
+## Common traps and hard-won lessons
+
+These are the judgment calls that the mechanical steps below don't capture — failure modes that recur across upgrades and cost time to relearn. Each notes the step it reinforces.
+
+### Versions and relationships (Step 2)
+
+- **Compatibility cliffs — sometimes no released version works.** Cross-dependency API coupling can leave no *released* version of one key dep compiling against any single release of another. Pin a pre-release tag or a `main`-branch snapshot to bridge them; prefer a tagged RC over a raw snapshot. *Example:* otel-contrib v0.147–v0.149 calls the old 5-arg `scrape.NewManager` while v0.150+ needs the new `AppendableV2` signature (prom #17872), so no released prom v0.311.x fit either — bridged with `prometheus v0.312.0-rc.0`.
+
+- **A GA release can re-couple the stack.** A GA cut often runs a blanket CVE-driven `go get -u` that floats the whole OTel stack to a newer minor, so the last RC before GA can be the cleaner pin. *Example:* prom GA `v0.312.0` pulled deltatocumulative→v0.153 and collector core→v1.59 via a 46-dep refresh (prometheus#18812); `v0.312.0-rc.0` stayed v0.151-aligned.
+
+- **Forced minimum bumps carry their own breaks.** An older version of a key dep may import a package a newer transitive dep removed, forcing a minimum bump — and that bump can carry an unrelated user-facing breaking change. *Example:* Loki v3.6.x imports the removed `tsdb/errors` package → forced to v3.7.1, which also brings loki#19991 (`fix!`: parsed labels should not override structured metadata).
+
+### Forks (Step 3)
+
+- **A minimal one-commit fork can be lower-risk than upstream main.** The reflex "drop the fork, point at upstream main" frequently breaks the build, because main has moved an unrelated transitive dep the rest of the graph isn't ready for. A single-patch fork that stays on the graph's version line is sometimes the safer choice. Remove it on a coordinated, graph-wide uplift — not because upstream tagged a release. *Example:* prom-operator main has the slog `rulefmt.Parse` fix but also bumped `client-go` to v0.36.2 (`HasSyncedChecker`), breaking pinned `controller-runtime v0.20.4` and otel-contrib's `k8sattributesprocessor` (still client-go v0.35.4 through otel v0.154) — kept a 1-line `v0.91.0`+slog fork instead.
+
+### Modules and conflict resolution (Step 4)
+
+- **Don't let `tidy` downgrade main.** After a rebase or merge, `go mod tidy` can silently select versions *below* what main already requires (renovate bumps). Reconcile take-higher per module and audit explicitly for downgrades before proceeding — `tidy` won't raise a direct dep back up on its own. *Example:* one rebase surfaced 16 such traps (secretsmanager, pgx/v5, go-git, zerolog, …).
+
+- **Hold the OTel stack uniform.** When main introduced a new OTel component since your branch diverged, pin it to your target version in `collector/builder-config.yaml` so `tidy` doesn't float the whole stack to a newer minor. After regenerating, grep for any otel module off the uniform version and bump the stragglers. *Example:* `nginxreceiver`, `signaltometricsconnector`, and a few collector-core modules came in at the old minor and had to be pinned to v0.151.0.
+
+- **Bump dependencies at their source of truth, not the generated file.** Forcing a version straggler directly in `go.mod` when `collector/builder-config.yaml` owns it gets reverted on the next regeneration and surfaces as `make generate` drift. Fix it in the builder-config. *Example:* the lone `httpsprovider` straggler was raised in builder-config, not go.mod, to keep the stable stack uniform without drift.
+
+- **Verify the commit, not just the working tree.** `go build`, `test`, and `lint` run against the working tree, so they stay green even when a merge or amend captured pre-regeneration files. Stage regenerated files *after* regenerating, and check `git status` before pushing. *Example:* a merge commit once captured a stale `components.go` (missing a connector) while every local check passed against the regenerated tree.
+
+### Compilation (Step 6)
+
+- **Build the upstream Config from `CreateDefaultConfig()`, not a zero-value struct.** When upstream adds a new *required* field with a non-zero factory default, a fresh struct ships a zero value that breaks at runtime. ~39 of ~50 otelcol components build from a fresh struct, so this recurs. *Example:* googlecloudpubsub `flow_control` shipped zeros → `StreamAckDeadline=0` rejected by the API and `TriggerAckBatchDuration=0` hot-loops.
+
+### Tests, docs, and change notes (Steps 7, 9, and 10)
+
+- **Newly-required upstream config: expose it, don't hardcode.** If upstream makes a field required, expose it as Alloy config defaulting to the upstream factory value (zero behavior change unless set) rather than hardcoding the default. Keep scope tight: defer *purely-optional* new fields to a follow-up. *Example:* exposed kafka `record_partitioner` and googlecloudpubsub `flow_control`, parked the optional ones; tail_sampling `sampling_strategy` couldn't be exposed at all — upstream keeps the type unexported through v0.154, so it needs a reflection shim.
+
+- **A major bump can regress runtime *defaults*, not just APIs.** Even when nothing fails to compile, check that default behaviors still match the prior major — build config via `Load` on empty input so populated defaults (like the scrape fallback protocol) survive. *Example:* after the Prometheus v3 bump, `prometheus.operator.*` scraping silently stopped for ServiceMonitors with no explicit protocol (#4291).
+
+- **A changed test expectation is a behavior change until proven otherwise.** It usually means a user-facing change (default flip, type rename) that needs a release note — surface it, don't just update the fixture. Conversely, a changelog-flagged default flip that Alloy explicitly overrides is *no-impact* — say so rather than assuming impact. *Examples:* v0.153 flipped k8sattributes `deployment_name_from_replicaset` to true (note it); the filter/transform `error_mode` propagate→ignore flip has no impact because Alloy defaults to `PropagateError`.
+
+- **Dropping a fork whose patch landed upstream is not a user-facing change.** When you retire a fork because its fix is now in the target version, behavior is continuous across the swap — don't flag it in the release notes. Verify continuity before calling something a break. *Example:* dropping the `grafana/prometheus` fork looked like a `sent_batch_duration` change but wasn't — both the fork and the target rc.0 carried the same fix.
+
+- **Flag no-op-upstream fields as deprecations.** A field that still parses but does nothing upstream now needs a deprecation note, not silent removal. *Example:* kafka `resolve_canonical_bootstrap_servers_only` became a no-op after the franz-go migration.
+
+- **Scope release notes to components actually in the released distro.** Before calling a change user-facing, check the component even ships in the latest release — skip notes for ones not yet exposed. *Example:* the `otelcol.receiver.awss3` change was left out of the changelog because the component wasn't in the latest release (#5128).
+
+- **Link component docs to the `OTEL_VERSION` variable, not a hardcoded tag.** When upgrade docs reference upstream source, use `{{< param "OTEL_VERSION" >}}` so the link tracks each bump instead of rotting. *Example:* raised in review on `otelcol.processor.filter.md` (#5784); the templated form is the repo convention, though the stale `<OTEL_VERSION>` angle-bracket form still lingers in shared links.
+
 ## Steps to update key dependencies
 
 Don't write anything in this document. Create a separate document called 'deps-update-<version>-YYYY-MM-DD.md' in the root of the repository. This is your output file where you will write all the output as required.
--- a/docs/developer/key-deps-update/key-dep-updates.md
+++ b/docs/developer/key-deps-update/key-dep-updates.md
@@ -128,49 +128,97 @@ Here is a summary of the relationships between the key dependencies of Alloy.
 
 ## Common traps and hard-won lessons
 
-These are the judgment calls that the mechanical steps below don't capture — failure modes that recur across upgrades and cost time to relearn. Each notes the step it reinforces.
+The bullets here are judgment calls the "Steps to update key dependencies" section does not spell out. They are recurring failure modes that cost time to relearn on every run. Each subsection names which step they apply to.
 
 ### Versions and relationships (Step 2)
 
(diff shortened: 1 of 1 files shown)


docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **Dropping a fork whose patch landed upstream is not a user-facing change.** When you retire a fork because its fix is now in the target version, behavior is continuous across the swap — don't flag it in the release notes. Verify continuity before calling something a break. *Example:* dropping the `grafana/prometheus` fork looked like a `sent_batch_duration` change but wasn't — both the fork and the target rc.0 carried the same fix.
+
+- **Flag no-op-upstream fields as deprecations.** A field that still parses but does nothing upstream now needs a deprecation note, not silent removal. *Example:* kafka `resolve_canonical_bootstrap_servers_only` became a no-op after the franz-go migration.
+
+- **Scope release notes to components actually in the released distro.** Before calling a change user-facing, check the component even ships in the latest release — skip notes for ones not yet exposed. *Example:* the `otelcol.receiver.awss3` change was left out of the changelog because the component wasn't in the latest release (#5128).
```
clayton-cornell:
```suggestion
- **Scope release notes to components actually in the released distro.** Before calling a change user-facing, check that the component even ships in the latest release — skip notes for ones not yet exposed. *Example:* the `otelcol.receiver.awss3` change was left out of the changelog because the component wasn't in the latest release (#5128).
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
    - Depends on: OpenTelemetry Collector Contrib (internal/exp/metrics, pkg/pdatautil, processor/deltatocumulativeprocessor)
    - Does NOT depend on: Beyla
 
+## Common traps and hard-won lessons
+
+These are the judgment calls that the mechanical steps below don't capture — failure modes that recur across upgrades and cost time to relearn. Each notes the step it reinforces.
```
clayton-cornell:
```suggestion
The bullets here are judgment calls the "Steps to update key dependencies" section does not spell out. They are recurring failure modes that cost time to relearn on every run. Each subsection names which step they apply to.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+### Versions and relationships (Step 2)
+
+- **Compatibility cliffs — sometimes no released version works.** Cross-dependency API coupling can leave no *released* version of one key dep compiling against any single release of another. Pin a pre-release tag or a `main`-branch snapshot to bridge them; prefer a tagged RC over a raw snapshot. *Example:* otel-contrib v0.147–v0.149 calls the old 5-arg `scrape.NewManager` while v0.150+ needs the new `AppendableV2` signature (prom #17872), so no released prom v0.311.x fit either — bridged with `prometheus v0.312.0-rc.0`.
```
clayton-cornell:
```suggestion
- **Compatibility cliffs — sometimes no released version works.** Cross-dependency API coupling can leave no *released* version of one key dependency compiling against any single release of another. Pin a pre-release tag or a `main`-branch snapshot to bridge them. Prefer a tagged RC over a raw snapshot. *Example:* otel-contrib v0.147–v0.149 calls the old 5-arg `scrape.NewManager` while v0.150+ needs the new `AppendableV2` signature (prom #17872), so no released prom v0.311.x fit either. Bridged with `prometheus v0.312.0-rc.0`.
```

Would it make more sense to split the solutions off to sub-bullets here to make them stand out a bit more vs being lost in the paragraph?

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **A GA release can re-couple the stack.** A GA cut often runs a blanket CVE-driven `go get -u` that floats the whole OTel stack to a newer minor, so the last RC before GA can be the cleaner pin. *Example:* prom GA `v0.312.0` pulled deltatocumulative→v0.153 and collector core→v1.59 via a 46-dep refresh (prometheus#18812); `v0.312.0-rc.0` stayed v0.151-aligned.
+
+- **Forced minimum bumps carry their own breaks.** An older version of a key dep may import a package a newer transitive dep removed, forcing a minimum bump — and that bump can carry an unrelated user-facing breaking change. *Example:* Loki v3.6.x imports the removed `tsdb/errors` package → forced to v3.7.1, which also brings loki#19991 (`fix!`: parsed labels should not override structured metadata).
```
clayton-cornell:
```suggestion
- **Forced minimum bumps carry their own breaking changess.** An older version of a key dependency may import a package a newer transitive dependency removed, forcing a minimum bump. That bump can carry an unrelated user-facing breaking change. *Example:* Loki v3.6.x imports the removed `tsdb/errors` package → forced to v3.7.1, which also brings loki#19991 (`fix!`: parsed labels should not override structured metadata).
```

We should try to avoid using `→`  Are you trying to say `so we had to move to v3.7.1`?

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+### Forks (Step 3)
+
+- **A minimal one-commit fork can be lower-risk than upstream main.** The reflex "drop the fork, point at upstream main" frequently breaks the build, because main has moved an unrelated transitive dep the rest of the graph isn't ready for. A single-patch fork that stays on the graph's version line is sometimes the safer choice. Remove it on a coordinated, graph-wide uplift — not because upstream tagged a release. *Example:* prom-operator main has the slog `rulefmt.Parse` fix but also bumped `client-go` to v0.36.2 (`HasSyncedChecker`), breaking pinned `controller-runtime v0.20.4` and otel-contrib's `k8sattributesprocessor` (still client-go v0.35.4 through otel v0.154) — kept a 1-line `v0.91.0`+slog fork instead.
```
clayton-cornell:
```suggestion
- **A minimal one-commit fork can be lower-risk than upstream main.** The reflex "drop the fork, point at upstream main" frequently breaks the build, because main has moved an unrelated transitive dependency the rest of the graph isn't ready for. A single-patch fork that stays on the graph's version line is sometimes the safer choice. Remove it on a coordinated graph-wide uplift, not because upstream tagged a release. *Example:* prom-operator main has the slog `rulefmt.Parse` fix but also bumped `client-go` to v0.36.2 (`HasSyncedChecker`), breaking pinned `controller-runtime v0.20.4` and otel-contrib's `k8sattributesprocessor` (still client-go v0.35.4 through otel v0.154) — kept a 1-line `v0.91.0`+slog fork instead.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+### Modules and conflict resolution (Step 4)
+
+- **Don't let `tidy` downgrade main.** After a rebase or merge, `go mod tidy` can silently select versions *below* what main already requires (renovate bumps). Reconcile take-higher per module and audit explicitly for downgrades before proceeding — `tidy` won't raise a direct dep back up on its own. *Example:* one rebase surfaced 16 such traps (secretsmanager, pgx/v5, go-git, zerolog, …).
```
clayton-cornell:
```suggestion
- **Don't let `tidy` downgrade main.** After a rebase or merge, `go mod tidy` can silently select versions *below* what main already requires (renovate bumps). Take the higher version per module, reconcile, and check explicitly for downgrades before you continue. `tidy` won't raise a direct dependency back up on its own. *Example:* one rebase surfaced 16 such traps (secretsmanager, pgx/v5, go-git, zerolog, …).
```

I struggled with what `Reconcile take` meant here.

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **Hold the OTel stack uniform.** When main introduced a new OTel component since your branch diverged, pin it to your target version in `collector/builder-config.yaml` so `tidy` doesn't float the whole stack to a newer minor. After regenerating, grep for any otel module off the uniform version and bump the stragglers. *Example:* `nginxreceiver`, `signaltometricsconnector`, and a few collector-core modules came in at the old minor and had to be pinned to v0.151.0.
+
+- **Bump dependencies at their source of truth, not the generated file.** Forcing a version straggler directly in `go.mod` when `collector/builder-config.yaml` owns it gets reverted on the next regeneration and surfaces as `make generate` drift. Fix it in the builder-config. *Example:* the lone `httpsprovider` straggler was raised in builder-config, not go.mod, to keep the stable stack uniform without drift.
```
clayton-cornell:
```suggestion
- **Bump dependencies at their source of truth, not the generated file.** If you force a version straggler in `go.mod` when `collector/builder-config.yaml` owns it, the next regenerate will revert it and it surfaces as `make generate` drift. Fix it in the builder-config. *Example:* the lone `httpsprovider` straggler was raised in builder-config, not go.mod, to keep the stable stack uniform without drift.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **Verify the commit, not just the working tree.** `go build`, `test`, and `lint` run against the working tree, so they stay green even when a merge or amend captured pre-regeneration files. Stage regenerated files *after* regenerating, and check `git status` before pushing. *Example:* a merge commit once captured a stale `components.go` (missing a connector) while every local check passed against the regenerated tree.
```
clayton-cornell:
```suggestion
- **Verify the commit, not just the working tree.** `go build`, `test`, and `lint` run against the working tree, so they stay green even when a merge or amend captured pre-regeneration files. Regenerate the files first, then stage the updated files, and check `git status` before you push. *Example:* a merge commit once captured a stale `components.go` (missing a connector) while every local check passed against the regenerated tree.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+### Compilation (Step 6)
+
+- **Build the upstream Config from `CreateDefaultConfig()`, not a zero-value struct.** When upstream adds a new *required* field with a non-zero factory default, a fresh struct ships a zero value that breaks at runtime. ~39 of ~50 otelcol components build from a fresh struct, so this recurs. *Example:* googlecloudpubsub `flow_control` shipped zeros → `StreamAckDeadline=0` rejected by the API and `TriggerAckBatchDuration=0` hot-loops.
```
clayton-cornell:
```suggestion
- **Build the upstream Config from `CreateDefaultConfig()`, not a zero-value struct.** When upstream adds a new *required* field with a non-zero factory default, a fresh struct ships a zero value that breaks at runtime. Roughly 39 of 50 otelcol components build from a fresh struct, so this recurs. *Example:* googlecloudpubsub `flow_control` shipped zeros, so `StreamAckDeadline=0` was rejected by the API and `TriggerAckBatchDuration=0` caused a hot-loop.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+### Tests, docs, and change notes (Steps 7, 9, and 10)
+
+- **Newly-required upstream config: expose it, don't hardcode.** If upstream makes a field required, expose it as Alloy config defaulting to the upstream factory value (zero behavior change unless set) rather than hardcoding the default. Keep scope tight: defer *purely-optional* new fields to a follow-up. *Example:* exposed kafka `record_partitioner` and googlecloudpubsub `flow_control`, parked the optional ones; tail_sampling `sampling_strategy` couldn't be exposed at all — upstream keeps the type unexported through v0.154, so it needs a reflection shim.
```
clayton-cornell:
```suggestion
- **Newly-required upstream config: expose it, don't hardcode.** If upstream makes a field required, expose it as Alloy config defaulting to the upstream factory value (zero behavior change unless set) rather than hardcoding the default. Keep the scope tight. Defer *purely-optional* new fields to a follow-up. *Example:* exposed kafka `record_partitioner` and googlecloudpubsub `flow_control`, parked the optional ones. tail_sampling `sampling_strategy` couldn't be exposed at all because upstream keeps the type unexported through v0.154, so it needs a reflection shim.
```

What is "tail_sampling `sampling_strategy`" here? Specifically tail_sampling... should that be code formatted? Or written differently?

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **A major bump can regress runtime *defaults*, not just APIs.** Even when nothing fails to compile, check that default behaviors still match the prior major — build config via `Load` on empty input so populated defaults (like the scrape fallback protocol) survive. *Example:* after the Prometheus v3 bump, `prometheus.operator.*` scraping silently stopped for ServiceMonitors with no explicit protocol (#4291).
```
clayton-cornell:
```suggestion
- **A major bump can regress runtime *defaults*, not just APIs.** Even when everything compiles, check that default behaviors still match the prior major. Build config via `Load` on empty input so populated defaults (like the scrape fallback protocol) survive. *Example:* after the Prometheus v3 bump, `prometheus.operator.*` scraping silently stopped for ServiceMonitors with no explicit protocol (#4291).
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **A changed test expectation is a behavior change until proven otherwise.** It usually means a user-facing change (default flip, type rename) that needs a release note — surface it, don't just update the fixture. Conversely, a changelog-flagged default flip that Alloy explicitly overrides is *no-impact* — say so rather than assuming impact. *Examples:* v0.153 flipped k8sattributes `deployment_name_from_replicaset` to true (note it); the filter/transform `error_mode` propagate→ignore flip has no impact because Alloy defaults to `PropagateError`.
```
clayton-cornell:
```suggestion
- **A changed test expectation is a behavior change until proven otherwise.** It usually means a user-facing change (default flip, type rename) that needs a release note. Document it in the release notes, don't just update the fixture. Conversely, a changelog-flagged default flip that Alloy explicitly overrides is *no-impact*. Document this in the release notes rather than assuming impact. *Examples:* v0.153 flipped k8sattributes `deployment_name_from_replicaset` to true (note it); the filter/transform `error_mode` propagate→ignore flip has no impact because Alloy defaults to `PropagateError`.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **Dropping a fork whose patch landed upstream is not a user-facing change.** When you retire a fork because its fix is now in the target version, behavior is continuous across the swap — don't flag it in the release notes. Verify continuity before calling something a break. *Example:* dropping the `grafana/prometheus` fork looked like a `sent_batch_duration` change but wasn't — both the fork and the target rc.0 carried the same fix.
```
clayton-cornell:
```suggestion
- **Dropping a fork whose patch landed upstream is not a user-facing change.** When you retire a fork because its fix is now in the target version, behavior is continuous across the swap. Don't flag this in the release notes. Verify continuity before calling something a breaking change. *Example:* dropping the `grafana/prometheus` fork looked like a `sent_batch_duration` change but wasn't. Both the fork and the target rc.0 carried the same fix.
```

docs/developer/key-deps-update/key-dep-updates.md
```diff
@@ -126,6 +126,52 @@ Here is a summary of the relationships between the key dependencies of Alloy.
...
+
+- **Flag no-op-upstream fields as deprecations.** A field that still parses but does nothing upstream now needs a deprecation note, not silent removal. *Example:* kafka `resolve_canonical_bootstrap_servers_only` became a no-op after the franz-go migration.
+
+- **Scope release notes to components actually in the released distro.** Before calling a change user-facing, check the component even ships in the latest release — skip notes for ones not yet exposed. *Example:* the `otelcol.receiver.awss3` change was left out of the changelog because the component wasn't in the latest release (#5128).
+
+- **Link component docs to the `OTEL_VERSION` variable, not a hardcoded tag.** When upgrade docs reference upstream source, use `{{< param "OTEL_VERSION" >}}` so the link tracks each bump instead of rotting. *Example:* raised in review on `otelcol.processor.filter.md` (#5784); the templated form is the repo convention, though the stale `<OTEL_VERSION>` angle-bracket form still lingers in shared links.
```
clayton-cornell:
```suggestion
- **Link component docs to the `OTEL_VERSION` variable, not a hardcoded tag.** When upgrade docs reference upstream source, use `{{< param "OTEL_VERSION" >}}` so the link tracks each bump instead of rotting. *Example:* raised in review on `otelcol.processor.filter.md` (#5784), the templated form is the repo convention, though the stale `<OTEL_VERSION>` angle-bracket form still lingers in shared links.
```

clayton-cornell (commented):
We should drop all en/em dashes in the docs.  A few other minor tweaks and clarification questions.

docs/developer/key-deps-update/key-dep-updates.md
kgeckhart: Yes, reworded it to "forcing a bump to v3.7.1"

docs/developer/key-deps-update/key-dep-updates.md
kgeckhart: Good point, tail_sampling should have been code formatted here and I changed it to the `tail_sampling` processor's `sampling_strategy`

docs/developer/key-deps-update/key-dep-updates.md
kgeckhart: I ended up rewording this because I think it's somewhat contradicting. I think the release notes should capture user facing changes. If an upstream default changed but we chose not to inherit I ended up noting it in the component docs as nothing changes for our users.

blewis12 (dismissed):

thampiotr (dismissed):
Nice. I like that these were kept short and feels like LLM will make sense of them.

blewis12 (approved):

PR #6599, opened 2026-06-25
fix(loki.process): Log logfmt decode failures at debug instead of error

mstumpfx:

Lowers the log level of logfmt decode failures in the `loki.process` logfmt stage from `error` to `debug`. A single malformed log line is bad input, not an Alloy fault, so an unparseable entry shouldn't spam the logs at error level.

Somewhat related, the JSON parser is set to debug when those logs fail to parse (https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/json.go#L140-L145)

### Pull Request Details

https://github.com/grafana/deployment_tools/pull/625507#discussion_r3472542269

> So when parsing json alloy only ever emits debug logs, even when we fail to parse a line. But logfmt is a problem since it will produce error logs for every failed parse, we should probably change that to debug as well so you don't have to do this workaround just to reduce noisy logs

N/A

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/logfmt.go
+++ b/internal/component/loki/process/stages/logfmt.go
@@ -127,7 +127,7 @@ func (j *logfmtStage) Process(labels model.LabelSet, extracted map[string]any, t
 	}
 
 	if decoder.Err() != nil {
-		j.logger.Error("failed to decode logfmt", "err", decoder.Err())
+		j.logger.Debug("failed to decode logfmt", "err", decoder.Err())
 		return
 	}
 


internal/component/loki/process/stages/logfmt.go
```diff
@@ -127,7 +127,7 @@ func (j *logfmtStage) Process(labels model.LabelSet, extracted map[string]any, t
 	}
 
 	if decoder.Err() != nil {
-		j.logger.Error("failed to decode logfmt", "err", decoder.Err())
+		j.logger.Debug("failed to decode logfmt", "err", decoder.Err())
```
mstumpfx, line 130: Wasn't sure if this needed to be wrapped in a 
```
if debugEnabled(j.logger) {
...
}
```
like the JSON counterpart

kalleep: It's not needed, I have been meaning to get rid of most of those guards. It's only really necessary if building any of the args is expensive and here it's not

kalleep (approved):

PR #6610, opened 2026-06-29
feat: Add start_timestamp_zero_ingestion to Prometheus metric components

### LINKED ISSUE #4045: Support created-timestamp-zero-ingestion feature flag for prometheus.scrape component
### Request
Add support for the  `created-timestamp-zero-ingestion` feature flag in `prometheus.scrape` Alloy component.

### Context
Prometheus added an experimental support Add support zeros as [created timestamps](https://github.com/prometheus/prometheus/pull/14356), this is important to 
Otel collector has [adopted](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36660) this feature flag as well.

### Workaround
Current workaround is to use prometheus in agent mode to scape the data and enable that flag there.

 

### Use case

This feature allows better support for [extended range selectors semantics by Prometheus](https://github.com/prometheus/proposals/blob/main/proposals/2025-04-04_extended-range-selectors-semantics.md#qa).

madaraszg-tulip:

Adds an experimental `start_timestamp_zero_ingestion` argument to `otelcol.exporter.prometheus`, `prometheus.scrape`, and the `prometheus.operator.*` components. When enabled, a synthetic zero sample is injected at the start timestamp of cumulative metrics (counters, histograms, and summaries) ahead of the real sample, marking a counter reset so `rate` and friends account for the series' initial value. The zero sample is emitted only when a series first appears or its start timestamp advances, not on every sample, and gauges are unaffected.

### Pull Request Details

- **Static metrics WAL** (`internal/static/metrics/wal`): implements the previously-stubbed `AppendSTZeroSample` and `AppendHistogramSTZeroSample`. These were a no-op, so start-timestamp zero samples were silently dropped on the way to `prometheus.remote_write`. The implementation mirrors the Prometheus TSDB head: it returns `ErrSTNewerThanSample` when the start timestamp isn't older than the sample and `ErrOutOfOrderST` when it would be out of order, so the zero is injected exactly once per start-timestamp value (first appearance or counter reset).
- **`otelcol.exporter.prometheus`**: reads the OTLP start timestamp and injects the zero sample during conversion for counters, classic and native histograms, and summaries. Gauges are skipped.
- **`prometheus.scrape` and `prometheus.operator.*`**: enable `EnableStartTimestampZeroIngestion` (and `ParseST`) on the Prometheus scrape `Options`, which drives the scrape loop to parse the start timestamp and inject the zero via the WAL.
- The `testappender` and `collectingappender` test utilities now implement the start-timestamp methods (previously a panic / no-op), with the injected zeros recorded separately so they're observable in tests.
- The argument is experimental and requires `stability.level=experimental`, matching the existing experimental arguments on these components.

The argument is named `start_timestamp_zero_ingestion` to align with the `StartTimestamp` terminology Prometheus renamed the concept to. (The upstream `--enable-feature` flag keeps the old `created-timestamp-zero-ingestion` name for stability, but aligning the Alloy argument with the newer terminology keeps it futureproof.)

Fixes #4045

### Notes to the Reviewer

- The argument is off by default. The injected zero is an ordinary sample, so it works over both Remote Write 1.0 and 2.0.
- **Naming**: `start_timestamp_zero_ingestion` aligns with the `StartTimestamp` terminology Prometheus renamed the concept to (`AppendSTZeroSample`, `ParseST`). The upstream `--enable-feature` flag itself keeps the old `created-timestamp-zero-ingestion` name for stability.
- **Scrape options**: for the flow Alloy uses today (AppenderV1), `EnableStartTimestampZeroIngestion` alone is sufficient — the scrape manager ORs `ParseST` in. I also set `ParseST` explicitly to future-proof for the AppenderV2 migration, where `EnableStartTimestampZeroIngestion` becomes a no-op and is slated for removal. This mirrors what Prometheus' own `--enable-feature=created-timestamp-zero-ingestion` sets.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


internal/component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -358,9 +366,47 @@ func (conv *Converter) consumeGauge(app storage.Appender, memResource *memorySer
...
 
+// writeCreatedTimestampZeroSample injects a synthetic zero sample at the data
+// point's start timestamp ahead of the real sample, marking the point at which
+// the cumulative series started or reset. It is a no-op unless the feature is
+// enabled and the data point carries a start timestamp. The appender rejects
+// start timestamps that are not strictly older than the sample
+// (ErrSTNewerThanSample) or that would be out of order (ErrOutOfOrderST); both
+// are expected - the latter is the steady-state case for long-lived counters -
+// and are not logged.
+func (conv *Converter) writeCreatedTimestampZeroSample(app storage.Appender, series *memorySeries, dp otelcolDataPoint) {
```
ptodev: Does the OTLP translation work ok if there are multiple datapoints from the same series in the same batch, and every one of them has a start timestamp?

ptodev (commented):
@madaraszg-tulip thank you for the PR, and apologies for the late review! It mostly looks ok. I'm not too sure how the OTLP translations should work and whether it's ok for some edge cases. Did you copy the code from the translators that Contrib uses?

Also, would you mind to indeed rename "created" to "start" so that we align with the new terminology please? Thanks for pointing this out! I see that in the [upstream docs](https://prometheus.io/docs/prometheus/latest/feature_flags/#start-created-timestamps-zero-injection) this is also labeled as a deprecated name, and want us to be more futureproof in case there's a new feature flag upstream.

internal/component/otelcol/exporter/prometheus/internal/convert/convert.go
madaraszg-tulip: Good catch — it wasn't handled originally. The appender-side out-of-order guard doesn't cover this within a single batch, because the WAL's last-timestamp is only updated on commit, so every datapoint in the batch saw the same pre-batch value and each would emit its own zero.

Fixed in c633346c1: the converter now tracks the last injected start timestamp per series (`lastZeroST`) and injects the zero at most once per start timestamp. Multiple datapoints of the same series sharing a start timestamp now produce a single zero, while a genuine mid-batch counter reset (start timestamp advancing) still injects a new one. Added `TestConverterStartTimestampZeroIngestionDedup` covering exactly this case.

For what it's worth, this mirrors how Prometheus' own OTLP handler did it before the AppenderV2 switch — the `CombinedAppender` gated injection on `updateRefs := !exists || series.st != st` (i.e. new series or changed start timestamp), which is the same per-series dedup.

madaraszg-tulip:
Thanks for the review!

**Rename**: done in 78849520f — the argument (and struct fields, converter option, docs, tests) is now `start_timestamp_zero_ingestion`, matching the `StartTimestamp` terminology. (Note the upstream `--enable-feature` flag itself keeps the old `created-timestamp-zero-ingestion` name for stability — it's the concept that was renamed — but aligning our new option with the new terminology keeps it futureproof.)

**Did I copy from Contrib?** No — and there's nothing there to copy. Contrib's `prometheusremotewrite`/`prometheus` translators don't implement start/created-timestamp zero injection at all: the `_created` const in `helper.go` is dead code and there's no `StartTimeUnixNano` handling (confirmed on latest `main`, `9b25cc2bd64`) — they just drop the start timestamp.

The logic is modeled on Prometheus core instead. Alloy's exporter is still on AppenderV1, so I inject explicitly via `AppendSTZeroSample`, exactly like Prometheus' pre-AppenderV2 OTLP handler did (the `CombinedAppender`: `AppendSTZeroSample` gated by `st != 0 && st < t`, per-series dedup, OOO/duplicate errors ignored). Prometheus only moved this to AppenderV2 — passing `st` inline to `Append(ref, ls, st, t, …)` and letting the appender own the injection — recently (https://github.com/prometheus/prometheus/pull/17996, 2026-02-03, first in v0.310.0). Migrating Alloy's exporter to AppenderV2 would be the natural long-term follow-up, but it's a much larger change (the whole appender stack here is V1), so this PR stays on V1 and mirrors the established V1 pattern. The `st == t` "unknown start time" OTLP convention is handled correctly (no zero injected) via the `st < t` gate.

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
```diff
@@ -78,6 +79,17 @@ To enable and use an experimental feature, you must set the `stability.level` [f
...
+{{< admonition type="warning" >}}
+**EXPERIMENTAL**: The `start_timestamp_zero_ingestion` argument is an [experimental][] feature.
+
+When the `start_timestamp_zero_ingestion` argument is enabled, a zero sample is injected at the OTLP start timestamp for cumulative metrics, marking a counter reset so `rate` accounts for the initial value.
+
+To enable and use an experimental feature, you must set the `stability.level` [flag][] to `experimental`.
+
+[experimental]: https://grafana.com/docs/release-life-cycle/
+[flag]: https://grafana.com/docs/alloy/<ALLOY_VERSION>/reference/cli/run/
+{{< /admonition >}}
```
clayton-cornell, line 91:
```suggestion
> **EXPERIMENTAL**: The `start_timestamp_zero_ingestion` argument is an [experimental][] feature.
>
> When the `start_timestamp_zero_ingestion` argument is enabled, a zero sample is injected at the OTLP start timestamp for cumulative metrics, marking a counter reset so `rate` accounts for the initial value.
>
> Experimental features are subject to frequent breaking changes, and may be removed with no equivalent replacement.
> To enable and use an experimental feature, you must set the `stability.level` [flag][] to `experimental`.

[experimental]: https://grafana.com/docs/release-life-cycle/
[flag]: https://grafana.com/docs/alloy/<ALLOY_VERSION>/reference/cli/run/
```

We aren't using Warning or Caution admonitions for the Experimental notices.  To keep it consistent, use the `>` notation like is used in the prometheus.scrape topic.

ptodev: There are two `{{< admonition type="warning" >}}` for experimental features right above this one - for `honor_metadata` and for `convert_classic_histograms_to_nhcb`. We could merge the PR as it is, and change all 3 later?

ptodev (approved):
Thank you!

docs/sources/reference/components/otelcol/otelcol.exporter.prometheus.md
ptodev: @clayton-cornell  I merge it as it is so that it's consistent, I hope you don't mind 😅

madaraszg-tulip: I did the second one in a previous contribution, just followed the `honor_metadata` lead

clayton-cornell: I should have dug a little deeper before I comented.  

Dismiss the suggestion. Merge the PR with the text as-is.

I've opened: https://github.com/grafana/alloy/issues/6728 to fix the inconsistencies.

PR #6614, opened 2026-06-29
fix(local.file_match): Handle exclude paths case-insensitively on Windows and macOS

### LINKED ISSUE #6607: local.file_match: __path_exclude__ is matched case-sensitively while __path__ is case-insensitive on case-insensitive operating systems (Windows, macOS)
### Component(s)

local.file_match

### What's wrong?

On case-insensitive operating systems (Windows, macOS), local.file_match discovers files via `__path__` using case-insensitive glob matching that follows the OS's native path rules, but it evaluates `__path_exclude__` using a case-sensitive matcher.

Because the two labels disagree about case, an exclude pattern can silently fail to exclude files that __path__ happily discovered. Any path_targets entry whose directory or file casing differs from the on-disk casing will still discover files through `__path__`, while the otherwise-correct `__path_exclude__` matches nothing and the unwanted files are ingested anyway.

This is surprising because __path__ and __path_exclude__ accept the same glob syntax and are typically written with the identical path prefix, yet they behave inconsistently with respect to case.

## Expected behavior

`app.log` is discovered and tailed. `app-backup-2024.01.01.log` matches `*-backup*.log` and is excluded. Net result: only `app.log` is forwarded.

`__path_exclude__` should follow the operating system's native path semantics — the same rules used to discover `__path__`. On a case-insensitive OS, that means the exclude comparison should also be case-insensitive, so an exclude that lines up with a successful include reliably removes the same files.

## Actual behavior

Both files are discovered and tailed, and lines from `app-backup-2024.01.01.log` are forwarded.

`__path__` matches case-insensitively, so `Myservice` correctly resolves to the on-disk `MyService` and both files are found. But `__path_exclude__` is matched case-sensitively, so the pattern segment `Myservice\...\*-backup*.log` never matches the real path `...\MyService\...\app-backup-2024.01.01.log`, and the backup file is never excluded.

Note: the case typo in the example is only there to make the inconsistency easy to observe. The underlying defect is that discovery folds case (consistent with the OS) while the exclude check does not, so the two sides can always disagree on a case-insensitive filesystem.

### Steps to reproduce

1. On a Windows host (case-insensitive NTFS), create the following files:
   - `C:\AppLogs\MyService\Logs\app.log`
   - `C:\AppLogs\MyService\Logs\app-backup-2024.01.01.log`

2. Configure `local.file_match` so that one directory segment is written in a **different case** than what exists on disk — here the on-disk directory is `MyService` but the config uses `Myservice` (lowercase `s`) in **both** `__path__` and `__path_exclude__`:

   ```alloy
   local.file_match "app" {
     sync_period  = "10s"
     path_targets = [{
       "__address__"      = "localhost",
       // On disk the directory is "MyService"; this config uses "Myservice".
       "__path__"         = "C:\\AppLogs\\Myservice\\Logs\\*.log",
       "__path_exclude__" = "C:\\AppLogs\\Myservice\\Logs\\*-backup*.log",
       "job"              = "app",
     }]
   }

   loki.source.file "app" {
     targets    = local.file_match.app.targets
     forward_to = [loki.echo.app.receiver]
   }

   loki.echo "app" {}
   ```

### System information

Windows Server 2022 Standard

### Software version

Grafana Alloy v1.17.0

### Configuration

```text
local.file_match "app" {
  sync_period  = "10s"
  path_targets = [{
    "__address__"      = "localhost",
    // On disk the directory is "MyService"; this config uses "Myservice".
    "__path__"         = "C:\\AppLogs\\Myservice\\Logs\\*.log",
    "__path_exclude__" = "C:\\AppLogs\\Myservice\\Logs\\*-backup*.log",
    "job"              = "app",
  }]
}

loki.source.file "app" {
  targets    = local.file_match.app.targets
  forward_to = [loki.echo.app.receiver]
}

loki.echo "app" {}
```

### Logs

```text

```

### Tip

_No response_

vishalmore90:
### PR Description
Fixes an issue where file matches specified in `__path_exclude__` are matched case-sensitively, causing exclusions to fail on case-insensitive filesystems (such as Windows and macOS).
On case-insensitive file systems like Windows and macOS, the file matcher's discovery pattern (`__path__`) resolves files case-insensitively. However, the exclusion pattern (`__path_exclude__`) is evaluated purely in-memory using `doublestar.PathMatch`, which matches case-sensitively by default. This makes it impossible to reliably exclude files using mixed-case or capitalized extensions.

Fixes #6607
### Note
Automated unit tests for `file_match` pass successfully on Windows. A new test `TestCaseInsensitiveExcludeMatching` has been added to verify this fix.

### Verification
- Added and ran automated unit tests under `internal/component/local/file_match`: `go test -v -tags="nodocker" ./internal/component/local/file_match/...`.
- Confirmed that `TestCaseInsensitiveExcludeMatching` successfully filters out lowercase files using uppercase exclude patterns on case-insensitive OSes.
### PR Checklist
- [x] Verified changes with `go test`
- [x] Checked with `go vet` and format code with `go fmt`

### DIFF
--- a/internal/component/local/file_match/file_windows_test.go
+++ b/internal/component/local/file_match/file_windows_test.go
@@ -59,3 +59,25 @@ func TestCaseInsensitiveGlobMatchingUppercasePattern(t *testing.T) {
 	require.Len(t, foundFiles, 1, "Windows should be case-insensitive: *.LOG should match test.log")
 	require.True(t, testutil.PathEndsWith(foundFiles, "test.log"))
 }
+
+// TestCaseInsensitiveExcludeMatching verifies that exclude patterns are case-insensitive on Windows.
+func TestCaseInsensitiveExcludeMatching(t *testing.T) {
+	dir := path.Join(os.TempDir(), "alloy_testing", "case_insensitive_exclude")
+	err := os.MkdirAll(dir, 0755)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		os.RemoveAll(dir)
+	})
+
+	// Create a file with lowercase extension
+	testutil.WriteFile(t, dir, "test.log")
+
+	// Search with *.log, but exclude *.LOG - should NOT return any files since the exclude is case-insensitive.
+	c := testCreateComponent(t, dir, []string{path.Join(dir, "*.log")}, []string{path.Join(dir, "*.LOG")})
+	c.args.SyncPeriod = 10 * time.Millisecond
+	err = c.Update(c.args)
+	require.NoError(t, err)
+
+	foundFiles := c.getWatchedFiles()
+	require.Empty(t, foundFiles, "Windows should be case-insensitive: *.LOG exclude pattern should match test.log")
+}
--- a/internal/static/server/tls_certstore_stub.go
+++ b/internal/static/server/tls_certstore_stub.go
@@ -1,10 +1,29 @@
-//go:build !windows
+//go:build !windows || !cgo
 
 package server
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"log/slog"
+)
 
 type WinCertStoreHandler struct {
 }
 
-func (w WinCertStoreHandler) Run() {}
+func (w *WinCertStoreHandler) Run() {}
 
-func (w WinCertStoreHandler) Stop() {}
+func (w *WinCertStoreHandler) Stop() {}
+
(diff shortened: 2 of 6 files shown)


kgeckhart:
👋  earlier today we introduced a genai policy for Alloy, https://github.com/grafana/alloy/blob/main/docs/developer/genai.md ([referenced in AGENTS.md as well](https://github.com/grafana/alloy/blame/main/AGENTS.md#L10-L11)). Could you please spend some time to ensure this PR adheres to our AI policy? Thanks.

PR #6620, opened 2026-06-30
chore: Add integration tests for User Agent

ptodev:
This PR will change the User-Agent header of the Alloy OTel Extension to start with "Alloy OTel Extension" as opposed to just "Alloy". That way it can be distinguished when we gather statistics about user agents.

It will also add tests to make sure that the User Agents are what they should be. For example:
* Default engine: `Alloy/v1.18.0 (linux; helm)`
* OTel Extension: `Alloy OTel Extension/v1.18.0 (linux; docker)`
* OTel Engine native components: `Alloy OTel Collector distribution./v1.18.0 (linux/amd64)`

### DIFF
--- a/integration-tests/k8s/deps/manifests/alloy-otel.yaml
+++ b/integration-tests/k8s/deps/manifests/alloy-otel.yaml
@@ -1,0 +1,46 @@
+---
+# Alloy running the OTel engine (`alloy otel`) with the alloyengine extension.
+# The extension runs an inner Alloy config (mounted as config.alloy) whose
+# prometheus.remote_write carries the "Alloy OTel Extension" User-Agent. The image
+# ENTRYPOINT is /bin/alloy, so args run `alloy otel --config=...`. __ALLOY_IMAGE__
+# is substituted by the AlloyOtel dependency at install time.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: alloy-otel
+  labels:
+    app: alloy-otel
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: alloy-otel
+  template:
+    metadata:
+      labels:
+        app: alloy-otel
+    spec:
+      containers:
+        - name: alloy
+          image: __ALLOY_IMAGE__
+          imagePullPolicy: Never
+          args:
+            - otel
+            - --config=/etc/alloy/collector.yaml
+          ports:
+            - name: health
+              containerPort: 13133
+          readinessProbe:
+            tcpSocket:
+              port: 13133
+            initialDelaySeconds: 2
+            periodSeconds: 3
+            timeoutSeconds: 3
+            failureThreshold: 30
+          volumeMounts:
+            - name: config
+              mountPath: /etc/alloy
+      volumes:
+        - name: config
+          configMap:
+            name: alloy-otel-config
--- a/integration-tests/k8s/tests/metamonitor-otel/k8s_test.go
+++ b/integration-tests/k8s/tests/metamonitor-otel/k8s_test.go
@@ -1,0 +1,39 @@
+package metamonitorotel
+
+import (
+	"testing"
+
+	"github.com/grafana/alloy/integration-tests/k8s/deps"
+	"github.com/grafana/alloy/integration-tests/k8s/harness"
+)
+
+// TestMetamonitorOtel runs Alloy with the OTel engine (`alloy otel`), which runs the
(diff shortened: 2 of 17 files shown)


integration-tests/k8s/deps/manifests/alloy-otel.yaml
```diff
@@ -0,0 +1,46 @@
+---
+# Alloy running the OTel engine (`alloy otel`) with the alloyengine extension.
```
blewis12, line 2: Out of curiosity, could we not test the same behavior using docker integration tests? k8s testing has a higher cost in terms of set up complexity, but does it provide us with something that docker tests can't?

ptodev: We could! But ever since the k8s tests were refactored, we decided to migrate all tests to k8s and to use k8s by default in the future. The only tests that would stay on docker are the docker-specific ones. The reason is that k8s tests are usually more flexible and suitable for a wide array of use cases.

thampiotr: I think k8s tests are good choice here, because we have option to shard them among multiple GH Actions to reduce the time. They are a bit more expensive, but the difference is not as big. While k8s is the most frequent way to deploy Alloy, so it's good to cover this. 

We could ask if this could be instead a pipeline test or a unit test though. 
We have this guide for choosing the level for tests: https://github.com/grafana/alloy/blob/main/docs/developer/writing-tests.md#choosing-the-level-of-tests-to-write 

Thoughts on this @ptodev ?

integration-tests/k8s/tests/metamonitor-otel/k8s_test.go
```diff
@@ -0,0 +1,39 @@
+package metamonitorotel
+
+import (
+	"testing"
+
+	"github.com/grafana/alloy/integration-tests/k8s/deps"
+	"github.com/grafana/alloy/integration-tests/k8s/harness"
+)
+
+// TestMetamonitorOtel runs Alloy with the OTel engine (`alloy otel`), which runs the
```
thampiotr, line 10: Is the "metamonitor" naming here the right choice if these tests are verifying user agents retention?

internal/useragent/useragent.go
```diff
@@ -15,13 +15,20 @@ import (
 
 const (
 	ProductName = "Alloy"
+	// ExtensionProductName is the product name reported when Alloy runs as an OTel
+	// Collector extension (the OTel engine), distinguishing it from native Alloy.
```
thampiotr:
```suggestion
	// Collector extension, distinguishing it from native Alloy.
```

Technically it could run as an OTel Collector extension of any OTel Collector distro

integration-tests/k8s/deps/manifests/alloy-otel.yaml
blewis12: > we decided to migrate all tests to k8s 

ah! cool I had missed that

integration-tests/k8s/tests/metamonitor-otel/k8s_test.go
ptodev: In the future I've been [planning](https://github.com/grafana/alloy/pull/6620/changes/BASE..1884234569a14b4215130d7489ce4ebf6719778d#diff-7bdbbd81be871b02623f37092dbef998a00d7adb66dec88da8c68d46c26032d1R37) to add tests which sanity check that the metamonitoring components and dashboards are working fine. We currently don't even have a doc for OTel Engine metamonitoring - having tests would be nice so that we avoid doing independent local testing.

integration-tests/k8s/deps/manifests/alloy-otel.yaml
ptodev: The content of the User Agent header depends on global variables which don't lend themselves to accurate unit testing. Starting Alloy as a whole process and treating it like a black box is guaranteed to test this accurately.

thampiotr: sgtm!

integration-tests/k8s/tests/metamonitor-otel/k8s_test.go
thampiotr: I think that basic 'it's not obviously broken' tests for metamonitoring would be fine in k8s integrateion tests. But for better coverage and more exhaustive tests, maybe it's better to write tests at a lower level.

thampiotr (approved):

PR #6627, opened 2026-07-01
fix(windows): Allow non-admin service account to start Alloy

### LINKED ISSUE #6626: windows: non-admin service account fails to start Alloy (error 1053) due to eventlog registration
### What's wrong?

When Alloy is installed on Windows with a dedicated non-admin service account (via `/USERNAME=` / `/PASSWORD=` on the silent installer, added in #3160), `sc.exe start Alloy` fails with:

```
[SC] StartService FAILED 1053:
The service did not respond to the start or control request in a timely fashion.
```

The Windows System event log shows the SCM pair:
- `Service Control Manager` **7000** — *The Alloy service failed to start due to the following error: The service did not respond to the start or control request in a timely fashion.*
- `Service Control Manager` **7009** — *A timeout was reached (30000 milliseconds) while waiting for the Alloy service to connect.*

No entries appear from the `Alloy` event source (i.e. Alloy itself never logs anything after the first attempt).

### Root cause

`alloy-service.exe` calls `eventlog.InstallAsEventCreate(serviceName, ...)` unconditionally on every start (`internal/cmd/alloy-service/writer_windows.go:27`). Under the hood that opens `HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application` with `CREATE_SUB_KEY` access, which requires admin. A dedicated non-admin service account does **not** have this right, so the call returns `Access is denied.` even when the `Alloy` source is already registered.

The current error check only tolerates the string `"already exists"`:

```go
err := eventlog.InstallAsEventCreate(serviceName, eventTypes)
if err != nil && !strings.Contains(err.Error(), "already exists") {
    return nil, err
}
```

So `newWriter()` returns the access-denied error, `main()` writes to stderr and calls `os.Exit(1)` **before `svc.Run`** ever contacts the SCM. From the SCM's point of view the process never connected, and after 30 s it reports the 1053 / 7009 pair.

This is a related but distinct problem from #529 (which was about slow Go package init under high CPU). The symptom (1053) is the same but the trigger is running as a non-admin account, not startup timing.

### Steps to reproduce

1. Fresh Windows Server 2019 (or similar); no prior Alloy install.
2. Create a local (or domain) service account, e.g. `EC2AMAZ-XXXXXX\svc_grafana_alloy`.
3. Grant it **Log on as a service** (`secpol.msc` → Local Policies → User Rights Assignment).
4. Run the silent installer:
   ```
   alloy-installer-windows-amd64.exe /S /USERNAME="EC2AMAZ-XXXXXX\svc_grafana_alloy" /PASSWORD="…" /FORCEREGISTRY=yes
   ```
5. `sc.exe start Alloy` → `1053`.
6. Check `Get-WinEvent -FilterHashtable @{LogName='System';ProviderName='Service Control Manager'}` → the 7000 / 7009 pair above.

### Confirmed workarounds

Any of these unblock the service:

- **Grant the account write access to the parent EventLog key** (mirrors what `InstallAsEventCreate` needs):

  ```powershell
  $key = 'HKLM:\SYSTEM\CurrentControlSet\Services\EventLog\Application'
  $acl = Get-Acl $key
  $rule = New-Object System.Security.AccessControl.RegistryAccessRule(
      'EC2AMAZ-XXXXXX\svc_grafana_alloy',
      'CreateSubKey,SetValue,ReadKey',
      'ContainerInherit,ObjectInherit','None','Allow'
  )
  $acl.AddAccessRule($rule)
  Set-Acl $key $acl
  ```

- **Run once as `LocalSystem` to install the source, then switch back to the service account.**
- **Pre-create the source subkey manually** (`HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy` with `EventMessageFile` = `%SystemRoot%\System32\EventCreate.exe`, `TypesSupported` = 7).

### Suggested fix

Two complementary changes:

1. In `newWriter()`, try `eventlog.Open(serviceName)` first. If it succeeds, use it. Only fall back to `InstallAsEventCreate` when the source isn't registered yet. This avoids ever needing `CREATE_SUB_KEY` at runtime when the source already exists.

2. In `packaging/windows/install_script.nsis`, pre-create `HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy` during install (the installer runs elevated). This closes the gap for a fresh install where the source doesn't exist yet on first service start.

Both changes together let the service start on first boot under a dedicated non-admin account, without documenting extra manual registry steps for operators.

I have a working patch and will open a PR referencing this issue.

### System information

Windows Server 2019 (EC2 `t3.medium`), amd64.

### Software version

Alloy v1.17.0 (installer `alloy-installer-windows-amd64.exe`).

### Configuration

Default `config.alloy` (unmodified from installer).

### Logs

```
# System log (relevant only)
Service Control Manager  7000  The Alloy service failed to start due to the following error: The service did not respond to the start or control request in a timely fashion.
Service Control Manager  7009  A timeout was reached (30000 milliseconds) while waiting for the Alloy service to connect.

# Application log filtered on ProviderName=Alloy → no events after the initial LocalSystem test run
```

### AI disclosure

fabiiw05:

Alloy fails to start with SCM error 1053 (`The service did not respond to the start or control request in a timely fashion.`) when installed with a dedicated non-admin service account via the silent installer's `/USERNAME=` / `/PASSWORD=` flags. Fix the runtime code path in `alloy-service.exe` and have the Windows installer pre-register the event source, so a fresh install under a dedicated service account works end to end.

### Pull Request Details

Two changes:

**1. `internal/cmd/alloy-service/writer_windows.go` — runtime fix**

`alloy-service.exe` called `eventlog.InstallAsEventCreate` unconditionally on every start. Internally that opens `HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application` with `CREATE_SUB_KEY` access, which non-admin accounts do not have — even when the `Alloy` source is already registered. The call returned `Access is denied`, `newWriter()` failed, `main()` exited before `svc.Run`, and the SCM surfaced the timeout as 1053.

Try `eventlog.Open` first (which does not need registry write access). If the source is already registered, the fast path returns immediately and the service starts. Only fall back to `InstallAsEventCreate` when the source isn't registered yet.

**2. `packaging/windows/install_script.nsis` — installer fix**

Pre-create `HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy` with `EventMessageFile` and `TypesSupported` during install (the installer runs elevated). This closes the gap for a fresh install where the source doesn't exist yet on the first service start, and it also completes the existing TODO in `internal/runtime/logging/eventlog/eventlog_windows.go:64-65`:

```go
// TODO: Also do the registration in the Alloy Windows installer.
// That way users don't have to run Alloy with admin rights one time
// just for this to be installed.
```

The parent (`alloy-service.exe`) and the child (`alloy.exe`'s `logging {}` block via `openEventLogWindows`) both benefit: neither needs admin at runtime, and Event Viewer displays messages cleanly because the source has a message file registered.

Fixes #6626

### Notes to the Reviewer

- Validated end-to-end on Windows Server 2019 (EC2) with a dedicated service account (`svc_grafana_alloy`) that has only the **Log on as a service** right — no admin, no write access to the EventLog registry hive.
- Manually removed the `Alloy` event source subkey and any temporary ACL grants to reproduce a "fresh install" state, then confirmed that the fixed `alloy-service.exe` starts, the child `alloy.exe` initializes the `logging {}` block, and Alloy reaches `msg="{^_^} Alloy is running"` with the HTTP server listening — all without any admin-level runtime privileges.
- The change is Windows-only (build-tagged `_windows.go`) and does not affect other platforms.
- No new tests were added: the failure mode is a Windows-specific registry ACL interaction that requires a real non-admin service account to reproduce, which the existing test harness doesn't model. The change is a small ordering swap (`Open` before `Install`) that preserves the original behavior when `Open` fails.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/cmd/alloy-service/writer_windows.go
+++ b/internal/cmd/alloy-service/writer_windows.go
@@ -20,12 +20,26 @@ var (
 // newWriter creates a new writer which writes to the Windows Event
 // Logger.
 func newWriter() (*writer, error) {
+	// Fast path: if the event source is already registered, open it directly.
+	// eventlog.InstallAsEventCreate opens the parent EventLog\Application key
+	// with CREATE_SUB_KEY access, which requires admin. When Alloy runs under a
+	// dedicated non-admin service account, that call fails with "Access is
+	// denied" even though the source already exists, causing the process to
+	// exit before svc.Run can report back to the SCM (surfacing as service
+	// error 1053).
+	if el, err := eventlog.Open(serviceName); err == nil {
+		runtime.SetFinalizer(el, func(li *eventlog.Log) {
+			_ = li.Close()
+		})
+		return &writer{el: el}, nil
+	}
+
 	eventTypes := uint32(eventlog.Info | eventlog.Warning | eventlog.Error)
 
 	// Install the event source. This will fail with an error string saying "already
 	// exists" if it has been installed before.
-	err := eventlog.InstallAsEventCreate(serviceName, eventTypes)
-	if err != nil && !strings.Contains(err.Error(), "already exists") {
+	if err := eventlog.InstallAsEventCreate(serviceName, eventTypes); err != nil &&
+		!strings.Contains(err.Error(), "already exists") {
 		return nil, err
 	}
 
--- a/internal/cmd/alloy-service/writer_windows.go
+++ b/internal/cmd/alloy-service/writer_windows.go
@@ -17,15 +17,6 @@ var (
 )
 
 // newWriter creates a new writer which writes to the Windows Event Log.
-//
-// It uses the shared event log opener from internal/runtime/logging/eventlog,
-// which pre-checks whether the source is already registered before attempting
-// to install it. That lets a dedicated non-admin service account start the
-// service when the Alloy installer (or a prior admin run) has already
-// registered the "Alloy" event source, avoiding the CREATE_SUB_KEY access
-// requirement of eventlog.InstallAsEventCreate that would otherwise cause the
-// service to exit before svc.Run can report back to the SCM (surfacing as
-// service error 1053).
 func newWriter() (*writer, error) {
 	el, err := eventlog.GetEventLogOpener()(serviceName)
 	if err != nil {
--- a/internal/cmd/alloy-service/writer_windows.go
+++ b/internal/cmd/alloy-service/writer_windows.go
@@ -2,57 +2,35 @@ package main
 
 import (
 	"io"
-	"runtime"
 	"strings"
 
-	"golang.org/x/sys/windows/svc/eventlog"
+	"github.com/grafana/alloy/internal/runtime/logging/eventlog"
 )
 
 // writer sends writes to the Windows Event Log.
 type writer struct {
-	el *eventlog.Log
+	el eventlog.EventLog
 }
 
 var (
 	_ io.Writer = (*writer)(nil)
 )
 
-// newWriter creates a new writer which writes to the Windows Event
-// Logger.
+// newWriter creates a new writer which writes to the Windows Event Log.
+//
+// It uses the shared event log opener from internal/runtime/logging/eventlog,
+// which pre-checks whether the source is already registered before attempting
+// to install it. That lets a dedicated non-admin service account start the
+// service when the Alloy installer (or a prior admin run) has already
+// registered the "Alloy" event source, avoiding the CREATE_SUB_KEY access
+// requirement of eventlog.InstallAsEventCreate that would otherwise cause the
+// service to exit before svc.Run can report back to the SCM (surfacing as
+// service error 1053).
 func newWriter() (*writer, error) {
-	// Fast path: if the event source is already registered, open it directly.
-	// eventlog.InstallAsEventCreate opens the parent EventLog\Application key
-	// with CREATE_SUB_KEY access, which requires admin. When Alloy runs under a
-	// dedicated non-admin service account, that call fails with "Access is
-	// denied" even though the source already exists, causing the process to
-	// exit before svc.Run can report back to the SCM (surfacing as service
-	// error 1053).
-	if el, err := eventlog.Open(serviceName); err == nil {
-		runtime.SetFinalizer(el, func(li *eventlog.Log) {
-			_ = li.Close()
-		})
-		return &writer{el: el}, nil
-	}
-
-	eventTypes := uint32(eventlog.Info | eventlog.Warning | eventlog.Error)
-
-	// Install the event source. This will fail with an error string saying "already
-	// exists" if it has been installed before.
-	if err := eventlog.InstallAsEventCreate(serviceName, eventTypes); err != nil &&
-		!strings.Contains(err.Error(), "already exists") {
-		return nil, err
-	}
-
-	el, err := eventlog.Open(serviceName)
+	el, err := eventlog.GetEventLogOpener()(serviceName)
 	if err != nil {
 		return nil, err
 	}
-
-	// Ensure the logger gets closed when GC runs.
-	runtime.SetFinalizer(el, func(li *eventlog.Log) {
-		_ = li.Close()
-	})
-
 	return &writer{el: el}, nil
 }
 
--- a/packaging/windows/install_script.nsis
+++ b/packaging/windows/install_script.nsis
@@ -119,12 +119,13 @@ Section "install"
   Call CreateConfig
   Call CreateDataDirectory
   Call InitializeRegistry
+  Call RegisterEventSource
 
   ${If} $User != ""
     StrCpy $AuthFlag "obj= $\"$User$\""
     ${If} $Password != ""
       StrCpy $AuthFlag "$AuthFlag password= $\"$Password$\""
-    ${EndIf}  
+    ${EndIf}
   ${EndIf}
 
   # Create the service.
@@ -246,6 +247,20 @@ Function InitializeRegistry
   ${EndIf}
 
   Return
+FunctionEnd
+
+# RegisterEventSource pre-creates the Windows Event Log source for Alloy so
+# that a non-admin service account can open the source without needing write
+# access to HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application. Without
+# this, alloy-service.exe fails on first start under a dedicated service
+# account and the SCM reports error 1053 (timeout waiting to connect).
+Function RegisterEventSource
+  !define EVENTLOGKEY "HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy"
+
+  nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v EventMessageFile /t REG_EXPAND_SZ /d "%SystemRoot%\System32\EventCreate.exe" /f'
+  Pop $0
+  nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v TypesSupported   /t REG_DWORD     /d 7 /f'
+  Pop $0
 FunctionEnd
 
 Function SetFolderPermissions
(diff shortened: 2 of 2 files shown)


internal/cmd/alloy-service/writer_windows.go
```diff
@@ -20,12 +20,26 @@ var (
...
+	}
+
 	eventTypes := uint32(eventlog.Info | eventlog.Warning | eventlog.Error)
 
 	// Install the event source. This will fail with an error string saying "already
 	// exists" if it has been installed before.
-	err := eventlog.InstallAsEventCreate(serviceName, eventTypes)
-	if err != nil && !strings.Contains(err.Error(), "already exists") {
+	if err := eventlog.InstallAsEventCreate(serviceName, eventTypes); err != nil &&
+		!strings.Contains(err.Error(), "already exists") {
```
ptodev:
```suggestion
	el, err := eventlog.GetEventLogOpener()(serviceName)
	if err != nil {
```

We could use the same method that Alloy itself uses in [its own logger](https://github.com/grafana/alloy/blob/main/internal/runtime/logging/logger.go#L84)? It has a [mechanism](https://github.com/grafana/alloy/blob/main/internal/runtime/logging/eventlog/eventlog_windows.go) to avoid using elevated access. This was added recently so that Alloy can [natively](https://grafana.com/docs/alloy/latest/reference/config-blocks/logging/#destination) log to the event log without a windows service.

We are still keeping the Windows service code and its event logger because there could be code in Alloy's dependencies that sends output to stdout/stderr and the Windows service is a nice fallback to capture this and send it to the event log too.

ptodev:
Thank you! I'd also noticed this in the past but never had the time to test a change for it 😅

internal/cmd/alloy-service/writer_windows.go
```diff
@@ -2,43 +2,35 @@ package main
...
+// newWriter creates a new writer which writes to the Windows Event Log.
+//
+// It uses the shared event log opener from internal/runtime/logging/eventlog,
+// which pre-checks whether the source is already registered before attempting
+// to install it. That lets a dedicated non-admin service account start the
+// service when the Alloy installer (or a prior admin run) has already
+// registered the "Alloy" event source, avoiding the CREATE_SUB_KEY access
+// requirement of eventlog.InstallAsEventCreate that would otherwise cause the
+// service to exit before svc.Run can report back to the SCM (surfacing as
+// service error 1053).
```
ptodev:
```suggestion
```

packaging/windows/install_script.nsis
```diff
@@ -248,6 +249,20 @@ Function InitializeRegistry
   Return
 FunctionEnd
 
+# RegisterEventSource pre-creates the Windows Event Log source for Alloy so
+# that a non-admin service account can open the source without needing write
+# access to HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application. Without
+# this, alloy-service.exe fails on first start under a dedicated service
+# account and the SCM reports error 1053 (timeout waiting to connect).
```
ptodev:
```suggestion
# RegisterEventSource pre-creates the Windows Event Log source for Alloy so
# that a non-admin service account can open the source without needing write
# access to HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application.
```

packaging/windows/install_script.nsis
```diff
@@ -248,6 +249,18 @@ Function InitializeRegistry
...
+# that a non-admin service account can open the source without needing write
+# access to HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application.
+Function RegisterEventSource
+  !define EVENTLOGKEY "HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy"
+
+  nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v EventMessageFile /t REG_EXPAND_SZ /d "%SystemRoot%\System32\EventCreate.exe" /f'
+  Pop $0
+  nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v TypesSupported   /t REG_DWORD     /d 7 /f'
+  Pop $0
+FunctionEnd
```
ptodev, line 272:
```suggestion
# RegisterEventSource pre-creates the Windows Event Log source for Alloy so
# that a non-admin service account can open the source without needing write
# access to HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application.
#
# These values mirror what eventlog.InstallAsEventCreate writes at runtime
# (golang.org/x/sys/windows/svc/eventlog). We only write the values when the
#  source key doesn't already exist. This matches eventlog.Install and preserves
#  customizations a user has made.
Function RegisterEventSource
  nsExec::ExecToLog 'Reg.exe query "${EVENTLOGKEY}"'
  Pop $0
  ${If} $0 == 1
    nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v CustomSource     /t REG_DWORD     /d 1 /f'
    Pop $0 # Ignore return result
    nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v EventMessageFile /t REG_EXPAND_SZ /d "%SystemRoot%\System32\EventCreate.exe" /f'
    Pop $0 # Ignore return result
    nsExec::ExecToLog 'Reg.exe add "${EVENTLOGKEY}" /v TypesSupported   /t REG_DWORD     /d 7 /f'
    Pop $0 # Ignore return result
  ${EndIf}
FunctionEnd
```

I just noticed that the Go code currently assigns one more value (CustomSource):

Image

Let's update nsis to match what Go does. In the beginning of the nsis script you can also add this, after the other `!define `:

```
!define EVENTLOGKEY "HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\Alloy"
```

And at the end of this file you can add this for the uninstaller:
```
  nsExec::ExecToLog 'Reg.exe delete "${EVENTLOGKEY}" /f'
```

fabiiw05:
@ptodev Thanks for the thorough review, and I'm sorry for the slow and
incomplete response — I missed a few pieces of your suggestion for
`RegisterEventSource` the first time around.

The latest commit addresses the remaining items:

- Moved `!define EVENTLOGKEY` to the top-level `!define` block so both the
  installer and uninstaller can share it.
- Added a `Reg.exe query` guard so we only write values when the source key
  doesn't yet exist, matching `eventlog.Install` semantics and preserving any
  user customizations.
- Added the `CustomSource=1` `REG_DWORD` value to mirror what
  `eventlog.InstallAsEventCreate` writes at runtime.
- Added a corresponding `Reg.exe delete "${EVENTLOGKEY}" /f` in the uninstall
  section.

### Verification on Windows Server 2022 (AWS EC2)

I built two installers — one from `main` (`9f3ae13c1`) and one from this branch
with the changes above — and ran them against a Windows Server 2022 instance
with a dedicated non-admin service account (`AlloyUser` with
`SeServiceLogonRight`).

| Check | Pre-fix (`main`) | Post-fix (this PR) |
|---|---|---|
| Service starts under non-admin account | Fails (System log 7000/7009, `Start-Service` errors out) | Service reaches Running |
| `Application` log entries from Alloy under non-admin | 0 net new (source can't be opened) | Alloy writes `INFO msg="starting program"` |
| `HKLM\...\EventLog\Application\Alloy` after install | Not created (Go fails to `InstallAsEventCreate` as non-admin) | Pre-registered by installer with `CustomSource=1`, `EventMessageFile`, `TypesSupported=7` |
| `HKLM\...\EventLog\Application\Alloy` after uninstall | Left behind | Removed by the new uninstall step |

Please take another look when you get a chance.

ptodev (approved):
Thank you!

ptodev:
@fabiiw05 would you mind signing your commits please? I cannot merge otherwise.

fabiiw05:
@ptodev I've re-signed the commits to satisfy the verified-signature requirement.

One note: I accidentally also re-signed the review-suggestion commit authored by you, so it now shows as authored by you and committed/signed by me. Sorry for the extra churn. Please let me know if you'd prefer me to squash or rework the branch to avoid that.

ptodev (approved):

ptodev:
@fabiiw05 thank you!

> Please let me know if you'd prefer me to squash or rework the branch to avoid that.

It's ok, no need to be this thorough. When it's merged to the main branch it's all going to get squashed into one commit anyway.

PR #6632, opened 2026-07-01
feat: Add OpAmp Extension to OTel Engine

blewis12:
As we're gearing up now to move the OpAmp FM solution in GC out of private preview and into GA, now is a good time to add the `opamp` extension to the OTel engine. This means that users will be able to manage their Alloy instances (with the [supervisor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/opampsupervisor/specification/README.md)) as well as other open telemetry collector distributions

This is also a pre-req for implementation for our built-in/alloy-native solution


collector/go.mod
```diff
@@ -675,6 +676,7 @@ require (
 	github.com/mdlayher/wifi v0.1.0 // indirect
 	github.com/metalmatze/signal v0.0.0-20210307161603-1c9aa721a97a // indirect
 	github.com/mholt/archives v0.1.5 // indirect
+	github.com/michel-laterman/proxy-connect-dialer-go v0.1.0 // indirect
```
x1unix, line 679: This repo [was archived](https://github.com/michel-laterman/proxy-connect-dialer-go) recently.

blewis12: It's already updated to a non-archived repo in the `main` branch of [opamp-client](https://github.com/open-telemetry/opamp-go/blob/main/go.mod) - so once that is released and opamp extension pulls it in we'll be good there

security wise it's [very light](https://github.com/michel-laterman/proxy-connect-dialer-go/blob/main/go.mod) in what it brings in regardless, so I think it's fine to wait

blewis12: I guess we could always just add a replace for it but I feel it's unnecessary right now, lmk if you disagree

thampiotr (approved):

collector/go.mod
ptodev: I think it's ok to not add a replace. It wasn't archived that long ago.

ptodev (approved):

PR #6634, opened 2026-07-02
fix(pyroscope.ebpf): Skip empty build ID debug info uploads

korniltsev-grafanista:
Skips Pyroscope eBPF debug info uploads when the GNU build ID is empty. Adds coverage to ensure those files are not opened or queued for upload. Verified with GOOS=linux go test -run TestUpload_EmptyBuildID -exec /usr/bin/true ./internal/component/pyroscope/ebpf/reporter/parca/reporter.


jake-kramer (approved):
LGTM

marcsanmi (approved):

PR #6636, opened 2026-07-02
feat!: Expose HTTP server idle/read/read_header/write timeouts in OTel Components

kgeckhart:

Add `idle_timeout`, `read_timeout`, `read_header_timeout`, and `write_timeout` to the shared `HTTPServerArguments` block, covering all 8 components that run an HTTP server. Each component's `SetToDefault` now sets `idle_timeout`/`read_header_timeout`/`write_timeout` to `confighttp`'s own defaults (`1m`/`1m`/`30s`); `Convert()` passes the fields straight through unchanged, so an explicit `"0s"` genuinely means unbounded. 

### Pull Request Details

These values have existed upstream since 142 but we didn't expose them causing every otelcol HTTP server to run with unbounded regardless of config. These new defaults bring otelcol's HTTP servers closer inline with Alloy's own components with an HTTP server based on dskit which uses: `ServerReadTimeout=30s`, `ServerWriteTimeout=30s`, `ServerIdleTimeout=120s`. Write timeout matches, idle is shorter (`1m` vs dskit `2m`); read is split into `read_header_timeout` (`1m`) with an unbounded `read_timeout`, where dskit bounds the whole read at `30s` combined.

BREAKING-CHANGE: `otelcol.receiver.otlp`, `otelcol.receiver.faro`, `otelcol.receiver.jaeger`, `otelcol.receiver.zipkin`, `otelcol.receiver.influxdb`, `otelcol.receiver.splunkhec`, `otelcol.receiver.datadog`, and `otelcol.extension.jaeger_remote_sampling` now default to `idle_timeout = "1m"`, `read_header_timeout = "1m"`, and `write_timeout = "30s"` on their HTTP servers, matching upstream default. These were previously being defaulted to `"0s"`. You can explicitly set them to `"0s"` to restore the previous behavior.

### Notes to the Reviewer

`keep_alives_enabled` is a pre-existing converter gap on `HTTPServerArguments` which is left out deliberately to avoid a further sprawling change.

`otelcol.receiver.datadog`'s standalone `read_timeout` was also consolidated into the shared block, preserving its existing 60s default.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.receiver.datadog.md
```diff
@@ -45,11 +45,14 @@ You can use the following arguments with `otelcol.receiver.datadog`:
...
 | `endpoint`               | `string`                   | `host:port` to listen for traffic on.                                        | `"localhost:8126"`                                         | no       |
+| `idle_timeout`           | `duration`                 | Maximum idle time before closing a keep-alive connection.                    | `"1m"`                                                     | no       |
 | `include_metadata`       | `bool`                     | Propagate incoming connection metadata to downstream consumers.              | `false`                                                    | no       |
 | `keep_alives_enabled`    | `boolean`                  | Whether or not HTTP keep-alives are enabled                                  | `true`                                                     | no       |
 | `max_request_body_size`  | `string`                   | Maximum request body size the server will allow.                             | `"20MiB"`                                                  | no       |
```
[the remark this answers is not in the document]
clayton-cornell: Out of scope for this PR

internal/converter/internal/otelcolconvert/converter_datadogreceiver.go
```diff
@@ -46,15 +46,15 @@ func toDatadogReceiver(state *State, id componentstatus.InstanceID, cfg *datadog
...
+	args := &datadog.Arguments{
+		HTTPServer:   *toHTTPServerArguments(&cfg.ServerConfig),
 		DebugMetrics: common.DefaultValue[datadog.Arguments]().DebugMetrics,
 
 		Output: &otelcol.ConsumerArguments{
 			Metrics: ToTokenizedConsumers(nextMetrics),
 			Traces:  ToTokenizedConsumers(nextTraces),
 		},
 	}
+	args.HTTPServer.ReadTimeout = cfg.ReadTimeout
```
blewis12: is it worth documenting that we override this here? I guess the behavior is the same as upstream?

blewis12 (commented):
One small question - IMO these breaking changes are permissible because they fix an unintended bug of mismatching defaults, and will be marked in our changelog/release notes too so folks are aware

internal/converter/internal/otelcolconvert/converter_datadogreceiver.go
kgeckhart: This ended up being a bit of a rabbit hole, turns out I shouldn't do it at all. upstream stopped applying the outer field in [open-telemetry/opentelemetry-collector-contrib#46657)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46657/changes). The component no longer sets it and I cleaned up some broken pieces in the converter around this.

blewis12 (approved):

PR #6647, opened 2026-07-06
feat(otelcol): Add otelcol.processor.redaction

### LINKED ISSUE #5829: Add support for redaction processor
### Component(s)

_No response_

### Request

I would like to propose the inclusion of the [redaction processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/redactionprocessor) in Alloy.

It can help in cases where instrumentation libraries [produce high cardinality span names](https://github.com/vercel/next.js/issues/54694) , and without it, users need to add regexp patterns for each route one by one and keep them up to date.

The logic also mirrors what obi/beyla do (same underlying [library](https://github.com/grafana/clusterurl)), which has been battle-tested in production.

While other solutions are possible (remap span_names from other attributes), this is a great tool in the toolbox to have that covers 90% of the cases and can help alleviate cases in which instrumentation libraries cause cardinality explosions.

### Use case

An instrumentation library produces high cardinality span_names, for example, because the full route url (including IDs and query string params) is included (instead of the route path).

Span metrics connector (either in the collector or Tempo) produces metrics out the traces and the cardinality of the span_name label explodes. This can lead to failed ingestion due to rate limits or, in the worst case, database OOM crashes.

Example from the OTEL demo, where a workaround has been implemented to mitigate the issue, but does not scale to real world usage: https://github.com/open-telemetry/opentelemetry-demo/issues/1665

### Tip

andreacappadona17:

Add `otelcol.processor.redaction`, wrapping the upstream OpenTelemetry Collector `redactionprocessor` so Alloy users can redact, hash, and sanitize
  telemetry attributes, URLs, and database query values in Alloy pipelines.

### Pull Request Details

This PR adds the experimental `otelcol.processor.redaction` component and registers it with Alloy’s default component set.

  It also adds the upstream redaction processor to the embedded OpenTelemetry Collector distro used by `alloy otel`, updates the OpenTelemetry config
  converter with golden test coverage, and adds component reference documentation.

  The component supports the upstream redaction options, including allowed and ignored keys, blocked key and value patterns, HMAC hash functions, URL
  sanitization, database query sanitization, output routing, and debug metrics.

Fixes #5829 

### Notes to the Reviewer

 The Alloy component is marked experimental, matching the upstream processor’s mixed alpha/beta stability.

  Validation delegates to the upstream config where possible, with an additional `hash_function` enum check because the Alloy conversion path doesn’t invoke
  upstream text unmarshalling for that field.

  Tested with:

  ```sh
  go test ./internal/component/otelcol/processor/redaction ./internal/converter/internal/otelcolconvert
```

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### CODE REVIEW

**Comment by clayton-cornell** [comment]
Ping me for a doc review after an initial code review is done.

File: docs/sources/reference/components/otelcol/otelcol.processor.redaction.md [hunk 1]
```diff
@@ -0,0 +1,289 @@
...
+    - oss
+title: otelcol.processor.redaction
+---
+
+# `otelcol.processor.redaction`
+
+{{< docs/shared lookup="stability/experimental.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+`otelcol.processor.redaction` deletes span, log, and metric attributes that don't match a list of allowed attributes.
+It also masks attribute values that match a blocked value list, and can sanitize high-cardinality URLs and database queries into stable, low-cardinality forms.
```
clayton-cornell:
```suggestion
`otelcol.processor.redaction` removes span, log, and metric attributes that don't match an allowlist.
It masks attribute values that match blocklist expressions and sanitizes high-cardinality URLs and database queries into stable, low-cardinality forms.
```

Some digging suggests that the wrapped OpenTelemetry processor also applies redaction and sanitization to log bodies.  Should this be added here?

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+The `>` symbol indicates deeper levels of nesting.
+For example, `db_sanitizer` > `sql` refers to a `sql` block defined inside a `db_sanitizer` block.
+
+[output]: #output
+[db_sanitizer]: #db_sanitizer
+[db]: #db-blocks
+[debug_metrics]: #debug_metrics
+[url_sanitizer]: #url_sanitizer
+
```
clayton-cornell, line 108:
```suggestion
[url_sanitizer]: #url_sanitizer

{{< /docs/alloy-config >}}

```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+Each `blocked_values` entry is a regular expression that's matched against attribute values.
+Only the substring that matches the expression is replaced, with `****` by default or with a hash of the matched text when `hash_function` is set.
+Regular expressions use the [RE2][] syntax, which doesn't support lookarounds or backreferences.
+
+[RE2]: https://github.com/google/re2/wiki/Syntax
+
+## Blocks
+
+You can use the following blocks with `otelcol.processor.redaction`:
+
```
clayton-cornell, line 83:
```suggestion
You can use the following blocks with `otelcol.processor.redaction`:

{{< docs/alloy-config >}}

```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+| `ignored_key_patterns` | `list(string)` | Regular expressions for attribute keys that pass through unchanged.                  | `[]`    | no       |
+| `allowed_values`       | `list(string)` | Regular expressions for values of blocked keys that are not masked.                  | `[]`    | no       |
+| `blocked_values`       | `list(string)` | Regular expressions for values of allowed keys that are masked.                      | `[]`    | no       |
+| `hash_function`        | `string`       | Function used to hash redacted values instead of masking them with a fixed string.   | `""`    | no       |
+| `hmac_key`             | `secret`       | Secret key used for HMAC hashing when `hash_function` is an HMAC variant.            | `""`    | no       |
+| `redact_all_types`     | `bool`         | Redact non-string attributes as well, by converting them to a string representation. | `false` | no       |
+| `summary`              | `string`       | Verbosity of the diagnostic attributes the processor adds. One of `debug`, `info`, `silent`. | `""` | no       |
```
clayton-cornell:
```suggestion
| Name                   | Type           | Description                                                                                                                                        | Default | Required |
| ---------------------- | -------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
| `allow_all_keys`       | `bool`         | Allow all attribute keys. Disables the `allowed_keys` list.                                                                                        | `false` | no       |
| `allowed_keys`         | `list(string)` | Allowed attribute keys. Keys not on the list are removed.                                                                                          | `[]`    | no       |
| `allowed_values`       | `list(string)` | Regular expressions for values that the processor leaves unchanged, even when they also match blocked value patterns.                              | `[]`    | no       |
| `blocked_key_patterns` | `list(string)` | Regular expressions for attribute keys whose values are masked.                                                                                    | `[]`    | no       |
| `blocked_values`       | `list(string)` | Regular expressions for value substrings that the processor masks after key filtering.                                                             | `[]`    | no       |
| `hash_function`        | `string`       | Function used to hash redacted values instead of masking them with a fixed string.                                                                 | `""`    | no       |
| `hmac_key`             | `secret`       | Secret key used for HMAC hashing when `hash_function` is an HMAC variant.                                                                          | `""`    | no       |
| `ignored_key_patterns` | `list(string)` | Regular expressions for attribute keys that pass through unchanged.                                                                                | `[]`    | no       |
| `ignored_keys`         | `list(string)` | Attribute keys that pass through unchanged.                                                                                                        | `[]`    | no       |
| `redact_all_types`     | `bool`         | Redact non-string attributes as well, by converting them to a string representation.                                                               | `false` | no       |
| `summary`              | `string`       | Controls diagnostic attributes that describe redaction activity. Use `debug` or `info` to send diagnostics. Use `silent` or `""` to suppress them. | `""`    | no       |
```

Alpha sort rows

Rewrote the descriptions for summary, allowed_values, and blocked_values to better align with implementation/upstream.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+If `allow_all_keys` is `false`, only attributes whose keys appear in `allowed_keys` are kept.
+The `allowed_keys` list fails closed: if it's empty and `allow_all_keys` is `false`, all attributes are removed.
+
```
clayton-cornell, line 63: Should we note that the processor removes keys before it checks against `blocked_values`? And to mask values without removing keys, set `allow_all_keys` to `true` or add them to `allowed_keys`?

Reference upstream docs here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/redactionprocessor/README.md

Possible text to add:

```

The processor removes disallowed keys before it checks remaining values against `blocked_values`.
To mask values without removing keys, set `allow_all_keys = true` or include the keys in `allowed_keys`.

```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+`hash_function` accepts one of `sha1`, `sha3`, `md5`, `hmac-sha256`, or `hmac-sha512`.
+When `hash_function` is `hmac-sha256`, `hmac_key` is required and must be at least 32 bytes long.
+When `hash_function` is `hmac-sha512`, `hmac_key` is required and must be at least 64 bytes long.
```
clayton-cornell:
```suggestion
{{< param "PRODUCT_NAME" >}} validates `hash_function` during configuration parsing.
When you set `hash_function` to `hmac-sha256`, you must also set `hmac_key` to at least 32 bytes.
When you set `hash_function` to `hmac-sha512`, you must also set `hmac_key` to at least 64 bytes.
```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+Each `blocked_values` entry is a regular expression that's matched against attribute values.
+Only the substring that matches the expression is replaced, with `****` by default or with a hash of the matched text when `hash_function` is set.
+Regular expressions use the [RE2][] syntax, which doesn't support lookarounds or backreferences.
```
clayton-cornell:
```suggestion
The processor matches each `blocked_values` expression against attribute values.
It replaces only the substring that matches the expression.
By default, the processor replaces matching text with `****`.
When you set `hash_function`, the processor replaces matching text with a hash of the matched substring.
Regular expressions use [RE2][] syntax, which doesn't support lookarounds or backreferences.
```

Just re-worked to be more active

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+### `output`
+
+{{< badge text="Required" >}}
+
+{{< docs/shared lookup="reference/components/output-block.md" source="alloy" version="<ALLOY_VERSION>" >}}
+
+### `url_sanitizer`
+
+The `url_sanitizer` block collapses high-cardinality URLs, such as routes with embedded IDs or query strings, into stable low-cardinality forms.
```
clayton-cornell:
```suggestion
The `url_sanitizer` block sanitizes high-cardinality URLs, such as routes with embedded IDs or query strings, into stable low-cardinality forms.
```

What about matching span names?

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+| Name                 | Type           | Description                                              | Default | Required |
+| -------------------- | -------------- | -------------------------------------------------------- | ------- | -------- |
+| `enabled`            | `bool`         | Enable URL sanitization.                                 | `false` | no       |
+| `attributes`         | `list(string)` | Attributes to sanitize.                                  | `[]`    | no       |
+| `sanitize_span_name` | `bool`         | Whether span names should also be sanitized. Only applies when `enabled` is `true`. Set to `false` to opt out. | `true` | no |
+
+### `db_sanitizer`
+
+The `db_sanitizer` block enables sanitization of database queries.
```
clayton-cornell:
```suggestion
The `db_sanitizer` block sanitizes database queries and commands.
```

What about matching span names?

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+
+`otelcol.processor.redaction` doesn't expose any component-specific debug information.
+
+## Debug metrics
+
+`otelcol.processor.redaction` doesn't expose any component-specific debug metrics.
+
+## Examples
+
+### Redact all keys and mask credit card numbers
```
clayton-cornell:
```suggestion
### Keep selected keys and mask credit card numbers
```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+{{< admonition type="note" >}}
+The processor replaces the entire matched substring, and RE2 doesn't support lookarounds, so you can't keep only the domain part and produce `****@example.com`.
+If you only want to hide the local part and keep the domain, match the local part together with the `@`, for example `` `[a-zA-Z0-9._%+\-]+@` ``.
+This turns `john.doe@example.com` into `****example.com` because the `@` is part of the match and is replaced as well.
+{{< /admonition >}}
+
+### Hash values instead of masking
```
clayton-cornell:
```suggestion
### Hash matched values
```

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
```diff
@@ -0,0 +1,289 @@
...
+  output {
+    traces = [otelcol.exporter.otlp.default.input]
+  }
+}
+```
+
+{{< admonition type="note" >}}
+The processor replaces the entire matched substring, and RE2 doesn't support lookarounds, so you can't keep only the domain part and produce `****@example.com`.
+If you only want to hide the local part and keep the domain, match the local part together with the `@`, for example `` `[a-zA-Z0-9._%+\-]+@` ``.
+This turns `john.doe@example.com` into `****example.com` because the `@` is part of the match and is replaced as well.
```
clayton-cornell:
```suggestion
If you want to hide the local part and keep the domain, match the local part together with `@`, for example `` `[a-zA-Z0-9._%+\-]+@` ``.
That pattern turns `john.doe@example.com` into `****example.com` because the match includes `@`.
```

clayton-cornell:
Decided to do an initial doc review before code review happened.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Added this clarification, including the `allow_all_keys` and `allowed_keys` guidance.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

docs/sources/reference/components/otelcol/otelcol.processor.redaction.md
andreacappadona17: Addressed in commit d158771d3.

clayton-cornell:
Docs look ok. Over to @grafana/grafana-alloy-maintainers for code review now.

internal/component/otelcol/processor/redaction/redaction.go
```diff
@@ -0,0 +1,208 @@
...
+	args.DebugMetrics.SetToDefault()
+}
+
+// Validate implements syntax.Validator. It adds a parse-time enum check on
+// hash_function (mapstructure.Decode bypasses HashFunction.UnmarshalText, so the
+// value is otherwise unvalidated), then delegates the remaining checks — notably
+// the HMAC key requirements — to the upstream Config.Validate.
+func (args *Arguments) Validate() error {
+	switch redactionprocessor.HashFunction(args.HashFunction) {
+	case redactionprocessor.None,
```
blewis12: actually I think upstream is case-insensitive, they always convert to [lowercase before parsing](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8d47043fe56afcb35273a898e6f20efbb2ab8c64/processor/redactionprocessor/config.go#L104). I think we should do the same here so that the behavior is matched

blewis12 (commented):
This looks good to me, apologies for the long delay in review! I left one small comment in there

edit: you'll need to run `make generate-gh-issue-templates` as per the failing CI [description](https://github.com/grafana/alloy/actions/runs/30243939330/job/89945492788?pr=6647)

internal/component/otelcol/processor/redaction/redaction.go
andreacappadona17: I opted for referencing directly the upstream component's hash function so that any behaviour is inherited from there.

blewis12: yeah I think that makes sense here

blewis12 (approved):

PR #6652, opened 2026-07-07
wal: recover from corrupt checkpoint on WAL truncation

### LINKED ISSUE #6166: prometheus.remote_write WAL truncation permanently disabled by single-byte checkpoint corruption
### Component(s)

prometheus.remote_write

### What's wrong?

A one-byte non-zero write into the page-padded region of any checkpoint segment file causes every subsequent truncate cycle to fail. WAL segments accumulate forever; the data path keeps working, so the failure is invisible in the standard remote_write counters.

This is an edge case which was discovered while trying to reproduce reported failures of Alloy related to this WAL, where the failures were suspected to relate to corrupted filesystems in some way.

Note: the only way to spot this issue seems to be in logs, I didn't find metrics that expose the condition.

### Steps to reproduce

I've attached a test environment that uses a stub service to receive Prometheus Remote Write data, for ease of reproduction. I also reproduced this separately while sending to Grafana Cloud, to verify the results in real Prometheus storage. Note that the wal settings have short truncate_frequency and keepalive times, again for ease of reproduction.

The repro-truncation-disabled.sh script runs the reproduction. The attached tarball has a docker-compose file and Alloy configuration.

[reproduction.tar.gz](https://github.com/user-attachments/files/27262113/reproduction.tar.gz)

```
mkdir -p ./data
docker compose up -d                           # start Alloy + stub
sleep 60                                       # warmup: at least one rotation
./repro-truncation-disabled.sh                 # write a non zero byte into the page-padded region of a checkpoint segment
```

## Finding: A single-byte corruption inside a checkpoint file permanently disables WAL truncation

**Repro:** `./repro-truncation-disabled.sh`

**One-line repro:** with at least one `checkpoint.NNN/` directory present, write any non-zero byte into the page-padded region (offset ≥ 32768) of a checkpoint segment file.

```sh
printf '\xff' | dd of=./data/.../wal/checkpoint.00000005/00000000 bs=1 count=1 seek=32768 conv=notrunc
```

**What happens**

Each truncation cycle (every `truncate_frequency`) tries to create a new checkpoint by reading the existing one and merging in newer segments. The read fails immediately on the corrupt page, so the new checkpoint is never written, and no segments are reaped. The same error fires every 30 s for the lifetime of the process. Throughput on the data path is unaffected — the watcher does not need to re-read this checkpoint to ship samples — but on-disk segments accumulate forever.

### System information

MacOS 26.4.1, Docker

### Software version

Alloy v1.16.0

### Configuration

```text
// Minimal Alloy config used by the WAL-corruption repros.
// We only need:
//   - A scrape source that produces enough series to grow the WAL and trigger
//     checkpoints on the schedule below.
//   - A remote_write to the local stub.
// The truncation knobs are tuned aggressively so a checkpoint exists within
// ~30 seconds of startup (default truncate_frequency is 2h).

prometheus.scrape "self" {
  targets         = [{"__address__" = "127.0.0.1:12345", "job" = "alloy-self"}]
  forward_to      = [prometheus.remote_write.stub.receiver]
  scrape_interval = "1s"
  scrape_timeout  = "500ms"
}

prometheus.remote_write "stub" {
  endpoint {
    url = "http://stub:8080/api/v1/push"
  }

  wal {
    truncate_frequency = "30s"
    min_keepalive_time = "10s"
    max_keepalive_time = "1m"
  }
}
```

### Logs

```text
ts=...:46:45 level=info msg="Creating checkpoint" from_segment=6 to_segment=8 ...
ts=...:46:45 level=warn msg="could not truncate WAL"
             err="create checkpoint: read segments: corruption in segment .../checkpoint.00000005/00000000
                  at 32768: unexpected non-zero byte in padded page"
ts=...:47:15 level=info msg="Creating checkpoint" from_segment=6 to_segment=8 ...
ts=...:47:15 level=warn msg="could not truncate WAL"
             err="... unexpected non-zero byte in padded page"

Sample throughput (data path unaffected):
  samples_out=113239  watcher_cur_seg=12
```

### Tip

harsh543:

Recover from a corrupt WAL checkpoint instead of permanently disabling
truncation.

### Pull Request Details

Issue #6166 describes a situation where a single corrupt byte in a
checkpoint file causes every subsequent WAL truncation to fail silently.
The WAL keeps growing until disk runs out, and there is no metric that
surfaces the problem — it only appears in logs.

The root cause is that `wlog.Checkpoint` reads the existing checkpoint
to merge it with newer WAL segments. If that read fails with a
`CorruptionErr`, the call returns immediately and stays broken on every
future truncation cycle.

The fix: when we see a `CorruptionErr` after a checkpoint failure, delete
the corrupt checkpoint file and retry. This is safe because checkpoint
data is always fully reconstructible from the WAL segments that are still
on disk. The retry produces a clean checkpoint from those segments alone.

Changes are in `wal.go` (~10 lines). Added a regression test in
`wal_test.go` that creates a WAL, runs `Truncate`, bitflips the
checkpoint file, then asserts the second `Truncate` succeeds and
produces a new checkpoint index.

Fixes #6166

### Notes to the Reviewer

The retry only fires once. If the second checkpoint attempt also fails
(e.g. disk full), the error is returned as normal and truncation stops
again — no infinite loop.

I modelled the approach on how Prometheus handles similar WAL corruption
scenarios. Happy to discuss if there is a preferred recovery pattern in
Alloy.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/static/metrics/wal/wal.go
+++ b/internal/static/metrics/wal/wal.go
@@ -657,8 +657,26 @@ func (w *Storage) Truncate(mint int64) error {
 		var cerr *wlog.CorruptionErr
 		if errors.As(err, &cerr) {
 			w.metrics.walCorruptionsTotal.Inc()
+			// A corrupt checkpoint file permanently blocks every subsequent truncation
+			// cycle because wlog.Checkpoint re-reads the existing checkpoint before
+			// writing a new one. The checkpoint is a compaction of WAL segments that
+			// are still present on disk, so deleting it is safe: the next
+			// wlog.Checkpoint call will rebuild from those segments instead.
+			if _, cpIdx, cpErr := wlog.LastCheckpoint(w.wal.Dir()); cpErr == nil {
+				if delErr := wlog.DeleteCheckpoints(w.wal.Dir(), cpIdx+1); delErr != nil {
+					w.logger.Error("could not delete corrupt checkpoint", "err", delErr)
+				} else {
+					w.logger.Warn("deleted corrupt checkpoint, retrying", "checkpoint_index", cpIdx, "corruption_err", err)
+					_, err = wlog.Checkpoint(w.logger, w.wal, first, last, keep, mint, false)
+					if err != nil {
+						w.metrics.checkpointCreationFail.Inc()
+					}
+				}
+			}
 		}
-		return fmt.Errorf("create checkpoint: %w", err)
+		if err != nil {
+			return fmt.Errorf("create checkpoint: %w", err)
+		}
 	}
 	if err := w.wal.Truncate(last + 1); err != nil {
 		// If truncating fails, we'll just try again at the next checkpoint.
--- a/internal/static/metrics/wal/wal_test.go
+++ b/internal/static/metrics/wal/wal_test.go
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/prometheus/tsdb/wlog"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/prometheus/prometheus/model/exemplar"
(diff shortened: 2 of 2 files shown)


harsh543:
Hey @kgeckhart @thampiotr 👋 — tagging you since you've both been active on the `prometheus.remote_write` / WAL area recently.

**What this fixes:** A one-line checkpoint file corruption permanently disables WAL truncation. Once `wlog.Checkpoint` fails reading a corrupt existing checkpoint, every subsequent truncation call short-circuits at the same error, causing the WAL to grow unboundedly until disk exhaustion.

**Root cause:** `wlog.Checkpoint` always reads the *current* checkpoint to merge it with new WAL segments. A corrupt read → permanent failure. But checkpoint data is always fully reconstructible from WAL segments on disk — so deleting the corrupt checkpoint and retrying is safe.

**The fix (3 additions to `wal.go`):**
1. Detect `wlog.CorruptionErr` on checkpoint failure
2. Delete the corrupt checkpoint via `wlog.DeleteCheckpoints`
3. Retry once — producing a clean checkpoint from segments alone

**Test added:** `TestTruncate_CorruptCheckpointRecovery` in `wal_test.go` — creates a WAL, appends series, runs `Truncate`, bitflips the checkpoint file, then asserts the second `Truncate` succeeds and advances the checkpoint index.

This matches how Prometheus itself handles similar WAL recovery scenarios. Happy to adjust the retry strategy or add more test cases if useful.

PR #6654, opened 2026-07-07
prometheus.scrape: add labels_to_hash for stable cluster ownership

### LINKED ISSUE #6484: prometheus.scrape: allow selecting target labels used for clustering ownership
### Request

When clustering is enabled, `prometheus.scrape` assigns each target to a node by hashing the target's labels (`NonMetaLabelsHash`, which keeps everything except `__meta_*`, so `__address__`, `__metrics_path__`, and `__param_*` are all part of the key). There's no way to restrict which labels feed that hash.

This breaks ownership whenever a target carries a **volatile** label that differs across nodes or over time. Each node runs service discovery independently, so if the discovered target identity isn't stable, the nodes compute different hashes for the same logical target and disagree on who owns it. The target then gets scraped by multiple nodes (duplicate samples) or none (gaps).

### Where we hit it

We scrape PlanetScale's Postgres metrics via `discovery.http`. Their discovery endpoint returns a pre-signed scrape URL per target with a fresh `?exp=…&sig=…` on every poll, which become `__param_exp` / `__param_sig` labels and end up in the ownership hash. With a 2-replica clustered metrics collector, each replica polls discovery on its own and gets a different signed URL for the same database, so they disagree on ownership and some databases are scraped twice while others aren't scraped at all.

We confirmed this via the component debug API (`/api/v0/web/components/prometheus.scrape.<id>`): the same database showed as owned by both replicas in one snapshot and neither in another, rotating each discovery refresh. Moving the scrape to a single non-clustered collector made the gaps disappear, which isolates it to clustered ownership. The signed URL is auth material, so we can't relabel it away before the scrape, and there's currently no config to keep it out of the hash.

### Proposed solution

Expose a config option to choose which target labels are used for the clustering ownership hash, falling back to the current behavior (all non-meta labels) when unset.

The mechanism already exists: `discovery.NewDistributedTargetsWithCustomLabels(...)` hashes an allowlist of labels via `SpecificLabelsHash(labels)`, and `loki.source.kubernetes` already uses it with a hardcoded `kubetail.ClusteringLabels` slice. The change for `prometheus.scrape` is to pass a user-configurable label list into that same constructor in `distributeTargets`. In our case, hashing on `__metrics_path__` (the stable per-database branch path) would shard deterministically and ignore the rotating `sig`/`exp`.

This also brings clustered scrape closer to what Prometheus already allows. Prometheus shards via `hashmod` on `source_labels` the operator chooses, so you control which labels determine a target's shard. Alloy clustering automates ownership but hardcodes the label set to all non-meta labels. This option gives that control back.

### Open design question

The `clustering` block is the shared `cluster.ComponentBlock` (currently just `enabled`), embedded by several components (`prometheus.scrape`, `pyroscope.scrape`, `prometheus.operator.*`, the loki sources). The option could live on the shared block so every clustered scraper benefits (consistent, but wider), or as a `prometheus.scrape`-local field (narrower). The problem isn't scrape-specific, so the shared block seems like the more natural home, but it touches more components. Happy to go whichever way you prefer.

### Willing to implement

Happy to open a PR with tests and docs once there's agreement on the config surface. Claiming this per the contributing guide.

harsh543:

Add `labels_to_hash` to the `clustering` block so users can restrict
which target labels feed the ownership hash.

### Pull Request Details

When clustering is enabled, `prometheus.scrape` assigns each target to
a node by hashing all non-meta labels (`__address__`, `__metrics_path__`,
`__param_*`, etc.). There is no way to exclude labels from that hash.

This breaks ownership whenever a target carries a volatile label that
differs per node or per discovery poll. Issue #6484 has a concrete
example: PlanetScale's discovery endpoint returns a pre-signed scrape URL
with a fresh `?sig=...` on every poll. Each Alloy replica polls discovery
independently and gets a different `__param_sig`, so they compute
different ownership keys for the same database and disagree on who should
scrape it — resulting in duplicates on some targets and gaps on others.

The infrastructure to fix this already exists:
`NewDistributedTargetsWithCustomLabels` and `SpecificLabelsHash` are both
in the discovery package but not exposed to the component config.

This PR adds an optional `labels_to_hash` attribute to `ComponentBlock`.
When set, only those labels are used for the ownership hash. When empty
(the default), behaviour is unchanged — all non-meta labels are hashed
as before.
On the cluster.ComponentBlock placement: I went back and forth on this — flagged it as an open question in #6484 before implementing. I put LabelsToHash on the shared block because the underlying problem (volatile discovery labels breaking ownership hashing) isn't specific to prometheus.scrape — pyroscope.scrape and the loki.source.* clustered components could hit the same failure mode with their own volatile labels.

That said, this PR only wires the field into prometheus.scrape's distributeTargets. I didn't touch the other five consumers because I don't have a concrete volatile-label case for them, and I'd rather not guess at behavior nobody's asked for. Happy to either (a) scope the field down to prometheus.scrape for now and move it to the shared block if/when another component needs it, or (b) leave it shared as forward-looking API surface with a comment noting it's currently only consumed by scrape. Leaning toward (a) — smaller surface, easier to review — but open to your call.

Example config:

```alloy
prometheus.scrape "db_metrics" {
  clustering {
    enabled        = true
    labels_to_hash = ["__metrics_path__", "__address__"]
  }
}
```

Added a test `TestLabelsToHash_VolatileLabelsIgnored` that creates two
targets with identical `__metrics_path__` but different `__param_sig`,
and asserts they collapse to one ownership key when `labels_to_hash` is
set.

Fixes #6484

### Notes to the Reviewer

No new hashing logic is introduced — this just threads the existing
`SpecificLabelsHash` path through the component config. The attribute is
optional with no default change, so existing deployments are unaffected.

I am aware of the feedback about the PR template and AI disclosure —
updated accordingly. The change itself I have reviewed and understand
end to end.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/scrape/scrape.go
+++ b/internal/component/prometheus/scrape/scrape.go
@@ -431,7 +431,7 @@ func (c *Component) Run(ctx context.Context) error {
 
 func (c *Component) distributeTargets(targets []discovery.Target, jobName string, args Arguments) (map[string][]*targetgroup.Group, []*scrape.Target) {
 	var (
-		newDistTargets        = discovery.NewDistributedTargets(args.Clustering.Enabled, c.cluster, targets)
+		newDistTargets        = discovery.NewDistributedTargetsWithCustomLabels(args.Clustering.Enabled, c.cluster, targets, args.Clustering.LabelsToHash)
 		oldDistributedTargets *discovery.DistributedTargets
 	)
 
--- a/internal/component/prometheus/scrape/scrape_clustering_test.go
+++ b/internal/component/prometheus/scrape/scrape_clustering_test.go
@@ -383,3 +383,55 @@ func (f *fakeCluster) Peers() []peer.Peer {
 func (f *fakeCluster) Ready() bool {
 	return true
 }
+
+// TestLabelsToHash_VolatileLabelsIgnored verifies that setting LabelsToHash
+// makes two targets with identical stable labels but different volatile labels
+// (e.g. pre-signed URL params that change per node) hash to the same shard key
+// and be treated as the same logical target (issue #6484).
+//
+// Without LabelsToHash, the volatile __param_sig label enters the hash and
+// the two targets get different keys → one is owned locally, one remotely,
+// producing gaps or duplicates in a two-replica cluster.
+//
+// With LabelsToHash = ["__metrics_path__"], the volatile label is excluded and
+// both targets hash identically → consistent ownership on both replicas.
+func TestLabelsToHash_VolatileLabelsIgnored(t *testing.T) {
+	stablePath := "/__internal/db1/metrics"
+
+	// target A: what replica 1 sees after its discovery poll (sig=aaaa)
+	tgtA := discovery.Target{
+		"__address__":      "db.example.com:9104",
+		"__metrics_path__": stablePath,
+		"__param_sig":      "aaaa",
+		"__scheme__":       "https",
+	}
+	// target B: what replica 2 sees — identical except the rotating signature
+	tgtB := discovery.Target{
+		"__address__":      "db.example.com:9104",
+		"__metrics_path__": stablePath,
+		"__param_sig":      "bbbb",
+		"__scheme__":       "https",
+	}
+
+	t.Run("without LabelsToHash volatile labels cause different hash", func(t *testing.T) {
+		dt := discovery.NewDistributedTargets(true, nil, []discovery.Target{tgtA, tgtB})
+		// When clustering is enabled but cluster is nil, disabledCluster is used
+		// and all targets are assigned locally — but both must appear as separate
+		// targets because they hash differently.
+		require.Equal(t, 2, dt.TargetCount(),
+			"targets with different volatile labels must be treated as distinct (old behaviour)")
+	})
+
+	t.Run("with LabelsToHash volatile labels are ignored", func(t *testing.T) {
+		dt := discovery.NewDistributedTargetsWithCustomLabels(
+			true, nil,
+			[]discovery.Target{tgtA, tgtB},
+			[]string{"__metrics_path__"},
+		)
+		// With LabelsToHash restricting the hash to __metrics_path__ only, both
+		// targets produce the same key and are de-duplicated to one logical target.
+		require.Equal(t, 1, dt.TargetCount(),
+			"targets sharing stable labels must collapse to one with LabelsToHash set")
+	})
+}
+}
--- a/internal/service/cluster/cluster.go
+++ b/internal/service/cluster/cluster.go
@@ -127,6 +127,22 @@ type Component interface {
 // "clustering".
 type ComponentBlock struct {
 	Enabled bool `alloy:"enabled,attr"`
+
+	// LabelsToHash, when non-empty, restricts the set of target labels used to
+	// compute the ownership hash for clustered target distribution. Only the
+	// listed label names are hashed; all other labels are ignored for sharding
+	// purposes.
+	//
+	// Use this when targets carry volatile labels (e.g. pre-signed URL
+	// parameters such as __param_sig or __param_exp) that differ across nodes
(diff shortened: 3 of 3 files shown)


blewis12:
Thanks for your contribution @harsh543 - from your PR description (and a couple other PR descriptions recently opened), it looks like this PR was generated with claude code - please refer to our [policy](https://github.com/grafana/alloy/blob/main/docs/developer/genai.md) on this. We welcome contributions that lean on AI for implementation, but we ask that PR's themselves both conform to our PR templates and ideally have a description/reasoning written by yourself

harsh543:
Hey @kgeckhart @thampiotr — this one's in the `prometheus.scrape` clustering path, tagging you for awareness.

**The problem:** In a clustered scrape setup, `NewDistributedTargets` hashes *all* non-meta target labels to determine ownership. Any volatile label — pre-signed URL params, rotating auth tokens, timestamp-based signatures — causes the same logical target to hash differently across replicas, breaking consistent ownership: one replica thinks it owns the target, the other does too, producing duplicate scrapes or gaps.

**The fix:** Exposes `labels_to_hash` on `cluster.ComponentBlock` (optional, defaults to existing full-hash behaviour). When set, `NewDistributedTargetsWithCustomLabels` (already exists in discovery) uses `SpecificLabelsHash` instead of `NonMetaLabelsHash` — restricting the ownership key to stable labels only.

**No new hashing logic:** The infrastructure (`NewDistributedTargetsWithCustomLabels`, `SpecificLabelsHash`) was already in place. This PR just wires the field through `ComponentBlock` → `distributeTargets`.

**Test:** `TestLabelsToHash_VolatileLabelsIgnored` — two targets sharing `__metrics_path__` but differing on `__param_sig`: asserts 2 distinct keys without `labels_to_hash`, 1 key with `labels_to_hash=["__metrics_path__"]`.

Config ergonomics example:
```alloy
prometheus.scrape "db_metrics" {
  clustering {
    enabled       = true
    labels_to_hash = ["__metrics_path__", "__address__"]
  }
}
```

Fixes #6484. Happy to add docs or expand test cases.

blewis12:
@harsh543 To be clear, what we're asking is that the PR description and any comments are written in your own words and reflect your own understanding of the change, rather than pasted-in generated text that pings maintainers.

please take this as a friendly but firm heads-up: if AI-generated comments and maintainer tagging continue across your PRs, we'll close them under the policy and may restrict further contributions. We'd rather keep working with you on this, so if you re-engage in your own words we're happy to review/discuss the change.

What i'd like to see in this PR is reasoning in your own words on why you took the approach you did (e.g. embedding the  new config  in `cluster.ComponentBlock` vs just `prometheus.scrape` as was raised in the related issue), and the tradeoffs/impact of those decisions. This can help us understand where you're coming from and engage in discussion as part of the review.

harsh543:
@blewis12  #6484 is a distributed observability correctness bug—two Alloy replicas can disagree about which node owns a scrape target, causing duplicate metrics or missing metrics. Since I have had similar experiences in the past where I have worked on GPU reliability for Inference and training workfloads so that's why the distributed observability makes relevance to my career. I have added my personal motivation and understanding in PR description.

thampiotr:
Hi @harsh543 

> is a distributed observability correctness bug

as per our documentation https://grafana.com/docs/alloy/next/get-started/clustering/

> For target auto-distribution to work:
> 
> 1. All Alloy deployments in the same cluster must access the same service discovery APIs.
> 2. All deployments must scrape the same targets.

So this is not a bug. I see that https://github.com/grafana/alloy/issues/6484 is open and perhaps it should become a proposal that we can discuss with the team. Clustering is critical in many deployments and there are some other considerations we would like to take into account before moving to implementation. 

I'm going to close this PR _for now_, which we can re-open in the future if the proposal / issue reaches a conclusion. Note we have also this doc about the process: https://github.com/grafana/alloy/blob/main/docs/design/908-proposal-process.md

PR #6655, opened 2026-07-07
prometheus/operator: fail when managers stop unexpectedly

### LINKED ISSUE #6586: prometheus.operator.*  silently stops discovering targets when discovery/informer dies at runtime (no recovery, pod stays Ready)
### Component(s)

prometheus.operator.servicemonitors, prometheus.operator.probes, prometheus.operator.podmonitors

### What's wrong?

The prometheus.operator.* components can permanently stop discovering targets after a healthy start, with no recovery.

In crdManager.Run(), the discovery and scrape managers are launched as fire-and-forget goroutines. If one dies at runtime (e.g. a transient apiserver error such as "no route to host" to the kubernetes ClusterIP), the error is only logged and the goroutine exits:

    go func() {
        err := c.discoveryManager.Run()
        if err != nil {
            level.Error(c.logger).Log("msg", "discovery manager stopped", "err", err)
        }
    }()

There is no restart or retry. Because a dead component does not crash the Alloy process, the pod stays Running/Ready while reporting 0 scrape targets indefinitely.

Observed in production: a pod sat with 0 ServiceMonitor targets and 0 restarts for 13 days; a manual `rollout restart` fixed it instantly. Meanwhile `kube_node_info` kept flowing from a separate kube-state-metrics source, so the cluster looked alive while reporting 0% availability on our SLO dashboards.

This is distinct from #1853 / #3688, which only added retry to informer *initialization* ("On node restart, the API server is not always immediately available"). That fix does not cover a manager/informer that dies *after* successful startup. I verified the runtime handling in crdmanager.go is identical across v1.12.0, v1.16.3, and main.

Suggested fix: detect runtime death of the discovery/scrape managers (or the informer cache) and either restart them with backoff, or surface it as component-unhealthy / return from Run() so the supervisor recreates the component — the existing code comment already notes "on update, this entire thing should be recreated and restarted".

### Steps to reproduce

1. Configure prometheus.operator.servicemonitors and let it discover targets (prometheus_sd_discovered_targets{...servicemonitors...} > 0).
2. After it's healthy, induce a transient apiserver connectivity failure that kills the discovery manager / watch (e.g. briefly break the route to the kubernetes ClusterIP ).
3. Observe: the pod stays Ready, prometheus_sd_discovered_targets{...servicemonitors...} drops to 0, and it never recovers without a manual pod restart.

### System information

Linux x86_64 (Kubernetes)

### Software version

Grafana Alloy v1.12.0

### Configuration

```text
prometheus.operator.servicemonitors "default" {
  forward_to = [prometheus.remote_write.default.receiver]
}

prometheus.remote_write "default" {
  endpoint {
    url = "http://<remote-write-endpoint>/api/v1/write"
  }
}
```

### Logs

```text

```

### Tip

harsh543:

Make `prometheus.operator.*` components exit (and restart) when the
internal discovery or scrape manager dies at runtime.

### Pull Request Details

Issue #6586 describes a production incident: a pod sat with 0
ServiceMonitor targets and 0 restarts for 13 days after a transient
apiserver blip killed the discovery goroutine. The pod stayed Ready the
whole time because the component did not crash — it just silently stopped
discovering targets.

The problem is in `crdManager.Run()`: the discovery and scrape managers
are launched as fire-and-forget goroutines. If either exits, the error is
logged and the goroutine is gone, but `Run` keeps blocking. Alloy's
supervisor never sees an error and never restarts the component.

The fix replaces the two fire-and-forget goroutines with an `errgroup`.
If either manager returns an error, the errgroup cancels the shared
context and `Run` returns that error. Alloy's component supervision then
restarts the component, which re-initialises discovery from scratch.

Also added a regression test that injects a failing manager and asserts
`Component.Run` returns a non-nil error.

Fixes #6586

### Notes to the Reviewer

The context cancellation on manager exit means the other manager (scrape
or discovery) also gets a cancel signal and shuts down cleanly before
`Run` returns — no goroutine leak.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/prometheus/operator/common/component.go
+++ b/internal/component/prometheus/operator/common/component.go
@@ -89,6 +89,10 @@ func (c *Component) Run(ctx context.Context) error {
 			return nil
 		case err := <-errChan:
 			c.reportHealth(err)
+			if cancel != nil {
+				cancel()
+			}
+			return err
 		case <-c.onUpdate:
 			c.mut.Lock()
 			manager := c.crdManagerFactory.New(c.opts, c.cluster, c.opts.Logger, c.config, c.kind, c.ls)
--- a/internal/component/prometheus/operator/common/component_test.go
+++ b/internal/component/prometheus/operator/common/component_test.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
@@ -55,6 +56,32 @@ func (c *crdManagerHungRun) DebugInfo() any {
 }
 
 func (c *crdManagerHungRun) GetScrapeConfig(ns, name string) []*config.ScrapeConfig {
+	return nil
+}
+
+type crdManagerFactoryErrorRun struct {
+	err error
+}
+
+func (m crdManagerFactoryErrorRun) New(_ component.Options, _ cluster.Cluster, _ *slog.Logger, _ *operator.Arguments, _ string, _ labelstore.LabelStore) crdManagerInterface {
+	return &crdManagerErrorRun{err: m.err}
+}
+
+type crdManagerErrorRun struct {
+	err error
+}
+
+func (c *crdManagerErrorRun) Run(_ context.Context) error {
+	return c.err
+}
+
+func (c *crdManagerErrorRun) ClusteringUpdated() {}
+
+func (c *crdManagerErrorRun) DebugInfo() any {
+	return nil
+}
+
+func (c *crdManagerErrorRun) GetScrapeConfig(ns, name string) []*config.ScrapeConfig {
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


harsh543:
Hey @kgeckhart @jharvey10 — tagging you on this one since it touches `prometheus/operator`.

**What this does:** When a manager goroutine inside `prometheus.operator.*` exits unexpectedly (panic, context-unrelated error), the component currently silently degrades — it stops reconciling without surfacing any signal to the Alloy runtime or the user.

This PR makes the component's `Run` method return the manager's error instead, so Alloy can restart the component and the error surfaces in logs/UI.

Straightforward change: replace the fire-and-forget goroutine pattern with an errgroup so any unexpected manager exit propagates up to `Run`.

Happy to walk through the before/after or adjust the error-propagation strategy if there's a preferred pattern in the codebase.

PR #6658, opened 2026-07-08
feat(discovery.aws): Add AWS unified service discovery component

### LINKED ISSUE #5160: AWS Service Discovery (discovery.aws)
### Component(s)

_No response_

### Request

Prometheus has recently added AWS Service Discovery https://github.com/prometheus/prometheus/pull/17406.

This is intended to work similar to the Kubernetes Service Discovery, where a user specifies a `role` and then this dictates the resources that they discover. Currently this supports EC2, Lightsail and ECS, but there are plans to add more (MSK and Elasticache). 

I'd like to contribute this as it will be expanded in the future, and currently it will give users the ability to discover ECS services .

To be clear this **isn't** intended to replace the existing `discovery.ec2` or `discovery.lightsail` as these are still supported in Prometheus.

So I propose we add a new component called `discovery.aws`.

### Use case

Currently there is no way of doing Service Discovery in ECS unless you write a http service. 

One of the benefits of AWS Service Discovery is that it should be a lot easier to add more services in the future, eg MSK, Elasticache, RDS etc. Many exporters in `prometheus.exporter.*` already support [multi-target](https://prometheus.io/docs/guides/multi-target-exporter/) and this would mean users can pass targets directly to these when the services are added.

### Tip

kgeckhart:

Adds a `discovery.aws` component that wraps Prometheus's unified AWS service discovery, where a `role` argument selects which service to discover: `ec2`, `ecs`, `elasticache`, `lightsail`, `msk`, or `rds`. This is the way to discover Amazon ECS targets, and it doesn't replace the existing `discovery.ec2` or `discovery.lightsail`.

Fixes #5160

### Notes to the Reviewer

Converter support (`aws_sd_configs` → `discovery.aws` for `alloy convert`) is deliberately deferred to a follow-up, which is why "Config converters updated" is left unchecked.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


internal/component/discovery/aws/aws.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if args.Region == "" {
+		region, err := resolveRegion()
+		if err != nil {
+			return err
+		}
+		args.Region = region
+	}
+
```
jharvey10, line 229: Should there be a guard here too for `request_concurrency`? For example if a user wrote a config like this:

```
discovery.aws "example" {
  role               = "ec2"
  region             = "us-east-1"
  request_concurrency = 50
}
```

`request_concurrency` would get silently dropped and the user would have no indication that their configuration had no effect.

jharvey10 (commented):
Looks good! Just one comment about validation.

internal/component/discovery/aws/aws.go
kgeckhart: I might be blind but I can't find where we are going to ignore the users value. AFAICT we take the configured value or default if it's zero. 

edit: Is it because it's odd we are doing work only for region here? If it's empty we do have a way to try to infer it via loaded config / other things like upstream does.

jharvey10: `promaws.EC2SDConfig` and `promaws.LightsailSDConfig` don't appear to have a `RequestConcurrency` field at all. So for those two roles, `args.RequestConcurrency` is never referenced anywhere in `Convert()`.

So right now, `Validate()` says "yep looks good", then `Convert()` builds an `EC2SDConfig` which has no `RequestConcurrency` field, so the value is silently dropped.

It's just that a user combining `request_concurrency` with the ec2 role is misunderstanding their config settings, and we have the potential to help out and say "hey actually that field doesn't apply to what you're trying to do".

kgeckhart: Got it! Fixed it and Clusters which has the same problem

jharvey10 (approved):
🚀

PR #6677, opened 2026-07-10
feat: Check imports of inline Alloy config in OTel mode

### LINKED ISSUE #6420: Explore Implementation for Inline Config in Alloy Engine Extension

x1unix:

Check configs imported by `import.*` components.

### Pull Request Details

This PR is a follow-up for #6442

The OTel extension mode imposes certain limitations:

- `module_path` keyword resolves to a current work dir, if `config.inline.module_path` is undefined.
  - As this behavior diverges from Alloy mode, a warning log is thrown.
- The `remotecfg` block cannot be used.

This PR extends config check for files that can be imported using `import.*` components.

#### Notes

- The `remotecfg` block can't be used inside imports and their `declare` blocks, therefore the import hook only validates the `module_path` usage.
- Only the `import.string` inherits global `module_path`, therefore warning isn't thrown for the rest of `import.*` components as they define their own path based on import URL.

Fixes #6420

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -272,7 +272,7 @@ In the meantime, use the CLI or Helm options for testing.
 1. **Server ports**: The {{< param "DEFAULT_ENGINE" >}} exposes its HTTP server on port `12345`.
    The {{< param "OTEL_ENGINE" >}} exposes its HTTP server on port `8888`.
    The {{< param "OTEL_ENGINE" >}} HTTP server doesn't expose a UI, support bundles, or reload endpoint functionality like the {{< param "DEFAULT_ENGINE" >}} does.
-1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` Alloy config keyword resolves to the process current working directory.
+1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` Alloy configuration keyword resolves to the process current working directory.
 1. **Fleet management**: [Grafana Fleet Management](https://grafana.com/blog/opentelemetry-and-grafana-labs-whats-new-and-whats-next-in-2026/#fleet-management) doesn't support the {{< param "OTEL_ENGINE" >}} yet.
    You must define and manage the input configuration manually.
 
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -272,7 +272,7 @@ In the meantime, use the CLI or Helm options for testing.
 1. **Server ports**: The {{< param "DEFAULT_ENGINE" >}} exposes its HTTP server on port `12345`.
    The {{< param "OTEL_ENGINE" >}} exposes its HTTP server on port `8888`.
    The {{< param "OTEL_ENGINE" >}} HTTP server doesn't expose a UI, support bundles, or reload endpoint functionality like the {{< param "DEFAULT_ENGINE" >}} does.
-1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` Alloy configuration keyword resolves to the process current working directory.
+1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` {{< param "PRODUCT_NAME" >}} configuration keyword resolves to the process current working directory.
 1. **Fleet management**: [Grafana Fleet Management](https://grafana.com/blog/opentelemetry-and-grafana-labs-whats-new-and-whats-next-in-2026/#fleet-management) doesn't support the {{< param "OTEL_ENGINE" >}} yet.
    You must define and manage the input configuration manually.
 
--- a/docs/sources/set-up/otel_engine.md
+++ b/docs/sources/set-up/otel_engine.md
@@ -89,9 +89,10 @@ Since the {{< param "DEFAULT_ENGINE" >}} isn't running, the UI and metrics aren'
 
 You can also run the {{< param "OTEL_ENGINE" >}} with the {{< param "DEFAULT_ENGINE" >}}.
 Modify your YAML configuration to include the `alloyengine` extension.
-This extension accepts a path to the {{< param "DEFAULT_ENGINE" >}} configuration and starts a {{< param "DEFAULT_ENGINE" >}} pipeline alongside the {{< param "OTEL_ENGINE" >}} pipeline.
+This extension accepts a path to the {{< param "DEFAULT_ENGINE" >}} configuration or an inline {{< param "DEFAULT_ENGINE" >}} configuration.
+It starts a {{< param "DEFAULT_ENGINE" >}} pipeline alongside the {{< param "OTEL_ENGINE" >}} pipeline.
 
-The following example shows the configuration:
+The following example uses `config.path` to load the {{< param "DEFAULT_ENGINE" >}} configuration from a file or directory:
 
 ```yaml
 extensions:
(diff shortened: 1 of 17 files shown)


docs/sources/set-up/otel_engine.md
```diff
@@ -258,6 +272,7 @@ In the meantime, use the CLI or Helm options for testing.
 1. **Server ports**: The {{< param "DEFAULT_ENGINE" >}} exposes its HTTP server on port `12345`.
    The {{< param "OTEL_ENGINE" >}} exposes its HTTP server on port `8888`.
    The {{< param "OTEL_ENGINE" >}} HTTP server doesn't expose a UI, support bundles, or reload endpoint functionality like the {{< param "DEFAULT_ENGINE" >}} does.
+1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` Alloy config keyword resolves to the process current working directory.
```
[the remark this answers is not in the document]
x1unix: Fixed in https://github.com/grafana/alloy/pull/6677/changes/9c0c54a6013680bb0f5564a9e4b18b6d2ec4af4b

docs/sources/set-up/otel_engine.md
```diff
@@ -258,6 +272,7 @@ In the meantime, use the CLI or Helm options for testing.
 1. **Server ports**: The {{< param "DEFAULT_ENGINE" >}} exposes its HTTP server on port `12345`.
    The {{< param "OTEL_ENGINE" >}} exposes its HTTP server on port `8888`.
    The {{< param "OTEL_ENGINE" >}} HTTP server doesn't expose a UI, support bundles, or reload endpoint functionality like the {{< param "DEFAULT_ENGINE" >}} does.
+1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` Alloy configuration keyword resolves to the process current working directory.
```
clayton-cornell:
```suggestion
1. **Inline configuration module path**: If `config.inline.module_path` isn't defined, the `module_path` {{< param "PRODUCT_NAME" >}} configuration keyword resolves to the process current working directory.
```

internal/runtime/internal/controller/node_config_import.go
```diff
@@ -221,6 +219,15 @@ func (cn *ImportConfigNode) onContentUpdate(importedContent map[string]string) {
...
+		if cn.globals.OnImportContent != nil {
+			if err := cn.globals.OnImportContent(f, parsedImportedContent, cn.source); err != nil {
+				cn.logger.Error("imported content rejected", "file", f, "err", err)
+				cn.setContentHealth(component.HealthTypeUnhealthy, fmt.Sprintf("imported content from %q rejected: %s", f, err))
+				return
+			}
+		}
+
 		// populate importedDeclares and importConfigNodesChildren
 		err = cn.processImportedContent(parsedImportedContent)
```
thampiotr, line 227: This will already return `return fmt.Errorf("only declare and import blocks are allowed in a module")` - so do we have some redundant checks? 

Could this be simplified?

extension/alloyengine/extension.go
```diff
@@ -325,24 +333,80 @@ func buildAlloyConfig(cfg AlloyConfig) (modulePath string, files map[string][]by
...
+		if name := block.GetBlockName(); name == remotecfg.ServiceName {
+			return fmt.Errorf("block %q is not supported in Alloy extension", name)
 		}
 	}
-
 	return nil
 }
+
+// newImportContentHook returns a hook that validates imported module content for extension mode.
+// It rejects unsupported blocks (e.g. remotecfg). When warnModulePathDefault is set, it also warns
```
thampiotr: A boolean parameter that flips behaviour is a mild code smell IMO. Could we have callers emit the warning, or split this into two checks? 

Any ohter ways to simplify?

extension/alloyengine/extension.go
```diff
@@ -325,24 +333,80 @@ func buildAlloyConfig(cfg AlloyConfig) (modulePath string, files map[string][]by
...
+// if an imported module references module_path while it still resolves to the defaulted (cwd)
+// value of an inline config.
+//
+// initialModulePath is the module_path used in the root Alloy config.
+func newImportContentHook(logger *zap.Logger, warnModulePathDefault bool, initialModulePath string) importsource.ImportContentHook {
```
thampiotr: Also we use warnModulePathDefault, warnIfModulePathUsed, and usesDefaultedModulePath - can we standardise on one if we really need to keep this?

extension/alloyengine/extension_test.go
```diff
@@ -127,26 +134,17 @@ func newRetryTrackingCommand(failCount int, err error) (func(string, map[string]
 	return factory, state
 }
 
-func TestConfig_MissingContent(t *testing.T) {
```
blewis12, line 130: how come this test was removed?

x1unix: @blewis12 I've moved the test to `config_test.go` and extended it.

internal/runtime/internal/controller/node_config_import.go
x1unix: @thampiotr thanks for feedback. Removed redundant check in https://github.com/grafana/alloy/pull/6677/changes/c68e21cf3b4841772efa3d445b03baa931b67059

extension/alloyengine/extension.go
x1unix: Done in https://github.com/grafana/alloy/pull/6677/changes/c68e21cf3b4841772efa3d445b03baa931b67059

extension/alloyengine/extension.go
x1unix: Addressed together with a comment above

flowcmd/flowcmd.go
```diff
@@ -40,9 +41,12 @@ func RootCommand() *cobra.Command {
...
 // RunAsExtensionCommand returns a standalone cobra command to run Alloy inside Otel collector.
-func RunAsExtensionCommand(modulePath string, configs map[string][]byte) *cobra.Command {
+//
+// configImportHook is a hook that called when `import.*` component loads additional config files.
+func RunAsExtensionCommand(modulePath string, configs map[string][]byte, configImportHook importsource.ImportContentHook) *cobra.Command {
 	return alloycli.NewRunAsExtensionCommand(alloycli.ExtensionModeParams{
-		Configs:    configs,
-		ModulePath: modulePath,
+		Configs:        configs,
+		ModulePath:     modulePath,
```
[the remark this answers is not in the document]
thampiotr: I do not expect that this is something used widely, but also an upgrade path is trivial. So a false positive imo.

extension/alloyengine/README.md
```diff
@@ -17,19 +17,19 @@ The extension accepts the following configuration fields:
 
 ### Config Object
 
-The `config` object specifies the Alloy configuration source. Either `path` or `inline` must be set, but not both.
+The `config` object specifies the Alloy configuration source. Exactly one of `path` or `inline.content` must be set.
```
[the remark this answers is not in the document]
thampiotr: Might be good to clarify this.

thampiotr (dismissed):

thampiotr (approved):

clayton-cornell (approved):

blewis12 (approved):

PR #6695, opened 2026-07-14
feat: Implement Beyla as a subprocess (simple)

rafaelroquetto:
### (Not So) Brief description of Pull Request

This is an alternative to https://github.com/grafana/alloy/pull/6477

https://github.com/grafana/alloy/pull/6477 features a generator that downloads the Beyla config schema, and auto-generates both the Alloy facing and the Beyla facing config structs/YAML, at the expense of having to maintain a code generator - please see https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/README.md

The generator also outputs a reference file: https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/sample_config.alloy

In the general case, it mirrors the beyla config struct verbatim into Alloy. For example, Beyla has:

```
ebpf:
  payload_extraction:
    http:
      genai:
        openai:
          enabled: true
```

for which it would generate something like:

```
ebpf {
  payload_extraction {
    http {
      genai {   
        openai {
          enabled = true
        }
      }
    }
  }
}
```

However, that is not always the case with the Alloy side of things, for legacy reasons or otherwise - sometimes the Alloy config side does not 1-1 map to Beyla config side, so the generator needs to account for that - these exceptions are specified in https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/mappings.json

The file https://github.com/grafana/alloy/blob/beyla_hackathon/internal/component/beyla/ebpf/internal/config/gen/mappings_examples.md outlines what these differences are.

Bumping the Beyla version is then pretty much a matter of bumping the `BEYLA_VERSION` makefile var - the generator will do the rest.

----
### What this (simple) PR does

It trades off the complexity of the generator for a very obvious hand-rolled config implementation similar to the original (pre-subprocess) `beyla.ebpf` approach featuring multiple `Convert()` functions. The difference is that these `Convert()` functions fill it a `map` that will be serialised as the final Beyla YAML config rather than filling in the Beyla config struct themselves (to avoid pulling Beyla as a dependency, which would make the subprocess a moot point).

The code is very straightforward to maintain. The Beyla config schema is still downloaded, but it is just used as a test fixture to validate the current handl-rolled implementation - if a newly added field to the Beyla config is missing, the test will fail (and thus CI will fail). This PR demoes exactly that by bumping the Beyla version to 3.28 but missing to update the code to cover a few new fields. To run the test:

```
go test ./internal/component/beyla/ebpf/...
```

It will fail telling us exactly what is missing:
```
+ebpf/payload_extraction/http/enrichment/rules/match/response_status_code
```

(so we missed adding the code to handle that configuration field)

See https://github.com/grafana/alloy/actions/runs/29375745246/job/87228709577?pr=6695

### PR Checklist

- [x] Documentation added
- [x] Tests updated


Makefile
ptodev, line 1: We are currently migrating to Taskfile. Could you add the new targets as Tasks instead? We can still have make targets that reference the task targets, although I'm not sure if it's necessary.

internal/component/beyla/ebpf/internal/config/gen/download.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	for _, p := range paths {
+		if _, err := os.Stat(p); err != nil {
+			return false
+		}
+	}
+	return true
+}
+
+// fetchAssetDigests returns each release asset's published "sha256:..." digest,
```
ptodev, line 181: Can we store the sha in the repo, rather then retrieving it? This would give us guarantees that the file is the one we expect. It'd be similar to go.sum. 

I recently implemented a similar mechanism for a [zip file we download](https://github.com/grafana/alloy/blob/69cd783eb474aad6518150cc688d08c04745653e/.github/workflows/integration-test-windows-service.yml#L58-L68) to protect the CI from supply chain attacks.

internal/component/beyla/ebpf/internal/config/gen/README.md
ptodev, line 1: It's ok for this file to state how to use this utility. However, the doc on upgrading beyla need to be in the [docs/developer](https://github.com/grafana/alloy/tree/69cd783eb474aad6518150cc688d08c04745653e/docs/developer) folder.

docs/sources/_index.md.t
```diff
@@ -8,7 +8,7 @@ cascade:
   OTEL_VERSION: v0.153.0
   PROM_WIN_EXP_VERSION: v0.31.3
   SNMP_VERSION: v0.29.0
-  BEYLA_VERSION: v3.9.2
+  BEYLA_VERSION: v3.28.0
```
ptodev, line 11: It's a bummer how the beyla version is repeated several times, and sometimes is out of date. It'd be nice to have only one copy or to have the CI double check the correctness, but this is definitely not necessary for this PR. It's so large as it is 😅

internal/component/beyla/ebpf/internal/subprocess/fd_linux.go
ptodev, line 1: Can some of the subprocess code be moved to a more generic subprocess utility? It'd make it more clear what the contract is between Alloy and the subprocessed binary, and it'd make it more reusable for the future.

internal/component/beyla/ebpf/internal/config/config_build.go
```diff
@@ -0,0 +1,904 @@
+package config
+
+// Convert builds the discovery configuration.
+func (d Discovery) Convert() map[string]any {
```
ptodev, line 4: Why are we creating `map[string]any`? Is it not enough just to convert to the yaml struct from the upstream Beyla repo, and then marshal the config to yaml that we pass to the subprocess?

internal/component/beyla/ebpf/watchdog_linux.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+		},
+	}
+	defer client.CloseIdleConnections()
+
+	// Beyla loads eBPF programs before opening its Prometheus port (~15-20s)
+	select {
+	case <-ctx.Done():
+		return nil
+	case <-time.After(20 * time.Second):
```
ptodev, line 31: Should the 20 seconds be configurable? Can this take longer on certain hosts?

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -1148,379 +308,184 @@ func getLatestArgsFromChannel[A any](ch chan A, current A) A {
...
-	return attr, nil
-}
+func (c *Component) scheduleRestart() {
+	backoff, count, ok := c.subprocess.NextBackoff()
 
-func stringToGlobAttr(s string) (services.GlobAttr, error) {
-	if s == "" {
-		return services.GlobAttr{}, nil
+	if !ok {
+		c.opts.Logger.Error("Beyla subprocess exceeded maximum restart attempts, giving up", "max_restarts", subprocess.MaxRestarts)
```
ptodev: Is this in the Run() function? Should that ever give up? Can it not keep running until the context is done?

internal/component/beyla/ebpf/beyla_linux.go
ptodev, line 1: This file feels bigger than necessary. Ideally I'd like to have a short file where the we call an abstraction that's well documented and explains how Beyla is going to be ran as a sub process, and beyla_linux just feeds it some parameters. Then it's clear what the expected behaviour will be from a user point of view.

internal/component/beyla/ebpf/exec_linux.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return binPath, func() { unix.Close(fd) }, nil
+}
+
+func findFreePort() (int, error) {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return 0, err
+	}
+	defer listener.Close()
```
ptodev, line 110: In theory something else can acquire this port between the time that it's closed and the time that it gets aquired by the function that called `findFreePort()`. Can Beyla just bind to `127.0.0.1:0`?

internal/component/beyla/ebpf/exec_linux.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"go.uber.org/atomic"
+	"golang.org/x/sys/unix"
+)
+
+var socketCounter atomic.Uint64
+
+func abstractSocketAddr(role, componentID string) string {
+	return fmt.Sprintf("@alloy-beyla-%s-%s-%d-%d", role, componentID, os.Getpid(), socketCounter.Add(1))
+}
```
ptodev, line 24: I suppose we are not worried about man in the middle attacks related to the security of the sockets? I can't think of a way it can be explorable, although in general it's best to avoid ingesting unauthenticated data.

internal/component/beyla/ebpf/internal/config/config_build.go
ptodev: Btw I hope that we can move the Beyla component to the OTel Engine some day. Then we can just pass through yaml config as it is, without converting it from Alloy syntax.

internal/alloycli/resources_collector.go
ptodev, line 1: IIUC, the issue here is that beyla and alloy use different collectors for their process metrics? Even if we make the help strings the same, do we know that the numbers mean the same thing?  It feels like we're making a lot of assumptions here. Can we just prefix the name of the beyla metrics with something? Like "beyla_subproc"? Alternatively, Alloy could use a collector which can also track metrics for its subprocesses? That way you also know all the metrics are from the same time.

internal/component/beyla/ebpf/internal/config/config_build.go
rafaelroquetto: So, unfortunately we cannot import the upstream Beyla struct - as soon as we do it, we end up importing (parts of Beyla), which generates transient dependencies between Alloy and Beyla (i.e. beyla/obi will appear in the go.mod) - which brings back the dependency hell and defeats the purpose of this PR.

Makefile
rafaelroquetto: I am not too familiar with that, so I may need help. Claude told me this introduces a dependency on `mise`, is that truly the case?

internal/component/beyla/ebpf/internal/config/gen/download.go
rafaelroquetto: Yeah that is a good point - I changed it so now the sha256 are stored alongside the schema - it can be updated manually or with `make update-beyla-checksums` (which will fetch them online).

internal/component/beyla/ebpf/internal/config/gen/README.md
rafaelroquetto: done

docs/sources/_index.md.t
rafaelroquetto: Yeah, it kind of sucks, because to fix it we'd need somehow to have both the docs and the non-docs files to derive the version from somewhere. Maybe we can come up with something more elaborate, but for now I extended sync-beyla-docs-version to keep both `_index.md.t` and `_index.md` in sync from the `Makefile`'s `BEYLA_VERSION`, so `make beyla` handles it.

internal/component/beyla/ebpf/internal/subprocess/fd_linux.go
rafaelroquetto: I've split/reorganised the code - not everything got moved into a reusable component, because some things are Beyla specific, but hopefully it makes more sense now

internal/component/beyla/ebpf/watchdog_linux.go
rafaelroquetto: I don't think so, in all my tests this has been more than enough - this is not a kill timeout, but a grace period before the health-probing of the subprocess stats because Beyla loads its eBPF programs before it opens the health port (~15–20s). The actual "unresponsive -> restart" threshold is much later -> after the 20s we probe every 10s and only restart after 5 consecutive failures, so a subprocess effectively gets ~70s to become responsive. A genuine early crash (the process exiting) is caught immediately by the Wait() on the subprocess, independent of this grace period, the 20s only delays detecting a process that's running but hung.

I reckon almost no one will touch this if it were made a config.

internal/component/beyla/ebpf/beyla_linux.go
rafaelroquetto: This was the original behaviour - but during my tests, I noticed that in the cases in which it would crash indefinitely (i.e. forever), it would just be polluting the POD logs (or whatever logs). So I explicitly added a more than reasonable (IMHO) max attempts - if the subprocess crashes more than N times, then we have other issues at this stage (and the backup period is exponential for this very same reason)

internal/component/beyla/ebpf/beyla_linux.go
rafaelroquetto: Done as part of the reshuffling

internal/component/beyla/ebpf/exec_linux.go
rafaelroquetto: Beyla cannot, it's unfortunate. @kgeckhart raise dthe same point on the other original PR, so I will just paste here the answer I have there for the record:

"yeah that's indeed unfortunate - the problem here is that we aren't the ones binding the port - OBI/Beyla is - all we are doing is guessing a port that we can pass to down to Beyla and say "hey, bind your exporter to this port" - and then we reverse-proxy it.

In the unlikely event of a race, Beyla will fail, the process will restart and try again at a different port, so it should be pretty harmless - albeit not ideal.

I suppose a more proper would be to change OBI/Beyla to accept file descriptors instead - this way, we could do what you are proposing and bind/hold the port, then pass down the fd to Beyla - no race. I can work on that, but that'd be best handled as a separate/follow up PR IMHO. For now, I believe this should be acceptable (but feel free to disagree)."

(except for the last point, I can no longer work on that in a timely fashion, so we'll need to be pragmatic about this)

internal/component/beyla/ebpf/exec_linux.go
rafaelroquetto: That's right - the sockets are abstract-namespace and unauthenticated, so _in theory_ any process in in the same netns, but in practice I think that's not exploitable: the OTLP socket only ingests telemetry (worst case, a local process in the same netns injects fake metrics). Something like impersonation would require pre-binding a name that embeds our PID + a runtime counter and winning a startup race. I wouldn't worry about it.

internal/alloycli/resources_collector.go
rafaelroquetto: This is intentional and documented in the component docs (from #6477) the TLDR being that  Beyla's process metrics are exposed under the same names as Alloy's own (`alloy_resources_process_*`) with a `subprocess="beyla"` label, specifically so we don't fork a parallel metric namespace, allowing existing resource dashboards keep working and just filter/group by the label.

Prefixing them (`beyla_subproc_*`) was considered but dropped because it breaks backwards compatibility, something that was explicitly noted in the design doc. We even fixed some dashboards in dev as far as I remember to accomodate this (@skl may remember, it was a long time ago already).

WRT "do the numbers mean the same thing", these are standard OS process metrics (CPU seconds, RSS/VMS, start time) - the meaning is the same regardless of which collector reads them, and the label just scopes them to the child process. The help-string match is the cost of sharing the name (otherwise the collector goes belly up)

Having said that, your second suggestion is the clean way to drop the two-collector assumption, but it requires a lot of refactoring that I honestly think it should be done on a follow up, if push really comes to shove. At this stage, I am trying to take the very pragmatic road with this current PR (which is already a deviation from the battle tested #6477). All in all, this merits a discussion of its own IMHO.

internal/component/beyla/ebpf/beyla_linux.go
```diff
@@ -61,1466 +30,143 @@ func init() {
...
 func New(opts component.Options, args Arguments) (*Component, error) {
 	c := &Component{
-		opts:       opts,
-		args:       args,
-		argsUpdate: make(chan Arguments, 1),
+		opts:           opts,
+		args:           args,
+		argsUpdate:     make(chan Arguments, 1),
+		subprocessExit: make(chan error),
+		subprocess:     subprocess.New(),
```
ptodev, line 61: Should we start the subprocess in the Run() function? Usually New() functions shouldn't have side effects. They do, however, check for "catastrophic" issues that could prevent the component from starting in the first place.

docs/developer/updating-beyla.md
```diff
@@ -0,0 +1,50 @@
...
+
+   This writes `internal/component/beyla/ebpf/internal/config/gen/beyla-checksums.txt`
+   (committed, sha256sum format). `make download-beyla` verifies downloads against it,
+   so a compromised upstream can't swap in a binary we didn't review — the committed
+   checksum is the trust anchor, like `go.sum`.
+
+3. Download the binaries and schema, and sync the docs version:
+
+   ```
+   make beyla
```
ptodev:
```suggestion
   make update-beyla
```
Why not just run a "update-beyla" target that does the whole update, including updating the checksums? You could even state the Beyla version that you want to update to, in the command line argument?

docs/developer/updating-beyla.md
```diff
@@ -0,0 +1,50 @@
+# Updating Beyla
+
```
ptodev, line 2: We could put a note at the top to state that this action can only be done by maintainers, to prevent issues with supply chain attacks.

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,109 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
 You can configure the component to collect telemetry data from a specific port or executable path, and other criteria from Kubernetes metadata.
 The component exposes metrics that can be collected by a Prometheus scrape component, and traces that can be forwarded to an OTel exporter component.
 
-{{< admonition type="note" >}}
-To run this component, {{< param "PRODUCT_NAME" >}} requires administrative privileges, or at least it needs to be granted the following capabilities: `BPF`, `SYS_PTRACE`, `NET_RAW`, `CAP_CHECKPOINT_RESTORE`, `DAC_READ_SEARCH`, and `PERFMON`.
-The number of required capabilities depends on the specific use case.
-Refer to the [Beyla capabilities](https://grafana.com/docs/beyla/latest/security/#list-of-capabilities-required-by-beyla) for more information.
+## Permissions
```
ptodev, line 23: @clayton-cornell I don't know if you're ok with the placement of this title and whether you'd like to standardise permissions sections across reference docs? In this case I think it's ok to place it in the beginning. A lot of people might wonder about permissions in the context of eBPF and it'd be handy to state it first thing in the doc, before the reference tables.

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -20,15 +20,109 @@ The `beyla.ebpf` component is a wrapper for [Grafana Beyla][] which uses [eBPF][
...
+Treat this as a deliberate security decision.
 {{< /admonition >}}
 
+### Kubernetes: privileged
+
+Set `privileged: true` in the container's `securityContext`.
+This grants all capabilities and disables `seccomp` and AppArmor profiles.
+This approach is **not recommended** for production environments.
+
+### Kubernetes: unprivileged, root user
```
ptodev, line 65: The subprocess actually feels like a good opportunity to decrease the amount of capabilities Alloy has. I wonder if this would be doable? In theory you'd need to run a separate container for Beyla to achieve this and so I think the current subprocess solution wouldn't achieve this?

If customers ask for this, I suppose we'll just recommend them to run an Alloy instance just for Beyla. That way Alloy will have less priviledges.

A few months ago we'd considered running Beyla as a separate container in the Helm chart, but if I remember correctly we didn't do it since Alloy offers better Fleet Management integration.

Makefile
ptodev: No, mise is a different tool that's used for setting up the local dev environment. For example, it will make sure you use the same go lint binary so that every developer (and the CI server) are in sync.

You can try vibe coding a taskfile migration for your make steps on a different branch? Just point the new branch to this one and I'll review it seperately. The config looks very similar to github actions. You can use the existing examples to guide you.

internal/component/beyla/ebpf/internal/config/gen/.schema-version
ptodev, line 1: Why is this a `.` file, and why is it separate from the checksum one? Can we just have one file called something like "beyla_version.yaml", which contains all this information in a yaml format? It'd make it very easy to tell what needs to be updated.

internal/alloycli/resources_collector.go
ptodev: I'm ok with keeping them as a `subprocess="beyla"` label of the same metric family. But I'm worried about inconsistencies between the collectors. Even if they're similar now, they could diverge over time. For example, one collector might change its units while the other one doesn't.

We could open an issue to change the code to Alloy using its own collector to track subprocess metrics like RSS, CPU, network?

internal/component/beyla/ebpf/internal/config/gen/schema.json
ptodev, line 1: Can we put in downloaded files like this in a folder? E.g. something like `vendor`?

internal/component/beyla/ebpf/internal/config/config_build.go
ptodev: There are a couple of solutions to this:
* The config generator can be moved to the [tools cli](https://github.com/grafana/alloy/tree/main/tools), which has its own `go.mod`. The tools CLI is for Alloy build and release tooling, so this tool should probably go there anyway.
* The Beyla config could be exposed in a module in a Beyla repo that's separate from the other module. It could make it easy to import it?

It'd make it easier to correlate our configs with upstream's if we can import the struct? And the compiler will enforce extra type checks so that at least we're not exposing attributes that no longer exist.

ptodev (commented):
Thank you! I wrote a few comments and I'll get back to the PR again at the end of the day to see if I can add a few more. In general it looks ok and I think it could just use more organisation to make it more clear what is generated, what is manually updated, etc. I wrote a few suggestions but I don't yet have a concrete suggestion on how to reorganise it. I'll have a think at the end of the day. In the meantime, feel free to respond!

internal/component/beyla/ebpf/beyla_linux.go
rafaelroquetto: I reckon this is already the case - the subprocess is only started inside `Run()` by the call `c.handleSubprocessStart(ctx)` - `New()` here just sets it up, registers the process collector (the so called catastrophic check would fall in here) and publishes the initial scrape target. No subprocess is run at this stage.

internal/component/beyla/ebpf/internal/config/config_build.go
rafaelroquetto: So, just to disambiguate anything: there's no more config generator, this is all hand crafted code - the generator was part of the other PR. So there is nothing to move, this code is the core part of the beyla/ebpf component, just more organised to make it easier to read (rather than having everything inside beyla_linux.go)

With that out of the way...

> will enforce extra type checks so that at least we're not exposing attributes that no longer exist

that's already covered by the schema validation test and the coverage tests. So CI will fail in that case. Importing the Beyla module would add little to no value in this case (just more maintenance).

We floated the idea of exposing the Beyla config in a module a few weeks ago - but that won't cut it - we'd need to do OBI first (which may or may not be acceptable to the other maintainers) and that would reintroduce Beyla types/dependency hell (because the Beyla / OBI configs depend on types from packages like prometheus and otel - which defeats the purpose).

The current approach is entirely orthogonal - the only glue between Beyla and Alloy is a config file and a binary - I consider that to be a feature. Zero dependencies. Fully flexible. And IMHO, if we wanted to take that to the next level, I'd resurrect #6477 - which automates the generation from the schema. Anything that reimports dependencies defeat the purpose of the subprocess.

internal/alloycli/resources_collector.go
rafaelroquetto: https://github.com/grafana/alloy/issues/6721 (disclaimer: drafted with Claude for convenience)

ptodev: Thank you, looks good!

internal/component/beyla/ebpf/internal/config/config_build.go
ptodev: The problem is not just the config generator, but the schema itself. Everyone agrees that schema is nice and that we can use is for generating docs and code. But no one wants to have to write it and to maintain schema-related code. It's hard to understand, it's hard to read. I did a lot of work in this in #5539 and in various upstream PRs. It was in the pre-AI days and it took so much manual work just to write those schema files and to figure out how to expose Alloy-specific info in the parser, etc. Much of this can be automated now, but a person still has to read it and understand it. I'd like to avoid that if possible.

I think it's possible to avoid the schema. If the goal is just to test Alloy's config converter, then there could be a util in the [tools](https://github.com/grafana/alloy/tree/main/tools) directory which converts Alloy's config to a Beyla yaml config and double checks that it works. Then there'd be no schema and the PR will be easier to review, and the code will be easier to maintain. We can also reuse this approach for similar components in the future, like prometheus exporters.

rafaelroquetto: So I removed the schema completely and I vibe-coded (at this stage, I don't have time to tackle it properly) the validator tool as you proposed (`tools/beyla-config-validator`).

The trade-off is now that the go.mod/replaces appear again and need to be maintained, and it cannot enumerate any missing Beyla fields (because it cannot reflect on Alloy's config).

Makefile
```diff
@@ -257,7 +267,53 @@ test-pyroscope:
...
+
+$(BEYLA_SCHEMA_STAMP): $(BEYLA_SCHEMA)
+	@echo $(BEYLA_VERSION) > $@
+
+$(BEYLA_SCHEMA):
+	@mkdir -p $(BEYLA_SCHEMA_DIR)
+	curl -fsSL -o $@ \
+	  "https://raw.githubusercontent.com/grafana/beyla/$(BEYLA_VERSION)/docs/config-schema.json"
+
+alloy: generate-ui generate-source-code beyla
```
jharvey10, line 311: [🤖 Joebot automated comment] I’m wondering if we should mirror the existing update-vs-build split here. For example, `generate-snmp` does the “fetch/sync docs version” work as part of the explicit `generate` path, while `make alloy` only runs the generation needed to build. The existing Beyla bump workflow similarly treats `make alloy` as a verification step after the update work is done. Could we make this a consolidated `update-beyla`/`generate-beyla` style target for checksum/schema/docs updates, while keeping the `alloy` dependency closer to “ensure the pinned binary exists for embedding”?

internal/component/beyla/ebpf/internal/config/validation.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, feature := range args.Features {
+		if feature == "network" {
+			return true
+		}
+	}
+	return false
+}
+
+// hasAppFeature checks if any application feature is enabled in metrics
+func (args Metrics) hasAppFeature() bool {
```
jharvey10, line 59: [🤖 Joebot automated comment] This looks like a regression for wildcard feature semantics. The previous `hasAppFeature()` treated `"*"` and `"all"` as app features, and `TestMetrics_Validate_FeatureWildcard` covered that. In this PR those values are still accepted by `Metrics.Validate()`, but they no longer trigger the app discovery validation. Should we preserve the old wildcard behavior here? Related question: should `hasNetworkFeature()` also treat `"*"`/`"all"` (and maybe the network-specific feature names) as network-enabled, since the YAML path only emits `network.enable = true` through this helper?

internal/component/beyla/ebpf/internal/config/validation.go
```diff
@@ -0,0 +1,223 @@
...
+		hasKubernetes := svc.Kubernetes.Namespace != "" ||
+			svc.Kubernetes.PodName != "" ||
+			svc.Kubernetes.DeploymentName != "" ||
+			svc.Kubernetes.ReplicaSetName != "" ||
+			svc.Kubernetes.StatefulSetName != "" ||
+			svc.Kubernetes.DaemonSetName != "" ||
+			svc.Kubernetes.OwnerName != "" ||
+			len(svc.Kubernetes.PodLabels) > 0
+
+		if svc.OpenPorts == "" && svc.Path == "" && !hasKubernetes {
```
jharvey10: [🤖 Joebot automated comment] Looks like `cmd_args` fell out of the “has at least one selector” check. The previous validation allowed cmd-args-only service selectors, and there was a test covering `discovery.instrument { cmd_args = "--serve*" }` with application metrics enabled. This PR still exposes `CmdArgs` and still emits `cmd_args` in `Services.Convert()`, so I think validation is now rejecting a config shape the converter otherwise supports. Could we add `svc.CmdArgs == ""` back into this check and keep/add a focused test for cmd-args-only selectors?

tools/beyla-config-validator/main.go
ptodev, line 1: Thanks for uploading this! It's close to what I expected, but not quite there. I was hoping for this utility to import both beyla and the converter code from `internal/component/beyla/ebpf/internal/config/config.go`. Then this utility can validate that every Alloy block and attribute is a valid beyla config parameter. The simplest way would be to have a single config file in Alloy syntax and to translate it to a beyla yaml config file. It'd look similar to the [otel converters](https://github.com/grafana/alloy/tree/main/internal/converter/internal/otelcolconvert/testdata). If you want to make it smarter, I suppose you could also use reflection to find out if Alloy is missing any parameters.

grcevski: I think it's a good idea, but it can be done in a follow-up PR.

rafaelroquetto: The problem is that I don't think this is possible in this case - because `tools/beyla-config-validator/main.go` lives in its own go mod, orthogonal and oblivious to everything, it is not able to import `internal/component/beyla/ebpf/internal/config/config.go`. The otel converters work because they are part of the main Alloy go.mod - doing that would bring us back to square 0, defeating the subprocess purpose.

I am going to be very frank: after looking at this again, I realised I really dislike this approach. I find having to maintain a separate `beyla-config-validator`, including its `go.mod` and dependency tree highly unpractical: now one needs to reason about:
 - the Beyla code, manually
 - this validator, including the dependency tree, which I was _very happy to see gone_.

I feel the schema test that we had was a good middle ground because the schema is a more universal language: any JSON parser can consume it, there's no implied reasoning about go dependencies, no management, it was all automated. Design-wise, it was just an input file that came from the Beyla release alongside with the binaries and checksums, nothing deep. Input `(beyla_bin, checksum, schema) -> beyla.ebpf`. Very straightforward. 

The handling code was convoluted, I agree with that - but to me, the path forward would be trying to make it simpler and more maintainable, to the extent possible, not getting rid of it altogether in favour of this. I will go as far as I say that the code that walked the JSON schema to validate ~60 lines, so perhaps not as complicated as it appears - and looking into it, we could make it even easier by using ` github.com/xeipuuv/gojsonschema`. It would look like this:

```
func TestEmittedConfigMatchesSchema(t *testing.T) {
      var args Arguments
      fillValue(reflect.ValueOf(&args).Elem(), 0)
      cfg := buildYAML(t, args, Runtime{Port: 12345})

      schemaBytes, err := os.ReadFile("gen/beyla/schema.json")
      require.NoError(t, err)

      result, err := gojsonschema.Validate(
              gojsonschema.NewBytesLoader(schemaBytes),
              gojsonschema.NewGoLoader(cfg),
      )
      require.NoError(t, err)

      for _, e := range result.Errors() {
              t.Errorf("%s: %s", e.Field(), e.Description())
      }
}
```

plus  a few dozen of other extra lines for completeness.

I will push a commit on top of this one for you to compare both approaches. The file in question: https://github.com/grafana/alloy/blob/beyla_hackathon_simple/internal/component/beyla/ebpf/internal/config/schema_validation_test.go

This test does exactly what this tool aims to do. What neither do is validating if Alloy is missing a newly added option to Beyla. That would entail resurrecting `coverage_test.go`, which is the more complicated bit. 

Anyhow, the test as it is seems to be a good compromise. Let me know your thoughts.

docs/sources/reference/components/beyla/beyla.ebpf.md
rafaelroquetto: TLDR: Alloy does not run with these capabilities on its effective set. 

Alloy now _never_ runs with those capabilities - it only has the ability to acquire them, but only the Beyla subprocess has the capabilities on the effective set. So unless Alloy elevates it capabilities on purpose, it's effectively running without them.

It does need to have its capabilities present somehow on its ambient set, so that the subprocess can elevate to these capabilities and run. Alternatively, the binary can have the capabilities set as described in the next session:Kubernetes: unprivileged, non-root user."

It is impossible to start with 0 capabilities in the inheritable or the ambient set and then "elevate" to these capabilities.

Makefile
rafaelroquetto: So, I did have a look but it need to duplicate a lot of the Makefile. It doesn't look good - I think this should be done on a follow up PR, I won't have time this week to look into this as part of this PR.

internal/component/beyla/ebpf/internal/config/gen/.schema-version
rafaelroquetto: It's now in `beyla_version.yaml`

internal/component/beyla/ebpf/internal/config/gen/schema.json
rafaelroquetto: they are in the `binaries` folder now - `vendor` cannot work here

docs/developer/updating-beyla.md
rafaelroquetto: done, `make update-beyla TAG=<ver>` now records checksums, downloads+verifies the binary, fetches the schema, and syncs the docs version, all in one

docs/developer/updating-beyla.md
rafaelroquetto: done

Makefile
rafaelroquetto: fixed

internal/component/beyla/ebpf/internal/config/validation.go
rafaelroquetto: fixed

internal/component/beyla/ebpf/internal/config/validation.go
rafaelroquetto: fixed

rafaelroquetto:
@ptodev @jharvey10 I've addressed most of the comments/deferred one or two of them to follow up PRs.

I've also rebased. Hopefully this is satisfactory now.

Makefile
ptodev: Ok, no problem, it's not very important.

docs/developer/updating-beyla.md
```diff
@@ -0,0 +1,47 @@
...
+> **Maintainers only.** Bumping Beyla records new binary checksums in
+> `beyla_version.yaml` — the trust anchor for verifying the download. Run
+> `make update-beyla` and review the checksum diff only as a maintainer, to guard
+> against supply-chain tampering.
```
ptodev, line 6:
```suggestion
> **Only maintainers should update Beyla** 
> Bumping Beyla records new binary checksums in `beyla_version.yaml`.
> To guard against supply-chain attacks, only maintainers should change the Beyla binaries that Alloy incorporates.
> If a community PR is opened for this, maintainers should close it and reopen a new one made by a maintainer.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
ptodev: > So unless Alloy elevates it capabilities on purpose, it's effectively running without them.

Yes, what I mean is that it'd great if Alloy can't even elevate its own capabilities. Then it'd be entirely sandboxed.  It's obviously not something we should do for this PR, and I don't know if it's even possible, but I thought it's worth mentioning.

internal/component/beyla/ebpf/internal/config/config_build.go
ptodev: Will close this since we're discussing it on the [other thread](https://github.com/grafana/alloy/pull/6695#discussion_r3617967114).

internal/component/beyla/ebpf/internal/config/config_test.go
```diff
@@ -0,0 +1,233 @@
...
+					CacheLen: 256,
+				},
+			},
+		},
+	}
+
+	config := buildYAML(t, args, Runtime{Port: 12345})
+
+	// Verify network
+	networkFlows, ok := config["network"].(map[string]any)
```
ptodev, line 178: It'd be more readable to use `require.YAMLEq` instead?

docs/developer/updating-beyla.md
```diff
@@ -0,0 +1,47 @@
...
+Review the `beyla_version.yaml` diff before committing — the committed checksum is the
+trust anchor, like `go.sum`.
+
+## Updating the config translation
+
+Reconcile the hand-written types and translation with any new or changed Beyla
+options. Two checks guard this:
+
+- **`golden_test.go`** locks the full emitted YAML byte-for-byte (regenerate with
+  `UPDATE_GOLDEN=1` when a translation change is intended).
```
ptodev, line 42: Feels like the update of the golden config should also be part of the above "make update-beyla" command? Otherwise we could foget to update it?

rafaelroquetto:
For the record, this version is now running on the dev clusters without any issues.

tools/beyla-config-validator/main.go
ptodev: > The problem is that I don't think this is possible in this case - because tools/beyla-config-validator/main.go lives in its own go mod, orthogonal and oblivious to everything, it is not able to import internal/component/beyla/ebpf/internal/config/config.go.

It could work with a replace directive [like this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b808b2e0574106bb965f726f8fafbc4f09bc41a6/receiver/kafkareceiver/go.mod#L128).

> we could make it even easier by using  github.com/xeipuuv/gojsonschema

Yes, this is much better! Now the schema is acceptable because the Alloy team neither has to maintain a schema file nor to maintain schema-related code. And I agree that it's less overhead than the converter I proposed above. We can keep the schema and to use it for further automations for docs, etc.

ptodev (approved):
I added a few super minor comments, but we can resolve them in a subsequent PR.

Thanks a lot for working on this!

Issue #6570, opened 2026-06-22
Improve performance of Prometheus metadata pipelines


ptodev:
### Component(s)

prometheus.scrape, prometheus.remote_write

### Add a description

I observed a roughly 10% increase in memory and network consumption when I set `honor_metadata` to `true` in `prometheus.scrape` in a few clusters. We need to look into reducing the performance impact of metadata. Otherwise we can not enable it by default in Alloy. It may even be cost prohibitive to enable it in big clusters.

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

ptodev:
This issue is closely related to an [upstream issue ](https://github.com/prometheus/prometheus/issues/17619)for performance improvements. I also opened [a PR](https://github.com/prometheus/prometheus/pull/19015) with one potential upstream improvement to memory consumption.

ptodev:
The most up to date information I have on this is that it's discussed upstream on a [proposal](https://docs.google.com/document/d/1PCDrUSmxOxR8u_CHde5x9r0Nc07c4bUhe1TLjIkDhfs/edit?tab=t.0#heading=h.953vjrm0r8pj) and on a [recurring call](https://docs.google.com/document/d/1ZcyKVF87Xf6FhaGkb33LgVLb27FH5NIlRXWEGO2Cc4Q/edit?tab=t.0#heading=h.1w7ckakwf0kf), but the current conversation is about optimising overhead of processing type and unit metadata (not help strings).

kgeckhart:
> the current conversation is about optimising overhead of processing type and unit metadata (not help strings).

Just a small clarification, the current conversations revolve around agreeing on type and unit labels as the path forward for those and needing PoCs to better understand our options for help text solutions.

PR #6698, opened 2026-07-15
fix(prometheus): Reduce metric pipeline allocation overhead

BominRahmani:
Reduces avoidable heap allocations when Alloy forwards Prometheus metadata while live debugging is inactive.

### Pull Request Details

Prometheus metadata flows through the scrape and remote-write interceptors. Previously each update created a `livedebugging.Data` value before passing it to PublishIfActive.
The debug data contains a function that formats the entry only when its needed. Although formatting is deferred, creating that function still captures the labels and metadata in a closure, causing them to escape to the heap.

Since Go evaluates function arguments before making the call, this allocation occurred before PublishIfActive could determine that nobody was listening, including when even when live debugging was inactive:

Related to #6570

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


kgeckhart:
👋  thanks for the PR I would love to see some benchmarks on this before we proceed. Could you please add some and share the benchstat comparison results, thanks!

BominRahmani:
> 👋 thanks for the PR I would love to see some benchmarks on this before we proceed. Could you please add some and share the benchstat comparison results, thanks!

Hey @kgeckhart thanks for looking at the PR!

I wrote two small benchmarks to measure the heap allocation caused by constructing the deferred-rendering closure while live debugging is inactive. I ran the same benchmark code against the PR merge base and this branch using `-benchmem -benchtime=2s -count=10`
Here is the scrape interceptor benchmark; the remote-write benchmark uses the same setup around its interceptor:

```
func BenchmarkInterceptorUpdateMetadataInactive(b *testing.B) {
	interceptor := NewInterceptor(
		"prometheus.scrape.test",
		livedebugging.NewLiveDebugging(),
		appenders.Noop{},
	)
	appender := interceptor.Appender(b.Context())
	metricLabels := labels.FromStrings("__name__", "test_metric")
	metricMetadata := metadata.Metadata{
		Type: "gauge",
		Unit: "items",
		Help: "test metric",
	}

	b.ReportAllocs()
	b.ResetTimer()

	for b.Loop() {
		if _, err := appender.UpdateMetadata(0, metricLabels, metricMetadata); err != nil {
			b.Fatal(err)
		}
	}
}
```
Here is the benchstat comparison
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/alloy/internal/component/prometheus/remotewrite
cpu: Apple M2 Pro
                                     │     base     │                 pr                  │
                                     │    sec/op    │   sec/op     vs base                │
InterceptorUpdateMetadataInactive-12   100.85n ± 3%   63.97n ± 1%  -36.57% (p=0.000 n=10)

                                     │    base    │                 pr                 │
                                     │    B/op    │   B/op     vs base                 │
InterceptorUpdateMetadataInactive-12   80.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                                     │    base    │                 pr                  │
                                     │ allocs/op  │ allocs/op   vs base                 │
InterceptorUpdateMetadataInactive-12   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)

pkg: github.com/grafana/alloy/internal/component/prometheus/scrape
                                     │    base     │                 pr                  │
                                     │   sec/op    │   sec/op     vs base                │
InterceptorUpdateMetadataInactive-12   46.52n ± 6%   18.23n ± 1%  -60.80% (p=0.000 n=10)

                                     │    base    │                 pr                 │
                                     │    B/op    │   B/op     vs base                 │
InterceptorUpdateMetadataInactive-12   80.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                                     │    base    │                 pr                  │
                                     │ allocs/op  │ allocs/op   vs base                 │
InterceptorUpdateMetadataInactive-12   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```
I wasn’t sure whether you’d prefer the benchmark files to be committed or whether including the benchmark and results here is sufficient. I’m happy to add them to the PR if they would be useful to keep for future comparisons.

BominRahmani:
@kgeckhart Hey just wanted to check in to see if this was still on the radar for review?

kgeckhart:
👋  I'm catching up after being off for awhile. Thanks for the benchmarks I think it's fine to keep them off the PR I was just interested because I couldn't remember seeing the interceptor pop up on profiles internally before I went on vacation. Now that I'm back I was able to verify it is indeed a rather high allocator itself. 

 Could you apply this same pattern to the Append, Histogram, and Exemplar hooks as well? They are much more active than metadata is today so it's a much more impactful to limit them. 

FTR this won't resolve the metadata overhead referenced in the issue (I changed your resolve to references). The same overhead referenced there has been observed in prometheus itself where they do not have this.

BominRahmani:
> 👋 I'm catching up after being off for awhile. Thanks for the benchmarks I think it's fine to keep them off the PR I was just interested because I couldn't remember seeing the interceptor pop up on profiles internally before I went on vacation. Now that I'm back I was able to verify it is indeed a rather high allocator itself.
> 
> Could you apply this same pattern to the Append, Histogram, and Exemplar hooks as well? They are much more active than metadata is today so it's a much more impactful to limit them.
> 
> FTR this won't resolve the metadata overhead referenced in the issue (I changed your resolve to references). The same overhead referenced there has been observed in prometheus itself where they do not have this.

No worries! I've updated the PR and fixed my commits so everything should be signed and good to go. Let me know if it needs anything else!

kgeckhart:
Thank you! It looks like CI is unhappy because the test interface needs to implement appenderV2,

```
internal/component/prometheus/remotewrite/interceptor_test.go:65:7: cannot use interceptorTestStorage{…} (value of struct type interceptorTestStorage) as "github.com/prometheus/prometheus/storage".Storage value in argument to NewInterceptor: interceptorTestStorage does not implement "github.com/prometheus/prometheus/storage".Storage (missing method AppenderV2)
internal/component/prometheus/remotewrite/interceptor_test.go:113:5: cannot use interceptorTestStorage{…} (value of struct type interceptorTestStorage) as "github.com/prometheus/prometheus/storage".Storage value in argument to NewInterceptor: interceptorTestStorage does not implement "github.com/prometheus/prometheus/storage".Storage (missing method AppenderV2) (typecheck)
```

kgeckhart (approved):
Thanks again for the PR on this. I'm interested to see the impact it has!

PR #6709, opened 2026-07-16
docs(loki.secretfilter): Document custom Gitleaks configuration and false positives

kleimkuhler:

Adds documentation for `loki.secretfilter`'s custom Gitleaks configuration: a section on extending the default ruleset with `[extend]` / `useDefault = true`, and a section on handling false positives by allowlisting a built-in rule (by matched text and by secret value). Also fixes a stale link to the default ruleset.

### PR Checklist

- [x] Documentation added


docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,6 +135,98 @@ The following fields are exported and can be referenced by other components:
...
+Point the component at this file:
+
+```alloy
+loki.secretfilter "secret_filter" {
+    forward_to      = [loki.write.local_loki.receiver]
+    gitleaks_config = "/etc/alloy/gitleaks.toml"
+}
+```
+
+When `useDefault = true`, redefining a `[[rules]]` block with the same `id` as a built-in rule merges your changes into that rule instead of replacing it.
```
kelnage, line 172: Perhaps we can also mention that rules can be disabled in this way too? Potentially with an example for disabling `generic-api-key`?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,6 +135,98 @@ The following fields are exported and can be referenced by other components:
...
+
+[[rules]]
+id = "secret-internal-token"
+description = "Secret internal service token"
+regex = '''secret_tok_[0-9a-zA-Z]{32}'''
+keywords = ["secret_tok_"]
+
+# Reduce false positives from the built-in generic-api-key rule!
+[[rules]]
+id = "generic-api-key"
```
kelnage, line 223: Add a line to show how to change the entropy (as suggested above)?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,6 +135,98 @@ The following fields are exported and can be referenced by other components:
...
+  [[rules.allowlists]]
+  description = "Ignore UUID values"
+  regexTarget = "secret"
+  regexes = [
+    '''^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$''',
+  ]
+```
+
+With this configuration, `generic-api-key` keeps detecting real secrets, but no longer redacts log lines such as `token_id=a1B2c3D4e5F6g7H8i9J0` or values shaped like the UUID `8f14e45f-ceea-167a-9c2b-1f0a3e4d5c6b`.
+
```
kelnage, line 245: A section discussing performance and using the `rate`/`processing_timeout` settings would be worthwhile

docs/sources/reference/components/loki/loki.secretfilter.md
kelnage: Isn't this obvious? It drops the log line, so what would it label?

mikefat (commented):
LGTM, I'd probably add a link if we don't already to the official gitleaks documentation at the end of the section

kelnage (commented):
Great addition, a couple of extra suggestions for examples or clarity

kleimkuhler:
@mikefat I've changed where we reference Gitleaks documentation in https://github.com/grafana/alloy/pull/6709/commits/eac74fa2f6cc0509fcac87da6e975753c6bcfe2f. I replaced where we point to the default ruleset (which we already did at the top of the section) with actually pointing to the configuration documentation as part of the "Use a custom Gitleaks configuration" section.

kleimkuhler:
@kelnage I added a "Manage Performance" section to talk about `rate` and `processing_duration`.

I also merged "Usage" and "Example" sections as they kind of overlapped, but "Usage" really was not that helpful. I kept it at the top, but it's more detailed now.

clayton-cornell:
Just a note... there's a PR in progress to fix the noisy Vale linting bot so it only looks at lines that were changed in the PRs... but it's not yet finalized/merged.

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: Mentioned below now

```
To turn off a built-in rule entirely rather than adjust it, list its `id` in `disabledRules`.
Disabling a rule stops it from detecting anything, so to keep a rule but ignore specific values, use an allowlist instead.
```

I didn't include an actual TOML example since it's in the Gitleaks configuration documentation linked directly below now!

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: Added in "Manage performance"

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,45 +168,133 @@ The following fields are exported and can be referenced by other components:
...
+Monitor `loki_secretfilter_entries_bypassed_total` to see how many entries were skipped.
+
+**Bound per-line work with `processing_timeout`.**
+Large lines can be expensive to scan, so set `processing_timeout` to cap how long the component spends on any single entry.
+By default, a timed-out line is forwarded unredacted so no data is lost, or you can set `drop_on_timeout = true` to drop it instead.
+Monitor `loki_secretfilter_lines_timed_out_total` and `loki_secretfilter_lines_dropped_total`, and use `loki_secretfilter_processing_duration_seconds` to track overall processing time.
+
+**Detect fewer secrets.**
+Every rule adds regular expression work, so removing rules you don't need, or raising a noisy rule's `entropy` threshold, also lowers CPU usage.
+Refer to [Use a custom Gitleaks configuration](#use-a-custom-gitleaks-configuration) to disable or tune rules.
```
mikefat: Maybe we can add something like this
```
[extend]
useDefault = true
disabledRules = ["generic-api-key"]
...
```

kleimkuhler: I mention `disabledRules` [here](https://github.com/grafana/alloy/pull/6709/changes#diff-cb42a8225fc40261b4b72273b2a59706fba149ef5baf5e8f026d064e35261bb9R205-R206)! If you still think it's worth including I can, but trying not to replicate too much of Gitleaks configuration again in here.

mikefat: I would still include a specific example with removing generic-api-key. 
The rest looks good to me

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -34,14 +34,48 @@ Place `loki.secretfilter` after components that reduce log volume so it processe
 
 [gitleaks-config]: https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml
 
-## Usage
+## Example usage
```
clayton-cornell: This breaks the format and layout of the component documentation. The Usage section is only for the high level syntax/format of the component config (analogous to the Synopsis section of a man page). Working examples are either lower down in the component topic, or alternatively we can:

* write a new task topic that deep dives into the configuration possibilities and discusses the how, why, etc.
* develop a new scenario and add it to the alloy-scenarios

Minimum here is to restore the Usage section and move the example down to where examples are placed in Component topics.

There is a WIP that is early stages to remove comprehensive examples out of the component docs and shift the configuration details/discussions and in-depth examples to a task based topic elsewhere in the Alloy docs.

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: https://github.com/grafana/alloy/pull/6709/commits/20b9abebfcab64497231cd8871087aa0e04edaf8

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: Ah okay thanks for the heads up. I can just restore the previous "Usage" section and keep these changes in the existing "Example" section that was at the bottom!

clayton-cornell: That works. Ping me when it's ready for doc review :-)

kleimkuhler: @clayton-cornell https://github.com/grafana/alloy/pull/6709/commits/8dde505cbcf772c9c6ad5752a0b8eb0a79bff6cf restores the original sections layout!

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
 | `loki_secretfilter_secrets_redacted_total`         | Counter | Total number of secrets redacted.                                                              |
 | `loki_secretfilter_secrets_redacted_by_category_total` | Counter | Number of secrets redacted, partitioned by rule name and origin label value. The `origin` label is empty when `origin_label` is not set or the label is absent on the entry. |
 
+## Use a custom Gitleaks configuration
+
+By default, the component detects secrets using the upstream [Gitleaks default ruleset][gitleaks-config], which is compiled into {{< param "PRODUCT_NAME" >}}.
```
```suggestion
By default, the component detects secrets with the upstream [Gitleaks default ruleset][gitleaks-config].
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+{{< /admonition >}}
+
+## Handle false positives
+
+Because detection is regular expression-based, `loki.secretfilter` sometimes redacts values that aren't secrets, such as identifiers, UUIDs, or hashes that happen to look like an API key.
+When this happens, identify which rule is firing, then add an _allowlist_ to that rule so the matching values are ignored.
+
+To find the responsible rule, inspect the `rule` label on the `loki_secretfilter_secrets_redacted_by_category_total` metric.
+
+For example, the built-in `generic-api-key` rule is the most common source of false positives because it matches high-entropy strings broadly.
```
```suggestion
For example, the built-in `generic-api-key` rule can produce false positives because it matches high-entropy strings broadly.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
 | `loki_secretfilter_secrets_redacted_total`         | Counter | Total number of secrets redacted.                                                              |
 | `loki_secretfilter_secrets_redacted_by_category_total` | Counter | Number of secrets redacted, partitioned by rule name and origin label value. The `origin` label is empty when `origin_label` is not set or the label is absent on the entry. |
 
+## Use a custom Gitleaks configuration
+
+By default, the component detects secrets using the upstream [Gitleaks default ruleset][gitleaks-config], which is compiled into {{< param "PRODUCT_NAME" >}}.
+To change which secrets are detected, set the `gitleaks_config` argument to the path of a custom Gitleaks TOML file.
```
```suggestion
To change which secrets the component detects, set the `gitleaks_config` argument to the path of a custom Gitleaks TOML file.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+}
+```
+
+When `useDefault = true`, redefining a `[[rules]]` block with the same `id` as a built-in rule merges your changes into that rule instead of replacing it.
+This is what makes it possible to adjust a built-in rule, for example, to add allowlists to it, without losing its detection logic.
+
+To turn off a built-in rule entirely rather than adjust it, list its `id` in `disabledRules`.
+Disabling a rule stops it from detecting anything, so to keep a rule but ignore specific values, use an allowlist instead.
+
+Adjusting a built-in rule with allowlists is the most common way to reduce false positives, as the [following section](#handle-false-positives) demonstrates.
```
```suggestion
Adjust built-in rules with allowlists to reduce false positives, as shown in [Handle false positives](#handle-false-positives).
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+
+## Handle false positives
+
+Because detection is regular expression-based, `loki.secretfilter` sometimes redacts values that aren't secrets, such as identifiers, UUIDs, or hashes that happen to look like an API key.
+When this happens, identify which rule is firing, then add an _allowlist_ to that rule so the matching values are ignored.
+
+To find the responsible rule, inspect the `rule` label on the `loki_secretfilter_secrets_redacted_by_category_total` metric.
+
+For example, the built-in `generic-api-key` rule is the most common source of false positives because it matches high-entropy strings broadly.
+You may benefit from just raising the default entropy value.
```
```suggestion
Start by raising the entropy value.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+
+For more precise control, keep the rule enabled and allowlist only the specific values that cause false positives.
+
+Allowlist the false positives by extending the same configuration from the previous section.
```
```suggestion
Allowlist false positives by extending the same configuration from the previous section.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+  # Allowlist by secret value: ignore when the captured value is a UUID.
+  [[rules.allowlists]]
+  description = "Ignore UUID values"
+  regexTarget = "secret"
+  regexes = [
+    '''^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$''',
+  ]
+```
+
+With this configuration, `generic-api-key` keeps detecting real secrets, but no longer redacts log lines such as `token_id=a1B2c3D4e5F6g7H8i9J0` or values shaped like the UUID `8f14e45f-ceea-167a-9c2b-1f0a3e4d5c6b`.
```
```suggestion
With this configuration, `generic-api-key` still detects real secrets, but it no longer redacts log lines such as `token_id=a1B2c3D4e5F6g7H8i9J0` or values shaped like the UUID `8f14e45f-ceea-167a-9c2b-1f0a3e4d5c6b`.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+
+## Manage performance
+
+Secret detection runs a set of regular expressions against every log line, so its cost scales with both your log volume and the number of active rules.
+The following options help you control that cost.
```
```suggestion
Secret detection runs regular expressions against every log line.
The cost rises with your log volume and the number of active rules.
```

Is it CPU cost?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+Place `loki.secretfilter` after components that filter or drop logs, such as `loki.process`, so it only scans the lines you care about.
+Fewer lines in means less work for the component.
+
+**Sample high-volume streams with `rate`.**
+Set `rate` to a value below `1.0` to process only a fraction of entries, for example, `0.1` to scan 10% of lines.
+Entries that {{< param "PRODUCT_NAME" >}} doesn't select are forwarded unchanged, so this reduces CPU usage at the cost of leaving some lines unscanned.
+Monitor `loki_secretfilter_entries_bypassed_total` to see how many entries were skipped.
+
+**Bound per-line work with `processing_timeout`.**
+Large lines can be expensive to scan, so set `processing_timeout` to cap how long the component spends on any single entry.
```
```suggestion
Large lines take longer to scan, so set `processing_timeout` to cap how long the component spends on any single entry.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+By default, a timed-out line is forwarded unredacted so no data is lost, or you can set `drop_on_timeout = true` to drop it instead.
+Monitor `loki_secretfilter_lines_timed_out_total` and `loki_secretfilter_lines_dropped_total`, and use `loki_secretfilter_processing_duration_seconds` to track overall processing time.
```
clayton-cornell: Is this true? Line 352 in internal/component/loki/secretfilter/secretfilter.go says // Redact any partial findings before forwarding, even if the timeout was hit.

I asked Claude to analyse this part of the source code and it suggested this:

```suggestion
When a timeout occurs, `loki_secretfilter_lines_timed_out_total` increments.
If `drop_on_timeout = false`, {{< param "PRODUCT_NAME" >}} forwards the entry.
If the detector returns findings before timeout, {{< param "PRODUCT_NAME" >}} redacts those findings before it forwards the entry.
If `drop_on_timeout = true`, {{< param "PRODUCT_NAME" >}} drops the entry and increments `loki_secretfilter_lines_dropped_total`.
Use `loki_secretfilter_processing_duration_seconds` to track overall processing time.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+[[rules]]
+id = "secret-internal-token"
+description = "Secret internal service token"
+regex = '''secret_tok_[0-9a-zA-Z]{32}'''
+keywords = ["secret_tok_"]
+
+# Reduce false positives from the built-in generic-api-key rule.
+[[rules]]
+id = "generic-api-key"
+# Raise the generic-api-key entropy threshold above its default of 3.5.
```
clayton-cornell: Should we pin to a specific default here or be more generic... maybe:


```suggestion
# Raise the generic-api-key entropy threshold above its default.
```

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+
+```alloy
+loki.secretfilter "secret_filter" {
+    forward_to      = [loki.write.local_loki.receiver]
+    gitleaks_config = "/etc/alloy/gitleaks.toml"
+}
+```
+
+When `useDefault = true`, redefining a `[[rules]]` block with the same `id` as a built-in rule merges your changes into that rule instead of replacing it.
+This is what makes it possible to adjust a built-in rule, for example, to add allowlists to it, without losing its detection logic.
```
```suggestion
This lets you add allowlists or tune fields without losing the detection logic for that rule.
```

clayton-cornell:
I've opened: https://github.com/grafana/writers-toolkit/pull/1475 to add Gitleaks to the Vale dictionary.

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: Yea it's the compute cost of introducing a component that will lead to more processing time in the Alloy pipeline. I think I'd like to keep these 2 lines as is if that's okay? I think the second line stating clearly that the following options help control cost is helpful at least.

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: True yea good catch. It does redact what it's found, but will forward the rest (that it did not finish scanning) unredacted. What do you think of keeping it closer to the original wording though?

```
By default, a timed-out line is still forwarded, with any secrets found before the deadline already redacted, so no data is lost.

Set `drop_on_timeout = true` to drop it instead. Monitor `loki_secretfilter_lines_timed_out_total` and `loki_secretfilter_lines_dropped_total`, and use `loki_secretfilter_processing_duration_seconds` to track overall processing time.
```

kleimkuhler: I made the change in https://github.com/grafana/alloy/pull/6709/commits/9a369f3fd30b393334673bb3cf8955e75c9d4ca6 anyways, let me know your thoughts!

docs/sources/reference/components/loki/loki.secretfilter.md
clayton-cornell:
```suggestion
Secret detection runs regular expressions against every log line.
Each active rule adds scan work, and CPU cost rises with log volume and rule count.
The following options help you control that cost.
```

Trying to keep it all in active voice... what about this?

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,156 @@ The following fields are exported and can be referenced by other components:
...
+Fewer lines in means less work for the component.
+
+**Sample high-volume streams with `rate`.**
+Set `rate` to a value below `1.0` to process only a fraction of entries, for example, `0.1` to scan 10% of lines.
+Entries that {{< param "PRODUCT_NAME" >}} doesn't select are forwarded unchanged, so this reduces CPU usage at the cost of leaving some lines unscanned.
+Monitor `loki_secretfilter_entries_bypassed_total` to see how many entries were skipped.
+
+**Bound per-line work with `processing_timeout`.**
+Large lines can take longer to scan, so set `processing_timeout` to cap how long the component spends on any single entry.
+By default, a timed-out line is still forwarded, with any secrets found before the deadline already redacted, so no data is lost. Set `drop_on_timeout = true` to drop it instead.
```
```suggestion
By default, a timed out line is forwarded, and {{< param "PRODUCT_NAME" >}} redacts any findings returned before the timeout, so the line is not dropped.
Set `drop_on_timeout` to `true` to drop timed out lines instead.
```

A small tweak to tighten it up.

docs/sources/reference/components/loki/loki.secretfilter.md
clayton-cornell: No prob. I've added a small tweak to the change to tighten it up.

clayton-cornell (dismissed):

clayton-cornell:
Do we need any more approves before merging?

docs/sources/reference/components/loki/loki.secretfilter.md
kelnage: I know you didn't change this line in this PR, but I think it suffers from the same [issue identified here](https://github.com/grafana/alloy/pull/6709#discussion_r3631741862) - timeouts don't necessarily mean the original unredacted entry is forwarded.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,158 @@ The following fields are exported and can be referenced by other components:
...
+Monitor `loki_secretfilter_entries_bypassed_total` to see how many entries were skipped.
+
+**Bound per-line work with `processing_timeout`.**
+Large lines can take longer to scan, so set `processing_timeout` to cap how long the component spends on any single entry.
+By default, a timed out line is forwarded, and {{< param "PRODUCT_NAME" >}} redacts any findings returned before the timeout, so the line is not dropped.
+Set `drop_on_timeout` to `true` to drop timed out lines instead.
+Monitor `loki_secretfilter_lines_timed_out_total` and `loki_secretfilter_lines_dropped_total`, and use `loki_secretfilter_processing_duration_seconds` to track overall processing time.
+
+**Detect fewer secrets.**
+Every rule adds regular expression work, so removing rules you don't need, or raising a noisy rule's `entropy` threshold, also lowers CPU usage.
```
kelnage:
```
or raising a noisy rule's `entropy` threshold
```

This won't reduce the CPU usage significantly, given the regular expression will still need to run _before_ calculating the entropy.

kelnage (commented):
Two minor pedantic issues that I thought should be flagged.

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: Yep good catch I removed mentioning entropy in this section. It's still mentioned in the False Positives section.

docs/sources/reference/components/loki/loki.secretfilter.md
```diff
@@ -135,17 +134,158 @@ The following fields are exported and can be referenced by other components:
...
+  regexes = [
+    '''^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$''',
+  ]
+```
+
+With this configuration, `generic-api-key` still detects real secrets, but it no longer redacts log lines such as `token_id=a1B2c3D4e5F6g7H8i9J0` or values shaped like the UUID `8f14e45f-ceea-167a-9c2b-1f0a3e4d5c6b`.
+
+## Manage performance
+
+Secret detection runs regular expressions against every log line.
```
kelnage: By using the Gitleaks implementation, the secretfilter now uses a trie (the prefilter) to [scan log lines with fixed string matches](https://github.com/gitleaks/gitleaks/blob/master/detect/detect.go#L322), and only uses regular expressions if there's at least one match for the corresponding keyword (or a rule that lacks any keywords). The only rule without keywords in the current default Gitleaks config is `pkcs12-file` - and that only matches on the path (which secretfilter doesn't currently use).

This detail seems important to me,  because if users want to introduce their own rules, if they don't add keywords, they'll get worse performance than the default rules provide.

kelnage (commented):
Apologies, but I had one more pedantic point - secretfilter won't necessarily run regular expressions against every line (the use of a trie was a significant factor in the performance improvements). But I don't know how to phrase that simply 😄

docs/sources/reference/components/loki/loki.secretfilter.md
kleimkuhler: No worries yes good clarification! I split up the explanation into the sections where it felt important to mention this... let me know if that still looks good to you.

kelnage (approved):
LGTM!

PR #6716, opened 2026-07-20
fix(loki.source.docker): Apply label changes to already-running tailers

### LINKED ISSUE #6714: loki.source.docker: label changes silently ignored until restart
### Component(s)

loki.source.docker

### What's wrong?

    if s.Contains(key) {
        continue
    }

rebuilds the tailer.

    }

  v1.13.0 unannounced.

### Scope

### Why it matters

by value into `process()`

```go
package docker

import (
	"testing"

)

		cfg := `
			targets    = [
			]
			forward_to = []
		`
		var a Arguments
		return a
	}

		DataPath:   t.TempDir(),
	}, mk("staging"))
	require.NoError(t, err)

	labelsOf := func() string {
			return s.(*tailer).labelsStr
		}
		return "<none>"
	}

}
```

   error is logged.

### System information

### Software version

### Configuration

```text
}

}

loki.echo "demo" {}
```

### Logs

```text
=== RUN   TestStaleLabels
    stale_labels_test.go:48:
        	Test:       	TestStaleLabels
FAIL

```

### AI disclosure

### Tip

zhaow-de:

`loki.source.docker` silently ignored label changes for containers it was already tailing, so a
running tailer kept the label set it was constructed with until Alloy restarted. This affected the
component's own `labels` and `relabel_rules` arguments as well as target label changes arriving from
upstream `discovery.*` components.

The newly computed labels are now pushed to running tailers instead of being discarded. Labels are
updated in place rather than by restarting the tailer, so the container's log stream is not
interrupted and its read offset is preserved.

### Pull Request Details

`source.Reconcile` skips keys that are already scheduled:

```go
if s.Contains(key) {
    continue
}
```

`loki.source.docker` keys its tailers on the container ID alone, so for a running container the
source factory is never re-invoked and the freshly computed `defaultLabels` / `c.rcs` in `Update()`
are discarded. `Update()` only calls `c.scheduler.Reset()` when the Docker client changes (a `host`
edit), so nothing else rebuilds the tailer.

Three things were therefore ignored for an already-running container:

- any target label change from an upstream `discovery.*` component that did not also change
  `__meta_docker_container_id`

#### Why not restart the tailer

Treating a label change as an identity change, so the tailer is stopped and rebuilt, looks like the
smaller fix but is the wrong one. `tailer.stop()` removes the container's positions entry when the
component is not shutting down, so the replacement would start from `since=0` and re-ship the whole
container log. That is also the behaviour #5026 set out to reduce, and the direction #5135 and #5717
are moving away from.

#### What this does instead

- `tailer` computes its two relabelled per-stream label sets together and publishes them in an
  `atomic.Pointer`. `process` reads them per entry, so the hot path costs one atomic load rather
  than a lock per log line.
- `updateLabels` swaps the label set and relabel rules under the existing mutex and republishes.
- `Update()` pushes the newly computed labels to every running tailer after `Reconcile`.

The log stream is never interrupted, no entries are duplicated or dropped, and the positions file is
untouched.

`labelsStr` deliberately stays fixed at the label set the tailer was created with, because it is the
positions key: the read offset describes how far we have read in the container's log and does not
depend on how entries are labelled. Re-keying it on a label change would orphan the existing entry.
`DebugInfo` now reports the current labels, while the read offset stays keyed by the original.

#### Scope

`loki.source.file` is unaffected — it folds labels into its key
(`positions.Entry{Path, Labels.String()}`) and so already restarts on a label change.

`loki.source.kubernetes_events` has the same class of problem (`job_name` and `log_format` are
frozen for already-running namespaces). I left it out to keep this PR to one component, and am happy
to follow up.

Fixes #6714

### Notes to the Reviewer

**This is a regression, first shipped in v1.13.0.** Before #5026 the tailer identity had two parts:
`internal/runner`'s `Task` required both `Hash()` (which bucket) and `Equals()` (has it changed?),
implemented for this component as the label fingerprint and a `LabelsStr()` comparison. A label
change produced a non-equal task, so the old worker was stopped and a new one started with fresh
labels. The replacement `source.Scheduler` has only `Key()`, so the change-detection half of the
identity was dropped. Neither #5026 nor #5095 carried a CHANGELOG entry, so the behaviour change
shipped unannounced.

**Design question for you:** I considered adding a general change-detection callback to
`source.Reconcile` so all three callers get this behaviour, which would be the broader fix. I chose
the localized change to leave the shared reconciler untouched. Happy to switch if you prefer the
general approach.

**Tests.** Two are added, and I confirmed both fail on `main` before the fix:

- `TestComponentUpdatesLabelsOfRunningTailer` — the component-level case from the issue: same
  container ID, changed `labels` argument and changed target label. Also asserts the tailer is not
  restarted.
- `TestTailerUpdateLabelsAppliesToNewEntries` — feeds `process` directly so labels can change while
  the stream stays open, then asserts entries read after the change carry the new labels, entries
  read before keep the old ones, and `labelsStr` is unchanged.

`TestComponentDuplicateTargets` still passes, so the existing first-one-wins de-duplication of
targets sharing a container ID is preserved.

Verified locally:

```
go test -race -tags="nodocker gore2regex" ./internal/component/loki/...   # all pass
golangci-lint run internal/component/loki/source/docker/...              # 0 issues
go vet -tags="nodocker gore2regex" ./internal/component/loki/source/docker/
alloylint github.com/grafana/alloy/internal/component/loki/source/docker/...
```

No CHANGELOG entry, since release tooling generates those from the PR title. No documentation change:
the reference page does not state that label changes require a restart, so nothing there is made
stale by this fix.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -214,6 +214,18 @@ func (c *Component) Update(args component.Arguments) error {
 		return promTargets[i].fingerPrint < promTargets[j].fingerPrint
 	})
 
+	// Retain the winning target for each container ID so the labels of tailers
+	// which are already running can be refreshed below. Reconcile applies the
+	// same first-one-wins rule over the sorted slice.
+	targetsByID := make(map[string]promTarget, len(promTargets))
+	for _, target := range promTargets {
+		containerID := string(target.labels[dockerLabelContainerID])
+		if _, seen := targetsByID[containerID]; containerID == "" || seen {
+			continue
+		}
+		targetsByID[containerID] = target
+	}
+
 	source.Reconcile(
 		c.opts.Logger,
 		c.scheduler,
@@ -239,6 +251,18 @@ func (c *Component) Update(args component.Arguments) error {
 			)
 		},
 	)
+
+	// Reconcile only builds sources for container IDs that are not running yet,
+	// so a tailer which is already running would otherwise keep the labels it
+	// was created with forever. Push the newly computed labels and relabel
+	// rules to them instead of restarting them, which would reset their read
+	// offset and re-ship the container's log.
+	for s := range c.scheduler.Sources() {
+		t := s.(*tailer)
+		if target, ok := targetsByID[t.containerID]; ok {
+			t.updateLabels(target.labels.Merge(defaultLabels), c.rcs)
+		}
+	}
 
 	c.args = newArgs
 	return nil
--- a/internal/component/loki/source/docker/docker_test.go
+++ b/internal/component/loki/source/docker/docker_test.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/alloy/internal/component"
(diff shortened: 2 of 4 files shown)


ptodev:
Hi @zhaow-de, thank you for the PR! As you say, this is a wider problem that impacts other components. I'll see if I can open a PR which has a more general approach. Let's keep this one open until I can also open mine. Then I can invite another maintainer to take a look.

zhaow-de:
Thanks @ptodev. The general approach is what I asked about at the end of the PR description, so that answers it.

If it helps, I can implement it here instead of you opening a second PR, which leaves one PR to review rather than two. If you'd rather write the shared reconciler part yourself, that works too. I'd rebase this branch on top of yours and keep only the docker parts and the tests. Let me know which you prefer.

Rough plan, in case you want to correct the shape before I write any code:

Before #5026, `runner.Task` identified a source by `Hash()` and `Equals()`. `source.Scheduler` only has `Key()`, so `Reconcile` can tell whether a source is running but not whether it changed, and all it can do is `continue`. I'd put the second half back. When a key is already scheduled, pass the new input to the running source instead of discarding it, through an optional hook such as an `Updater[Input]` interface, with a by-key lookup added to `Scheduler`. Sources that don't implement the hook keep today's behaviour, so `loki.source.file` is unaffected, its labels being part of its key already. `updateLabels` in `loki.source.docker` becomes its implementation of the hook. `loki.source.kubernetes_events` can then do the same for `job_name` and `log_format`, which are frozen there for the same reason. The argument for updating in place rather than restarting carries over unchanged: a restart deletes the positions entry and re-ships the whole container log.

On how this sits with the other open work, since three separate defects are behind the "docker logs go dark" reports and only this one is about labels:

* #6309 covers the 5 second `No such container` retry spam, which is the second half of #3054. It doesn't depend on this PR and this PR doesn't depend on it.
* prometheus/prometheus#19236 and prometheus/prometheus#19237 (merged 21 July) cover the other half of #3054, where new containers are never discovered. On `unix://` and `npipe://` hosts the Docker client ends up with no request timeout at any layer, so a single stalled API call blocks the refresh loop for good and the target set stays frozen on the last good snapshot. It isn't in a tagged Prometheus release yet, so Alloy picks it up on a future dependency bump. That's also the upstream report you noted was missing back in March 2025.

Both of those mattered for the outage that got me here. Discovery froze, so the replacement container never showed up, and the tailers that survived kept stale labels through a `discovery.relabel` change.

ptodev:
@zhaow-de thanks a lot for pointing me to the other issues and for the suggestions! I reviewed #6309 - it should be ok for merging. On Monday I'll have a more in depth look at #3054.

I opened #6761 for the `loki.source.docker` label issue. It turns out that even `loki.source.file` is somewhat affected, but not for labels - only config arguments are not reloaded. Thank you for the help but it ended up being easier to just open the PR myself - I'm not super familiar with this code and I had to spend a few hours understanding the issue, writing tests, and making sure we are fixing the wider problem as a whole.

The other side of this issue is that the positions file still doesn't identify targets solely via container IDs. This could be fixed by something like #5717, but I'd prefer that we don't expose config options to fix it. I commented on it and will get back to it when @kalleep is back from vacation in a few days.

PR #6725, opened 2026-07-21
add per-thread frame and label options

evkuzin:

Addition to a https://github.com/grafana/pyroscope-jfr-parser/pull/116
Still WIP, for discussion

### DIFF
--- a/docs/sources/reference/components/pyroscope/pyroscope.java.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.java.md
@@ -146,6 +146,14 @@ The following arguments are supported:
 
 Refer to [profiler-options](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options) for more information about async-profiler configuration.
 
+The following blocks are supported inside `profiling_config`:
+
+| Block               | Description                                        | Required |
+| ------------------- | -------------------------------------------------- | -------- |
+| [`thread`][thread]  | Surface the sampled thread in the profile.         | no       |
+
+[thread]: #thread
+
 #### `event`
 
 The `event` argument configures the profiling mode used by async-profiler.
@@ -159,6 +167,41 @@ The `per_thread` option doesn't apply when using JFR output format.
 Since `pyroscope.java` uses JFR format exclusively, this option has no effect.
 For more details, refer to [Options applicable to any output format except JFR](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilerOptions.md#options-applicable-to-any-output-format-except-jfr) in the async-profiler documentation.
 {{< /admonition >}}
+
+#### `thread`
+
+The `thread` block surfaces the thread each sample ran on. It requires `per_thread` to be enabled.
+
+The following arguments are supported:
+
+| Name         | Type     | Description                                                                                                            | Default | Required |
+| ------------ | -------- | -------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `frame`      | `bool`   | Add the thread name as a root frame, so flame graphs split by thread.                                                 | `false` | no       |
+| `label_name` | `string` | Add a sample label under this name holding the thread name, for filtering and grouping. Empty disables it.            | `""`    | no       |
+| `regex`      | `string` | Collapse the thread name to this regular expression's first capture group, for example a pool name. Applies to both `frame` and `label_name`. | `""`    | no       |
+
+The `regex` regular expression must contain at least one capture group. It applies to both `frame` and `label_name`, so without it the raw thread name is used. Because it depends on the output medium, the following combinations produce different views:
+
+* `frame` on, no `regex`: the graph splits by each individual thread.
+* `frame` on with `regex`: the graph splits by pool.
+* `label_name` set: the graph is unchanged, but you can filter or group by thread or pool in {{< param "PRODUCT_NAME" >}}.
+
+```alloy
+pyroscope.java "java" {
+  targets    = discovery.relabel.java.output
+  forward_to = [pyroscope.write.staging.receiver]
+
+  profiling_config {
+    per_thread = true
+
+    thread {
+      frame      = false
+      label_name = "thread_pool"
+      regex      = "^(.*?)-\\d"
+    }
+  }
+}
+```
 
 ### `custom_arguments`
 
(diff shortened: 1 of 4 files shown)


docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -153,6 +143,9 @@ The following arguments are supported:
...
 | `quiet`                           | `bool`          | If set, suppresses the `Profiling started/stopped` log message.                                                                    | `false`    | no       |
 | `sample_rate`                     | `int`           | CPU profiling sample rate. It's converted from Hz to interval and passed as an `-i` argument to async-profiler.                    | `100`      | no       |
+| `thread_frame`                    | `bool`          | Adds the sampled thread's name as a root frame so flame graphs split by thread. Requires `per_thread`.                             | `false`    | no       |
+| `thread_label_name`               | `string`        | Name of a sample label derived from the thread name via `thread_label_regex`. Requires `per_thread`. Empty disables the label.     | `""`       | no       |
+| `thread_label_regex`              | `string`        | Regex whose first capture group, applied to the thread name, is the `thread_label_name` value. Must have a capture group.          | `""`       | no       |
```
clayton-cornell:
```suggestion
| `thread_label_regex`              | `string`        | Regular expression whose first capture group, applied to the thread name, is the `thread_label_name` value. Must have a capture group. | `""`       | no       |
```

We should spell out regular expression.

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -156,6 +146,14 @@ The following arguments are supported:
 
 Refer to [profiler-options](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options) for more information about async-profiler configuration.
 
+The following blocks are supported inside `profiling_config`:
+
+| Block               | Description                                        | Required |
+| ------------------- | -------------------------------------------------- | -------- |
+| [`thread`][thread]  | Surface the sampled thread in the profile.         | no       |
+
+[thread]: #thread
```
clayton-cornell: Blocks are all given in a single table. Refer to other component topics for examples of how this is done. Basically this block section needs to be moved up to the main blocks table and added as a new row (and formatted correctly to show its relationship to the parent.

internal/component/pyroscope/java/loop.go
```diff
@@ -209,6 +211,32 @@ func (p *profilingLoop) push(jfrBytes []byte, startTime time.Time, endTime time.
...
 
+// threadParseOptions builds the jfr-parser thread option from the profiling
+// config. The regex is validated at config time (Arguments.Validate), so
+// compilation here is expected to succeed; a compile error drops the transform
+// rather than failing.
+func (p *profilingLoop) threadParseOptions() []jfrpprof.Option {
+	t := p.getConfig().Thread
+	if !t.Frame && t.LabelName == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
evkuzin: fixed

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -171,13 +162,46 @@ Since `pyroscope.java` uses JFR format exclusively, this option has no effect.
...
+* `label_name` set: the graph is unchanged, but you can filter or group by thread or pool in {{< param "PRODUCT_NAME" >}}.
+
+```alloy
+pyroscope.java "java" {
+  targets    = discovery.relabel.java.output
+  forward_to = [pyroscope.write.staging.receiver]
+
+  profiling_config {
+    thread {
+      frame      = false
```
clayton-cornell: The example shows `frame=false`, but this is just the default. Could we show an example with this set to `true` instead?

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -171,13 +162,46 @@ Since `pyroscope.java` uses JFR format exclusively, this option has no effect.
 For more details, refer to [Options applicable to any output format except JFR](https://github.com/async-profiler/async-profiler/blob/master/docs/ProfilerOptions.md#options-applicable-to-any-output-format-except-jfr) in the async-profiler documentation.
 {{< /admonition >}}
```
clayton-cornell, line 175: Should we add something like this here: 

``To surface per-thread data in profiles, use the [`thread`](#thread) block instead.``

The warning says "this doesn't work", but doesn't make it clear what the reader can do.. that `thread` can be used.

docs/sources/reference/components/pyroscope/pyroscope.java.md
```diff
@@ -171,13 +162,46 @@ Since `pyroscope.java` uses JFR format exclusively, this option has no effect.
...
+| `frame`      | `bool`   | Add the thread name as a root frame, so flame graphs split by thread.                                                 | `false` | no       |
+| `label_name` | `string` | Add a sample label under this name holding the thread name, for filtering and grouping. Empty disables it.            | `""`    | no       |
+| `regex`      | `string` | Collapse the thread name to this regular expression's first capture group, for example a pool name. Applies to both `frame` and `label_name`. | `""`    | no       |
+
+The `regex` regular expression must contain at least one capture group. It applies to both `frame` and `label_name`, so without it the raw thread name is used. Because it depends on the output medium, the following combinations produce different views:
+
+* `frame` on, no `regex`: the graph splits by each individual thread.
+* `frame` on with `regex`: the graph splits by pool.
+* `label_name` set: the graph is unchanged, but you can filter or group by thread or pool in {{< param "PRODUCT_NAME" >}}.
+
```
clayton-cornell, line 198: What happens if both `frame` is `false` and `label_name` is empty? Will this block do anything in this case?

docs/sources/reference/components/pyroscope/pyroscope.java.md
evkuzin: added exactly that line.

docs/sources/reference/components/pyroscope/pyroscope.java.md
evkuzin: Nothing, the block is a no-op in that case, and regex on its own doesn't enable anything either. threadParseOptions returns no options unless frame is true or label_name is set. I've spelled that out in the docs. I also documented the no-match behavior of regex, since that's the neighboring question: a thread name that doesn't match keeps its original name rather than becoming empty.

docs/sources/reference/components/pyroscope/pyroscope.java.md
evkuzin: Switched the example to frame = true and added a sentence explaining what it produces with an example

PR #6743, opened 2026-07-22
feat(database_observability.sql_server): Add `query_metrics` collector

cristiangreco:

Add a basic `query_metrics` collector implementation that reads per-query execution stats from Query Store and emits them as Prometheus counters (only RED metrics), keyed by `database` and `query_hash`.

### Pull Request Details

Notes on initial design decisions:
- Rank the top-N queries by duration within the recent lookback window, but sum each query's full retained Query Store history. This keeps the emitted counters monotonic across Query Store interval rollover: the windowed ranking changes as intervals roll over, but the summed totals only grow.

- Turn Query Store's cumulative totals into monotonic Prometheus counters by tracking a per-series baseline. A first sighting is created at zero so that historical totals don't surface as a startup spike, and a decrease in any total (e.g. Query Store cleared, retention or plan cleanup) re-baselines the series instead of emitting a spurious jump.

- Bound registry cardinality by pruning queries/series not seen within a TTL window, so queries that drop out of the top-N are removed. A query that flaps around the boundary will keep its baseline within the TTL, so activity accrued while it was absent is counted on re-entry rather than lost to a re-baseline.

- On instances with large Query Stores the collection query might be potentially slow as it intentionally sums the full retained history per ranked hash. This is deemed acceptable for now.

- Scope is limited to only the connected database. Also, when Query Store is not available or not `READ_WRITE`, we skip collection while preserving existing baselines, as a reasonable tradeoff.

None of these decisions are final, and the implementation is a work in progress. The internal state might be reworked to be exposed to other collectors in the future.

Relates to https://github.com/grafana/grafana-dbo11y-app/issues/3023

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/sql_server/collector/query_metrics.go
+++ b/internal/component/database_observability/sql_server/collector/query_metrics.go
@@ -1,0 +1,309 @@
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+const QueryMetricsCollector = "query_metrics"
+
+// selectQueryStoreState reads the connected database's Query Store state
+const selectQueryStoreState = `
+SELECT
+	DB_NAME(),
+	actual_state_desc,
+	query_capture_mode_desc,
+	readonly_reason
+FROM sys.database_query_store_options`
+
+// selectQueryMetrics ranks the top-N queries by duration within the recent lookback
+// window, then sums their full retained Query Store history so the emitted
+// counters stay monotonic across interval rollover.
+//
+// Note: the final SELECT sums the full retained history per ranked hash
+// (no lookback filter on the aggregation), while the lookback window only
+// constrains the top-N ranking/eligibility in the eligible/ranked CTEs.
+// In internal state, on a Query Store bucket rollover (when interval ends),
+// the cumulative totals keep growing → deltas stay non-negative → the
+// Prometheus counters remain monotonic. The only time totals decrease
+// is when Query Store is cleared, or when retention/cleanup drops an old
+// interval: in that case the state re-baselines rather than emitting a spike.
+const selectQueryMetrics = `
+WITH eligible AS (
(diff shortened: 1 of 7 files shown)


internal/component/database_observability/sql_server/collector/query_metrics.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+func (c *QueryMetrics) Start(ctx context.Context) error {
+	c.logger.Debug("collector started")
+
+	var registered []prometheus.Collector
+	for _, m := range c.metrics {
+		if err := c.registry.Register(m); err != nil {
+			for _, done := range registered {
+				c.registry.Unregister(done)
```
matthewnolf, line 159: Doesn't this have the potential to unregister those we just registered - even if it's not the same metric

internal/component/database_observability/sql_server/collector/query_metrics_state.go
```diff
@@ -0,0 +1,135 @@
...
+
+	for _, source := range sources {
+		key := queryMetricsKey{database: source.database, queryHash: source.queryHash}
+
+		previous, exists := s.entries[key]
+		if !exists {
+			// First sighting: create all three series at zero so downstream
+			// ratios resolve to 0 rather than no-data, and store the baseline
+			// without emitting the historical totals as a startup spike.
+			s.executions.WithLabelValues(source.database, source.queryHash).Add(0)
```
matthewnolf, line 72: This has the potential to always be behind by the first delta which can lead to inaccurate reporting

matthewnolf (approved):
Approved on the grounds we acknowledge the present limitations and will work to address in followup given more test data can be collected

internal/component/database_observability/sql_server/collector/query_metrics.go
cristiangreco: Yes, this is intended cleanup - this will unregister all metrics from this collector and return an error, preventing collector start.

internal/component/database_observability/sql_server/collector/query_metrics_state.go
cristiangreco: Potential undercounting by design, a tradeoff that will be re-evaluated.

PR #6750, opened 2026-07-23
fix(loki.source.file, loki.source.journal): Fail loudly on unreadable legacy positions file

### LINKED ISSUE #5493: Kill Alloy if 'permission denied' on legacy positions file when running with `config.format`
### Component(s)

_No response_

### Background

I am following the [promtail migration guide](https://grafana.com/docs/alloy/latest/set-up/migrate/from-promtail/). I confirmed the conversion tool does properly set the legacy positions file.
However, when I `alloy run --config.format=promtail promtail-config.yml`, I got:

```
ts=2026-02-10T14:01:20.207556224Z level=error msg="error reading legacy positions file" component_path=/ component_id=loki.source.file.varlogs path=/home/ubuntu/promtail/positions.yml error="open /home/ubuntu/promtail/positions.yml: permission denied"
```

That part's on me!
But Alloy then went ahead and ignored the positions file and blew up my server by re-logging every file we've ever created. 

### Proposal

At _least_ in the 'direct conversion of promtail config' mode, kill Alloy if the positions file can't be read. I'd rather it fall over than blow up my server.

(I think it makes sense to always die if the positions file can't be read, but that's a much bigger change)

### Tip

allocsys:

Fixes `loki.source.file` and `loki.source.journal` silently re-ingesting all
tailed logs from the beginning when the legacy promtail positions file exists
but can't be read (e.g. permission denied). `readLegacyFile` previously
logged the read error and returned `nil`, which callers treated identically
to "file doesn't exist" — this now distinguishes the two cases and propagates
a real error, failing component construction instead of proceeding silently.
Adds regression tests covering the permission-denied case for both the file
and journal conversion paths.

### Pull Request Details

`readLegacyFile` now returns `(*LegacyFile, error)` instead of just
`*LegacyFile`. A non-nil error means the legacy file exists but a genuine
I/O error prevented reading it (e.g. permission denied); this is distinct
from the file simply not existing, which remains `(nil, nil)`. Malformed
YAML content is still treated as non-fatal, matching prior behavior — the
scope of this fix is I/O errors on the file itself, not validation of its
contents.

`ConvertLegacyPositionsFile` and `ConvertLegacyPositionsFileJournal` now
propagate that error to their callers (`loki.source.file` and
`loki.source.journal` respectively), which fail component construction
instead of proceeding as if no legacy positions file existed.

Fixes #5493

### Notes to the Reviewer

The permission-denied regression tests (`TestLegacyConversionUnreadableFile`,
`TestConvertLegacyPositionsFileJournalUnreadableFile`) use `os.Chmod` and are
skipped on Windows and when running as root, since neither environment
enforces the permission bits the test relies on.

### PR Checklist

- [x] Tests updated


internal/component/loki/source/internal/positions/positions_test.go
```diff
@@ -113,6 +114,64 @@ func TestLegacyConversionWithNoLegacyFile(t *testing.T) {
...
+	legacy := writeLegacy(t, tmpDir)
+	require.NoError(t, os.Chmod(legacy, 0000))
+	defer func() {
+		_ = os.Chmod(legacy, 0644) // restore so t.TempDir() cleanup can remove it
+	}()
+
+	positionsPath := filepath.Join(tmpDir, "positions")
+	err := ConvertLegacyPositionsFile(legacy, positionsPath, logging.NewSlogNop())
+	require.Error(t, err)
+	require.True(t, strings.Contains(err.Error(), legacy))
```
dehaansa: The test should confirm that the expected error (failed to read/check) is returned, not just that the error contains the filename.

internal/component/loki/source/file/file.go
```diff
@@ -216,8 +216,14 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	}
 	newPositionsPath := filepath.Join(o.DataPath, "positions.yml")
 	// Check to see if we can convert the legacy positions file to the new format.
+	// A non-nil error here means the legacy file exists but couldn't be read
+	// (e.g. permission denied) -- fail component construction rather than
+	// silently proceeding as if there were no legacy positions, which would
+	// re-ingest every tailed file from the beginning (#5493).
```
ptodev: There's no need to have such lenghty comments.

internal/component/loki/source/internal/positions/positions_test.go
```diff
@@ -113,6 +114,68 @@ func TestLegacyConversionWithNoLegacyFile(t *testing.T) {
 	}
 }
 
+func TestLegacyConversionUnreadableFile(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("chmod-based permission test is not supported on windows")
+	}
+	if os.Geteuid() == 0 {
+		t.Skip("test requires running as a non-root user so file permissions are enforced")
+	}
```
ptodev: These restrictions make the test harder to run. It'd be better to have a test that can run anywhere. The only ways to do it is either to mock the file system or to make a proper integration test. 

WDYT about keeping the test but changing it so that the error isn't due to permissions? E.g. it can just be that the file is actually a directory? Something like:

```
func TestConvertLegacyPositionsFileUnreadableFile(t *testing.T) {
	tmpDir := t.TempDir()
	legacy := filepath.Join(tmpDir, "legacy.yaml")
	require.NoError(t, os.Mkdir(legacy, 0750))

	positionsPath := filepath.Join(tmpDir, "positions")
	err := ConvertLegacyPositionsFile(legacy, positionsPath, logging.NewSlogNop())
	require.ErrorContains(t, err, "could not be read")

	_, statErr := os.Stat(positionsPath)
	require.True(t, os.IsNotExist(statErr))
}
```

allocsys:
@ptodev Both addressed, thanks for the suggestions:

- Trimmed the comment in `file.go` down to the single-line summary.
- Replaced the `chmod`-based permission tests with the directory-as-legacy-file approach you suggested (`os.Mkdir` instead of writing+chmod'ing a file) — dropped the Windows/root `t.Skip` guards along with it, and removed the now-unused `runtime` import.

Note the branch was rebased afterward to get proper commit signatures, so the SHAs your comments are anchored to (`2bb0150`, `65e172a`) have moved — the diffs may show as outdated but the fixes are in the current tip.

ptodev (dismissed):

internal/component/loki/source/internal/positions/positions.go
```diff
@@ -103,18 +103,27 @@ type LegacyFile struct {
...
 // 1. There is no file at the newpath
 // 2. There is a file at the legacy path and that it is valid yaml
-func ConvertLegacyPositionsFile(legacyPath, newPath string, l *slog.Logger) {
-	legacyPositions := readLegacyFile(legacyPath, l)
+//
+// Returns an error if the legacy file exists but could not be read (e.g.
+// permission denied) -- as opposed to simply not existing, which is not an
+// error. Callers should treat a non-nil error as fatal: proceeding despite
+// an unreadable legacy positions file risks silently re-ingesting every
+// tailed file from the beginning (#5493).
```
ptodev: Would you mind trimming the comments please?

allocsys:
Good call, thanks! Trimmed the flagged comment and a few other spots in the same file that were repeating the same explanation — pushed in 41deef8.

ptodev:
@allocsys don't worry, there's no need to update the branch with main if it's just a bit out of date :)

allocsys:
@ptodev Sorry about that — this was from an agentic automation job I had running, and I've now disabled it.

ptodev (approved):
Thank you!

ptodev:
@allocsys I'm sorry but I can't merge because there's one commit that doesn't have a signature (41deef847936d515ec0670111e148ccd265e175c). Do you mind signing it please?

allocsys:
@ptodev The commit is signed now — should be good to merge. Thanks for flagging it!

ptodev (approved):

PR #6754, opened 2026-07-23
feat(pyroscope.java): Add tlab profiling config option for TLAB-based allocation events

github-hamza-bouqal:

Adds a `tlab` boolean to `profiling_config` that passes `--tlab` to async-profiler alongside `--alloc`, switching allocation profiling from the statistically sampled `jdk.ObjectAllocationSample` events (produced by the JVMTI ObjectSampler, default since v4.0) to the fully-instrumented `jdk.ObjectAllocationOutsideTLAB` and `jdk.ObjectAllocationInNewTLAB` events (produced by the TLAB-hook AllocTracer engine). The TLAB engine captures every allocation without sampling gaps, at the cost of higher overhead — prefer it when sampling bias would skew results, not as a general production default.

**Requires async-profiler v4.4+**. The `--tlab` engine-selection flag was introduced in v4.4 (issue #[1671](https://github.com/async-profiler/async-profiler/pull/1671)). On earlier versions the flag is unrecognised; no config-layer version check guards against this.

Without this, users who want TLAB-based allocation events had to abandon structured config and fall back to `custom_arguments`. 

Setting `tlab = true` without alloc is rejected at config parse time, since `--tlab` only affects engine selection inside the `EM_ALLOC` event path; without `--alloc` it has no effect.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/pyroscope/pyroscope.java.md
+++ b/docs/sources/reference/components/pyroscope/pyroscope.java.md
@@ -153,6 +153,7 @@ The following arguments are supported:
 | `per_thread`                      | `bool`          | Sets per thread mode on async profiler. It's passed as an `-t` argument to async-profiler.                                         | `false`    | no       |
 | `quiet`                           | `bool`          | If set, suppresses the `Profiling started/stopped` log message.                                                                    | `false`    | no       |
 | `sample_rate`                     | `int`           | CPU profiling sample rate. It's converted from Hz to interval and passed as an `-i` argument to async-profiler.                    | `100`      | no       |
+| `tlab`                            | `bool`          | Enables TLAB-based allocation events. Passed as `--tlab` to async-profiler. Requires `alloc` to be set.                            | `false`    | no       |
 
 Refer to [profiler-options](https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options) for more information about async-profiler configuration.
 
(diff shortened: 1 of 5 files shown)


github-hamza-bouqal:
Before / After: `--tlab` in action

[Screenshot 1] — Standard `profiling_config` with alloc = "512k". Uses the default ObjectSampler engine (jdk.ObjectAllocationSample).

image


[Screenshot 2] — Grafana dashboard with that config. Only `memory/alloc_in_new_tlab_*` metrics are available, **outside-TLAB allocations** (large objects allocated directly on the heap, bypassing TLAB) **are invisible**.

image

[Screenshot 3] — Same config rewritten with **custom_arguments** and `--tlab`. Switches to the `AllocTracer` engine (jdk.ObjectAllocationInNewTLAB + jdk.ObjectAllocationOutsideTLAB).

image

[Screenshot 4] — Grafana dashboard now exposes both `memory/alloc_in_new_tlab_*` and `memory/alloc_outside_tlab_*`. Large object allocations are now visible and attributable to call stacks.

image


This is exactly the gap **tlab = true** closes, without it, any allocation that bypasses TLAB (objects too large for the thread-local buffer) is silently dropped from the profile.

I used this [example](https://github.com/grafana/alloy-scenarios/tree/main/java-profiling) to validate the changes locally.

korniltsev-grafanista:
please  sign your git commit with gpg or ssh

korniltsev-grafanista (approved):

PR #6761, opened 2026-07-24
fix(loki.source.docker): Updates in target labels are now taken into account

### LINKED ISSUE #6714: loki.source.docker: label changes silently ignored until restart

ptodev:
### Pull Request Details

Seems like #5026 broke label updates in `loki.source.docker`. Prior to #5026, a label update would be considered a new target even if the container ID is the same. This meant there was also a new positions entry, and I believe the container logs would be re-read. This was because `manager.syncTargets` would use `Runner.ApplyTasks` which seems to use the hash of the label set.

#5026 seems like a step in the right direction, because restarting the whole task when the container ID doesn't change doesn't seem efficient. It'd be good to keep this change, to fix the update mechanism (what this PR does), and in the future to also somehow fix the positions file so that when Alloy is restarted it doesn't re-read the same container logs again.

IIUC Alloy would previously re-read the logs when Update() was called. With this PR, this would be postponed to the next time Alloy restarts (that's when it will find out that the position in the file is different).

The PR consists of two commits. The first commit adds tests which fail. The second commit fixes the tests.

Fixes #6714

### Notes to the Reviewer

This is the Docker-specific part of #6759. The other changes in #6759 will be merged in separate pull requests.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -238,6 +238,11 @@ func (c *Component) Update(args component.Arguments) error {
 				func() bool { return c.exited.Load() },
 			)
 		},
+		func(existing source.Source[string], target promTarget) (source.Source[string], error) {
+			t := existing.(*tailer)
+			t.updateLabels(target.labels.Merge(defaultLabels), c.rcs)
+			return nil, nil
+		},
 	)
 
 	c.args = newArgs
--- a/internal/component/loki/source/docker/docker_test.go
+++ b/internal/component/loki/source/docker/docker_test.go
@@ -1,13 +1,23 @@
 package docker
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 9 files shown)


ptodev:
Closing this since #6818 fixed it in a more minimal way.

PR #6766, opened 2026-07-27
fix: Prevent loki.write WAL watcher panic when a record spans a page boundary

### LINKED ISSUE #6757: loki.write WAL panics in LiveReader.readRecord when a record spans a page and metrics is nil
### Component(s)

loki.write

### What's wrong?

When the experimental WAL is enabled for `loki.write`, Alloy can panic while reading a WAL segment whose next record appears to span a page boundary.

  We observed 19 container restarts over approximately 35 hours. All 19 panics had the same stack frame:

  `github.com/prometheus/prometheus/tsdb/wlog.(*LiveReader).readRecord`
  at `tsdb/wlog/live_reader.go:311`.

  The immediate cause appears to be a nil metrics pointer:

  1. Alloy creates the live reader with nil metrics:

     https://github.com/grafana/alloy/blob/v1.16.1/internal/component/common/loki/wal/watcher.go#L203-L210

     `wlog.NewLiveReader(..., nil, segment)`

  2. When a record appears to span a page boundary, LiveReader unconditionally dereferences the metrics pointer:

     https://github.com/grafana/prometheus/blob/87200e297b57/tsdb/wlog/live_reader.go#L304-L313

     `r.metrics.readerCorruptionErrors.WithLabelValues("record_span_page").Inc()`

  Because `r.metrics` is nil, this branch panics instead of reporting or tolerating the page-boundary condition. The SIGSEGV address was `0x10`, which
  is consistent with dereferencing the second field of a nil `LiveReaderMetrics` value.

  After the final restart, the following WAL segment repeatedly produced:

  `segment 31: unexpected non-zero byte in page term bytes`

  This indicates that the panic is triggered by a WAL structural/alignment problem, but a corrupt or unusual WAL record should not crash the entire
  Alloy process.

  The repeated restarts also caused delayed WAL replay and downstream `too_far_behind` rejections from Loki.

  The same nil metrics argument and unconditional dereference appear to still be present on the current main branches.

  Expected behavior:

  - The page-boundary/corruption condition should be reported as a WAL read error or warning.
  - Alloy should not panic because optional metrics were not provided.
  - For a closed corrupt segment, the watcher should follow its existing error-handling path and continue or skip the segment as appropriate.

  Possible fixes include either:

  - passing a non-nil `LiveReaderMetrics`, such as one created with a nil registerer; or
  - guarding the metric increment when `r.metrics == nil`.

### Steps to reproduce

  The exact WAL byte sequence has not yet been reduced to a standalone fixture, but the failing code path can be reproduced conceptually as follows:

  1. Run Grafana Alloy v1.16.1 with WAL enabled for `loki.write`.
  2. Use persistent storage for the Alloy storage path.
  3. Produce enough logs for multiple WAL segments to be created.
  4. Cause or provide a WAL segment where the next record header makes
     `readIndex + recordHeaderSize + length > pageSize`.
  5. Start or restart Alloy so that the `loki.write` WAL watcher reads that segment.
  6. Observe a nil pointer panic at `live_reader.go:311`.

  A minimal unit test should also be possible by constructing a `wlog.LiveReader` with nil metrics and input that enters the permissive
  `record_span_page` branch, then calling `Next()`.

  In our deployment, every one of the 19 observed restarts reached the same `live_reader.go:311` frame.

### System information

Linux/amd64   Ubuntu 22.04.5 LTS   Kernel 5.15.0-127-generic   Kubernetes v1.33.5   containerd 1.7.28   Persistent volume-backed Alloy storage path

### Software version

  Embedded WAL reader dependency:   github.com/grafana/prometheus@v1.8.2-0.20260313093229-87200e297b57

### Configuration

```text
loki.write "default" {
    endpoint {
      url                 = "http://loki.example/loki/api/v1/push"
      max_backoff_period  = "30s"
      max_backoff_retries = 20
    }

    wal {
      enabled         = true
      max_segment_age = "2h"
    }
  }
```

### Logs

```text
panic: runtime error: invalid memory address or nil pointer dereference
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x10]

  goroutine 554 [running]:
  github.com/prometheus/prometheus/tsdb/wlog.(*LiveReader).readRecord(...)
      /go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20260313093229-87200e297b57/tsdb/wlog/live_reader.go:311 +0x14e
  github.com/prometheus/prometheus/tsdb/wlog.(*LiveReader).buildRecord(...)
      /go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20260313093229-87200e297b57/tsdb/wlog/live_reader.go:194 +0x52
  github.com/prometheus/prometheus/tsdb/wlog.(*LiveReader).Next(...)
      /go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20260313093229-87200e297b57/tsdb/wlog/live_reader.go:143 +0x1c
  github.com/grafana/alloy/internal/component/common/loki/wal.(*Watcher).readSegment(...)
      /src/alloy/internal/component/common/loki/wal/watcher.go:315 +0x4a
  github.com/grafana/alloy/internal/component/common/loki/wal.(*Watcher).watch(...)
      /src/alloy/internal/component/common/loki/wal/watcher.go:278 +0x739
  github.com/grafana/alloy/internal/component/common/loki/wal.(*Watcher).run(...)
      /src/alloy/internal/component/common/loki/wal/watcher.go:180 +0x46b
  github.com/grafana/alloy/internal/component/common/loki/wal.(*Watcher).mainLoop(...)
      /src/alloy/internal/component/common/loki/wal/watcher.go:134 +0x18f
  created by github.com/grafana/alloy/internal/component/common/loki/wal.(*Watcher).Start
      /src/alloy/internal/component/common/loki/wal/watcher.go:121 +0x6c

  After restart:

  level=error msg="error tailing WAL" err="segment 31: unexpected non-zero byte in page term bytes"
```

### AI disclosure

### Tip

emmayusufu:
When the experimental WAL is enabled for loki.write, the watcher builds its LiveReader with nil metrics, so the moment a record spans a page boundary the reader panics on the nil counter instead of just logging the corruption. This is what #6757 hit, 19 restarts in 35 hours from the same stack frame.

The fix passes a LiveReaderMetrics built with a nil registerer, meaning that the branch increments an unregistered counter instead of crashing. Also added a regression test that feeds the reader a crafted page where a record header runs past the page size. It panics without the fix and passes with it.

Closes #6757

### DIFF
--- a/internal/component/common/loki/wal/watcher.go
+++ b/internal/component/common/loki/wal/watcher.go
@@ -185,15 +185,6 @@ func (w *Watcher) run() error {
 	return nil
 }
 
-// newLiveReader builds a wlog.LiveReader for the given segment. It always
-// passes a non-nil *wlog.LiveReaderMetrics (with a nil registerer, since
-// there is nowhere to register it): wlog.LiveReader.readRecord dereferences
-// the metrics unconditionally when a record spans a page boundary, so
-// passing literal nil there panics instead of just skipping the metric.
-func newLiveReader(logger *slog.Logger, r io.Reader) *wlog.LiveReader {
-	return wlog.NewLiveReader(logger, wlog.NewLiveReaderMetrics(nil), r)
-}
-
 // watch will start reading from the segment identified by segmentNum.
 // If an EOF is reached and tail is true, it will keep reading for more WAL records with a wlog.LiveReader. Periodically,
 // it will check if there's a new segment, and if positive read the remaining from the current one and return.
@@ -208,7 +199,8 @@ func (w *Watcher) watch(segmentNum int, tail bool) error {
 	}
 	defer segment.Close()
 
-	reader := newLiveReader(w.logger, segment)
+	// Metrics must be non-nil (built with a nil registerer), the LiveReader panics otherwise.
+	reader := wlog.NewLiveReader(w.logger, wlog.NewLiveReaderMetrics(nil), segment)
 
 	readTimer := newBackoffTimer(w.minReadFreq, w.maxReadFreq)
 
--- a/internal/component/common/loki/wal/watcher.go
+++ b/internal/component/common/loki/wal/watcher.go
@@ -185,6 +185,15 @@ func (w *Watcher) run() error {
 	return nil
 }
 
+// newLiveReader builds a wlog.LiveReader for the given segment. It always
+// passes a non-nil *wlog.LiveReaderMetrics (with a nil registerer, since
+// there is nowhere to register it): wlog.LiveReader.readRecord dereferences
+// the metrics unconditionally when a record spans a page boundary, so
+// passing literal nil there panics instead of just skipping the metric.
+func newLiveReader(logger *slog.Logger, r io.Reader) *wlog.LiveReader {
+	return wlog.NewLiveReader(logger, wlog.NewLiveReaderMetrics(nil), r)
+}
+
 // watch will start reading from the segment identified by segmentNum.
 // If an EOF is reached and tail is true, it will keep reading for more WAL records with a wlog.LiveReader. Periodically,
 // it will check if there's a new segment, and if positive read the remaining from the current one and return.
@@ -199,7 +208,7 @@ func (w *Watcher) watch(segmentNum int, tail bool) error {
 	}
 	defer segment.Close()
 
-	reader := wlog.NewLiveReader(w.logger, nil, segment)
+	reader := newLiveReader(w.logger, segment)
 
 	readTimer := newBackoffTimer(w.minReadFreq, w.maxReadFreq)
 
--- a/internal/component/common/loki/wal/watcher_test.go
+++ b/internal/component/common/loki/wal/watcher_test.go
@@ -1,7 +1,10 @@
 package wal
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
+	"hash/crc32"
 	"log/slog"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


emmayusufu:
The Govulncheck failure is not from this change. It flags GO-2026-6061 in google.golang.org/grpc plus three Moby advisories in github.com/docker/docker, all from the dependency set on main, and this PR only touches the WAL watcher and its test with no go.mod changes. So any PR on the current base should hit the same failure until the deps are bumped.

internal/component/common/loki/wal/watcher.go
```diff
@@ -185,6 +185,15 @@ func (w *Watcher) run() error {
...
 }
 
+// newLiveReader builds a wlog.LiveReader for the given segment. It always
+// passes a non-nil *wlog.LiveReaderMetrics (with a nil registerer, since
+// there is nowhere to register it): wlog.LiveReader.readRecord dereferences
+// the metrics unconditionally when a record spans a page boundary, so
+// passing literal nil there panics instead of just skipping the metric.
+func newLiveReader(logger *slog.Logger, r io.Reader) *wlog.LiveReader {
+	return wlog.NewLiveReader(logger, wlog.NewLiveReaderMetrics(nil), r)
+}
```
kalleep: No need for a seperate function here you can just inline `wlog.NewLiveReaderMetrics(nil)` at the only callsite we have

kalleep (commented):
Thanks left one comment. The comments added are way to excessive for this. Just a simple comment that we need to pass wlog.NewLiveReaderMetrics with `nil` for the register is enough. Same for the test tone down or remove the comment

kalleep:
> The Govulncheck failure is not from this change. It flags GO-2026-6061 in google.golang.org/grpc plus three Moby advisories in github.com/docker/docker, all from the dependency set on main, and this PR only touches the WAL watcher and its test with no go.mod changes. So any PR on the current base should hit the same failure until the deps are bumped.

That check is not required and if you rebase on main it should pass anyway.

kalleep (approved):
LGTM thanks!

PR #6784, opened 2026-07-29
perf(loki.process): Cache compiled regexes in template stage functions to improve performance

pikachuabc:

The `regexReplaceAll` and `regexReplaceAllLiteral` template functions used by the `loki.process` template stage compiled their regular expression via `regexp.MustCompile` on every invocation — i.e. once per log line — which was a significant source of CPU usage and heap allocations on high-volume pipelines.

We cache a regex instance to speed up the pipeline

A microbenchmark of `regexReplaceAll` shows roughly 18µs → 2.8µs per call and 57 → 5 allocations/op. Tests cover cache reuse, distinct expressions, invalid-expression handling, and the cache bound.

### Pull Request Details

We found that some times Alloy falling behind on a high-volume file. After investigating with runtime pprof:
The CPU profile shows 80% of the time is spent on compiling the regex.
```
templateStage.Process
  → text/template (*Template).Execute → evalFunction → evalCall → safeCall
    → reflect.Value.Call
      → stages.init.func3          (= regexReplaceAll)
          → regexp.MustCompile        4.83s  (82% of the function)
          → regexp.(*Regexp).ReplaceAllString  1.05s  (18%, the actual work)
```
The Heap profile shows regex compilation takes 30% of the all heap 
```
regexp/syntax.(*compiler).inst        15.31%
regexp/syntax.(*parser).newRegexp      9.45%
regexp/syntax.(*Regexp).Simplify       6.48%
```

Thus we are thinking of using cache for the regex to speed up the pipeline

### Notes to the Reviewer

This only works for static regex (which should be the most case?). However it does introduce more logic. Increase the cache size may be better.

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/loki/process/stages/template.go
+++ b/internal/component/loki/process/stages/template.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"text/template"
 	"time"
 
@@ -25,6 +26,24 @@ import (
 var (
 	ErrTemplateSourceRequired = errors.New("template source value is required")
 )
+
+// cachedRegexp is an immutable (expression, compiled) pair stored in regexpCache.
+type cachedRegexp struct {
+	expr string
+	re   *regexp.Regexp
+}
+
+// regexpCache keeps only the most recently compiled expression
+var regexpCache atomic.Pointer[cachedRegexp]
+
+func mustCompileCachedRegexp(expr string) *regexp.Regexp {
+	if c := regexpCache.Load(); c != nil && c.expr == expr {
+		return c.re
+	}
+	re := regexp.MustCompile(expr)
+	regexpCache.Store(&cachedRegexp{expr: expr, re: re})
+	return re
+}
 
 var extraFunctionMap = template.FuncMap{
 	"ToLower":    strings.ToLower,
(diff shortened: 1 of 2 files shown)


kalleep:
So this would work if you only have one regex in templates but would fall apart and make it worse than before if you have two or more. So if we really want to fix this we need to be able to cache all of them

pikachuabc:
Thanks for the reply! Yes we could cache them all the regex in a map. However, if we have variable in the regex expression then it will cause the cache grow unlimited. 

And you are absolutely right on the multi-regex case that this has extra logic. But I assume this is a cheap operation which has limit impact to the performance.

So a solution might be we use a fixed size cache? This really improves a lot performance in our single regex case.

kalleep:
Yes you would need a bounded cache, maybe something like LRU

kalleep:
>And you are absolutely right on the multi-regex case that this has extra logic. But I assume this is a cheap operation which has limit impact to the performance.

Still it would be extra work and you would never get any benefit from the cache because it would swap the cached regex all the time

pikachuabc:
Yeah... tradeoffs. What's you suggestion? shall we make it configurable or leave it.

kalleep:
So the current solution is not really something I would consider adding. 

I see you also created a [issue for this](https://github.com/grafana/alloy/issues/6792) and mark it as proposal, but it's really not. It would be better to propose a few different ideas you have to solve it in that issue and we can discuss it there before jumping into a pr. That way us in alloy have a chance to approve a solution before even looking at code.

PR #6796, opened 2026-07-30
docs: Validate beyla.ebpf component topic and consolidate examples

clayton-cornell:
Cleanup on the beyla.ebpf topic

1. Added HTML comments to the output block section regarding some issues found during validation that could not be confirmed well enough for doc changes
2. Added some other HTML comments for future followup
3. Added a review date in the frontmatter
4. Found a deprecated arg and updated arg table
5. Fixed product variable name use
6. Added missing supported values for `instrumentations`
7. Removed duplicated example
8. Stripped out the big examples at the bottom of the topic
9. Added a link out to the Alloy scenario that provides a fully runnable example showing the same features as the old example that was stripped out.


docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -7,6 +7,7 @@ labels:
   stage: general-availability
   products:
     - oss
+review_date: 2026-07-30
```
ptodev, line 10: Isn't every doc supposed to be up to date for its respective release? 😄

docs/sources/reference/components/beyla/beyla.ebpf.md
```diff
@@ -1060,136 +1072,41 @@ A scrape job that targets only the {{< param "PRODUCT_NAME" >}} pprof endpoints
...
-```
-
-Replace the following:
+* _`<LABEL>`_: A unique label for the component instance.
+* _`<OPEN_PORT>`_: The port of the running service to instrument with eBPF.
+* _`<SERVICE_NAME>`_: The name to assign to the instrumented service in exported metrics and traces.
 
-* _`<OPEN_PORT>`_: The port of the running service for Beyla automatically instrumented with eBPF.
-* _`<OTLP_ENDPOINT>`_: The endpoint of the OpenTelemetry Collector to send traces to.
+For a complete working example that collects both metrics and traces from a running service using `beyla.ebpf`, refer to the [beyla-zero-code-instrumentation Alloy scenario][beyla-zero-code-instrumentation scenario].
```
ptodev: Nice to see links to Alloy scenarios in the example sections!

It'd be great to get feedback from the Beyla team on this PR and on the example in the scenarios repo. Currently Beyla is mostly for k8s, so a docker-only example may not be very representative or real world usage. But it might be ok.

docs/sources/reference/components/beyla/beyla.ebpf.md
clayton-cornell: Yes... this is a metadata attribute that has existed for a long time and I wanted to see how it'd work as a "freshness" date for docs.

There's a lot of documentation that hasn't had a validation review for who knows how long. With this date, we can scan the markdown for topics that might need a refresh (AI is really good at this. but we can also script it).

This isn't a hard requirement, just something Jack and I have been chatting about off and on for a couple of months

docs/sources/reference/components/beyla/beyla.ebpf.md
clayton-cornell: 100%. There needs to be some Beyla eyes on this. The validity review (code comparison) turned up some questions that need some expert eyes.

clayton-cornell:
```suggestion
For a complete working example that collects both metrics and traces from a running service using `beyla.ebpf`, refer to the [Zero-code instrumentation with Beyla][beyla-zero-code-instrumentation scenario] scenario.
```

docs/sources/reference/components/beyla/beyla.ebpf.md
ptodev: I don't mind if it's there, but not sure what it means. An old doc might still be accurate. It might be old just because it doesn't need a refresh.

I suppose AI can find a lot of out of date content by just looking for contradictions between different pages, and usage of attributes and blocks that are labeled as deprecated.

clayton-cornell: That's true. If an old doc is still valid and verified, the date is updated, and we let it ride for 6 months or whatever cycle we want to use to flag a "Hey is this still valid" check.

It's still up for discussion whether it'll be a good idea. :-)

PR #6797, opened 2026-07-30
docs: Validate database_observability.mysql  topic and consolidate examples

clayton-cornell:
Cleanup on the database_observability.mysql  topic

1. Fix datatypes
2. Add missing Exported fields section
3. Strip big example
4. Create new task topic and import stripped example
5. Fix critical config error/mistake found in the original example,  `discovery.relabel.orders_db.targets` is invalid, must be `discovery.relabel.orders_db.output`
6. Develop/build the new task topic.


docs/sources/collect/mysql-database-metrics.md
```diff
@@ -0,0 +1,304 @@
...
+  Refer to the [MySQL performance_schema documentation](https://dev.mysql.com/doc/refman/8.0/en/performance-schema.html) for setup instructions.
+* Have write access to a Prometheus-compatible endpoint or Grafana Cloud for storing metrics.
+* Have write access to a Loki-compatible endpoint or Grafana Cloud for storing logs.
+* Be familiar with the concept of [Components][] in {{< param "PRODUCT_NAME" >}}.
+
+## Collect MySQL database observability data
+
+The `database_observability.mysql` component connects to a MySQL database and collects performance schema data.
+This data includes query details, execution plans, and lock information forwarded as logs to Loki.
+The component also exports targets that can be scraped with Prometheus to collect MySQL metrics.
```
ptodev, line 47:
```suggestion
The `database_observability.mysql` component connects to a MySQL database and collects performance telemetry:
* Loki logs for query details, execution plans, and lock information.
* Prometheus metrics for <>
```
Nit: it'd be nice to avoid general and ambigous terms such as "data" and to be more concise and straight to the point. People probably just wonder what telemetry they can collect. 

Also, logs and metrics are mentioned as sort of two different things, and it doesn't seem clear that these are the two signals that are being gathered.

In any case, it's probably best to mention this at the beginning of the page? There's already a list of bullet points there for similar things.

docs/sources/collect/mysql-database-metrics.md
```diff
@@ -0,0 +1,304 @@
...
+   }
+   ```
+
+   Replace the following:
+
+   * _`<LABEL>`_: Must match the label used for the `database_observability.mysql` component.
+
+   For more information, refer to the [`loki.relabel`][loki.relabel] documentation.
+
+4. Add a `discovery.relabel` component to standardize labels on Prometheus targets.
```
ptodev, line 119: It's sad how we have to spend so much time describing relabelling and remote write. I agree that a user who wants to set up this component has to understand these concepts, but it feels like we're making the experience needlessly complicated. It would get especially repetitive if we have to mention this in every doc. Then a user who has to set up both MSSQL and MySQL has to read through or scroll through identical content. Is there a way to reduce this? E.g. we could just show a simple example which contains those components and include a link to a doc which explains how to do relabeling?

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -197,87 +198,38 @@ The `data_source_name` is inherited from the parent block.
...
-* _`<GRAFANA_CLOUD_HOSTED_METRICS_URL>`_: The URL for your Grafana Cloud hosted metrics.
-* _`<GRAFANA_CLOUD_HOSTED_METRICS_ID>`_: The user ID for your Grafana Cloud hosted metrics.
-* _`<GRAFANA_CLOUD_RW_API_KEY>`_: Your Grafana Cloud API key.
-* _`<GRAFANA_CLOUD_HOSTED_LOGS_URL>`_: The URL for your Grafana Cloud hosted logs.
-* _`<GRAFANA_CLOUD_HOSTED_LOGS_ID>`_: The user ID for your Grafana Cloud hosted logs.
+* _`<LABEL>`_: The Alloy component label (for example, `prod_mysql`).
+* _`<DATA_SOURCE_NAME>`_: The MySQL connection [Data Source Name] (for example, `user:pass@tcp(mysql:3306)/`).
+* _`<AWS_RDS_ARN>`_: The ARN of your AWS RDS database (for example, `arn:aws:rds:us-east-1:123456789:db/prod-mysql`).
+
+For a complete end-to-end example that demonstrates sending database observability metrics to Grafana Cloud, refer to [Collect MySQL database metrics and logs](../../../../collect/mysql-database-metrics/).
```
ptodev, line 305: Should we make this more prominent? Maybe to have a (sub)section for it? For some components there could be more than one such link, so maybe we can just always list them as bullet points?

docs/sources/collect/mysql-database-metrics.md
clayton-cornell: Definitely. If we can identify the common "plumbing" components and document the concepts for them (vs the raw reference we have now) we'd have something to xref and it'd significantly reduce the middle content in any other task topic that used... loki.write, dicsovery.relabel and so on.

docs/sources/reference/components/database_observability/database_observability.mysql.md
clayton-cornell: Yeah, of some components, there might be 2 or 3 scenarios that cover common implementations... 

There's also a growing number of Learning Journeys and Learning Paths that could be used for comprehensive example links.

ptodev: WDYT about making the main examples section title "Examples" (always plural, never "Example"), and listing the links in bullet points right at the first paragraph, right under "Examples" and above any subsections with examples?

PR #6802, opened 2026-08-03
fix(validation): Typecheck attribute object

### LINKED ISSUE #6801: alloy validate rejects object literals decoded into structs: "expected object, got object"
### What's wrong?

`alloy validate` rejects an object literal whose target Go type is a struct, with the self-contradictory message `expected object, got object`. `alloy run` loads the exact same config without complaint.

The case I hit is `resource_filters` on `otelcol.exporter.googlecloud`, but the cause is generic: any attribute of the form `attr = [{...}]` whose element type is a struct rather than a map fails validation.

### Steps to reproduce

`config.alloy`:

```alloy
otelcol.exporter.googlecloud "t" {
	project = "p"
	log { resource_filters = [{regex = ".+"}] }
}
```

```console
$ alloy validate --stability.level=experimental --feature.community-components.enabled=true config.alloy
Error: config.alloy:3:28: expected object, got object

2 |     project = "p"
3 |     log { resource_filters = [{regex = ".+"}] }
  |                               ^^^^^^^^^^^^^^
4 | }
Error: validation failed

$ alloy run --stability.level=experimental --feature.community-components.enabled=true config.alloy
ts=... level=info msg="finished node evaluation" node_id=otelcol.exporter.googlecloud.t duration=414.583µs
ts=... level=info msg="finished complete graph evaluation"
```

`resource_filters = []` validates fine. Every object literal I tried fails identically (`{regex = ".+"}`, `{prefix = ""}`, `{prefix = "", regex = ".+"}`, `{"regex" = ".+"}`), and the block form is rejected with `"resource_filters" must be an attribute, but is used as a block`, so there is no spelling that satisfies both `validate` and `run`.

### Expected behaviour

`alloy validate` accepts any config that `alloy run` accepts.

### Cause

`validate` and `run` do not share a decode path. `internal/alloycli/cmd_validate.go` goes through `internal/validator`, which static-type-checks each block (`internal/validator/graph.go:96`, `typecheck.Block(node.block, reg.CloneArguments())`), whereas `run` decodes through `syntax/vm`.

In `syntax/typecheck/typecheck.go`, `typecheckObjectExpr` only accepts a map as the target of an object literal:

```go
// Check if the expected type is actually a map before trying to get its element type
if rv.Kind() != reflect.Map {
	expectedType := value.AlloyType(rv.Type())
	return diag.Diagnostics{{
		...
		Message: fmt.Sprintf("expected %s, got object", expectedType),
	}}
}
```

The target here is `[]ResourceFilter`, so `rv` is a struct. `value.AlloyType` maps `reflect.Struct` to `TypeObject` (`syntax/internal/value/type.go:129`), which renders as `"object"` — hence `expected object, got object`. Decoding an object literal into a struct is something the runtime supports, so the check looks too strict rather than the config being wrong.

The same function already carries a `FIXME(kalleep)` about not type checking further, and `typecheck.Block` has a `FIXME(kallep)` about not yet considering `value.Defaulter` / `value.Unmarshaler` / `value.ConvertibleFromCapsule` / `encoding.TextUnmarshaler`, so this is presumably known territory.

Checked against `main` at f7e4cf8c85e90d736dd59643e4eac7532fd12f15.

### Environment

- Alloy v1.15.1 and v1.18.0, both reproduce (darwin/arm64 and linux/amd64 release builds)

### Impact

This blocks using `alloy validate` in CI. We render our Alloy configs from Helm templates and wanted to validate all of them pre-merge; 7 of 40 rendered configs fail on this alone while running healthily in production, so we currently have to special-case the error message.

kalleep:
### Pull Request Details
Before it was assumed that all objects was either a capsule or a map, so it missed the case there a struct is marked as a attribute e.g:
```
struct = {
  str = "1"
  int = 1
}
```

We need to do a similar check as we do for blocks here, checking all sub expressions and check that required attributes are set for the struct.

To check this we need to perform similar checks as we do for block

Fixes: #6801

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/syntax/typecheck/typecheck_test.go
+++ b/syntax/typecheck/typecheck_test.go
@@ -386,14 +386,16 @@ func TestObjectAttr(t *testing.T) {
 	}
 
 	type Args struct {
-		Structs []Struct `alloy:"structs,attr"`
-		Struct  Struct   `alloy:"struct,attr"`
+		Structs   []Struct `alloy:"structs,attr"`
+		Struct    Struct   `alloy:"struct,attr"`
+		StructPtr *Struct  `alloy:"struct_ptr,attr,optional"`
 	}
 
 	src := []byte(`
 		test "name" {
 			struct =  {str = "test"}
 			structs = [{str= "test"}]	
+			struct_ptr =  {str = "test"}
 		}
 	`)
 
--- a/syntax/typecheck/typecheck_test.go
+++ b/syntax/typecheck/typecheck_test.go
@@ -380,6 +380,29 @@ func TestBlockMap(t *testing.T) {
 	require.Len(t, diag, 0)
 }
 
+func TestObjectAttr(t *testing.T) {
+	type Struct struct {
+		Str string `alloy:"str,attr"`
+	}
+
+	type Args struct {
+		Structs []Struct `alloy:"structs,attr"`
+		Struct  Struct   `alloy:"struct,attr"`
+	}
+
+	src := []byte(`
+		test "name" {
+			struct =  {str = "test"}
+			structs = [{str= "test"}]	
+		}
+	`)
+
+	file, err := parser.ParseFile("", src)
+	require.NoError(t, err)
+	diag := Block(file.Body[0].(*ast.BlockStmt), &Args{})
+	require.Len(t, diag, 0, diag)
+}
+
 type ValidatedString string
 
 func (s *ValidatedString) UnmarshalText(text []byte) error {
--- a/syntax/typecheck/typecheck.go
+++ b/syntax/typecheck/typecheck.go
@@ -314,6 +314,7 @@ func typecheckObjectExpr(expr *ast.ObjectExpr, rv reflect.Value) diag.Diagnostic
 		return nil
 	}
 
+	rv = reflectutil.DeferencePointer(rv)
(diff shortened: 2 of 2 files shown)


syntax/typecheck/typecheck_test.go
```diff
@@ -380,6 +380,31 @@ func TestBlockMap(t *testing.T) {
...
+	src := []byte(`
+		test "name" {
+			struct =  {str = "test"}
+			structs = [{str= "test"}]	
+			struct_ptr =  {str = "test"}
+		}
+	`)
+
+	file, err := parser.ParseFile("", src)
+	require.NoError(t, err)
```
ptodev, line 403: Do you want also a test that fails validation?

ptodev (approved):
Seems ok, but do you think in the future more code could be shared between the `run` and `validate` commands? The fact that they run different code paths could cause another such issues sooner or later.

kalleep:
>Seems ok, but do you think in the future more code could be shared between the run and validate commands? The fact that they run different code paths could cause another such issues sooner or later.

Yes there are most likely other issues too. The problem is that the code used by run have side effects (e.g. envs) and that is why validate cannot perform all the checks done by run. But we should be able to share more code

PR #6808, opened 2026-08-03
chore(loki.relabel): Implement consumer interface

kalleep:
Implement `loki.Consumer` for `loki.relabel` component.

### Pull Request Details
Instead of doing one big pr to loki pipelines I started splitting the changes up in several prs and only at the end we will change at the pipeline level. 

So in this pr I implemented `Consume` and `ConsumeEntry` functions without forwarding, since we have not made the changes at pipeline level yet. 

I fixed some existing issues:
1. We did not use read lock for relabel config
2. We never updated the cached `maxCacheSize` when that value changed.

`ConsumeEntry` is doing identical work like we do in the run loop but `Consume` will instead iterate on streams within a batch since all entries within a stream shares labels.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -53,14 +53,9 @@ func (b *Batch) add(labels model.LabelSet, entries ...push.Entry) {
 // entry is kept, if fn returns false the entry is dropped. Kept entries are
 // written back, and entries whose labels change are moved to a different stream.
 func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
-	type movedEntry struct {
-		labels model.LabelSet
-		entry  push.Entry
-	}
-
 	var (
 		newLen int
-		moves  []movedEntry
+		moves  []Entry
 	)
 
 	// Process each entry and compact each stream in place.
@@ -69,6 +64,9 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
 	// so we do not mutate the stream set while iterating it.
 	for i := range b.streams {
 		var (
+			// dst is where the next kept entry is written. It only moves forward
+			// when an entry is kept, so it never gets ahead of the entry being read
+			// and we only write to slots the loop has already read.
 			dst    = 0
 			stream = &b.streams[i]
 		)
@@ -84,10 +82,7 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
 			}
 
 			if !stream.Labels.Equal(entry.Labels) {
-				moves = append(moves, movedEntry{
-					labels: entry.Labels,
-					entry:  entry.Entry,
-				})
+				moves = append(moves, entry)
 				newLen++
 
 				continue
@@ -98,12 +93,14 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
 			newLen++
 		}
 
+		// Entries from dst onwards were either dropped or saved in moves, so it is
+		// safe to cut them here and let add reuse that spare capacity.
 		stream.Entries = stream.Entries[:dst]
 	}
 
 	// Reinsert entries whose labels changed into their destination streams.
 	for _, moved := range moves {
-		b.add(moved.labels, moved.entry)
+		b.add(moved.Labels, moved.Entry)
 	}
 	b.entryLen = newLen
 
@@ -125,10 +122,17 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
 func (b *Batch) FilterMapStreams(fn func(stream *Stream) (keep bool)) {
 	var (
 		newLen int
-		dst    int
-		moves  []Stream
+		// dst is where the next kept stream is written. It only moves forward when
+		// a stream is kept, so it never gets ahead of i and we only write to slots
+		// the loop has already read.
+		dst   int
+		moves []Stream
 	)
 
+	// Process each stream and compact the stream slice in place. Dropped streams
+	// are skipped, and streams whose labels changed are deferred into moves so we
+	// do not mutate the stream set while iterating it. They are reinserted below
+	// and may merge into an existing stream.
 	for i := range b.streams {
 		stream := Stream{
 			// FIXME(kalleep): Once the new batched pipeline owns this path, consider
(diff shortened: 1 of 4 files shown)


kalleep:
Ah need to rework this a bit and use `InterceptorConsumer`.

internal/component/common/loki/batch.go
```diff
@@ -119,6 +119,50 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
...
+		}
+
+		keep := fn(&stream)
+		if !keep {
+			continue
+		}
+
+		newLen += len(stream.Entries)
+
+		if !b.streams[i].Labels.Equal(stream.Labels) {
```
ptodev, line 151: Isn't checking label equality so often resource intensive?

internal/component/common/loki/batch.go
```diff
@@ -119,6 +119,50 @@ func (b *Batch) FilterMap(fn func(entry *Entry) (keep bool)) {
...
+			moves = append(moves, stream)
+			continue
+		}
+
+		b.streams[dst] = stream
+		dst++
+	}
+
+	b.streams = b.streams[:dst]
```
ptodev, line 162: It might be good to write a few comments. Maybe I'm slow but it took me a few min to figure out why this won't clobber `b.streams`.

internal/component/loki/relabel/relabel.go
```diff
@@ -209,60 +279,65 @@ func (c *Component) relabel(e loki.Entry) (loki.Entry, bool) {
...
-	return e, true
+	return relabeled, true
 }
 
-func (c *Component) process(e loki.Entry) model.LabelSet {
-	c.builder.Reset()
-	for k, v := range e.Labels {
-		c.builder.Add(string(k), string(v))
+func (c *Component) process(lset model.LabelSet) model.LabelSet {
+	br := labels.NewBuilder(labels.EmptyLabels())
```
ptodev, line 315: Is it ok not to reuse the builder? Can it lead to performance problems?

internal/component/loki/relabel/relabel.go
```diff
@@ -119,23 +194,31 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
-			componentID,
+			livedebugging.ComponentID(c.opts.ID),
 			livedebugging.LokiLog,
 			count,
 			func() string {
 				if !ok {
 					return fmt.Sprintf("entry: %s, labels: %s => <dropped>", entry.Line, entry.Labels.String())
 				}
-				return fmt.Sprintf("entry: %s, labels: %s => %s", entry.Line, entry.Labels.String(), relabeled.Labels.String())
+				return fmt.Sprintf("entry: %s, labels: %s => %s", entry.Line, entry.Labels.String(), relabeled.String())
```
ptodev, line 221: Could there be race conditions between live debugging and the consume hooks?

internal/component/loki/relabel/relabel.go
```diff
@@ -197,10 +269,8 @@ type cacheItem struct {
 // between model.LabelSet (map) and labels.Labels (slice). Promtail does
 // not have this issue as relabel config rules are only applied to targets.
 // Do we want to use labels.Labels in loki.Entry instead?
-func (c *Component) relabel(e loki.Entry) (loki.Entry, bool) {
-	c.metrics.entriesProcessed.Inc()
-
-	hash := e.Labels.Fingerprint()
+func (c *Component) relabel(lset model.LabelSet) (model.LabelSet, bool) {
+	hash := lset.Fingerprint()
```
ptodev: Can we use `FastFingerprint` instead? Hash collisions are in any case handled below.

ptodev (commented):
Mostly nitpicks. The only thing that could be a genuine concern is the comment on live debugging.

internal/component/loki/relabel/relabel.go
kalleep: Yes we for sure can.

internal/component/loki/relabel/relabel.go
kalleep: No, this does not run in a background go routine. LiveDebugging works this way to avoid unnessisary allocations (formatting the string) if it's not enabled or no one is listening. So no issue with concurrency here

internal/component/common/loki/batch.go
kalleep: Sure I can do that. I just implemented the same pattern I did for `FilterMap`.

internal/component/loki/relabel/relabel.go
kalleep: Sure we could use a pool of builders. TBH this should be better [than what we had before](https://github.com/grafana/alloy/blob/b168d0ec5e5d2db665c5495d01fc6b12f97d2997/internal/component/loki/relabel/relabel.go#L247-L255).

Before this pr we did re-use `ScratchBuilder` but we always allocated new labels from it and then created a builder anyway.

internal/component/common/loki/batch.go
kalleep: Yes it could be and we would then have the same problem when iterating through entries with `FilterMap` too. 

At some point I was considering that the caller should notify that it changed labels but that seems really error prone. I will investigate this a bit but would like to do that outside the scope of this pr if that is ok.

kalleep: Another thing I want do do is to no longer use `model.LabelSet` but instead use `labels.Labels` throughout loki components. That is a much bigger change and I want to do that as a entirely different effort, since we use stringlabels the equality check should be much more trivial. But still before that is done I will investigate if this can become a real problem.

dehaansa (dismissed):

internal/component/common/loki/batch.go
kalleep: Added in [6c5621c](https://github.com/grafana/alloy/pull/6808/commits/6c5621c55a892ff1e5adb88505ded09787716b61)

internal/component/loki/relabel/relabel.go
kalleep: Fixed in [f5e61af](https://github.com/grafana/alloy/pull/6808/commits/f5e61afa0150edda85b3855d06c10c1d58ed3933)

kalleep: Reverted to use Fingerprint again. Need to recompute values for test to work. Since it used Fingerprint before we should do this change in another pr.

internal/component/common/loki/batch.go
kalleep: Added a proposal here https://github.com/grafana/alloy/issues/6835

ptodev (approved):

Issue #6421, opened 2026-06-04
Implement First Experimental Solution + Tests


blewis12:
### Add a description

A PoC already exists [here](https://github.com/grafana/alloy/blob/embedded-supervisor-poc/collector/supervisor_command.go) - we should follow whatever is the outcome of [this issue](https://github.com/grafana/alloy/issues/6416)

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

devzl:
@blewis12 @x1unix I assume this is already done in #6815 and should be closed right?

PR #6815, opened 2026-08-03
feat: Add opamp supervisor

x1unix:

Add embedded opamp supervisor via `otel-supervisor` subcommand.

### Pull Request Details

The embedded supervisor is responsible for retrieving configuration via opamp protocol and managing the otel engine subprocess.

The supervisor runs 2 modes:

#### Standard mode

Standard mode requires user to supply a full supervisor yaml configuration via `--config` flag.

Config should contain opamp settings and any environment variables that might be required.

NOTE: The `agent.executable` has no effect as a path to collector is set by supervisor.

<details>
<summary>Example config provided by Fleet Management</summary>

```yaml
server:
  endpoint: "https://fleet-management-prod-014.grafana.net/v1/opamp"
  headers:
    Authorization: "Basic <auth string>"

capabilities:
  reports_effective_config: true
  accepts_remote_config: true
  reports_remote_config: true

agent:
  executable: ../../bin/otelcontribcol_linux_amd64
  description:
    identifying_attributes:
      service.name: "otelcol"
  args: [--feature-gates, service.AllowNoPipelines]
  env:
    GCLOUD_FM_URL: "https://fleet-management-prod-014.grafana.net"
    GCLOUD_BASIC_AUTH_BASE64: "<your token>"

storage:
  directory: .
```
</details>

#### Simple mode

In simple mode, supervisor config is generated in runtime with prefilled Opamp auth block, capabilities, path to a collector and other fields.

Supervisor runs in a simple mode when `--config` flag is unset and all required environment variables are defined:

| Variable | Description |
| - | - |
| `GCLOUD_FM_URL` | OpAmp endpoint. The `/v1/opamp` suffix is appended automatically if needed. |
| `GCLOUD_INSTANCE_ID` | Grafana Cloud instance (stack) ID. |
| `GCLOUD_RW_API_KEY` | Grafana Cloud API token. |
| `STORAGE_DIR` | Path to a directory to store supervisor data. |

In addition to that, the supervisor sets `GCLOUD_BASIC_AUTH_BASE64` and `OTLP_TOKEN` environment variables as some FM configuration pipelines might require them. 

#6421 

### Notes to the Reviewer

Documentation will be added in a separate PR, therefore the `otel-supervisor` subcommand is hidden.

### PR Checklist

- [x] Tests updated


collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+		return fmt.Errorf("failed to build supervisor logger: %w", err)
+	}
+	defer func() {
+		_ = logger.Sync()
+	}()
+
+	if cfg.Agent.Executable != "" && cfg.Agent.Executable != exe {
+		logger.Sugar().Warnf("warning: ignoring agent.executable %q from supervisor config; forcing the running Alloy binary %q\n", cfg.Agent.Executable, exe)
+	}
+	cfg.Agent.Executable = exe
```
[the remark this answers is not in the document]
x1unix: False-positive. Path to collector is expected to be overridden by default.

packaging/docker/otel-supervisor.cmd
```diff
@@ -0,0 +1,5 @@
+@echo off
+REM Wrapper to run Alloy's embedded OpAMP supervisor as PID 1.
+REM Delegate to Alloy's otel-supervisor subcommand and pass through all arguments.
+REM Using exec makes the supervisor PID 1 so it receives SIGTERM directly
```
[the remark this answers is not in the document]
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/9a187210735c242edd1a9a0721d2bf8de5149dc2

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+
+	token := strings.TrimSpace(os.Getenv(envAPIToken))
+	if token == "" {
+		missing = append(missing, envAPIToken)
+	}
+
+	storageDir := os.Getenv(envStorageDir)
+	if storageDir == "" {
+		missing = append(missing, envStorageDir)
+	}
```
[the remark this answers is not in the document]
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/9a187210735c242edd1a9a0721d2bf8de5149dc2

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+
+	resolverSettings := confmap.ResolverSettings{
+		DefaultScheme: "env",
+		URIs:          []string{cfgPath},
+		ProviderFactories: []confmap.ProviderFactory{
+			fileprovider.NewFactory(),
+			envprovider.NewFactory(),
+		},
+	}
```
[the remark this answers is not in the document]
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/468cfd6174c9a8144ba77c9c99dd5402e628eebc

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (err *missingEnvVarsError) Error() string {
+	return fmt.Sprintf("missing environment variables: %s", strings.Join(err.missing, ", "))
+}
+
+const opAmpEndpoint = "/v1/opamp"
+
+type authCredentials struct {
+	basicAuthCredentials string
```
jharvey10: What are your thoughts about naming this something like `basicAuthBase64` so it's clear that its intention is to only ever hold base64 encoded creds?

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+
+	if creds != nil {
+		// Expose auth credentials to the supervised collector process.
+		// The FM collector health config uses this env var in its Basic auth header.
+		if err := os.Setenv(envBasicAuthBase64, creds.basicAuthCredentials); err != nil {
+			return fmt.Errorf("cannot set %s environment variable: %w", envBasicAuthBase64, err)
+		}
+
+		// Fallback env var for "ping_otlp" configuration.
+		if err := os.Setenv("OTLP_TOKEN", creds.apiToken); err != nil {
```
jharvey10: There's some constants up on line 24 that seem like could contain this string as well. This would match the pattern established in the block above for setting the envvars for basic auth.

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+	sup, err := supervisor.NewSupervisor(runCtx, logger.Named("supervisor"), *cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create supervisor: %w", err)
+	}
+
+	if err := sup.Start(runCtx); err != nil {
+		return fmt.Errorf("failed to start supervisor: %w", err)
+	}
+
+	interrupt := make(chan os.Signal, 2)
```
jharvey10, line 115: AI indicated that upstream is using 1 here, but is also only using `os.Interrupt`. Would it be worth adding a comment here for the reason for the deviation from upstream and going with 2 and two signals?

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+	cfg.Server.Endpoint = fmURL
+	authStr := base64.StdEncoding.EncodeToString([]byte(instanceID + ":" + token))
+	cfg.Server.Headers = http.Header{
+		"Authorization": []string{"Basic " + authStr},
+	}
+	cfg.Capabilities.AcceptsRemoteConfig = true
+	cfg.Capabilities.ReportsRemoteConfig = true
+	cfg.Agent.Arguments = []string{"otel"}
+	cfg.Agent.PassthroughLogs = true
+	cfg.Storage.Directory = storageDir
```
jharvey10, line 220: I'm not super familiar with this, but is there a default value for the storage dir that can be obtained from the default supervisor to use if `STORAGE_DIR` is omitted? Just trying to think of if there's any way to reduce the number of required vars in some cases.

packaging/docker/otel-supervisor.cmd
```diff
@@ -0,0 +1,4 @@
+@echo off
+REM Wrapper to run Alloy's embedded OpAMP supervisor as PID 1.
+REM Delegate to Alloy's otel-supervisor subcommand and pass through all arguments.
+"C:\Program Files\GrafanaLabs\Alloy\alloy.exe" otel-supervisor %*
```
jharvey10, line 3: From AI:

The POSIX script uses `exec` so the supervisor becomes PID 1 and receives `SIGTERM` directly, per its comment. `.cmd` files can't replace the process image the same way, so this spawns `alloy.exe` as a child of `cmd.exe`. Do we know whether `docker stop` on Windows containers reaches `alloy.exe` promptly through that extra hop, or could it end up waiting on `cmd.exe` first? The two wrapper comments claim the same PID-1 behavior, but the underlying mechanism differs — might be worth a quick manual check if that hasn't happened already.

collector/builder-config.yaml
```diff
@@ -100,11 +99,22 @@ providers:
...
+  - github.com/knadh/koanf v1.5.0
+
 replaces:
   # Local modules.
   - github.com/grafana/alloy => ../
   - github.com/grafana/alloy/syntax => ../syntax
   # <BEGIN_SHARED_REPLACE_DIRECTIVES>
+  # OTel supervisor that embedded in Alloy, imports cmd/opampsupervisor, but OCB does not list it as a component,
+  # so `go mod tidy` would otherwise resolve it to @latest and break the filelogreceiver family.
+  # Pin it to the v0.153 contrib train
```
jharvey10: Wording suggestion:

```suggestion
  # The OTel supervisor that's embedded in Alloy imports cmd/opampsupervisor, but OCB does not list it as a component.
  # This would cause `go mod tidy` to resolve it to @latest and break the filelogreceiver family.
  # Pin it to the v0.153 contrib train to prevent this.
```

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+
+type missingEnvVarsError struct {
+	missing []string
+}
+
+func (err *missingEnvVarsError) Error() string {
+	return fmt.Sprintf("missing environment variables: %s", strings.Join(err.missing, ", "))
+}
+
+const opAmpEndpoint = "/v1/opamp"
```
jharvey10: Does this make sense to move to the top with other constants?

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+			return runSupervisor(configPath)
+		},
+	}
+
+	cmd.Flags().StringVar(&configPath, "config", "", "path to a configuration file. If omitted, uses environment-based simple mode")
+
+	return cmd
+}
+
+func runSupervisor(cfgPath string) error {
```
jharvey10, line 79: This function (rightfully so) brings together a lot of pieces and executes them in a controlled sequence, but it's hard to parse where one step ends and another starts. One potential improvement could be to break the high-level steps into purpose-named functions to slim it down and increase readability while keeping the same intent. Perhaps something like:

```go
func runSupervisor(cfgPath string) error {
	exe, err := os.Executable()
	if err != nil {
		return fmt.Errorf("unable to get executable path: %w", err)
	}

	cfg, creds, err := resolveSupervisorConfig(cfgPath)
	if err != nil {
		return err
	}

	if err := exportCredentialEnv(creds); err != nil {
		return err
	}

	logger, err := supervisortelemetry.NewLogger(cfg.Telemetry.Logs)
	if err != nil {
		return fmt.Errorf("failed to build supervisor logger: %w", err)
	}
	defer func() { _ = logger.Sync() }()

	overrideAgentExecutable(cfg, exe, logger)

	return startAndWaitForShutdown(cfg, logger)
}
```

collector/supervisor.go
x1unix: @jharvey10 there is a default one but I made that required as the default one is linux-specific otel path.

packaging/docker/otel-supervisor.cmd
x1unix: @jharvey10 true, I also saw that note from reviewer clanker, but I probably forgot to commit the fix. Thanks :)

collector/builder-config.yaml
```diff
@@ -100,11 +99,22 @@ providers:
...
+  - github.com/knadh/koanf v1.5.0
+
 replaces:
   # Local modules.
   - github.com/grafana/alloy => ../
   - github.com/grafana/alloy/syntax => ../syntax
   # <BEGIN_SHARED_REPLACE_DIRECTIVES>
+  # The OTel supervisor that's embedded in Alloy imports cmd/opampsupervisor, but OCB does not list it as a component.
+  # This would cause `go mod tidy` to resolve it to @latest and break the filelogreceiver family.
+  # Pin it to the v0.153 contrib train to prevent this.
```
blewis12:
```suggestion
  # The OTel supervisor that's embedded in Alloy imports cmd/opampsupervisor, but OCB does not list it as a component.
  # This would cause `go mod tidy` to resolve it to @latest
  # Pin it to current contrib train version to prevent this.
```

just to make it less specific

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	if creds != nil {
+		// Expose auth credentials to the supervised collector process.
+		// The FM collector health config uses this env var in its Basic auth header.
+		if err := os.Setenv(envBasicAuthBase64, creds.basicAuthCredentials); err != nil {
+			return fmt.Errorf("cannot set %s environment variable: %w", envBasicAuthBase64, err)
+		}
+
+		// Fallback env var for "ping_otlp" configuration.
```
blewis12: I don't think need to consider `OTLP_TOKEN` here, this was only used in pre-GA OpAmp code snippets - `ping_otlp` isn't included in the GA version of FM (discussed [here](https://raintank-corp.slack.com/archives/C07EWQY7USE/p1785767440414149?thread_ts=1785765333.471209&cid=C07EWQY7USE))

collector/supervisor.go
```diff
@@ -0,0 +1,257 @@
...
+		"Authorization": []string{"Basic " + authStr},
+	}
+	cfg.Capabilities.AcceptsRemoteConfig = true
+	cfg.Capabilities.ReportsRemoteConfig = true
+	cfg.Agent.Arguments = []string{"otel"}
+	cfg.Agent.PassthroughLogs = true
+	cfg.Storage.Directory = storageDir
+
+	if err := cfg.Validate(); err != nil {
+		return nil, nil, fmt.Errorf("cannot validate generated supervisor config: %w", err)
```
blewis12: Is this message confusing? the `err` isn't that it can't validate but that validation failed

collector/supervisor.go
x1unix: @jharvey10 `os.Interrupt` is `syscall.SIGINT` which is different from `SIGTERM`.

The `Ctrl+C` hotkey in terminal produces `SIGINT`, but afaik container runtimes like docker or k8s are using `SIGTERM`.

Therefore the code handles both cases.

collector/supervisor.go
x1unix: @jharvey10 I've removed that env var as it was apparently deprecated (see: https://github.com/grafana/alloy/pull/6815#discussion_r3766139576 )

collector/supervisor.go
x1unix: @jharvey10 I usually prefer to keep longer functions to avoid jumping between function definitions, but this is a matter of taste.

I've decoupled supervisor and config construction into separate functions in https://github.com/grafana/alloy/pull/6815/changes/81a721c4af507e3d949ec2b93aa31e9a053dca5c

collector/supervisor.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/e8f2a4df40beedc08f968a1d014e100d22876d74

collector/supervisor.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/e8f2a4df40beedc08f968a1d014e100d22876d74

collector/supervisor.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/e8f2a4df40beedc08f968a1d014e100d22876d74

packaging/docker/otel-supervisor.cmd
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/9518237a7855d21a0467b97b5654a2d683e5d5b0

collector/supervisor.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/e8f2a4df40beedc08f968a1d014e100d22876d74

collector/supervisor.go
x1unix: Addressed in https://github.com/grafana/alloy/pull/6815/changes/e8f2a4df40beedc08f968a1d014e100d22876d74

collector/supervisor.go
jharvey10: Fair enough! I will say this current version reads super cleanly to me and was very straightforward to review, but I agree it's definitely a taste/preference thing.

jharvey10 (approved):

PR #6818, opened 2026-08-04
fix(loki.source.docker): Respect label changes

### LINKED ISSUE #6714: loki.source.docker: label changes silently ignored until restart

kalleep:
### Pull Request Details
Before https://github.com/grafana/alloy/pull/5026 any change to labels would restart a tailer and re-ingest logs. It was unintended to get rid of this behavior. While it is a problem that logs are re-ingested when any labels changes that should be handled in a seperate pr / issue, we already track it here https://github.com/grafana/alloy/issues/5135

So this pr will revert to how it worked before _and_ fix the issue when relabel config change, i.e. not being picked up. We can probably rework this later to not require a restart but I suspect that chaining e.g relabel config is a thing that don't happen too often.

Fixes: https://github.com/grafana/alloy/issues/6714

### Notes to the Reviewer

Added `ReconcileWithDedup`, that Reconcile will call internally. This functions allows caller to provide a dedup key that can be different from the key used to schedule sources. This way we can keep the existing behavior of not scheduling duplicated tailers by container id but still respect the key (container id + labels).

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -215,12 +215,15 @@ func (c *Component) Update(args component.Arguments) error {
 		return targets[i].fingerPrint < targets[j].fingerPrint
 	})
 
-	source.Reconcile(
+	source.ReconcileWithDedup(
 		c.opts.Logger,
 		c.scheduler,
 		slices.Values(targets),
 		func(target containerTarget) positions.Entry {
 			return positions.Entry{Path: string(target.labels[dockerLabelContainerID]), Labels: target.labels.Merge(defaultLabels).String()}
+		},
+		func(target containerTarget) string {
+			return string(target.labels[dockerLabelContainerID])
 		},
 		func(entry positions.Entry, target containerTarget) (source.Source[positions.Entry], error) {
 			if entry.Path == "" {
--- a/internal/component/loki/source/docker/docker.go
+++ b/internal/component/loki/source/docker/docker.go
@@ -171,6 +171,8 @@ func (c *Component) Run(ctx context.Context) error {
 }
 
 type containerTarget struct {
+	containerID string
+	// labels is the target's own labels merged with the component's default labels.
 	labels      model.LabelSet
 	fingerPrint model.Fingerprint
 }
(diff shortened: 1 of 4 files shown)


internal/component/loki/source/docker/docker.go
```diff
@@ -189,7 +191,10 @@ func (c *Component) Update(args component.Arguments) error {
 		return err
 	}
 
-	if client != c.client {
+	rcs := alloy_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelRules)
+
+	if requiresReset(client, c.client, rcs, c.rcs) {
```
ptodev, line 196: I don't know if this check is robust enough. I'm not sure if it even works at the moment. For example, if `refresh_interval`, is changed then it seems `getClient` will still return the old client because `reflect.DeepEqual(c.args.Host, args.Host)` still holds? This seems like an old problem, but I'm just saying it's easy to introduce a change in the config attributes which can break the logic on whether a tailer should be reset. If you think there's an issue with the comparison of the clients, then we could fix it in a subsequent PR.

ptodev (approved):
Thanks, LGTM!

internal/component/loki/source/docker/docker.go
kalleep: Yeah I don't think so either. Also not really sure this works correctly for relabel rules in all cases. Will do a follow up and fix this :+1:

PR #6804, opened 2026-08-03
chore(database_observability): Set logs emit interval for explain plans

cristiangreco:
### Brief description of Pull Request
Bring the same approach to `explain_plans` as we have for `schema_details` collectors (and `query_details` in sql_server), where logging is throttled to at most one emission per EmitInterval. Now every processing result (success, skipped, error) throttles uniformly, and a failure that emits nothing stays unthrottled.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated
- [ ] This pull request was substantially generated with AI assistance (see the [GenAI policy](https://github.com/grafana/alloy/blob/main/docs/developer/genai.md))

PR #6819, opened 2026-08-04
fix(database_observability.postgres): Use of defer in `explain_plans` collector

cristiangreco:
Fix `explain_plans` collector so that a cached query is finalised immediately instead of being deferred (which prevented correct update of processedCount and thus made batch-size limit ineffective).

### Pull Request Details

Followup from https://github.com/grafana/alloy/pull/6804#discussion_r3703333119

n.a.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/explain_plans.go
+++ b/internal/component/database_observability/postgres/collector/explain_plans.go
@@ -497,8 +497,6 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
 
 	processedCount := 0
 	for _, qi := range c.queryCache {
-		generatedAt := time.Now().Format(time.RFC3339)
-		nonRecoverableFailureOccurred := false
 		if c.isThrottled(qi.uniqueKey) {
 			c.markFinished(qi)
 			delete(c.queryCache, qi.uniqueKey)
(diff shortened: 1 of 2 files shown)


internal/component/database_observability/postgres/collector/explain_plans.go
```diff
@@ -507,126 +505,127 @@ func (c *ExplainPlans) fetchExplainPlans(ctx context.Context) error {
...
+		processedCount++
+	}
 
-		containsReservedWord, err := database_observability.ContainsReservedKeywords(qi.queryText, database_observability.ExplainReservedWordDenyList, sqllexer.DBMSPostgres)
+	return nil
+}
+
+// processExplainPlan processes a single query from the cache.
+// Returns true if the query encountered a non-recoverable failure and should be denylisted.
+func (c *ExplainPlans) processExplainPlan(ctx context.Context, qi *queryInfo) bool {
```
matthewnolf, line 523: Wouldn't this make more sense (semantically) to return as an error rather than a bool?

cristiangreco: Hmm unsure whether it's worth the change. The caller would have to log the error in place of the callee logging it, but in this function we almost always log an error before any type of return, whether true or false (the real difference being only whether the error was deemed recoverable or not).

We could add a sort of type wrapper for recoverable vs non-recoverable error, but feels an overkill.

wdyt?

matthewnolf: Okay that makes sense, let's leave as is.

matthewnolf (approved):

PR #6822, opened 2026-08-04
feat(otelcol.receiver.prometheus): Return stable refs

kgeckhart:

Sync `Append`/`AppendHistogram` in the copied `prometheusreceiver` transaction.go so it returns a stable ref rather than a hardcoded `1`. 

### Pull Request Details

This mirrors [opentelemetry-collector-contrib#46589](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46589) and the AppenderV2 rewrite ([opentelemetry-collector-contrib#46426](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46426)). The AppenderV2 rewrite is what had Append return a stable ref which is mirrored here. 

### PR Checklist

- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/receiver/prometheus/internal/transaction.go
+++ b/internal/component/otelcol/receiver/prometheus/internal/transaction.go
@@ -103,7 +103,7 @@ func newTransaction(
 	}
 }
 
-// Append always returns 0 to disable label caching.
+// Append returns a stable series reference so Prometheus can track series staleness.
 func (t *transaction) Append(_ storage.SeriesRef, ls labels.Labels, atMs int64, val float64) (storage.SeriesRef, error) {
 	t.addingNativeHistogram = false
 	t.addingNHCB = false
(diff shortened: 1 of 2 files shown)


internal/component/otelcol/receiver/prometheus/internal/transaction.go
```diff
@@ -179,6 +179,7 @@ func (t *transaction) Append(_ storage.SeriesRef, ls labels.Labels, atMs int64,
 	curMF := t.getOrCreateMetricFamily(*rKey, scope, metricName)
 
 	seriesRef := t.getSeriesRef(ls, curMF.mtype)
+	cacheRef := ls.Hash()
```
kalleep, line 182: I know this is synced from upstream but this really should be computed where we return since this would not be used if next call fails.

But maybe better to leave as is

kalleep (approved):
Nice, left a comment but feel free to ignore

internal/component/otelcol/receiver/prometheus/internal/transaction.go
kgeckhart: There were a few things I was looking at that seem worth a patch upstream. Ignoring the incoming series ref to create a new one and then returning a completely different value as the ref seems really odd.

Issue #5913, opened 2026-03-27
chore: update stage configs in loki.process to use syntax.Validator


kalleep:
### Component(s)

loki.process

### Add a description

# Move `loki.process` stage validation to `syntax.Validator`

## Summary

`loki.process` stage configuration validation is currently split across two places. Small subset of stage configs implement `syntax.Validator` butmany stage configs are only validated later when `stages.NewPipeline` constructs runtime stages during component update.

This issue is to move stage config validation to `syntax.Validator` so invalid configurations fail during config validation instead of during pipeline construction.

## Stages to migrate

This should cover all stages in `internal/component/loki/process/stages`:

- [x] `stage.cri`
- [x] `stage.decolorize`
- [x] `stage.docker`
- [ ] `stage.drop`
- [x] `stage.eventlogmessage`
- [ ] `stage.geoip`
- [ ] `stage.json`
- [ ] `stage.label_keep`
- [ ] `stage.label_drop`
- [ ] `stage.labels`
- [ ] `stage.limit`
- [ ] `stage.logfmt`
- [ ] `stage.luhn`
- [ ] `stage.match`
- [ ] `stage.metrics`
- [ ] `stage.multiline`
- [ ] `stage.output`
- [x] `stage.pack`
- [ ] `stage.pattern`
- [ ] `stage.regex`
- [ ] `stage.replace`
- [x] `stage.sampling`
- [ ] `stage.static_labels`
- [ ] `stage.structured_metadata`
- [ ] `stage.structured_metadata_drop`
- [ ] `stage.template`
- [ ] `stage.tenant`
- [ ] `stage.timestamp`
- [ ] `stage.truncate`
- [x] `stage.windowsevent`

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

michael-duren:
@kalleep, I'm working on `stage.static_labels` to get a feel for this. Will pick up more of this once I can get something in.

PR #6823, opened 2026-08-05
feat(loki.process): Add stage.split_json to split top-level JSON arrays

### LINKED ISSUE #6816: loki.process: add a stage to split top-level JSON arrays
### Component(s)

loki.process, loki.source.azure_event_hubs

### Background

Some producers send bare top-level JSON arrays. `loki.source.azure_event_hubs` forwards each as one line, or drops it when `disallow_custom_messages = true`.

The array's elements can't be filtered or timestamped independently. No stage splits such an array into entries.

### Proposal

Add `stage.split_json`, a stage that splits top-level JSON arrays. Each element's JSON text becomes the log line of its own entry, in order; each entry gets its own copy of the original labels, timestamp, structured metadata, and extracted map. An empty array emits nothing.

`[{"a":1},{"b":2}]` becomes lines `{"a":1}` and `{"b":2}`.

An optional `source` attribute follows the existing stage convention (`stage.json`, `stage.regex`, `stage.logfmt`): unset, the stage splits the log line; set, it splits the named value from the extracted map. If the configured source key is absent, or the selected input is not a JSON array, the entry passes through unchanged.

```alloy
// Array nested inside JSON: {"something":"a","my_array":[{"a":1},{"b":2}]}
stage.json {
  expressions = { my_array = "" }
}

stage.split_json {
  source = "my_array"
}
```

For Event Hubs, set `disallow_custom_messages = false` and forward the source to `loki.process.split.receiver`. Later stages can set element timestamps:

```alloy
loki.process "split" {

  stage.split_json {}

  stage.json {
    expressions = { ts = "time" }
  }
  stage.timestamp {
    source            = "ts"
    format            = "RFC3339"
    action_on_failure = "skip"
  }
}
```

`action_on_failure = "skip"` keeps the existing timestamp when `ts` is missing or invalid.

Options considered:

- A new stage (chosen): works with any source that forwards the array and keeps splitting separate from extraction. For Event Hubs, allowing custom messages also forwards unrelated payloads.
- Extending `stage.json`: keeps splitting and extraction in one block, but that stage only extracts into the extracted map and doesn't change the log record, so a fan-out mode there is a bigger behavior change than a dedicated stage.
- A source-side option on `loki.source.azure_event_hubs`: would split valid Azure records with `disallow_custom_messages = true` and reuse Azure's timestamp handling and category relabel input. It's Azure-only; nothing stops adding it later if the Event Hubs path still needs it.

I've only tested the Event Hubs source-side prototype; `stage.split_json` will be a fresh implementation. I'm happy to implement it once accepted.

DeeGeeGit:

Adds `stage.split_json`, a new `loki.process` stage that splits a top-level JSON array into one entry per element, per the proposal in #6816. By default the stage parses the log line; the optional `source` argument selects a value from the extracted map instead, following the `stage.json`/`stage.regex` convention. Each element's JSON text becomes its own entry's log line — the element's raw JSON value bytes, in array order — each with its own labels, timestamp, structured metadata, and extracted map. An empty array emits nothing; malformed or non-array input passes through unchanged.

### Pull Request Details

- Non-final child entries are cloned and sent one at a time, so channel backpressure applies per child rather than to a materialized batch; the final child reuses the original entry's label, metadata, and extracted-map allocations, which the stage never touches again. The parsed array's raw elements are held in memory while children are being sent.
- I've left v1 without an amplification cap or metrics (parser-stage precedent): the stage is opt-in, but a hostile line can still fan out CPU, transient memory, and output volume — flagging that trade-off for reviewers.
- Elements are parsed with stdlib `encoding/json` rather than the package's usual jsoniter: jsoniter's `RawMessage` decodes a `null` element to an empty line and mis-handles some numeric literals.

Fixes #6816

### Notes to the Reviewer

"Config converters updated" is unchecked because Promtail has no equivalent stage to convert from. Local verification ran `gofmt`, `golangci-lint`, `alloylint`, and `go test -race -tags="gore2regex nodocker"` over `./internal/component/loki/process/...`.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// Config Errors
+const (
+	ErrEmptySplitJSONStageSource = "empty split_json stage source"
+)
+
+// SplitJSONConfig represents a split_json stage configuration.
+type SplitJSONConfig struct {
+	Source *string `alloy:"source,attr,optional"`
```
kalleep: We don't have this for other stages that have Source yet but we want to remove as much "runtime" errors as possible and move to syntax parsing error.

I would suggest adding a new type to type.go 

```go
var (
	_ encoding.TextMarshaler   = Source("")
	_ encoding.TextUnmarshaler = (*Source)(nil)
)

type Source string

// UnmarshalText implements encoding.TextUnmarshaler.
func (s *Source) UnmarshalText(text []byte) error {
	if len(text) == 0 {
		return errors.New("source cannot be empty")
	}
	*s = Source(text)
	return nil
}

// MarshalText implements encoding.TextMarshaler.
func (s Source) MarshalText() (text []byte, err error) {
	return []byte(s), nil
}
```
And use a pointer to that instead of string. That way we cannot create this stage from config with empty source set so we don't have to return an error from this stage

internal/component/loki/process/stages/split_json_test.go
```diff
@@ -0,0 +1,366 @@
...
+	return Entry{
+		Extracted: extracted,
+		Entry: loki.NewEntryWithCreatedUnixMicro(
+			model.LabelSet{"app": "test"},
+			splitJSONTestCreated,
+			push.Entry{
+				Timestamp:          splitJSONTestTime,
+				Line:               line,
+				StructuredMetadata: push.LabelsAdapter{{Name: "trace_id", Value: "123"}},
+				Parsed:             push.LabelsAdapter{{Name: "level", Value: "info"}},
```
kalleep: We don't need to set / check Parsed. This property is internal to loki codebase and not used in alloy

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			for _, raw := range elems {
+				child := e               // struct copy keeps the timestamp and the private created field
+				child.Line = string(raw) // element bytes verbatim, no re-encode
+				child.Labels = e.Labels.Clone()
+				child.Extracted = maps.Clone(e.Extracted)
+				child.StructuredMetadata = slices.Clone(e.StructuredMetadata)
+				child.Parsed = slices.Clone(e.Parsed)
+				out <- child
+			}
```
kalleep:
```suggestion
			for i, raw := range elems {
				var child Entry
				// We don't need to clone entry for the last element.
				if i == len(elems)-1 {
					child = e
				} else {
					child.Entry = e.Entry.Clone()
					child.Extracted = maps.Clone(e.Extracted)
				}

				// #nosec G103 nosemgrep: use-of-unsafe-block
				// Safety: json.Unmarshal allocates a fresh byte slice per RawMessage
				// element rather than aliasing the input, and elems is never mutated after
				// this point, so the string's bytes stay immutable.
				child.Line = unsafe.String(unsafe.SliceData(raw), len(raw))
				out <- child
			}
```

A couple of things we can optimize here:
1. We don't need to clone the last element since it can re-use already allocated labels, structured_metadata and extracted map.
2. Calling string(raw) will allocate a new string, json.RawMessage is just a byte slice that is allocated during Unmarshal so it's safe to just convert it into a immutable string
3. Use the clone function from loki.Entry

kalleep (commented):
Nice work. Left a couple of comment on how we can optimize this a bit

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,115 @@
...
+
+// splitJSONStage splits a top-level JSON array into one entry per element.
+type splitJSONStage struct {
+	cfg    SplitJSONConfig
+	logger *slog.Logger
+}
+
+// newSplitJSONStage creates a new split_json pipeline stage from a config.
+// An Alloy-decoded config cannot carry an empty source (Source rejects it at
+// decode time), so construction cannot fail.
```
kalleep: I don't think this comment is that usefull

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,115 @@
...
+	go func() {
+		defer close(out)
+		for e := range in {
+			elems, ok := s.split(e)
+			if !ok {
+				out <- e
+				continue
+			}
+			for i, raw := range elems {
+				child := e // struct copy keeps scalar fields, incl. the private created field
```
kalleep: Same as above, not a useful comment

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done — added `Source` to types.go per your snippet and switched the config to `*Source`, so an empty source now fails at configuration decode time and the constructor no longer returns an error. The empty-source test moved to a decode-level assertion.

internal/component/loki/process/stages/split_json_test.go
DeeGeeGit: Done — dropped the `Parsed` handling (no Alloy production code reads the field directly, and `loki.process` ingress clears it via `Entry.Clone()` before the pipeline).

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done — removed.

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done — removed.

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done — applied the suggestion: the final child reuses the original entry, non-final children go through `loki.Entry.Clone()` + `maps.Clone`, and the line conversion uses `unsafe.String` (verified stdlib `RawMessage` copies each element into its own backing slice). Two small notes: the suppression ended up split across lines — semgrep's preceding-line matcher needs `nosemgrep` first, so the combined `#nosec … nosemgrep` form failed the scan — and I kept `child := e` as the starting copy, overwriting `Entry`/`Extracted` for non-final children, so any future field on `stages.Entry` survives the copy (behavior identical today).

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,113 @@
...
+	// Skip only the four JSON whitespace bytes (RFC 8259) — not
+	// strings.TrimSpace, which would also strip non-JSON Unicode whitespace —
+	// and pass through anything that is not a top-level array.
+	i := 0
+	for i < len(input) && (input[i] == ' ' || input[i] == '\t' || input[i] == '\r' || input[i] == '\n') {
+		i++
+	}
+	if i == len(input) || input[i] != '[' {
+		return nil, false
+	}
```
thampiotr: Instead of writing code like this, can we use idiomatic options like:
```
!strings.HasPrefix(strings.TrimLeft(input, " \t\r\n"), "[")
```
?

I would expect there must be a cleaner code solution for this.

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,113 @@
...
+				child := e
+				if i < len(elems)-1 {
+					child.Entry = e.Entry.Clone()
+					child.Extracted = maps.Clone(e.Extracted)
+				}
+				// Safety: json.RawMessage's UnmarshalJSON contract copies each
+				// value into its own backing slice, which is never exposed or
+				// mutated after this point, so the string's bytes stay immutable.
+				// nosemgrep: use-of-unsafe-block
+				child.Line = unsafe.String(unsafe.SliceData(raw), len(raw)) // #nosec G103
```
thampiotr, line 54: I wonder if we should run some benchmarks that can tell us if it is worth complicating the code here for the gain of skipping one allocation @kalleep what do you think?

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,113 @@
...
+		cfg:    cfg,
+		logger: logger.With("stage", "split_json"),
+	}
+}
+
+// Run implements Stage. Non-final children are cloned and sent one at a time
+// rather than materialized as a batch, so channel backpressure applies per
+// child; the final child reuses the original entry's allocations, which the
+// stage never touches again.
+func (s *splitJSONStage) Run(in chan Entry) chan Entry {
```
thampiotr, line 34: There is `RunWithSkipOrSendMany` in `pipeline.go` that could probably be used instead. We should reuse the framework code where possible

internal/component/loki/process/stages/types.go
```diff
@@ -2,9 +2,35 @@ package stages
...
 import (
 	"encoding"
+	"errors"
 	"fmt"
 )
 
+// Source names a value in the extracted map for stages that read their input
+// from it. Unlike SourceType below (which selects the kind of location a
+// stage reads from), a Source is the non-empty key itself; rejecting the
+// empty string at decode time keeps misconfigurations out of the pipeline.
```
thampiotr, line 12: A good follow-up would be to use this more broadly in other stages.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1534,6 +1536,84 @@ stage.sampling {
...
+By default, this is the log line itself, but it can also be a previously extracted value.
+{{< param "PRODUCT_NAME" >}} rejects an empty `source` string when it loads the configuration.
+
+When the input is a top-level JSON array, the stage replaces the log entry with one entry for each array element, in array order.
+Each new log line is the raw text of its array element, exactly as it appears in the input.
+The stage doesn't re-encode the element.
+The stage also doesn't split nested arrays: an element that's itself an array becomes the log line of a single entry.
+Each new entry keeps the timestamp of the original entry, and the emitted entries have independent labels, top-level extracted maps, and structured metadata, so later stages can modify one entry without affecting the others.
+
+An empty array produces no entries: the original entry is dropped and the stage emits nothing.
```
```suggestion
An empty array produces no entries.
The stage drops the original entry and emits nothing.
```

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1534,6 +1536,84 @@ stage.sampling {
...
+
+In all other cases, the log entry passes through the stage unchanged.
+This happens when the input isn't valid JSON, or is valid JSON that's not an array, such as an object, a string, or a number.
+When `source` is set, this also happens when the extracted value is missing or can't be converted to a string.
+The stage never falls back to the log line when `source` is set.
```
```suggestion
When you set `source`, the log entry also passes through unchanged when the extracted value is missing or can't be converted to a string.
When you set `source`, the stage never falls back to the log line and only reads the extracted value named by `source`.
```

internal/component/loki/process/stages/types.go
kalleep: I created a issue a while back https://github.com/grafana/alloy/issues/5913. This is for moving all validation code to parsing errors instead of runtime validation. But yes we can follow up just with a pr to update other stages to use this one.

internal/component/loki/process/stages/split_json.go
kalleep: Benchmarks are always good! I don't think this is that complicated but it can be a bit _scary_ when you see unsafe usage. In this case I don't see why we would not do it since it would skip one allocation per item in the array we split.

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done.

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Done — BenchmarkSplitJSONStage, 10x1KiB and 100x120B shapes. `string(raw)` instead of `unsafe.String`: +1 alloc/element (+10/+100 allocs/op), +36%/+12.5% B/op, ~+5% ns/op — unsafe stays.

internal/component/loki/process/stages/split_json.go
DeeGeeGit: Benchmarked it: ns/op parity but +4%/+12% B/op at 10/100 elements from materializing the batch — keeping the loop; the comment on Run documents the trade-off.

docs/sources/reference/components/loki/loki.process.md
DeeGeeGit: Done.

docs/sources/reference/components/loki/loki.process.md
DeeGeeGit: Done.

internal/component/loki/process/stages/split_json.go
thampiotr: @kalleep could you chime in a second opinion? the motivation to use `RunWithSkipOrSendMany` instead of a brand new custom `Run` is to reuse more code for obvious reasons, not just the performance concern. I do see that it would add one alloc per unsplit line, but that's not that big.

internal/component/loki/process/stages/split_json.go
thampiotr: > Done — BenchmarkSplitJSONStage, 10x1KiB and 100x120B shapes. string(raw) instead of unsafe.String: +1 alloc/element (+10/+100 allocs/op), +36%/+12.5% B/op, ~+5% ns/op — unsafe stays.

I'll share my decoding of this, but it does include some guesses 😛 
- Tested string(raw) vs. the current unsafe.String
- Tested using two scenarios: 10 array elements of 1KiB each and 100 array elements of 120B each
- raw string is 1 extra alloc per element - so +10 allocs for 10 element array and +100 for 100 elements. This is +36% and +12.5% more allocs for raw string compared to unsafe
- speed is roughly the same for both

internal/component/loki/process/stages/split_json.go
```diff
@@ -0,0 +1,111 @@
...
+	trimmed := strings.TrimLeft(input, " \t\r\n")
+	if !strings.HasPrefix(trimmed, "[") {
+		return nil, false
+	}
+
+	// stdlib RawMessage keeps every element byte-for-byte, null included;
+	// jsoniter would decode a null element to an empty line. A failed
+	// unmarshal yields no children at all.
+	var elems []json.RawMessage
+	if err := json.Unmarshal([]byte(input), &elems); err != nil {
```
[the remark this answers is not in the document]
thampiotr: From my understanding, `utf8.ValidString` would be too pedantic here. We would force an error, where the library is handling it fine. RFC 8259 does require JSON text to be UTF-8, so ["\xff"] isn't strictly valid JSON — but Go's encoding/json doesn't enforce that

internal/component/loki/process/stages/split_json.go
thampiotr: I ran this change locally and conclusion is that it's not saving nearly as much code as I have hoped. And nothing too complex. So happy to ignore this.

thampiotr (approved):

kalleep (approved):

PR #6826, opened 2026-08-05
fix(loki.source.docker): Restart tailers when config changes

kalleep:
### Pull Request Details

Follow up to https://github.com/grafana/alloy/pull/6818#event-29005879673. Try to cover all cases where we need to restart log tailers.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/common/config/types.go
```diff
@@ -220,6 +260,15 @@ func (p *Headers) Validate() error {
 	return promHeaders.Validate()
 }
 
+// Equal reports whether p and other hold the same headers. A nil map and an
+// empty map are considered equal.
+func (p *Headers) Equal(other *Headers) bool {
```
ptodev: Is it really necessary to have these Equal functions?

kalleep: Hm maybe not, for some reason I thought that reflect.DeepEqual would not work but I will give it a try

kalleep: Okay so using `reflect.DeepEqual` worked. I made changes in [d840261](https://github.com/grafana/alloy/pull/6826/commits/d840261bfd45af24ed73039592e54041dda5ee43) and added some tests for this function.

ptodev (approved):

PR #6829, opened 2026-08-05
docs(loki.process): Clarify default log timestamp without stage.timestamp

### LINKED ISSUE #6286: Docs feedback: Loki.process stage.timestamp contains incorrect statement
### URL

https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagetimestamp

### Component(s)

loki.process

### Feedback

On this page it states directly under `stage.timestamp`:
```
When no timestamp stage is set, the log entry timestamp defaults to the time when the log entry was scraped.
```
Is this even true? In my testing Alloy refused to process logs that it thinks are in the future because of a timezone difference (these are syslogs, they do not contains what the timezone is). Not having a `stage.timestamp` entry does not prevent Alloy from throwing away those logs.

### Tip

locker95:
The old note said the timestamp always defaults to scrape time if you skip `stage.timestamp`. That's not quite right — the entry keeps whatever the upstream source already set, and Loki can still drop out-of-window timestamps either way.

Fixes #6286

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -1823,7 +1823,7 @@ stage.tenant {
 ### `stage.timestamp`
 
 The `stage.timestamp` inner block configures a processing stage that sets the timestamp of log entries before they're forwarded to the next component.
-When no timestamp stage is set, the log entry timestamp defaults to the time when the log entry was scraped.
+When no timestamp stage is set, `loki.process` keeps the timestamp already attached to the entry by the upstream component (for example a `loki.source.*` component). Many sources default that to the time the entry was received; others can use a timestamp embedded in the log when configured to do so. Leaving out `stage.timestamp` does not by itself rewrite or sanitize timestamps, and Loki may still reject entries that fall outside its configured acceptance window (for example far-future or out-of-order timestamps).
 
 The following arguments are supported:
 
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -1823,7 +1823,7 @@ stage.tenant {
 ### `stage.timestamp`
 
 The `stage.timestamp` inner block configures a processing stage that sets the timestamp of log entries before they're forwarded to the next component.
-When no timestamp stage is set, the log entry timestamp defaults to the time when the log entry was scraped.
+When no timestamp stage is set, `loki.process` keeps the timestamp already attached to the entry by the upstream component (for example a `loki.source.*` component). Many sources default that to the time the entry was received; others can use a timestamp embedded in the log when configured to do so. Leaving out `stage.timestamp` does not by itself rewrite or sanitize timestamps, and Loki may still reject entries that fall outside its configured acceptance window (for example far-future or out-of-order timestamps).
```
```suggestion
When no timestamp stage is set, `loki.process` keeps the timestamp that the upstream component already attached to the entry, for example a `loki.source.*` component.
Many sources set that timestamp to the time they receive the entry by default.
Other sources can use a timestamp embedded in the log when you enable incoming-timestamp options.
If you leave out `stage.timestamp`, `loki.process` does not rewrite or sanitize timestamps by itself, and Loki may still reject entries that fall outside its configured acceptance window, for example far-future or out-of-order timestamps.
```

clayton-cornell (commented):
Some rewriting to clarify subject and use active voice.

locker95:
took the rewrite — splits the default-timestamp note into shorter active-voice sentences. thanks for the pass.

clayton-cornell (approved):

PR #6837, opened 2026-08-06
docs(loki.process): Document extracting docker attrs before stage.docker

### LINKED ISSUE #4609: Docs feedback: Document about fetching attrs docker field (or augment docker stage to do so automatically)
### URL

https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagedocker

### Component(s)

loki.source.docker

### Feedback

I add some tags through docker attrs log field, like this in docker compose:

```
        logging:
          driver: "json-file"
          options: 
            tag: "{{ '{{.ImageName}}|{{.Name}}|{{.ImageFullID}}|{{.FullID}}' }}"
```

To parse it back in alloy it needs to be done like this (and it took some time through trial and error to discover)

```
    stage.json {
        expressions = {
            attrs = null,
        }
    }

    stage.docker { }

    stage.json {
        expressions = {
            tag = "",
        }
        source = "attrs"
    }
```
the important bit is stage.json for attrs must come first before stage.docker or else it doesn't work.

Previously I didn't use stage.docker and with only stage.json it was much more intuitive. However stage.docker seems to be necessary to deal with cases where docker splits (json) log line into multiple when it exceeds some limit.

### Tip

locker95:
stage.docker only unpacks log/stream/time and then replaces the line with log, so anything in attrs is gone unless you pull it out first.

Added a short note plus the extract-before-docker pattern for json-file tag/attrs.

Fixes #4609

### DIFF
--- a/docs/sources/reference/components/loki/loki.process.md
+++ b/docs/sources/reference/components/loki/loki.process.md
@@ -188,6 +188,29 @@ Docker log entries are formatted as JSON with the following keys:
 * `log`: The content of log line.
 * `stream`: Either `stdout` or `stderr`.
 * `time`: The timestamp string of the log line.
+* `attrs` (optional): Extra attributes from the logging driver, for example when the json-file `tag` option is set.
+
+`stage.docker` only unpacks `log`, `stream`, and `time`. It rewrites the log line to the value of `log` and sets the stream label and timestamp. Any other top-level JSON fields, including `attrs`, are dropped unless you copy them out first.
+
+If you need values from `attrs` (or another sidecar field) **and** the docker stage (for example to reassemble split lines), extract those fields with `stage.json` **before** `stage.docker`:
+
+```alloy
+// docker compose json-file tag example: tag = "{{.ImageName}}|{{.Name}}|..."
+stage.json {
+  expressions = {
+    attrs = "",
+  }
+}
+
+stage.docker {}
+
+stage.json {
+  expressions = {
+    tag = "",
+  }
+  source = "attrs"
+}
+```
 
 Given the following log line, the subsequent key-value pairs are created in the shared map of extracted data:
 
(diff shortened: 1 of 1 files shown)


docs/sources/reference/components/loki/loki.process.md
```diff
@@ -188,6 +188,29 @@ Docker log entries are formatted as JSON with the following keys:
 * `log`: The content of log line.
 * `stream`: Either `stdout` or `stderr`.
 * `time`: The timestamp string of the log line.
+* `attrs` (optional): Extra attributes from the logging driver, for example when the json-file `tag` option is set.
+
+`stage.docker` only unpacks `log`, `stream`, and `time`. It rewrites the log line to the value of `log` and sets the stream label and timestamp. Any other top-level JSON fields, including `attrs`, are dropped unless you copy them out first.
```
```suggestion
`stage.docker` only unpacks `log`, `stream`, and `time`.
It rewrites the log line to the value of `log` and sets the stream label and timestamp.
Any other top-level JSON fields, including `attrs`, are dropped unless you copy them out first.
```

One sentence per line - simplicity and makes reviews easier.

docs/sources/reference/components/loki/loki.process.md
```diff
@@ -188,6 +188,29 @@ Docker log entries are formatted as JSON with the following keys:
...
+
+If you need values from `attrs` (or another sidecar field) **and** the docker stage (for example to reassemble split lines), extract those fields with `stage.json` **before** `stage.docker`:
```
```suggestion
If you need values from `attrs` or another sidecar field **and** the docker stage, for example to reassemble split lines, extract those fields with `stage.json` **before** `stage.docker`:
```

locker95:
split those into one sentence per line and took your wording on the extract-before-docker line.

clayton-cornell (approved):

PR #6842, opened 2026-08-07
feat!: Bump otel to v0.158.0 and prometheus to  v3.13.2

blewis12:
Bumps the contrib stack from v0.153.0 to v0.158.0, and prometheus to v3.13.2

### Pull Request Details
In terms of why versions were chosen - v0.158.0 is the lowest contrib release that compiles against prometheus 3.13.x. Contrib's prometheusreceiver calls proms `web/api/v1.NewAPI`, whose signature changed in v3.13.0, and contrib itself did not adopt the new signature until v0.158.0 (which depends on v3.13.x prometheus)

This PR deals with _only_ the version upgrade. Like the last upgrade we did, I will open follow up PR's that address the any new config/knobs for components that arose out of the version change.

Most of the diff in this PR are mechanical: upstream un-embedded structs in their config so there is a lot of refactoring to match that, core's `confmap/xconfmap.Validate` was removed in v0.158.0 in favor of `confmap.Validate` which caused some diffs. The behavior should be identical though

BEGIN_COMMIT_OVERRIDE
feat!: Bump otel to v0.158.0 and prometheus to v3.13.2 (#6842)

BREAKING-CHANGE: `otelcol.exporter.kafka`: `max_message_bytes` values above 100 MiB now fail validation. otel-contrib v0.158.0 added `max_broker_write_bytes`, whose franz-go minimum is 100 MiB, and requires `max_message_bytes` to be less than or equal to it. Alloy pins `max_broker_write_bytes` to that minimum and does not expose it yet, so configurations above 100 MiB no longer load. Lower `max_message_bytes` to 104857600 (100MiB) or below.

BREAKING-CHANGE: Components that build HTTP clients through `prometheus/common` no longer send credentials when following a redirect to a different host. This includes `prometheus.scrape`, `prometheus.remote_write`, `discovery.kubelet`, `loki.source.docker`, `pyroscope.scrape`, and the `remotecfg` service. prometheus/common v0.69.0 strips `Authorization` and other sensitive headers on cross-host redirects, matching net/http behavior. Endpoints that relied on credentials surviving such a redirect no longer authenticate; point the configuration at the final host instead

BREAKING-CHANGE: `otelcol.processor.k8sattributes`: when `metadata` is not set in the `extract` block, the processor now adds the documented default resource attributes (`container.image.name`, `container.image.tag`, `k8s.deployment.name`, `k8s.namespace.name`, `k8s.node.name`, `k8s.pod.name`, `k8s.pod.start_time`, and `k8s.pod.uid`). Alloy previously passed an empty list so no attributes were added, contrary to the component documentation. Set `metadata = []` in the `extract` block to keep the previous behavior.

BREAKING-CHANGE: `otelcol.exporter.kafka`: the producer now waits up to 10ms to batch records before sending, matching the upstream default. Alloy previously set the linger time to 0, so records were sent immediately

BREAKING-CHANGE:  The memory_limiter processor's internal telemetry metrics were renamed upstream in OpenTelemetry Collector v0.155.0 to carry a memory_limiter prefix
END_COMMIT_OVERRIDE

### PR Checklist
- [x] Documentation added
- [x] Tests updated


internal/converter/internal/otelcolconvert/converter_attributesprocessor.go
```diff
@@ -53,7 +53,7 @@ func toAttributesProcessor(state *State, id componentstatus.InstanceID, cfg *att
 
 	return &attributes.Arguments{
 		Match:   toMatchConfig(cfg),
-		Actions: toAttrActionKeyValue(encodeMapslice(cfg.Actions)),
```
blewis12, line 56: diffs are a bit noisy because of changes like this: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49972

There's been work upstream to un-embed config structs, so when referencing upstream config structs now we have to use the full path instead of relying on field promotion

In this case: [before](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42f949127580c0d00088b785cd0b35842dc0ddb8/processor/attributesprocessor/config.go#L28) & [after](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/processor/attributesprocessor/config.go#L28)

internal/component/otelcol/exporter/kafka/kafka.go
```diff
@@ -202,14 +202,14 @@ type Producer struct {
...
 func (args Producer) Convert() configkafka.ProducerConfig {
-	return configkafka.ProducerConfig{
-		MaxMessageBytes:        args.MaxMessageBytes,
-		RequiredAcks:           configkafka.RequiredAcks(args.RequiredAcks),
-		Compression:            args.Compression,
-		CompressionParams:      args.CompressionParams.Convert(),
-		FlushMaxMessages:       args.FlushMaxMessages,
-		AllowAutoTopicCreation: args.AllowAutoTopicCreation,
-	}
+	cfg := configkafka.NewDefaultProducerConfig()
```
blewis12, line 205: This was previously building the config struct from zero, which meant that any fields that weren't explicitly being assigned here were being assigned their zero-value instead of using upstream defaults. `MaxBrokerWriteBytes` was introduced as a new config param upstream which is required to have a value of 100 MiB. I think instead of trying to create a default value here in alloy we can just use upstream defaults? 

The one change here would be that now Linger (previously unexposed) will pick up the upstream default of [10s](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/pkg/kafka/configkafka/config.go#L334)

internal/component/pyroscope/write/promhttp2/testdata/http.conf.basic-auth.bad-username.yaml
```diff
@@ -1,4 +1,4 @@
 basic_auth:
   username: user
-  username_file: testdata/basic-auth-username
```
blewis12, line 3: all these tests changed because [LoadHTTPConfigFile](https://github.com/grafana/alloy/blob/7e3e5aad4cf2c60687632809a6b76c5efc402049/internal/component/pyroscope/write/promhttp2/http_config_test.go#L1403) used to look like [this](https://github.com/prometheus/common/blob/934ff3789ee17026206fe4f7e5f59c0a09fbe511/config/http_config.go#L327) and now looks like [this](https://github.com/prometheus/common/blob/v0.69.0/config/http_config.go#L328).

internal/converter/internal/staticconvert/testdata-v2/integrations_v2.alloy
```diff
@@ -731,6 +712,25 @@ prometheus.scrape "integrations_gcp" {
 	job_name   = "integrations/gcp"
 }
 
+prometheus.exporter.mssql "integrations_mssql" {
```
blewis12, line 715: This is the same content just moved, bumping `prometheus/common` and `client_golang` changed the order for which the static converter emits integrations

internal/converter/internal/staticconvert/testdata/integrations.alloy
```diff
@@ -739,6 +709,36 @@ prometheus.scrape "integrations_gcp_exporter" {
 	}
 }
 
+prometheus.exporter.mssql "integrations_mssql" {
```
blewis12, line 712: Same as [above](https://github.com/grafana/alloy/pull/6842/changes#r3750277319)

collector/main.go
```diff
@@ -38,11 +38,11 @@ func main() {
...
-			httpprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/httpprovider v1.59.0",
-			httpsprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/httpsprovider v1.59.0",
-			yamlprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/yamlprovider v1.59.0",
+			envprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/envprovider v1.64.0",
+			fileprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/fileprovider v1.64.0",
+			httpprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/httpprovider v1.64.0",
+			httpsprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/httpsprovider v1.64.0",
+			yamlprovider.NewFactory().Create(confmap.ProviderSettings{}).Scheme(): "go.opentelemetry.io/collector/confmap/provider/yamlprovider v1.64.0",
     	},
```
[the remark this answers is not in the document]
blewis12: ignore - this code is generated

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -92,7 +92,12 @@ Supported strategies are:
   It minimizes the number of partition movements, which can be beneficial for stateful consumers.
   For more information, refer to the Kafka StickyAssignor documentation, see [StickyAssignor][].
 - `cooperative-sticky`: This strategy uses incremental cooperative rebalancing to reduce partition movement during rebalances.
-  For more information, refer to the Kafka CooperativeStickyAssignor documentation, see [CooperativeStickyAssignor][].
+  For more information, refer to the Kafka CooperativeStickyAssignor documentation, refer to [CooperativeStickyAssignor][].
+
+{{< admonition type="note" >}}
+The upstream OpenTelemetry Collector setting behind `group_rebalance_strategy` is deprecated in favor of an ordered list of strategies.
+`group_rebalance_strategy` continues to work, and Alloy's `range` default is unchanged.
```
clayton-cornell:
```suggestion
`group_rebalance_strategy` continues to work, and the `range` default is unchanged.
```

grcevski (dismissed):
LGTM!

internal/component/otelcol/exporter/kafka/kafka.go
kgeckhart: Just wondering if https://github.com/grafana/alloy/blob/main/docs/developer/key-deps-update/key-dep-updates.md?plain=1#L179-L182 helped on this one.

There's a follow up I didn't do to try to invert the config for our components but it's a massive task and hopefully it will happen slowly through upgrades. 

I think changing to the upstream default is okay and is defensible. Especially if it was a zero value that could produce unexpected behavior.s

internal/component/otelcol/exporter/kafka/kafka.go
```diff
@@ -357,7 +362,7 @@ func (args Arguments) Convert() (otelcomponent.Config, error) {
 
 	if args.TLS != nil {
 		tlsCfg := args.TLS.Convert()
-		result.TLS = tlsCfg
+		result.ClientConfig.TLS = tlsCfg
 	}
```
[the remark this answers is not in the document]
kgeckhart: This one looks valid line 338 and 358 are fully duplicated.

internal/component/otelcol/processor/attributes/attributes_test.go
kgeckhart, line 1: I'm noticing a few places where we are changing config struct layout on the OTel side. Are these configs we don't expose today? If we do expose them should we try to mirror them? 

Edit: I found https://github.com/grafana/alloy/pull/6842/changes#r3747995417 which looks like it explains it. I still wondering about mirroring upstream if it makes sense.

kgeckhart (commented):
Sad to see some of my short lived forks might not be so short lived after all 🙈 . Just a question.

internal/component/otelcol/exporter/kafka/kafka.go
blewis12: most likely it did, especially if this didn't pop up before in any upgrades 🤔 

I also think copying upstream defaults is the way to go, but wanted to make sure that we didn't accidentally break existing users by doing that. I think in this case it's the reasonable thing to do though

internal/component/otelcol/processor/attributes/attributes_test.go
blewis12: Yeah that comment addresses why the changes were made - upstream un-embedded a bunch of structs across config so we have to match that when we reference individual fields. All these configs are indeed exposed by alloy, this PR doesn't introduce any new knobs - those will be follow up PR's. Worth saying as well that the _user_ facing config is unchanged, it's the struct definition that changed.

I think we already mirror what's going on upstream, even moreso now with the upstream change. Using that attributes processor comment as an example, upstream looks like

```
type Config struct {
	MatchConfig filterconfig.MatchConfig `mapstructure:",squash"`
...
}
```

and alloy looks like

```
type Arguments struct {
	Match otelcol.MatchConfig `alloy:",squash"`
...
}
```

Let me know if you meant something else by mirroring, but imo we're already matching upstream as we are

internal/component/otelcol/exporter/kafka/kafka.go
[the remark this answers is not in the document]
blewis12: oh yeah, looks like this has been here a while - i'll fix that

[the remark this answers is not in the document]
blewis12: https://github.com/grafana/alloy/pull/6842/commits/8a2464656c1f57ac895a8ce512f784438a88b4ea

kgeckhart (dismissed):

internal/component/otelcol/processor/k8sattributes/k8sattributes_test.go
```diff
@@ -467,3 +467,41 @@ func Test_WatchSyncPeriod(t *testing.T) {
 		require.Equal(t, 30*time.Second, otelObj.WatchSyncPeriod)
 	})
 }
+
+func Test_ExtractMetadata(t *testing.T) {
+	t.Run("omitted metadata inside an extract block is nil", func(t *testing.T) {
```
blewis12: @kgeckhart an interesting bug I found - claude pointed out that because the `k8sattributeprocessor` builds from a [zero struct](https://github.com/grafana/alloy/blob/4633351c14d5b414161098d29b56aa2a181d8bd9/internal/component/otelcol/processor/k8sattributes/k8sattributes.go#L117), and because upstream introduced a [new config param](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.158.0/processor/k8sattributesprocessor/config.go#L57) for default grace period that was previously a [const](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42f949127580c0d00088b785cd0b35842dc0ddb8/processor/k8sattributesprocessor/internal/kube/kube.go#L85) - that means that if a user doesn't pass in any config for the grace period (or the knob just isn't exposed, like it isn't now) then it's going to get zero'd. So users go from having 120s as the `pod_delete_grace_period` to having 0s. 

The fix for this would be the same as the kafka component, which is to just ensure we construct the default struct instead. but now we get a _different_ issue: previously if [extract.metadata](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.k8sattributes/#extract) was left blank by a user, then it's zero'd and we pass through nil to the underlying otel config. This is actually the opposite of what we document [here](https://grafana.com/docs/alloy/latest/reference/components/otelcol/otelcol.processor.k8sattributes/#extract)

> By default, if metadata isn’t specified, the following fields are extracted and > added to spans, metrics, and logs as resource attributes ... (lists 8 attributes)

so I think the fact that we pass through nil right now is itself a bug. It would be fixed by using the default struct, but the only thing is that users who previously didn't supply metadata and now going to find themselves going from zero metadata attributes being added to the 8 default ones. I would list that as a "fix" in the changelog, though it might border on breaking change

I pushed this test here to demonstrate what I mean - excluding metadata results in nil metadata on the underlying otel config

kgeckhart: 👍 thanks for all the context. I think it's safe and perfectly reasonable to list it as a breaking change with a caveat that it now matches the behavior we documented.

kgeckhart (dismissed):

internal/component/otelcol/processor/k8sattributes/k8sattributes_test.go
blewis12: sweet, pushed a fix [here](https://github.com/grafana/alloy/pull/6842/changes/9f37dcfd739f640897137f61ea70587ce039b8c2) which includes a test and updating our docs (the defaults we list are slightly outdated)

clayton-cornell (approved):

kgeckhart (approved):

PR #6868, opened 2026-08-12
feat: Expose new kafka receiver/exporter knobs after upgrade to v0.158.0

blewis12:
A follow up to the main upgrade: https://github.com/grafana/alloy/pull/6842

This introduces some further knobs that are available after the upgrade from contrib v0.153.0 to v0.158.0

### Pull Request Details
**otelcol.kafka.receiver** => exposes the `group_rebalance_strategies` which is intended to be used instead of the now deprecate `group_rebalance_strategy` - the deprecated value is still valid so this is not a breaking change

**otelcol.kafka.exporter** => exposes `max_broker_write_bytes` and `linger`. `max_broker_write_bytes` must be greater than or equal to `max_message_bytes` which we hardcode to [1000000](https://github.com/grafana/alloy/pull/6868/changes#diff-6a9be317468d7fad9ccd739a250747799a743c675a7ea43807548eba22f1a92aR260). This condition is a breaking change from upstream that is documented in the changelog for the upgrade

### Notes to the Reviewer
This PR also includes a fix to our OTel dashboards, addressing a breaking change upstream that renames the `memory_limiter` processor internal metrics

### PR Checklist
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
```diff
@@ -52,6 +52,7 @@ You can use the following arguments with `otelcol.receiver.kafka`:
 | `group_id`                                 | `string`       | Consumer group to consume messages from.                                                                              | `"otel-collector"` | no       |
 | `group_instance_id`                        | `string`       | A unique identifier for the consumer instance within a consumer group.                                                | `""`               | no       |
 | `group_rebalance_strategy`                 | `string`       | The strategy used to assign partitions to consumers within a consumer group.                                          | `"range"`          | no       |
+| `group_rebalance_strategies`               | `list(string)` | The ordered list of strategies to advertise to the group coordinator.                                                 | `[]`               | no       |
```
kgeckhart: Nit: Might be good document strategy as deprecated here and note in each that it's mutually exclusive with the other.

internal/component/otelcol/exporter/kafka/kafka.go
```diff
@@ -175,6 +175,11 @@ type Producer struct {
 	// Maximum message bytes the producer will accept to produce.
 	MaxMessageBytes int `alloy:"max_message_bytes,attr,optional"`
 
+	// MaxBrokerWriteBytes is the maximum bytes the producer will write to a broker
+	// in a single request. Must be greater than or equal to max_message_bytes, and
+	// at least 100 MiB
```
kgeckhart, line 180: The wording on here is a bit awkward, checking upstream I think it's suggesting this value needs to be changed when max_message_bytes goes above 100MiB which is the default value here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/kafkaexporter/README.md

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -215,15 +215,22 @@ The `producer` block configures how to retry retrieving metadata when retrieval
...
+| `compression`            | `string`   | The level of compression to use on messages.                               | `"none"`    | no       |
+| `flush_max_messages`     | `number`   | The maximum number of messages in one request.                             | `10000`     | no       |
+| `linger`                 | `duration` | How long a topic partition waits for more records before building a request. | `"10ms"`    | no       |
+| `max_broker_write_bytes` | `number`   | The maximum permitted size of a single write to a broker in bytes.         | `104857600` | no       |
+| `max_message_bytes`      | `number`   | The maximum permitted size of a message in bytes.                          | `1000000`   | no       |
+| `required_acks`          | `number`   | Controls when a message is regarded as transmitted.                        | `1`         | no       |
 
 Refer to the [Go sarama documentation][RequiredAcks] for more information on `required_acks`.
 
+`max_broker_write_bytes` must be at least `104857600` (100 MiB), and `max_message_bytes` must be less than or equal to it.
```
clayton-cornell:
```suggestion
`max_broker_write_bytes` must be at least `104857600` (100 MiB), and `max_message_bytes` must be less than or equal to `max_broker_write_bytes`.
```

Using `it` makes the subject ambiguous.

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
```diff
@@ -215,15 +215,22 @@ The `producer` block configures how to retry retrieving metadata when retrieval
...
 
-| Name                 | Type     | Description                                         | Default   | Required |
-| -------------------- | -------- | --------------------------------------------------- | --------- | -------- |
-| `compression`        | `string` | The level of compression to use on messages.        | `"none"`  | no       |
-| `flush_max_messages` | `number` | The maximum number of messages in one request.      | `10000`   | no       |
-| `max_message_bytes`  | `number` | The maximum permitted size of a message in bytes.   | `1000000` | no       |
-| `required_acks`      | `number` | Controls when a message is regarded as transmitted. | `1`       | no       |
+| Name                     | Type       | Description                                                                | Default     | Required |
+| ------------------------ | ---------- | -------------------------------------------------------------------------- | ----------- | -------- |
+| `compression`            | `string`   | The level of compression to use on messages.                               | `"none"`    | no       |
```
clayton-cornell: Is this level of compression? or the compression algorithm?

kgeckhart (commented):
AI noted that introducing `max_broker_write_bytes` without exposing it in the converter could break previously working configs like,

```
producer:
  max_message_bytes: 209715200
  max_broker_write_bytes: 209715200
```

which would have ignored the field previously but will now error because `max_broker_write_bytes` didn't come along. It also found a second converter issues that came with the bump since upstream no longer defaults a `group_rebalance_strategy` an exist config which did not set it will fail to load. 

I don't know that the converters are a big deal but might be good to ensure it's not going to break an existing config.

docs/sources/reference/components/otelcol/otelcol.exporter.kafka.md
blewis12: ahh good point, it's the compression algorithm

blewis12: https://github.com/grafana/alloy/pull/6868/commits/dc597e3c81aad5b3a3d8b23cb7687f052f9e3b76

docs/sources/reference/components/otelcol/otelcol.receiver.kafka.md
blewis12: https://github.com/grafana/alloy/pull/6868/commits/c817f68d262d3f0f8273220fe491d85833a77792

blewis12:
@kgeckhart yeah that's true, `max_broker_write_bytes` needs to be included in the conversion otherwise a valid upstream config (where `max_broker_write_bytes` < `max_message_bytes`) could turn into an invalid alloy config - i'll add it to the converter

> upstream no longer defaults a `group_rebalance_strategy`

banged my head over this a bit, upstream _never_ set a default ([v0.153.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.153.0/pkg/kafka/configkafka/config.go#L159-L174) and [v0.158.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.158.0/pkg/kafka/configkafka/config.go#L171-L186)). I think this was already broken, I tested it on main - If you pass through otel config currently and omit `group_rebalance_strategy` then the resulting config fails to load. 

Alloy sets a default to `range` in it's own config, but upstream just uses the default set within `go-franz` is [cooperative-sticky](https://github.com/twmb/franz-go/blob/v1.21.5/pkg/kgo/config.go#L689). 

I think then if we get otel config which omits `group_rebalance_strategy/ies` we should fallback to `cooperative-sticky` as part of the conversion - then at least behavior is true to the config. We could also change our default value in alloy from `range` to `cooperative-sticky` though that's another breaking change

internal/component/otelcol/receiver/kafka/kafka.go
```diff
@@ -86,20 +86,19 @@ func (args *Arguments) SetToDefault() {
...
-		ClientID:               "otel-collector",
-		GroupID:                "otel-collector",
-		InitialOffset:          "latest",
-		SessionTimeout:         10 * time.Second,
-		HeartbeatInterval:      3 * time.Second,
-		MinFetchSize:           1,
-		MaxFetchSize:           1048576,
-		MaxPartitionFetchSize:  1048576,
-		MaxFetchWait:           250 * time.Millisecond,
-		GroupRebalanceStrategy: "range",
```
blewis12, line 99: this default setting was moved to below so that we can ensure it's only set if `GroupRebalanceStrategies` isn't set

internal/component/otelcol/exporter/kafka/kafka.go
blewis12: hmm, we could just simplify it to this maybe

```
the maximum bytes the producer will write to a broker in a single request. Must be greater than or equal to max_message_bytes
```

there _is_ a minimum value for `max_broker_write_bytes` [of 100MiB though](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/821a9d9c2c1623c4a0ceba5d47b57c48879c3f84/pkg/kafka/configkafka/config.go#L364) - it's enforced upstream but not clearly documented

kgeckhart: Ugh how annoying, from a purely practical standpoint maybe we mirror the upstream documentation for now? Edit: It;s a code comment that's accurate for the code so probably fine.

kgeckhart (approved):

PR #6872, opened 2026-08-12
POC: New structure for loki.process pipeline

kalleep:
Most loki components are trivial to change in order to support the new loki pipeline we are moving too but `loki.process` is not.

We have a couple of different types of stages:
1. Simple stages that just process entries and passes to next e.g. [stage.static_labels](https://github.com/grafana/alloy/blob/a7696d5565a78563a1f36714165ecaed1fa0794a/internal/component/loki/process/stages/static_labels.go#L63)
2. Stages that can drop labels e.g. [stage.drop](https://github.com/grafana/alloy/blob/a7696d5565a78563a1f36714165ecaed1fa0794a/internal/component/loki/process/stages/drop.go)
3. Stages that can produce multiple entries, [stage.split_json](https://github.com/grafana/alloy/blob/a7696d5565a78563a1f36714165ecaed1fa0794a/internal/component/loki/process/stages/split_json.go) and [stage.cri](https://github.com/grafana/alloy/blob/a7696d5565a78563a1f36714165ecaed1fa0794a/internal/component/loki/process/stages/cri.go#L98)
4. Stages that creates a sub pipeline, the only stage we have is [stage.match](https://github.com/grafana/alloy/blob/a7696d5565a78563a1f36714165ecaed1fa0794a/internal/component/loki/process/stages/match.go#L127)
5. Stages that can flush entries on it's own, the only one with this capability is [stage.multiline](https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/multiline.go#L178)

I am pretty sure the last one is the reason why this component today is built up by starting each stage in it's own goroutine and chain them with channels.

This is problematic for the loki pipeline changes since with this structure a batch / entry sent down the pipeline would be disconnected from the original caller and this would still cause the same head-of-line blocking that exists today.

What we really want is for the stages them self just be function calls, just like component are going to be. In this pr I have experimented a bit with a new stage / pipeline interface. The best way to check how it would be used is to check the added test `TestPipeline2` where I have added these different kind of stages we have today (simple ports).

A couple of things I want to highlight:
1. All stages that carries internal state now needs to be safe to call concurrently, for this POC made them safe with a sync.Mutex but we need to review this stage by stage.
2. Since `stage.multiline` can flush entries on it's own we need an external trigger, this would most likely be up to the component to scan for nearest flush deadline from potentially several stages and call the flush function and forward the batch on it's own.
3. For this POC I opted to supply a `Emitter` interface to every Process call, this allows a stage to emit multiple entries during on call. Another solution instead of this would be to change the signature of each stage to work on slices of entries instead and loop on each stage.

### DIFF
--- a/internal/component/loki/process/stages/pipeline_4.go
+++ b/internal/component/loki/process/stages/pipeline_4.go
@@ -1,0 +1,442 @@
+package stages
+
+import (
+	"bytes"
+	"context"
+	"maps"
+	"regexp"
+	"slices"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/grafana/alloy/internal/component/common/loki"
+	crip "github.com/grafana/alloy/internal/component/loki/process/stages/cri"
+	"github.com/grafana/loki/pkg/push"
+	"github.com/prometheus/common/model"
+)
+
+type Emitter2 func(ctx context.Context, entries []Entry) error
+
+type Stage4 interface {
+	Process(ctx context.Context, entries []Entry) error
+	SetNext(next Emitter2)
+	Cleanup()
+}
+
+var _ Stage4 = (*Pipeline4)(nil)
+
+type Pipeline4 struct {
+	stages []Stage4
+}
+
+func NewPipeline4(stages []Stage4) *Pipeline4 {
+	return &Pipeline4{stages: stages}
+}
+
+func (p *Pipeline4) ProcessBatch(ctx context.Context, batch loki.Batch) error {
+	entries := make([]Entry, 0, batch.EntryLen())
+	_ = batch.ConsumeStreams(func(stream loki.Stream, created int64) error {
+		for _, e := range stream.Entries {
+			entries = append(entries, Entry{
+				Extracted: make(map[string]any, len(stream.Labels)),
+
+				Entry: loki.NewEntryWithCreatedUnixMicro(stream.Labels.Clone(), created, e),
+			})
+		}
+		return nil
+	})
+
+	return p.stages[0].Process(ctx, entries)
+}
+
+func (p *Pipeline4) ProcessEntry(ctx context.Context, entry loki.Entry) error {
+	return p.stages[0].Process(ctx, []Entry{
+		{
+			Extracted: make(map[string]any, len(entry.Labels)),
+			Entry:     entry,
+		},
+	})
+}
+
+func (p *Pipeline4) Process(ctx context.Context, entries []Entry) error {
+	return p.stages[0].Process(ctx, entries)
+}
+
+func (p *Pipeline4) SetNext(next Emitter2) {
+	for i, s := range slices.Backward(p.stages) {
+		if i == len(p.stages)-1 {
+			s.SetNext(next)
+			continue
+		}
+
+		s.SetNext(p.stages[i+1].Process)
+	}
+}
+
+func (p *Pipeline4) Cleanup() {}
+
+var (
+	_ Stage4 = (*staticLabelsStage4)(nil)
+)
+
+func newStaticLabelsStage4(values []string) *staticLabelsStage4 {
+	return &staticLabelsStage4{values: values}
+}
+
+// staticLabelsStage4 is the simplest kind of stage. It just processes the
+// current entry and always forwards it to next.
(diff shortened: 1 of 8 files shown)


internal/component/loki/process/stages/pipeline_4.go
```diff
@@ -0,0 +1,442 @@
...
+func (p *Pipeline4) SetNext(next Emitter2) {
+	for i, s := range slices.Backward(p.stages) {
+		if i == len(p.stages)-1 {
+			s.SetNext(next)
+			continue
+		}
+
+		s.SetNext(p.stages[i+1].Process)
+	}
+}
```
thampiotr, line 75: This is problematic, because it's a setter with a side-effect AND `SetNext` is mandatory. It does two things:
- Sets the next for the whole pipeline - but pipeline is not really useful without next, so this should be made a mandatory parameter in a constructor in `NewPipeline4`
- Sets the internal state of each stage to point to its next stage - that's an unexpected side-effect and I would like this to live somewhere else. 

This brings also another ask. For this POC to visualise better how it would work, could you maybe add some code to show how Pipeline4 would be used in `process.go`?

internal/component/loki/process/stages/pipeline_4.go
```diff
@@ -0,0 +1,442 @@
...
+
+func (c *criStage4) Cleanup() {}
+
+func newMatchStage4(match func(Entry) bool, action string, stages []Stage4) Stage4 {
+	switch action {
+	case MatchActionKeep:
+		return newMatchStageKeep4(match, stages)
+	default:
+		return newMatchStageDrop4(match)
```
thampiotr, line 183: Let's avoid the use of word 'drop' as it indicates we will drop the logs. 
Maybe it should be `matchStage(Select|Skip)` or something similar?

kalleep: Well that stage can drop logs and our actions are called `action = "drop|keep"`

internal/component/loki/process/stages/pipeline_4.go
```diff
@@ -0,0 +1,442 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-m.stop:
+			return
+		case <-ticker.C:
+			entries := m.expired()
+			if len(entries) > 0 {
+				_ = m.next(context.Background(), entries)
```
thampiotr, line 422: One thing to consider here is that this can be slow and can error. We should think how to handle this.

internal/component/loki/process/stages/pipeline_4.go
```diff
@@ -0,0 +1,442 @@
...
+	if dst == 0 {
+		return nil
+	}
+	return m.next(ctx, entries[:dst])
+}
+
+func (m *multilineStage4) processEntry(key model.Fingerprint, e Entry) (Entry, bool) {
+	state, hasState := m.streams[key]
+
+	isFirstLine := m.regex.MatchString(e.Line)
```
thampiotr, line 326: Shouldn't we check for expired timer and flush lines that would otherwise be flushed by the goroutine?

internal/component/loki/process/stages/pipeline_4.go
thampiotr: Ah, true. Not a fan of this haha. 
So, am I right in understanding that:
```
if selector matches:
  run through sub-stages
else if action is keep:
  skip sub-stages and continue processing
else if action is drop:
  drop the log line
```
?

Would be nice to make it more obvious in naming.

kalleep: That is not really how this stage works.

1. `action = "keep"` - All matching entries passes through sub-pipeline, rest skips sub-pipeline.
2. `action = "drop"` - All matching entries are dropped rest is forwarded, no sub-pipeline involved

That is why I create one stage for each action

internal/component/loki/process/stages/pipeline_4.go
kalleep: We could, but we could also just let the go-routine take care of that, the logic becomes much simpler that way.

kalleep:
Closing in favor of https://github.com/grafana/alloy/pull/6913

PR #6876, opened 2026-08-13
feat: Part 2 - Expose knobs for OTel v0.158.0 upgrade

blewis12:
Exposes some config now available upstream with the v0.153.0 -> v0.158 contrib bump

This PR covers a handful of components, there will be a couple other follow up PR's covering other components as well, I just didn't want to have one massive PR and break things out into reviewable chunks

### Pull Request Details
**otelcol.processor.k8sattributes** => exposes `pod_delete_grace_period`. This PR also fixes the converter, which wasn't mapping `watch_sync_period`

**otelcol.connector.spanmetrics** => exposes `glob` on the dimension block

 **otelcol.exporter.file** => exposes the `compression_params` block

 **otelcol.receiver.awscloudwatch** => exposes `initial_lookback` on the logs block

 **otelcol.receiver.googlecloudpubsub** / **otelcol.exporter.googlecloudpubsub** => `expose universe_domain` on both

**shared tls block** => exposes `include_insecure_cipher_suites`, which allows `cipher_suites` to select suites Go considers insecure. It lives on `TLSSetting`, so it reaches every component with a client or server tls block, and is documented in the two shared doc snippets.

### PR Checklist
- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/otelcol/otelcol.exporter.file.md
```diff
@@ -68,6 +70,18 @@ You can use the following blocks with `otelcol.exporter.file`:
 
 You can't enable `append` and `rotation` together.
 
+### `compression_params`
+
+The `compression_params` block configures the compression codec selected by `compression`.
```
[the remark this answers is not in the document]
blewis12: I think this should be valid? it's a technical term that makes sense here

docs/sources/reference/components/otelcol/otelcol.processor.k8sattributes.md
```diff
@@ -46,6 +46,7 @@ You can use the following arguments with `otelcol.processor.k8sattributes`:
 | `wait_for_metadata_timeout` | `duration` | How long to wait for Kubernetes metadata to arrive.                            | `"10s"`            | no       |
 | `wait_for_metadata`         | `bool`     | Whether to wait for Kubernetes metadata to arrive before processing telemetry. | `false`            | no       |
 | `watch_sync_period`         | `duration` | The resync period for the Kubernetes informers. `0` disables periodic resync.  | `"5m"`             | no       |
```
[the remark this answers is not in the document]
blewis12: (i'm resolving any comments here that pop up under changes I didn't make)

internal/component/otelcol/processor/k8sattributes/k8sattributes.go
```diff
@@ -70,6 +74,8 @@ func (args *Arguments) SetToDefault() {
 	}
 	args.WaitForMetadataTimeout = 10 * time.Second
 	args.WatchSyncPeriod = 5 * time.Minute
+	args.PodDeleteGracePeriod = k8sattributesprocessor.NewFactory().
+		CreateDefaultConfig().(*k8sattributesprocessor.Config).PodDeleteGracePeriod
```
kalleep: Could we use default values for other fields here?

internal/component/otelcol/receiver/googlecloudpubsub/googlecloudpubsub_test.go
```diff
@@ -121,3 +121,18 @@ func TestValidate(t *testing.T) {
 		})
 	}
 }
+
+func TestUniverseDomain(t *testing.T) {
```
kalleep: So we add a bunch of these small tests in this pr.

Ideally we would have one or more test covering all setting for a component and not just these small specialized tests for properties that was just added, since we miss coverage for basically everything else. WDYT?

internal/component/otelcol/processor/k8sattributes/k8sattributes.go
blewis12: looks like it actually yes 👍 these all match upstream defaults

internal/component/otelcol/receiver/googlecloudpubsub/googlecloudpubsub_test.go
blewis12: yeah that's a good point, I was trying to keep it specific to these changes but makes sense to have tests grouped together

internal/component/otelcol/connector/spanmetrics/spanmetrics.go
```diff
@@ -138,26 +138,17 @@ func (args *Arguments) SetToDefault() {
...
-		return fmt.Errorf("invalid aggregation_temporality: %v", args.AggregationTemporality)
-	}
-
-	if args.AggregationCardinalityLimit < 0 {
-		return fmt.Errorf("invalid aggregation_cardinality_limit: %v, the limit should be positive", args.AggregationCardinalityLimit)
-	}
-
-	if args.AggregationTemporality == AggregationTemporalityDelta && args.TimestampCacheSize <= 0 {
-		return fmt.Errorf("invalid metric_timestamp_cache_size: %v, the cache size should be positive", args.TimestampCacheSize)
+	cfg, err := args.Convert()
```
blewis12, line 147: we can actually just delegate to the upstream validation here, instead of duplicating it here - it's a pattern we use elsewhere in otel components in alloy

docs/sources/reference/components/otelcol/otelcol.exporter.googlecloudpubsub.md
```diff
@@ -45,15 +45,16 @@ Refer to the [Google Cloud Pub/Sub Exporter][] and [Google Cloud Exporter][] doc
...
+| `project`         | `string`   | Google Cloud Platform project identifier.                                                                                                                              | Fetch from credentials                          | no       |
+| `timeout`         | `Duration` | Timeout for calls to the Pub/Sub API.                                                                                                                                  | `"12s"`                                         | no       |
+| `universe_domain` | `string`   | Universe domain for the Pub/Sub service. Set to support Sovereign Cloud regions.                                                                                       | `"googleapis.com"`                              | no       |
+| `user_agent`      | `string`   | Override the user agent string on requests to Cloud Monitoring. This only applies to metrics. Specify `{{version}}` to include the application version number.         | `"opentelemetry-collector-contrib {{version}}"` | no       |
```
clayton-cornell:
```suggestion
| `user_agent`      | `string`   | Override the user agent string used for Pub/Sub requests. Specify `{{version}}` to include the application version number.         | `"opentelemetry-collector-contrib {{version}}"` | no       |
```

Remove the Cloud Monitoring ref and the implication that this is metrics-only. (this is based on a Claude analysis which finds that it's referencing the wrong backend, and that stating metrics only implies a signal specific behaviour that the component does not expose)

blewis12: that's true, good catch

internal/component/otelcol/receiver/googlecloudpubsub/googlecloudpubsub_test.go
blewis12: https://github.com/grafana/alloy/pull/6876/changes/5007545571281d559976a960a40794e2ebbd9951 / https://github.com/grafana/alloy/pull/6876/changes/5a34903b9fca223f5c01bf23812046584da03990

tried to combine tests as best I could, sometimes it's still quite verbose but I think that's about acceptable?

kalleep (approved):

PR #6885, opened 2026-08-13
fix(prometheus.operator.servicemonitors)!: Sync ServiceMonitor permission semantics with upstream

jharvey10:

Sync `prometheus.operator.servicemonitors` with Prometheus Operator's arbitrary filesystem access guard for ServiceMonitor endpoints. ServiceMonitor endpoints now reject `bearerTokenFile`, `tlsConfig.caFile`, `tlsConfig.certFile`, and `tlsConfig.keyFile` by default, while allowing trusted deployments to opt out with `allow_arbitrary_file_access = true`.

This also fixes a bug where a failing endpoint would leave earlier endpoints from the same ServiceMonitor already written into the shared scrape/discovery config

BEGIN_COMMIT_OVERRIDE
fix(prometheus.operator.servicemonitors)!: Sync ServiceMonitor permission semantics with upstream

BREAKING CHANGE: `prometheus.operator.servicemonitors` now rejects ServiceMonitor endpoints that reference local files through `bearerTokenFile`, `tlsConfig.caFile`, `tlsConfig.certFile`, or `tlsConfig.keyFile` by default. Set `allow_arbitrary_file_access = true` to preserve the previous behavior for trusted ServiceMonitor authors.
END_COMMIT_OVERRIDE

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
+++ b/docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
@@ -30,7 +30,7 @@ In these cases, the secrets are loaded and refreshed only when the ServiceMonito
 By default, `disallow_arbitrary_file_access` prevents ServiceMonitor endpoints from referencing files on the {{< param "PRODUCT_NAME" >}} filesystem through `bearerTokenFile`, `tlsConfig.caFile`, `tlsConfig.certFile`, or `tlsConfig.keyFile`.
 This is important in multi-tenant Kubernetes clusters where ServiceMonitor resources may be created by users who don't have the same permissions as {{< param "PRODUCT_NAME" >}}.
 
-Use `bearerTokenSecret`, `authorization`, or TLS secret and config map references in the ServiceMonitor instead of local file references.
+Use `bearerTokenSecret`, `authorization`, or TLS secret and `ConfigMap` references in the ServiceMonitor instead of local file references.
 
 ## Usage
 
--- a/docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
+++ b/docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
@@ -25,6 +25,13 @@ You can run it from outside the cluster by supplying connection info in the `cli
 ServiceMonitors may reference secrets for authenticating to targets to scrape them.
 In these cases, the secrets are loaded and refreshed only when the ServiceMonitor is updated or when this component refreshes its' internal state, which happens on a 5-minute refresh cycle.
 
+## Security considerations
+
+By default, `disallow_arbitrary_file_access` prevents ServiceMonitor endpoints from referencing files on the {{< param "PRODUCT_NAME" >}} filesystem through `bearerTokenFile`, `tlsConfig.caFile`, `tlsConfig.certFile`, or `tlsConfig.keyFile`.
+This is important in multi-tenant Kubernetes clusters where ServiceMonitor resources may be created by users who don't have the same permissions as {{< param "PRODUCT_NAME" >}}.
+
+Use `bearerTokenSecret`, `authorization`, or TLS secret and config map references in the ServiceMonitor instead of local file references.
+
 ## Usage
 
 ```alloy
@@ -37,12 +44,13 @@ prometheus.operator.servicemonitors "<LABEL>" {
 
 You can use the following arguments with `prometheus.operator.servicemonitors`:
 
-| Name                    | Type                    | Description                                                                                               | Default       | Required |
-| ----------------------- | ----------------------- | --------------------------------------------------------------------------------------------------------- | ------------- | -------- |
-| `forward_to`            | `list(MetricsReceiver)` | List of receivers to send scraped metrics to.                                                             |               | yes      |
-| `informer_sync_timeout` | `duration`              | Timeout for initial sync of ServiceMonitor resources.                                                     | `"1m"`        | no       |
-| `kubernetes_role`       | `string`                | The Kubernetes role used for discovery. Supports `endpoints` or `endpointslice`.                          | `"endpoints"` | no       |
-| `namespaces`            | `list(string)`          | List of namespaces to search for ServiceMonitor resources. If not specified, all namespaces are searched. |               | no       |
+| Name                               | Type                    | Description                                                                                               | Default       | Required |
+| ---------------------------------- | ----------------------- | --------------------------------------------------------------------------------------------------------- | ------------- | -------- |
+| `disallow_arbitrary_file_access`   | `bool`                  | Disallow ServiceMonitor endpoints that reference arbitrary files on the {{< param "PRODUCT_NAME" >}} filesystem. | `true`        | no       |
+| `forward_to`                       | `list(MetricsReceiver)` | List of receivers to send scraped metrics to.                                                             |               | yes      |
+| `informer_sync_timeout`            | `duration`              | Timeout for initial sync of ServiceMonitor resources.                                                     | `"1m"`        | no       |
+| `kubernetes_role`                  | `string`                | The Kubernetes role used for discovery. Supports `endpoints` or `endpointslice`.                          | `"endpoints"` | no       |
+| `namespaces`                       | `list(string)`          | List of namespaces to search for ServiceMonitor resources. If not specified, all namespaces are searched. |               | no       |
 
 ## Blocks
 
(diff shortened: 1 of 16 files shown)


docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
```diff
@@ -25,6 +25,13 @@ You can run it from outside the cluster by supplying connection info in the `cli
 ServiceMonitors may reference secrets for authenticating to targets to scrape them.
 In these cases, the secrets are loaded and refreshed only when the ServiceMonitor is updated or when this component refreshes its' internal state, which happens on a 5-minute refresh cycle.
 
+## Security considerations
```
clayton-cornell: I'm not sure this is the right place for this.  The location here is before Usage and Arguments/Blocks, before the arguments are defined in the topic. I would suggest it is moved to the Arguments section, below the table and wrapped in a Caution (or possibly a Warning) admonition instead of a standalone section under an H2 (my preference).  If it needs a heading level, I'd make it an H3 under Arguments or as an H2 after Blocks, before Exported fields.

There's a info gap here. It doesn't say what happens. It's vague about the failure. Does the endpoint get skipped but the scrape continues? Or does it fail completely? Do they lose one target or the whole pipeline?  Where does the warning show up? If someone is trying to understand why the targets disappeared, they need some direction.

What happens if you set this new arg to `false`? Does opting out silence the warning log?

This is breaking behaviour, do we document how to deal with the breakage? (changelog? WhatsNew? here in the docs?)

docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
```diff
@@ -37,12 +44,13 @@ prometheus.operator.servicemonitors "<LABEL>" {
...
-| `kubernetes_role`       | `string`                | The Kubernetes role used for discovery. Supports `endpoints` or `endpointslice`.                          | `"endpoints"` | no       |
-| `namespaces`            | `list(string)`          | List of namespaces to search for ServiceMonitor resources. If not specified, all namespaces are searched. |               | no       |
+| Name                               | Type                    | Description                                                                                               | Default       | Required |
+| ---------------------------------- | ----------------------- | --------------------------------------------------------------------------------------------------------- | ------------- | -------- |
+| `disallow_arbitrary_file_access`   | `bool`                  | Disallow ServiceMonitor endpoints that reference arbitrary files on the {{< param "PRODUCT_NAME" >}} filesystem. | `true`        | no       |
+| `forward_to`                       | `list(MetricsReceiver)` | List of receivers to send scraped metrics to.                                                             |               | yes      |
```
clayton-cornell:
```suggestion
| `forward_to`                       | `list(MetricsReceiver)` | List of receivers to send scraped metrics to.                                                             |               | yes      |
| `disallow_arbitrary_file_access`   | `bool`                  | Disallow ServiceMonitor endpoints that reference arbitrary files on the {{< param "PRODUCT_NAME" >}} filesystem. | `true`        | no       |
```

Optional should be below required

docs/sources/reference/components/prometheus/prometheus.operator.servicemonitors.md
jharvey10: Thanks! Updated this for clarity. The `BREAKING CHANGE` block in the PR description will trigger an entry at the top of the changelog for this release once merged.

clayton-cornell:
Doc part looks good

internal/component/prometheus/operator/common/component_test.go
```diff
@@ -58,6 +58,20 @@ func (c *crdManagerHungRun) GetScrapeConfig(ns, name string) []*config.ScrapeCon
 	return nil
 }
 
+func TestNewValidatesOptions(t *testing.T) {
```
blewis12: nit: rename this to just `TestValidation` or so, since the logic is only "new" now

internal/component/prometheus/operator/common/component.go
```diff
@@ -30,11 +30,45 @@ type Component struct {
...
+}
+
+type Options struct {
+	Kind                   string
+	ServiceMonitorSettings *ServiceMonitorSettings
+}
+
+func DefaultOptions(kind string) Options {
+	return Options{
+		Kind: kind,
```
blewis12: nit: explicitly defining service monitor settings as `nil` here makes it more obvious that _that_ is the default value, that's my personal preference for readability though

internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -135,19 +139,27 @@ func newCrdManager(opts component.Options, cluster cluster.Cluster, logger *slog
 	default:
 		panic(fmt.Sprintf("Unknown kind for crdManager: %s", kind))
 	}
+	serviceMonitorSettings := (*ServiceMonitorSettings)(nil)
```
blewis12: nit: can we rename this to `defaultServiceMonitorSettings` since that's the intention of this function?

internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -595,16 +617,20 @@ func (c *crdManager) addServiceMonitor(sm *promopv1.ServiceMonitor) {
...
 		c.addDebugInfo(sm.Namespace, sm.Name, err)
 		return
 	}
 	c.mut.Lock()
+	for k, v := range discoveryConfigs {
+		c.discoveryConfigs[k] = v
+	}
+	for k, v := range scrapeConfigs {
+		c.scrapeConfigs[k] = v
+	}
```
blewis12, line 635: IIUC this is a bug fix ensuring that we don't have partial scrape config, probably something we need to add to the PR description too just to be clear

blewis12: Also this same bug exists in [addPodMonitor](https://github.com/grafana/alloy/blob/main/internal/component/prometheus/operator/common/crdmanager.go#L542-L543) but I vote to address this in a follow up PR to avoid cluttering this one

internal/component/prometheus/operator/common/crdmanager.go
```diff
@@ -603,12 +603,14 @@ func (c *crdManager) addServiceMonitor(sm *promopv1.ServiceMonitor) {
 		ScrapeOptions:               c.args.Scrape,
 	}
 
+	for i, ep := range sm.Spec.Endpoints {
+		c.observeServiceMonitorArbitraryFileAccess(sm, i, ep)
+	}
```
blewis12, line 608: I moved this here so that we're actually logging all endpoints that are attempting to use arbitrary filesystem access - previously only the first would error and then we'd bail out of the loop after `GenerateServiceMonitorConfig`

internal/component/prometheus/operator/configgen/config_gen_servicemonitor.go
```diff
@@ -21,6 +21,12 @@ func (cg *ConfigGenerator) GenerateServiceMonitorConfig(m *promopv1.ServiceMonit
 	cfg = cg.generateDefaultScrapeConfig()
 
 	cfg.JobName = fmt.Sprintf("serviceMonitor/%s/%s/%d", m.Namespace, m.Name, i)
+	if cg.DisallowArbitraryFileAccess {
+		if field := ServiceMonitorEndpointArbitraryFileField(ep); field != "" {
+			return nil, fmt.Errorf("serviceMonitor %s/%s endpoint %d uses %s, which is disallowed by disallow_arbitrary_file_access; use secret or authorization fields instead", m.Namespace, m.Name, i, field)
```
dehaansa: Is this error clear enough? Reading it without the context of a bad config it doesn't seem very clear how "use" the other fields solves the problem.

blewis12: yeah that's true, maybe we could have it be something that looks like 

```
return nil, fmt.Errorf("serviceMonitor %s/%s endpoint %d uses %s, which is disallowed because allow_arbitrary_file_access is false; use %s instead", m.Namespace, m.Name, i, field, arbitraryFileFieldReplacement(field)
```

where arbitraryFileFieldReplacement does the mapping between the arbitrary file `field` and the proper ServiceMonitor field to be used instead?

blewis12: https://github.com/grafana/alloy/pull/6885/commits/858093270e9786b24cd46b16078e944c1368950a

dehaansa (approved):

clayton-cornell (approved):
Approving for docs

PR #6889, opened 2026-08-14
feat: Part 3 - Expose knobs for OTel v0.158.0 upgrade - add exposed_headers to otel http config

blewis12:
Adds the `exposed_headers` setting to `http.cors` otel config so users can set the [Access-Control-Expose-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) response header, indicating which headers are safe to expose to the API of a CORS response.

### Pull Request Details
This config is present in a couple different components so is spread across

### PR Checklist
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/reference/components/otelcol/otelcol.receiver.influxdb.md
+++ b/docs/sources/reference/components/otelcol/otelcol.receiver.influxdb.md
@@ -87,11 +87,12 @@ The `cors` block configures CORS settings for an HTTP server.
 
 The following arguments are supported:
 
-| Name              | Type           | Description                              | Default                | Required |
-|-------------------|----------------|------------------------------------------|------------------------|----------|
-| `allowed_origins` | `list(string)` | Allowed values for the `Origin` header.  |                        | no       |
-| `allowed_headers` | `list(string)` | Accepted headers from CORS requests.     | `["X-Requested-With"]` | no       |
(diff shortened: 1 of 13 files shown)


docs/sources/reference/components/otelcol/otelcol.receiver.influxdb.md
```diff
@@ -87,11 +87,12 @@ The `cors` block configures CORS settings for an HTTP server.
 
 The following arguments are supported:
 
-| Name              | Type           | Description                              | Default                | Required |
-|-------------------|----------------|------------------------------------------|------------------------|----------|
-| `allowed_origins` | `list(string)` | Allowed values for the `Origin` header.  |                        | no       |
-| `allowed_headers` | `list(string)` | Accepted headers from CORS requests.     | `["X-Requested-With"]` | no       |
-| `max_age`         | `number`       | Configures the `Access-Control-Max-Age`. |                        | no       |
+| Name              | Type           | Description                                                     | Default                | Required |
```
kalleep, line 90: We should probably have shared doc blocks that can be reused, not for this pr but something we can consider

kalleep (approved):

docs/sources/reference/components/otelcol/otelcol.receiver.influxdb.md
blewis12: yeah I agree 👍

PR #967, opened 2024-05-30
feat(prom/exp/snmp): add support to pass multiple SNMP config files to `prometheus.exporter.snmp`

hainenber:
#### Which issue(s) this PR fixes

Closes #916 

#### Notes to the Reviewer

I decided not to add another test case into `snmp_test.go` as it doesn't test whole component with argument but rather just a unit test for `LoadSNMP` function. Incoming integration test in #954 will address this.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Config converters updated

Issue #916, opened 2024-05-23
prometheus.exporter.snmp to support multiple config files


ptodev:
### Request

The `config_file` (and maybe `config`) options can be changed from `string` to `list(string)`, so that users can specify more than one config file. This is a feature which was [added upstream](https://github.com/prometheus/snmp_exporter/pull/970) some time ago.

This was requested by @bastischubert 

### Use case

It could be helpful to split out a config file into multiple files.

rfratto:
Is there a way that this can be done without a breaking change, like a new `config_files` argument that becomes mutually exclusive with the `config_file` argument?

hainenber:
That's also more intuitive, imo. From user perspective, it'd be pretty weird passing a list of SNMP config files to `config_file` rather `config_files`

Or maybe we can mark `config_file` as deprecated and enforce people passing single-entry list of SNMP config file to `config_files`? 🤔

ptodev:
> Is there a way that this can be done without a breaking change, like a new config_files argument that becomes mutually exclusive with the config_file argument?

I would be happy with that. I also think we should document the `config_file` argument as "deprecated", as @hainenber  suggested. Having both arguments side by side will give people time to migrate prior to 2.0.

> From user perspective, it'd be pretty weird passing a list of SNMP config files to config_file rather config_files

True! It was sort of implied in the original issue above, but I didn't mention it. Thanks for clearing it. up :)

Btw I don't think this is super high priority, because in theory `config_file` accepts a glob pattern. So even with the current config it should be possible for folks to use multiple files. I haven't tested it though. Hopefully we can use the integration tests in #954 for this purpose in the future.

colinedwardwood:
I'd like this feature for my customers. They want parity with the upstream and it just makes life a lot easier for them to maintain configs separately based on all kinds of things like manufacturers and auth.

LarsStegman:
@jharvey10 
@blewis12 

I want to spend some time on getting this PR back in sync with main to make a PR. Do you think the team will have time to review it?

https://github.com/grafana/alloy/pull/967

jharvey10:
@LarsStegman Apologies for the delayed response. A few of us have some time off here during the summer. I can't promise a timeline for review, but I will say that we definitely have more team bandwidth at the moment than we did ~6-12 months ago and have been reviewing PRs at a pretty decent pace.

LarsStegman:
I have created the PR for this.

I also created another issue that I think might be nice to have to properly handle secrets: #6893

PR #6892, opened 2026-08-14
feat(prometheus.snmp_exporter): Support multiple config files

LarsStegman:

Add `config_files` option to read multiple config files

### Pull Request Details

- Add `config_files` option to read multiple config files.
- "Hide" `config_file` option by not documenting it, so people will migrate over time. 
- Update static converter to handle the new `config_files` option properly

If both `config_files` and `config_file` are specified, an error is given to the user. These options are mutually exclusive.

Based on #967 

#916 

### Notes to the Reviewer

Please check whether I updated all required code areas. I might have missed some.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -45,15 +45,18 @@ prometheus.exporter.snmp "<LABEL>" {
...
 | `config`                | `string` or `secret` | SNMP configuration as inline string.                                                                                         |             | no       |
 | `targets`               | `list(map(string))`  | SNMP targets.                                                                                                                |             | no       |
 
-The `config_file` argument points to a YAML file defining which snmp_exporter modules to use.
-Refer to [snmp_exporter](https://github.com/prometheus/snmp_exporter/tree/{{< param "SNMP_VERSION" >}}?tab=readme-ov-file#configuration) for details on how to generate a configuration file.
+The `config_files` argument points to YAML files defining which snmp_exporter
+modules to use. Refer to [snmp_exporter]
+(https://github.com/prometheus/snmp_exporter/tree/{{< param "SNMP_VERSION" >
+}}?tab=readme-ov-file#configuration) for details on how to generate
+configuration files.
```
clayton-cornell:
```suggestion
The `config_files` argument points to YAML files that define which `snmp_exporter` modules to use.
Refer to [Configuration](https://github.com/prometheus/snmp_exporter/tree/{{< param "SNMP_VERSION" >}}?tab=readme-ov-file#configuration) in the Prometheus `snmp_exporter` documentation for configuration-generation details.
```

Semantic line breaks at sentence level instead of arbitrary character counts - ensures that weird spaces etc don't sneak into URLs and makes it easier to review.

LarsStegman:
@clayton-cornell I've incorporated your feedback. 

I seem to be having trouble with the cla-assistant, but there are more people who are having trouble.

LarsStegman:
@clayton-cornell do you have time to take a look to get this merged?

kalleep:
>"Hide" config_file option by not documenting it, so people will migrate over time.

Instead of hiding it we mark it as deprecated, it will be confusing for existing users of this component that we no longer have this attribute documented.

clayton-cornell:
> > "Hide" config_file option by not documenting it, so people will migrate over time.
> 
> Instead of hiding it we mark it as deprecated, it will be confusing for existing users of this component that we no longer have this attribute documented.

Ah I didn't catch that detail in the description.  This means that the attribute row for `config_file` should be restored and the description updated to `Deprecated. Use `config_files` instead.`

There also must be a line added to the text below the table stating that you can't config both, and that if you do, the error that will be shown/given.

LarsStegman:
Feedback incorporated.

kalleep (approved):
LGTM but will wait for @clayton-cornell to do another pass

docs/sources/reference/components/prometheus/prometheus.exporter.snmp.md
```diff
@@ -44,16 +44,19 @@ prometheus.exporter.snmp "<LABEL>" {
...
-| `targets`               | `list(map(string))`  | SNMP targets.                                                                                                                |             | no       |
+| Name                    | Type                 | Description                                                                                                                      | Default     | Required |
+|-------------------------|----------------------|----------------------------------------------------------------------------------------------------------------------------------|-------------|----------|
+| `concurrency`           | `int`                | SNMP exporter concurrency.                                                                                                       | `1`         | no       |
+| `config_file`           | `string`             | SNMP configuration file defining custom modules. Deprecated. Use `config_files` instead. Mutually exclusive with `config_files`. |             | no       |
```
clayton-cornell, line 50:
```suggestion
| `config_file`           | `string`             | Deprecated. Use `config_files` instead.                                                                                         |             | no       |
```

We should be able to drop the old description and just use the deprecation note.

clayton-cornell (approved):
Looks good. One minor nit that should be fixed, but non-blocking.

PR #6905, opened 2026-08-17
fix(beyla): Unix socket name could be too long

skl:

Beyla subprocess Unix socket name must not exceed sun_path length limit.

### Pull Request Details

When Alloy creates a Unix socket for the purposes of communicating with the Beyla subprocess, the name of the socket must not be greater than the limit:

> The sun_family field always contains AF_UNIX.  On Linux, sun_path is 108 bytes in size; see also BUGS, below.
> https://man7.org/linux/man-pages/man7/unix.7.html

Previously, `componentID` had a variable length, causing in some cases the `sun_path` length limit to be breached, which resulted in socket creation failure.

To prevent this from happening, while keeping the socket name unique enough to support many different sockets, we take the first 16 bytes of a hash of the `componentID`.

This was raised in internal [slack](https://raintank-corp.slack.com/archives/C0903J6CUSV/p1785854543938309?thread_ts=1785423027.291409&cid=C0903J6CUSV).

### Notes to the Reviewer

See details above. Verified on an Ubuntu ARM VM locally (OrbStack Machine).

### PR Checklist

- [X] Tests updated
- [X] This pull request was substantially generated with AI assistance (see the [GenAI policy](https://github.com/grafana/alloy/blob/main/docs/developer/genai.md))

### DIFF
--- a/internal/component/beyla/ebpf/exec_linux.go
+++ b/internal/component/beyla/ebpf/exec_linux.go
@@ -4,6 +4,7 @@ package beyla
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"log/slog"
 	"net"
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

Issue #6835, opened 2026-08-06
loki: Use labels.Labels for entries instead of model.LabelSet


kalleep:
### Component(s)

Loki components

### Background

Today `loki.Entry` stores [model.LabelSet](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/common/loki/entry.go#L38) as its representation for labels, so that entries can be mutated as they move through the pipeline. But almost nothing in the pipeline actually wants a map.

Most of our `loki.source.*` components support relabeling and convert between `labels.Labels` and `model.LabelSet`, e.g. [loki.source.api](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/loki/source/api/routes.go#L95-L116).

`loki.relabel` has this exact same pattern, [converting in both directions](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/loki/relabel/relabel.go#L314-L329) on every cache miss.

`loki.write` [needs to sort](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/common/loki/client/batch.go#L168-L198) these labels in another slice to properly build the required label string before sending to Loki.

Our WAL implementation will convert to `labels.Labels` [when writing](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/common/loki/wal/writer.go#L261) and back to `model.LabelSet` [after reading](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/common/loki/client/consumer_wal.go#L190-L192).

`loki.process` [prematurely clones](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/loki/process/process.go#L191-L195) the entry because mutating it in place would be considered unsafe since we can fan out to multiple components. There is already a `TODO` on that line proposing copy-on-write so the copy only happens on the first stage that needs to modify labels.

And with the new batch structure we want to use for the pipeline changes described in https://github.com/grafana/alloy/issues/4953 we could have some potentially [expensive clones / comparisons](https://github.com/grafana/alloy/blob/1fef867c159db775d0758e55b8d2d63e87bc1cff/internal/component/common/loki/batch.go#L76-L102).

[...]

PR #6913, opened 2026-08-18
chore(loki.process): Introduce new pipeline structure and entryProcess interface

kalleep:
This implements pipeline4 from https://github.com/grafana/alloy/pull/6872. We want to migrate away from the channels + goroutine setup our `loki.process` have today.

### Pull Request Details

In this pr the new pipeline primities and interfaces are introduces. On top of that I have implemented the new `stage` interface for `stage.cri`. Since this new pipeline will be called concurrently we need to make sure it's also safe for concurrent use.

We have experimented a lot with different solutions:
1. Single mutex - Simplest but will serialize calls (only one batch can be processed at one time) 
2. Sharded state - Idea was to be able to pass a "shard" id and construct a sharded state, no mutex required but it also forces all stages to know that i's sharded.
3. Local state with single mutex - The idea was to only touch the state once for a stream in each batch, works pretty well but is something we can implement at a later point if we want to.
4. Existing concurrent map implementation - Became really complicated since we still had to use a RWLock on top of it.
5. Striped map - Experimented with in https://github.com/grafana/alloy/pull/7025. 

I have gone with the striped map implementation for partialLines, that one is only used when new pipeline is used, otherwise the regular map implementation is used.

`TestCRI` and `TestCRI_tags` are collapsed into the same test `TestCRIStage`, with all cases preserved. For this I added a new helper function `runPipelineTest` that will run passed stages with both the old pipeline and the one I introduced in this pr and assert that we get expected entries.

For the new pipeline I also implemented `stopper`. This will allow cri stage to flush buffered partial lines when done, something that we don't do today.

For benchmark I introduced a similar helper to run benchmarks using both pipelines, this helper will create a "worker" go routine for each batch passed and concurrently call the pipeline.

Result for running benchmark for CRI
```
goos: linux
goarch: amd64
BenchmarkCRIStage/single_stream/Pipeline-16         	  779700	      1405 ns/op	     888 B/op	      11 allocs/op
BenchmarkCRIStage/single_stream/New_Pipeline-16     	 1350559	       885.4 ns/op	    1144 B/op	      15 allocs/op
BenchmarkCRIStage/multiple_streams/Pipeline-16      	   16069	     74275 ns/op	   42720 B/op	     600 allocs/op
BenchmarkCRIStage/multiple_streams/New_Pipeline-16  	  101698	     11894 ns/op	   55680 B/op	     559 allocs/op
BenchmarkCRIStage/flush_pressure/Pipeline-16        	   49099	     24991 ns/op	   13697 B/op	     190 allocs/op
BenchmarkCRIStage/flush_pressure/New_Pipeline-16    	  339879	      3648 ns/op	   16355 B/op	     187 allocs/op
```

### Notes to the Reviewer

I renamed cri struct and constructor function to make them match other stages and no longer export them.

Asuming all downstream entries are going through a shared `ShardingConsumer` there is still one case that could cause OOO for entires, when `max_partial_lines` overflows. Since the batch that sees this first will collected them and forward within the batch they could race against other concurrent batches that have entries belonging to the same stream.
### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/loki/process/stages/pipeline_test.go
```diff
@@ -56,6 +62,184 @@ func newPipelineFromConfig(cfg string) (*Pipeline, error) {
...
+func runPipelineTest(t *testing.T, cfgs []StageConfig, entries []Entry, expected []Entry, expectedMetrics string) {
+	// Pipeline.Run seeds the extracted map with each entry's initial labels
+	// before running any stage. process (called directly below, bypassing
+	// ProcessBatch/ProcessEntry) does not. Seed it here once so both
+	// pipeline implementations start from the same state.
+	for i := range entries {
+		for labelName, labelValue := range entries[i].Labels {
+			entries[i].Extracted[string(labelName)] = string(labelValue)
+		}
+	}
```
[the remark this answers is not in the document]
kalleep: Not relevant and all tests are updated in other pr:s

internal/component/loki/process/stages/pipeline_test.go
```diff
@@ -56,6 +62,184 @@ func newPipelineFromConfig(cfg string) (*Pipeline, error) {
...
+		var collected []Entry
+		next := func(_ context.Context, entries []Entry) error {
+			collected = append(collected, entries...)
+			return nil
+		}
+
+		p, err := NewPipeline2(logging.NewSlogNop(), registry, featuregate.StabilityGenerallyAvailable, cfgs, next)
+		require.NoError(t, err)
+
+		p.process(context.Background(), entries)
```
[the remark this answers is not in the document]
kalleep: Relevant but fix in next pr

internal/component/loki/process/stages/cri.go
```diff
@@ -173,7 +183,131 @@ func (c *cri) Run(in chan Entry) chan Entry {
...
+				c.logger.Warn("partial lines upperbound exceeded, merging it to single line", "threshold", c.cfg.MaxPartialLines)
+				if c.partialLinesFlushedMetric != nil {
+					c.partialLinesFlushedMetric.Add(float64(len(c.partialLines)))
+				}
+
+				// Add existing partialLines
+				for _, v := range c.partialLines {
+					extra = append(extra, v)
+				}
+
```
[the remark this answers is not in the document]
kalleep: fixed in [950616f](https://github.com/grafana/alloy/pull/6913/commits/950616fc2e29b769ff21d35a69de635fb147984a)

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/grafana/alloy/internal/component/common/loki"
+	"github.com/grafana/alloy/internal/featuregate"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// NextFn forwards a batch of entries to whatever comes next in a pipeline.
+type NextFn func(ctx context.Context, entries []Entry) error
+
+// stage is a single step in a pipeline.
+type stage interface {
```
thampiotr: Can we call this `entryProcessor` instead? Makes more sense given the method in it.

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+	// process runs the stage's work on entries and forwards the result.
```
thampiotr:
```suggestion
	// process performs work on entries. The result is not returned. Typically the implementations call a `NextFn` configured at construction time. Errors are propagated.
```

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+	return &Pipeline2{
+		next:   next,
+		stages: stages,
+	}, nil
+}
+
+// ProcessBatch runs every entry in batch through the pipeline.
+func (p *Pipeline2) ProcessBatch(ctx context.Context, batch loki.Batch) error {
+	entries := make([]Entry, 0, batch.EntryLen())
+	_ = batch.ConsumeStreams(func(stream loki.Stream, created int64) error {
```
thampiotr: I see that the error returned would always be `nil` but having to check that is a cognitive load on the reader, so let's check the error and return it, there's nothing preventing us from it. It should never happen so dead code, but it will be easier to read as there's no need to be verifying 'is this a bug?'

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+		}
+		return nil
+	})
+
+	return p.process(ctx, entries)
+}
+
+// ProcessEntry runs a single entry through the pipeline.
+func (p *Pipeline2) ProcessEntry(ctx context.Context, entry loki.Entry) error {
```
thampiotr: Is this even used? When would we need this instead of `ProcessBatch`?

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+	return p.process(ctx, []Entry{
+		{
+			Extracted: extracted,
+			Entry:     entry,
+		},
+	})
+}
+
+func (p *Pipeline2) process(ctx context.Context, entries []Entry) error {
+	return p.next(ctx, entries)
```
thampiotr, line 147: This function looks pointless. Do we need it?

internal/component/loki/process/stages/cri.go
```diff
@@ -173,7 +183,133 @@ func (c *cri) Run(in chan Entry) chan Entry {
 	})
 }
 
-func (c *cri) ensureTruncateIfRequired(e *Entry) {
+// process implements stage and is only used by our new pipeline.
+func (c *criStage) process(ctx context.Context, entries []Entry) error {
```
thampiotr, line 130: That function is a bit too large, making the code rather hard to read.

internal/component/loki/process/stages/cri.go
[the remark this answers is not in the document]
thampiotr: I think that the change has flipped the order the other way, but didn't fix it. We can still get out-of-order within the same stream. 

Instead of adding the partialLines to `extra` when `len(c.partialLines) >= c.cfg.MaxPartialLines`, we could instead append them directly to `out` - this way they get flushed when MaxPartialLines is reached at exact place where that limit is reached. 

This also exposes a gap in testing. I would like to see a test case for this:
```
# with max_partial_lines = 1 and following log lines:
{s="1"} F FULL-FIRST
{s="1"} P PARTIAL-SECOND
{s="2"} P other-stream      <- trips the limit, flushes stream s=1
# output 
out[0] {s="1", stream="stdout"} "PARTIAL-SECOND "
out[1] {s="1", stream="stdout"} "FULL-FIRST"
```

internal/component/loki/process/stages/cri.go
```diff
@@ -173,7 +183,133 @@ func (c *cri) Run(in chan Entry) chan Entry {
...
+			e.Line = builder.String()
+			c.ensureTruncateIfRequired(&e)
+			delete(c.partialLines, fingerprint)
+		}
+
+		entries[dst] = e
+		dst++
+	}
+
+	c.mut.Unlock()
```
thampiotr: This makes me wonder about the concurrency model. We should document what do we intend. I'm not 100% sure myself. 

Right now we have:
- `ProcessBatch` can be called concurrently via sharding consumers
- `ProcessBatch` serialises all the streams in a batch into one array of entries and processes them in that order
- But some stages like `cri` are sharing state between calls to `ProcessBatch` - such as the partial lines count. 
- I think that current `cri` implementation will result in another bug: two calls to `ProcessBatch` concurrently. The first one creates some partial lines. The second one triggers the flush due to max lines limit. This results in second call outputting lines from the streams from the first call. And they can come out out-of-order. Again I think we would want to have some tests for this.

In parallel processing this shared limit in `cri` has some tradeoffs:
- more precise to limit the memory
- but unpredictable - same input may hit it sometimes and other times not
- adding comlexity to the code

Maybe we should document that in the new, concurrent Loki pipelines, this limit is per-shard. And implement it that way? Each shard would need to have its own limit 🤔 Maybe we need to create a Pipeline2 per each shard?

I'm not sure how other stateful stages need to be handled in this concurrency model - something to think about.

internal/component/loki/process/stages/pipeline2.go
kalleep: Yes we do need it for [match](https://github.com/grafana/alloy/blob/249c3da262a95dd6f64f19579f262afa217cda99/internal/component/loki/process/stages/match.go#L284).

internal/component/loki/process/stages/pipeline2.go
kalleep: I should probably have commented on this one. Since we start with single entry and not batches in the beginning I added this entry point so we don't have to allocate a temporary batch just to call into the pipeline.

internal/component/loki/process/stages/pipeline2.go
kalleep: Sure I am fine with that

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// ProcessBatch runs every entry in batch through the pipeline.
+func (p *Pipeline2) ProcessBatch(ctx context.Context, batch loki.Batch) error {
+	entries := make([]Entry, 0, batch.EntryLen())
+	_ = batch.ConsumeStreams(func(stream loki.Stream, created int64) error {
+		extracted := make(map[string]any, len(stream.Labels))
+		for k, v := range stream.Labels {
+			extracted[string(k)] = string(v)
+		}
```
thampiotr, line 68: In the future, we could maybe avoid / materialise this extracted map only when it is used - make it lazy? something to think about, but I wouldn't want to do this now. Maybe just leave a comment.

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,127 @@
...
+		for i, e := range stream.Entries {
+			if i == len(stream.Entries)-1 {
+				entries = append(entries, Entry{
+					Extracted: extracted,
+					Entry:     loki.NewEntryWithCreatedUnixMicro(stream.Labels, created, e),
+				})
+			} else {
+				entries = append(entries, Entry{
+					Extracted: maps.Clone(extracted),
+					Entry:     loki.NewEntryWithCreatedUnixMicro(stream.Labels.Clone(), created, e),
```
thampiotr: Also here with `stream.Labels.Clone()` - maybe we don't need to clone until there is a need to modify / mutate them. That would require some extra abstraction. Fine to just leave a comment for the future.

kalleep: That is what this proposal is for https://github.com/grafana/alloy/issues/6835

internal/component/loki/process/stages/pipeline2.go
kalleep: Yes that is a good idea

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,159 @@
...
+		return p.inner.process(ctx, entries)
+	})
+}
+
+func (p *PipelineConsumer) Stop() {
+	p.inner.stop()
+}
+
+func (p *PipelineConsumer) collect(ctx context.Context, entries []Entry) error {
+	// FIXME(kallep): Currently this would set time.Now as creation time for a batch.
```
kalleep, line 94: Currently we store create in batch, but I think it would be more correct to store it for each stream instead.

And then when we assemble the batch again we set creation time to the lowest one for each stream.

I left it here as a fixme since I would want to do that change in a follow up pr

internal/component/loki/process/stages/pipeline2.go
kalleep: Added a comment

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -0,0 +1,159 @@
...
+	batch := loki.NewBatch()
+	for _, e := range entries {
+		batch.AddEntry(e.Labels, e.Entry.Entry)
```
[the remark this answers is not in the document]
kalleep: That is fine for now and the comment hints about this. I want to go back and fix this later

internal/component/loki/process/stages/pipeline2.go
thampiotr: How is the creation time of a batch used? what is the impact of it being not exactly right?

kalleep: Its used to track `loki_write_entry_propagation_latency_seconds`.

thampiotr: I think it's fine to keep it on a batch level. Wouldn't rename the metric, but we can change the help comment of this metric metadata to say that it is the upper bound of latency of entries in a batch. It is still useful and simpler to maintain.

internal/component/loki/process/stages/state.go
```diff
@@ -0,0 +1,179 @@
+package stages
```
thampiotr: Can we put this in an internal package? It seems like we have a `loki.process.stage` otherwise ;)

internal/component/loki/process/stages/state.go
```diff
@@ -0,0 +1,179 @@
...
+
+import (
+	"sync"
+
+	"go.uber.org/atomic"
+)
+
+func newSharedState[T any](sizeHint int, concurrentSafe bool) sharedState[T] {
```
thampiotr: I personally agree with the articles like this: https://dev.to/rweisleder/the-flag-parameter-anti-pattern-1j82 
Would be best to avoid the `bool` parameters entirely. 

These are two implementations, can we have them:
1. In their own files
2. have the caller do the check what implementation it needs and call that using the correct `new*` method.
3. have their own tests - this makes the files smaller and more focuessed
4. can you also drop a comment about thread safety of these structures so it's all easier to read?

internal/component/loki/process/stages/cri.go
```diff
@@ -94,92 +97,166 @@ const (
...
+
+	if len(out) == 0 {
+		return nil
+	}
+
+	return c.next(ctx, out)
+}
+
+func (c *criStage) addPartialLine(fp model.Fingerprint, e Entry) {
+	c.partialLines.Update(uint64(fp), func(prev Entry) Entry {
```
thampiotr: Because `func` used in `Update`, the entry escapes to the heap and causes allocation. I think it's fine to have a more specific method like `partialLines.UpdateLine` instead?

internal/component/loki/process/stages/state.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// Update runs fn under key's stripe lock with the current value stored at
+// key, or the zero value of V if key is not present, and stores fn's result.
+func (m *stripedMapState[T]) Update(key uint64, fn func(v T) T) {
+	s := m.stripe(key)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	_, existed := s.data[key]
```
thampiotr: You don't need to throw away the value with `_`, because you use it right away below when calling fn.

internal/component/loki/process/stages/cri.go
```diff
@@ -94,92 +97,166 @@ const (
...
+	}
+}
+
+func (c *criStage) ensureTruncateIfRequired(prevLine, newLine string) string {
+	if !c.cfg.MaxPartialLineSizeTruncate {
+		return prevLine + newLine
+	}
+
+	// If prev line is already at max size we don't have to concatenate new line.
+	if len(prevLine) == int(c.cfg.MaxPartialLineSize) {
```
thampiotr: To prevent any bugs, this should be a `>=`, even if it's correct now, just for future issues.

internal/component/loki/process/stages/cri.go
```diff
@@ -94,92 +97,166 @@ const (
...
+		if c.linesTruncatedMetric != nil {
```
thampiotr: Do we need this nil check? I don't see it done consistently everywhere.

internal/component/loki/process/stages/cri.go
```diff
@@ -94,92 +97,166 @@ const (
...
+			// We got full-line 'F'.
+			entries = []Entry{c.completeFullLine(fingerprint, e)}
 		}
 
-		e.Labels[criStream] = model.LabelValue(parsed.Stream.String())
+		if extra := c.flushPartialLinesIfExceeded(); len(extra) > 0 {
+			entries = append(entries, extra...)
+		}
+
+		return entries, len(entries) == 0
```
thampiotr: previously we returned `entries, false` always. Can't we go back to this?

internal/component/loki/process/stages/cri.go
```diff
@@ -94,92 +97,166 @@ const (
...
+		if parsed.Flag == crip.FlagPartial {
+			// it's a partial-line buffer it and move on.
+			c.addPartialLine(fingerprint, e)
+		} else {
+			// We got full-line 'F'.
+			entries = []Entry{c.completeFullLine(fingerprint, e)}
 		}
 
-		e.Labels[criStream] = model.LabelValue(parsed.Stream.String())
+		if extra := c.flushPartialLinesIfExceeded(); len(extra) > 0 {
```
thampiotr: I think you already noted that this is change in behaviour... let's go back to as little changes to the current cri implementation as possible.

internal/component/loki/process/stages/state.go
```diff
@@ -0,0 +1,179 @@
...
+func newSharedState[T any](sizeHint int, concurrentSafe bool) sharedState[T] {
+	if concurrentSafe {
+		return newStripedMapState[T](sizeHint)
+	}
+	return newMapState[T](sizeHint)
+}
+
+type sharedState[T any] interface {
+	Take(key uint64) (T, bool)
+	Update(key uint64, fn func(v T) T)
```
thampiotr: I know you are trying to prepare this for reuse in `multiline`, but I think it's becoming a bit too early. It's not possible to see the full shape yet. So maybe let's make this the simplest implementation needed for `cri` and when we get to `multiline` we will create the necessary abstractions and benchmark them to be sure nothing regressed, and this way we will have the smallest necessary abstraction for what we need.

internal/component/loki/process/stages/cri.go
kalleep: We had it before so I left it, but we do don't need them

internal/component/loki/process/stages/cri.go
kalleep: Sure I reverted to old behavior here, not sure why `RunWithSkipOrSendMany` also have a bool since an empty slice or nil slice would not forward anything anyway.

thampiotr (approved):
🎉 thanks for the patience iterating on this. Nice work, this was a tough one :)

PR #6914, opened 2026-08-18
chore(loki.process): Implement new stage interface and update tests for simpler stages

kalleep:

This will migare all "simple" stages to implement the new `entryProcessor` and optionally the new `stopper` interfaces.

I know this is a large pr but I think most of the changes should be pretty straight forward since they all follow the same pattern.  

### Pull Request Details

All stages that is updated in this pr is simple processing stages, stages that dont really hold state. There is a lot of changes since it touches almost all stages we have.

At a component level nothing have changed yet, these stages are just changed so that they implement the old and the new pipeline interface. To verify that they work properly almost all test had to be rewritten to use `runPipelineTest` and alternative `runPipelineBenchmark`. These test helpers accepts config for several stages and run one test for each pipeline and assert that entries in matches entries out for both of them. `runPipelineTest` can also verify expected metrics after a run.

Generally every test for a stage is now ran as a table test e.g. `TestRegexStage` and it should contain all other tests we had before. There are some exception for e.g. `stage.limit` where we need to test a specific behavior such as stopping while blocked and not just proccessing.

Because `runPipelineTest` now assert on everything for one entry (`extracted`, `labels`, `line`, `structured_metadata` and `time`) we now verify much more stuff on each test than we did before. Some stages just verified lines while others only verified labels etc.

There is also some general cleanup such as no longer exporting a bunch of errors and removal of unused variables. I also renamed config and stage in `labels_keep` from allow to keep to match what we have documented and use in other places.

I did the implementation for every stage my self. But for test I did the first couple of stages so I got a general pattern and then used AI to help me convert the rest as I moved through stages.

Part of: #4953

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

---


internal/component/loki/process/stages/timestamp.go
```diff
@@ -157,87 +157,103 @@ type timestampStage struct {
...
 			if parsedTs.Equal(entry.lastParsed) {
-				*t = entry.lastAdjusted.Add(1 * time.Nanosecond)
+				e.Timestamp = entry.lastAdjusted.Add(1 * time.Nanosecond)
 			}
 		}
 	}
-	if (ts.config.ActionOnFailure == TimestampActionOnFailureFudge || ts.config.ActionOnDuplicateTimestamp == TimestampActionOnDuplicateTimestampFudge) && ts.lastKnownTimestamps != nil {
-		ts.lastKnownTimestamps.Add(labelsStr, timestampCacheEntry{lastParsed: *parsedTs, lastAdjusted: *t})
+	if (ts.config.ActionOnFailure == timestampActionOnFailureFudge || ts.config.ActionOnDuplicateTimestamp == timestampActionOnDuplicateTimestampFudge) && ts.lastKnownTimestamps != nil {
+		ts.lastKnownTimestamps.Add(labelsStr, timestampCacheEntry{lastParsed: *parsedTs, lastAdjusted: e.Timestamp})
```
thampiotr: Make timestamp cache updates atomic: concurrent calls with identical label sets can read the same cached value and emit the same fudged timestamp. Please use one stage mutex around the complete lookup → adjustment → update sequence, covering both duplicate-fudge and failure-fudge paths. Keep timestamp parsing and downstream processing outside the lock.

internal/component/loki/process/stages/pattern.go
```diff
@@ -52,80 +50,87 @@ func validatePatternConfig(c PatternConfig) (*pattern.Matcher, error) {
...
 	}
 
 	names := r.matcher.Names()[:len(matches)]
 	for i, m := range matches {
 		name := names[i]
-		extracted[name] = string(m)
+		e.Extracted[name] = string(m)
 		if r.config.LabelsFromGroups {
 			labelName := model.LabelName(name)
 			labelValue := model.LabelValue(m)
```
thampiotr: Protect the stateful pattern matcher: the matcher reuses capture storage, so concurrent calls can overwrite results that another call is still reading. Please protect both matching and consuming the captures (through the extracted values and labels), or use independent matcher instances. A concurrent test with distinct input values would catch cross-contamination.

internal/component/loki/process/stages/limit.go
```diff
@@ -112,13 +122,42 @@ func (m *limitStage) Run(in chan Entry) chan Entry {
 	return out
 }
 
+func (m *limitStage) process(ctx context.Context, entries []Entry) error {
+	var dst int
+
+	for _, e := range entries {
+		if m.shouldThrottle(e.Labels) {
```
thampiotr: Honor caller cancellation in limit waits: this new method accepts ctx, but shouldThrottle waits using the stage-owned background context (see lines 169–176). A canceled caller can therefore remain blocked waiting for rate-limit capacity. Please connect caller cancellation to the wait while preserving stage shutdown cancellation, and propagate the wait error from process. This gap arises with the new context-aware path; the old channel API only supported shutdown cancellation.

internal/component/loki/process/stages/stage.go
```diff
@@ -175,53 +152,50 @@ func newStageWithOpts(
...
 		}
 	case cfg.StaticLabelsConfig != nil:
-		s, err = newStaticLabelsStage(*cfg.StaticLabelsConfig)
+		s, err = newStaticLabelsStage(*cfg.StaticLabelsConfig, opts)
 		if err != nil {
 			return nil, err
 		}
 	case cfg.GeoIPConfig != nil:
-		s, err = newGeoIPStage(opts.slogger, *cfg.GeoIPConfig)
+		s, err = newGeoIPStage(*cfg.GeoIPConfig, opts)
```
thampiotr, line 175: Clean up after failed construction: migrating resource-owning stages such as GeoIP exposes an existing infrastructure issue in [newPipeline, pipeline2.go lines 126–142](https://github.com/grafana/alloy/blob/e8803804d00f11577c83ef3dd8ff2fbe95963d6b/internal/component/loki/process/stages/pipeline2.go#L126-L142). Please release already-created stages before returning a construction error, including a successfully constructed stage rejected for lacking entryProcessor. Otherwise resources acquired during a partially successful pipeline construction are left unreleased.

internal/component/loki/process/stages/split_json.go
```diff
@@ -59,6 +67,34 @@ func (s *splitJSONStage) Run(in chan Entry) chan Entry {
...
+			}
+			// Safety: json.RawMessage's UnmarshalJSON contract copies each
+			// value into its own backing slice, which is never exposed or
+			// mutated after this point, so the string's bytes stay immutable.
+			// nosemgrep: use-of-unsafe-block
+			child.Line = unsafe.String(unsafe.SliceData(raw), len(raw)) // #nosec G103
+			out = append(out, child)
+		}
+	}
+	return s.next(ctx, out)
```
thampiotr, line 100: Skip empty split output: when an entry contains [], out can be empty, but this still calls the downstream callback. Please add `if len(out) == 0 { return nil }` before calling next. Add a test asserting that [] causes zero downstream calls; checking only the collected entries misses an empty callback.

thampiotr (commented):
Leaving some suggestions from my review with an agent. I think there are a few stages that are not exactly simple ;)

internal/component/loki/process/stages/stage.go
kalleep: Right, I fixed this in a later pr in the stack but will move it here :+1:

internal/component/loki/process/stages/split_json.go
kalleep: Fixed in [13d2c8c](https://github.com/grafana/alloy/pull/6914/commits/13d2c8c2642e265bbe8cc7bf29a5f6f24258cd8c)

internal/component/loki/process/stages/stage.go
kalleep: Fixed in [fc7e8dd](https://github.com/grafana/alloy/pull/6914/commits/fc7e8dd20a7983fb9879a3e594aa0d155a5a02f4)

internal/component/loki/process/stages/pattern.go
kalleep: Fixed in [aca602c](https://github.com/grafana/alloy/pull/6914/commits/aca602c6409d10c16ea0a8ab333f239c632811f0)

I opted to use a sync.Pool here, since the validate function guarantees that we can build the matcher.

internal/component/loki/process/stages/limit.go
kalleep: I pushed [047f012](https://github.com/grafana/alloy/pull/6914/commits/047f012edcfbfe4bb8816c7d9edbe437cf3f423d)

But this will not be correct either. For our new pipeline we [document that we cannot have any in-flight entries when Stopped is called](https://github.com/grafana/alloy/blob/6d3f0b9607e1e0173e9292d0833ac21e4605c5ed/internal/component/loki/process/stages/pipeline2.go#L89-L93) and this makes sense especially since we could end up buffering entries that would be forever lost. But that also means we can never cancel any rate limited entries during update.

Another option would be to do the same kind of change I did here at component level instead, but that would also cancel the context for any entries that have moved on too the next component and have not yet been stored in WAL or queued. Maybe this is acceptable and sources would have to retry at that point?

internal/component/loki/process/stages/timestamp.go
kalleep: Fixed in [c1d8318](https://github.com/grafana/alloy/pull/6914/commits/c1d8318200f3668a3b3d6684260b9334bc9f4bcf)

I tried to clean up the (unreadable) code a bit. Since I switch to a mutex I also swapped it out to use v2 (uses generics) and instead of the `lru.Cache` I used `simplelru.LRU` from the same package since we don't need locks.

internal/component/loki/process/stages/split_json.go
thampiotr: I think there is still a chance that you will get empty `out` and try to call `next` with it. I think we need:

```
if len(out) == 0 {
    return nil
}
```

in case the input line contains `[]`

kalleep: Yeah so I don't know what happened :facepalm: 
I added that line and the test, yes the test passes since we go an empty batch but still.

internal/component/loki/process/stages/limit.go
```diff
@@ -73,84 +84,119 @@ func newLimitStage(logger *slog.Logger, cfg LimitConfig, registerer prometheus.R
...
+
+// stop implements stopper.
+func (m *limitStage) stop() {
+	m.cancel()
+}
+
+// Stop implements Stopper
+func (m *limitStage) Stop() {
+	m.stop()
+}
```
thampiotr: This looks odd, no way to settle on one interface?

kalleep: So for the current pipeline we have [Stopper](https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/stage.go#L34). Exported interface with Exported method, and we cant reuse the same one since we [implemented partial line flushing on stop for CRI for the new pipeline](https://github.com/grafana/alloy/blob/main/internal/component/loki/process/stages/cri.go#L204). I will eventually remove the other interface

internal/component/loki/process/stages/split_json.go
kalleep: Fix in [90b533c](https://github.com/grafana/alloy/pull/6914/commits/90b533c0a61da16d8416556f14e2d880d57fa966)

internal/component/loki/process/stages/limit.go
```diff
@@ -73,84 +84,119 @@ func newLimitStage(logger *slog.Logger, cfg LimitConfig, registerer prometheus.R
...
 }
 
-func (m *limitStage) shouldThrottle(labels model.LabelSet) bool {
+func (m *limitStage) process(ctx context.Context, entries []Entry) error {
+	var dst int
+
+	sctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	stop := context.AfterFunc(m.ctx, cancel)
+	defer stop()
```
thampiotr: Can we make these context names easier to follow? We have:
- batchCtx - for processing the entries
- stageCtx - for entire stage
- processCtx + processCancel - for this call to process with its cancel
- stopJoin to stop joining the stageCtx with processCancel function via AfterFunc

using synonyms, ctx, m.ctx, sctx, stop, cancel is making it hard to read, we should avoid confusing names as a rule

internal/component/loki/process/stages/limit.go
```diff
@@ -73,84 +84,119 @@ func newLimitStage(logger *slog.Logger, cfg LimitConfig, registerer prometheus.R
...
+
+	for _, e := range entries {
+		err := m.throttle(sctx, e.Labels)
```
thampiotr: Ok, so regarding the question of how to handle cancellation in limit stage, I think we can go with something like this:

- **When updating the component, block new batches from entering the old pipeline.** New calls should wait or reach the replacement pipeline once it is ready.

- **Give active calls a short grace period to finish normally.** This avoids interrupting processing unnecessarily. Like 5 seconds maybe?

- **If the grace period expires, cancel the remaining calls’ batch contexts.** A blocked `limit` wait will return an error. 

- **Wait for those calls to actually return.** Canceling a context only signals cancellation; it does not mean processing has finished. So we wait for the processing to be done.

- **Report failed batches without claiming every entry was lost.** If there is error returned - Log the error, the original batch entry count, and that processing or delivery may have been partial. 

- **Once all active calls have returned, call `Stop()`.** It flushes buffered entries and releases resources. Flushing in all stages that do it needs a hard timeout too. Maybe it should be controlled directly in the loki.process component? TBD

If we take this approach, I think one of the harder parts we need to look into soon is how the `Update` is implemented for loki.process.

internal/component/loki/process/stages/limit.go
thampiotr: hmmm, not sure I follow why we can't use one with some refactor, but if we are planning to clean it up eventually, that's fine. Maybe adding a todo?

internal/component/loki/process/stages/limit.go
kalleep: Yes I had something like this in mind :+1:

thampiotr (approved):
🎉 great work!

PR #6916, opened 2026-08-18
chore: Update js dependencies for dependabot alerts for 1.19

dehaansa:


internal/web/ui/package-lock.json
```diff
@@ -1541,9 +1541,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
```
dehaansa, line 1544: npm audit fix dropped these, I'm not sure if they are relevant or not, tbh. I think I'll revert them, just to be safe.

blewis12 (approved):

PR #6918, opened 2026-08-18
fix: Use default upstream config for vcenter receiver

blewis12:
Ensures that we use the upstream config for the vcenter receiver, instead of building from a zero-struct and overriding with alloy args. This will fix an existing bug where some unexposed `MetricsBuilderConfig.Metrics.*` fields are zero'd out due to not being defined anywhere, which is causing some labels/attributes to be dropped.

### PR Checklist
- [ ] Documentation added
- [x] Tests updated


internal/component/otelcol/receiver/vcenter/vcenter_test.go
```diff
@@ -260,43 +260,87 @@ func TestArguments_UnmarshalAlloy(t *testing.T) {
 	require.False(t, otelArgs.MetricsBuilderConfig.Metrics.VcenterClusterCPUEffective.Enabled)
 	require.True(t, otelArgs.MetricsBuilderConfig.Metrics.VcenterClusterCPULimit.Enabled)
 	require.True(t, otelArgs.MetricsBuilderConfig.Metrics.VcenterClusterHostCount.Enabled)
+	require.Equal(t, "sum", otelArgs.MetricsBuilderConfig.Metrics.VcenterClusterHostCount.AggregationStrategy)
```
blewis12, line 263: since we already have these manual-ish checks, I decided to just extend them to include checks against the fields that weren't being populated. if anything changes upstream and there's drift we should catch it here

kgeckhart (approved):

PR #6951, opened 2026-08-21
chore(loki.process): Implement new stage interface for multiline

kalleep:

Implement the new stage and stopper interface for `stage.multiline`.

### Pull Request Details

For the new stage interface the implementation of `stage.multiline` is much simpler. I decided to not deal with dynamic timers and just run the backgroud flush loop every `MaxWaitTime`, this means that expired entries can be buffered for a bit longer than `MaxWaitTime` but I think that is a fair trade of compared to what the complexity it would bring in otherwise.

I tried to make the logic in process easy to follow compared to Run. But the only thing that is a bit hard to reason about is why the last check is safe and cannot overwrite entries. I added a comment to explain why it is safe.

All test have been update to use `runPipelineTest` function so that every case is verified against old and new pipeline.

Part  of: https://github.com/grafana/alloy/issues/4953

### Notes to the Reviewer

* `TestMultilineStageStreamsMapCleanedUpOnClose` and `TestMultilineStageStreamsMapCleanedUpAfterTimeout` was merged into the same test since they tested the same thing since we always stopped the pipeline before checking stream, race detector fails if pipeline is runnning.

* All timeout related test was collapsed into `TestMultilineStageMaxWaitTime`.

* For now possible OOO entries are accepted for flushed entries but is something we can revisit in the future.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

---


internal/component/loki/process/stages/multiline.go
```diff
@@ -199,6 +225,150 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+
+	for _, e := range entries {
+		key := e.Labels.FastFingerprint()
+
+		if m.streams == nil {
+			m.streams = make(map[model.Fingerprint]*multilineState)
+		}
+		state, hasState := m.streams[key]
+
+		isFirstLine := m.regex.MatchString(e.Line)
```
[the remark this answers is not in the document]
kalleep: fixed in [1ca2ba5](https://github.com/grafana/alloy/pull/6951/commits/1ca2ba5397d6fcea5203b42beb7caf41a38112a1)

internal/component/loki/process/stages/pipeline2.go
```diff
@@ -58,6 +65,10 @@ func NewPipeline2(
 			return nil, errors.New("stage has not been migrated to new interface")
 		}
 
+		if ss, ok := newStage.(starter); ok {
+			ss.start()
+		}
```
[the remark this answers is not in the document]
kalleep: fixed in [1fb12b9](https://github.com/grafana/alloy/pull/6951/commits/1fb12b97670a431338b373523d7a1db2bb1ae8ce)

internal/component/loki/process/stages/multiline.go
```diff
@@ -199,6 +225,150 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+// start implements starter.
+func (m *multilineStage) start() {
+	// NOTE: If we can only buffer one line this stage is skipped so there will never
+	// exist any buffered lines to flush.
+	if m.cfg.MaxLines == 1 {
+		return
+	}
+	m.wg.Go(func() {
+		ticker := time.NewTicker(m.cfg.MaxWaitTime)
+		defer ticker.Stop()
```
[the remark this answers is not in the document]
kalleep: Yes IMO this is acceptable and part of the pr description:
>I decided to not deal with dynamic timers and just run the backgroud flush loop every MaxWaitTime.

internal/component/loki/process/stages/multiline.go
```diff
@@ -199,6 +225,150 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+
+		for {
+			select {
+			case <-m.done:
+				return
+			case <-ticker.C:
+				entries := m.expired()
+				if len(entries) > 0 {
+					ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+					if err := m.next(ctx, entries); err != nil {
```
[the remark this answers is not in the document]
kalleep: fixed in [5dcf1ee](https://github.com/grafana/alloy/pull/6951/commits/5dcf1ee0213b27d29e228d151c157b83586a5212)

internal/component/loki/process/stages/multiline.go
```diff
@@ -53,49 +54,74 @@ func (args *MultilineConfig) Validate() error {
...
+
 func (m *multilineStage) Run(in chan Entry) chan Entry {
+	// NOTE: If we can only buffer one line there is no point to doing it at all.
+	if m.cfg.MaxLines == 1 {
+		return RunWith(in, func(e Entry) Entry {
+			return e
+		})
+	}
+
 	out := make(chan Entry)
```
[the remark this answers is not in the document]
kalleep: fixed in [be480e2](https://github.com/grafana/alloy/pull/6951/commits/be480e260821f1590f008ad1e3274f5116892ea7)

internal/component/loki/process/stages/multiline.go
```diff
@@ -199,6 +218,168 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+		}
+	}
+
+	m.mut.Unlock()
+
+	if dst == 0 {
+		return nil
+	}
+
+	return m.next(ctx, entries[:dst])
```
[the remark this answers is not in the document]
kalleep: We have still not moved over to the new pipeline where we can actually return errors and this is not something we handle today. But yes we should fix this. I will leave a FIXME comment and get back to this in a future pr.

internal/component/loki/process/stages/multiline.go
```diff
@@ -199,6 +218,168 @@ func (m *multilineStage) Run(in chan Entry) chan Entry {
...
+					expired []Entry
+					now     = time.Now()
+				)
+
+				for key, state := range m.streams {
+					if !state.lastSeen.Add(m.cfg.MaxWaitTime).After(now) {
+						if state.currentLines > 0 {
+							expired = append(expired, m.flushState(state))
+						}
+						delete(m.streams, key)
```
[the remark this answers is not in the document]
kalleep: We have still not moved over to the new pipeline where we can actually return errors and this is not something we handle today. But yes we should fix this. I will leave a FIXME comment and get back to this in a future pr.

internal/component/loki/process/stages/multiline.go
```diff
@@ -261,37 +272,256 @@ func (m *multilineStage) processEntry(key model.Fingerprint, e Entry) []Entry {
...
+		s.stripes[i].data = make(map[model.Fingerprint]*multilineState)
+	}
+	return s
+}
+
+type multilineStreamsStriped struct {
+	stripes [stripeCount]multilineStripe
+}
+
+func (m *multilineStreamsStriped) Mutate(fp model.Fingerprint, fn func(s *multilineState, ok bool) *multilineState) {
```
kalleep, line 444: I decided to go with the callback approach here since we need to do a couple of things to the state.

I experimented with a different approach where I had `Take` and `Store`. But if we for some reason would get calls here for the same stream they would race so it's better to lock the state for one loop iteration. I confirmed with benchmarks that this does not do any extra allocations like it did with the initial implementation for CRI

internal/component/loki/process/stages/multiline.go
```diff
@@ -53,48 +56,58 @@ func (args *MultilineConfig) Validate() error {
...
-	lastSeen       time.Time
-}
+	_ entryProcessor = (*multilineStage)(nil)
+	_ starter        = (*multilineStage)(nil)
+	_ stopper        = (*multilineStage)(nil)
+)
 
-// newMultilineStage creates a MulitlineStage from config
-func newMultilineStage(logger *slog.Logger, config MultilineConfig) (Stage, error) {
+// newMultilineStage creates a mulitlineStage from config
```
dehaansa:
```suggestion
// newMultilineStage creates a multilineStage from config
```

internal/component/loki/process/stages/multiline_test.go
```diff
@@ -1,637 +1,733 @@
...
-func TestMultilineStageProcessLeaveNewlines(t *testing.T) {
-	mcfg := MultilineConfig{Expression: "^START", MaxWaitTime: 3 * time.Second, TrimNewlines: false}
-	regex, err := validateMultilineConfig(mcfg)
-	require.NoError(t, err)
-
-	stage := &multilineStage{
-		cfg:    mcfg,
-		regex:  regex,
-		logger: logging.NewSlogNop(),
+// TestMultilineStageStreamsCleaned verifies that the streams is empty after the stopping.
```
dehaansa:
```suggestion
// TestMultilineStageStreamsCleanup verifies that the streams is empty after the stopping.
```

internal/component/loki/process/stages/multiline.go
```diff
@@ -261,37 +274,256 @@ func (m *multilineStage) processEntry(key model.Fingerprint, e Entry) []Entry {
...
+			case state.currentLines > 0 && time.Since(state.lastSeen) >= m.cfg.MaxWaitTime:
+				entries[dst] = state.flush()
+				dst++
+
+				line := e.Line
+				if m.cfg.TrimNewlines {
+					line = strings.TrimRight(line, "\r\n")
+				}
+
+				state.buffer.WriteString(line)
```
dehaansa, line 331: If I'm understanding this correctly, in this case we've identified that this is _not_ a firstline of a multiline log, but we're adding it to a buffer after flushing the previous  multiline log? It seems like this line should just be written to entries directly and state should be empty on next line evaluation. Is that right?

internal/component/loki/process/stages/multiline.go
```diff
@@ -261,37 +274,256 @@ func (m *multilineStage) processEntry(key model.Fingerprint, e Entry) []Entry {
...
+			s.startLineEntry.Created(),
+			push.Entry{
+				Timestamp:          s.startLineEntry.Entry.Entry.Timestamp,
+				Line:               s.buffer.String(),
+				StructuredMetadata: slices.Clone(s.startLineEntry.Entry.Entry.StructuredMetadata),
+			}),
 	}
 
 	s.buffer.Reset()
 	s.currentLines = 0
```
dehaansa, line 526: Not resetting startLine means the `case state.currentLines > 0 && time.Since(state.lastSeen) >= m.cfg.MaxWaitTime:` case could leak startLine between batches, though this may be a non-issue depending on how the other comment is handled.

dehaansa: In re-reading this logic I guess the existing behavior that we've kept consistent is that if the multiline takes too long to come in, each block of slow multiline logs will still be multiline chunked until the next start is detected.... I don't love it but I guess if it's the existing behavior 🤷

kalleep: Yes this part is really strange and I think we should revisit it. For now I went with compatibility with the old logic and we can make that change in it's own pr.

Should have pointed this out to make the review easier, my bad.

internal/component/loki/process/stages/multiline.go
kalleep: This is related to the comment below right? As mentioned there I wanted to keep existing behavior even though it's a bit strange. We should address this as a separate pr.

PR #6953, opened 2026-08-21
fix(database_observability.mysql): Resolve query-parsed table names against known-table casing

gaantunes:

query_details.go emits `query_parsed_table_name` using the table name's literal casing as it appears in the query text, while schema_details.go emits `create_statement` using `information_schema`'s casing for the table. On a MySQL cluster configured with `lower_case_table_names=1` or `2`, these two casings can differ for the same table, so anything matching the two labels together (e.g. an exact-match join/filter) can miss an existing table.

This PR adds a `TableRegistry` to `schema_details.go`, populated from `information_schema.tables` and shared with `query_details.go` (mirroring the `TableRegistry` the postgres collector already uses for its own identifier-folding case). `query_details.go` now resolves each parsed table name against the registry — exact match first, case-insensitive fallback — before emitting `query_parsed_table_name`, so it's emitted with the same casing as `create_statement` whenever the table exists.

`query_details.go` also now emits a `validated="true"/"false"` field on `query_parsed_table_name` (mirroring the postgres collector), based on whether the parsed name was actually found in the `TableRegistry`. This lets a downstream consumer filter out a table name the SQL parser/tokenizer misidentified (e.g. a column or alias parsed as if it were a table reference) the same way it already does for postgres.

Follow-ups from review:
- `TableRegistry.IsValid` now splits a schema-qualified parsed table name (a cross-database reference like `otherdb.sometable`) and resolves the schema part against known schema names the same way the table part is resolved, instead of treating the whole qualified string as an opaque, never-matching table name scoped to the connection's active schema.
- `extractSchema` now tracks whether it scanned every `information_schema.tables` row without bailing out early, and skips updating the table registry / pruning throttle entries when it didn't — mirroring the postgres collector's `scanComplete` guard. Previously a scan failure partway through would still replace the whole registry and prune throttle entries from the partial list collected so far, incorrectly treating existing-but-unscanned tables as unknown/gone until the next successful tick.
- The case-insensitive fallback in `TableRegistry` used last-write-wins when two distinct schemas or tables collided on the same lowercased name (possible under the MySQL default `lower_case_table_names=0`, where such objects can genuinely coexist). `SetTables` now tracks such collisions explicitly, and the resolvers treat an ambiguous lowercased name as no match rather than guessing; an exact match against either colliding name is unaffected.

New tests: `TableRegistry.IsValid` (schema-scoped exact + case-insensitive resolution, schema-qualified resolution, case-insensitive collision handling, empty/unknown-schema/unknown-table cases, exact match taking priority over the fallback), `TestSchemaDetails_PopulatesTableRegistry`, `TestSchemaDetails_SkipsRegistryAndThrottleUpdatesOnPartialScan`, `TestQueryDetails_ResolvesTableNameCasingViaTableRegistry`, `TestQueryDetails_MarksAnUnresolvedParsedNameAsNotValidated`, and `TestQueryDetails_ResolvesASchemaQualifiedCrossDatabaseReference`.

Fixes grafana/grafana-dbo11y-app#3142

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/mysql/collector/schema_details.go
+++ b/internal/component/database_observability/mysql/collector/schema_details.go
@@ -242,11 +242,20 @@ func (c *SchemaDetails) extractSchema(ctx context.Context) error {
 	now := c.now()
 	tables := []*tableInfo{}
 	seenTables := map[string]struct{}{}
+	// scanComplete tracks whether we scanned every row without bailing out early. If false,
+	// tables is a partial view of information_schema.tables, so we must skip replacing the
+	// table registry and pruning throttle entries below: doing either from a partial list
+	// would incorrectly mark existing-but-unscanned tables as unknown/unvalidated, or evict
+	// their throttle entries as if they no longer existed. The per-table create_statement
+	// processing further down still runs over whatever partial list we did get — those rows
+	// scanned fine and are real tables, just possibly not an exhaustive set this tick.
+	scanComplete := true
 	for rs.Next() {
 		var schema, tableName, tableType string
 		var createTime, updateTime time.Time
 		if err := rs.Scan(&schema, &tableName, &tableType, &createTime, &updateTime); err != nil {
 			c.logger.Error("failed to scan tables", "err", err)
+			scanComplete = false
 			break
 		}
 		tables = append(tables, &tableInfo{
@@ -271,14 +280,16 @@ func (c *SchemaDetails) extractSchema(ctx context.Context) error {
 		return fmt.Errorf("failed to iterate over tables result set: %w", err)
 	}
 
-	c.tableRegistry.SetTables(tables)
+	if scanComplete {
+		c.tableRegistry.SetTables(tables)
 
-	// Cleanup: drop throttle entries for tables that no longer exist (e.g.
-	// tables dropped or renamed). Done before the empty-tables early return so
-	// the map is also pruned when every monitored table disappears.
-	for k := range c.lastEmittedAt {
-		if _, ok := seenTables[k]; !ok {
-			delete(c.lastEmittedAt, k)
+		// Cleanup: drop throttle entries for tables that no longer exist (e.g.
+		// tables dropped or renamed). Done before the empty-tables early return so
+		// the map is also pruned when every monitored table disappears.
+		for k := range c.lastEmittedAt {
+			if _, ok := seenTables[k]; !ok {
+				delete(c.lastEmittedAt, k)
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


fridgepoet:
I think this will fix https://github.com/grafana/grafana-dbo11y-app/issues/3142

gaantunes:
> I think this will fix [grafana/grafana-dbo11y-app#3142](https://github.com/grafana/grafana-dbo11y-app/issues/3142)

I have added the isValid field to it, mimicking  the postgres implementation, so it solves this ticket you referenced too.

fridgepoet:
Thanks @gaantunes 

Can you ensure the registry behavior will be acceptable for schema-qualified as well as unqualified tables? I think Postgres might split the qualified tables before lookup. We might want to match that behavior.

internal/component/database_observability/mysql/collector/schema_details.go
fridgepoet, line 259: @gaantunes I think this is an edge case, so probably not a big deal... but

Would you just check here that the failure to scan and break is consistent with what we do in the Postgres collector? (or we may choose not to be consistent)

Just want to make sure that what we do here will make sense for setting the registry table below `c.tableRegistry.SetTables(tables)` and we don't have something partial or incoherent on the app side.

gaantunes: addressed that in the last commit, thanks for bringing it up.

gaantunes:
@fridgepoet this should be ready for another round of review.

fridgepoet (dismissed):
look good, may just need to take a decision on the copilot review https://github.com/grafana/alloy/pull/6953#discussion_r3874885431

internal/component/database_observability/mysql/collector/schema_details.go
```diff
@@ -545,3 +568,133 @@ func (c *SchemaDetails) fetchSchemaSpecs(ctx context.Context, schema string, tab
...
+	tr.schemaByLower = make(map[string]string, len(tables))
+	for _, t := range tables {
+		if tr.exact[t.schema] == nil {
+			tr.exact[t.schema] = make(map[string]struct{})
+			tr.byLower[t.schema] = make(map[string]string)
+		}
+		tr.exact[t.schema][t.tableName] = struct{}{}
+		tr.byLower[t.schema][strings.ToLower(t.tableName)] = t.tableName
+		tr.schemaByLower[strings.ToLower(t.schema)] = t.schema
+	}
```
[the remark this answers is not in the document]
gaantunes: fixed: SetTables tracks lowercase collisions explicitly; resolvers return "no match" instead of guessing.

gaantunes:
> look good, may just need to take a decision on the copilot review [#6953 (comment)](https://github.com/grafana/alloy/pull/6953#discussion_r3874885431)

 
Sorry I missed this, it should be addressed.

fridgepoet (approved):

PR #6957, opened 2026-08-24
chore(loki.process): Implement new stage interface for match

kalleep:

Implement the new stage and stopper interface for `stage.match`.

### Pull Request Details

In this pr I implemented the new `stage` interface for matchStage, to make the logic easier to reason about I split the stage into `matchKeepStage` and `matchDropStage` since these two behave differently.

Change should pr pretty straight forward but there is one caveat to `matchKeepStage`. Since we need to call into the inner pipeline for only entries that matches logql expression we have two options:
1. Split entries into two different calls, this would result in a batch split in the new arcitecture.
2. Merge entries from inner pipeline back into entries slice so we keep batch intact.

I went with the second option since it's feels better to keep batches intact but I am open to option 1 as well.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

---


internal/component/loki/process/stages/limit_test.go
```diff
@@ -147,42 +147,70 @@ loki_process_dropped_lines_by_label_total{label_name="app",label_value="poki"} 4
 // TestLimitStageShutdown verifies that an entry blocked in rateLimiter.Wait
 // is released promptly when the pipeline shuts down.
 func TestLimitStageShutdown(t *testing.T) {
```
kalleep, line 149: We only run this test for the previous pipeline. Canceling a blocked limit call will be handled differently in the new pipeline and will be fixed by next pr.

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
 	}
-	return e, false
+
+	return m.next(ctx, entries[:dst])
 }
 
-func (m *matcherStage) Cleanup() {
-	if m.pipeline != nil {
-		m.pipeline.Cleanup()
+// collected is passed as the next function to the inner pipeline.
```
dehaansa, line 308:
```suggestion
// collect is passed as the next function to the inner pipeline.
```

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
-func (m *matcherStage) Stop() {
-	if m.pipeline != nil { // nil for MatchActionDrop matchers, see Cleanup
-		m.pipeline.Stop()
+// stop implements stopper.
+func (m *matchKeepStage) stop() {
+	m.pipeline2.stop()
+}
+
+func matchLogQL(e Entry, matchers []*labels.Matcher, filter logql.Filter) bool {
+	for _, filter := range matchers {
```
dehaansa: Let's not overload the name filter in this function?

internal/component/loki/process/stages/pack_test.go
```diff
@@ -14,91 +15,122 @@ import (
...
-	w = &Packed{}
-	assert.NoError(t, json.Unmarshal([]byte(out2.Entry.Entry.Line), w))
-	expectedPackedLabels = map[string]string{
-		"pod":       "foo-vvsdded",
-		"container": "bar",
-	}
-	assert.Equal(t, expectedPackedLabels, w.Labels)
-	assert.Equal(t, regexLogFixture, w.Entry)
+		require.NoError(t, p.process(context.Background(), []Entry{entry1()}))
+		time.Sleep(1 * time.Millisecond)
```
dehaansa, line 121: a 1ms sleep seems strange, can we at least provide a comment why it's necessary?

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
-			return e, false
+// Cleanup implements Stage.
+func (m *matchKeepStage) Cleanup() {
+	m.pipeline.Cleanup()
+}
+
+type matchMergeKey struct{}
+
+// withMatchMerge is used to store a pointer to a entry slice so that we can later
+// merge it back to the original silce after inner pipeline have finished.
```
dehaansa, line 252:
```suggestion
// merge it back to the original slice after inner pipeline have finished.
```

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
+		if matched == nil {
+			matched = make([]Entry, 0, len(entries)-i)
+		}
+		matched = append(matched, e)
 	}
 
-	if m.filter == nil || m.filter([]byte(e.Line)) {
-		return e, true
+	if len(matched) > 0 {
+		// We store a pointer to buf in context so that we can later merged it back after
```
dehaansa, line 285:
```suggestion
		// We store a pointer to buf in context so that we can merge it back after
```

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
 }
 
-func (m *matcherStage) Cleanup() {
-	if m.pipeline != nil {
-		m.pipeline.Cleanup()
+// collected is passed as the next function to the inner pipeline.
+func (m *matchKeepStage) collect(ctx context.Context, entries []Entry) error {
+	// If context contains a pointer to a entry slice that means we called it directly.
+	// In this case we should just set it to resulting entries from inner pipeline
+	// so that process can merged it back.
```
dehaansa, line 312:
```suggestion
	// so that process can merge it back.
```

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
-				out <- e
-				continue
-			}
-			m.dropCount.WithLabelValues(m.dropReason).Inc()
-		}
-	}()
-	return out
+// Stop implements Stopper.
+func (m *matchKeepStage) Stop() {
+	m.pipeline.Stop()
```
dehaansa, line 241: m.pipeline can be nil, right? Same for Cleanup()

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
+		var buf []Entry
+		if err := m.pipeline2.process(withMatchMerge(ctx, &buf), matched); err != nil {
+			return err
+		}
+		entries = append(entries[:dst], buf...)
+		dst += len(buf)
+
+		// Entries that took different branches can still share a stream so we need to
+		// make sure they are sorted to prevent OOO.
+		slices.SortStableFunc(entries[:dst], func(x, y Entry) int {
```
dehaansa, line 296: Adding a sort here is new behavior, can reorder log streams if timestamp changed in the inner pipeline. Unsure if that's okay or not, should probably at least be exercised in tests.

internal/component/loki/process/stages/match.go
kalleep: No it cannot, Stop and Cleanup are for the current pipeline, new only have `stop`. For the next pr these will be removed since we can replace the current one with the new one internally.

internal/component/loki/process/stages/match.go
kalleep: > Unsure if that's okay or not, should probably at least be exercised in tests

This would be better than we had before, since with a single entry we can still hit this case but don't even have a chance to fix it, the fix is limited to one batch but should still be an improvement. I just need to think a bit about what happens if entries within the same stream share the same timestamp.

But yes a test here would be good :+1:

internal/component/loki/process/stages/pack_test.go
kalleep: Agreed, this was keep from the old test, will look into it.

internal/component/loki/process/stages/match.go
```diff
@@ -162,42 +236,124 @@ func (m *matcherStage) runKeep(in chan Entry) chan Entry {
...
+			return x.Timestamp.Compare(y.Timestamp)
+		})
```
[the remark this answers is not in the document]
kalleep: We should not try to keep "original interleaving", that would be pretty hard to even do since more entries could be produced by the inner pipeline. 

>Second-precision logs in one stream get regrouped
Not important since that would not be OOO.

IMO this is better than we had before since it would fix the order for streams before we add it to the batch again.

internal/component/loki/process/stages/match.go
kalleep: fixed in [2d7f492](https://github.com/grafana/alloy/pull/6957/commits/2d7f49217a43d3ffcd7bb4db85042d44ac7f15a3)

internal/component/loki/process/stages/match.go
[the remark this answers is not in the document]
kalleep: Okay added some more tests in [bde22d1](https://github.com/grafana/alloy/pull/6957/commits/bde22d1cda119116c660b999ec4514f51709d68f).

The last one is a bit annoying since they will be re-ordered if they share timestamp. In practice I don't really think this is a problem because I don't think loki will guarantee any order at query time when this happens, except if `increment_duplicate_timestamp` is enabled (off by default).

If we want to fix this we would need to pass entries one by one through the sub-pipeline, WDYT?

PR #6959, opened 2026-08-24
test: Migrate static prometheus exporter integration test to pipelinetest

thampiotr:

Adds Prometheus signal support to `internal/pipelinetest`, mirroring the existing Loki support, then
moves the `static` integration test down a level from Docker to a pipeline test.

`prometheus.exporter.static` connects to no external dependency, so the Docker test only verified
that components work together inside a pipeline. Its Mimir leg is already covered by other
integration tests.

So following our test-writing guidelines, it's best suited for a pipeline test.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/pipelinetest/harness/service.go
```diff
@@ -2,30 +2,51 @@ package harness
...
 		&mockService{
-			name: httpservice.ServiceName,
-			data: httpservice.Data{
-				HTTPListenAddr:   "127.0.0.1:0",
-				MemoryListenAddr: "alloy.internal:0",
-				BaseHTTPPath:     "/",
-				DialFunc:         (&net.Dialer{}).DialContext,
-			},
+			name: remotecfg.ServiceName,
 		},
```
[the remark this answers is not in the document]
thampiotr: done

internal/pipelinetest/harness/assertion_prometheus.go
```diff
@@ -0,0 +1,192 @@
+package harness
```
kalleep, line 1: I guess it would be nice to have `assertion_loki.go` and move loki specific assertions there?

internal/pipelinetest/harness/alloy.go
```diff
@@ -39,7 +39,7 @@ func NewAlloy(cfg Config) (*Alloy, error) {
 		Logger:       logger,
 		DataPath:     cfg.DataPath,
 		MinStability: featuregate.StabilityExperimental,
-		Services:     defaultServices(logger.Slog()),
+		Services:     defaultServices(logger),
```
kalleep, line 42: 👍

internal/pipelinetest/harness/service.go
```diff
@@ -2,31 +2,50 @@ package harness
...
-				MemoryListenAddr: "alloy.internal:0",
-				BaseHTTPPath:     "/",
-				DialFunc:         (&net.Dialer{}).DialContext,
-			},
-		},
+		labelstore.New(l.Slog(), prometheus.NewRegistry()),
+		// The real HTTP service, not a mock. prometheus.exporter.* components serve
+		// their metrics over the component HTTP path and export a target pointing at
+		// the in-memory listener, so prometheus.scrape can only reach them when the
+		// service actually serves and its DialFunc routes in-memory traffic.
```
kalleep: Do we really need this comment?

internal/pipelinetest/schema_prometheus.go
```diff
@@ -0,0 +1,121 @@
+package pipelinetest
```
kalleep, line 1: Same as the other comment, maybe a `schema_loki.go` file?

kalleep (commented):
This is a good start, left a couple of comments but nothing blocking.

There are other things we want to support here in the future such as adding a endpoint to the sink that accepts remote write requests (both v1 and v2 I guess).

internal/pipelinetest/harness/service.go
thampiotr: no, removed

thampiotr:
> This is a good start, left a couple of comments but nothing blocking.
> 
> There are other things we want to support here in the future such as adding a endpoint to the sink that accepts remote write requests (both v1 and v2 I guess).

Yeah, I think we can add remote writes later or maybe keep them on the level of integration tests. Have not fully decided yet.

kalleep (approved):
🎉

PR #6968, opened 2026-08-24
docs: Clean up the cli run topic and align with Clustering tasks topics

clayton-cornell:
This PR combines a review pass top clean up language, spelling and word use with a refactor of the Clustering section.

The Clustering section is stripped down (previously very word dense and hard to read/understand) and an xref to the Clustering topics is added.

Depends on: https://github.com/grafana/alloy/pull/5589


docs/sources/reference/cli/run.md
```diff
@@ -200,9 +152,13 @@ For example: `dnssrvnoa+_alloy-memberlist._tcp.service.consul`
 
 If no prefix is provided, Alloy will attempt to resolve the name using both A/AAAA and DNSSRV queries.
```
dehaansa: I know it's not new here, but missing {{< param "PRODUCT_NAME" >}}
```suggestion
If no prefix is provided, {{< param "PRODUCT_NAME" >}} will attempt to resolve the name using both A/AAAA and DNSSRV queries.
```

clayton-cornell: When you stare at a topic for so long these things fade into the paragraphs. Thanks.. fixed.

dehaansa (approved):

PR #6969, opened 2026-08-24
feat: switch systemd service type to notify

defanator:

Send sd_notify(READY=1) via github.com/coreos/go-systemd/v22/daemon once the HTTP listener is bound. The call is wired through a new NotifyFunc callback on httpservice.Options, keeping the HTTP service decoupled from the systemd package.

Set Type=notify in packaging/deb/alloy.service and packaging/rpm/alloy.service so systemd waits for the readiness signal before considering the service started, rather than declaring it active the moment the process is forked.

### Pull Request Details

This change is intended to address potential race between systemd started alloy service and other processes trying to subsequently call e.g. `systemctl reload alloy`. Currently it could lead to the effect of alloy being killed with SIGHUP (default golang behavior) instead of initiating proper configuration reload, if SIGHUP comes in _before_ alloy sets its signal handlers (proved to be an issue on slow cloud VMs where alloy process could take 10-20 seconds to start due to the size of a binary).

With sd_notify, we clearly define at which point service is considered active and ready - when HTTP management API is up.

N/A

### Notes to the Reviewer

This was tested on Ubuntu 26.04 VM. Corresponding log entries:
```
Aug 24 23:32:42 ubuntu2604 systemd[1]: Starting alloy.service - Vendor-agnostic OpenTelemetry Collector distribution with programmable pipelines...
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.272951652Z level=info msg="Alloy is starting"
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.273013764Z level=info msg="boringcrypto enabled" enabled=false
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.27315794Z level=info msg="memory is not limited, skipping" package=github.com/KimMachineGun/automemlimit/memlimit
[..]
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.276251681Z level=info msg="{^_^} Alloy is running"
[..]
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.277128823Z level=info msg="now listening for http traffic" service=http addr=127.0.0.1:12345
Aug 24 23:32:42 ubuntu2604 alloy[73860]: ts=2026-08-24T23:32:42.277154609Z level=debug msg="sent readiness notification to systemd"
```

Service status:
```
root@ubuntu2604:/etc/alloy# systemctl status alloy
● alloy.service - Vendor-agnostic OpenTelemetry Collector distribution with programmable pipelines
     Loaded: loaded (/usr/lib/systemd/system/alloy.service; enabled; preset: enabled)
     Active: active (running) since Mon 2026-08-24 23:32:42 UTC; 16min ago
 Invocation: 1ed47aeb9d6b413882c46921f18c8cc4
       Docs: https://grafana.com/docs/alloy
   Main PID: 73860 (alloy)
      Tasks: 8 (limit: 15648)
     Memory: 52.8M (peak: 53.9M)
        CPU: 3.633s
     CGroup: /system.slice/alloy.service
             └─73860 /usr/bin/alloy run --storage.path=/var/lib/alloy/data /etc/alloy/config.alloy
[..]

root@ubuntu2604:/etc/alloy# systemctl show alloy -p Type -p ActiveEnterTimestamp
ActiveEnterTimestamp=Mon 2026-08-24 23:32:42 UTC
Type=notify
```

### PR Checklist

N/A

### DIFF
--- a/internal/alloycli/cmd_run.go
+++ b/internal/alloycli/cmd_run.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/fatih/color"
 	"github.com/grafana/ckit/advertise"
 	"github.com/grafana/ckit/peer"
@@ -523,6 +524,13 @@ func (fr *alloyRun) run(ctx context.Context, fset *pflag.FlagSet, params runPara
 		ReadyFunc: func() bool { return ready() },
 		ReloadFunc: func() error {
 			return reload()
+		},
+		NotifyFunc: func() {
+			if sent, err := daemon.SdNotify(false, daemon.SdNotifyReady); err != nil {
+				slogger.Warn("failed to send systemd readiness notification", "err", err)
+			} else if sent {
+				slogger.Debug("sent readiness notification to systemd")
+			}
 		},
 
 		HTTPListenAddr:   fr.httpListenAddr,
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -53,6 +53,11 @@ type Options struct {
 	ReadyFunc  func() bool
 	ReloadFunc func() error
 
+	// NotifyFunc, if non-nil, is called once after the HTTP listener is
+	// successfully bound and the server is ready to accept connections.
+	// It is intended for process-readiness signalling (e.g. sd_notify).
+	NotifyFunc func()
+
 	HTTPListenAddr   string                // Address to listen for HTTP traffic on.
 	MemoryListenAddr string                // Address to accept in-memory traffic on.
 	EnablePProf      bool                  // Whether pprof endpoints should be exposed.
@@ -300,6 +305,10 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 	srv := &http.Server{Handler: r, Protocols: protos}
 
 	s.log.Info("now listening for http traffic", "addr", s.opts.HTTPListenAddr)
+
+	if s.opts.NotifyFunc != nil {
+		s.opts.NotifyFunc()
+	}
 
 	listeners := []net.Listener{s.publicLis, s.memLis}
 	for _, lis := range listeners {
--- a/internal/service/http/http.go
+++ b/internal/service/http/http.go
@@ -306,10 +306,6 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 
 	s.log.Info("now listening for http traffic", "addr", s.opts.HTTPListenAddr)
 
-	if s.opts.NotifyFunc != nil {
-		s.opts.NotifyFunc()
-	}
-
 	listeners := []net.Listener{s.publicLis, s.memLis}
 	for _, lis := range listeners {
 		wg.Add(1)
@@ -324,6 +320,10 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 	}
 
 	defer func() { _ = srv.Shutdown(ctx) }()
+
+	if s.opts.NotifyFunc != nil {
+		s.opts.NotifyFunc()
+	}
 
 	<-ctx.Done()
 	return nil
--- a/packaging/deb/alloy.service
+++ b/packaging/deb/alloy.service
@@ -5,6 +5,7 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
+Type=notify
 Restart=always
 User=alloy
 Environment=HOSTNAME=%H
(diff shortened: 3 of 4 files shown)


internal/alloycli/cmd_run.go
```diff
@@ -524,6 +525,13 @@ func (fr *alloyRun) run(ctx context.Context, fset *pflag.FlagSet, params runPara
 		ReloadFunc: func() error {
 			return reload()
 		},
+		NotifyFunc: func() {
+			if sent, err := daemon.SdNotify(false, daemon.SdNotifyReady); err != nil {
+				slogger.Warn("failed to send systemd readiness notification", "err", err)
+			} else if sent {
+				slogger.Debug("sent readiness notification to systemd")
+			}
```
ryepup, line 533: nit: [`SdNotify`](https://pkg.go.dev/github.com/coreos/go-systemd/daemon#SdNotify) has a third outcome that might be nice to log

```suggestion
			} else {
				slogger.Debug("notification not supported, check NOTIFY_SOCKET")
			}
```

internal/service/http/http.go
```diff
@@ -301,6 +306,10 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
 
 	s.log.Info("now listening for http traffic", "addr", s.opts.HTTPListenAddr)
 
+	if s.opts.NotifyFunc != nil {
+		s.opts.NotifyFunc()
+	}
+
```
ryepup: I think we want this a few lines lower, the listeners aren't started until after the `range listeners` loop. Maybe right above the `<-ctx.Done()`?

internal/service/http/http.go
```diff
@@ -53,6 +53,11 @@ type Options struct {
 	ReadyFunc  func() bool
 	ReloadFunc func() error
 
+	// NotifyFunc, if non-nil, is called once after the HTTP listener is
+	// successfully bound and the server is ready to accept connections.
+	// It is intended for process-readiness signalling (e.g. sd_notify).
+	NotifyFunc func()
```
ryepup, line 59: future/probably out of scope: `http_test.go` could benefit from this feature. Currently they have a lot of `utils.Eventually` loops retrying the test until the HTTP listener has started.

The pattern looks like:

```golang
env, err := newTestEnvironment(t) // <-- creates a http.Service
require.NoError(t, err)
// <other env setup omitted>
go func() {
	require.NoError(t, env.Run(ctx)) // <-- calls http.Service.Run
}()

util.Eventually(t, func(t require.TestingT) {
	// make requests to env.ListenAddr()
}
```

With this new `NotifyFunc`, that could get deterministic and drop the `util.Eventually`. The `newTestEnvironment` could make a channel, have the `NotifyFunc` close that channel, then expose a method so tests can wait until the listener is ready.

```golang
env, err := newTestEnvironment(t)
require.NoError(t, err)
// <other env setup omitted>
go func() {
	require.NoError(t, env.Run(ctx))
}()
env.Wait(t) <-- blocks until NotifyFunc is called
// make requests to env.ListenAddr()
```

internal/alloycli/cmd_run.go
defanator: alloy still can be running as standalone binary, or in a k8s deployment, where this 3rd one could be misleading; this whole thing only makes sense in a scenario when alloy is running as a systemd service on a real server / VM. Containerized environments already rely on proper readiness/health checks.

internal/service/http/http.go
defanator: Agree - done.

internal/service/http/http.go
defanator: Agree, though it definitely should be addressed separately.

defanator:
@dehaansa would appreciate your eyes on this one, TIA.

packaging/deb/alloy.service
```diff
@@ -5,6 +5,7 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
+Type=notify
```
dehaansa, line 8: Are there any concerns/side effects of having the notify type and _not_ seeking to send a `stopping` notification?

defanator: UPD: please ignore below response - I misread the question :) for stopping notification it shouldn't be a problem; it would be nice to also send it e.g. when alloy gets a signal for shutdown - that would make `systemctl status` response more accurate in hypothetic scenario when some component(s) take long to shut themselves down. Minor, but would be nice to add.

--- IGNORE THIS ---
@dehaansa yes - if systemd won't receive ack from a service of type notify, the service will not be marked as active and may stuck in "activating" state.

So, if at some point alloy won't have http API always enabled, we'll need another place when/where such notification can be sent.

Issue #6970, opened 2026-08-25
Windows container crashes on startup with STATUS_DLL_NOT_FOUND (0xC0000135) since v1.19.0


petewall:
### What's wrong?

Starting with `v1.19.0-windowsservercore-ltsc2022`, the Alloy Windows container image fails to start. The `alloy` process crashes immediately, with **no log output at all**, it dies before it can write anything.

Pod exit code: `-1073741515`, which decodes to NTSTATUS `0xC0000135` (`STATUS_DLL_NOT_FOUND`). Windows couldn't resolve a DLL dependency of `alloy.exe` at process launch.

Pinning back to `v1.18.1-windowsservercore-ltsc2022` resolves the issue immediately with the exact same Kubernetes manifests, node, and config.

An init container in the same pod, using the *same* `v1.19.0` image but only running a PowerShell `New-NetFirewallRule` command, completes successfully. So the base OS layer itself is fine. Only `alloy.exe` fails to launch.

#### Suspected root cause

Comparing `Dockerfile.windows` between `v1.18.1` and `v1.19.0`, the CGO compiler used to build the Windows binary changed:

```diff
-RUN choco install 7zip --version 22.1 -y
-ADD https://github.com/jmeubank/tdm-gcc/releases/download/v10.3.0-tdm64-2/tdm64-gcc-10.3.0-2.exe ...
-RUN mkdir C:\TDM-GCC-64; ... (unpack TDM-GCC) ...
+RUN choco install mingw --version 14.2.0 -y
```

TDM-GCC 10.3.0 was replaced with a WinLibs-packaged mingw-w64 14.2.0. The final runtime stage is:

```dockerfile
FROM ${BASE_IMAGE_WINDOWS}   # mcr.microsoft.com/windows/nanoserver:ltsc2022
COPY --from=builder ["/src/alloy/build/alloy", "C:/Program Files/GrafanaLabs/Alloy/alloy.exe"]
```

[...]

PR #6972, opened 2026-08-25
ci: Add Sanity Tests For Windows Builds

blewis12:
Fixes some dynamic linking failures we're seeing in Alloy v1.19.0, due to the switch from `TDM-GCC` to `mingw-w64` toolchains in our build pipeline

### Pull Request Details
There are two different paths to consider, both broke due to missing dll's in the final release (in Alloy v1.19.0). 

[Standalone executable/service installation](https://github.com/grafana/alloy/blob/4cd37219590b1623c6843402cfddde5ad4d6ce12/build-tools/build-image/Dockerfile#L54-L56)
This build path includes an older version of `mingw-64` via the viceroy base image. This build has the [-lssp](https://github.com/grafana/alloy/blob/66ee50305df5f3f60b58e7150a53848ab4310867/build-tools/make/packaging.mk#L33) `CGO_LDFLAGS ` flag enabled so that it can properly build alloy, _but_ the dll that this requires, `libssp-0.dll`, isn't shipped anywhere. This PR enforces static linking so that we don't need to ship any extra dll's

[Image build](https://github.com/grafana/alloy/blob/f001028f246b526c59b9af0149260c58808f79d2/Dockerfile.windows#L39)
This build path has a newer version of the `mingw-64` (migrated from `TDM-GCC`) which OOTB statically links against `libssp` , so we don't get the same issue as above. But the introduction of the new toolchain also introduced a new dynamic dependency that was not included in the container runtime (according to claude: `libwinpthread-1.dll`). This PR enforces static linking in order to prevent those errors.

`TDM-GCC` defaulted to static linking by default, whilst `mingw-w64` defaulted to dynamic linking

BEGIN_COMMIT_OVERRIDE
fix(windows): Statically link Windows builds to fix missing DLL errors on startup (#6972)
END_COMMIT_OVERRIDE

https://github.com/grafana/alloy/issues/6970

### PR Checklist
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/.github/workflows/check-windows-release-build.yml
+++ b/.github/workflows/check-windows-release-build.yml
@@ -1,0 +1,71 @@
+name: Check Windows release build
+
+# Builds alloy-windows-amd64.exe through the same toolchain the real release
+# pipeline uses (grafana/alloy-build-image, cross-compiling from Linux via
+# viceroy), then actually runs it on Windows to make sure it starts
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'Dockerfile.windows'
+      - 'scripts/docker-containers-windows'
+      - '.github/workflows/check-windows-container.yml'
+      - '.github/workflows/publish-alloy.yml'
+      - '.github/workflows/publish-alloy-devel.yml'
+      - '.github/workflows/release-publish-alloy-artifacts.yml'
+      - '.github/workflows/publish-alloy-windows.yml'
+  pull_request:
+    paths:
+      - 'Dockerfile.windows'
+      - 'scripts/docker-containers-windows'
+      - '.github/workflows/check-windows-container.yml'
+      - '.github/workflows/publish-alloy.yml'
+      - '.github/workflows/publish-alloy-devel.yml'
+      - '.github/workflows/release-publish-alloy-artifacts.yml'
+      - '.github/workflows/publish-alloy-windows.yml'
+permissions:
+  contents: read
+
+jobs:
+  build_windows_binary:
+    name: Build alloy-windows-amd64.exe
+    container: grafana/alloy-build-image:v0.1.35@sha256:9fa2a341b53503ce42cf9900c401d689a68ea67cdec6a20f53d72e3665fb8dc6
+    runs-on:
+      labels: github-hosted-ubuntu-x64-large
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
+
+    - name: Set ownership
+      # https://github.com/actions/runner/issues/2033#issuecomment-1204205989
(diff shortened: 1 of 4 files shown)


.github/workflows/check-windows-release-build.yml
```diff
@@ -0,0 +1,75 @@
...
+    needs: [build_windows_binary]
+    runs-on: windows-latest
+    steps:
+    - name: Download alloy-windows-amd64.exe
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+      with:
+        name: windows-binary-smoke-test
+        path: .
+
+    - name: Run alloy.exe --version
```
ptodev, line 70: Why not run this test as part of `check-windows-container.yml`? It already reuses the same workflow as the publish workflow.

.github/workflows/check-windows-release-build.yml
```diff
@@ -0,0 +1,75 @@
...
+      # Scrub PATH down to the stock Windows locations so this
+      # check reflects a plain Windows install, not this runner's tooling.
+      run: |
+        $env:Path = "$env:SystemRoot\System32;$env:SystemRoot;$env:SystemRoot\System32\Wbem;$env:SystemRoot\System32\WindowsPowerShell\v1.0\"
+        .\alloy-windows-amd64.exe --version
```
ptodev: This feels hacky but I'm not sure if there's a better alternative. Maybe something like `dumpbin /dependents alloy-windows-amd64.exe`? Or we could run dumpbin in addition to this?

Dockerfile.windows
```diff
@@ -66,7 +66,7 @@ SHELL ["cmd", "/S", "/C"]
 
 # Creating new layers can be really slow on Windows so we clean up any caches
 # we can before moving on to the next step.
-RUN ""C:\Program Files\git\bin\bash.exe" -c "RELEASE_BUILD=${RELEASE_BUILD} VERSION=${VERSION} GO_TAGS=\"embedalloyui ${GO_TAGS}\" make alloy && rm -rf internal/web/ui/node_modules""
+RUN ""C:\Program Files\git\bin\bash.exe" -c "RELEASE_BUILD=${RELEASE_BUILD} VERSION=${VERSION} GO_TAGS=\"embedalloyui ${GO_TAGS}\" CGO_LDFLAGS=\"-static\" make alloy && rm -rf internal/web/ui/node_modules""
```
ptodev, line 69: It'd be nice to have a comment to clarify why `CGO_LDFLAGS=\"-static\"`  is needed.

dehaansa (approved):

.github/workflows/check-windows-release-build.yml
blewis12: I was thinking to move this there as part of cleanup! I will create a follow up PR

PR #6984, opened 2026-08-26
chore(loki): Remove ConsumeEntry and only work witch batches

kalleep:

Remove `ConsumeEntry` and only support batches in `Consumer` interface

### Pull Request Details

While thinking of potential ways to perform roll-out of the new pipeline architecture that is in progress the safest way would be to use a feature toggle. With that we would support current channel based pipeline and Consumer pipelines for a while. There will be a lot of work refactoring component and it would be ideal if we internally can switch between the two while testing. With this we don't need the temporary `ConsumeEntry` function, whose original purpose  was to be able to change at pipeline level first but keep changes to components minimal at first. Since this is no longer the plan there is no need to have this function since it will just complicate things.

Part of: #4953

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/loki/echo/echo.go
thampiotr: This is duplication of the code from above, maybe we can have a function for this?

thampiotr (dismissed):

internal/component/loki/echo/echo.go
kalleep: It's not fully duplicated since here we only call string once for labels for a stream. But yes I can create a shared function

kalleep: fixed in [3848479](https://github.com/grafana/alloy/pull/6984/commits/3848479ad833ddf6e29a1d3bf50ccbca6942d9bc)

thampiotr (dismissed):

thampiotr (approved):

PR #7001, opened 2026-08-27
CRI stage POC

thampiotr:
Wrote `process2` in `cri.go` to demonstrate a few options.
There are 3 commits currently, each showing a slightly different appraoch building on each other.

### DIFF
--- a/internal/component/loki/process/stages/cri.go
+++ b/internal/component/loki/process/stages/cri.go
@@ -92,8 +92,8 @@ type criStage struct {
 	partialLines map[model.Fingerprint]Entry
 
 	// flushMu is a phase lock for pendingLines, used by process2 and its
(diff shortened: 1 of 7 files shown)


internal/component/loki/process/stages/cri.go
```diff
@@ -123,23 +141,255 @@ func (c *cri) Run(in chan Entry) chan Entry {
...
+	if flushed := c.flushPendingLinesIfExceeded(); len(flushed) > 0 {
+		output = append(flushed, output...)
+	}
+
+	// Send the output to the next stage.
+	if len(output) == 0 {
+		return nil
+	}
+	return c.next(ctx, output)
+}
```
thampiotr, line 243: That's the main part. 
Whatever we do, I would like to have it organised logically into methods like that. It is cheap (happy to add benchmarks) and helps a ton with maintainability.

PR #7002, opened 2026-08-27
feat: POC OTel Engine support bundle implementation

dehaansa:

Adds a `supportbundle` OpenTelemetry Collector extension that serves an HTTP
endpoint returning a zip archive of diagnostics from a running collector:
runtime metadata and telemetry resource attributes, the redacted running
configuration, per-component status, an allowlist of environment variables,
feature-gate states, best-effort self-telemetry metrics, pprof profiles
(heap/goroutine always; CPU and mutex sampled over a window), and — when
configured — collector logs captured during the request. It ships as its own
separately versioned Go module under `extension/supportbundle` and is registered
in the collector distribution.

### Pull Request Details

- Ships as a standalone, separately released Go module (its own `go.mod`, its own
  release-please package mirroring `syntax`), wired into the distro via a local
  replace directive.
- The gatherers live behind a `Gatherer`/`AsyncGatherer` contract in an
  `internal/gather` subpackage; the extension package is just factory/config and
  the HTTP handler.
- Config capture uses the `ConfigSnapshotWatcher` interface and writes only the
  collector's unexpanded (secret-redacted) configuration; the effective/expanded
  config is never written to the bundle. Component status uses
  `componentstatus.Watcher`.
- CPU profiling, mutex sampling, and optional log capture share one collection
  window (`?duration=`, default 30s, configurable max 60s); metrics are sampled at the start
  and end of the window so counter deltas can be computed. Bundle generation is
  serialized process-wide.
- Log capture is opt-in via a registered zap sink (`supportbundle://` in
  `service::telemetry::logs::output_paths`); its idle path is lock-free so it
  adds no steady-state logging overhead.
- Security: the endpoint has no authentication by default and binds to loopback;
  `logs.txt` and captured environment variables can contain secrets. This is
  documented in the extension README.

### PR Checklist

- [X] Documentation added
- [X] Tests updated
- [X] This pull request was substantially generated with AI assistance (see the [GenAI policy](https://github.com/grafana/alloy/blob/main/docs/developer/genai.md))


dehaansa:
An example config to see all the behavior of the support bundle:

```
extensions:
  basicauth:
    client_auth:
      username: admin
      password: admin

  health_check:
    endpoint: localhost:13133

  supportbundle:
    endpoint: localhost:55680

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:4317
      http:
        endpoint: localhost:4318

  hostmetrics:
    collection_interval: 10s
    scrapers:
      cpu:
      load:
      memory:
      network:
      paging:
      filesystem:

  prometheus:
    config:
      scrape_configs:
        - job_name: otelcol-self
          scrape_interval: 15s
          static_configs:
            - targets: [localhost:8888]

processors:
  memory_limiter:
    check_interval: 1s
    limit_percentage: 80
    spike_limit_percentage: 20

  batch:
    send_batch_size: 512
    timeout: 5s

  attributes/tag:
    actions:
      - key: demo.source
        value: zpages-demo
        action: insert

  probabilistic_sampler:
    sampling_percentage: 100

  transform/logs:
    log_statements:
      - context: log
        statements:
          - set(attributes["demo.seen_by"], "transform")

  filter/drop_noisy:
    error_mode: ignore
    metrics:
      metric:
        - 'name == "system.paging.faults"'

connectors:
  spanmetrics:
    histogram:
      explicit:
        buckets: [1ms, 10ms, 100ms, 1s, 10s]
    metrics_flush_interval: 15s

exporters:
  debug:
    verbosity: basic
    sampling_initial: 2
    sampling_thereafter: 500

  otlp_http/unreachable:
    endpoint: localhost:4319
    auth:
      authenticator: basicauth
    tls:
      insecure: true
    timeout: 2s
    retry_on_failure:
      enabled: false
    sending_queue:
      enabled: false

service:
  extensions: [health_check, supportbundle, basicauth]

  pipelines:
    traces:
      receivers: [otlp]
      processors: [memory_limiter, probabilistic_sampler, attributes/tag, batch]
      exporters: [debug, spanmetrics]

    traces/passthrough:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp_http/unreachable]

    metrics/spanmetrics:
      receivers: [spanmetrics]
      processors: [batch]
      exporters: [debug]

    metrics:
      receivers: [hostmetrics, prometheus, otlp]
      processors: [memory_limiter, filter/drop_noisy, attributes/tag, batch]
      exporters: [debug]

    logs:
      receivers: [otlp]
      processors: [memory_limiter, transform/logs, batch]
      exporters: [debug]

  telemetry:
    logs:
      level: info
      output_paths: [stderr, supportbundle://]

    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: localhost
                port: 8888

    traces:
      level: detailed
      propagators: [tracecontext, b3]
      processors:
        - batch:
            schedule_delay: 2000
            exporter:
              otlp:
                protocol: http/protobuf
                endpoint: http://localhost:4318
                insecure: true

```

extension/supportbundle/config.go
```diff
@@ -0,0 +1,74 @@
...
+
+type Config struct {
+	confighttp.ServerConfig `mapstructure:",squash"`
+
+	// Path is the HTTP path that serves the support bundle.
+	Path string `mapstructure:"path"`
+
+	// CollectionDuration is the default window for windowed collectors, such as
+	// the CPU profile.
+	CollectionDuration time.Duration `mapstructure:"collection_duration"`
```
thampiotr: Maybe we call this DefaultCollectionDuration, as it's overridable in the URL query.

extension/supportbundle/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Log capture
+
+The extension registers a zap log sink named `supportbundle`. To capture collector logs in the bundle, add `supportbundle://` to `service::telemetry::logs::output_paths`. Keep the existing output paths so logs still reach their normal destination.
+
+```yaml
+service:
+  telemetry:
+    logs:
+      output_paths: [stderr, supportbundle://]
```
thampiotr, line 115: That's nice! I think we had an idea to include logs from the past up to 1MiB or some configurable limit. With this it seems it should be doable if we want to :) Should we consider this option right away? 30s of logs is not that much.

extension/supportbundle/README.md
```diff
@@ -0,0 +1,168 @@
...
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [debug]
+```
+
+See the [`confighttp`][confighttp] documentation for every available server field.
+
+### Log capture
```
thampiotr, line 103: Could we use similar trick to capture traces? may be useful, not something needed from day one.

extension/supportbundle/README.md
```diff
@@ -0,0 +1,168 @@
...
+otelcol-support-bundle/
+├── metadata.yaml
+├── config.yaml            # present once the collector sends a config snapshot
+├── component-status.yaml  # health of each pipeline component
+├── environment.txt        # allowlisted environment variables that are set
+├── feature-gates.txt      # every feature gate and its state
+├── metrics-start.txt      # metrics at window start (present with a pull endpoint and duration > 0)
+├── metrics-end.txt        # metrics at window end (present with a pull endpoint and duration > 0)
+├── metrics.txt            # single metrics sample (present with a pull endpoint and duration = 0)
+├── pprof/
```
thampiotr, line 180: Had to check this, but it seems indeed the case: pprof is available in every OTel collector distro based on core.

collector/builder-config.yaml
```diff
@@ -9,6 +9,10 @@ extensions:
   - gomod: github.com/grafana/alloy v1.19.0 # x-release-please-version
     import: github.com/grafana/alloy/extension/alloyengine
     name: alloyengine
+  # supportbundle is a separately released module. Its version is managed by its
+  # own release-please package, so it does not carry the x-release-please-version
+  # marker (which tracks the root module version).
+  - gomod: github.com/grafana/alloy/extension/supportbundle v0.1.0
```
jharvey10, line 15: Ooo this makes me realize that there's a deficiency here with how release versions propagate: We can't do multiple modules using `x-release-please-version` alone (as I'm sure you discovered).

I may try to implement a little utility or release-please plugin to sync these versions from all of our extensions so it can work automatically for more than just the alloyengine extension.

In the absence of that, we'll have to hand-update this every time a new version goes out.

extension/supportbundle/README.md
dehaansa: Added a ring buffer for logs to capture a configured amount of logs from before the bundle.

extension/supportbundle/README.md
dehaansa: Added a trace capture similar to how its implemented in zpages. 👍

thampiotr:
What's the plan with this one? should we continue revising this to land? or keep it as a POC? I think it's achievable to get it into a working state and add docs in follow-up?

Issue #4725, opened 2025-10-30
[otel-engine] Service installations


jharvey10:
### Add a description

Ensure service installations of Alloy work: systemd, windows, etc - make sure we can configure the OTel engine

### Tip

<sub>React with 👍 if this issue is important to you.</sub>

blewis12:
This [open PR](https://github.com/grafana/alloy/pull/6551) relates to this as it will allow us to enable the OTel engine through homebrew installations via an environment variable

blewis12:
Currently writing up a design document to go over the current state of enabling the OTel engine via service installations and proposals on how to fill remaining gaps - there are some parts here that are intertwined with the [supervisor ergonomics](https://github.com/grafana/alloy/issues/6416) work as well so this should be ironed out a bit more before moving on further on that side

blewis12:
^^ I figured a design doc might be overkill, so instead went with creating some PoC's that i'll link here. As part of this work i'm considering, going off our [installation docs](https://grafana.com/docs/alloy/latest/set-up/install/), the following process managers: `systemd` (deb/rpm installations), `launchd` (brew installations), and the `windows service manager` (windows installations). 

For each of those I am considering three different points: 
1. Toggling between engines using some environment variable
2. Ensuring that the toggle also accounts for switching betweem `.alloy` and `.yaml` config, that user can also edit
3. Ensuring that the user can also pass through any extra args to the running service in OTel mode

Launchd/homebrew already has a working solution to this (done here https://github.com/grafana/alloy/pull/6723) and so for the other two managers I want there to be parity, just so there's not a huge amount of effort for users to learn how it works in each manager. Everything in the `launchd` column is already done/released prior to this - though I can't find documentation on it.

```
+====================+==================================================+=====================================================+===================================================+
|                    |                     Launchd                      |                       Systemd                       |                  Windows service                  |
+====================+==================================================+=====================================================+===================================================+
| Engine toggle      | ALLOY_OTEL_MODE in config.env                    | ALLOY_OTEL_MODE in the environment file             | ALLOY_OTEL_MODE, a new registry value             |
|                    |                                                   | (/etc/default/alloy or /etc/sysconfig/alloy)        | under HKLM\Software\GrafanaLabs\Alloy             |
+--------------------+--------------------------------------------------+-----------------------------------------------------+---------------------------------------------------+
[...]

PR #7006, opened 2026-08-28
feat: OTel service installation support for systemd 

blewis12:
This PR introduces a new wrapper script that will be used in `systemd` installations to launch alloy and toggle between the Default and OTel engines based on environment variables the user sets. 

Part of https://github.com/grafana/alloy/issues/4725

### Pull Request Details
A simplified image of the flow:

```
        +-----------+
        |  systemd  |
        +-----------+
              |
              v
       +-----------------+
       |  alloy-wrapper  |
       +-----------------+
          |           |
          v           v
   +-----------+ +------------+
   | alloy run | | alloy otel |
   +-----------+ +------------+
```

Instead of systemd directly invoking the alloy binary, which currently hardcodes to the `run` sub command, we can introduce a new environment variable called `ALLOY_OTEL_MODE` (mirroring the same name of the env var introduced in the launchd service installation to achieve the same thing). Systemd can then instead invoke a wrapper script (`packaging/systemd/alloy-wrapper`) which uses the `ALLOY_OTEL_MODE` to branch between either the `run` or `otel` sub command.

Currently, an env variable called `CONFIG_FILE` exists which the user can set to override the default config path of `/etc/alloy/config.alloy`. This PR introduces a sister env variable called `OTEL_CONFIG_FILE` that the user can set to define a custom OTel YAML config file, the default value being `/etc/alloy/config.yaml`.

Lastly, the current implementation is such that a user can set `CUSTOM_ARGS` to define any extra flags to pass through as argv. This PR introduces a sister env variable called `OTEL_CUSTOM_ARGS` to achieve the same for the OTel engine.

A user can define env variable values for both engines - and then toggle between engines by switching `ALLOY_OTEL_MODE`. 

This is also designed to be backwards compatible. If a user hasn't set any value for `ALLOY_OTEL_MODE` then we fall back to the default engine, so current installations should remain the same on upgrade.

I've purposely left docs out of this, and would like to add them once the windows implementation is done (aka the remaining PR) so that docs can be drafted with a complete picture. LMK if you disagree with that approach though

https://github.com/user-attachments/assets/306d7932-055b-4b3f-adcb-54e3172671e8

### PR Checklist
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/tools/packaging_test/alloy_linux_packages_test.go
+++ b/internal/tools/packaging_test/alloy_linux_packages_test.go
@@ -158,6 +158,21 @@ func (env *AlloyEnvironment) TestEngineToggle(t *testing.T) {
 			env:      `CONFIG_FILE=/custom/config.alloy ALLOY_OTEL_MODE=1`,
 			expected: "otel --config=/etc/alloy/config.yaml\n",
 		},
+		{
+			name:     "otel engine, CUSTOM_OTEL_ARGS applies",
+			env:      `CONFIG_FILE=/etc/alloy/config.alloy ALLOY_OTEL_MODE=1 CUSTOM_OTEL_ARGS="--set=processors.batch.timeout=2s"`,
+			expected: "otel --set=processors.batch.timeout=2s --config=/etc/alloy/config.yaml\n",
+		},
+		{
+			name:     "otel engine ignores CUSTOM_ARGS, the default engine's flags",
+			env:      `CONFIG_FILE=/etc/alloy/config.alloy ALLOY_OTEL_MODE=1 CUSTOM_ARGS="--storage.path=/should/not/appear"`,
+			expected: "otel --config=/etc/alloy/config.yaml\n",
+		},
+		{
+			name:     "default engine ignores CUSTOM_OTEL_ARGS, the OTel engine's flags",
+			env:      `CONFIG_FILE=/etc/alloy/config.alloy CUSTOM_OTEL_ARGS="--set=processors.batch.timeout=2s"`,
+			expected: "run --storage.path=/var/lib/alloy/data /etc/alloy/config.alloy\n",
+		},
 	}
 
 	for _, tc := range tt {
--- a/internal/tools/packaging_test/alloy_linux_packages_test.go
+++ b/internal/tools/packaging_test/alloy_linux_packages_test.go
@@ -130,8 +130,8 @@ func (env *AlloyEnvironment) TestEngineToggle(t *testing.T) {
 	res = env.ExecScript(`[ -f /etc/alloy/config.yaml ]`)
 	require.Equal(t, 0, res.ExitCode, "expected the default OTel engine config to be installed")
 
-	res = env.ExecScript(`f=/etc/default/alloy; [ -f "$f" ] || f=/etc/sysconfig/alloy; grep -qF 'ALLOY_OTEL_MODE=""' "$f" && grep -qF 'ALLOY_CONFIG=""' "$f"`)
-	require.Equal(t, 0, res.ExitCode, "expected the installed environment file to declare ALLOY_OTEL_MODE and ALLOY_CONFIG as empty by default")
+	res = env.ExecScript(`f=/etc/default/alloy; [ -f "$f" ] || f=/etc/sysconfig/alloy; grep -qF 'ALLOY_OTEL_MODE=""' "$f"`)
+	require.Equal(t, 0, res.ExitCode, "expected the installed environment file to declare ALLOY_OTEL_MODE as empty by default")
 
 	tt := []struct {
 		name     string
@@ -144,14 +144,19 @@ func (env *AlloyEnvironment) TestEngineToggle(t *testing.T) {
 			expected: "run --storage.path=/var/lib/alloy/data /etc/alloy/config.alloy\n",
 		},
 		{
-			name:     "otel engine, default sibling config",
+			name:     "default engine, custom CONFIG_FILE",
+			env:      `CONFIG_FILE=/custom/config.alloy`,
+			expected: "run --storage.path=/var/lib/alloy/data /custom/config.alloy\n",
+		},
+		{
+			name:     "otel engine, default config",
 			env:      `CONFIG_FILE=/etc/alloy/config.alloy ALLOY_OTEL_MODE=1`,
 			expected: "otel --config=/etc/alloy/config.yaml\n",
 		},
 		{
-			name:     "otel engine, ALLOY_CONFIG override",
-			env:      `CONFIG_FILE=/etc/alloy/config.alloy ALLOY_OTEL_MODE=1 ALLOY_CONFIG=/custom/config.yaml`,
-			expected: "otel --config=/custom/config.yaml\n",
+			name:     "otel engine ignores a custom CONFIG_FILE, always uses the installed config.yaml",
+			env:      `CONFIG_FILE=/custom/config.alloy ALLOY_OTEL_MODE=1`,
+			expected: "otel --config=/etc/alloy/config.yaml\n",
 		},
 	}
 
--- a/internal/tools/packaging_test/alloy_linux_packages_test.go
+++ b/internal/tools/packaging_test/alloy_linux_packages_test.go
@@ -31,6 +31,7 @@ func TestAlloyLinuxPackages(t *testing.T) {
 		{"install package", (*AlloyEnvironment).TestInstall},
 		{"ensure existing config doesn't get overridden", (*AlloyEnvironment).TestConfigPersistence},
 		{"test data folder permissions", (*AlloyEnvironment).TestDataFolderPermissions},
+		{"test engine toggle", (*AlloyEnvironment).TestEngineToggle},
 
 		// TODO: a test to verify that the systemd service works would be nice, but not
 		// required.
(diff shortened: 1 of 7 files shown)


thampiotr:
Since this is not fully tested in CI, were you able to try `make dist/alloy-linux-amd64 && make test-packages` or something similar and verify this a bit further manually?

packaging/environment-file
```diff
@@ -6,11 +6,26 @@
...
+# The configuration file holding the Alloy Default Engine config.
+# when ALLOY_OTEL_MODE=1; use OTEL_CONFIG_FILE
 CONFIG_FILE="/etc/alloy/config.alloy"
 
-# User-defined arguments to pass to the run command.
+# The configuration file holding the Alloy OTel Engine config.
+# when ALLOY_OTEL_MODE=0; use CONFIG_FILE
+OTEL_CONFIG_FILE="/etc/alloy/config.yaml"
+
+# User-defined arguments to pass to the default engine (`alloy run`). Has
```
thampiotr:
```suggestion
# User-defined arguments to pass to the default engine (`alloy run`).
```

packaging/environment-file
```diff
@@ -6,11 +6,26 @@
...
+# when ALLOY_OTEL_MODE=1; use OTEL_CUSTOM_ARGS
 CUSTOM_ARGS=""
 
+# User-defined arguments to pass to the OTel engine (`alloy otel`). Only
```
thampiotr:
```suggestion
# User-defined arguments to pass to the OTel engine (`alloy otel`).
```

packaging/environment-file
```diff
@@ -6,11 +6,26 @@
...
+# when ALLOY_OTEL_MODE=1; use OTEL_CONFIG_FILE
 CONFIG_FILE="/etc/alloy/config.alloy"
 
-# User-defined arguments to pass to the run command.
+# The configuration file holding the Alloy OTel Engine config.
+# when ALLOY_OTEL_MODE=0; use CONFIG_FILE
+OTEL_CONFIG_FILE="/etc/alloy/config.yaml"
+
+# User-defined arguments to pass to the default engine (`alloy run`). Has
+# when ALLOY_OTEL_MODE=1; use OTEL_CUSTOM_ARGS
```
thampiotr: I think something is missing after `use OTEL_CUSTOM_ARGS` here and below

packaging/environment-file
```diff
@@ -6,11 +6,26 @@
...
 CUSTOM_ARGS=""
 
+# User-defined arguments to pass to the OTel engine (`alloy otel`). Only
+# when ALLOY_OTEL_MODE=0; use CUSTOM_ARGS
+OTEL_CUSTOM_ARGS=""
+
+# Set to (lowercase) 1, true, yes, or on to run the OTel engine (`alloy otel`)
+# instead of the default engine (`alloy run`). Unset or any other value keeps
```
thampiotr: > Unset or any other value keeps the default engine

but above we talk about `ALLOY_OTEL_MODE=0/1`... so makes me wonder. I think in practice `0` isn't a special value. Maybe it's okay, but there are more ways to set or not set it.

thampiotr:
There's a sentence in otel_engine.md that we may want to update:

> Service installation support for systemd, launchd, and similar systems isn't included in the initial experimental release.

packaging/systemd/alloy-wrapper
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+set -euf
+
+ALLOY_BIN="${ALLOY_BIN:-/usr/bin/alloy}"
+
+otel_mode=""
+case "${ALLOY_OTEL_MODE:-}" in
+  # check against truthy values
+  1 | true | yes | on) otel_mode="1" ;;
```
thampiotr, line 9: Note this will be case-sensitive match. Maybe that's okay.

packaging/systemd/alloy-wrapper
```diff
@@ -0,0 +1,21 @@
...
+esac
+
+if [ -n "$otel_mode" ]; then
+  OTEL_CONFIG_FILE="${OTEL_CONFIG_FILE:-/etc/alloy/config.yaml}"
+  # shellcheck disable=SC2086
```
thampiotr, line 17: Should we log/echo which mode is starting or something like that?

internal/tools/packaging_test/alloy_linux_packages_test.go
```diff
@@ -103,11 +104,107 @@ func (env *AlloyEnvironment) TestConfigPersistence(t *testing.T) {
...
+			expected: "run --storage.path=/var/lib/alloy/data /etc/alloy/config.alloy\n",
+		},
+		{
+			name:     "default engine, custom CONFIG_FILE",
+			env:      `CONFIG_FILE=/custom/config.alloy`,
+			expected: "run --storage.path=/var/lib/alloy/data /custom/config.alloy\n",
+		},
+		{
+			name:     "otel engine, default config",
+			env:      `CONFIG_FILE=/etc/alloy/config.alloy ALLOY_OTEL_MODE=1`,
```
thampiotr, line 156: Would be good to test the other ways to set the variable than just `1`

packaging/systemd/alloy-wrapper
thampiotr, line 17: I liked a comment like this: 

>       # extra_args is intentionally unquoted so a file with multiple arguments
>      # word-splits into separate argv entries.

packaging/environment-file
blewis12: ah sorry, I iterated on this a couple times and so the comment is out of date - it originally only accepted 0/1 but then I realized the launchd implementation was anything truthy. I'll align the comments back

packaging/systemd/alloy-wrapper
blewis12: the [launchd/homebrew](https://github.com/grafana/alloy/blob/afce20979892ecc8bb0f9acab5e9fbc0dd872e37/packaging/homebrew/alloy.rb.tpl#L69-L71) implementation is also case sensitive which is why I opted for that here - but I don't seem the harm in making it case insensitive just to avoid confusion, and modify it in the other .tpl file too

packaging/systemd/alloy-wrapper
blewis12: yeah that's a good idea 👍

packaging/systemd/alloy-wrapper
thampiotr: I think this is actually very minor, but would be good to keep everything consistent. So if you can make it easily be case insensitive everywhere that's cool, but otherwise it can stay the same :)

packaging/environment-file
blewis12: ok, i've re-jigged it a bit now

blewis12:
> Since this is not fully tested in CI, were you able to try `make dist/alloy-linux-amd64 && make test-packages` or something similar and verify this a bit further manually?

yeah, `test-packages` is what i've been appending to and is what i'm using to assert that the state on the host machine is what you'd expect after unpackaging the deb/rpm artifact with the changes. We don't have any sort of integration/smoke test for systemd itself, so for that I actually spun up a container - installed systemd - installed the deb/rpm package - and played around with it. I've attached a video to the PR description so you can see what I mean.

I'm looking into whether we could setup some sort of CI smoke test for this so that we don't have to rely on manual QA

thampiotr:
Awesome, thanks @blewis12! The video looks 🥇 Let me know when ready for final review.

packaging/systemd/alloy-wrapper
blewis12: alright, decided to make it case insensitive 👍 I don't see why not at this point

blewis12:
Hi @thampiotr - this is ready for a re-review :) 

Whilst looking into smoke testing, I actually realized that our `make test-packages` has been [failing](https://github.com/grafana/alloy/actions/workflows/test_linux_system_packages.yml?query=branch%3Amain). The reason for that is that our debian image base, `debian:buster-20240612`, reached EOL as of 2024 and the build stopped working at all as we were getting 404's when running `apt-get update`. I updated the image base to a successor `debian:12-slim`. After that the tests themselves passed. For the RPM test path, we were using `centos:8` which is also EOL as of 2021, so I decided to also update this to rockylinux` which is actively maintained. 

I added smoke tests which actually install (via rpm or deb) and launch alloy with `systemd`, and check that it's running and that it was ran with the expected arguments. It's deliberately kept minimal, with the installation-only path doing most of the heavy lifting for argument parsing checks. I just wanted something that ran the e2e service flow.

> There's a sentence in otel_engine.md that we may want to update:

I was thinking to do that as a follow up PR, since launchd is also not documented but is released it would be nice to also document that. I don't want to crowd this PR with too many changes though, wdyt?

internal/tools/packaging_test/environment_test.go
```diff
@@ -83,18 +102,22 @@ func DEBEnvironment(t *testing.T, packageName string, pool *dockertest.Pool) Env
...
 
+	// The images run systemd as PID 1 instead of a shell, so the tests can
+	// exercise the installed alloy.service
 	container, err := pool.BuildAndRunWithOptions(
 		dockerfile,
 		&dockertest.RunOptions{
 			Name:       name,
-			Entrypoint: []string{"/bin/bash"},
-			Tty:        true,
-			Mounts:     []string{"/sys/fs/cgroup:/sys/fs/cgroup:ro"},
```
blewis12, line 92: `Entrypoint` and `Tty` was no longer necessary, and from browsing some issues online I discovered that manually mounting `/sys/fs/cgroup` can cause issues when running systemd in a container

.github/workflows/test_linux_system_packages.yml
```diff
@@ -8,6 +8,14 @@ on:
     - internal/tools/packaging_test/**
     - Makefile
     - build-tools/make/*.mk
+    - .github/workflows/test_linux_system_packages.yml
+  pull_request:
```
blewis12, line 12: These tests only ran on `main` which I think makes visibility quite low, i'd much rather have a feedback loop happen in a relevant PR

thampiotr:
> I was thinking to do that as a follow up PR, since launchd is also not documented but is released it would be nice to also document that. I don't want to crowd this PR with too many changes though, wdyt?

sounds good to me!

packaging/systemd/alloy-wrapper
```diff
@@ -0,0 +1,25 @@
...
+
+ALLOY_BIN="${ALLOY_BIN:-/usr/bin/alloy}"
+
+otel_mode=""
+case "$(printf '%s' "${ALLOY_OTEL_MODE:-}" | tr '[:upper:]' '[:lower:]')" in
+  # check against truthy values
+  1 | true | yes | on) otel_mode="1" ;;
+esac
+
+# extra_args is intentionally unquoted so a file with multiple arguments
```
thampiotr:
```suggestion
# OTEL_CUSTOM_ARGS is intentionally unquoted so a file with multiple arguments
```

thampiotr (approved):

PR #7007, opened 2026-08-28
feat: OTel service installation support for windows service

blewis12:
Implements the ability for users to toggle between the default engine and the otel engine in windows service installations

### Pull Request Details
Part of https://github.com/grafana/alloy/issues/4725

Mirroring the systemd and launchd implementations, this PR introduces the registry value `ALLOY_OTEL_MODE` which, when empty or set to falsy, will mean that the windows service launches `alloy run...`. If the user toggles this flag to a truthy value (true|1|yes, case insensitive) then the OTel engine will run with the config defined at `$INSTDIR\config.yaml`. The user can also specify the `OTelArguments` registry value in which they can pass through extra arguments to the `otel` command after the default `--config` argv. 

### Notes to the Reviewer
We already have some integration tests set up, so i extended those to smoke test toggling between the engines. Once we have a RC published I will also do some manual testing with the installer on our windows VM

Much like the systemd [PR](https://github.com/grafana/alloy/pull/7006), this PR does not include documentation. I will create a follow up PR which includes documentation for all our service installations that can be reviewed independently. 

### PR Checklist
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/windows-service/windows_service_test.go
+++ b/integration-tests/windows-service/windows_service_test.go
@@ -99,6 +99,70 @@ func TestWindowsService(t *testing.T) {
 	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		util.AssertEventLogLine(c, "msg=\"{^_^} Alloy is running\"")
 	}, waitTimeout*waitAttempts, waitTimeout)
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		v, err := util.RegistryStringValue(registryPath, "ALLOY_OTEL_MODE")
+		assert.NoError(c, err)
+		assert.Equal(c, "", v)
+	}, waitTimeout*waitAttempts, waitTimeout)
+}
+
+func TestWindowsServiceOtelMode(t *testing.T) {
+	installerPath := os.Getenv(envVarInstallerPath)
+	if installerPath == "" {
+		t.Fatalf("%s not set; skipping Windows service integration test", envVarInstallerPath)
+	}
+	if _, err := os.Stat(installerPath); err != nil {
+		t.Fatalf("%s %q not found: %v", envVarInstallerPath, installerPath, err)
+	}
+
+	uninstallerPath := filepath.Join(installDir, "uninstall.exe")
+	cleanup := os.Getenv(envVarStateful) != "true"
+	if cleanup {
+		t.Logf("Stateful mode: skipping cleanup (service will remain installed) env=%s", envVarStateful)
+	} else {
+		defer uninstallAlloy(t, uninstallerPath)
+	}
+
+	if isAlloyInstalled(t, installDir) {
+		cleanIfExists := os.Getenv(envVarCleanIfExists) == "true"
+		if !cleanIfExists {
+			t.Fatalf("Alloy already present on the system. Uninstall manually or set %s=true to remove and continue", envVarCleanIfExists)
+		}
+		t.Logf("Alloy already present on the system. Uninstalling...")
+		uninstallAlloy(t, uninstallerPath)
+		if isAlloyInstalled(t, installDir) {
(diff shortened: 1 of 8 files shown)


integration-tests/windows-service/windows_service_test.go
```diff
@@ -46,9 +46,9 @@ func TestWindowsService(t *testing.T) {
 	uninstallerPath := filepath.Join(installDir, "uninstall.exe")
 	cleanup := os.Getenv(envVarStateful) != "true"
 	if cleanup {
-		t.Logf("Stateful mode: skipping cleanup (service will remain installed) env=%s", envVarStateful)
-	} else {
```
blewis12, line 50: this if statement was inverted, fixing this sorted out errors that popped up when running multiple statement tests in the same suite

thampiotr (dismissed):
It's all looking good, no comments 🎉 
Hope we have a good way to test it on some Windows machine.

integration-tests/windows-service/util/service.go
```diff
@@ -90,3 +100,30 @@ func EnsureServiceRunning(c *assert.CollectT, t *testing.T, serviceName string)
...
+		if status.State != svc.StopPending {
+			t.Logf("Stopping service (not stopped)")
+			if _, err := s.Control(svc.Stop); err != nil {
+				t.Logf("Stop failed err=%v", err)
+				assert.NoError(c, err, "stop Alloy service")
+				return
+			}
+			t.Logf("Stop requested successfully")
+		} else {
+			t.Logf("Service is stop pending, waiting")
```
thampiotr, line 124: Ah, just noticed - we are not waiting for anything here actually - is that okay? 

When called from Eventually functions, I think it will be considered success, so it may consider stopped, while it is stop pending, which could result in some test race condition and flakiness?

integration-tests/windows-service/windows_service_test.go
```diff
@@ -31,31 +31,25 @@ const (
...
 
-	uninstallerPath := filepath.Join(installDir, "uninstall.exe")
-	cleanup := os.Getenv(envVarStateful) != "true"
-	if cleanup {
+	uninstallerPath = filepath.Join(installDir, "uninstall.exe")
+	if os.Getenv(envVarStateful) == "true" {
 		t.Logf("Stateful mode: skipping cleanup (service will remain installed) env=%s", envVarStateful)
 	} else {
-		defer uninstallAlloy(t, uninstallerPath)
+		t.Cleanup(func() { uninstallAlloy(t, uninstallerPath) })
```
thampiotr: Should we not uninstall if the install failed or Alloy is already installed?

integration-tests/windows-service/util/service.go
blewis12: ahh yes, the implementation isn't right 🤦 both both the pre-existing "running" and stopped" methods. I'll push a fix so that it actually asserts the state is expected at the end of the method

integration-tests/windows-service/windows_service_test.go
blewis12: that's true, I can tack on a `if (isAlloyInstalled)` there to avoid errors

integration-tests/windows-service/util/service.go
blewis12: https://github.com/grafana/alloy/pull/7007/changes/2562f28feb81e77af54b68da98922ded6e794ad4

integration-tests/windows-service/windows_service_test.go
blewis12: https://github.com/grafana/alloy/pull/7007/changes/b6f96a5b075f67f2474e0864ca0b413dc244a1c6

thampiotr (approved):

PR #7031, opened 2026-09-02
fix(otelcol.auth.google): Set valid authorization token header

odam06:

  Fixes a regression in otelcol.auth.google introduced by an upstream OpenTelemetry dependency upgrade.

  Upstream commit e2a9f694 (https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/commit/e2a9f6942a31c3092247b4c42b905cc4b18b4ee1) added the
  token_header field and validation requiring either authorization or proxy-authorization. This change was released in opentelemetry-operations-go v0.56.0
  (https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/releases/tag/v0.56.0).

  Alloy constructs the upstream configuration directly rather than through its default configuration factory, leaving the new field empty and causing
  previously valid configurations to fail validation.

  The converter now explicitly sets token_header to the upstream default, authorization, preserving the existing Alloy API and ensuring Google access and ID
  tokens continue to work with HTTP and gRPC exporters. Tests cover ID tokens, access tokens, default configuration, and OTLP handler compatibility.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/auth/google/google.go
+++ b/internal/component/otelcol/auth/google/google.go
@@ -42,6 +42,10 @@ type Arguments struct {
 	// default: access_token
 	TokenType string `alloy:"token_type,attr,optional"`
 
+	// TokenHeader controls which HTTP header carries the token. It must be
+	// authorization or proxy-authorization. The default is authorization.
+	TokenHeader string `alloy:"token_header,attr,optional"`
+
 	// Audience specifies the audience claim used for generating ID token.
 	Audience string `alloy:"audience,attr,optional"`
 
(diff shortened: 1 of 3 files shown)


internal/component/otelcol/auth/google/google.go
```diff
@@ -68,6 +74,7 @@ func (args *Arguments) SetToDefault() {
 		"https://www.googleapis.com/auth/trace.append",
 	}
 	args.TokenType = upstreamDefault.TokenType
+	args.TokenHeader = defaultTokenHeader
```
kalleep: I would imagine that using upstreamDefault.TokenHeader would be better here instead of hardcoding default in alloy?

odam06: You're right, upstreamDefault is a better fit here. Fixed.

dehaansa (approved):
LGTM. Assuming CI agrees this should be fine to merge. Thanks for the contribution!

dehaansa:
govulncheck failure is unrelated to the PR, merging.

odam06:
Thanks for the approval and the merge @dehaansa

PR #7039, opened 2026-09-04
test(prometheus.exporter.cadvisor): Add integration test

dehaansa:

Adds a k8s integration test for `prometheus.exporter.cadvisor` under `integration-tests/k8s/tests/prometheus-exporter-cadvisor`. The test scrapes a single cAdvisor exporter into Mimir and asserts that cAdvisor metrics appear. cAdvisor uses its raw cgroup driver, so the test needs a privileged container but no container runtime.

### Pull Request Details

Purpose: add end-to-end coverage for the cAdvisor exporter so future bumps of the `grafana/cadvisor` fork are guarded against regressions in the exporter's collectors.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


integration-tests/docker/tests/cadvisor/config.alloy
```diff
@@ -0,0 +1,25 @@
+// No container runtime sockets are configured. cAdvisor uses its raw cgroup
+// driver, which reads /sys/fs/cgroup. This needs a privileged container but no
+// docker or containerd runtime. See test.yaml.
```
kgeckhart: Maybe I missed it but I don't see anything here about raw cgroup drivers, probably best to nix this one and keep the other.

If it was AI, I had to add a rule which forces AI to only put why comments next to the piece which needs the "why" explained.

integration-tests/docker/tests/cadvisor/cadvisor_metrics_test.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"runtime"
+	"testing"
+
+	"github.com/grafana/alloy/integration-tests/docker/common"
+)
+
+func TestCadvisorMetrics(t *testing.T) {
+	// cAdvisor only runs on Linux. The test exercises the Grafana cAdvisor fork's
+	// collectors, so it must run against a real Linux cgroup tree.
```
kgeckhart: Is this truly a limitation of our fork or is it true of cadvisor in general?

integration-tests/docker/tests/cadvisor/config.alloy
dehaansa: Ope yeah, AI, I'll re-review that 😞

integration-tests/docker/tests/cadvisor/cadvisor_metrics_test.go
dehaansa: Not a limitation of our fork, but I see how the comment reads that way. Will rewrite.

integration-tests/k8s/tests/prometheus-exporter-cadvisor/config/alloy-values.yaml
```diff
@@ -0,0 +1,37 @@
+# cAdvisor reads the node's cgroup tree and machine stats. The Alloy pod runs
+# privileged and mounts the host paths read-only. This follows the canonical
+# cAdvisor DaemonSet, but does not mount /var/run. A read-only /var/run blocks
+# Kubernetes from mounting the service account token and the pod fails to start.
+# The raw driver does not need /var/run; only the docker and containerd plugins
+# use it, and this test does not use them.
```
thampiotr:
```suggestion
# This follows the canonical
# cAdvisor DaemonSet, but does not mount /var/run. A read-only /var/run blocks
# Kubernetes from mounting the service account token and the pod fails to start.
# The raw driver does not need /var/run; only the docker and containerd plugins
# use it, and this test does not use them.
```

Nit: the first two sentences are obvious in code. The rest is useful, but maybe should be closer to the mount line?

integration-tests/k8s/tests/prometheus-exporter-cadvisor/config/alloy-values.yaml
```diff
@@ -0,0 +1,37 @@
...
+        hostPath:
+          path: /dev/disk
+
+alloy:
+  stabilityLevel: experimental
+  initialDelaySeconds: 1
+  # The raw cgroup driver needs a privileged container to read /sys/fs/cgroup.
+  securityContext:
+    privileged: true
+  mounts:
```
thampiotr, line 33: I did some digging with an agent and I think that the current test returns metrics without container labels. We could also add container labels by adding the mount config:
```
controller:
  volumes:
    extra:
      - name: containerd-sock
        hostPath:
          path: /run/containerd/containerd.sock
          type: Socket
alloy:
  mounts:
    extra:
      - name: containerd-sock
        mountPath: /run/containerd/containerd.sock
        readOnly: true
```
And we would also need to add some extra assertion methods to verify labels exist (or match a value) on metrics, not just that metrics exist. 

Here's a little table of what I think we can expect to see:

label | current PR | with socket mounted |
|---|---|---|
| `name` | *(empty)* | container IDs |
| `image` | *(empty)* | `docker.io/grafana/alloy:latest`, `docker.io/grafana/mimir:2.17.9`, `docker.io/kindest/kindnetd:v20240813-c6f155d6`, … |
| `container_label_io_kubernetes_pod_name` | *(empty)* | `coredns-6f6b679f8f-pj72q`, `etcd-…-control-plane`, `kindnet-hhs62`, … |
| `container_label_io_kubernetes_pod_namespace` | *(empty)* | `kube-system`, `local-path-storage`, `test-prometheus-exporter-cadvisor` |
| `container_label_io_kubernetes_container_name` | *(empty)* | `alloy`, `coredns`, `etcd`, `kube-apiserver`, … (12)

integration-tests/k8s/deps/mimir.go
```diff
@@ -136,6 +152,72 @@ func (m *Mimir) QueryMetrics(t *testing.T, testName string, expectedMetrics []st
...
+			}
+			require.Greaterf(c, maxValue, 0.0, "%s: expected a positive sample", metric)
+		}
+	}, timeout, retryInterval)
+}
+
+// QueryLabelsPresent asserts at least one series for testName carries every
+// given label with a non-empty value. Passing multiple labels requires them on
+// the same series. It checks cAdvisor attaches container labels.
+func (m *Mimir) QueryLabelsPresent(t *testing.T, testName string, labelNames ...string) {
```
thampiotr, line 198: Nit: That's a pretty narrow solution. I was thinking of something we can use more broadly like `QueryMetricWithLabels(t, metricName string, labelNames ...string)` that would check the labels exist for the given metric name. 

Leaving this as optional nit, because that can be done as a refactor as we add other tests.

thampiotr (approved):

PR #7040, opened 2026-09-04
fix(loki.source.docker): Preserve positions for stopped containers

### LINKED ISSUE #7036: loki.source.docker: stopping a container deletes its read position, causing a full log re-read on next start
### Component(s)

loki.source.docker

### What's wrong?

Stopping a container deletes its read position, so the next time that container starts,
Alloy re-reads its log from the beginning and re-sends the whole retained history to Loki.
The container does not need to be removed — an ordinary docker stop is enough.

When a container leaves discovery while the component is still running, tailer.stop()
removes its cursor:

// internal/component/loki/source/docker/tailer.go:176-179 (v1.16.1)
// If the component is not stopping, then it means that the target for this component is
// gone and that we should clear the entry from the positions file.
if !t.componentStopping() {
t.positions.Remove(positions.CursorKey(t.containerID), t.labelsStr)
}

componentStopping is func() bool { return c.exited.Load() } (docker.go:241) — it reports
whether Alloy's component is shutting down, and nothing else. It cannot distinguish "the
container was removed", where discarding the cursor is right, from "the container was stopped
and is coming back", where it is wrong. discovery.docker filtered on status=running drops a
stopped container from its targets, so the second case is reached by a plain docker stop.

The next start is silent about it. The lookup miss returns zero with no error:

// internal/component/loki/source/internal/positions/positions.go:241-249
pos, ok := p.positions[Entry{path, labels}]
if !ok {
return 0, nil
}

so since is 0 and the tailer asks Docker for the entire log (tailer.go:145,
Since: strconv.FormatInt(t.since.Load(), 10)). Everything still held locally is re-sent.
Entries older than Loki's reject_old_samples_max_age are refused; everything inside that
window is ingested a second time.

This is worst on any host that restarts its containers — a nightly reboot, a Docker daemon
restart, a compose-wide stop. Containers stop before Alloy does, and for as long as Alloy
outlives them it watches each one vanish and purges its cursor. Alloy then exits with a
positions file missing exactly those containers. Because it depends on shutdown order, the
affected set differs on every restart, which makes the symptom look intermittent and
container-specific when it is neither.

On one host with 40 containers and a scheduled nightly restart, 17 tailers stopped in a
1.8-second window while Alloy kept running for a further 134 seconds. The following morning
36,132 entries were rejected in a single bucket, then zero for the rest of the day. It recurs
every morning.

EXPECTED: a stopped container's read position survives, so restarting it resumes where the
tailer left off. Only a container that is genuinely gone should have its cursor discarded.

SUGGESTED FIX: the distinction is available from Docker. On target loss, ContainerInspect
the id — IsErrNotFound means removed, so discard the cursor; any other result (exited,
paused, created) means it still exists, so keep it. Alternatively drop the removal from
stop() entirely and let positions cleanup handle it, noting cursor keys are already exempt
from cleanup() by design (positions.go:303-322). At minimum, log the purge and the
subsequent since=0 full read at warn level — both are silent today, and the only external
signal is Loki rejecting weeks-old entries.

POSSIBLY THE SAME ROOT CAUSE AS https://github.com/grafana/alloy/issues/4321: that issue reports a CPU and disk-read spike for a few
minutes after every Alloy restart, with a positions file that "looks the same before and
after". That is consistent with this mechanism — the entries are removed and then rewritten by
the re-read itself, so a file compared after the fact shows no difference. Worth checking
whether the reporter's containers stop before Alloy does. https://github.com/grafana/alloy/issues/5135 asks for the same property for
loki.source.file, triggered by label churn rather than container stops.

### Steps to reproduce

No restart or multi-container setup needed. One container is enough.

With a running Alloy using discovery.docker + loki.source.docker (config below), start a
container Alloy will discover:

docker run -d --name probe alpine sh -c 'while true; do echo line; sleep 2; done'

Wait ~20s for the 5s discovery refresh and the 10s positions sync, then count the cursors:

docker exec alloy grep -c 'path: cursor-'
/var/lib/alloy/loki.source.docker.containers/positions.yml

The count has gone up by one; the entry for probe is present.

STOP the container. Do not remove it.

docker stop probe

Wait ~25s and count again. The entry is gone, although docker ps -a still lists the
container as Exited (137).

docker start probe and watch Alloy replay the container's whole log.

Measured on v1.16.1:

T0 cursor entries: 41 probe entries: 0
T1 docker run probe
T2 cursor entries: 42 probe entries: 1
T3 docker stop probe (docker ps -a: Exited (137))
T4 cursor entries: 41 probe entries: 0 <-- entry deleted

### System information

macOS 26 (Darwin 25.6.0) x86_64, Docker via OrbStack; Alloy container linux/amd64

### Software version

Grafana Alloy v1.16.1

### Configuration

```text
  host             = "unix:///var/run/docker.sock"
  refresh_interval = "5s"

  filter {
    name   = "status"
    values = ["running"]
  }
}

discovery.relabel "containers" {
  targets = discovery.docker.containers.targets

  rule {
    source_labels = ["__meta_docker_container_name"]
    regex         = "/(.*)"
    target_label  = "container"
  }
  rule {
    source_labels = ["__meta_docker_container_label_com_docker_compose_project"]
    target_label  = "compose_project"
  }
  rule {
    source_labels = ["__meta_docker_container_label_com_docker_compose_service"]
    target_label  = "service"
  }
  rule {
    replacement  = "mbp-lab"
    target_label = "host"
  }
  rule {
    action = "labelkeep"
    regex  = "__meta_docker_container_id|container|stream|compose_project|service|host"
  }
}

  host             = "unix:///var/run/docker.sock"
  targets          = discovery.relabel.containers.output
  forward_to       = [loki.write.default.receiver]
  refresh_interval = "5s"
}

// started with --storage.path=/var/lib/alloy, positions on a named volume
```

### Logs

```text
17 tailers stopped while the component kept running, during a host restart:

ts=2026-09-01T08:00:00.873927387Z level=info msg="finished transferring logs" component_id=loki.source.docker.containers component=tailer container=docker/0c9536ad5fcc written=324409
ts=2026-09-01T08:00:00.876043916Z level=info msg="finished transferring logs" component_id=loki.source.docker.containers component=tailer container=docker/d2d9214c640c written=691484
ts=2026-09-01T08:00:00.899424046Z level=info msg="finished transferring logs" component_id=loki.source.docker.containers component=tailer container=docker/8523b40d9335 written=23290
ts=2026-09-01T08:00:00.902208563Z level=info msg="finished transferring logs" component_id=loki.source.docker.containers component=tailer container=docker/76155f5eed20 written=35446
ts=2026-09-01T08:00:01.334729461Z level=info msg="finished transferring logs" component_id=loki.source.docker.containers component=tailer container=docker/b747c5dad490 written=36534
... 12 more, through 08:00:02.679877069Z ...

Alloy itself did not restart until 08:02:16.298Z -- 134 seconds after the first container stopped.

On the next start, the re-read is rejected by Loki:

ts=2026-09-01T08:02:39.888361898Z level=error msg="final error sending batch, no retries left, dropping data" component_id=loki.write.default component=endpoint host=loki:3100 status=400 error="server returned HTTP status 400 Bad Request (400): 3 errors like: entry for stream '{compose_project=\"nut-client\", container=\"nut-exporter\", host=\"mbp-lab\", service=\"nut-exporter\"}' has timestamp too old: 2026-08-11T14:15:57Z, oldest acceptable timestamp is: 2026-08-25T08:02:39Z; 12 errors like: entry for stream '{compose_project=\"portainer\", container=\"portainer\", ...}' has timestamp too old: 2026-08-21T17:02:55Z, oldest acceptable timestamp is: 2026-08-25T08:02:39Z"

sum by (instance) (increase(loki_write_dropped_entries_total[10m])) on that host:
  04:10 bucket   36132
  every other bucket that day   0
```

### AI disclosure

### Tip

al1re3a:

Preserve Docker log positions when a container stops but still exists, so restarting the container resumes from the stored cursor instead of replaying retained logs.

### Pull Request Details

`loki.source.docker` now inspects a target after its tailer stops. It removes the stored position only when Docker reports that the container is not found; stopped containers and transient inspection failures keep their positions. Tests cover stopped, removed, inspection-error, and component-shutdown paths.

Fixes #7036

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/loki/source/docker/tailer.go
+++ b/internal/component/loki/source/docker/tailer.go
@@ -177,11 +177,27 @@ func (t *tailer) stop() {
 		t.wg.Wait()
 		t.logger.Debug("stopped Docker target", "container", t.containerID)
 
-		// If the component is not stopping, then it means that the target for this component is gone and that
-		// we should clear the entry from the positions file.
-		if !t.componentStopping() {
-			t.positions.Remove(positions.CursorKey(t.containerID), t.labelsStr)
-		}
+		t.removePositionIfContainerDeleted()
+	}
+}
+
+func (t *tailer) removePositionIfContainerDeleted() {
+	if t.componentStopping() || t.client == nil {
+		return
+	}
+
+	_, err := t.client.ContainerInspect(context.Background(), t.containerID, client.ContainerInspectOptions{})
+	switch {
+	case err == nil:
+		// The target may have disappeared because the container stopped. Keep its
+		// position so a later restart resumes where the tailer left off.
+		return
+	case cerrdefs.IsNotFound(err):
+		t.positions.Remove(positions.CursorKey(t.containerID), t.labelsStr)
+	default:
+		// Preserve the position when the container's state cannot be determined.
+		// Re-reading logs is more harmful than retaining a stale cursor.
+		t.logger.Warn("could not determine whether to remove Docker container position", "id", t.containerID, "error", err)
 	}
 }
 
--- a/internal/component/loki/source/docker/tailer_test.go
+++ b/internal/component/loki/source/docker/tailer_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


kalleep:
This would break some behavior we have right now, if a target is removed then we would keep the position with this pr.

PR #7043, opened 2026-09-05
fix(positions): use deterministic temp file to prevent dentry slab growth

### LINKED ISSUE #6938: positions file write uses a uniquely-named temp file every 10s (renameio), causing unbounded kernel dentry slab growth under cgroup v2
## Summary

Every component that uses the shared `positions` package (`loki.source.kubernetes`, `loki.source.file`, `loki.source.docker`, `loki.source.podlogs`, …) rewrites its positions file **every 10 seconds** through `renameio.WriteFile`. `renameio` always suffixes its temp file with a **fresh random number**, so each write creates a filename that has never been seen before.

Under **cgroup v2 the kernel memory for those dentries is charged to the container** and counts toward `memory.max`. The result is a linear, unbounded growth of `slab_reclaimable` while the Go heap and RSS stay completely flat — so it is invisible to `kubectl top`, to RSS-based dashboards, and to Alloy's own `go_memstats_*` metrics.

In our production EKS cluster this reached **90% of the container memory limit in 52 days** and fired our OOM-risk alert.

## Evidence

Alloy **v1.17.0**, DaemonSet, EKS / Amazon Linux 2023, kernel `6.12.x`, cgroup v2, `loki.source.kubernetes`.

**The Go heap is identical between a growing pod and a flat pod — only kernel memory differs:**

| | `memory.current` | `anon` | `slab_reclaimable` | Go `heap_inuse` |
|---|---|---|---|---|
| pod A (growing) | 233 MiB / 256 | 59.5 MiB | **171.2 MiB** | 41.9 MiB |
| pod B (flat) | 89 MiB / 256 | 56.9 MiB | 7.8 MiB | 42.6 MiB |

Goroutines 214 vs 171, open fds 13 vs 12. No application-level leak.

**Growth is perfectly linear over 30 days at ~+3.3 MiB/day**, with no plateau and no correlation to log throughput — in fact the *flat* pod shipped 35x more log entries (3.67 GiB / 17.7 entries-per-second vs 0.77 GiB / 0.5 eps).

**`/proc/slabinfo` identifies the cache as `dentry`:**

| cache | growing node | flat node |
|---|---|---|
| `dentry` | **594.3 MiB** (3,245,907 objs) | 30.1 MiB (164,640 objs) |
| `kmalloc-rcl-64` | 121.3 MiB (**1,986,944** objs) | not in top 8 |

`/proc/sys/fs/dentry-state` → `nr_dentry` 3,245,300 vs `inode-nr` 300,852 = **10.8:1 ratio, 99.1% unused** → negative-dentry dominated.

**The temp files, captured by polling the positions directory — 10 s apart, every name unique:**

```
07:10:56  .positions.yml4038530999602394749
07:11:06  .positions.yml4215553290718380405
07:11:16  .positions.yml8956570913662244016
07:11:26  .positions.yml9016088893600132773
```

Each name is **33 characters**, which exceeds `DNAME_INLINE_LEN` (32), so every dentry also needs an external name buffer — that is the `kmalloc-rcl-64` population above (~192 B dentry + 64 B name ≈ 256 B per entry).

**The arithmetic closes:** 8,640 writes/day × ~256 B ≈ **2.2 MiB/day** steady-state; we measure 3.3 MiB/day on average, the remainder coming from short bursts during target churn.

## Root cause

`internal/component/loki/source/internal/positions/write_positions_unix.go`:

```go
// This code is copied from Promtail. ...
import renameio "github.com/google/renameio/v2"

func writePositionFile(filename string, positions map[Entry]string) error {
    ...
    return renameio.WriteFile(target, buf, os.FileMode(positionFileMode))
}
```

`google/renameio` `tempfile.go`:

```go
// nextrandom is a function generating a random number.
var nextrandom = rand.Int64
// The name is always suffixed with a random number.
name := prefix + strconv.FormatInt(nextrandom(), 10)
```

`internal/component/loki/source/internal/positions/positions.go`:

```go
f.DurationVar(&cfg.SyncPeriod, prefix+"positions.sync-period", 10*time.Second, "Period with this to sync the position file.")
...
ticker := time.NewTicker(p.cfg.SyncPeriod)
```

The file comment says the code was copied from Promtail — but [Promtail's `positions.go`](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/positions/positions.go) writes to a **fixed** name (`<target>-new`) and then renames it. A fixed temp name reuses the same dentry forever and does not grow. **Substituting `renameio` is what introduced this.**

## Why users cannot work around it in config

`SyncPeriod` exists in the Go struct (`mapstructure:"sync_period"`), but [`loki.source.kubernetes`](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.kubernetes/) exposes only `forward_to`, `targets` and the `client` / `authorization` / `basic_auth` / `oauth2` / `tls_config` / `clustering` blocks. There is no way to lengthen the 10 s period from Alloy configuration.

The only mitigations available today are raising the memory limit (buys time proportionally) or restarting the container periodically (slab is per-cgroup and resets on restart).

## Suggested fix

Use a deterministic temp filename so the dentry is reused, e.g. restore Promtail's `<target>-new` pattern, or keep a single reusable temp file per positions file. Optionally also expose `sync_period` on the components.

## Note on reproducibility

The growth is only *observable* on nodes where the kernel is not already reclaiming dentries. On our memory-pressured nodes (JVM workloads with large page cache) shrinkers run often enough to keep slab near zero, which is why some pods in the same DaemonSet appear flat. The allocation churn happens everywhere; whether it accumulates depends on node-level pressure. This also means it can go unnoticed for a long time and then surface only on the least-pressured nodes.

## Possibly related

- #6100 — "Alloy self reported RSS memory does not match cAdvisor rss memory". This mechanism would explain that class of discrepancy: `gopsutil` RSS and cAdvisor `container_memory_rss` do not include the kernel slab that cgroup v2 charges to the container, while `container_memory_working_set_bytes` / `memory.current` do.
- #4421 — same symptom reported, closed as `frozen-due-to-age` with no root cause identified.

## Useful diagnostic

Kernel memory can be recovered from existing cAdvisor metrics without a new exporter:

```promql
container_memory_usage_bytes - container_memory_rss - container_memory_cache
```

This matched `memory.stat`'s `kernel` field to within 0.2% in our measurements. (Note: using `working_set` instead of `usage` yields `kernel - inactive_file` and can go negative.)

AruneshDwivedi:
## Summary

Replaces renameio.WriteFile with a custom atomicWriteFile that uses a deterministic temp path (<target>.tmp) instead of a unique random name on every call. This prevents unbounded kernel dentry slab growth under cgroup v2 when positions are written every 10 seconds.

## Changes

- : Replaced renameio.WriteFile with atomicWriteFile using deterministic temp path
- : Added TestWritePositionFile_ReusesDentry test

Fixes #6938

### DIFF
--- a/internal/component/loki/source/internal/positions/positions_test.go
+++ b/internal/component/loki/source/internal/positions/positions_test.go
@@ -423,6 +423,32 @@ positions:
 	}, out)
 }
 
+// TestWritePositionFile_ReusesDentry verifies that writePositionFile uses a
+// deterministic temp path (<target>.tmp) rather than a unique random name on
+// every call, so the kernel dentry slab does not grow unboundedly under
+// cgroup v2.
+func TestWritePositionFile_ReusesDentry(t *testing.T) {
+	tmpDir := t.TempDir()
+	target := filepath.Join(tmpDir, "positions.yaml")
+
+	// Write multiple times and verify the temp file path stays the same.
+	for i := 0; i < 3; i++ {
+		err := writePositionFile(target, map[Entry]string{
+			{Path: "/tmp/test.log", Labels: ""}: "100",
+		})
+		require.NoError(t, err)
+
+		// The temp file should either not exist (already renamed) or be the
+		// deterministic <target>.tmp path.
+		entries, readErr := os.ReadDir(tmpDir)
+		require.NoError(t, readErr)
+		for _, entry := range entries {
+			require.NotContains(t, entry.Name(), "positions.yaml.",
+				"unexpected uniquely-named temp file (renameio-style)")
+		}
+	}
+}
+
 func TestReadEmptyLabels(t *testing.T) {
 	temp := tempFilename(t)
 	defer func() {
--- a/internal/component/loki/source/internal/positions/write_positions_unix.go
+++ b/internal/component/loki/source/internal/positions/write_positions_unix.go
@@ -10,11 +10,50 @@ import (
 	"os"
 	"path/filepath"
 
-	renameio "github.com/google/renameio/v2"
 	yaml "gopkg.in/yaml.v2"
 )
 
 const positionFileMode = 0600
+
+// atomicWriteFile writes buf to filename using a deterministic temp file
+// (<target>.tmp), fsyncs, renames over the target, then fsyncs the directory.
+// This reuses the same dentry on every write instead of allocating a new one
+// (as renameio.WriteFile does), which prevents unbounded kernel dentry slab
+// growth under cgroup v2.
+func atomicWriteFile(filename string, buf []byte) error {
+	target := filepath.Clean(filename)
+	tmp := target + ".tmp"
+
+	f, err := os.OpenFile(tmp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(positionFileMode))
+	if err != nil {
+		return err
+	}
+
+	if _, err := f.Write(buf); err != nil {
+		f.Close()
+		return err
+	}
+
+	if err := f.Sync(); err != nil {
+		f.Close()
+		return err
+	}
+
+	if err := f.Close(); err != nil {
+		return err
+	}
+
+	if err := os.Rename(tmp, target); err != nil {
+		return err
+	}
+
+	dir, err := os.Open(filepath.Dir(target))
+	if err != nil {
+		return err
+	}
+	defer dir.Close()
+	return dir.Sync()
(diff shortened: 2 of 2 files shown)


kalleep:
Hey I believe we can use https://github.com/natefinch/atomic instead of having our own implementation. We already have this as a dependency in alloy

PR #7047, opened 2026-09-07
fix: Ensure Windows exporter uses correct text file defaults when not specified

blewis12:
### Pull Request Details
We're currently falling back to incorrect text file defaults in the windows prometheus exporter. This PR ensures that we're correctly following the defaults we describe in our [docs](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.windows/#textfile). This has been broken since Alloy v1.11.0.

This is a breaking change as the defaults will be updated for users, but since it's a fix to match our docs I think that is acceptable 

### Note to reviewer
Also windows tests [are failing](https://github.com/grafana/alloy/actions/runs/34120368055/job/102032357523?pr=7047) for reasons unrelated to this PR, but I ran then just to check that the test added in this PR passes (it does). Happy to investigate those other windows failures outside of this PR

BREAKING CHANGE: `prometheus.exporter.windows` now applies the documented default `textfile` directory (`<install dir>\textfile_inputs`) when neither `textfile.directories` nor the deprecated `text_file.text_file_directory` is set, restoring behavior that was lost in v1.11.0. Deployments that were relying on the collector silently returning nothing will start receiving any .prom files found there, or an error if that directory doesn't exist, either of which can be avoided by setting directories explicitly or removing textfile from enabled_collectors.

### PR Checklist
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/static/integrations/windows_exporter/config_windows.go
+++ b/internal/static/integrations/windows_exporter/config_windows.go
@@ -80,7 +80,9 @@ func (c *Config) ToWindowsExporterConfig() (collector.Config, error) {
 	cfg.SMTP.ServerExclude, err = regexp.Compile(coalesceString(c.SMTP.Exclude, c.SMTP.BlackList))
 	errs = append(errs, err)
 
-	cfg.Textfile.TextFileDirectories = strings.Split(c.TextFile.TextFileDirectory, ",")
+	if c.TextFile.TextFileDirectory != "" {
+		cfg.Textfile.TextFileDirectories = strings.Split(c.TextFile.TextFileDirectory, ",")
+	}
 
 	cfg.PhysicalDisk.DiskInclude, err = regexp.Compile(c.PhysicalDisk.Include)
 	errs = append(errs, err)
--- a/internal/static/integrations/windows_exporter/config_windows_test.go
+++ b/internal/static/integrations/windows_exporter/config_windows_test.go
@@ -1,0 +1,26 @@
+package windows_exporter
+
+import (
+	"testing"
+
+	"github.com/prometheus-community/windows_exporter/pkg/collector"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfig_TextFileDefaultsWhenUnset(t *testing.T) {
+	c := &Config{}
+
+	cfg, err := c.ToWindowsExporterConfig()
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


internal/static/integrations/windows_exporter/config_windows.go
```diff
@@ -80,7 +80,9 @@ func (c *Config) ToWindowsExporterConfig() (collector.Config, error) {
 	cfg.SMTP.ServerExclude, err = regexp.Compile(coalesceString(c.SMTP.Exclude, c.SMTP.BlackList))
 	errs = append(errs, err)
 
-	cfg.Textfile.TextFileDirectories = strings.Split(c.TextFile.TextFileDirectory, ",")
+	if c.TextFile.TextFileDirectory != "" {
+		cfg.Textfile.TextFileDirectories = strings.Split(c.TextFile.TextFileDirectory, ",")
+	}
```
blewis12, line 85: Explanation: If a user doesn't specify any `text_file` or `textfile` value - in [SetToDefaults](https://github.com/grafana/alloy/blob/bff3d36eb18e62891ddf25a51b66d7f2dde2869c/internal/component/prometheus/exporter/windows/config_windows.go#L84-L85) we leave both the Alloy input fields (`TextFile` and `TextFileDeprecated`) as [nil](https://github.com/grafana/alloy/blob/bff3d36eb18e62891ddf25a51b66d7f2dde2869c/internal/component/prometheus/exporter/windows/config_windows.go#L84-L85), when we then convert to `windows_integration.TextFileConfig` we return an [empty struct](https://github.com/grafana/alloy/blob/bff3d36eb18e62891ddf25a51b66d7f2dde2869c/internal/component/prometheus/exporter/windows/config.go#L100-L102), and when we then convert to the underlying prometheus exporter config we first set the [correct defaults](https://github.com/grafana/alloy/blob/bff3d36eb18e62891ddf25a51b66d7f2dde2869c/internal/static/integrations/windows_exporter/config_windows.go#L52) and then override them with an [invalid value](https://github.com/grafana/alloy/blob/bff3d36eb18e62891ddf25a51b66d7f2dde2869c/internal/static/integrations/windows_exporter/config_windows.go#L83)

The fix here means that we _only_ override the defaults if the user actually specified input values

thampiotr (approved):

PR #7050, opened 2026-09-08
chore(loki): Move created timestamp to stream instead of batch.

kalleep:
### Pull Request Details

From https://github.com/grafana/alloy/pull/6913#discussion_r3880197335. When passing entries through `loki.process` we need to rebuild the batch. Here we can get entries from other streams/batches we want to store the oldest created time we can find. To make this a bit more accurate I changed it so that we store created in every stream instead for every batch.  

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/batch.go
+++ b/internal/component/common/loki/batch.go
@@ -9,8 +9,6 @@ import (
 )
 
 type Batch struct {
-	// created is a unix timestamp in micro seconds.
-	created  int64
 	entryLen int
 	streams  []Stream
 }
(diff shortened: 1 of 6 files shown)


ptodev (approved):
Thanks, makes sense.

I wonder if there could be weird situations with OOO entries where a batch with old entry is processed later than a batch with more recent entry for the same stream. I don't expect there to be an issues, but it's worth mentioning. It's the only edge case I can think of that could be affected.

kalleep:
> Thanks, makes sense.
> 
> I wonder if there could be weird situations with OOO entries where a batch with old entry is processed later than a batch with more recent entry for the same stream. I don't expect there to be an issues, but it's worth mentioning. It's the only edge case I can think of that could be affected.


Yeah so this is important and we want to try to avoid it as much as possible. But nothing stops someone from relabeling streams into the same stream (both in the old and new Pipeline) and then we can't really guard against OOO.

Issue #2906, opened 2025-03-04
Debug compression ratio in loki.write


dehaansa:
### Request

Loki client records compressed log bytes in the `loki_write_encoded_bytes_total` metric, we should also track the uncompressed bytes or just calculate the compression ratio.

### Use case

Users have concerns about egress & ingress costs, this can help understand how batch configuration contributes to network costs.

thampiotr:
That sounds like a good idea. We would be open to contributions.

PR #7052, opened 2026-09-08
feat(loki.write): Add loki_write_batch_size_bytes histogram

grobinson-grafana:

`loki_write_request_size_bytes` measures the batch after proto encoding and snappy compression, so it can't be compared against the configured batch_size. Add `loki_write_batch_size_bytes`, which observes the uncompressed sum of entry sizes in a batch — the same quantity the batch compares against batch_size when deciding whether another entry fits.

The histogram is observed once per batch in sendBatch, before encoding, and is labelled by host and tenant with the same buckets as `loki_write_request_size_bytes`.

### Pull Request Details

This is important so Alloy users can see their batch size, for which there are limits in Loki. The existing metric `loki_write_request_size_bytes` counts the number of bytes after Snappy compression, and so unfortunately is not adequate for this purpose.

https://github.com/grafana/alloy/issues/2906

### Notes to the Reviewer

Created with Claude Opus 5, tested locally on my Mac.

### PR Checklist

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/shards.go
+++ b/internal/component/common/loki/client/shards.go
@@ -403,6 +403,10 @@ func (s *shards) sendBatch(tenantID string, batch *batch, protoBuf, snappyBuf *[
 	obs := s.metrics.entryLatency.WithLabelValues(s.cfg.URL.Host, tenantID)
 	defer batch.reportAsSentData(s.tracker, obs)
 
+	// batch.size is the uncompressed size of the log lines, which is what's
+	// compared against the configured batch_size when filling the batch.
+	s.metrics.batchSize.WithLabelValues(s.cfg.URL.Host, tenantID).Observe(float64(batch.size))
+
 	r, entriesCount := batch.request()
 
 	size := r.Size()
--- a/docs/sources/reference/components/loki/loki.write.md
+++ b/docs/sources/reference/components/loki/loki.write.md
@@ -202,6 +202,7 @@ The following fields are exported and can be referenced by other components:
 * `loki_write_dropped_entries_total` (counter): Number of log entries dropped because they failed to be sent to the ingester after all retries.
 * `loki_write_sent_bytes_total` (counter): Number of bytes sent.
 * `loki_write_sent_entries_total` (counter): Number of log entries sent to the ingester.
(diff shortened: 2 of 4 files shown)


internal/component/common/loki/client/shards.go
```diff
@@ -403,6 +403,10 @@ func (s *shards) sendBatch(tenantID string, batch *batch, protoBuf, snappyBuf *[
 	obs := s.metrics.entryLatency.WithLabelValues(s.cfg.URL.Host, tenantID)
 	defer batch.reportAsSentData(s.tracker, obs)
 
+	// batch.size is the uncompressed size of the log lines, which is what's
+	// compared against the configured batch_size when filling the batch.
+	s.metrics.batchSize.WithLabelValues(s.cfg.URL.Host, tenantID).Observe(float64(batch.size))
```
kalleep, line 408: The size we use in batch.size is not fully accurate. When we add entries to a batch we use [this method](https://github.com/grafana/alloy/blob/0d731b0e22c620202c6a55874981176180908762/internal/component/common/loki/entry.go#L61-L70). 

This only accounts for some of the bytes for each entry. I does not count labels, timestamp and protobuf padding. I experimented a while back with actually computing the correct size but it was quite expensive to do for each entry.

I guess it's fine to use here but the actual byte size will be a bit bigger once it is encoded as protobuf

grobinson-grafana: What if instead we emit the size of `protoBuf`, would that be better?

grobinson-grafana: I suppose one disadvantage is that this metric could then exceed the configured `batch_size` value in the Alloy configuration, and that would be confusing to users?

kalleep: >What if instead we emit the size of protoBuf, would that be better?

Yes I was leaning towards this initially but I don't think that would help users here. IMO let start with that you added. At some point I want to get back to fix the sizing when adding to a batch and then that would also fix the metric.

kalleep (approved):

clayton-cornell (approved):

PR #7057, opened 2026-09-08
feat(helm): Add service.trafficDistribution to values.yaml

### LINKED ISSUE #7017: helm: support spec.trafficDistribution on the Alloy Service
### What's missing

The chart's `templates/service.yaml` renders `internalTrafficPolicy` from
`.Values.service.internalTrafficPolicy`, but there is no way to set
`spec.trafficDistribution` on the Service.

For a DaemonSet-mode Alloy acting as a node-local OTLP receiver, that leaves two
unsatisfactory options:

- `internalTrafficPolicy: Cluster` (the default) — every export is load-balanced across
  all DaemonSet pods, so telemetry usually leaves the node that produced it.
- `internalTrafficPolicy: Local` — node-local, but with no fallback. While a node's
  Alloy pod is unready — notably during any DaemonSet rollout — that node's telemetry is
  dropped rather than routed elsewhere.

### What would fix it

`spec.trafficDistribution: PreferSameNode` prefers an endpoint on the caller's own node
and falls back when there isn't one, which is exactly the semantics a node-local
collector wants.

It is GA as of Kubernetes 1.35 (alpha in 1.33, beta and enabled by default in 1.34). See
KEP-3015, https://github.com/kubernetes/enhancements/issues/3015.

Suggested shape: a `service.trafficDistribution` value, rendered only when non-empty and
guarded on the cluster version the same way `internalTrafficPolicy` already is:

```yaml
service:
  # -- Traffic distribution preference. e.g. "PreferSameNode" (k8s >= 1.34).
  trafficDistribution: ""
```

```gotemplate
{{- with .Values.service.trafficDistribution }}
  trafficDistribution: {{ . }}
{{- end }}
```

### Workaround

Declare a second Service via `extraObjects` selecting the same pods, and point exporters
at that name instead of the chart's Service.

### Context

A nine-service fleet where a DaemonSet Alloy receives OTLP from application-level
OpenTelemetry SDKs. We want node-local delivery without introducing a per-rollout
span-loss window.

agermel:
Adds a `service.trafficDistribution` value to the Alloy Helm chart. When set, it renders `spec.trafficDistribution` on the Service, guarded to Kubernetes 1.34+.
 
Fixes #7017
 
### Notes to the Reviewer
This is my first contribution to this project, and I’m really excited to be involved. I’ve enjoyed working on this change and hope to continue contributing to the project in the future.
 
Thank you for taking the time to review my PR!
 
### PR Checklist
* [x]  Documentation added
* [ ]  Tests updated
* [ ]  Config converters updated
* [x]  This pull request was substantially generated with AI assistance (see the [GenAI policy](https://github.com/grafana/alloy/blob/main/docs/developer/genai.md))

### DIFF
--- a/operations/helm/charts/alloy/CHANGELOG.md
+++ b/operations/helm/charts/alloy/CHANGELOG.md
@@ -13,6 +13,7 @@ Unreleased
 ### Enhancements
 
 - Add `controller.dnsConfig` to configure the pod's DNS settings (`nameservers`, `searches`, `options`). (@younsl)
+- Allow setting `trafficDistribution` in the service spec. (@agermel)
 
 1.12.1 (2026-08-26)
 ----------
--- a/operations/helm/charts/alloy/README.md
+++ b/operations/helm/charts/alloy/README.md
@@ -188,6 +188,7 @@ useful if just using the default DaemonSet isn't sufficient.
 | service.externalTrafficPolicy | string | `"Cluster"` | Value for external traffic policy. 'Cluster' or 'Local' |
 | service.internalTrafficPolicy | string | `"Cluster"` | Value for internal traffic policy. 'Cluster' or 'Local' |
 | service.nodePort | int | `31128` | NodePort port. Only takes effect when `service.type: NodePort` |
+| service.trafficDistribution | string | `""` | Value for traffic distribution. 'PreferSameNode' or 'PreferSameZone' (k8s >= 1.34) |
 | service.type | string | `"ClusterIP"` | Service type |
 | serviceAccount.additionalLabels | object | `{}` | Additional labels to add to the created service account. |
 | serviceAccount.annotations | object | `{}` | Annotations to add to the created service account. |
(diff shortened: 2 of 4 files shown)


dehaansa:
Thanks for looking to contribute! We'll want a test added for this in the helm tests, and it appears the key that you've signed your commit with isn't recognized by github. You'll either need to update github or use a different key that it recognizes!

agermel:
> Thanks for looking to contribute! We'll want a test added for this in the helm tests, and it appears the key that you've signed your commit with isn't recognized by github. You'll either need to update github or use a different key that it recognizes!

Thank you for your review! I’ve added the test for this contribution, and the commit is now verified on GitHub.

dehaansa:
There's a tiny conflict in the changelog (we'll be using release please for the helm changelog soon and this won't be a tedious thing to solve in the future 🙃 ) that needs resolving.

agermel:
Thanks! and fixed🫡

dehaansa (approved):

dehaansa:
Thanks again for the contribution, and helping drive this one that's had a couple lost attempts in the past!

PR #7064, opened 2026-09-09
fix: Debug log the reason an exporter metrics handler fails instead of discarding it

thampiotr:

When a metrics handler fails, `promhttp` puts the reason in the HTTP response body. The Prometheus scrape client discards that body and reports only `server returned HTTP status 500 Internal Server Error`. The cause never reaches the logs, so a broken exporter shows `up=0` with no diagnostics.

This PR adds `util.PromHTTPErrorLogger`, an adapter from `*slog.Logger` to `promhttp.Logger` that logs at debug level, and passes it as `promhttp.HandlerOpts.ErrorLog` at every metrics handler Alloy owns.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


kgeckhart (commented):
Nice find! Is this,

> logs at debug level

a default or is it something we can control. Having a log at debug is better than nothing but if we can log it at error it might be more ideal.

thampiotr:
> Nice find! Is this,
> 
> > logs at debug level
> 
> a default or is it something we can control. Having a log at debug is better than nothing but if we can log it at error it might be more ideal.

It's easy to make it any level we want, but I wanted to make it off-by-default because this could spam your logs a lot in some cases. Like you have an exporter that connects over the network to something and that something is not available due to restarts/rollouts, etc. It could be a warning maybe, not an error. But there's a reason why every scrape error is not logged in Prometheus. I thought it would be safer to keep it on debug as a troubleshooting tool. 

What do you think? I'm slightly leaning towards the safe route of debug level.

internal/util/promhttp_errorlog.go
kgeckhart, line 1: Could we expand this to pull in some of the `promhttp.HandlerFor` ceremony that's repeated? This way we can put a test on it cleanly. There's some bits in the implementation about log formatting and it would be nice to see how it looks.

kgeckhart (commented):
> What do you think? I'm slightly leaning towards the safe route of debug level.

I feel like it's a damned if you do, damned if you don't conundrum  😅. Debug logging is better than silently dropping it though.

internal/util/promhttp_errorlog.go
thampiotr: That actually helped reduce code, good call. 
I also added a test that shows the multiline handling

internal/util/promhttp_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.Empty(t, buf.String())
+}
+
+// recordingLogger is a promhttp.Logger a caller may already have.
+type recordingLogger struct{ lines []string }
+
+func (l *recordingLogger) Println(v ...any) { l.lines = append(l.lines, fmt.Sprintln(v...)) }
+
+// TestPromHTTPHandlerFor_KeepsCallerErrorLog makes sure the helper fills in a
+// default rather than overwriting a destination the caller chose.
```
kgeckhart: Is the comment on this one backwards? I think it's testing that when the logger is provided we don't override it with the default.

kgeckhart (dismissed):
Just a small test comment comment. I'll leave it up to you on if you want to fix it if my comment is accurate.

blewis12 (approved):

PR #7068, opened 2026-09-09
feat(database_observability.postgres): Add index-usage collectors

gaantunes:

Split out of #7003 (originally combined postgres + mysql; splitting per one-logical-change-per-PR).

Adds two opt-in collectors to `database_observability.postgres`, disabled by default and enabled via `enable_collectors`, emitting only the metrics needed for missing-index and unused-index KG insights:

- `table_stats` — `pg_stat_user_tables_seq_scan`/`idx_scan`/`n_live_tup`.
- `index_stats` — `pg_stat_user_indexes_idx_scan_total`, `pg_index_properties{is_primary}`, `pg_index_size_bytes`.

Both fan out across every database the connection can reach, reusing the discovery/connection mechanism already used by `schema_details` (extracted into a shared `multi_database.go`). This fixes the vendored `postgres_exporter`'s single-DSN limitation for these system views natively, rather than waiting on the unmerged upstream `prometheus-community/postgres_exporter#1378`.

Metric names/labels match the existing `pg_stat_user_tables_*` fields, and, for the new index metrics, the shape proposed by the unmerged upstream `prometheus-community/postgres_exporter#1071`.

Also fixes a redundant-connection issue in the multi-database fan-out (`connectToDatabase`): it always opened a fresh connection via `factory()`, even for the one database that's already the database the initial connection points to, since a freshly-`sql.Open`'d `*sql.DB` is never pointer-equal to the initial one. It now detects that case up front and reuses the initial connection directly instead.

Validated against a live `db-o11y-postgres-16` test instance hosting real `books_store` data: every emitted series across both collectors matched the database source exactly.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated


clayton-cornell:
All looks good for doc changes :-)

internal/component/database_observability/postgres/collector/multi_database.go
```diff
@@ -0,0 +1,67 @@
...
+	"database/sql"
+	"fmt"
+)
+
+// discoverDatabases lists databases the current connection can reach, via
+// pg_database (readable from any single connection) -- used to fan out
+// per-database connections, since most stat views only report on the
+// database a connection is actually established to.
+func discoverDatabases(ctx context.Context, conn *sql.DB, excludeDatabases []string) ([]string, error) {
+	query := fmt.Sprintf(selectAllDatabases, buildExcludedDatabasesClause(excludeDatabases))
```
cristiangreco: Couple of thoughts:
- `selectAllDatabases` should live close to this function now
- maybe this code could be moved into `dsn.go`, I feel dsn parsing and database listing almost always go hand in hand

internal/component/database_observability/postgres/collector/index_stats.go
```diff
@@ -0,0 +1,173 @@
...
+		pg_relation_size(s.indexrelid) AS index_size_bytes
+	FROM pg_stat_user_indexes s
+	JOIN pg_index i ON i.indexrelid = s.indexrelid`
+
+var indexLabels = []string{labelDatname, "schemaname", "relname", "indexrelname"}
+
+var (
+	// Field names match the proposed (unmerged) upstream pg_stat_user_indexes
+	// collector; the database_observability namespace keeps them from
+	// colliding with that collector's own names if it ever ships.
```
cristiangreco: Comment might be unnecessary

internal/component/database_observability/postgres/collector/index_stats.go
```diff
@@ -0,0 +1,173 @@
...
+	)
+	indexPropertiesDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("database_observability", "pg", "index_properties"),
+		"Properties of an index; a constant 1 with is_primary set to whether the index backs a primary key",
+		append(append([]string{}, indexLabels...), "is_primary"), nil,
+	)
+	indexSizeBytesDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("database_observability", "pg", "index_size_bytes"),
+		"Total disk space used by this index, in bytes",
+		indexLabels, nil,
```
cristiangreco: Naming of these metrics seems a bit inconsistent? (`pg_stat_user_indexes` vs `pg`)

internal/component/database_observability/postgres/collector/index_stats.go
```diff
@@ -0,0 +1,173 @@
...
+	// colliding with that collector's own names if it ever ships.
+	indexUsageIdxScanTotalDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("database_observability", "pg_stat_user_indexes", "idx_scan_total"),
+		"Number of index scans initiated on this index",
+		indexLabels, nil,
+	)
+	indexPropertiesDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("database_observability", "pg", "index_properties"),
+		"Properties of an index; a constant 1 with is_primary set to whether the index backs a primary key",
+		append(append([]string{}, indexLabels...), "is_primary"), nil,
```
cristiangreco: Why the double append here?

cristiangreco: On second thought, do we actually need this metric at all? We can get this info from `schema_details` collector, if necessary. And this metric doesn't match any other metric we added for mysql in https://github.com/grafana/alloy/pull/7069

WDYT?

internal/component/database_observability/postgres/collector/multi_database.go
gaantunes: All applied, thanks for the direction.

internal/component/database_observability/postgres/collector/index_stats.go
gaantunes: Moving that info to `index_size_bytes` as a label, together with `is_unique` and `is_partial`. These indexes are necessary (first 2 are constraints, third maps to a subset of rows) but might not have invocations to it, so these properties are important to map, so to avoid false positives.

internal/component/database_observability/postgres/collector/index_stats.go
gaantunes: indeed, will name it more consistently with what we used in mysql.

cristiangreco (approved):

PR #7069, opened 2026-09-09
feat(database_observability.mysql): Add index-usage collectors

gaantunes:

Split out of #7003 (originally combined postgres + mysql; splitting per one-logical-change-per-PR).

Adds two opt-in collectors to `database_observability.mysql`, disabled by default and enabled via `enable_collectors`, emitting only the metrics needed for missing-index and unused-index:

- `table_stats` — `mysql_table_stats_seq_scan_total{schema,table}`, the fetch count of the `index="NONE"` ("no index used") row per table.
- `index_stats` — `mysql_index_stats_idx_scan_total{schema,table,index}`, the same counter for each named-index row, plus (MySQL >= 5.6.6) `mysql_index_stats_size_bytes{schema,table,index}` from `mysql.innodb_index_stats` (InnoDB's persistent optimizer statistics, refreshed by MySQL itself, never triggered by us).

Both read `performance_schema.table_io_waits_summary_by_index_usage` and bind the excluded-schema list as query parameters rather than formatting it into the SQL text. Each collector owns its own metric name rather than sharing one upstream-style name (`mysqld_exporter`'s `perf_schema.indexiowaits`) across two independently-toggleable collectors, which would mean one collector's data being silently incomplete under a shared name whenever only one is enabled.

**Known dependency**: `mysql_index_stats_size_bytes` requires `GRANT SELECT ON mysql.innodb_index_stats` on the monitoring role, which is not part of today's default grant set (verified against the actual `db-o11y` MySQL role's live grants). Missing the grant degrades gracefully (logged error, the metric just doesn't appear).

Validated against a live `db-o11y` MySQL test instance hosting real `books_store`/`employees` data: every emitted series across both collectors, including the size metric, matched the database source exactly.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -48,6 +48,8 @@ The following collectors are configurable:
 | `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
 | `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
 | `explain_plans`   | Collect explain plans information.                           | yes                |
+| `table_stats`     | Collect the no-index (`INDEX_NAME` is `NULL`) fetch count per table from `performance_schema.table_io_waits_summary_by_index_usage`, for missing-index detection. | no |
+| `index_stats`     | Collect the per-index fetch count from `performance_schema.table_io_waits_summary_by_index_usage`, and (MySQL 5.6.6+) per-index size from `mysql.innodb_index_stats`, for unused-index detection. | no |
 
 ## Blocks
 
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -49,7 +49,7 @@ The following collectors are configurable:
 | `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
 | `explain_plans`   | Collect explain plans information.                           | yes                |
 | `table_stats`     | Collect table-level index usage statistics.                  | no                 |
-| `index_stats`     | Collect per-index usage statistics.                           | no                 |
+| `index_stats`     | Collect per-index usage statistics.                          | no                 |
 
 ## Blocks
 
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -48,8 +48,8 @@ The following collectors are configurable:
 | `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
 | `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
 | `explain_plans`   | Collect explain plans information.                           | yes                |
-| `table_stats`     | Collect the no-index (`INDEX_NAME` is `NULL`) fetch count per table from `performance_schema.table_io_waits_summary_by_index_usage`, for missing-index detection. | no |
-| `index_stats`     | Collect the per-index fetch count from `performance_schema.table_io_waits_summary_by_index_usage`, and (MySQL 5.6.6+) per-index size from `mysql.innodb_index_stats`, for unused-index detection. | no |
+| `table_stats`     | Collect table-level index usage statistics, for missing-index detection. | no |
+| `index_stats`     | Collect per-index usage statistics, for unused-index detection. | no |
 
 ## Blocks
 
--- a/docs/sources/reference/components/database_observability/database_observability.mysql.md
+++ b/docs/sources/reference/components/database_observability/database_observability.mysql.md
@@ -48,8 +48,8 @@ The following collectors are configurable:
 | `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
 | `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
 | `explain_plans`   | Collect explain plans information.                           | yes                |
-| `table_stats`     | Collect table-level index usage statistics, for missing-index detection. | no |
-| `index_stats`     | Collect per-index usage statistics, for unused-index detection. | no |
+| `table_stats`     | Collect table-level index usage statistics.                  | no                 |
+| `index_stats`     | Collect per-index usage statistics.                           | no                 |
 
 ## Blocks
 
--- a/internal/component/database_observability/mysql/collector/exclude_schemas.go
+++ b/internal/component/database_observability/mysql/collector/exclude_schemas.go
@@ -1,6 +1,10 @@
 package collector
 
-import "github.com/grafana/alloy/internal/component/database_observability"
+import (
+	"strings"
+
+	"github.com/grafana/alloy/internal/component/database_observability"
+)
 
 var excludedSchemas = []string{"mysql", "performance_schema", "sys", "information_schema"}
 
@@ -15,4 +19,30 @@ func buildExcludedSchemasClause(schemas []string) string {
 	all = append(all, excludedSchemas...)
 	all = append(all, schemas...)
 	return database_observability.BuildExclusionClause(all)
+}
+
+// excludedSchemasArgs returns the combined default + custom excluded schemas
+// as query arguments, for callers that bind them as placeholders (`?`)
+// instead of formatting them into the SQL text.
+func excludedSchemasArgs(schemas []string) []any {
+	all := make([]string, 0, len(excludedSchemas)+len(schemas))
+	all = append(all, excludedSchemas...)
+	all = append(all, schemas...)
+
+	args := make([]any, len(all))
+	for i, s := range all {
+		args[i] = s
+	}
+	return args
+}
+
+// sqlPlaceholders returns a comma-separated list of n "?" placeholders, for
+// building a parameterized `IN (...)` / `NOT IN (...)` clause whose argument
+// count is dynamic.
+func sqlPlaceholders(n int) string {
+	placeholders := make([]string, n)
+	for i := range placeholders {
+		placeholders[i] = "?"
+	}
+	return strings.Join(placeholders, ", ")
 }
--- a/internal/component/database_observability/mysql/collector/exclude_schemas.go
+++ b/internal/component/database_observability/mysql/collector/exclude_schemas.go
@@ -22,8 +22,8 @@ func buildExcludedSchemasClause(schemas []string) string {
 }
 
 // excludedSchemasArgs returns the combined default + custom excluded schemas
-// as query arguments, for callers that bind them as placeholders (`?`)
-// instead of formatting them into the SQL text.
+// as query arguments, for binding as `?` placeholders instead of formatting
+// them into the SQL text.
 func excludedSchemasArgs(schemas []string) []any {
 	all := make([]string, 0, len(excludedSchemas)+len(schemas))
 	all = append(all, excludedSchemas...)
--- a/internal/component/database_observability/mysql/collector/exclude_schemas.go
+++ b/internal/component/database_observability/mysql/collector/exclude_schemas.go
@@ -25,13 +25,12 @@ func buildExcludedSchemasClause(schemas []string) string {
 // as query arguments, for binding as `?` placeholders instead of formatting
 // them into the SQL text.
 func excludedSchemasArgs(schemas []string) []any {
-	all := make([]string, 0, len(excludedSchemas)+len(schemas))
-	all = append(all, excludedSchemas...)
-	all = append(all, schemas...)
-
-	args := make([]any, len(all))
-	for i, s := range all {
-		args[i] = s
+	args := make([]any, 0, len(excludedSchemas)+len(schemas))
+	for _, s := range excludedSchemas {
+		args = append(args, s)
+	}
+	for _, s := range schemas {
+		args = append(args, s)
 	}
 	return args
 }
--- a/internal/component/database_observability/mysql/collector/exclude_schemas.go
+++ b/internal/component/database_observability/mysql/collector/exclude_schemas.go
@@ -1,10 +1,6 @@
 package collector
 
-import (
-	"strings"
-
-	"github.com/grafana/alloy/internal/component/database_observability"
-)
+import "github.com/grafana/alloy/internal/component/database_observability"
 
 var excludedSchemas = []string{"mysql", "performance_schema", "sys", "information_schema"}
 
@@ -19,29 +15,4 @@ func buildExcludedSchemasClause(schemas []string) string {
 	all = append(all, excludedSchemas...)
 	all = append(all, schemas...)
 	return database_observability.BuildExclusionClause(all)
-}
-
-// excludedSchemasArgs returns the combined default + custom excluded schemas
-// as query arguments, for binding as `?` placeholders instead of formatting
-// them into the SQL text.
-func excludedSchemasArgs(schemas []string) []any {
-	args := make([]any, 0, len(excludedSchemas)+len(schemas))
-	for _, s := range excludedSchemas {
-		args = append(args, s)
-	}
-	for _, s := range schemas {
-		args = append(args, s)
-	}
-	return args
-}
-
-// sqlPlaceholders returns a comma-separated list of n "?" placeholders, for
-// building a parameterized `IN (...)` / `NOT IN (...)` clause whose argument
-// count is dynamic.
-func sqlPlaceholders(n int) string {
-	placeholders := make([]string, n)
-	for i := range placeholders {
-		placeholders[i] = "?"
-	}
-	return strings.Join(placeholders, ", ")
 }
--- a/internal/component/database_observability/mysql/collector/index_stats.go
+++ b/internal/component/database_observability/mysql/collector/index_stats.go
@@ -1,0 +1,177 @@
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log/slog"
+
+	"github.com/blang/semver/v4"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+// IndexStatsCollector emits the minimal per-index metrics needed for the
+// unused-index KG insight: the fetch count of each named-index row per
+// index, from performance_schema.table_io_waits_summary_by_index_usage, and
+// (MySQL >= 5.6.6 only, see minIndexSizeEngineVersion) its size in bytes,
+// from mysql.innodb_index_stats.
+const IndexStatsCollector = "index_stats"
+
+const selectIndexIOWaits = `
+	SELECT OBJECT_SCHEMA, OBJECT_NAME, INDEX_NAME, COUNT_FETCH
+	FROM performance_schema.table_io_waits_summary_by_index_usage
+	WHERE INDEX_NAME IS NOT NULL AND OBJECT_SCHEMA NOT IN (%s)`
+
+// mysql.innodb_index_stats holds InnoDB's persistent optimizer statistics,
+// refreshed by MySQL itself (never by us) -- introduced in MySQL 5.6.6, its
+// schema (and the "size" stat, in pages) hasn't changed since. Reading it
+// requires the monitoring role to be granted SELECT on this one table, which
+// is not part of today's default grant set (see deployment_tools).
+const selectIndexSizeBytes = `
+	SELECT database_name, table_name, index_name, stat_value * @@innodb_page_size
+	FROM mysql.innodb_index_stats
+	WHERE stat_name = 'size' AND database_name NOT IN (%s)`
+
+// minIndexSizeEngineVersion is the first MySQL version where
+// mysql.innodb_index_stats exists at all (persistent optimizer statistics,
+// introduced 5.6.6) -- versions older than this have no size source, so the
+// query is skipped entirely below it rather than attempted and failing.
+var minIndexSizeEngineVersion = semver.MustParse("5.6.6")
+
+var indexLabels = []string{labelSchema, labelTable, "index"}
+
+var (
+	indexStatsIdxScanDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("mysql", "index_stats", "idx_scan_total"),
+		"Number of row fetches against this index",
+		indexLabels, nil,
+	)
+	indexStatsSizeBytesDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("mysql", "index_stats", "size_bytes"),
+		"Total disk space used by this index, in bytes",
+		indexLabels, nil,
+	)
+)
+
+type IndexStatsArguments struct {
+	DB             *sql.DB
+	ExcludeSchemas []string
+	EngineVersion  semver.Version
+	Registry       *prometheus.Registry
+
+	Logger *slog.Logger
+}
+
+type IndexStats struct {
+	dbConnection   *sql.DB
+	excludeSchemas []string
+	engineVersion  semver.Version
+	registry       *prometheus.Registry
+
+	logger  *slog.Logger
+	running *atomic.Bool
+}
+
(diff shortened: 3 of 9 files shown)


clayton-cornell:
Doc changes checks out and look OK as-is

internal/component/database_observability/mysql/collector/exclude_schemas.go
```diff
@@ -16,3 +20,29 @@ func buildExcludedSchemasClause(schemas []string) string {
...
+	all := make([]string, 0, len(excludedSchemas)+len(schemas))
+	all = append(all, excludedSchemas...)
+	all = append(all, schemas...)
+
+	args := make([]any, len(all))
+	for i, s := range all {
+		args[i] = s
+	}
+	return args
+}
```
cristiangreco: Feel that we can avoid `all` as it's just a temporary variable? e.g.


```go
func excludedSchemasArgs(schemas []string) []any {
	args := make([]any, 0, len(excludedSchemas)+len(schemas))
	for _, s := range excludedSchemas {
		args = append(args, s)
	}
	for _, s := range schemas {
		args = append(args, s)
	}
	return args
}
```

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,173 @@
...
+// one table, which isn't in today's default grant set.
+const selectIndexSizeBytes = `
+	SELECT database_name, table_name, index_name, stat_value * @@innodb_page_size
+	FROM mysql.innodb_index_stats
+	WHERE stat_name = 'size' AND database_name NOT IN (%s)`
+
+// minIndexSizeEngineVersion is the first MySQL version with
+// mysql.innodb_index_stats; older versions have no size source, so the
+// query is skipped below it rather than attempted and failing.
+var minIndexSizeEngineVersion = semver.MustParse("5.6.6")
```
cristiangreco: We officially support mysql 8+, maybe could avoid these compatibility checks?

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,173 @@
...
+	c.collectIdxScan(ctx, ch)
+
+	if c.engineVersion.GE(minIndexSizeEngineVersion) {
+		c.collectIndexSize(ctx, ch)
+	}
+}
+
+func (c *IndexStats) collectIdxScan(ctx context.Context, ch chan<- prometheus.Metric) {
+	args := excludedSchemasArgs(c.excludeSchemas)
+	query := fmt.Sprintf(selectIndexIOWaits, sqlPlaceholders(len(args)))
```
cristiangreco: really a nitpick, but it seems that instead of building placeholders and args we could just use something like:

```go
query := fmt.Sprintf(selectIndexIOWaits, buildExcludedSchemasClause(c.excludeSchemas))
```

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: Yeah that makes sense, applied it.

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: I was doing it this way before on the previous PR, but semgrep pointed a risk of SQL Injection.

docs/sources/reference/components/database_observability/database_observability.mysql.md
```diff
@@ -48,6 +48,8 @@ The following collectors are configurable:
 | `setup_actors`    | Check and update `performance_schema.setup_actors` settings. | yes                |
 | `locks`           | Collect queries that are waiting/blocking other queries.     | no                 |
 | `explain_plans`   | Collect explain plans information.                           | yes                |
+| `table_stats`     | Collect table-level index usage statistics.                  | no                 |
+| `index_stats`     | Collect per-index usage statistics.                           | no                 |
```
cristiangreco:
```suggestion
| `index_stats`     | Collect per-index usage statistics.                          | no                 |
```

internal/component/database_observability/mysql/collector/index_stats.go
cristiangreco: We have generally discarded those reports so far, given the context where this code runs and the fact that these arguments are user-controlled.

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,160 @@
...
+	SELECT database_name, table_name, index_name, stat_value * @@innodb_page_size
+	FROM mysql.innodb_index_stats
+	WHERE stat_name = 'size' AND database_name NOT IN (%s)`
+
+var indexLabels = []string{labelSchema, labelTable, "index"}
+
+var (
+	indexStatsIdxScanDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("database_observability", "mysql_index_stats", "idx_scan_total"),
+		"Number of row fetches against this index",
```
cristiangreco: Wondering about:
- instead of `idx_scan_total` could use `idx_fetch_total`
- we could describe it as "Count of index I/O wait events for fetch operations"

Don't feel too strong, just a thought about consistency with mysql terminology and docs.

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: I have just applied it, and it is a blocking failure on ci. Are you over writing it somehow?

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: Good catch, it makes sense to rename. Applying it.

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: I see there is a lint ignore we can use. Wonder if that is overkill in this case though?

internal/component/database_observability/mysql/collector/index_stats.go
gaantunes: Applying the same principle to rename `database_observability_mysql_table_stats_seq_scan_total` to `database_observability_mysql_table_stats_no_index_fetch_total`

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,158 @@
...
+const IndexStatsCollector = "index_stats"
+
+const selectIndexIOWaits = `
+	SELECT OBJECT_SCHEMA, OBJECT_NAME, INDEX_NAME, COUNT_FETCH
+	FROM performance_schema.table_io_waits_summary_by_index_usage
+	WHERE INDEX_NAME IS NOT NULL AND OBJECT_SCHEMA NOT IN %s`
+
+// mysql.innodb_index_stats holds InnoDB's persistent optimizer statistics,
+// refreshed by MySQL itself, never by us. Reading it requires SELECT on this
+// one table, which isn't in today's default grant set.
```
cristiangreco:
```suggestion
// mysql.innodb_index_stats holds InnoDB's persistent optimizer statistics,
// refreshed by MySQL itself.
```

Just to make it less verbose.

internal/component/database_observability/mysql/collector/table_stats.go
```diff
@@ -0,0 +1,114 @@
...
+
+const (
+	labelSchema = "schema"
+	labelTable  = "table"
+)
+
+var tableStatsNoIndexFetchDesc = prometheus.NewDesc(
+	prometheus.BuildFQName("database_observability", "mysql_table_stats", "no_index_fetch_total"),
+	"Number of row fetches against this table that did not use an index",
+	[]string{labelSchema, labelTable}, nil,
```
cristiangreco:
```suggestion
	prometheus.BuildFQName("database_observability", "mysql_table_stats", "no_idx_fetch_total"),
	"Count of index I/O wait events for fetch operations that did not use an index",
	[]string{labelSchema, labelTable}, nil,
```

internal/component/database_observability/mysql/collector/table_stats.go
```diff
@@ -0,0 +1,114 @@
...
+// Describe implements prometheus.Collector.
+func (c *TableStats) Describe(ch chan<- *prometheus.Desc) {
+	ch <- tableStatsSeqScanDesc
+}
+
+// Collect implements prometheus.Collector. It runs synchronously at scrape time.
+func (c *TableStats) Collect(ch chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	query := fmt.Sprintf(selectTableIOWaitsNoIndex, buildExcludedSchemasClause(c.excludeSchemas))
```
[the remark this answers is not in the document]
gaantunes: /fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,158 @@
...
+// Collect implements prometheus.Collector. It runs synchronously at scrape time.
+func (c *IndexStats) Collect(ch chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	c.collectIdxFetch(ctx, ch)
+	c.collectIndexSize(ctx, ch)
+}
+
+func (c *IndexStats) collectIdxFetch(ctx context.Context, ch chan<- prometheus.Metric) {
+	query := fmt.Sprintf(selectIndexIOWaits, buildExcludedSchemasClause(c.excludeSchemas))
```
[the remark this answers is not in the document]
gaantunes: /fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,158 @@
...
+		ch <- prometheus.MustNewConstMetric(indexStatsIdxScanDesc, prometheus.CounterValue, float64(countFetch), objectSchema, objectName, indexName)
+	}
+
+	if err := rows.Err(); err != nil {
+		c.logger.Error("error iterating table_io_waits_summary_by_index_usage rows", "err", err)
+	}
+}
+
+func (c *IndexStats) collectIndexSize(ctx context.Context, ch chan<- prometheus.Metric) {
+	query := fmt.Sprintf(selectIndexSizeBytes, buildExcludedSchemasClause(c.excludeSchemas))
```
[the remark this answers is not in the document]
gaantunes: /fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/index_stats.go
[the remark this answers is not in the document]
gaantunes: > **Semgrep identified a blocking 🔴 issue in your code:**
> 
> `c.excludeSchemas` is interpolated into `selectIndexIOWaits` with `fmt.Sprintf` before execution by `QueryContext`. If an attacker can control that configuration value, they may inject SQL that bypasses schema filtering or changes the query executed during metric collection.
> 
> More details about this
> `collectIdxFetch` builds the SQL text with `fmt.Sprintf(selectIndexIOWaits, buildExcludedSchemasClause(c.excludeSchemas))` and then executes the result directly through `c.dbConnection.QueryContext`. The value returned by `buildExcludedSchemasClause` is inserted into the query as SQL syntax rather than passed as a parameter, so any attacker-controlled or tampered value in `c.excludeSchemas` can alter the statement.
> 
> A plausible attack is:
> 
> 1. An attacker gains control of the configuration that populates `c.excludeSchemas` (for example, through an exposed admin/configuration endpoint) and supplies a value such as `app') OR 1=1 -- `.
> 2. `buildExcludedSchemasClause(c.excludeSchemas)` incorporates that value into the exclusion clause.
> 3. `fmt.Sprintf` places the resulting clause into `selectIndexIOWaits`, producing a query whose `WHERE` condition can be changed to `OR 1=1`, with the remainder commented out.
> 4. During the next Prometheus scrape, `collectIdxFetch` sends the modified query to MySQL via `QueryContext`, potentially returning indexes from schemas that were meant to be excluded. Depending on the surrounding SQL and database permissions, the injected statement could also expose additional data or perform unauthorized database operations.
> 
> **To resolve this comment:**
> 
> ✨ Commit fix suggestion
> 1. Replace `buildExcludedSchemasClause(c.excludeSchemas)` with bound SQL arguments. Combine the default and configured schemas into `[]any`, preserving the same order used by the query.
> 2. Generate one `?` placeholder per schema, for example with a helper such as `sqlPlaceholders(n int) string` that joins `n` question marks with commas. Only generate the placeholder list; never insert schema values into the SQL string.
> 3. Update `selectIndexIOWaits` so the exclusion clause includes parentheses: `OBJECT_SCHEMA NOT IN (%s)`.
> 4. Build and execute the index query with the placeholders and arguments:
>    `args := excludedSchemasArgs(c.excludeSchemas)`
>    `query := fmt.Sprintf(selectIndexIOWaits, sqlPlaceholders(len(args)))`
>    `rows, err := c.dbConnection.QueryContext(ctx, query, args...)`
> 5. Apply the same parameterized approach to `collectIndexSize`: change its SQL template to `NOT IN (%s)`, create the schema arguments, generate the placeholders, and pass the arguments to `QueryContext`.
> 6. Keep the SQL structure static and pass all schema values through `QueryContext`; values supplied as query arguments are handled separately from the SQL syntax, preventing schema names from altering the query.
> 
> 💬 Ignore this finding
> 
> Reply with Semgrep commands to ignore this finding.
> 
> * `/fp <comment>` for false positive
> * `/ar <comment>` for acceptable risk
> * `/other <comment>` for all other reasons
> 
> Alternatively, triage in [Semgrep AppSec Platform](https://semgrep.dev/orgs/grafana_labs/findings/963452532) to ignore the finding created by [string-formatted-query](https://semgrep.dev/playground/r/ZRTKA2q/go.lang.security.audit.database.string-formatted-query.string-formatted-query?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule).
> 
> We're currently testing semgrep's diff-aware PR comment feature on a subset of our repos-- if you run into issues or find this spammy, please reach out to @danny.cooper in slack and give feedback.
> 
> For backwards compatability with gosec, its best to use polyglot suppression comments of the following format for false positives: `// #nosec <gosec rule ID> nosemgrep: <semgrep rule ID>`
> 
> _You can view more details about [this finding](https://semgrep.dev/orgs/grafana_labs/findings/963452532) in the Semgrep AppSec Platform._

/fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/table_stats.go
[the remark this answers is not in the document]
gaantunes: > **Semgrep identified a blocking 🔴 issue in your code:**
> 
> `fmt.Sprintf` assembles the SQL executed by `QueryContext` using a clause derived from `c.excludeSchemas`. If that configuration is attacker-controlled and the clause is not safely represented as SQL literals, an attacker can alter the collector query and potentially run SQL with the database account’s privileges.
> 
> More details about this
> `Collect` builds the SQL text in `query` with `fmt.Sprintf`, inserting the result of `buildExcludedSchemasClause(c.excludeSchemas)`, and then executes that text through `c.dbConnection.QueryContext`. If any value in `c.excludeSchemas` can be influenced by an operator, deployment setting, or another untrusted input—and `buildExcludedSchemasClause` places it into SQL without safe SQL-literal handling—the input can change the query structure rather than merely select a schema.
> 
> A plausible exploit is:
> 
> 1. An attacker gains control of the configuration that populates `c.excludeSchemas` and supplies a schema value such as `public') OR 1=1 --` (the exact payload depends on the clause format produced by `buildExcludedSchemasClause`).
> 2. `buildExcludedSchemasClause(c.excludeSchemas)` incorporates that value into the SQL fragment passed to `fmt.Sprintf`, so `query` contains attacker-controlled SQL syntax.
> 3. `c.dbConnection.QueryContext(ctx, query)` sends the altered query to PostgreSQL during every Prometheus scrape. The attacker could make the collector return rows it was intended to exclude, alter the filtering logic, or—if the database driver and resulting statement permit stacked statements—execute additional SQL using the database account's privileges.
> 
> The finding is specifically at the `fmt.Sprintf` call because the eventual database execution receives one assembled SQL string; the safety of this code depends on `c.excludeSchemas` and the escaping behavior of `buildExcludedSchemasClause`, neither of which is visible here.
> 
> **To resolve this comment:**
> 
> ✨ Commit fix suggestion
> 1. Inspect `buildExcludedSchemasClause` and ensure the schema list cannot contain user-controlled input. Do not concatenate raw schema names into SQL.
> 2. Refactor the helper to return a SQL fragment and bound arguments, for example `(string, []any)`. Generate placeholders for each excluded schema, such as `... NOT IN ($1, $2)`, and return the schema names separately.
> 3. Pass the returned arguments to `QueryContext` instead of embedding values in the formatted query:
>    `clause, args := buildExcludedSchemasClause(c.excludeSchemas)`
>    `query := fmt.Sprintf(selectTableIOWaitsNoIndex, clause)`
>    `rows, err := c.dbConnection.QueryContext(ctx, query, args...)`
> 4. If the helper must generate SQL identifiers rather than values, validate each identifier against an allowlist and quote it with the database driver's identifier-quoting mechanism. Never insert unchecked configuration or request data into the query string.
> 5. If `buildExcludedSchemasClause` is proven to produce only fixed, internally generated SQL and cannot receive external data, keep the formatting and document the review with a targeted suppression on the query line, such as `// #nosec G201 nosemgrep: go.lang.security.audit.database.string-formatted-query.string-formatted-query`. Do not use this suppression if schema names or the generated clause can be influenced externally.
> 
> 💬 Ignore this finding
> 
> Reply with Semgrep commands to ignore this finding.
> 
> * `/fp <comment>` for false positive
> * `/ar <comment>` for acceptable risk
> * `/other <comment>` for all other reasons
> 
> Alternatively, triage in [Semgrep AppSec Platform](https://semgrep.dev/orgs/grafana_labs/findings/963448769) to ignore the finding created by [string-formatted-query](https://semgrep.dev/playground/r/ZRTKA2q/go.lang.security.audit.database.string-formatted-query.string-formatted-query?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule).
> 
> We're currently testing semgrep's diff-aware PR comment feature on a subset of our repos-- if you run into issues or find this spammy, please reach out to @danny.cooper in slack and give feedback.
> 
> For backwards compatability with gosec, its best to use polyglot suppression comments of the following format for false positives: `// #nosec <gosec rule ID> nosemgrep: <semgrep rule ID>`
> 
> _You can view more details about [this finding](https://semgrep.dev/orgs/grafana_labs/findings/963448769) in the Semgrep AppSec Platform._

/fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/index_stats.go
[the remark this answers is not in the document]
gaantunes: > **Semgrep identified a blocking 🔴 issue in your code:**
> 
> `buildExcludedSchemasClause(c.excludeSchemas)` is interpolated into a SQL query before `QueryContext` executes it. An attacker who can control the excluded-schema configuration may inject SQL and expose or alter the index statistics returned by the exporter.
> 
> More details about this
> `collectIndexSize` builds the SQL text with `fmt.Sprintf(selectIndexSizeBytes, buildExcludedSchemasClause(c.excludeSchemas))` and then executes that text through `c.dbConnection.QueryContext`. The result of `buildExcludedSchemasClause` is therefore inserted into the SQL structure rather than passed as a bound value; if an attacker can influence `c.excludeSchemas` (for example through exporter configuration or a writable config source) and that helper does not safely encode the values, the attacker can change the query predicate.
> 
> A plausible attack is:
> 
> 1. Set an excluded schema value to `analytics') OR 1=1 -- `.
> 2. `buildExcludedSchemasClause(c.excludeSchemas)` incorporates that value into the exclusion clause, producing SQL with an injected `OR 1=1` and comment marker.
> 3. During the next metrics scrape, `collectIndexSize` formats that clause into `query` and sends it to MySQL via `c.dbConnection.QueryContext`.
> 4. MySQL evaluates the altered predicate, potentially returning index statistics that the operator intended to exclude or allowing other query manipulation depending on the surrounding `selectIndexSizeBytes` template. Those rows are then exported through `indexStatsSizeBytesDesc`, exposing the unexpected database, table, and index names to anyone who can read the metrics endpoint.
> 
> The same string-construction pattern also appears in `collectIdxScan`, where `buildExcludedSchemasClause(c.excludeSchemas)` is formatted into `selectIndexIOWaits` before `QueryContext` executes it.
> 
> **To resolve this comment:**
> 
> ✨ Commit fix suggestion
> 1. Build the excluded schema values as query arguments with `excludedSchemasArgs(c.excludeSchemas)` instead of inserting them into the SQL string.
> 2. Update `selectIndexSizeBytes` so its exclusion clause uses `NOT IN (%s)`, leaving `%s` only for the generated placeholders.
> 3. Generate one `?` placeholder per argument with `sqlPlaceholders(len(args))`.
> 4. Replace the formatted query with `query := fmt.Sprintf(selectIndexSizeBytes, sqlPlaceholders(len(args)))`.
> 5. Pass the values separately to the database call with `c.dbConnection.QueryContext(ctx, query, args...)`.
> 6. Apply the same parameterized pattern to any related collector query that still calls `buildExcludedSchemasClause` when constructing SQL. This keeps schema names out of the SQL text and prevents external values from changing the query structure.
> 
> 💬 Ignore this finding
> 
> Reply with Semgrep commands to ignore this finding.
> 
> * `/fp <comment>` for false positive
> * `/ar <comment>` for acceptable risk
> * `/other <comment>` for all other reasons
> 
> Alternatively, triage in [Semgrep AppSec Platform](https://semgrep.dev/orgs/grafana_labs/findings/963448770) to ignore the finding created by [string-formatted-query](https://semgrep.dev/playground/r/ZRTKA2q/go.lang.security.audit.database.string-formatted-query.string-formatted-query?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule).
> 
> We're currently testing semgrep's diff-aware PR comment feature on a subset of our repos-- if you run into issues or find this spammy, please reach out to @danny.cooper in slack and give feedback.
> 
> For backwards compatability with gosec, its best to use polyglot suppression comments of the following format for false positives: `// #nosec <gosec rule ID> nosemgrep: <semgrep rule ID>`
> 
> _You can view more details about [this finding](https://semgrep.dev/orgs/grafana_labs/findings/963448770) in the Semgrep AppSec Platform._

/fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

internal/component/database_observability/mysql/collector/index_stats.go
```diff
@@ -0,0 +1,158 @@
...
+// Collect implements prometheus.Collector. It runs synchronously at scrape time.
+func (c *IndexStats) Collect(ch chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	c.collectIdxScan(ctx, ch)
+	c.collectIndexSize(ctx, ch)
+}
+
+func (c *IndexStats) collectIdxScan(ctx context.Context, ch chan<- prometheus.Metric) {
+	query := fmt.Sprintf(selectIndexIOWaits, buildExcludedSchemasClause(c.excludeSchemas))
```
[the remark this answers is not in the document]
gaantunes: > String-formatted SQL query detected. This could lead to SQL injection if the string is not sanitized properly. Audit this call to ensure the SQL is not manipulable by external data.
> 
> 🧹 _Fixed in commit [f16eed9](https://github.com/grafana/alloy/commit/f16eed96c19658c24ec1747411f5f9da7f663f9f)_ 🧹

/fp since this is running within an alloy instance and these values are fully under the user control within a config file, the risk is very low.

cristiangreco (approved):

PR #7070, opened 2026-09-10
docs: Fix bad link in the OTel components inclusion guide

blewis12:
The current link in the OTel component inclusion docs to the OCB guide takes users to the .md file, but that file is supposed to be rendered in the web docs since it includes `{{ ... }}` params. This PR fixes the link so it points to the web docs instead of the raw file.


clayton-cornell (approved):
Ooops missed that in the reviews.

PR #7075, opened 2026-09-10
fix(database_observability.postgres): Set explain plan `search_path` to all schemas of the database

cristiangreco:
The `explain_plans` collector set the session `search_path` to the database name plus `public` before preparing each query. `search_path` takes schemas, not databases, so unqualified table names that live in any other schema failed to resolve: EXPLAIN returned "relation does not exist" and no plan was produced.

Here we source the database's schemas from the TableRegistry that `schema_details` already refreshes periodically and put them all on the `search_path` (`public` last). This reuses the existing cache instead of adding a per-query schema lookup. When the registry is unavailable it falls back to using just `public`.

Caveat: table names duplicated across different schemas resolve by `search_path` order. `queryid` calculation does not encode the app's `search_path`, so exact per-schema attribution not possible; the plan is generated against one representative schema.

Fixes https://github.com/grafana/grafana-dbo11y-app/issues/2307

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/database_observability/postgres/collector/explain_plans.go
+++ b/internal/component/database_observability/postgres/collector/explain_plans.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"math"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


gaantunes (approved):
LGTM.

PR #7078, opened 2026-09-10
refactor(common/net): Use native Go HTTP/2 for receiver servers

### LINKED ISSUE #6347: Migrate common/net HTTP/2 setup off deprecated x/net/http2/h2c
Migrate `internal/component/common/net` HTTP/2 setup off the deprecated `golang.org/x/net/http2/h2c` package to `http.Server.Protocols` + `http.Server.HTTP2` (stdlib). The migration needs to decide what to do with three Alloy `http2 { }` block attributes that have no stdlib equivalent — `max_handlers` (always a no-op upstream), `idle_timeout`, and `read_idle_timeout` (closest fit is `SendPingTimeout`) — which is why a `//nolint:staticcheck` was added in #6336 instead of doing the migration there.

thampiotr:

Replace the deprecated `x/net/http2/h2c` receiver handler with Go’s native HTTP/2 support, retaining existing configuration names without adding new options or user-facing documentation.

### Pull Request Details

The upgrade removed support for `Upgrade: h2c` handshake. It was never documented by Alloy and has [limited deployment](https://www.rfc-editor.org/rfc/rfc9113.html#section-3.1), so breakage is expected to be very rare and within our backward compatibility guarantees. Decided to not document this to reduce confusion. 

There are some undocumented, existing HTTP/2 knobs which originated with Pyroscope’s former upload transport, but this implementation swap doesn’t promote them into a broader documented configuration surface. This is by choice, as for the purposes of Alloy, we don't need UI and APIs to support the HTTP/2 at the moment. 

Fixes #6347.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/shared/reference/components/server-http.md
+++ b/docs/sources/shared/reference/components/server-http.md
@@ -16,5 +16,3 @@ You can use the following arguments to configure the `http` block. Any omitted f
 | `server_idle_timeout`  | `duration` | Idle timeout for HTTP server.                                                                                    | `"120s"` | no       |
 | `server_read_timeout`  | `duration` | Read timeout for HTTP server.                                                                                    | `"30s"`  | no       |
 | `server_write_timeout` | `duration` | Write timeout for HTTP server.                                                                                   | `"30s"`  | no       |
-
-For unencrypted HTTP/2, clients must use prior knowledge rather than the legacy `Upgrade: h2c` handshake.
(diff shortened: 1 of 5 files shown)


docs/sources/shared/reference/components/server-http.md
```diff
@@ -16,3 +16,5 @@ You can use the following arguments to configure the `http` block. Any omitted f
 | `server_idle_timeout`  | `duration` | Idle timeout for HTTP server.                                                                                    | `"120s"` | no       |
 | `server_read_timeout`  | `duration` | Read timeout for HTTP server.                                                                                    | `"30s"`  | no       |
 | `server_write_timeout` | `duration` | Write timeout for HTTP server.                                                                                   | `"30s"`  | no       |
+
+For unencrypted HTTP/2, clients must use prior knowledge rather than the legacy `Upgrade: h2c` handshake.
```
thampiotr: I think that we never documented the `h2c` specific upgrades, so this will be fine to remove entirely.

kalleep (approved):

PR #4613, opened 2025-10-14
fix: stop loki.source.kubernetes discarding log lines with duplicate timestamps (#3959)

ciaranj:
#### PR Description
This code should help stop the loki.source.kubernetes from discarding log lines from kubernetes pods with the same timestamp.

#### Which issue(s) this PR fixes
#3959 

#### Notes to the Reviewer
I'm amending this to contain some tests.
#### PR Checklist

- [X] CHANGELOG.md updated

PR #7080, opened 2026-09-10
fix(loki.source.kubernetes): Forward log lines with out-of-order timestamps

### LINKED ISSUE #7076: "loki.source.kubernetes" drops log lines whose timestamp is earlier than the previous line (stdout/stderr interleaving)
### Component(s)

loki.source.kubernetes

### What's wrong?

`loki.source.kubernetes` drops any line with a timestamp older than the previous line. This is the check in [kubetail/tailer.go](https://github.com/grafana/alloy/blob/main/internal/component/loki/source/kubernetes/kubetail/tailer.go#L284):

```
if entryTimestamp.Before(lastReadTime) {
    continue
}
lastReadTime = entryTimestamp
```

Containerd stamps `stdout` and `stderr` in two goroutines and writes both to the same file. When an app writes to both at the same moment, a stderr line with a newer timestamp can land before a stdout line with an older one. Alloy skips the stdout line, silently: no metric, no log, no discard in Loki.

I understand `lastReadTime` is there to avoid resending lines after a reconnect (`SinceTime` has second precision). But since it moves on every line it also filters the live stream. Any app writing stdout and stderr at the same time is affected.

In our production, with Loki Canary in `stdout` mode: 199 of 5035 entries lost in 14 hours (4%). A patched image that compares against the resume point instead of the previous line lost 0 of 1435 in 2 hours on two clusters. The control cluster on the upstream image lost 25 of 717 in the same window.

### Steps to reproduce

The easiest way to see it is with Loki Canary in `stdout` mode, because the canary writes the entries to `stdout` and its own messages to `stderr`, and from time to time both happen in the same instant.

1. Run loki.source.kubernetes sending to loki.write. Clustering on or off does not matter.
2. Deploy loki-canary with `-push=false`, so it writes to `stdout` and expects the agent to ship the lines. Keep the pod label indexed for the canary and use `-streamname` container `-streamvalue ` canary, because the API tailer does not add a stream label. You can add `-query-append '|~ "^[0-9]+ p"'` so it only compares real entries.
3. Wait some minutes and compare kubectl logs --timestamps of a canary pod with what is in Loki.

Not every canary pod shows it, depends on the timing of the pod. When a pod is affected, one of its two 30 second slots loses lines, around 1 of 3.

This is what we got on one pod. kubectl had 31 entries in 15 minutes, Loki had 23. The kubelet log around one of the missing entries:

```
2026-09-09T21:44:01.011721089Z Querying loki for logs with query: ...   <- stderr
2026-09-09T21:44:01.011798230Z Querying loki for logs with query: ...   <- stderr
2026-09-09T21:44:01.011741569Z 1788990241011511645 ppppppppp...         <- stdout, 57us older than the line before, written after it
2026-09-09T21:44:01.020814497Z Querying loki for logs with query: ...
```

**The entry with .011741569 never reached Loki.**

### System information

Amazon Linux 2023, kernel 6.12.103 x86_64, Kubernetes v1.35.7 (EKS), containerd 2.2.5

### Software version

Grafana Alloy v1.16.1 (Helm chart alloy 1.8.1)

### Configuration

```text
loki.source.kubernetes "pod_logs" {
  targets    = discovery.relabel.pod_logs.output
  forward_to = [loki.process.pipeline.receiver]
  clustering {
    enabled = true
  }
}
```

### Logs

```text
# Alloy with logging level info. These are all the events for the affected target in the 20 minutes where 7 entries were lost. One stream opened, no reconnect, no retry.

ts=2026-09-09T21:29:53Z level=info msg="tailer running"     target=loki/loki-canary-5bc777fb96-4pj2r:canary
ts=2026-09-09T21:30:06Z level=info msg="opened log stream"  target=loki/loki-canary-5bc777fb96-4pj2r:canary start_time=0001-01-01T00:00:00Z

# Nothing else is logged when a line is skipped.
```

### AI disclosure

### Tip

johnpinto-clarity:
The tailer compared every line with the previous one and skipped the older ones. This dropped stdout lines that **containerd** wrote after a `stderr` line with a newer timestamp. Now it only compares with the resume point of the stream, and the saved position moves with the highest timestamp seen. **Two tests added:** an out-of-order line on a live stream is forwarded, a line older than the resume time is still skipped.

**Pull Request Details:**
`lastReadTime` is there to not resend lines after a reconnect (`SinceTime` has second precision). Moving it on every line made it a filter on the live stream, and Kubernetes does not guarantee order between `stdout` and `stderr` in the CRI log file.

Same behaviour kept for the resume: lines before the resume point are skipped, equal ones go through (like [#4613](https://github.com/grafana/alloy/pull/4613)). Positions and `target.LastEntry` use max() so a re-tail never resumes from an older timestamp.

**One thing not fixed here:** a line out of order exactly at a re-tail boundary can still be lost, a timestamp position cannot tell "already sent" from "written later with an older timestamp". A small slack on the resume point would fix it at the cost of some duplicates per re-tail. Open to do it if you prefer.

**Tested in production:** 0 of 1435 canary entries lost in 2 hours with this patch, 25 of 717 lost in the same window on a control cluster with the upstream image.

Issue(s) fixed:
Fixes #7076 

### PR Checklist
- [ ] Documentation added
- [x] Tests updated
- [x] This pull request was substantially generated with AI assistance

### DIFF
--- a/internal/component/loki/source/kubernetes/kubetail/tailer.go
+++ b/internal/component/loki/source/kubernetes/kubetail/tailer.go
@@ -266,9 +266,16 @@ func (t *tailer) tail(ctx context.Context, handler loki.EntryHandler) error {
 
 // processLogStream reads log lines from a reader and processes them.
 // It returns when the context is done, the stream ends, or an error occurs.
-func (t *tailer) processLogStream(ctx context.Context, stream io.ReadCloser, handler loki.EntryHandler, lastReadTime time.Time, positionsEnt positions.Entry, calc *rollingAverageCalculator) error {
+//
+// sinceTime is the time the stream was opened from. The API server applies
+// SinceTime with second precision, so the head of the stream can repeat lines
+// that were already shipped; those are the only lines skipped. Lines are never
+// compared against the previous line: the container runtime stamps stdout and
+// stderr independently, so a live stream is not ordered by timestamp.
+func (t *tailer) processLogStream(ctx context.Context, stream io.ReadCloser, handler loki.EntryHandler, sinceTime time.Time, positionsEnt positions.Entry, calc *rollingAverageCalculator) error {
 	ch := handler.Chan()
 	reader := bufio.NewReader(stream)
+	lastReadTime := sinceTime
 	for {
 		line, err := reader.ReadString('\n')
 
@@ -278,13 +285,11 @@ func (t *tailer) processLogStream(ctx context.Context, stream io.ReadCloser, han
 			calc.AddTimestamp(time.Now())
 
 			entryTimestamp, entryLine := parseKubernetesLog(line)
-			// Skip only if the timestamp is strictly before lastReadTime.
-			// This allows multiple log lines with the same timestamp to be processed,
-			// which is common in Windows containers that log rapidly.
-			if entryTimestamp.Before(lastReadTime) {
+			// Skip only if the timestamp is strictly before sinceTime, so that
+			// lines sharing the resume timestamp are still processed.
+			if entryTimestamp.Before(sinceTime) {
 				continue
 			}
-			lastReadTime = entryTimestamp
 
 			entry := loki.NewEntry(t.lset.Clone(), push.Entry{
 				Timestamp: entryTimestamp,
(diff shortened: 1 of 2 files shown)


internal/component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -266,9 +266,16 @@ func (t *tailer) tail(ctx context.Context, handler loki.EntryHandler) error {
 
 // processLogStream reads log lines from a reader and processes them.
 // It returns when the context is done, the stream ends, or an error occurs.
-func (t *tailer) processLogStream(ctx context.Context, stream io.ReadCloser, handler loki.EntryHandler, lastReadTime time.Time, positionsEnt positions.Entry, calc *rollingAverageCalculator) error {
+//
+// sinceTime is the time the stream was opened from. The API server applies
+// SinceTime with second precision, so the head of the stream can repeat lines
+// that were already shipped; those are the only lines skipped. Lines are never
+// compared against the previous line: the container runtime stamps stdout and
+// stderr independently, so a live stream is not ordered by timestamp.
```
kalleep: IMO this is the wrong placement for this comment, internal implementation detail to the function.

johnpinto-clarity: Agree, moved it next to the check and made it shorter.

johnpinto-clarity: Could you approve the workflow runs?

kalleep (approved):
So this change makes sense but I don't love that we will produce OOO logs but as far as I know there is not really a good way to prevent that without skipping logs.

Since we are in the middle of a release we currently have a code freeze. Will merge when that is over.

johnpinto-clarity:
Thanks @kalleep! Yeah, I couldn't find a clean way around it without dropping lines. Loki takes out-of-order writes by default anyway, and it can be tuned per tenant, so I'd rather forward them than lose them. No rush, happy to wait for the freeze.

Issue #6866, opened 2026-08-11
OTel Components: Wrapped config should start from OTel Defaults


kgeckhart:
### Component(s)

Most OTel Collector components

### Add a description

In the last two upgrades it was noted that most OTel components we wrap start from a "zero" config with no defaults applied. This can cause problems during upgrades when new fields are introduced and instead of us inheriting an upstream default we inject a "zero" value resulting in unexpected behavior. 

We should invert all of these so that we start from upstream defaults and override based on our exposed configuration. Doing this isn't a small task I did use AI to research and group these for us to start,

## Group 1: pure no-ops (7)

Upstream `createDefaultConfig` returns `&Config{}`, so seeding changes nothing. No goldens move. Cheapest way to establish the pattern.

| Component | `Convert()` |
|---|---|
| `otelcol.auth.basic` | `auth/basic/basic.go:112` |
| `otelcol.auth.headers` | `auth/headers/headers.go:50` |
| `otelcol.auth.sigv4` | `auth/sigv4/sigv4.go:46` |
| `otelcol.processor.attributes` | `processor/attributes/attributes.go:62` |
| `otelcol.processor.redaction` | `processor/redaction/redaction.go:157` |
| `otelcol.processor.span` | `processor/span/span.go:81` |
| `otelcol.receiver.fluentforward` | `receiver/fluentforward/fluentforward.go:63` |

## Group 2: one-line seed (7)

Already build into a named var, so the fix is one line at the top. 

`mapstructure.Decode` into a zero struct, identical shape to the #6842 k8sattributes fix:

| Component | `Convert()` |
|---|---|
| `otelcol.exporter.kafka` | `exporter/kafka/kafka.go:276` |
| `otelcol.receiver.kafka` | `receiver/kafka/kafka.go:207` |
| `otelcol.receiver.vcenter` | `receiver/vcenter/vcenter.go:404` |

`var result X.Config` + field assignment:

| Component | `Convert()` | Note |
|---|---|---|
| `otelcol.exporter.googlecloud` | `exporter/googlecloud/googlecloud.go:66` | half-done: seeds embedded `collector.DefaultConfig()` and hardcodes the factory's 12s timeout, but never calls the exporter factory default |
| `otelcol.exporter.googlecloudpubsub` | `exporter/googlecloudpubsub/googlecloudpubsub.go:71` | |
| `otelcol.processor.cumulativetodelta` | `processor/cumulativetodelta/cumulativetodelta.go:109` | |
[...]

PR #7085, opened 2026-09-10
refactor(otelcol): Seed factory defaults where Alloy already assigns them (group 2 of 5)

kgeckhart:

Group 2 of 5 of the upstream-defaults sweep. Seed the wrapped config from the component factory's default instead of a zero struct in five `otelcol` components. Their upstream defaults are non-empty, unlike group 1's, but Alloy already assigns every one of them, so this batch also changes no behavior. 

Related to https://github.com/grafana/alloy/issues/6866

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/exporter/googlecloud/googlecloud.go
+++ b/internal/component/otelcol/exporter/googlecloud/googlecloud.go
@@ -2,9 +2,6 @@
 package googlecloud
 
 import (
-	"time"
-
-	"github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/collector"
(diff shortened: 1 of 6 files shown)


kalleep (commented):
I think we should add tests to verify default config.

When we update it will be much easier to detect breaking changes

kalleep:
Nice I like it, do we already have these kind of tests for the exporters and the processor?

kgeckhart:
> Nice I like it, do we already have these kind of tests for the exporters and the processor?

I'm not 100% sure I'll update the ticket to add a final step to do a canary config test pass on all components for coverage.

kalleep (approved):

PR #7086, opened 2026-09-10
fix(database_observability.sql_server): Fix explain_plans timeout

rgeyer:

Rewrites the `explain_plans` collector's hash-to-plan resolution query
(`selectExplainPlansTemplate` in `explain_plans.go`) to stop timing out in
production. The query no longer joins `sys.query_store_runtime_stats` /
`sys.query_store_runtime_stats_interval` to rank plans by recency, and no
longer groups by the plan's `nvarchar(max)` XML across the resulting
fanned-out rows. Instead it ranks by `sys.query_store_plan.last_execution_time`
(with `plan_id` as a tiebreaker) and fetches the winning plan's XML via a
single `plan_id`-keyed join.

### Pull Request Details

**Root cause**: `sys.query_store_runtime_stats` has one row per
`(plan_id, interval, execution_type)`, and Query Store retains this history
for days/weeks. The old query joined through both runtime-stats views purely
to compute `MAX(interval.end_time)` for "most recent plan", then had to
`GROUP BY` on the fanned-out result including the full plan XML blob -
sorting/hashing that blob across a large fanned-out row set is what timed
out (confirmed via `context deadline exceeded`, 100s, in production logs).

**Semantic note**: ordering shifts from "latest runtime-stats interval end
time" to the plan's own `last_execution_time` (same intent - most recently
active plan wins - just computed without the fan-out). `last_execution_time`
is `NULL` for a plan that has never executed, which sorts last under `DESC`,
so an already-executed plan is still always preferred; `plan_id DESC` breaks
remaining ties deterministically.

**Validated compatibility**: this collector's "report on change, or at least
every 30 minutes" behavior lives entirely downstream of this query, in a
separate emission gate keyed off a structural hash of whichever plan this
query returns (`shouldEmit`, using the shared `EmitInterval` constant). This
query still runs every `collect_interval` regardless; the rewrite only
changes which `plan_id` wins per `query_hash` before that gate ever sees it,
so there's no interaction with or change to that cadence.

**Manual verification** (one-off, not part of this diff): seeded a throwaway
local SQL Server with 50 query_hashes, 2 plan variants each, and ~2.8
`runtime_stats_interval` rows per plan on average, then timed the old vs. new
query text with `SET STATISTICS TIME, IO ON`, 3 runs each:

| Query | Mean elapsed (ms) | Mean CPU (ms) | Logical + LOB reads |
|-------|--------------------|-----------------|----------------------|
| OLD   | 90.0               | 88.3            | 622                  |
| NEW   | 14.7               | 14.3            | 128                  |

~6x faster and ~5x fewer reads at this reduced scale. Since the old query's
cost scales multiplicatively with `plans x intervals`, and this session's
history (minutes) is far smaller than production's (days/weeks), this is a
conservative lower bound on the actual production regression, not a claim of
matching its exact magnitude.

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/database_observability/sql_server/collector/explain_plans.go
+++ b/internal/component/database_observability/sql_server/collector/explain_plans.go
@@ -30,28 +30,30 @@ const (
 // SET SHOWPLAN_XML ON compilation - the plan is read from the same Query
 // Store data query_metrics already joins). A query_hash can map to more than
 // one query_id/plan_id (recompiles, parameter sniffing), so ROW_NUMBER picks
-// exactly one plan per hash rather than enumerating every match.
+// exactly one plan per hash rather than enumerating every match. Recency is
+// ranked by the plan's own last_execution_time rather than the latest
+// runtime-stats interval end time, since that avoids joining through
+// runtime_stats/runtime_stats_interval entirely - see the plan_id-keyed fetch
+// below. plan_id DESC breaks ties (including NULL last_execution_time for a
+// plan that has never executed, which sorts last under DESC) deterministically.
 const selectExplainPlansTemplate = `
 	WITH ranked_plans AS (
 		SELECT
 			q.query_hash,
-			p.query_plan,
+			p.plan_id,
 			ROW_NUMBER() OVER (
 				PARTITION BY q.query_hash
-				ORDER BY MAX(i.end_time) DESC
+				ORDER BY p.last_execution_time DESC, p.plan_id DESC
 			) AS rn
 		FROM sys.query_store_query q
 		JOIN sys.query_store_plan p ON p.query_id = q.query_id
-		JOIN sys.query_store_runtime_stats rs ON rs.plan_id = p.plan_id
-		JOIN sys.query_store_runtime_stats_interval i
-			ON i.runtime_stats_interval_id = rs.runtime_stats_interval_id
 		WHERE q.is_internal_query = 0
 			AND q.query_hash IN (%s)
-		GROUP BY q.query_hash, p.plan_id, p.query_plan
 	)
-	SELECT query_hash, query_plan
-	FROM ranked_plans
-	WHERE rn = 1`
+	SELECT rp.query_hash, p.query_plan
+	FROM ranked_plans rp
+	JOIN sys.query_store_plan p ON p.plan_id = rp.plan_id
+	WHERE rp.rn = 1`
 
 type ExplainPlansArguments struct {
 	DB              *sql.DB


rgeyer:
A bit more detail from the one-off performance test used to validate the change.

## explain_plans query timing comparison (manual, one-off)

**Environment**: `mcr.microsoft.com/mssql/server:2022-latest` (Developer edition), single throwaway
container, local docker run. Not persisted — container removed after this run.

**Seed**: 50 distinct `query_hash`es (one per indexed column on a 50k-row synthetic table, matching
the component's real `StatementsLimit: 50` default), 2 distinct `plan_id`s per hash (seek plan, then
a scan plan forced by dropping the index), Query Store `INTERVAL_LENGTH_MINUTES = 1`, `QUERY_CAPTURE_MODE
= ALL`, ~2.8 `runtime_stats_interval` rows accumulated per plan on average (max 5) across 4 batches
spaced by real 65s sleeps.

**Caveat**: a third plan variant (filtered index) was attempted but the optimizer reused the existing
scan plan rather than compiling a new one (parameter-sniffing caution on the filtered predicate), so
fan-out here is 2 plans/hash, not 3. More importantly, interval/plan fan-out magnitude in this short
session (minutes of history, ~2.8 intervals/plan) is much smaller than production's (days/weeks of
retention, likely dozens-hundreds of intervals/plan). The old query's cost scales multiplicatively with
`plans × intervals`, so this is a conservative lower bound on the real production regression, not a
claim of matching its exact magnitude — production's actual timeout is consistent with a much larger
multiplier than this session reproduces.

### Timing

| Query | Run | Elapsed (ms) | CPU (ms) |
|-------|-----|---------------|-----------|
| OLD   | 1   | 87            | 86        |
| OLD   | 2   | 95            | 92        |
| OLD   | 3   | 88            | 87        |
| NEW   | 1   | 8             | 9         |
| NEW   | 2   | 17            | 15        |
| NEW   | 3   | 19            | 19        |

| Query | Mean elapsed (ms) | Mean CPU (ms) |
|-------|--------------------|----------------|
| OLD   | 90.0               | 88.3           |
| NEW   | 14.7               | 14.3           |

**~6.1x faster** at this reduced scale.

### Logical reads (SET STATISTICS IO, identical across the 3 runs of each query — deterministic against unchanged seed data)

| Query | Non-LOB logical reads | LOB logical reads (plan XML) | Combined |
|-------|------------------------|-------------------------------|----------|
| OLD   | 22 (`plan_persist_plan`: 7, `plan_persist_query`: 13, `plan_persist_runtime_stats_interval`: 2; `plan_persist_runtime_stats`/`_v2` scanned 100x each — 2 plans × 50 hashes — at 0 pages) | 600 | 622 |
| NEW   | 28 (`plan_persist_plan`: 14, `plan_persist_query`: 14) | 100 | 128 |

**~4.9x fewer logical reads.** The OLD query's LOB reads (600) reflect reading the plan XML blob
*during the GROUP BY*, once per fanned-out row (50 hashes × 2 plans × ~1.2 rows after the runtime-stats
join per plan in this session) rather than once per winning plan — this multiplier is exactly what
grows unbounded with retention/history in production. The NEW query's LOB reads (100 = 50 winning
plans × ~2 average bytes-unit reads per blob fetch) are bounded by the tracked hash count only, and do
not grow with history at all.

### Conclusion

The rewrite is directionally and substantially faster even at this reduced scale, and the mechanism
(eliminating a read of the plan XML blob per fanned-out runtime-stats/interval row, in favor of exactly
one blob read per winning plan) predicts the gap widens further, not narrows, as production's actual
history/fan-out grows — consistent with the rewrite resolving the observed 100s timeout.

cristiangreco (approved):

PR #7087, opened 2026-09-10
fix(loki.write): Don't hang on shutdown when the send queue is full

kgeckhart:

Let `loki.write` shut down while an endpoint's send queue is full.

### Pull Request Details

`Stop` waited on the goroutine feeding `endpoint.enqueue` before releasing it, so a full queue deadlocked shutdown and config reloads. Blocked writes now get `drain_timeout` to find room before they're dropped. Also corrects three arguments on the reference page whose documented defaults and behavior disagreed with the code.

### PR Checklist

- [x] Documentation added
- [x] Tests updated


kgeckhart:
Replaced by https://github.com/grafana/alloy/pull/7091

PR #7088, opened 2026-09-10
fix(otelcol.receiver.cloudflare): Apply the documented default arguments

kgeckhart:

Seed the cloudflare receiver's config from the factory default so `timestamp_field`, `timestamp_format` and `separator` get the defaults the reference docs already document, and rename the argument table's `delimiter` to `separator`, the name the component actually accepts. `SetToDefault` was an empty stub, so configs that left these unset got no log timestamp and nested attribute keys flattened with no separator between them. Those records now carry a `Timestamp`, and those keys gain a `.`.

Related to https://github.com/grafana/alloy/issues/6866

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/otelcol/receiver/cloudflare/cloudflare.go
+++ b/internal/component/otelcol/receiver/cloudflare/cloudflare.go
@@ -47,24 +47,25 @@ type Arguments struct {
 
 // SetToDefault implements syntax.Defaulter.
 func (args *Arguments) SetToDefault() {
-	// Defaults filled by upstream OTel receiver in a factory.
+	cfg := cloudflarereceiver.NewFactory().CreateDefaultConfig().(*cloudflarereceiver.Config)
+	*args = Arguments{
+		TimestampField:  cfg.Logs.TimestampField,
+		TimestampFormat: cfg.Logs.TimestampFormat,
+		Separator:       cfg.Logs.Separator,
+	}
 }
 
 func (args Arguments) receiverConfig() *cloudflarereceiver.Config {
 	tlsCfg := args.TLS.Convert()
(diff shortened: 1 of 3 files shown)


clayton-cornell:
Minimal doc change. Good as-is

internal/component/otelcol/receiver/cloudflare/cloudflare.go
```diff
@@ -47,24 +47,25 @@ type Arguments struct {
 
 // SetToDefault implements syntax.Defaulter.
 func (args *Arguments) SetToDefault() {
-	// Defaults filled by upstream OTel receiver in a factory.
+	cfg := cloudflarereceiver.NewFactory().CreateDefaultConfig().(*cloudflarereceiver.Config)
```
ptodev, line 50: This could make it easy for our docs to drift over time? If Alloy sets its own default values, then it's less likely to drift.

If there's a test to check for drift then that could be ok, but with the way the "minimal configuration" test case is currently written it wouldn't catch a change in an upstream default. We'd have to list the expected defaults explicitly in the test case.

kgeckhart: I went with option 2 because I don't think it's safe for us to carry our own defaults as we have no real defensible reason to drift from upstream. I renamed the test, it checks against static values, and the comment on the test says you need to update the docs then the values here. 

I added a follow up task to the defaults flip to verify we do this on all of them.

ptodev: Thank you!

ptodev (approved):

PR #7091, opened 2026-09-11
fix(loki.write): Potential deadlock when queue is full and  component shuts down

kalleep:

Be able to shutdown in respect to drain timeouts when `loki.write` is blocked on full queue.

### Pull Request Details

Before this pr loki.write would not be able to perform updates or shutdown if any entry was blocked being enqueued. So the drain timeouts would not be respected.

To be able to handle both of these cases I had to remove channel usage from both of our consumer clients (WAL and non-WAL). IMO channel usage where the consumer of that channel is also responsible to close that channel is a bad pattern.

So instead I updated to just a plain function call `ConsumeEntry`. This one will propagate context from component so we can detect shutdown. We also "snapshot" consumer under read lock before calling `ConsumeEntry` so that we can unblock a potential update call. If update happens `loki.ErrConsumerStopped` will be returned and we have to abandon the entry. 

Alternative implementation to https://github.com/grafana/alloy/pull/7087

Related to: https://github.com/grafana/alloy/issues/7094

### Notes to the Reviewer

First commit in this pr adds 2 test that failed without the fix.

### PR Checklist

- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/common/loki/client/consumer_fanout.go
+++ b/internal/component/common/loki/client/consumer_fanout.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -16,10 +17,12 @@ func NewFanoutConsumer(logger *slog.Logger, reg prometheus.Registerer, cfgs ...C
 		return nil, fmt.Errorf("at least one endpoint config must be provided")
 	}
 
-	m := &FanoutConsumer{
+	c := &FanoutConsumer{
 		endpoints: make([]*endpoint, 0, len(cfgs)),
 		recv:      make(chan loki.Entry),
 	}
+
+	c.ctx, c.cancel = context.WithCancel(context.Background())
 
 	var (
 		metrics        = newMetrics(reg)
@@ -39,27 +42,30 @@ func NewFanoutConsumer(logger *slog.Logger, reg prometheus.Registerer, cfgs ...C
 			return nil, fmt.Errorf("error starting endpoint: %w", err)
 		}
 
-		m.endpoints = append(m.endpoints, endpoint)
+		c.endpoints = append(c.endpoints, endpoint)
 	}
 
-	m.wg.Go(m.run)
-	return m, nil
+	c.wg.Go(c.run)
+	return c, nil
 }
 
 var _ Consumer = (*FanoutConsumer)(nil)
 
 type FanoutConsumer struct {
 	endpoints []*endpoint
-	wg        sync.WaitGroup
-	once      sync.Once
-	recv      chan loki.Entry
+
+	wg     sync.WaitGroup
+	once   sync.Once
+	recv   chan loki.Entry
+	ctx    context.Context
+	cancel context.CancelFunc
 }
 
 func (c *FanoutConsumer) run() {
 	for e := range c.recv {
-		for _, c := range c.endpoints {
+		for _, endpoint := range c.endpoints {
 			// NOTE: For now it's fine to ignore error because we can't act on it.
-			_ = c.enqueue(e, 0)
+			_ = endpoint.enqueue(c.ctx, e, 0)
 		}
 	}
 }
@@ -70,7 +76,11 @@ func (c *FanoutConsumer) Chan() chan<- loki.Entry {
 
 func (c *FanoutConsumer) Stop() {
 	// First stop the receiving channel.
-	c.once.Do(func() { close(c.recv) })
+	c.once.Do(func() {
+		close(c.recv)
+		c.cancel()
+	})
+
 	c.wg.Wait()
 
 	var stopWG sync.WaitGroup
--- a/internal/component/common/loki/client/consumer_wal.go
+++ b/internal/component/common/loki/client/consumer_wal.go
@@ -194,7 +194,7 @@ func (c *walEndpointAdapter) StoreSeries(series []record.RefSeries, segment int)
 	}
 }
 
-func (c *walEndpointAdapter) AppendEntries(entries wal.RefEntries, segment int) error {
+func (c *walEndpointAdapter) AppendEntries(ctx context.Context, entries wal.RefEntries, segment int) error {
 	c.seriesLock.RLock()
 	l, ok := c.series[entries.Ref]
 	c.seriesLock.RUnlock()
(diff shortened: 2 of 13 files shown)


internal/component/common/loki/client/endpoint.go
```diff
@@ -57,30 +43,34 @@ func newEndpoint(metrics *metrics, cfg Config, logger *slog.Logger, markerHandle
...
 
-		e.backoff.Wait()
-		if !e.backoff.Ongoing() {
-			return e.backoff.Err()
-		}
+		bo.Wait()
 	}
 
-	return nil
+	return bo.Err()
```
kgeckhart, line 75: If we got here we dropped data due to a cancellation on the context breaking the bo, right?

kgeckhart (commented):
Looks like both PRs still hang on a config update when the queue is full. Mine was supposed to fix it but it did not and goroutine dump on this one shows,

| Goroutine | Parked at | State |
|---|---|---|
| fanout `run` | `client/endpoint.go:67` (`bo.Wait()`) via `client/consumer_fanout.go:68` | retrying a full queue |
| mutator forwarder | `common/loki/entry_handler.go:86` | blocked on `nextChan <- f(e)` |
| `Component.Run` | `loki/write/write.go:139` | blocked on `c.sink.Chan() <- entry`, **holding `c.mut.RLock()`** |
| `Component.Update` | `loki/write/write.go:159` | blocked on `c.mut.Lock()` |

`Run`'s inner select only escapes on `ctx.Done()`, and a reload does not cancel the component context, so the read lock is never released.

internal/component/common/loki/client/endpoint.go
kalleep: Yeah so if WAL is enabled we dont since we dont update data as sent. But for non WAL we do drop data

internal/component/common/loki/client/consumer_fanout.go
```diff
@@ -39,40 +40,37 @@ func NewFanoutConsumer(logger *slog.Logger, reg prometheus.Registerer, cfgs ...C
...
+		err := e.enqueue(ctx, entry, 0)
+
+		if err != nil {
+			// If we get errQueueIsFull we skipped the entry for this endpoints
+			// and should try next.
+			if errors.Is(err, errQueueIsFull) {
+				continue
+			}
+			// The other errors we can get are the context error and loki.ErrConsumerStopped.
+			// In both these cases there is no point trying to enqueue for other endpoints.
```
kgeckhart:
```suggestion
			// For any other error we assume it's not useful to send to the next endpoint
```

The comment about what errors are possible is liable to rot

internal/component/common/loki/client/consumer_wal.go
```diff
@@ -108,8 +110,9 @@ type WALConsumer struct {
 	pairs  []endpointWatcherPair
 }
 
-func (m *WALConsumer) Chan() chan<- loki.Entry {
-	return m.writer.Chan()
+// ConsumeEntry implements DrainableConsumer.
+func (m *WALConsumer) ConsumeEntry(ctx context.Context, entry loki.Entry) error {
+	return m.writer.WriteEntry(entry)
 }
```
kgeckhart, line 116: Is there any value is checking/propagating the context here or is something else going to cancel this during a drain/shutdown scenario?

internal/component/common/loki/client/endpoint.go
```diff
@@ -35,52 +29,53 @@ func newEndpoint(metrics *metrics, cfg Config, logger *slog.Logger, markerHandle
...
+	for bo.Ongoing() {
+		err := e.shards.enqueue(tenantID, entry, segmentNum)
+
+		if err == nil {
+			return nil
+		}
+
+		if errors.Is(err, loki.ErrConsumerStopped) {
+			return err
+		}
```
kgeckhart, line 64: Could we add a test on this one. It doesn't looks like it's currently tested and shutdown was another fun edge case on this one.

internal/component/common/loki/client/util_test.go
```diff
@@ -0,0 +1,52 @@
...
+		case <-done:
+		case <-r.Context().Done():
+		}
+	}))
+
+	return server, blocked, release
+}
+
+func feedUntilBlocked(t *testing.T, blocked *atomic.Bool, consumer Consumer) {
+	e := loki.NewEntry(model.LabelSet{"A": "b"}, push.Entry{
```
kgeckhart, line 43: Nit
```suggestion
func feedUntilBlocked(t *testing.T, blocked *atomic.Bool, consumer Consumer) {
    t.Helper()
	e := loki.NewEntry(model.LabelSet{"A": "b"}, push.Entry{
```

Is this worth having a timeout on? There's nothing that couples it to newBlockedServer so something using it incorrectly would hang for a long time.

internal/component/common/loki/client/consumer_wal.go
kalleep: No not really, if we are in a drain state it's still fine to write entry to the WAL and it will be picked up when it's started again. If we have stopped it get back an error.

internal/component/common/loki/client/endpoint.go
kalleep: Yes will add one :+1:

kalleep: added in [43ed462](https://github.com/grafana/alloy/pull/7091/commits/43ed462837693da97ccde1538ddfaddf44d930f6)

internal/component/common/loki/client/util_test.go
kalleep: Sure I don't see why not, added a hard timeout in [840c140](https://github.com/grafana/alloy/pull/7091/commits/840c1407994680e5a32d04374ed250656b87f48b)

internal/component/loki/write/write.go
```diff
@@ -199,22 +183,24 @@ func (c *Component) Update(args component.Arguments) error {
 	}
 
 	if err != nil {
-		return fmt.Errorf("failed to create cliens: %w", err)
+		return fmt.Errorf("failed to create clients: %w", err)
```
kgeckhart, line 186: For a follow up because it's an existing issue, AI called out that if creating the consumer fails `c.consumer` is going to be nil leading to a panic 😬 . Looks pretty legit since we do have some failure modes with empty/duplicate endpoints on consumer creation.

kgeckhart (approved):
Nice work!

clayton-cornell (approved):
Doc input is small.. no suggestions. Good as-is

internal/component/loki/write/write.go
kalleep: Yes I saw that too. We also leak goroutines if we successfully create at least one endpoint but fails for another https://github.com/grafana/alloy/blob/main/internal/component/common/loki/client/consumer_fanout.go#L37.

Can fix both in another pr.

PR #7093, opened 2026-09-11
fix(loki.write): WAL writer metrics is now collected after reload

kalleep:

Fix WAL writer metrics after config reload.

### Pull Request Details

Writer metrics was registered with `reg.Register`. This one will return an error if we have already registered these metrics before, but the code ignored this error.

Because of that writer would update metrics that was not collected anymore after a reload happened. I changed it to accept a metrics struct instead to match other code here and use our util function when registering the metric.

Part of: https://github.com/grafana/alloy/issues/7094

### PR Checklist

- [ ] Documentation added
- [x] Tests updated


internal/component/common/loki/wal/writer_test.go
thampiotr, line 1: I think the tests are not able still to discover this as a bug. Would it be possible to add a check for this?

kalleep: Done

thampiotr (approved):

PR #7098, opened 2026-09-11
ci: Add action to validate items that make a PR unmergable before CI is approved

kgeckhart:

Add a `PR Preflight` workflow validating pull request title, brief description, and commit signatures. These are three relatively easy things to validate without a maintainer needing to approve CI surfacing issues that make a PR unable to be merged earlier. 

### Notes to the Reviewer

The new workflow duplicates the existing release-lint-pr-title which is currently required. If this new workflow works as expected it would be come required and that flow would be deleted afterwards.

I chose not to add another comment because we already get comments from CLA + commit signing. I don't think we can easily control those enough to do a single checklist comment so for now the failures are surfaced as an annotation which should be one click from the checks list. 

### PR Checklist

- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/.github/workflows/pr-preflight.yml
+++ b/.github/workflows/pr-preflight.yml
@@ -1,0 +1,132 @@
+name: PR Preflight
+
+on:
+  # SECURITY: This workflow uses pull_request_target so these checks report on fork
+  # pull requests without waiting for a maintainer to approve the workflow run.
+  #
+  # - Nothing is checked out and nothing from the contributor is executed
+  # - Only the base branch's copy of this workflow runs
+  # - The PR body reaches the shell through env, never template interpolation
+  # - Jobs grant at most pull-requests: read
+  #
+  # zizmor: ignore[dangerous-triggers] see above
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions: {}
+
+concurrency:
+  # Editing a title or description fires this repeatedly; only the last run matters.
+  # Keyed on PR number because branch names are not unique across forks.
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  title:
+    name: PR title
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read # Read the title to validate.
+    steps:
+      - name: Validate PR title 🔎
+        uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50 # v6
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        with:
+          # Require conventional commit types
+          types: |
+            feat
+            fix
+            docs
+            style
+            refactor
+            perf
+            test
+            ci
+            chore
+            revert
+            proposal
+          # Scope is optional
+          requireScope: false
+          # Disallow uppercase first letter in subject
+          subjectPattern: ^[A-Z].+$
+          subjectPatternError: |
+            The subject "{subject}" must start with an uppercase letter.
+            Example: "feat: Add new component" not "feat: add new component"
+
+  brief-description:
+    name: Brief description
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for a brief description 📝
+        # The default shell can fall back to sh and omits pipefail; this needs both.
+        shell: bash
+        env:
+          BODY: ${{ github.event.pull_request.body }}
+          HEADING: '### Brief description of Pull Request'
+          CONTRIBUTING: https://github.com/grafana/alloy/blob/main/docs/developer/contributing.md
+        run: |
+          # GitHub sends CRLF bodies, which would leave a stray CR on every line and
+          # make an "empty" section look filled. Drop HTML comments too, so the
+          # template's own guidance text doesn't count as content; perl reads the
+          # whole body at once, which a line-based strip could not do for a comment
+          # spanning several lines.
+          text=$(printf '%s\n' "${BODY}" | tr -d '\r' | perl -0777 -pe 's/<!--.*?-->//gs')
+
+          if grep -qF "${HEADING}" <<<"${text}"; then
+            section=$(awk -v h="${HEADING}" 'index($0,h)==1 {f=1; next} f && /^###/ {exit} f' <<<"${text}")
+            what='The "Brief description of Pull Request" section'
+          else
+            # The template was removed. What we guard against is an empty squash
+            # commit message, not a missing heading, so accept prose in its place.
+            section="${text}"
+            what='The pull request description'
+          fi
+
+          if [[ -z "${section//[[:space:]]/}" ]]; then
+            echo "::error::${what} is empty. Alloy squashes with the pull request body as the commit message, so it needs a factual summary of what changed. See ${CONTRIBUTING}"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-preflight.yml
```diff
@@ -0,0 +1,132 @@
...
+            chore
+            revert
+            proposal
+          # Scope is optional
+          requireScope: false
+          # Disallow uppercase first letter in subject
+          subjectPattern: ^[A-Z].+$
+          subjectPatternError: |
+            The subject "{subject}" must start with an uppercase letter.
+            Example: "feat: Add new component" not "feat: add new component"
```
jharvey10: This is a dupe of [release-lint-pr-title](https://github.com/grafana/alloy/blob/main/.github/workflows/release-lint-pr-title.yml). We could either switch that workflow to pull_request_target or we could sunset that one in favor of this one. I kind of like having that check in its own file, but I don't feel strongly either way.

.github/workflows/pr-preflight.yml
```diff
@@ -0,0 +1,132 @@
...
+            section="${text}"
+            what='The pull request description'
+          fi
+
+          if [[ -z "${section//[[:space:]]/}" ]]; then
+            echo "::error::${what} is empty. Alloy squashes with the pull request body as the commit message, so it needs a factual summary of what changed. See ${CONTRIBUTING}"
+            exit 1
+          fi
+
+          echo "${what} is filled in."
```
jharvey10: This scripting scares me a bit. I kinda sorta kinda follow the logic, but it's very obtuse. If this were an internal-only tool I'd probably be more okay with it, but given that we're in sensitive territory here with respect to security, my preference would be to make the logic look as simple and straightforward as possible for human reviewers.

Using `github-script` is perhaps contentious in its own right, but I'd be curious of your thoughts on this sort of thing as an alternative without going to a full-blow go script for this.

```yaml
- name: Check for a brief description 📝
  uses: actions/github-script@v7
  env:
    HEADING: "### Brief description of Pull Request"
    CONTRIBUTING: https://github.com/grafana/alloy/blob/main/docs/developer/contributing.md
  with:
    script: |
      const body = (context.payload.pull_request.body ?? "")
        .replace(/\r/g, "")
        .replace(/<!--[\s\S]*?-->/g, "");

      const heading = process.env.HEADING;
      const headingOffset = body.indexOf(heading);

      let description;
      let label;

      if (headingOffset >= 0) {
        const sectionStart = headingOffset + heading.length;
        const nextHeading = body.indexOf("\n###", sectionStart);

        description = body.slice(
          sectionStart,
          nextHeading === -1 ? undefined : nextHeading,
        );
        label = `The "${heading}" section`;
      } else {
        // A PR may deliberately omit the template; require prose somewhere.
        description = body;
        label = "The pull request description";
      }

      const hasNonWhitespace = /\S/.test(description);

      if (hasNonWhitespace === false) {
        core.setFailed(
          `${label} is empty. Alloy squashes with the pull request body as ` +
          `the commit message, so it needs a factual summary of what changed. ` +
          `See ${process.env.CONTRIBUTING}`,
        );
        return;
      }

      core.info(`${label} is filled in.`);
```

.github/workflows/pr-preflight.yml
```diff
@@ -0,0 +1,132 @@
...
+          # offending commits is the part a contributor needs in order to act.
+          {
+            echo '### Unsigned commits'
+            echo
+            echo '```'
+            echo "${unsigned}"
+            echo '```'
+            echo
+            echo "Re-sign these commits in your fork and force-push. See [Signed commits](${CONTRIBUTING})."
+          } >>"${GITHUB_STEP_SUMMARY}"
```
jharvey10: If we find users get confused about trying to figure out why these checks are failing, we have some good tooling for adding issue comments, but this expands the permissions scope of the jobs and might not be worth that tradeoff.

https://github.com/grafana/alloy/blob/main/tools/release/internal/github/client.go#L331

.github/workflows/pr-preflight.yml
```diff
@@ -0,0 +1,132 @@
...
+
+          echo "${what} is filled in."
+
+  signatures:
+    name: Commit signatures
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read # List the commits on the PR.
+    steps:
+      - name: Check commit signatures 🔑
```
jharvey10: Suggestion for a slight wording tweak that will show up in the checks on the PR without having to go to the job logs itself.

```suggestion
      - name: Ensure all commits are signed 🔑
```

.github/workflows/pr-preflight.yml
kgeckhart: Ahhh something that got missed in the PR description was that I want to condense around a single pre-flight vs having a bunch of different failed jobs. I'm not strongly opinionated about it though.

.github/workflows/pr-preflight.yml
kgeckhart: I would be fine with any language option TBH. What I was very much trying to avoid was needing this to checkout / clone the repo which is what led to inline scripting.

.github/workflows/pr-preflight.yml
kgeckhart: Yep I should have put this in the notes to reviewer. I considered doing a comment but if we go that route then we are going to get CLA comment, the commit signer leaves a comment, and then this leaves a comment. I would love a single comment with a checklist but didn't dig on how complex that would be with CLA and signing not being ours directly.

.github/workflows/pr-preflight.yml
jharvey10: 100% agree avoiding a checkout is ideal. I could be biased since I'm very comfortable parsing javascript/typescript. The github API within it is perhaps _slightly_ nicer, but the main benefit is functional slicing/dicing of the string.

kgeckhart: Done

.github/workflows/pr-preflight.yml
```diff
@@ -0,0 +1,139 @@
...
+  cancel-in-progress: true
+
+jobs:
+  preflight:
+    name: PR Preflight
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read # Read the title, body, and commit list.
+    steps:
+      - name: Ensure the title follows Conventional Commits 🔎
```
blewis12, line 34: nit: Should this just entirely replace the [existing action](https://github.com/grafana/alloy/blob/main/.github/workflows/release-lint-pr-title.yml) which can just run in this pre-flight check rather than a separate flow? Otherwise we have a duplicate check

blewis12: oh 🤦 duplicate comment to joe's

.github/workflows/pr-preflight.yml
blewis12: accidentally made a duplicate comment but I agree that we can just condense around a single pre-flight check, seems simpler as a user to jus check a single place

kgeckhart: Yeah so my plan if we merge this is to see how it behaves and if it works as we would like then make this one required and delete the other one. This way if something isn't quite right with this it's non-blocking.

jharvey10 (approved):
LGTM. Eager to see it in action on PRs!

PR #7102, opened 2026-09-14
docs(cloudwatch): Add iam:ListAccountAliases to the recommended IAM policy

diegobelotti:
 
  Add `iam:ListAccountAliases` to the recommended IAM policy for `prometheus.exporter.cloudwatch`, and document the effect of the permission being missing.

  ### Pull Request Details

  The component embeds yet-another-cloudwatch-exporter, which calls `iam:ListAccountAliases` to populate the account alias label on the `aws_account_info` metric. The permission is part of the quick-start IAM policy in the upstream YACE README, but it's
  absent from the policy recommended on this page.
 
  As a result, users who apply the documented policy get a `Couldn't get account alias` warning logged at every scrape, with an empty alias label, while metric collection itself works normally. There's no configuration argument on the component to
  disable the lookup, so granting the permission is the only fix.

  This adds the action to the policy, plus a short paragraph in the same style as the existing API Gateway and DMS permission notes, explaining what the permission is for and what happens without it.
  
  ### Notes to the Reviewer
  
  
  
  ### PR Checklist
 
  
 
  - [x] Documentation added
  - [ ] Tests updated


docs/sources/reference/components/prometheus/prometheus.exporter.cloudwatch.md
```diff
@@ -58,6 +58,12 @@ The following IAM permissions are required to discover tagged [Database Migratio
 "dms:DescribeReplicationTasks"
 ```
 
+The following IAM permission is required to retrieve the AWS account alias, which is exported as a label on the `aws_account_info` metric. If it's missing, {{< param "PRODUCT_NAME" >}} logs a `Couldn't get account alias` warning at each scrape and the alias label is empty. Metric collection isn't affected.
```
clayton-cornell: This mirrors upstream, but... is this scoped to just this metric (`aws_account_info`) or is this done on all metrics?

What label is applied? Is it `account_alias`?

diegobelotti: Hi, I've corrected the description: there is no aws_account_info metric. 
The alias is applied to all metrics, and the label name should be account_alias. I can't test it directly, because none of my accounts has an account alias set. 
I looked into this to avoid logs full of warnings. 
Label name was taken from source code.

clayton-cornell (approved):

